PR #494, opened 2016-12-03
PICARD-577: Albums with same name by different artists are clustered together

emhazen:
Initially, clustering would cluster all tracks with the same album title. 

There are two scenarios regarding tracks with the same album title but different artists
1. There are two separate albums with the same name by different artists
2. There is one album which two different artists worked on

I figured the best way to handle this is to allow the user to determine if they want the tracks clustered or not. I've added conflict detection to `picard/cluster.py`. It grabs the first artist that it finds and dubs this the "main artist" of the track. Then it raises an alert if a conflict is found; that is, if it finds a track within the album cluster with a conflicting artist title. A message box appears informing the user of the error and displaying the conflicting track. The user can view the cluster that it is conflicting with the details button.
messagebox
They may also select to perform the same action for all remaining conflicts.


scripts/picard
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python
+from picard.tagger import main; main('/usr/share/locale', True)
```
zas: What's that ?

picard/cluster.py
```diff
@@ -245,23 +245,83 @@ def cluster(files, threshold):
...
+
+        msg.setDetailedText(cluster_list)
+
+        do_all = QCheckBox()
+        do_all.setText(_(u"Do this for all conflicts"))
+
+        layout.addWidget(do_all, layout.rowCount()- 3, 1)
+
+        ret = msg.exec_()
+        return ret, do_all.isChecked();
```
zas: `ret` is useless here

picard/cluster.py
```diff
@@ -245,23 +245,83 @@ def cluster(files, threshold):
...
+        yes = msg.addButton("Yes", QMessageBox.YesRole)
+        msg.setDefaultButton(yes)
+
+        cluster_list = "Current cluster:"
+        for cluster_id in album:
+            if cluster_id < track_id:
+                artist = files[cluster_id].metadata["artist"]
+                title = files[cluster_id].metadata["title"]
+                track = files[cluster_id].metadata["tracknumber"]
+                cluster_list = cluster_list + '\n' + '{:4}{:12}{:10}'.format(track, title, artist)
```
zas: Instead of concatenating strings this way, use a list, append elements to it, and join them after the loop.

picard/cluster.py
```diff
@@ -245,23 +245,83 @@ def cluster(files, threshold):
...
+        title = _(u"Album Artist Conflict")
+        text = _(u"This track shares an album title with a cluster, "
+            "but does not share an artist name. Would you still like to "
+            "cluster this track?\n\n")
+
+        # Conflicting Track
+        artist_name = files[track_id].metadata["artist"]
+        album_name = files[track_id].metadata["album"]
+        song_title = files[track_id].metadata["title"]
+        text = text + album_name + '\n' + song_title + '\n' + artist_name
```
zas: Better add elements to a list, and join them with`\n` at the end

picard/cluster.py
```diff
@@ -245,23 +245,83 @@ def cluster(files, threshold):
...
+
+
```
zas: Extra blank line, please remove

picard/cluster.py
```diff
@@ -245,23 +245,83 @@ def cluster(files, threshold):
             artist_max = 0
             artist_id = None
             artist_hist = {}
+            main_artist = ""
+            do_all = False
+            do_cluster = True
+            to_remove = []
             for track_id in album:
+                artist = tracks[track_id][0]
+                if main_artist is "":
```
zas: Better initialize and compare to `None`

scripts/picard
emhazen: Looks like auto-generated code

picard/cluster.py
```diff
@@ -288,31 +288,35 @@ def cluster_warning(files, track_id, album):
...
 
-        msg = QMessageBox(QMessageBox.Question, title, text)
+        msg = QMessageBox(QMessageBox.Question, title, '\n'.join(text))
         layout = msg.layout()
         no = msg.addButton("No", QMessageBox.NoRole)
         yes = msg.addButton("Yes", QMessageBox.YesRole)
         msg.setDefaultButton(yes)
 
-        cluster_list = "Current cluster:"
+        cluster_list = ["Current cluster:"]
```
zas: This string should be marked for translation.

zas:
I don't understand why there are unrelated commits in this PR. Diff files is ok, but commit list is weird.

picard/cluster.py
emhazen: How would I go about doing that?

emhazen:
There was some stuff that was required for a university course; should have undone all of that though.

zas:
> How would I go about doing that?

Wrap string with `_()` (mark for translation + immediate translation function)

zas:
> There was some stuff that was required for a university course; should have undone all of that though.

Without a proper cleanup (removing unrelated stuff) this PR will not be merged.
Unrelated commits can be removed by rebasing interactively (`git rebase -i`)
At the end you should only have changes related to PICARD-577 touching only file `picard/cluster.py`

emhazen:
I'll get working on that. I'm still learning this stuff, thanks for your patience.

emhazen:
Alright, I think I got the rebase to work. Couldn't take care of that first commit though, any advice?

Sophist-UK:
I am only just reviewing  this PR because I have experienced the same issue, and I have several reservations:

1. I suspect that there are several different scenarios that the clustering code has to cope with:
a. Files with partial tagging - ideally album title and album artist but it could be only one of these
b. Files without any tags - have to rely on directory names for clustering
c. A mix of the above in the same directory

2. To be honest, I can't understand how the clustering code can cluster together files from two directories just because the directory names are the same e.g.
```
artist1 -> greatest hits
artist 2 -> greatest hits
```
And of course you might get
```
artist1-> greatest hits cd 1
       -> greatest hits cd 2
```
and
```
artist -> greatest hits normal edition
        -> greatest hits special edition
```
where both greatest hits directories have files tagged with the same album title and artist, but which to our intelligent eyes are clearly different situations. To my eye the only way to differentiate these is to NEVER cluster files with the same album title and artist together if they have different directories because it is easier to merge clusters by dragging one onto the top of the other than it is to split a cluster into two.

And you might also get
```
top dir -> artist1 greatest hits
        -> artist 2 greatest hits
```
where the files only have the album title and no album artist.

What is needed here is an algorithm that works as well as possible, but errs on the side of splitting an album into separate clusters (where the user can easily drag one to another to merge the clusters) rather than merging them.

My first cut algorithm would be:
a. Start by grouping 2 or more files in the same directory which have both album title and artist tags and group these. If you have two albums in the same directory, you might get multiple clusters from one directory. Cluster Title / Artist taken from tags.
b. Any files which have either the album title or artist but not both, if there is exactly one group from a. where the tag and directory both match, then add the file to that group.
c. Any files which have neither the album title or artist, if there is exactly one group from a. in the same directory then add to that group.
d. Any files without either album title or artist, then if there is exactly one group in that directory from a. add the files to it.

e. Any files in the same directory which have album title or artist but not both, group them together. Cluster title / artist taken from the tag that exists - if album artist, then title taken from last two directories in the path (which would typically be artist / album or album / CDx - if album title, look at last and second last directory names in the path and see if they equal or have the title at the start  end and determine artist from rest of directory or parent directory.
f. Any files without album title or artist, where directory matches exactly one group from e., add to that group.
g. Any directories where no files are grouped yet, group those files together - and take cluster title from last 2 directories and cluster artist from 2 directories further up.

Any files left over, remain Unmatched.

Releases split across directories (e.g. CD1/CD2) are never grouped together even if they have the same album title and artist in tags. If you want to assist the user with these, then you can add some clever functionality in Lookup to look for same cluster artist and similar cluster titles i.e. equal or first / last parts of title match and do a lookup as if the two clusters were merged.

This won't solve every situation (my directory naming structure is "album artist/album artist (year) album title") but I think it would match most of the default structures created by taggers.

Sophist-UK:
P.S. I haven't read the clustering code so I have no idea what the current clustering algorithm is.

There are also some technical issues with the code in this PR. I haven't looked at it closely, but I note that cluster.py now has QtGui imported - and if this ever runs on a worker thread, that is a no no.

Perhaps we can discuss the algorithm here and then decide whether it is worth a new PR to implement the algorithm and close this one.

samj1912:
It's been almost an year since this was updated. Closing this with a 'Needs more work' tag. You can reopen this if you want to continue working on it. Also we have updated to Qt5 and py3 so be mindful of that while reworking on this PR.

PR #496, opened 2016-12-12
PICARD-851: Correctly update parent items on save

samj1912:
When multiple files were linked to the same track and the changes to
the items were saved, the parent album icon didn't update to reflect
the change.

This was because of a buggy item.save() function which led to an update
call to the update() method of files. The files on update did not recu-
rsively call their parent items to be updated too.

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -756,7 +756,6 @@ class TrackItem(TreeItem):
             self.setBackground(i, bgcolor)
         if self.isSelected():
             TreeItem.window.update_selection()
-
         if update_album:
             self.parent().update(update_tracks=False)
 
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -775,8 +775,9 @@ class FileItem(TreeItem):
         if self.isSelected():
             TreeItem.window.update_selection()
 
-        if isinstance(self.parent(), TrackItem) and update_track:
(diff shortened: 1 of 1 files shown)


mwiencek:
Looks good, I think, though I haven't touched this code in years.

picard/ui/itemviews.py
```diff
@@ -728,25 +728,26 @@ def update(self, update_album=True):
...
-                for i in xrange(newnum - 1, oldnum - 1, -1):
-                    item = FileItem(track.linked_files[i], False)
-                    item.update()
-                    items.append(item)
-                self.addChildren(items)
+            if update_files:
+                oldnum = self.childCount()
+                newnum = track.num_linked_files
+                if oldnum > newnum:  # remove old items
+                    for i in xrange(oldnum - newnum):
```
zas: `xrange()` doesn't exist in python 3, perhaps better to use `range()`.

samj1912: I just shifted that bit from the existing code. Will make the changes :)

picard/ui/itemviews.py
```diff
@@ -755,13 +756,14 @@ def update(self, update_album=True):
             self.setBackground(i, bgcolor)
         if self.isSelected():
             TreeItem.window.update_selection()
+
```
mwiencek: Sorry to nitpick, but can you remove the unrelated whitespace change?

PR #497, opened 2016-12-13
PICARD-855: Adds errorhandler for missing meta-data

samj1912:
When a track with a non valid total-tracks has multiple release matches
on Picard, picard tries to apply a linear combination  of weights to
detect the best match. In the process it tried to typecast totaltracks
into and int.

This causes a ValueError when totaltracks is not castable into an int.
Added a try except clause to take care of the same.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -71,7 +71,7 @@ class Metadata(dict):
             if img.is_front_image():
                 return [img]
         return []
-        
+
     def remove_image(self, index):
         self.images.pop(index)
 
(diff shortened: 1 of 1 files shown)


samj1912:
The thing is, the above assumes if the totaltracks meta tag exists, it must be valid and then proceeds to act on it.
The following calculation of a and b depends entirely on whether total tracks exist or not.

According to my proposed changes if the above code encounters a value error then it skips the calculation of a and b and doesn't add totaltracks to weights when calculating the linear combination of weights to determine the item's appropriate match in case of multiple matches.

I think setting a value 0 to a might affect the linear combination algorithm in a way it is not supposed to when totaltracks tag is not present. Hence I did not assign it a default value and added the entire block under the try clause.

mwiencek:
I still agree with @y-van-z here: "I suggest to handle error for the first line only since it is the one that deals with untrusted data." You can do that with `try ... except ... else`:

```Python
try:
    a = int(self["totaltracks"])
except ValueError:
    pass
else:
    # this block runs only when there is no exception
```

Python has function scope, not block scope, so you can still access `a` underneath the `else`.

PR #499, opened 2016-12-14
Add support for nosetests

samj1912:
When the tests are run using nosetests, the Qt uses sip API v1 by
default which causes nosetests to fail when running version_from_string
function which returns a QVariant object instead of a string.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -107,10 +107,6 @@ class picard_test(Command):
 
     def run(self):
         import unittest
-        import sip
-
-        sip.setapi("QString", 2)
-        sip.setapi("QVariant", 2)
 
         names = []
         for filename in glob.glob("test/test_*.py"):
--- a/test/__init__.py
+++ b/test/__init__.py
@@ -1,5 +1,10 @@
 import glob
 import os.path
 
+import sip
+
+sip.setapi("QString", 2)
+sip.setapi("QVariant", 2)
+
 for filename in glob.glob(os.path.join(os.path.dirname(__file__), "test_*.py")):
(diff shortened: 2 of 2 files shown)


mineo:
To be honest, I'd prefer moving https://github.com/metabrainz/picard/blob/master/setup.py#L110-L113 from `setup.py` to `test/__init__.py`, which fixes the problem for nose2 (nose1 is no longer supported upstream), pytest, test execution via setup.py and `python2 -m unittest discover` instead of hiding a conversation in the config code.

samj1912:
Mineo, I tried doing that as the first thing, it didn't work out however, as it said the api levels had already been set to level 1.

samj1912:
The problem is if pyQt is import before importing sip, the api version defaults to 1. I'll have to go through the stack trace and see where it is being imported first.

mineo:
Are you using nose or nose2? Nose itself recommends using something else (https://nose.readthedocs.io/en/latest/#note-to-users) and changing `test/__init__.py` works for all recommended test runners AFAICT.

samj1912:
Ah no, I was using nose 1.3.7, just shifted to nose2 , it works out fine :D

samj1912:
Updated the files accordingly :)

We can specify nose2 specifically in the docs just in case.

mineo:
Great, thanks! Let's keep other test runners out of INSTALL.txt for now - if they're compatible with Pythons unittest library, that's fine and is probably the case for nearly all of them, but I don't feel like making sure they work.

PR #500, opened 2016-12-14
PICARD-857: Remove deprecated checks and add…

samj1912:
…support for GNOME 3

Gnome 3 recommends using gsettings to get and set values, as a result
gconftool-2 does not return a value in gnome3. Also GNOME_SESSION_ID
as a variable has been deprecated, so the detection of Gnome as a DE
has been changed accordingly.

### DIFF
--- a/picard/util/icontheme.py
+++ b/picard/util/icontheme.py
@@ -30,8 +30,10 @@ else:
         '/usr/share/pixmaps',
     ]
 
-if 'GNOME_DESKTOP_SESSION_ID' in os.environ:
-    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or None
+if 'XDG_CURRENT_DESKTOP' in os.environ and os.environ['XDG_CURRENT_DESKTOP'].lower() == 'gnome':
+    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or os.popen('gsettings get org.gnome.desktop.interface icon-theme').read().strip()[1:-1] or None
+elif 'XDG_CURRENT_DESKTOP' in os.environ and os.environ['XDG_CURRENT_DESKTOP'].lower() == 'unity':
+    _current_theme = os.popen('dconf read /desktop/gnome/interface/icon_theme').read().strip()[1:-1] or None
 elif os.environ.get('KDE_FULL_SESSION'):
     _current_theme = os.popen("kreadconfig --file kdeglobals --group Icons --key Theme --default crystalsvg").read().strip() or None
 else:
--- a/picard/util/icontheme.py
+++ b/picard/util/icontheme.py
@@ -32,13 +32,11 @@ else:
 
 _current_theme = None
 if 'XDG_CURRENT_DESKTOP' in os.environ:
-    if os.environ['XDG_CURRENT_DESKTOP'].lower() == 'gnome':
+    desktop = os.environ['XDG_CURRENT_DESKTOP'].lower()
+    if desktop in ('gnome', 'unity'):
         _current_theme = (os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip()
-                          or os.popen('gsettings get org.gnome.desktop.interface icon-theme').read().strip()[1:-1]
+                          or os.popen('dconf read /org/gnome/desktop/interface/icon-theme').read().strip()[1:-1]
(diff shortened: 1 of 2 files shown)


picard/util/icontheme.py
```diff
@@ -30,8 +30,8 @@
         '/usr/share/pixmaps',
     ]
 
-if 'GNOME_DESKTOP_SESSION_ID' in os.environ:
-    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or None
+if 'DESKTOP_SESSION' in os.environ and os.environ['DESKTOP_SESSION'].lower() == 'gnome':
+    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or os.popen('gsettings get org.gnome.desktop.interface icon-theme').read().strip() or None
```
mineo: The new command returns `'Adwaita'` on my machine - note the apostrophes. Do they need to be removed as well?

mineo:
I'm curious: according to https://bugzilla.gnome.org/show_bug.cgi?id=542880 `GNOME_DESKTOP_SESSION_ID` was no longer set in 2008 already and this is the first time somebody has reported a problem with it, so what changes with the new variable?

picard/util/icontheme.py
samj1912: Mineo, the thing is although the above has been patched, it has been deprecated. 

So it still exists as a legacy variable so that programmes that use it are not affected.
But I think we should switch to something more consistent.
 
Also the above variable exists even though I am using unity as a DE instead of gnome.

samj1912:
Updated the commit to remove the quotes.

mineo:
According to http://unix.stackexchange.com/a/116694/4699 and http://askubuntu.com/a/227669 `XDG_CURRENT_DESKTOP` is better suited to properly detect a running gnome desktop.

samj1912:
Yes, but according http://askubuntu.com/questions/72549/how-to-determine-which-window-manager-is-running/227669#227669 it return gnome when running mint under cinnamon.

samj1912:
Also, another thing that I have noticed is I am running Unity on Ubuntu 16.10 and it returns unity using both the variables, but I do use a theme.

Should I add unity too to the list?

It uses gsettings with the same path as gnome to set the icon theme.

samj1912:
Updated gsettings to dconf for a more low level access to icon themes in case gsettings is not installed.

picard/util/icontheme.py
```diff
@@ -30,8 +30,10 @@
         '/usr/share/pixmaps',
     ]
 
-if 'GNOME_DESKTOP_SESSION_ID' in os.environ:
-    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or None
+if 'XDG_CURRENT_DESKTOP' in os.environ and os.environ['XDG_CURRENT_DESKTOP'].lower() == 'gnome':
+    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or os.popen('gsettings get org.gnome.desktop.interface icon-theme').read().strip()[1:-1] or None
+elif 'XDG_CURRENT_DESKTOP' in os.environ and os.environ['XDG_CURRENT_DESKTOP'].lower() == 'unity':
+    _current_theme = os.popen('dconf read /desktop/gnome/interface/icon_theme').read().strip()[1:-1] or None
```
zas: I would rewrite it a bit differently:

```python
        _current_theme = None
        if 'XDG_CURRENT_DESKTOP' in os.environ:
            desktop = os.environ['XDG_CURRENT_DESKTOP'].lower()
            if desktop == 'gnome':
                _current_theme = ....
```

picard/util/icontheme.py
```diff
@@ -30,12 +30,19 @@
         '/usr/share/pixmaps',
     ]
 
-if 'GNOME_DESKTOP_SESSION_ID' in os.environ:
-    _current_theme = os.popen('gconftool-2 -g /desktop/gnome/interface/icon_theme').read().strip() or None
+_current_theme = None
+if 'XDG_CURRENT_DESKTOP' in os.environ:
```
zas: You can go a bit further and shorten lines, reducing code redundancy:

```python
desktop = os.environ['XDG_CURRENT_DESKTOP'].lower()

if desktop == 'gnome':
....
elif desktop == 'unity':
...
```

PR #502, opened 2016-12-16
PICARD-858: Remove apparent non-functionality of the tagger button

suhas2go:
While using the 'Lookup in the Browser' option, clicking the tagger button loads the album into Picard but since Picard doesn't come to surface, it looks like as if nothing is happening. This PR just adds a bring_tagger_front() function and makes a call it to it when one clicks on the tagger button.

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -67,6 +67,7 @@ class BrowserIntegration(QtNetwork.QTcpServer):
             action, args = line[1].split("?")
             args = [a.split("=", 1) for a in args.split("&")]
             args = dict((a, unicode(QtCore.QUrl.fromPercentEncoding(b))) for (a, b) in args)
+            self.tagger.bring_tagger_front()
             if action == "/openalbum":
                 self.tagger.load_album(args["id"])
             elif action == "/opennat":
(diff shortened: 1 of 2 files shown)


suhas2go:
Here's the link to the issue https://tickets.metabrainz.org/browse/PICARD-858

mineo:
Just a thought: there are pages (like release search results, `artist/<mbid>/releases` and maybe others where not stealing window focus might be the desired option.

suhas2go:
@mineo On Ubuntu, this nicely displays a notification saying Picard is ready. It doesn't just pop up and steal focus. Also, the notification appears only once, even on artist/<mbid>/releases , it doesn't notify everytime you click on the tagger button.

zas:
@mineo: do you think we can merge this PR ?

suhas2go:
Just to add..I've tested behavior on Ubuntu 16.10 Gnome and Windows 10. I've mentioned about Ubuntu above, on Windows the taskbar entry of Picard changes color and blinks for some time (can be a little annoying). This is mentioned in the [qt documentation](http://doc.qt.io/qt-4.8/qwidget.html#activateWindow) as well. 
Unfortunately, I cannot test behavior on Mac.

zas:
@mwiencek : can you test the behavior on Mac ?

mwiencek:
On macOS 10.12.2 it switches to Picard instantly (steals focus).

zas:
> On macOS 10.12.2 it switches to Picard instantly (steals focus).

@mineo , @mwiencek 
Is this an acceptable behavior ? Sstealing focus is not worse than apparent non-functionnality imho.

PR #385, opened 2014-12-07
Initial port of picard/* to Qt5

mwiencek:
I'd like Picard to have proper VoiceOver/accessibility support on Mac OS X (we've gotten several support@ emails from users who cannot use Picard at all). This might even be a prerequisite for submitting Picard to the app store.

This requires porting Picard to Qt5. The carbon version of Qt4 actually had some accessibility features enabled, but it was gutted from the cocoa version (only to be mentioned in a footnote somewhere).

This will break things as far as plugins/packaging are concerned, so I would recommend bumping the major version of Picard to 2.0.

I haven't ported any of the plugins, but can work on that if this is merged.

PR #503, opened 2016-12-21
PICARD-817, PICARD-960: Port Picard to Qt5(Continued)

samj1912:
Continues and finishes #385


mwiencek:
I think the commits should be named "Port Picard to Qt5" and "Port Picard to Qt5 (continued)", where the commit message for the second commit explains that it's a continuation of e41fb3d.

picard/config.py
```diff
@@ -65,13 +65,6 @@ def remove(self, key):
     def raw_value(self, key):
         """Return an option value without any type conversion."""
         value = self.__config.value("%s/%s" % (self.__name, key))
-
-        # XXX QPyNullVariant does not exist in all PyQt versions, and was
-        # removed entirely in PyQt5. See:
-        # http://pyqt.sourceforge.net/Docs/PyQt5/pyqt_qvariant.html
-        if str(type(value)) == "<class 'PyQt4.QtCore.QPyNullVariant'>":
```
mwiencek: If someone had such a value in their configuration and upgraded Picard, what would `str(type(value))` return? We should figure that out.

picard/ui/ui_infodialog.py
```diff
@@ -1,132 +1,73 @@
...
 # Use `python setup.py build_ui` to update it.
 
-from PyQt4 import QtCore, QtGui
-
-try:
-    _fromUtf8 = QtCore.QString.fromUtf8
-except AttributeError:
-    _fromUtf8 = lambda s: s
-
-class ArtworkTable(QtGui.QTableWidget):
```
mwiencek: Huh, this shouldn't be removed. Are you sure you rebased things against latest master?

mwiencek: Ah, the file was edited manually in 7451602f8400bb32206e423286475b56e4cffa27 (not good). So we'll have to move the custom code to somewhere else, without changing any behavior.

picard/config.py
samj1912: http://pyqt.sourceforge.net/Docs/PyQt5/pyqt_qvariant.html

According to this it will just convert it to a python object. So I suppose it will just return a str type?

samj1912:
Updated the Qt5 port. (All of picard till 1.4 is ported now) Although all of my commits got merged into @mwiencek 's original commit while rebasing.

mineo:
At the moment, the settings can't be opened:

```
  File "./picard/ui/mainwindow.py", line 746, in show_options
    dialog = OptionsDialog(page, self)
  File "./picard/ui/options/dialog.py", line 121, in __init__
    page.load()
  File "./picard/ui/options/scripting.py", line 402, in load
    self.ui.script_list.setItemSelected(last_selected_script, True)
AttributeError: 'QListWidget' object has no attribute 'setItemSelected'
```

picard/webservice.py
```diff
@@ -512,7 +512,7 @@ def browse_releases(self, handler, priority=True, important=True, **kwargs):
     def submit_ratings(self, ratings, handler):
         host = config.setting['server_host']
         port = config.setting['server_port']
-        path = '/ws/2/rating/'
+        path = '/ws/2/rating/?client=' + CLIENT_STRING
```
mineo: See 2113a13d60888bfced7c3bc7c960944c50b1be31, this is not correct.

mineo:
We should probably decide which minimum version of (Py)Qt5 we want to require. Most of the HighDPI things seem to require Qt 5.6 (or at a minimum 5.4), redirection (I think this came up somewhere recently, but don't remember where) support in QNetworkAccessManager is 5.6+ as well.

picard/webservice.py
samj1912: I did correct it. No idea why it didn't update O.o.

Freso:
> We should probably decide which minimum version of (Py)Qt5 we want to require.

Given the amount of things we want to change, Picard 2.0 probably won't be out for a while yet, so my vote is for not putting the bar on a too old version. 5.6, or even 5.7, would get my vote. (Qt 5.8 was released in January, so PyQt 5.8 will likely get released soon. While I would be happy to go for this version as well, that might be a tad too recent for even non-Debian systems. :))

zas:
> Given the amount of things we want to change, Picard 2.0 probably won't be out for a while yet, so my vote is for not putting the bar on a too old version. 5.6, or even 5.7, would get my vote. (Qt 5.8 was released in January, so PyQt 5.8 will likely get released soon. While I would be happy to go for this version as well, that might be a tad too recent for even non-Debian systems. :))

Qt5.4 is the first version with HiDPI support, but it was greatly reworked in 5.6. But 5.6 was released in March 2016, it is a LTS version. 5.4 supports for HiDPI was implemented differently, it was released back 2014. Ubuntu 16.04 comes with Qt 5.5.
So i think we should target the lowest version for now, perhaps adding tests for more recent versions (especially because HiDPI-related features are quite different among 5.x versions).
It also means we can't use 5.6+-only features for now.

Sophist-UK:
Sounds to me like we should go with 5.6.

samj1912:
Ditto with 5.6 given the above points in its favor.

Freso:
> Ubuntu 16.04

If people with old Ubuntu's want to run Picard, Picard 1.4.x should continue to work fine for them for some time to come. What does Ubuntu 16.10 ship? Or what will Ubuntu 17.04 and 17.10 have? I doubt Picard 2.0 will be out before October, so non-hardcore-LTS'ers should be able to have switched to a Ubuntu version with a more recent version.

Also, this is one of the reasons I really dislike non-rolling release distros: they hamper software development. Let's not let the "most outdated denominator" determine what library to use, esp. if it means more work 6–12+ months from now. If @samj1912 suddenly gets caught up with something (e.g., paid work, studies, …), then we can wait 2+ years for someone else to pick up the mantle and move us onwards from (Py)Qt 5.4. Let's try and not stay too far behind the curve, please?

(See also: The [GoPHP5 mission](https://web.archive.org/web/20110723121850/http://gophp5.org/).)

mineo:
Trying to use the OAuth authentication only gets me

```
  File "./picard/ui/options/general.py", line 91, in login
    authorization_code, scopes, self.on_authorization_finished)
  File "./picard/oauth.py", line 138, in exchange_authorization_code
    url.setEncodedQuery(url_query)
AttributeError: 'QUrl' object has no attribute 'setEncodedQuery'
zsh: abort (core dumped)  python2 tagger.py
```

samj1912:
@mineo fixed :)

Dockerfile
```diff
@@ -1,16 +1,18 @@
...
+
 RUN mkdir -p /build/
 ADD . /build/
 WORKDIR /build/
 ENV PYTHONPATH /usr/lib/python2.7/dist-packages
-RUN python setup.py build_ext -i										
-RUN python setup.py build_locales -i										
-CMD python setup.py test
+RUN python setup.py build_ext -i 
+RUN python setup.py build_locales -i
```
zas: Still extra whitespace at end of line (it was worse before though).

picard/ui/coverartbox.py
```diff
@@ -63,17 +63,22 @@ def mouseReleaseEvent(self, event):
             self.clicked.emit()
 
     def dragEnterEvent(self, event):
-        for url in event.mimeData().urls():
-            if url.scheme() in ('https', 'http', 'file'):
```
zas: Test about scheme is gone ?

picard/ui/infodialog.py
```diff
@@ -297,7 +297,7 @@ def _display_info_tab(self):
         if album.errors:
             tabWidget.setTabText(tab_index, _("&Errors"))
             text = '<br />'.join(map(lambda s: '<font color="darkred">%s</font>' %
-                                     '<br />'.join(unicode(QtCore.Qt.escape(s))
+                                     '<br />'.join(unicode(cgi.escape(s))
```
zas: I propose to create `picard.util.htmlescape()` so we can change it to whatever fits the best.

picard/ui/coverartbox.py
samj1912: yeah, it is checked again in dropEvent.

zas: Yep, just noticed. Thanks.

picard/ui/coverartbox.py
```diff
@@ -238,8 +243,8 @@ def open_release_page(self):
         lookup = self.tagger.get_file_lookup()
         lookup.albumLookup(self.release)
 
-    def fetch_remote_image(self, url):
-        return self.parent().fetch_remote_image(url)
+    def fetch_remote_image(self, url, dropped_data):
+        return self.parent().fetch_remote_image(url, dropped_data)
```
antlarr: Reviewing this I noticed a problem in master, so I created https://github.com/metabrainz/picard/pull/665 . Can you review that PR and make the corresponding rebase here? (note that the PR removes fetch_remote_image from CoverArtThumbnail)

picard/ui/ui_infodialog.py
```diff
@@ -67,6 +56,7 @@ def setupUi(self, InfoDialog):
         InfoDialog.setTabOrder(self.tabWidget, self.buttonBox)
 
     def retranslateUi(self, InfoDialog):
+        _translate = QtCore.QCoreApplication.translate
```
antlarr: is _translate used somewhere?

picard/ui/ui_options.py
```diff
@@ -3,37 +3,33 @@
...
         self.buttonbox.setOrientation(QtCore.Qt.Horizontal)
-        self.buttonbox.setObjectName(_fromUtf8("buttonbox"))
+        self.buttonbox.setObjectName("buttonbox")
         self.vboxlayout.addWidget(self.buttonbox)
 
         self.retranslateUi(Dialog)
         QtCore.QMetaObject.connectSlotsByName(Dialog)
 
     def retranslateUi(self, Dialog):
+        _translate = QtCore.QCoreApplication.translate
```
antlarr: Same here, maybe you wanted to set _ ?

picard/ui/ui_options_advanced.py
```diff
@@ -3,75 +3,62 @@
...
         self.vboxlayout.addWidget(self.groupBox_completeness)
-        spacerItem = QtGui.QSpacerItem(181, 21, QtGui.QSizePolicy.Minimum, QtGui.QSizePolicy.Expanding)
+        spacerItem = QtWidgets.QSpacerItem(181, 21, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
         self.vboxlayout.addItem(spacerItem)
 
         self.retranslateUi(AdvancedOptionsPage)
         QtCore.QMetaObject.connectSlotsByName(AdvancedOptionsPage)
 
     def retranslateUi(self, AdvancedOptionsPage):
+        _translate = QtCore.QCoreApplication.translate
```
antlarr: And also here and all other retranslateUi methods in other classes

picard/ui/ui_options_renaming.py
```diff
@@ -173,7 +160,7 @@ def retranslateUi(self, RenamingOptionsPage):
 "<html><head><meta name=\"qrichtext\" content=\"1\" /><style type=\"text/css\">\n"
```
antlarr: This line seems to have the only call to _translate in all picard's code (at least from a quick grep I did). Maybe it can be changed to _ and remove the _translate entry from _translate_re in setup.py ?

mwiencek:
You removed my authorship from the initial commit. :(

samj1912:
> You removed my authorship from the initial commit. :(
Sorry, it had too many merge conflicts, I had to checkout every file from the current qt5 branch.
Umm, can you rebase the master and fix it?

mwiencek:
It's okay, I don't think it's worth rewriting history. :)

PR #511, opened 2016-12-26
Dockerize the Travis build

samj1912:
Note: Related PR-510 to fix tests for docker

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,9 @@
-language: python
-python:
-  # Picard currently supports Python >=2.6 and <3
-  # However, since apt-get only installs PyQt4 for Py2.7, we only test against that :/
-  - "2.7"
-virtualenv:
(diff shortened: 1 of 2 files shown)


mineo:
What's the benefit of this change? We were previously running 4 test
configurations, each in about 50 seconds, now we're running a single one in
twice that time, covering only a subset of our dependencies (at least I
couldn't find anything about discid bindings in the Dockerfile) and with a
very old version of mutagen (1.25) that wouldn't have uncovered the recent
incompatibilities.

samj1912:
This change was to pave a way for the upcoming qt5/py3 changes.

samj1912:
Although yes, the previous travis build was more extensive of a test than this docker one. Maybe we can revert this change and switch to docker for qt5.

And if we do plan on shifting to py3 we can drop docker altogether.

PR #514, opened 2016-12-27
PICARD-861: Rename 'save_only_front_images_to_tags' option appropriately 

samj1912:


zas:
Please create a matching ticket in JIRA, as we want to document such change (we use JIRA to generate Changelogs)

samj1912:
Done :)

PR #515, opened 2016-12-28
PICARD-860: Add Coverart auto-fit options

samj1912:


samj1912:
Just an experimental branch. Needs more work.

samj1912:
@mwiencek it was decided after a discussion with @zas  and @Freso that we should keep pillow optional and let the user install it.

mwiencek:
It'll be optional for Linux users only (maybe, depending on how it's packaged), since we'll be forced to bundle it with Mac/Windows executables.

mwiencek:
Also, anyone on Linux who has a window manager plus Qt installed is going to have all or most of the PIL dependencies anyway.

picard/file.py
```diff
@@ -170,6 +171,25 @@ def save(self):
...
+        if config.setting["autofit_coverart"]:
+            try:
+                from PIL import Image, ImageOps
+                import io
+            except ImportError:
+                log.debug("Pillow not found. Please make sure Pillow(Python) is installed to allow\nresizing and cropping coverart images.")
+            else:
+                for image in metadata.images:
+                    resize_values = (config.setting['resize_width'], config.setting['resize_height'])
+                    image_info = imageinfo.identify(image.data)
```
zas: Image object has already those infos, see https://github.com/metabrainz/picard/blob/master/picard/coverart/image.py#L212

picard/file.py
```diff
@@ -170,6 +171,25 @@ def save(self):
...
+                    if image_info[0] > resize_values[0] or image_info[1] > resize_values[1]:
+                        buff = io.BytesIO(image.data)
```
zas: This intermediate `buff` variable is useless

picard/file.py
```diff
@@ -170,6 +171,25 @@ def save(self):
...
+                        im = Image.open(buff)
+                        im = ImageOps.fit(im, resize_values)
+                        image_type = image_info[2].split("/")[1]
```
zas: Type is already provided by image object, see https://github.com/metabrainz/picard/blob/master/picard/coverart/image.py#L212

picard/file.py
```diff
@@ -170,6 +171,25 @@ def save(self):
...
+            try:
+                from PIL import Image, ImageOps
+                import io
+            except ImportError:
+                log.debug("Pillow not found. Please make sure Pillow(Python) is installed to allow\nresizing and cropping coverart images.")
+            else:
+                for image in metadata.images:
+                    resize_values = (config.setting['resize_width'], config.setting['resize_height'])
+                    image_info = imageinfo.identify(image.data)
+                    if image_info[0] > resize_values[0] or image_info[1] > resize_values[1]:
```
samj1912: Tuple comparison will falsely evaluate (2,3) > (1,5), this returns True.

picard/file.py
samj1912: pillow uses JPEG as a type for .jpg objects, JPG returns invalid. But yes, I can eliminate the need for calling imageinfo again :)

picard/file.py
Sophist-UK: If resize_width < image.width and `resize_height > image.height what is the result of `(image.width, image.height) > (resize_width, resize_height)`?

From Python:
```
>>> (0,3) > (1,2)
False
```

samj1912: Python does tuple comparison index wise, if first indices are equal, it uses the next ones to evaluate. Otherwise returns on the basis of first index comparison.

zas:
I'm still wondering if it is the correct approach, it will work for basic user, let's imagine a more complex usage case:
- i want to downsize big images before storing them to tags (this PR)
- i want to unify image types (only jpeg) (=convert from PNG to JPEG in certain cases)
- i want to squarify images (= crop, eventually smart crop)

Adding options for all this = UI cluttering

Using ImageMagick API through http://docs.wand-py.org/en/latest/ together with a plugin hook (warning! threads) would allow much more possibilities and better performance

Sophist-UK:
Better IMO to keep the more esoteric stuff in plugins and add support for new types of plugins if needed.

So downsizing, converting  to specific types, squarifying etc. might best be done by a new coverart plugin type.

samj1912:
Having some problems with circular dependency import which is causing the build to fail.

samj1912:
Temporarily fixed it by import it under the try block. Any suggestions on how to fix this?

samj1912:
Test plugins : https://gist.github.com/samj1912/24a8c2d35c70285280a3921d3ef539ae

mwiencek:
Would it be useful to also pass the file metadata to the processor functions? Maybe in case someone only wants to process certain images.

samj1912:
Hmm ,interesting, I will have to look for different extension points then. I haven't checked if the current ones have a reference to metadata.

Also any thoughts on the circular dependency issue?

picard/coverart/image.py
```diff
@@ -266,6 +266,12 @@ def save(self, dirname, metadata, counters):
         else:
             filename = config.setting["cover_image_filename"]
             log.debug("Using default cover image filename %r", filename)
+        try:
+            from picard.coverart import run_coverart_file_save_action
+            run_coverart_file_save_action(self)
+        except Exception as e:
+            log.debug(e)
```
zas: I think it would be better to use `log.error()` and prefix the message with the hook name. Messages from `log.debug()` aren't shown by default, but here i think we need to let the user know something went wrong with (possibly) one of his plugins.

picard/metadata.py
```diff
@@ -60,7 +60,15 @@ def images_to_be_saved_to_tags(self):
...
             front_image = self.get_single_front_image(images)
             if front_image:
-                return front_image
+                images = front_image
+
+        # Run coverart tag embed actions
+        try:
+            from picard.coverart import run_coverart_tag_embed_action
+            for image in images:
+                run_coverart_tag_embed_action(image)
```
zas: Looking at https://gist.github.com/samj1912/24a8c2d35c70285280a3921d3ef539ae and this code, i think this isn't the right way to do it, there is no processing caching.
Of course in the case of resize, since dimensions are tested, the process is done once, but it wouldn't work with, for example, a black and white conversion.

We need a way to identify the process, and see if it was applied or not on one image.

zas:
Replaced by https://github.com/metabrainz/picard/pull/555

PR #516, opened 2016-12-28
PICARD-253: Show only supported tags in UI

samj1912:


picard/formats/apev2.py
```diff
@@ -204,6 +204,12 @@ def _get_tag_name(self, name):
         else:
             return name.title()
 
+    def supports_tag(self, name):
+        if self._get_tag_name(name):
```
zas: Not important, but i would shorten the function to:
```python
def supports_tag(self, name):
    return bool(self._get_tag_name(name))
```

zas:
Looks ok to me :+1:

samj1912:
Changed according to review :)

samj1912:
Changed according to review :)

mineo:
I might be wrong here, but isn't `_get_tag_name` the wrong method to use
here? It will always return a value that evaluates to True in a boolean
context if it's passed something that's not falsy - in this case, an empty
string/tag name.

samj1912:
In all of the above cases supports_tag has been modelled in such a way that it returns true only if the code in self._save saves that tag in the file.

mwiencek:
Well, I guess it "works" because apev2 and vorbis have free-form tags, so any tag will be saved. But you might as well change the implementation to `bool(name)` in that case...there's no point in calling `_get_tag_name` at present.

In actuality I'm guessing these formats have character or length limitations on the tag names (but I'm just guessing here). But I don't know if mutagen rejects those or just munges the names.

samj1912:
Improved as personal reviews.

PR #517, opened 2016-12-28
PICARD-706: Add option to ignore existing AcoustID fingerprints

samj1912:


zas:
Looks ok to me :+1:

mineo:
The name of the setting is a bit misleading, imho. It should be `
ignore_existing_acoustid_fingerprints` because those are what's
used/ignored.

samj1912:
Changed as per review :)

zas:
@mineo : if the PR is ok for you, please merge

mineo:
Sorry to only come back to this now, I did not have much internet access the last two weeks.

I can't help but feel that the setting is on the wring options page - the one it's tied to the most is "Automatically scan all new files" and the one that's nearly the same one ("Ignore MBIDs when loading new files") are both on the "General" page, whereas this one is hidden on the "Fingerprinting" one.

zas:
> I can't help but feel that the setting is on the wring options page - the one it's tied to the most is "Automatically scan all new files" and the one that's nearly the same one ("Ignore MBIDs when loading new files") are both on the "General" page, whereas this one is hidden on the "Fingerprinting" one.

True, though this option makes sense if fingerprints are enabled, which isn't the case the other options you cite.
For me, it is the correct page, but i agree it is a bit "hidden".

PR #519, opened 2016-12-29
PICARD-229: Implement case-insensitive move_additional_files

samj1912:


picard/file.py
```diff
@@ -353,16 +354,17 @@ def _move_additional_files(self, old_filename, new_filename):
...
+            pattern_regex = re.compile(fnmatch.translate(pattern), re.IGNORECASE)
+            for old_file in os.listdir(old_path):
+                if pattern_regex.match(old_file):
+                    new_file = os.path.join(new_path, old_file)
+                    old_file = os.path.join(old_path, old_file)
+                    # FIXME we shouldn't do this from a thread!
+                    if self.tagger.files.get(decode_filename(old_file)):
+                        log.debug("File loaded in the tagger, not moving %r", old_file)
+                        continue
+                    log.debug("Moving %r to %r", old_file, new_file)
```
zas: Unrelated, but `new_file = os.path.join(new_path, old_file)` line could be moved here, since `new_file` is only used here.

zas:
Looks ok to me :+1:

samj1912:
Made changes according to the review :)

mineo:
Note that fnmatch.translate and glob.glob1 do not treat wildcards the same
- * and ? match even filenames beginning with a dot when using
fnmatch.translate, but not when using glob.glob1.

>

samj1912:
@mineo yes, just went through the implementation of glob1

`    if pattern[0] != '.':
        names = filter(lambda x: x[0] != '.', names)
`

will add the appropriate checks :)

samj1912:
Added :)

zas:
@mineo : if this PR is now ok for you, please merge

picard/file.py
```diff
@@ -353,16 +354,23 @@ def _move_additional_files(self, old_filename, new_filename):
...
-                old_file = os.path.join(old_path, old_file)
-                # FIXME we shouldn't do this from a thread!
-                if self.tagger.files.get(decode_filename(old_file)):
-                    log.debug("File loaded in the tagger, not moving %r", old_file)
-                    continue
-                log.debug("Moving %r to %r", old_file, new_file)
-                shutil.move(old_file, new_file)
+            pattern_regex = re.compile(fnmatch.translate(pattern), re.IGNORECASE)
+            try:
+                names = os.listdir(old_path)
```
zas: `old_path` isn't modified in the loop, please move following lines outside the `for pattern...` loop.
The same goes for `names = filter(...` you may just do it once since `names` doesn't change.
Also in which case does it create an exception ? It shouldn't be hidden imho.

picard/file.py
```diff
@@ -352,17 +353,27 @@ def _move_additional_files(self, old_filename, new_filename):
         new_path = encode_filename(os.path.dirname(new_filename))
         patterns = encode_filename(config.setting["move_additional_files_pattern"])
         patterns = filter(bool, [p.strip() for p in patterns.split()])
+        try:
+            names = os.listdir(old_path)
+        except os.error:
+            log.debug("Error: {} directory not found".format(old_path))
```
zas: Use `log.error()` instead, as debug level messages are only shown when debug mode is on.

zas:
Since this PR changes slightly the usual behavior, i think we should add "(case insensitive)" to https://github.com/metabrainz/picard/blob/master/ui/options_renaming.ui#L75

OR make this an option, using a checkbox so the user can choose to match ignoring case or not, keeping the legacy behavior for smooth upgrade (case sensitive).

What do you think ?

samj1912:
I think the major use of the feature is moving associated files. So users should mainly be using it to move files with particular extensions. I think adding a simple "(case insensitive)" is enough.

zas:
> I think the major use of the feature is moving associated files. So users should mainly be using it to move files with particular extensions. I think adding a simple "(case insensitive)" is enough.

Agreed, if users complain about that, we'll still be able to add a checkbox and make a new PR ;)

samj1912:
Cool. I am done with this PR on my side then :)

picard/ui/ui_options_renaming.py
```diff
@@ -163,7 +163,7 @@ def retranslateUi(self, RenamingOptionsPage):
         self.move_files.setTitle(_("Move files when saving"))
         self.label.setText(_("Destination directory:"))
         self.move_files_to_browse.setText(_("Browse..."))
-        self.move_additional_files.setText(_("Move additional files:"))
+        self.move_additional_files.setText(_("Move additional files(case insensitive):"))
```
zas: Please add a space before the first parenthesis: `"Move additional files (case insensitive):"`

zas:
> Cool. I am done with this PR on my side then :)

Almost ;) See https://github.com/metabrainz/picard/pull/519/files#r94570323

picard/file.py
```diff
@@ -352,17 +353,27 @@ def _move_additional_files(self, old_filename, new_filename):
...
-                    continue
-                log.debug("Moving %r to %r", old_file, new_file)
-                shutil.move(old_file, new_file)
+            pattern_regex = re.compile(fnmatch.translate(pattern), re.IGNORECASE)
+            file_names = names
+            if pattern[0] != '.':
+                file_names = filtered_names
+            for old_file in file_names:
+                if pattern_regex.match(old_file):
+                    old_file = os.path.join(old_path, old_file)
```
zas: `old_file` is overwritten here

PR #520, opened 2016-12-29
PICARD-116: Add restore default button to options

samj1912:


samj1912:
I tried to make a generic way to load defaults without changing config.setting values but there was a lot of problem with plugin options since their ui elements and config.setting options are not consistently named.

This seemed the other most obvious way.

zas:
The difference between both buttons isn't very obvious.
I think we need a confirmation box for those actions, explaining exactly what will happen when confirmed.

samj1912:
Done :)

picard/ui/options/dialog.py
```diff
@@ -79,11 +79,19 @@ def __init__(self, default_page=None, parent=None):
         self.ui = Ui_Dialog()
         self.ui.setupUi(self)
 
+        self.ui.reset_all_button = QtGui.QPushButton(_("&Restore all Defaults"))
+        self.ui.reset_button = QtGui.QPushButton(_("Restore Defaults"))
```
zas: Accelerator (`&`) is missing here.

samj1912:
Hadn't added it to Reset Defaults, because of the conflicting Rs :P, added it to D instead.

PR #521, opened 2016-12-29
PICARD-834: Properly handle ValueErrors in case of invalid length in script

samj1912:


mwiencek:
Since we already have precedent of catching ValueErrors in the math functions, LGTM.

PR #522, opened 2016-12-30
PICARD-748: Fix duplication of tag fields in mp3 files

samj1912:


samj1912:
To replicate the bug, simply add a field for performer and add any custom value. Picard used to append it each time you saved.

mwiencek:
I wonder if we shouldn't be using `delall` with clear_existing_tags disabled, and instead just check that we're not adding duplicate values. But I see this is already done for `TCMP`.

samj1912:
The problem lies in the fact that metadata contains all the performers. The new ones as well as the existing ones. The tags.add(tmcl) line simply adds another frame with duplicate data.

The easiest way is to remove the existing frame and add a new one. 

The other way (checking each tmcl if the given performer already exists will cause an overhead)

zas:
> To replicate the bug, simply add a field for performer and add any custom value. Picard used to append it each time you saved.

Can you add a test for that ?

samj1912:
As discussed in IRC the above bug is only related to our implementation of saving tags while id3v2.3 is used. It is not apparent in v2.4 which most of our tests are based on. I am looking closely at our code to identify the issue because as per @lazka the tags.add() method replaces instead of appending, but that doesn't happen with our implementation. This corresponds to a bigger problem therefore.

samj1912:
The main bug lies in the fact that we have added the "if tipl.people" and "if tmcl.people" statements.

Here is how to replicate the bug. 

Enable id3v23 for saving.
Take a new mp3 file without performer and 
`
        'engineer': 'engineer',
        'arranger': 'arranger',
        'producer': 'producer',
        'DJ-mix': 'djmixer',
        'mix': 'mixer',
`
tags.

Add a new performer tag and click on save multiple times. Remove the file and load it again. You will see the performer field has been duplicated and gets duplicated each time you save.

Now add any of the above mentioned tags ( producer for eg) and save.
The duplicates dont occur anymore.

samj1912:
Added tests to replicate the bug behaviour :)

samj1912:
This PR is complete from my side :)

PR #524, opened 2016-12-30
PICARD-865: Add limit to length in $num

samj1912:
An arbitrarily large value if entered in GUI would cause the system to become unresponsive.


mwiencek:
Looks reasonable to me. 👍 Perhaps it should be documented on the website, but I don't know who would want to exceed that limit.

samj1912:
Rebased against upstream, ready to merge :)

PR #525, opened 2016-12-31
PICARD-864: Add support for autodetection of supported file formats

samj1912:


picard/formats/__init__.py
```diff
@@ -38,6 +38,23 @@ def supported_formats():
...
+    """Select the best matching file type amongst supported formats."""
+    fileobj = file(filename, "rb")
+    results = []
+    try:
+        header = fileobj.read(128)
+        options = [option for option in options if not option in (OggAudioFile, OggVideoFile) and option._File is not None]
+        results = [
+            (option._File.score(filename, fileobj, header), option.__name__, option)
+            for option in options]
+    finally:
```
zas: Is there a reason to use `try/finally` here instead of `with open(filename, "rb") as fileobj` ?

samj1912: The function is taken directly from vorbis file's detect ogg audio file. I didn't change it.

picard/formats/__init__.py
```diff
@@ -38,6 +38,23 @@ def supported_formats():
...
+def _guess_format(filename, options):
+    """Select the best matching file type amongst supported formats."""
+    fileobj = file(filename, "rb")
+    results = []
+    try:
+        header = fileobj.read(128)
+        options = [option for option in options if not option in (OggAudioFile, OggVideoFile) and option._File is not None]
+        results = [
+            (option._File.score(filename, fileobj, header), option.__name__, option)
+            for option in options]
```
zas: I have the feeling code can be improved here, we have somehow 2 iterations within `options`

picard/formats/__init__.py
```diff
@@ -38,6 +38,23 @@ def supported_formats():
...
+    finally:
+        fileobj.close()
+    results.sort()
```
zas: I would only call `sort()` on non-empty `results`. Function calls have a cost.

picard/formats/__init__.py
```diff
@@ -38,6 +38,23 @@ def supported_formats():
...
 
+def _guess_format(filename, options):
+    """Select the best matching file type amongst supported formats."""
+    fileobj = file(filename, "rb")
+    results = []
+    try:
+        header = fileobj.read(128)
+        options = [option for option in options if not option in (OggAudioFile, OggVideoFile) and option._File is not None]
+        results = [
+            (option._File.score(filename, fileobj, header), option.__name__, option)
```
zas: I think those lines deserver a comment explaining what they are doing

mineo:
This looks like a breaking change in the plugins API to me, because format
plugins now need to prevent a `_File` attribute (imho the leading
underscore should not be there if it's required).

mineo:
There's now a comment here, but it's still not immediately obvious why
Ogg{Audio,Video}File are hardcoded as exceptional cases here.

Sophist-UK:
@mineo said: "This looks like a breaking change in the plugins API to me, because format
plugins now need to prevent a `_File` attribute"

It looks like that to me, though it has been added to all the standard format plugins in Picard, and AFAIK there are no format plugins in Picard_Plugins - so it will only break a format plugin if someone has written one for themselves, and I am not aware of any such.

However perhaps we could handle this unlikely situation by providing some sort of error message if _File does not exist saying that _File needs to be added to the plugin.

Sophist-UK:
@mineo said: "There's now a comment here, but it's still not immediately obvious why
Ogg{Audio,Video}File are hardcoded as exceptional cases here."

I was wondering that too.

samj1912:
I kept it as _File as the existing formats kept it that way. We can remove the leading underscore if needed. 

Also we can create a Format class if needed which and all existing formats can inherit from it.

samj1912:
Updated the filter to make it more generic. Now if a format class doesn't have the attribtue _File it wont break the code, just that format wont be considered when autodetecting a file's format, however it will still open it using that format if it matches the extension.

picard/formats/__init__.py
```diff
@@ -20,7 +20,6 @@
 import sys
 from mutagen import _util
 from picard.plugin import ExtensionPoint
-
```
zas: This blank line should stay

picard/formats/__init__.py
```diff
@@ -39,6 +38,27 @@ def supported_formats():
     return formats
 
 
+def guess_format(filename, options):
+    """Select the best matching file type amongst supported formats."""
+    results = []
+    with file(filename, "rb") as fileobj:
+        header = fileobj.read(128)
+        # Calls the score method of a particulat format's associated filetype
```
zas: s/particulat/particular/

picard/formats/__init__.py
```diff
@@ -39,6 +38,27 @@ def supported_formats():
...
+                   for option in options
+                   if hasattr(option, "_File")
+                   and option._File is not None]
+    if results:
+        results.sort()
+    # No postive score i.e. the fileobj's header did not match any supported format
+    if not results or results[-1][0] <= 0:
+        raise Exception("Unknown audio format")
+    # return the format with the highest matching score
+    return results[-1][2]
```
zas: I may miss something, but can it be rewritten like this ? 

```python
if results:
   results.sort()
   # Positive score i.e. the fileobj's header did match a supported format
   if results[-1][0] > 0:
       # return the format with the highest matching score
       return results[-1][2]

raise Exception("Unknown audio format")
```

samj1912:
Updated as per reviews :)

samj1912:
And rebased the last one :P

zas:
@mineo , @Sophist-UK : anything to add before i proceed to merge ?

Sophist-UK:
Not tested this, but I have no further comments from code read through.

picard/formats/__init__.py
```diff
@@ -49,6 +71,8 @@ def open(filename):
         format = _extensions[ext]
     except KeyError:
         return None
```
zas: @samj1912 : according to this code, if extension isn't found, KeyError is raised and `open()` returns None
So there's no attempt of using the file magic number detection.
If another exception is raised, then we use `guess_format()`, but if it fails we raise an exception ("Unknown audio format").
I may miss something, but i don't hink it is the behavior we expect, and return/exception should be consistent whichever method we use.

Since using extension is prolly faster, we should fallback on magic number method if the extension isn't known, but if the format cannot be identified using this method, still return None.

samj1912: I am facing a bit of a problem. Apparently the else block is executing every time anything other than a KeyError is encountered.

Now if I place the guess_format function under the except clause it causes an error later in the thread when the file's _load method is called and I can't do anything about the format then.

How should I proceed?

zas:
> Apparently the else block is executing every time anything other than a KeyError is encountered.

Yes, this is what i'm talking about.

I think we should try extensions first, then fallback on magic number, but we need consistency.

KeyError should be the only exception possible in this code, so code should be something like:

```python
     try:
         format = _extensions[ext]
     except KeyError:
       try:
         format = guess_format(filename, _formats)
       except FileFormatUnknown:
          return None
     return format(filename)
 ```

FileFormatUnknown exception class needs to be defined.

File: picard/formats/__init__.py [hunk 9]
```diff
@@ -39,17 +39,44 @@ def supported_formats():
...
+            # return the format with the highest matching score
+            return results[-1][2](filename)
+
+    # No positive score i.e. the fileobj's header did not match any supported format
+    return None
+
+
 def open(filename):
     """Open the specified file and return a File instance with the appropriate format handler, or None."""
     i = filename.rfind(".")
```
zas: Those lines are detecting extension, if none is found at all it will return `None`, but imho we should still try magic number in this case.
So move those lines after `if not audio_file`

samj1912:
Done. Also updated the new test with no extension.

picard/formats/__init__.py
```diff
@@ -39,17 +39,44 @@ def supported_formats():
...
+        audio_file = guess_format(filename)
+        if not audio_file:
+            # Switch to extension based opening
+            audio_file = _extensions[ext](filename)
     except KeyError:
-        return None
-    return format(filename)
+        # None is returned if both the methods fail
+        pass
+    return audio_file
```
zas: I'm thinking about something like:
```python
try:
    audio_file = guess_format(filename)
    if not audio_file:
        # Switch to extension based opening
        i = filename.rfind(".")
        if i < 0:
           return None
       ext = filename[i+1:].lower()
       audio_file = _extensions[ext](filename)
    return audio_file
except KeyError:
   return None
except Exception:
  # log the exception message but still return None
  ....
  return None
```

picard/formats/__init__.py
```diff
@@ -39,17 +39,47 @@ def supported_formats():
...
+                return None
+            ext = filename[i+1:].lower()
+            # Switch to extension based opening
+            audio_file = _extensions[ext](filename)
+        return audio_file
     except KeyError:
+        # None is returned if both the methods fail
+        return None
+    except Exception as error:
+        log.debug("Error occured:\n{}".format(error.message))
```
zas: Use `log.error()`, since this case is unlikely to happen better not limit this to debug mode.

picard/formats/__init__.py
```diff
@@ -20,7 +20,7 @@
 import sys
 from mutagen import _util
 from picard.plugin import ExtensionPoint
-
+from picard import log
 _formats = ExtensionPoint()
```
zas: Add blank lines here

samj1912:
Squashed those last few changes.

test/test_formats.py
```diff
@@ -158,6 +157,16 @@ def test_ratings(self):
...
 
+    def test_guess_format(self):
+        if self.original:
+            fd, temp_file = mkstemp()
+            os.close(fd)
+            shutil.copy(self.original, temp_file)
+            audio = picard.formats.guess_format(temp_file)
+            audio_original = picard.formats.open(self.filename)
+            self.assertEqual(type(audio), type(audio_original))
+            os.unlink(temp_file)
```
zas: There's a risk of remaining temp files in case of failures, registring a cleanup method is much safer, see https://docs.python.org/2/library/unittest.html#unittest.TestCase.addCleanup
It can be done in another PR though.

picard/formats/__init__.py
```diff
@@ -39,17 +40,47 @@ def supported_formats():
...
     """Open the specified file and return a File instance with the appropriate format handler, or None."""
-    i = filename.rfind(".")
-    if i < 0:
-        return None
-    ext = filename[i+1:].lower()
     try:
-        format = _extensions[ext]
+        # First try to guess the format on the basis of file headers
+        audio_file = guess_format(filename)
+        if not audio_file:
```
mineo: This should have been `if audio_file is not None`, as `guess_format` returns `None` in case of failure.

test/test_formats.py
```diff
@@ -158,6 +157,16 @@ def test_ratings(self):
...
 
+    def test_guess_format(self):
+        if self.original:
+            fd, temp_file = mkstemp()
+            os.close(fd)
+            shutil.copy(self.original, temp_file)
+            audio = picard.formats.guess_format(temp_file)
+            audio_original = picard.formats.open(self.filename)
+            self.assertEqual(type(audio), type(audio_original))
+            self.addCleanup(os.unlink,temp_file)
```
mineo: This is just the same as calling `os.unlink(temp_file)` directly because all the interesting things happen before the call to `addCleanup` - imagine `guess_format` or `open` raising an error, the execution will never reach `addCleanup`. `addCleanup` is also only available in Python 2.7.

zas:
> This is just the same as calling os.unlink(temp_file) directly because all the interesting things happen before the call to addCleanup - imagine guess_format or open raising an error, the execution will never reach addCleanup. addCleanup is also only available in Python 2.7.

Oh, i missed the fact it is called too late, and yes it is python2.7.
Question: can't we drop 2.6 support for the next 1.4 release, 2.7 is out there since ages now.

zas:
> I actually agree with it - in nearly all cases, matching by file extension is probably good enough.

Yes, but we discussed about it on IRC and conclude the overhead is low enough to not rely on extensions by default, and do the reverse (that is falling back on extensions if magic number fails, as last chance to identify the format...).

Old code was relying on correct extension (that is, filename with a known extension), the new code will allow to provide a totally unknown file to Picard (`somefile`) and the user will not have to bother with file identification (which may a complex task for newbie). For now, file renaming will not change extension to the correct one without help, but at least, the file will be handled and tagged.

Overhead is a non-issue on modern systems (and even if it was, since we'll read the file at some point, it is very likely first blocks will be already cached, so the time spent here is earnt later).

zas:
> This should have been if audio_file is not None, as guess_format returns None in case of failure.

+1

picard/formats/__init__.py
```diff
@@ -39,17 +40,47 @@ def supported_formats():
...
+    results = []
+    with file(filename, "rb") as fileobj:
+        header = fileobj.read(128)
+        # Calls the score method of a particular format's associated filetype
+        # and assigns a positive score depending on how closely the fileobj's header matches
+        # the header for a particular file format.
+        results = [(option._File.score(filename, fileobj, header), option.__name__, option)
+                   for option in options
+                   if hasattr(option, "_File")
+                   and option._File is not None]
```
zas: > These two lines are the same as getattr(option, "_File", None).

+1

zas:
> I'm pretty sure that setting a local variable called _File in this function does not give the function an attribute called _File. This only works because the code that uses _File falls back to the same value assigned to _File here. For the same reason, EXTENSIONS and NAME are not set in this function, but outside of it.

Ooops i missed that, @mineo is totally right here.

picard/formats/vorbis.py
```diff
@@ -333,28 +333,11 @@ def _info(self, metadata, file):
...
-        raise mutagen.ogg.error("unknown Ogg audio format")
-    return results[-1][2](filename)
-
-
 def OggAudioFile(filename):
     """Generic Ogg audio file."""
     options = [OggFLACFile, OggSpeexFile, OggVorbisFile]
-    return _select_ogg_type(filename, options)
-
+    _File = None
```
samj1912: Oops, I added it(the variable _File) before and forgot to remove it, it is not included while testing options in guess_format anyways.

I'll remove it.

picard/formats/__init__.py
samj1912: Nope, it is supposed to be if audio_file is None as we are trying to open that file using its extension if _score method fails to determine the file type as it doesn't include all the corner cases.

mineo:
> the new code will allow to provide a totally unknown file to Picard (somefile) and the user will not have to bother with file identification (which may a complex task for newbie).

This is not clear from the UI. The file browser doesn't *show* files with unsupported extensions, but does load them if a whole directory is loaded. The "Add Files" dialog simply doesn't allow loading files with unsupported extensions at all AFAICT. "Add Folder" loads them from a directory and dropping a file with an unsupported extension onto the main window works as well, but I think only the last options is one that someone would actually use.

samj1912:
@mineo I brought this point up in discussion with @mwiencek  in IRC. The above does help when someone adds a folder(much more likely to have mismatched extensions which are difficult for the user to identify unlike a single file) or drags and drops a file.

samj1912:
We can remove the filters from the file dialog if that helps?

zas:
> will allow

@mineo ^^

picard/formats/__init__.py
```diff
@@ -39,17 +40,47 @@ def supported_formats():
...
-    i = filename.rfind(".")
-    if i < 0:
-        return None
-    ext = filename[i+1:].lower()
     try:
-        format = _extensions[ext]
+        # First try to guess the format on the basis of file headers
+        audio_file = guess_format(filename)
+        if not audio_file:
+            i = filename.rfind(".")
```
Sophist-UK: Just reviewed this code (as part of attempting to find the cause of another issue).

I think this code will fail on a file like '/my.dir/file-no-ext'.

I would suggest that this is changed to use os.path.splitext.

Sophist-UK: P.S. Will also fail on `/dir/.hidden-file/ because .hidden-file does not have an ext.

samj1912: I dont think it will? since we are attempting guess format before checking for extension.

And since only supported extensions are matched, the above file, if not identified through guess_format can't be identified by checking its extension(since it doesn't have one)

Sophist-UK:
I am also a little bit uncomfortable about this PR opening the file multiple times, because of any performance impact it might have on slow access remote files

Would it be better to open the file, guess the format move the pointer back to the beginning, run mutagen using the existing file handle and then close the file?

samj1912:
It is not opening it multiple times? Just once to read the header which is just 128 bytes and then closing it.
And then loading it using the appropriate container using mutagen.
How do you propose to run mutagen use the existing file object? AFAIK mutagen doesn't provide a method for tag editing with already open file objects.
Also, we decided the overhead is worth it in the IRC discussion regarding this PR.

picard/formats/__init__.py
samj1912: Yes, just tested, doesn't fail on the cases you specified.

Sophist-UK: OK

Sophist-UK:
We open it once. Then mutagen opens it again. So that's twice.

But if the performance impact was considered that's good enough for me.

picard/formats/__init__.py
```diff
@@ -39,17 +40,47 @@ def supported_formats():
     return formats
 
 
+def guess_format(filename, options=_formats):
+    """Select the best matching file type amongst supported formats."""
+    results = []
+    with file(filename, "rb") as fileobj:
```
Sophist-UK: From https://docs.python.org/2/library/functions.html#file:

> **file(name[, mode[, buffering]])**
>Constructor function for the file type, described further in section File Objects. The constructor’s arguments are the same as those of the open() built-in function described below.
>
>When opening a file, it’s preferable to use open() instead of invoking this constructor directly. file is more suited to type testing (for example, writing isinstance(f, file)).

Of course there are other open methods used in this code other than the native Python file open() - so using open instead of file might reduce readability rather than improve it.

picard/formats/__init__.py
Sophist-UK: Use of os.path.splitext is IMO still preferred as the code is more readable and shorter.

However this is a minor point.

picard/formats/__init__.py
samj1912: We already have an open() method defined in the file. Which is why I opted to use file()

zas: > We already have an open() method defined in the file. Which is why I opted to use file()

Yes, and it is perhaps not the best thing. This module level `open()` isn't great at all vs the builtin `open()`.

About `file()`, it is removed from Python 3.0, with comment "Use open" (https://docs.python.org/release/3.0/whatsnew/3.0.html#builtins)

picard/formats/__init__.py
zas: > Yes, just tested, doesn't fail on any of the cases you specified.

If not the case already, add actual tests in test/ ;)
`os.path.splitext()` vs  `rfind()` ... well, i guess both will work quite well, i don't know any system using anything but dot to separate extension. Though i would tend to use `os.path.splitext()` if it simplifies the code.

Note that Python 3.4 comes with a new lib for manipulating paths : https://docs.python.org/3.4/library/pathlib.html#pathlib.PurePath.suffix (not sure on what will be min Python version requirement for Picard 2.0)

PR #526, opened 2017-01-01
PICARD-872: Optionnaly use language from release for tracks without work language

yaymeh:
(ui / turned off by default)

https://tickets.metabrainz.org/browse/PICARD-872


mineo:
How does this work for tracks with a language set on the work, but when
track/recording relationships are disabled?

yaymeh:
If the "use track relationship" option is inactive, tracks don't get a language from the work (as far as I can tell); so in that case, with the "language from release" option activated, all tracks will get the release language.

Was that what you meant? Should that be mentioned explicitly? I couldn't think of a shorter option name that describes it better. Maybe "Use release language as default for tracks"?

zas:
You need to link this PR to a matching Picard issue at https://tickets.metabrainz.org

mineo:
In the model of MB, "Use the release language as default for tracks" makes sense to me.

zas:
> In the model of MB, "Use the release language as default for tracks" makes sense to me.

You mean as text at https://github.com/metabrainz/picard/pull/526/files#diff-493c462bfbbfdff6021c7fe1c9cbaa40R108 ?
I agree.

yaymeh:
Just noticed that the folksonomy option text says: "Fall back", so I used "fallback" - it's probably most accurate?

Added jira ticket: 
https://tickets.metabrainz.org/browse/PICARD-872

zas:
> Just noticed that the folksonomy option text says: "Fall back", so I used "fallback" - it's probably most accurate?

To fall back is a verb, fallback is a noun. So it depends on context, see discussion at http://english.stackexchange.com/questions/90549/difference-between-fallback-and-fall-back

yaymeh:
Sorry, I don't understand what you're trying to tell me... if there's still something wrong with the UI text, I can't find it. 

Or do I have to rebase / squash the commits or something?

zas:
> Sorry, I don't understand what you're trying to tell me... if there's still something wrong with the UI text, I can't find it.
> 

@mineo ^^ ?

> Or do I have to rebase / squash the commits or something?

Nope, that's ok for me, but this PR will have to wait after 1.4.0 release, strings are frozen until release.

yaymeh:
With that:
> > Sorry, I don't understand what you're trying to tell me... if there's still something wrong with the UI text, I can't find it.

...I meant the "needs more work" tag. 


If the link to:
> http://english.stackexchange.com/questions/90549/difference-between-fallback-and-fall-back

...was a clue, I didn't get it.

What exactly needs more work?

Freso:
Note that I don't think this should be merged into Picard's core: https://tickets.metabrainz.org/browse/PICARD-872?focusedCommentId=41975&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-41975

And since @samj1912 at least wants to move Picard 2.0 in a more modular direction, it seems like going down the wrong path to introduce something like this to the 1.4.x branch as a core feature rather than as a plugin.

samj1912:
I agree that this needs to be a plugin rather than an inclusion to the core picard code. Please feel free to rework this PR as a plugin and submit to the [Plugin Repo](https://github.com/metabrainz/picard-plugins). As such I am closing this PR for now.

Sophist-UK:
Actually, it can easily be done in a tagging script.

phw:
> Actually, it can easily be done in a tagging script.

This.

```
$if($not(%language%),$set(language,%_releaselanguage%))
```

PR #528, opened 2017-01-06
PICARD-868,307: Fix TMOO,TSST frame not being saved

samj1912:


samj1912:
TMOO frame was introduced in id3v2.4 and is not saved when id3v2.3 mode is enabled.

We have 2 options, either add TMOO frame to our compatid3 file or prevent it from being displayed in picard if v2.3 is selected. Or we can skip this and add TMOO as a TXXX:mood frame if saving to 2.3 and upgrade  it to TMOO when loading.

mwiencek:
Do any applications read/write TMOO with v2.3? Searching Google for "TXXX:mood" gives some results.

samj1912:
@mwiencek not sure, I had a talk with @lazka about this and his opinion was that picard decide how it wants to handle TMOO frames. Mutagen by default deletes all TMOO data( which I don't think is correct, as people might lose some existing metadata on their files if they switch)

It is misleading though, that we include mood tags in the ui and they can't be saved if we are using id3v2.3.

So we can:
1. Hide the mood tag when id3v2.3 is enabled.
2. Convert TMOO to TXXX:mood and upgrade it to TMOO again when loading.
3. Add TMOO frames as I have above.

What should we do?

mwiencek:
I would do whatever is interoperable with other applications, otherwise the change won't be useful. Google says that foobar, mp3tag, jaikoz, and musicbee all write TXXX:mood.

samj1912:
So I will just downgrade TMOO to TXXX:mood when saving with id3v2.3 and upgrade it to TMOO while processing :)

Sophist-UK:
Leaving aside the issue of TMOO, I believe that we should convert any ID3 tags we don't know about to metadata of the form ~id3:tag:ident:value and then convert it back again when we save.

samj1912:
@Sophist-UK TMOO is included in __translate so its dealt with that way, apart from that the existing code already does what you suggested.

Sophist-UK:
I am not sure that this is the case - its a while since I looked at the code. So for example, if there is a lyrics tag, how is that handled currently?

samj1912:
@Sophist-UK in form of USLT frames.

we handle it as 

```python
            elif name.startswith('~id3:'):
                name = name[5:]
                if name.startswith('TXXX:'):
                    tags.add(id3.TXXX(encoding=encoding, desc=name[5:], text=values))
                else:
                    frameclass = getattr(id3, name[:4], None)
                    if frameclass:
                        tags.add(frameclass(encoding=encoding, text=values))
```
while saving.

samj1912:
Similar issue also plagues TSST(in all id3 formats) and TSOP(only in AIFF format) frames.

samj1912:
Nevermind, just figured, the compatid3 we have is only applicable for ID3 files, AIFF is not affected by it.

I think we should update the supported tags then, since our AIFF implementation doesn't support any of the added fromes in compatid3

samj1912:
We haven't ever supported the compatid3 tags for AIFF. It was discovered right now because I added a couple of tests to check id3v23 compatibility. And adding support for them in AIFF seems to be out of the scope for this PR. I have created a new JIRA ticket for the same.

I think solving AIFF tests is part of a bigger issue with AIFF/id3 compatibility and I will try to address it in a separate PR.

zas:
Should the documentation updated after this PR ? See https://picard.musicbrainz.org/docs/mappings/

samj1912:
@zas yes we can update the documentation there, update the tsst and tmoo frames and remove the sort order frames from aiff files.

PR #529, opened 2017-01-06
PICARD-835: Remove contrib/plugins from repo

samj1912:
I have currently removed plugins from the installer completely.

Although I am not entirely sure if that was what was intended by this ticket. 
Did the task also include getting the latest plugins form git and adding them to the installer?


zas:
> Did the task also include getting the latest plugins form git and adding them to the installer?

Users can now installed plugins from UI, so i think it is enough.

zas:
@mwiencek @mineo : ok to merge ?

mineo:
I don't know how the installer works, but making the plugins options page the primary way of installing plugins is ok with me.

PR #531, opened 2017-01-07
PICARD-870,873,874: Fix for List options not updating and saving on reset

suhas2go:
With reference to [Issue 870](https://tickets.metabrainz.org/browse/PICARD-870), Options in Cover Art Providers and taggers scripts were not removing old entries during a reset. This PR simply does that.

https://tickets.metabrainz.org/browse/PICARD-870
https://tickets.metabrainz.org/browse/PICARD-873
https://tickets.metabrainz.org/browse/PICARD-874

### DIFF
--- a/picard/ui/options/cover.py
+++ b/picard/ui/options/cover.py
@@ -28,55 +28,6 @@ from picard.ui.sortablecheckboxlist import (
 )
 
 
-class ProviderList(SortableCheckboxListWidget):
-
-    def __init__(self, parent=None):
-        self.all_marked = False
-        self.item_marking = []
-        super(ProviderList, self).__init__(parent)
-
-    def addItem(self, item):
-        self.item_marking.append(False)
-        super(ProviderList, self).addItem(item)
-
-    def addItems(self, items):
-        for item in items:
-            self.addItem(item)
-
-    def moveItem(self, from_row, to_row):
-        self.item_marking[from_row], self.item_marking[to_row] = self.item_marking[to_row], self.item_marking[
-            from_row]
-        super(ProviderList, self).moveItem(from_row, to_row)
-
-    def set_item_marking(self, row, value):
-        self.item_marking[row] = value
-
-    def set_all_marked(self, value):
-        self.all_marked = value
-
-    def remove_marked(self):
-        if self.all_marked:
-            for i in reversed(range(len(self._SortableCheckboxListWidget__items))):
-                self._remove(i)
-            self._SortableCheckboxListWidget__items = []
-            self.all_marked = False
-            self.item_marking = []
-        else:
-            indices = []
-            for i in reversed(range(len(self._SortableCheckboxListWidget__items))):
-                if self.item_marking[i]:
-                    indices.append(i)
-                    self._remove(i)
-            self._SortableCheckboxListWidget__items = [i for j, i in enumerate(self._SortableCheckboxListWidget__items)
-                                                       if j not in indices]
-            self.item_marking = [i for j, i in enumerate(self.__item_marking) if j not in indices]
-
-    def _remove(self, row):
-        self.layout().itemAtPosition(row, self._CHECKBOX_POS).widget().setParent(None)
-        self.layout().itemAtPosition(row, self._BUTTON_UP).widget().setParent(None)
-        self.layout().itemAtPosition(row, self._BUTTON_DOWN).widget().setParent(None)
-
-
 class CoverOptionsPage(OptionsPage):
 
     NAME = "cover"
@@ -106,7 +57,7 @@ class CoverOptionsPage(OptionsPage):
         self.ui.setupUi(self)
         self.ui.save_images_to_files.clicked.connect(self.update_filename)
         self.ui.save_images_to_tags.clicked.connect(self.update_save_images_to_tags)
-        self.provider_list_widget = ProviderList()
+        self.provider_list_widget = SortableCheckboxListWidget()
         self.ui.ca_providers_list.insertWidget(0, self.provider_list_widget)
         self.ca_providers = []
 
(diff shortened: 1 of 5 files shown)


picard/ui/options/cover.py
```diff
@@ -61,6 +61,10 @@ def __init__(self, parent=None):
     def load_cover_art_providers(self):
         """Load available providers, initialize provider-specific options, restore state of each
         """
+        # Remove current provider items during a reset
+        if self.ui.ca_providers_list.count() > 1:
+            self.ui.ca_providers_list.itemAt(0).widget().setParent(None)
```
zas: I'm not fond of this approach.
If i understand correctly you test for `self.ui.ca_providers_list.count() > 1` because `ca_providers_list` has an element added by https://github.com/suhas2go/picard/blob/fc36cb0a28c150c051182bda654557316a1cf195/picard/ui/ui_options_cover.py#L62-L65

The thing is this can be modified in https://github.com/suhas2go/picard/blob/fc36cb0a28c150c051182bda654557316a1cf195/ui/options_cover.ui#L80 and if an element is added, this code will break.

Perhaps a better approach would be to use our own Widget-derived class with a marker, and iterating over widgets, selecting marked ones, remove them. So this "reset" will not depend on anything else.

suhas2go:
@zas, I've implemented ProviderList which extends SortableCheckboxListWidget to now handle markings on items, this now helps to remove the previous hardcoding. The new widget seems useful should I move it out of cover.py or perhaps add the feature to SortableCheckboxListWidget itself?

suhas2go:
In cover.py, I noticed a [minor bug](https://tickets.metabrainz.org/browse/PICARD-874). I have fixed it in this PR itself.

suhas2go:
Summary of the pull

List kind options were not being updated and/or saved on reset
- Created to new ProviderList widget for ca providers which allows to mark and remove items on reset ([PICARD 870](https://tickets.metabrainz.org/browse/PICARD-870))
- Fixed broken reset option tagger scripts and plugins([PICARD-873](https://tickets.metabrainz.org/browse/PICARD-873)) 
- Few lines of code to make ca providers option to save properly on reset ([PICARD-874](https://tickets.metabrainz.org/browse/PICARD-874))

picard/ui/options/plugins.py
```diff
@@ -159,7 +158,19 @@ def _populate(self):
 
         self._user_interaction(True)
 
+    def _remove_all(self):
+        self._user_interaction(False)
+        for i, p in self.items.items():
+            idx = self.ui.plugins.indexOfTopLevelItem(i)
+            item = self.ui.plugins.takeTopLevelItem(idx)
+            del item
```
mineo: I'm not entirely sure why you're calling `del item` here, but not, for example, `del idx`. In general, you don't need to call `del` on small variables that go out of scope in a few lines.

suhas2go: I guess that is because item here is a TreeWidgetItem. (also I just used the code from reload_list_of_plugins function)

picard/ui/options/cover.py
```diff
@@ -28,6 +28,53 @@
 )
 
 
+class ProviderList(SortableCheckboxListWidget):
```
suhas2go: Done. Fixed an underlying bug in the process :)

suhas2go:
Changes made.

suhas2go:
Oops, closed the PR by mistake.

picard/ui/sortablecheckboxlist.py
```diff
@@ -101,6 +106,33 @@ def _emit_changed(self):
         if not self.__no_emit:
             self.changed.emit(self.__items)
 
+    def set_item_marking(self, row, value):
```
mineo: I still do not understand what this marking code does :(

suhas2go: It's not really required for this PR as I can just remove all items.. I implemented it initially and didn't want to remove the feature of delayed delete. Should I?

mineo: Are there any plans to use that feature in the near future? If not, remove it, it's easy enough to add it again if needed. That would also allow merging set_all_marked and remove_marked into one function (looking at the documentation of a few qt widgets, this functions seems to be called `clear` most of the time).

mineo:
Additionally, this is supposed to solve [PICARD-873](https://tickets.metabrainz.org/browse/PICARD-873), but afaict it doesn't do that, all plugins that were enabled before restoring the default settings are still enabled.

suhas2go:
@mineo Interesting. Actually, the items in plugins and scripts list were duplicating just like [PICARD 870](https://tickets.metabrainz.org/browse/PICARD-870). I only the fixed that. Seems like plugins are more broken than other lists. I will fix this and update soon.

suhas2go:
@mineo Plugin manager had to be updated manually during a reset. I've done `plugin.enabled = False` but I'm not sure if this is enough.

picard/ui/options/dialog.py
```diff
@@ -170,6 +170,12 @@ def _restore_default(self, page):
             if option.section == 'setting':
                 old_options[option.name] = config.setting[option.name]
                 config.setting[option.name] = option.default
+
+        # Plugin manager has to be updated
```
mineo: This change (and #551) both implement an operation to perform changes for properly resetting a page in addition to resetting the options. Instead of cluttering up this class with `if isinstance` or `if page.NAME` checks, there should be a method that the various `OptionsPage`s can implement that does this (similar to load, check, save, ...).

samj1912: we can have a load_default method method implemented. I guess, I'll update the implementation and add a load_defaults method.

zas:
@suhas2go : can you fix conflicts please ?

suhas2go:
@zas Rebased and resolved.

PR #532, opened 2017-01-07
PICARD-222: Add keyboard shortcuts for tag editing

samj1912:


picard/ui/metadatabox.py
```diff
@@ -186,6 +186,9 @@ def __init__(self, parent):
         self.changes_first_action.setChecked(config.persist["show_changes_first"])
         self.changes_first_action.toggled.connect(self.toggle_changes_first)
         self.browser_integration = BrowserIntegration()
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+R")), self, self.remove_selected_tags)
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+A")), self, partial(self.edit_tag, ""))
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+E")), self, partial(self.edit_selected_tag))
```
zas: The lack of context comment for translators and the lack of portability between platforms makes me think we should perhaps provide a way to re-assign such shortcuts.

samj1912: I can add the context comment for translators here?

picard/ui/metadatabox.py
```diff
@@ -186,6 +186,12 @@ def __init__(self, parent):
         self.changes_first_action.setChecked(config.persist["show_changes_first"])
         self.changes_first_action.toggled.connect(self.toggle_changes_first)
         self.browser_integration = BrowserIntegration()
+        # TR: Keyboard shortcut for "Add New Tag..."
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+A")), self, partial(self.edit_tag, ""))
+        # TR: Keyboard shortcut for "Edit..." (tag)
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+E")), self, partial(self.edit_selected_tag))
+        # TR: Keyboard shortcut for "Remove" (tag)
+        QtGui.QShortcut(QtGui.QKeySequence(_("Alt+Shift+R")), self, self.remove_selected_tags)
```
zas: Ok, better than nothing, in the future i hope we'll be able to use gettext contexts (see https://bugs.python.org/issue2504)

Sophist-UK:
>Ok, better than nothing, in the future i hope we'll be able to use gettext contexts (see https://bugs.python.org/issue2504)

Given the long-running nature of discussion on this "bug" (or rather enhancement request) I doubt that Python is going to be enhanced any time soon.

The problem with translating "Alt+Shift+R" is that you cannot be certain that you want the same translation everywhere you might use this key combination.

So Alt-Shift-R for "Right" would be Alt-Shift-D for "Droite" in french.
But Alt-Shift-R for "Redo" would be Alt-Shift-R for "Refaire" in french. 

However, I don't think you can use a language independent string like "Right-Alt-Shift-R" and convert this to "Alt-Shift-R" in english and "Alt-Shift-D" in french, can you?

And the alternative of creating a Options UI-based mapping seems like major overkill. So this may well be the best of a bad selection of alternatives.

mineo:
Is there any way for users to discover these shortcuts without reading the source code or changelog?

Also, what's the reason for using other shortcuts than the ones mentioned in the ticket?

Sophist-UK:
Tooltips?

samj1912:
I can add tooltips, the reason I did not add all these shortcuts to context menu options is because they only worked when the menu was instantiated.

So I added independent functions to do the job for selected tags.

I can possibly add them as tooltips.

Also the reason for using shortcuts other than the ones specified in the ticket is because some of them were already assigned(alt+e) and I wanted uniformity

Sophist-UK:
Perhaps we are looking at this wrong. What does e.g. Microsoft Word do for a save Shortcut in different countries? In English versions it is Ctrl-S. What about in other countries? Is it still Ctrl-S? In other words, does other software simply ave a standard set of shortcuts regardless of whether they match the local words for e.g. Save?

samj1912:
@Sophist-UK the problem is not that, sometimes people use non-US keyboard layouts, I think it is maybe to tackle that?

Sophist-UK:
I have never used e.g. an arabic or hebrew keyboard - do they have latin characters on them as well?

However, my same suggestion still stands? What does e.g. Microsoft Word do on these keyboards?

(The idea being to use the same shortcuts commonly used in other programs that the users are likely to be using.)

PR #490, opened 2016-11-18
PICARD-840: Warn people who try to load very large amounts of files

emhazen:
Default maximum: 5000

The default maximum can be changed by the user in Options > General within the range 1 - 100,000,000.

If a user tries to load more files than the set maximum, a message box will alert the user and ask if they would like to continue.

PR #448, opened 2015-12-13
PICARD-226: Automatically cluster new files

opatel99:

PR #533, opened 2017-01-07
PICARD-226: Add option for auto-clustering new files

samj1912:


picard/ui/options/general.py
```diff
@@ -41,6 +41,7 @@ class GeneralOptionsPage(OptionsPage):
         config.TextOption("persist", "oauth_refresh_token", ""),
         config.BoolOption("setting", "analyze_new_files", False),
         config.BoolOption("setting", "ignore_file_mbids", False),
+        config.BoolOption("setting", "cluster_new_files", False),
```
zas: We have a choice to make here. I think the goal here is to help new comers, so setting this to True would make sense, though it would change behavior for "old" users, which is a bad idea.
What do you think?

samj1912: I think we should leave it to False for 1.4 and introduce all these "automatic" changes from further versions onwards.

Sophist-UK: Could we set it to True, but use an upgrade function to set it to false for existing users?

samj1912: Clustering files after each addition works slightly different than if all the files were loaded and then the cluster button was clicked ( because of the way the cluster function works if a single file is passed to it). It gives slightly more preference to "Same origin folder" based clustering. 

That is why I set it as false.

mineo:
This looks *very* similar to #448 and (judging from a quick test) it has the same problems when loading a large number of files (but less than the limit suggested in #490). I think I had a branch somewhere that solved PICARD-226 and didn't have those problems - lets see if I can find it somewhere.

samj1912:
> This looks very similar to #448 and (judging from a quick test) it has the same problems when loading a large number of files (but less than the limit suggested in #490). I think I had a branch somewhere that solved PICARD-226 and didn't have those problems - lets see if I can find it somewhere.

Let me know if you find it, otherwise I will implement it as you suggested :)

PR #536, opened 2017-01-08
PICARD-876, 881: Fix tag display in metadatabox

samj1912:


picard/formats/asf.py
```diff
@@ -228,7 +228,9 @@ def _remove_deleted_tags(self, metadata, tags):
                     del tags[real_name]
 
     def supports_tag(self, name):
-        return name in self.__TRANS
+        return (name in self.__TRANS
+                or name in ('~rating', '~length', 'totaldiscs', 'totaltracks')
```
mineo: Do you know to which extent totaltracks is supported by ASF? It's not listed as supported on https://picard.musicbrainz.org/docs/mappings/ and I can't see anything that loads or saves it.

samj1912: Yeah, checked WMP specification. Removed totaltracks from delete method as well in PR #541

PR #537, opened 2017-01-08
PICARD-207: Edit preserve tags list from metadatabox context menu

samj1912:


picard/ui/metadatabox.py
```diff
@@ -309,11 +309,23 @@ def contextMenuEvent(self, event):
...
         event.accept()
 
+    def add_to_preserved_tags(self):
+        tags = self.selected_tags()
+        preserved_tags = config.setting['preserved_tags'].split(',')
+        for tag in preserved_tags:
+            if tag.strip() == "":
+                preserved_tags.remove(tag)
+        preserved_tags.extend(tags)
+        config.setting['preserved_tags'] = ", ".join(preserved_tags)
```
mineo: This adds an additional space after every comma in the setting each time this function is used because the values in `preserved_tags` contain spaces after being split.

picard/ui/metadatabox.py
```diff
@@ -309,11 +309,23 @@ def contextMenuEvent(self, event):
             if len(tags) == 1 or removals or useorigs:
                 menu.addSeparator()
             menu.addAction(self.add_tag_action)
+            add_to_preserved_tags_action = QtGui.QAction(_(u"Add selected tags to 'Preserve Tags' list"), self.parent)
```
mineo: All other actions capitalize the first letter of every word.

samj1912:
Fixed as per reviews. As for musicbrainz_* tags, they exist in the auto-complete for preserved tags list, so I am using the same list used for the completer there.

Should we change both?

picard/ui/metadatabox.py
```diff
@@ -309,16 +309,29 @@ def contextMenuEvent(self, event):
...
         event.accept()
 
+    def add_to_preserved_tags(self):
+        # Get all selected tags including '~length'
+        tags = self.selected_tags()
+        preserved_tags = [tag for tag in
+                          map(lambda x: x.strip(), config.setting['preserved_tags'].split(','))
+                          if tag != ""]
+        preserved_tags.extend(tags)
+        config.setting['preserved_tags'] = ", ".join(filter(lambda x: x in TAG_NAMES.keys(),
```
mineo: Unfortunately, `TAG_NAMES` can't be used like this because it's not a complete list of tag names that can be written (such a list would be impossible to make with formats where tag names can be anything).

Additionally, enabling the content menu action and then silently not adding the tag to the list is not good.

mineo: Minor nit: this filter is a bit too complicated, you could just write (I did not actually test this ;-) ):

```python
[...].join(tag for tag in set(preserved_tags) if tag in TAG_NAMES.keys())
```

samj1912: Looking at the code more closely it, seems we can add arbitrary tags to preserved tags and not just auto-completed ones. I will remove the filter then :)

samj1912:
I looked at the code again properly and apparently it allows all tags, not just the autocompleted ones to be added to the preserved tags list.

Updated as needed.

picard/ui/metadatabox.py
```diff
@@ -309,16 +309,28 @@ def contextMenuEvent(self, event):
             if len(tags) == 1 or removals or useorigs:
                 menu.addSeparator()
             menu.addAction(self.add_tag_action)
+            add_to_preserved_tags_action = QtGui.QAction(_(u"Add Selected Tags to 'Preserve Tags' List"), self.parent)
```
mineo: Thinking about this some more (again), I think you can drop "Selected Tags" here because neither "Edit" nor "Remove" have it.

samj1912:
Edit disappears because it is shown only for 1 tag. We can change the behavior to allow adding to preserved tags list only when a single tag is selected. Then it is possible.

samj1912:
Updated :) I think this alternative should be fine?

picard/ui/metadatabox.py
```diff
@@ -317,11 +323,18 @@ def contextMenuEvent(self, event):
         menu.exec_(event.globalPos())
         event.accept()
 
+    def add_to_preserved_tags(self, name):
+        preserved_tags = [tag for tag in
+                          map(lambda x: x.strip(), config.setting['preserved_tags'].split(','))
+                          if tag != ""]
+        preserved_tags.append(name)
+        config.setting['preserved_tags'] = ", ".join(list(set(preserved_tags)))
```
zas: Just a minor thing i just noticed, we don't need to convert to list, as `str.join()` takes an iterable as argument: `", ".join(set(preserved_tags))`

Ref: https://docs.python.org/2/library/stdtypes.html#str.join

samj1912: Cool! 
Didn't know that, have always been using the quick hack of list(set()) to remove duplicates :P

samj1912:
Updated :)

picard/ui/metadatabox.py
```diff
@@ -328,7 +328,7 @@ def add_to_preserved_tags(self, name):
                           map(lambda x: x.strip(), config.setting['preserved_tags'].split(','))
                           if tag != ""]
         preserved_tags.append(name)
-        config.setting['preserved_tags'] = ", ".join(list(set(preserved_tags)))
+        config.setting['preserved_tags'] = ", ".join((set(preserved_tags)))
```
zas: You can also drop extra parenthesis ;)

samj1912:
Updated :)

picard/ui/metadatabox.py
```diff
@@ -268,12 +268,18 @@ def closeEditor(self, editor, hint):
     def contextMenuEvent(self, event):
         menu = QtGui.QMenu(self)
         if self.objects:
-            tags = self.selected_tags()
+            tags = self.selected_tags(discard=('~length'))
             if len(tags) == 1:
+                selected_tag = list(tags)[0]
```
mineo: `list(tags)` is now done twice.

picard/ui/metadatabox.py
```diff
@@ -268,12 +268,18 @@ def closeEditor(self, editor, hint):
...
                 edit_tag_action = QtGui.QAction(_(u"Edit..."), self.parent)
                 edit_tag_action.triggered.connect(partial(self.edit_tag, list(tags)[0]))
                 edit_tag_action.setShortcut(self.edit_tag_shortcut.key())
+                edit_tag_action.triggered.connect(partial(self.edit_tag, selected_tag))
```
mineo: This looks to be the same as 2 lines above?

picard/ui/metadatabox.py
```diff
@@ -317,11 +323,18 @@ def contextMenuEvent(self, event):
         menu.exec_(event.globalPos())
         event.accept()
 
+    def add_to_preserved_tags(self, name):
+        preserved_tags = [tag for tag in
+                          map(lambda x: x.strip(), config.setting['preserved_tags'].split(','))
+                          if tag != ""]
+        preserved_tags.append(name)
+        config.setting['preserved_tags'] = ", ".join(set(preserved_tags))
```
mineo: Using a `set` here doesn't preserve the order of elements in the setting. It's desirable to keep that because not all tags are conceptually independent, so you might have `replaygain_album_gain`, `replaygain_track_gain` that you enter one after another, but building the set shuffles the order. Fortunately, there's a `uniquify` function in `picard.util` that removes duplicates and keeps the order of elements.

picard/ui/metadatabox.py
```diff
@@ -268,12 +268,18 @@ def closeEditor(self, editor, hint):
...
-            tags = self.selected_tags()
+            tags = self.selected_tags(discard=('~length'))
             if len(tags) == 1:
+                selected_tag = list(tags)[0]
                 edit_tag_action = QtGui.QAction(_(u"Edit..."), self.parent)
                 edit_tag_action.triggered.connect(partial(self.edit_tag, list(tags)[0]))
                 edit_tag_action.setShortcut(self.edit_tag_shortcut.key())
+                edit_tag_action.triggered.connect(partial(self.edit_tag, selected_tag))
                 menu.addAction(edit_tag_action)
+                if selected_tag in TAG_NAMES:
```
mineo: Sorry, I think you added this check after my comment about enabling the action unconditionally. Looking back, that has to be one of the most vague comments I've ever written ;-) What I actually meant was that the action shouldn't be enabled or shouldn't be in the menu at all if the tag is already in the preserved tags list (the context menu here seems to use the latter option).

samj1912:
Also added an option to remove the tag from preserved list if it exists.

picard/ui/metadatabox.py
```diff
@@ -317,13 +329,21 @@ def contextMenuEvent(self, event):
         menu.exec_(event.globalPos())
         event.accept()
 
+    def add_to_preserved_tags(self, name, preserved_tags):
+        preserved_tags.append(name)
+        config.setting['preserved_tags'] = ", ".join(uniqify(preserved_tags))
+
+    def remove_from_preserved_tags(self, name, preserved_tags):
+        preserved_tags = filter(lambda x: x != name, preserved_tags)
+        config.setting['preserved_tags'] = ", ".join(uniqify(preserved_tags))
```
zas: This small code duplication (line 334 & 338) could be solved using a small static method.

samj1912:
Updated :)

picard/ui/metadatabox.py
```diff
@@ -317,13 +329,25 @@ def contextMenuEvent(self, event):
         menu.exec_(event.globalPos())
         event.accept()
 
+    @staticmethod
+    def _make_preserved_tags_string(preserved_tags):
+        return ", ".join(uniqify(preserved_tags))
```
zas: I was thinking about:

```python
@staticmethod
def _save_preserved_tags(preserved_tags):
    config.setting['preserved_tags'] = ", ".join(uniqify(preserved_tags))
```

samj1912:
Updated :)

picard/ui/metadatabox.py
```diff
@@ -317,13 +329,25 @@ def contextMenuEvent(self, event):
...
+    @staticmethod
+    def _save_preserved_tags(preserved_tags):
+        config.setting['preserved_tags'] = ", ".join(uniqify(preserved_tags))
+
+    def add_to_preserved_tags(self, name, preserved_tags):
+        preserved_tags.append(name)
+        self._save_preserved_tags(preserved_tags)
+
+    def remove_from_preserved_tags(self, name, preserved_tags):
+        preserved_tags = filter(lambda x: x != name, preserved_tags)
```
mineo: `preserved_tags.remove(name)`

samj1912: Did not do that in case there are multiple instances of the name ( maybe entered by the user)

PR #538, opened 2017-01-08
PICARD-291: Non-recursively add a directory

samj1912:


picard/tagger.py
```diff
@@ -412,7 +412,11 @@ def process(result=None, error=None):
                     self.add_files(result)
                 thread.run_task(get_files, process)
 
-        process(True, False)
+        if config.setting['recursively_add_files']:
+            process(True, False)
+        else:
+            self.add_files(os.path.join(path, f) for f in os.listdir(path))
```
mineo: Calling `add_files` here creates a new codepath without all the statusbar and debugging messages that `get_files` has, which is not good.

picard/ui/ui_options_general.py
```diff
@@ -80,6 +80,9 @@ def setupUi(self, GeneralOptionsPage):
         self.ignore_file_mbids = QtGui.QCheckBox(self.groupBox_2)
         self.ignore_file_mbids.setObjectName(_fromUtf8("ignore_file_mbids"))
         self.verticalLayout.addWidget(self.ignore_file_mbids)
+        self.recursively_add_files = QtGui.QCheckBox(self.groupBox_2)
+        self.recursively_add_files.setObjectName(_fromUtf8("recursively_add_files"))
```
mineo: Is the general options page really the correct one for this setting? I'm not sure why anyone would want such a feature in the first place, but I imagine it would be better to have it on the Advanced page next to the ignore hidden files checkbox.

picard/tagger.py
mineo: Trying this I saw the following error in the log:

> E: 17:55:57 Error occured:

I could only find "Error occured" in `picard/formats/__init__.py`. I guess this is because this also passes directory names to `add_files`.

samj1912:
Updated :)

picard/tagger.py
```diff
@@ -412,7 +412,36 @@ def process(result=None, error=None):
...
+                log.debug("Adding %(count)d files from '%(directory)s'" %
+                          mparms)
+                self.window.set_statusbar_message(
+                    ungettext(
+                        "Adding %(count)d file from '%(directory)s' ...",
+                        "Adding %(count)d files from '%(directory)s' ...",
+                        number_of_files),
+                    mparms,
+                    translate=None,
+                    echo=None
```
samj1912: I was replicating the existing behaviour for get_files, should that be changed too?

zas: Yes, but in another PR, since the change will be unrelated.

samj1912: Okay :)

picard/tagger.py
```diff
@@ -412,7 +412,32 @@ def process(result=None, error=None):
                     self.add_files(result)
                 thread.run_task(get_files, process)
 
-        process(True, False)
+        def get_files_non_recursive():
+            listings = [f for f in os.listdir(path)]
+            if ignore_hidden:
+                listings = filter(lambda x: not is_hidden(x), listings)
```
mineo: This is not correct. `is_hidden` requires the full path to work correctly on all platforms, but x is only the filename at this point.

picard/tagger.py
```diff
@@ -412,7 +412,32 @@ def process(result=None, error=None):
                     self.add_files(result)
                 thread.run_task(get_files, process)
 
-        process(True, False)
+        def get_files_non_recursive():
+            listings = [f for f in os.listdir(path)]
```
mineo: I think it would be better to not use all these lambdas here, but instead just use a single loop:

```python
files = []
for filename in os.listdir(path):
   join()
   if is_hidden() or not isfile():
     continue
   files.append()
```

picard/tagger.py
```diff
@@ -412,7 +412,32 @@ def process(result=None, error=None):
...
+                self.window.set_statusbar_message(
+                    ungettext(
+                        "Adding %(count)d file from '%(directory)s' ...",
+                        "Adding %(count)d files from '%(directory)s' ...",
+                        number_of_files),
+                    mparms,
+                    translate=None)
+            return files
+
+        if config.setting['recursively_add_files']:
```
mineo: At this point, `add_directory` only dispatches to either `process` or `get_files_non_recursive`. Imho it should instead dispatch to either `_add_directory_recursive` or `_add_directory_non_recursive` that are new methods on the tagger class. That would also help in avoiding unnecessary variables that are not needed in both cases (like `walk`).

samj1912:
Updated :)

PR #539, opened 2017-01-09
PICARD-305: Add files to 'Pending Files' status on Lookup action

samj1912:


picard/file.py
```diff
@@ -511,6 +511,7 @@ def _lookup_finished(self, lookuptype, document, http, error):
         if self.state == File.REMOVED:
             return
 
+        self.clear_pending()
```
mineo: Why is this necessary? AFAICT all paths through this function call `clear_pending` already.

samj1912:
Updated :)

Sophist-UK:
Resolves: [PICARD-305](https://tickets.metabrainz.org/browse/PICARD-305)

PR #540, opened 2017-01-10
Remove redundant debug log

samj1912:


mineo:
This is zas' territory, but I though the point of `set_statusbar_message` was to not log.debug translated messages (for example in acoustid.py:167, acoustidmanager.py:102, collection.py:{63,85}).

zas:
> This is zas' territory, but I though the point of set_statusbar_message was to not log.debug translated messages (for example in acoustid.py:167, acoustidmanager.py:102, collection.py:{63,85}).

Hmmm, you're totally right. Better keep debug messages untranslated. I forgot about this :(
@samj1912 : ^^ sorry

PR #542, opened 2017-01-11
PICARD-883: Improved Multiple Scripts UI

suhas2go:
The multiple scripting UI that I first created was fairly basic. This PR drastically improves the UI :)
Here's the [jira ticket](https://tickets.metabrainz.org/browse/PICARD-883) 

Screenshots attached
new_ui1
new_ui2


zas:
Good job on this, i have few remarks:
- wouldn't be better to have a normal  button "Add new script" instead of "+" with "Add script" tooltip button ?
- The small "..." button shows a context menu with remove/rename options, it lacks a tooltip, also i wonder if usual right-click context menu would be enough.
- I think remove script option should ask for confirmation (usual Are you sure ...), since it is a destructive operation.

suhas2go:
@zas I'll add the confirmation message and see how 'Add new script' looks. About "..." button, I made it keeping in mind that many other options like 'open in other editor', 'save to file' ..all can be added within it.

scripts/picard
```diff
@@ -0,0 +1,3 @@
+#!/usr/bin/env python
+from picard.tagger import main; main('/usr/share/locale', True)
```
zas: What is the purpose of this ?

suhas2go: I'm sorry, while building the ui, I once mistakenly pressed enter at 'python setup.py build' instead of ..build_ui. This has been generated as a by-product of this. I have removed the file and updated the PR.

suhas2go:
@zas Fixed.

picard/ui/options/scripting.py
```diff
@@ -98,6 +99,88 @@ def highlightBlock(self, text):
...
+        super(AdvancedScriptItem, self).__init__(parent)
+        layout = QtGui.QGridLayout()
+        layout.setHorizontalSpacing(5)
+        layout.setVerticalSpacing(2)
+        layout.setContentsMargins(0, 0, 0, 0)
+        self.setLayout(layout)
+
+        checkbox = QtGui.QCheckBox()
+        checkbox.setChecked(state)
+        checkbox.setText(name)
```
zas: This is a bit awkward, because when one wants to select another script by click, clicking on the text check/uncheck the checkbox, instead of selecting the line.
I think you should use a checkbox without text in first column and use another column for script name.

suhas2go: Done

Sophist-UK:
Just tried this functionality, and the UI needs a minor tweak to make the divider between the list of scripts and the scripts themselves movable and ideally to make the default width of the list of scripts smaller so that the width of the scripts themselves are larger when the options window is at default size.

zas:
> Just tried this functionality, and the UI needs a minor tweak to make the divider between the list of scripts and the scripts themselves movable and ideally to make the default width of the list of scripts smaller so that the width of the scripts themselves are larger when the options window is at default size.

If you can provide a fix as a PR before tomorrow, we can merge it. Else it will have to wait 1.4.1.

Sophist-UK:
No – I don’t have time to code and fix. But so long as there is going to be a 1.4.1 soon, its not a problem.

zas:
> Just tried this functionality, and the UI needs a minor tweak to make the divider between the list of scripts and the scripts themselves movable and ideally to make the default width of the list of scripts smaller so that the width of the scripts themselves are larger when the options window is at default size.

@suhas2go : can you have a look ? If it is trivial enough, just do it so we include it in 1.4.0 else it'll wait.
@Sophist-UK : in any case, please submit a ticket to JIRA

suhas2go:
@samj1912 Thanks. :+1:

suhas2go:
By default, the list of scripts is still larger. #575 fixes this.

PR #543, opened 2017-01-11
PICARD-319: Add an action toolbar to file browser

samj1912:


samj1912:
I want to add a couple more buttons to this toolbar 

1. A set home/set starting directory button
2. A move files here button
3. Maybe a toggle hidden files button?

zas:
It would be nice to provide a screenshot for such UI changes in the PR description.

picard/ui/filebrowser.py
```diff
@@ -23,7 +23,51 @@
...
+        paths = self._selected_paths()
+        files = []
+        directories = []
+        for path in paths:
+            if os.path.isfile(path):
+                files.append(path)
+            else:
+                directories.append(path)
+
+        if len(files):
```
zas: `if files:` should be enough.

picard/ui/filebrowser.py
```diff
@@ -23,7 +23,51 @@
...
+            self.tagger.add_files(files)
+        if len(directories):
```
zas: `if directories:` should be enough

samj1912:
Screenshot

Screenshot of the same. I plan to add 3 more buttons to that toolbar

zas:
> Screenshot of the same. I plan to add 3 more buttons to that toolbar

Imho this should be discussed, i'm not an UI guy at all, but i have the feeling adding more buttons don't make things easier.
I'll all for changes in the UI. What about creating a thread about this on forums so non-devs can participate ?

samj1912:
I modified the code to show text labels under icons, but I don't think that is meant for the filebrowser toolbar.

filebrowser


This is how it looks and I think it kind of defeats the purpose of having a small toolbar. Also the '+' sign is intuitive enough and I have tooltips for more info.

mineo:
The + sign is intuitive for the single add-items-to-picard function, but if more buttons are going to be added, not all of them are going to have an obvious meaning. We should take this into consideration because either more buttons are going to be added or the Add one is going to stay the only one.
In the latter case, having it just say "Add" and use the whole width of the toolbar instead of being a lonely icon on the left might be preferrable.

samj1912:
I am thinking of "Add new folder" button. Which I think will be intuitive as it is used a lot. and a move files here icon like this? move files

picard/ui/filebrowser.py
```diff
@@ -23,7 +23,51 @@
...
+
+    def add_items(self):
+        paths = self._selected_paths()
+        files = []
+        directories = []
+        for path in paths:
+            if os.path.isfile(path):
+                files.append(path)
+            else:
+                directories.append(path)
```
zas: Why not add directories directly here ?

I understand why you are building a list of files to pass to `tagger.add_files()`, but you may skip this for directories and just do `self.tagger.add_directory(path)` here.

picard/ui/filebrowser.py
```diff
@@ -23,7 +23,51 @@
...
+        self.file_browser.save_state()
+
+    def restore_state(self):
+        self.file_browser.restore_state()
+
+    def _selected_paths(self):
+        paths = []
+        for index in self.file_browser.selectedIndexes():
+            paths.append(self.file_browser.model.filePath(index))
+        return paths
```
zas: A list comprehension is prolly faster:
```python
return [self.file_browser.model.filePath(index) for index in self.file_browser.selectedIndexes()]
```

samj1912:
image

Updated look for add items button. I think we can add other buttons after more discussion.

samj1912:
This PR is complete from my side pending any reviews :)

mineo:
To be honest, I still can't get myself to like that toolbar very much :( On your screenshot it looks really good, but on a higher resolution the toolbar is 70%+ empty whitespace. Maybe this is just because there's only one button now there at the moment...

Since I don't have any other suggestions, I'll leave it to zas to decide about this.

zas:
> To be honest, I still can't get myself to like that toolbar very much :( On your screenshot it looks really good, but on a higher resolution the toolbar is 70%+ empty whitespace. Maybe this is just because there's only one button now there at the moment...

I have the same feeling. I think we can do better, not sure yet how.

mineo:
Let's postpone this until we have more operations that we'd like to put into that toolbar.

PR #544, opened 2017-01-11
PICARD-884: Add the parent directory if file selected from filebrowser

samj1912:


picard/ui/filebrowser.py
```diff
@@ -134,10 +134,16 @@ def move_files_here(self):
...
+        config.setting["move_files_to"] = destination
 
     def set_as_starting_directory(self):
         indexes = self.selectedIndexes()
         if indexes:
             path = self.model.filePath(indexes[0])
-            config.setting["starting_directory_path"] = os.path.normpath(unicode(path))
+            destination = os.path.normpath(unicode(path))
+            if not os.path.isdir(destination):
+                destination = os.path.dirname(destination)
```
zas: You could reduce code redundancy here:

```python
def path2destination(path):
            destination = os.path.normpath(unicode(path))
            if not os.path.isdir(destination):
                destination = os.path.dirname(destination)
           return destination
...
config.setting["move_files_to"] = path2destination(path)
...
config.setting["starting_directory_path"] = path2destination(path)
```

Feel free to find a better function name, i suck at it ;)

samj1912:
Updated :)

PR #545, opened 2017-01-11
PICARD-885: Remove unsupported MPEG4 files from tagger

samj1912:
https://tickets.metabrainz.org/browse/PICARD-885

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -108,6 +108,12 @@ class File(QtCore.QObject, Item):
         if error is not None:
             self.error = str(error)
             self.state = self.ERROR
+            from picard.formats import supported_extensions
+            file_name, file_extension = os.path.splitext(self.base_filename)
+            if file_extension not in supported_extensions():
+                self.remove()
+                log.error('Unsupported media file {} wrongly loaded. Removing ...'.format(self))
+                return
         else:
             self.error = None
             self.state = self.NORMAL
--- a/picard/file.py
+++ b/picard/file.py
@@ -112,7 +112,7 @@ class File(QtCore.QObject, Item):
             file_name, file_extension = os.path.splitext(self.base_filename)
             if file_extension not in supported_extensions():
                 self.remove()
-                log.error('Unsupported media file {} wrongly loaded. Removing ...'.format(self))
+                log.error('Unsupported media file %r wrongly loaded. Removing ...',self)
                 return
         else:
             self.error = None
--- a/picard/formats/__init__.py
+++ b/picard/formats/__init__.py
@@ -40,6 +40,11 @@ def supported_formats():
     return formats
 
 
+def supported_extensions():
+    extensions = [ext for exts, name in supported_formats() for ext in exts]
+    return extensions
+
+
 def guess_format(filename, options=_formats):
     """Select the best matching file type amongst supported formats."""
     results = []
--- a/picard/formats/__init__.py
+++ b/picard/formats/__init__.py
@@ -34,15 +34,12 @@ def register_format(format):
 
 def supported_formats():
     """Returns list of supported formats."""
-    formats = []
-    for format in _formats:
-        formats.append((format.EXTENSIONS, format.NAME))
-    return formats
+    return [(format.EXTENSIONS, format.NAME) for format in _formats]
 
 
 def supported_extensions():
-    extensions = [ext for exts, name in supported_formats() for ext in exts]
-    return extensions
+    """Returns list of supported extensions."""
+    return [ext for exts, name in supported_formats() for ext in exts]
 
 
 def guess_format(filename, options=_formats):
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -118,14 +118,6 @@ class MP4File(File):
             file.add_tags()
 
         metadata = Metadata()
-        if not tags:
-            # MPEG-4 file in an mov container is unsupported and is one of
-            # the formats which returns a positive score in guess_format method
-            # If we are unable to add tags, it is not a supported mp4 file
-            # remove it and log an error
-            log.error("Unable to load tags for file {}".format(self))
-            self.remove()
-            return metadata
 
         for name, values in tags.items():
             if name in self.__text_tags:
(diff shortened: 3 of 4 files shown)


picard/formats/mp4.py
```diff
@@ -118,6 +118,14 @@ def _load(self, filename):
             file.add_tags()
 
         metadata = Metadata()
+        if not tags:
+            # MPEG-4 file in an mov containter is unsupported and is one of
```
zas: s/containter/container/

picard/formats/mp4.py
```diff
@@ -118,6 +118,14 @@ def _load(self, filename):
...
+            # the formats which returns a positive score in guess_format method
+            # If we are unable to add tags, it is not a supported mp4 file
+            # remove it and log an error
+            log.error("Unable to load tags for file {}".format(self))
+            self.remove()
+            return
```
zas: I wonder about that, what do callers expect as return value ?

samj1912: metadata

samj1912:
Returning a blank dictionary(metadata parent type) now instead and moved the metadata instantiation after checking if tags exist.

mineo:
Just making sure: you did test this, right? AFAICT, after `_load`, `_loading_finished` is called, which calls `_copy_loaded_metadata` which accesses a `length` attribute that doesn't exist on dicts.

mineo:
Also, can't this now be a problem for every file type? Their score methods are all only an indicator but no promise that the file can be loaded.

samj1912:
> Just making sure: you did test this, right? AFAICT, after _load, _loading_finished is called, which calls _copy_loaded_metadata which accesses a length attribute that doesn't exist on dicts.

Yes I did test it. I call a self.remove on it which causes _loading_finished to return without calling _copy_loaded_metadata since the file state is REMOVED and not PENDING.

picard/formats/mp4.py
```diff
@@ -117,6 +117,14 @@ def _load(self, filename):
...
             file.add_tags()
 
+        if not tags:
+            # MPEG-4 file in an mov container is unsupported and is one of
+            # the formats which returns a positive score in guess_format method
+            # If we are unable to add tags, it is not a supported mp4 file
+            # remove it and log an error
+            log.error("Unable to load tags for file {}".format(self))
+            self.remove()
+            return {}
```
zas: I would just return `Metadata()`. Assuming parent class is a dict looks weird to me.

samj1912:
Updated :)

zas:
> Also, can't this now be a problem for every file type? Their score methods are all only an indicator but no promise that the file can be loaded.

@samj1912 ^^

I also think it can occur with any file type, so we have to handle it for all formats.

samj1912:
Here's a better way to handle such files which are loaded. Since the only way such files can be loaded is if the guess_format returns true, it has errors and the extension is not among our supported extension list.
Can't we just remove such files? Since they wouldn't have been added in the first place?

samj1912:
Updated. Reviews?

picard/formats/__init__.py
```diff
@@ -40,6 +40,11 @@ def supported_formats():
     return formats
 
 
+def supported_extensions():
+    extensions = [ext for exts, name in supported_formats() for ext in exts]
+    return extensions
```
zas: Huh? Not sure why there is 2 `for` but i think the following should give the same result:

```python
return [ext for ext, _ in supported_formats()]
```

Also no need of intermediate variable.

picard/ui/options/about.py
```diff
@@ -44,10 +44,7 @@ def load(self):
         }
         args.update(versions.as_dict(i18n=True))
 
-        formats = []
-        for exts, name in supported_formats():
-            formats.extend(exts)
-        args["formats"] = ", ".join(formats)
+        args["formats"] = ", ".join(supported_extensions())
```
zas: I think we should drop the dot in front of each extension to shorten the display: `.mp3, .mp2, ...` to `mp3, mp2`

Also the text still says **Supported formats**, but we list extensions, and in practice we care more about **format** than **extension** (since recent changes). But i think it isn't something dramatic ;)

picard/formats/mp4.py
```diff
@@ -118,6 +118,7 @@ def _load(self, filename):
             file.add_tags()
 
         metadata = Metadata()
+
```
zas: This blank line should be removed.

picard/formats/__init__.py
```diff
@@ -40,6 +40,11 @@ def supported_formats():
     return formats
```
zas: Not totally related, but i guess you can just rewrite this to:

```python 
return [(format.EXTENSIONS, format.NAME) for format in _formats]
```

picard/formats/__init__.py
mineo: `EXTENSIONS` is a list, not a single string, because some formats support more than one extension.

picard/file.py
```diff
@@ -108,6 +108,12 @@ def _loading_finished(self, callback, result=None, error=None):
         if error is not None:
             self.error = str(error)
             self.state = self.ERROR
+            from picard.formats import supported_extensions
+            file_name, file_extension = os.path.splitext(self.base_filename)
+            if file_extension not in supported_extensions():
+                self.remove()
+                log.error('Unsupported media file {} wrongly loaded. Removing ...'.format(self))
```
mineo: `log.error` does the formatting on its own.

picard/file.py
```diff
@@ -108,6 +108,12 @@ def _loading_finished(self, callback, result=None, error=None):
         if error is not None:
             self.error = str(error)
             self.state = self.ERROR
+            from picard.formats import supported_extensions
```
mineo: Can this be moved to a top-level import?

samj1912: Nope, circular dependencies.

samj1912:
Updated. Didn't change the supported extensions/supported formats thing in about page. Will update it after deciding if we want supported formats and list all the formats or supported extensions and display the extensions.

zas:
> Updated. Didn't change the supported extensions/supported formats thing in about page. Will update it after deciding if we want supported formats and list all the formats or supported extensions and display the extensions.

It can come later, not much an issue.

PR #546, opened 2017-01-11
PICARD-882: Remove deleted tags from UI on save

samj1912:


zas:
What if file save is failing ?
From what i understand, tags are deleted from UI before actually saving the file, so it is possible they are removed in the UI before they are removed from the file.

samj1912:
They are still stored in metadata.deleted_tags so it wont affect the save from progressing.
They will just be removed from the UI.

PR #548, opened 2017-01-12
PICARD-887: Allow some scripting functions to have an arbitrary number of arguments

samj1912:


picard/script.py
```diff
@@ -475,57 +492,78 @@ def func_trim(parser, text, char=None):
         return text.strip()
 
 
-def func_add(parser, x, y):
-    """Add ``y`` to ``x``."""
+def func_add(parser, x, y, *args):
+    """Adds ``y`` to ``x``.
+       Can be used with arbitrary number of arguements. The result is
```
zas: s/arguements/arguments/

picard/script.py
```diff
@@ -475,57 +492,78 @@ def func_trim(parser, text, char=None):
...
+       $add($add(x,y)...)
+    """
     try:
-        return str(int(x) + int(y))
+        return _compute('add', locals())
```
zas: I'm not fond of using `locals()`, why not use `*args` and drop `x` and `y` ?

picard/script.py
```diff
@@ -475,57 +492,78 @@ def func_trim(parser, text, char=None):
...
     except ValueError:
         return ""
 
 
-def func_sub(parser, x, y):
-    """Subtracts ``y`` from ``x``."""
+def func_sub(parser, x, y, *args):
+    """Subtracts ``y`` from ``x``.
+       Can be used with arbitrary number of arguements. The result is
+       $sub($sub(x,y)...)
```
zas: This isn't very clear.

picard/script.py
```diff
@@ -307,6 +308,22 @@ def register_script_function(function, name=None, eval_args=True,
...
+def _compute(operation, operands):
+
+    operand_list = [operands.get('x'), operands.get('y')] + list(operands.get('args'))
+    int_operators = {'add': operator.add,
+                     'sub': operator.sub,
+                     'mul': operator.mul,
+                     'div': operator.div,
+                     'mod': operator.mod}
+    logic_operators = {'and': (lambda x, y: x and y),
+                       'or': (lambda x, y: x or y)}
```
zas: `int_operators` and `logic_operators` are constant dicts, right? I would move them outside the function, no need to re-evaluate them on each run.

picard/script.py
```diff
@@ -307,6 +308,22 @@ def register_script_function(function, name=None, eval_args=True,
...
+    if operation in int_operators:
+        return str(reduce(int_operators[operation], map(int, operand_list)))
+    elif operation in logic_operators:
```
zas: Instead of having a `if` here, i would just make 2 functions `_compute_int` and `_compute_logic` and call them appropriately.

picard/script.py
samj1912: I wanted 2 variables to be necessary.

I can just check for the length of args and raise an error though.

Any thoughts on how else it can be implemented?

mineo: You can pass them on the same way you receive them: `_compute('add', x, y, *args)` and receive them in `def _compute(operation, *args)` and use them as `args`.

picard/script.py
```diff
@@ -307,6 +308,24 @@ def register_script_function(function, name=None, eval_args=True,
     )
 
 
+_int_operators = {'add': operator.add,
```
mineo: You don't need _int_operators, you can just pass the function directly to _compute_int.

picard/script.py
```diff
@@ -307,6 +308,24 @@ def register_script_function(function, name=None, eval_args=True,
...
+                  'mul': operator.mul,
+                  'div': operator.div,
+                  'mod': operator.mod}
+
+_logic_operators = {'and': (lambda x, y: x and y),
+                    'or': (lambda x, y: x or y)}
+
+
+def _compute_int(operation, *args):
+    return str(reduce(_int_operators[operation], map(int, list(args))))
```
mineo: No need for list() here, map works on any sequence, including tuples.

picard/script.py
```diff
@@ -307,6 +308,24 @@ def register_script_function(function, name=None, eval_args=True,
...
+
+
+def _compute_logic(operation, *args):
+    return reduce(_logic_operators[operation], list(args))
```
mineo: You can skip `reduce` here (and `_logic_operators`) and implement this as `return operation(args)` where operation is either `any` or `all`.

samj1912:
Updated :)

samj1912:
Merged with https://github.com/metabrainz/picard/commit/9af52e4156ade78ef5423bb1176449f381bec07e

samj1912:
Closing this.

PR #549, opened 2017-01-12
PICARD-384: Allow wildcards for unsetting certain tags

samj1912:


picard/script.py
```diff
@@ -413,6 +413,13 @@ def func_unset(parser, name):
     """Unsets the variable ``name``."""
     if name.startswith("_"):
         name = "~" + name[1:]
+    # Allow wild-card unset for certain keys
+    if name in ('performer:*', 'comment:*', 'lyrics:*'):
```
zas: I wonder if we could make this even more generic, `anytag:*` could be used to delete all `anytag`.

samj1912: We can make it more generic, it is also possible to introduce regex.

samj1912: I chose the above ones just to include the common 'tag:name' type tags that we have.

zas: I think it is fine as is for now. We'll see later if we want to extend this.

test/test_script.py
```diff
@@ -324,4 +324,11 @@ def test_cmd_with_not_arguments(self):
             self.parser.eval("$noargstest()")
         except ScriptError:
             self.fail("Function noargs raised ScriptError unexpectedly.")
-        
+
+    def test_cmd_unset(self):
```
zas: There was no test for `unset()` ? :(
Can you please test the simple cases too (including the _/~ case) ?

samj1912: okay :)

samj1912: Done :)

PR #551, opened 2017-01-12
PICARD-888:Logout when restoring defaults for general page

samj1912:


picard/ui/options/dialog.py
```diff
@@ -174,6 +174,8 @@ def _restore_default(self, page):
         # Restore the config values incase the user doesn't save after restoring defaults
         for key in old_options:
             config.setting[key] = old_options[key]
+        if isinstance(page, general.GeneralOptionsPage):
```
mineo: See https://github.com/metabrainz/picard/pull/531#discussion_r96072631

samj1912: Fixed in #554

PR #553, opened 2017-01-13
PICARD-700: Support albumartists tag

samj1912:
As discussed with @Freso an elementary PR to provide support for albumartists tag.

### DIFF
--- a/picard/formats/asf.py
+++ b/picard/formats/asf.py
@@ -138,6 +138,7 @@ class ASFFile(File):
         'djmixer': 'WM/DJMixer',
         'mixer': 'WM/Mixer',
         'artists': 'WM/ARTISTS',
+        'albumartists': 'WM/ALBUMARTISTS',
         'work': 'WM/Work',
         'website': 'WM/AuthorURL',
     }
--- a/picard/mbxml.py
+++ b/picard/mbxml.py
@@ -195,7 +195,7 @@ def artist_credit_to_metadata(node, m, release=False):
         m["musicbrainz_albumartistid"] = ids
         m["albumartist"] = artist
         m["albumartistsort"] = artistsort
-        m["~albumartists"] = artists
+        m["albumartists"] = artists
         m["~albumartists_sort"] = artistssort
     else:
         m["musicbrainz_artistid"] = ids
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -153,6 +153,7 @@ class ID3File(File):
         'ASIN': 'asin',
         'MusicMagic Fingerprint': 'musicip_fingerprint',
         'Artists': 'artists',
+        'Album Artists': 'albumartists',
         'Work': 'work',
         'Writer': 'writer',
     }
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -102,6 +102,7 @@ class MP4File(File):
         "----:com.apple.iTunes:SCRIPT": "script",
         "----:com.apple.iTunes:LANGUAGE": "language",
         "----:com.apple.iTunes:ARTISTS": "artists",
+        "----:com.apple.iTunes:ALBUMARTISTS": "albumartists",
         "----:com.apple.iTunes:WORK": "work",
         "----:com.apple.iTunes:initialkey": "key",
     }


picard/mbxml.py
```diff
@@ -195,7 +195,7 @@ def artist_credit_to_metadata(node, m, release=False):
         m["musicbrainz_albumartistid"] = ids
         m["albumartist"] = artist
         m["albumartistsort"] = artistsort
-        m["~albumartists"] = artists
+        m["albumartists"] = artists
         m["~albumartists_sort"] = artistssort
```
Freso: What about this one?

picard/mbxml.py
```diff
@@ -195,7 +195,7 @@ def artist_credit_to_metadata(node, m, release=False):
         m["musicbrainz_albumartistid"] = ids
         m["albumartist"] = artist
         m["albumartistsort"] = artistsort
-        m["~albumartists"] = artists
```
Freso: Maybe we'll want to keep `~albumartists` (alongside `albumartists`) for the 1.4.x releases, marking it as deprecated and remove for 2.0?

picard/formats/asf.py
```diff
@@ -138,6 +138,7 @@ class ASFFile(File):
         'djmixer': 'WM/DJMixer',
         'mixer': 'WM/Mixer',
         'artists': 'WM/ARTISTS',
```
Freso: Huh. I wonder why this is like this, when it's `'artistsort': 'WM/ArtistSortOrder',` further up for `artistsort`. Maybe `'artists': 'WM/Artists',` would have been better. Does mutagen use all-uppercase? :/

picard/formats/asf.py
```diff
@@ -138,6 +138,7 @@ class ASFFile(File):
...
+        'albumartists': 'WM/ALBUMARTISTS',
```
Freso: Maybe `'albumartists': 'WM/AlbumArtists',`? (Based on `'albumartistsort': 'WM/AlbumArtistSortOrder',`.)

Freso:
Would be good to get feedback from @lazka and @sampsyo and maybe other audio tagger/player software writers so we can get some common ground for the tag names.

picard/mbxml.py
samj1912: the '~' simple means its a 'hidden/protected' tag and is appeneded to hide it from UI/saving from file tags.

picard/formats/asf.py
lazka: What's the difference between "artist" and "artists"? mutagen properly supports multiple "Author" values since 1.26.

picard/formats/asf.py
lazka: Similar question, what's the difference between "albumartist" and "albumartists"?

picard/formats/asf.py
samj1912: artists is simply a multi valued field with all the artists, artist includes things like 'feat.' etc

picard/formats/asf.py
samj1912: @lazka  same as artist and artists. artist is a string, artists is a multi valued field.

samj1912: eg Artist: Chakuza & RAF Camora
and Artists: Chakuza/RAF Camora

picard/formats/asf.py
lazka: OK, we don't have something like that in quodlibet. We have a "artist" tag containing a list of artists and save it in "Author" as is.

samj1912: Yeah, we have had some discussion on this see [PICARD-586](https://tickets.metabrainz.org/browse/PICARD-586)

I think @zas @mwiencek @mineo should be consulted since this is a big change

samj1912:
@ijabz thanks for that, I was looking for it but couldn't find it on the site :)

sampsyo:
Interesting! Multi-value tags for this kind of situation have been a long-standing request for beets, but I always assumed that we'd just do that in the current `artist` tag instead of adding a separate plural version. I see the argument from the JIRA issue about compatibility with basically all other music software on Earth, which expects the artist to be a single string, and that makes lots of sense.

But can I ask about the long-term plan here? Is the idea to create plural versions of every field that might want multiple values—for example, all the "person" fields like composer, performer, arranger, etc., and even other fields like genre?

ijabz:
Sampyso No, the key thing is the artist and album artist fields contain additional important information, the join phrases . i.e we can record if it is 'Johnny Cash & June Carter;, 'Johnny Cash with June Carter' or 'Johnny Cash Featuring June Carter' so if we just made artist/albumartist a list of artists we would lose that information which is why both the artist and artists fields are required. But for composers, performer ectera we just have  a list of people so we can just store the composers in the composer field using the formats method for making mutivalue (i.e multiple COMPOSER tags for OggVorbis, null separated frame for ID3)

picard/formats/asf.py
Sophist-UK: See [Jaudiotagger Tag Mappings](http://www.jthink.net/jaudiotagger/tagmapping.html)

Sophist-UK: Can;t remember whether tags in ASF format are case independent or not.

Sophist-UK: A little bit of internet research suggests that in WMP 11 and below they were case insensitive, but in WMP they are case sensitive. I suspect that Jaikoz created the tag first and we then adopted it with the case defined, but that may be wrong. Regardless, the case is now fixed by historical usage.

picard/mbxml.py
Sophist-UK: I created ~albumartists.

I would suggest that we dispose of this now, and instead create update code that replaces `%_albumartists%` with `%albumartists%` .

It might break plugins, but with the exception of my `sort_multivalue_tags` plugin (which strangely already has `albumartists` instead of `~albumartists`) there are no other plugins I know of which use this.

picard/mbxml.py
Sophist-UK: I created this one too. I vote for removing the `~`.

picard/formats/asf.py
ijabz: jaudiotagger tag mappings doc is out of date actually best to use the spreadsheet link below, the jaudiotagger source code on the songkong help http://www.jthink.net/songkong/help/help.html#d5e2051for uptodate mappings

samj1912:
Closing this for now.

Sophist-UK:
I think that the "needs more work" is limited to not deleting line 198. We are NOT removing `albumartist` (singular) which has the join phrases, but instead adding a new multi-value tag which is the list of artists.

ijabz:
and using the same internal format specific name to match what we already used in Jaikoz/Songkong

Sophist-UK:
@ijabz - Yes - same as Jaikoz/Songkong but only if that makes sense with Picard.

ijabz:
What do you mean, why would it not make sense, im talking about how the data is written to file it has no effect on the UI

Sophist-UK:
@ijabz - I mean that we (Picard) are not automatically going to follow what Jaikoz / Songkong have done if it does not make long-term sense for Picard. If it is no worse than other choices then of course we should try to be compatible.

JuniorJPDJ:
You guys should reconsider this as it's inconsistent in my opinion that `artist` has it's plural veresion and `albumartist` doesn't when we already have mbid for both anyway.
Making this tag sort of standard is cool idea and makes music library handling much easier in software.

Example is funkwhale which now has pretty big problems with handling multiple artists and big mess which could probably be handled more easily when decisions like this would be done long time ago: [1](https://dev.funkwhale.audio/funkwhale/funkwhale/-/merge_requests/1279) [2](https://dev.funkwhale.audio/funkwhale/funkwhale/-/issues/1170)
Also Kodi already adapted and uses this tag to make library browsable: https://kodi.wiki/view/Music_tagging#albumartists
There's a need for this tag!

PR #555, opened 2017-01-13
PICARD-860: Implement Cover-Art plugins

samj1912:


samj1912:
Related test plugins - https://gist.github.com/samj1912/82122aa4f13369b9629abdc41baa39cb#file-test_coverart-py

picard/coverart/plugins.py
```diff
@@ -0,0 +1,88 @@
...
+    try:
+        m = md5()
+        m.update(coverartimage.original_data)
+        for option in options:
+            m.update("{}:{}".format(option.name, config.setting[option.name]))
+        if extra_settings:
+            m.update(json.dumps(extra_settings))
+        m.update(action_id.hex)
+        process_hash = m.hexdigest()
+        print process_hash
```
zas: Debug print to remove, or use `log.debug()`

picard/coverart/image.py
```diff
@@ -216,6 +219,8 @@ def set_data(self, data):
 
         try:
             self.datahash = DataHash(data, suffix=self.extension)
+            if not self.original_datahash:
+                self.original_datahash = DataHash(data, suffix=self.extension)
```
zas: Hmmm actually a bad idea, it will recalculate the same stuff for nothing.
See https://github.com/metabrainz/picard/blob/master/picard/coverart/image.py#L56

Imho`original_data` and its hash should only be set and calculated when needed, which is on first processor run.

Imho we should limit to the bare minumum:
- if a processor has to run on the image
- test if image's `original_datahash` is None, set it to current hash
- copy the Datahash object (copying hash & linked file)
- run processor and use set_data() to update the copied object (if it isn't modified it will not create a new file)
- store `original_datahash` in new object

Basically we have cases where processors will do nothing, we shouldn't create more objects or files in this case.

Instead of providing test plugins as gist, create actual tests. Each hook should use a different processing (ie. different sizes).

picard/coverart/plugins.py
samj1912: Ah sorry, left overs from testing.

samj1912:
Did not include tests since that will require PIL in travis.

samj1912:
Hmm, I just realized a very glaring problem in my implementation.

Since we are using set_data, we are actually changing the data for the coverartimage file all across, whether its file save action or tag save action.

And I tested the above implementation properly, although it is working without any threading issues if we are using the download_action, for the other 2 it is actually creating problems.

So I think we need to resolve that. 

I will look into reimplementing the tag and file save actions, but as for download actions, the above is working well including the caching.

samj1912:
Although I managed to fix the problem and seperate the actions, I dont think my implementation is the most efficient. 

Suggestions to improve the set_data problem appreciated :)

picard/coverart/plugins.py
```diff
@@ -0,0 +1,89 @@
...
+
+
+def is_cached(action_id, coverartimage, options, extra_settings):
+    _cache_mutex.lock()
+    result = False
+    try:
+        m = md5()
+        if not coverartimage.original_datahash:
+            coverartimage.original_datahash = coverartimage.datahash
+        m.update(coverartimage.original_data)
```
zas: Every DataHash has already a hash of the data, you should be able to skip this one. Use the already existing hash as part of your new md5 (instead of the full data).
BTW, the whole point of DataHash for images was to limit duplication of the data and having a hash for the data for comparaisons.

I'll test this code and see if i can find a way to improve.

samj1912: I was thinking this and a solution for that deepcopy everytime is that we instead have attributes file_data and tag_data for CoverArtImage which are initialized to none and when we are saving coverart images to file/tags if they are set to none we can just use the original data and if they are not we can use them. Also we can provide a wrapper to set_data which takes care of setting data accordingly to these attributes when used by plugins.

samj1912: Thoughts?

samj1912:
Closing this for now.

PR #557, opened 2017-01-14
PICARD-897: Rename "Submit" button to "Submit AcoustIDs".

Freso:
The submit button doesn't actually submit any information to MusicBrainz, but instead submits acoustic fingerprint information to AcoustID.org. The title of this button comes up over and over as a point of confusion for new users of Picard, so renaming it to be more precise will hopefully alleviate this point of confusion.

Jira ticket: https://tickets.metabrainz.org/browse/PICARD-897

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -360,7 +360,7 @@ class MainWindow(QtGui.QMainWindow):
         self.save_action.setEnabled(False)
         self.save_action.triggered.connect(self.save)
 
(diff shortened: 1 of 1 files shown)


mineo:
I'd like to get @lalinsky's opinion on this. IIRC the submit button does not submit AcoustIDs, but only associates acoustic fingerprints with a bunch of metadata. The AcoustIDs are then generated on the AcoustID server.

lalinsky:
@mineo I think that's a technical detail that most people don't need to know. I'm fine with the change.

PR #559, opened 2017-01-14
PICARD-898: Change default id3v24 encoding to utf8

samj1912:
https://tickets.metabrainz.org/browse/PICARD-898

### DIFF
--- a/picard/ui/options/tags.py
+++ b/picard/ui/options/tags.py
@@ -106,6 +106,7 @@ class TagsOptionsPage(OptionsPage):
             self.ui.id3v23_join_with.setEnabled(True)
         else:
             self.ui.enc_utf8.setEnabled(True)
+            self.ui.enc_utf8.setChecked(True)
             self.ui.label_id3v23_join_with.setEnabled(False)
             self.ui.id3v23_join_with.setEnabled(False)
 


Sophist-UK:
I know it seems like we should default to the latest version, HOWEVER the general consensus is that v23 is better and many media players only support v23 most notably Windows.

What is YOUR rationale for wanting to set the default to  v24 and UTF-8?

samj1912:
A lot of our tests revolve around v2.4 - see test_formats and also there are a lot of frames which are absent  in 2.3 that we claim we support. Basically the TS* (sort order) series. Although compatid3 was added to provide them for id3tags there are no existing hooks for AIFF files thus some of the tags are broken (basically all the tags in compatid3)

However, looking at the fact that Windows currently just supports id3v23 we can decide on what to do.

I will update the PR to remove id3v24 as the default version, but I think the rest of the 2 tickets are still valid.

Sophist-UK:
Thank you.

Do you have any evidence that you can add here and / or Jira to support the switch to UTF-8 as default for ID3v24?

Sophist-UK:
I think you need code to set UTF-8 as default if we switch from v23 to v24, and possible code if you switch from v24 to v23.

samj1912:
@Sophist-UK already done? My proposed changes change the encoding to utf8 if we switch from id3v23 and to utf16 if we switch back. These changes are proposed in https://tickets.metabrainz.org/browse/PICARD-898

Sophist-UK:
Haven't tested this myself, but otherwise its got my vote.

jistme:
This might be of interest:
http://getmusicbee.com/forum/index.php?topic=20536.0

If there is any group of users who take tagging seriously, it will be MusicBee users.
While I would personally prefer 2.4 as default, I must admit the tides are still a bit against it considering some lack of support, understanding, awareness.

An idea comes to mind to perhaps help to improve those three, the Picard installer could present the option if a user wants to start off using ID3v2.4, explaining shortly what the advantages and possible disadvantages are.

Sophist-UK:
I like the idea - but perhaps as part of the first use wizard that we are talking about.

zas:
> https://tickets.metabrainz.org/browse/PICARD-900

Can you split this PR in two, one for each issue (898 and 902) ?

samj1912:
Sure

samj1912:
@zas https://github.com/metabrainz/picard/pull/579

PR #561, opened 2017-01-14
PICARD-357: Fix the way Picard handles filenames with only extensions

samj1912:
https://tickets.metabrainz.org/browse/PICARD-357

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -290,14 +290,11 @@ class File(QtCore.QObject, Item):
 
         if settings["rename_files"]:
             new_filename, ext = os.path.splitext(new_filename)
-            try:
-                # In case the filename is blank and only has the extension
-                # the real extension is in new_filename and ext is blank
-                if ext == '' and new_filename.lower() in map(unicode, self.EXTENSIONS):
-                    ext = new_filename
-                    new_filename = ''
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -284,6 +284,14 @@ def _make_filename(self, filename, metadata, settings=config.setting):
 
         if settings["rename_files"]:
             new_filename, ext = os.path.splitext(new_filename)
+            try:
+                # In case the filename is blank and only has the extension
+                # the real extension is in new_filename and ext is blank
+                if ext == '' and new_filename.lower() in map(unicode, self.EXTENSIONS):
+                    ext = new_filename
+                    new_filename = ''
+            except AttributeError:
```
mineo: Under which circumstances can the lines above raise an AttributeError?

samj1912: In case extensions is not defined

mineo: `register_formats` and `supported_formats` already don't work if `EXTENSIONS` is not a defined attribute.

samj1912:
Updated :)

zas:
@samj1912 : can you update tests so this case is tested ?

samj1912:
sure :)

PR #562, opened 2017-01-15
PICARD-906:Use globalInstance of threadpool as default

samj1912:


samj1912:
@mwiencek I think this was a bug in pyqt? It is working fine for me now.

Tagging since you made this TO-DO

mwiencek:
I don't remember the specifics of that comment, but if the `start()` method was broken, then `run_task` would also be completely broken, and I guess you'd notice right away that things didn't work. So maybe it was a Qt bug (hopefully not macOS-specific, since I can't run Picard at all on 10.12 right now to test anything).

mineo:
Unless someone wants to test that this works on all PyQt versions that we claim to support, I vote for not merging this. If we bump the requirements at some point and can guarantee that it works on all platforms, this change is small enough to reapply.

zas:
https://github.com/metabrainz/picard/pull/562#issuecomment-272922587

PR #565, opened 2017-01-16
PICARD-911: Messagebox wraps and displays title inappropriately

suhas2go:
Informative text wraps based on size of text

Before
before

After
after


suhas2go:
Here's the [jira ticket](https://tickets.metabrainz.org/browse/PICARD-911)

PR #491, opened 2016-11-18
PICARD-408:  Tree selector in Options window is partially obscured, pane too narrow

emhazen:
Changed default size of the options window so that the left tree fits.

PR #566, opened 2017-01-16
PICARD-408: Options pane is obscured

suhas2go:
I know that another [PR](https://github.com/metabrainz/picard/pull/491) already exists for this [issue](https://tickets.metabrainz.org/browse/PICARD-408), I just wanted to put forward this approach. Rather than simply increasing the default options dialog size, I have set the minimum width according to the contents of the pages_tree. 
This allows the splitter to nicely snap at the perfect pane size.
reszied_pane
Note that I have expanded all the items to show that it fits, they're not expanded by default.

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -103,6 +103,9 @@ class OptionsDialog(PicardDialog):
         self.default_item = None
         self.add_pages(None, default_page, self.ui.pages_tree)
 
+        max_top_page_name = self.ui.pages_tree.sizeHintForColumn(0) + 2*self.ui.pages_tree.frameWidth()
+        self.ui.pages_tree.setMinimumWidth(max_top_page_name)
+
         self.ui.pages_tree.setHeaderLabels([""])
         self.ui.pages_tree.header().hide()
         self.ui.pages_tree.itemSelectionChanged.connect(self.switch_page)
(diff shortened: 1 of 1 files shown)


zas:
Imho a much better approach than https://github.com/metabrainz/picard/pull/491 :+1:

suhas2go:
@zas At the default size, resizing is not possible. On increasing the size of the options dialog the pane also grows, the splitter allows to reduce the size of the option pane up to the minimum.

zas:
> @zas At the default size, resizing is not possible. On increasing the size of the options dialog the pane also grows, the splitter allows to reduce the size of the option pane up to the minimum.

Yup, i still have problems with Options dialog:
- i can't reduce size
- i can increase the size, but the pane on the left can only be downized to nothing or default, and increased a bit or to max dialog width, no smooth resize
- the behavior is somehow unpredicable :(

That said this stuff is plagued with such issues since years, not sure the actual problem is in this PR.

suhas2go:
@zas I'll work on the issue of resizing in the Options Dialog when I get time. Is there a corresponding ticket?

zas:
> @zas I'll work on the issue of resizing in the Options Dialog when I get time. Is there a corresponding ticket?
https://tickets.metabrainz.org/browse/PICARD-930

PR #577, opened 2017-01-18
Do not translate widget names

phw:
Do we need a ticket for this? It is basically just a fixup for the builtin search.

Speaking of that: I could not find a ticket for the builtin search, but I guess we should have one for the release notes, right?


zas:
> Do we need a ticket for this? It is basically just a fixup for the builtin search.
> 
> Speaking of that: I could not find a ticket for the builtin search, but I guess we should have one for the release notes, right?

You're right, I created missing tickets:
https://tickets.metabrainz.org/browse/PICARD-927
https://tickets.metabrainz.org/browse/PICARD-928
https://tickets.metabrainz.org/browse/PICARD-929

No need to create one for this fix.

PR #580, opened 2017-01-18
PICARD-923: Add startswith and endswith script functions

samj1912:
https://tickets.metabrainz.org/browse/PICARD-923

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -684,14 +684,14 @@ def func_firstwords(parser, text, length):
         return text[:length].rsplit(' ', 1)[0]
 
 
-def func_startswith(parser, text, pattern):
-    if text.startswith(pattern):
+def func_startswith(parser, text, prefix):
+    if text.startswith(prefix):
         return "1"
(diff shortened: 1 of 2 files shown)


picard/script.py
```diff
@@ -684,6 +684,18 @@ def func_firstwords(parser, text, length):
         return text[:length].rsplit(' ', 1)[0]
 
 
+def func_startswith(parser, text, pattern):
```
zas: Instead of `pattern`, use `prefix`.
Because it isn't a pattern.

picard/script.py
```diff
@@ -684,6 +684,18 @@ def func_firstwords(parser, text, length):
...
+    if text.startswith(pattern):
+        return "1"
+    return "0"
+
+
+def func_endswith(parser, text, pattern):
```
zas: Instead of `pattern`, use `suffix`.
Because it isn't a pattern.

samj1912:
Done :)

PR #583, opened 2017-01-20
Add mimetypes in py2exe module includes

samj1912:


mineo:
For commits like this, including a reason in the commit message would be
great.

PR #586, opened 2017-01-22
PICARD-258: Visual Feedback for coverart changes 

samj1912:
Rename branch of #585 

http://tickets.musicbrainz.org/browse/PICARD-258

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -84,6 +84,12 @@ class CoverArtThumbnail(ActiveLabel):
         self.clicked.connect(self.open_release_page)
         self.imageDropped.connect(self.fetch_remote_image)
 
+    def __eq__(self, other):
+        if self.data and other.data:
+            return self.data.data == other.data.data
+        else:
+            return False
+
     def show(self):
         self.set_data(self.data, True)
 
@@ -167,7 +173,9 @@ class CoverArtBox(QtGui.QGroupBox):
         self.setLayout(self.layout)
 
     def _show(self):
-        if self.cover_art.data == self.orig_cover_art.data:
+        # We want to show the 2 coverarts only if they are different
+        # and orig_cover_art is not None
+        if getattr(self.orig_cover_art, 'data', None) is None or self.cover_art == self.orig_cover_art:
             self.orig_cover_art.setHidden(True)
             self.cover_art_label.setText('')
             self.orig_cover_art_label.setText('')
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -181,8 +181,8 @@ class CoverArtBox(QtGui.QGroupBox):
             self.orig_cover_art_label.setText('')
         else:
             self.orig_cover_art.setHidden(False)
-            self.cover_art_label.setText('New Cover-Art')
-            self.orig_cover_art_label.setText('Original Cover-Art')
+            self.cover_art_label.setText(_(u'New Cover-Art'))
+            self.orig_cover_art_label.setText(_(u'Original Cover-Art'))
 
     def show(self):
         self.cover_art.show()
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -83,12 +83,10 @@ class CoverArtThumbnail(ActiveLabel):
         self.setAlignment(QtCore.Qt.AlignTop | QtCore.Qt.AlignHCenter)
         self.clicked.connect(self.open_release_page)
         self.imageDropped.connect(self.fetch_remote_image)
+        self.related_images = list()
 
     def __eq__(self, other):
-        if self.data and other.data:
-            return self.data.datahash == other.data.datahash
-        else:
-            return False
+        return self.related_images == other.related_images
 
     def show(self):
         self.set_data(self.data, True)
@@ -121,9 +119,8 @@ class CoverArtThumbnail(ActiveLabel):
 
     def set_metadata(self, metadata):
         data = None
-        # Check if metadata has any images. It might have images even if
-        # metadata's dictionary is None
-        if getattr(metadata, 'images', None):
+        if metadata and metadata.images:
+            self.related_images = metadata.images
             for image in metadata.images:
                 if image.is_front_image():
                     data = image
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -86,7 +86,10 @@ class CoverArtThumbnail(ActiveLabel):
         self.related_images = list()
 
     def __eq__(self, other):
-        return self.related_images == other.related_images
+        if len(self.related_images) or len(other.related_images):
+            return self.related_images == other.related_images
+        else:
+            return True
 
     def show(self):
         self.set_data(self.data, True)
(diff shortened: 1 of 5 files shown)


samj1912:
When no new coverart or new coverart has same data as old or after save

image

when different cover arts
image

picard/ui/coverartbox.py
```diff
@@ -130,27 +125,76 @@ def set_metadata(self, metadata, item):
...
+        # We want to show the 2 coverarts only if they are different
+        # and orig_cover_art is not None
+        if getattr(self.orig_cover_art, 'data', None) is None or self.cover_art == self.orig_cover_art:
+            self.orig_cover_art.setHidden(True)
+            self.cover_art_label.setText('')
+            self.orig_cover_art_label.setText('')
+        else:
+            self.orig_cover_art.setHidden(False)
+            self.cover_art_label.setText('New Cover-Art')
+            self.orig_cover_art_label.setText('Original Cover-Art')
```
zas: `_(u'string')` ?

samj1912: Ah yes!

samj1912:
Updated :)
New preview

image

picard/ui/coverartbox.py
```diff
@@ -130,27 +125,88 @@ def set_metadata(self, metadata, item):
...
+        self.setLayout(self.layout)
+        self.view_changes_button.clicked.connect(self.show_cover_art_info)
+
+    def show_cover_art_info(self):
+        self.parent.view_info(default_tab=1)
+
+    def _show(self):
+        # We want to show the 2 coverarts only if they are different
+        # and orig_cover_art is not None
+        if getattr(self.orig_cover_art, 'data', None) is None or self.cover_art == self.orig_cover_art:
```
samj1912: orig_cover_art doesn't have a data attribute when the default CD shadow is shown 
image

If I were to remove that check if data is None, the code shows the new coverart image and a blank CD shadow( in this case, IMO it should not show the blank CD shadow indicating that there was no coverart previously)

samj1912:
I can't show the diff on an album level since there might be a case that related files have different CAs.

zas:
> I think this is confusing, because cover art is a property of the release in MB

Yes, that's an issue, but Picard is managing files, and cover art, when it comes to tagging, is also a property of tracks.
This is something that always bothered me, but it is complicated to list changes related to a release, when there is no such release on file system, but a bunch of files.
In fact, the only cover art linked to release is not stored in tags, but, optionnally, as a file (cover.jpg or the like in the release directory if files are grouped by release).

Still we should show upcoming changes for all stuff that may be saved/modified, for user convenience.
Ideas (and code) are welcome to improve the situation.

picard/ui/coverartbox.py
mineo: I wasn't suggesting removing the check altogether, just replacing the `getattr` call with `self.orig_cover_art.data` directly. AFAICT, `orig_cover_art` is always an instance of `CoverArtThumbnail` that always initializes `data` to `None` in its initializer.

mineo:
> I can't show the diff on an album level since there might be a case that related files have different CAs.

But in the majority of all cases, this will never happen, because not a single cover art provider allows setting per-track cover art.

samj1912:
The last commit adds album level diffs, if all the related files for all the album tracks have the same set of images.

It causes a slight slow-down in performance while updating though. I will revert the last commit if it is not worth the performance loss.

picard/ui/coverartbox.py
```diff
@@ -201,8 +200,21 @@ def show(self):
             self._show()
         QtGui.QGroupBox.show(self)
 
+    def _extract_common_metadata(self, item):
+        related_files = [track.linked_files[0] for track in item.tracks
+                         if track.num_linked_files == 1]
+        image_hashes_list = [[image.datahash for image in related_file.orig_metadata.images]
```
zas: Using a list of sets would be prolly faster in the `all()` equality test. Also it would prevent order of elements to have any impact (not sure if we have any guarantee about order of images in orig_metadata.images, but we don't care order here anyway).
What do you think ?

samj1912:
Removed the album level diffs. Working with @antlarr on https://github.com/antlarr/picard/tree/samj1912-picard258 He has a better implementation in mind so we will go with that :).

As for the scope of this PR, it will just serve to provide track level diffs and address @mineo's comments which are not related to album level covert art diffs. 

The rest will be continued in the branch linked above and a PR can be made once this PR is merged.

samj1912:
Addressed 1 and 4. 2 requires a complete refactoring of the info dialog to show image property diffs ( it is implemented in a weird way currently).

As for 3, currently I am only showing diff for files/tracks, images saved to separate files instead of mp3 files themselves need to handled in a different way. (Also since they are more album level than file/track level can be handled with @antlarr s album/cluster level img diff PR)

zas:
@samj1912 : please fix conflicts
@mineo : are your points addressed by latest changes ?

samj1912:
Rebased.

samj1912:
@mineo @zas fixed. Can we merge this PR so @antlarr can continue from here?

he has a better implementation for Album and drag n drop related img diffs and changes.

mineo:
I really would prefer not merging things to master just to have a base for another pull request (github even allows opening pull requests against branches in forks, I think). On the other hand can I not guarantee too much time for reviewing things over the next two weeks, so I'll dismiss my review.

But please be aware that their are still several things wrong with this.

1. Unfortunately I don't know how to reproduce this. I did a lot of loading files, saving, switching cover art settings and releases etc., but I now get multiple entries like

```python
  File "./picard/ui/coverartbox.py", line 185, in show_cover_art_info
    self.parent.view_info(default_tab=1)
  File "./picard/ui/mainwindow.py", line 830, in view_info
    if isinstance(self.selected_objects[0], Album):
```

in the console I've started Picard from. The "Show more details" button is also displayed all the time now. Clicking it only makes more of the above errors appear.

2. Set your settings to load only front images from the CAA. Load a release that has multiple images in the CAA and wait for the front cover art to load. Do not save. Notice that no cover art changes are indicated. Select more CAA types. Reload/Refresh the release. The "Show more details" button now appears.

3. Load a release that has CAA images and at least 2 tracks, save it, change the cover art size, reload the release. Select the release item. Select any track of the release. "New Cover Art" and "Original Cover Art" appear, but no "Show more details" button. Select another track, button appears. This doesn't happen if I make the metadata box larger (350px seem to be necessary).

4. I am pretty sure that I need to click "Save" twice with these changes to actually save a release. I did try to the same release on both master and with this. This happens with my config and even if I remove that.

picard/ui/mainwindow.py
```diff
@@ -836,6 +836,10 @@ def view_info(self):
         else:
             file = self.tagger.get_files_from_objects(self.selected_objects)[0]
             dialog = FileInfoDialog(file, self)
+        try:
+            dialog.ui.tabWidget.setCurrentIndex(default_tab)
+        except IndexError:
```
mineo: Under which circumstances does the `setCurrentIndex` call above raise an `IndexError`? It doesn't do it for 50000 and -5.

samj1912:
@mineo, I'd like to test some more. The errors are way too much to be merged. 

Also, as for 2,show more details button is now shown when new cover art images have more than 1 image.

samj1912:
@mineo all issues fixed. Show more details button is supposed to be shown when there is more than 1 CA image.

PR #589, opened 2017-01-23
PICARD-944: Wait for save thread pool to be finished before exit

samj1912:
https://tickets.metabrainz.org/browse/PICARD-944

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -274,6 +274,7 @@ class Tagger(QtGui.QApplication):
         self.stopping = True
         self._acoustid.done()
         self.thread_pool.waitForDone()
+        self.save_thread_pool.waitForDone()
         self.browser_integration.stop()
         self.xmlws.stop()
         for f in self.exit_cleanup:


picard/formats/__init__.py
```diff
@@ -84,107 +84,6 @@ def open(filename):
         return None
 
 
-def _insert_bytes_no_mmap(fobj, size, offset, BUFFER_SIZE=2**16):
```
mineo: Why?

samj1912: https://tickets.metabrainz.org/browse/PICARD-933

Removed it as per the discussion. Can undo it if required :)

zas: Imho, this removal should go in another PR. Perhaps i miss something but it doesn't look it is directly related to PICARD-933.

@samj1912 : can you create another ticket and another PR for the monkey patching removal ?

picard/tagger.py
```diff
@@ -274,6 +274,7 @@ def exit(self):
         self.stopping = True
         self._acoustid.done()
         self.thread_pool.waitForDone()
+        self.save_thread_pool.waitForDone()
```
zas: I did a test, inserting a 60s sleep in `file._save_and_rename()` with and without this line.

I load a file, tag it, press Save and exit Picard.
Without this line, the file isn't written to new destination (i put the sleep at start of function), Picard exits totally.

With this line, the file is written to new destination, Picard UI is closed immediatly on quit action, but it stays as a processus (without UI), until the delay expires.

In the first case, i'm not sure when and where the save process is stopped. In the second, we are pretty sure it goes to the end.

I'm worried though about possibly stuck save threads (and therefore stuck Picard processes), so i went the evil way, i ran one Picard, save a file (with sleep), quit, and run another Picard, load/save the same file (possible, sleep is at the start, file isn't moved yet), and when the save finally occurs i get an exception in `info = os.stat(encoded_old_filename)` file not found of course, which is expected.

So :+1:

Sophist-UK: I assume that if you have 20 files queued up it doesn't start any new saves from the queue before quitting.

Sophist-UK:
P.S. Anyone know how many threads can be used for loading (reading directories, loading existing metadata) and saving.

picard/tagger.py
zas: > I assume that if you have 20 files queued up it doesn't start any new saves from the queue before quitting.

Nope, but queued ones will finish.

samj1912:
@Sophist-UK for saving the thread pool is set to 1, for other things, it depends on your system. On mine it is 4.

Sophist-UK:
Hmm ... nothing to do with this PR, but we should reserve one thread for UI.

PR #590, opened 2017-01-23
Bump minimum Python version to 2.7.9 for Picard 1.4.

Freso:
Since the various MetaBrainz projects are increasingly moving to HTTPS-mostly (and hopefully -only before too long!), it would be nice to know that Picard 1.4 won't break because of this anytime soon. Since several of the MetaBrainz sites share an IP address they're relying on SNI for sending the proper certificate. SNI support wasn't added in Python 2.7 before the .9 release, so bumping minimum Python version to this specific one, so that we can be sure that SNI is supported.

I feel like Picard 1.4 might be a bit of an "LTS" type release for us, since there will undoubtedly be some people unhappy with whatever changes we make for Picard 2.0, so I feel like we'll want to make sure it'll work for as long as possible—which includes after musicbrainz.org and acousticbrainz.org going full-HTTPS. For that we'll want SNI, and it's easier to just depend on 2.7.9 rather than add another dependency and change code for that, and we probably don't want to touch Picard either at that time.

Ubuntu Trusty (from April 2014) still ships a pre-2.7.9 Python, but are any other distros of relevance still on this old Python? Ubuntu Trusty is 3 years old soon, and not likely used on a lot of desktops. It probably also ships an old Picard version anyway and isn't likely to pick up Picard 1.4 either.

A slight bit of IRC discussion: https://chatlogs.metabrainz.org/brainzbot/metabrainz/msg/3800847/

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -6,7 +6,7 @@ Dependencies
 
 Before installing Picard, you need to have these:
 
- * Python 2.7 or newer (Picard will not work with Python 3)
+ * Python 2.7.9 or newer (Picard will not work with Python 3)
    http://python.org/download
 
  * PyQt 4.5 or newer
--- a/setup.py
+++ b/setup.py
@@ -10,8 +10,8 @@ import sys
 import subprocess
 from picard import __version__, compat
 
-if sys.version_info < (2, 7):
-    print("*** You need Python 2.7 or higher to use Picard.")
+if sys.version_info < (2, 7, 9):
+    print("*** You need Python 2.7.9 or higher to use Picard.")
 
 
 args = {}


zas:
:+1: from me
@phw , @mwiencek , @mineo , @Sophist-UK , @samj1912 : any objection on this ?

Sophist-UK:
As usual @F.reso makes a good case.

Whilst I don't consider my vote to be essential to this decision, I am certainly in favour

mineo:
Unless I'm severely mistaken, this is not necessary because Qt provides SNI support since 2011 (https://bugreports.qt.io/browse/QTBUG-1352). Since 8e0118c6de5ebe3b5a8b1faa53d9638345d03ed0 requests to picard.mbo by the plugins stuff are done over HTTPS, so if the windows nightly builds that use Python 2.7.6 AFAICT work, there's nothing to worry about here.

zas:
> Unless I'm severely mistaken, this is not necessary because Qt provides SNI support since 2011 (https://bugreports.qt.io/browse/QTBUG-1352). Since 8e0118c requests to picard.mbo by the plugins stuff are done over HTTPS, so if the windows nightly builds that use Python 2.7.6 AFAICT work, there's nothing to worry about here.

That's a good point.
If there is no actual issue with 2.7.6 for now and since we'll move to Python3 / Qt5 for next major release, that's probably a useless move.

Thanks @mineo for those infos.

lalinsky:
There will be no Windows or OS X packages built by me if you do this. I might be able to upgrade Python on Windows, but I won't even attempt doing that on OS X.

mineo:
Let's not do this.

PR #597, opened 2017-01-25
Fix OS X package

lalinsky:
It seems that if the ini-based config was initialized before creating the app instance, Qt would do a different initialization sequence and some configs would not be loaded. That caused it to not look for plugins in the app bundle.

@zas @mwiencek Can you please review?

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -92,11 +92,11 @@ class Config(QtCore.QSettings):
 
     """Configuration."""
 
-    def __init__(self):
+    def __init__(self, app):
         """Initializes the configuration."""
 
         QtCore.QSettings.__init__(self, QtCore.QSettings.IniFormat,
-                                  QtCore.QSettings.UserScope, PICARD_ORG_NAME, PICARD_APP_NAME)
+                                  QtCore.QSettings.UserScope, PICARD_ORG_NAME, PICARD_APP_NAME, app)
         # If there are no settings, copy existing settings from old format
         # (registry on windows systems)
         if not self.allKeys():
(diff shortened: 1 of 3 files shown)


mwiencek:
It seems fine as long as it doesn't break invocations the module-level, e.g. [here](https://github.com/metabrainz/picard/blob/9cc69acf25f1b9970e112d3f2807c0305cdfbb22/picard/coverart/providers/caa.py#L134-L139) or [here](https://github.com/metabrainz/picard/blob/9cc69acf25f1b9970e112d3f2807c0305cdfbb22/picard/coverart/providers/local.py#L39-L39). I can't run Picard here to test that, but it's possible those modules aren't loaded until after the options are opened (so config will no longer be None).

zas:
I tested, and the PR looks ok, i will merge it, and if issues arise we'll know soon.

zas:
After this patch, i get such error on save :

```
E: 13:21:15 Traceback (most recent call last):
  File "./picard/util/thread.py", line 46, in run
    result = self.func()
  File "./picard/file.py", line 191, in _save_and_rename
    new_filename = self._rename(old_filename, metadata)
  File "./picard/file.py", line 333, in _rename
    self._make_filename(old_filename, metadata))
  File "./picard/file.py", line 292, in _make_filename
    if settings["move_files"]:
TypeError: 'NoneType' object has no attribute '__getitem__'
```

mwiencek:
There might be more issues than I remember. Those variables were intended to be usable anywhere with https://github.com/metabrainz/picard/pull/117.

mwiencek:
Perhaps it's fixable if the modules don't import `settings` etc. directly, but reference `config.settings` so that the attribute lookup is delayed.

lalinsky:
Sorry about that, I thought I reviewed all cases where this might be imported, but I missed this kind of import. We should change the imports. I'll do that today evening.

zas:
> Sorry about that, I thought I reviewed all cases where this might be imported, but I missed this kind of import. We should change the imports. I'll do that today evening.

Thanks ;)

zas:
@mwiencek , @lalinsky : this issue is the last big one to fix before we are able to release, any update ?

PR #598, opened 2017-01-26
PICARD-942: Fix explore/play remote folders/files

Sophist-UK:
Resolves [PICARD-942](https://tickets.metabrainz.org/browse/PICARD-942).

Make "Open Containing Folder" and "Open in Player" work for remote files.

See [QTBUG-13359](https://bugreports.qt.io/browse/QTBUG-13359) for cause and workaround.

Tested on Windows. Needs testing on Mac / Linux.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -785,6 +785,8 @@ class MainWindow(QtGui.QMainWindow):
         return self.tagger.analyze(self.selected_objects)
 
     def _openUrl(self,url):
+        # Resolves a bug in Qt opening remote URLs - QTBUG-13359
+        # See https://bugreports.qt.io/browse/QTBUG-13359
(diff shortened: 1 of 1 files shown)


samj1912:
Tested on Linux. Works :)

picard/ui/mainwindow.py
```diff
@@ -784,17 +784,21 @@ def analyze(self):
                 return
         return self.tagger.analyze(self.selected_objects)
 
+    def _openUrl(self,url):
+        if url.startswith("\\\\") or url.startswith("//"):
```
zas: Please add a comment about the reason of this, it is a workaround for a Qt bug, so it may be removed at some point.

Sophist-UK:
I am using PyQt 4.8.6.

The Qt ticket says it should have been fixed in Qt 4.8 - so obviously this fix has not made it into PyQt.

PR #599, opened 2017-01-26
PICARD-950: Don't open same folder multiple times

Sophist-UK:
Resolves [PICARD-950](https://tickets.metabrainz.org/browse/PICARD-950)

This will need rebasing after #598 is merged.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -804,7 +804,7 @@ class MainWindow(QtGui.QMainWindow):
             url = self._openUrl(os.path.dirname(file.filename))
             if not url in opened:
                 QtGui.QDesktopServices.openUrl(url)
-                opened.append(url)
+                opened.add(url)
 
     def show_analyze_settings_info(self):
         ret = QtGui.QMessageBox.question(self,
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -799,8 +799,12 @@ class MainWindow(QtGui.QMainWindow):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
+        opened = []
         for file in files:
-            QtGui.QDesktopServices.openUrl(self._openUrl(os.path.dirname(file.filename)))
+            url = self._openUrl(os.path.dirname(file.filename))
+            if not url in opened:
+                QtGui.QDesktopServices.openUrl(url)
+                opened.append(url)
 
     def show_analyze_settings_info(self):
         ret = QtGui.QMessageBox.question(self,
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -799,7 +799,7 @@ class MainWindow(QtGui.QMainWindow):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
-        opened = []
+        opened = set()
         for file in files:
             url = self._openUrl(os.path.dirname(file.filename))
             if not url in opened:
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -799,12 +799,9 @@ class MainWindow(QtGui.QMainWindow):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
-        opened = set()
-        for file in files:
-            url = self._openUrl(os.path.dirname(file.filename))
-            if not url in opened:
-                QtGui.QDesktopServices.openUrl(url)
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -792,9 +792,12 @@ def play_file(self):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
+        opened = []
```
zas: You should use a `set()` here, we don't care about order, elements are unique by definition, and search is done in O(1) vs O(n).

Sophist-UK: Given that use generally won't select more than a few files - and we are avoiding attempting to open the same directory several times which will take much more CPU, the performance impact of a list will be negligible. 

However this would nonetheless be better code so I will change it. 👍

picard/ui/mainwindow.py
```diff
@@ -792,9 +792,12 @@ def play_file(self):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
+        opened = set()
         for file in files:
             url = QtCore.QUrl.fromLocalFile(os.path.dirname(file.filename))
-            QtGui.QDesktopServices.openUrl(url)
+            if not url in opened:
+                QtGui.QDesktopServices.openUrl(url)
+                opened.append(url)
```
zas: s/append/add/
https://docs.python.org/2/library/sets.html

Sophist-UK:
Rebased - should be ready to merge.

picard/ui/mainwindow.py
```diff
@@ -799,8 +799,12 @@ def play_file(self):
 
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
+        opened = set()
         for file in files:
-            QtGui.QDesktopServices.openUrl(self._openUrl(os.path.dirname(file.filename)))
+            url = self._openUrl(os.path.dirname(file.filename))
+            if not url in opened:
+                QtGui.QDesktopServices.openUrl(url)
+                opened.add(url)
```
zas: Did you actually test this code ?

This doesn't work, because `url` is an object, so each set element is considered different.

I propose the following code to solve this issue:

```python
    def open_folder(self):
        files = self.tagger.get_files_from_objects(self.selected_objects)
        folders = set([os.path.realpath(os.path.dirname(f.filename)) for f in files])
        for folder in folders:
            QtGui.QDesktopServices.openUrl(self._openUrl(folder))
```

Advantages:
- it works
- it handles symlinked directories case
- it calls self._openUrl() once per dir

Sophist-UK: Why are you proposing use of os.path,.realpath? Do you have any examples of why this is necessary?

Sophist-UK: Aside from this I like the style of code - very pythonesque.

Sophist-UK:
Tested on both local and remote files and ready for merge.

picard/ui/mainwindow.py
Sophist-UK: Upon consideration, I am wondering whether use of os.path.realpath is actually wrong (as opposed to unnecessarily).

I think @zas thinking was to avoid duplication of two directories which are mapped by symbolic links to the same directory - and at first sight this seems correct.

However I think this has an unintended consequence of displaying the real location rather than the one the user would be expecting. So for example:

* Suppose I have two artist directories mapped to the same directory - for the sake of argument suppose `/Music/Annie Lennox` and `/Music/Dave Stewart` are both symbolic links to `/Music/Eurythmics`.
* And suppose I have an Annie Lennox album loaded in Picard from  `/Music/Annie Lennox` and a Dave Stewart album loaded from `/Music/Dave Stewart`. If I select all the tracks and Open containing Folder I would expect to get two file explorer windows opening, one in each directory, but instead I get one file explorer window opening in `/Music/Eurythmics`.

So the results would not be as I would expect. Indeed it is possible that I have permissions to view the linked from directories but don't have security permissions to view the linked to directory - and I might not get a window opening at all.

So, on the whole I am thinking that it might be better to drop use of `os.path.realpath`.

@zas what do you think?

zas: @Sophist-UK : i thought about it yesterday, and yes, i think now it is a bad move, it may cause more issues than it solves. I will drop the `os.path.realpath` part.

zas: `os.path.realpath()` removed by https://github.com/metabrainz/picard/commit/0180d0cf4111d03e932d66fef3501e0b62cafe56

PR #600, opened 2017-01-26
PICARD-951: Always use HTTPS for musicbrainz.org.

Freso:
https://tickets.metabrainz.org/browse/PICARD-951

MusicBrainz.org is eventually going to go all HTTPS-only, though the webservice for now can still be reached via HTTP. To help the move to HTTPS, this commit makes URLs built via `build_qurl()` always use HTTPS when going to a MusicBrainz server (as defined by the `MUSICBRAINZ_SERVERS` constant).

This also means that the check in the `build_qurl()` function for whether a given page/URL is/requires an authentication for MusicBrainz is now redundant, as we're always using HTTPS on MusicBrainz, regardless of login/authentication status.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -395,9 +395,9 @@ def build_qurl(host, port=80, path=None, queryargs=None):
     url.setHost(host)
     url.setPort(port)
     if (# We're contacting a MusicBrainz server
-        (host in MUSICBRAINZ_SERVERS and port == 80) or
+        host in MUSICBRAINZ_SERVERS or
         # Or we're contacting some other server via HTTPS.
-         port == 443):
+        port == 443):
             url.setScheme("https")
             url.setPort(443)
     else:
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -394,12 +394,10 @@ def build_qurl(host, port=80, path=None, queryargs=None):
     url = QtCore.QUrl()
     url.setHost(host)
     url.setPort(port)
-    if (# We're contacting a MusicBrainz server
-        host in MUSICBRAINZ_SERVERS or
-        # Or we're contacting some other server via HTTPS.
-        port == 443):
-            url.setScheme("https")
-            url.setPort(443)
+
+    if (host in MUSICBRAINZ_SERVERS or port == 443):
+        url.setScheme("https")
+        url.setPort(443)
     else:
         url.setScheme("http")
 
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -383,12 +383,10 @@ def album_artist_from_path(filename, album, artist):
     return album, artist
 
 
-def build_qurl(host, port=80, path=None, mblogin=False, queryargs=None):
+def build_qurl(host, port=80, path=None, queryargs=None):
     """
     Builds and returns a QUrl object from `host`, `port` and `path` and
     automatically enables HTTPS if necessary.
-
-    Setting `mblogin` to True forces HTTPS on MusicBrainz' servers.
 
     Encoded query arguments can be provided in `queryargs`, a
     dictionary mapping field names to values.
@@ -396,8 +394,8 @@ def build_qurl(host, port=80, path=None, mblogin=False, queryargs=None):
     url = QtCore.QUrl()
     url.setHost(host)
     url.setPort(port)
-    if (# Login is required and we're contacting an MB server
-        (mblogin and host in MUSICBRAINZ_SERVERS and port == 80) or
+    if (# We're contacting a MusicBrainz server
+        (host in MUSICBRAINZ_SERVERS and port == 80) or
         # Or we're contacting some other server via HTTPS.
          port == 443):
             url.setScheme("https")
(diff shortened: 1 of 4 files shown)


picard/util/__init__.py
```diff
@@ -383,21 +383,19 @@ def album_artist_from_path(filename, album, artist):
...
     Encoded query arguments can be provided in `queryargs`, a
     dictionary mapping field names to values.
     """
     url = QtCore.QUrl()
     url.setHost(host)
     url.setPort(port)
-    if (# Login is required and we're contacting an MB server
-        (mblogin and host in MUSICBRAINZ_SERVERS and port == 80) or
+    if (# We're contacting a MusicBrainz server
+        (host in MUSICBRAINZ_SERVERS and port == 80) or
```
Freso: Does it make sense to check for `port == 80`? Maybe it should just check for `host in MUSICBRAINZ_SERVERS` to prevent typos etc.?

zas: If we always use https for servers listed in `MUSICBRAINZ_SERVERS`, i see no point to keep `and port == 80`.

Freso: I removed the `port` check with https://github.com/metabrainz/picard/pull/600/commits/e85febfd2357b558b3f83d41e2459b062ff29d46.

The second part of your comment feels like something beyond this PR. Note that `build_qurl()` is used for all requests, not just for {beta.,musicbrainz.org} ones.

Freso:
Lots of issues related to MBH-466 on IRC over night. Maybe this can go into Picard 1.4 instead of waiting for 1.4.1?

zas:
> Lots of issues related to MBH-466 on IRC over night. Maybe this can go into Picard 1.4 instead of waiting for 1.4.1?

Hmmm yes, prolly safer. I'll merge it, so we can have a wider test, i was expecting to release 1.4 on 1st feb, but i may delay it by few days so we are sure all important issues are fixed.

zas:
@mwiencek @mineo : ok to merge this for 1.4 ?

Sophist-UK:
Suggest you change the default port from 80 to 443 in `picard/ui/options/general.py` at line 40.

Sophist-UK:
@zas Are we thinking of a full release of 1.4 or releasing 1.4 as a "beta" release for wider testing?

zas:
> @zas Are we thinking of a full release of 1.4 or releasing 1.4 as a "beta" release for wider testing?

1.4, if we call it beta, people will stay on 1.3.2 and packagers will wait final ... 
But i expect to release 1.4.1 soon after, because i'm pretty sure we'll have tons of new bug reports.

1.4 dev is quite stable (after all we all use it since... years)

zas:
> Suggest you change the default port from 80 to 443 in picard/ui/options/general.py at line 40.

:+1: 

@Freso : can you do it ? https://github.com/metabrainz/picard/blob/master/picard/ui/options/general.py#L40

Sophist-UK:
I am running 1.4 from source for a complete re-tag and it is pretty stable - a few oddities which are so odd that I doubt I will ever get even a hint of the cause.

But I agree with your reasoning.

Sophist-UK:
P.S. The oddities are I suspect a consequence of me doing stuff to an album (like refreshing or selecting Other version) whilst file saves are queued. Ideally Picard would cope with users (like me) doing (stupid) things like this - but probably not worth spending the effort fixing them.

Freso:
> Suggest you change the default port from 80 to 443

I did consider this. The reason I opted not to do it is that a not insignificant number of people run their own copy of the MB server locally (or even use one of the couple/handful of unofficial MB mirrors) and those are all likely to stay on port 80 for the foreseeable future. If you change the host to "localhost" it is not unlikely you will forget to change port 443 to port 80.

Sophist-UK:
Yes - but these are advanced users and they will quickly spot the need to change the port.

Let's not forget that the majority of folks will be using Picard as novices, and it will seem odd for PIcard to use port 443 when options says port 80.

Freso:
> these are advanced users

I wouldn't assume that. I've seen people pointing to unofficial mirrors and people just copying that without a second thought. Just because we don't see it often in MetaBrainz land doesn't mean it's not a common thing to happen outside (e.g., in Kodi land). I'd prefer to not touch it and completely revamp how alternative servers are selected for Picard 2.0.

zas:
> Let's not forget that the majority of folks will be using Picard as novices, and it will seem odd for PIcard to use port 443 when options says port 80.

I think this is the main point, Options UI will default to musicbrainz.org / port 80, but in practice, it will use musicbrainz.org / port 443.

People using their own servers have to change the hostname, and eventually the port, i guess that perfectly acceptable.

So, for the sake of consistency, having musicbrainz.org / port 443 when using https proto by default makes more sense imho.

Freso:
Alright then. Changed as per https://github.com/metabrainz/picard/pull/600/commits/256d0ef3ead0f9c7036c9fcbe87cd33977b88ff2

Sophist-UK:
Ok - you've convinced me. So this is ready to merge then?

PR #602, opened 2017-01-26
Update links to ticket tracker.

Freso:
The URL for the ticket tracker has changed, so let's reflect this in the files.

I realise that the command also changed a number of .po files and that these changes will likely get lost, but that it was likely still less effort to just include them anyway rather than try and weed them out of the commit. However, not all .po files have this line and it isn't included in the .pot either, apparently, so not sure exactly where to change the .po file link properly. Maybe on Transifex somewhere.

Using:
`sed -i 's|http://tickets.musicbrainz.org|https://tickets.metabrainz.org|' $(git grep -l tickets.musicbrainz.)`

### DIFF
--- a/README.md
+++ b/README.md
@@ -14,4 +14,4 @@ Binary downloads are available [here](http://picard.musicbrainz.org/downloads/).
 Support and issue reporting
 ---------------------------
 
-Please report all bugs and feature requests in the [MusicBrainz issue tracker](http://tickets.musicbrainz.org/browse/PICARD). If you need support in using Picard please read the [documentation](https://picard.musicbrainz.org/docs/) first and have a look at the [MusicBrainz community forums](https://community.metabrainz.org/c/picard).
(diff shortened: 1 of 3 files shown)


po/af.po
```diff
@@ -7,7 +7,7 @@
 msgid ""
 msgstr ""
 "Project-Id-Version: MusicBrainz\n"
-"Report-Msgid-Bugs-To: http://tickets.musicbrainz.org/\n"
+"Report-Msgid-Bugs-To: https://tickets.metabrainz.org/\n"
```
zas: Not sure about this, we are pulling those files from transifex, the only way i'm aware of is to push .po files, with the (slight) risk to overwrite changes made through transifex.

In the mean time, can you split source code changes (.py) and translation file changes ?

Freso: I already commented on this in the PR description. :) I'd rather not mess with the commit anymore, and if the changes get wiped by the next Transifex pull, then that's fine too.

zas:
> I already commented on this in the PR description. :) I'd rather not mess with the commit anymore, and if the changes get wiped by the next Transifex pull, then that's fine too.

But we don't want unneeded changes in the merge, so please just drop changes concerning .po files.
I'll see later on how to update urls in .po (and we need to add the same url for all, or just drop such urls).

PR #603, opened 2017-01-26
PICARD-953: Show album with extra tracks unmatched

Sophist-UK:
Resolves [PICARD-953](https://tickets.metabrainz.org/browse/PICARD-953).

If an album has unmatched files it is now shown as incomplete even if all album tracks are matched to one file. Although the terminology would suggest that such an album is "complete", in practical terms the album is not perfect because in all likelihood the user needs to select a different release with more tracks.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -485,6 +485,8 @@ class Album(DataObject, Item):
         for track in self.tracks:
             if not track.is_complete():
                 return False
+        if self.get_num_unmatched_files():
+            return False
         else:
             return True
 


zas:
It is changing the behavior on albums with video tracks for example ? (used to be complete when all audio files were there, and ignoring missing video tracks).

Do you fix a regression ? when did it happen ?

Sophist-UK:
This does not change the behaviour when e.g. video tracks are missing - that is handled in the tracks loop above and depending on settings in Options / Advanced. All this does is mark albums with unmatched tracks as imperfect.

As an example, if you have files from Adele 19 Extended Edition which have been mistakenly tagged with the release MBID for the non-extended edition, then it will load CD1 perfectly against the non-extended edition, put CD2 into unmatched files and show the album as perfect. This is clearly not a perfect album.

Sophist-UK:

PR #604, opened 2017-01-27
PICARD-936: Improve UI responsiveness whilst busy loading, saving etc.

Sophist-UK:
Resolves [PICARD-936](https://tickets.metabrainz.org/browse/PICARD-936).

Improves the responsiveness during worker thread operations by triggering main thread to process events after events are sent to it.

Improves file loading performance by moving operations into separate threadpools.

**This PR was brought to you as a joint effort between Sophist-UK and @samj1912.**

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -285,11 +285,13 @@ class Tagger(QtGui.QApplication):
         log.debug("Picard stopping")
         self.stopping = True
         self._acoustid.done()
-        self.thread_pool.waitForDone()
-        self.save_thread_pool.waitForDone()
         self.browser_integration.stop()
         self.xmlws.stop()
+        self.thread_pool.waitForDone()
+        self.load_thread_pool.waitForDone()
+        self.save_thread_pool.waitForDone()
         self.run_cleanup()
+        # Ensure that log messages etc. are processed before quitting
         QtCore.QCoreApplication.processEvents()
 
     def _run_init(self):
@@ -303,7 +305,7 @@ class Tagger(QtGui.QApplication):
                     files.append(decode_filename(file))
             del self._cmdline_files
             if files:
-                self._add_files([files])
+                self.add_files(files)
             if directories:
                 self.add_directory_list(directories)
 
@@ -321,14 +323,101 @@ class Tagger(QtGui.QApplication):
             event.run()
         elif event.type() == QtCore.QEvent.FileOpen:
             f = str(event.file())
-            self._add_files([[f]])
+            self.add_files([f])
             # We should just return True here, except that seems to
             # cause the event's sender to get a -9874 error, so
             # apparently there's some magic inside QFileOpenEvent...
             return 1
         return QtGui.QApplication.event(self, event)
 
+    def move_files(self, files, target):
+        if target is None:
+            log.debug("Aborting move since target is invalid")
+            return
+        if isinstance(target, (Track, Cluster)):
+            for file in files:
+                file.move(target)
+        elif isinstance(target, File):
+            for file in files:
+                file.move(target.parent)
+        elif isinstance(target, Album):
+            self.move_files_to_album(files, album=target)
+        elif isinstance(target, ClusterList):
+            self.cluster(files)
+
+    def add_files(self, filenames, target=None):
+        # Run in a worker thread in case there are large numbers of directories which will make UI non-responsive
+        # Run on main thread_pool rather than file load pool because it is low overhead
+        thread.run_task(partial(self._add_files, [self._add_files_from_directory("", filenames)], target=target),
+                None,
+                thread_pool=self.thread_pool)
+
+    def _add_files(self, list_of_filenames, target=None):
+        if self.stopping:
+            return
+        """Add files to the tagger."""
+        number_of_files = sum([len(x) for x in list_of_filenames])
+        number_of_directories = len(list_of_filenames)
+        self.window.set_statusbar_message(
+            N_("Adding a total of %(files)d files in %(directories)d directories ..."),
+            {'files': number_of_files, 'directories': number_of_directories}
+        )
+        for filenames in list_of_filenames:
+            # Run open_files at lower queuing priority (0) than loading files (1)
+            thread.run_task(partial(self._open_files, filenames),
+                    partial(self._open_files_finished, target),
+                    priority=0,
+                    thread_pool=self.load_thread_pool)
+            # This function runs in a worker thread.
+            # Queuing threads for a large number of directories can make UI unresponsive
+            # Allow UI to process events after each folder to keep it responsive
+            QtCore.QCoreApplication.processEvents()
+
+    def _open_files(self, filenames):
+        if self.stopping:
+            return
+        new_files = {}
+        for filename in filenames:
+            filename = os.path.normpath(os.path.realpath(filename))
+            if filename not in self.files:
+                file = open_file(filename)
+                if file:
+                    new_files[filename] = file
+                # This function runs in a worker thread.
+                # Opening a large number of files in a single directory can make UI unresponsive
+                # Allow UI to process events after each file to keep it responsive
+                QtCore.QCoreApplication.processEvents()
+        return new_files
+
+    def _open_files_finished(self, target=None, result=None, error=None):
+        if self.stopping:
+            return
+        if result:
+            new_files = [result[f] for f in sorted(result)]
+            log.debug("Adding files %r", new_files)
+            self.files.update(result)
+            if target is None or target is self.unmatched_files:
+                self.unmatched_files.add_files(new_files)
+                target = None
+            # Run in a worker thread in case there are large numbers of files which will make UI non-responsive
+            # Run on main thread_pool rather than file load pool because it is low overhead
+            thread.run_task(partial(self._files_load, new_files, target=target),
+                    None,
+                    thread_pool=self.thread_pool)
+        return
+
+    def _files_load(self, files, target=None):
+        if self.stopping:
+            return
+        for file in files:
+            file.load(partial(self._file_loaded, target=target))
+            # Queuing threads for a large number of files in a directory can make UI unresponsive
+            # Allow UI to process events after each file to keep it responsive
+            QtCore.QCoreApplication.processEvents()
+
     def _file_loaded(self, file, target=None):
+        if self.stopping:
+            return
         if file is not None and not file.has_error():
             recordingid = file.metadata.getall('musicbrainz_recordingid')[0] \
                 if 'musicbrainz_recordingid' in file.metadata else ''
@@ -349,81 +438,19 @@ class Tagger(QtGui.QApplication):
             elif config.setting['analyze_new_files'] and file.can_analyze():
                 self.analyze([file])
 
-    def move_files(self, files, target):
-        if target is None:
-            log.debug("Aborting move since target is invalid")
-            return
-        if isinstance(target, (Track, Cluster)):
-            for file in files:
-                file.move(target)
-        elif isinstance(target, File):
-            for file in files:
-                file.move(target.parent)
-        elif isinstance(target, Album):
-            self.move_files_to_album(files, album=target)
-        elif isinstance(target, ClusterList):
-            self.cluster(files)
-
-    def add_files(self, filenames, target=None):
-        self._add_files([self._add_files_from_directory("", filenames)], target=target)
-
-    def _add_files(self, list_of_filenames, target=None):
-        """Add files to the tagger."""
-        number_of_files = sum([len(x) for x in list_of_filenames])
-        number_of_directories = len(list_of_filenames)
-        log.debug("Adding a total of %d files in %d directories",
-                number_of_files, number_of_directories)
-        self.window.set_statusbar_message(
-            N_("Adding a total of %(files)d files in %(directories)d directories ..."),
-            {'files': number_of_files, 'directories': number_of_directories},
-            echo=None
-        )
-        for filenames in list_of_filenames:
-            thread.run_task(partial(self._open_files, filenames),
-                    partial(self._open_files_finished, target),
-                    priority=2,
-                    thread_pool=self.load_thread_pool)
-            # Queuing threads for a large number of directories can make UI unresponsive
-            # Allow UI to process events after each folder to keep it responsive
-            QtCore.QCoreApplication.processEvents()
-
-    def _open_files(self, filenames):
-        new_files = {}
-        for filename in filenames:
-            filename = os.path.normpath(os.path.realpath(filename))
-            if filename not in self.files:
-                file = open_file(filename)
-                if file:
-                    new_files[filename] = file
-                # This function runs in a thread.
-                # Opening a large number of files in a single directory can make UI unresponsive
-                # Allow UI to process events after each file to keep it responsive
-                QtCore.QCoreApplication.processEvents()
-        return new_files
-
-    def _open_files_finished(self, target=None, result=None, error=None):
-        if result:
-            new_files = []
-            for filename in sorted(result):
-                file = result[filename]
-                self.files[filename] = file
-                new_files.append(file)
-            log.debug("Adding files %r", new_files)
-            if target is None or target is self.unmatched_files:
-                self.unmatched_files.add_files(new_files)
-                target = None
-            for file in new_files:
-                file.load(partial(self._file_loaded, target=target))
-                # Queuing threads for a large number of files in a directory can make UI unresponsive
-                # Allow UI to process events after each file to keep it responsive
-                QtCore.QCoreApplication.processEvents()
-
     def _add_directory_method(self):
         if config.setting['recursively_add_files']:
             return self._add_directory_recursive
         return self._add_directory_non_recursive
 
     def add_directory_list(self, dir_list, parent=None):
+        # Run in a worker thread in case there are large numbers of directories which will make UI non-responsive
+        # Run on main thread_pool rather than file load pool because it is low overhead
+        thread.run_task(partial(self._add_directory_list, dir_list, parent),
+                None,
+                thread_pool=self.thread_pool)
+
+    def _add_directory_list(self, dir_list, parent=None):
         if parent:
             parent = os.path.normpath(parent)
         number_of_dirs = len(dir_list)
@@ -434,7 +461,7 @@ class Tagger(QtGui.QApplication):
                     {'directory': parent, 'count': number_of_dirs}
                 )
             else:
-                self.window.set_statusbar_message(
+                self.window.set_statusbar_-message(
                     N_("Adding %(count)d directories ..."),
                     {'directory': parent, 'count': number_of_dirs}
                 )
@@ -456,13 +483,13 @@ class Tagger(QtGui.QApplication):
         list_to_add = []
         # walk topdown to allow modifying the sub-dirs to remove hidden dirs
         for root, dirs, files in os.walk(unicode(path), topdown=True):
-            # Keep UI responsive and ensure that log messages are written to stderr
-            QtCore.QCoreApplication.processEvents()
             if ignore_hidden:
                 dirs[:] = [d for d in dirs if not is_hidden(os.path.join(root, d))]
             files_to_add = self._add_files_from_directory(root, files)
             if files_to_add:
                 list_to_add.append(files_to_add)
+            # Keep UI responsive
+            QtCore.QCoreApplication.processEvents()
         return list_to_add
 
     def _add_directory_non_recursive(self, path):
@@ -491,15 +518,11 @@ class Tagger(QtGui.QApplication):
         if number_of_files:
             if path:
                 path = os.path.normpath(path)
-                log.debug("Adding %d files from %r",
-                          number_of_files, path)
                 message = ungettext(
                         "Adding %(count)d file from '%(directory)s' ...",
                         "Adding %(count)d files from '%(directory)s' ...",
                         number_of_files)
             else:
-                log.debug("Adding %d files" %
-                          number_of_files)
                 message = ungettext(
                         "Adding %(count)d file ...",
                         "Adding %(count)d files ...",
@@ -511,8 +534,7 @@ class Tagger(QtGui.QApplication):
             self.window.set_statusbar_message(
                 message,
                 mparms,
-                translate=None,
-                echo=None
+                translate=None
             )
         return files_to_add
 
@@ -583,6 +605,13 @@ class Tagger(QtGui.QApplication):
     def save(self, objects):
         """Save the specified objects."""
         files = self.get_files_from_objects(objects, save=True)
+        # Run in a worker thread in case there are large numbers of files which will make UI non-responsive
+        # Run on main thread_pool rather than file save pool because it is low overhead
+        thread.run_task(partial(self._save, files),
+                None,
+                thread_pool=self.thread_pool)
+
+    def _save(self, files):
         for file in files:
             file.save()
             # Queuing threads for a large number of files can make UI unresponsive
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -295,14 +295,17 @@ class Tagger(QtGui.QApplication):
     def _run_init(self):
         if self._cmdline_files:
             files = []
+            directories = []
             for file in self._cmdline_files:
                 if os.path.isdir(file):
-                    self.add_directory(decode_filename(file))
+                    directories.append(decode_filename(file))
                 else:
                     files.append(decode_filename(file))
             del self._cmdline_files
             if files:
                 self._add_files([files])
+            if directories:
+                self.add_directory_list(directories)
 
     def run(self):
         if config.setting["browser_integration"]:
@@ -366,9 +369,15 @@ class Tagger(QtGui.QApplication):
 
     def _add_files(self, list_of_filenames, target=None):
         """Add files to the tagger."""
-        # If we want to know the total number of files e.g. to warn the user
-        # sum the lengths of the lists.
-        # number_of_files = sum([len(x) for x in list_of_filenames])
+        number_of_files = sum([len(x) for x in list_of_filenames])
+        number_of_directories = len(list_of_filenames)
+        log.debug("Adding a total of %d files in %d directories",
+                number_of_files, number_of_directories)
+        self.window.set_statusbar_message(
+            N_("Adding a total of %(files)d files in %(directories)d directories ..."),
+            {'files': number_of_files, 'directories': number_of_directories},
+            echo=None
+        )
         for filenames in list_of_filenames:
             thread.run_task(partial(self._open_files, filenames),
                     partial(self._open_files_finished, target),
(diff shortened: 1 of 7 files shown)


Sophist-UK:
This is definitely better, however there are two threading issues that still need fixing:

1. Most importantly, loading the file (file.py#load) is queued on a thread, but the new code (in util/_init_.py#open_file) to quess the file format is not and is therefore blocking. IMO the code in add_files needs refactoring to queue it on a thread.

2. Less importantly, the non-recursive add_directory may need to be queue it on a thread - but I need to get add_files working properly before I can test to see if this is necessary if we are already busy loading large directories recursively.

Any suggestions welcome.

picard/tagger.py
```diff
@@ -416,7 +426,11 @@ def process(result=None, error=None):
             if result:
                 if error is None:
                     self.add_files(result)
-                thread.run_task(get_files, process)
+                thread.run_task(
+                    get_files,
+                    process,
+                    priority=2,
+                    thread_pool=self.load_thread_pool)
```
Sophist-UK: We run this on the load threadpool to avoid it blocking, and run it at priority 2 so that it runs ahead of queued tasks to load metadata from files.

But I wonder whether it might make more sense to have another threadpool just for this.

Sophist-UK: I tried this and it did not seem to make much difference.

Sophist-UK:
Whilst the UI is definitely a bit more responsive with the PR in its current state, the UI is still pretty unresponsive, and I am not sure how to proceed further.

Any help gratefully received.

Sophist-UK:
Ok - I have found the primary cause for UI non-responsiveness.

In `picard/util/thread.py#to_main` When we send a task to the main UI "thread" as an event, Picard does not process that event until the main thread gets activated, and because of the Python LGL stopping real multi-tasking, that doesn't happen until potentially a lot later.

So I have added a `QCoreApplication.processEvents()` immediately after sending the event, in order to give control to the UI thread to process the event, and it makes the whole of Picard much more lively.

That's the good news.

**The bad news is that it exposes other bug(s), most notably when loading a lot of files I get a slew of errors `RuntimeError: maximum recursion depth exceeded`.**

The UI is also too lively, showing updates to objects too fast for the eye to process - so we probably need to do something to only display them every so many miliseconds - and leaving some objects blank,

So there is more work to do, but I am on the right track.

Sophist-UK:
I suggest that we delay further work on #613 until this PR is completed so that we can evaluate whether this resolves the performance issues with large number of files loaded, thus avoiding the need for #613.

Sophist-UK:
Example recursion errors: [picard-recursion-errors.txt](https://github.com/metabrainz/picard/files/750040/picard-recursion-errors.txt)

Sophist-UK:
Ok - I worked out that the cause was calling processEvents when we were on the main thread rather than leaving the main thread to get to it in its own time.

So, basic functionality is now working great. I just need to see what tweaks I need to do to which functions run on which threads and this will be ready for wider testing and merge.

Sophist-UK:
**At the current commit this PR seems stable - any further tweaks will be to threading to make the GUI even more responsive.

So I would welcome anyone else giving this a good test - especially on non-windows systems - and to give feedback about whether it adversely impacts file load or save performance.

Sophist-UK:
P.S. Github is showing the last two commits in the wrong sequence if anyone is wondering about a missing import on the first one.

Sophist-UK:
Remote loads still leave UI unresponsive - need to agree with @samj1912 about borrowing some of his code to try to address this.

samj1912:
An alt approach in #616 I am not sure which one will allow the UI to be more responsive. Can you check my PR's performance once too?

We will go with the better alt. :)

samj1912:
Since its a UI based thing, I just tested with a stopwatch from clicking the load button to when the UI is responsive again. I am running picard on Ubuntu 16.10, on a SSD , ext4.

While loading ~1000 files using your code, it takes picard about ~25 seconds before it is responsive again.
With my PR it is somewhere around ~20 seconds.

I think we can improve on this code further by finding where the UI is being blocked while loading and separating the load code from the main thread as much as possible :)

Sophist-UK:
I am working on the final piece - which I suspect is quite similar to what you have just done - in which case I will borrow your code if I may. I have just reworked the whole file loading code to eliminate duplications and prepare for running open_file in a thread which should make the UI immediately responsive. :-)

samj1912:
Feel free to borrow my code if it helps :) :+1:

Sophist-UK:
Thank you. I did.

The code in #616 was pretty much exactly what I was about to write - so this PR now includes both @samj1912's worker thread code and his Add Directory iteration rewrite - both with some restructuring to handle hidden and excluded files in the right place and also to facilitate a check on number of files to be added easily. See what you think.

**This PR is now ready for code review and testing by others on systems other than windows.**

In the mean while I will see whether this helps with slow-down when you have loaded large numbers of files.

picard/tagger.py
```diff
@@ -345,30 +359,36 @@ def move_files(self, files, target):
...
-                log.info("File ignored (matching %r): %r" % (pattern, filename))
-                continue
             if filename not in self.files:
                 file = open_file(filename)
                 if file:
-                    self.files[filename] = file
-                    new_files.append(file)
-        if new_files:
+                    new_files[filename] = file
+                QtCore.QCoreApplication.processEvents()
```
samj1912: Please add a comment explaining this line.

picard/tagger.py
```diff
@@ -345,30 +359,36 @@ def move_files(self, files, target):
...
+            thread.run_task(partial(self._open_files, filenames),
+                    partial(self._open_files_finished, target),
+                    priority=2,
+                    thread_pool=self.load_thread_pool)
+
+    def _open_files(self, filenames):
+        new_files = {}
         for filename in filenames:
             filename = os.path.normpath(os.path.realpath(filename))
-            if ignore_hidden and is_hidden(filename):
```
samj1912: Why was this code block removed and moved to add directory code? 
What if a user selects a group of files? Ignore regex and hidden file ignores won't work then?

Sophist-UK:
The responsiveness of the UI is still not perfect - for example when loading files and processing webservice responses it can still be sluggish - however I doubt that there is much more that we can do with threading - I suspect that the cause is now the LGL preventing pre-emptive multi-tasking and that, If we can identify where to put them, a few QCoreApplication.processevent() calls will help further, Once this PR has been merged and I have the time I will take a look at some of these code paths and see if I can profile them.

Sophist-UK:
P.S. I have just finished loading 8594 pre-tagged files in 529 releases from a local drive and have not noticed any slowdown in file loading. It is early days for this code, but I am hopeful that the performance problems with large number of files have been significantly reduced.

However, the handles leak still persists - at the end of this load, task manager reports Python as having 58,708 open handles!!!! To put this in context, after Picard has initialised completely, Python shows 755 handles.

Sophist-UK:
P.S. I am going on holiday next week for 2 weeks - so if you want code changes before I go away and before the release of 1.4 on 14 Feb, please let me have them in the next day or two at the latest.

picard/tagger.py
```diff
@@ -287,9 +301,9 @@ def _run_init(self):
                     self.add_directory(decode_filename(file))
                 else:
                     files.append(decode_filename(file))
-            if files:
-                self.add_files(files)
             del self._cmdline_files
+            if files:
+                self.add_files([files])
```
zas: Isn't `files` already a list here ?

zas:
> P.S. I am going on holiday next week for 2 weeks - so if you want code changes before I go away and before the release of 1.4 on 14 Feb, please let me have them in the next day or two at the latest.

Such PR will not be merged in 1.4 but rather in 1.4.1

picard/tagger.py
Sophist-UK: I must have missed that. But it is an easy one to fix.

picard/tagger.py
Sophist-UK: add_files now takes a list of lists of filenames. We group files in each directory into a list and then create a list of these - so we can queue these for guess_format and loading on a directory by directory basis so the user does not have to wait until we have guess_format on every file they have selected to load before they start appearing in the UI.

Sophist-UK: P.S. About to rename tagger.py#add_files to _add_files and create a stub add_files to handle the Add Files dialog which @samj1912 picked up on.

Sophist-UK:
> Such PR will not be merged in 1.4 but rather in 1.4.1

No problem.

picard/tagger.py
zas: I don't get why you need to pass list of lists of filenames, but i didn't check the code in details yet. I'm worried about such change concerning functions like `tagger.add_files()` that may be used in plugins (we really need to clarify plugins API in future versions).

picard/util/thread.py
```diff
@@ -36,12 +36,15 @@ def run(self):
 
 class Runnable(QRunnable):
 
-    def __init__(self, func, next):
+    def __init__(self, func, next, os_priority=0):
```
zas: What is `os_priority` ?

picard/util/thread.py
```diff
@@ -36,12 +36,15 @@ def run(self):
...
         QRunnable.__init__(self)
         self.func = func
         self.next = next
+        self.thread_priority = min(QThread.IdlePriority,max(QThread.TimeCriticalPriority,
```
zas: Explain in a comment please.

picard/util/thread.py
```diff
@@ -52,12 +55,14 @@ def run(self):
...
     if thread_pool is None:
         thread_pool = QCoreApplication.instance().thread_pool
-    thread_pool.start(Runnable(func, next), priority)
+    thread_pool.start(Runnable(func, next, os_priority=os_priority), priority)
 
 
 def to_main(func, *args, **kwargs):
     QCoreApplication.postEvent(QCoreApplication.instance(),
                                ProxyToMainEvent(func, *args, **kwargs))
+    if QCoreApplication.instance().thread() != QThread.currentThread():
```
zas: Explain in a comment please.

picard/tagger.py
```diff
@@ -377,56 +397,47 @@ def add_files(self, filenames, target=None):
...
-            if result:
-                if error is None:
-                    self.add_files(result)
-                thread.run_task(get_files, process)
-
-        process(True, False)
+        list_to_add = []
+        # walk topdown to allow modifying the sub-dirs to remove hidden dirs
+        for root, dirs, files in os.walk(unicode(path), topdown=True):
+            QtCore.QCoreApplication.processEvents()
```
zas: This line deserves a comment on its role.

picard/util/thread.py
Sophist-UK: I accept that this is extra to the functionality of this PR. The idea is that it adjusts the **operating system priority** of the thread down (or up) in order to influence the way threads are dispatched by Python. But now I understand more about how the Python GIL works and impacts which thread gets the GIL (which is essentially random), I am not sure that it would be of any use. (And we don't know what else is running on the user's system and whether it would make sense to do it.)

So I will be happy to revert this Commit.

picard/util/thread.py
Sophist-UK: I will if we don't decide to revert this functionality.

picard/util/thread.py
Sophist-UK: Ok.

picard/tagger.py
Sophist-UK: OK

picard/tagger.py
Sophist-UK: This has been fixed. I renamed this to _add_files and created a stub `tagger.add_files()` which provides the same interface and functionality as before.

samj1912:
I think it will be better to remove the OS priority commits in this pr and make a separate pr for them. 

Also since you are rebasing would it be possible to cherry pick my commits to this PR so that the commits are kept. 

It will be nice if you could add me to collaborators for this pr, I might have a few changes I would like to make.

Sophist-UK:
I am not rebasing - I am going to revert the commit which is different.

Sorry - I cannot cherry pick your commits - because my code is not quite the same. However I will try to remember to do that next time.

But I have added you as a collaborator to my fork which allows you to make commits into the branch on my fork which will then become part of this PR.

zas:
> I accept that this is extra to the functionality of this PR. The idea is that it adjusts the operating system priority of the thread down (or up) in order to influence the way threads are dispatched by Python. But now I understand more about how the Python GIL works and impacts which thread gets the GIL (which is essentially random), I am not sure that it would be of any use. (And we don't know what else is running on the user's system and whether it would make sense to do it.)
> 
> So I will be happy to revert this Commit.

Yes, if it isn't totally related better do it in another PR and keep this one simple.

picard/tagger.py
Sophist-UK: Fixed.

Sophist-UK:
I have just reverted the os_priority commit.

Because of the way that the Python GIL works, having threads at different priorities doesn't actually work - the choice of which thread to run is essentially random - so you cannot influence it by os priorities. And since the GIL makes Python essentially single threaded, running a thread at a lower priority essentially makes the application as a whole run at that lower priority until the thread finishes or gets paused. 

So until the GIL is changed to a fine-grained locking mechanism which allows proper parallel pre-emptive multi-tasking - a change which is likely to be never!! - there is no point to this functionality IMO and so no point in submitting it as a separate PR.

Sophist-UK:
See http://www.dabeaz.com/python/UnderstandingGIL.pdf if you want more detail about the GIL issue. Indeed I would say that this should be mandatory reading for anyone wanting to work on threading functionality because it is very counter-intuitive to how you expect Python multithreading to work. You expect it to be pre-emptive and it isn't. You expect it to utilise multiple-cores and it doesn't. So whilst it useful for running blocking (i.e. disk) I/O, you still need to use processEvents to keep the UI responsive for any code - regardless of whether it is on the main or worker threads - which uses CPU for at least .1sec (which if course depends on the speed of your CPU core), and (in this PR) whenever you want the GUI to reflect any changes (thread.to_main() in Picard's case).

mineo:
With this pull request applied I get the same behaviour as described in https://github.com/metabrainz/picard/pull/613#issuecomment-277433096. The last commit I have is dd3ac5f4ceaf01af0745004a0fb2d1ca4706b8db.

Sophist-UK:
Yes - that is the last commit. I will look into this if I have time before I go away. How does this compare to the behaviour in the current master branch?

Sophist-UK:
Just tried it on windows using a few albums on a remote share and it worked fine - UI was responsive and I could kill it quickly.

Sophist-UK:
But I think it uses a different code path when you drag and drop files on a Mac - but I can't test that because I only have windows.

Sophist-UK:
I have found where drop events are handled - so it may be the same code for windows and macs. 

If we are still needing to warn user about loading a lot of files (which no longer seems to be as much of an issue with this PR) then it is not difficult to consolidate multiple directories into a list before calling tagger.py to load them so that we can the total number of files in all the directories to pop-up a warning.

Sophist-UK:
Code added to handle Add of multiple directories in a consistent way:

Add Directory button
Drag and drop
Command line

Sophist-UK:
P.S. I know this is not in scope of the performance issues but it is a tidy up based on having to change the Add DIrectory functionality because of performance issues.

picard/tagger.py
```diff
@@ -399,6 +408,36 @@ def _open_files_finished(self, target=None, result=None, error=None):
...
+        else:
+            self.window.set_statusbar_message(
+                N_("Adding directory: '%(directory)s' ..."),
+                {'directory': dir_list[0]}
+            )
+
+        list_of_files = []
+        recursively = config.setting['recursively_add_files']
+        for directory in dir_list:
+            if recursively:
```
zas: I would get rid of this test in loops, and make use of the already imported `itertools.chain`:

```python
    def _add_directory_method(self):
        if config.setting['recursively_add_files']:
            return self._add_directory_recursive
        return self._add_directory_non_recursive

    def add_directory_list(self, dir_list, parent=None):
        if parent:
            parent = os.path.normpath(parent)
        number_of_dirs = len(dir_list)
        if number_of_dirs > 1:
            if parent:
                self.window.set_statusbar_message(
                    N_("Adding %(count)d directories from '%(directory)s' ..."),
                    {'directory': parent, 'count': number_of_dirs}
                )
            else:
                self.window.set_statusbar_message(
                    N_("Adding %(count)d directories ..."),
                    {'directory': parent, 'count': number_of_dirs}
                )
        else:
            self.window.set_statusbar_message(
                N_("Adding directory: '%(directory)s' ..."),
                {'directory': dir_list[0]}
            )

        add_directory = self._add_directory_method()
        self._add_files(list(chain.from_iterable([add_directory(d) for d in dir_list])))

    def add_directory(self, path):
        add_directory = self._add_directory_method()
        self._add_files(add_directory(path))
```

picard/tagger.py
```diff
@@ -346,105 +363,152 @@ def move_files(self, files, target):
...
+        number_of_files = sum([len(x) for x in list_of_filenames])
+        number_of_directories = len(list_of_filenames)
+        log.debug("Adding a total of %d files in %d directories",
+                number_of_files, number_of_directories)
+        self.window.set_statusbar_message(
+            N_("Adding a total of %(files)d files in %(directories)d directories ..."),
+            {'files': number_of_files, 'directories': number_of_directories},
+            echo=None
+        )
+        for filenames in list_of_filenames:
```
zas: According to line 371 `number_of_directories = len(list_of_filenames)` so i guess you should rename `filenames` here to `directory`.

picard/tagger.py
```diff
@@ -346,105 +363,152 @@ def move_files(self, files, target):
...
-                continue
             if filename not in self.files:
                 file = open_file(filename)
                 if file:
-                    self.files[filename] = file
-                    new_files.append(file)
-        if new_files:
+                    new_files[filename] = file
+                # Allow UI to process events to keep it responsive
+                QtCore.QCoreApplication.processEvents()
```
zas: Does it have to be done on each loop ?

picard/tagger.py
```diff
@@ -346,105 +363,152 @@ def move_files(self, files, target):
...
+        return new_files
+
+    def _open_files_finished(self, target=None, result=None, error=None):
+        if result:
+            new_files = []
+            for filename in sorted(result):
```
zas: I wonder if this sort is really useful. Is there a reason to keep it ?

picard/tagger.py
```diff
@@ -346,105 +363,152 @@ def move_files(self, files, target):
...
-        def process(result=None, error=None):
-            if result:
-                if error is None:
-                    self.add_files(result)
-                thread.run_task(get_files, process)
-
-        process(True, False)
+        list_to_add = []
+        # walk topdown to allow modifying the sub-dirs to remove hidden dirs
+        for root, dirs, files in os.walk(unicode(path), topdown=True):
```
zas: Just a note: I just noticed `os.walk()` has no followlinks parameter, and therefore defaults to False, meaning it will not follow symlinked directories (on systems supporting them). I did a fast test, and verified it. Of course, enabling followlinks would imply to handle infinite loops.

picard/tagger.py
samj1912: Nope, this is just left over code from #490

picard/tagger.py
Sophist-UK: I wasn't sure what to call it. 

"filename" would typically refer to string, and "filenames" would refer to a list of "filename". This is a list of filenames, where each member is a list of files in a single directory. 

Similarly "directory" suggests a string containing a directory path. 

That said `for directory in list_of_filenames:` would at first sight seem a bit odd, but you could probably work it out.

How about `filenames_in_directory`?

picard/tagger.py
Sophist-UK: I have no idea - I didn't benchmark how often it needed to be done on my system for the UI to stay responsive, but it would depend mainly on the time taken to read the 128 bytes for the file (so on whether you had an SSD or were accessing a file on a share) and to a lesser extent on your CPU clock rate. 

Similarly I have not benchmarked whether the python overhead of keeping a count so we could call this every x calls would be greater than the overhead of switching to the main thread finding there are no events to process and switching back.

If we did it once a directory instead of once per file, then it would depend on how many files there are in the directory - which or some releases on MB is several hundred.

And remember it is better to do it more often than necessary to keep the UI responsive rather than less often (which would create unresponsiveness).

picard/tagger.py
Sophist-UK: Yes - it is necessary - files are not necessarily provided in track order - so what this does is display the tracks in correct sequence if they have track numbers (which in most cases will be true).

picard/tagger.py
Sophist-UK: Yes - I spotted that and thought the same. Most people will not be using symbolic links in the midst of their music library directory structure.

picard/tagger.py
Sophist-UK: Yes - use of a method like this is good pythonesque code. However doesn't this code produce a list rather than a list of lists required by _add_files. But I think it only needs a quick tweak to make it work.

Sophist-UK: No it was perfect.

zas:
> Similarly I have not benchmarked whether the python overhead of keeping a count so we could call this every x calls would be greater than the overhead of switching to the main thread finding there are no events to process and switching back.

May be stupid, but wouldn't be possible to run a timer in a high priority thread executing `processEvents()` on a regular basis ?

Something like (in a dedicated thread):

```python
        def p():
        timer = QtCore.QTimer()
        timer.timeout.connect(p)
        timer.start(100)
```

Sophist-UK:
No - not a stupid idea, however it won't work. Aqs I said in an earlier comment, the GIL makes multithreading work counter-intuitively. In essence it works more like Windows 3 (which required tasks to cooperate in multi-tasking) rather than todays o/s which is pre-emptive, so this thread would only be competing with the main task to get processor time rather than forcing the main task to get processor time. And priorities don't work anyway. See the previous link to an explanation of how the GIL works and why this won't.

Sophist-UK:
Off on holiday for 2 weeks tomorrow, so if there are any questions, please say now.

P.S. Good luck with the 1.4 release next Tuesday.

Sophist-UK:

Sophist-UK:
Ok - I have fixed the merge issue created by another PR.
I previously addressed the changes requested by @samj1912 but Github is still showing this as unresolved.

IMO this is ready to merge again - and I think it should be merged ASAP in order to allow for as much testing by others before 1.4.1 is released.

picard/tagger.py
Sophist-UK: Apparently NOT fixed. Sorry. Will fix it now.

Sophist-UK: Sorry again. It was fixed - but it is in `_add_files_from_directory` not here.

Sophist-UK:
I have been using this for the last few days for some serious tagging and it seems very stable.

Sophist-UK:
@mineo - I have managed to duplicate the issue with dragging and dropping large numbers of files and will try to diagnose what is causing it.

Sophist-UK:
@mineo I believe I have now addressed the non-responsiveness when loading large numbers of files - and it should behave the same whether you load 1 directory with lots of sub-directories or all the sub-directories explicitly, and should behave the same whether you use `Add Directory ` or drag and drop.

In essence there were several loops which monopolised the GIL without cedeing control to the UI. So we now do a ProcessEvents in the loop and it seems very responsive in my tests on Win 10.

Can you please retest on your own system and let me know what you find?

Thx.

picard/tagger.py
Sophist-UK: @samj1912 Since I have fixed this review comment can you please mark it as fixed?

Sophist-UK:
The UI responsiveness also seems to depend on whether you are running from the console and / or have debug enabled - presumably because writing to the console makes calls to ProcessEvents because it has to be done in the main thread. 

More importantly under some conditions I have been getting `RuntimeError: maximum recursion depth exceeded` errors - which is indicative of running ProcessEvents when you are already on the main thread, which if there are events to process can run other code rather than returning to whatever called the ProcessEvents method. This may happen if I try to save files whilst I am also loading them.

Which gives us a bit of a catch-22: we need to call processEvents in CPU bound code running on the main thread in order to keep the GUI responsive, but this can dispatch some other piece of work (rather than returning to the thread which called ProcessEvents) which can then call ProcessEvents itself and thus create significant stack depth.

The solution may be to run some of the CPU-intensive stuff in a worker thread - which has its own headaches in updating the UI as work is done. Alternatively I may be able to find some method of making sure that we return to the code that called ProcessEvents. Or maybe this is just a bug in the file save code.

Sophist-UK:
I have at least worked out in my head what the cause of the `RuntimeError: maximum recursion depth exceeded` is.

When a ProcessEvents is called in the main thread from the call-back routine, instead of returning to the statement after the ProcessEvents, instead a new worker thread is dispatched which then does the same, and eventually when you have a large number of call-back routine calls to ProcessEvents stacked up you get the error message.

The solution seems to me to be:

1. Don't call ProcessEvents in the main thread - so to keep the UI responsive we must avoid CPU intensive activities in the main thread, which means... 

2. Moving all the loops for starting the load of files into a worker thread.

3. Ensuring that the call-back routines which have to run in the main thread because they update the UI are run per file so that we don't have the possibility of large loops;

4. Make the `log` code work in the same way as far as ProcessEvents calls regardless of whether debug is on or there is a console to write to.

I won't have time to look at this until this evening - and maybe not even until later this week.

But I welcome comments from anyone as to whether this sounds right or not, and if anyone has any brighter ideas I would love to hear them.

Sophist-UK:
P.S. I also wondered whether we should maintain our own queues for worker threads rather than use the inbuilt threadpool queues, but there doesn't seem much point in this to me - unless,of course, other's can see reasons that I can't (in which case please speak up).

Sophist-UK:
This should now hopefully be working at volume and give improved UI responsiveness. There are probably more changes that can be made to improve responsiveness elsewhere (like summarising the metadata for selected tracks, and possibly for handling webservice responses), but as it stands this should still be a stable improvement.

I don't really want to rebase this again, but I don't want to get too far behind on changes to master before looking at further improvements. So I would like to make sure that this is stable and get it merged and then work on the next set of improvements in a separate PR from a fresh baseline.

@mineo - Would you mind testing this again?

Sophist-UK:
This is NOT yet ready for merging - whilst the UI generally is responsive even with big loads, some functions cease working during saves - so this needs to be fixed. It may be due to track analysis calling processEvents from the main thread - I need to check.

Sophist-UK:
There is still a problem with UI responsiveness loading large numbers of files. The symptoms are that one of the worker threads (doing the new file evaluation) monopolises execution despite issuing a processEvents. Debug messages from worker threads (which also run to_main) seem to appear promptly, but the call-back function doesn't run promptly. Other worker threads like loading files are not getting a look in and I suspect that neither is the main thread. I am guessing that when it issues a local disk I/O for the 128 bytes the read finishes so quickly that the thread is ready to execute when Python issues a wakeup to all threads to decide what to execute next - and when the thread runs to_main and processEvents, again the thread is ready to run as soon

This may be because processEvents only runs events posted from the worker thread (as suggested by the documentation). Or it may be simply because Python doesn't handle threading well and the same thread gets the execution back again by accident (putting work into worker threads typically slows things down - see presentation on how the GIL works). So moving some code back to main thread might actually improve things.

Or perhaps there is some way to force the open_threadpool to give up CPU and allow other threads to execute (we could use time.sleep but that introduces delays when there is no other work). 

This is likely to be a bit trial and error (unless anyone else has a bright idea that could help), so please bear with me.

Sophist-UK:
Ok - I am getting there. However, I could do with a bit of advice.

According to http://www.qtcentre.org/printthread.php?t=7167&pp=100 we should be using signals and slots to do calling between worker threads and main thread. I haven't tried switching yet, but it looks like this switches thread and executes the slot immediately - and if so that would seem to be a better way of doing `to_main`. Whilst this would separate `to_main` execution from ensuring that the main thread gets sufficient CPU to keep the UI responsive, I think I have a different solution to that.

I am wondering whether our current approach in `threads.to_main` is just a legacy or whether there is a specific reason that we are not using signals and slots?

Sophist-UK:
Ok - This is ready for further testing.

Please let me know if I have not resolved any previous review comments.

samj1912:
@Sophist-UK build  is failing. Lots of attribute errors.

Sophist-UK:
@samj1912 Damn. It was running perfectly on my PC. But I didn't run unit tests as I should have done - rusty at this. And when I do, I get the same errors. I will look into it.

Sophist-UK:
I suspect that this is more about how unit testing is setup - without Qt initialisation to_main won't work.

So I either need to modify setup.py so that it initialises Qt (prefered) or have additional checks in thread.py to handle unit test situations (which I don't like the idea of).

Sophist-UK:
Hopefully that will fix it - will check that Travis ran clean in the morning.

Sophist-UK:
Still more work needed.

1. A bug where metadata panel stops updating.
2. When processing lots of webservice responses, UI becomes unresponsive - so need to move some of the heavier webservice handling into worker-threads.

samj1912:
It's been almost 8 months since this was updated. Closing this with a 'Needs more work' tag. You can reopen this if you want to continue working on it. Also we have updated to Qt5 and py3 so be mindful of that while reworking on this PR.

Sophist-UK:
I agree that this should be closed for three reasons:

a. It was written for P1.4 and needs to be retrofitted to P2.
b. It has multi-threading & intermittent bugs which are hellishly difficult to resolve.
c. It was way too big and needs to be broken into smaller PRs.

When I get time I will rework it for P2 in smaller chunks.

PR #605, opened 2017-01-27
PICARD-958: Avoid dir/file name unicode errors

Sophist-UK:
Resolves [PICARD-958](https://tickets.metabrainz.org/browse/PICARD-958).

When dir or file names contain unicode characters and debug logging is on, then debug.log statements which use %s placeholders cause the following error:
```
  File ".\picard\tagger.py", line 315, in event
    event.run()
  File ".\picard\util\thread.py", line 34, in run
    self.func(*self.args, **self.kwargs)
  File ".\picard\log.py", line 116, in _stderr_receiver
    sys.stderr.write(formatted_log_line(level, time, msg + os.linesep))
  File "C:\Program Files (x86)\Python2\lib\encodings\cp850.py", line 12, in encode
    return codecs.charmap_encode(input,errors,encoding_map)
UnicodeEncodeError: 'charmap' codec can't encode character u'\u2010' in position 77: character maps to <undefined>
```

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -151,7 +151,7 @@ class Tagger(QtGui.QApplication):
         log.debug("Platform: %s %s %s", platform.platform(),
                   platform.python_implementation(), platform.python_version())
         log.debug("Versions: %s", versions.as_string())
-        log.debug("Configuration file path: %s", config.config.fileName())
+        log.debug("Configuration file path: %r", config.config.fileName())
 
         # TODO remove this before the final release
         if sys.platform == "win32":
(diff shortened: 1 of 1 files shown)


mineo:
There are other occurences of log.debug with "%s" in the format string where the  value inserted (for example in `webservice.py` with the cacheDirectory, a lot in `acoustid.py`), do these work correctly?

Sophist-UK:
Hmm... unfortunately the trace stack for these doesn't show where the error is occurring (because it is handed off to a separate thread to print it). But I am indeed still getting errors.

I think perhaps an alternative approach is needed i.e. switch back to using %s for strings in tagger.py, file.py etc. and implement a fix in log.py to print the message to stderr with %r which puts the conversion where it is needed not for e.g. the error/debug window where it adds lots of extra \\\\

PR #606, opened 2017-01-29
PICARD-958 Better fix for unicode errors when writing to stderr

Sophist-UK:
This replacement fix for PICARD-958 addresses the root cause of unicode errors when writing to stderr, and avoids poluting the error/debug log (which can handle unicode characters) with double `\\` which occur when you format any string with %r.

I have addressed all the relevant log statements that I could find, but it is entirely possible that I have missed some.

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -113,7 +113,7 @@ def formatted_log_line(level, time, message, timefmt='hh:mm:ss',
 
 def _stderr_receiver(level, time, msg):
     try:
-        sys.stderr.write(formatted_log_line(level, time, msg + os.linesep))
+        sys.stderr.write("%r%s" % (formatted_log_line(level, time, msg), os.linesep))
     except UnicodeDecodeError:
         import traceback
         traceback.print_exc()
--- a/picard/log.py
+++ b/picard/log.py
@@ -113,7 +113,7 @@ def formatted_log_line(level, time, message, timefmt='hh:mm:ss',
 
 def _stderr_receiver(level, time, msg):
     try:
-        sys.stderr.write("%r%s" % (formatted_log_line(level, time, msg), os.linesep))
+        sys.stderr.write(formatted_log_line(level, time, msg + os.linesep))
     except UnicodeDecodeError:
         import traceback
         traceback.print_exc()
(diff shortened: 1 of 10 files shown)


Sophist-UK:
This PR reverts #605 and provides a better solution.

picard/log.py
```diff
@@ -113,7 +113,7 @@ def formatted_log_line(level, time, message, timefmt='hh:mm:ss',
 
 def _stderr_receiver(level, time, msg):
     try:
-        sys.stderr.write(formatted_log_line(level, time, msg + os.linesep))
+        sys.stderr.write("%r%s" % (formatted_log_line(level, time, msg), os.linesep))
```
zas: It will ouput the whole line as 'u"..."', not what we want (at all)

```
u"D: 14:23:36 WSREQ: Last request to (u'musicbrainz.org', 80) was 1000 ms ago, starting another one"
```
The issue is related to encodings and win console (and perhaps others), and we are using unicode internally, the exception you get is because you output to a console with another encoding and not all chars can be converted to the destination charmap.

See:
- http://stackoverflow.com/questions/5419/python-unicode-and-the-windows-console
- https://wiki.python.org/moin/PrintFails
- https://www.python.org/dev/peps/pep-0528/
- http://stackoverflow.com/questions/1545263/utf-8-in-python-logging-how

Can you try without this PR changes, but adding line:
```
sys.stderr = codecs.getwriter(locale.getpreferredencoding())(sys.stderr)
```
just before `sys.stderr.write(...)` call ?

Sophist-UK: Yes - I will try it.

However it doesn't mean we won't need the rest of the PR.

zas: > Yes - I will try it.
> 
> However it doesn't mean we won't need the rest of the PR.

The rest of the PR doesn't help to fix the issue, those changes will not go in 1.4, but the console encoding output fix may.

Sophist-UK:
OK - I have reverted the change to convert to unicode when writing to stderr and left the remainder in here.

I will create a second PR with @zas proposed changes so that we can test and merge that first, and then test this PR against Picard with that change merged.

Once I have submitted this second PR, I will rebase this one against that change so it is made pre-requisite.

Sophist-UK:
Unfortunately, @zas' proposed fix doesn't work reliably, and tweaks don't work either.

I have coded a different solution  which I have submitted as PR #607. The rest of this PR is also wrong, so closing.

PR #607, opened 2017-01-31
PICARD-958: Fix-stderr-unicode encoding-errors

Sophist-UK:
This is a further fix to [PICARD-958](https://tickets.metabrainz.org/browse/PICARD-958).

No point in having traceback.

Instead if we get a unicode error, we try using a repr format for the message instead.

This replaces [PR #606](https://github.com/metabrainz/picard/pull/606).

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -20,7 +20,6 @@
 from __future__ import print_function
 import sys
 import os
-import codecs, locale
 from collections import deque
 from PyQt4 import QtCore
 from picard.util import thread
--- a/picard/log.py
+++ b/picard/log.py
@@ -20,6 +20,7 @@
 from __future__ import print_function
 import sys
 import os
+import codecs, locale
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Note: This whole mess should go away if / when we move to Picard 3.

Sophist-UK:
It is better for Picard code to use repr formatting whenever file names or directories are sent to log.debug as doing it for the whole message puts the `u'string'` around the whole message instead of just the file name or directory, but this avoids any errors and outputs the log message to stderr even if the formatting is not perfect.

picard/log.py
```diff
@@ -20,6 +20,7 @@
 from __future__ import print_function
 import sys
 import os
+import codecs, locale
```
Sophist-UK: Yes - from previous attempts as per @zas suggestion.

I will remove.

PR #608, opened 2017-02-01
PICARD-970 Read files unbuffered in guess_format

Sophist-UK:
Resolves [PICARD-970](https://tickets.metabrainz.org/browse/PICARD-970).

### DIFF
--- a/picard/formats/__init__.py
+++ b/picard/formats/__init__.py
@@ -45,7 +45,9 @@ def supported_extensions():
 def guess_format(filename, options=_formats):
     """Select the best matching file type amongst supported formats."""
     results = []
-    with file(filename, "rb", 128) as fileobj:
+    # Since we are reading only 128 bytes and then immediately closing the file,
+    # use unbuffered mode.
+    with file(filename, "rb", 0) as fileobj:
         header = fileobj.read(128)
         # Calls the score method of a particular format's associated filetype
         # and assigns a positive score depending on how closely the fileobj's header matches
--- a/picard/formats/__init__.py
+++ b/picard/formats/__init__.py
@@ -45,7 +45,7 @@ def supported_extensions():
 def guess_format(filename, options=_formats):
     """Select the best matching file type amongst supported formats."""
     results = []
-    with file(filename, "rb") as fileobj:
+    with file(filename, "rb", 128) as fileobj:
         header = fileobj.read(128)
         # Calls the score method of a particular format's associated filetype
         # and assigns a positive score depending on how closely the fileobj's header matches


picard/formats/__init__.py
```diff
@@ -45,7 +45,7 @@ def supported_extensions():
 def guess_format(filename, options=_formats):
     """Select the best matching file type amongst supported formats."""
     results = []
-    with file(filename, "rb") as fileobj:
+    with file(filename, "rb", 128) as fileobj:
```
zas: I fear it would have a counter effect. Actually it may be much slower. Therefore please profile this code (that is using buffer of 128 instead of specifying none which is equal to using system default).

See https://docs.python.org/2/library/functions.html#open and http://stackoverflow.com/a/29712601

Sophist-UK:
I doubt it will be slower. AFAIK, setting a buffer size does not change the logic used, just the size of the buffer from the default.

Most windows file systems use NTFS with a 4K cluster size, so a default buffer of 8192 means it will read two clusters. And if the file is remote it will transmit 8192 bytes over the network.

However, having read the description again, it may be better to use zero i.e. unbuffered - after all we are going to read exactly 128 bytes and then close the file, so not worth buffering anything.

All that said, I have no idea how to profile this, nor the time to invest in that. So I will change buffer to zero and then leave it to others to decide whether to merge or close unmerged.

zas:
Without profiling, such change is too much a bet, and since it depends much on systems and filesystems, very hard to tell what is the exact effect.
So, until we are able to profile this in various envs, i prefer to not merge this PR.

@mwiencek @mineo @samj1912 : what do you think ?

samj1912:
I agree on not merging unless tested.

Also http://stackoverflow.com/questions/29712445/what-is-the-use-of-buffering-in-pythons-built-in-open-function

Doesn't seem like it will cause much of a change when reading small amounts of data.

Sophist-UK:
I don't see this as any more of a risk than adding the file reads when recently creating guess_format.

In the overall elapsed time for reading the file either locally or remotely, the cost of CPU cycles referred to in the links above for buffering / not buffering is trivial.

My main objective here is to reduce network traffic from 8192 to 128 bytes for remote files.

zas:
> I don't see this as any more of a risk than adding the file reads when recently creating guess_format.
> 
> In the overall elapsed time for reading the file either locally or remotely, the cost of CPU cycles referred to in the links above for buffering / not buffering is trivial.
> 
> My main objective here is to reduce network traffic from 8192 to 128 bytes for remote files. For an album of 20 tracks, a reduction of 20 x 8064B = c. 1.3Mb is reasonably significant.

I can think of so many reasons it wouldn't work as you expect... i will just give you a hint: i don't see any protocol, frame size, header size, encryption algorithm, compression algorithm, windowing algorithm, filesystem version and options, language version, library versions, memory size, processor specifications, disk cache technology, physical link specifications, etc... in this calculation.

Make proper measurements, not such simplistic calculations.

Note: i don't say the change you propose is good or bad, i don't know, but i know the only way to know is to measure.

Sophist-UK:
I think we are getting over focussed here. 

We were quite happy to add the guess_format function which adds a whole extra file operation for every file without benchmarking the consequences, but not happy to do a simple change to minimise the impact of it without an extensive benchmark.

I think this needs a quick suck it and see if it has any noticeable adverse impact just like the original change.

Sophist-UK:
> I can think of so many reasons it wouldn't work as you expect... i will just give you a hint: i don't see any protocol, frame size, header size, encryption algorithm, compression algorithm, windowing algorithm, filesystem version and options, language version, library versions, memory size, processor specifications, disk cache technology, physical link specifications, etc... in this calculation.

This was a quick back of the fag packet estimate. Which is better than no estimate at all for adding the guess_format code.

zas:
> I think we are getting over focussed here.
> 
> We were quite happy to add the guess_format function which adds a whole extra file operation for every file without benchmarking the consequences, but not happy to do a simple change to minimise the impact of it without an extensive benchmark.
> 
> I think this needs a quick suck it and see if it has any noticeable adverse impact just like the original change.

I don't understand why it is so hard for you to make measurements. Since you are lazy, and i think this patch *can* improve the situation, i wrote a test, and ran it on linux (kernel 3.19.0-32-generic, trusty, ext4 partition/RAID1 software, on WDC WD1003FZEX-0 hardrive)
Using this test script: https://gist.github.com/zas/fbaa942164b569067120895c60a4c0d5

And here the results:

python 2.7.6
```
python2 test_read.py /tmp/
Default buffer size: 8192
datasize: 16384
buf=    -1 files=  5000 secs=   0.110 usecsperfile=  22.089
buf=     0 files=  5000 secs=   0.056 usecsperfile=  11.259
buf=   128 files=  5000 secs=   0.057 usecsperfile=  11.325
buf= 16384 files=  5000 secs=   0.083 usecsperfile=  16.584
buf=   128 files=  5000 secs=   0.060 usecsperfile=  12.012
buf=     0 files=  5000 secs=   0.055 usecsperfile=  11.082
buf=    -1 files=  5000 secs=   0.095 usecsperfile=  18.923
```

python 3.4.3
```
python3 test_read.py /tmp/
Default buffer size: 8192
datasize: 16384
buf=    -1 files=  5000 secs=   0.123 usecsperfile=  24.618
buf=     0 files=  5000 secs=   0.077 usecsperfile=  15.379
buf=   128 files=  5000 secs=   0.096 usecsperfile=  19.185
buf= 16384 files=  5000 secs=   0.121 usecsperfile=  24.275
buf=   128 files=  5000 secs=   0.100 usecsperfile=  19.904
buf=     0 files=  5000 secs=   0.081 usecsperfile=  16.104
buf=    -1 files=  5000 secs=   0.106 usecsperfile=  21.275
```

Conclusion: **in my test env, the proposed change gives the best performance.**

Now, instead of arguing, you should run this test in your own env (network filesystem/windows i think), and post results.
Use wireshark or the like to sniff network traffic during tests.

Note 1 : when you submit a PR, it is up to you to convince it is a positive change, this time i took time to do proper tests, next time i will just ignore the PR.

Note 2: if you think using magic number instead extensions has a significant negative impact in your env, what about proposing a patch implementing an option to make it optional ?

Note 3: interesting, python3 vs python2 differences

Sophist-UK:
@zas I don't appreciate personal insults suggesting that "I am lazy". I have a lot of things I need to do that compete for my time, and compared to running a business and preparing to move house, Picard is a lower priority and gets a limited amount of time.

Secondly, I am great at enhancing existing code, but my skills to write stand-alone test cases are not there - so it would take me a disproportionate amount of time to create such a test myself.

Thirdly, there was nothing to suggest that my proposal would create a performance problem rather than make minor improvements - and as I said, no one bothered to test the impact of an extra file operation on every file or indeed bother discussing whether it was an issue that needed consideration before the code was merged.

However, I will certainly run your code against a remote test directory and post the results here.

Sophist-UK:
Possible minor bug in the test code: Does L49 `secs += t.secs` need to be indented to be within the `with` block that defines t.

mwiencek:
> Thirdly, there was nothing to suggest that my proposal would create a performance problem rather than make minor improvements - and as I said, no one bothered to test the impact of an extra file operation on every file or indeed bother discussing whether it was an issue that needed consideration before the code was merged.

I think the difference is that that PR was adding a feature rather than claiming to improve performance, whereas this PR is claiming to improve performance (without any measurements). I definitely agree it would've been wise to measure the performance of the previous PR too, but this one is claiming a specific improvement.

Sophist-UK:
I think what I was saying (or at least was thinking and trying to say) was that the PR was to select what should be the optimum parameters based on theory. I was hoping for rather than promising performance improvements, though even I knew they would be likely to be minor.

But I know that @zas wants code that is optimum - a recent comment from him on a PR was to use a set rather than a list for a likely len of < 10 because there as no need for duplicates in the list. He was right about that - we should code optimally - but this is a similar situation.

Sophist-UK:
Here are the results:
```
Using` '\\\\D630\\D630-D\\Paul' as temporary directory
No of files: 500
Filesize: 32768

Write took 60 seconds datarate=2107kbps
Read: buf=    -1 files=   500 secs=  22.053 usecsperfile=44105.994 effective-datarate=22kbps
Read: buf=     0 files=   500 secs=  21.431 usecsperfile=42862.001 effective-datarate=23kbps
Read: buf=   128 files=   500 secs=  21.440 usecsperfile=42880.006 effective-datarate=23kbps
Read: buf=  4096 files=   500 secs=  22.189 usecsperfile=44377.999 effective-datarate=22kbps
Read: buf=  8192 files=   500 secs=  41.663 usecsperfile=83325.996 effective-datarate=12kbps
Read: buf= 16384 files=   500 secs=  73.017 usecsperfile=146033.998 effective-datarate=6kbps
Read: buf=    -1 files=   500 secs=  23.115 usecsperfile=46230.000 effective-datarate=21kbps
Read: buf=     0 files=   500 secs=  22.401 usecsperfile=44801.996 effective-datarate=22kbps
Read: buf=   128 files=   500 secs=  22.209 usecsperfile=44418.001 effective-datarate=22kbps
Read: buf=  4096 files=   500 secs=  21.571 usecsperfile=43142.002 effective-datarate=23kbps
Read: buf=  8192 files=   500 secs=  42.298 usecsperfile=84595.998 effective-datarate=11kbps
Read: buf= 16384 files=   500 secs=  74.390 usecsperfile=148779.994 effective-datarate=6kbps
```

It looks to me like it is transmitting 4096 bytes minimum - perhaps because this is the cluster size of the hard drive on the remote server or possibly because that is some sort of default network block size.

So it appears that my PR makes little difference - and I am happy for it to be merged or dropped as others think best.

Sophist-UK:
P.S. Effective data rate is based on 128 bytes per file of usable data.

zas:
> @zas I don't appreciate personal insults suggesting that "I am lazy". I have a lot of things I need to do that compete for my time, and compared to running a business and preparing to move house, Picard is a lower priority and gets a limited amount of time.

Sorry i didn't want to insult you or anything, i'm myself quite lazy ;)

But everyone has limited time, and you often answer, when requested for more information or modifications, or in this case for measurements, that you lack of time and therefore will not do it, time is always a question of priorities.
If you lack of time to discuss/improve/modify/document your proposals, better not propose anything.
When you propose a patch "improving the performance" better give proofs of the actual improvement, it makes things easier for everyone.

> 
> Secondly, I am great at enhancing existing code, but my skills to write stand-alone test cases are not there - so it would take me a disproportionate amount of time to create such a test myself.
> 

If you don't measure, there is no point in attempting to improve performance, you'll lose your time.

> Thirdly, there was nothing to suggest that my proposal would create a performance problem rather than make minor improvements - and as I said, no one bothered to test the impact of an extra file operation on every file or indeed bother discussing whether it was an issue that needed consideration before the code was merged.
> 

That's not my opinion, especially when it comes to filesystems/low level operations/buffer sizes. Any tuning in this field needs to be carefully evaluated. And i had plenty of reasons thinking it would not achieve your goal (which was reducing the network traffic).
Also the gain (or loss) in performance of such change is very dependent on the environment.

> However, I will certainly run your code against a remote test directory and post the results here.

Yes. :+1:

zas:
> Here are the results:
> 
> Using` '\\\\D630\\D630-D\\Paul' as temporary directory
> No of files: 500
> Filesize: 32768
> 
> Write took 60 seconds datarate=2107kbps
> Read: buf=    -1 files=   500 secs=  22.053 usecsperfile=44105.994 effective-datarate=22kbps
> Read: buf=     0 files=   500 secs=  21.431 usecsperfile=42862.001 effective-datarate=23kbps
> Read: buf=   128 files=   500 secs=  21.440 usecsperfile=42880.006 effective-datarate=23kbps
> Read: buf=  4096 files=   500 secs=  22.189 usecsperfile=44377.999 effective-datarate=22kbps
> Read: buf=  8192 files=   500 secs=  41.663 usecsperfile=83325.996 effective-datarate=12kbps
> Read: buf= 16384 files=   500 secs=  73.017 usecsperfile=146033.998 effective-datarate=6kbps
> Read: buf=    -1 files=   500 secs=  23.115 usecsperfile=46230.000 effective-datarate=21kbps
> Read: buf=     0 files=   500 secs=  22.401 usecsperfile=44801.996 effective-datarate=22kbps
> Read: buf=   128 files=   500 secs=  22.209 usecsperfile=44418.001 effective-datarate=22kbps
> Read: buf=  4096 files=   500 secs=  21.571 usecsperfile=43142.002 effective-datarate=23kbps
> Read: buf=  8192 files=   500 secs=  42.298 usecsperfile=84595.998 effective-datarate=11kbps
> Read: buf= 16384 files=   500 secs=  74.390 usecsperfile=148779.994 effective-datarate=6kbps
> 
> It looks to me like it is transmitting 4096 bytes minimum - perhaps because this is the cluster size of the hard drive on the remote server or possibly because that is some sort of default network block size.
> 
> So it appears that my PR makes little difference - and I am happy for it to be merged or dropped as others think best.

Setting the buffering to 0 didn't show any degradation in performance, either on your tests or mine, so i think, since it is actually improving performance in some cases, that this change can go in.

zas:
> I think the difference is that that PR was adding a feature rather than claiming to improve performance, whereas this PR is claiming to improve performance (without any measurements). I definitely agree it would've been wise to measure the performance of the previous PR too, but this one is claiming a specific improvement.

Exactly.
But everyone seems to think i didn't evaluate the impact of the original PR. The fact is i did, after testing, found out it was very low, we discuss about it with @samj1912 , being perfectly aware of the impact of this move.
Next time, i'll attach the results of my tests to the PR, to make it clear. Sorry but @Sophist-UK isn't the sole person here having limited time.

That said, an option to toggle this feature would be convenient, as each user has a different env. It would make everyone happy.

mwiencek:
> But everyone seems to think i didn't evaluate the impact of the original PR. The fact is i did, after testing, found out it was very low, we discuss about it with @samj1912 , being perfectly aware of the impact of this move.

👍 Sorry for implying you didn't, I didn't follow the discussion closely.

Sophist-UK:
Oh no - not ANOTHER options checkbox!!

Sophist-UK:
> I think the difference is that that PR was adding a feature rather than claiming to improve performance, whereas this PR is claiming to improve performance (without any measurements). I definitely agree it would've been wise to measure the performance of the previous PR too, but this one is claiming a specific improvement.

> If you don't measure, there is no point in attempting to improve performance, you'll lose your time.

I think if you look back you will see that I have never suggested it would improve performance. I simply said that this looked like the optimal way of setting the parameters given the profile of the guess_format file reads and that it probably wouldn't be detrimental to performance.

Anyway, this is done now, so let's move on.

Sophist-UK:
How's that?

samj1912:
@Sophist-UK I think he meant in the commit message for https://github.com/metabrainz/picard/pull/608/commits/5acf8c5dffff750bf085a6c37b2f74d69d53e698

but that will require a rebase and force push to reword so I guess let's leave it for now and note it for future purposes?

PR #609, opened 2017-02-01
PICARD-972: Avoid save if file removed

Sophist-UK:
Handles both removal before save starts and removal whilst file is being
saved.

Resolves [PICARD-972](https://tickets.metabrainz.org/browse/PICARD-972).

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -226,8 +226,8 @@ class File(QtCore.QObject, Item):
 
     def _saving_finished(self, result=None, error=None):
         # Handle file removed before save
-        # Result is None if save was skipped because file was removed. 
-        if self.state == File.REMOVED and result is not None:
+        # Result is None if save was skipped because file was removed.
+        if self.state == File.REMOVED and result is None:
             return
         old_filename = new_filename = self.filename
         if error is not None:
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -222,6 +225,9 @@ def _rmdir(dir):
             raise OSError
 
     def _saving_finished(self, result=None, error=None):
+        # Handle file removed
+        if self.state == File.REMOVED and not result:
```
zas: In the previous comment, please explain `and not result`

picard/file.py
```diff
@@ -247,8 +253,9 @@ def _saving_finished(self, result=None, error=None):
             self.clear_pending()
             self._add_path_to_metadata(self.orig_metadata)
 
-        del self.tagger.files[old_filename]
-        self.tagger.files[new_filename] = self
+        if self.state != File.REMOVED:
```
samj1912: Any reason we are checking here again? Didn't we already return earlier?

Sophist-UK: Yes - we get here if file was removed **during** the save - and we only add the file to the tagger if it wasn't removed.

Sophist-UK:
I still need to test this latest incarnation, but providing it is working OK it will be ready to merge.

Sophist-UK:
I still need to test this - so please wait on the merge until I do so later today.

Sophist-UK:
Ok - I have tested this - and the functionality works,fine.

But it has identified another bug which is that the log.debug messages are not flushed to stderr before the executable quits.

PR #613, opened 2017-02-02
PICARD-840: Warn user when loading a large number of files

samj1912:
[PICARD-840](https://tickets.metabrainz.org/browse/PICARD-840)


Sophist-UK:
PR #604 is not quite complete, and may still impact the Add Directory functionality that this PR changes, but delivers a much more responsive UI and is stable at the current commit and can be used as a basis for considering whether this PR is still required.

That said, I am currently testing it against 8000+ mp3 files, And although I have not been measuring performance as it loads them it is definitely still slowing down - currently at c. 6300 files. I suspect that there is an issue with a File Handle leak somewhere that is causing it to slow down, but I have not been able to track this down as yet. But again, if we can track it down and fix it, that might be a better solution than this PR.

samj1912:
You might still benefit from the iterative add directory I have implemented. Let me know if it's helps.

Sophist-UK:
Yes - the iterative add directory is MUCH better than the recursive one. The only question is whether a recursive add should be run on the UI main thread (which is what this does) or on a worker thread.

Because it is potentially time consuming to iterate over a lot of directories - particularly if they are on a remote network share - you may not want to lock up the UI by running this on the main thread. On the other hand, if you want to pop-up a "Do you want to load this many files?" it probably has to run on the main thread.

Either way, the guess_format should run on a worker thread.

I think that the ideal scenario would be:

a. Have an option checkbox for whether to pop-up if loading to many files and a value for how many to check.

b. If the option is checked, then run Add Directory on the main thread and run the guess_format on a worker thread. If the option is not-checked, run Add Directory on the worker thread and do the guess_format as you load each directory.

Sophist-UK:
Hmm - when it came to the end of the recursive directory walk, my load suddenly speeded up again - so (ironically) it might be that switching to an iterative Add Directory might solve the problem of it all slowing down anyway.

Sophist-UK:
Alternatively, if you are happy for me to do so, I will copy your iterative directory add and integrate it into my PR to see if it fixes the slow-down. (Until I can move the guess_format into a separate thread, UI will continue to be unresponsive on loading of remote files.)

mineo:
I tried this and noticed a few things:

1) This only works if a single recursively added directory contains more than $threshold files. If I drag two or three folders from the file browser/"Add directory"/file manager of my choice, no popup appears.

2) When adding multiple folders that have (in total) more than $threshold files (in my case ~7000), the UI doesn't do anything for more than a minute. If I switch to a different window, then back to Picard, the window is just grey (probably gets the "Not responding" title on windows/beachball on macOS). This doesn't happen with the same folders on master.

3) Killing Picard during 2) with Ctrl-C doesn't work.

Sophist-UK:
1. This is because I think that each directory is presented to Picard as a separate Open File event tagger.py#L305 and each event is processed separately (same for directories on the command line tagger.py#L284). 

2. & 3. See #604 for the issue about Picard being unresponsive. Please try it with this PR and see if it helps in this use case.

Sophist-UK:
Oops sorry - just seen that you have already done this.

PR #614, opened 2017-02-02
PICARD-798: Handle empty collection list

samj1912:
[PICARD-798](https://tickets.metabrainz.org/browse/PICARD-798)

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -141,12 +141,14 @@ class Album(DataObject, Item):
 
         # Add album to collections
         if "collection_list" in release_node.children:
-            for node in release_node.collection_list[0].collection:
-                if node.editor[0].text.lower() == config.persist["oauth_username"].lower():
-                    if node.id not in user_collections:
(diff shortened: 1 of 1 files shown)


picard/album.py
```diff
@@ -141,12 +141,14 @@ def _parse_release(self, document):
...
         # Add album to collections
         if "collection_list" in release_node.children:
-            for node in release_node.collection_list[0].collection:
-                if node.editor[0].text.lower() == config.persist["oauth_username"].lower():
-                    if node.id not in user_collections:
-                        user_collections[node.id] = \
-                            Collection(node.id, node.name[0].text, node.release_list[0].count)
-                    user_collections[node.id].releases.add(self.id)
+            # Check for empty collection list
+            if "collection" in release_node.collection_list.children:
```
Freso: Maybe this could have been combined with the previous `if`? We're having a lot of indentation going on here already.

samj1912: definitely better to do that

PR #616, opened 2017-02-03
PICARD-936: Allow UI to be more responsive when loading files

samj1912:
An alt. approach to loading files in a load thread


Sophist-UK:
This includes the check for number of files - so it mixes approaches. However once I have finished my version I will certainly compare the two.

Sophist-UK:
#936 is complete and combines @samj1912 's changes here and my own. I suggest this PR is closed.

PR #617, opened 2017-02-03
Refactor add_release_to_user_collections

samj1912:


picard/collection.py
```diff
@@ -134,16 +134,16 @@ def request_finished(document, reply, error):
...
-                if node.editor[0].text.lower() == username:
-                    if node.id not in user_collections:
-                        user_collections[node.id] = \
-                            Collection(node.id, node.name[0].text, node.release_list[0].count)
-                    user_collections[node.id].releases.add(release_node.id)
-                    log.debug("Adding release %r to %r" %
-                              (release_node.id, user_collections[node.id]))
+    """Add album to collections"""
+    # Check for empy collection list
+    if ("collection_list" in release_node.children and
```
alastair: Why parentheses? They're not needed in such a simple condition (unless this is a picard style thing)

zas: https://www.python.org/dev/peps/pep-0008/#maximum-line-length

> The preferred way of wrapping long lines is by using Python's implied line continuation inside parentheses, brackets and braces. Long lines can be broken over multiple lines by wrapping expressions in parentheses. These should be used in preference to using a backslash for line continuation.

alastair: huh. I never knew that :) 👍

PR #618, opened 2017-02-04
Fix 'date' to id3 mapping

pR0Ps:
According to the [date docs](https://musicbrainz.org/doc/Release/Date), the 'date' tag holds "the date in which a release was made available through some sort of distribution mechanism".

Relevant excerpts from the [id3v2.4 spec](http://id3.org/id3v2.4.0-frames):
 - `TDRC`: "Recording time [...] when the audio was recorded".
 - `TDRL`: "Release time [...] when the audio was first released"
 - `TDOR`: "Original release time [...] when the original recording of the audio was released"

Currently the 'date' tag is mapped to `TDRC` (recording time). This commit changes the mapping to `TDRL` (release time) to bring the two tag types more in line with each other.

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -115,7 +115,7 @@ class ID3File(File):
         'TCOP': 'copyright',
         'TPUB': 'label',
         'TDOR': 'originaldate',
-        'TDRC': 'date',
+        'TDRL': 'date',
         'TSSE': 'encodersettings',
         'TSOA': 'albumsort',
         'TSOP': 'artistsort',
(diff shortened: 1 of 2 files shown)


samj1912:
It is kept as TDRC for program compatibility reasons.
from foobar2000 tag mappings.
> The TDRC frame is officially to be used for the recording time, TDRL is for the release time but TDRC is generally used by most programs.
The same goes for the rest of tagging programs.

Freso:
Two things:
1. Is this how the tags are actually used in the wild?
2. Does the current mapping originate from mutage?

If 1., then certainly, +100 from me.
If 1. *and* 2., then please submit a patch upstream to mutagen as well. :)

Sophist-UK:
I seem to remember asking this myself a few years ago. I guess we need a comment in the code to point these things out.

pR0Ps:
I submitted an issue (well, question) to mutagen asking about this behavior as well: https://github.com/quodlibet/mutagen/issues/290

samj1912:
Looking at the reply by @lazka on quodlibet/mutagen#290 I think we can close this.

PR #619, opened 2017-02-08
PICARD-978: Distinguish between unclustered unmatched files

Sophist-UK:
Calling it unclustered in the UI will help users to realise that they need to Cluster files, and in debug messages helps to distinguish between Unclustered files just loaded, and an album with Unmatched files.

https://tickets.metabrainz.org/browse/PICARD-978


Sophist-UK:

mwiencek:
They don't need to be (or can't be) clustered when using scan or lookup on individual tracks, so this seems incorrect to me.

Sophist-UK:
They are equally as much unclustered as they are unmatched. But (in the absence of a UI which guides them) using the word clustered points novices to what they should try next. As I understand it, "How do I use  Picard?" is the most common question asked.

mwiencek:
> They are equally as much unclustered as they are unmatched.

You could just as well say they are equally "unscanned." And I wonder what % of users are tagging complete albums, because clustering isn't as useful otherwise.

Sophist-UK:
Not sure that it is still the case, but most users USED to rip Casanova so whole albums. 

And yes you could equally say "unscanned" but clustering is the first thing people should try so we should IMO lead them in that direction.

samj1912:
IMO we should not merge this. We will be rethinking the UI with 2.0 anyways. The above might cause some confusion and seems incorrect.

@zas @mineo thoughts?

Sophist-UK:
As I said previously, this renaming is more important now than later - once we have a better UI it may not be needed, but right now with the 1.4 UI it will help people press the Cluster button.

mineo:
I don't have a strong opinion for either name and also don't have any idea how many of our users use lookup directly, cluster first or scan directly. I fear however we phrase it, the name will be wrong for the other 2 use cases.

Sophist-UK:
If we are going to have a new UI for v2.0.0, then I agree with @samj1912 that we should close this unmerged.

samj1912:
Still of the opinion that we should not merge this. Rather open up a UX ticket on Jira and rethink some of labels we use in Picard currently.

Sophist-UK:
This is a very small change. Why not merge this and see whether it helps.

As and when we revamp the UI in Picard 2 we can think again about this.

PR #620, opened 2017-02-08
PICARD-969: Queue search dialog xmlws to run first

Sophist-UK:
If you are loading a lot of albums and have a long queue of releases being looked up, then the xmlws request for Search for similar albums... is queued behind them.

This commit runs these request as priority=True important=True which means that they will run as the very next xmlws.

https://tickets.metabrainz.org/browse/PICARD-969


Sophist-UK:

picard/webservice.py
```diff
@@ -484,30 +486,55 @@ def _find(self, entitytype, handler, kwargs):
...
-    def find_artists(self, handler, **kwargs):
-        return self._find('artist', handler, kwargs)
-
-    def _browse(self, entitytype, handler, kwargs, inc=[], priority=False, important=False):
+        return self.get(host, port, path, handler, queryargs=queryargs,
+                xml=xml, priority=priority, important=important, mblogin=mblogin,
+                cacheloadcontrol=cacheloadcontrol, refresh=refresh)
+
+    def find_releases(self, handler,
+            xml=True, priority=False, important=False, mblogin=False,
```
mineo: The repitition of keyword arguments in the new code looks unnecessary, why not just pass `**kwargs` through?

Sophist-UK:
Because the search terms are also kwargs so this distinguishes between keywords which are Picard internal and those that are part of the search term. 

But not wedded to this. If we still want it changed then it will have to wait for 3-4 weeks until I am back in the uk. Or someone else can change it for me.

picard/webservice.py
Sophist-UK: Because the search terms are also kwargs so this distinguishes between keywords which are Picard internal and those that are part of the search term.

@mineo I still believe that this helps readers distinguish between Picard internal kwargs and search kwargs, but I am not wedded to this.

Now back in the UK and if you still want this simplified after reading my justification, let me know and I will make the changes at some point next week.

mineo: Please do, thank you.

Sophist-UK:
Ok - ready for re-review.

picard/webservice.py
```diff
@@ -453,7 +453,9 @@ def lookup_discid(self, discid, handler, priority=True, important=True, refresh=
         return self._get_by_id('discid', discid, handler, inc, queryargs={"cdstubs": "no"},
                                priority=priority, important=important, refresh=refresh)
 
-    def _find(self, entitytype, handler, kwargs):
+    def _find(self, entitytype, handler, kwargs,
```
mineo: You could move the arguments that are only passed to `self.get` to kwargs here as well. Also, are there plugins that make use of this function? AFAICT all users set `priority` and `important` to `True`, so the defaults should reflect that.

Sophist-UK: I suspect that the issue causing the error may be that sometimes we have `**kwargs` and sometimes we have only `kwargs`. though I will have to test this.

I am not quite sure what you mean by:
> You could move the arguments that are only passed to self.get to kwargs here as well. Also, are there plugins that make use of this function? 

I think your point about defaults is well made - I will make this change as soon as I get time to test it.

Sophist-UK:
Ok - I think this is ready to merge now. Apologies for poor quality last time.

picard/webservice.py
Sophist-UK: I hope I have addressed all review comments - please let me know if I have missed any.

Sophist-UK:
Travis failure is because `Error: Failed to download resource "qt"` not because of any coding issue as far as I can tell.

picard/webservice.py
```diff
@@ -484,30 +486,35 @@ def _find(self, entitytype, handler, kwargs):
...
-        return self._find('recording', handler, kwargs)
+        return self._find('recording', handler, **kwargs)
 
     def find_artists(self, handler, **kwargs):
-        return self._find('artist', handler, kwargs)
+        return self._find('artist', handler, **kwargs)
 
-    def _browse(self, entitytype, handler, kwargs, inc=[], priority=False, important=False):
+    def _browse(self, entitytype, handler, inc=[],
+            xml=True, priority=False, important=False, mblogin=False,
```
mineo: One more thing: why is _browse now neither important nor has it priority? It's used for the other versions context menu item, which seems to be of similar importance as the search for similar tracks...

Sophist-UK: @mineo Thanks for the PR to the branch fr this in my repo. I will look at it more carefully as soon as I have time.

As for this comment - I agree and will change this at the same time as I review your PR.

Sophist-UK:
@mineo please re-review so that this can be merged for 1.4.1

mineo:
@Sophist-UK Just in case you're wondering what the status of this is: it looks good now and I'm ok with merging it, but I won't do it right now because using any of the internal search functions of picard immediately crashes it :( (reported as [PICARD-1051](https://tickets.metabrainz.org/browse/PICARD-1051)).

Sophist-UK:
It missed 1.4.1 so there is no rush.

Sophist-UK:
Don't forget to cherry-pick this to 1.4.x.

PR #621, opened 2017-02-09
PICARD-980: Fix various cover drop issues

antlarr:
This patch adds https support, so users can drag covers from amazon.

Also fixes the imgurl query handling so users can drag covers from
google images thumbnails (but large resolution images will be downloaded).

It also adds support for dropped octet-stream data, so chromium users
can drag "opened" images from google images even if the link actually
points to the url containing the image, since chromium adds the
image data to the drop event. Unfortunately, firefox doesn't do that,
so firefox users will have to drag thumbnails (but the same large
resolution image will be used in any case)

https://tickets.metabrainz.org/browse/PICARD-980


picard/ui/coverartbox.py
```diff
@@ -169,16 +177,29 @@ def fetch_remote_image(self, url):
...
+               if fallbackData[:2]=='\xff\xd8':
+                   mime = 'image/jpeg'
+               elif fallbackData[:8]=='\x89PNG\x0d\x0a\x1a\x0a':
+                   mime = 'image/png'
+               else:
+                   mime = None
+
+               if mime:
+                   log.warning("Trying the dropped %s data", mime)
+                   self.load_remote_image(url, mime, fallbackData)
```
samj1912: I don't think this should be a warning. Change the log level to info or debug instead?

picard/ui/coverartbox.py
```diff
@@ -33,7 +33,7 @@ class ActiveLabel(QtGui.QLabel):
     """Clickable QLabel."""
 
     clicked = QtCore.pyqtSignal()
-    imageDropped = QtCore.pyqtSignal(QtCore.QUrl)
+    imageDropped = QtCore.pyqtSignal(QtCore.QUrl, QtCore.QByteArray)
```
samj1912: Please use snake case for all variable names. See https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md

picard/ui/coverartbox.py
```diff
@@ -150,15 +154,19 @@ def open_release_page(self):
         lookup = self.tagger.get_file_lookup()
         lookup.albumLookup(self.release)
 
-    def fetch_remote_image(self, url):
+    def fetch_remote_image(self, url, fallbackData=None):
```
samj1912: Snake case again.

picard/ui/coverartbox.py
```diff
@@ -54,16 +54,20 @@ def mouseReleaseEvent(self, event):
...
+            if url.scheme() in ('https', 'http', 'file'):
                 event.acceptProposedAction()
                 break
 
     def dropEvent(self, event):
         accepted = False
+        # Chromium includes the actual data of the dragged image in the drop event. This
+        # is useful for Google Images, where the url links to the page that contains the image
+        # so we use it if the downloaded url is not an image.
+        droppedData = event.mimeData().data('application/octet-stream')
```
samj1912: Snake case for this too.

picard/ui/coverartbox.py
```diff
@@ -169,16 +177,29 @@ def fetch_remote_image(self, url):
...
             # This may be a google images result, try to get the URL which is encoded in the query
-            url = QtCore.QUrl(reply.url().queryItemValue("imgurl"))
+            url = QtCore.QUrl(url.queryItemValue("imgurl"))
             self.fetch_remote_image(url)
         else:
-            log.warning("Can't load image with MIME-Type %s", mime)
+            log.warning("Can't load remote image with MIME-Type %s", mime)
+            if fallbackData:
+               # Tests for image format obtained from file-magic
+               if fallbackData[:2]=='\xff\xd8':
```
samj1912: A startswith is more readable instead in this scenario?

samj1912:
Please make a relevant jira ticket at  https://tickets.metabrainz.org and edit the PR title with the issue number and link the ticket in the description.

picard/ui/coverartbox.py
```diff
@@ -169,16 +177,29 @@ def fetch_remote_image(self, url):
...
+                   mime = 'image/jpeg'
+               elif fallbackData[:8]=='\x89PNG\x0d\x0a\x1a\x0a':
+                   mime = 'image/png'
+               else:
+                   mime = None
+
+               if mime:
+                   log.warning("Trying the dropped %s data", mime)
```
samj1912: I don't think this should be a warning. Change the log level to info or debug instead?

picard/ui/coverartbox.py
antlarr: changed

picard/ui/coverartbox.py
antlarr: changed

picard/ui/coverartbox.py
antlarr: Note that fallbackData is a QByteArray object, so it doesn't have a startswith method, but startsWith. In fact, I was tempted to use startsWith but then thought using [:2] works for byte arrays and QByteArray, so it would be more generic and allow both kind of objects to be used in case it's neccesary in the future. So... Do you prefer me to change that to startsWith or should I keep it as it is ?

picard/ui/coverartbox.py
antlarr: changed to debug, indeed

picard/ui/coverartbox.py
antlarr: Sorry, I'm so used to Qt code and I saw camel case being used in this file a lot, so I didn't notice snake-case is preferred in picard. Btw, note that I didn't change the name of that variable, and that is actually a signal, so I'm not sure if it falls under the category of "pre-generated PyQt functions". Do you really think it should use snake-case ?

samj1912:
Can you please rebase the branch against master instead? The merge commit(2nd commit) is unrelated to the PR.

picard/ui/coverartbox.py
samj1912: Sorry, didn't notice.

https://github.com/metabrainz/picard/blob/master/picard/util/imageinfo.py#L72 
Looking at above, since there seems to be a precedent of using [:n], keep it as is.

picard/ui/coverartbox.py
samj1912: Yes, this seems to have gone under the radar since we recently solidified our Contribution Guide.

Look at https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L95

and https://github.com/metabrainz/picard/blob/master/picard/plugin.py#L220

We should switch this over to snake case too for uniformity :)

antlarr: In that case... fixed :)

antlarr:
Hehe, don't worry about nit-picking, I understand the importance of an homogeneous codebase :).

picard/ui/coverartbox.py
```diff
@@ -169,16 +177,29 @@ def fetch_remote_image(self, url):
...
             # This may be a google images result, try to get the URL which is encoded in the query
-            url = QtCore.QUrl(reply.url().queryItemValue("imgurl"))
+            url = QtCore.QUrl(url.queryItemValue("imgurl"))
             self.fetch_remote_image(url)
         else:
-            log.warning("Can't load image with MIME-Type %s", mime)
+            log.warning("Can't load remote image with MIME-Type %s", mime)
+            if fallback_data:
+               # Tests for image format obtained from file-magic
+               if fallback_data[:2]=='\xff\xd8':
```
mineo: We already have something similar to this in `picard.util.imageinfo.py::identify`

antlarr: Thanks, I didn't know that function. I added another commit to use it.

picard/ui/coverartbox.py
```diff
@@ -189,16 +189,13 @@ def on_remote_image_fetched(self, url, data, reply, error, fallback_data=None):
...
-               else:
-                   mime = None
-
-               if mime:
-                   log.debug("Trying the dropped %s data", mime)
+               try:
+                   mime = imageinfo.identify(fallback_data)[2]
+                   log.warning("Trying the dropped %s data", mime)
                    self.load_remote_image(url, mime, fallback_data)
+               except:
```
zas: This catchall except isn't great imho.
The error message should be more informative.

picard/ui/coverartbox.py
```diff
@@ -169,16 +177,26 @@ def fetch_remote_image(self, url):
...
+            url = QtCore.QUrl(url.queryItemValue("imgurl"))
             self.fetch_remote_image(url)
         else:
-            log.warning("Can't load image with MIME-Type %s", mime)
+            log.warning("Can't load remote image with MIME-Type %s", mime)
+            if fallback_data:
+               # Tests for image format obtained from file-magic
+               try:
+                   mime = imageinfo.identify(fallback_data)[2]
+                   log.warning("Trying the dropped %s data", mime)
```
samj1912: Not sure if this should be a warning log. Also where exactly do you expect the error to occur in these 3 lines? Maybe a try except else would be better?

picard/ui/coverartbox.py
```diff
@@ -169,16 +177,27 @@ def fetch_remote_image(self, url):
                     data = f.read()
                 self.load_remote_image(url, mime, data)
 
-    def on_remote_image_fetched(self, url, data, reply, error):
+    def on_remote_image_fetched(self, url, data, reply, error, fallback_data=None):
         mime = reply.header(QtNetwork.QNetworkRequest.ContentTypeHeader)
         if mime in ('image/jpeg', 'image/png'):
             self.load_remote_image(url, mime, data)
-        elif reply.url().hasQueryItem("imgurl"):
+        elif url.hasQueryItem("imgurl"):
```
mineo: I wonder if `reply.url()` was used here because it might be different from `url` in case of redirects?

antlarr: Probably, yes, but here it's used to test for Google Images' use of the imgurl query item, and that's lost in the redirect, so url() (and not reply.url() ) has to be used.

PR #623, opened 2017-02-12
PICARD-922: Fix inmulti behaviour with multi-values

antlarr:
inmulti arguments where evaluated before inmulti was executed, which
resulted in strings being passed as arguments, which breaks cases in which
the strings in multivalue variables contain the separator used to separate
multiple values (see PICARD-922).
This commit fixes the behaviour of inmulti so arguments are not evaluated
before they're passed, so it can extract the argument value as a list and
work with that.
This fixes https://tickets.metabrainz.org/browse/PICARD-922

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -387,9 +387,30 @@ def func_in(parser, text, needle):
         return ""
 
 
-def func_inmulti(parser, text, value, separator=MULTI_VALUED_JOINER):
-    """Splits ``text`` by ``separator``, and returns true if the resulting list contains ``value``."""
-    return func_in(parser, text.split(separator) if separator else [text], value)
+def func_inmulti(parser, haystack, needle, separator=MULTI_VALUED_JOINER):
+    """Searches for ``needle`` in ``haystack``, supporting a list variable for ``haystack``.
+       If a string is used instead, then a ``separator`` can be used to split it.
+       In both cases, it returns true if the resulting list contains ``needle``."""
+
+    needle = needle.eval(parser)
+    if type(text)==ScriptExpression and len(text)==1 and type(text[0])==ScriptVariable:
+        text=text[0]
+    if type(text)==ScriptVariable:
+        if text.name.startswith(u"_"):
+            name = u"~" + text.name[1:]
+        else:
+            name = text.name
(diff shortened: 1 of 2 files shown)


picard/script.py
```diff
@@ -387,9 +387,30 @@ def func_in(parser, text, needle):
...
-def func_inmulti(parser, text, value, separator=MULTI_VALUED_JOINER):
-    """Splits ``text`` by ``separator``, and returns true if the resulting list contains ``value``."""
-    return func_in(parser, text.split(separator) if separator else [text], value)
+def func_inmulti(parser, haystack, needle, separator=MULTI_VALUED_JOINER):
+    """Searches for ``needle`` in ``haystack``, supporting a list variable for ``haystack``.
+       If a string is used instead, then a ``separator`` can be used to split it.
+       In both cases, it returns true if the resulting list contains ``needle``."""
+
+    needle = needle.eval(parser)
+    if type(text)==ScriptExpression and len(text)==1 and type(text[0])==ScriptVariable:
```
samj1912: I would prefer using isinstance() here

picard/script.py
```diff
@@ -387,9 +387,30 @@ def func_in(parser, text, needle):
...
+        text=text[0]
+    if type(text)==ScriptVariable:
```
samj1912: Same here

picard/script.py
```diff
@@ -387,9 +387,30 @@ def func_in(parser, text, needle):
...
+        if text.name.startswith(u"_"):
+            name = u"~" + text.name[1:]
+        else:
+            name = text.name
+        values=parser.context.getall(name)
```
samj1912: PEP8, please put spaces around "="

picard/script.py
samj1912: Please run the code through a pep8 checker and fix any errors :)

picard/script.py
samj1912: text not defined? I think you mean haystack.

samj1912: Tests seem to be failing because of the above.

samj1912:
Please edit the PR title with PICARD-922 and add a link to jira :)

PR #624, opened 2017-02-14
Add Picard 1.4 release notes

samj1912:


Sophist-UK:
Was going to suggest that we add a list of known issues i.e. pending nerves, and suggest users keep a look out for 1.4.1.

PR #626, opened 2017-02-15
Revert "PICARD-922: Fix inmulti behaviour with multi-values"

zas:
Reverts metabrainz/picard#623


antlarr:
Just for the records, I'm the author of #623 and I agree. Please see #627

PR #628, opened 2017-02-15
PICARD-922: Add inmulti2 function to fix inmulti behaviour with multi-values

antlarr:
Add a inmulti2 function which works as inmulti but gets a variable name
as first parameter, instead of the contents of the variable

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -391,6 +391,16 @@ def func_inmulti(parser, text, value, separator=MULTI_VALUED_JOINER):
     """Splits ``text`` by ``separator``, and returns true if the resulting list contains ``value``."""
     return func_in(parser, text.split(separator) if separator else [text], value)
 
+def func_inmulti2(parser, haystack, needle):
+    """Searches for ``needle`` in the contents of the variable named
+       ``haystack`` . It returns true if the resulting list
+       contains ``needle``."""
+
+    if haystack.startswith("_"):
+        haystack = "~" + haystack[1:]
+    values = parser.context.getall(haystack)
+
+    return func_in(parser, values, needle)
 
 def func_rreplace(parser, text, old, new):
     return re.sub(old, new, text)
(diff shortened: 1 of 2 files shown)


Sophist-UK:
I am unclear why e cannot fix $in multi to work as it should. I assume it is an issue of  backwards compatibility but an explanation as to why we need to introduce an alternative version would be helpful.

samj1912:
@Sophist-UK yes, we decided to move to a new function to allow for backwards compatibility with userscripts.
We can include a comment in the picard scripting docs for in_multi2?

antlarr:
IMHO the problem is not a backwards compatibility issue, since it was broken and the previous PR made it work as the documentation said inmulti should work. Also, nobody should have ever expected it to return what it did for the broken cases mentioned in PICARD-922. The problem is the fix is quite hackish and so, it's much nicer (code quality-wise) to create an inmulti2 new function and recommend to use it, than fix inmulti and having to maintain ugly code (but that's just my thoughts).

Sophist-UK:
I can't understand why we cannot use existing inmulti name and name it work properly. 

We should be able to distinguish whether haystack is a single variable or a calculated string and act accordingly.

antlarr:
The ideal would be to have ScriptVariable return the correct type of a variable when it's evaluated, but afaik, that's not possible because in ```Metadata.__setitem__```  there's code that does:
```
            values = [values]
```
Thus effectively losing the information of the original type, so it's not possible to know if a variable was originally a string or a list.

Sophist-UK:
1. We should be able to distinguish between needle being a calculated string e.g. %v1%;%v2%

2. If we wanted to we can treat a list of len==1 as a string with ; separators and len>1 as a multi,  but personally I don't like this because it will mishandle multi-values which happen to have a single value which contains a ;

This should be both backwards compatible and

mineo:
Just fix this properly after 1.4.1 :-)

Sophist-UK:
This issues was previously fixed by #623 without introducing a new function and then reverted by #626 for an unknown reason.

Unless anyone can see a reason why this should be kept open, I recommend it is closed.

Sophist-UK:
IMO this should be closed since it has been superseded by merged #710

PR #631, opened 2017-02-19
PICARD-988: Extract file paths from NSURL while drag'n'drop

samj1912:
https://tickets.metabrainz.org/browse/PICARD-988

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -467,6 +467,7 @@ class BaseTreeView(QtGui.QTreeWidget):
         files = []
         new_files = []
         for url in urls:
+            log.debug("Adding the URL: %r", url)
             if url.scheme() == "file" or not url.scheme():
                 # Dropping a file from iTunes gives a filename with a NULL terminator
                 filename = os.path.normpath(os.path.realpath(unicode(url.toLocalFile()).rstrip("\0")))
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -32,6 +32,14 @@ from picard.plugin import ExtensionPoint
 from picard.ui.ratingwidget import RatingWidget
 from picard.ui.collectionmenu import CollectionMenu
 
+if sys.platform == 'darwin':
+    try:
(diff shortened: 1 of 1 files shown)


picard/ui/itemviews.py
```diff
@@ -467,9 +468,21 @@ def drop_urls(urls, target):
         files = []
         new_files = []
         for url in urls:
+            log.debug("Dropped the URL: %r", url.toString(QtCore.QUrl.RemoveUserInfo))
             if url.scheme() == "file" or not url.scheme():
-                # Dropping a file from iTunes gives a filename with a NULL terminator
-                filename = os.path.normpath(os.path.realpath(unicode(url.toLocalFile()).rstrip("\0")))
+                # Workaround for https://bugreports.qt.io/browse/QTBUG-40449
+                # OSX Urls follow the NSURL scheme and need to be converted
+                if sys.platform == 'darwin' and unicode(url.path()).startswith('/.file/id='):
```
zas: I would move the platform test and try/import stuff outside the main loop, it can be done once at module level, just use a flag and test for it here. We may need the same kind of workaround elsewhere.

picard/ui/itemviews.py
```diff
@@ -467,9 +475,16 @@ def drop_urls(urls, target):
         files = []
         new_files = []
         for url in urls:
+            log.debug("Dropped the URL: %r", url.toString(QtCore.QUrl.RemoveUserInfo))
             if url.scheme() == "file" or not url.scheme():
-                # Dropping a file from iTunes gives a filename with a NULL terminator
-                filename = os.path.normpath(os.path.realpath(unicode(url.toLocalFile()).rstrip("\0")))
+                # Workaround for https://bugreports.qt.io/browse/QTBUG-40449
+                # OSX Urls follow the NSURL scheme and need to be converted
+                if unicode(url.path()).startswith('/.file/id=') and NSURL_IMPORTED:
```
zas: Move test for `NSURL_IMPORTED` in first position in the `and` expression (faster on other platforms)

PR #634, opened 2017-02-20
PICARD-990: Avoid multiple value TXXX frames 

tungol:
Multiple values in TXXX frames are supported by Mutagen, but violate the official ID3 specification. This would allow all compliant ID3 implementations to read these tags, as opposed to just the ones violating the spec in this way. I believe this change will work, but I wasn't able to build picard to test it. Whether it's the best approach is another question. Related to https://tickets.metabrainz.org/browse/PICARD-990

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -175,6 +175,13 @@ class ID3File(File):
         'TPOS': re.compile(r'^(?P<discnumber>\d+)(?:/(?P<totaldiscs>\d+))?$')
     }
 
+    def build_TXXX(self, encoding, desc, values):
+        """Construct and return a TXXX frame."""
+        # This is here so that plugins can customize the behavior of TXXX
+        # frames in particular via subclassing.
+        # discussion: https://github.com/metabrainz/picard/pull/634
+        return id3.TXXX(encoding=encoding, desc=desc, text=values)
+
     def _load(self, filename):
         log.debug("Loading file %r", filename)
         file = self._get_file(encode_filename(filename))
@@ -365,7 +372,7 @@ class ID3File(File):
                     if frameid == 'WCOP':
                         # Only add WCOP if there is only one license URL, otherwise use TXXX:LICENSE
                         if len(values) > 1 or not valid_urls:
-                            tags.add(id3.TXXX(encoding=encoding, desc=self.__rtranslate_freetext[name], text=values))
+                            tags.add(self.build_TXXX(encoding, self.__rtranslate_freetext[name], values))
                         else:
                             tags.add(id3.WCOP(url=values[0]))
                     elif frameid == 'WOAR' and valid_urls:
@@ -374,7 +381,7 @@ class ID3File(File):
                 elif frameid.startswith('T'):
                     if config.setting['write_id3v23']:
                         if frameid == 'TMOO':
-                            tags.add(id3.TXXX(encoding=encoding, desc='mood', text=values))
+                            tags.add(self.build_TXXX(encoding, 'mood', values))
                     # No need to care about the TMOO tag being added again as it is
                     # automatically deleted by Mutagen if id2v23 is selected
                     tags.add(getattr(id3, frameid)(encoding=encoding, text=values))
@@ -385,18 +392,18 @@ class ID3File(File):
                     elif frameid == 'TSO2':
                         tags.delall('TXXX:ALBUMARTISTSORT')
             elif name in self.__rtranslate_freetext:
-                tags.add(id3.TXXX(encoding=encoding, desc=self.__rtranslate_freetext[name], text=values))
+                tags.add(self.build_TXXX(encoding, self.__rtranslate_freetext[name], values))
             elif name.startswith('~id3:'):
                 name = name[5:]
                 if name.startswith('TXXX:'):
-                    tags.add(id3.TXXX(encoding=encoding, desc=name[5:], text=values))
+                    tags.add(self.build_TXXX(encoding, name[5:], values))
                 else:
                     frameclass = getattr(id3, name[:4], None)
                     if frameclass:
                         tags.add(frameclass(encoding=encoding, text=values))
             # don't save private / already stored tags
             elif not name.startswith("~") and name not in self.__other_supported_tags:
-                tags.add(id3.TXXX(encoding=encoding, desc=name, text=values))
+                tags.add(self.build_TXXX(encoding, name, values))
 
         tags.add(tmcl)
         tags.add(tipl)


samj1912:
We need to investigate whether or not it breaks compatibility with taggers and which delimiter to use if we are switching to a string.

@lazka thoughts?

tungol:
I don't have strong feelings about the delimiter personally, but picked one to get the ball rolling. As I noted in the ticket, an alternate approach would be to create multiple TXXX frames with incremented names: TXXX:foo 1, TXXX:foo 2 or something like that, since TXXX frames can't have the same name. I could make a patch for that approach instead if there's a consensus for it.

Sophist-UK:
The premier that multi-values txxx games are not part of the spec is not completely true. 

AFAICR, the ID3v24 spec explicitly defines these whilst the ID3v24 spec requires them joined by "/". 

I am away from home at present so can you please quote or link the relevant v23 and v24 specs here for reference.

Sophist-UK:
P.S. We should forgery that there are many examples where de-facto standards trump formal ones. Use of some ID3v24 tags in v23 files,  misuse of date fields and single value "artist" tag are 3 that come immediately to mind. 

This seems to me to be something that really belongs in a plugin.

tungol:
Here's the relevant parts form the 2.4 spec. It defines "Text information frames" which do support multiple items but excludes TXXX from that definition, and then separately defines "User defined text information frame". From http://id3.org/id3v2.4.0-frames:

 4.2.   Text information frames

   The text information frames are often the most important frames,
   containing information like artist, album and more. There may only be
   one text information frame of its kind in an tag. All text
   information frames supports multiple strings, stored as a null
   separated list, where null is reperesented by the termination code
   for the charater encoding. All text frame identifiers begin with "T".
   Only text frame identifiers begin with "T", with the exception of the
   "TXXX" frame. All the text information frames have the following
   format:

     <Header for 'Text information frame', ID: "T000" - "TZZZ",
     excluding "TXXX" described in 4.2.6.>
     Text encoding                $xx
     Information                  <text string(s) according to encoding>

4.2.6.   User defined text information frame

   This frame is intended for one-string text information concerning the
   audio file in a similar way to the other "T"-frames. The frame body
   consists of a description of the string, represented as a terminated
   string, followed by the actual string. There may be more than one
   "TXXX" frame in each tag, but only one with the same description.

     <Header for 'User defined text information frame', ID: "TXXX">
     Text encoding     $xx
     Description       <text string according to encoding> $00 (00)
     Value             <text string according to encoding>

tungol:
Also, for me what seperates this from the other non-specification uses is that those change what's put in the string, but the end software is going to read the string just fine, maybe not interpret the value correctly. In this case, I discovered the issue because the implementation used by some software I was looking at follows the spec and only reads the first value string. I checked several other ID3 implementations and found that to be the case there as well. I think there's a difference between an informal standard that results in all the data getting through anyway versus one that doesn't.

lazka:
It's a violation of the spec and I can't find any discussion regarding it.

* taglib can read/write them
* foobar2000 can read/write them, but will join them with ";" if changed.

The way the id3 spec is structured makes it look like it was just an oversight.

Your call.

samj1912:
I think we can then merge this PR. Just 1 little update required. Can you convert the " / " as the seperator to the value config.setting['id3v23_join_with']

lazka:
> I checked several other ID3 implementations and found that to be the case there as well.

Which implementation supports multi value text frames but not for TXXX?

picard/formats/id3.py
```diff
@@ -365,7 +365,7 @@ def _save(self, filename, metadata):
                     if frameid == 'WCOP':
                         # Only add WCOP if there is only one license URL, otherwise use TXXX:LICENSE
                         if len(values) > 1 or not valid_urls:
-                            tags.add(id3.TXXX(encoding=encoding, desc=self.__rtranslate_freetext[name], text=values))
+                            tags.add(id3.TXXX(encoding=encoding, desc=self.__rtranslate_freetext[name], text=[' / '.join(values)]))
```
zas: Instead of repeating this 5 times, i would create a local function after line 283:

```python
def addTXXX(desc, values):
  tags.add(id3.TXXX(encoding=encoding, desc=desc, text=[' / '.join(values)]))

...
addTXXX(desc=self.__rtranslate_freetext[name], text=values)
```

The separator should be configurable per option imho.

tungol:
> Which implementation supports multi value text frames but not for TXXX?

The one that tripped me up is libid3tag. It's used by musicpd, which has specific support for MBID fields otherwise.

tungol:
Changes made per request: I created the addTXXX function with two arguments (desc and values) and one optional keyword argument (sep). sep defaults to config.setting['id3v23_join_with'] per @lazka's suggestion, which makes it doubly configurable - in the picard settings but can be overridden in the code if a situation is found where it makes sense to specify the separator.

I do have questions about the use of config.setting['id3v23_join_with'] though. I can't modify that setting in the Picard UI if id3 version is set to 2.4; is it still available for use like this in that case? Either way the UI needs a minor update there which I can add to the pull request if that's the consensus.

Alternatively, if we don't use that setting, is there any value in matching foobar's use of ";"?

samj1912:
@tungol  maybe we can change the option to "Join unsupported multi valued tags with"?

zas:
> Changes made per request: I created the addTXXX function with two arguments (desc and values) and one optional keyword argument (sep). sep defaults to config.setting['id3v23_join_with'] per @lazka's suggestion, which makes it doubly configurable - in the picard settings but can be overridden in the code if a situation is found where it makes sense to specify the separator.
> 
> I do have questions about the use of config.setting['id3v23_join_with'] though. I can't modify that setting in the Picard UI if id3 version is set to 2.4; is it still available for use like this in that case? Either way the UI needs a minor update there which I can add to the pull request if that's the consensus.
> 
> Alternatively, if we don't use that setting, is there any value in matching foobar's use of ";"?

Ok, we need to sort out this:
- is there any interest about having "id3v23_join_with" and "join_multivalued_tags_with" ?
- if we can replace "id3v23_join_with" with a less specific option it will require following changes:
  - modify UI so the new option is not dependent on id3 versions
  - add an upgrade hook using the value of "id3v23_join_with" to initnitalize new option value

tungol:
I'll take your lead on this, but I'm thinking it over and I don't see the usefulness of making the separator configurable. Having a separate setting complicates the UI and is one more thing a Picard user has to process, but using the same setting is difficult to describe in a way that isn't going to be confusing.

I dug back through the history to find more about how the id3v23_join_with setting happened so I could better understand the context. You all probably know this already since you were involved in the process, but for my own sake thinking out loud here, I see that it came from https://github.com/metabrainz/picard/pull/107 which improved data matching. There's not much discussion of the separator there or in the linked chat log, mostly it looks like there was no consensus on what to use, plus possible concern about pre-existing tags from elsewhere, so a config setting was easier.

Gets brought up again in [PICARD-585]( https://tickets.metabrainz.org/browse/PICARD-585) which raised the issue of removing the setting for compliance with the spec - again no consensus and some voiced support for non-compliant "; " specifically.

Next I had a look at what tags might actually be affected by this issue. First case in the code is TXXX:LICENSE, used only if non-URL or multiple otherwise WCOP. It's populated in Picard from URL relationships, so I'm not sure how you'd end up with a non-URL there (manual editing or pre-existing I guess?). Mostly going to be multiples then, but nothing else particular to be gleaned here.

Second case is TXXX:mood, which is 2.3 only and should definitely use id3v23_join_with.

Third case is __rtranslate_freetext. More than half of these are MB specific identifiers. Since Picard is the official musicbrainz tagger, I think for these there's value in picking something and sticking to it - we can set the standard, and it's better to have a standard than not.

Fourth case is tag names that start with "TXXX:". I'm not sure what might fall into that other than end-user entered manual tags which are hard to say much about.

Fifth case is __other_supported_tags: discnumber, tracknumber, totaldiscs, totaltracks. None of these should be multiples I don't think (???). Not relevant.

From which I take these points:

- the configuration setting is path of least resistance but also adds complexity for questionable value (imo)
- the majority of the tags affected by this are MB-specific and we can reasonably pick a standard for those
- people seem to like "; " in particular
- 2.3 tags should use id3v23_join_with, otherwise it would be really confusing

So my proposal is id3v23_join_with for 2.3 tags, and the non-configurable "; " for 2.4. That adds no new complexity while remaining consistent.

zas:
> So my proposal is id3v23_join_with for 2.3 tags, and the non-configurable "; " for 2.4. That adds no new complexity while remaining consistent.

Ok for me ;)

samj1912:
Okay for me too :)

picard/formats/id3.py
```diff
@@ -281,6 +281,12 @@ def _save(self, filename, metadata):
             tags.delall('APIC')
 
         encoding = {'utf-8': 3, 'utf-16': 1}.get(config.setting['id3v2_encoding'], 0)
+        
+        def addTXXX(desc, values, sep=config.setting['id3v23_join_with']):
```
samj1912: Can you rename this function to snake case? see https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#picard-specific-code

tungol:
So I've been looking into making this a plugin because while I've got the use case for this and is the official spec, I'm not sure it should be the default. Like @lazka said, it does seem like an oversight, and it seems that most of the modern implementations support multiple. And I'm also hesitant to say it should get a setting in the already overstuffed settings dialog.

That said, I can only see two possible ways to implement it with the current plugins system, neither of them satisfying:

1. I could use a track metadata processor to join certain tags before sending them off, but there's no way to get complete coverage of all possible TXXX frames this way.
2. I could register my own version of the affected file formats, if it even lets me overwrite them there? But this would involve a lot of duplicate code and need to be updated every time the official id3.py file gets updated, or else lose feature parity. That seems hacky and bad.

Going outside the current plugin API:

3. If we add the add_TXXX function where I can get at it, I could skip the plugins API and monkeypatch that function in a plugin, which seems also pretty hacky but the maintenance for the plugin would be a lot less

4. Alternatively I could add a new ExtensionPoint for something like register_plugin_setting, and then have the add_TXXX function check for a setting in there and switch behavior appropriately. That feels cleanest to me but I don't know how anyone else feels about the idea of hidden settings that way or making a new plugin API. Personally I think hidden but documented settings are a good compromise between the desire maximum configurability for pro users and the desire to keep the UI clean and just have good defaults for new users. But I don't think picard has any hidden settings like that already so I'd understand if it's not a popular idea.

5. A less clean but maybe less dramatic hidden setting option would be a variable in id3.py that the plugin could change.

I need more opinions to figure out what's best. is there any option I missed there? Which option would you all prefer I follow up on? I like 4 because it keeps it formally a plugin with no monkeypatching, but it's got downsides too.

Options 6 and 7 are forget the plugin and either add a setting or don't.

tungol:
Accidentally made a mess. Continued in https://github.com/metabrainz/picard/pull/635.

ijabz:
I find this change problematic for a number of reasons:

1.Multi string ID3v24 TXXX frames have become a defacto standard and are now quite well supported.
2. You break other aspects of the standards such as support for TCMP field
3. This is the sensible approach for multivalue fields such as artists and albumartists, and concerned that in Tungols comments he seems totally unaware of these fields.
4. By essentially changing from using a null separator to a ';' you break the logic for all taggers that do understand null separators, and change the logic across formats. So now a tagger such as Jaikoz would continue to show a Flac file containing multiple values for ARTISTS field as two values, but would show an MP3 version of file as one value
5. So now these taggers would have to have special option of treating a user defined value ';' as multi value separator, but only for some fields since you wouldn't want to do this for ARTIST (without the S), making it more difficult for users.
6. It doesn't fix taggers that dont understand multi value TXXX frames since now they will display the whole value but still wont understand they are multiple values.
7. Since there will never be an ID3 v2.5 specification there is no way to officlally become part of specification, but it is a clear improvement.

I request this change be reversed.

Sophist-UK:
This change hasn't been made - and so it cannot be reversed.

Besides which, the change in Picard #635 is on an enabler, with the change actually implemented in this plugin so every user would have a choice of whether to install it or not.

ijabz:
Phew, so nothing has changed for default, you have just added a plugin in hook right ?
I thought it had been implemented because it is listed as an issue resolved in the https://picard.musicbrainz.org/changelog/

PR #635, opened 2017-02-26
PICARD-990: Avoid multiple value TXXX frames

tungol:
Sorry, I was trying to squash the commits down and accidentally closed the old pull request, can't reopen it. Following on from https://github.com/metabrainz/picard/pull/634.

So I tested it out and if you're okay with making the TXXX function at the class level instead of a local function, I can subclass in a plugin, add a custom version, and then replace the mp3 format with my custom one. I've tested it locally and this works fine. I've got the plugin ready to go if you're happy with this.

https://tickets.metabrainz.org/browse/PICARD-990

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -175,6 +175,16 @@ class ID3File(File):
         'TPOS': re.compile(r'^(?P<discnumber>\d+)(?:/(?P<totaldiscs>\d+))?$')
     }
 
+    def build_TXXX(self, encoding, desc, values):
+        """Construct and return a TXXX frame."""
+        # This is here so that plugins can customize the behavior of TXXX
+        # frames in particular via subclassing.
+        # discussion: https://github.com/metabrainz/picard/pull/634
+        # discussion: https://github.com/metabrainz/picard/pull/635
+        # Used in the plugin "Compatible TXXX frames"
+        # PR: https://github.com/metabrainz/picard-plugins/pull/83
+        return id3.TXXX(encoding=encoding, desc=desc, text=values)
+
     def _load(self, filename):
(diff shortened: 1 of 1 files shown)


Sophist-UK:
I think that this is exactly the right approach. 

Can you submit the plugin and then add a comment with its name and PR to ťhe code in this PR?

tungol:
Okay, that's up at https://github.com/metabrainz/picard-plugins/pull/83

The one remaining thing that's a bit tenuous here is that I can't actually deregister the original MP3File class. The plugin works as I've written it only because "MP3FileCompatible" sorts after "MP3File", and the guess_formats function picks the last item from a sorted list. I can't see an obvious way to allow for deregistering default format modules, but maybe I haven't stared at it enough yet.

tungol:
Here's the ability to unregister a format, which lets me replace MP3File et al cleanly. 

Two new functions: unregister_format in picard.formats and a new method ExtensionPoint.unregister in picard.plugin. I can rewrite it as replace_format and/or ExtensionPont.replace if you'd prefer that. Remove is more flexible, replace makes it more obvious why you'd want to, I think.

zas:
> Here's the ability to unregister a format, which lets me replace MP3File et al cleanly.
> 
> Two new functions: unregister_format in picard.formats and a new method ExtensionPoint.unregister in picard.plugin. I can rewrite it as replace_format and/or ExtensionPont.replace if you'd prefer that. Remove is more flexible, replace makes it more obvious why you'd want to, I think.

Not very obvious how it relates to initial PR, but i think those changes should go in a different PR

tungol:
Backed it out.

PR #640, opened 2017-03-03
PICARD-995: Avoid error on OSX invalid file path

Sophist-UK:
Resolves a minor bug in #631.

If file path is invalid we need to skip the rest of the code in the for loop since filename is not set.

Ticket https://tickets.metabrainz.org/browse/PICARD-995.


samj1912:
LGTM

PR #642, opened 2017-03-05
PICARD-1000: Show album/cluster cover art

antlarr:
This PR adds support to show album/cluster cover art on the CoverArtBox.
I think it's better shown with an example, so...
When an album is looked up and the cover image has changed (slightly, but it has change) it shows up as http://i.imgur.com/M6dFXna.png
When one track is selected, and a different cover image is dropped for that track: http://i.imgur.com/W6xfS2L.png
Then when the user clicks on the album... http://i.imgur.com/ITX4e3g.png
Another file from another album (which has a different cover art) is added to track 9 of that album: http://i.imgur.com/jAt7ZLJ.png
And then another file is added to another track and other covers are dropped to other tracks on that album: http://i.imgur.com/XSig6BS.png

https://tickets.metabrainz.org/browse/PICARD-1000

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -557,6 +557,19 @@ class Album(DataObject, Item):
             self.tagger.albums[mbid] = self
             self.load(priority=True, refresh=True)
 
+    def update_metadata_images(self):
+        new_images = []
+        for track in self.tracks:
+            for file in list(track.linked_files):
+                for image in file.metadata.images:
+                    if image not in new_images:
+                        new_images.append(image)
+        for file in list(self.unmatched_files.files):
+            for image in file.metadata.images:
+                if image not in new_images:
+                    new_images.append(image)
+        self.metadata.images = new_images
+
 
 class NatAlbum(Album):
 
--- a/picard/album.py
+++ b/picard/album.py
@@ -58,6 +58,7 @@ class Album(DataObject, Item):
     def __init__(self, id, discid=None):
         DataObject.__init__(self, id)
         self.metadata = Metadata()
+        self.orig_metadata = Metadata()
         self.tracks = []
         self.loaded = False
         self.load_task = None
@@ -382,6 +383,10 @@ class Album(DataObject, Item):
     def _add_file(self, track, file):
         self._files += 1
         self.update(update_tracks=False)
+        # Fixme: The next lines can probably be moved to update()
+        for image in file.orig_metadata.images:
+            if image not in self.orig_metadata.images:
+                self.orig_metadata.append_image(image)
 
     def _remove_file(self, track, file):
         self._files -= 1
--- a/picard/album.py
+++ b/picard/album.py
@@ -561,11 +561,15 @@ class Album(DataObject, Item):
             for image in file.metadata.images:
                 if image not in new_images:
                     new_images.append(image)
-            for image in file.orig_metadata.images:
-                if image not in orig_images:
-                    orig_images.append(image)
+            try:
+                for image in file.orig_metadata.images:
+                    if image not in orig_images:
+                        orig_images.append(image)
+            except AttributeError:
+                pass
 
         for track in self.tracks:
+            process_file_images(track)
             for file in list(track.linked_files):
                 process_file_images(file)
         for file in list(self.unmatched_files.files):
--- a/picard/album.py
+++ b/picard/album.py
@@ -383,10 +383,6 @@ class Album(DataObject, Item):
     def _add_file(self, track, file):
         self._files += 1
         self.update(update_tracks=False)
-        # Fixme: The next lines can probably be moved to update()
-        for image in file.orig_metadata.images:
-            if image not in self.orig_metadata.images:
-                self.orig_metadata.append_image(image)
 
     def _remove_file(self, track, file):
         self._files -= 1
@@ -559,16 +555,24 @@ class Album(DataObject, Item):
 
     def update_metadata_images(self):
         new_images = []
+        orig_images = []
         for track in self.tracks:
             for file in list(track.linked_files):
                 for image in file.metadata.images:
                     if image not in new_images:
                         new_images.append(image)
+                for image in file.orig_metadata.images:
+                    if image not in orig_images:
+                        orig_images.append(image)
         for file in list(self.unmatched_files.files):
             for image in file.metadata.images:
                 if image not in new_images:
                     new_images.append(image)
+            for image in file.orig_metadata.images:
+                if image not in orig_images:
+                    orig_images.append(image)
         self.metadata.images = new_images
+        self.orig_metadata.images = orig_images
 
 
 class NatAlbum(Album):
--- a/picard/album.py
+++ b/picard/album.py
@@ -556,21 +556,21 @@ class Album(DataObject, Item):
     def update_metadata_images(self):
         new_images = []
         orig_images = []
-        for track in self.tracks:
-            for file in list(track.linked_files):
-                for image in file.metadata.images:
-                    if image not in new_images:
-                        new_images.append(image)
-                for image in file.orig_metadata.images:
-                    if image not in orig_images:
-                        orig_images.append(image)
-        for file in list(self.unmatched_files.files):
+
+        def process_file_images(file):
             for image in file.metadata.images:
                 if image not in new_images:
                     new_images.append(image)
             for image in file.orig_metadata.images:
                 if image not in orig_images:
                     orig_images.append(image)
+
+        for track in self.tracks:
+            for file in list(track.linked_files):
+                process_file_images(file)
+        for file in list(self.unmatched_files.files):
+            process_file_images(file)
+
         self.metadata.images = new_images
         self.orig_metadata.images = orig_images
 
--- a/picard/album.py
+++ b/picard/album.py
@@ -72,6 +72,7 @@ class Album(DataObject, Item):
         self.errors = []
         self.status = None
         self._album_artists = []
+        self.update_metadata_images_enabled = True
 
     def __repr__(self):
         return '<Album %s %r>' % (self.id, self.metadata[u"album"])
@@ -255,6 +256,7 @@ class Album(DataObject, Item):
             self._tracks_loaded = True
 
         if not self._requests:
+            self.update_metadata_images_enabled = False
             # Prepare parser for user's script
             if config.setting["enable_tagger_scripts"]:
                 for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"]:
@@ -276,6 +278,7 @@ class Album(DataObject, Item):
                         self._new_metadata.strip_whitespace()
 
             for track in self.tracks:
+                track.metadata_images_changed.connect(self.update_metadata_images)
                 for file in list(track.linked_files):
                     file.move(self.unmatched_files)
             self.metadata = self._new_metadata
@@ -285,6 +288,7 @@ class Album(DataObject, Item):
             self.loaded = True
             self.status = None
             self.match_files(self.unmatched_files.files)
+            self.update_metadata_images_enabled = True
             self.update()
             self.tagger.window.set_statusbar_message(
                 N_('Album %(id)s loaded: %(artist)s - %(album)s'),
@@ -379,14 +383,19 @@ class Album(DataObject, Item):
     def update(self, update_tracks=True):
         if self.item:
             self.item.update(update_tracks)
+        self.update_metadata_images()
 
     def _add_file(self, track, file):
         self._files += 1
         self.update(update_tracks=False)
+        file.metadata_images_changed.connect(self.update_metadata_images)
+        self.update_metadata_images()
 
     def _remove_file(self, track, file):
         self._files -= 1
         self.update(update_tracks=False)
+        file.metadata_images_changed.disconnect(self.update_metadata_images)
+        self.update_metadata_images()
 
     def match_files(self, files, use_recordingid=True):
         """Match files to tracks on this album, based on metadata similarity or recordingid."""
@@ -554,6 +563,8 @@ class Album(DataObject, Item):
             self.load(priority=True, refresh=True)
 
     def update_metadata_images(self):
+        if not self.update_metadata_images_enabled:
+            return
         new_images = []
         orig_images = []
 
--- a/picard/album.py
+++ b/picard/album.py
@@ -557,23 +557,23 @@ class Album(DataObject, Item):
         new_images = []
         orig_images = []
 
-        def process_file_images(file):
-            for image in file.metadata.images:
+        def process_images(obj):
+            for image in obj.metadata.images:
                 if image not in new_images:
                     new_images.append(image)
             try:
-                for image in file.orig_metadata.images:
+                for image in obj.orig_metadata.images:
                     if image not in orig_images:
                         orig_images.append(image)
             except AttributeError:
                 pass
 
         for track in self.tracks:
-            process_file_images(track)
+            process_images(track)
             for file in list(track.linked_files):
-                process_file_images(file)
+                process_images(file)
         for file in list(self.unmatched_files.files):
-            process_file_images(file)
+            process_images(file)
 
         self.metadata.images = new_images
         self.orig_metadata.images = orig_images
--- a/picard/album.py
+++ b/picard/album.py
@@ -85,6 +85,11 @@ class Album(DataObject, Item):
             for file in self.unmatched_files.iterfiles():
                 yield file
 
+    def enable_update_metadata_images(self, enabled):
+        self.update_metadata_images_enabled = enabled
+        if enabled:
+            self.update_metadata_images()
+
     def append_album_artist(self, id):
         """Append artist id to the list of album artists
         and return an AlbumArtist instance"""
@@ -256,7 +261,7 @@ class Album(DataObject, Item):
             self._tracks_loaded = True
 
         if not self._requests:
-            self.update_metadata_images_enabled = False
+            self.enable_update_metadata_images(False)
             # Prepare parser for user's script
             if config.setting["enable_tagger_scripts"]:
                 for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"]:
@@ -288,8 +293,8 @@ class Album(DataObject, Item):
             self.loaded = True
             self.status = None
             self.match_files(self.unmatched_files.files)
-            self.update_metadata_images_enabled = True
             self.update()
+            self.enable_update_metadata_images(True)
             self.tagger.window.set_statusbar_message(
                 N_('Album %(id)s loaded: %(artist)s - %(album)s'),
                 {
--- a/picard/album.py
+++ b/picard/album.py
@@ -570,29 +570,48 @@ class Album(DataObject, Item):
     def update_metadata_images(self):
         if not self.update_metadata_images_enabled:
             return
-        new_images = []
-        orig_images = []
 
-        def process_images(obj):
-            for image in obj.metadata.images:
-                if image not in new_images:
-                    new_images.append(image)
-            try:
-                for image in obj.orig_metadata.images:
-                    if image not in orig_images:
-                        orig_images.append(image)
-            except AttributeError:
-                pass
+        class State:
+            new_images = []
+            orig_images = []
+            has_common_new_images = True
+            has_common_orig_images = True
+            first_new_obj = True
+            first_orig_obj = True
+
+        state = State()
+
+        def process_images(state, obj):
+            # Check new images
+            if state.first_new_obj:
+                state.new_images = obj.metadata.images[:]
+                state.first_new_obj = False
+            else:
+                if state.new_images != obj.metadata.images:
+                    state.has_common_new_images = False
+                    state.new_images.extend([image for image in obj.metadata.images if image not in state.new_images])
+            if isinstance(obj, Track):
+                return
+            # Check orig images, but not for Tracks (which don't have orig_metadata)
+            if state.first_orig_obj:
+                state.orig_images = obj.orig_metadata.images[:]
+                state.first_orig_obj = False
+            else:
+                if state.orig_images != obj.orig_metadata.images:
+                    state.has_common_orig_images = False
+                    state.orig_images.extend([image for image in obj.orig_metadata.images if image not in state.orig_images])
 
         for track in self.tracks:
-            process_images(track)
+            process_images(state, track)
             for file in list(track.linked_files):
-                process_images(file)
+                process_images(state, file)
         for file in list(self.unmatched_files.files):
-            process_images(file)
+            process_images(state, file)
 
-        self.metadata.images = new_images
-        self.orig_metadata.images = orig_images
+        self.metadata.images = state.new_images
+        self.metadata.has_common_images = state.has_common_new_images
+        self.orig_metadata.images = state.orig_images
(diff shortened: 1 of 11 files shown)


picard/album.py
```diff
@@ -552,6 +553,27 @@ def switch_release_version(self, mbid):
             self.tagger.albums[mbid] = self
             self.load(priority=True, refresh=True)
 
+    def update_metadata_images(self):
+        new_images = []
+        orig_images = []
+        for track in self.tracks:
+            for file in list(track.linked_files):
+                for image in file.metadata.images:
```
samj1912: This same for loop has been repeated 4 times, maybe you can make a private function and call it instead?

picard/ui/coverartbox.py
```diff
@@ -103,35 +117,49 @@ def set_data(self, data, force=False, pixmap=None):
...
+                    thumb.loadFromData(image.data)
+                    thumb = self.decorate_cover(thumb)
+                    painter.drawPixmap(x - thumb.width() / 2, y - thumb.height() / 2, thumb)
+                    x += displacements
+                    y += displacements
                 painter.end()
-        self.setPixmap(cover)
+                pixmap = pixmap.scaled(w, h, QtCore.Qt.KeepAspectRatio, QtCore.Qt.SmoothTransformation)
+                self.setPixmap(pixmap)
+                pixmap = None
```
samj1912: Why are we setting pixmap to None here?

IHMO we can move the if clause below to the top and move the above block to an else.

picard/ui/coverartbox.py
```diff
@@ -103,35 +117,49 @@ def set_data(self, data, force=False, pixmap=None):
...
         data = None
         self.related_images = []
         if metadata and metadata.images:
             self.related_images = metadata.images
-            for image in metadata.images:
-                if image.is_front_image():
-                    data = image
-                    break
-            else:
+            log.debug("%s using images:" % (self.name), metadata.images)
```
samj1912: If we are adding a debug log, maybe an additional parameter display the display string of the related object will be more beneficial than just new cover art and original coverart.

Also the above changes on each selection update, it might clutter the log a bit too much.

samj1912: On that, maybe we can cache the pixmap. Since it is updated each time the selection is updated, it can lead to too much wasted processing. Especially since a call to self.data[i].data involves I/O.

picard/ui/coverartbox.py
```diff
@@ -94,6 +95,19 @@ def __eq__(self, other):
     def show(self):
         self.set_data(self.data, True)
 
+    def decorate_cover(self, pixmap):
+        offx, offy, w, h = (1, 1, 121, 121)
```
zas: Just an unrelated note: I wonder about the future of this (hardcoded values), especially since we want to add high dpi compatibility in 2.x releases.

picard/ui/coverartbox.py
```diff
@@ -255,28 +283,29 @@ def load_remote_image(self, url, mime, data):
...
+                types=[u'front'],
                 data=data
             )
         except CoverArtImageError as e:
             log.warning("Can't load image: %s" % unicode(e))
             return
         if isinstance(self.item, Album):
             album = self.item
-            album.metadata.append_image(coverartimage)
+            album.metadata.set_front_image(coverartimage)
```
zas: Can you explain this ?

I'm low on coffee, but it seems to `set_front_image()` is actually **removing** other images having is_front_image set to True. How does this work when a release has multiple front images (currently only possible with CAA provider).

picard/ui/coverartbox.py
mineo: This line also doesn't work because not all arguments to log.debug are converted during string formatting.

picard/ui/coverartbox.py
antlarr: Sure, when an user drags an image and drops it on the cover art box, I think the natural thing he/she wants to do is to replace the cover art with something different, not add another image to a list of images. Maybe in the context menu (check my cover-art-context-menu branch) I can add a "radio button" where user can select between "Append images on drop" or "Replace images on drop". Would that be better? Still, I think the default should be "replace" since it's more intuitive.

picard/ui/coverartbox.py
antlarr: I will just remove that line. As samj1912 said, that would clutter the log too much. I just added while I was working on that and thought it would be ok to keep it, but it doesn't really give any information once this works as expected.

picard/ui/coverartbox.py
antlarr: Instead of radio buttons in the context menu, maybe an option in the "user interface" section of the options dialog would be better.

antlarr: I implemented the radio buttons on the context menu not to clutter the options dialog with a new option, so this will be addressed by this branch:
https://github.com/antlarr/picard/compare/album-cover-art...antlarr:cover-art-context-menu?expand=1
You can see the result here: http://i.imgur.com/lyI9SNs.png
I'll create the pull request for the cover-art-context-menu branch as soon as this one is merged (the work there is already finished) since the commits in that branch depend on this PR.

zas: But, depending on chosen option, the other front images should be kept imho. What about either prepending dropped image, or replacing only the first one ? And keep the rest as is.

picard/ui/coverartbox.py
```diff
@@ -28,6 +28,36 @@
 from picard.util import encode_filename, imageinfo
 
 
+class LRUCache(dict):
```
samj1912: Since this can be used by other modules too, mind moving this to a new utils.cache file?

picard/ui/coverartbox.py
```diff
@@ -74,16 +104,18 @@ def dropEvent(self, event):
 
 class CoverArtThumbnail(ActiveLabel):
 
-    def __init__(self, active=False, drops=False, *args, **kwargs):
+    def __init__(self, active=False, drops=False, name=None, pixmap_cache=None, *args, **kwargs):
         super(CoverArtThumbnail, self).__init__(active, drops, *args, **kwargs)
         self.data = None
+        self.name = name
```
samj1912: Is there any use for the name attribute?

picard/ui/coverartbox.py
antlarr: sure

picard/ui/coverartbox.py
antlarr: you're right. Not anymore, so I'll remove it

picard/ui/coverartbox.py
```diff
@@ -94,44 +95,70 @@ def __eq__(self, other):
...
-                painter = QtGui.QPainter(cover)
-                bgcolor = QtGui.QColor.fromRgb(0, 0, 0, 128)
-                painter.fillRect(QtCore.QRectF(offx, offy, w, h), bgcolor)
-                x = offx + (w - pixmap.width()) / 2
-                y = offy + (h - pixmap.height()) / 2
-                painter.drawPixmap(x, y, pixmap)
+                pixmap.loadFromData(self.data[0].data)
+                pixmap = self.decorate_cover(pixmap)
+            else:
+                stack_width, stack_height = (w + displacements * (len(self.data) - 1), h + displacements * (len(self.data) - 1))
```
zas: Please change this to:

```python
offset = displacements * (len(self.data) - 1)
stack_width, stack_height = (w + offset, h + offset)
```

picard/util/lrucache.py
```diff
@@ -0,0 +1,79 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""
```
mineo: This comment block should be on the class itself, not somewhere before it.

picard/ui/coverartbox.py
antlarr: done

picard/metadata.py
```diff
@@ -55,6 +55,11 @@ def __nonzero__(self):
     def append_image(self, coverartimage):
         self.images.append(coverartimage)
 
+    def set_front_image(self, coverartimage):
+        # First remove all front images
+        self.images[:] = [ img for img in self.images if not img.is_front_image() ]
```
zas: Please remove spaces at start and end of list comprehension:

```python
self.images[:] = [img for img in self.images if not img.is_front_image()]
```

picard/util/lrucache.py
```diff
@@ -0,0 +1,79 @@
...
+  File "<stdin>", line 1, in <module>
+  File "lrucache.py", line 48, in __getitem__
+    return super(LRUCache, self).__getitem__(key)
+KeyError: 'item3'
+>>> cache['item1']
+'some value'
+"""
+
+
+class LRUCache(dict):
```
mineo: PyQt provides a [QPixmapCache](http://pyqt.sourceforge.net/Docs/PyQt4/qpixmapcache.html) class, is that something we could use instead of rolling our own? (If we need an lru cache in the future, we can use the one in functools if that future involves python 3).

picard/metadata.py
antlarr: fixed

picard/util/lrucache.py
samj1912: @antlarr I didn't know about this but it seems like the way to go. It has all the functionality we need.

picard/album.py
```diff
@@ -552,6 +553,31 @@ def switch_release_version(self, mbid):
...
+
+        def process_file_images(file):
+            for image in file.metadata.images:
+                if image not in new_images:
+                    new_images.append(image)
+            try:
+                for image in file.orig_metadata.images:
+                    if image not in orig_images:
+                        orig_images.append(image)
+            except AttributeError:
```
mineo: Under which circumstances does this raise an AttributeError? Assuming that `file` is always a `File`, it will always have `orig_metadata` and that will always have `images`.

picard/album.py
```diff
@@ -552,6 +553,31 @@ def switch_release_version(self, mbid):
...
+                pass
+
+        for track in self.tracks:
+            process_file_images(track)
+            for file in list(track.linked_files):
```
mineo: Is `linked_files` ever not a list?

picard/album.py
```diff
@@ -552,6 +553,31 @@ def switch_release_version(self, mbid):
...
+                process_file_images(file)
+        for file in list(self.unmatched_files.files):
```
mineo: Is `unmatched_files.files` ever not a list?

picard/coverart/image.py
```diff
@@ -70,6 +70,9 @@ def __init__(self, data, prefix='picard', suffix=''):
     def __eq__(self, other):
         return self._hash == other._hash
 
+    def hash(self):
```
mineo: Why isn't this `__hash__`?

picard/ui/infodialog.py
```diff
@@ -296,6 +297,20 @@ def _display_info_tab(self):
             tabWidget.setTabText(tab_index, _("&Info"))
             self.tab_hide(tab)
 
+class TrackInfoDialog(InfoDialog):
```
mineo: Maybe something's wrong on my end, but it doesn't seem to be possible to access this dialog from the Info button in the toolbar or the context menu of tracks without files?

picard/util/lrucache.py
samj1912: Update: See https://chatlogs.metabrainz.org/brainzbot/metabrainz/2017-03-07/?msg=3848492&page=5

picard/ui/mainwindow.py
```diff
@@ -944,6 +948,10 @@ def update_selection(self, objects=None):
                         }
                     self.set_statusbar_message(msg, mparms, echo=None,
                                                history=None)
+            elif isinstance(obj, Album):
+                obj.update_metadata_images()
```
mineo: Does this really need to be called every time an album is selected? It really feels like this should be done after the images on the metadata objects are updated.

picard/ui/infodialog.py
```diff
@@ -101,7 +102,7 @@ def __init__(self, obj, parent=None):
         self.ui = Ui_InfoDialog()
         self.display_existing_artwork = False
         if isinstance(obj, File) and isinstance(obj.parent, Track) or \
-                isinstance(obj, Track):
+                isinstance(obj, Track) or isinstance(obj, Album):
```
samj1912: Better to use isinstance(obj, (File, Track, Album))

antlarr: Hmmm, I can merge Track and Album but isinstance(obj, File) has an "and" with isinstance(obj.parent, Track), so I would leave it as:
```
if isinstance(obj, File) and isinstance(obj.parent, Track) or \
        isinstance(obj, (Track, Album)):
```
Deal?

samj1912: Remove the "\" for a pair of parentheses and you got it! :P

picard/album.py
antlarr: file is not always a File, but can be a Track too, and tracks don't have orig_metadata. Should I change that to "obj" ?

picard/album.py
antlarr: I had a reason for those, but I forgot... so ok, I'll change it (and the other case)

picard/coverart/image.py
antlarr: Because IMHO that function doesn't return the hash of that object, but the hash of the CoverArtImage stored in that DataHash object.

picard/ui/infodialog.py
antlarr: hmm, I tested it with zas' example release with 3 images (e5e80fb7-f6b7-3c51-9183-5af44a6c6195) and there it works, but you're right, for single image albums it's not accesible. I'll try to fix that

picard/ui/mainwindow.py
antlarr: I didn't find any way to call an album's update_metadata_images when the images are changed in any track/file related to that album. Do you have any suggestion?

picard/album.py
antlarr: In fact, there are other previous cases in album.py where linked_files is converted to a list, so I'm not so sure that's not needed

picard/ui/mainwindow.py
samj1912: We can emit a signal on new CA images which can then call the update_metadata_images

picard/album.py
zas: If it isn't always a File, yes, better change it to `obj`.

picard/ui/mainwindow.py
antlarr: Hmm, the parent of a file is the track, not the album, right? I think I prefer your previous solution (emitting a signal from track/files and have the album's update_metadata_images called). Would that be ok?

samj1912: Works :)

antlarr:
> The order in which the covers are stacked seems wrong: the frontmost is the cover of the last track that does not have the same cover as the first track. I would have expected the covers to follow the track order, with the frontmost one being the cover of the first track.

It's strange but for me, the current implementation follows track order. Imagine that you have the track's covers in your hands. First, the first track's cover is laid down, then, the second track's cover is laid down (thus over the first one), and so on. Can someone else vote here which one looks more natural for you?

In any case, if everyone else votes to have the first track at the top, I would arrange the stack from bottom-left to top-right instead of the current top-left to bottom-right, so at least the tracks order is somehow "left to right". Although that's probably cultural and some countries with other writing systems would prefer something else :(.

> One other thing that I really don't like is how this album-level view of the cover art is different from the album-level view of track metadata: the latter one shows (different across x items), even if x-1 items have the same value for a tag, the cover art now looks as if all tracks of an album have the same y pictures. Is this the direction we want to go?

Do you have a proposal?

antlarr:
Please review again

picard/ui/infodialog.py
antlarr: This should be fixed now

zas:
> 
> It's strange but for me, the current implementation follows track order. Imagine that you have the track's covers in your hands. First, the first track's cover is laid down, then, the second track's cover is laid down (thus over the first one), and so on. Can someone else vote here which one looks more natural for you?
> 


First track cover image on top, second under it, etc....

> 
> I guess you have a point there. Would it be ok if we show the album cover art only when all tracks have the same front images and if any is different we show one of the following options?
> 
>     The default disc shadow with a red tint
>     A split screen-like pixmap of the different covers (as opposed to the current stack of covers)
>     The current stack with some symbol in one of the corners
> 
> I think I prefer 1 or 3.

Adding symbols is a bad idea imho (it complicates things, needs explanations, etc...). So i'll not go for 3.

When all tracks share the same image, just display this image for the release.

When one or more tracks have different images, display the one shared among most tracks or the first one on top, plus a special shadow showing there are more images (like a stack of images).

antlarr:
When the tracks/files have no common images, a yellow glow (same color as used in the MetadataBox)  is added to the front image: http://i.imgur.com/aXF4lUO.png
When there are four images to show, they are shown normally (in this case, with a glow, because there are different images in different tracks): http://i.imgur.com/RU3M4B0.png
When there are five (or more) images to show, only 3 are shown and a "grey stack" is added to represent "there are more": http://i.imgur.com/5TlkS2o.png
I think that should satisfy everyone

picard/ui/coverartbox.py
```diff
@@ -94,51 +95,119 @@ def __eq__(self, other):
...
         if metadata:
             release = metadata.get("musicbrainz_albumid", None)
         if release:
             self.setActive(True)
-            self.setToolTip(_(u"View release on MusicBrainz"))
+            text = _(u"View release on MusicBrainz")
+            if hasattr(metadata, 'has_common_images'):
+                text += '<br />'
+                if has_common_images:
+                    text += _(u'<i>Common images on all tracks</i>')
```
zas: I would exclude tags from translatable string here:

```python
if has_common_images:
    note =  _(u'Common images on all tracks')
else:
    note = _(u'Tracks contain different images')
text += '<br /><i>%s</i>' % note
```

picard/ui/infodialog.py
```diff
@@ -100,10 +101,13 @@ def __init__(self, obj, parent=None):
         self.obj = obj
         self.ui = Ui_InfoDialog()
         self.display_existing_artwork = False
-        if isinstance(obj, File) and isinstance(obj.parent, Track) or \
-                isinstance(obj, Track):
+        if (isinstance(obj, File) and isinstance(obj.parent, Track) or
+                isinstance(obj, Track)):
```
zas: While at it, move operators in front:

```python
if (isinstance(obj, File)
    and isinstance(obj.parent, Track)
    or isinstance(obj, Track)):
```

picard/ui/coverartbox.py
```diff
@@ -94,51 +95,119 @@ def __eq__(self, other):
...
+                    painter.drawPixmap(x, y, thumb)
+                    cx -= displacements
+                    cy += displacements
+                if not has_common_images:
+                    color = QtGui.QColor("darkgoldenrod")
+                    border_length = 10
+                    for k in range(border_length):
+                        color.setAlpha(255 - k * 255 / border_length)
+                        painter.setPen(color)
+                        painter.drawLine(x, y - k - 1, x + 121 + k + 1, y - k - 1)
```
antlarr: AFAIK, there isn't any method in Qt for that. I think there may be methods in kde frameworks in some style library like breeze, but that's definitely of no use here. About high dpi scaling, that's "not supported" on Qt4, Qt5 brings support for hidpi devices from Qt 5.6.0, but that's definitely something to worry about with picard 2.0 . In the meantime, since the coverartbox is fixed to use 128x128 covers, that means users with hidpi devices will see small covers and small cover stacks (which Qt5 will scale automatically)

picard/ui/coverartbox.py
```diff
@@ -94,51 +95,119 @@ def __eq__(self, other):
...
-                painter = QtGui.QPainter(cover)
-                bgcolor = QtGui.QColor.fromRgb(0, 0, 0, 128)
-                painter.fillRect(QtCore.QRectF(offx, offy, w, h), bgcolor)
-                x = offx + (w - pixmap.width()) / 2
-                y = offy + (h - pixmap.height()) / 2
-                painter.drawPixmap(x, y, pixmap)
+                pixmap.loadFromData(self.data[0].data)
+                pixmap = self.decorate_cover(pixmap)
+            else:
+                limited = len(self.data) > 4
```
antlarr: ok

picard/ui/coverartbox.py
antlarr: ok

picard/ui/infodialog.py
antlarr: ok

picard/ui/coverartbox.py
```diff
@@ -73,6 +73,7 @@ def dropEvent(self, event):
 
 
 class CoverArtThumbnail(ActiveLabel):
+    MAX_COVERS_TO_STACK = 4
```
samj1912: Move this to picard.const

samj1912:
Please don't include any merge commits in a PR. Always rebase against the master :)

picard/ui/coverartbox.py
antlarr: Sorry, I didn't know about picard.const. I moved it there.

picard/ui/coverartbox.py
antlarr: With respect to the drawing, I tried saving an image for the glow under the resources directory and just loading/painting it. Since the resolution is fixed, it's doable, but I found my drawing skills suck, so I ended up making the glow in code, mainly because we can experiment with making it bigger, changing color or whatever. But if you think it's better to stick to that glow, it's also possible to store the glow and the cover stack as images and just use them as is done with CoverArtShadow.png . Only note that I would still use code to generate the images instead of having a svg image as source. Since there are not many cases (afaik) with so many cover art images and anyway the generated images are cached, I'm not sure it's worth to change that.

zas:
> since the coverartbox is fixed to use 128x128 covers, that means users with hidpi devices will see small covers and small cover stacks (which Qt5 will scale automatically)

That's exactly my point, current cover art box code (and the one you are adding) are far from being "scalable", with all those fixed numbers, so a complete rewrite will be needed.

To make it clear: 128x128 on a 4k screen is a small icon, not a thumbnail, scaling it up will increase the size but quality will be awful, since we have 500x500 pixels images from CAA we could at least use those without the downscale/upscale process.

I was just wondering if we could avoid to rewrite the code here in one month...

antlarr:
Yes, that has to be rewritten anyway for Qt5 (just like the previous code in there had to be rewritten too) as everything not managed by layouts has to use "device-independent pixels". I'll try to help with that once the port to Qt5 starts.

zas:
@antlarr : please fix conflicts and sort out this PICARD-1000 / description / whatever @mineo reported
@mineo : are changes you requested done ? If yes, please approve

antlarr:
I think the only thing reported by mineo I didn't address is that DataHash.hash should be __hash__, but I was waiting for him to agree or disagree with my opinion that it doesn't feel ok that hash(image) would be equal to hash(image.datahash). But if it's necessary, I can of course change that.

PR #644, opened 2017-03-06
PICARD-998: Fix toolbar toggle action

samj1912:
https://tickets.metabrainz.org/browse/PICARD-998


samj1912:
The toggle action toolbar button wasn't working because it was binded to toggle view action of the initial toolbar. Since they are now dynamic, we need to bind the toggle to a function call.

mineo:
Thaanks for that comment, the commit didn't make sense to me before it. Maybe put it in the commit message next time?;)

PR #647, opened 2017-03-08
PICARD-1006: Fix Drag'n'Drops for cover art drops

samj1912:
https://tickets.metabrainz.org/browse/PICARD-1006


samj1912:
Although we had fixed the drag and drop for file drops in the main window, a similar problem existed when dropping covers. So I made a function for the same in picard.utils since it be used in the future.

PR #648, opened 2017-03-08
Fix circular imports in picard.util

samj1912:
Fixes crash due to #647


mineo:
Why?

samj1912:
Some circular dependencies causing a crash due to picard log not being imported.

mineo:
Well, that would be good information for the commit message ;-)

samj1912:
Updated PR and commit :)

Sorry for the hiccup!

PR #654, opened 2017-03-12
PICARD-1012: Change QToolButtons to QPushButtons for GUI uniformity

samj1912:
QToolButtons in User Interface and Scripting looked smalled in comparison
to the rest of the UI.

See: https://tickets.metabrainz.org/browse/PICARD-1012


mineo:
The up/down buttons are still larger than (for example) the remove button (but seem to be about the same size as the Cancel/Ok buttons, which are again larger than the Restore buttons - when did this happen?)

bla

samj1912:
image
Rendering fine for me?

zas:
Still a difference between buttons here too:
Platform: Linux-4.8.0-39-generic-x86_64-with-LinuxMint-18.1-serena CPython 2.7.12
Versions: Picard 1.4.1dev1, PyQt 4.11.4, Qt 4.8.7, Mutagen 1.36.2, Discid discid 1.1.0 (compat layer from python-libdiscid 0.4.1), libdiscid 0.6.1

image

PR #567, opened 2017-01-16
PICARD-353: Add customizable toolbars

samj1912:
https://tickets.metabrainz.org/browse/PICARD-353

How it looks 
image

Example updated toolbar when customized 

image

The list widget also allows drag and drop reordering apart from the up and down buttons. 

Update - Added icons 
--------
image

Eg drag and drop 
image

PR #655, opened 2017-03-12
PICARD-1011: Set action toolbar tab order according to custom order

samj1912:
With #567 support for customizable toolbars was added.
This commit changes the set_tab_order function to correspond with the
custom toolbar layout dynamically.

See: https://tickets.metabrainz.org/browse/PICARD-1011

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -647,27 +647,30 @@ class MainWindow(QtGui.QMainWindow):
         hbox.addWidget(self.search_button)
         toolbar.addWidget(search_panel)
 
-
     def set_tab_order(self):
         tab_order = self.setTabOrder
         tw = self.toolbar.widgetForAction
+        prev_action = None
+        current_action = None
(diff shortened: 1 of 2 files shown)


picard/ui/mainwindow.py
```diff
@@ -647,27 +647,30 @@ def create_search_toolbar(self):
...
 
-
     def set_tab_order(self):
         tab_order = self.setTabOrder
         tw = self.toolbar.widgetForAction
+        prev_action = None
+        current_action = None
+        # Setting toolbar widget tab-orders for accessibility
+        for action in config.setting['toolbar_layout']:
+            if action != 'seperator':
```
zas: `s/seperator/separator/` ?

https://github.com/metabrainz/picard/blob/4c270fc76fe7bd325169f14a720aabc06703c034/picard/ui/options/interface.py#L104

samj1912: yes. fixed and rebased.

PR #657, opened 2017-03-14
PICARD-1001, PICARD-1014, PICARD-1015, PICARD-1016, PICARD-1018: Fix album cover art corner cases

antlarr:
This pull request fixes PICARD-1014, PICARD-1015 and PICARD-1018 (which was actually reported by samj1912 on irc)

Related tickets:
https://tickets.metabrainz.org/browse/PICARD-1001 (Partially fixed)
https://tickets.metabrainz.org/browse/PICARD-1014
https://tickets.metabrainz.org/browse/PICARD-1015
https://tickets.metabrainz.org/browse/PICARD-1016
https://tickets.metabrainz.org/browse/PICARD-1018


mineo:
No code review yet:

- fixes PICARD-1014
- seems to fix PICARD-1018, I think was @samj1912's though
- I can still reproduce PICARD-1001 with the repro from the ticket
- PICARD-1015 is somewhat fixed, the following screenshots are from after loading the album, saving it and dragging a new image to one track 
bla
bla2

antlarr:
PICARD-1015 and PICARD-1001 should be fixed now. Thanks for testing and for the screenshots.

picard/album.py
```diff
@@ -534,8 +531,12 @@ def column(self, column):
                 unsaved = self.get_num_unsaved_files()
                 if unsaved:
                     text += '; %d*' % (unsaved,)
-                text += ungettext("; %i image", "; %i images",
-                                  len(self.metadata.images)) % len(self.metadata.images)
+                if getattr(self.metadata, 'has_common_images', True):
+                    text += ungettext("; %i image", "; %i images",
+                                      len(self.metadata.images)) % len(self.metadata.images)
+                else:
+                    text += ungettext("; %i image", "; %i different images among tracks",
```
zas: SIngular and plural forms have different meanings here, is that expected ?

picard/album.py
```diff
@@ -534,8 +531,12 @@ def column(self, column):
                 unsaved = self.get_num_unsaved_files()
                 if unsaved:
                     text += '; %d*' % (unsaved,)
-                text += ungettext("; %i image", "; %i images",
-                                  len(self.metadata.images)) % len(self.metadata.images)
+                if getattr(self.metadata, 'has_common_images', True):
+                    text += ungettext("; %i image", "; %i images",
+                                      len(self.metadata.images)) % len(self.metadata.images)
```
zas: `len(self.metadata.images)` is repeated 4 times, please store the value in a short variable.

picard/album.py
antlarr: Originally I thought it was not possible to reach that case, but left it there just in case. After a second thought, it is possible when a release doesn't have any cover art and the user adds one image only to one track. So I'll change the singular text to be "%i image not in all tracks"

picard/album.py
antlarr: ok

antlarr:
Please test/review again, I fixed a few other corner cases

mineo:
PICARD-1015 appears to be fixed, thanks! PICARD-1001 still happens.

antlarr:
Can you tell me exactly the steps you follow to reproduce PICARD-1001? Also what release are you testing with, because I can't reproduce it anymore.

Also, note that sometimes the cover looks to be the same, but if you open the info dialog,you'll see that the original and new cover although looking as if they are the same, have different size (not talking about dimensions, but size in KB). So in those cases it's impossible for picard to detect they are the same cover since the images are encoded differently, even if they look the same. I saw a few of those cases, in covers that were previously downloaded from the same provider which now provides the same image encoded a bit differently. Please check if that's what's happening to you.

mineo:
I use the release and configuration file from [this comment](https://tickets.metabrainz.org/browse/PICARD-1001?focusedCommentId=42683&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-42683). Reproducing it is easy: load the release after tagging it once. Changes on the release and all tracks are indicated, but there's no change to any metadata.

antlarr:
I can't reproduce it with that release and my configuration (see http://i.imgur.com/HXwVOSB.png ). I'll try later tonight with your configuration.

zas:
@antlarr : i think about merging those fixes, for 1.4.1, even if some corner cases remain. This PR is meant to fix PICARD-1014, PICARD-1015, PICARD-1018, and PICARD-1001, right ? I would have preferred one PR per ticket if fixes are independent enough.

picard/cluster.py
```diff
@@ -269,6 +279,30 @@ def cluster(files, threshold):
 
             yield album_name, artist_name, (files[i] for i in album)
 
+    def update_metadata_images(self):
```
zas: This code is duplicated in https://github.com/metabrainz/picard/pull/657/files#diff-d6e5ee884fbd0a82f4c76dcd8584b6aaR227 
Not sure if this is stricly needed.

antlarr: This is needed since when removing a file from a cluster, we have to iterate over all files and check if we can remove (or not) the covers from that file from the cluster's metadata.images (checking if other files have the same cover image or not).

Also, this one in Cluster is much simpler than the one in Album, it could be moved together but I think that's a decision I wasn't sure you'd like. Would you prefer to move both Album.update_metadata_images and Cluster.update_metadata_images to a static method in Metadata that would get an obj member (since it has to change album.metadata and album.orig_metadata, but only cluster.metadata) and a list of objects to iterate on ? (so for clusters, it would iterate over its files, but for albums it would iterate over tracks/files/unmatched_files). I'm not sure if it's worth to add even more complexity to have a function that works in the general case as opposed to having similar but simpler code in the two classes.

antlarr:
I mentioned it yesterday on irc, but I'll write it down here for the records and expand on it a bit.

Mineo, your issue is introduced by commit f99abadd10531d7122d9c82487ffa9e188803777 . The problem is that I think that commit is right and we have to rethink the whole image handling to fix this (since reverting that commit would break much more common cases).

That commit sets the file state to CHANGED when self.orig_metadata.images (which contain only images embedded into the files) is different from self.metadata.images (which contain the images from the cover art provider). In your configuration, self.orig_metadata is empty since the file doesn't have any embedded image, so there IS actually a difference in the metadata. The problem is that the user doesn't care if there IS a difference in the metadata but if there WILL BE a difference saved if he pressed the save button.

I propose the next solution:
1) In File._copy_loaded_metadata, set metadata.images (that is, orig_metadata.images) to:

```
if config["save_images_to_tags"] is set:
 - the embedded images (that is, don't change the current behaviour)
elif config["save_images_to_files"] is set:
 - the images in files in the directory (where the filenames would be obtained from config["cover_image_filename"] and, if config["caa_image_type_as_filename"] is set, also files named like "DB:cover_art_archive.art_type/name:*" values in MB_ATTRIBUTES ).
```
So all images comparisons will be done with respect to what will be saved.

2) Change the comparison in the commit above so it's only done when config["save_images_to_tags"] or config["save_images_to_files"] is set.

3) Similarly to 2), stop showing differences in CoverArtBox/InfoDialog if config["save_images_to_tags"] and config["save_images_to_files"] are not set (maybe even not use CA providers in that case?)

The problem with the issue in general is that both config options (save to tags and safe to files) can be true and both embedded and external images can differ (as in your case, Mineo). So it's a complex concept.

antlarr:
(I just rebased the branch against master)

picard/cluster.py
```diff
@@ -94,6 +100,10 @@ def remove_file(self, file):
         self.item.remove_file(file)
         if not self.special and self.get_num_files() == 0:
             self.tagger.remove_cluster(self)
+        self.update_metadata_images()
+        if self.related_album:
```
samj1912: These same 3 lines have been repeated thrice, we can make a private function instead?

picard/cluster.py
```diff
@@ -269,6 +279,30 @@ def cluster(files, threshold):
 
             yield album_name, artist_name, (files[i] for i in album)
 
+    def update_metadata_images(self):
+        class State:
```
samj1912: Do we really need a local class?

picard/track.py
```diff
@@ -220,6 +224,30 @@ def _get_ignored_folksonomy_tags(self):
             tags = [s.strip().lower() for s in ignore_tags.split(',')]
         return tags
 
+    def update_orig_metadata_images(self):
+        class State:
```
samj1912: Same local class here?

picard/ui/infodialog.py
```diff
@@ -99,18 +99,37 @@ class InfoDialog(PicardDialog):
...
+        def get_image_type(image):
+            return image.types_as_string()
+
+        if (isinstance(obj, File) and
+                isinstance(obj.parent, Track) or
+                isinstance(obj, Track)):
             # Display existing artwork only if selected object is track object
             # or linked to a track object
-            if getattr(obj, 'orig_metadata', None) is not None:
+            if (getattr(obj, 'orig_metadata', None) is not None and
```
samj1912: Code duplication in this line and the next elif, you can combine them into one if else

picard/ui/infodialog.py
```diff
@@ -206,18 +225,17 @@ def _display_artwork_type(self):
     def arrange_images(self):
         def get_image_type(image):
             return image.types_as_string()
-        self.obj.metadata.images.sort(key=get_image_type)
-        if self.display_existing_artwork:
-            self.obj.orig_metadata.images.sort(key=get_image_type)
+        self.images.sort(key=get_image_type)
```
samj1912: Looking at the number of times this has been done, maybe we can make an ImageList class and implement __cmp__ along with other default initializations for the class?

Will help in reducing the ```if elses``` and make it easier to fix bugs if all of the code is compacted to one python module.

picard/cluster.py
zas: `add_file()` and `add_files()` share a lot of code, we can reduce code redundancy, but perhaps after merging this PR (since it isn't its purpose).

picard/ui/infodialog.py
antlarr: Nop, we can't because metadata.images is a list generated in mutagen

picard/ui/infodialog.py
antlarr: Indeed, I tried several options there until reaching that point where it can be easily merged

picard/track.py
antlarr: And also in Album, yes. See https://github.com/metabrainz/picard/pull/657#discussion_r106848223 . If you think State/process_images/update_metadata_images/update_orig_metadata_images should be unified for Track/Album/Cluster in a generic way that works for all use cases, I'll do it, but first confirm that you really want to add such a complex function, please. (The signature for the function would be something similar to ```def update_metadata_images(obj, sources,  update_metadata=True, update_orig_metadata=True)``` and would be used like:
```Metadata.update_metadata_images(self, self.linked_files, False, True)``` in Track

File: picard/cluster.py (hunk 5 above)
Comment (antlarr) [inline picard/cluster.py]: Commented below.

File: picard/cluster.py (hunk 4 above)
Comment (antlarr) [inline picard/cluster.py]: fixed

File: picard/ui/infodialog.py (hunk 8 above)
Comment (antlarr) [inline picard/ui/infodialog.py]: Ok, good idea

File: picard/coverart/imagelist.py [hunk 9]
```diff
@@ -0,0 +1,38 @@
...
+
+class ImageList(list):
+
+    def __init__(self):
+        super(ImageList, self).__init__()
+
+    def __eq__(self, other):
+        return sorted(self, key=get_image_type) == sorted(other, key=get_image_type)
+
+    def __getslice__(self, i, j):
```
zas: Isn't that deprecated ?
https://docs.python.org/2.7/reference/datamodel.html#object.__getslice__

An alternative method is shown at http://pythoncentral.io/how-to-slice-custom-objects-classes-in-python/

antlarr: Yes, it's deprecated but it has to be used in this case: (from https://docs.python.org/2.7/reference/datamodel.html#object.__getslice__ ) "Deprecated since version 2.0: Support slice objects as parameters to the __getitem__() method. (However, built-in types in CPython currently still implement __getslice__(). Therefore, you have to override it in derived classes when implementing slicing.)" . So in general it shouldn't be used but for classes derived from lists and dicts, __getslice__ has to be used (only in Python 2).

samj1912:
Latest build fails due to import errors 
See https://travis-ci.org/metabrainz/picard/jobs/213100232#L459

picard/track.py
antlarr: I just submitted a common function to imagelist.py that works for all cases. I'm afraid it's using local imports which I'm not a fan of, but I found no other way to avoid circular imports.

picard/cluster.py
antlarr: Fixed with 68b3eeb

picard/util/imagelist.py
```diff
@@ -0,0 +1,107 @@
...
+                state.has_common_orig_images = False
+                state.orig_images.extend([image for image in src_obj.orig_metadata.images if image not in state.orig_images])
+
+
+def update_metadata_images(obj):
+    from picard.track import Track
+    from picard.cluster import Cluster
+    from picard.album import Album
+
+    class State:
```
samj1912: Redefine class at module level and make the variables instance level rather than class variables.

picard/util/imagelist.py
```diff
@@ -0,0 +1,107 @@
...
+        if state.first_orig_obj:
+            state.orig_images = src_obj.orig_metadata.images[:]
+            state.first_orig_obj = False
+        else:
+            if state.orig_images != src_obj.orig_metadata.images:
+                state.has_common_orig_images = False
+                state.orig_images.extend([image for image in src_obj.orig_metadata.images if image not in state.orig_images])
+
+
+def update_metadata_images(obj):
```
samj1912: I have been wondering if we could use single dispatch from functools to take care of the isinstance if elses.

samj1912: Ah, the decorator was introduced in 3.4. Will redo the code in picard 2.0 then.

picard/util/imagelist.py
antlarr: Done

picard/util/imagelist.py
antlarr: Added a TODO

picard/album.py
```diff
@@ -33,6 +32,7 @@
 from picard.script import ScriptParser
 from picard.ui.item import Item
 from picard.util import format_time, mbid_validate
+from picard.util.imagelist import ImageList
```
zas: You forgot two occurences:
picard/ui/coverartbox.py:27:from picard.coverart.imagelist import ImageList
picard/ui/infodialog.py:29:from picard.coverart.imagelist import ImageList

picard/util/imagelist.py
```diff
@@ -0,0 +1,110 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+
+def get_image_type(image):
+    return image.types_as_string()
+
+
+class ImageList(list):
+
+    def __init__(self):
```
zas: I don't think it is really needed to override `list.__init__()` here, and it disallows to call it with a sequence of images.

picard/util/imagelist.py
```diff
@@ -0,0 +1,110 @@
...
+        super(ImageList, self).__init__()
+
+    def __eq__(self, other):
+        return sorted(self, key=get_image_type) == sorted(other, key=get_image_type)
+
+    def __getslice__(self, i, j):
+        i = max(0, min(i, len(self)))
```
zas: Store `len(self)` in a variable as it it called twice ?

picard/util/imagelist.py
```diff
@@ -0,0 +1,110 @@
...
+        j = max(0, min(j, len(self)))
+        r = ImageList()
+        r[:] = [self[it] for it in range(i, j)]
+        return r
```
zas: If  `__init__()` is not overriden, doing `return ImageList([self[it] for it in range(i, j)])` should be possible (untested).

antlarr: I tested it, and indeed, it worked, so I committed the three changes you requested

samj1912:
Okay, there are still bugs apparently. If you disable all CA providers and load the following test file -> 

I am unable to save the file at all. (No green ticks)


[test.flac.zip](https://github.com/metabrainz/picard/files/866930/test.flac.zip)

PR #658, opened 2017-03-16
PICARD-1021: Skip file loads if Picard is stopping

Sophist-UK:
Resolves https://tickets.metabrainz.org/browse/PICARD-1021.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -98,12 +98,26 @@ class File(QtCore.QObject, Item):
 
     def load(self, callback):
         thread.run_task(
-            partial(self._load, self.filename),
+            partial(self._load_check, self.filename),
             partial(self._loading_finished, callback),
             priority=1)
 
+    def _load_check(self, filename):
+        # Check that file has not been removed since thread was queued
+        # Don't load if we are stopping.
+        if self.state != File.PENDING or self.tagger.stopping:
+            log.debug("File not loaded because %s: %r",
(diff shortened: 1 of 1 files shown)


picard/file.py
```diff
@@ -98,12 +98,26 @@ def new_metadata(self):
...
         thread.run_task(
-            partial(self._load, self.filename),
+            partial(self._load_check, self.filename),
             partial(self._loading_finished, callback),
             priority=1)
 
+    def _load_check(self, filename):
+        # Check that file has not been removed since thread was queued
+        # Don't load if we are stopping.
+        if self.state != File.PENDING or self.tagger.stopping:
```
zas: At the expense of a minor code duplication, for the sake of clarity, please rewrite it as:

```python
if self.state != File.PENDING:
    log.debug("File not loaded because it was removed: %r", self.filename)
    return None
if self.tagger.stopping:
    log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
    return None
return self._load(filename)
```

Also don't use hardcoded Picard in strings (well, i guess there are already too many occurences), we have a constant for it. It will ease a fork under another name (open source!!). Or just use "application" instead of "Picard).

Also, we may want to translate those strings at some point, and building strings as you did here isn't very i18n friendly.

Sophist-UK:
@zas comments addressed.

PR #659, opened 2017-03-18
PICARD-1024: Allow specifying a configuration file path

mineo:
This is mostly useful for debugging to avoid touching the main configuration
file while trying different settings, but could also be used for switching
between multiple tagging profiles.

https://tickets.metabrainz.org/browse/PICARD-1024

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -18,8 +18,6 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from __future__ import print_function
-import re
-import sys
 from operator import itemgetter
 from PyQt4 import QtCore
 from picard import (PICARD_APP_NAME, PICARD_ORG_NAME, PICARD_VERSION,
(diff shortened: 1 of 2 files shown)


picard/ui/options/renaming.py
```diff
@@ -22,11 +22,11 @@
 import sys
 from functools import partial
 from PyQt4 import QtGui
-from PyQt4.QtGui import QPalette, QColor
+from PyQt4.QtGui import QPalette
 from picard import config
 from picard.const import PICARD_URLS
 from picard.file import File
-from picard.script import ScriptParser, SyntaxError, ScriptError, UnknownFunction
+from picard.script import ScriptParser, SyntaxError, ScriptError
```
zas: @mineo: this cleanup is totally unrelated to the rest of the PR imho. Can you make a standalone import cleanup PR ? I agree with those trivial changes of course, but it is better to not mix them with feature PRs.

zas:
:+1:

picard/ui/options/renaming.py
mineo: I won't make an import cleanup PR across all of Picard because this requires a decent amount of testing because not all imports are side effect free (that is, simply removing the otherwise unused import might already run code).

PR #661, opened 2017-03-19
PICARD-1023: Tweak Add Directory path persistance

Sophist-UK:
Stop Add Directory opening one directory higher if you don't select any sub-directories.

Resolves https://tickets.metabrainz.org/browse/PICARD-1023

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -735,7 +735,7 @@ class MainWindow(QtGui.QMainWindow):
 
         dir_count = len(dir_list)
         if dir_count:
-            parent = os.path.dirname(dir_list[0])
+            parent = os.path.dirname(dir_list[0]) if dir_count > 1 else dir_list[0]
             config.persist["current_directory"] = parent
             if dir_count > 1:
                 self.set_statusbar_message(


zas:
Can you provide screenshots before/after this PR ?

Sophist-UK:
I could - but I would need 6 screen shots to show the before and after of three user cases, and it seems like a lot of effort for a 1 line change. The three cases are:

a. Add Directory, click Choose without selecting anything
    Before: Next time opens one directory up (so you need to select the sub-directory).
    After: Next time opens in same directory.

b. Add Directory, select multiple sub-directories, click Choose
    Before: Next time opens in same directory.
    After: Next time opens in same directory.

c. Add Directory, select one sub-directory click Choose
    Before: Next time opens same directory.
    After: Next time opens in selected directory (so you can click choose without selecting anything).

Unfortunately dialog returns single directory for both a. and c. so no way to distinguish between them to avoid change in c. if you want change in a.

My belief is that the new behaviour is less annoying than the old behaviour.

flint7flint:
I do not want you to have problems. Therefore, I do not come to your site. I study myself. I have successes.
  But the material to regret was only adult content.

sorry. But I did not quite understand what you wrote to me, what exactly should I do.
Not to harm you.

zas:
> I could - but I would need 6 screen shots to show the before and after of three user cases, and it seems like a lot of effort for a 1 line change. The three cases are:

I can't reproduce issues you are trying to fix, comparing master and this branch don't show any change to me, this is why i asked for screenshots, to be sure i'm testing the same things you do.

Sophist-UK:
Its a 1-line change at line 738 in mainwindow.py.

https://github.com/metabrainz/picard/pull/661/files

zas:
> Its a 1-line change at line 738 in mainwindow.py.

I noticed, still i don't know what it fixes exactly, comparing master and this branch i can't find the case where it makes a difference. Hence my ask for screenshots.

mineo:
@zas You need to deactivate Options -> Options -> User Interface -> Begin browsing [...].

PR #663, opened 2017-03-24
PICARD-1030: Add cover art context menu and "keep original images" option

antlarr:
This adds a cover art context menu that allows users to select if they want dropped images to replace the current images or they want them to be appended to the current images. The context menu also adds an option to "Keep original images" which allows to ignore downloaded images and keep the original images if users prefer the images they already have.

Fixes [PICARD-1030](https://tickets.metabrainz.org/browse/PICARD-1030)

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -380,9 +380,11 @@ class CoverArtBox(QtGui.QGroupBox):
             return
 
         if config.setting["behaviour_on_image_drop"] == 'replace':
-            drop_image = lambda obj: obj.metadata.set_front_image(coverartimage)
+            def drop_image(obj):
+                obj.metadata.set_front_image(coverartimage)
         else:
-            drop_image = lambda obj: obj.metadata.append_image(coverartimage)
+            def drop_image(obj):
+                obj.metadata.append_image(coverartimage)
 
         if isinstance(self.item, Album):
             album = self.item
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -21,7 +21,7 @@ import os
 import sys
 from functools import partial
 from PyQt4 import QtCore, QtGui, QtNetwork
-from picard import log
+from picard import config, log
 from picard.album import Album
 from picard.coverart.image import CoverArtImage, CoverArtImageError
 from picard.track import Track
@@ -378,14 +378,21 @@ class CoverArtBox(QtGui.QGroupBox):
         except CoverArtImageError as e:
             log.warning("Can't load image: %s" % unicode(e))
             return
+
+        if config.setting["behaviour_on_image_drop"] == 'replace':
+            drop_image = lambda obj: obj.metadata.set_front_image(coverartimage)
+        else:
+            drop_image = lambda obj: obj.metadata.append_image(coverartimage)
+
         if isinstance(self.item, Album):
             album = self.item
             album.enable_update_metadata_images(False)
+            drop_image(album)
             for track in album.tracks:
-                track.metadata.set_front_image(coverartimage)
+                drop_image(track)
                 track.metadata_images_changed.emit()
             for file in album.iterfiles():
-                file.metadata.set_front_image(coverartimage)
+                drop_image(file)
                 file.metadata_images_changed.emit()
                 file.update()
             album.enable_update_metadata_images(True)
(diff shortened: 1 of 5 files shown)


picard/ui/coverartbox.py
```diff
@@ -394,19 +403,54 @@ def load_remote_image(self, url, mime, data):
         elif isinstance(self.item, Track):
             track = self.item
             track.album.enable_update_metadata_images(False)
-            track.metadata.set_front_image(coverartimage)
+            drop_image(track)
             track.metadata_images_changed.emit()
             for file in track.iterfiles():
-                file.metadata.set_front_image(coverartimage)
+                drop_image(file)
```
zas: I think `drop_image()` is a misleading name.

picard/ui/coverartbox.py
```diff
@@ -378,14 +378,23 @@ def load_remote_image(self, url, mime, data):
         except CoverArtImageError as e:
             log.warning("Can't load image: %s" % unicode(e))
             return
+
+        if config.setting["behaviour_on_image_drop"] == 'replace':
```
zas: Since this option only applies to `load_remote_image()` and isn't specific to drag'n'drop (well, it is at the moment, since `load_remote_image()` is only called when dropping, but we could imagine other ways to add a remote image), i think the option should be `load_remote_image_behavior` instead.

picard/ui/coverartbox.py
```diff
@@ -394,19 +403,54 @@ def load_remote_image(self, url, mime, data):
...
         elif isinstance(self.item, File):
             file = self.item
-            file.metadata.set_front_image(coverartimage)
+            drop_image(file)
             file.metadata_images_changed.emit()
             file.update()
         self.cover_art.set_metadata(self.item.metadata)
         self.show()
+
+    def setBehaviourOnImageDrop(self, behaviour):
```
zas: https://github.com/metabrainz/picard/pull/663/files#r107887135

picard/ui/coverartbox.py
```diff
@@ -394,19 +403,54 @@ def load_remote_image(self, url, mime, data):
...
+        config.setting["behaviour_on_image_drop"] = behaviour
+
+    def contextMenuEvent(self, event):
+        menu = QtGui.QMenu(self)
+        if self.show_details_button.isVisible():
+            name = _(u'Show more details')
```
zas: Unrelated, but it should be: Show more details... as it opens a dialog

picard/ui/coverartbox.py
antlarr: It's not neccesarily "remote", local files can also be dropped from any file manager. Maybe ```load_image_behavior``` ?

picard/ui/coverartbox.py
antlarr: Hmmm, I see what you mean... any suggestion for a better name? process_image ?

picard/ui/coverartbox.py
antlarr: Sure, I noticed it but forgot to fix it, thanks for the reminder

picard/ui/coverartbox.py
zas: `set_image()` ?

picard/ui/coverartbox.py
zas: I'm ok for `load_image_behavior`, you're right about local files.

picard/ui/coverartbox.py
```diff
@@ -378,14 +378,23 @@ def load_remote_image(self, url, mime, data):
         except CoverArtImageError as e:
             log.warning("Can't load image: %s" % unicode(e))
             return
+
+        if config.setting["behaviour_on_image_drop"] == 'replace':
+            def drop_image(obj):
```
samj1912: I am not so sure about defining functions inside if/else

picard/ui/coverartbox.py
samj1912: any particular reason this is camelCase and not snake_case?

samj1912: snake_case here?

antlarr: That function is a slot called by the Qt signal-slot mechanism, so I thought it was ok to use camelCase, but ok, I'll change that.

PR #664, opened 2017-03-25
PICARD-558: Initial port to Py3

samj1912:
https://tickets.metabrainz.org/browse/PICARD-588


picard/ui/infodialog.py
```diff
@@ -297,7 +297,7 @@ def _display_info_tab(self):
         if album.errors:
             tabWidget.setTabText(tab_index, _("&Errors"))
             text = '<br />'.join(map(lambda s: '<font color="darkred">%s</font>' %
-                                     '<br />'.join(unicode(QtCore.Qt.escape(s))
+                                     '<br />'.join(str(cgi.escape(s))
```
zas: https://wiki.qt.io/Transition_from_Qt_4.x_to_Qt5 says `Qt.escape()` is deprecated in favor of `QString.toHtmlEscaped()`, `cgi.escape()` looks weird here, and it doesn't escape `"` by default.

There is also Py3 `html.escape()` https://docs.python.org/3/library/html.html 
https://wiki.python.org/moin/EscapingHtml

Not sure which one is better.

samj1912: QString.toHtmlEscaped() in doesn't work python it's a C only method. We can use html.escape however.

picard/metadata.py
```diff
@@ -267,7 +267,7 @@ def set(self, name, values):
     def __setitem__(self, name, values):
         if not isinstance(values, list):
             values = [values]
-        values = filter(None, map(unicode, values))
+        values = [value for value in map(str, values) if value is not None]
```
zas: `[str(value) for value in values if value is not None]` ?

picard/script.py
```diff
@@ -425,7 +426,8 @@ def func_unset(parser, name):
     # Allow wild-card unset for certain keys
     if name in ('performer:*', 'comment:*', 'lyrics:*'):
         name = name[:-1]
-        for key in parser.context.keys():
+        key_list = list(parser.context.keys())
```
zas: `for key in parser.context:` ?

samj1912: Can't do that since we are modifying parser.context in the loop. Storing it in an initial list just to be safe.

zas:
> Can't do that since we are modifying parser.context in the loop. Storing it in an initial list just to be safe.
I didn't read next lines... ;)

picard/util/webbrowser2.py
```diff
@@ -23,65 +23,6 @@
 from PyQt5 import QtWidgets
 from picard.const import PICARD_URLS
 
-"""
-A webbrowser extension that respects user's preferred browser on each
-platform. Python 2.5 already has *some* support for this, but it's not
-enough, in my opinion. See also:
-http://sourceforge.net/tracker/index.php?func=detail&aid=1681228&group_id=5470&atid=105470
-"""
```
Freso: Please replace docstring instead of removing it. It's hard enough to get documentation into the codebase, so let's not remove what we have without at least putting new documentation in place.

picard/util/webbrowser2.py
```diff
@@ -23,65 +23,6 @@
 from PyQt5 import QtWidgets
```
Freso: I can't comment on code above this line. :( But I wanted to note that with the removed code, `os` and `sys` are no longer used, so they also no longer need to be `import`ed here.

picard/ui/options/interface.py
```diff
@@ -223,7 +223,7 @@ def _all_list_items(self):
 
     def _added_actions(self):
         actions = self._all_list_items()
-        actions = filter(lambda x: x != 'separator', actions)
+        actions = [action for action in actions if action != 'separator']
```
mineo: FWIW, you can just do `actions = set(action for .... )` here.

picard/ui/options/plugins.py
```diff
@@ -33,10 +34,6 @@
 from picard.ui.ui_options_plugins import Ui_PluginsOptionsPage
 
 
-def cmp_plugins(a, b):
```
mineo: The benefit of the explicitly defined function is that there's only one place it's defined, instead of twice.

picard/util/astrcmp.c
```diff
@@ -170,8 +170,17 @@ static PyMethodDef AstrcmpMethods[] = {
...
 };
 
-PyMODINIT_FUNC
-initastrcmp(void)
+static struct PyModuleDef AstrcmpModule =
 {
-	(void)Py_InitModule("astrcmp", AstrcmpMethods);
-}
+    PyModuleDef_HEAD_INIT,
+    "astrcmp", /* name of module */
```
mineo: Nitpick: either let all comments here begin one space after the comma or align the slashes.

picard/util/imageinfo.py
```diff
@@ -17,8 +17,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-import StringIO
 import struct
+from picard.compat import StringIO
```
mineo: If we're moving to Python 3, just remove `compat.py` and use whatever the correct import is.

picard/ui/options/plugins.py
samj1912: Py3 deprecates cmp.

picard/ui/options/interface.py
```diff
@@ -125,8 +125,8 @@ def __init__(self, parent=None):
         self.ui.setupUi(self)
         self.ui.ui_language.addItem(_('System default'), '')
         language_list = [(l[0], l[1], _(l[2])) for l in UI_LANGUAGES]
-        for lang_code, native, translation in sorted(language_list, key=operator.itemgetter(2),
-                                                     cmp=locale.strcoll):
+        fcmp = lambda x: locale.strxfrm(x[2])
+        for lang_code, native, translation in sorted(language_list, key=fcmp):
```
samj1912: py3 deprecates cmp, thus the problem.

samj1912:
This is not a working port, I am currently working on it. It wont run and it doesn't pass the tests.

picard/ui/options/interface.py
samj1912: Set does not preserve order.

picard/i18n.py
```diff
@@ -93,13 +93,13 @@ def _ugettext_countries(msg):
         def _ugettext_attributes(msg):
             return msg
 
-    builtins.__dict__['ungettext'] = _ungettext
+    builtins.__dict__['lngettext'] = _lngettext
```
zas: This change requires few more changes:

https://github.com/metabrainz/picard/blob/master/setup.cfg#L8

In order to extract strings used as `lngettext()` arguments, you may need to modify https://github.com/metabrainz/picard/blob/master/setup.py#L423
because `lngettext` isn't part of default keywords (see https://github.com/python-babel/babel/blob/master/babel/messages/extract.py#L32)

samj1912:
Working now and passes all tests except filenaming ones.

samj1912:
@xlotlu can you help me with porting the filenaming util to py3?

xlotlu:
Sure, but I probably won't be able to work on it unil next week.

>@xlotlu can you help me with porting the filenaming util to py3?

picard/ui/options/interface.py
```diff
@@ -223,8 +222,8 @@ def _all_list_items(self):
 
     def _added_actions(self):
         actions = self._all_list_items()
-        actions = [action for action in actions if action != 'separator']
-        return set(actions)
+        actions = set(action for action in actions if action != 'separator')
+        return actions
```
zas: Make it one line:
`return set(action for action in self._all_list_items() if action != 'separator')`

picard/formats/asf.py
```diff
@@ -221,10 +220,7 @@ def _remove_deleted_tags(self, metadata, tags):
         for tag in metadata.deleted_tags:
             real_name = self._get_tag_name(tag)
             if real_name and real_name in tags:
-                if tag == 'totaldiscs':
-                    tags[real_name] = metadata['discnumber']
```
zas: Not needed anymore ?

samj1912: Something changed in the way mutagen behaves with the WM/PartOfSet tag, one is useless without the other. (disc number and total discs)

samj1912:
@xlotlu fixed it by myself :)

picard/ui/__init__.py
```diff
@@ -27,3 +28,16 @@ class PicardDialog(QtWidgets.QDialog):
 
     def __init__(self, parent=None):
         QtWidgets.QDialog.__init__(self, parent, self.flags)
+
+
+class HashableTreeWidgetItem(QtWidgets.QTreeWidgetItem):
```
mineo: This commit and 9ffc73b0bb6facc0895f3dd73508abae196b5e40 are really in need of an explanation: neither the class documentation nor the commit message explain why this is necessary. I guess it's because of [this](https://stackoverflow.com/questions/42140366/qtreewidgetitem-not-hashable-in-python3)?

zas: Yes, it is. And i agree it deserves a comment in the code.

zas:
Tested with py3.5 and pyqt5.8, reported issues were all fixed. Good job, @samj1912 !

picard/ui/options/plugins.py
Freso: I agree with @mineo. If we're making the same function twice, it would be better to make it once.

`fcmp = lambda x: locale.strxfrm(x[2])` is exactly the same as
```python
fcmp(x):
    return locale.strxfrm(x[2])
```
And can be used in the same manner, but the latter has the benefit of being unittestable and docstringable.

PR #666, opened 2017-03-29
PICARD-1034: Add support for TOAL id3 frame

samj1912:
https://tickets.metabrainz.org/browse/PICARD-1034

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -112,6 +112,7 @@ class ID3File(File):
         'TCON': 'genre',
         'TMED': 'media',
         'TMOO': 'mood',
+        'TOAL': 'work',
         'TCOP': 'copyright',
         'TPUB': 'label',
         'TDOR': 'originaldate',
@@ -153,7 +154,6 @@ class ID3File(File):
         'ASIN': 'asin',
         'MusicMagic Fingerprint': 'musicip_fingerprint',
         'Artists': 'artists',
-        'Work': 'work',
         'Writer': 'writer',
     }
     __rtranslate_freetext = dict([(v, k) for k, v in __translate_freetext.iteritems()])


picard/formats/id3.py
```diff
@@ -112,6 +112,7 @@ class ID3File(File):
         'TCON': 'genre',
         'TMED': 'media',
         'TMOO': 'mood',
+        'TOAL': 'work',
```
Sophist-UK: Work is also used in freetext.

When saving, the work metadata will be saved both as TOAL and TXXX:work.

When loading and TOAL and TXXX:work are both set, I have not tested which takes precedence or whether you get multi-value metadata.

ijabz:
Please see  discussion at https://community.metabrainz.org/t/picard-using-toal-for-work-in-id3/239131/10

In ID3 TOAL is described as:

The 'Original album/movie/show title' frame is intended for the title
of the original recording (or source of sound), if for example the
music in the file should be a cover of a previously released song.

whereas TIT1 is defined as:

The 'Content group description' frame is used if the sound belongs to
a larger category of sounds/music. For example, classical music is
often sorted in different musical sections (e.g. "Piano Concerto",
"Weather - Hurricane").

So why write works to TOAL it makes no sense to me ?

Also seeing as you are already writing to TXXX:work why are you considering duplicating the information

Also ID3 is case sensitive I thought Picard wrote to TXXX:work not TXXX:Work. 

FWIW Jaikoz/SongKong uses TXXX:WORK rather than TXXX:Work simply because I find that tools such as DbPoweramp Transcoder and MinimServer UPnP expect the same value to be used for different formats otherwise things gets broken.  Also although we use TXXX:WORK we only use it for Classical works rather than any work as that works better for our users, we have a separate field MUSICBRAINZ_WORK_COMPOSITION for any recording that is linked to a MusicBrainz work

Minor point I use TXXX:WORK rather than TXXX:Work simply because I find that tools such as DbPoweramp Transcoder and MinimServer UPnP expect the same value to be used for different formats otherwise things gets broken. But I don't suppose you can/would want to change it at this point.

phw:
We had this change already in the past and reverted again. @Sophist-UK, you did that, maybe you can explain about the reasoning for both introducing and reverting this.

Otherwise I agree with ijabz

Sophist-UK:
I can find where I introduced support for Work - but nothing about using TOAL as the tag for Work.

Can you give me a link to the PR or commit?

Sophist-UK:
P.S. My vote is to use whatever tag has greatest compatibility with other taggers and media players. If they use TOAL for Work, I can go with that - but if they use it as per spec, then I would go with that instead.

samj1912:
No longer relevant after #667

Sophist-UK:
Just as well given the PR no. It was doomed from the start really. :-)

PR #667, opened 2017-03-30
PICARD-1034: Add support for loading and saving TOAL & TOPE id3 frames

Sophist-UK:
This PR adds support for loading and saving the TOAL and TOPE tags.

It does not add support for populating these from MB - see ticket for details.

Resolves https://tickets.metabrainz.org/browse/PICARD-1034


Sophist-UK:
Can we merge this also into 1.4.x please so I can work on next set of stuff without needing to install Qt5?

PR #668, opened 2017-03-30
PICARD-1041: Replace Ok button text by Make It So! in Options dialog

zas:


picard/ui/options/dialog.py
```diff
@@ -84,7 +84,9 @@ def __init__(self, default_page=None, parent=None):
         self.ui.reset_button = QtGui.QPushButton(_("Restore &Defaults"))
         self.ui.reset_button.setToolTip(_("Reset all settings for current option page"))
 
-        self.ui.buttonbox.addButton(StandardButton(StandardButton.OK), QtGui.QDialogButtonBox.AcceptRole)
```
Sophist-UK:
```
# April fool
self.ui.buttonbox.addButton(StandardButton(StandardButton.OK).setText(_("Make It So!")), QtGui.QDialogButtonBox.AcceptRole)
```
?

zas: @Sophist-UK :  it's not an April fool.

Sophist-UK: Ok. :-)

mineo:
I have to admit that I don't like this. It's bad enough to make one of the most important buttons of the whole application an easter egg that not everybody might get, but this is not even translated in a single language. I'm sorry, but do we expect everybody using any of the enabled languages in picard to get english star trek references?

Sophist-UK:
Admittedly it is a bit geeky - ok, a lot geeky - and of course, @mineo's common sense should prevail.

But I hope it doesn't - a little quirkiness and individuality is becoming a rarity these days.

Of course, @mineo can always submit translations so that "Make it so" translates to OK in every language other than English. ;-)

Sophist-UK:
P.S. Should `Make it so` be in quotation marks `"Make it so"`?

mineo:
> Of course, @mineo can always submit translations so that "Make it so" translates to OK in every language other than English. ;-)

I could, but would already fail at providing one for my native language german because AFAIK "Make it so" is not Picards catch-phrase in german and doesn't have a consistent translation across all episodes.

Sophist-UK:
Thinking again about this - how about the button text being "OK, Make it so" as a compromise - because some people won't get the reference and will wonder where the OK button is.

PR #672, opened 2017-04-03
PICARD-1047: Mark plugins with api_versions<2 as incompatible

samj1912:
Fixes https://tickets.metabrainz.org/browse/PICARD-1047


Freso:
This is not what PICARD-1046 is about though. You can still write a plugin for API 2.0.0 with badly written code that will make Picard crash. PICARD-1046 is about catching such a crash and letting Picard continue regardless.

samj1912:
Updated description and created a different ticket :)

Sophist-UK:
1. Presumably we need to stop Picard 1 plugins loading in Picard 2 and vice versa.

2. Do we have a new definition for Picard2 APIs? Do we expect these to be significantly different? It will be a pain to rewrite existing plugins to meet a different API, however a clean break could be very beneficial. 

One area I would like to see APIs change is in presentation of raw metadata for a couple of reasons:

a. It would be nice to have an internal representation which can survive the switch from WS2 XML to WS3 JSON.

b. Profiling shows that /picard/webservice.py#_read_xml is a CPU hog - because it is python rather than a C library. If we can find a C-library which can do point A. that would be highly beneficial.

If we were sticking with XML for the foreseeable future, I would recommend using ElementTree, but not if we are switching to JSON.

MB documentation suggests that a JSON interface is available for WS2 - is this stable enough and fully working to switch to in Picard2? Or is this a step too far? And do we know enough about WS3 to know how to have a common internal representation?

P.S. Apologies for the brain dump here.

zas:
> MB documentation suggests that a JSON interface is available for WS2 - is this stable enough and fully working to switch to in Picard2? Or is this a step too far? And do we know enough about WS3 to know how to have a common internal representation?

JSON interface should be stable enough on WS2.
WS3 isn't defined for now, but for sure it will be JSON only.

So it is actually a good idea to support JSON WS2 in Picard 2.0.

About "common internal representation", the thing is that plugins currently can use anything in the Picard code (data & functions), and it is an issue because a plugin may use non-stable functions.

We could start listing data & functions currently used by most plugins, and see if we can provide a more stable API and how.

PR #675, opened 2017-04-04
PICARD-1052: Only enable search for similar tracks if the selection contains objects

mineo:
https://tickets.metabrainz.org/browse/PICARD-1052

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -873,6 +873,7 @@ class MainWindow(QtWidgets.QMainWindow):
         can_view_info = bool(single and single.can_view_info())
         can_browser_lookup = bool(single and single.can_browser_lookup())
         have_files = len(self.tagger.get_files_from_objects(self.selected_objects)) > 0
+        have_objects = len(self.selected_objects) > 0
         for obj in self.selected_objects:
             if obj is None:
                 continue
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -902,6 +903,7 @@ def update_actions(self):
         self.cut_action.setEnabled(bool(self.selected_objects))
         files = self.get_selected_or_unmatched_files()
         self.tags_from_filenames_action.setEnabled(bool(files))
+        self.track_search_action.setEnabled(have_objects)
```
zas: `self.track_search_action.setEnabled(bool(self.selected_objects))` and drop `have_objects` ?

mineo: I much prefer the similarity to `have_files` instead of the implicit `bool` thing.

zas: In the proposed code, `bool()` is explicit, not implicit.

The point is that testing non-emptyness of a sequence using `len(x) > 0` isn't pythonic, as documented in PEP8 for example, and also read https://docs.python.org/3/reference/datamodel.html#object.__bool__ (or equivalent __nonzero__ for py2).

I'm not totally against intermediate variable like `have_objects` when it clarifies the code, but this code is pretty trivial. Also look at line 901, `self.cut_action.setEnabled(bool(self.selected_objects))` for example, so, while you are consistent with `have_files`, you aren't with previous lines.
Imho `have_files`  shouldn't use `len() > 0` either.


I have the feeling we need to unify this stuff a bit, because when looking at those lines we have:
- useless intermediate variables
- non-pythonic way of testing (non-)emptyness of sequences
- inconsistent way of doing it from lines to lines

It doesn't change things much in practice, but i would like us to unify the code style a bit.

@samj1912, @Sophist-UK, @mwiencek : any advice on the style here ?

mineo: I've changed the code to use `bool(sequence)`, but kept the intermediate variables because both `have_files` and `have_objects` are used twice.

PR #676, opened 2017-04-04
PICARD-1043: Add support for iTunes Movement tags - 1.4.x branch

Sophist-UK:
Resolves https://tickets.metabrainz.org/browse/PICARD-1043

Note 1: This is targetted at the 1.4.x branch in order to avoid issues of installing Qt5 with Py2. 
Note 2: This PR currently includes the commits from #667 pending their inclusion in the 1.4.x branch. Once Py3 support is in master or #667 cherry picked into 1.4.x I will rebase without these.
Note 3: See #695 for equivalent PR for master branch.

### DIFF
--- a/INSTALL.txt
+++ b/INSTALL.txt
@@ -12,7 +12,7 @@ Before installing Picard, you need to have these:
  * PyQt 4.5 or newer
    http://www.riverbankcomputing.co.uk/software/pyqt/download
 
- * Mutagen 1.22 or newer (1.23 or newer for AIFF support)
+ * Mutagen 1.37 or newer
    https://bitbucket.org/lazka/mutagen/downloads
 
  * gettext
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -146,6 +146,7 @@ class ID3File(File):
         'MusicBrainz Release Track Id': 'musicbrainz_trackid',
         'MusicBrainz Disc Id': 'musicbrainz_discid',
         'MusicBrainz Work Id': 'musicbrainz_workid',
+        'MusicBrainz Movement Work Id': 'musicbrainz_movementid',
         'MusicBrainz Release Group Id': 'musicbrainz_releasegroupid',
         'MusicBrainz Album Release Country': 'releasecountry',
         'MusicIP PUID': 'musicip_puid',
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -123,6 +123,7 @@ class ID3File(File):
         'WCOP': 'license',
         'WOAR': 'website',
         'COMM': 'comment',
+        'TOAL': 'originalalbum',
         'TOPE': 'originalartist',
 
         # The following are informal iTunes extensions to id3v2:
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -209,7 +209,7 @@ class ID3File(File):
             frameid = frame.FrameID
             if frameid in self.__translate:
                 name = self.__translate[frameid]
-                if frameid.startswith('T'):
+                if frameid.startswith('T') or frameid in ["MVNM", "GRP1"]:
                     for text in frame.text:
                         if text:
                             metadata.add(name, unicode(text))
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -40,6 +40,8 @@ from urlparse import urlparse
 id3.TCMP = compatid3.TCMP
 id3.TSO2 = compatid3.TSO2
 id3.TSOC = compatid3.TSOC
+id3.MVNM = compatid3.MVNM
+id3.MVIN = compatid3.MVIN
 
 __ID3_IMAGE_TYPE_MAP = {
     "other": 0,
@@ -130,6 +132,7 @@ class ID3File(File):
         'TCMP': 'compilation',
         'TSOC': 'composersort',
         'TSO2': 'albumartistsort',
+        'MVNM': 'movementname',
     }
     __rtranslate = dict([(v, k) for k, v in __translate.iteritems()])
 
@@ -171,10 +174,12 @@ class ID3File(File):
     _rtipl_roles = dict([(v, k) for k, v in _tipl_roles.iteritems()])
 
     __other_supported_tags = ("discnumber", "tracknumber",
-                              "totaldiscs", "totaltracks")
+                              "totaldiscs", "totaltracks",
+                              "movementnumber", "movementtotal")
     __tag_re_parse = {
         'TRCK': re.compile(r'^(?P<tracknumber>\d+)(?:/(?P<totaltracks>\d+))?$'),
-        'TPOS': re.compile(r'^(?P<discnumber>\d+)(?:/(?P<totaldiscs>\d+))?$')
+        'TPOS': re.compile(r'^(?P<discnumber>\d+)(?:/(?P<totaldiscs>\d+))?$'),
+        'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$'),
     }
 
     def build_TXXX(self, encoding, desc, values):
@@ -308,6 +313,13 @@ class ID3File(File):
                 text = metadata['discnumber']
             tags.add(id3.TPOS(encoding=0, text=id3text(text, 0)))
 
+        if 'movementnumber' in metadata:
+            if 'movementtotal' in metadata:
+                text = '%s/%s' % (metadata['movementnumber'], metadata['movementtotal'])
+            else:
+                text = metadata['movementnumber']
+            tags.add(id3.MVIN(encoding=0, text=id3text(text, 0)))
+
         # This is necessary because mutagens HashKey for APIC frames only
         # includes the FrameID (APIC) and description - it's basically
         # impossible to save two images, even of different types, without
@@ -383,7 +395,7 @@ class ID3File(File):
                     elif frameid == 'WOAR' and valid_urls:
                         for url in values:
                             tags.add(id3.WOAR(url=url))
-                elif frameid.startswith('T'):
+                elif frameid.startswith('T') or frameid == "MVNM":
                     if config.setting['write_id3v23']:
                         if frameid == 'TMOO':
                             tags.add(self.build_TXXX(encoding, 'mood', values))
@@ -500,6 +512,8 @@ class ID3File(File):
             return 'TRCK'
         elif name == 'discnumber':
             return 'TPOS'
+        elif name == 'movementnumber':
+            return 'MVIN'
         else:
             return None
 
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -123,6 +123,7 @@ class ID3File(File):
         'WCOP': 'license',
         'WOAR': 'website',
         'COMM': 'comment',
+        'TOPE': 'originalartist',
 
         # The following are informal iTunes extensions to id3v2:
         'TCMP': 'compilation',
@@ -432,7 +433,7 @@ class ID3File(File):
                         if frame.FrameID in ('TMCL', 'TIPL', 'IPLS'):
                             for people in frame.people:
                                 if people[0] == role:
-                                    frame.people.remove(people)   
+                                    frame.people.remove(people)
                 elif name.startswith('comment:'):
                     desc = name.split(':', 1)[1]
                     if desc.lower()[:4] != 'itun':
@@ -453,7 +454,7 @@ class ID3File(File):
                         if frame.FrameID in ('TIPL', 'IPLS'):
                             for people in frame.people:
                                 if people[0] == role:
-                                    frame.people.remove(people)   
+                                    frame.people.remove(people)
                 elif name == 'musicbrainz_recordingid':
                     for key, frame in tags.items():
                         if frame.FrameID == 'UFID' and frame.owner == 'http://musicbrainz.org':
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -42,6 +42,7 @@ id3.TSO2 = compatid3.TSO2
 id3.TSOC = compatid3.TSOC
 id3.MVNM = compatid3.MVNM
 id3.MVIN = compatid3.MVIN
+id3.GRP1 = compatid3.GRP1
 
 __ID3_IMAGE_TYPE_MAP = {
     "other": 0,
@@ -133,6 +134,7 @@ class ID3File(File):
         'TSOC': 'composersort',
         'TSO2': 'albumartistsort',
         'MVNM': 'movementname',
+        'GRP1': 'itunesgrouping',
     }
     __rtranslate = dict([(v, k) for k, v in __translate.iteritems()])
 
@@ -396,7 +398,7 @@ class ID3File(File):
                     elif frameid == 'WOAR' and valid_urls:
                         for url in values:
                             tags.add(id3.WOAR(url=url))
-                elif frameid.startswith('T') or frameid == "MVNM":
+                elif frameid.startswith('T') or frameid in ["MVNM", "GRP1"]:
                     if config.setting['write_id3v23']:
                         if frameid == 'TMOO':
                             tags.add(self.build_TXXX(encoding, 'mood', values))
--- a/picard/log.py
+++ b/picard/log.py
@@ -48,8 +48,7 @@ class Logger(object):
         self._receivers.append(receiver)
 
     def unregister_receiver(self, receiver):
-        if receiver in self._receivers:
-            self._receivers.remove(receiver)
+        self._receivers.remove(receiver)
 
     def message(self, level, message, *args):
         if not self.log_level(level):
--- a/picard/log.py
+++ b/picard/log.py
@@ -48,7 +48,8 @@ class Logger(object):
         self._receivers.append(receiver)
 
     def unregister_receiver(self, receiver):
-        self._receivers.remove(receiver)
+        if receiver in self._receivers:
+            self._receivers.remove(receiver)
 
     def message(self, level, message, *args):
         if not self.log_level(level):
--- a/picard/mbxml.py
+++ b/picard/mbxml.py
@@ -350,10 +350,10 @@ def workparts_to_metadata(partof, m):
 
 def work_to_metadata(work, m):
     m.add_unique("musicbrainz_workid", work.id)
-    if 'title' in work.children:
-        m.add_unique("work", work.title[0].text)
     if 'language' in work.children:
         m.add_unique("language", work.language[0].text)
+    if 'title' in work.children:
+        m.add_unique("work", work.title[0].text)
     if 'disambiguation' in work.children:
         m.add_unique("~workcomment", work.disambiguation[0].text)
     if 'relation_list' in work.children:
--- a/picard/mbxml.py
+++ b/picard/mbxml.py
@@ -108,6 +108,8 @@ def _relations_to_metadata(relation_lists, m):
                 if relation.type == 'performance':
                     performance_to_metadata(relation, m)
                     work_to_metadata(relation.work[0], m)
+                elif relation.type == 'parts':
+                    workparts_to_metadata(relation, m)
         elif relation_list.target_type == 'url':
             for relation in relation_list.relation:
                 if relation.type == 'amazon asin' and 'asin' not in m:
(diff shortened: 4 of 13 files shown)


samj1912:
@Sophist-UK tests are failing, please check

Sophist-UK:
Units tests should now work. This PR is now ready for review. 

However, there are two caveats:

a. For "Work is part no x of y parts" we need y in XML. I have submitted an MBS enhancement for it, In the mean time we use the maximum part number on the album for the container work.

b. This has only been tested on a release with a single level of "work is part of work" relationship, and where recordings only have a single work associated. I could do with some example releases of multi-level part-of combined with multiple works in a recording to see whether we need to be cleverer in determining the overall work,

Sophist-UK:
Spotted that - works on Mutagen 1.24 not 1.23 - fix submitted.

Sophist-UK:
Bugger - fix makes it work on Mutagen 1.23 but breaks it on 1.24!!!!

I am off to bed - fix it another day.

Sophist-UK:
One other potential issue - external documentation shows `(c)wrk` being used for `work` and we currently use `----:com.apple.iTunes:WORK`. Not sure what to do here - have used both and switched to add_unique so that loading will not result in duplicate values.

Sophist-UK:
Movement support is provided in Mutagen 1.37. So to provide this support, we need to bump minimum version to this.

INSTALL.txt
```diff
@@ -12,7 +12,7 @@ Before installing Picard, you need to have these:
  * PyQt 4.5 or newer
    http://www.riverbankcomputing.co.uk/software/pyqt/download
 
- * Mutagen 1.22 or newer (1.23 or newer for AIFF support)
+ * Mutagen 1.37 or newer
```
zas: 1.37 is required for MVNM/MVIN support only ?
Why is this exact version required exactly ?

I propose to stay on 1.22, and just add (Mutagen 1.37 or newer for MVNM/MVIN support), together with a version test in the code to enable/disable the support of those tags.

picard/log.py
```diff
@@ -48,7 +48,8 @@ def register_receiver(self, receiver):
         self._receivers.append(receiver)
 
     def unregister_receiver(self, receiver):
-        self._receivers.remove(receiver)
+        if receiver in self._receivers:
```
zas: This change is unrelated to the PR, please take it out.

picard/formats/id3.py
```diff
@@ -432,7 +446,7 @@ def _remove_deleted_tags(self, metadata, tags):
                         if frame.FrameID in ('TMCL', 'TIPL', 'IPLS'):
                             for people in frame.people:
                                 if people[0] == role:
-                                    frame.people.remove(people)
```
zas: Unrelated change.

picard/formats/id3.py
```diff
@@ -453,7 +467,7 @@ def _remove_deleted_tags(self, metadata, tags):
                         if frame.FrameID in ('TIPL', 'IPLS'):
                             for people in frame.people:
                                 if people[0] == role:
-                                    frame.people.remove(people)   
+                                    frame.people.remove(people)
```
zas: Unrelated change.

picard/mbxml.py
```diff
@@ -329,12 +331,31 @@ def performance_to_metadata(relation, m):
...
+
+
 def work_to_metadata(work, m):
     m.add_unique("musicbrainz_workid", work.id)
-    if 'language' in work.children:
-        m.add_unique("language", work.language[0].text)
     if 'title' in work.children:
         m.add_unique("work", work.title[0].text)
+    if 'language' in work.children:
+        m.add_unique("language", work.language[0].text)
```
zas: Is there a reason to move the language test block here ? If not, please don't.

picard/util/tags.py
```diff
@@ -88,6 +88,16 @@
     '~rating': N_('Rating'),
     'artists': N_('Artists'),
     'work': N_('Work'),
+    'originalartist': N_('Original Artist'),
+    'musicbrainz_originalartistid': N_('MusicBrainz Original Artist Id'),
+    'originalalbum': N_('Original Album'),
+    'musicbrainz_originalalbumid': N_('MusicBrainz Orignal Release Id'),
```
zas: 'MusicBrainz Orignal Release Id' -> 'MusicBrainz Original Release Id'

picard/util/tags.py
```diff
@@ -88,6 +88,16 @@
...
+    'movementname': N_('Movement name'),
```
zas: Other names have first letter capitalized on all words.

picard/util/tags.py
```diff
@@ -88,6 +88,16 @@
...
+    'musicbrainz_movementid': N_('MusicBrainz Movement Work Id'),
+    'movementnumber': N_('Movement number'),
```
zas: Other names have first letter capitalized on all words.

picard/util/tags.py
```diff
@@ -88,6 +88,16 @@
...
+    'movementtotal': N_('Movement total'),
```
zas: Other names have first letter capitalized on all words.

picard/log.py
Sophist-UK: It is related to the provision of logging in test_formats.

picard/formats/id3.py
Sophist-UK: I don't remember doing this - will back it out.

picard/mbxml.py
Sophist-UK: No - will move it back.

picard/util/tags.py
Sophist-UK: Will fix

picard/util/tags.py
Sophist-UK: Will fix.

picard/util/tags.py
Sophist-UK: Will fix.

picard/util/tags.py
Sophist-UK: Will fix.

Sophist-UK:
I tried loading https://musicbrainz.org/release/88ea0473-d785-4ef4-83d1-fed6bce61785 and it doesn;t work properly when a recording has several works assigned.

I think we need to process all the works first, and then process `parts` relationships afterwards.

ijabz:
Does this include support for equivalent mp4 fields @mvn, @mvi, @mvc ?
What about Work, adding movements with associated work doesnt have any effect in iTunes, AFAIK the work Picard currently adds to its work field is actually the movement rather than the work

Sophist-UK:
Yes - it adds support for mp4.
And yes - it will populate the Work with the overall work, and put the movement where it belongs.

But it gets complicated when a track has multiple movements from the same overall work or is a medley from one or more different overall works.

So I am just recoding to handle this.

I could do with an example of a multi-level part of hierarchy (if they exist) i.e. the recording work is part of a movement which is part of an overall work.

ijabz:
Here is an example of a recording that is a 'recording of' a movement (called work in musicbrainz) that is part of a work which in turn is part of a larger work and so on.

https://musicbrainz.org/recording/69037c4f-c0c8-4523-845b-a513bca69a03

Sophist-UK:
@ijabz There is bad news and good news:

Bad news: The xml returned by MB WS only gives the immediate containing work. So Picard cannot get the overall containing work.
Good news: The xml returned by MB WS only gives the immediate containing work. So I don't have to make the already reasonably complex code even more complex.

See Track 9 on https://musicbrainz.org/release/788806e9-2afe-474d-a574-60c3882c6b11

XML: https://musicbrainz.org/ws/2/release/788806e9-2afe-474d-a574-60c3882c6b11?inc=release-groups+media+recordings+artist-credits+artists+aliases+labels+isrcs+collections+artist-rels+release-rels+url-rels+recording-rels+work-rels+recording-level-rels+work-level-rels

ijabz:
Can you give example of the movement, work fields that would set for track 1 using that as example

Sophist-UK:
Track 9:
```
movementname: Götterdämmerung: Act III, Scene II. Trauermusik beim Tode Siegfrieds (Funeral March)
musicbrainz_movementworkid: b4d7c421-e730-3a74-8e28-be58b35e51ba

work: Götterdämmerung: Act III, Scene II
musicbrainz_workid: 2304132a-d28d-408c-b443-94a3822fef72
```
If you want to replace Work with root work i.e. 
```
work: Götterdämmerung, WWV 86D
musicbrainz_workid: c906f057-e479-418d-a96a-ffa639c8f8c8
```
then this requires one or more additional calls to MB WS to work your way up the hierarchy and can be achieved with a plugin (use my Artist Official Website plugin as a starting point).

ijabz:
That is okay, what I would ask you to consider though is that musicbrainz_movementworkid is also a new field not currently in Picard right ?

We already have a equivalent field in Jaikoz called MUSICBRAINZ_WORK_COMPOSITION_ID so it would be better to use this for better compatability.

https://docs.google.com/spreadsheets/d/1afugW3R1FRDN-mwt5SQLY4R7aLAu3RqzjN3pR1497Ok/edit#gid=0

Also can I ask what do you do if a recordng has a 'recording of' relationship to a work, but that work is not part of any other work. Does  anything get added to work fields in this case ?

Sophist-UK:
Work fields are left as current if we cannot identify a parent work, or if the recording is of two or more works and we cannot identify a common parent work.

I will look at the field names again - as better to change them before we go live, however I am not going to make changes which don't make sense.

I have looked at your spreadsheet, and could do with some explanations about Movement fields. Also I will be trying to make Picard tags more comprehensive over the next few months, so we might want to touch base on IRC and exchange email addresses.

Sophist-UK:
Hmm - I have no idea why 1 of 4 Travis runs failed - because it couldn't find PyQt4.

ijabz:
And what is current, my understanding is you put the work that the recording has a recording of relationship into the work and  musicbrainz_workid fields. If this is correct I think putting into the Work field is very confusing because any user interested in classical works  (but not particulary in MusicBrainz) would assume that if the Work was filled in this indicated a classical work with movements which it does not in this case. 

Happy to explain anything in the spreadsheet, what is your particular question about movement fields ?

ijabz:
You may find this discussion about work names useful https://community.metabrainz.org/t/how-to-explain-works/89790 if you havent see it.

Sophist-UK:
Thanks for the discussion pointer - however I wish it were that simple. 

Firstly you can have recordings which are two or more scenes from the same act or perhaps from two acts or maybe a scene work and an act work from different levels, or in the case of a medley from two or more pieces by different artists even. Some of these could be partial recordings.

Or there could be more than two sub-levels, three or maybe even four.

Then you need to think about how to tag multiple tracks from the same work - where you want work to be as descriptive as possible for the commonality of the tracks. So, an album containing all parts of a full work would want the full work as common description, whereas a multi-CD-set with one CD per act might want the common description to be the act - same for a (hypothetical) compilation CD with several tracks from Acts 1 of various operettas.

The bottom line is that there are a huge number of different scenarios, some of which make sense to have compositions, some of which have different relationships between main work and composition, and some of which don't make sense to have compositions. 

If iTunes had not already defined this, then I would have gone with your proposal to use Work as Scene and Composition as Main work, but we are too late. And I don't think it makes sense to use Composition for the sub-work because both main- and sub-works are compositions. So my vote is still for Movement and following the iTunes two level approach.

Sophist-UK:
@ijabz - Do you know if the iTunes Movement tags in ID3 and MP4 are multivalue?

Sophist-UK:
Ready for another review - because the algorithm is two-phase, the code is a bit more arcane.

Sophist-UK:
P.S. Travis issue is due to Travis configuration, NOT this PR. Or a glitch.

mineo:
Travis on macOS is failing because homebrews pyqt formula is now pyqt5 (https://github.com/Homebrew/homebrew-core/commit/11500473cedf4120d303aa271ac8fe8df939bbb6), which probably overrides the one from cartr/pyqt4 or something like that.

Sophist-UK:
@mineo I don't know how to fix this? Can you help?

ijabz:
Hi, i think we are talking  at cross purposes here because Im well aware of the iTunes tags and the main thrust of my argument is for Picard to respect the use of Work and Movement as used by iTunes and others.

Can you clarify something for me, my concern was that I thought if Picard had a recording with a 'recording of' relationship to a work and that work was not part of another work it would still add this work to the work field. If this is the case this is wrong wrt to iTunes since the work field should only be filled if it is a  work encompassing movements.  

But I have just tried Picard both with songs that are linked to such a work (should not add), and to songs linked to a work that is part of a work (should add) and in neither case can I get Picard to adds any works, so confused.

Sophist-UK:
@ijabz Picard currently saves any "recording of" relationships to Work - this predates the iTunes stuff so we are not going to change it.

If the work is part of another work, this PR moves the existing work to Movement Name and replaces it with the containing work.

Sophist-UK:
Thanks to @ilovezfs for help with fixing Travis.

ilovezfs:
You're welcome!

ijabz:
Right that s what I thought (although I don't understand why I cant get Picard to do this). So that is my point you are using WORK field to both represent an encompassing Work (like iTunes does) and a non-encompassing Work, which is not what iTunes does and is now what most users would consider a work at all in this context. Using the same field to mean different things is always a bad idea.

Work is conceptually understood as something containing movements you should not be filling in the Work field with single level MusicbrainzWorks it breaks compatability with any other work aware tagging application. That is the purpose of MUSICBRAINZ_WORK_COMPOSITION field to have a place to store any WORK linked directly to a recording regardless of what type it is, without messing up the metadata for other applications. I dont love the name MUSICBRAINZ_WORK_COMPOSITION but because we are tagging a song the name should be from the song perspective, and from a song perspective the work that the recording is directly linked to is always a composition of sorts, it is not always a movement/work ectera. 

I'm still not clear what fields PIcard is actually writing for ID3 and Mp4, for ID3 you need to write to TIT1 field for iTunes see the work whereas for mp4 you need to write to @wrk for it to see the work. With Songkong we write to TXXX:WORK (ID3), @wrk (mp4) for a multi-level work, then we have an option to allow the value to be copied to TIT1 for ID3 to allow compatability with iTunes, but we make this optional as many users are using TIT1 field for other things.

Sophist-UK:
My point is that Picard has already tagged millions of tracks with Work, and we shouldn't change to meet the iTunes spec. without careful consideration. Personally I am quite happy with the way this PR works, and indeed it works the same as iTunes.

©wrk is NOT currently supported by mutagen, though I have raised an issue for this. In Picard TIT1 is mapped to the `grouping` variable and this is a better match than `work` IMO.

ijabz:
I'm not saying you should change Work just because of iTunes, I'm saying you should change it because the way Picard does it makes little  sense. and is odd with everyone else, I raised an issue about this https://tickets.metabrainz.org/browse/PICARD-1050

I totally agree that it is more logical to map GROUPING to TIT1, http://blog.jthink.net/2016/11/the-reason-why-is-grouping-field-no.html . But if you don't provide an option to write work to TIT1 (and you don't write to @wrk for mp4) then actually you have not added proper iTunes compatabilty since as far as iTunes is concerned you  have only added the movement part without the work part.

Sophist-UK:
I don't think Picard claims to be fully compatible with iTunes - instead IMO we are compatible with iTunes where it makes sense.

©wrk support should be added to mp4 - but this is not directly related to this and should be in a separate PR (which I will do). Also, there will need to be compatibility code to handle the switch away from existing mp4 work field.

ijabz:
The i iTunes  movement, movementno, movementtotal all work all work together so if you don't plan to support them all  better to support none at all.

With this PR implemented for mp4s you will never set the @wrk field even when set the @mvi, @mvn and @mvc fields, if you wait for the mutagen fix then it will work correctly for these songs, but you will then have the issue that songs linked to only a single defined work will now be adding a a value for @wrk but not 
for  @mvi, @mvn and @mvc fields. Both situations are inconsistent and not helpful to users.

Sophist-UK:
@ijabz As I previously said, I agree with you regarding ©wrk - but IMO this needs to be in a separate PR.

At present this supports reading and writing of movement, movementnumber, movementtotal. It currently supports population of movement & movementnumber from MB data, but pending a change to MBS it cannot yet support population of movementtotal because the data is not available to Picard to do so.

It does only sets movementnumber when there is one movement i.e. if the recording has two movements from the same piece then it cannot set movement numbers because ID3 only supports a single movement number.

It does not set movement data at all if the recording is part of multiple overall works - though in this situation I suppose we could put all the containing works in work and all the recording works in movementname. I am happy to make this change if voted for.

ijabz:
"At present this supports reading and writing of movementname, movementnumber, movementtotal in ID3 and MP4 formats (which are the ones supported by iTunes).", yes but it doesnt set WORK, I understand this is a temporary issue for Mp4s, but what are you going to do about ID3. 

"It does not set movement data at all if the recording is part of multiple overall works - though in this situation I suppose we could put all the containing works in work and all the recording works in movementname. I am happy to make this change if voted for."

wrt I think when we have multiple parents either do nothing or pick just one, its too complicated to capture this level of complexity within an essentially flat tagging structure.

Sophist-UK:
See also https://tickets.metabrainz.org/browse/PICARD-1050 for further discussion about this.

Sophist-UK:
"Yes but it doesnt set WORK, I understand this is a temporary issue for Mp4s, but what are you going to do about ID3."

It DOES set Work (as metadata) and Work is already written to most formats including ID3 and MP4.

The fact that MP4 should write to (c)wrk and not a TXXX field is a bug - which is why it will be in a separate PR.

Sophist-UK:
Essentially the PR currently looks at all "recording of" relationships and if they have a common parent it uses that.

So one "recording of" work with multiple parents won't get Movement.

But two "recording of" works with a common parent will get Movement and common parent.

Two recording-of works each with two parents will get a Movement if there is one common parent but not if there are two common parents. This is a real corner case - because there won't be many of these.

However it would be easy to add all common parents rather than only if there is one.

ijabz:
>>"Yes but it doesnt set WORK, I understand this is a temporary issue for Mp4s, but what are you going to do about ID3."
>>It DOES set Work (as metadata) and Work is already written to most formats including ID3 and MP4.

I mean as far as iTunes is concerned it doesn't set WORK because iTunes uses TIT1 not TXXX:WORK so it wont see the work Picard is adding, which is weird as the other fields that Picard is adding (MVIN, MVNM) are iTunes fields. I quite agree it shoudn't do this by default and i wished iTunes used a different field for works bu there should be a way to set it for TIT1

Sophist-UK:
I am going to do nothing about where ID3 stores Work - IMO iTunes should not be using TIT1 for this.

ijabz:
Right I agree they shouldn't as well, but, how does that help Picard users who want to use iTunes, they are not going to change iTunes because of what you think. For a long time MusicBrainz Picard defined compilations differently to iTunes but eventually adopted the iTunes approach,

Sophist-UK:
Yes - and I was the one who fixed the iTunes Compilations code.

TCMP is an iTunes defined ID3 tag - not one defined by ID3 specification. Ditto for MVNM and MVIN. So iTunes can make their own definitions.

TIT1, however, is NOT defined by Apple, but instead has a formal definition which is closer to Album Genre than to Work:

"TIT1: The 'Content group description' frame is used if the sound belongs to a larger category of sounds/music. For example, classical music is often sorted in different musical sections (e.g. "Piano Concerto", "Weather - Hurricane")."

To confine it to the overall work is to constrain a lot of its flexibility. And since Work is already widely used, IMO iTunes should have used that. And of course in PICARD-1049 you point out that everyone but iTunes uses TXXX:WORK, and we use Work for non-classical pieces where it won;t be any sort of grouping for the tracks.

But if the Picard community says we should switch to TIT1 for Work, I will of course change the code. But I don't think they will want to become compatible with iTunes but completely incompatible with everyone else.

ijabz:
I only said it should be an option (write to TIT1 as well as TXXX:WORK) for those who use iTunes not the default.

Sophist-UK:
@ijabz  Hmm - yes I agree we could have a "Maximise Picard compatibility with iTunes" option.

Please feel free to write this once this PR has been merged.

ijabz:
@Sophist-UK I don't think they are multivalue fields in iTunes, but then I dont think iTunes supports multiple values for any fields really,

Sophist-UK:
Not asking about iTunes but about ID3 and MP4 - I suspect that they are multi-value in both due to the underlying types of tag used.

ijabz:
The ID3 tags are not text frames (i.e begin with T ) so I would consider to not be multivalue, and for mp4 I dont expect they were designed to be multivalue either.

Sophist-UK:
IIRC Mp4 tags are multi-value by definition. And since MVxx tags are based on ID3 common text ormat they are multivalue in id3v24 by default.

INSTALL.txt
lazka: fyi, all changes ever made to mutagen are relicensed, so theoretically any previous version is now gplv2+ as well. But IANAL...

PR #685, opened 2017-04-07
Add documentation for string conversion and load_json functions

samj1912:


picard/util/__init__.py
```diff
@@ -442,15 +442,14 @@ def union_sorted_lists(list1, list2):
...
-    return cgi.escape(string)
-
-
-def json_load(data):
-    return json.loads(bytes(data).decode())
-
-
 def convert_to_string(obj):
+    """
+    Appropriately converts the input `obj` to a string.
```
mineo: Looking at the documentation for google style docstrings, you're missing a blank line between each section here and in `json_load`.

picard/util/__init__.py
```diff
@@ -459,4 +458,20 @@ def convert_to_string(obj):
         return str(obj)
 
 
+def htmlescape(string):
+    return cgi.escape(string)
+
+
+def json_load(data):
```
mineo: Other functions in this file (`build_qurl`, `htmlescape`, ...) are named `verb_something`, not the other way around.

picard/util/__init__.py
```diff
@@ -459,4 +458,20 @@ def convert_to_string(obj):
...
+    """
+    Appropriately convert and returns a python dictionary from
```
mineo: +`a`

picard/util/__init__.py
```diff
@@ -459,4 +458,20 @@ def convert_to_string(obj):
...
+    json string.
```
mineo: According to https://tools.ietf.org/html/rfc7159#section-7 a JSON string is just a sequence of characters enclosed in quotation marks, the input expected should be described similarly to how the `json.loads` function does it.

picard/util/__init__.py
```diff
@@ -459,4 +458,20 @@ def convert_to_string(obj):
...
+    Args:
+        data (QByteArray): The json response
```
mineo: Since this uses `convert_to_string`, the input data can at least be of the same types that function supports.

PR #691, opened 2017-04-08
Fix crash on Browser lookups

samj1912:
Crash due to unexpected type string.
Bytes/str issue with QT


Sophist-UK:
I thought Python3's use of universal unicode was to make things easier not harder. :-)

PR #692, opened 2017-04-08
PICARD-1061: Use different plugins directory for v2

Sophist-UK:
Avoid unavoidable Python errors attempting to import plugins which are not designed for Picard v2 and which e.g. try to import PyQt4.

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -31,7 +31,7 @@ else:
     USER_DIR = os.environ.get("XDG_CONFIG_HOME", "~/.config")
 
 USER_DIR = os.path.join(
-    os.path.expanduser(USER_DIR), "MusicBrainz", "Picard"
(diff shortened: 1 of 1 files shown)


mineo:
See [PICARD-1059](https://tickets.metabrainz.org/browse/PICARD-1059). I think we need to first figure out how we want the upgrade path to 2.x to look like before merging things like this, because neither the current solution nor the one proposed here work really well.

Sophist-UK:
I guess what we really need is some upgrade code that upgrades a users plugins when they have installed v2. And this is regardless of whether we have separate directories or not.

We also need to cater for users who have manually installed plugins that we cannot automatically upgrade - and these could potentially have e.g. PyQt4 calls in them which would give errors.

The only time you **need** separate directories is if anyone is running v1 and v2 in parallel for some reason.

However, for all of the above, having a separate directory is going to make things easier to achieve and avoid error messages.

zas:
Plugins API V2 will be reworked

PR #694, opened 2017-04-10
PICARD-1062: Sort results by similarity 

samj1912:
Fixes https://tickets.metabrainz.org/browse/PICARD-1062

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -19,6 +19,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 import traceback
+from operator import itemgetter
 from PyQt5 import QtCore, QtNetwork
 from picard import config, log
 from picard.coverart import coverart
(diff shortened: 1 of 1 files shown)


picard/track.py
```diff
@@ -60,6 +60,9 @@ def __init__(self, id, album=None):
     def __repr__(self):
         return '<Track %s %r>' % (self.id, self.metadata["title"])
 
+    def __lt__(self, other):
```
zas: Wouldn't it make sense to implement other methods of the same family for consistency ?
```
  def __eq__(self, other):
  def __ne__(self, other):
  def __gt__(self, other):
  def __ge__(self, other):
  def __le__(self, other):
```

zas: At least,  `__eq__()`, `__lt__()` and `__le__()` are needed

samj1912: __lt__ and __le__ were implemented, I am leaving __eq__ as python does a default id(obj) for that, which is what is the intended behaviour for these items currently.

PR #695, opened 2017-04-10
PICARD-1043: Add support for iTunes Movement tags - master branch

Sophist-UK:
This resolves https://tickets.metabrainz.org/browse/PICARD-1043 for Picard v2.

See also #676 for 1.4.x branch equivalent.


Sophist-UK:
Seeking review - ready for merge.

test/test_formats.py
```diff
@@ -74,8 +79,12 @@ def setUp(self):
         shutil.copy(self.original, self.filename)
         config.setting = settings.copy()
         QtCore.QObject.tagger = FakeTagger()
+        self.log_messages = []
+        log.main_logger.unregister_receiver(log._stderr_receiver)
```
mineo: Why?

picard/mbxml.py
```diff
@@ -320,6 +350,33 @@ def recording_to_metadata(node, m, track=None):
         elif name == 'video' and nodes[0].text == 'true':
             m['~video'] = '1'
     m['~length'] = format_time(m.length)
+    # Analyse to see if we should populate the movement variables
+    if m.partsof:
+        target_work = None
+        for work, parts_of in m.partsof.items():
+            works = set([p['workid'] for p in parts_of])
```
mineo: You can drop the intermediate list here.

picard/mbxml.py
```diff
@@ -17,8 +17,27 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+# NOTES ON MOVEMENTS (because logic is not straight forward):
+# If a recording is a movement i.e. part of an overall work, then the movement fields
+# store the movement details and the work fields store details of the overall work.
+# The movement details consist of:
+#   Movement Work MBID
+#   Movement Title
```
mineo: This appears to be called Movement Name in the UI and the rest of this comment.

picard/mbxml.py
```diff
@@ -17,8 +17,27 @@
...
+#   Movement Number
+#   Movement Total (i.e. number of movments in work) - not currently available in MB xml
```
mineo: movements

picard/mbxml.py
```diff
@@ -320,6 +350,33 @@ def recording_to_metadata(node, m, track=None):
         elif name == 'video' and nodes[0].text == 'true':
             m['~video'] = '1'
     m['~length'] = format_time(m.length)
+    # Analyse to see if we should populate the movement variables
+    if m.partsof:
+        target_work = None
+        for work, parts_of in m.partsof.items():
+            works = set([p['workid'] for p in parts_of])
+            if target_work is None:
```
mineo: You can just initialize `target_work` to the empty set and skip this "is this the first iteration of the loop" check.

picard/mbxml.py
```diff
@@ -107,7 +126,17 @@ def _relations_to_metadata(relation_lists, m):
             for relation in relation_list.relation:
                 if relation.type == 'performance':
                     performance_to_metadata(relation, m)
+                    if ('work' not in relation.children
+                        or 'id' not in relation.work[0].attribs
+                        ):
+                        log.warning('MBXML: Recording work skipped - no target work MBID present')
```
mineo: On a general note, I think this and some other warnings do more harm than good because you're testing for the very unlikely case that a recording-work relationship of a recording doesn't have a work associated with it or that work doesn't have an MBID. If we start adding logging warnings on that level, we have a lot of code to add.

picard/mbxml.py
Sophist-UK: Agreed

picard/mbxml.py
Sophist-UK: Agreed

picard/mbxml.py
Sophist-UK: I am in two minds about this sort of warning message. On the one hand, MB XML is generally reliable, but on the other hand if we start to get these sorts of inconsistencies we would want to know about them.

But I am not adding these consistently across Picard in this PR, so happy to remove in order to be consistent. Please advise.

picard/mbxml.py
Sophist-UK: I did this deliberately as empty set means no common target i.e. the Union operations in the loop eventually result in an empty set - so needed something else to differentiate.

test/test_formats.py
Sophist-UK: To avoid e.g. warning messages appearing in console during test. But I don't see why that is necessary. Will delete.

Sophist-UK:
@mineo said:
> One complaint I have about this is that it's redefining what the musicbrainz work id tag stands for: it could either be the work id of the work the recording is a recording of or the (what the code comments call) container work. As an application developer, I can now run into the situation where a work tag is present, but no movement id tag. This means that either the work is not part of another work or the tags were written with an older version of Picard (or the relationships have not yet been added to MB yet, but let's ignore that one). We had a similar problem with the recording/track id situation and just left the old tag alone and added a new one.

Yes - I agonised over this myself. Unfortunately, Apple's implementation in iTunes is incompatible with what we have been doing for years - and this PR is trying to create compatibility with it i.e. open a Picard tagged track in iTunes and the fields populate properly. But on reflection I think you are right - we shouldn't be redefining the meaning of our Work tag.

Perhaps we should split this into two pieces - a Picard PR which defines the tags and provides support for reading and writing the tags, and a plugin which populates the tags from the webservice data.

Or equally admit that Picard is incompatible with iTunes Movements, and only provide support for reading and writing their non-standard tags.

> Another problem is that classical music is not the only part of MB where the parts of relationship is used - at least soundtracks also used it because the style guideline mentions it. I'm not sure how common the "movement" terminology is when talking about soundtracks - for movies, maybe, but for tv series soundtracks it sound very weird imho. But this also leads to another problem: soundtrack works, especially main themes, are reused, like on the first track of https://musicbrainz.org/release/cc9d60db-20a7-498c-ba79-be5ba6e63d39?tport=8000. This means that the first track is not considered a movement, whereas the others on that release are :(

This should actually be OK with the current implementation because where the recording work is part of more than one higher work, then no changes are made. But that was a design decision as I could equally have moved the recording work to Movement Name and created a list of overall works.

Sophist-UK:
@mineo was absolutely right - we should not compromise the way Picard works in order to pander to an extremely poor and incompatible iTunes implementation (and the description is being kind - see https://2manyrobots.com/YateResources/InAppHelp/English.lproj/pgs/iTunesGrouping.html for a description of how iTunes uses an additional GRP1 ID3 tag and changes the use of the TIT1 ID3 tag depending on music type - yurghhhhhh!!!!).

I have taken the decision to revert the code which populates the Movement fields from MB XML so the scope of this PR is now only to read and write iTunes fields and not populate them.

There are some further changes that are needed - to support ID3 GRP1 and to read and save the MP4 shwm tag. Also to address the remaining @mineo comments.

Sophist-UK:
Ready for rereview.

picard/formats/id3.py
```diff
@@ -143,6 +148,7 @@ class ID3File(File):
         'MusicBrainz Release Track Id': 'musicbrainz_trackid',
         'MusicBrainz Disc Id': 'musicbrainz_discid',
         'MusicBrainz Work Id': 'musicbrainz_workid',
+        'MusicBrainz Movement Work Id': 'musicbrainz_movementid',
```
samj1912: Is this needed?

picard/formats/mp4.py
```diff
@@ -78,6 +86,7 @@ class MP4File(File):
         "----:com.apple.iTunes:MusicBrainz Disc Id": "musicbrainz_discid",
         "----:com.apple.iTunes:MusicBrainz TRM Id": "musicbrainz_trmid",
         "----:com.apple.iTunes:MusicBrainz Work Id": "musicbrainz_workid",
+        "----:com.apple.iTunes:MusicBrainz Movement Work Id": "musicbrainz_movementid",
```
samj1912: Again, is this needed?

picard/util/tags.py
```diff
@@ -92,6 +92,14 @@
     'musicbrainz_originalartistid': N_('MusicBrainz Original Artist Id'),
     'originalalbum': N_('Original Album'),
     'musicbrainz_originalalbumid': N_('MusicBrainz Original Release Id'),
+    'movementname': N_('Movement Name (iTunes Classical)'),
+    'musicbrainz_movementid': N_('MusicBrainz Movement Work Id'),
```
samj1912: Same for this.

test/test_formats.py
```diff
@@ -491,6 +489,11 @@ class MP3Test(ID3Test):
         #'podcast': '1',
         #'podcasturl': 'Foo',
         #'show': 'Foo',
+        'work': 'Foo',
+        'movementname': 'Bar',
+        'movementnumber': '1',
+        'movementtotal': '2',
+        'itunesgrouping': 'Foo',
```
samj1912: Please add tests to add and delete all these movement tags. Just adding them to the dict does not check if it gets deleted or not.

picard/formats/id3.py
Sophist-UK: IMO yes. A plugin to replicate the iTunes functionality from MB data would move the Picard Work field to Movement so it needs a Work Id field to match,.

picard/formats/mp4.py
Sophist-UK: See above.

picard/util/tags.py
Sophist-UK: See above

test/test_formats.py
Sophist-UK: Will do.

Sophist-UK:
Ready for re-review.

samj1912:
This LGTM. @Sophist-UK Let me know if you want to update this, otherwise I am okay to merge.

samj1912:
@mineo can you re-review this please?

Sophist-UK:
I have no changes to make. I don't think that the decision on multi-level work fields impacts this because it is only an enabling PR needing a plugin to populate the fields.

picard/formats/mp4.py
```diff
@@ -182,7 +192,13 @@ def _save(self, filename, metadata):
...
                     tags[self.__r_int_tags[name]] = [int(value) for value in values]
-                except ValueError:
+                except (ValueError, TypeError) as e:
+                    log.error("MP4: Cannot save tag %s from %s as %r in %r: %s",
+                        self.__r_int_tags[name],
+                        name,
+                        [int(value) for value in values],
+                        filename,
+                        e)
                     pass
```
mineo: Remove the `pass` statement.

picard/formats/mp4.py
```diff
@@ -182,7 +192,13 @@ def _save(self, filename, metadata):
             elif name in self.__r_int_tags:
                 try:
                     tags[self.__r_int_tags[name]] = [int(value) for value in values]
-                except ValueError:
+                except (ValueError, TypeError) as e:
+                    log.error("MP4: Cannot save tag %s from %s as %r in %r: %s",
```
mineo: The only other call to `log.error` in this file doesn't have an `MP4` prefix in the message.

picard/formats/mp4.py
```diff
@@ -182,7 +192,13 @@ def _save(self, filename, metadata):
             elif name in self.__r_int_tags:
                 try:
                     tags[self.__r_int_tags[name]] = [int(value) for value in values]
-                except ValueError:
+                except (ValueError, TypeError) as e:
```
mineo: It's not clear to me why a `TypeError` would be raised here.

picard/util/tags.py
```diff
@@ -92,6 +92,15 @@
...
+    'movementnumber': N_('Movement Number (iTunes Classical)'),
+    'movementtotal': N_('Movement Total (iTunes Classical)'),
+    'itunesgrouping': N_('Grouping (iTunes Classical)'), # ID3 GRP1 introduced in iTunes 12.5.4.42 Dec 2016
+    'ituneswork': N_('Work (iTunes Classical)'), # mp4 ©wrk introduced in iTunes 2016
+    '~movementshow': N_('Movement Show (iTunes Classical)'),
+    '~workcomment': N_('Work Disambiguation'),
+    '~movementcomment': N_('Movement Disambiguation (iTunes Classical)'),
```
mineo: I can never remember if `~` tags are not written to files for all formats, or if id3 writes them under some circumstances, but if they're not written, how can this be an `iTunes Classical` tag, if it's only used internally by picard?

picard/util/tags.py
```diff
@@ -92,6 +92,15 @@
     'musicbrainz_originalartistid': N_('MusicBrainz Original Artist Id'),
     'originalalbum': N_('Original Album'),
     'musicbrainz_originalalbumid': N_('MusicBrainz Original Release Id'),
+    'movementname': N_('Movement Name (iTunes Classical)'),
+    'musicbrainz_movementid': N_('MusicBrainz Movement Work Id'),
+    'movementnumber': N_('Movement Number (iTunes Classical)'),
+    'movementtotal': N_('Movement Total (iTunes Classical)'),
+    'itunesgrouping': N_('Grouping (iTunes Classical)'), # ID3 GRP1 introduced in iTunes 12.5.4.42 Dec 2016
+    'ituneswork': N_('Work (iTunes Classical)'), # mp4 ©wrk introduced in iTunes 2016
+    '~movementshow': N_('Movement Show (iTunes Classical)'),
```
mineo: If this is written to files, remove the `~`.

picard/formats/mp4.py
Sophist-UK: I received a TypeError during testing so added it as an exception.

picard/formats/mp4.py
Sophist-UK: Will add to other call.

picard/formats/mp4.py
Sophist-UK: Agreed.

picard/util/tags.py
Sophist-UK: It is written to MP4 but not ID3 which is why I used a `~`. But on balance I agree it should be without it.

picard/util/tags.py
Sophist-UK: It is data available from iTunes but there are no tags to map it to. So it is not written to any formats unless user provides a tagging script to copy it to a variable which will end up as an e.g. TXXX tag.

Sophist-UK:
Requested changes made - ready for re-review.

sambhav:
@Sophist-UK, this PR has been pending for a while. You need to rebase this for 2.0. I am closing this for now. You can reopen this or start off fresh from master.

PR #696, opened 2017-04-10
PICARD-1063: Fix astrcmp tests if not compiled

Sophist-UK:
After #689 tests fail if astrcmp is not compiled.

Resolves https://tickets.metabrainz.org/browse/PICARD-1063

### DIFF
--- a/test/test_util_astrcmp.py
+++ b/test/test_util_astrcmp.py
@@ -3,7 +3,6 @@
 import os
 import os.path
 import unittest
-from picard.util.astrcmp import astrcmp_c, astrcmp_py
 
 
 class AstrcmpBase(object):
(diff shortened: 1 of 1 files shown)


test/test_util_astrcmp.py
```diff
@@ -20,9 +19,15 @@ def test_astrcmp(self):
         self.assertAlmostEqual(0.7083333333333333, astrcmp(u"The Great Gig in the Sky", u"Great Gig In The sky"))
 
 
-class AstrcmpCTest(AstrcmpBase, unittest.TestCase):
-    func = astrcmp_c
-
+from picard.util.astrcmp import astrcmp_py
```
samj1912: Why was this shifted from the top of the module?

test/test_util_astrcmp.py
```diff
@@ -20,9 +19,15 @@ def test_astrcmp(self):
...
 class AstrcmpPyTest(AstrcmpBase, unittest.TestCase):
     func = astrcmp_py
+
+try:
+    from picard.util.astrcmp import astrcmp_c
+
+    class AstrcmpCTest(AstrcmpBase, unittest.TestCase):
+        func = astrcmp_c
+except:
+    pass
```
samj1912: Add an endline to the file. Please rebase the above changes.

test/test_util_astrcmp.py
Sophist-UK: Because it goes with the test immediately below and compares and contrasts with the one for the C module.

Sophist-UK:
I hope that this is what you wanted.

Sophist-UK:
Agreed - I will make this change.

Sophist-UK:
@mineo Definitely a better solution than mine - fix pushed.

PR #699, opened 2017-04-10
PICARD-1065: Fix patch_version

samj1912:
https://tickets.metabrainz.org/browse/PICARD-1065


setup.py
```diff
@@ -584,7 +584,7 @@ def run(self):
 
     def patch_version(self, filename):
         regex = re.compile(r'^PICARD_BUILD_VERSION_STR\s*=.*$', re.MULTILINE)
-        with open(filename, 'r+b') as f:
```
zas: Doing so will possibly change end of lines of the read then written file. Instead if  i think we should keep binary mode, but convert source to unicode.

PR #700, opened 2017-04-10
Fix astrcmp tests when C version of astrcmp() isn't available

zas:


Sophist-UK:
Actually I quite like this version.

PR #709, opened 2017-04-26
PICARD-1075: Add unit tests for Metadata object variables

Sophist-UK:
Resolves https://tickets.metabrainz.org/browse/PICARD-1075

### DIFF
--- a/test/test_metadata.py
+++ b/test/test_metadata.py
@@ -1,12 +1,8 @@
 # -*- coding: utf-8 -*-
-import os.path
-import picard.formats
 import unittest
-import shutil
 
 from picard import config
-from picard.coverart.image import CoverArtImage, TagCoverArtImage
-from picard.metadata import Metadata
+from picard.metadata import Metadata, MULTI_VALUED_JOINER
 
 
 settings = {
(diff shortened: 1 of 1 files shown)


test/test_metadata.py
```diff
@@ -0,0 +1,112 @@
...
+        self.assertEqual(MULTI_VALUED_JOINER.join(self.multi1), self.metadata["multi1"])
+        self.assertEqual(MULTI_VALUED_JOINER.join(self.multi1), self.metadata.get("multi1"))
+        self.assertEqual(self.multi1, self.metadata.getall("multi1"))
+
+        self.assertEqual("", self.metadata["nonexistent"])
+        self.assertEqual(None, self.metadata.get("nonexistent"))
+        self.assertEqual([], self.metadata.getall("nonexistent"))
+
+        self.assertEqual(dict.items(self.metadata), self.metadata.rawitems())
+        metadata_items = [(x, z) for (x, y) in dict.items(self.metadata) for z in y]
```
samj1912: Any reason you are using this and not list(metadata.items( ))?

test/test_metadata.py
```diff
@@ -0,0 +1,112 @@
...
+        self.metadata.delete("single1")
+        m.update(self.metadata)
+        self.assertIn("old", m)
+        self.assertNotIn("single1", m)
+        self.assertIn("single1", m.deleted_tags)
+        self.assertEqual("single2-value", m["single2"])
+        self.assertEqual(self.metadata.deleted_tags, m.deleted_tags)
+
+        self.metadata["old"] = "old-value"
+        for (key, value) in dict.items(self.metadata):
```
samj1912: same for this and all the other dict.items instances

test/test_metadata.py
```diff
@@ -0,0 +1,112 @@
...
+        for (key, value) in dict.items(m):
+            self.assertIn(key, self.metadata)
+            self.assertEqual(value, dict.get(self.metadata, key))
+
+    def test_metadata_clear(self):
+        self.metadata.clear()
+        self.assertEqual(0, len(self.metadata))
+
+    def test_metadata_applyfunc(self):
+        func = lambda x: x[1:]
```
samj1912: Not really a fan of assigning lambda functions.

test/test_metadata.py
Sophist-UK: Yes - this is to create (using dict) the data that metadata.items is tested against. If I used metadata items to create it then it wouldn't be a valid test.

test/test_metadata.py
Sophist-UK: Same answer.

test/test_metadata.py
Sophist-UK: I want to use the same lambda function multiple times - and because it is a function to use as test data and not a function which is part of the testing functionality (if I have explained that well enough), this seemed to be the best approach. But I can change it if you want.

samj1912:
LGTM :+1:

PR #713, opened 2017-04-30
Remove unneeded unicode prefix

samj1912:


picard/const/attributes.py
```diff
@@ -3,101 +3,101 @@
 # Use `python setup.py update_constants` to update it.
```
mineo: That code needs to be updated.

PR #714, opened 2017-05-01
Reject functions with required keyword-only arguments

mineo:
Python 3 / PEP3102 introduced keyword-only arguments, like in

    def foo(a, *, b):
      pass

where `b` can't be set as a positional argument. This can't be expressed in
tagger script, so reject calls to register such functions.

This also gets rid of a DeprecationWarning for inspect.getargspec.

### DIFF
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -384,3 +384,17 @@ class ScriptParserTest(unittest.TestCase):
         self.parser.eval("$unset(performer:*)", context)
         self.assertNotIn('performer:bar', context)
         self.assertNotIn('performer:foo', context)
+
+    def test_required_kwonly_parameters(self):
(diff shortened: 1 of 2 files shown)


test/test_script.py
```diff
@@ -384,3 +384,17 @@ def test_cmd_unset_multi(self):
...
+    def test_required_kwonly_parameters(self):
+        def func(a, *, required_kwarg):
+            pass
+
+        with self.assertRaises(TypeError,
+                               msg="Functions with required keyword-only parameters are not supported"):
+            register_script_function(func)
+
+    def test_optional_kwonly_parameters(self):
+        def func(a, *, required_kwarg=1):
```
Sophist-UK: s/required_kwarg/optional_kwarg/

PR #721, opened 2017-05-09
Update setup script to allow submitting pip packages

samj1912:
Since all of Picard's requirements are available on pip,
this PR makes changes to the setup script to allow submitting a
distribution file to PyPi. A MANIFEST.in was included to add data dirs.
New packages can be submitted via python setup.py sdist upload -r pypi

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -76,3 +76,12 @@ To run the included tests, follow the instructions for "Running From
 the Source Tree".  Afterward you can run the tests using setup.py:
 
     python3 setup.py test
+
+Packaging
+---------
+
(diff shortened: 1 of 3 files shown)


zas:
> python setup.py sdist upload -r pypi

Perhaps worth to add it somewhere in embedded documents.

samj1912:
Another problem that popped up. Now that we updated to setuptools, there is an additional warning about normalization of versions.
2.0.0dev1 gets normalized to 2.0.0.dev1

It shows up as a warning everytime a command is run via setup.py (test, build* etc)

zas:
> We can update the version string to the above form to avoid errors.

Yes, but carefully check it doesn't break anything.

PR #724, opened 2017-05-10
Avoid mutable default arguments

samj1912:
See http://python-guide-pt-br.readthedocs.io/en/latest/writing/gotchas/#mutable-default-arguments


zas:
I hate python for this unintuitive behavior.

PR #728, opened 2017-05-11
Update @2x icons and add attribution to the About dialog

samj1912:


mwiencek:
Looks good on my MacBook.

PR #729, opened 2017-05-11
Support High DPI icons via SVG

samj1912:
Although requires less files, it has aliasing issues, atleast on my test run.
See https://usercontent.irccloud-cdn.com/file/hjiuhJOF/image.png

Maybe @mwiencek can try it out on OSX and see which version is better #729 or #728


mwiencek:
Looks pixely on my MacBook.

Also get some errors in the console:

```
Cannot open file ':/img-src/system-search.svg', because: No such file or directory
```

picard-icons-svg

samj1912:
Closed in favor of #728

PR #730, opened 2017-05-11
PICARD-1085: Fix multi-directory load on Windows

Sophist-UK:
On Windows on Qt5 it appears you need to avoid native dialog.

As far as I can tell, this was actually a bug under Qt4 (because native
dialog wasn't used anyway) and it should always have been specified.

I have only added win32 to the exception list. This may be required on
all platforms, but I am unable to test this.

Resolves https://tickets.metabrainz.org/browse/PICARD-1085

### DIFF
--- a/picard/ui/util.py
+++ b/picard/ui/util.py
@@ -104,7 +104,7 @@ class MultiDirsSelectDialog(QtWidgets.QFileDialog):
         super(MultiDirsSelectDialog, self).__init__(*args)
         self.setFileMode(self.Directory)
         self.setOption(self.ShowDirsOnly)
-        if sys.platform == "darwin":
+        if sys.platform in ["darwin", "win32"]:
             # The native dialog doesn't allow selecting >1 directory
             self.setOption(self.DontUseNativeDialog)
         for view in self.findChildren((QtWidgets.QListView, QtWidgets.QTreeView)):


picard/ui/util.py
```diff
@@ -104,7 +104,7 @@ def __init__(self, *args):
         super(MultiDirsSelectDialog, self).__init__(*args)
         self.setFileMode(self.Directory)
         self.setOption(self.ShowDirsOnly)
-        if sys.platform == "darwin":
+        if sys.platform in ["darwin", "win32"]:
```
Sophist-UK: I have no idea what happens on Linux. All I know is that Windows used to get non-native dialog and in QT5 it didn't - so I fixed it for Windows and left it for Linux because I had no knowledge to do anything different.

As noted, Windows native dialog doesn't provide multi-select under Qt4 or Qt5. Qt4 apparently gave non-native dialog as standard whilst Qt5 doesn't.

samj1912:
I am not sure about this. Given that I did explore the unified file dialog and it turned out that Qt has horrible support for a combined X-platform file/folder dialog. I am slightly inclined to merge this.Thoughts on it @zas  @mineo ?

Sophist-UK:
This PR is a bug fix to make P2 work like P1 and not a functionality change, and is already specific only to Windows and should not impact Linux. If we want to make a functional change, IMO that should be done in a separate PR.

PR #731, opened 2017-05-12
Change directory and config names to correspond to dev version

samj1912:


picard/const/__init__.py
```diff
@@ -31,7 +31,7 @@
     USER_DIR = os.environ.get("XDG_CONFIG_HOME", "~/.config")
 
 USER_DIR = os.path.join(
-    os.path.expanduser(USER_DIR), "MusicBrainz", "Picard"
+    os.path.expanduser(USER_DIR), "MusicBrainz", "Picard Dev"
```
zas: Can't we use `PICARD_APP_NAME` ?

setup.py
```diff
@@ -662,7 +662,7 @@ def _picard_packages():
         'regen_pot_file': picard_regen_pot_file,
         'patch_version': picard_patch_version,
     },
-    'scripts': ['scripts/picard'],
+    'scripts': ['scripts/picard_dev'],
```
zas: I think we should define it once at top, and use variables instead.

mineo:
How are people running development builds supposed to find out about this change which makes their Picard looks like it's been reset?

samj1912:
@mineo PW is to be updated with the info before release. We will add this over there and link it to the blog announcing the release of Dev version.

Sophist-UK:
My concern with this is that the change in name is hard coded.

Can we not determine from the version whether we are running dev and change the name by code logic.

samj1912:
@Sophist-UK, we can and it will hardly take an if else, but I wonder the value of this addition? The builds are packaged after a version change and we can very well change the constants that time. 

But if others feel it necessary to add the logic to determine this automatically, I don't mind doing the addition.

Sophist-UK:
It's just so that at release time we only need to change one thing i.e. version number.

PR #732, opened 2017-05-13
Detect package name from version string

samj1912:


setup.py
```diff
@@ -35,8 +35,13 @@
...
 
-# Change to 'picard' when building for production
-PACKAGE_NAME = "picard_dev"
+from picard import PICARD_VERSION
+
+# Decides package name on the basis of version
+if PICARD_VERSION[3] == 'dev':
+	PACKAGE_NAME = "picard_dev"
+else:
+	PACKAGE_NAME = "picard"
```
Freso: The rest of the file uses 4-space indentation, but here you use tab indentation…?

PR #733, opened 2017-05-13
Remove unused imports

samj1912:


picard/ui/ui_infodialog.py
```diff
@@ -16,7 +16,6 @@ def setupUi(self, InfoDialog):
         self.info_tab = QtWidgets.QWidget()
         self.info_tab.setObjectName("info_tab")
         self.vboxlayout = QtWidgets.QVBoxLayout(self.info_tab)
-        self.vboxlayout.setContentsMargins(0, 0, 0, 0)
```
mineo: This is not an import.

samj1912: I regenerated the UI files from .ui files. Guess that isn't needed anymore and wasn't updated in the repo. Should I remove it from this PR and make another one?

picard/config_upgrade.py
```diff
@@ -191,7 +191,6 @@ def upgrade_to_v1_4_0_dev_4():
 def upgrade_to_v1_4_0_dev_5():
     """Using Picard.ini configuration file on all platforms"""
     # this is done in Config.__init__()
-    pass
```
mineo: I think a function declaration without any function body is actually invalid syntax.

picard/ui/ui_infodialog.py
mineo: Hm, calling `build_ui` on master doesn't remove it, but calling it on this branch doesn't add it back in again, so it's not a big problem.

picard/config_upgrade.py
samj1912: If it has a doc string then it is not.

PR #738, opened 2017-05-14
Picard 1087: Lookup CD UI improvements

Sophist-UK:
1. Improve menu when more than one CD drive
2. Give warning message if discid not found or no CD drive.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -421,6 +421,13 @@ class MainWindow(QtWidgets.QMainWindow):
         # TR: Keyboard shortcut for "Lookup CD"
         self.cd_lookup_action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
         self.cd_lookup_action.triggered.connect(self.tagger.lookup_cd)
+        self.cd_lookup_action.setEnabled(False)
+        if discid is None:
+            log.warning("CDROM: discid library not found - Lookup CD functionality disabled")
+        elif len(get_cdrom_drives()) == 0:
+            log.warning("CDROM: No CD-ROM drives found - Lookup CD functionality disabled")
+        else:
+            self.cd_lookup_action.setEnabled(True)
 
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -422,6 +422,24 @@ def create_actions(self):
...
         self.cd_lookup_action.triggered.connect(self.tagger.lookup_cd)
 
+        self.cd_lookup_menu = QtWidgets.QMenu(_("Lookup &CD..."))
+        self.cd_lookup_menu.triggered.connect(self.tagger.lookup_cd)
+        self.cd_lookup_action.setEnabled(False)
+        if discid is None:
+            log.warning("CDROM: discid library not found - Lookup CD functionality disabled")
+        else:
+            drives = get_cdrom_drives()
+            if len(drives) == 0:
```
zas: `if not drives:` would be more pythonic

picard/ui/mainwindow.py
```diff
@@ -422,6 +422,24 @@ def create_actions(self):
...
+                log.warning("CDROM: No CD-ROM drives found - Lookup CD functionality disabled")
+            else:
+                shortcut_drive = config.setting["cd_lookup_device"].split(",")[0] if len(drives) > 1 else ""
+                self.cd_lookup_action.setEnabled(True)
+                for drive in drives:
+                    action = self.cd_lookup_menu.addAction(drive)
+                    if drive == shortcut_drive:
+                        action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
```
zas: The same shortcut is used at line 422:
```python
self.cd_lookup_action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
```
I think it would be more obvious to have only one place for this shortcut, instead of defining then undefining

Edit: understood, but a bit late

picard/ui/mainwindow.py
Sophist-UK: Agreed.

picard/ui/mainwindow.py
Sophist-UK: I will swap these lines.

PR #751, opened 2017-06-07
PICARD-1106: Discard the second return value of QFileDialog.getOpenFileNames

mineo:
getOpenFileNames was previously called getOpenFileNamesAndFilter (see [0]) and
returns a tuple (filenames, filter). Since we're not interested in the filter,
discard it.

This is similar to other fixes already done for QFileDialogs in mainwindow.py
and fingerprinting.py.

[0] http://pyqt.sourceforge.net/Docs/PyQt5/pyqt4_differences.html#qfiledialog

### DIFF
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -306,7 +306,7 @@ class PluginsOptionsPage(OptionsPage):
         self.ui.details.setText("<p>%s</p>" % "<br/>\n".join(text))
 
     def open_plugins(self):
-        files = QtWidgets.QFileDialog.getOpenFileNames(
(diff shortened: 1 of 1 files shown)


picard/ui/options/plugins.py
```diff
@@ -306,7 +306,7 @@ def change_details(self):
         self.ui.details.setText("<p>%s</p>" % "<br/>\n".join(text))
 
     def open_plugins(self):
-        files = QtWidgets.QFileDialog.getOpenFileNames(
+        files, _ = QtWidgets.QFileDialog.getOpenFileNames(
```
samj1912: Not so sure about the use of "_" for discarding because of its use as a builtin for gettext.

But other than that LGTM

PR #756, opened 2017-07-11
Fix variable names and error handling in WebService module

samj1912:


picard/webservice/__init__.py
```diff
@@ -223,46 +233,51 @@ def _handle_reply(self, reply, request, handler, parse_response_format, refresh)
...
+                        response_parser = self.get_response_parser(parse_response_type)
                     except UnknownResponseParserError as e:
-                        log.error("Attempting to parse the response body without a proper parser")
+                        log.error(e.args[0])
                         document = reply.readAll()
+                    else:
+                        try:
+                            document = response_parser(reply)
+                        except TypeError as e:
+                            log.error("Unable to parse the response. %s" % e.message)
```
mineo: The logger does the formatting itself, you don't necessarily need to do this here.

picard/webservice/__init__.py
```diff
@@ -223,46 +233,51 @@ def _handle_reply(self, reply, request, handler, parse_response_format, refresh)
...
-                        document = response_parser(reply)
+                        response_parser = self.get_response_parser(parse_response_type)
                     except UnknownResponseParserError as e:
-                        log.error("Attempting to parse the response body without a proper parser")
+                        log.error(e.args[0])
                         document = reply.readAll()
+                    else:
+                        try:
+                            document = response_parser(reply)
+                        except TypeError as e:
```
mineo: If an exception other than a TypeError is raised, document will still be undefined later on. It's also not entirely clear to me why TypeError is special here.

samj1912: Type error will be raised in case the response parser is not a callable

PR #759, opened 2017-07-12
Add helper functions for the JSON MB API

samj1912:


picard/mbjson.py
```diff
@@ -0,0 +1,473 @@
...
+_REPLACE_MAP = {}
+_EXTRA_ATTRS = ['guest', 'additional', 'minor']
+_BLANK_SPECIAL_RELTYPES = {'vocal': 'vocals'}
+
+
+def _parse_attributes(attrs, reltype):
+    attrs = [_decamelcase(_REPLACE_MAP.get(a, a)) for a in attrs]
+    prefix = ' '.join([a for a in attrs if a in _EXTRA_ATTRS])
+    attrs = [a for a in attrs if a not in _EXTRA_ATTRS]
+    if len(attrs) > 1:
```
zas: Nitpicking: calculate `len(attrs)` once, and cache

test/test_mbjson.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertEqual(formats, '12" Vinyl')
+
+
+class RecordingTest(unittest.TestCase):
+
+    def setUp(self):
+        config.setting = settings
+        self.json_doc = None
+        with open(os.path.join('test', 'data', 'ws_data', 'recording.json')) as f:
+            self.json_doc = json.loads(f.read())
```
zas: You may use `json.load(f)` instead of `json.loads(f.read())` here

test/test_mbjson.py
```diff
@@ -0,0 +1,182 @@
...
+        self.assertEqual(m['~recordingtitle'], 'Thinking Out Loud')
+
+
+class TrackTest(unittest.TestCase):
+
+    def setUp(self):
+        config.setting = settings
+        self.json_doc = None
+        with open(os.path.join('test', 'data', 'ws_data', 'track.json')) as f:
+            self.json_doc = json.loads(f.read())
```
zas: `json.load(f)` again
`os.path.join('test', 'data', 'ws_data')` is repeated at lot

I think you could do something like (untested):

```python
class JsonTests(object):
    def setup(self, filename):
        config.setting = settings
        self.json_doc = None
        with open(os.path.join('test', 'data', 'ws_data', filename)) as f:
            self.json_doc = json.load(f)

class TrackTest(unittest.TestCase, JsonTests):

    def setUp(self):
        self.setup('track.json')
...
````

File: test/test_mbjson.py [hunk 4]
```diff
@@ -0,0 +1,167 @@
...
+class TrackTest(MBJSONTest):
+
+    def setUp(self):
+        self.init_test('track.json')
+
+    def test_track(self):
+        t = Track("1")
+        m = t.metadata
+        track_to_metadata(self.json_doc, t)
+        self.assertEqual(m['discsubtitle'], 'Speak to Me')
```
mineo: This is wrong, `discsubtitle` is the subtitle of the disc, not the trck title.

test/test_mbjson.py
```diff
@@ -0,0 +1,167 @@
...
+        self.assertEqual(m['musicbrainz_recordingid'], 'bef3fddb-5aca-49f5-b2fd-d56a23268d63')
+        self.assertEqual(m['musicbrainz_trackid'], 'd4156411-b884-368f-a4cb-7c0101a557a2')
+        self.assertEqual(m['title'], 'Speak to Me')
+        self.assertEqual(m['~length'], '1:08')
+        self.assertEqual(m['~musicbrainz_tracknumber'], 'A1')
```
mineo: This should also test the standard `tracknumber` tag.

picard/mbjson.py
```diff
@@ -0,0 +1,460 @@
...
+    "sound": "engineer",
+    "live sound": "engineer",
+    "mix": "mixer",
+    #"Recording": "engineer",
+    "mix-DJ": "djmixer",
+}
+
+_TRACK_TO_METADATA = {
+    'title': 'discsubtitle',
+    'position': 'tracknumber',
```
mineo: Tracks in the JSON output do not appear to have a `position`, only a `number`. It appears the JSON web service is not aware of the changes from https://tickets.metabrainz.org/browse/MBS-842: compare https://musicbrainz.org/ws/2/release/b84ee12a-09ef-421b-82de-0441a926375b?inc=recordings&fmt=json with https://musicbrainz.org/ws/2/release/b84ee12a-09ef-421b-82de-0441a926375b?inc=recordings&fmt=xml, the JSON output only has `number`.

samj1912: This was already reported and fix in the beta. It will roll out to the main site this Monday.

samj1912: see https://tickets.metabrainz.org/browse/MBS-9403

PR #763, opened 2017-07-18
Display astrcmp implementation (C or Python) in About and debug log

zas:
It can also be displayed using -V command line option along various libs versions


samj1912:
LGTM

PR #764, opened 2017-07-18
Build repo pushes only on master and 1.4.x branch

samj1912:


mineo:
Does this still build pull requests? I couldn't figure that out from the docs :( But if so, +1.

I'd prefer to not have a lot of branches in the metabrainz repository in the first place because (realistically) noone's going to clean them up and everyone gets a ton of useless branches on a clone.

PR #771, opened 2017-07-23
Fix error while fetching collections in a release group

samj1912:


picard/mbjson.py
```diff
@@ -273,9 +273,9 @@ def media_formats_from_node(node):
     formats_count = {}
     formats_order = []
     for medium in node:
-        try:
-            text = medium['format']
-        except KeyError:
+
+        text = medium.get('format', "(unknown)")
```
zas: I think you can replace:
```python
text = medium.get('format', "(unknown)")
if not text:
    text = "(unknown)"
```

by:

```python
text = medium.get('format', "") or "(unknown)"
```

PR #773, opened 2017-07-25
Fix null catalog-number and/or label cases, add tests

zas:


test/test_mbjson.py
```diff
@@ -166,3 +172,25 @@ def setUp(self):
...
         self.assertEqual(['GB'], country_list)
+
+
+class LabelInfoTest(MBJSONTest):
+
+    def setUp(self):
+        self.init_test('label_info.json')
+
+    def test_label_info_from_node_0(self):
+        label_info = label_info_from_node(self.json_doc['releases'][0]['label-info'])
```
samj1912: You might want to make self.json_doc['releases'] into a releases variable.

Or make a private function 
```python
def _get_label_info_node(json_doc, i):
    return json_doc['releases'][i]['label-info']
```

samj1912:
Appveyor is failing since the commit is not mergable.
Rebasing from master should fix it.

PR #776, opened 2017-07-26
Fix null type errors while parsing API response

samj1912:


picard/mbjson.py
```diff
@@ -248,9 +248,12 @@ def country_list_from_node(node):
         country = []
         for release_event in node['release-events']:
             try:
-                country.append(release_event['area']['iso-3166-1-codes'][0])
-            except KeyError:
+                country_code = release_event['area']['iso-3166-1-codes'][0]
+            # TypeError in case object is None
+            except (KeyError, IndexError, TypeError):
                 pass
+            if country_code:
```
zas: This should be in the try/except block after `country_code = ...` else it may be undefined

picard/mbjson.py
```diff
@@ -260,7 +263,7 @@ def label_info_from_node(node):
     for label_info in node:
         if 'label' in label_info and label_info['label'] and 'name' in label_info['label']:
             label = label_info['label']['name']
-            if label not in labels:
+            if label not in labels and label:
```
zas: Better test `label` first imho: `if label and label not in labels:`

PR #783, opened 2017-09-23
PICARD-1140: Add a confirmation box while exiting the picard

vishichoudhary:
Add a confirmation box while exiting the picard. So user will not loose his/her data if he/she accidently hit either alt+f4 or exit button.


zas:
There is already a Quit Confirmation, see https://github.com/vishichoudhary/picard/blob/f6bc695141cc576422e3143c6f2374f35c4f983f/picard/ui/mainwindow.py#L150
An option allows one to enable/disable it though.

PR #785, opened 2017-10-04
PICARD-1143: Prevent loading .smbdelete* files

Sophist-UK:
.smbdelete files are temporary versions of files that should have been deleted but cannot be due to  samba bug(?). We should not load these files even if we have ignore-hidden-files unchecked.

Note: I am not running on a Mac and have not tested this code.

This PR resolves [PICARD-1143](https://tickets.metabrainz.org/browse/PICARD-1143).

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -367,7 +367,7 @@ class Tagger(QtWidgets.QApplication):
             if ignore_hidden and is_hidden(filename):
                 log.debug("File ignored (hidden): %r" % (filename))
                 continue
-            # Ignore .smbdelete* files which Applie iOS creates by renaming a file when it cannot delete it
+            # Ignore .smbdelete* files which Applie iOS SMB creates by renaming a file when it cannot delete it
             if os.path.basename(filename).startswith(".smbdelete"):
                 log.debug("File ignored (.smbdelete): %r", filename)
                 continue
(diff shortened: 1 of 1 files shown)


picard/tagger.py
```diff
@@ -367,6 +367,9 @@ def add_files(self, filenames, target=None):
             if ignore_hidden and is_hidden(filename):
                 log.debug("File ignored (hidden): %r" % (filename))
                 continue
+            if os.path.basename(filename).startswith(".smbdelete") and is_hidden(filename):
```
samj1912: Does not make sense to have just 1 extension as a special case. Can we instead create a new constant list in picard.constant, and check force all the files with the extension in that extension list?

Sophist-UK: I agree in general - but can we leave it until we have a second case to create a general solution?

samj1912: Okay, can you add a comment explaining this special case then? And a TODO for future reference when we do need to shift to a general case

zas: Is there a specific need to test if the file is hidden, i think we want to ignore all files started with `.smbdelete` whether they are hidden or not.
For reference, this kind of files are created by http://www.opensource.apple.com/source/smb/smb-759.40.1/kernel/smbfs/smbfs_smb.c

About handling general case, no need for now, just add a comment about why we need to ignore such files.

PR #786, opened 2017-10-18
PICARD-1147: Handle media with only data tracks

mineo:
In the JSON representation, these media do not have a 'tracks' element
containing the list of tracks.

@samj1912 I'm not sure if this is intentional in the JSON format and needs to be handled here or in MBS.


samj1912:
LGTM

PR #790, opened 2017-12-06
PICARD-1159: Don't pass filenames as bytes to wave.open

mineo:
wave.open treats everything that's not a str as a file-like object.
encode_filename returns bytes, but they're not file-like.


vishichoudhary:
@mineo if you don't have time . Can i rebase this pr?

PR #793, opened 2017-12-22
Add a PR template

samj1912:


mineo:
How does this relate to the existing CONTRIBUTING.md, which also is only about code and pull requests?

Sophist-UK:
I think this is the default text provided when you submit a PR.

samj1912:
@mineo yes, this is shown in the PR text box. Makes it easy to add info to PRs.

PR #797, opened 2018-01-17
Minor code cleanup, related to track durations/lengths

zas:


picard/metadata.py
```diff
@@ -30,6 +30,21 @@
...
+# 20000 0     0.333333333333
+# 20000 10000 0.666666666667
+# 20000 20000 1.0
+# 20000 30000 0.666666666667
+# 20000 40000 0.333333333333
+# 20000 50000 0.0
+LENGTH_SCORE_THRES_MS = 30000
+
+def length_score(a, b):
+    return (1.0 - min(abs(a - b), LENGTH_SCORE_THRES_MS) / float(LENGTH_SCORE_THRES_MS)
```
samj1912: Missing a bracket here.

picard/metadata.py
```diff
@@ -30,6 +30,21 @@
...
+# equal lengths will give a score of 1.0
+# example
+# a     b     score
+# 20000 0     0.333333333333
+# 20000 10000 0.666666666667
+# 20000 20000 1.0
+# 20000 30000 0.666666666667
+# 20000 40000 0.333333333333
+# 20000 50000 0.0
+LENGTH_SCORE_THRES_MS = 30000
```
samj1912: We can move this to picard.constants?

picard/metadata.py
```diff
@@ -30,6 +30,21 @@
...
+
+def length_score(a, b):
```
samj1912: Not sure if this should be a public functions, maybe _length_score or maybe move this to picard.util?

picard/metadata.py
zas: I don't think it is needed, this value is very unlikely to be used elsewhere. If it happens in the future, we'll move it to picard.constants.

picard/metadata.py
zas: I'll make it a static method

picard/metadata.py
zas: Fixed, thanks

picard/ui/metadatabox.py
```diff
@@ -527,7 +526,10 @@ def _update_items(self, result=None, error=None):
                 self.setItem(i, 2, new_item)
             tag_item.setText(display_tag_name(name))
             self.set_item_value(orig_item, self.tag_diff.orig, name)
-            new_item.setFlags(orig_flags if length else new_flags)
+            if name == "~length":
+                new_item.setFlags(orig_flags)
+            else
```
samj1912: Missing colon

samj1912:
Dont mean to be picky but `Fix missing semicolon` can you amend the commit message to a colon?

zas:
> Dont mean to be picky but Fix missing semicolon can you amend the commit message to a colon?

I rebased and squashed typos.

picard/metadata.py
mineo: Maybe I'm stupid, but none of these these calculations appear to be correct (and I'm not sure how they should be with the current implementation of `length_score`):

```
In [3]:     def length_score(a, b):
   ...:         return (0.0 - min(abs(a - b), LENGTH_SCORE_THRES_MS) /
   ...:                 float(LENGTH_SCORE_THRES_MS))
   ...:     
In [7]: for a,b in [(20000, 0), (20000, 10000), (20000, 20000), (20000, 30000), (20000, 40000), (20000, 50000)]:
   ...:     print(f"{a}, {b} = {length_score(a,b)}")
   ...:     
20000, 0 = -0.6666666666666666
20000, 10000 = -0.3333333333333333
20000, 20000 = 0.0
20000, 30000 = -0.3333333333333333
20000, 40000 = -0.6666666666666666
20000, 50000 = -1.0
```

In fact, if the score needs to be between 0.0 and 1.0, how is substracting an always positive number from 0.0 going to achieve that (no, changing `-` to `+` doesn't get you the numbers from the docstring).

samj1912: I think this happened sometime later while rebasing. It was 1.0 when I reviewed it.

PR #798, opened 2018-01-17
Fix removal of ~length from sets, incorrect single-element tuple syntax

zas:
The `discard` argument is expected to be a sequence, but the syntax `discard=('elem')` is equivalent to `discard='elem'`.
A trailing comma is needed in this case: `discard=('elem',)` so discard is actually a tuple with a single element.


samj1912:
:+1:  LGTM

PR #799, opened 2018-01-18
[PICARD-1174] Add an option to control the tolerance of track times difference

zas:
The option `ignore_track_duration_difference_under` can be set through Advanced options panel.
If old and new Length tag have a difference under this value, they are considered equal, and this isn't accounted as a track difference.

The default value remains the same (2 seconds).

https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-1174


picard/ui/metadatabox.py
```diff
@@ -427,7 +428,7 @@ def _update_tags(self):
         if not (files or tracks):
             return None
 
-        tag_diff = TagDiff()
+        tag_diff = TagDiff(max_length_diff=config.setting["ignore_track_duration_difference_under"])
```
samj1912: Any reason to initialize the value here instead of just fetching it in the __init__ for TagDiff?

picard/ui/metadatabox.py
```diff
@@ -96,18 +96,19 @@ def display_value(self, tag):
 
 class TagDiff(object):
 
-    __slots__ = ("tag_names", "new", "orig", "status", "objects")
+    __slots__ = ("tag_names", "new", "orig", "status", "objects", "max_length_delta_ms")
 
-    def __init__(self):
+    def __init__(self, max_length_diff=4):
```
samj1912: Do we really need this to be a KW argument? Any reason why we are not just setting the value below at line 107 from config.setting ? It has a default value of 2, so it won't be undefined anytime?

samj1912:
Can you rebase the branch against master?

picard/ui/metadatabox.py
obtext: Also, if it _is_ to be a KW argument, the default should probably be 2.

picard/ui/metadatabox.py
zas: > Any reason to initialize the value here instead of just fetching it in the __init__ for TagDiff?

I prefer to not link the class TagDiff to config options directly.

picard/ui/metadatabox.py
zas: > Do we really need this to be a KW argument? Any reason why we are not just setting the value below at line 107 from config.setting ? It has a default value of 2, so it won't be undefined anytime?

I prefer to make this argument optional with a safe default.

zas:
> Can you rebase the branch against master?

Done, and fixed the default (2 vs 4).

picard/ui/ui_options_advanced.py
```diff
@@ -51,18 +80,19 @@ def setupUi(self, AdvancedOptionsPage):
...
+        self.vboxlayout.addItem(spacerItem1)
 
         self.retranslateUi(AdvancedOptionsPage)
         QtCore.QMetaObject.connectSlotsByName(AdvancedOptionsPage)
 
     def retranslateUi(self, AdvancedOptionsPage):
         _translate = QtCore.QCoreApplication.translate
         self.groupBox.setTitle(_("Advanced options"))
+        self.recursively_add_files.setText(_("Recursively add files and folders from directory"))
+        self.label_track_duration_diff.setText(_("Ignore track duration difference under this number of seconds "))
```
obtext: Remove trailing space?

zas: > Remove trailing space?

Done, thanks for noticing it.

PR #800, opened 2018-01-18
Re-define or set tabstops to improve keyboard navigation through options

zas:


samj1912:
LGTM, rebase and merge

PR #803, opened 2018-01-20
Simplify search dialog save/restore states code

zas:
Track/Album/ArtistSearchDialog share the same code, move it to parent class, and reduce code redundacy.


samj1912:
LGTM

PR #804, opened 2018-01-20
PICARD-1177 : Add a option to remove plugins

vishichoudhary:
# Summary

* This is a…
    * [ ] Bug fix
    * [X] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **This will add a button to uninstall a plugin**:

# Problem

* JIRA ticket (_optional_): [PICARD-1177](https://tickets.metabrainz.org/browse/PICARD-1177)

# Solution

Added a new column in Plugin gui , to uninstall a plugin . Added a confirmation box too

# Action

### DIFF
--- a/picard/releasegroup.py
+++ b/picard/releasegroup.py
@@ -77,25 +77,14 @@ class ReleaseGroup(DataObject):
                 "id":      node['id'],
                 "year":    node['date'][:4] if "date" in node else "????",
                 "country": "+".join(countries) if countries
-                    else node['country'] if "country" in node
-                    else "??",
+                    else node.get('country', '') or "??",
                 "format":  media_formats_from_node(node['media']),
                 "label":  ", ".join([' '.join(x.split(' ')[:2]) for x in set(labels)]),
                 "catnum": ", ".join(set(catnums)),
                 "tracks":  "+".join([str(m['track-count']) for m in node['media']]),
-                "barcode":
-                    node['barcode']
-                    if "barcode" in node
-                    and node['barcode'] != ""
-                    else _("[no barcode]"),
-                "packaging":
-                    node['packaging']
-                    if "packaging" in node
-                    else None,
-                "disambiguation":
-                    node['disambiguation']
-                    if "disambiguation" in node
-                    else None,
+                "barcode": node.get('barcode', '') or _('[no barcode]'),
+                "packaging": node.get('packaging', '') or '??',
+                "disambiguation": node.get('disambiguation', ''),
                 "_disambiguate_name": list(),
                 "totaltracks": sum([m['track-count'] for m in node['media']]),
                 "countries": countries,


picard/ui/options/plugins.py
```diff
@@ -269,6 +269,22 @@ def download_button_process():
             else:
                 label = _("Installed")
             self.ui.plugins.setItemWidget(item, 2, QtWidgets.QLabel(label))
+            button = QtWidgets.QPushButton(_("Uninstall"))
+            button.setMaximumHeight(button.fontMetrics().boundingRect(label).height() + 7)
```
samj1912: Why this arbitrary +7? Do not make UI changes in non ui_*.py files.

picard/ui/options/plugins.py
```diff
@@ -269,6 +269,22 @@ def download_button_process():
...
+            self.ui.plugins.setItemWidget(item, 3, button)
+
+            def plugin_remove():
+                self.ui.plugins.setCurrentItem(item)
+                plugin = self.items[self.ui.plugins.selectedItems()[0]]
+                self.tagger.pluginmanager.remove_plugin(plugin.module_name + '.zip')
```
samj1912: It isn't always a zip. It could also be a single py file or a folder.

picard/ui/options/plugins.py
```diff
@@ -269,6 +269,22 @@ def download_button_process():
             else:
                 label = _("Installed")
             self.ui.plugins.setItemWidget(item, 2, QtWidgets.QLabel(label))
+            button = QtWidgets.QPushButton(_("Uninstall"))
+            button.setMaximumHeight(button.fontMetrics().boundingRect(label).height() + 7)
+            self.ui.plugins.setItemWidget(item, 3, button)
+
+            def plugin_remove():
```
samj1912: Do not introduce function definitions inside other functions.

picard/ui/options/plugins.py
```diff
@@ -269,6 +269,22 @@ def download_button_process():
...
+                plugin = self.items[self.ui.plugins.selectedItems()[0]]
+                self.tagger.pluginmanager.remove_plugin(plugin.module_name + '.zip')
+                msgbox = QtWidgets.QMessageBox(self)
+                msgbox.setText(
+                    _("The plugin '%s' will be remove on next run of Picard.")
+                    % (plugin.name))
+                msgbox.setStandardButtons(QtWidgets.QMessageBox.Ok)
+                msgbox.setDefaultButton(QtWidgets.QMessageBox.Ok)
+                msgbox.exec_()
+            button.released.connect(plugin_remove)
```
samj1912: You might want to ask for user confirmation here.

picard/ui/options/plugins.py
```diff
@@ -269,6 +269,22 @@ def download_button_process():
...
+            button.setMaximumHeight(button.fontMetrics().boundingRect(label).height() + 7)
+            self.ui.plugins.setItemWidget(item, 3, button)
+
+            def plugin_remove():
+                self.ui.plugins.setCurrentItem(item)
+                plugin = self.items[self.ui.plugins.selectedItems()[0]]
+                self.tagger.pluginmanager.remove_plugin(plugin.module_name + '.zip')
+                msgbox = QtWidgets.QMessageBox(self)
+                msgbox.setText(
+                    _("The plugin '%s' will be remove on next run of Picard.")
```
samj1912: s/remove/removed

samj1912:
Please fix the PR template and link to the Jira ticket.

samj1912:
image
This is how my picard looks. Why do we need an extra Remove column? you could simply modify the install button to become the uninstall buttons once its installed.

picard/ui/options/plugins.py
samj1912: Please address this.

picard/ui/options/plugins.py
```diff
@@ -230,6 +230,25 @@ def plugin_updated(self, plugin_name):
                 self.change_details()
                 break
 
+    def plugin_remove(self):
+            plugin = self.items[self.ui.plugins.selectedItems()[0]]
+            buttonReply = QtWidgets.QMessageBox.question(self,'Uninstall?', "Do you want to uninstall?",
+                    QtWidgets.QMessageBox.Yes | QtWidgets.QMessageBox.No, QtWidgets.QMessageBox.No)
+            if buttonReply == QtWidgets.QMessageBox.Yes:
+                if os.path.isfile(USER_PLUGIN_DIR+plugin.module_name+'.zip'):
```
samj1912: I cannot fathom why you need to add extensions to the plugin names. It will only cause erroneous results with remove_plugin. Remove_plugin just expects the plugin name by default. Nothing else. It handles stripping of extensions and deletion of files.

See https://github.com/metabrainz/picard/blob/master/picard/plugin.py#L351

chhavijustme: I think we can do something like this.
screen shot 2018-01-21 at 7 17 28 pm
Instead of course, following=> uninstall and follow=>install. The important thing here is to keep it looking like a button so that people know it can be clicked.

vishichoudhary:
There is already a second operation UPDATE on it so i can't add REMOVE on the same button. :-(

picard/ui/options/plugins.py
chhavijustme: @vishichoudhary 
1. I don't know the case here, but if update is available for only a few plugins at a time,
You can add an "update" link in front of all plugins whose update is available.

2. If an update is available for all the plugins all the time, then having install/uninstall button and an update button makes more sense.

vishichoudhary:
@samj1912 , i am extracting the module name only, then i am checking that whether this moudule name exists as .zip or .py or as a folder?

samj1912:
@vishichoudhary yes and you don't need to do any of that. Simply the last line will suffice.
`self.tagger.pluginmanager.remove_plugin(plugin.module_name)`

You don't need to do the extension checks.

vishichoudhary:
ok @samj1912  i will make changes accordingly. :-)

vishichoudhary:
@chhavijustme  i didn't get you on your last point. Can you please explain it bit more? :-)

picard/ui/options/plugins.py
vishichoudhary: @chhavijustme  what about having only one button for install / uninstall and update automatically?

PR #806, opened 2018-01-21
PICARD-1179 Fix release builder logic

samj1912:
When releases are built from node, fields are assigned values None at
times which doesn't play well with the string join in a couple of lines
below. This fixes that and reduces code by using default values.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1179](https://tickets.metabrainz.org/browse/PICARD-1179)

### DIFF
--- a/picard/releasegroup.py
+++ b/picard/releasegroup.py
@@ -77,25 +77,14 @@ class ReleaseGroup(DataObject):
                 "id":      node['id'],
                 "year":    node['date'][:4] if "date" in node else "????",
                 "country": "+".join(countries) if countries
-                    else node['country'] if "country" in node
-                    else "??",
+                    else node.get('country', '') or "??",
                 "format":  media_formats_from_node(node['media']),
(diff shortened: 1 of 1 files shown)


picard/releasegroup.py
```diff
@@ -77,25 +77,14 @@ def _parse_versions(self, document):
                 "id":      node['id'],
                 "year":    node['date'][:4] if "date" in node else "????",
                 "country": "+".join(countries) if countries
-                    else node['country'] if "country" in node
-                    else "??",
+                    else node.get('country', '') or "??",
```
mineo: Why not just `node.get('country', '??')` (similarly for barcode and packaging)?

samj1912: There are times when node has a value for country, but it is set to None. It is to handle cases like that.

samj1912: The alternative would be to use something like node.get('country', '??') or '??'

PR #810, opened 2018-01-24
Search dialog cleanup

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Rework search dialog, many parts weren't working as expected

# Problem

- incorrect indentation
- missing whitelines
- Python2-style super()
- hardcoded column positions
- cover art fetching tasks continue after the dialog is closed
- cover art is fetched at once with no limit on number of releases
- sorting by columns is disabled

# Solution

- PEP8 fixes
- Python3-style super()
- use named columns
- store fetch tasks and remove them on dialog close
- load cover art only for visible cells
- basic sorting by column is enabled, search score added

# Action

Review


picard/ui/searchdialog.py
```diff
@@ -58,15 +58,19 @@ def __init__(self, parent, column_titles):
             QtWidgets.QHeaderView.Stretch)
         self.horizontalHeader().setSectionResizeMode(
             QtWidgets.QHeaderView.Interactive)
+        @throttle(1000)
+        def emit_scrolled(x):
```
samj1912: Can you add a comment on why this is throttled?

picard/ui/searchdialog.py
```diff
@@ -594,27 +666,38 @@ def handle_reply(self, document, http, error):
         self.fetch_coverarts()
```
samj1912: Fixed this by changing it to         
```
QTimer().singleShot(2000, self.fetch_coverarts)
```
The problem is that redering the cells takes time. So they are not visible immediately after the call to handle_reply.
Although this is a bit hacky, a better way would be to connec to the cell's show event. But I guess you get what the problem is.

zas: It was my guess, i'll try to find a good way to do it. Thanks ;)

zas: Fixed in 766c31a

picard/ui/searchdialog.py
```diff
@@ -705,21 +709,21 @@ def parse_releases(self, releases):
...
-            self.table.setItem(row, self.colpos('tracks'), table_item(release.get("tracks", "")))
-            self.table.setItem(row, self.colpos('date'), table_item(release.get("date", "")))
-            self.table.setItem(row, self.colpos('country'), table_item(release.get("country", "")))
-            self.table.setItem(row, self.colpos('labels'), table_item(release.get("label", "")))
-            self.table.setItem(row, self.colpos('catnums'), table_item(release.get("catalognumber", "")))
-            self.table.setItem(row, self.colpos('barcode'), table_item(release.get("barcode", "")))
-            self.table.setItem(row, self.colpos('language'), table_item(release.get("~releaselanguage", "")))
-            self.table.setItem(row, self.colpos('type'), table_item(release.get("releasetype", "")))
-            self.table.setItem(row, self.colpos('status'), table_item(release.get("releasestatus", "")))
+            self.set_table_item(row, 'name',     release, "album")
```
samj1912: This violates PEP8 https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

picard/ui/searchdialog.py
```diff
@@ -463,13 +463,13 @@ def display_results(self):
...
             table_item = QtWidgets.QTableWidgetItem
             self.table.insertRow(row)
-            self.table.setItem(row, self.colpos('name'), table_item(track.get("title", "")))
-            self.table.setItem(row, self.colpos('length'), table_item(track.get("~length", "")))
-            self.table.setItem(row, self.colpos('artist'), table_item(track.get("artist", "")))
-            self.table.setItem(row, self.colpos('release'), table_item(track.get("album", "")))
-            self.table.setItem(row, self.colpos('date'), table_item(track.get("date", "")))
-            self.table.setItem(row, self.colpos('country'), table_item(track.get("country", "")))
-            self.table.setItem(row, self.colpos('type'), table_item(track.get("releasetype", "")))
+            self.set_table_item(row, 'name',    track, "title")
```
samj1912: Ditto here about PEP8

picard/ui/searchdialog.py
```diff
@@ -146,6 +147,8 @@ def trigger_search_action(self):
 
 class CoverWidget(QtWidgets.QWidget):
 
+    shown = pyqtSignal()
```
samj1912: Correct me if I am wrong but doesn't this make the scrolled signal based fetching of CAA redundant?

picard/ui/searchdialog.py
zas: > This violates PEP8 https://www.python.org/dev/peps/pep-0008/#whitespace-in-expressions-and-statements

Those spaces are meant to make things easier to read (and it does, really).

And it doesn't violate PEP8, if one reads it carefully:

> Some other good reasons to ignore a particular guideline:
> 
> When applying the guideline would make the code less readable, even for someone who is used to reading code that follows this PEP.

samj1912: Hmm, I found it weird to read, since I am not used to aligned `=` or `,` in python. If we are making it an exception to PEP8 we should add it to our contribution guide.

picard/ui/searchdialog.py
zas: > Ditto here about PEP8

Ditto
https://www.python.org/dev/peps/pep-0008/#a-foolish-consistency-is-the-hobgoblin-of-little-minds ;)

picard/ui/searchdialog.py
zas: That's pretty usual to "columnize" such repeated statements (using spaces, not tabs) in all languages.
We don't want to make a rule, as it is very dependent on context.

One has yet to prove that:

```python
self.set_table_item(row, 'name', release, "album")
self.set_table_item(row, 'artist', release, "albumartist")
self.set_table_item(row, 'format', release, "format")
self.set_table_item(row, 'tracks', release, "tracks")
self.set_table_item(row, 'date', release, "date")
self.set_table_item(row, 'country', release, "country")
self.set_table_item(row, 'labels', release, "label")
self.set_table_item(row, 'catnums', release, "catalognumber")
self.set_table_item(row, 'barcode', release, "barcode")
self.set_table_item(row, 'language', release, "~releaselanguage")
self.set_table_item(row, 'type', release, "releasetype")
self.set_table_item(row, 'status', release, "releasestatus")
self.set_table_item(row, 'score', release, "score", conv=int)
```
is easier to read than:

```python
self.set_table_item(row, 'name',     release, "album")
self.set_table_item(row, 'artist',   release, "albumartist")
self.set_table_item(row, 'format',   release, "format")
self.set_table_item(row, 'tracks',   release, "tracks")
self.set_table_item(row, 'date',     release, "date")
self.set_table_item(row, 'country',  release, "country")
self.set_table_item(row, 'labels',   release, "label")
self.set_table_item(row, 'catnums',  release, "catalognumber")
self.set_table_item(row, 'barcode',  release, "barcode")
self.set_table_item(row, 'language', release, "~releaselanguage")
self.set_table_item(row, 'type',     release, "releasetype")
self.set_table_item(row, 'status',   release, "releasestatus")
self.set_table_item(row, 'score',    release, "score", conv=int)
```

But you have a point, this is an intentional exception to the rule you cited, and you know i'm used to enforce PEP8.

picard/ui/searchdialog.py
zas: > Correct me if I am wrong but doesn't this make the scrolled signal based fetching of CAA redundant?

That's a good point, and you may be right, i need to verify this.

zas: According to my experiments, it is required, i'll see if we can improve that later, but for now, i'd let this as is.

PR #814, opened 2018-01-26
Move throttling code to its own file

zas:
API is modified to use methods to get/set delays, instead of setting global variables

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -41,6 +41,7 @@ from picard import (PICARD_APP_NAME,
 from picard.oauth import OAuthManager
 from picard.util import build_qurl, parse_json
 from picard.util.xml import parse_xml
+from picard.webservice import ratecontrol
 
 COUNT_REQUESTS_DELAY_MS = 250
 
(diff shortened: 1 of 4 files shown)


picard/webservice/throttle.py
```diff
@@ -118,13 +118,13 @@ def incr_requests(hostkey):
     _remember_request_time(hostkey)
     # Increment the number of unack'd requests on sending a new one
     CONGESTION_UNACK[hostkey] += 1
-    log.debug("throttle: %s: outstanding reqs: %d",
+    log.debug("throttle: %s: incr reqs: %d",
```
samj1912: Maybe we can be more verbose in logging? `Incrementing requests to: %d`?

picard/webservice/throttle.py
```diff
@@ -0,0 +1,205 @@
...
+              hostkey, elapsed, delay)
+    return (True, delay)
+
+
+def _remember_request_time(hostkey):
+    if REQUEST_DELAY[hostkey]:
+        LAST_REQUEST_TIMES[hostkey] = time.time()
+
+
+def incr_requests(hostkey):
```
samj1912: Do not like the function name that much. Maybe `increment_unacked_requests`? Or is that too long?

picard/webservice/throttle.py
```diff
@@ -128,12 +128,15 @@ def decr_requests(hostkey):
...
-        log.debug("throttle: Setting the minimum rate limit for %s to %i" %
-                  (redirect_host_key, REQUEST_DELAY_MINIMUM[original_host_key]))
-        REQUEST_DELAY_MINIMUM[redirect_host_key] = REQUEST_DELAY_MINIMUM[original_host_key]
+def copy_minimal_delay(from_hostkey, to_hostkey):
+    """Copy minimal delay from one hostkey to another
+        Useful for redirections
+    """
+    if (from_hostkey in REQUEST_DELAY_MINIMUM
+            and to_hostkey not in REQUEST_DELAY_MINIMUM):
+        log.debug("throttle: %s: Setting the minimum rate limit to %i" %
```
samj1912: We can avoid % when logging and simply pass the values as parameters.

picard/webservice/throttle.py
```diff
@@ -159,12 +159,9 @@ def adjust(hostkey, slow_down):
                     slow_down: `True` if we encountered intermittent server trouble
                     and need to slow down.
     """
-    def in_backoff_phase(hostkey):
-        return CONGESTION_UNACK[hostkey] > CONGESTION_WINDOW_SIZE[hostkey]
-
     if slow_down:
         _slow_down(hostkey)
-    elif not in_backoff_phase(hostkey):
+    elif CONGESTION_UNACK[hostkey] <= CONGESTION_WINDOW_SIZE[hostkey]:
```
samj1912: Although it was trivial, it was more readable, maybe add a comment here?

picard/webservice/throttle.py
zas: > Do not like the function name that much. Maybe increment_unacked_requests? Or is that too long?

I don't like it much either, and yes,  `increment_unacked_requests` is a bit long.
Why not `increment_requests` and `decrement_requests` ? mid-sized & explicit.

picard/webservice/throttle.py
zas: > We can avoid % when logging and simply pass the values as parameters.

You're totally right, i'll fix that.

picard/webservice/throttle.py
zas: > Although it was trivial, it was more readable, maybe add a comment here?

Yes, i agree. It was more readable, but a function def was a bit overkilling here imho. I'll add a comment.

samj1912:
I think it's a good idea to rename throttle to ws_throttle.

PR #816, opened 2018-01-28
Optimize imports and follow PEP-8

samj1912:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Import clean up for picard

# Solution

# Action


picard/coverart/__init__.py
```diff
@@ -182,22 +183,22 @@ def next_in_queue(self):
...
         )
         log.debug("Downloading %r" % coverartimage)
         self.album.tagger.webservice.download(
-            coverartimage.host,
-            coverartimage.port,
-            coverartimage.path,
-            partial(self._coverart_downloaded, coverartimage),
-            priority=True,
-            important=False
+                coverartimage.host,
```
zas: Why additional spaces here ?

picard/const/attributes.py
```diff
@@ -3,93 +3,93 @@
 # Use `python setup.py update_constants` to update it.
```
zas: ^^

picard/coverart/image.py
```diff
@@ -27,16 +27,17 @@
 import tempfile
 from hashlib import md5
 
-from PyQt5.QtCore import QUrl, QObject, QMutex
+from PyQt5.QtCore import (QMutex,
```
zas: Just a thing, i prefer following syntax:
```python
from PyQt5.QtCore import (
    QMutex,
    QObject,
    QUrl,
)
```

Reason: if you change the list, ie. sort it, or insert/delete lines, diffs are much easier to read. Sorting is easier too (select all, sort). The extra comma is legit in Python, and convenient when you re-order.
Also it requires only 4 spaces to indent.

picard/coverart/image.py
```diff
@@ -282,7 +282,7 @@ def save(self, dirname, metadata, counters):
         if not self.can_be_saved_to_disk:
             return
         if (config.setting["caa_image_type_as_filename"] and
-            not self.is_front_image()):
+                not self.is_front_image()):
```
zas: In this case, i prefer to have the `and` at start of the line:

```python
if (config.setting["caa_image_type_as_filename"]
    and not self.is_front_image()):
```

Reason: having all logical operators aligned vertically is much easier to read.

```python
if (expr1
    or not expr2
    and (expr3
         or expr4)
    and expr4):
```

Also it is consistent with http://legacy.python.org/dev/peps/pep-0008/#should-a-line-break-before-or-after-a-binary-operator

picard/coverart/image.py
```diff
@@ -355,7 +355,6 @@ def types_as_string(self, translate=True, separator=', '):
 
 
 class CaaCoverArtImage(CoverArtImage):
-
```
zas: Not sure about this removal.

picard/acoustid/__init__.py
```diff
@@ -102,24 +105,24 @@ def _lookup_fingerprint(self, next_func, filename, result=None, error=None):
...
+                    N_("AcoustID lookup returned no result for file '%(filename)s'"),
+                    mparms,
+                    echo=None
             )
             file.clear_pending()
             return
         log.debug(
-            "AcoustID: looking up the fingerprint for file '%(filename)s'" %
-            mparms
+                "AcoustID: looking up the fingerprint for file '%(filename)s'" %
```
zas: Actually i think it is a bad mix.
Either we align on first char after the `(` or we wrap all and indent with 4 spaces.
I prefer the second approach, because it allows longer lines (less indentation spaces), and allow to keep args aligned (one per line), which is much more readable.
It limits impact of changes (modification of an argument name, adding/remove args, etc...) which makes diffs more easier to read (diffs are line-oriented mostly).

My preferred style, especially with many args:
```python
somefunc(
    "the string %s" %
    arg
)
```

Or, in certain cases (only 2 lines):
```python
somefunc("the string %s" %
         arg)
```

picard/acoustid/manager.py
```diff
@@ -45,7 +47,9 @@ def add(self, file, recordingid):
         if not hasattr(file, 'acoustid_length'):
             return
         puid = file.metadata['musicip_puid']
-        self._fingerprints[file] = Submission(file.acoustid_fingerprint, file.acoustid_length, recordingid, recordingid, puid)
+        self._fingerprints[file] = Submission(file.acoustid_fingerprint, file.acoustid_length, recordingid,
```
zas: Here, wrap all arguments, one per line.

picard/acoustid/manager.py
```diff
@@ -76,38 +80,39 @@ def submit(self):
...
         log.debug("AcoustID: submitting ...")
         self.tagger.window.set_statusbar_message(
-            N_('Submitting AcoustIDs ...'),
-            echo=None
+                N_('Submitting AcoustIDs ...'),
+                echo=None
         )
         self.tagger.acoustid_api.submit_acoustid_fingerprints(fingerprints,
-            partial(self.__fingerprint_submission_finished, fingerprints))
+                                                              partial(self.__fingerprint_submission_finished,
```
zas: No, it moves things too far on right, you'll end with wrapping a lot etc...
I know we don't enforce line width, but still, better have shorter lines.

Extreme example:

```python
somefunc(someverylongshitkjsdljfksldkfjskdjflkjdskfjlsdkjfkjsdkfksdjkfjsjjfsdk(ddddddddd,
                                                                               xxxxxxxxxx,
                                                                               yyyyyyyyyy),
                                                                               func(you_are_doomed,
                                                                                       messy))
```

```python
somefunc(
    someverylongshitkjsdljfksldkfjskdjflkjdskfjlsdkjfkjsdkfksdjkfjsjjfsdk(
        ddddddddd,
        xxxxxxxxxx,
        yyyyyyyyyy
    ),
    func(
        you_are_doomed,
        messy
    )
)

File: picard/acoustid/manager.py [hunk 9]
```diff
@@ -76,38 +80,39 @@ def submit(self):
...
+                                                                      fingerprints))
 
     def __fingerprint_submission_finished(self, fingerprints, document, http, error):
         if error:
             try:
                 error = load_json(document)
                 message = error["error"]["message"]
-            except :
+            except:
```
zas: Yeeessss :)

picard/album.py
```diff
@@ -20,30 +20,33 @@
...
+
+from picard import (config,
+                    log)
+from picard.cluster import Cluster
+from picard.collection import add_release_to_user_collections
+from picard.const import VARIOUS_ARTISTS_ID
 from picard.coverart import coverart
+from picard.dataobj import DataObject
+from picard.file import File
+from picard.mbjson import (medium_to_metadata,
```
zas: Better wrap and indent with 4 spaces.

picard/album.py
```diff
@@ -54,7 +57,6 @@ def __init__(self, album_artist_id):
 
 
 class Album(DataObject, Item):
-
```
zas: Now i think it is wrong, because of https://www.python.org/dev/peps/pep-0257/#multi-line-docstrings

> Insert a blank line after all docstrings (one-line or multi-line) that document a class -- generally speaking, the class's methods are separated from each other by a single blank line, and the docstring needs to be offset from the first method by a blank line.


```python
class TheClass(object):

   def method(self):
      ...
```

```python
class TheClass(object):
   """docstring"""

   def method(self):
      ...
```

```python
class TheClass(object):
   """docstring"""

   somevar = 123

   def method(self):
      ...
```

```python
class TheClass(object):

   somevar = 123

   def method(self):
      ...
```

picard/cluster.py
```diff
@@ -63,7 +67,7 @@ def __repr__(self):
             return '<Cluster %s %r>' % (
                 self.related_album.id,
                 self.related_album.metadata[u"album"] + '/' + self.metadata['album']
-                )
+            )
```
zas: Yes.

picard/cluster.py
```diff
@@ -213,14 +217,14 @@ def lookup_metadata(self):
...
         )
         self.lookup_task = self.tagger.mb_api.find_releases(self._lookup_finished,
-            artist=self.metadata['albumartist'],
-            release=self.metadata['album'],
-            tracks=string_(len(self.files)),
-            limit=QUERY_LIMIT)
+                                                            artist=self.metadata['albumartist'],
+                                                            release=self.metadata['album'],
+                                                            tracks=string_(len(self.files)),
+                                                            limit=QUERY_LIMIT)
```
zas: Nope, instead of inserting tons of spaces, wrap first arg and indent with 4 spaces.
I hate when everything is pushed on the right for no reason.

zas:
@samj1912 : we can re-introduce this in smaller bits, the import clean up was good

PR #817, opened 2018-01-28
Use pyinstaller to package builds

samj1912:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
Uses pyinsatller to build executables for various systems.
# Problem

Our current setup uses py2exe and py2app both of which do not produce working executables with py3 and pyqt5. Thus the shift to pyinstaller was needed. It also simplifies the build process and allows us to make Picard a portable app in windows.

# Solution

Use pyinstaller and appveyor to build packages


picard.spec
```diff
@@ -0,0 +1,69 @@
...
+            locale = os.path.splitext(filename)[0]
+            locales.append((domain, locale, filepath))
+    return locales
+
+
+def get_local_messages():
+    data_files = []
+    for locale in _picard_get_locale_files():
+        data_files.append(
+            ("build/locale/" + locale[1] + "/LC_MESSAGES/" + locale[0] + ".mo",
```
zas: Why no `os.path.join()` here ?

picard.spec
```diff
@@ -0,0 +1,57 @@
...
+             win_private_assemblies=False,
+             cipher=block_cipher)
+pyz = PYZ(a.pure, a.zipped_data,
+             cipher=block_cipher)
+exe = EXE(pyz,
+          a.scripts,
+          a.binaries,
+          a.zipfiles,
+          a.datas,
+          name='picard',
```
zas: Is the version set in `picard/__init__.py` used somewhere ?

picard.spec
```diff
@@ -0,0 +1,57 @@
...
+          debug=False,
+          strip=False,
+          upx=True,
+          runtime_tmpdir=None,
+          console=False,
+          icon=os.path.join('.', 'picard.ico'),
```
zas: Is this really needed since the file is searched in the same dir ? `icon='picard.ico'` should be enough, no ?

tagger.py
```diff
@@ -4,7 +4,13 @@
...
+
 from picard.tagger import main
 
-localedir = os.path.join(os.path.dirname(sys.argv[0]), 'locale')
-main(localedir, True)
+# This is needed to find resources when using pyinstaller
+if getattr(sys, 'frozen', False):
+    basedir = sys._MEIPASS
+else:
+    basedir = os.path.dirname(os.path.abspath(__file__))
```
zas: Not sure if the change from `os.path.dirname(sys.argv[0])` to `os.path.dirname(os.path.abspath(__file__))` is strictly needed to support pyinstaller.
I don't know it is better or worse, but this branch isn't used by pyinstaller, right ?
 Is there a specific reason ?

samj1912: It's not needed to support pyinstaller, but not sure how sys.argv[0] is a better path for basedir?
sys.argv[0] in this case will be the path to tagger.py if we run it via python tagger.py , which is the same as what the above does.

picard.spec
samj1912: We will have to set it up using python setup.py version

appveyor.yml
```diff
@@ -10,28 +15,36 @@ environment:
...
+    python_version: 3.5.3
+    python_arch: 64
+  - PYTHON: C:\Python36-x64
+    PYTHON_VERSION: 3.6.4
+    PYTHON_ARCH: 64
+build_script:
 - cmd: >-
     appveyor DownloadFile https://github.com/acoustid/chromaprint/releases/download/v%CHROMAPRINT_FPCALC_VERSION%/chromaprint-fpcalc-%CHROMAPRINT_FPCALC_VERSION%-windows-i686.zip -FileName fpcalc.zip
 
+    appveyor DownloadFile https://indy.fulgan.com/SSL/openssl-%OPENSSL_VERSION%-x64_86-win64.zip -FileName openssl.zip
```
mineo: Where is OPENSSL_VERSION defined? I can't find it on https://www.appveyor.com/docs/environment-variables/ and I think it's also not visible in the build log (https://ci.appveyor.com/project/MetaBrainz/picard/build/2.0.0.dev.2.49/job/8w22578qj5bf7wgh).

mineo: Also, do we need to change anything to fulfill the [OpenSSL license](https://www.openssl.org/source/license.html), in particular paragraph 2:

>  * 2. Redistributions in binary form must reproduce the above copyright
>  *    notice, this list of conditions and the following disclaimer in
>  *    the documentation and/or other materials provided with the
>  *    distribution.

appveyor.yml
```diff
@@ -10,28 +15,36 @@ environment:
...
-    PYTHON_ARCH: 32
-  - PYTHON: C:\Python35
-    PYTHON_VERSION: 3.5.3
-    PYTHON_ARCH: 32
-build: off
-test_script:
+  - PYTHON: C:\Python35-x64
+    python_version: 3.5.3
+    python_arch: 64
+  - PYTHON: C:\Python36-x64
```
mineo: Does this mean we're now only providing x86_64 builds? AFAICT 1.4.2 was still compatible with 32bit Windows versions. I don't have strong feelings about this, but I'm afraid this will not show up in any changelog when 2.0 is released if there's no jira ticket or something.

appveyor.yml
```diff
@@ -10,28 +15,36 @@ environment:
...
+    PYTHON_ARCH: 64
+build_script:
 - cmd: >-
     appveyor DownloadFile https://github.com/acoustid/chromaprint/releases/download/v%CHROMAPRINT_FPCALC_VERSION%/chromaprint-fpcalc-%CHROMAPRINT_FPCALC_VERSION%-windows-i686.zip -FileName fpcalc.zip
 
+    appveyor DownloadFile https://indy.fulgan.com/SSL/openssl-%OPENSSL_VERSION%-x64_86-win64.zip -FileName openssl.zip
+
     7z x fpcalc.zip -y
 
     copy /Y chromaprint-fpcalc-%CHROMAPRINT_FPCALC_VERSION%-windows-i686\fpcalc.exe %PYTHON%
```
mineo: Does this provide a working fpcalc.exe anywhere that can be picked up by Picard? At least in Wine I wasn't able to configure that.

appveyor.yml
samj1912: Its configurable in AppVeyor settings along with the other env variables.

mineo: Could we move this to the environment section in this file or echo the version at some point? Otherwise it's nearly impossible to figure out which openssl version is downloaded for normal people. At the moment it appears to be set to 1.0.2, but 1.0.2n is already available.

appveyor.yml
samj1912: #824 was made to fix that

PR #819, opened 2018-01-29
Use Fusion style for Picard

samj1912:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem
Qt5 abandoned the old Plastique and motif styles and added 'Fusion' a cross platform style that looks native and absolutely amazing on each of the OSes. 
Here's the difference in Windows with and without - http://imgur.com/a/UgMt8
See http://blog.qt.io/blog/2012/10/30/cleaning-up-styles-in-qt5-and-adding-fusion/

# Solution

Making Picard default to this.


zas:
I see no change on my platform with or without this.
Can you provide screenshots showing difference on OSes it changes something ?

PR #822, opened 2018-01-29
Fix sorting in SearchDialog

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Sorting in SearchDialog wasn't enabled at start, enabling it revealed various issues:
- returned row was incorrect after a re-ordering of rows
- sort order of certain columns was incorrect (Length column of Track search for example), because the default sort key is a string.

# Solution

Subclass QTableWidgetItem and optionnaly set a per-column sort method ('duration' or 'number')
Store row in the data array as UserRole on first column and retrieve it on accept.

# Action

Test & review

### DIFF
--- a/picard/ui/searchdialog/__init__.py
+++ b/picard/ui/searchdialog/__init__.py
@@ -336,7 +336,8 @@ class SearchDialog(PicardDialog):
 
     def accept(self):
         if self.table:
-            row = self.table.selectionModel().selectedRows()[0].row()
+            idx = self.table.selectionModel().selectedRows()[0]
+            row = self.table.itemFromIndex(idx).data(QtCore.Qt.UserRole)
             self.accept_event(row)
         self.save_state()
         QtWidgets.QDialog.accept(self)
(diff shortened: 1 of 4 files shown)


picard/ui/searchdialog/__init__.py
```diff
@@ -133,6 +133,28 @@ def trigger_search_action(self):
...
+BY_NUMBER, BY_DURATION = range(2)
+
+
+class SortableTableWidgetItem(QtWidgets.QTableWidgetItem):
+
+    def __init__(self, sort_key):
+        super().__init__()
+        self.sort_key = sort_key
+
+    def __lt__(self, other):
```
samj1912: Is this everything it needs to generate a sort order? I thought you needed atleast `__eq__` and functools.total_ordering

zas: Yes, it is everything needed.
See https://docs.python.org/3/library/stdtypes.html#list.sort for example:
"This method sorts the list in place, using only < comparisons between items."

PR #824, opened 2018-01-31
Bundle fpcalc with distributables

samj1912:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

We were not bundling fpcalc properly while packaging. This PR fixes it.

# Action

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -35,7 +35,7 @@ os_name = platform.system()
 fpcalc_name = 'fpcalc'
 
 if os_name == 'Windows':
-    fpcalc_name += '.exe'
+    fpcalc_name = 'fpcalc.exe'
 
 if os.path.isfile(fpcalc_name):
     data_files += [(fpcalc_name, fpcalc_name)]
--- a/picard.spec
+++ b/picard.spec
@@ -2,6 +2,7 @@
 
 import os
 import glob
+import platform
 
 
 def _picard_get_locale_files():
@@ -28,9 +29,17 @@ def get_locale_messages():
     return data_files
 
 
-
 block_cipher = None
 data_files = get_locale_messages()
+os_name = platform.system()
+fpcalc_name = 'fpcalc'
+
+if os_name == 'Windows':
+    fpcalc_name += '.exe'
+
+if os.path.isfile(fpcalc_name):
+    data_files += [(fpcalc_name, fpcalc_name)]
+
 
 a = Analysis(['tagger.py'],
              pathex=['picard'],
--- a/picard.spec
+++ b/picard.spec
@@ -30,20 +30,22 @@ def get_locale_messages():
 
 
 block_cipher = None
-data_files = get_locale_messages()
 os_name = platform.system()
-fpcalc_name = 'fpcalc'
+binaries = []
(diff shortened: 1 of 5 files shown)


picard.spec
```diff
@@ -28,9 +29,17 @@ def get_locale_messages():
...
 block_cipher = None
 data_files = get_locale_messages()
+os_name = platform.system()
+fpcalc_name = 'fpcalc'
+
+if os_name == 'Windows':
+    fpcalc_name += '.exe'
+
+if os.path.isfile(fpcalc_name):
+    data_files += [(fpcalc_name, fpcalc_name)]
```
zas: Not sure it such program should go in datas or binaries: https://pythonhosted.org/PyInstaller/spec-files.html#spec-file-operation
How does it handle file perms ? (on systems like macosx)

zas: Also second argument of the tuple is supposed to be a folder:
https://pyinstaller.readthedocs.io/en/stable/spec-files.html#adding-data-files
"The second specifies the name of the **folder** to contain the files at run-time."

picard.spec
```diff
@@ -28,9 +29,17 @@ def get_locale_messages():
     return data_files
 
 
-
 block_cipher = None
 data_files = get_locale_messages()
+os_name = platform.system()
```
zas: Just to be picky i would just write things this way:

```python
if platform.system() == 'Windows':
  fpcalc_name = 'fpcalc.exe'
else:
  fpcalc_name = 'fpcalc'
```

I don't see the point of using string concatenation to "build" the file name, `os_name` would be useful if used more than once, which isn't the case here.

picard/util/__init__.py
```diff
@@ -197,6 +197,9 @@ def find_executable(*executables):
         executables = [e + '.exe' for e in executables]
     paths = [os.path.dirname(sys.executable)] if sys.executable else []
     paths += os.environ.get('PATH', '').split(os.pathsep)
+    # This is for searching for executables bundled in packaged builds
+    if getattr(sys, 'frozen', False):
+        paths += sys._MEIPASS
```
zas: What's `sys._MEIPASS` ? Perhaps add a link to doc or smt.

picard/acoustid/__init__.py
```diff
@@ -34,7 +35,10 @@ def __init__(self):
         self._running = 0
         self._max_processes = 2
 
-        if not config.setting["acoustid_fpcalc"]:
+        # The second condition is checked because in case of a packaged build of picard
+        # the temp directory that pyinstaller decompresses picard into changes on every
+        # launch, thus we need to ignore the existing config values.
+        if not config.setting["acoustid_fpcalc"] or getattr(sys, 'frozen', False):
```
zas: What about using something more explicit than `getattr(sys, 'frozen', False)` like `is_pyinstaller()`

picard.spec
samj1912: It is used in the OSX pr again, so I thought I'd just include it here.

PR #826, opened 2018-02-01
PICARD-1175: save and restore CD Lookup dialog and headers states

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Save CD Lookup dialog size and header state to persist options, and restore them when needed. 

# Problem

Every time a CD lookup is done you have to scroll or make the resulting window larger to see all the information to distinguish releases, especially the catalog number and barcode (as the last columns). The barcode column often is cut off.
Picard should store and remember that window setting (and maybe auto-size the smaller columns).

# Solution

Save CD Lookup dialog size and header state to persist options, and restore them when needed. 

* fix [PICARD-1175](https://tickets.metabrainz.org/browse/PICARD-1175)

### DIFF
--- a/picard/ui/cdlookup.py
+++ b/picard/ui/cdlookup.py
@@ -18,12 +18,21 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from PyQt5 import QtCore, QtWidgets
+
+from picard import config, log
 from picard.ui import PicardDialog
 from picard.ui.ui_cdlookup import Ui_Dialog
 from picard.mbjson import artist_credit_from_node, label_info_from_node
 
 
 class CDLookupDialog(PicardDialog):
+    dialog_window_size = "cdlookupdialog_window_size"
+    dialog_header_state = "cdlookupdialog_header_state"
+
+    options = [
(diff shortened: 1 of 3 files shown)


picard/ui/cdlookup.py
```diff
@@ -57,13 +67,46 @@ def __init__(self, releases, disc, parent=None):
...
         lookup = self.tagger.get_file_lookup()
         lookup.disc_lookup(self.disc.submission_url)
-        QtWidgets.QDialog.accept(self)
+        self.save_and_accept()
+
+    def reject(self):
+        self.save_state()
+        QtWidgets.QDialog.reject(self)
+
+    def restore_state(self):
```
samj1912: Any reason why save_state saves both size and header state but restore state only restores size? We should maintain consistency in behavior or function name

zas: Because restoring is a two-steps action, the size of dialog is restored just after it is created, but the header stuff is done after the header is actually created. I didn't dig if it is still needed, but all save/restore are made like this. I don't want to bother with that for now, i have ideas on how to make this much simpler (and with less code redundancy).

PR #829, opened 2018-02-01
PICARD-1184: Fix multi click weird issues

vishichoudhary:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Currently in picard if you click multiple time install plugin button it will behave weird. It shows the upgrade dialog and then updated status

* JIRA ticket : [PICARD-1184](https://tickets.metabrainz.org/browse/PICARD-1184)

# Solution
Reason for this problem was in picard/plugin.py there is only single option to handle both install a plugin or update a plugin .
Clicking the install a plugin call this function which emits a singal to picard/ui/options/plugins.py.plugin_installed to update the status from install button to "installed", but if in mean time you click install again this function will call picard/ui/options/plugins.py.plugin_updated method as there exist some downloaded data in plugin folder. I made two function one for handling install actions and another for update action

# Action

### DIFF
--- a/picard/plugin.py
+++ b/picard/plugin.py
@@ -394,16 +394,10 @@ class PluginManager(QtCore.QObject):
                 plugin_name = os.path.splitext(zip_plugin)[0]
         if plugin_name:
             try:
-                dirpath, filepaths = self._get_existing_paths(plugin_name)
-                update = dirpath or filepaths
                 if plugin_data and plugin_name:
                     # zipped module from download
                     zip_plugin = plugin_name + '.zip'
                     dst = os.path.join(USER_PLUGIN_DIR, zip_plugin)
-                    if update:
-                        dst += '.update'
-                        if os.path.isfile(dst):
-                            os.remove(dst)
                     ziptmp = tempfile.NamedTemporaryFile(delete=False,
                                                          dir=USER_PLUGIN_DIR).name
                     try:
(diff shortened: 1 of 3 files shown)


picard/ui/options/plugins.py
```diff
@@ -257,10 +257,13 @@ def add_plugin_item(self, plugin, item=None):
             button.setMaximumHeight(button.fontMetrics().boundingRect(label).height() + 7)
             self.ui.plugins.setItemWidget(item, 2, button)
 
-            def download_button_process():
+            def download_button_process(action):
                 self.ui.plugins.setCurrentItem(item)
-                self.download_plugin()
-            button.released.connect(download_button_process)
+                self.download_plugin(action)
+            if label == _("Update"):
```
zas: Don't do that. Instead mark string for translation, using `N_()`, call `_()` the later possible, and define constants for possible actions. 

```python

PLUGIN_ACTION_NONE, PLUGIN_ACTION_UPDATE, PLUGIN_ACTION_INSTALL = range(3)

...

bt_action = PLUGIN_ACTION_NONE
 if plugin.can_be_updated:
    bt_action = PLUGIN_ACTION_UPDATE
 elif plugin.can_be_downloaded:
    bt_action = PLUGIN_ACTION_INSTALL
    item.setFlags(item.flags() ^ QtCore.Qt.ItemIsUserCheckable)

if bt_action != PLUGIN_ACTION_NONE:
    labels = {
      PLUGIN_ACTION_UPDATE:  N_("Update"),
      PLUGIN_ACTION_INSTALL: N_("Install"),
    }
    button = QtWidgets.QPushButton(_(labels[bt_action]))
  
...
    button.released.connect(partial(download_button_process, bt_action))
...


def download_handler(self, action, response, reply, error, plugin):

    ...

    if action == PLUGIN_ACTION_INSTALL:

        ...

```

Instead of calling functions and comparing strings, you compare integers, and don't call useless functions.

picard/plugin.py
```diff
@@ -421,24 +415,63 @@ def install_plugin(self, path, overwrite_confirm=None, plugin_name=None,
...
+                installed_plugin = self.load_plugin(zip_plugin or plugin_name, USER_PLUGIN_DIR)
+                if installed_plugin is not None:
+                    self.plugin_installed.emit(installed_plugin, False)
+            except (OSError, IOError):
+                log.warning("Unable to copy %s to plugin folder %s" % (path, USER_PLUGIN_DIR))
+
+    def update_plugin(self, path, overwrite_confirm=None, plugin_name=None,
+                       plugin_data=None):
+
+        zip_plugin = False
```
zas: Most of the code is duplicated from above... that's bad.
Instead create a function with all the common code, called by those two.

picard/ui/options/plugins.py
```diff
@@ -341,11 +344,18 @@ def download_handler(self, response, reply, error, plugin):
             log.error("Error occurred while trying to download the plugin: '%s'" % plugin.module_name)
             return
 
-        self.tagger.pluginmanager.install_plugin(
-            None,
-            plugin_name=plugin.module_name,
-            plugin_data=response
-        )
+        if action == "install":
+            self.tagger.pluginmanager.install_plugin(
```
zas: Since those share same arguments, you may do:

```python
if condition:
  func = self.tagger.pluginmanager.install_plugin
else:
  func = self.tagger.pluginmanager.update_plugin

func(None, plugin_name=plugin.module_name, plugin_data=response)
```

vishichoudhary:
Requested changes made. Ready for re review

picard/plugin.py
```diff
@@ -39,6 +39,7 @@
 from picard.const import USER_PLUGIN_DIR, PLUGINS_API
 from picard.util import load_json
 
+PLUGIN_ACTION_UPDATE = 1
```
zas: Since you need this here and in `picard/ui/options/plugins.py`, i would set those in `picard/const.py`:

```python
PLUGIN_ACTION_NONE, PLUGIN_ACTION_UPDATE, PLUGIN_ACTION_INSTALL = range(3)
```

and import those where needed:

```
from picard.const import USER_PLUGIN_DIR, PLUGINS_API, PLUGIN_ACTION_UPDATE
```



Note: we could start to use Python3 Enum for this, but i'm not sure which version supports what etc..

picard/ui/options/plugins.py
```diff
@@ -34,6 +34,7 @@
 from picard.ui.options import OptionsPage, register_options_page
 from picard.ui.ui_options_plugins import Ui_PluginsOptionsPage
 
+PLUGIN_ACTION_NONE, PLUGIN_ACTION_UPDATE, PLUGIN_ACTION_INSTALL = range(3)
```
zas: See my comment about moving this to `picard.const`

vishichoudhary:
@zas done with requested changes

PR #830, opened 2018-02-04
[WIP] Improve log view so one can filter out messages based on verbosity or domains

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Application log view is really user-friendly, but underlying log system neither.
To improve it, few features will be added:
- log "domains", allowing devs to regroup logs in sets
- log verbosity filtering in log view
- log domains filtering in log view
- log highlighting in log view
- saving current log to file

Related to [PICARD-1185](https://tickets.metabrainz.org/browse/PICARD-1185)

# Solution

- log "domains" are implemented, every `log.(error|warn|debug|info)()` now accepts an optional set of domains using `domains` kv parameter
- log messages are stored in a list of `MessageObj`, which has `domains` and `level` properties
- log verbosity can be set through a menu in log view
- log domains can be hidden through a menu in log view
- user-defined text can be highlighted in log view

capture du 2018-02-04 18-19-43

# Action

Review & test, the code is still very experimental, feedback welcome

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -26,6 +26,7 @@ from picard.util import find_executable, is_frozen
 from picard.acoustid.json_helpers import parse_recording
 
 
+
 class AcoustIDClient(QtCore.QObject):
 
     def __init__(self):
@@ -48,6 +49,7 @@ class AcoustIDClient(QtCore.QObject):
     def done(self):
         pass
 
+    @log.domain('acoustid')
     def _on_lookup_finished(self, next_func, file, document, http, error):
 
         doc = {}
@@ -58,7 +60,7 @@ class AcoustIDClient(QtCore.QObject):
             }
             log.error(
                 "AcoustID: Lookup network error for '%(filename)s': %(error)r" %
-                mparms, domains='acoustid')
+                mparms)
             self.tagger.window.set_statusbar_message(
                 N_("AcoustID lookup network error for '%(filename)s'!"),
                 mparms,
@@ -77,7 +79,7 @@ class AcoustIDClient(QtCore.QObject):
                             parsed_recording = parse_recording(recording)
                             if parsed_recording is not None:
                                 recording_list.append(parsed_recording)
-                        log.debug("AcoustID: Lookup successful for '%s'", file.filename, domains='acoustid')
+                        log.debug("AcoustID: Lookup successful for '%s'", file.filename)
             else:
                 mparms = {
                     'error': document['error']['message'],
@@ -85,7 +87,7 @@ class AcoustIDClient(QtCore.QObject):
                 }
                 log.error(
                     "AcoustID: Lookup error for '%(filename)s': %(error)r" %
-                    mparms, domains='acoustid')
+                    mparms)
                 self.tagger.window.set_statusbar_message(
                     N_("AcoustID lookup failed for '%(filename)s'!"),
                     mparms,
@@ -106,7 +108,7 @@ class AcoustIDClient(QtCore.QObject):
         if not result:
             log.debug(
                 "AcoustID: lookup returned no result for file '%(filename)s'" %
-                mparms, domains='acoustid'
+                mparms
             )
             self.tagger.window.set_statusbar_message(
                 N_("AcoustID lookup returned no result for file '%(filename)s'"),
@@ -117,7 +119,7 @@ class AcoustIDClient(QtCore.QObject):
             return
         log.debug(
             "AcoustID: looking up the fingerprint for file '%(filename)s'" %
-            mparms, domains='acoustid'
+            mparms
         )
         self.tagger.window.set_statusbar_message(
             N_("Looking up the fingerprint for file '%(filename)s' ..."),
@@ -167,9 +169,7 @@ class AcoustIDClient(QtCore.QObject):
                     "Fingerprint calculator failed exit code = %r, exit status = %r, error = %s",
                     exit_code,
                     exit_status,
-                    process.errorString(),
-                    domains='acoustid'
-                )
+                    process.errorString())
         finally:
             next_func(result)
 
@@ -182,11 +182,7 @@ class AcoustIDClient(QtCore.QObject):
         try:
             self._running -= 1
             self._run_next_task()
-            log.error("Fingerprint calculator failed error = %s (%r)",
-                      process.errorString(),
-                      error,
-                      domains='acoustid'
-                     )
+            log.error("Fingerprint calculator failed error = %s (%r)", process.errorString(), error)
         finally:
             next_func(None)
 
@@ -202,7 +198,7 @@ class AcoustIDClient(QtCore.QObject):
         process.finished.connect(partial(self._on_fpcalc_finished, next_func, file))
         process.error.connect(partial(self._on_fpcalc_error, next_func, file))
         process.start(fpcalc, ["-length", "120", file.filename])
-        log.debug("Starting fingerprint calculator %r %r", fpcalc, file.filename, domains='acoustid')
+        log.debug("Starting fingerprint calculator %r %r", fpcalc, file.filename)
 
     def analyze(self, file, next_func):
         fpcalc_next = partial(self._lookup_fingerprint, next_func, file.filename)
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -58,7 +58,7 @@ class AcoustIDClient(QtCore.QObject):
             }
             log.error(
                 "AcoustID: Lookup network error for '%(filename)s': %(error)r" %
-                mparms)
+                mparms, domains='acoustid')
             self.tagger.window.set_statusbar_message(
                 N_("AcoustID lookup network error for '%(filename)s'!"),
                 mparms,
@@ -77,7 +77,7 @@ class AcoustIDClient(QtCore.QObject):
                             parsed_recording = parse_recording(recording)
                             if parsed_recording is not None:
                                 recording_list.append(parsed_recording)
-                        log.debug("AcoustID: Lookup successful for '%s'", file.filename)
+                        log.debug("AcoustID: Lookup successful for '%s'", file.filename, domains='acoustid')
             else:
                 mparms = {
                     'error': document['error']['message'],
@@ -85,7 +85,7 @@ class AcoustIDClient(QtCore.QObject):
                 }
                 log.error(
                     "AcoustID: Lookup error for '%(filename)s': %(error)r" %
-                    mparms)
+                    mparms, domains='acoustid')
                 self.tagger.window.set_statusbar_message(
                     N_("AcoustID lookup failed for '%(filename)s'!"),
                     mparms,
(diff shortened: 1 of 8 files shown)


picard/acoustid/__init__.py
```diff
@@ -77,15 +77,15 @@ def _on_lookup_finished(self, next_func, file, document, http, error):
...
+                        log.debug("AcoustID: Lookup successful for '%s'", file.filename, domains='acoustid')
             else:
                 mparms = {
                     'error': document['error']['message'],
                     'filename': file.filename
                 }
                 log.error(
                     "AcoustID: Lookup error for '%(filename)s': %(error)r" %
-                    mparms)
+                    mparms, domains='acoustid')
```
samj1912: Instead of manually specifying the domains, we can use the built in logging module from python stdlib and take advantage of its attributes https://docs.python.org/3/library/logging.html#logrecord-attributes

This will save us all the overhead and the repetitions

samj1912: I'd say we can simply drop the domains part and just keep the UI additions and maybe we can continue this in another PR

picard/ui/logview.py
```diff
@@ -97,31 +125,159 @@ def restoreWindowState(self, position, size):
...
+
+        self.verbosity_menu = QtWidgets.QMenu(self)
+        for level, label in self._log_level_names.items():
+            act = QtWidgets.QAction(_(label), self.verbosity_menu)
+            act.setCheckable(True)
+            act.setChecked(level in self.verbosity)
+            act.triggered.connect(partial(self._verbosity_changed, level))
+            self.verbosity_menu.addAction(act)
+
+        self.verbosity_menu_button = QtWidgets.QPushButton(_("Verbosity"))
```
samj1912: Any reason for having the gui layouts part here instead of a ui file?

picard/acoustid/__init__.py
zas: > Instead of manually specifying the domains, we can use the built in logging module from python stdlib and take advantage of its attributes https://docs.python.org/3/library/logging.html#logrecord-attributes
> 
> This will save us all the overhead and the repetitions

I don't see how LogRecord provides this feature, as is.
Using module is not what i want, since, in the same module, i want to be able to specify the domain of the log line. Filtering on module can be good, but that's not my goal.
In fact, to implement this you need to subclass `logging.Logger`, something like that:

```python

def _onestr2set(domains):
    if not domains:
        return set()
    if isinstance(domains, str):
        # this is a shortcut: allow one to pass a domains as sole
        # argument, but use it as first argument of set
        return set((domains,))
    else:
        return set(domains)


class CustomLogger(logging.Logger):

    def _log(self, level, msg, args, exc_info=None, extra=None):
        if extra is None:
            extra = { 'domains': set() }
        super(CustomLogger, self)._log(level, msg, args, exc_info, extra)


logging.setLoggerClass(CustomLogger)

def debug(message, *args, domains=None):
    domains = _onestr2set(domains)
     root.debug(message, *args, extra={'domains': domains})


def info(message, *args, domains=None):
    domains = _onestr2set(domains)
    root.info(message, *args, extra={'domains': domains})


def warning(message, *args, domains=None):
    domains = _onestr2set(domains)
    root.warning(message, *args, extra={'domains': domains})


def error(message, *args, domains=None):
    domains = _onestr2set(domains)
    root.error(message, *args, extra={'domains': domains})

root = logging.getLogger('picard')
root.setLevel(logging.DEBUG)

handler = logging.StreamHandler()
bf = logging.Formatter('{asctime} {levelname:8s} {message} {domains}', style='{')
handler.setFormatter(bf)
root.addHandler(handler)

```

picard/ui/logview.py
zas: Nope, but since menus are generated, you'll not have much more than few widgets in ui file anyway.

picard/ui/logview.py
```diff
@@ -25,6 +27,8 @@
 
 class LogViewCommon(PicardDialog):
 
+    default_verbosity = {log.LOG_INFO, log.LOG_WARNING, log.LOG_ERROR, log.LOG_DEBUG}
```
samj1912: I am not sure how much sense this makes. Generally debug includes all the levels above it. So if debug is checked all errors, info and warning logs should come too.
This is default behaviour in almost all loggers.

picard/ui/logview.py
```diff
@@ -37,7 +41,11 @@ def __init__(self, title, logger, w=740, h=340, parent=None):
         self.browser.setDocument(self.doc)
         self.vbox = QtWidgets.QVBoxLayout(self)
         self.vbox.addWidget(self.browser)
-        self._display()
+        self._setup_formats()
+        self.verbosity = self.default_verbosity
```
samj1912: Dont really get the use of the debug checkbox and this drop down. We can keep only this and remove the checkbox?

picard/ui/logview.py
```diff
@@ -37,7 +41,11 @@ def __init__(self, title, logger, w=740, h=340, parent=None):
...
+        self.hidden_domains = set()
```
samj1912: Again, don't get why both domains and show only domains are needed? Does selecting domains discard messages altogether and `show only` create a temp filter? If so, we should keep the distinction between filters and records separate.

picard/ui/logview.py
```diff
@@ -97,31 +125,159 @@ def restoreWindowState(self, position, size):
...
+        self.show_only_domains_menu = QtWidgets.QMenu(self)
+        self.show_only_domains_menu_button = QtWidgets.QPushButton(_("Show Only"))
+        self.show_only_domains_menu_button.setMenu(self.show_only_domains_menu)
+        self.show_only_domains_menu.aboutToShow.connect(partial(self.set_domains_menu_can_update, False))
+        self.show_only_domains_menu.aboutToHide.connect(partial(self.set_domains_menu_can_update, True))
+
+        self.hbox.addWidget(self.domains_menu_button)
+        self.hbox.addWidget(self.show_only_domains_menu_button)
+
+        self.highlight_text = QtWidgets.QLineEdit()
```
samj1912: Pressing enter after inputting text shifts the UI to verbosity selection instead of applying the HL.

picard/ui/logview.py
```diff
@@ -97,31 +125,159 @@ def restoreWindowState(self, position, size):
...
+        self.hbox.addWidget(self.highlight_text)
+
+        self.highlight_button = QtWidgets.QPushButton(_("Highlight"))
+        self.hbox.addWidget(self.highlight_button)
+        self.highlight_button.clicked.connect(self._highlight_do)
+
+        self.clear_log_button = QtWidgets.QPushButton(_("Clear Log"))
```
samj1912: Can we add some tooltips for all these options? It is unclear to a user whether these are destructive or just temp.

picard/ui/logview.py
zas: That's not the behaviour of the logger here, but the behaviour of the log viewer: each log message is set to a certain level (and that's the default behaviour of almost all loggers), whether you want to log or display a certain level is up to you.
For example, i may want to write only debug messages to one debug file, and the rest in a common file, or each level in a separate file, that's not defined by the logger, but by the log handler.

The idea here is to reduce the view to what is interesting, if i know the thing i'm looking for is a debug message, i want to hide the other levels. And this is possible here.

picard/ui/logview.py
zas: > Dont really get the use of the debug checkbox and this drop down. We can keep only this and remove the checkbox?

That's 2 different things, once again, you mix up the logger and the log viewer:
- debug mode controls if we log debug messages (if disabled, they aren't written anywhere at all)
- verbosity debug unchecked will hide the debug messages from the view, but they are still outputted to console for example

samj1912: Hmm, can we separate the logger config from the log viewer config?

picard/ui/logview.py
zas: Because domains is a tuple (even if a trick allows to pass just a string, when the tuple should have only one element), example:

- a message A with domains `('ws', 'ratecontrol')`
- a message B with domains `('ws')`
- a message C with domains `('ratecontrol')`
- a message D with domains `('ws', 'ratecontrol', 'subratecontrol')`

Show only `(ws)` shows A,B,D (only)
Show only `(ws, ratecontrol)` shows A, B, C,D

Show only `(ws, ratecontrol)` + hide `('subratecrontrol')` shows A, B, C but not D

Show only totally unset + hide ('ws') shows C only

That said, i didn't write tests to ensure the actual behaviour is the one i expect ;)

Do you get why both are needed ? Include list + exclude list, both working together to provide more than you could do with just one.

picard/ui/logview.py
zas: Yes, i know, i'll clean up those UI quirks later when we are sure which features we want and how (experimental PR)

picard/ui/logview.py
zas: Yes, of course, this option actually clears the circular buffer. UI tuning will come at the end.

picard/ui/logview.py
zas: > Hmm, can we separate the logger config from the log viewer config?

What do you mean ?

Debug mode is just a trick to allow user to enable the debug mode from within the app (before it was only possible using the -d command line option).
Logging level setting could be moved elsewhere in Options, but then it would be a bit cumbersome to toggle debug mode on/off.

Think about the end user:
1. open Log View
2. check Debug mode checkbox
3. do some action
4. press Save as
5. uncheck Debug mode checkbox

Instead of:
1. Open Options
2. go to panel "Advanced"
3. set the log level to Debug (if it wasn't already...)
4. Press OK
5. open the Log View
6. do some action
7. press Save as
8. Open Options
9. go to panel "Advanced"
10. set the log level back to what it was (if you remember)
11. Press OK

samj1912:
Thought of a good idea to add domains to a function - 

```python
def domain(*domains):
    def wrapper(f):
        @wraps(f)
        def func(*args, **kwargs):
            log.debug = partial(log.debug, domains=domains)
            log.info = partial(log.info, domains=domains)
            log.warning = partial(log.warning, domains=domains)
            log.error = partial(log.error, domains=domains)
            return f(*args, **kwargs)
        return func
    return wrapper
```

It can be used as follows
```python
@domain('random', 'more')
def random_func(*args):
    log.debug('logs as the above domains')
```

picard/log.py
```diff
@@ -19,111 +19,142 @@
...
 
-LOG_INFO = 1
-LOG_WARNING = 2
-LOG_ERROR = 4
-LOG_DEBUG = 8
+def debug_mode(enabled):
+    if enabled:
+        main_logger.setLevel(logging.DEBUG)
+    else:
+        main_logger.setLevel(logging.INFO)
```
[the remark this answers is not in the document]
zas: Historically, log was showing anything but debug level

zas:
Replaced by simpler https://github.com/metabrainz/picard/pull/834

PR #832, opened 2018-02-08
Improve search feature and related code

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other
* **Description**:

- simplify code, reduce redundancy
- when mbid / entity search is triggered by the search string, rewrite it to make it obvious (ie. pasted url)

Before the search:
capture du 2018-02-08 11-36-57

After the search:
capture du 2018-02-08 11-37-30

- add a contextual menu (right-click) on the search button to toggle search options (advanced syntax & builtin search). It is much faster than digging Options ...

capture du 2018-02-08 11-35-43


picard/browser/filelookup.py
```diff
@@ -115,8 +117,8 @@ def tag_lookup(self, artist, release, track, trackNum, duration, filename):
     def collection_lookup(self, userid):
         return self._build_launch('/user/%s/collections' % userid)
 
-    def _search(self, type_, query, adv=False):
-        if self.mbid_lookup(query, type_):
+    def search_entity(self, type_, query, adv=False, mbid_matched_callback=None):
```
samj1912: PR looks good to me, although I wonder if it would be nice to have a pyqtsignal for this instead.
Not sure.

samj1912:
1 little bug, in case you haven't typed anything, the search button is disabled and you can't get the contextual menu. Also it is a bit hidden to have this option.

samj1912:
Another bug - copy any random mbid, for eg an artist mbid and paste it in the search box. It will search for that mbid as a release and then it will load a corrupt album. Like so - 
https://i.imgur.com/3ucxMra.png

zas:
> 1 little bug, in case you haven't typed anything, the search button is disabled and you can't get the contextual menu. Also it is a bit hidden to have this option.

Yes, i'm aware of it, but it isn't a bug, it is a limitation because when the button is disabled... it is disabled, and events on it are ignored.
Since that's just a shortcut, i think we can live with it, i don't want to enable the button if there's nothing to search.

zas:
> Another bug - copy any random mbid, for eg an artist mbid and paste it in the search box. It will search for that mbid as a release and then it will load a corrupt album. Like so - https://i.imgur.com/8IFOj3d.png

There's no way to know if the mbid is valid for this entity or no, so it tries to load the album with the specified mbid, and it fails.
Searches require entity and mbid; when you paste just an mbid it will use the current type of search selected (Album->release, etc... attempting to guess).

If you select Artist and paste a release mbid it will fail too, but since it doesn't try to load a release in Picard, that will be an error on MB, the result is different.

Note: this "feature" was added for convenience, it allows one to paste an album url for example, and load the album in Picard.
The search box is also able to parse other kind of entities (try to paste https://musicbrainz.org/area/dc10c22b-e510-4006-8b7f-fecb4f36436e) but then it will just open the browser

samj1912:
But if its an unknown entity, it should not trigger the callback and try to load the album no?

We can easily check if the call returned a valid entity.

picard/browser/filelookup.py
zas: > PR looks good to me, although I wonder if it would be nice to have a pyqtsignal for mbid_match instead.
> Not sure.

Not a bad idea, but FileLookup doesn't inherit from QObject. I'll test this approach, dunno if it is worth.

zas:
> But if its an unknown entity, it should not trigger the callback and try to load the album no?
> 
> We can easily check if the call returned a valid entity.

It matches for known entities: https://github.com/zas/picard/blob/4f58a45a3fcdc7a9c706ccaa2c5cc0ea7c7c3059/picard/browser/filelookup.py#L89
It checks mbid too.

But it doesn't know if the mbid is valid or not, for this entity.

Do i miss something?

PR #833, opened 2018-02-09
Add support for 1200 px CAA thumbnail size

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Add support for 1200 px CAA thumbnail size

# Problem

- 1200px thumbnails aren't supported
- size was stored as index in the combo box, compatibility issue 
- items were set in .ui, not dynamically built
- QtDesigner doesn't support setting user data to QComboBox items

* JIRA ticket: [PICARD-1189](https://tickets.metabrainz.org/browse/PICARD-1189)

# Solution

- add 1200 px thumbnail choice
- compatibility preserved using a trick: any value between 0 and 2 is old way, new way uses -1 and values >= 250
- safe default if value is outside range is read from config file
- fallback on smaller thumbnails if desired size isn't available

# Action

Review & test


picard/coverart/providers/caa.py
```diff
@@ -148,7 +191,18 @@ def __init__(self, parent=None):
         self.ui.select_caa_types.clicked.connect(self.select_caa_types)
 
     def load(self):
-        self.ui.cb_image_size.setCurrentIndex(config.setting["caa_image_size"])
+        self.ui.cb_image_size.clear()
+        for item_id, item in _CAA_THUMBNAIL_SIZE_MAP.items():
+            self.ui.cb_image_size.addItem(_(item.label), userData=item_id)
+
+        size = config.setting["caa_image_size"]
+        if size in _CAA_SIZE_COMPAT:
```
samj1912: Rather than introducing this, wouldn't it be better to just have a plugin upgrade hook?

zas: It was to avoid an upgrade hook. I pondered both approaches, and this one is much simpler and reliable.

samj1912: I just feel that this more of a compat issue with picard versions and should be kept in upgrade hooks, it makes it easier to track config errors. This approach adds unrelated code to the view IMHO.

Also, why are plugin hooks unreliable or this approach more reliable?

zas: Mostly because upgrade hooks don't work well when someone switch between versions, this code would handle a switch, or even a buggy entry.
This is perfectly possible to do with an upgrade hook though, feel free to propose a patch.

PR #834, opened 2018-02-09
Rework logging system and in-app log view

zas:
- use standard logging module as backend
- implement a circular buffer handler for in-app log view
- use default logging levels
- debug mode is now tied to verbosity
- default log format changed, adding source/lineno infos and milliseconds precision
- history log changed to logging/circular buffer handler too
 
- log view was changed accordingly
- Save As button added
- Clear Log (only the in-memory buffer and the view) button added, with confirmation dialog
- Highlight text and button added
- Verbosity toggling menu added
- Debug mode check box removed (replaced by verbosity = Debug)

Replace https://github.com/metabrainz/picard/pull/830

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -29,7 +29,7 @@ from PyQt5 import QtCore
 
 _MAX_TAIL_LEN = 10**6
 
-VERBOSITY_DEFAULT = logging.INFO
+VERBOSITY_DEFAULT = logging.WARNING
 
 
 def debug_mode(enabled):
--- a/picard/log.py
+++ b/picard/log.py
@@ -17,113 +17,206 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-import sys
+import io
+import logging
 import os
-from collections import deque
+import sys
+import traceback
+
+from collections import deque, namedtuple, OrderedDict
 from PyQt5 import QtCore
-from picard.util import thread
 
 
-LOG_INFO = 1
-LOG_WARNING = 2
-LOG_ERROR = 4
-LOG_DEBUG = 8
+_MAX_TAIL_LEN = 10**6
+
+VERBOSITY_DEFAULT = logging.INFO
 
 
-class Logger(object):
-
-    def __init__(self, maxlen=0):
-        self._receivers = []
-        self.maxlen = maxlen
-        self.reset()
-
-    def reset(self):
-        if self.maxlen > 0:
-            self.entries = deque(maxlen=self.maxlen)
-        else:
-            self.entries = []
-
-    def register_receiver(self, receiver):
-        self._receivers.append(receiver)
-
-    def unregister_receiver(self, receiver):
-        self._receivers.remove(receiver)
-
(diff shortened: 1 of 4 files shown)


picard/ui/logview.py
```diff
@@ -97,48 +54,271 @@ def restoreWindowState(self, position, size):
...
+            act.triggered.connect(partial(self.verbosity_changed.emit, level))
+            self.action_group.addAction(act)
+            self.addAction(act)
+            self.actions[level] = act
+
+    def set_verbosity(self, level):
+        self.actions[level].setChecked(True)
+
 
 class LogView(LogViewCommon):
```
samj1912: I get this warning from Qt - `QLayout: Attempting to add QLayout "" to LogView "", which already has a layout`
should it matter?

picard/ui/logview.py
```diff
@@ -97,48 +54,271 @@ def restoreWindowState(self, position, size):
...
+        self.clear_log_button = QtWidgets.QPushButton(_("Clear Log"), self)
+        self.hbox.addWidget(self.clear_log_button)
+        self.clear_log_button.clicked.connect(self._clear_log_do)
+
+        # save as
+        self.save_log_as_button = QtWidgets.QPushButton(_("Save As..."), self)
+        self.hbox.addWidget(self.save_log_as_button)
+        self.save_log_as_button.clicked.connect(self._save_log_as_do)
+
+        # ------
```
samj1912: Do we need this?

picard/ui/logview.py
zas: That's a parenting conflict, when you set explicitly a parent and Qt think different.
Too much parenting kills parenting ;)
Try to see if it is better with https://github.com/metabrainz/picard/pull/834/commits/5699bb14d1ee81f6053c91c08c18d2f01d0d2982

picard/ui/logview.py
zas: > Do we need this?

No, removed in 5699bb1

picard/log.py
```diff
@@ -17,113 +17,206 @@
...
-                import traceback
-                traceback.print_exc()
+class OurLogger(logging.getLoggerClass()):
 
-    def log_level(self, level):
-        return True
 
+    # copied from https://github.com/python/cpython/blob/3.5/Lib/logging/__init__.py#L1353-L1381
+    # see https://stackoverflow.com/questions/4957858/how-to-write-own-logging-methods-for-own-logging-levels
+    def findCaller(self, stack_info=False):
```
mineo: Maybe I'm missing something, but why do we need our own copy of this method? Replacing this whole class with

```python
class OurLogger(logging.getLoggerClass()):
    pass
```

appears to result in exactly the same log messages.

Also, technically, there's a license declaration at the top of `logging.py` in cpython that this copy here appears to violate (but I am not a lawyer).

zas: We need it because logging module is wrapped (log.* methods) and therefore its findCaller method is returning log method file/lineno etc.. instead of the method executing log.*().


About the license thing i think you're right, it should be copied as well, just above this code (or?).

zas: https://github.com/metabrainz/picard/pull/839 removes the useless code ;)

PR #835, opened 2018-02-09
PICARD-1187: Add DSF file support

zas:
Fix https://tickets.metabrainz.org/browse/PICARD-1187


zas:
Do not pass tests yet.

zas:
@phw i'm not at ease with formats in Picard, if you can give me a hand on this ... ;)

picard/formats/id3.py
```diff
@@ -595,6 +596,36 @@ def _info(self, metadata, file):
         metadata['~format'] = self.NAME
 
 
+class DSFFile(ID3File):
```
samj1912: We can over-ride the supports_tag method in the class to exclude the unsupported tags so that they don't appear in the UI.

PR #837, opened 2018-02-12
PICARD-1192: Add .txt extension in path if no extension is there

vishichoudhary:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Currently if you click on save as option in log dialog. And if you doesn't specify .txt or .TXT format it will not save that file

* JIRA ticket (_optional_): [PICARD-1192](https://tickets.metabrainz.org/browse/PICARD-1192)

# Solution
I simply check the file and if there isn't any .txt or .TXT extension add '.txt' extension

# Action

### DIFF
--- a/picard/ui/logview.py
+++ b/picard/ui/logview.py
@@ -248,6 +248,8 @@ class LogView(LogViewCommon):
             filter=_("Text Files (*.txt *.TXT)"),
             options=QtWidgets.QFileDialog.DontConfirmOverwrite
         )
+        if not path.endswith('.txt') and not path.endswith('.TXT'):
+            path = path + '.txt'
         if ok and path:
             if os.path.isfile(path):
                 reply = QtWidgets.QMessageBox.question(


picard/ui/logview.py
```diff
@@ -248,6 +248,8 @@ def _save_log_as_do(self):
             filter=_("Text Files (*.txt *.TXT)"),
             options=QtWidgets.QFileDialog.DontConfirmOverwrite
         )
+        if not path.endswith('.txt') and not path.endswith('.TXT'):
+            path = path + '.txt'
```
samj1912: Not necessarily wanted behaviour. Logs may also be stored with a different extension like picard.log or even without an extension.

samj1912: Although I am not entirely sure the intentions with which this feature was implemented since there is a text file filter above. @zas any thoughts?

vishichoudhary: @samj1912 i don't know why this happens. This was a small fix. Though what i found interesting is that write operation gives output False if you havn't specifiy any extension

zas: Hmmm, imho we should remove `filter=_("Text Files (*.txt *.TXT)")`, this is why it fails if none of those extensions is set.

vishichoudhary: @zas removing the filter doesn't seem to have any effect on it. Let me just check the reason behind this behaviour.

zas: I implemented a fix in https://github.com/metabrainz/picard/pull/840

PR #838, opened 2018-02-13
Handle caa_image_size option change in an upgrade hook, bump version to 2.0.0dev3

zas:
Suggested by sambhav in https://github.com/metabrainz/picard/pull/833/files#r167294289

It also fixes the issue reported at https://tickets.metabrainz.org/browse/PICARD-1189?focusedCommentId=45336&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-45336

### DIFF
--- a/picard/config_upgrade.py
+++ b/picard/config_upgrade.py
@@ -222,6 +222,25 @@ def upgrade_to_v1_4_0_dev_7():
         _s.remove(old_opt)
 
 
+def upgrade_to_v2_0_0_dev_3():
+    """Option "caa_image_size" value has different meaning."""
+    _s = config.setting
+    opt = "caa_image_size"
+    if opt in _s:
+        # caa_image_size option was storing index of a combobox item as size
+        # therefore it depends on items order and/or number, which is bad
+        # To keep the option as is, values >= 250 are stored for thumbnails and -1 is
+        # used for full size.
+        _CAA_SIZE_COMPAT = {
+            0: 250,
+            1: 500,
+            2: -1,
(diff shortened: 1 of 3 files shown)


picard/config_upgrade.py
```diff
@@ -222,6 +222,25 @@ def upgrade_to_v1_4_0_dev_7():
...
+        # To keep the option as is, values >= 250 are stored for thumbnails and -1 is
+        # used for full size.
+        _CAA_SIZE_COMPAT = {
+            0: 250,
+            1: 500,
+            2: -1,
+        }
+        value = _s[opt]
+        if value in _CAA_SIZE_COMPAT:
+            _s[opt] = _CAA_SIZE_COMPAT[value]
```
samj1912: There seems to be some sort of bug here. Here's how to reproduce it -
Put the value for caa_image_size as 0 in the config file. Open picard. Dont open settings box. 
Load an album, it loads the full size CA image.

zas: I can't reproduce this issue:
- i set caa_image_size as 0 in the config file, and version=2.0.0.dev2
- i start picard, caa_image_size is set to 250 as per upgrade hook
- i load a release having 250px, 500px and 1200px thumbnails
- i check the thumbnail loaded is <=250px
- in Options > Cover Art > CAA, the combo box is on 250px

samj1912: Ah yes, sorry, it had updated the version on the second test. Didn't notice.

PR #839, opened 2018-02-13
Remove unnecessary code to figure out logger name

samj1912:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -17,11 +17,9 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-import io
 import logging
 import os
-import sys
-import traceback
+import pathlib
 
 from collections import deque, namedtuple, OrderedDict
 from PyQt5 import QtCore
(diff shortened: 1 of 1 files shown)


picard/log.py
```diff
@@ -56,67 +54,8 @@ def debug_mode(enabled):
...
-else:  # pragma: no cover
-    def currentframe():
-        """Return the frame object for the caller's stack frame."""
-        try:
-            raise Exception
-        except Exception:
-            return sys.exc_info()[2].tb_frame.f_back
-
-
 class OurLogger(logging.getLoggerClass()):
```
mineo: I think you can get rid of this class completely - at least doing that didn't change any of the messages during picard startup with debug mode enabled.

picard/log.py
```diff
@@ -165,11 +104,9 @@ def clear(self):
 
 def name_filter(record):
     # provide a significant name, because module sucks
-    prefix = os.path.dirname(os.path.normcase(_srcfile))
-    name = record.pathname
-    if name.startswith(prefix):
-        name = name[len(prefix) + 1:].replace(os.path.sep, '.').replace('.__init__', '')
-    record.name, _ = os.path.splitext(name)
+    name = pathlib.Path(record.pathname)
```
mineo: It took me a while to figure out what this function does - I see that it strips `picard` from the beginning of a path, removes the file extension and any occurences of `.__init__` and for plugins adds the whole path to the file at the beginning of the line, right? In that case, sure, looks reasonable.

Note that before this change, plugin logs were like

> /home/wieland/.config/MusicBrainz/Picard/plugins/keep.transltag:19:

now they're like

>  home.wieland..config.MusicBrainz.Picard.plugins.keep.transltag:19:

zas: > Note that before this change, plugin logs were like
> 
>     /home/wieland/.config/MusicBrainz/Picard/plugins/keep.transltag:19:
> 
> now they're like
> 
>     home.wieland..config.MusicBrainz.Picard.plugins.keep.transltag:19:

Not very fancy, that's only for plugins right ?
Can you post a bigger part of your log ?

PR #840, opened 2018-02-13
PICARD-1192: set QTextDocumentWriter format to plaintext, and allow any filename

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Saving log file with anything but .txt extension doesn't work

* JIRA ticket: fix [PICARD-1192](https://tickets.metabrainz.org/browse/PICARD-1192)

# Solution

- plaintext format was set from filename, which doesn't work if the file isn't a .txt
- filter prevents selecting other filename, one may want to save to picard.log
- an error message will be displayed if something prevented the file to be written

# Action

Test & review


zas:
Replace https://github.com/metabrainz/picard/pull/837

vishichoudhary:
@zas . We can still go with filters, to hide other extension file from user. Also we can show message if the file has been created successfully. Something like "logs have been saved in ----- file ."

zas:
> @zas . We can still go with filters, to hide other extension file from user. Also we can show message if the file has been created successfully. Something like "logs have been saved in ----- file ."

About filters, if we decide the user can choose whatever extensions he wants (including none), there's no need of filters, that would mislead the user in thinking only those extensions can be used. Note that meaning of extensions is mostly a windows thing.

Since this PR displays a message if something is wrong, there's no point to show a message in case of success (that would add one more action to close the message).
http://www.faqs.org/docs/artu/ch01s06.html#id2878450

picard/ui/logview.py
```diff
@@ -271,7 +276,7 @@ def _clear_log_do(self):
         reply = QtWidgets.QMessageBox.question(
             self,
             _("Clear Log"),
-            _("Are you sure you want to clear the log?"),
+            ("Are you sure you want to clear the log?"),
```
samj1912: Why remove i18n?

zas: > Why remove i18n?

Is "because i suck" an acceptable answer ? ;)
Fixed and rebased, thanks for spotting this typo.

PR #841, opened 2018-02-14
PICARD-1194: Show all files in details section of plugin options page

vishichoudhary:
screenshot from 2018-02-15 03-16-45
screenshot from 2018-02-15 03-16-20

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Currently picard shows only one file of plugin if there exists many. Screenshot has been added for more explanation. It shows correct values for plugins which are not installed.

* JIRA ticket (_optional_): [PICARD-1194](https://tickets.metabrainz.org/browse/PICARD-1194)

# Solution
I simply check if plugin is in zip form list all the files inside it else return the same file name.

# Action

### DIFF
--- a/picard/plugin.py
+++ b/picard/plugin.py
@@ -213,12 +213,11 @@ class PluginWrapper(PluginShared):
 
     @property
     def files_list(self):
-        path = os.path.join(USER_PLUGIN_DIR,self.module_name)
-        path = path + '.zip'
+        path = os.path.join(USER_PLUGIN_DIR, self.module_name + '.zip')
         if os.path.isfile(path):
-            return ', '.join(zipfile.ZipFile(path).namelist())
+            return zipfile.ZipFile(path).namelist()
         else:
-            return self.module_name + '.py'
+            return [self.module_name + '.py']
 
 class PluginData(PluginShared):
 
--- a/picard/plugin.py
+++ b/picard/plugin.py
@@ -213,8 +213,12 @@ class PluginWrapper(PluginShared):
 
     @property
     def files_list(self):
-        return self.file[len(self.dir)+1:]
-
+        path = os.path.join(USER_PLUGIN_DIR,self.module_name)
+        path = path + '.zip'
+        if os.path.isfile(path):
+            return ', '.join(zipfile.ZipFile(path).namelist())
+        else:
+            return self.module_name + '.py'
 
 class PluginData(PluginShared):
 
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -352,7 +352,7 @@ class PluginsOptionsPage(OptionsPage):
             text.append("<b>" + _("Authors") + "</b>: " + plugin.author)
         if plugin.license:
             text.append("<b>" + _("License") + "</b>: " + plugin.license)
-        text.append("<b>" + _("Files") + "</b>: " + plugin.files_list)
+        text.append("<b>" + _("Files") + "</b>: " + ', '.join(plugin.files_list))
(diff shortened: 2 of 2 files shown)


picard/plugin.py
```diff
@@ -213,8 +213,12 @@ def license_url(self):
 
     @property
     def files_list(self):
-        return self.file[len(self.dir)+1:]
-
+        path = os.path.join(USER_PLUGIN_DIR,self.module_name)
```
zas: Please add a space after comma (PEP8).

Also you can do it in one line:
```python
path = os.path.join(USER_PLUGIN_DIR, self.module_name + '.zip')
if os.path.isfile(path):
```

picard/plugin.py
```diff
@@ -213,8 +213,12 @@ def license_url(self):
...
+        path = path + '.zip'
+        if os.path.isfile(path):
+            return ', '.join(zipfile.ZipFile(path).namelist())
```
zas: Imho, since the method is called `files_list()`, we should always return a list:

```python
if os.path.isfile(path):
    return zipfile.ZipFile(path).namelist()
else:
    return [self.module_name + '.py']
```

And `join()` at display time. It will be easier, if, for example, we want to change from a comma-separated list to a list view or whatever.

zas:
You also need to change: https://github.com/metabrainz/picard/pull/841/files#diff-737a9b322eaff0f4ac8a07dc80c1d554R239

```python
def files_list(self):
    return self.files
```

samj1912:
This is how the output looks like for non-installed plugins - https://i.imgur.com/0PvliBK.png

samj1912:
I very much doubt the use-fullness of such display of information to a user. This is only helpful to a developer. 
For eg. Let's take a look at - https://i.imgur.com/SJW5uih.png

I really doubt the user needs to know what files a plugin uses to give him a particular functionality,
@zas any thoughts on this?

mineo:
I think it could be useful if a user has problems with a plugin because it makes it easy to check that all necessary files are actually in the zip file (especially on platforms where it's not easily possible to calculate checksums for files).

samj1912:
@mineo  An easier way to do that would be to simply display a package verified value here instead that calculates the checksum from the list of files online and the list of files download. And anyway, @zas is continuing the work I did on 2.0 plugins which does exactly this.

zas:
> @zas any thoughts on this?

It can be useful, but the current way it is displayed isn't great (a long list of files separated by a comma...)

zas:
> @mineo An easier way to do that would be to simply display a package verified value(boolean) here instead that calculates the checksum from the list of files online and the list of files download. And anyway, @zas is continuing the work I did on 2.0 plugins which does exactly this.

Well, i'm not continuing much, tried to clean up existing stuff. But i'm not really happy how plugins V2 turns out, it seems overcomplicated to me (and the code reflects that imho).

I'd like we discuss about that on IRC with @mineo, etc..

vishichoudhary:
screenshot from 2018-02-16 23-59-33
@mineo @zas @samj1912 What about this? each file in new line

zas:
> What about this? each file in new line

Better imho, you could even add a newline after `Files:`. The area is scrollable so long lists arent an issue.

vishichoudhary:
Ok @zas i will add.

sambhav:
Any updates on this? Otherwise I will be closing it with a WIP tag.

vishichoudhary:
@sambhav . @zas told me to leave it as we are gonna fix these things with new plugin structure. So yeah you can close it.

PR #842, opened 2018-02-17
PICARD-1196: Allow symlinks in plugin directories

vishichoudhary:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
replaygain is a symlink to /home/vishal/replaygain.zip but yet not listed in installed plugin. 
screenshot from 2018-02-17 14-25-47
screenshot from 2018-02-17 14-28-20

* JIRA ticket (_optional_): [PICARD-1196](https://tickets.metabrainz.org/browse/PICARD-1196)

# Solution
Simply load the target of symlink.
screenshot from 2018-02-17 14-27-55

# Action

### DIFF
--- a/picard/plugin.py
+++ b/picard/plugin.py
@@ -275,6 +275,9 @@ class PluginManager(QtCore.QObject):
             name = is_zip(path)
             if not name:
                 name = _plugin_name_from_path(path)
+                # check for symlink
+                if os.path.islink(path):
+                    name = os.path.realpath(path)
             if name:
                 names.add(name)
         log.debug("Looking for plugins in directory %r, %d names found",


samj1912:
Fix the commit message and PR title. This clearly goes against the contribution guidelines.

vishichoudhary:
@samj1912 apologize for the same. And title and commit message has been updated.

picard/plugin.py
```diff
@@ -275,6 +275,9 @@ def load_plugindir(self, plugindir):
             name = is_zip(path)
             if not name:
                 name = _plugin_name_from_path(path)
+                # check for symlink
+                if os.path.islink(path):
```
zas: Shouldn't the real path passed to `_plugin_name_from_path()` ?

```python
if os.path.islink(path):
    path = os.path.realpath(path)

name = _plugin_name_from_path(path)
```

If it is correct, `islink()` check and following should be moved to `_plugin_name_from_path()`.
And, imho, the comment is not needed, `islink()` is pretty obvious.

vishichoudhary: @zas yes its better to pass original path to the ```_plugin_name_from_path``` as  currently if it is a link then i don't need to call it.

vishichoudhary: @zas one issue arrises while doing this.  while loading the plugin by calling 
```def load_plugin(self, name, plugindir):```
it will use the value of plugindir while it's not correct.

vishichoudhary: One thing we can do is make a tupe in ```names``` list indicating names and directory it is present in. Any thoughts?

zas: I had a look at the code, and it looks like it wasn't meant to support symlinks at all ;)

All those name + plugindir splits & join and manipulation are awful.
I'm not sure it is worth trying to support symlinks with current code :(

vishichoudhary: @zas upto you. I will continue to PICARD-1194.

PR #843, opened 2018-02-20
Codacy fixes

zas:


vishichoudhary:
@zas there are still many files where we can use super(). Should i send a pr for this? or point those file here.

vishichoudhary:
like in this case https://github.com/metabrainz/picard/blob/f73dbe1be3e42f010a4ad5317db2cb2dfb2b283a/picard/ui/cdlookup.py#L42

zas:
> @zas there are still many files where we can use super(). Should i send a pr for this? or point those file here.

@vishichoudhary : yes, if you want, but please, in many small commits in a PR so we can easily review

PR #847, opened 2018-02-24
Use python3 super function instead of calling by class name

vishichoudhary:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
While understanding the picard codebase i am encountering various python2 oriented code. This pr will port them to python3.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -599,7 +599,7 @@ class Album(DataObject, Item):
 class NatAlbum(Album):
 
     def __init__(self):
-        Album.__init__(self, "NATS")
+        super().__init__("NATS")
         self.loaded = True
         self.update()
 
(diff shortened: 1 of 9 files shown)


zas:
@vishichoudhary : what's the status of this ?

vishichoudhary:
@zas i dont know what to do with this pr. As the codebase is big and i can't really look each module. I think we should merge it . But what should we do if i found more ?

samj1912:
@vishichoudhary I think it is very much possible to port individual python 2 patterns to python3 if you do them systematically. Otherwise Picard 2.0 wouldn't be py3 ;).

So that leaves that out as an excuse to merge this PR. I really am not in the favor the the process you described - you finding things and making a commit with 1-2 lines of similar changes. It pollutes the git history.

vishichoudhary:
@samj1912 give me just 3 - 4 days  i will look in each file.

samj1912:
@vishichoudhary  a better way would be to list out all the possible python2 remnants you can find and change them across all the modules (possibly use a code refactoring tool or IDE for this)

vishichoudhary:
@sambhav @zas  . I think it's complete now.

PR #848, opened 2018-02-25
Cleanuptests

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Rework and simplify format tests

# Problem

A lot of duplication, hard to maintain, tests skipped silently, etc...

# Solution

- Use a common tags dict definition, modify it when needed
- raise SkipTest exception instead of silently returning
- wrap classes that inherit from unittest.TestCase but are meant to declare common stuff, preventing invalid tests to be run
- general cleanup

# Action


test/test_formats.py
```diff
@@ -126,265 +126,267 @@ def save_and_load_metadata(filename, metadata):
     'website': 'http://example.com',
 }
 
+# prevent unittest to run tests in those classes
+class CommonTests:
```
samj1912: We have two ways to approaching this without a wrapper class - 
1. Make a mixin
2. Set `__test__` to False

zas: Mixin can work, but using a wrapper class is even simpler, imho.

`__test__` approach isn't really documented, i'd stay away from it.

zas: I tested the `__test__ = False` approach, it doesn't work.

PR #849, opened 2018-02-26
Do not show un-supported DSF tags in UI

samj1912:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

PR #835 added support for DSF tags(\o/), but users are still shown unsupported tags in the UI. This fixes that.

# Solution

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -625,6 +625,10 @@ class DSFFile(ID3File):
             tags.update_to_v24()
             tags.save(filename, v2_version=4)
 
+    def supports_tag(self, name):
+        return (super().supports_tag(name)
+                and name not in ['albumsort', 'artistsort', 'discsubtitle', 'titlesort'])
+
 
 if mutagen.aiff:
     class AiffFile(ID3File):


picard/formats/id3.py
```diff
@@ -625,6 +625,10 @@ def _save_tags(self, tags, filename):
             tags.update_to_v24()
             tags.save(filename, v2_version=4)
 
+    def supports_tag(self, name):
+        return (super().supports_tag(name)
+                and name not in ['albumsort', 'artistsort', 'discsubtitle', 'titlesort'])
```
zas: It looks like it also applies to `AiffFile`.
We could reduce the code duplication between `DSFFile` & `AiffFile`.

Also, since unsupported tags are now all listed in `tests/test_formats.py`, perhaps we should just move those lists there, and use them in tests.

See, for example, https://github.com/metabrainz/picard/blob/master/test/test_formats.py#L379-L395
If those tags aren't supported by the format, listing them in format declaration would make much more sense.

What do you think ?

zas: Just a quick note about something i just thought about.

In expressions like:

```python
name not in ['albumsort', 'artistsort', 'discsubtitle', 'titlesort']
```

We should use a set instead:

```python
name not in {'albumsort', 'artistsort', 'discsubtitle', 'titlesort'}
```

In practice, it wouldn't not make much difference (since this isn't executed many times), but i see no reason to use a tuple or a list when we can use a set.

Btw, with the `in` operator a set is faster of course.

```python
number = 100000000
t = OrderedDict()

t['list'] = timeit(
    "e in s",
    setup="e='b'; s=['a','b','c']",
    number=number
)

t['tuple'] = timeit(
    "e in s",
    setup="e='b'; s=('a','b','c')",
    number=number
)

t['set'] = timeit(
    "e in s",
    setup="e='b'; s={'a','b','c'}",
    number=number
)

for k, v in t.items():
    print("%-5s: %0.3f" % (k, v))
```

```
list : 8.208
tuple: 7.599
set  : 5.580
```

test/test_formats.py
```diff
@@ -146,13 +147,14 @@ class FormatsTest(unittest.TestCase):
 
         def setUp(self):
             self.tags = TAGS.copy()
-            self.setup_tags()
+            _, self.testfile_ext = os.path.splitext(self.testfile)
```
zas: Just a minor thing, using `_` as place holder for unused value is pretty standard in Python, but since we are using `_()` for gettext/i18n, it is a bad idea to overwrite it.
Please use something else (like `root`, used in `os.path.splitext()` documentation, but you lose the unused info, or just `unused`)

picard/formats/apev2.py
```diff
@@ -204,8 +204,22 @@ def _get_tag_name(self, name):
...
+            'showsort',
+            'show',
+            'arranger',
+            'musicip_fingerprint',
+            'podcast',
+            'totaltracks'}
+        return (bool(name)
+                and name not in unsupported_tags
+                and not name.startswith('performer')
+                and not name.startswith('comment')
```
zas: Shouldn't we use `name.startswith('performer:') and not name.startswith('comment:')` instead ?

PR #850, opened 2018-02-27
PICARD-1200: fix Options tree pane resizing

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Options tree pane resizing is impossible, the behaviour is unexpected.

* JIRA ticket : [PICARD-1200](https://tickets.metabrainz.org/browse/PICARD-1200)

# Solution

Size policy of both panes plus the minimum width enforced for the tree pane are responsible of the issue

Set horizontal policies to Ignored, and use QSplitter.setSizes() to set a smart default.
After this change, it is possible to move the splitter smoothly.

# Action

Review & test


samj1912:
Hmm, while playing with this I noticed a pretty bad GUI bug, we can completely hide the left side options pane, which is very counter-intuitive imo.

zas:
> Hmm, while playing with this I noticed a pretty bad GUI bug, we can completely hide the left side options pane, which is very counter-intuitive imo. This is also present in master.

Yes, we can prevent it by setting a minimum width, though it isn't a real issue, since the splitter handle is still available. Not worth to bother imho.

> Another bug (Also present in master) is that if I expand a menu item and close options and open options again, it is again collapsed.

That's not a bug, the trick to get correct tree pane width is expanding then collapsing the tree, so it always starts collapsed.
There's a feature request to save the current option pane and restore it when the options dialog is open again, it could be part of such improvement.

PR #851, opened 2018-02-27
PICARD-1201: Add -N/--no-restore option to not restore persisted sizes and positions

zas:
It is useful for:
- users having UI issues on startup (too small, too big, out of screen, etc...)
- devs for testing purposes

Fix [PICARD-1201](https://tickets.metabrainz.org/browse/PICARD-1201)


zas:
@samj1912 , @mineo : tell me what you think about this

Related ticket: https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-592
Note: it will not prevent badly designed plugins to create oversized option pages.

PR #854, opened 2018-03-01
PICARD-1204: Read and dump values to QSettings only when necessary

sambhav:
Reading and dumping too often from QSettings causes picard to lock up.
This makes sure we only read on start and dump on change.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1204](https://tickets.metabrainz.org/browse/PICARD-1204)

# Solution

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -34,8 +34,23 @@ class ConfigSection(LockableObject):
 
     def __init__(self, config, name):
         super().__init__()
-        self.__config = config
+        self.__qt_config = config
+        self.__config = {}
         self.__name = name
+        self.load_keys()
+        self.__qt_config.parent().register_cleanup(self.dump_keys)
+
+    def load_keys(self):
+        for key in self.__qt_config.allKeys():
+            self.__config[key] = self.__qt_config.value(key)
+
+    def dump_keys(self):
+        log.debug('Dumping config(%s) to file.', self.__name)
+        for key in self.__qt_config.allKeys():
+            if key not in self.__config:
+                self.__qt_config.remove(key)
+        for key, value in self.__config.items():
+            self.__qt_config.setValue(key, value)
 
     def __getitem__(self, name):
         opt = Option.get(self.__name, name)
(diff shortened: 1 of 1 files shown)


picard/config.py
```diff
@@ -34,8 +34,18 @@ class ConfigSection(LockableObject):
...
     def __init__(self, config, name):
         super().__init__()
-        self.__config = config
+        self.__qt_config = config
+        self.__config = {}
         self.__name = name
+        self.load_keys()
+
+    def __qt_keys(self):
+        return filter(lambda key: key.startswith('%s/' % self.__name),
```
zas: You may want to cache `'%s/' % self.__name`:

```python
def __qt_keys(self):
   prefix = '%s/' % self.__name
   def prefixed(key):
       return key.startswith(prefix)
    return filter(prefixed, self.__qt_config.allKeys())
```

picard/tagger.py
```diff
@@ -230,7 +231,6 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
         self.unclustered_files = UnclusteredFiles()
         self.nats = None
         self.window = MainWindow()
-        self.exit_cleanup = []
```
zas: This change isn't needed anymore.

picard/config.py
```diff
@@ -34,8 +34,19 @@ class ConfigSection(LockableObject):
...
     def __init__(self, config, name):
         super().__init__()
-        self.__config = config
+        self.__qt_config = config
+        self.__config = {}
         self.__name = name
+        self.load_keys()
+
+    def __qt_keys(self):
+        prefix = '%s/' % self.__name
```
zas: Instead of using % format, since key building here are very simple, using concatenation will waste less cpu cycles.
Since config options can be read in loops, optimization makes sense, and doesn't cost much in readibility.

```python
prefix = '%s/' % self.__name
key = "%s/%s" % (self.__name, name)
```

vs

```python
prefix = self.__name + '/'
key = self.__name + '/' + name
```

```python
from timeit import timeit

runs = 10000000


def print_results(time, start_string):
    print('%s\nTotal: %.4f\nAvg: %.4fns\n' % (start_string, time,
                                              (time/runs)*1000000000))

t1 = timeit('"%s/%s" % (a, b)',
            setup='a="xxx"; b="yyy"',
            number=runs)
t2 = timeit("a + '/' + b",
            setup='a="xxx"; b="yyy"',
            number=runs)
t3 = timeit('a + "/" + b',
            setup='a="xxx"; b="yyy"',
            number=runs)
t4 = timeit('"{}/{}".format(a, b)',
            setup='a="xxx"; b="yyy"',
            number=runs)

print_results(t1, '% replacement')
print_results(t2, 'concatenation single quotes')
print_results(t3, 'concatenation double quotes')
print_results(t4, '.format method')
```

```
% replacement
Total: 3.1121
Avg: 311.2064ns

concatenation single quotes
Total: 1.6318
Avg: 163.1769ns

concatenation double quotes
Total: 1.7592
Avg: 175.9204ns

.format method
Total: 4.3462
Avg: 434.6209ns
```

picard/config.py
```diff
@@ -44,35 +55,42 @@ def __getitem__(self, name):
...
+        self.lock_for_write()
+        try:
+            if key in self.__config:
+                self.__config.pop(key)
+                self.__qt_config.remove(key)
+        finally:
+            self.unlock()
 
-    def raw_value(self, key):
+    def raw_value(self, name):
```
sambhav: Did this so that there is a consistency between `name` and `key`

zas: Ok ;)

PR #858, opened 2018-03-04
InfoDialog: reduce code redundancy, subclassing QWidget

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -32,6 +32,30 @@ from picard.ui import PicardDialog
 from picard.ui.ui_infodialog import Ui_InfoDialog
 
 
+class ArtworkCoverWidget(QtWidgets.QWidget):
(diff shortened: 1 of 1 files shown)


picard/ui/infodialog.py
```diff
@@ -32,6 +32,33 @@
 from picard.ui.ui_infodialog import Ui_InfoDialog
 
 
+class ArtworkCoverWidget(QtWidgets.QWidget):
+    """A QWidget that can be added to artwork column cell of ArtworkTable."""
+
+    SIZE = 170
```
sambhav: Not sure if this should be caps.
Also maybe use a tuple here instead?

zas: I used caps because it is constant, and since height=width in this case, only one value is needed.
I don't feel the need to change it.

PR #170, opened 2013-10-20
PICARD-259: Allow tagger script to access file metadata

kepstin:
This moves the taggerscript evaluation to happen when a file is matched (i.e. when a file is added to a track), where it has access to the both the file metadata _and_ the track metadata.

In order to make the UI work correctly, track.metadata has the tagger script run at load time, but a copy of the pre-tagger-script metadata is saved in track.orig_metadata. At file matching time, the metadata is copied from the track.orig_metadata to the file, and the tagger script is re-run in a context that has file information present.

PR #859, opened 2018-03-05
Allow tagger script to access file metadata

kepstin:
Fixes PICARD-259 (and associated sub-tickets)

This moves the taggerscript evaluation to happen when a file is matched (i.e. when a file is added to a track), where it has access to the both the file metadata *and* the track metadata.

In order to ensure that the titles in the user interface are correct, this is done by saving an extra copy of the original musicbrainz metadata in track.orig_metadata. The taggerscript is run as per normal
on track.metadata, but the file matching will use track.orig_metadata and re-run the taggerscript on the merged data.

This is an updated version of the code previously submitted as #170 

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-259](https://tickets.metabrainz.org/browse/PICARD-259)

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -25,6 +25,7 @@ import operator
 from functools import reduce
 from collections import namedtuple
 from inspect import getfullargspec
+from picard import config
 from picard.metadata import Metadata
 from picard.metadata import MULTI_VALUED_JOINER
 from picard.plugin import ExtensionPoint
@@ -282,6 +283,13 @@ Grammar:
         if key not in ScriptParser._cache:
             ScriptParser._cache[key] = self.parse(script, True)
         return ScriptParser._cache[key].eval(self)
+
+def enabled_tagger_scripts_texts():
+    """Returns an iterator over the enabled tagger scripts.
+    For each script, you'll get a tuple consisting of the script name and text"""
+    if not config.setting["enable_tagger_scripts"]:
+        return []
+    return [ (s_name, s_text) for _s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"] if s_enabled and s_text]
 
 
 def register_script_function(function, name=None, eval_args=True,
--- a/picard/track.py
+++ b/picard/track.py
@@ -25,7 +25,7 @@ from picard.metadata import Metadata, run_track_metadata_processors
 from picard.dataobj import DataObject
 from picard.util.textencoding import asciipunct
 from picard.mbjson import recording_to_metadata
-from picard.script import ScriptParser
+from picard.script import ScriptParser, enabled_tagger_scripts_texts
 from picard.const import VARIOUS_ARTISTS_ID, SILENCE_TRACK_TITLE, DATA_TRACK_TITLE
 from picard.ui.item import Item
 from picard.util.imagelist import update_metadata_images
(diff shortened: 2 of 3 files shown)


sambhav:
From what I test this seems to work nicely. @zas @mineo will you also test this please?

picard/track.py
```diff
@@ -71,8 +71,23 @@ def add_file(self, file):
...
+        # Re-run tagger scripts with updated metadata
+        if config.setting["enable_tagger_scripts"]:
+            for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"]:
+                if s_enabled and s_text:
+                    parser = ScriptParser()
+                    try:
+                        parser.eval(s_text, file.metadata)
+                        parser.eval(s_text, self.metadata)
+                    except:
+                        log.error(traceback.format_exc())
```
zas: Since we moved to logging module you should be able to use `log.exception()` in place of `log.error(traceback.format_exc())`
https://github.com/metabrainz/picard/blob/master/picard/log.py#L137
https://docs.python.org/3/library/logging.html#logging.exception

picard/track.py
```diff
@@ -85,6 +100,23 @@ def remove_file(self, file):
...
+
+        # Restore to non-associated state
+        if config.setting["enable_tagger_scripts"]:
+            for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"]:
+                if s_enabled and s_text:
+                    parser = ScriptParser()
+                    try:
+                        parser.eval(s_text, self.metadata)
+                    except:
+                        log.error(traceback.format_exc())
```
zas: https://github.com/metabrainz/picard/pull/859/files#r172114919

picard/track.py
```diff
@@ -71,8 +71,23 @@ def add_file(self, file):
...
             return
-        file.copy_metadata(self.metadata)
+        file.copy_metadata(self.orig_metadata)
         file.metadata['~extension'] = file.orig_metadata['~extension']
+        self.metadata.copy(file.metadata)
+
+        # Re-run tagger scripts with updated metadata
+        if config.setting["enable_tagger_scripts"]:
+            for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"]:
+                if s_enabled and s_text:
```
zas: To reduce code redundancy i'd introduce a `enabled_tagger_scripts_list()`:

```python
def enabled_tagger_scripts_texts():
    if not config.setting["enable_tagger_scripts"]:
        return []
    return [ s_text for s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"] if s_enabled and s_text]
```

So you can just do:

```python
for s_text in enabled_tagger_scripts_texts():
   parser = ScriptParser()
   try:
        parser.eval(s_text, file.metadata)
        parser.eval(s_text, self.metadata)
   except:
....
```

sambhav: Since pyt3, we can do this - 
```python
def enabled_tagger_scripts_texts():
    if not config.setting["enable_tagger_scripts"]:
        return []
    return [ s_text for *_discard, s_enabled, s_text in config.setting["list_of_scripts"] if s_enabled and s_text]
```

sambhav:
@kepstin  mind writing some tests for this?

picard/track.py
kepstin: I like this change, will apply. I'll put the new function into `picard/script.py`.

kepstin:
I'm honestly not sure where to start with tests. This part of the code doesn't have any existing tests as far as I can tell, and it interacts with the gui (it triggers the update/re-render of list rows).
If you have any suggestions or pointers please let me know.

kepstin:
Refactoring the code to get a list of enabled scripts didn't really reduce code size that much, but it did significantly reduce the code complexity - there's no longer multiple complicated conditions to check for enabled scripts in every location where they get run.

picard/script.py
```diff
@@ -283,6 +284,13 @@ def eval(self, script, context=None, file=None):
             ScriptParser._cache[key] = self.parse(script, True)
         return ScriptParser._cache[key].eval(self)
 
+def enabled_tagger_scripts_texts():
+    """Returns an iterator over the enabled tagger scripts.
+    For each script, you'll get a tuple consisting of the script name and text"""
+    if not config.setting["enable_tagger_scripts"]:
+        return []
+    return [ (s_name, s_text) for _, s_name, s_enabled, s_text in config.setting["list_of_scripts"] if s_enabled and s_text]
```
sambhav: We generally don't use `_` as a discard-able variable in Picard since picard changes `_` to gettext as a builtin.

zas: Yes, please change it, otherwise the PR looks good to me.

kepstin:
I've introduced some variable naming problems in my last update, I'll respin it once I figure out the issue.

PR #862, opened 2018-03-06
PICARD-1206: force style for GroupBox::title

zas:
It is a workaround for a Qt bug, resulting on partially truncated group box titles.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1206](https://tickets.metabrainz.org/browse/PICARD-1206)

# Solution

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -118,6 +118,9 @@ class Tagger(QtWidgets.QApplication):
         # Allow High DPI Support
         self.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps)
         self.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
+        super().setStyleSheet(
+            'QGroupBox::title { /* PICARD-1206, Qt bug workaround */ }'
+        )
 
         self._cmdline_files = picard_args.FILE
         self._autoupdate = autoupdate


picard/tagger.py
```diff
@@ -118,7 +118,14 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
         # Allow High DPI Support
         self.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps)
         self.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
-
+        super().setStyleSheet(
```
sambhav: Can't really test this as it doesn't affect me. Maybe @vishichoudhary can test and see if it fixes it for him too?

vishichoudhary: @zas @sambhav no effect of adding these two
```
 subcontrol-origin: margin;
 subcontrol-position: top left;
```
this too works for me 

```
 super().setStyleSheet(
             """ 
             QGroupBox::title { }
             """
) 
```

weird.

zas: @vishichoudhary : i changed the patch to use just this, it works for me too.

sambhav: Isn't this a bit too weird and unpredictable?

zas: Yes it is... but it works ... don't ask me why.

sambhav:
@zas the new setStyle causes other bugs. See the following (Picard.ini is same for both but the PR causes the saved sizes to not reflect).

https://i.imgur.com/GwPIGHE.png
https://i.imgur.com/aOpqwX9.png

zas:
> @zas the new setStyle causes other bugs. See the following (Picard.ini is same for both but the PR causes the saved sizes to not reflect).

What is the issue ? Size of the dialog box ?

zas:
> I found the fix a bit hacky.

That's clearly an hacky workaround, fix has to be done upstream (Qt).

PR #863, opened 2018-03-07
Manage preserved tags as a set, with the help of new class PreservedTags

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -154,12 +154,8 @@ class PreservedTags:
         config.setting[self.opt_name] =  ", ".join(sorted(self._tags))
 
     def _from_config(self):
-        return set(
-            [tag for tag in map(lambda x: x.strip(),
-                                config.setting[self.opt_name].split(','))
-                    if tag != ""
-            ]
-        )
+        tags = config.setting[self.opt_name].split(',')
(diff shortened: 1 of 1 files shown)


picard/ui/metadatabox.py
```diff
@@ -143,6 +143,36 @@ def tag_status(self, tag):
...
+
+    def __init__(self):
+        self._tags = self._from_config()
+
+    def _to_config(self):
+        config.setting[self.opt_name] =  ", ".join(sorted(self._tags))
+
+    def _from_config(self):
+        return set(
+            [tag for tag in map(lambda x: x.strip(),
```
sambhav: Not a really big fan of this indentation :P

sambhav: You can replace lambda exp. with `str.strip`

sambhav: filter(bool, map(str.strip, config.setting[self.opt_name].split(','))) is prolly what  you want?

zas: See 7b94c79affb06b1d01bf28dfb649963751cbad13

PR #871, opened 2018-03-09
PICARD-1213: Correct the order of movement in coverart providers

vishichoudhary:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When you move top element of coverart providers. It is swapping it's position with last element. This implies that the least prior provider has now become most prior provider. Similar for lower most element.

# JIRA ticket : [PICARD-1213](https://tickets.metabrainz.org/browse/PICARD-1213)

# Solution
Disable the up button for top row and disable the down button for lowest row.

### DIFF
--- a/picard/ui/options/cover.py
+++ b/picard/ui/options/cover.py
@@ -71,6 +71,7 @@ class CoverOptionsPage(OptionsPage):
                 title = provider.NAME
             checked = is_provider_enabled(provider.NAME)
             self.provider_list_widget.addItem(SortableCheckboxListItem(title, checked=checked, data=provider.NAME))
+        self.provider_list_widget.disable_last_down_btn()
 
         def update_providers_options(items):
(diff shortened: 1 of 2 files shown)


picard/ui/options/cover.py
```diff
@@ -71,6 +71,7 @@ def load_cover_art_providers(self):
                 title = provider.NAME
             checked = is_provider_enabled(provider.NAME)
             self.provider_list_widget.addItem(SortableCheckboxListItem(title, checked=checked, data=provider.NAME))
+        self.provider_list_widget.disable_last_down_btn()
```
zas: I'd prefer to have everything self-contained in SortableCheckboxListWidget

zas:
@vishichoudhary : i went further and did a cleanup of the whole code: https://github.com/metabrainz/picard/pull/872 please test & review

PR #876, opened 2018-03-11
PICARD-1045 Add update checking framework

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* Add a framework for checking program updates by scraping the latest release version information from the Picard home page.  Add manual check to toolstrip under the "Help" section.

# Problem

Picard does not include a system to check for program updates.
* JIRA ticket: [PICARD-1045](https://tickets.metabrainz.org/browse/PICARD-1045): Check for new version

# Solution

Add a framework for checking program updates by scraping the latest release version information from the Picard home page.  This **partially** addresses **PICARD-1045: Check for new version**

This framework includes a new module with two functions.  The first function gets the latest release version number by scraping the Picard home page.  The information is stored in a module-level variable for re-use, thus avoiding additional calls to the web site during the same Picard session.

The second function compares the latest release version number (calling the first function if necessary) with the version number of the currently running program.  If the latest release version is higher than the currently running version, a dialog box is displayed informing the user of the new version available and offering to open the Picard website to facilitate the download.  Normally, nothing is displayed if there is no new version available (to allow for scheduled automatic update checks as requested in PICARD-1045); however, there is an override option to force displaying a dialog box with the check results even if there is no update available (for use when an update check is manually requested by the user).

In addition, an item has been added to the toolstrip under the "Help" section to allow the user to manually request an update check.

# Action

None.  An option for automatic checking, as requested in PICARD-1045, will be covered in a future pull request.

### DIFF
--- a/picard/util/checkupdate.py
+++ b/picard/util/checkupdate.py
@@ -52,9 +52,9 @@ def get_latest_version_number():
             if matches:
                 latest_version = matches[0]
             else:
-                log.debug("Unable to get the latest version information from", PICARD_URLS['home'])
+                log.warning("Unable to get the latest version information from %s" % PICARD_URLS['home'])
         except Exception as e:
-            log.debug("Exception while getting the latest version information from %s: %s", (PICARD_URLS['home'], e))
+            log.error("Exception while getting the latest version information from %s: %s" % (PICARD_URLS['home'], e,))
     return latest_version
 
 
--- a/picard/util/checkupdate.py
+++ b/picard/util/checkupdate.py
@@ -1,0 +1,82 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+# Copyright (C) 2018 Bob Swift
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from picard import (PICARD_VERSION_STR_SHORT, log)
+from picard.const import PICARD_URLS
+import picard.util.webbrowser2 as wb2
+from PyQt5.QtWidgets import QMessageBox
+#from PyQt5 import QtCore, QtGui, QtWidgets
+#from PyQt5.QtGui import QIcon
+import urllib.request
+import re
+
+_RE_TEST = r'.*infoText\s*=\s*\"v([0-9\.]*)'
+
+_NOTICE_TEXT = '''
+A new version of Picard is available.
+
+Your version: v%s
+New version: v%s
+
+Would you like to download the new version?
+'''
+
+latest_version = ""
+''' Use a module-level variable to store the latest release information
+    to avoid multiple calls to the web site in a single Picard session. '''
+
+
+def get_latest_version_number():
+    '''Scrapes the Picard home page to extract the latest release version number.'''
+    global latest_version
+    if not latest_version:
+        try:
+            page = urllib.request.urlopen(PICARD_URLS['home'])
+            matches = re.findall(_RE_TEST, page.read().decode('utf-8'), re.M | re.I)
+            if matches:
+                latest_version = matches[0]
+            else:
+                log.debug("Unable to get the latest version information from", PICARD_URLS['home'])
+        except Exception as e:
+            log.debug("Exception while getting the latest version information from %s: %s", (PICARD_URLS['home'], e))
+    return latest_version
+
+
+def check_update(show_always=False):
+    '''Checks if an update is available.
+
+    Compares the version number of the currently running instance of Picard
+    and displays a dialog box informing the user  if an update is available,
+    with an option of opening the Picard site in the browser to download the
+    update.  If there is no update available, no dialog will be shown unless
+    the "show_always" parameter has been set to True.  This allows for silent
+    checking during startup if so configured.
+    '''
+    latest_version = get_latest_version_number()
+    msg_title = "Picard Update"
+    if latest_version > PICARD_VERSION_STR_SHORT:
+        msg_text = _NOTICE_TEXT % (PICARD_VERSION_STR_SHORT, latest_version)
+        if QMessageBox.information(None, msg_title, msg_text, QMessageBox.Ok | QMessageBox.Cancel,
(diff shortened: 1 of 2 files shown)


picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+    Compares the version number of the currently running instance of Picard
+    and displays a dialog box informing the user  if an update is available,
+    with an option of opening the Picard site in the browser to download the
+    update.  If there is no update available, no dialog will be shown unless
+    the "show_always" parameter has been set to True.  This allows for silent
+    checking during startup if so configured.
+    '''
+    latest_version = get_latest_version_number()
+    msg_title = "Picard Update"
+    if latest_version > PICARD_VERSION_STR_SHORT:
```
mineo: Please don't do ordered comparison on strings containing version data, it will not work:

```python
In [1]: "3.1" > "3.10"
Out[1]: False

In [2]: "3.2" > "3.10"
Out[2]: True
```

Unfortunately I don't know of any library that makes comparing version numbers in Python easy.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+import picard.util.webbrowser2 as wb2
+from PyQt5.QtWidgets import QMessageBox
+#from PyQt5 import QtCore, QtGui, QtWidgets
+#from PyQt5.QtGui import QIcon
+import urllib.request
+import re
+
+_RE_TEST = r'.*infoText\s*=\s*\"v([0-9\.]*)'
+
+_NOTICE_TEXT = '''
```
mineo: The value of this variable should be wrapped in `_()`, so it can be translated.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+''' Use a module-level variable to store the latest release information
+    to avoid multiple calls to the web site in a single Picard session. '''
+
+
+def get_latest_version_number():
+    '''Scrapes the Picard home page to extract the latest release version number.'''
+    global latest_version
+    if not latest_version:
+        try:
+            page = urllib.request.urlopen(PICARD_URLS['home'])
```
mineo: Please only use Picards `webservice` module for performing HTTP requests. It's the only way to ensure a not-so-fast request won't block the whole application.

picard/util/checkupdate.py
rdswift: I'll come up with something based on the version_to_string and version_from_string code in the picard/__init__.py file.

picard/util/checkupdate.py
rdswift: Will do.  Should I also do something similar with the strings for the error messages?

picard/util/checkupdate.py
rdswift: Will do.  Sorry, I should have thought of that myself.

rdswift:
I'm going to put the changes on hold pending the outcome of PW-62 (picard-website PR # 112).  That will likely result in a significant rewrite of this PR.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+    msg_title = "Picard Update"
+    if latest_version > PICARD_VERSION_STR_SHORT:
+        msg_text = _NOTICE_TEXT % (PICARD_VERSION_STR_SHORT, latest_version)
+        if QMessageBox.information(None, msg_title, msg_text, QMessageBox.Ok | QMessageBox.Cancel,
+                                   QMessageBox.Cancel) == QMessageBox.Ok:
+            wb2.goto("home")
+    else:
+        if show_always:
+            msg_text = "There is no update currently available.  The latest release is %s." % (
+                'v' + latest_version if latest_version else 'Unknown',)
```
zas: Buillding strings like this doesn't play well with i18n and `'Unknown'` isn't translated here.

You have to do something like:
```python
if latest_version:
  msg_text = _("There is no update currently available. The latest release is v%s.") % latest_version
else:
  msg_text = _("There is no update currently available. The latest release is unknown")
QMessageBox.information(None, msg_title, msg_text, QMessageBox.Ok, QMessageBox.Ok)
```

picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+    and displays a dialog box informing the user  if an update is available,
+    with an option of opening the Picard site in the browser to download the
+    update.  If there is no update available, no dialog will be shown unless
+    the "show_always" parameter has been set to True.  This allows for silent
+    checking during startup if so configured.
+    '''
+    latest_version = get_latest_version_number()
+    msg_title = "Picard Update"
+    if latest_version > PICARD_VERSION_STR_SHORT:
+        msg_text = _NOTICE_TEXT % (PICARD_VERSION_STR_SHORT, latest_version)
```
zas: If `_NOTICE_TEXT` is only used once, move the text here and get rid of this variable.
For translation, use `_("...your...string...")`

picard/util/checkupdate.py
```diff
@@ -0,0 +1,82 @@
...
+'''
+
+latest_version = ""
+''' Use a module-level variable to store the latest release information
+    to avoid multiple calls to the web site in a single Picard session. '''
+
+
+def get_latest_version_number():
+    '''Scrapes the Picard home page to extract the latest release version number.'''
+    global latest_version
```
zas: Is a global mandatory ?

picard/util/checkupdate.py
rdswift: Will do.  I'll also translate the error message text lines.

picard/util/checkupdate.py
rdswift: Will do.

picard/util/checkupdate.py
rdswift: I was using it to hold the value so that subsequent calls are not made to the web site.  I'm open to suggestions for an alternate way of doing this.  I don't particularly like using a global either.

rdswift:
It seems that not everyone on the Picard leadership team shares the same vision of the end product, which makes trying to provide code a challenging task.  I've tried to summarize the issues and proposal in the gist at [https://gist.github.com/rdswift/6d6651a0a6c5ae1b5d619f6faab32a9b](https://gist.github.com/rdswift/6d6651a0a6c5ae1b5d619f6faab32a9b) and would appreciate your comments and input.  Thanks.

zas:
> It seems that not everyone on the Picard leadership team shares the same vision of the end product, **which makes trying to provide code a challenging task.**

Commenting a gist is a challenging task too. Why don't you post this on forum or as google docs so it is easier to comment / cite ?
Also "There are JIRA tickets requesting that Picard provide some form of new version availablity checking." please link to actual tickets...

If you refer to recent "beta1" it is an external version i used as a marketing tool... internal version is 2.0.0dev4, since Picard doesn't actually support beta in the name.

Since we are in feature freeze stage, any change to version naming or handling is very unlikely to happen before 2.0.

But if you want to contribute, you may want to look at recent bug reports with high priority, about Picard not working at all, crashing, etc... that would be much more useful at this stage.

Please discuss https://tickets.metabrainz.org/browse/PICARD-1045 in issue comments, and attach documents there if any.

rdswift:
Will re-think the process and create a new PR after the 2.0 release.

PR #880, opened 2018-03-18
Clean-up un-used code and errors

sambhav:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -17,7 +17,7 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-import cgi
+import html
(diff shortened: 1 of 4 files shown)


picard/util/__init__.py
```diff
@@ -469,7 +469,7 @@ def convert_to_string(obj):
 
 
 def htmlescape(string):
-    return cgi.escape(string)
+    return html.escape(string)
```
zas: Note that `cgi.escape()` differs from `html.escape()`, `quotes` defaults to `True` instead of `False`.

See https://docs.python.org/3.5/library/cgi.html#cgi.escape and https://docs.python.org/3.5/library/html.html?highlight=html#html.escape

I don't think it has any negative impact, but did you check actual calls to `htmlescape()` ?

PR #882, opened 2018-03-18
PICARD-1234: Update calls to `install_plugin` after method signature update

sambhav:
PR #829 refactored some of the plugin code and updated the method signature of install_plugin but all the calls to it were not updated causing crashes on drag'n'drop / manual installation of plugins. This fixes that.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1234](https://tickets.metabrainz.org/browse/PICARD-1234)

# Solution

# Action

### DIFF
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -366,7 +366,7 @@ class PluginsOptionsPage(OptionsPage):
         )
         if files:
             for path in files:
-                self.tagger.pluginmanager.install_plugin(path)
+                self.tagger.pluginmanager.install_plugin(path, action=PLUGIN_ACTION_INSTALL)
 
     def download_plugin(self, action):
         selected = self.ui.plugins.selectedItems()[COLUMN_NAME]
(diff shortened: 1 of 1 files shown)


picard/ui/options/plugins.py
```diff
@@ -366,7 +366,7 @@ def open_plugins(self):
         )
         if files:
             for path in files:
-                self.tagger.pluginmanager.install_plugin(path)
+	            self.tagger.pluginmanager.install_plugin(path, action=PLUGIN_ACTION_INSTALL)
```
zas: Indentation is incorrect here

sambhav: Ah sorry, my sublime settings got reset. It ended up inserting a tab there instead of 4 spaces. Will fix.

sambhav: Done.

PR #890, opened 2018-03-26
Metabrainz issue PICARD-1249

wjyoung65:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Filter for None strings when passing a sequence to the join function.

# Problem
Picard crashed using the Lookup CD function with "TypeError: sequence item 0: expected str instance, NoneType found".  For this CD, looks like the Country field is not set. 

* JIRA ticket (_optional_): [PICARD-1249](https://tickets.metabrainz.org/browse/PICARD-1249)

# Solution
Instead of return "\n".join(l), filter the 'l' variable for None type strings: join(filter(None,l))

# Action
No other action aside from reviewing and merging the one-liner change. Apologies for naming my branch in all-caps.

### DIFF
--- a/picard/ui/cdlookup.py
+++ b/picard/ui/cdlookup.py
@@ -52,7 +52,7 @@ class CDLookupDialog(PicardDialog):
                                               _("Labels"), _("Catalog #s"), _("Barcode")])
         if self.releases:
             def myjoin(l):
-                return "\n".join(l)
+                return "\n".join(filter(None,l))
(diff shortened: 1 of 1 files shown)


sambhav:
This is already fixed in #888. Use the latest build for the fix.

PR #892, opened 2018-03-30
PICARD-1252: properly escape strings used to build a regexp

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

An exception is raised because of invalid regexp

# Problem

If argument contains '*' (or any other re reserved and unescaped character), an invalid regexp can be build, leading to an exception

* JIRA ticket : [PICARD-1252](https://tickets.metabrainz.org/browse/PICARD-1252)

# Solution

Just use `re.escape()` on each argument before using them.

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -789,7 +789,8 @@ def _delete_prefix(parser, text, *prefixes):
     if not prefixes:
         prefixes = ('A', 'The')
     text = text.strip()
-    match = re.match('(' + r'\s+)|('.join(prefixes) + r'\s+)', text)
+    rx = '(' + r'\s+)|('.join(map(re.escape, prefixes)) + r'\s+)'
+    match = re.match(rx, text)
     if match:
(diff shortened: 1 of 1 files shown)


picard/script.py
```diff
@@ -789,7 +789,8 @@ def _delete_prefix(parser, text, *prefixes):
     if not prefixes:
         prefixes = ('A', 'The')
     text = text.strip()
-    match = re.match('(' + r'\s+)|('.join(prefixes) + r'\s+)', text)
+    rx = '(' + r'\s+)|('.join([re.escape(p) for p in prefixes]) + r'\s+)'
```
sambhav: `map(re.escape, prefixes)` ?

zas: > map(re.escape, prefixes) ?

Yes, shorter.

PR #899, opened 2018-06-03
PICARD-1122: Preffered release type settings are exclusive and should be inclusive

StasisInterrupted:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
This pull request fixes the the preferred release type functionality

# Problem
Prior to this, picard only used the primary release type and ignored the secondary release type. This algorithm improves on that by incorporating the secondary release types and averaging the weights of all the types

* JIRA ticket (_optional_): [PICARD-1122](https://tickets.metabrainz.org/browse/PICARD-1122)

# Solution
The algorithm gathers all the release types found in the release (both primary and secondary) into a list, adds their respective scores (from the user's settings) together and then divides that score by the number of types there are in total, thus averaging all the scores.

It is important to note that the preffered release types only works with lookups, and not with acousticid fingerprinting. It would be an improvement for the preffered release types to work in both cases but to implement this, it must be determined why fingerprinting does not return release types data while tag based lookup does

# Action
The changes are pep-8 compliant and have been tested successfully. No further action should be necessary.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -188,12 +188,22 @@ class Metadata(dict):
             parts.append((score, weights["format"]))
 
         if "releasetype" in weights:
+            # This section generates a score that determines how likely this release will be selected in a lookup.
+            # The score goes from 0 to 1 with 1 being the most likely to be chosen and 0 the least likely
+            # This score is based on the preferences of release-types found in this release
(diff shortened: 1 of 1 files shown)


mineo:
I believe this has been replaced by #920

PR #901, opened 2018-07-03
PICARD-1277 (workaround): translate_caa_type: Return the input if the type is unknown

mineo:
# Summary

That can happen if a new cover art type is added in MB that Picard doesn't yet
know about. Instead of raising a KeyError (and crashing), just return the raw
type, which is good enough until we have something better to show.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

See above.

* JIRA ticket (_optional_): [PICARD-1277](https://tickets.metabrainz.org/browse/PICARD-1277)

# Solution
See above.

# Action

Backport this to 1.4.something? Otherwise all users on older releases (Debian, Ubuntu, ...) will not be able to load albums with new cover art types.

### DIFF
--- a/picard/coverart/utils.py
+++ b/picard/coverart/utils.py
@@ -39,4 +39,5 @@ def translate_caa_type(name):
     if name == 'unknown':
         return _(CAA_TYPES_TR[name])
     else:
-        return gettext_attr(CAA_TYPES_TR[name], "cover_art_type")
+        title = CAA_TYPES_TR.get(name, name)
+        return gettext_attr(title, "cover_art_type")
--- a/picard/const/attributes.py
+++ b/picard/const/attributes.py
@@ -16,6 +16,7 @@ MB_ATTRIBUTES = {
     'DB:cover_art_archive.art_type/name:011': 'Poster',
     'DB:cover_art_archive.art_type/name:012': 'Liner',
     'DB:cover_art_archive.art_type/name:013': 'Watermark',
+    'DB:cover_art_archive.art_type/name:014': 'Raw/Unedited',
     'DB:medium_format/name:001': 'CD',
     'DB:medium_format/name:002': 'DVD',
     'DB:medium_format/name:003': 'SACD',
--- a/test/test_coverart_utils.py
+++ b/test/test_coverart_utils.py
@@ -1,0 +1,22 @@
+#!/usr/bin/env python
+# coding: utf-8
+import os.path
+import shutil
+import tempfile
+import unittest
+
+from picard.coverart.utils import translate_caa_type
+from picard.i18n import setup_gettext
+
+
+class CaaTypeTranslationTest(unittest.TestCase):
+    def setUp(self):
+        # we are using temporary locales for tests
+        self.tmp_path = tempfile.mkdtemp()
+        self.localedir = os.path.join(self.tmp_path, 'locale')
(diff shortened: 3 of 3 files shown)


picard/coverart/utils.py
```diff
@@ -39,4 +39,5 @@ def translate_caa_type(name):
     if name == 'unknown':
         return _(CAA_TYPES_TR[name])
     else:
-        return gettext_attr(CAA_TYPES_TR[name], "cover_art_type")
+        title = CAA_TYPES_TR.get(name, name)
```
zas: I'd just use try/except for this:

```python
try:
  return gettext_attr(CAA_TYPES_TR[name], "cover_art_type")
except KeyError:
  return name
```

BTW, to get the missing type, just run `python setup update_constants`

mineo: AFAICT from `rg "\.get\(.*\)" -A20 | rg keyerror` there's no place where we're using the "access a dict, fallback to a default value through exceptions"  pattern anywhere else.

mineo: Err, sorry, that search pattern is obviously useless because we're most likely accessing trying to retrieve a key through the standard `dictobj[key]` syntax if we don't pass a default value. Still, there are not really any places I can see where we're using KeyErrors for falling back to default values (ignoring plugin.py).

zas: I'd prefer the exception, because this should be logged, as a warning imho.
Returning a default value should never happen in normal conditions (caa types are not supposed to change often).
As is, this PR doesn't solve the core issue, it just prevents an exception: `python setup update_constants` will address the core issue (caa type unkown to Picard, but available through CAA)

mineo: I disagree with logging this as a warning. There's nothing an ordinary user can do about this, they'll have to wait until a new Picard release is made.

In the meantime though, Picard should certainly not crash when encountering a cover art type that it doesn't yet know about, which is why this PR has "workaround" in the title. I absolutely do not intend to close PICARD-1277 after this change because it seems that any unexpected state in the cover art queuing code can crash Picard, which certainly should not happen.

phw: I agree with mineo here. New cover art types, while seldom, is something that can happen anytime. Picard needs to deal in a clean way with that which does not break the user's workflow. Just running update constants is not a solution for this.

Of course logging this as a warning could be done additionaly. In this case maybe still throwing an exception could make sense, but then it would be in the responsibility of the caller to handle the exception, log it as a warning and use the raw type as text. Given this is a small utility function which should be easy to use I think it's better to have this function handle all of this transparently.

zas: > I disagree with logging this as a warning. There's nothing an ordinary user can do about this, they'll have to wait until a new Picard release is made.

I don't think warning log messages would bother the "ordinary user" at all.
But for sure advanced users need to know what is going on, and eventually report the issue, so maintainers can update cover art types list in Picard (until we have a web service for this).

> New cover art types, while seldom, is something that can happen anytime. Picard needs to deal in a clean way with that which does not break the user's workflow. Just running update constants is not a solution for this.

I agree, i think you didn't understand me, both are needed: handling the missing-type case elegantly AND updating constants.

mineo: I've updated the attributes file.

PR #904, opened 2018-07-17
PICARD-1045: Add program update checking

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Adds checking for updated version using the GitHub api.**

# Problem

There is currently no system included to allow checking for available updates, and for automatically notifying the user when a new version of the software is available.

* JIRA ticket: [PICARD-1045](https://tickets.metabrainz.org/browse/PICARD-1045)

# Solution

Added update checking class, which calls the GitHub api to get the latest available versions (i.e.: stable and beta) for Picard and compares them to the version of the currently running instance.  If an update is available, the user will be shown a dialog box with an option to download the new version.

Added a manual call to the update checker under the ***Help*** section of the main toolbar.

Added two new checkboxes to the ***General*** options tab to enable automatic checking during program startup, and whether to also check for beta (dev) versions.  Automatic checks will be silent and not display check results if no update is available.

Added one new spinbox to the ***General*** options tab to select the number of days to wait between automatic update checks during program startup.

# Action

No additional action required.


samj1912:
I would like Picard Website to have an end point to control updates instead. I think you made a PR for it a while back. Why not reopen that?

@zas any thoughts on Gh vs PW for update checking?

zas:
> @zas any thoughts on Gh vs PW for update checking?

Using a PW endpoint would make things much more simpler, using GH leads to a lot of extra code.
So yes, i'd prefer the PW endpoint way too.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,159 @@
...
+            none.
+
+        Raises:
+            none.
+        '''
+        self._show_always = show_always
+        self._update_level = update_level
+
+        # Gets list of releases from GitHub website api.
+        output_text = _("Getting release information from GitHub.")
```
phw: This and below is exactly the code from `query_available_updates`, should call that instead of reimplementing it here,

phw:
> Using a PW endpoint would make things much more simpler, using GH leads to a lot of extra code.
> So yes, i'd prefer the PW endpoint way too.

I don't think that's much the case. Given the pull request the actual Github specific code to handle the github request and the response is something below 20 LOC in total. The rest is UI and general update handling + logging.

Our own implementation might be simpler, especially as we could give back an optimized result from PW (e.g. return only the latest version), so the code to handle the response would be simplified a bit. But on the other side there is additional code on PW.

I actually would go with the GH pages. Of course it would also be possible to revive https://github.com/metabrainz/picard-website/pull/112/files and adapt this PR to use that endpoint, I don't think that would require much code changes.

rdswift:
> I actually would go with the GH pages. Of course it would also be possible to revive https://github.com/metabrainz/picard-website/pull/112/files and adapt this PR to use that endpoint, I don't think that would require much code changes.

It's funny because the reason that I abandoned that PR was because of zas's comment about using GitHub instead.  I assumed that using the GitHub api would result in less maintenance work because the release information would only need to be maintained in one place (i.e.: GitHub because that's where the releases were being managed).  If the consensus is to use PW instead, I'll resurrect the other PR and make the necessary changes here.

picard/util/checkupdate.py
rdswift: You're right.  That was a left over from a previous version and was supposed to have been removed.  I'll make the change.

rdswift: Change made, but I ended up having to force push it.  You should be able to see the corrected version in the current changes.

zas:
Picard may not be hosted on Github in the future (Microsoft bought it...)
Adding a level of abstraction through the PW web service would ease things  a lot if we need to move to another hosting solution and prolly reduce the code needed inside the application.

rdswift:
@phw, The change have been made to remove the duplicated code as requested, but I ended up having to force push it. You should be able to see the corrected version in the current changes.

rdswift:
@zas / @samj1912, I've changed the update version information source from GitHub to picard-website as requested.  I've tested it locally based on the current code in [picard-website PR #119](https://github.com/metabrainz/picard-website/pull/119), but won't be able to do any production testing until the picard-website PH has been merged.  I thought I would post the updated code for this PR now so that we can get a head start on any reviews and requested changes, at your convenience.

phw:
Code is looking good to me. I would like to run the code though to see it in action. Will do so later this week

rdswift:
@phw, the picard-website api PR #119 has now been merged and moved to production, so you should be able to test the code now using the production server rather than a local test server if you like.

picard/ui/options/general.py
```diff
@@ -45,6 +45,10 @@ class GeneralOptionsPage(OptionsPage):
         config.TextOption("persist", "oauth_access_token", ""),
         config.IntOption("persist", "oauth_access_token_expires", 0),
         config.TextOption("persist", "oauth_username", ""),
+        config.BoolOption("setting", "check_for_updates", False),
```
samj1912: Maybe set this to true?

picard/ui/options/general.py
```diff
@@ -45,6 +45,10 @@ class GeneralOptionsPage(OptionsPage):
...
+        config.IntOption("setting", "update_check_days", 0),
```
samj1912: And this to 10?

picard/util/checkupdate.py
```diff
@@ -0,0 +1,159 @@
...
+    def _releases_json_loaded(self, response, reply, error, callback=None):
+        '''Processes response from specified website api query.'''
+        if error:
+            log.error(
+                N_("Error loading releases list: %(error)s"),
+                {'error': reply.errorString()},
+            )
+            if self._show_always:
+                msg_title = _("Picard Update")
+                msg_text = _("Unable to retrieve the latest version informarmation.")
```
zas: `informarmation` -> `information` ;)

picard/util/checkupdate.py
```diff
@@ -0,0 +1,159 @@
...
+            VERSIONS_API['host'],
+            VERSIONS_API['port'],
+            VERSIONS_API['endpoint'],
+            partial(self._releases_json_loaded, callback=callback),
+            parse_response_type=None,
+            priority=True,
+            important=True
+        )
+
+    def _releases_json_loaded(self, response, reply, error, callback=None):
```
zas: `callback` is unused here.
I think it should be called, if set, and error/success passed to it, so you can move `config.persist['last_update_check'] = datetime.date.today().toordinal()` out of this (dropping dependency on datetime module), and move it to the caller.

picard/ui/mainwindow.py
```diff
@@ -147,6 +147,9 @@ def keyPressEvent(self, event):
     def show(self):
         self.restoreWindowState()
         super().show()
+        if config.setting['check_for_updates'] and datetime.date.today().toordinal() >= config.persist['last_update_check'] + config.setting['update_check_days']:
```
zas: `datetime.date.today().toordinal() >= config.persist['last_update_check'] + config.setting['update_check_days']`

If `update_check_days` is 0, and `>=` is used then it will check more than once per day.

So i think it should be `>` + ensure `update_check_days` is >= 0
Or `>=` and ensure `update_check_days` is >= 1

rdswift: With the zero interval I had intended to allow multiple checks per day, but I see now that might not be appropriate or required in a production environment.  I think your second suggested option is better because it allows checks on successive days with the 'update_check_days' setting at 1, whereas the first option will only do checks on successive days if the setting is 0.  It all comes down to how you interpret the meaning of 'update_check_days'.  The interpretation when I was writing it was that a value of 1 meant to check every 1 day (i.e.: every day).  Following this logic, I also suggest that the lower limit in the settings be changed from 0 to 1 for the 'update_check_days' setting.

picard/ui/options/general.py
rdswift: I'll make the change.

picard/ui/options/general.py
rdswift: I'll set the initial value to 7.

picard/util/checkupdate.py
rdswift: Okay, I've made the change as requested.  If I've misunderstood, please let me know.  Thanks.

picard/util/checkupdate.py
rdswift: Oops.  Fat fingers...  :grin:

rdswift:
@zas / @samj1912 / @phw / @Sophist-UK 

I think I have now captured all of your requested changes based on my understanding of the discussion.  Please have a look at your convenience and let me know if I've misunderstood or missed something.  Thanks.

rdswift:
If / when this is accepted, I have also made changes to the documentation on the picard-website and will submit a PR at that time.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,159 @@
...
+        i = 0
+        for test_key in ['stable', 'beta', 'dev']:
+            if self._update_level >= i and  compare_version_tuples(high_version, self._available_versions[test_key]['version']) > 0:
+                key = test_key
+                high_version = self._available_versions[test_key]['version']
+            i += 1
+        if key:
+            msg_text = _("A new version of Picard is available.\n\nOld version: %s\nNew version: %s\n\n"
+                         "Would you like to download the new version?") % (PICARD_FANCY_VERSION_STR, self._available_versions[key]['tag'],)
+            if QMessageBox.information(None, msg_title, msg_text, QMessageBox.Ok | QMessageBox.Cancel,
```
phw: All the message boxes should specify the main window as parent. Otherwise the dialog might open completely independent of the main window, as it happened to me on my multi-monitor setup.

You can get the main window with `self.tagger.window`.

rdswift: Done, I think.  Since I wasn't importing tagger into the ***checkupdate*** module, I included the parent information in the call to the ***check_update()*** method call from the main window.  Is this an acceptable alternative?

phw:
> I included the parent information in the call to the check_update() method call from the main window. Is this an acceptable alternative?

@rdswift Two answers to this :)

1. You actually really could have used just `self.tagger.window`. It is automagically available since Picard monkey patches `QObject`, see [tagger.py#L172](https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L172). But actually this is not the greatest design decision in regards to coupling :)

2. I think your approach of explicitly passing the parent is better. But I would change it to pass the parent only once in the constructor, I think this makes the code cleaner instead of passing around the parent internally to all functions needing it. You also missed one QMessageBox in `_releases_json_loaded`

rdswift:
I never noticed that patch in tagger.py before.  I still think I'd rather explicitly pass the parent than rely on a patch.  Good point about passing it to the constructor.  I'll make the change.

picard/ui/mainwindow.py
```diff
@@ -147,6 +147,9 @@ def keyPressEvent(self, event):
     def show(self):
         self.restoreWindowState()
         super().show()
+        if config.setting['check_for_updates'] and config.setting['update_check_days'] > 0 and datetime.date.today().toordinal() >= config.persist['last_update_check'] + config.setting['update_check_days']:
+            log.debug(_("Initiating start-up check for program updates."))
```
zas: May be add info concerning conditions in this debug message, like last check date, etc...

picard/ui/options/general.py
```diff
@@ -61,12 +65,18 @@ def load(self):
...
+        self.ui.update_level.setCurrentIndex(config.setting["update_level"])
+        self.ui.update_check_days.setValue(config.setting["update_check_days"])
 
     def save(self):
         config.setting["server_host"] = self.ui.server_host.currentText().strip()
         config.setting["server_port"] = self.ui.server_port.value()
         config.setting["analyze_new_files"] = self.ui.analyze_new_files.isChecked()
         config.setting["ignore_file_mbids"] = self.ui.ignore_file_mbids.isChecked()
+        config.setting["check_for_updates"] = self.ui.check_for_updates.isChecked()
+        config.setting["update_level"] = self.ui.update_level.currentIndex()
```
zas: It is a bad idea to use the list index, because if we re-order the list it will change the meaning of the option, you need to have some indirection here: index -> value, and store that value as option value. `setData()` and UserRole would work (but you can't set those in qt designer).

picard/util/checkupdate.py
```diff
@@ -0,0 +1,160 @@
...
+        if error:
+            log.error(
+                N_("Error loading releases list: %(error)s"),
+                {'error': reply.errorString()},
+            )
+            if self._show_always:
+                msg_title = _("Picard Update")
+                msg_text = _("Unable to retrieve the latest version information.")
+                QMessageBox.information(
+                    self._parent, msg_title, msg_text, QMessageBox.Ok, QMessageBox.Ok)
```
zas: Those intermediate variables aren't strictly needed:

```python
QMessageBox.information(
    self._parent,
    _("Picard Update"),
    _("Unable to retrieve the latest version information."),
    QMessageBox.Ok, QMessageBox.Ok
)
```

picard/util/checkupdate.py
```diff
@@ -0,0 +1,160 @@
...
+        key = ''
+        high_version = PICARD_VERSION
+        i = 0
+        for test_key in ['stable', 'beta', 'dev']:
+            if self._update_level >= i and  compare_version_tuples(high_version, self._available_versions[test_key]['version']) > 0:
+                key = test_key
+                high_version = self._available_versions[test_key]['version']
+            i += 1
+        if key:
+            msg_text = _("A new version of Picard is available.\n\nOld version: %s\nNew version: %s\n\n"
```
zas: Keep title and message within `QMessageBox.information()` it helps translators to have the context.
Also, now we moved to Python3, use placeholders/format() and wrap lines according to embedded newlines:

```python
if QMessageBox.information(
    self._parent,
    _("Picard Update"),
    _("A new version of Picard is available.\n\n"
      "Old version: {picard_old_version}\n"
      "New version: {picard_new_version}\n\n"
      "Would you like to download the new version?").format(
         picard_old_version=PICARD_FANCY_VERSION_STR,
         picard_new_version=self._available_versions[key]['tag']
       )
) == QMessageBox.Ok:
    wb2.open(self._available_versions[key]['urls']['download'])
```

picard/util/checkupdate.py
```diff
@@ -0,0 +1,160 @@
...
+                         "Would you like to download the new version?") % (PICARD_FANCY_VERSION_STR, self._available_versions[key]['tag'],)
+            if QMessageBox.information(self._parent, msg_title, msg_text, QMessageBox.Ok | QMessageBox.Cancel,
+                                       QMessageBox.Cancel) == QMessageBox.Ok:
+                wb2.open(self._available_versions[key]['urls']['download'])
+        else:
+            if self._show_always:
+                msg_text = _("There is no update currently available.\n\nYour version: %s\nLatest version: %s"
+                             % (PICARD_FANCY_VERSION_STR, self._available_versions['dev']['tag'],))
```
zas: See my comment above

picard/ui/mainwindow.py
rdswift: Changed to log a line showing:
- whether the automatic check was initiated or skipped
- today's date
- date of last check
- checking interval
- update check subscription level

picard/ui/options/general.py
rdswift: I reworked this a lot and now refer everything back to a constant dictionary defined in the picard `__init__.py` file.  The selection information is now taken from `self.ui.update_level.currentData(QtCore.Qt.UserRole)`.

picard/util/checkupdate.py
rdswift: Intermediate variables removed.

picard/util/checkupdate.py
rdswift: Reformatted as requested.

picard/util/checkupdate.py
rdswift: Reformatted.

picard/__init__.py
```diff
@@ -24,12 +24,26 @@
...
-
 # optional build version
 # it should be in the form '<platform>_<YYMMDDHHMMSS>'
 # ie. win32_20140415091256
 PICARD_BUILD_VERSION_STR = ""
 
+PROGRAM_UPDATE_LEVELS = {
+    'stable': {
+        'level': 0,
+        'title': 'Stable releases only',
```
zas: You need to mark strings for translation using `N_()` here

picard/__init__.py
```diff
@@ -24,12 +24,26 @@
 PICARD_APP_NAME = "Picard"
 PICARD_VERSION = (2, 0, 4, 'dev', 1)
 
-
 # optional build version
 # it should be in the form '<platform>_<YYMMDDHHMMSS>'
 # ie. win32_20140415091256
 PICARD_BUILD_VERSION_STR = ""
 
+PROGRAM_UPDATE_LEVELS = {
```
zas: `picard.const` would be a better place imho.

picard/util/checkupdate.py
```diff
@@ -102,8 +102,7 @@ def available_versions(self):
 
     def _query_available_updates(self, callback=None):
         '''Gets list of releases from specified website api.'''
-        output_text = _("Getting release information from %s." % (VERSIONS_API['host'],))
-        log.debug(output_text)
+        log.debug(_("Getting release information from {host_url}.".format(host_url=VERSIONS_API['host'],)))
```
zas: We don't usually translate debug messages (that would be extra work for translators).
And there's an error here:

```python
log.debug(_("Getting release information from {host_url}.".format(host_url=VERSIONS_API['host'],)))
```

Should be:

```python
log.debug(_("Getting release information from {host_url}.").format(host_url=VERSIONS_API['host']))
```

`_()` returns the translated string for `format()`, and no need of trailing comma (at least not on a single line)

picard/util/checkupdate.py
```diff
@@ -0,0 +1,172 @@
...
+            partial(self._releases_json_loaded, callback=callback),
+            parse_response_type=None,
+            priority=True,
+            important=True
+        )
+
+    def _releases_json_loaded(self, response, reply, error, callback=None):
+        '''Processes response from specified website api query.'''
+        if error:
+            log.error(N_("Error loading releases list: {error_message}".format(error_message=reply.errorString(),)))
```
zas: Again, `N_("Error loading releases list: {error_message}").format(...)`

But `N_()` is useless in this case, either `_()` or `t = N_("string")` then `display(_(t))`

picard/util/checkupdate.py
```diff
@@ -0,0 +1,172 @@
...
+                    version_key=key, version_information=self._available_versions[key],))
+            self._display_results()
+        if callback:
+            callback(not error)
+
+    def _display_results(self):
+        # Display results to user.
+        key = ''
+        high_version = PICARD_VERSION
+        for test_key in PROGRAM_UPDATE_LEVELS.keys():
```
zas: `.keys()` is not needed, as for/in with dict iterates over keys by default.

picard/tagger.py
```diff
@@ -97,6 +97,8 @@
 from picard.ui.searchdialog.artist import ArtistSearchDialog
 from picard.ui.searchdialog.track import TrackSearchDialog
 
+from picard.util.checkupdate import UpdateCheckManager
```
zas: Bad place, see `.isort.cfg`, and use `isort` to sort this out ;)

picard/ui/mainwindow.py
```diff
@@ -1087,3 +1107,11 @@ def paste(self):
             target = selected_objects[0]
         self.tagger.paste_files(target)
         self.paste_action.setEnabled(False)
+
+    def check_for_update(self):
+        self.tagger.updatecheckmanager.check_update(show_always=True, update_level=config.setting["update_level"], callback=update_last_check_date)
```
zas: Wrap long lines in this case:

```python
self.tagger.updatecheckmanager.check_update(
    show_always=True,
    update_level=config.setting["update_level"],
    callback=update_last_check_date
)
```

picard/ui/mainwindow.py
```diff
@@ -1087,3 +1107,11 @@ def paste(self):
...
+
+def update_last_check_date(is_success):
+    log.debug('Processing update_last_check_date callback with argument: %s' % (is_success,))
```
zas: I'd remove this `log.debug()` in final version of the PR, not very useful imho.
Instead be sure errors are logged in case of non-success ;)

picard/__init__.py
rdswift: Will do.  I need to do some reading so I understand the translation stuff better.

picard/__init__.py
rdswift: I was actually going to ask you about that.  I'll move it.

picard/ui/mainwindow.py
rdswift: Will do, but I have a question regarding log entries.  Should they all be translated?  The reason I ask is because I see them translated in some places and not in others in the existing code.  I'd like to do it properly in any new code or changes that I make.

zas: Well, the thing is that error/warning messages should be translated because they are targetted to the user mainly.
For debug messages, it's not that great, because we'll end with debug logs (provided by users having problems) with messages in other languages. We perhaps need to rethink the whole thing, especially now we have a nice log dialog, with an option to save an english-only log for sharing and still have translated messages for users.
Another concern is the fact it will be hard to translate debug messages for translators without a dev background, and those messages change often (as they are used mainly by devs).

So, for now, don't mark or translate debug levell messages, but do for anything not debug.
Which is more or less what we do since years.

I'd prefer to have more languages supported, so keeping the number of strings to translate to acceptable level is much better regarding this goal.

picard/__init__.py
zas: Ask me if you need, i18n is a rather complex matter.

rdswift: Thanks.  I'll try to do the research on my own first, so as not to waste your time.  That way when I do have a question it should be a short clarification only, and not require you to provide a long explanation.  You have much more important things to do with your time than spending it tutoring me.

picard/ui/mainwindow.py
```diff
@@ -174,6 +176,19 @@ def keyPressEvent(self, event):
     def show(self):
         self.restoreWindowState()
         super().show()
+        for key in PROGRAM_UPDATE_LEVELS:
+            if PROGRAM_UPDATE_LEVELS[key]['level'] == config.setting["update_level"]:
```
zas: Just a minor thing, use single quotes around `update_level` for consistency

picard/ui/mainwindow.py
```diff
@@ -174,6 +176,19 @@ def keyPressEvent(self, event):
...
+        do_auto_update_check = config.setting['check_for_updates'] and config.setting['update_check_days'] > 0 and datetime.date.today().toordinal() >= config.persist['last_update_check'] + config.setting['update_check_days']
+        log.debug(_("{check_status} start-up check for program updates.  Today: {today_date}, Last check: {last_check} (Check interval: {check_interval} days), Update level: {update_level}").format(
+            check_status=_('Initiating') if do_auto_update_check else _('Skipping'),
+            today_date=datetime.date.today(),
+            last_check=str(datetime.date.fromordinal(config.persist['last_update_check'])) if config.persist['last_update_check'] > 0 else 'never',
+            check_interval=config.setting['update_check_days'],
+            update_level=update_level_text if update_level_text else _('unknown'),
+            ))
+        if do_auto_update_check:
+            self.tagger.updatecheckmanager.check_update(show_always=False, update_level=config.setting["update_level"], callback=update_last_check_date)
```
zas: `'update_level'`

picard/util/checkupdate.py
```diff
@@ -0,0 +1,181 @@
...
+    compare_version_tuples,
+    load_json,
+)
+import picard.util.webbrowser2 as wb2
+
+# Used to strip leading and trailing text from version string.
+_RE_CLEAN_VERSION = re.compile('^[^0-9]*(.*)[^0-9]*$', re.IGNORECASE)
+
+
+# GitHub API information
```
zas: Not GitHub anymore

picard/util/checkupdate.py
```diff
@@ -0,0 +1,181 @@
...
+VERSIONS_API = {
+    'host': 'picard.musicbrainz.org',
+    'port': 443,
+    'endpoint': '/api/releases'
+}
```
zas: I'd move it to `picard.const` and we'll have to rename and merge with https://github.com/metabrainz/picard/blob/master/picard/const/__init__.py#L105-L111

picard/util/checkupdate.py
```diff
@@ -0,0 +1,181 @@
...
+            self._display_results()
+        if callback:
+            callback(not error)
+
+    def _display_results(self):
+        # Display results to user.
+        key = ''
+        high_version = PICARD_VERSION
+        for test_key in PROGRAM_UPDATE_LEVELS:
+            if self._update_level >= PROGRAM_UPDATE_LEVELS[test_key]['level'] and  compare_version_tuples(high_version, self._available_versions[test_key]['version']) > 0:
```
zas: Two spaces instead of one just before `compare_version_tuples` and i'd wrap it (with parenthesis) since it's quite long plus a bit of variable caching:

```python
test_version = self._available_versions[test_key]['version']
if (self._update_level >= PROGRAM_UPDATE_LEVELS[test_key]['level']
    and compare_version_tuples(high_version, test_version):
        key = test_key
        high_version = test_version
```

picard/util/checkupdate.py
```diff
@@ -0,0 +1,181 @@
...
+                      picard_old_version=PICARD_FANCY_VERSION_STR,
+                      picard_new_version=self._available_versions[key]['tag']
+                    ),
+                QMessageBox.Ok | QMessageBox.Cancel,
+                QMessageBox.Cancel
+            ) == QMessageBox.Ok:
+                wb2.open(self._available_versions[key]['urls']['download'])
+        else:
+            if self._show_always:
+                for key in PROGRAM_UPDATE_LEVELS.keys():
```
zas: `.keys()` not needed here.
But you could write:

```python
for key, update in PROGRAM_UPDATE_LEVELS.items():
    if self._update_level == update['level']:
        update_level_text = update['title']
```

**But i have the feeling you could simplify code even more if you modify `PROGRAM_UPDATE_LEVELS` to have numeric level as key:**

```python
PROGRAM_UPDATE_LEVELS = {
    0: {
        'name': 'stable',
        'title': N_('Stable releases only'),
    },
   ...
```

then this loop would be uneeded:

```python
update_level_text = PROGRAM_UPDATE_LEVELS[self._update_level]['title']
```

picard/ui/mainwindow.py
rdswift: Will do.  I have to go out to run some errands right now, but will fix that (along with the log translations) when I get back.

picard/util/checkupdate.py
rdswift: Argh!  I had changed that once, but somehow it seems that must have been reverted during one of my git sessions.  Will fix.  :slightly_smiling_face:

picard/util/checkupdate.py
rdswift: Okay, that makes a lot of sense.

picard/util/checkupdate.py
rdswift: I was toying with having numeric keys originally, but don't recall why I ended up not doing it.  I'll look at it again because it seems to make much more sense with the numeric keys.

picard/util/checkupdate.py
rdswift: Good point on the caching and such.  I think I'll still need to check if the return value from `compare_version_tuples()` > 0 because the function could return -1 which (I think) would be evaluated as True, and I only want to continue processing if the value is 1.

picard/ui/options/general.py
```diff
@@ -76,14 +76,13 @@ def load(self):
         self.ui.ignore_file_mbids.setChecked(config.setting["ignore_file_mbids"])
         self.ui.check_for_updates.setChecked(config.setting["check_for_updates"])
         self.ui.update_level.clear()
-        index = 0
-        for key in PROGRAM_UPDATE_LEVELS:
+        for index, key in enumerate(PROGRAM_UPDATE_LEVELS):
```
zas: Instead of testing for current level in the loop, i suggest to do it outside, and use dict features:

```python
        self.ui.update_level.clear()
        for level, description in PROGRAM_UPDATE_LEVELS.items():
            self.ui.update_level.addItem(_(description['title']), level)
        self.ui.update_level.setCurrentIndex(self.ui.update_level.findData(config.setting["update_level"]))
```

Note: since PROGRAM_UPDATE_LEVELS is a dict, it is not sure it will be ordered as written, you'd better use an ordered dict or use sorting (the former is simpler imho).

rdswift: Switched to use an OrderedDict.  That simplifies it a bit.  I also agree that it's much cleaner doing the test outside the loop as you suggested.

ui/options_general.ui
```diff
@@ -132,6 +132,112 @@
...
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+          <property name="editable">
+           <bool>false</bool>
+          </property>
+          <property name="currentIndex">
+           <number>0</number>
+          </property>
+          <item>
```
zas: Since items are added through code (from dict), better remove them from this file

picard/ui/mainwindow.py
```diff
@@ -535,6 +538,9 @@ def create_actions(self):
         self.open_folder_action.setEnabled(False)
         self.open_folder_action.triggered.connect(self.open_folder)
 
+        self.check_update_action = QtWidgets.QAction(_("&Check for Update"), self)
+        self.check_update_action.triggered.connect(partial(self.check_for_update, show_always=True))
```
zas: Since `show_always=True` by default, no need of `partial` here

rdswift: For some strange reason if I remove the `partial`, a `show_always` value of `False` is being passed to the function.

zas: > For some strange reason if I remove the partial, a show_always value of False is being passed to the function.

Nothing strange, i just forgot it was a pyqt slot, and those aren't supposed to have default argument.

Basically you should have a method:

```python
self.check_update_action.triggered.connect(self.do_check_update)

def do_check_update(self):
  self.check_for_update(show_always=True)
```

So, basically, `partial()` does the same, my bad.

Though i think you could make the `check_for_update()` argument mandatory, it is called in 2 places and argument is explicitely set each time.

rdswift: Right.  I should have seen that.  Changes made as suggested, including making the `check_for_update()` argument mandatory.  Thanks.

ui/options_general.ui
```diff
@@ -213,24 +213,6 @@
...
-           <property name="text">
-            <string>Stable and Beta releases</string>
-           </property>
-          </item>
-          <item>
-           <property name="text">
-            <string>Stable, Beta and Dev releases</string>
-           </property>
-          </item>
          </widget>
```
zas: Re-run `python setup.py build_ui` and commit changes to generated files.

rdswift: Sorry, I forgot all about that.  Done now.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,177 @@
...
+            # Release information already acquired from specified website api.
+            self._display_results()
+        else:
+            # Gets list of releases from specified website api.
+            self._query_available_updates(callback=callback)
+
+    @property
+    def available_versions(self):
+        '''Provide a list of the latest version tuples for each update type.'''
+        return self._available_versions
```
zas: Is this property used somewhere ?
Why not rename `self._available_versions` to `self.available_versions` instead ?

picard/util/checkupdate.py
```diff
@@ -0,0 +1,177 @@
...
+            test_version = self._available_versions[PROGRAM_UPDATE_LEVELS[test_key]['name']]['version']
+            if self._update_level >= test_key and  compare_version_tuples(high_version, test_version) > 0:
+                key = PROGRAM_UPDATE_LEVELS[test_key]['name']
+                high_version = test_version
+        if key:
+            if QMessageBox.information(
+                self._parent,
+                _("Picard Update"),
+                _("A new version of Picard is available.\n\n"
+                  "Old version: {picard_old_version}\n"
```
zas: s/Old version/Current version/?

picard/util/checkupdate.py
```diff
@@ -0,0 +1,177 @@
...
+                  "New version: {picard_new_version}\n\n"
+                  "Would you like to download the new version?").format(
```
zas: I'm a bit bothered by this, because if Picard was installed through a package manager it is a bad idea to suggest the user to download it.
I wonder if we should provide an easy way for packagers to disable this feature...

What do you think guys?

picard/util/checkupdate.py
```diff
@@ -0,0 +1,177 @@
...
+)
+from picard.const import (
+    PLUGINS_API,
+    PROGRAM_UPDATE_LEVELS,
+)
+from picard.util import (
+    compare_version_tuples,
+    load_json,
+)
+import picard.util.webbrowser2 as wb2
```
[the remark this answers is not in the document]
rdswift: You're right.  The way I had it doesn't make any sense at all.  The hazards of copy 'n' paste from another module.

picard/util/checkupdate.py
```diff
@@ -0,0 +1,177 @@
...
+        # Each of these keys contains a dictionary with the keys: 'tag' (string),
+        # 'version' (tuple) and 'urls' (dictionary).  The 'version' tuple comprises
+        # major (int), minor (int), micro (int), type (str) and development (int)
+        # as defined in PEP-440.  The Picard developers have standardized on using
+        # only 'dev' or 'final' as the str_type segment of the version tuple.  Each
+        # key in the 'urls' dictionary contains a string with the specified url.
+        # Valid keys include: 'download' and 'changelog'.  The only required key in
+        # the 'urls' dictionary is 'download'.
+
+        # Initialize empty dictionary for 'stable' key
```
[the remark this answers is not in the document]
rdswift: Not required at all.  Will set it to `None` and change the test as suggested.

picard/util/checkupdate.py
rdswift: Actually, it's not used anywhere, so I'll just remove it.

picard/util/checkupdate.py
rdswift: How about 'Your version' or 'This version' instead?  When I hear the term 'Current version' I automatically think of the latest version.

picard/util/checkupdate.py
rdswift: We could change this to something like, "Visit the website for more information?", and just link to `https://picard.musicbrainz.org/" or "https://picard.musicbrainz.org/changelog/" for stable releases (set in the PW configuration).

Another option is to leave out the link option entirely, and just have this as a notification only.  If the user wants to update, they'll find a way.

zas: Linking to the Changelog with some `View recent changes` title could do.

phw: I actually would leave it as is for the case that we show update information. As a user I expect the software that has built-in update checks to actually direct me to the download.

But @zas' concern about the need for packagers to disable this is valid. As a packager I ideally would want to have an option when running the install via setup.py to completely disable this feature (no visible option, no update checking performed).

But crippling the displayed message doesn't help both cases: Distribution users still get bothered with update information even though they depend on their distribution to update the packages and have no easy way to update themselves, and users who installed manually get an update information where they then have to hunt for the download page.

phw: Having said this IMHO we should probably merge this PR and open a separate PR for disabling the update check. I am also not really sure how to technically best solve this.

rdswift: I'm not very familiar with what happens when installing from a packager, but if there is (or could be) a specific file (e.g.: PACKAGER.txt) added to the program directory, we should be able to check for that and disable the update checking and hide the options in the toolbar and settings.  Would that work?

I've already coded the changes based on Picard finding a `PACKAGER.txt` file, and can include them here if you like.  Then the other PR could just sort out the `INSTALLED_BY_PACKAGER` test if we wanted something different.

samj1912:
The branch looks good to me, just 1 question/doubt of sorts. Do we need to add versioning to the releases API?
Currently it is /api/releases, should it be /api/v1/releases?

phw:
> But i agree, /api/v2/releases would be safer.

Why start with v2?

phw:
Ah, I think I see, you consider current `/api/` v1, right? Then I don't really see the point in now starting with `v2` since this API update does not provide any breaking change, it is a rather clean addition to the existing API. You can still introduce v2 if something actually requires an incompatible change. But appart from that I don't really care :)

zas:
> Why start with v2?

It is just because we re-use API made for plugins, and last iteration is v2 ... but the main issue here is this re-use. Perhaps we need to think about this stuff, especially because plugins stuff is likely to change (again) soon...

rdswift:
Are we talking about the releases endpoint on the PW api?  At one point I had it set to both `api/v1/releases` and `api/v2/releases` and it was decided to remove the api version number and just go with `api/releases`.  It's a simple change, however you decide to go.

zas:
> Are we talking about the releases endpoint on the PW api? At one point I had it set to both api/v1/releases and api/v2/releases and it was decided to remove the api version number and just go with api/releases. It's a simple change, however you decide to go.

@samj1912 @phw : Re-add `api/v1/releases` and `api/v2/releases` ?

phw:
@zas I vote for yes. At first I had though the plugin API would be versionless and we would only introduce this version for the releases API, but now I see we have both v1 and v2 for the plugin API. It just makes sense to have this in from the start, because it is not unlikely we will have changes to the structure.

samj1912:
Do we need v1/releases?

rdswift:
In my opinion, it should start with `v2/releases` because that's when it was first implemented (even though it could be included as `v1/releases` without breaking the current `v1` api).  I also agree that it should be versioned.

zas:
> Do we need v1/releases?

I don't think so. Plus, apart Picard, nothing will use those endpoints for a while.
@rdswift : can you re-introduce `v2/releases` endpoint and update this PR accordingly ? That's the sole thing remaining before a merge.

PR #910, opened 2018-07-26
PICARD-1300: Use a lock to ensure queue is not mutated while iteration

samj1912:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1300](https://tickets.metabrainz.org/browse/PICARD-1300)

# Solution

# Action

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -22,7 +22,7 @@ import os
 
 from collections import deque, namedtuple, OrderedDict
 from PyQt5 import QtCore
-
+from threading import Lock
 
 _MAX_TAIL_LEN = 10**6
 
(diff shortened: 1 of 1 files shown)


picard/log.py
```diff
@@ -55,20 +55,22 @@ def debug_mode(enabled):
...
         self.pos = 0
 
     def emit(self, record):
-        self.log_queue.append(
-            TailLogTuple(
-                self.pos,
-                self.format(record),
-                record.levelno
+        with self.log_queue_lock:
+            self.log_queue.append(
```
mineo: It's not quite clear to me why this is necessary. According to [the documentation](https://docs.python.org/3.6/library/collections.html#collections.deque), " Deques support thread-safe, memory efficient appends and pops from either side of the deque". Why is it required to protect `append` with a lock?

picard/log.py
```diff
@@ -55,20 +55,22 @@ def debug_mode(enabled):
...
+                TailLogTuple(
+                    self.pos,
+                    self.format(record),
+                    record.levelno
+                )
             )
-        )
         self.pos += 1
```
mineo: Does this line need to be protected by a lock as well, since `+=` is not atomic in Python, AFAIR.

picard/log.py
samj1912: The main problem was https://community.metabrainz.org/t/debug-logging-in-picard-2-0/391297/5?u=samj1912

Basically if the deque is changed while iterating, which happens when there are too many log events.

mineo: Ah, makes sense.

I think a crash like that deserves a JIRA ticket, though.

PR #914, opened 2018-07-30
PICARD-1273: Add an option to exclude new cover art type "Raw / Unedited"

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: This change provides the user with the option of excluding selected cover art types from downloading from CAA.

# Problem

[STYLE-980](https://tickets.metabrainz.org/browse/STYLE-980) has added a new cover art style "Raw / Unedited".  Although beneficial for identifying and allowing better images for archival purposes, the images within this type are generally not suitable for tagging purposes.  Picard currently allows the user to select the types of images to download from CAA, but does not provide the ability to exclude selected image types from being downloaded.

* JIRA ticket: [PICARD-1273](https://tickets.metabrainz.org/browse/PICARD-1273): Add an option to exclude new cover art type "Raw / Unedited"

# Solution

This change adds new user settings to indicate whether selected image types should be excluded from downloads from CAA and the list of image types to be excluded, similar to the option to only include selected image types.  Additional logic is applied to the list of images selected for download, and any images with a type intersecting the user's "exclusion" type list are removed from the download list.

Generally speaking, the current logic for determining whether to download an image is (where the image is downloaded when "types" evaluates to True):

    if self.restrict_types:
        # only keep enabled caa types
        types = set(image["types"]).intersection(
            set(self.caa_types))
    else:
        types = True

This change adds a further check for exclusion, as:

    if self.restrict_types:
        # only keep enabled caa types
        types = set(image["types"]).intersection(
            set(self.caa_types))
    else:
        types = True
    if types and self.omit_types:
        # omit selected caa types
        types = not set(image["types"]).intersection(
            set(self.caa_types_to_omit))

# Action

Currently, it appears that CAA does not return the new "Raw/Unedited" type in the list of image types associated with an image.  This is described (with an example) in [CAA-113](https://tickets.metabrainz.org/browse/CAA-113): CAA does not return "Raw/Unedited" in image type array.  This will need to be addressed before the proposed changes will work.

### DIFF
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -29,7 +29,6 @@ from collections import (
 from functools import partial
 
 from PyQt5 import (
-    Qt,
     QtCore,
     QtWidgets,
 )
@@ -199,6 +198,10 @@ class ListBox(QtWidgets.QListWidget):
         if callback:
             callback()
 
+    def all_items_data(self, role=QtCore.Qt.UserRole):
+        for index in range(self.count()):
+            yield self.item(index).data(role)
+
 
 class CAATypesSelectorDialog(QtWidgets.QDialog):
     """Display dialog box to select the CAA image types to include and exclude from download and use.
@@ -243,43 +246,31 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
         gridlayout = QtWidgets.QGridLayout()
         grid.setLayout(gridlayout)
 
-        row = 0
-        column = 0
-        item = QtWidgets.QLabel()
-        item.setText(_("Include types list"))
-        gridlayout.addWidget(item, row, column)
-
-        column = 4
-        item = QtWidgets.QLabel()
-        item.setText(_("Exclude types list"))
-        gridlayout.addWidget(item, row, column)
-
-        row = 1
-        column = 0
-        gridlayout.addWidget(self.list_include, row, column)
-
-        column = 1
         self.arrows_include = ArrowsColumn(
             self.list_include,
             self.list_ignore,
             callback=self.set_buttons_enabled_state,
         )
-        gridlayout.addWidget(self.arrows_include, row, column)
 
-        column = 2
-        gridlayout.addWidget(self.list_ignore, row, column)
-
-        column = 3
         self.arrows_exclude = ArrowsColumn(
             self.list_exclude,
             self.list_ignore,
             callback=self.set_buttons_enabled_state,
             reverse=True
         )
-        gridlayout.addWidget(self.arrows_exclude, row, column)
 
-        column = 4
-        gridlayout.addWidget(self.list_exclude, row, column)
+        def add_widget(row=0, column=0, widget=None):
+             gridlayout.addWidget(widget, row, column)
+
+        add_widget(row=0, column=0, widget=QtWidgets.QLabel(_("Include types list")))
+        add_widget(row=1, column=0, widget=self.list_include)
+
+        add_widget(row=1, column=1, widget=self.arrows_include)
+        add_widget(row=1, column=2, widget=self.list_ignore)
+        add_widget(row=1, column=3, widget=self.arrows_exclude)
+
+        add_widget(row=0, column=4, widget=QtWidgets.QLabel(_("Exclude types list")))
+        add_widget(row=1, column=4, widget=self.list_exclude)
 
         self.layout.addWidget(grid)
 
@@ -358,7 +349,7 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
         for caa_type in CAA_TYPES:
             name = caa_type['name']
             title = translate_caa_type(caa_type['title'])
-            item = Qt.QListWidgetItem(title)
+            item = QtWidgets.QListWidgetItem(title)
             item.setData(QtCore.Qt.UserRole, name)
             if name in includes:
                 self.list_include.addItem(item)
@@ -371,10 +362,10 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
         webbrowser2.goto('doc_cover_art_types')
 
     def get_selected_types_include(self):
-        return [self.list_include.item(index).data(QtCore.Qt.UserRole) for index in range(self.list_include.count())] or ['front']
+        return list(self.list_include.all_items_data()) or ['front']
 
     def get_selected_types_exclude(self):
-        return [self.list_exclude.item(index).data(QtCore.Qt.UserRole) for index in range(self.list_exclude.count())] or ['none']
+        return list(self.list_exclude.all_items_data()) or ['none']
 
     def clear_focus(self, lists):
         for temp_list in lists:
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -95,11 +95,20 @@ def caa_url_fallback_list(desired_size, thumbnails):
 
 
 class CAATypesSelectorDialog(QtWidgets.QDialog):
-    _columns = 4
+    _columns = 3        # Number of main display columns
+    _subcolumns = 3     # Number of columns comprising a main display column (i.e.: label, combobox, spacer)
+    _spacer_width = 35  # Width of the spacer columns in pixels
 
-    def __init__(self, parent=None, types=None):
-        if types is None:
-            types = []
+    # Combo box index definitions
+    IDX_INCLUDE_TYPE = 0
+    IDX_IGNORE_TYPE = 1
+    IDX_EXCLUDE_TYPE = 2
+
+    def __init__(self, parent=None, types_include=None, types_exclude=None):
+        if types_include is None:
+            types_include = []
+        if types_exclude is None:
+            types_exclude = []
         super().__init__(parent)
 
         self.setWindowTitle(_("Cover art types"))
@@ -108,15 +117,29 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
 
         grid = QtWidgets.QWidget()
         gridlayout = QtWidgets.QGridLayout()
+        # Set up spacer columns between type selection columns
+        spacer_column_index = self._subcolumns - 1  # index of first spacer column
+        while spacer_column_index < self._columns * self._subcolumns - 1:
+            gridlayout.setColumnMinimumWidth(spacer_column_index, self._spacer_width)
+            spacer_column_index += self._subcolumns
         grid.setLayout(gridlayout)
 
         for index, caa_type in enumerate(CAA_TYPES):
             row = index // self._columns
-            column = index % self._columns
+            column = (index % self._columns) * self._subcolumns
             name = caa_type["name"]
-            text = translate_caa_type(name)
-            item = QtWidgets.QCheckBox(text)
-            item.setChecked(name in types)
+            item = QtWidgets.QLabel()
+            item.setText(translate_caa_type(name))
+            gridlayout.addWidget(item, row, column)
+            column += 1
+            item = QtWidgets.QComboBox()
+            item.addItems([ _('Include'), _('Ignore'), _('Exclude'),])
+            if name in types_include:
+                item.setCurrentIndex(self.IDX_INCLUDE_TYPE) # Include coverart type
+            elif name in types_exclude:
+                item.setCurrentIndex(self.IDX_EXCLUDE_TYPE) # Exclude coverart type
+            else:
+                item.setCurrentIndex(self.IDX_IGNORE_TYPE)  # Ignore coverart type
             self._items[item] = caa_type
             gridlayout.addWidget(item, row, column)
 
@@ -132,8 +155,9 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
             StandardButton(StandardButton.HELP), QtWidgets.QDialogButtonBox.HelpRole)
 
         extrabuttons = [
-            (N_("Chec&k all"), self.checkall),
-            (N_("&Uncheck all"), self.uncheckall),
+            (N_("I&nclude all"), self.checkall),
+            (N_("E&xclude all"), self.uncheckall),
+            (N_("I&gnore all"), self.ignoreall),
         ]
         for label, callback in extrabuttons:
             button = QtWidgets.QPushButton(_(label))
@@ -152,26 +176,35 @@ class CAATypesSelectorDialog(QtWidgets.QDialog):
         webbrowser2.goto('doc_cover_art_types')
 
     def uncheckall(self):
-        self._set_checked_all(False)
+        self._set_checked_all(2)
 
     def checkall(self):
-        self._set_checked_all(True)
+        self._set_checked_all(0)
+
+    def ignoreall(self):
+        self._set_checked_all(1)
 
     def _set_checked_all(self, value):
         for item in self._items.keys():
-            item.setChecked(value)
+            item.setCurrentIndex(value)
 
-    def get_selected_types(self):
+    def get_selected_types_include(self):
         return [typ['name'] for item, typ in self._items.items() if
-                item.isChecked()] or ['front']
+                item.currentIndex() == self.IDX_INCLUDE_TYPE] or ['front']
+
+    def get_selected_types_exclude(self):
+        return [typ['name'] for item, typ in self._items.items() if
+                item.currentIndex() == self.IDX_EXCLUDE_TYPE] or ['none']
 
     @staticmethod
-    def run(parent=None, types=None):
-        if types is None:
-                types = []
-        dialog = CAATypesSelectorDialog(parent, types)
+    def run(parent=None, types_include=None, types_exclude=None):
+        if types_include is None:
+            types_include = []
+        if types_exclude is None:
+            types_exclude = []
+        dialog = CAATypesSelectorDialog(parent, types_include, types_exclude)
         result = dialog.exec_()
-        return (dialog.get_selected_types(), result == QtWidgets.QDialog.Accepted)
+        return (dialog.get_selected_types_include(), dialog.get_selected_types_exclude(), result == QtWidgets.QDialog.Accepted)
 
 
 class ProviderOptionsCaa(ProviderOptions):
@@ -187,6 +220,7 @@ class ProviderOptionsCaa(ProviderOptions):
                          _CAA_IMAGE_SIZE_DEFAULT),
         config.ListOption("setting", "caa_image_types", ["front"]),
         config.BoolOption("setting", "caa_restrict_image_types", True),
+        config.ListOption("setting", "caa_image_types_to_omit", ["raw/unedited"]),
     ]
 
     _options_ui = Ui_CaaOptions
@@ -231,10 +265,11 @@ class ProviderOptionsCaa(ProviderOptions):
         self.ui.select_caa_types.setEnabled(enabled)
 
     def select_caa_types(self):
-        (types, ok) = CAATypesSelectorDialog.run(
-            self, config.setting["caa_image_types"])
+        (types, types_to_omit, ok) = CAATypesSelectorDialog.run(
+            self, config.setting["caa_image_types"], config.setting["caa_image_types_to_omit"])
         if ok:
             config.setting["caa_image_types"] = types
+            config.setting["caa_image_types_to_omit"] = types_to_omit
 
 
 
@@ -253,6 +288,7 @@ class CoverArtProviderCaa(CoverArtProvider):
     def __init__(self, coverart):
         super().__init__(coverart)
         self.caa_types = list(map(str.lower, config.setting["caa_image_types"]))
+        self.caa_types_to_omit = list(map(str.lower, config.setting["caa_image_types_to_omit"]))
         self.len_caa_types = len(self.caa_types)
         self.restrict_types = config.setting["caa_restrict_image_types"]
 
@@ -346,6 +382,8 @@ class CoverArtProviderCaa(CoverArtProvider):
             except ValueError:
                 self.error("Invalid JSON: %s" % (http.url().toString()))
             else:
+                if self.restrict_types:
+                    log.debug('CAA types: included: %s, excluded: %s' % (self.caa_types, self.caa_types_to_omit,))
                 for image in caa_data["images"]:
                     if config.setting["caa_approved_only"] and not image["approved"]:
                         continue
@@ -364,6 +402,10 @@ class CoverArtProviderCaa(CoverArtProvider):
                         # only keep enabled caa types
                         types = set(image["types"]).intersection(
                             set(self.caa_types))
+                        if types and self.caa_types_to_omit:
+                            types = not set(image["types"]).intersection(
+                                set(self.caa_types_to_omit))
+                        log.debug('CAA Image %s: %s  %s' % ('accepted' if types else 'rejected', image['image'], image['types'],))
                     else:
                         types = True
                     if types:
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -26,8 +26,10 @@ from collections import (
     OrderedDict,
     namedtuple,
 )
+from functools import partial
 
 from PyQt5 import (
+    Qt,
     QtCore,
     QtWidgets,
 )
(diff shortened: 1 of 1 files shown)


samj1912:
I am not sure if creating another dialog is the right idea.

Let's take this for example - 

Current usage allows a user to select AND exclude a particular type of cover art.

Maybe fix the logic there or unify the dialogs somehow.

rdswift:
@samj1912, I can see how it could be confusing for the user.  What do you think about just having one selection list with the choices of "Ignore", "Include" and "Exclude" for each cover art image type?  The choices could be done using radio buttons or a combo box (my preference) for each image type.

rdswift:
@samj1912 Here is an example of the combo box option.  Is this more what you had in mind?

image

rdswift:
This is another option using radio buttons.  I actually think I prefer this one.  Comments?

image

zas:
I tend to prefer the combo box option. The radio buttons one looks overloaded.

rdswift:
Thanks, @zas.  I'll proceed on that basis and update the PR today or tomorrow.  Are the terms "Include", "Exclude" and "Ignore" okay, or can you suggest something more appropriate?

rdswift:
@zas / @samj1912 

I've updated the code to use the combo box option for selecting which file types to include or exclude.  That simplified the code significantly, with only one file being modified.

For testing, I have been using https://musicbrainz.org/release/a4a99919-0923-40aa-8f64-b17820ebc407/cover-art

picard/coverart/providers/caa.py
```diff
@@ -82,15 +91,30 @@ def __init__(self, parent=None, types=None):
...
+            gridlayout.addWidget(item, row, column)
+            column += 1
+            item = QtWidgets.QComboBox()
+            item.addItems([self._include, self._ignore, self._exclude,])
+            if name in types_include:
+                item.setCurrentIndex(0) # Include type
+            elif name in types_exclude:
+                item.setCurrentIndex(2) # Exclude type
+            else:
+                item.setCurrentIndex(1) # Ignore type
```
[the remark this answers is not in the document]
rdswift: Changed to use constants defined as per your earlier comment.

picard/coverart/providers/caa.py
```diff
@@ -69,11 +69,20 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
-        if types is None:
-            types = []
+    _columns = 3        # Number of main display columns
+    _subcolumns = 3     # Number of columns comprising  a main display column (i.e.: label, combobox, spacer)
+    _spacer_width = 35  # Width of the spacer columns in pixels
+
+    # Text displayed in combo box options.
+    _include = N_('Include')
+    _exclude = N_('Exclude')
+    _ignore = N_('Ignore')
```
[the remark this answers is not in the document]
rdswift: Dropped these entirely and moved them to the `addItems()` method call as `item.addItems([ _('Include'), _('Ignore'), _('Exclude'),])`.

picard/coverart/providers/caa.py
```diff
@@ -320,6 +357,9 @@ def _caa_json_downloaded(self, data, http, error):
             except ValueError:
                 self.error("Invalid JSON: %s" % (http.url().toString()))
             else:
+                if self.restrict_types:
+                    log.debug('Include types: %s' % (self.caa_types,))
+                    log.debug('Exclude types: %s' % (self.caa_types_to_omit,))
```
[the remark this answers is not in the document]
rdswift: Changed as suggested.  Also added clarification to the accepted / rejected debug log entry.

picard/coverart/providers/caa.py
```diff
@@ -82,15 +91,30 @@ def __init__(self, parent=None, types=None):
...
 
         for index, caa_type in enumerate(CAA_TYPES):
             row = index // self._columns
-            column = index % self._columns
+            column = (index % self._columns) * self._subcolumns
             name = caa_type["name"]
-            text = translate_caa_type(name)
-            item = QtWidgets.QCheckBox(text)
-            item.setChecked(name in types)
+            text = translate_caa_type(name) + ':'
```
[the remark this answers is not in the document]
rdswift: I dropped the colon and revised the code to remove the `text` variable.  You're right, in that it served no real purpose.  Moved the `translate_caa_type(name)` directly to the argument of the `item.setText()` method call.

picard/coverart/providers/caa.py
```diff
@@ -126,26 +151,35 @@ def help(self):
...
+
+    def get_selected_types_include(self):
+        return [typ['name'] for item, typ in self._items.items() if
+                item.currentIndex() == 0] or ['front']
 
-    def get_selected_types(self):
+    def get_selected_types_exclude(self):
         return [typ['name'] for item, typ in self._items.items() if
-                item.isChecked()] or ['front']
+                item.currentIndex() == 2] or ['none']
```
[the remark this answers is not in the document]
rdswift: Changed to use explicit constants as per your previous comment.

phw:
I think the average user will struggle understanding the difference between "ignore" and "exclude", and to be honest I am not 100% sure I got it myself. Can we come up with a better naming?

The old dialog with only the checkboxes was ok to understand: Just choose the cover art types you are interested in.

As I understand it the new "exclude" option will blacklist certain types, right? So if I have "Front" included, "Raw/Unedited" excluded and all other types ignored it will use images of e.g. type "Front", or "Front, Back", but never anything that has "Raw/Unedited"?

I think the main confusion will arise because people probably don't intuitively think about cover art having multiple types.  And "Ignore" sounds a lot like as if images of this type will not be used. So I think the dialog should explain a bit more.

One simple solution would be to just add a small help section with some info like:

> Include: Images having this type set will be downloaded and used as cover art
> Exclude: Images having this type will never be downloaded, even if they have also an included type
> Ignore: This type has no affect on choosing the cover art

I struggle with describing "Ignore", maybe somebody has a better idea.

Another option would be to rethink the UI. Maybe something like this?

grafik

Not sure really how to best solve this, but I am very sure Include, Ignore, Exclude will be confusing for many.

Apart from that I think this is really useful functionality. Not only for the raw type but also for other cases. E.g. for my files I basically never want to use an image labeled as front that also includes the back.

rdswift:
I agree that the terms could very well be confusing for most users, at least initially.  I have struggled all along with trying to find the right terms to use.  Your simple solution with the help section explanation could work.  Perhaps the explanation for "Ignore" could be something like, "This type is not considered when deciding whether or not to download an image."  Perhaps "Ignore" should be replaced with "Not considered"?

The intended logic is simply that an image will be downloaded only if a) one of its types matches a type in the "Include" list, **AND** b) none of its types match a type in the "Exclude" list.

I like your alternate UI suggestion because I think that it makes it much clearer.  Unfortunately, the way I struggle with UI design using Qt, I suspect it will take me quite a while to develop tight working code for it.  :frowning_face:

rdswift:
Does something like this make it a bit clearer for the users?

image

samj1912:
I like the solution proposed by phw and the one implemented by you in the 2nd image.

phw:
I just ran the code, that feels much better now! I like it very much.

zas:
It looks much better now ! Good job :+1: 

Few quirks with the UI:

capture du 2018-08-18 21-46-46

Since one can select one item in each list it becomes hard to understand, imho only one item should be selected at once.

Also items should be ordered consistently, for example, if i move one item from top of middle list to exclude list then back to the middle list, it appears at the end of the list. May be sorting alphabetically the (translated) items would improve that, dunno.

When an action isn't possible, matching button should be disabled (no item selected, or empty list).

I'll test a bit more tomorrow.

zas:
@rdswift : thanks :)

Let's nitpick a bit:
- it would be great to be able to select multiple items in the same list (using shift/control) and move them at once.
- instead of Clear All it may be better to have a Reset to Defaults button.
- may be start a thread on forums about this dialog, and see what users think about it.

What do you think?

rdswift:
@zas 

Just updated it to allow multiple selections.  I thought it was already doing that, but it seems that I forgot to specifically enable it.

Added a ***Restore Defaults*** button.  Great suggestion!  For now I've also left the ***Clear All*** button, but that can easily be removed if we want.

I'm just working on the thread in the forums now, so we'll see what the community has to say and what suggestions they might have.  Another great idea.  Thanks.

picard/coverart/providers/caa.py
```diff
@@ -46,6 +46,61 @@
...
+    for item in list1.selectedItems():
+        move_item(item, list1, list2)
+
+def move_all_items(list1, list2):
+    """Move all items from one listbox to another.  Used in the CAA image type selection dialog."""
+    while list1.count():
+        item = list1.item(0)
+        move_item(item, list1, list2)
+
+def clear_listbox_focus(list_to_clear):
```
zas: This one isn't strictly needed, you can call `clearSelection()` directly.

picard/coverart/providers/caa.py
```diff
@@ -46,6 +46,61 @@
...
+
+
+def make_list_box():
+    """Make standard list box for CAA image type selection dialog."""
+    list_box = QtWidgets.QListWidget()
+    list_box.setFixedSize(QtCore.QSize(150, 250))
+    list_box.setSizePolicy(QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
+    list_box.setSortingEnabled(True)
+    list_box.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
+    return list_box
```
zas: Instead of using `make_*()`, you could subclass Qt objects

```python
class ListBox(QtWidgets.QListWidget):
        super().__init__(parent=parent)
        self.setFixedSize(QtCore.QSize(150, 250))
        self.setSizePolicy(QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
        self.setSortingEnabled(True)
        self.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
```

```python
class ArrowButton(QtWidgets.QPushButton):
    def __init__(self, text, parent=None, command=None):
        super().__init__(text, parent=parent)
        if command:
            self.clicked.connect(command)
        self.setFixedSize(QtCore.QSize(35, 20))
```

And same for `make_arrows_column()`

picard/coverart/providers/caa.py
```diff
@@ -46,6 +46,61 @@
...
+    arrows_layout.addWidget(make_arrow_button('<' if reverse else '>', list_remove))
+    arrows_layout.addWidget(make_arrow_button('<<' if reverse else '>>', list_remove_all))
+    arrows_layout.addItem(QtWidgets.QSpacerItem(_spacer_item))
+    return arrows
+
+def move_item(item, list1, list2):
+    """Move the specified item from one listbox to another.  Used in the CAA image type selection dialog."""
+    clone = item.clone()
+    list2.addItem(clone)
+    list1.takeItem(list1.row(item))
```
zas: Perhaps you can skip the `clone()` if you remove it first from list1 then add it to list2:

```python
list2.addItem(list1.takeItem(list1.row(item)))
```

Untested though.

picard/coverart/providers/caa.py
rdswift: Right you are.  That was a holdover from when I was clearing each item individually.

picard/coverart/providers/caa.py
rdswift: Makes sense.  It seems I haven't completely embraced the concept of OOP yet.  Old habits die hard.  :grin:

picard/coverart/providers/caa.py
rdswift: It works, but I had to remove it from the source list before I could add it to the target list.

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+    Keyword Arguments:
+        label {string} -- Label to display on the button
+        command {command} -- Command to execute when the button is clicked (default: {None})
+        parent {[type]} -- Parent of the QPushButton object being created (default: {None})
+    """
+
+    ARROW_BUTTON_WIDTH = 35
+    ARROW_BUTTON_HEIGHT = 20
+
+    def __init__(self, label, command=None, parent=None ):
```
zas: There's an extra space before last parenthesis ;)

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
     return urls
 
 
+class arrow_button(QtWidgets.QPushButton):
```
zas: Class names should be camel-cased, so it should be `class ArrowButton`

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+    def __init__(self, label, command=None, parent=None ):
+        if label is None:
+            label = '.'
+        super().__init__(label, parent=parent)
+        if command is not None:
+            self.clicked.connect(command)
+        self.setFixedSize(QtCore.QSize(self.ARROW_BUTTON_WIDTH, self.ARROW_BUTTON_HEIGHT))
+
+
+class arrows_column(QtWidgets.QWidget):
```
zas: `ArrowColumn`

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        list_remove {command} -- Command to execute when the 'remove' button is clicked (default: {None})
+        list_remove_all {command} -- Command to execute when the 'remove all' button is clicked (default: {None})
+        reverse {bool} -- Determines whether the arrow directions should be reversed (default: {False})
+        parent {[type]} -- Parent of the QWidget object being created (default: {None})
+    """
+
+    def __init__(self, list_add=None, list_add_all=None, list_remove=None, list_remove_all=None, reverse=False, parent=None):
+        _spacer_item = QtWidgets.QSpacerItem(20, 20, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
+        _arrows_layout = QtWidgets.QVBoxLayout()
+        super().__init__(parent=parent)
```
zas: Call it as soon as possible.

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        _arrows_layout.addItem(QtWidgets.QSpacerItem(_spacer_item))
+        self.BUTTON_ADD = arrow_button('>' if reverse else '<', list_add)
```
zas: Since those are widgets, not constants, names should be lowercased.

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        list_add_all {command} -- Command to execute when the 'add all' button is clicked (default: {None})
+        list_remove {command} -- Command to execute when the 'remove' button is clicked (default: {None})
+        list_remove_all {command} -- Command to execute when the 'remove all' button is clicked (default: {None})
+        reverse {bool} -- Determines whether the arrow directions should be reversed (default: {False})
+        parent {[type]} -- Parent of the QWidget object being created (default: {None})
+    """
+
+    def __init__(self, list_add=None, list_add_all=None, list_remove=None, list_remove_all=None, reverse=False, parent=None):
+        _spacer_item = QtWidgets.QSpacerItem(20, 20, QtWidgets.QSizePolicy.Minimum, QtWidgets.QSizePolicy.Expanding)
+        _arrows_layout = QtWidgets.QVBoxLayout()
```
zas: Since `_arrows_layout` and `_space_item` are local to this method, no need to prefix with `_` (unlike a "private" property, ie. `self._arrows_layout`)

picard/coverart/providers/caa.py
```diff
@@ -123,8 +72,97 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        _arrows_layout.addWidget(self.BUTTON_ADD_ALL)
+        self.BUTTON_REMOVE = arrow_button('<' if reverse else '>', list_remove)
+        _arrows_layout.addWidget(self.BUTTON_REMOVE)
+        self.BUTTON_REMOVE_ALL = arrow_button('<<' if reverse else '>>', list_remove_all)
+        _arrows_layout.addWidget(self.BUTTON_REMOVE_ALL)
+        _arrows_layout.addItem(QtWidgets.QSpacerItem(_spacer_item))
+        self.setLayout(_arrows_layout)
+
+
+class list_box(QtWidgets.QListWidget):
```
zas: `class ListBox`

rdswift:
I think I've addressed all the style issues and such from your last request.  Sorry, I should have re-read the project guidelines again about style, so you wouldn't have to worry about that.  Thanks for your patience.  I think it's looking much better than it was, and I'm learning a lot in the process.  Thank-you all.

picard/coverart/providers/caa.py
rdswift: And that's something I even look for...  Argh!  :blush:

picard/coverart/providers/caa.py
rdswift: Fixed.

picard/coverart/providers/caa.py
rdswift: Fixed.

picard/coverart/providers/caa.py
rdswift: Moved to first line of __init__.  Same change made to other new classes.

picard/coverart/providers/caa.py
rdswift: Fixed.

picard/coverart/providers/caa.py
rdswift: Fixed.  Looks much better.

picard/coverart/providers/caa.py
rdswift: Fixed.

picard/coverart/providers/caa.py
```diff
@@ -68,34 +73,209 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
 
         self.setWindowTitle(_("Cover art types"))
-        self._items = {}
         self.layout = QtWidgets.QVBoxLayout(self)
 
+        # Populate display title / standard name dictionary
+        for caa_type in CAA_TYPES:
+            name = caa_type['name']
+            title = _(caa_type['title'])
+            self.image_types_by_display_title[title] = name
```
zas: Using translated string as key isn't a very good idea imho.
You should use the name which is meant for that (`caa_type['name']`)

Also `_()` will not work with caa types, because they come from another set of po files, you need to use `translate_caa_type()` (from coverart.utils)

Read about http://doc.qt.io/qt-5/qlistwidgetitem.html#setData and UserRole i think it is what you need here:
each item has caa type name as data, and you use `QListWidget.sortItems()`
It will prolly not work as expected because of QLocale stuff though (that's something we need to fix in Picard), don't bother with that for now.

picard/coverart/providers/caa.py
```diff
@@ -122,30 +304,79 @@ def __init__(self, parent=None, types=None):
...
+    def reset_to_defaults(self):
+        self.fill_lists(_CAA_IMAGE_TYPE_DEFAULT_INCLUDE, _CAA_IMAGE_TYPE_DEFAULT_EXCLUDE)
+        self.fix_buttons()
+
+    def fix_buttons(self):
+        has_items_include = self.list_include.count()
+        has_items_exclude = self.list_exclude.count()
+        has_items_ignore = self.list_ignore.count()
+
+        has_selected_include = True if self.list_include.selectedItems() else False
```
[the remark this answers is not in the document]
rdswift: I tried that before without the bool() and got the following error:

      File ".\picard\coverart\providers\caa.py", line 443, in select_caa_types
        self, config.setting["caa_image_types"], config.setting["caa_image_types_to_omit"])
      File ".\picard\coverart\providers\caa.py", line 380, in run
        dialog = CAATypesSelectorDialog(parent, types_include, types_exclude)
      File ".\picard\coverart\providers\caa.py", line 307, in __init__
        self.fix_buttons()
      File ".\picard\coverart\providers\caa.py", line 363, in fix_buttons
        self.arrows_include.button_add.setEnabled(has_items_ignore and has_selected_ignore)
    TypeError: setEnabled(self, bool): argument 1 has unexpected type 'list'

It seems that the error occurs because it's trying to AND a bool with a list.  That's why I went with forcing it to a True or False directly.

It works fine with the bool(), so I'll make that change.

picard/coverart/providers/caa.py
```diff
@@ -122,30 +304,79 @@ def __init__(self, parent=None, types=None):
...
+        self.fix_buttons()
 
-    def get_selected_types(self):
-        return [typ['name'] for item, typ in self._items.items() if
-                item.isChecked()] or ['front']
+    def reset_to_defaults(self):
+        self.fill_lists(_CAA_IMAGE_TYPE_DEFAULT_INCLUDE, _CAA_IMAGE_TYPE_DEFAULT_EXCLUDE)
+        self.fix_buttons()
+
+    def fix_buttons(self):
```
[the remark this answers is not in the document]
rdswift: You're right.  It doesn't sound good.  I'll change it to ***set_buttons_enabled_state***.

picard/coverart/providers/caa.py
samj1912: I am sure something like User Role will work. The easiest way is to just subclass it and add a field. See the toolbar customization options dialog in UI options for the relevant code.

picard/coverart/providers/caa.py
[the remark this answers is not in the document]
rdswift: I also just realized that I was no longer calling the method since moving the item moving code to ListBox methods.  I've set it up as a callback now and it's working properly again.

picard/coverart/providers/caa.py
rdswift: I struggled with that before but I think I have it now.  I'm storing the `caa_type['name']` in the list item ***UserRole*** role and the `caa_type['title']` as the display text.  I've also changed it to use `translate_caa_type()`.  It seems to be sorting the lists properly here for my testing, but everything is English so it's not much of a test.  I removed that extra dictionary.

This may sound strange, but one of my biggest problems is finding out which PyQt5 module a class or enumeration is in.  I had a heck of a time finding UserRole because I tried specifying it in PyQt5.Qt module instead of the PyQt5.QtCore.Qt module.  I guess I'll figure it out eventually.  :slightly_smiling_face: 

I'm just doing some final testing and clean-up and will upload shortly.

rdswift:
For some reason appveyor is stalling at the point of ***running test***, and then timing and failing after an hour.  Any idea what I might have done to cause it to fail?

image

zas:
> Any idea what I might have done to cause it to fail?

I tried to restart the build, etc.. and it still gets stuck at start of `python setup.py test -v` command, i see nothing in your PR that may cause this :(

@samj1912 ^^ any idea why this happens ??

phw:
For a quick test I ran the tests for that branch in my Windows 10 VM, but that works as expected. And I don't see anything either that looks suspicious :(

samj1912:
Appveyor complains that this pr is non mergable.

rdswift:
@samj1912:  I suspect that's because the master branch had changed with all the includes sorted with isort.  I rebased the PR on the updated master, and it appears that appveyor is stalling at the tests again.

zas:
> I rebased the PR on the updated master, and it appears that appveyor is stalling at the tests again

It is unrelated to conflicts, i emailed Appveyor support team about it, since we have no clue about what is happening.

rdswift:
Before I do something I shouldn't, would it be okay if I (temporarily) modified the `appveyor.yml` file and tried connecting as described in their response to see if I can get any more information?  I would then revert the changes.

phw:
@rdswift What I usually do to fix issues with  automated CI is create a separate branch just for experimentation of the builds. And once it is working I commit the final version to the target branch.

rdswift:
@phw: Thanks for the tip.  It worked like a charm.

@zas: The error message dialog box was:

image

When I clicked on the `Okay` button, the console showed:

image


I have no idea what in my code might be triggering that, but at least it's a place to start...

zas:
> I have no idea what in my code might be triggering that, but at least it's a place to start...

Nor i, i send the image to appveyor support ... let's hope they have an idea about what is going on ;)

picard/coverart/providers/caa.py
```diff
@@ -26,8 +26,10 @@
     OrderedDict,
     namedtuple,
 )
+from functools import partial
 
 from PyQt5 import (
+    Qt,
```
zas: This is prolly the cause of the appveyor issue, see https://stackoverflow.com/questions/48514027/spyder3-python-ide-does-not-start-this-windows-version-does-not-support-the-re

In fact, it is incorrect, and this import should be removed, because `QListWidgetItem` belongs to `QtWidgets`

So replace:

```python
 item = Qt.QListWidgetItem(title)
```

with:

```python
 item = QtWidgets.QListWidgetItem(title)
```

And let's see if it fixes the issue.

picard/coverart/providers/caa.py
```diff
@@ -94,34 +98,203 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        self.arrows_exclude = ArrowsColumn(
+            self.list_exclude,
+            self.list_ignore,
+            callback=self.set_buttons_enabled_state,
+            reverse=True
+        )
+        gridlayout.addWidget(self.arrows_exclude, row, column)
+
+        column = 4
+        gridlayout.addWidget(self.list_exclude, row, column)
```
zas: Code would be perhaps easier to read using a small local function, i was thinking about something like:

```python
        self.arrows_include = ArrowsColumn(
            self.list_include,
            self.list_ignore,
            callback=self.set_buttons_enabled_state,
        )

        self.arrows_exclude = ArrowsColumn(
            self.list_exclude,
            self.list_ignore,
            callback=self.set_buttons_enabled_state,
            reverse=True
        )

        def add_widget(row=0, column=0, widget):
             gridlayout.addWidget(widget, row, column)

        add_widget(row=0, column=0, QtWidgets.QLabel(_("Include types list")))
        add_widget(row=1, column=0, self.list_include)

        add_widget(row=1, column=1, self.arrows_include)
        add_widget(row=1, column=2, self.list_ignore)
        add_widget(row=1, column=3, self.arrows_include)

        add_widget(row=0, column=4, QtWidgets.QLabel(_("Exclude types list")))
        add_widget(row=1, column=4, self.list_exclude)
```

No strong opinion here though.

rdswift: I like it.  It definitely looks cleaner than the way it is currently.

picard/coverart/providers/caa.py
```diff
@@ -148,30 +323,97 @@ def __init__(self, parent=None, types=None):
...
     def help(self):
         webbrowser2.goto('doc_cover_art_types')
 
-    def uncheckall(self):
-        self._set_checked_all(False)
+    def get_selected_types_include(self):
+        return [self.list_include.item(index).data(QtCore.Qt.UserRole) for index in range(self.list_include.count())] or ['front']
+
+    def get_selected_types_exclude(self):
+        return [self.list_exclude.item(index).data(QtCore.Qt.UserRole) for index in range(self.list_exclude.count())] or ['none']
```
zas: A simple iterator in `ListBox` class could help to shorten lines here:

```python
class ListBox(QtWidgets.QListWidget):

    def all_items_data(self, role=QtCore.Qt.UserRole):
        for index in range(self.count()):
            yield self.item(index).data(role)
```

```python
    def get_selected_types_include(self):
        return [self.list_include.all_items_data()] or ['front']

    def get_selected_types_exclude(self):
        return [self.list_exclude.all_items_data()] or ['none']
```

picard/coverart/providers/caa.py
```diff
@@ -94,34 +98,195 @@ def caa_url_fallback_list(desired_size, thumbnails):
...
+        label {string} -- Label to display on the button
+        command {command} -- Command to execute when the button is clicked (default: {None})
+        parent {[type]} -- Parent of the QPushButton object being created (default: {None})
+    """
+
+    ARROW_BUTTON_WIDTH = 35
+    ARROW_BUTTON_HEIGHT = 20
+
+    def __init__(self, label, command=None, parent=None):
+        if label is None:
```
zas: Imho you can make `label` mandatory, and remove those 2 lines.

rdswift: As written, those lines didn't make sense because `label` was already mandatory.  I've removed them.

PR #915, opened 2018-07-31
PICARD-1302, PICARD-1303, PICARD-1304: Fix crashes

antlarr:
# Summary

Fix two crashes.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

- Fix crash when using drag&drop from Google Images to the coverartbox widget due to code not being fully ported from Qt4 to Qt5.
- Fix crash due to KeyError exceptions when trying to match a cluster with a release that doesn't have any track in musicbrainz.
- Fix crash when a cluster has a file with track number and a file without any track number information and the user opens the info dialog on the cluster.

# Solution

- Ported code to Qt5
- Catch and handle exceptions properly
- Treat files with no track number as having track number 0.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -150,10 +150,13 @@ class Metadata(dict):
         except (ValueError, KeyError):
             pass
         else:
-            if "title" in weights:
-                b = release['media'][0]['track-count']
-            else:
-                b = release['track-count']
+            try:
+                if "title" in weights:
+                    b = release['media'][0]['track-count']
+                else:
+                    b = release['track-count']
+            except KeyError:
+                b = 0
             score = 0.0 if a > b else 0.3 if a < b else 1.0
             parts.append((score, weights["totaltracks"]))
 
@@ -172,7 +175,7 @@ class Metadata(dict):
             parts.append((score, weights["releasecountry"]))
 
         total_formats = len(preferred_formats)
-        if total_formats:
+        if total_formats and 'media' in release:
             score = 0.0
             subtotal = 0
             for medium in release['media']:
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -19,6 +19,7 @@
 
 from collections import namedtuple
 import os.path
+import re
 import traceback
 from PyQt5 import QtGui, QtCore, QtWidgets
 from picard import log
@@ -357,7 +358,12 @@ class ClusterInfoDialog(InfoDialog):
             try:
                 return int(item.tracknumber)
             except:
-                return item.tracknumber or 0
+                try:
+                    # This allows to parse values like '3' but also '3/10'
+                    m = re.search('^\d+', item.tracknumber)
+                    return int(m.group(0))
+                except:
+                    return 0
 
         lines = ["%s %s - %s (%s)" % item for item in sorted(tracklist, key=sorttracknum)]
         info.append("<b>%s</b><br />%s" % (_('Tracklist:'),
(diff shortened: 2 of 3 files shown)


antlarr:
@samj1912 , I created the issues (I couldn't find any existing reference to those problems) and added them to the PR's title. Is that enough?

samj1912:
Works yes :)

picard/ui/infodialog.py
```diff
@@ -357,7 +357,7 @@ def sorttracknum(item):
             try:
                 return int(item.tracknumber)
             except:
-                return item.tracknumber
+                return item.tracknumber or 0
```
samj1912: Cant we directly return 0 here?

picard/metadata.py
```diff
@@ -150,10 +150,13 @@ def compare_to_release_parts(self, release, weights):
...
-            if "title" in weights:
-                b = release['media'][0]['track-count']
-            else:
-                b = release['track-count']
+            try:
+                if "title" in weights:
+                    b = release['media'][0]['track-count']
+                else:
+                    b = release['track-count']
+            except KeyError:
```
samj1912: Could this also raise an IndexError?

picard/ui/infodialog.py
antlarr: Yes, I usually try to change as little code as possible, but in this case, you're right that if the cast to int failed, then we should just return 0 in any case.

picard/metadata.py
antlarr: I don't think so. If the release contains a media, it must have a first element, and if it doesn't contain a media then it will already raise a KeyError when evaluating release['media']. I don't think there's such a case as release['media'] being an empty list. Or there is?

picard/ui/infodialog.py
antlarr: Hmmm, after thinking a bit more about it, I've made it a bit more complex and used a regular expression to extract a number if possible, since returning a string was used for tracks containing values like '3/10'. Previously, if all tracks in a cluster contained tracknumbers in such format, it worked fine since the sorting was done by comparing the strings, but if there are tracks with values '2' and '3/10', then that failed even with my commit.
I've just added another commit that parses all cases correctly and always try to convert item.tracknumber to an int.

picard/ui/infodialog.py
```diff
@@ -357,7 +358,12 @@ def sorttracknum(item):
             try:
                 return int(item.tracknumber)
             except:
-                return item.tracknumber or 0
+                try:
+                    # This allows to parse values like '3' but also '3/10'
+                    m = re.search('^\d+', item.tracknumber)
+                    return int(m.group(0))
+                except:
```
samj1912: We can do something better than a bare except here.

picard/ui/infodialog.py
```diff
@@ -357,7 +358,12 @@ def sorttracknum(item):
             try:
                 return int(item.tracknumber)
             except:
```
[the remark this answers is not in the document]
antlarr: Did you mean ValueError?

picard/ui/infodialog.py
antlarr: Anything more apart from AttributeError?

picard/ui/infodialog.py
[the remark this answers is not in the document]
samj1912: Ah yes, sorry, Value Error

PR #916, opened 2018-08-01
PICARD-1207: Fix handling paths with unicode characters on python3

antlarr:
# Summary

Fix a crash moving additional files when their full path contains non-ascii characters.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When moving additional files, the call to ```encode_filename``` encodes the path to the original filename converting the variable from str to bytes, provoking the next exception . ```shutil.move``` only accepts strings (the documentation don't specifically mention that for ```shutil.move```, but it does for [```copy```](https://docs.python.org/3/library/shutil.html#shutil.copy), ```copyfile```, ```copymode```, etc. and the error makes clear the same is true for ```shutil.move```)

  File "./picard/util/thread.py", line 47, in run
    result = self.func()
  File "./picard/file.py", line 222, in _save_and_rename
    self._move_additional_files(old_filename, new_filename)
  File "./picard/file.py", line 435, in _move_additional_files
    shutil.move(old_file, new_file)
  File "/usr/x86_64-pc-linux-gnu/lib/python3.6/shutil.py", line 551, in move
    if _destinsrc(src, dst):
  File "/usr/x86_64-pc-linux-gnu/lib/python3.6/shutil.py", line 565, in _destinsrc
    if not src.endswith(os.path.sep):
TypeError: endswith first arg must be bytes or a tuple of bytes, not str

* JIRA ticket: [PICARD-1207](https://tickets.metabrainz.org/browse/PICARD-1207)

# Solution

I removed the calls to ```encode_filename``` in ```file.py```, which convert the strings to bytes and breaks the call to ```shutil.move``` which require a string to be passed (as mentioned above).

Also, other functions like ```os.path.dirname```, ```os.path.isdir```, ```os.stat```, ```os.utime``` ... allow to be passed a string, instead of the encoded path in a bytes variable. So I removed the use of ```encode_filename``` for all cases that either require or allow a string to be passed so we don't do unnecessary processing neither.

### DIFF
--- a/test/test_filesystem.py
+++ b/test/test_filesystem.py
@@ -1,0 +1,89 @@
+# -*- coding: utf-8 -*-
+import os.path
+import picard.formats
+import unittest
+import shutil
+from contextlib import suppress
+
+from PyQt5 import QtCore
+from picard import config
+from tempfile import mkdtemp
+from test.test_formats import FakeTagger
+
+
+settings = {
+    'enabled_plugins': '',
+    'move_files': True,
+    'move_additional_files': True,
+    'move_additional_files_pattern': 'cover.jpg',
+}
+
+
+class TestFileSystem(unittest.TestCase):
+
+    def setUp(self):
+        self.src_directory = mkdtemp()
+        self.tgt_directory = mkdtemp()
+        QtCore.QObject.tagger = FakeTagger()
+        config.setting = settings.copy()
+
+    def tearDown(self):
+        shutil.rmtree(self.src_directory)
+        shutil.rmtree(self.tgt_directory)
+
+    def _set_up_src_file(self, filename, src_rel_path):
+        """Copy filename to the src directory under src_rel_path."""
+        path = os.path.join(self.src_directory, src_rel_path)
+        shutil.copy(filename, path)
+        return path
+
+    def _set_up_tgt_filename(self, tgt_rel_path):
+        """Return the absolute path to tgt_rel_path in the tgt directory."""
+        return os.path.join(self.tgt_directory, tgt_rel_path)
+
+    def _prepare_files(self, src_rel_path='', tgt_rel_path=''):
+        """Prepare src files and tgt filenames for a test."""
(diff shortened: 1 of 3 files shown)


antlarr:
Good idea. I'll try to write a test tomorrow that directly calls File._move_additional_files and checks that it doesn't raise any exceptions and that files are moved correctly.

test/test_filesystem.py
```diff
@@ -0,0 +1,97 @@
...
+    def setUp(self):
+        self.src_directory = mkdtemp()
+        self.tgt_directory = mkdtemp()
+        QtCore.QObject.tagger = FakeTagger()
+        config.setting = settings.copy()
+
+    def tearDown(self):
+        shutil.rmtree(self.src_directory)
+        shutil.rmtree(self.tgt_directory)
+        pass
```
samj1912: Leftovers?

test/test_filesystem.py
```diff
@@ -0,0 +1,97 @@
...
+            os.mkdir(os.path.join(self.src_directory, src_rel_path))
+
+        # Prepare the source directory structure under self.src_directory
+        # .../<src_rel_path>/test.mp3
+        # .../<src_rel_path>/cover.jpg
+
+        old_filename = \
+            self._set_up_src_file(os.path.join('test', 'data', 'test.mp3'),
+                                  os.path.join(src_rel_path, 'test.mp3'))
+        old_additional_filename = \
```
samj1912: We are not that strict on line lengths, even 120-130 chars is fine. Please avoid using `\`. It makes for some very ugly to read code.

test/test_filesystem.py
antlarr: This is to remove the temporary files from the tests

test/test_filesystem.py
antlarr: ok

PR #909, opened 2018-07-23
Cleanup Imports

Cabalist:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Minor / simple change (like a typo)
    * [ ] Other
* **Describe this change in 1-2 sentences**:
Used an automated tool to reorder the imports to be in line with PEP8 and remove unused ones.   

# Problem

I noticed some files had unused imports.   This should not affect functionality at all (the tests pass) and helps the code be better in line with PEP8.

# Solution

I ran an automated tool on the code and then reviewed each file.

# Action

PR #923, opened 2018-08-19
Import cleanup using isort

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Imports are clearly unordered
There was a cleanup attempt in https://github.com/metabrainz/picard/pull/909 already but the PR looks stale

# Solution

Use isort tool with proper configuration to clean up the mess
https://github.com/timothycrosley/isort

Since the config file can be stored in the tree, it will be easy to re-apply on regular basis (there are various ways to do so, including hooks, or editors plugins, see documentation).

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -18,32 +18,50 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-import traceback
 from operator import itemgetter
-from PyQt5 import QtCore, QtNetwork
-from picard import config, log
-from picard.coverart import coverart
-from picard.metadata import (Metadata,
-                             register_album_metadata_processor,
-                             run_album_metadata_processors,
-                             run_track_metadata_processors)
-from picard.dataobj import DataObject
-from picard.file import File
-from picard.track import Track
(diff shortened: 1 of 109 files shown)


picard/album.py
```diff
@@ -18,32 +18,49 @@
...
-import traceback
 from operator import itemgetter
-from PyQt5 import QtCore, QtNetwork
-from picard import config, log
+import traceback
+
+from PyQt5 import (
+    QtCore,
+    QtNetwork,
+)
```
phw: I am not sure I like how even simple imports like this get split over multiple lines. It has some advantages when editing the imports, but overall I think it hurts readability.

Could this be tweaked with isort?

zas: Yes, it can, but i prefer it like this, because it is consistent. If more than one module is imported, it will make one line for each, it helps a lot for readibility of diff patchs (one line added/removed/modified) and also helps with merges and conflict resolution.

Currently we have:

```python
from A import B
from A import B, C
from A import B, C, D
from A import (B, C)
from A import (
   B,
   C,
   D)
```
etc..

With this change we have:

```python
from A import B
from A import (
    B,
    C,
)
from A import (
   B,
   C,
   D,
)
```

samj1912:
Lgtm, rebase squash merge.

PR #925, opened 2018-08-23
PICARD-1308: fix incorrect version number in Info.plist (macosx)

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1308](https://tickets.metabrainz.org/browse/PICARD-1308)

# Solution

# Action

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -1,8 +1,20 @@
 # -*- mode: python -*-
 
-import os
 import glob
+import os
 import platform
+import sys
+
+
+# Get the version
+# and build a CFBundleVersion compatible version of it according to Apple dev documentation
+sys.path.append('.')
+from picard import PICARD_VERSION
+pv = [str(x) for x in PICARD_VERSION]
+macos_picard_version = '.'.join(pv[:3])
+macos_picard_short_version = macos_picard_version
+if pv[3] != 'final':
+	macos_picard_version += pv[3][0] + ''.join(pv[4:])
 
 
 def _picard_get_locale_files():
@@ -24,7 +36,7 @@ def get_locale_messages():
     data_files = []
     for locale in _picard_get_locale_files():
         data_files.append(
-            (os.path.join("build", "locale", locale[1], "LC_MESSAGES" , locale[0] + ".mo"),
+            (os.path.join("build", "locale", locale[1], "LC_MESSAGES", locale[0] + ".mo"),
              os.path.join("locale", locale[1], "LC_MESSAGES")))
     return data_files
 
@@ -38,7 +50,11 @@ data_files = get_locale_messages()
 fpcalc_name = 'fpcalc'
 if os_name == 'Windows':
     fpcalc_name = 'fpcalc.exe'
-    binaries += [('discid.dll', ''), ('ssleay32.dll', ''), ('libeay32.dll', '')]
+    binaries += [
+        ('discid.dll', ''),
+        ('ssleay32.dll', ''),
+        ('libeay32.dll', ''),
+    ]
 
 if os_name == 'Darwin':
     binaries += [('libdiscid.0.dylib', '')]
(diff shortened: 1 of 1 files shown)


samj1912:
Pass it here, it should write to the info plist https://github.com/metabrainz/picard/blob/master/picard.spec#L88

zas:
> Pass it here, it should write to the info plist https://github.com/metabrainz/picard/blob/master/picard.spec#L88

See last changes, i have no way to test it though.

phw:
I tested this, and showing the info from the app bundle looks correct now:

macos-app-info

I could not test if the app bundle runs as it should, though, since independent of this PR I am unable to generate a bundle that does not immediately crash after launch :(

picard.spec
```diff
@@ -85,8 +100,17 @@ coll = COLLECT(exe,
 
 
 if platform.system() == 'Darwin':
-    info_plist = {'NSHighResolutionCapable': 'True',
-                  'NSPrincipalClass': 'NSApplication'}
+    info_plist = {
+        'NSHighResolutionCapable': 'True',
+        'NSPrincipalClass': 'NSApplication',
+        'CFBundleName': 'picard',
```
phw: Must be `Picard` (uppercase P), because this will be used as displayed name of the macOS menu

picard.spec
```diff
@@ -85,8 +100,17 @@ coll = COLLECT(exe,
...
+        'CFBundleDisplayName': 'MusicBrainz Picard',
+        'CFBundleGetInfoString': 'Audio file tagger',
+        'CFBundleIdentifier': 'org.musicbrainz.picard',
+        'CFBundleVersion': macos_picard_version,
+        'CFBundleShortVersionString': macos_picard_version,
+        'NSHumanReadableCopyright': 'Copyright MetaBrainz Foundation',
```
phw: First "Copyright" should not be part of this string, otherwise it will display "Copyright: Copyright MetaBrainz Foundation" in the info dialog (see my previous screenshot). 

And second I am not sure we should set this, as it is actually not really correct because there are multiple copyright holders (all Picard contributors). So either we duplicate the copyright statement from the about dialog or we just leave this away entirely. In the latter case looking at the info of the app with Finder the Copyright line will not even be shown.

picard.spec
zas: Done.

picard.spec
zas: Removed, it was just to test.

picard.spec
```diff
@@ -85,8 +100,16 @@ coll = COLLECT(exe,
...
-                  'NSPrincipalClass': 'NSApplication'}
+    info_plist = {
+        'NSHighResolutionCapable': 'True',
+        'NSPrincipalClass': 'NSApplication',
+        'CFBundleName': 'Picard',
+        'CFBundleDisplayName': 'MusicBrainz Picard',
+        'CFBundleGetInfoString': 'Audio file tagger',
+        'CFBundleIdentifier': 'org.musicbrainz.picard',
+        'CFBundleVersion': macos_picard_version,
+        'CFBundleShortVersionString': macos_picard_version,
```
phw: If I understand the docs correctly the `CFBundleShortVersionString` should only be three parts without beta information. So `CFBundleVersion` currently would be 2.0.4d1 while `CFBundleShortVersionString` would only be `2.0.4`. Other than the docs I found no issue with how the package is handled and the version is displayed in macOS, though.

See:

- https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111349
- http://hboon.com/difference-between-cfbundleversion-and-cfbundleshortversionstring/

zas: Yes, you're right, should be fixed with last commit.

PR #927, opened 2018-08-24
PICARD-1319: Provide cover art metadata in cover art naming script

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Allow the file naming script for cover art to access additional metadata about the cover art itself, such as types, maintype and comment. This will allow e.g. the filename to be based upon the image type.

* JIRA ticket (_optional_): [PICARD-1319](https://tickets.metabrainz.org/browse/PICARD-1319)

# Solution

Add the following additional metadata variables for the cover ar file name script:

- `coverart_maintype`: The main type of the artwork as determined by `CoverArtImage.maintype`
- ` coverart_types`: Full multivalue list of all types
- ` coverart_comment`: The cover art comment

This will allow something like this:

```
$if($eq(%coverart_maintype%,front),AlbumArt,%album%_%coverart_maintype%)
```

# Action

With this change we can discuss to also remove the CAA option "Use the first image type as the filename", since setting the file name to just `%coverart_maintype%` is equivalent to this.

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -38,6 +38,7 @@ from picard import (
     log,
 )
 from picard.coverart.utils import translate_caa_type
+from picard.metadata import Metadata
 from picard.script import ScriptParser
 from picard.util import (
     decode_filename,
@@ -261,7 +262,15 @@ class CoverArtImage:
         # TODO: do something better than randomly using the first in the list
         return self.types[0]
 
-    def _make_image_filename(self, filename, dirname, metadata):
+    def _make_image_filename(self, filename, dirname, _metadata):
+        metadata = Metadata()
+        metadata.copy(_metadata)
+        metadata["coverart_maintype"] = self.maintype
+        metadata["coverart_comment"] = self.comment
+        if self.is_front:
+            metadata.add_unique("coverart_types", "front")
+        for cover_type in self.types:
+            metadata.add_unique("coverart_types", cover_type)
         filename = ScriptParser().eval(filename, metadata)
         if config.setting["ascii_filenames"]:
             if isinstance(filename, str):


picard/coverart/image.py
```diff
@@ -261,7 +262,15 @@ def maintype(self):
         # TODO: do something better than randomly using the first in the list
         return self.types[0]
 
-    def _make_image_filename(self, filename, dirname, metadata):
+    def _make_image_filename(self, filename, dirname, _metadata):
+        metadata = Metadata()
+        metadata.copy(_metadata)
+        metadata.add_unique("coverart_maintype", self.maintype)
```
phw: add_unique is not needed here. I will change this before we merge this

phw:
@zas Can you take a final look at my latest changes before I merge it?

zas:
> @zas Can you take a final look at my latest changes before I merge it?

Sure, LGTM. but rebase before merge.

phw:
@zas As a side note I find this setting of requiring rebasing before merge a bit annoying.  Because it means once any PR has  been merged all other PRs get stale, requiring rebasing. And rebasing causes reviews to become reset.

IMHO this causes more trouble than it helps.

samj1912:
Rebase before merge is a good idea. The stale review dismissal I get it's annoying. But it makes sure that the person merging it isn't the person who made the PR (which is good IMHO). In extreme cases, any admin can push without review approval.

phw:
@samj1912 But in the end it is an admin doing the rebase probably anyway. Because if you tell a contributor to rebase, they do so, other things gets merged, you tell again to rebase, they do so, other things again get into master before merging, etc. this becomes really annoying for the contributor.

samj1912:
Well its either that or adding a merge commit. I tend to alternate depending on how big the PR is but having so many merge commits clutters up the tree.

PR #928, opened 2018-08-24
PICARD-1245: Add an option to save grouping and work compatible with iTunes

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
For ID3 Picard is currently saving the `grouping` tag to `TIT1` and `work` to `TXXX:Work`. Originally iTunes actually used `TIT1` for grouping, but later repurposed `TIT1` to be used for the work name. For `grouping` they introduced the non-standard frame `GRP1`.

The result is that files tagged with Picard will show up wrong in iTunes: If you save the `grouping` tag it will instead be displayed as work name and saving `work` will actually not show up in iTunes at all.

The re-purposing of `TIT1` makes it difficult to deal with this in a generic manner. A lot of software followed Apple's original approach to use `TIT1` for grouping, so we can not simply change this to `GRP1`. Also using `TIT1` seems to be closer to the ID3 standard.

# Solution

This PR adds an option `itunes_compatible_grouping` to Tags -> Tag compatibility to change how Picard saves the tags `grouping` and `work` to ID3:

| itunes_compatible_grouping | False (default) | True   |
| -------------------------------------|--------------------|---------|
| grouping                                | TIT1                | GRP1 |
| work                                      | TXXX:Work      | TIT1 |

The default is disabled, which equals the existing behavior. The following screenshots demonstrate the result in iTunes. With standard settings the `grouping` tag will actually be displayed as the work name in iTunes, the `work` tag will not be displayed:

standard-grouping

When enabling itunes_compatible_grouping both the work name and grouping will display correctly in iTunes:

itunes-compatible-grouping

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -100,7 +100,7 @@ class ID3File(File):
 
     __translate = {
         # In same sequence as defined at http://id3.org/id3v2.4.0-frames
-        'TIT1': 'grouping',
+        # 'TIT1': 'grouping', # Depends on itunes_compatible_grouping
         'TIT2': 'title',
         'TIT3': 'subtitle',
         'TALB': 'album',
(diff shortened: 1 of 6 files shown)


phw:
@zas Allright, I refactored test_formats.py a bit

PR #936, opened 2018-08-28
PICARD-1267: Fixed Exception on Windows 10 when setting locale

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Sometimes Picard crashes without further output on Windows 10. The only output is in the Windows event logs, and that doesn't tell much. The crash can be traced back to the call of `locale.setlocale`.

In the issue this is reported only for running from source, but I could also reproduce this when building an installer and running that. But at the same time recent builds (exactly same branch) from Appveyor work.

* JIRA ticket (_optional_): [PICARD-1267](https://tickets.metabrainz.org/browse/PICARD-1267)

# Solution

The attached change of using `locale + '.' + locale.getpreferredencoding()` fixes the issue for me.

This is also consistent with the non-win32 specific call to `setlocale` before.

I am not sure what the root cause is, because this worked on my system just yesterday and I never had this before. I did do cleanup the source dir though and did a clean build, maybe this changed something.

# Action

- [x] will test an Appveyor build of this change: Test successful

### DIFF
--- a/picard/i18n.py
+++ b/picard/i18n.py
@@ -43,15 +43,16 @@ def setup_gettext(localedir, ui_language=None, logger=None):
         from ctypes import windll
         try:
             current_locale = locale.windows_locale[windll.kernel32.GetUserDefaultUILanguage()]
+            current_locale += '.' + locale.getpreferredencoding()
             locale.setlocale(locale.LC_ALL, current_locale)
-        except KeyError:
+        except KeyError as e:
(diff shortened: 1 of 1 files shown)


phw:
Closing in favor of https://github.com/metabrainz/picard/pull/938

The code was messy, easier to clean it up in one go.

PR #942, opened 2018-08-29
CI: Build against Python 3.7

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

Run builds against Python 3.7

# Solution

# Action

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ dist: trusty
 language: python
 sudo: required
 python:
+- '3.7'
 - '3.6'
 - '3.5'
 cache:
--- a/.travis.yml
+++ b/.travis.yml
@@ -16,6 +16,7 @@ env:
   - NOSE="$PIP_INSTALL nose2"
   - COVERAGE="$PIP_INSTALL nose-cov"
   - CODACY="$PIP_INSTALL codacy-coverage"
+  - LIBDISCID="libdiscid0 libdiscid0-dev"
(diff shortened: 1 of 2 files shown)


phw:
@samj1912 What was the reason for explicitly disabling PyQt 5.11 and sticking with PyQt 5.10? Just asking because I realized I have PyQT 5.11 on both my main Linux box and my Windows VM, and I wonder whether that is the reason I fail to produce working installer builds for Windows (the bundled openssl libs seem to break something, from source I also can only run it when those libs are not present).

phw:
Builds successfully now. One downside is that Travis does not yet support Python 3.7 properly, see https://github.com/travis-ci/travis-ci/issues/9815 . So currently it is it's own entry in the build matrix and only builds against the default PyQt from pip (5.10 currently for us). We could add Python 3.7 multiple times if we want to test against different PyQt versions also.

samj1912:
@phw it was a pyqt issue in Windows. Some SSL problem.

phw:
> it was a pyqt issue in Windows. Some SSL problem.

Ok, makes sense why I had those troubles during testing then. I will downgrade on my Windows box, too, and see if I actually can produce successful local installer builds with it. My guess is that it is "just" a binary compatibility issue then. No problems so far running with PyQt 5.11 on Linux.

PR #943, opened 2018-09-02
PICARD-1326: FLAC: Preserve waveformatextensible_channel_mask when clearing tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
FLAC files save the channel mapping inside a tag `WAVEFORMATEXTENSIBLE_CHANNEL_MASK`. When "Clear existing tags" is enabled this tag will be removed, destroying the defined channel header in the file.

* JIRA ticket (_optional_): [PICARD-1326](https://tickets.metabrainz.org/browse/PICARD-1326)

# Solution
Since `WAVEFORMATEXTENSIBLE_CHANNEL_MASK` is not really about track metadata but actually encodes some technical details for proper playback of the audio data I think Picard should not really handle this tag but make sure to preserve it. Hence this patch will copy the value of this tag over from the file when clearing existing tags.

It also makes sure to read this data as a hidden variable.

# Action


zas:
LGTM, but is it possible to add a test for this ?

phw:
Good point, I added a test.

FYI: I also experimented with actually reading this data and displaying it in a human readable form in the info dialog. Not sure about this yet, but maybe I will add some follow up PR. But let's focus on getting the issue this causes fixed first :D

PR #945, opened 2018-09-03
PICARD-1328: Fix loading OptimFROG files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1328](https://tickets.metabrainz.org/browse/PICARD-1328)

# Solution

# Action


phw:
Sorry, something got mixed up with my other branch. Obviously I was in a hurry, will fix this later

phw:
Now this works, on both Linux and Windows.

picard/formats/apev2.py
```diff
@@ -275,7 +275,8 @@ class OptimFROGFile(APEv2File):
 
     def _info(self, metadata, file):
         super()._info(metadata, file)
-        if file.filename.lower().endswith(".ofs"):
+        # mutagen.File.filename can be either a bytes or str object
+        if str(file.filename.lower()).endswith(".ofs"):
```
samj1912: If it can be a bytes object, it's better to us the builtin string_ from picard utils.

phw: Good catch, this actually caused setting `~format` wrong on Linux. Fixed now and added tests

zas: > If it can be a bytes object, it's better to us the builtin string_ from picard utils.

Unrelated to this PR, but i'm a bit bothered about `string_()`:
- that's a builtin aliasing `picard.util.convert_to_string()`
- it does 2 tests before eventually returning `str(arg)` plus the call  to the method, that's pretty expensive if not strictly needed
- explicit calls to conversion functions are better as they don't hide what is going on

For example, among 88 occurences, there's https://github.com/metabrainz/picard/blob/master/picard/cluster.py#L230

```python
        self.lookup_task = self.tagger.mb_api.find_releases(self._lookup_finished,
            artist=self.metadata['albumartist'],
            release=self.metadata['album'],
            tracks=string_(len(self.files)),
            limit=QUERY_LIMIT)
```

If i'm not wrong `len()` will hardly return anything but an integer, so `str(len(self.files))` should do.

`string_()` builitin is, in fact, https://github.com/metabrainz/picard/blob/master/picard/util/__init__.py#L453-L468


So, instead using `string_()` builitin everywhere, i'd rather not use it at all, but instead import explicitely the function it hides, and carefully review calls to get rid of useless ones.

What do you think?

phw:
@zas: I agree with your general ideas around `string_()`. I would keep it in, because this PR actually is a good example for a use case where you actually need a function to convert transparently between str or bytes to str. But for other cases I would suggest as you said to decide on a case by case basis. My guess is that in the majority of cases we now what gets handled.

Also I would probably do away with the builtins magic and just let it be imported on demand. I'm fine with giving the export a shorter name, but `from picard.util import string_` is IMHO better then making it available by magic :D

PR #954, opened 2018-09-04
PICARD-1332: When unmatching a file restore deleted tags from origina…

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Normally if you move a matched file from a track back to unmatched files all tag changed done on the release get reset.

But if tags got deleted on the matched files removing it from the track does not reset the deletion,

* JIRA ticket (_optional_): [PICARD-1332](https://tickets.metabrainz.org/browse/PICARD-1332)

# Solution

Do not preserve the deleted tags when restoring `file.orig_metadata`

# Action


phw:
There is a related issue: If you explicitly delete metadata from a loaded release, and then assign file to it, the deletion will not be reflected on the release. I noticed this when trying to update the "norelease" plugin.

The straight forward way to fix this is instead of doing:

    self.metadata.deleted_tags = deleted_tags

in `File.copy_metadata()` delete each tag individually:

    for tag in deleted_tags:
        self.metadata.delete(tag)

Which looks more correct. That has the side effect that switching between release versions actually does mark tags as deleted, and that is a bit confusing. In my example I had matched files to a release which had no catalognumber, but the (purchased) files have a catalognumber in their original data.

Since the release has no catalogue number the catalognumber tag will be marked as deleted as soon as the release gets loaded and the release event data gets applied.

Without the above patch matching and saving the files preserves the catalognumber, as it happens with all other tags present in the files but not the release. With the patch the catalognumber will be marked as deleted on all matched files also. That's probably not what the user wanted :( Not sure yet how to solve this.

phw:
I think this works now as one would expect it to work. I would still recommend not to include it in Picard 2.0.4 but keep it for a later release. The deleted tags sticking with files removed from albums is not a big bug, and also the other behavior is subtle.

It totally breaks the "norelease" plugin though :D

picard/file.py
```diff
@@ -163,7 +163,7 @@ def _copy_loaded_metadata(self, metadata):
         self.orig_metadata = metadata
         self.metadata.copy(metadata)
 
-    def copy_metadata(self, metadata):
+    def copy_metadata(self, metadata, preservedeleted=True):
```
zas: Just a stylistic remark, `preservedeleted` should be renamed `preserve_deleted` for the sake of consistency.

zas:
> I would still recommend not to include it in Picard 2.0.4 but keep it for a later release.

I agree.

picard/file.py
phw: Fixed

phw:
I adressed @zas' comment and added unit tests for the `Metadata` changes. As I find it difficult to comprehend myself, here is a short description of the three changes this PR actually makes:

1. When file metadata gets restored (i.e. when it gets removed from a track) this will also restore the status of deleted tags. Previously all tags deleted while attached to the track where still deleted when unmatching the file.

2. Setting metadata to empty by `metadata["sometag"] = ""` will no longer mark the tag as deleted, if it was not already present. E.g. this will no longer mark a deleted tag:

    ```
    m = Metadata()
    m["catalognumber"] = ""
    ```

    Previous to this PR the above caused the tag to be marked as deletd, even though it never was there in the first place. That caused all empty data loaded from MB.org to be marked as a deleted tag on album and tracks

3. Updating file metadata with `file.copy_metadata(metadata)` will also take into account all tags marked as deleted in `metadata`

phw:
I have been using this in different cases for a while now and it looks good, so I'm going to merge this. Turns out these fixes also partially address [PICARD-1112](https://tickets.metabrainz.org/browse/PICARD-1112), I will post a follow up PR that addresses the rest.

PR #957, opened 2018-09-05
Fix isort config to not skip __init__.py files (but few)

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Some files were forgotten during last isort pass, because isort ignores `__init__.py` files per default.

# Solution

Fix config file to include those but few exceptions.

# Action

### DIFF
--- a/.isort.cfg
+++ b/.isort.cfg
@@ -19,5 +19,6 @@ multi_line_output=3
 order_by_type=True
 sections=FUTURE,STDLIB,THIRDPARTY,QT,LOCALFOLDER,RESOURCES,PICARD_UI
 skip_glob=**/picard/ui/ui_*.py
-skip=./tagger.py,picard/const/attributes.py,picard/const/countries.py,picard/resources.py,scripts/picard.in
+skip=./tagger.py,picard/const/attributes.py,picard/const/countries.py,picard/resources.py,scripts/picard.in,picard/const/__init__.py,picard/formats/__init__.py,picard/coverart/providers/__init__.py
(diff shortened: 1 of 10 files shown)


.isort.cfg
```diff
@@ -19,5 +19,6 @@ multi_line_output=3
 order_by_type=True
 sections=FUTURE,STDLIB,THIRDPARTY,QT,LOCALFOLDER,RESOURCES,PICARD_UI
 skip_glob=**/picard/ui/ui_*.py
-skip=./tagger.py,picard/const/attributes.py,picard/const/countries.py,picard/resources.py,scripts/picard.in
+skip=./tagger.py,picard/const/attributes.py,picard/const/countries.py,picard/resources.py,scripts/picard.in,picard/const/__init__.py,picard/formats/__init__.py
```
phw: What's the issue with `formats/__init__.py` and `const/__init__.py`? Shouldn't it be possible to sort the imports to the start, too?

phw:
Sorry for breaking the tests, we need https://github.com/metabrainz/picard/pull/956 to fix that

.isort.cfg
zas: I just excluded problematic ones for now, either due to style or just crappy coding like caa providers

zas:
> Sorry for breaking the tests, we need #956 to fix that

Yes, i'll rebase once  #956 is in master

PR #958, opened 2018-09-06
Deprecate string_() and related cleanup

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

About string_():
- that's a builtin aliasing `picard.util.convert_to_string()`
- it does 2 tests before eventually returning `str(arg)` plus the call to the method, that's pretty expensive if not strictly needed
- in many cases, it is unneeded or equivalent to `str()`
- explicit calls to conversion functions are better as they don't hide what is going on
- it was introduced in https://github.com/metabrainz/picard/pull/664 to help the move to Python3

References:
https://github.com/metabrainz/picard/pull/945#discussion_r214912848
https://github.com/metabrainz/picard/pull/945#issuecomment-418523558

# Solution

Use `str()` or `bytes(arg).decode()` or drop it, depending on the case.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -130,7 +130,7 @@ class File(QtCore.QObject, Item):
         if self.state != File.PENDING or self.tagger.stopping:
             return
         if error is not None:
-            self.error = string_(error)
+            self.error = str(error)
             self.state = self.ERROR
             from picard.formats import supported_extensions
             file_name, file_extension = os.path.splitext(self.base_filename)
@@ -153,7 +153,7 @@ class File(QtCore.QObject, Item):
         if 'tracknumber' not in metadata:
             tracknumber = tracknum_from_filename(self.base_filename)
             if tracknumber != -1:
-                tracknumber = string_(tracknumber)
+                tracknumber = str(tracknumber)
                 metadata['tracknumber'] = tracknumber
                 if metadata['title'] == filename:
                     stripped_filename = filename.lstrip('0')
@@ -264,7 +264,7 @@ class File(QtCore.QObject, Item):
             return
         old_filename = new_filename = self.filename
         if error is not None:
-            self.error = string_(error)
+            self.error = str(error)
             self.set_state(File.ERROR, update=True)
         else:
             self.filename = new_filename = result
@@ -425,8 +425,9 @@ class File(QtCore.QObject, Item):
         """Move extra files, like playlists..."""
         old_path = os.path.dirname(old_filename)
         new_path = os.path.dirname(new_filename)
-        patterns = config.setting["move_additional_files_pattern"]
-        patterns = [string_(p.strip()) for p in patterns.split() if p.strip()]
+        patterns = [p for p in [p.strip() for p in
+                                config.setting["move_additional_files_pattern"].split()]
+                    if p]
         try:
             names = os.listdir(old_path)
         except os.error:
@@ -658,7 +659,7 @@ class File(QtCore.QObject, Item):
             release=metadata['album'],
             tnum=metadata['tracknumber'],
             tracks=metadata['totaltracks'],
-            qdur=string_(metadata.length // 2000),
+            qdur=str(metadata.length // 2000),
             isrc=metadata['isrc'],
             limit=QUERY_LIMIT)
 
--- a/picard/file.py
+++ b/picard/file.py
@@ -425,16 +425,16 @@ class File(QtCore.QObject, Item):
         """Move extra files, like playlists..."""
         old_path = os.path.dirname(old_filename)
         new_path = os.path.dirname(new_filename)
-        patterns = [p for p in [p.strip() for p in
-                                config.setting["move_additional_files_pattern"].split()]
-                    if p]
         try:
             names = os.listdir(old_path)
         except os.error:
             log.error("Error: {} directory not found".naming_format(old_path))
             return
         filtered_names = [name for name in names if name[0] != "."]
-        for pattern in patterns:
+        for pattern in config.setting["move_additional_files_pattern"].split():
+            pattern = pattern.strip()
+            if not pattern:
+                continue
             pattern_regex = re.compile(fnmatch.translate(pattern), re.IGNORECASE)
             file_names = names
             if pattern[0] != '.':
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -264,7 +264,7 @@ def format_file_info(file_):
     if '~sample_rate' in file_.orig_metadata:
         info.append((_('Sample rate:'), '%s Hz' % file_.orig_metadata['~sample_rate']))
     if '~bits_per_sample' in file_.orig_metadata:
-        info.append((_('Bits per sample:'), string_(file_.orig_metadata['~bits_per_sample'])))
+        info.append((_('Bits per sample:'), str(file_.orig_metadata['~bits_per_sample'])))
     if '~channels' in file_.orig_metadata:
         ch = file_.orig_metadata['~channels']
         if ch == '1':
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -418,7 +418,7 @@ def build_qurl(host, port=80, path=None, queryargs=None):
     if queryargs is not None:
         url_query = QtCore.QUrlQuery()
         for k, v in queryargs.items():
-            url_query.addQueryItem(k, string_(v))
+            url_query.addQueryItem(k, str(v))
         url.setQuery(url_query)
     return url
 
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -453,7 +453,7 @@ def union_sorted_lists(list1, list2):
     return union
 
 
-def convert_to_string(obj):
+def __convert_to_string(obj):
     """Appropriately converts the input `obj` to a string.
 
     Args:
@@ -471,6 +471,15 @@ def convert_to_string(obj):
         return str(obj)
 
 
+def convert_to_string(obj):
+    from picard import log
+    log.warning("string_() and convert_to_string() are deprecated, do not use")
+    return __convert_to_string(obj)
+
+
+builtins.__dict__['string_'] = convert_to_string
+
+
 def htmlescape(string):
     return html.escape(string, quote=False)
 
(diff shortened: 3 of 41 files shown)


zas:
@phw : there are more occurences, please review carefully
I'll get rid of `string_()| and `convert_to_string()` at the end

zas:
> No proper review yet, just some comments:

Good points.
Still i want to reduce the use (especially when totally unneeded).

picard/ui/infodialog.py
```diff
@@ -264,15 +264,15 @@ def format_file_info(file_):
...
+        info.append((_('Bits per sample:'), str(file_.orig_metadata['~bits_per_sample'])))
     if '~channels' in file_.orig_metadata:
         ch = file_.orig_metadata['~channels']
         if ch == 1:
             ch = _('Mono')
         elif ch == 2:
             ch = _('Stereo')
         else:
-            ch = string_(ch)
+            ch = str(ch)
```
phw: I looked at this recently, this code is broken right now. `ch` is always a `str` so the comparisson with integers fails. We don't need the `str()` here, but the comparisson needs to be changed to strings (`"1"`)

picard/util/__init__.py
```diff
@@ -486,7 +495,7 @@ def load_json(data):
         dict: Response data as a python dict
 
     """
-    return json.loads(convert_to_string(data))
+    return json.loads(bytes(data).decode())
```
phw: Can we really be sure to always get byte arrays here? I have checked the plugins, the current plugins that use it also just use it for webservice results. But this function actually could be handy for also loading JSON from string. I would keep it flexible to accept both.

picard/util/xml.py
```diff
@@ -71,7 +71,7 @@ def parse_xml(response):
 
             for i in range(attrs.count()):
                 attr = attrs.at(i)
-                node.attribs[_node_name(attr.name())] = string_(attr.value())
+                node.attribs[_node_name(attr.name())] = str(attr.value())
```
phw: I don't think this is needed, attr.value() should be a string

test/test_browser.py
```diff
@@ -32,7 +32,7 @@ def test_entity_lookups(self, mock_open):
         for i, type_ in enumerate(lookups):
             lookups[type_]['function']("123")
 
-            url = urlparse(string_(mock_open.call_args[0][0]))
+            url = urlparse(str(mock_open.call_args[0][0]))
```
phw: Not needed

picard/webservice/api_helpers.py
```diff
@@ -249,10 +249,10 @@ def submit_acoustid_fingerprints(self, submissions, handler):
         path_list = ['submit']
         args = {'user': config.setting["acoustid_apikey"]}
         for i, submission in enumerate(submissions):
-            args['fingerprint.%d' % i] = string_(submission.fingerprint)
-            args['duration.%d' % i] = string_(submission.duration)
-            args['mbid.%d' % i] = string_(submission.recordingid)
+            args['fingerprint.%d' % i] = str(submission.fingerprint)
+            args['duration.%d' % i] = str(submission.duration)
+            args['mbid.%d' % i] = str(submission.recordingid)
```
phw: Not sure here, but likely no `str()` needed

picard/file.py
```diff
@@ -419,8 +419,9 @@ def _move_additional_files(self, old_filename, new_filename):
         """Move extra files, like playlists..."""
         old_path = os.path.dirname(old_filename)
         new_path = os.path.dirname(new_filename)
-        patterns = config.setting["move_additional_files_pattern"]
-        patterns = [string_(p.strip()) for p in patterns.split() if p.strip()]
+        patterns = [p for p in [p.strip() for p in
+                                config.setting["move_additional_files_pattern"].split()]
+                    if p]
```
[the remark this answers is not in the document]
zas: Yes, but the new one get rid of an extra `strip()` call

picard/ui/infodialog.py
zas: But do we need such fancy conversion to "Mono" or "Stereo" at all ? Just displaying the number of channels is ok, that's just informative.

picard/util/__init__.py
samj1912: This is one of the reasons I introduced string_

samj1912: There will also be problems with the web browser module.

zas: Ok, i'll fix it, especially since the description specifies it ... ;)

picard/file.py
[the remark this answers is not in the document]
phw: I think this double list comprehension is worse than an extra `strip()`. But if you really want to get rid of the double `strip` I would use a normal `for p in ...` loop

picard/util/__init__.py
zas: > There will also be problems with the web browser module.

Which problems exactly?

picard/file.py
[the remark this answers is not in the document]
samj1912: A very useful case for assignment operator from py3.7 :)

picard/util/__init__.py
phw: It should mostly get called with `QByteArray`, because it is always used with the results from the webservice module. But I would still make it also accept strings in this case because of unknown plugin usage.

picard/ui/infodialog.py
phw: Maybe not, even though I find it is a nice touch. I also looked into visualizing the channel mapping data we find in some FLACs, and that would give more verbose strings (like "5.1 (side)") we could display here. So I would for now keep it but fix it.

picard/file.py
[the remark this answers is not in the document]
zas: See https://github.com/metabrainz/picard/pull/958/commits/1301301eda0810e1ce6a80ed9fe38f55e7cf7cf2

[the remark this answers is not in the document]
samj1912: A better way of doing this is

patterns = filter(bool, map(str.strip, config.setting["move_additional_files_pattern"].split()))

samj1912:
@zas, suggested a better way to write that strip expression in the comment above. Seems to be hidden now.

picard/file.py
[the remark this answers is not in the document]
phw: I think @zas solved this nicely in https://github.com/metabrainz/picard/pull/958/files/dbceae9e59e2704ce99d8189e535795d2af31cf7..1301301eda0810e1ce6a80ed9fe38f55e7cf7cf2

Three simple, easy to read lines

picard/ui/infodialog.py
zas: Simplified and fixed by https://github.com/metabrainz/picard/pull/958/commits/0915fad10f66053c5a6bbaa4d2be0d773bce5c27

picard/util/__init__.py
samj1912: The web browser module is used in different contexts iirc. Some of the times it has to handle bytes, sometimes qbytesarray and sometimes string. Check it's usage instances, idr all of it. But it was one of the primary reasons I introduced this dynamic conversion.

zas: > But I would still make it also accept strings in this case because of unknown plugin usage.

Whatever we change, that's likely to break an "unknown plugin" ... and i think it's a bad thing.
We definitively need to find a good solution for this.

For now, i'd rather bump plugin API version, rather that not doing change in core code...

zas:
> @zas, suggested a better way to write that strip expression in the comment above. Seems to be hidden now.

Wasn't your suggestion based on a Python 3.8 feature or i missed the comment ?

picard/util/__init__.py
samj1912: I frankly think this is premature optimization to break compatibility. Did we actually measure the performance degradation due to this call in normal picard usage and if it was significant? If it is not, I'd rather keep it in the api and webbrowser module. Also just by keeping the built-in I don't think we are hurting anybody, it's just 4 loc. We can add a warning to not use it in the future.

samj1912:
patterns = filter(bool, map(str.strip, config.setting["move_additional_files_pattern"].split()))

picard/util/__init__.py
phw: I don't think it is primarily a performance issue, but rather the frequent use of `string_` hides what data you are dealing with. And because at some point you stop knowing what you are handling you start adding string_ to even more code.

But for this specific case of `load_json` I agree that we should keep the string compatibility.

phw:
> patterns = filter(bool, map(str.strip, config.setting["move_additional_files_pattern"].split()))

Nah, keep the currently submitted code. Or if you must use this at least put it on two lines like

    patterns = config.setting["move_additional_files_pattern"]
    patterns = filter(bool, map(str.strip, patterns.split()))

picard/util/__init__.py
zas: Apart "unknown plugin" case, i don't think `json_load()` is used with anything but `bytes`.

So i'd rather move the "compatibility" to the call in `picard/plugin.py`.

And no, that's not a performance issue, just a bad idea that needs to be addressed before it's too late.

phw:
> You missed the point, it breaks nothing, since it was broken.

Yes, but I agree we need at least a ticket for it. And I still think it added a nice touch to the UI :( I always hoped that the info dialog would become more informative

zas:
> patterns = filter(bool, map(str.strip, config.setting["move_additional_files_pattern"].split()))

That's cheating, you just hide loops with extra function calls ;)
Currently submitted code addresses the issue and remove unneeded loops (my first patch was mainly because i missed the second loop actually using `patterns`), and is actually readable imho.

picard/util/__init__.py
phw: You don't have to add anything to the call in `picard/plugin.py`, that one is fine. I'm also ok with leaving this as is, since all calls to it (including inside plugins) I could find actually pass bytes.

zas:
@phw @samj1912 : please review current iteration, thanks

picard/ui/infodialog.py
zas: It will be fixed in another PR.
https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-1340

PR #963, opened 2018-09-07
PICARD-1340: fix display of Mono/Stereo for channels in info dialog

zas:
ch is a string

# Summary

In Info Dialog, Picard displayed Mono/Stereo for 1 or 2 channels, it was broken at some point

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Comparing int to string doesn't work, and thus the number of channels was displayed in all cases

* JIRA ticket (_optional_): [PICARD-1340](https://tickets.metabrainz.org/browse/PICARD-1340)

# Solution

Compare to "1" and "2"

# Action

### DIFF
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -267,12 +267,10 @@ def format_file_info(file_):
         info.append((_('Bits per sample:'), string_(file_.orig_metadata['~bits_per_sample'])))
     if '~channels' in file_.orig_metadata:
         ch = file_.orig_metadata['~channels']
-        if ch == 1:
+        if ch == '1':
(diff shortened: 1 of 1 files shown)


phw:
I'm glad this just gets fixed and not entirely removed. I already experimented showing more detailed data for multi-channel FLAC files here, maybe we can generalize this for more formats.

phw:
But let's get #958 merged first, because this will conflict and needs an update then

zas:
> But let's get #958 merged first, because this will conflict and needs an update then

I rebased #958 instead

PR #965, opened 2018-09-12
PICARD-518: Add a tip to sliders in Options > Metadata > Preferred Re…

zas:
…leases

- Display a tip with actual value
- 5 by 5 step (values from 0 to 100)
- page step (keyboard only set to 25)

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

There's no way to know in which direction sliders should be moved.

* JIRA ticket (_optional_): [PICARD-518](https://tickets.metabrainz.org/browse/PICARD-518)

# Solution

Not perfect solution, but display a tip over the slider handle with the actual number, and ease setting the value by moving 5 by 5.

The meaning of the number isn't really obvious but at least it shows when you increase or decrease the value.

capture du 2018-09-12 14-52-58

# Action


samj1912:
Rebase for test fix?

PR #970, opened 2018-09-16
PICARD-1344: Add $delete function

phw:
# Summary
There is no way currently for Picard to really delete a tag via scripting. You can unset a variable, but not add a tag to the deleted tags so it will actually be deleted.

This PR ads a `$delete` function that does that.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1344](https://tickets.metabrainz.org/browse/PICARD-XXX)

# Solution

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -486,6 +486,18 @@ def func_unset(parser, name):
     return ""
 
 
+def func_delete(parser, name):
+    """
+    Deletes the variable ``name``.
+    This will unset the tag with the given name and also mark the tag for
+    deletion on save.
+    """
+    if name.startswith("_"):
+        name = "~" + name[1:]
+    parser.context.delete(name)
+    return ""
+
+
 def func_set(parser, name, value):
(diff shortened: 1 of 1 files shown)


picard/script.py
```diff
@@ -486,6 +486,18 @@ def func_unset(parser, name):
     return ""
 
 
+def func_delete(parser, name):
+    """
+    Deletes the variable ``name``.
+    This will unset the tag with the given name and also mark the tag for
+    deletion on save.
+    """
+    if name.startswith("_"):
```
samj1912: These 2 lines repeat all over this file. We could possibly modify it before calling all the functions?

phw: @samj1912 Agreed, I added https://tickets.metabrainz.org/browse/PICARD-1347

PR #971, opened 2018-09-17
PICARD-1346: Fix exception with move additional files

phw:
When multiple patterns for move additional files matched the same file an exception was thrown when attempting to move the files again after it already had been moved.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1346](https://tickets.metabrainz.org/browse/PICARD-1346)

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -427,11 +427,11 @@ class File(QtCore.QObject, Item):
         old_path = os.path.dirname(old_filename)
         new_path = os.path.dirname(new_filename)
         try:
-            names = os.listdir(old_path)
+            names = set(os.listdir(old_path))
         except os.error:
             log.error("Error: {} directory not found".naming_format(old_path))
             return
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -442,6 +442,7 @@ def _move_additional_files(self, old_filename, new_filename):
                 file_names = filtered_names
             for old_file in file_names:
                 if pattern_regex.match(old_file):
+                    file_names.remove(old_file)
```
samj1912: That makes the loop O(n2). Can we make file_names a set instead so that it's O(n)?

phw:
@samj1912 Fixed to use sets instead. Also realized there was still the edge case of two list being used for the files (`names` and `filtered_names`). Once a pattern matched this file does not need to be considered for any match, not matter which list is used.

PR #972, opened 2018-09-17
PICARD-1350: fix drag'n'drop on cover art box

zas:
It wasn't working in most cases, mainly due to QByteArray vs bytes mismatch (actually since a long time)

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Dropping cover art image on cover art box doesn't work in many cases.

* JIRA ticket (_optional_): [PICARD-1350](https://tickets.metabrainz.org/browse/PICARD-1350)

# Solution

Tested working cases (linux):

- dropping local file (file://...) png & jpeg
- dropping remote url file (http://...) png & jpeg from Chrome & firefox
- dropping image url from Google Images, works from Chrome, doesn't from Firefox

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -371,7 +371,7 @@ class CoverArtBox(QtWidgets.QGroupBox):
             path = url.path()
             if url.hasQuery():
                 query = QtCore.QUrlQuery(url.query())
-                queryargs = {k:v for k, v in query.queryItems()}
+                queryargs = dict(query.queryItems())
             else:
                 queryargs = {}
             if url.scheme() == 'https':
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -368,16 +368,19 @@ class CoverArtBox(QtWidgets.QGroupBox):
             self.load_remote_image(url, None, fallback_data)
 
         if url.scheme() in ('http', 'https'):
-            path = url.path(QtCore.QUrl.FullyEncoded)
+            path = url.path()
             if url.hasQuery():
-                path += '?' + url.query(QtCore.QUrl.FullyEncoded)
+                query = QtCore.QUrlQuery(url.query())
+                queryargs = {k:v for k, v in query.queryItems()}
+            else:
+                queryargs = {}
             if url.scheme() == 'https':
                 port = 443
             else:
                 port = 80
             self.tagger.webservice.get(url.host(), url.port(port), path,
                                   partial(self.on_remote_image_fetched, url, fallback_data=fallback_data),
-                                  parse_response_type=None,
+                                  parse_response_type=None, queryargs=queryargs,
                                   priority=True, important=True)
         elif url.scheme() == 'file':
             path = os.path.normpath(os.path.realpath(url.toLocalFile().rstrip("\0")))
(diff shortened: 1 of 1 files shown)


phw:
Ok, first round of testing on Windows and Linux (macOS will follow):

1. Drag'n'drop in general works for me on current master on both Linux and Windows, but it prints warnings "Can't load image: Not enough data". The images are displayed fine, though, even with correct width, height etc. I did not investigate further, but your branch fixes this.

2. Images loaded from URLs work only, if the URL does not contain query strings. This is due to some crazy double encoding of paths and query parameters, I have a fix at https://github.com/phw/picard/commit/f701faeb13301a3878f4af898859e60d91127cba you can pull

3. Loading from Google images works only on Chrome, it passes the data differently (no URL, but actual image data). My fix above addresses this also and makes the image drop work in other browsers (tested Firefox on Linux, Edge on Windows). It works even better, since it will load the actual source image, while Chrome only provides the displayed thumbnail

4. On Windows dragging an image file on the coverartbox adds this file to the list of images (so dropping multiple times will result in multiple cover images), on Linux it replaced the existing one. I don't know why, is there a setting for this?

Apart from this regarding this fix definitely a +1. Maybe pull my change above into this PR as well.

zas:
>     4\. On Windows dragging an image file on the coverartbox adds this file to the list of images (so dropping multiple times will result in multiple cover images), on Linux it replaced the existing one. I don't know why, is there a setting for this?

Yes, you can switch Replace/Add by right click on cover art box, there are options in context menu to switch the behavior.

I added your fix for browser dnd, thanks.

picard/ui/coverartbox.py
```diff
@@ -75,33 +75,40 @@ def dropEvent(self, event):
...
+        if dropped_data:
+            try:
+                mime = imageinfo.identify(dropped_data)[2]
+                if mime in ('image/jpeg', 'image/png'):
+                    accepted = True
+                    log.debug("Dropped %s mime data (%d bytes)", mime,
+                              len(dropped_data or ''))
+                    self.image_dropped.emit(QtCore.QUrl(''), dropped_data)
+            except imageinfo.IdentificationError:
+                pass
```
phw: I think we should log this as a warning, at minimum debug. We have enough totally ignored exceptions already :D

picard/ui/coverartbox.py
```diff
@@ -357,20 +364,23 @@ def fetch_remote_image(self, url, fallback_data=None):
...
+        if fallback_data:
             self.load_remote_image(url, None, fallback_data)
 
         if url.scheme() in ('http', 'https'):
-            path = url.path(QtCore.QUrl.FullyEncoded)
+            path = url.path()
             if url.hasQuery():
-                path += '?' + url.query(QtCore.QUrl.FullyEncoded)
+                query = QtCore.QUrlQuery(url.query())
+                queryargs = {k:v for k, v in query.queryItems()}
```
phw: I know this is my code and a bit nit-picking, but I think `k: v` instead of `k:v` reads nicer. In [PEP 274](https://www.python.org/dev/peps/pep-0274/), which introduced dict comprehension, they even use `k : v` (but not consistently).

picard/ui/coverartbox.py
```diff
@@ -414,10 +425,14 @@ def load_remote_image(self, url, mime, data):
...
         if config.setting["load_image_behavior"] == 'replace':
             set_image = set_image_replace
+            debug_info = "Replacing with dropped %r in %r"
         else:
             set_image = set_image_append
+            debug_info = "Appending dropped %r to %r"
 
+        update = True
         if isinstance(self.item, Album):
+            log.debug(debug_info, coverartimage, self.item)
```
phw: This debug line gets repeated for every `isinstance` check, can we move it before the `if`?

picard/ui/coverartbox.py
zas: Addressed in 71737ed

picard/ui/coverartbox.py
zas: I agree for the style, but here the whole dict comprehension is useless, fixed in 
dac7c2

picard/ui/coverartbox.py
zas: Yup, i converted a (useless) warning to debug, and simplfied. See 
7dd6cf6

PR #975, opened 2018-09-19
PICARD-1339: Improve file removal performance

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
This is another attempt at addressing the performance issues Picard has when removing a lot of files.

The root cause was identified to be repeated calls to `imagelist.update_metadata_images()`. The previous attempt to fixing this was in https://github.com/metabrainz/picard/commit/d034e5c6fa7bee237892cba5265d0aedd2023d88 . This worked great for this single use case, but unfortunately missed some cases where files are also removed from a parent (album, cluster or track) and the image metadata needs to be updated.

This WIP PR is another attempt at fixing this. First I have reverted https://github.com/metabrainz/picard/commit/d034e5c6fa7bee237892cba5265d0aedd2023d88 , since I think we need a more generalized approach here which does not require special handling of parents on every `file.remove()` call.

Then as a first step I went to optimize `imagelist.update_metadata_images()` itself, which as it is now is just really inefficient. There is one side effect of this change: Since the updated code makes use of `set` internally the ordering of images for parents can change. But since we don't have a way to order images right now and make no guarantees on the order I think this is acceptable. Also if we will introduce ordering we can introduce some kind of sort attribute on `CoverArtImage`, which can be used for display.

Here are some performance results for removing a lot of unclustered files:

Picard 2.0.4:
3762 files removed in 72.710277 seconds

update_metadata_images optimized:
3762 files removed in 19.744973 seconds

with update_metadata_images delayed after removal:
3762 files removed in 0.238646 seconds

This shows optimizing `update_metadata_images` brings a big performance gain and should be done in any case (as this also has effect on other use cases). However we still want to optimize the number of calls to `update_metadata_images` on mass removal, ideally to be done only once per parent.

* JIRA ticket (_optional_): [PICARD-1339](https://tickets.metabrainz.org/browse/PICARD-1339)

# Solution

# Action


picard/util/imagelist.py
```diff
@@ -35,48 +35,51 @@ def __getitem__(self, k):
             return result
 
 
+class State:
```
zas: I'd prefer a longer name here, may be `ImageListState`, `State` is too general for such specific class

zas:
> Picard 2.0.4:
> 3762 files removed in 72.710277 seconds
> 
> update_metadata_images optimized:
> 3762 files removed in 19.744973 seconds
> 
> with update_metadata_images delayed after removal:
> 3762 files removed in 0.238646 seconds

Whoooa, impressive gain. I wonder if we have such other bottlenecks...

phw:
Some status update:

The current `update_metadata_images` is really a shotgun approach: Whenever anything is changed for a "parent", we rebuild the image list for the parent from the ground up. But we could do much better and optimize for the use cases where files are added / removed from the system. Adding is rather trivial, but removing is a bit more tricky. Nevertheless there are a couple of optimizations and shortcuts that are possible.

I have prototyped an updated implementation for the "remove a file and update cover art of parent use case". In the worst case this runs for my sample files as follows:

> 3762 files removed in 2.816399 seconds

But that basically is the case that a single cluster contains all those files, and all have basically different cover art. With optimizations, e.g. skip the iteration completely when there is no cover art for the parent or all files have same cover art, we get it down in many cases. E.g. the unclustered files cluster does not display the childrens cover art, so the update can basically be skipped:

> 3762 files removed in 0.717928 seconds

This is just a prototype for now. It needs some cleanup and extension to the use with Track and Album.

phw:
FYI: I have committed the reversal of https://github.com/metabrainz/picard/commit/d034e5c6fa7bee237892cba5265d0aedd2023d88 directly to master, since it is clear to me now that we will have a better solution for the issue with this PR. And since I still need some time to finalize it I think it is best to have a working master branch in the meantime.

phw:
What needs to be done:

- The removal function being extended to handle other parents than Cluster and update those classes to use it
- Probably add a similar function for adding cover art
- Cleaning up the unit tests
- If possible remove update_metadata_images. But currently I actually think we will have some cases where we still will need it which I won't address in this PR

picard/util/imagelist.py
```diff
@@ -35,73 +36,121 @@ def __getitem__(self, k):
...
-        self.orig_images = ImageList()
-        self.has_common_new_images = True
-        self.has_common_orig_images = True
-        self.first_new_obj = True
-        self.first_orig_obj = True
-        # The next variables specify what will be updated
-        self.update_new_metadata = False
-        self.update_orig_metadata = False
+def _update_state(obj, state):
+    is_first = True
```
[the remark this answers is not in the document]
zas: Still there ;)

picard/util/imagelist.py
```diff
@@ -35,73 +36,156 @@ def __getitem__(self, k):
...
+def _add_images(metadata, added_images):
+    if not added_images:
+        return
+
+    current_images = set(metadata.images)
+    if added_images == current_images:
+        return
+    else:
+        metadata.images = ImageList(current_images.union(added_images))
+        metadata.has_common_images = False
```
zas: You can write it:

```python
    if added_images != current_images:
        metadata.images = ImageList(current_images.union(added_images))
        metadata.has_common_images = False

```

picard/util/imagelist.py
```diff
@@ -35,73 +36,156 @@ def __getitem__(self, k):
...
+        return
+
+    current_images = set(metadata.images)
+
+    if metadata.has_common_images and current_images == removed_images:
+        return
+
+    common_images = True
+    previous_images = None
+    for source_metadata in sources:
```
zas: I think a comment explaining a bit what's the actual process could help, the code is clean but not really easy to grasp imho.
Perhaps just defining what are `source_images`, `previous_images`, `common_images`, `removed_images`, dunno

picard/util/imagelist.py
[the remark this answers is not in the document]
phw: Now it's finally gone :D

phw:
I will add some documentation to the code. I agree that this module would benefit from it a lot. I also struggled a bit with having variable names understandable without making them too long.

There are more intermediate and state variables here than I usually like, but they are needed to get this efficient while keeping it somewhat understandable :)

phw:
I added some documentation to the functions which should help a bit to clarify things. If this is ok I would suggest to merge this.

I have some improvement ideas to investigate, but I would prefer to do this once this is merged. One is that we can for sure improve `UnclusteredFiles`, because this special `Cluster` does not display its children's metadata and therefor does not need to call all those image update functions at all.

The other point would be to see if we can optimize the cases that currently rely on the `File.metadata_images_changed` signal. I think we can at least emit this signal only if we actually have changes to the images. Maybe we can even split into more specific `metadata_images_added` and `metadata_images_deleted` signals.

zas:
> If this is ok I would suggest to merge this.

LGTM

PR #976, opened 2018-09-20
WIP: PICARD-1313: Improve UI feedback regarding plugins management

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1313](https://tickets.metabrainz.org/browse/PICARD-1313)

# Solution

# Action


zas:
@phw can you give it a run and tell me if it improves things for you regarding PICARD-1313

phw:
Having the plugin enabled by default after installation definitely helps and is more inline with user expectations. Not sure about showing this "loaded" info behind the version number, I find this more confusing then helpful. What about putting the checkboxes in a separate column labeled "Activated"?

Another win for the usability would probably be fixing PICARD-1275 (allow reinstalling after uninstall).

A side note, more related to our recent discussion about issues with the current plugin system as a whole: In the recent bug reports I noted that quite a few people have every possible plugin (or at least a lot plugins) installed (and probably activated), even though some plugins clearly are for very specific use cases. That probably means people have the impression that the more plugins installed the better, because it will give them more functionality. But for Picard plugins this is not always true, since many plugins are not giving you **more** functionality but instead **changed** functionality. Maybe something to keep in mind for a future plugin system (also currently I have no idea how to better communicate these cases to the user).

zas:
> Not sure about showing this "loaded" info behind the version number, I find this more confusing then helpful

Removed in last iteration.

zas:
Update: this will be a complete rework of plugins UI, i want to fix all issues, and it appears the code is just not very friendly.

More to come, sorry.

picard/ui/options/plugins.py
```diff
@@ -100,103 +127,182 @@ def __init__(self, parent=None):
...
+
+    def find_item_by_plugin_name(self, plugin_name):
+        for item in self.items():
+            if plugin_name == item.plugin.module_name:
+                return item
+        return None
+
+    def selected_item(self):
+        try:
+            return self.ui.plugins.selectedItems()[0]
```
phw: I would use the `COLUMN_NAME` constant here again

phw: I would use the `COLUMN_NAME` here again instead of the magic `0`

picard/ui/options/plugins.py
```diff
@@ -100,103 +127,182 @@ def __init__(self, parent=None):
...
+            return item.plugin
+        else:
+            return None
 
     def save_state(self):
         header = self.ui.plugins.header()
         config.persist["plugins_list_state"] = header.saveState()
         config.persist["plugins_list_sort_section"] = header.sortIndicatorSection()
         config.persist["plugins_list_sort_order"] = header.sortIndicatorOrder()
-        try:
```
phw: If I see this correctly the saving and restoring of `plugins_list_selected` selected did not actually work, and in your branch it is fixed. But do we actually need it? While using I found it rather confusing that I was in the middle of the list when opening the options dialog again.

phw:
Regarding the UI itself I'm not entirely sure. Having plugins enabled after install is definitely a great improvement. Apart from this the two things already mentioned above would IMHO help a lot:

1. Have the activated checkbox in a separate "Activated" column
2. Fix the bug which does not allow you to reinstall a just uninstalled plugin (probably this is easier to fix now with the refactored code)

zas:
>     1. Pressing the Reload button crashes (see stacktrace below)

Yes, i was aware of it, should be fixed



>     2\. When you uninstall a plugin, leave the options page and load the options again the plugin's "uninstall" button is active again and can be clicked again. I think this is because you moved the states out of the plugins. Looks like `PluginManager.remove_plugin` needs to actually remove the plugin from its list
>     2\. Fix the bug which does not allow you to reinstall a just uninstalled plugin (probably this is easier to fix now with the refactored code)


Well, the whole uninstall/reinstall process doesn't work great at all, i'll rework this stuff, so one can uninstall and reinstall (online) plugins.




>     1. Have the activated checkbox in a separate "Activated" column

I'm not sure a separate column will change much, but i'll try

zas:
Still work on progress, many bits aren't working as they should yet, i'll try to address all issues within next days.

picard/ui/options/plugins.py
zas: >  But do we actually need it? While using I found it rather confusing that I was in the middle of the list when opening the options dialog again.

Me too, i think i'll get rid of it, i think it was meant for the Reload List option, but it will be less disturbing if the list was at top each time

phw: Did not think about plugin reload, for that it actually could be useful. But still that should not require any real persisting, just temporarily remembering the position.

zas: > Did not think about plugin reload, for that it actually could be useful. But still that should not require any real persisting, just temporarily remembering the position.

That's what i did, check last iteration.

phw:
One possible UI improvement would making the version columns smaller, it is really very wide for the content it displays. Maybe this would even solve [PICARD-1227](https://tickets.metabrainz.org/browse/PICARD-1227)

zas:
image

Current look, please test and comment

rdswift:
Here's how the current version looks under Windows 7.  It took a little while for me to figure out the meaning of the different fonts, bold text and two columns of buttons, but it looks like all the needed functionality and indication is all there.  Just different and will take a bit of getting used to.  Does look a bit cleaner overall.

The button graphic for ***install*** doesn't seem intuitive to me, but I don't know what to suggest as a replacement.  I'll think on that a bit.

image

zas:
> Here's how the current version looks under Windows 7.

Thanks, can you try last version with even less buttons ? ;)

zas:
image

phw:
In general I like the new look. But I experimented with 4 little changes, and I find the results a bit more pleasing:

- Move the version into a separate column again (but scaled to content size). I think that makes this information more visible and makes the name column to be less cluttered.
- Do not use italic font to mark installable plugins. I think this is adding visual noise and another piece of information the user has try to decode basically by guessing.
- Switch the buttons and moved uninstall to the right. I have no proper reason for this other than "I personally like it better", so ignore this idea if you disagree, I won't argue about this :)
- Use a down arrow button for install. I think this matches the general "download" symbolic better, but you said the icons are rather placeholders anyway.

Here is the result:

screenshot-2018-09-28-13-04-23

rdswift:
@zas, here's how the latest appears in Windows 7.

I'm not sure I like having separate buttons for ***Download and upgrade...*** and ***Download and install...*** in different columns.  I think it was clearer before when you had three columns.  Perhaps column 1 could be ***Enabled*** (check mark) or ***Disabled*** (cross / X) and blank if not installed.  Column 2 could be ***Install*** (down arrow) or ***Upgrade*** (refresh / recycle) and blank if installed but no update is available.  Finally, column 3 would be ***uninstall*** (trash can) with the button greyed out but still visible if the plugin is not installed (as you have it now).

My reasoning...  Column 1 is really an enabled status indicator which can be toggled, and only applies once a plugin has been installed.  Keeping with the status indicator notion, perhaps the image could be a warning symbol if the plugin fails to load properly?  Column 2 is an availability indicator which initiates an action (install or update).  Column 3 is a different action which is still a valid option even if an update is available, thus it can't really be combined with Column 2.  The reason I would like the buttons in Column 3 to always be visible is because it nicely frames the right hand side of the display (aesthetically pleasing).

I really like the tooltips that you provide, especially regarding the updates.  That's good information.  I sort of like the version number in a separate column as @phw has shown.  Somehow it just looks cleaner that way.

Of course, these are just my thoughts and opinions.  I can live with whatever you decide as long as the functionality is there.

image

zas:
>  4 little changes,

Can you provide a patch to this tree ? ;)

rdswift:
I can live with it.  One question though...  If there was an update available for a plugin, where would that show up?  It doesn't really matter, but I'm just curious.

zas:
> I can live with it. One question though... If there was an update available for a plugin, where would that show up? It doesn't really matter, but I'm just curious.

image

rdswift:
Looks good to me.  Thanks.

zas:
@phw can you share a final review and tell if the patch is mature enough to be merged ?
@rdswift  same^^

rdswift:
Looks good to me, and everything seems to work properly.  There does seem to be a display problem after uninstalling a plugin and then re-opening the Options page without closing the Picard session.  Hopefully the following help explain.

image

image

image

image

Furthermore, if you click ***Make It So!*** at Image 3, which you would think would save the first plugin as installed and enabled, then restart Picard, it still shows up as shown in Image 4 (not installed).

Also, I do have a (perhaps dumb) question though.  Does `_('some string')` return a string?  If so, should the lines like `error = _("Failed loading zipped plugin %r") % name` be changed to something like `error = _("Failed loading zipped plugin {0}").format(name)` as per Python 3?

zas:
@rdswift : the uninstall icon issue should be fixed by d2eace9e9af47, please re-test

> 
> Also, I do have a (perhaps dumb) question though. Does `_('some string')` return a string? If so, should the lines like `error = _("Failed loading zipped plugin %r") % name` be changed to something like `error = _("Failed loading zipped plugin {0}").format(name)` as per Python 3?

Both are valid in Python3. We still use the old % formatting for historical reasons, and we should move to format(), but this will mean to change almost all formatted strings which means a lot of translations work.

So, yes we should, but no, we won't yet because of translations.

rdswift:
> the uninstall icon issue should be fixed by [d2eace9](https://github.com/metabrainz/picard/commit/d2eace9e9af4730d77f3a7f4e5069f29852b2c1a), please re-test

I just tried it and it all works as expected.  Thanks.

> So, yes we should, but no, we won't yet because of translations.

Thanks for the explanation.  That makes perfect sense to me.  I just curious because that was one of the things that you asked me to change in one of my earlier PRs (to use `.format()` rather than `%`).  In my case, it was all new code and strings so there would not have been any existing translation.  Thanks again.

PR #977, opened 2018-09-21
i18n: Enable languages with more than 40% translation

phw:
# Summary

Enabled UI languages that according to Transifex have more than 40% of their strings translated.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


phw:
We should discuss this if we want to enable even less translated translations. I'm actually in favor of it, because if translations are missing that might motivate people to help with the translation. But the intention of the explicit whitelist was not to give the users false impression  that Picard is available in their language if it is nearly untranslated indeed. Opinions?

zas:
> Opinions?

Well, not sure what's the best here, hiding existing translations (even very incomplete) isn't great, but showing them may be deiciving for the user.

Perhaps we can have >40% translated shown by default, and add Other languages.... to the menu to display all.

But it can be done in another PR.

LGTM for this one.

Sophist-UK:
>> We should discuss this if we want to enable even less translated translations. I'm actually in favor of it, because if translations are missing that might motivate people to help with the translation. 

> Perhaps we can have >40% translated shown by default, and add Other languages.... to the menu to display all.

For what it is worth, I agree with both these sentiments.

phw:
I merged this for now. I also don't remember what percentage we used in the past, but I think it was more a rule of thumb / educated guess anyway (and still is). Let's roll with the 40% for now :D

PR #978, opened 2018-09-21
PICARD-1354: Allow using instruments and vocals as credited

phw:
# Summary
Currently Picard uses the standardized names for instruments and vocals. It should be possible to use the instruments / vocals as credited (e.g. if someone is credited to have played "drums (drum kit)" as just "drums").

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1354](https://tickets.metabrainz.org/browse/PICARD-1354)

# Solution
Add an option "Use instrument and vocal credits" in Options > Metadata (off by default). When activated the instrument / vocal credits are used instead of the default instrument / vocal name.

# Action


phw:
Looking at the other options we have for similar cases we should probably call the option "standardize_instruments" with reverted logic and enable it by default. What do you think?

picard/mbjson.py
```diff
@@ -101,8 +101,15 @@ def _decamelcase(text):
...
-    attrs = [_decamelcase(_REPLACE_MAP.get(a, a)) for a in attrs]
+def _transform_attribute(attr, credits):
+    if attr in credits:
+        return credits[attr]
+    else:
+        return _decamelcase(_REPLACE_MAP.get(attr, attr))
+
+
+def _parse_attributes(attrs, reltype, credits):
+    attrs = [_transform_attribute(a, credits) for a in attrs]
```
zas: There are 3 list comprehensions  based on `attrs` here, i wonder if this can be rewritten to something simpler using a normal loop.

picard/mbjson.py
```diff
@@ -131,7 +138,11 @@ def _relations_to_metadata(relations, m):
             if 'attributes' in relation:
                 attribs = [a for a in relation['attributes']]
             if reltype in ('vocal', 'instrument', 'performer'):
-                name = 'performer:' + _parse_attributes(attribs, reltype)
+                if config.setting['use_instrument_credits']:
```
zas: I'd cache `config.setting['use_instrument_credits']` using a variable outside the loop over `relations`

PR #983, opened 2018-09-25
PICARD-1285: Add Close Window menu entry for macOS

phw:
# Summary
Usually windows on macOS can be closed via a "Close Window" command, which is available both in the "File" menu and the keyboard shortcut ⌘W. This PR adds such a menu action to Picard on macOS

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1285](https://tickets.metabrainz.org/browse/PICARD-1285)

# Solution

# Action
While the change in this PR allows closing utility windows like activity protocol and debug log as well as the main window, it does not work as expected with the preferences window. This can be considered a bug caused by the fact that the preferences window is a modal dialog, which in Qt causes the menu items to get disabled. The issue is tracked at [PICARD-1282](https://tickets.metabrainz.org/browse/PICARD-1282). Still this PR provides the general functionality and is a prerequisite to fix PICARD-1282

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -377,6 +377,11 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.add_directory_action.setShortcut(QtGui.QKeySequence(_("Ctrl+D")))
         self.add_directory_action.triggered.connect(self.add_directory)
 
+        if sys.platform == "darwin":
+            self.close_window_action = QtWidgets.QAction(_("Close Window"), self)
+            self.close_window_action.setShortcut(QtGui.QKeySequence(_("Ctrl+W")))
+            self.close_window_action.triggered.connect(self.close_active_window)
+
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -377,6 +377,11 @@ def create_actions(self):
         self.add_directory_action.setShortcut(QtGui.QKeySequence(_("Ctrl+D")))
         self.add_directory_action.triggered.connect(self.add_directory)
 
+        if sys.platform == "darwin":
```
zas: Instead of multiple `sys.platform == "darwin"` tests, i think you should create a property `self.show_close_window` initialized in one place and test against it.

phw: Right, could also just check for `self.close_window_action`

PR #986, opened 2018-10-02
PICARD-1359: improve builtin server reliability and debugging output

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -27,6 +27,9 @@ from picard import (
     log,
 )
 
+OK_RESP = b"HTTP/1.1 200 OK\r\nCache-Control: max-age=0\r\n\r\nNothing to see here."
+ERR_RESP = b"HTTP/1.1 500 Internal Server Error\r\nCache-Control: max-age=0\r\n\r\nNothing to see here."
+
 
 class BrowserIntegration(QtNetwork.QTcpServer):
 
@@ -69,9 +72,9 @@ class BrowserIntegration(QtNetwork.QTcpServer):
         log.debug("Browser integration request: %r", rawline)
         try:
             line = rawline.decode()
-            conn.write(b"HTTP/1.1 200 OK\r\nCache-Control: max-age=0\r\n\r\nNothing to see here.")
+            conn.write(OK_RESP)
         except UnicodeDecodeError as e:
-            conn.write(b"HTTP/1.1 500 Internal Server Error\r\nCache-Control: max-age=0\r\n\r\nNothing to see here.")
+            conn.write(ERR_RESP)
             log.error(e)
             return
         finally:
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -65,12 +65,13 @@ class BrowserIntegration(QtNetwork.QTcpServer):
 
     def _process_request(self):
         conn = self.sender()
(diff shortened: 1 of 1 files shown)


picard/browser/browser.py
```diff
@@ -65,11 +65,18 @@ def stop(self):
...
     def _process_request(self):
         conn = self.sender()
-        line = bytes(conn.readLine()).decode()
         conn.write(b"HTTP/1.1 200 OK\r\nCache-Control: max-age=0\r\n\r\nNothing to see here.")
-        conn.disconnectFromHost()
+        rawline = conn.readLine().data()
+        log.debug("Browser integration request: %r", rawline)
+        try:
+            line = rawline.decode()
+        except UnicodeDecodeError as e:
```
phw: Upon further thinking we should probably return status 200 only on success in the try block and return a 500 in case of the UnicodeDecodeError

picard/browser/browser.py
```diff
@@ -65,11 +65,19 @@ def stop(self):
...
     def _process_request(self):
         conn = self.sender()
-        line = bytes(conn.readLine()).decode()
-        conn.write(b"HTTP/1.1 200 OK\r\nCache-Control: max-age=0\r\n\r\nNothing to see here.")
-        conn.disconnectFromHost()
+        rawline = conn.readLine().data()
+        log.debug("Browser integration request: %r", rawline)
+        try:
+            line = rawline.decode()
+            conn.write(b"HTTP/1.1 200 OK\r\nCache-Control: max-age=0\r\n\r\nNothing to see here.")
```
samj1912: Not really a big fan of hardcoded strings like these. Move it to constants maybe?

zas:
@samj1912 @phw: the code is returning now 400 Bad Request on malformed requests (ie. invalid mbid), please review again (and test)

phw:
I think that's a good reworking of this module, but I also think this is as far as we should go for now. E.g. one could nitpick about differentiating the error codes even more and maybe use 404 if the path is unknown, but for something as simple as this integration putting more complexity in it then needed is going to do more harm than it helps.

zas:
> I think that's a good reworking of this module, but I also think this is as far as we should go for now. E.g. one could nitpick about differentiating the error codes even more and maybe use 404 if the path is unknown, but for something as simple as this integration putting more complexity in it then needed is going to do more harm than it helps.

Yes, though i went a bit further to handle all corner cases, and have proper error logged when a badly formatted request is incoming, it is much harder to crash picard now.

picard/browser/browser.py
```diff
@@ -65,22 +83,45 @@ def stop(self):
...
-            elif action == "/opennat":
-                self.tagger.load_nat(args["id"])
+        rawline = conn.readLine().data()
+        log.debug("Browser integration request: %r", rawline)
+
+        def parse_line(line):
+            orig_line = line
+            try:
+                line = line.split()
+                if line[0] == "GET" and "?" in line[1]:
```
samj1912: This just seems to be doing a simple url parse. Why not use the python function instead of doing it manually? (I know you just moved this code, but just a suggestion)

zas: Done in faecb4cdd4b, not much shorter though, but still i think it's better.

PR #993, opened 2018-10-06
Isort fixes

phw:
See discussion in #991. isort was not sorting the picard modules as expected for me.

I also ran isort with this config on the source, and it did some changes. @zas: Are those the expected results, or do we miss some tweaking?


zas:
> Are those the expected results, or do we miss some tweaking?

Yes, those are expected, and that matches the current configuration

zas:
Hmmm, when running `isort -rc -ac -vb` with the change to `.isort.cfg`, i only have one change against current master tree, the one in `setup.py`.
Why do you get more changes ?

phw:
@zas: I still suspect you have some additional isort configuration somehwere. In https://github.com/timothycrosley/isort/wiki/isort-Settings it says you can also have `.isort.cfg` or `.editorconfig` files in your home directory. Can you check?

phw:
> Maybe change this to refer to its path within the picard package instead of using relative imports?

I'd like to get the isort configuration right in this PR. We can do changes to actual imports in a separate PR. Apart from this I don't see anything particular wrong of using relative imports in this specific case.

zas:
> @zas: I still suspect you have some additional isort configuration somehwere. In https://github.com/timothycrosley/isort/wiki/isort-Settings it says you can also have `.isort.cfg` or `.editorconfig` files in your home directory. Can you check?

Oh you're right, it happens i have `~/.isort.cfg` with following content:

```
[settings]
line_length=120
wrap_length=79
indent='    '
multi_line_output=3
use_parentheses=1
force_sort_within_sections=1
skip_glob=*__init__.py*
known_qt=PyQt5
known_resources=resources
sections=FUTURE,STDLIB,QT,LOCALFOLDER,RESOURCES
default_section=LOCALFOLDER
```

phw:
> But after removing this file, and testing again this PR, i still obtain the same result:

I can confirm this for my use case as well. I tested the other way around and added your `~/.isort.cfg`, but still get the same result with all the changes in this PR. Also isort 4.3.4, on Arch Linux with the `python-isort` package if this matters

phw:
Do we maybe still miss some place where configuration is pulled from?

But I agree, the current behavior is far from ideal. Actually if there is a project specific configuration I wld expect isort to use this and only this. That it is merging in user preferences really makes it hard to define reliable project defaults.

zas:
> @zas: Maybe another test: If you run isort on my branch here, does it suggest to revert my changes?

Using your own branch, and running `isort -rc` only changes `setup.py`.

phw:
>  running `isort -rc` only changes `setup.py`.

Yes, indeed I did not run it on this one but limited it on `picard/`. I saw that it proposed changes on setup.py but assumed you had excluded that one previously.

Actually I think we should merge this if this branch doesn't cause any changes when you run isort, what do you think?

zas:
> Actually I think we should merge this if this branch doesn't cause any changes when you run isort, what do you think?

Yes, you can merge it.

PR #995, opened 2018-10-07
PICARD-1369: Retry opening URLs

mineo:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Picard crashes opening any URL through its webbrowser2 module on Python 3.7.0.

* JIRA ticket (_optional_): [PICARD-1369](https://tickets.metabrainz.org/browse/PICARD-1369)

# Solution

This is a workaround for https://bugs.python.org/issue31014. The first time
webbrowser.open is called, an exception is thrown when initializing the list of
browsers on the system. Ignoring that first exception leaves the webbrowser
module in a possibly-not-optimal state regarding the list of browsers it uses,
but it does successfully open URLs. Since there's no point in fixing this in
Picard itself, just continue with a not optimal browser choice.

# Action


picard/util/webbrowser2.py
```diff
@@ -17,12 +17,14 @@
...
 
+from sys import version_info
 import webbrowser
 
-from PyQt5 import QtWidgets
-
+from picard import log
 from picard.const import PICARD_URLS
 
+from PyQt5 import QtWidgets
```
zas: There's no reason to move this import here, it was in correct place.

picard/util/webbrowser2.py
```diff
@@ -35,6 +37,12 @@ def open(url):
         webbrowser.open(url)
     except webbrowser.Error as e:
         QtWidgets.QMessageBox.critical(None, _("Web Browser Error"), _("Error while launching a web browser:\n\n%s") % (e,))
+    except TypeError as e:
+        if version_info.major == 3 and version_info.minor == 7 and version_info.micro == 0:
+            # See https://bugs.python.org/issue31014, webbrowser.open doesn't
+            # work on 3.7.0.
+            log.info("Working around https://bugs.python.org/issue31014 - URLs might not be opened in the correct browser")
+            webbrowser.open(url)
```
zas: Do i miss something here ? The workaround is to call `webbrowser.open(url)` a second time ??

picard/util/webbrowser2.py
mineo: This is what `isort` did to the file.

picard/util/webbrowser2.py
mineo: Yes. The second time `webbrowser.open` is called, the on-demand initialization in https://github.com/python/cpython/blob/7dfbd4967121ef6ecabe3fd53a430fd949f405fd/Lib/webbrowser.py#L72 is not executed.

picard/util/webbrowser2.py
phw: Yep, see #993. We are trying to sort this out :)

PR #1000, opened 2018-10-10
Removed a bunch of unused imports

zas:


picard/browser/browser.py
```diff
@@ -23,7 +23,6 @@
 )
 
 from PyQt5 import (
-    QtCore,
```
phw: Shouldn't this go on one line now? Or is it the different isort behavior again?

zas: Well, if it was me, all  imports will go with following formats:

```python
from X import (
  import1,
  import2,
)
```

Even with one single import:

```python
from X import (
  import1,
)
```

Reasons:

- if one changes one line, patch isn't polluted by non-real changes (like new line, or adding a comma)
- if an import is added, or removed, the patch only shows +/- this line

So, i don't think we should go to one format then another, and stick with one working for all cases...

If we go the "one line" import depending on the number of items to import, the syntax will continuously change.

From:

```python
from X import import1
```

To:

```python
from X import import1, import2
```

Then, at some point:
```python
from X import (
  import1,
  import2,
  import3,
)
```

Of course, we can enforce running `isort` on each commit... not sure this is what we want.

The `isort` current configuration was meant to satisfy people who cannot live with 3 lines single import. I'm not part of those, i prefer consistency.

```diff
--- t.py	2018-10-10 18:02:40.596944106 +0200
+++ t.py	2018-10-10 18:03:35.676220477 +0200
@@ -1 +1 @@
-from X import import1, import2
+from X import import2
```

vs.

```diff
--- t.py	2018-10-10 18:06:23.498018160 +0200
+++ t.py	2018-10-10 18:06:39.857803645 +0200
@@ -1,4 +1,3 @@
 from X import (
-    import1,
     import2,
 )
```

A matter of taste i guess

PR #1005, opened 2018-10-17
PICARD-1370: Remove Quick Launch shortcut from Windows installer

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The Quick Launch toolbar used to be a small area of fixed application icons shown by default left of the running applications. When Windows moved to a more dock like task bar with Windows 7 the Quick Launch icons were no longer shown by default.

Picard provides an option to add an icon to this Quick Launch list. Since the Quick Launch icons are no longer visible few users will know what this even is. In the best case it is useless, in the worst case people expect that the Picard icon will be pinned to the taskbar (see [this discussion](https://community.metabrainz.org/t/please-test-new-picard-windows-builds/400717/6)).

Hence this PR will remove the Quick Launch shortcut option from the installer. I intentionally left the uninstall part in. This ensures that the icon will be uninstalled if it was installed by an earlier Picard installer.

While there are ways to kind of [get the Quick Launch icons back to the taskbar](https://www.giga.de/downloads/windows-10/tipps/windows-10-schnellstartleiste-aktivieren-so-geht-s-quick-launch/), people being able to do this likely also can add a shortcut to it if needed.

* JIRA ticket (_optional_): [PICARD-1370](https://tickets.metabrainz.org/browse/PICARD-1370)

# Solution

# Action

### DIFF
--- a/installer/picard-setup.nsi.in
+++ b/installer/picard-setup.nsi.in
@@ -134,13 +134,6 @@ SubSection "Shortcuts" shortcuts
       "" "" "" SW_SHOWNORMAL "" "${PRODUCT_DESCRIPTION}"
   SectionEnd
 
-  Section "Quick Launch" quicklaunch
-    SetShellVarContext all
-    SetOutPath "$INSTDIR"
-    CreateShortCut "$QUICKLAUNCH\${PRODUCT_NAME}.lnk" "$INSTDIR\picard.exe" \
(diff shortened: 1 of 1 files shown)


Sophist-UK:
Personally I still use Quicklaunch icons on Windows 10 because they take less screen space than an inactive pinned application, but I still agree with this PR - users like me can add the Quicklaunch icons manually.

PR #1006, opened 2018-10-17
PICARD-1339: Further optimizations for update_metadata_images

phw:
# Summary

This is a follow up to #975 which introduces some more improvement how update_metadata_images is handled. The most important change is that for the special `UnclusteredFiles` cluster no metadata update is performed, since it does not show the cover art for the included files anyway.

With these changes I think we can safely close PICARD-1339

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1339](https://tickets.metabrainz.org/browse/PICARD-1339)

# Solution

# Action

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -94,11 +94,13 @@ class Cluster(QtCore.QObject, Item):
             self.metadata.length += file.metadata.length
             file._move(self)
             file.update(signal=False)
-            file.metadata_images_changed.connect(self.update_metadata_images)
+            if self.can_show_coverart:
+                file.metadata_images_changed.connect(self.update_metadata_images)
         self.files.extend(files)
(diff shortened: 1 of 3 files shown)


picard/cluster.py
```diff
@@ -332,6 +345,10 @@ def can_view_info(self):
     def can_remove(self):
         return len(self.files) > 0
 
+    @property
+    def can_show_coverart(self):
```
zas: Perhaps move this to Item class instead.

phw: Done in 8dba1fd7d9e86eff2b187975db83d53125960a65. I would also like to make all those `can_*` functions to properties, but this is probably something for a separate PR

PR #1008, opened 2018-10-18
PICARD-1379: Fix PyUnicode_GetSize deprecation warning in astrcmp

phw:
# Summary

Use `PyUnicode_GetLength` instead of deprecated `PyUnicode_GetSize`. Requires Python >= 3.3, but we list Python 3.5 as the minimum anyway.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1379](https://tickets.metabrainz.org/browse/PICARD-1379)

# Solution

# Action

### DIFF
--- a/picard/util/_astrcmp.c
+++ b/picard/util/_astrcmp.c
@@ -1,5 +1,6 @@
 /*
  * Picard, the next-generation MusicBrainz tagger
+ * Copyright (C) 2018 Philipp Wolfer
  * Copyright (C) 2006 Lukáš Lalinský
  * Copyright (C) 2003 Benbuck Nason
  *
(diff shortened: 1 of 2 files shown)


picard/util/_astrcmp.c
```diff
@@ -156,8 +156,8 @@ astrcmp(PyObject *self, PyObject *args)
 
 	us1 = PyUnicode_AS_UNICODE(s1);
 	us2 = PyUnicode_AS_UNICODE(s2);
```
zas: Isn't this also deprecated ?
https://docs.python.org/3/c-api/unicode.html#c.PyUnicode_AS_UNICODE

phw: Indeed it is. There was just no warning output, probably because these are macros. I will update the PR

phw:
I updated the module for the new Unicode API. For a quick speed comparison I measured running the unit test function 1 million times. On my system:

- Old implementation: 3.9155 seconds
- New implementation: 4.0373 seconds
- New implementation without https://github.com/metabrainz/picard/pull/1008/commits/6eb132ba3d133f3e2cc16c21eb37210d08b91483: 4.5491

So a slight performance decrease of 3%. But for comparison the Python implementation takes roughly 186 seconds for the same call count.

phw:
Sorry, this needs more work. I just realized my patch breaks the threading

PR #1010, opened 2018-10-18
PICARD-1379: Port _astrcmp to new Python Unicode API

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Fix deprecation warning during compilation by porting _astrcmp to the new Python Unicode API introduced with Python 3.3, see https://docs.python.org/3/c-api/unicode.html

* JIRA ticket (_optional_): [PICARD-1379](https://tickets.metabrainz.org/browse/PICARD-1379)

# Solution

# Action

### DIFF
--- a/picard/util/_astrcmp.c
+++ b/picard/util/_astrcmp.c
@@ -88,6 +88,7 @@ float LevenshteinDistance(int k1, const void * s1, Py_ssize_t len1,
 
 	for (index1 = 1; index1 <= len1; index1++)
 	{
+		Py_UCS4 s1_previous = 0;
 		Py_UCS4 s1_current = PyUnicode_READ(k1, s1, index1 - 1);
 
 		/* Step 4 */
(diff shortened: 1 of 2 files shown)


phw:
See also https://github.com/metabrainz/picard/pull/1008

phw:
The final commit changed the code again a bit so that it looks more closely to the old one. But there is a small difference: We create an actual copy of the unicode data. This is a bit slower of course, but it ensures we can safely call Py_UNBLOCK_THREADS without risking the string data being altered while we iterate over it. This actually was a potential issue with the old code, which operated on the internal unicode representation directly after Py_UNBLOCK_THREADS got called.

My first approach in https://github.com/metabrainz/picard/pull/1010/commits/ac8d8049a10a767460d7581bd4b1050e2b611ef1 worked also without copy, but I think we have to take the performance hit here.

PR #1012, opened 2018-10-21
PICARD-1381: Remove dependencies between tests

mineo:
# Summary
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

At the moment, some test files depend on other tests to be run before them to pass. `test_imagelist` for example was depending on `test_filesystem`.

* JIRA ticket (_optional_): [PICARD-1381](https://tickets.metabrainz.org/browse/PICARD-1381)

# Solution

## Execute tests in a random order on Travis.

This is done by switching to pytest (with pytest-randomly) instead of nose. I tried really hard to understand our current nose setup, but I failed. We're currently installing `nose2-0.8.0`, which automatically pulls in `coverage-4.5.1`. Then, if `CODACY_PROJECT_TOKEN` is set, we're installing `nose-cov`, which depends on `nose-1.3.7` (note this is not nose2, but nose). I could not make sense of that. There's [nose-randomly](https://pypi.org/project/nose-randomly/) that is similar to pytest-randomly, but requires nose instead of nose2, so I guess it doesn't work with the `nose2` executable we currently use. With pytest, things are less confusing.

## Show dependencies early

A new class, `PicardTestCase` is introduced whose `setUp` method sets up a `FakeTagger` at `QtCore.QObject.tagger`, which was where things broke for `test_imagelist` if executed before `test_filesystem`. Config settings are also cleared so every test has to make clear what its expected settings are.

# Action


test/test_coverart_utils.py
```diff
@@ -3,14 +3,15 @@
...
+from test.picardtestcase import PicardTestCase
 
 from picard.coverart.utils import translate_caa_type
 from picard.i18n import setup_gettext
 
 
-class CaaTypeTranslationTest(unittest.TestCase):
+class CaaTypeTranslationTest(PicardTestCase):
     def setUp(self):
+        super(CaaTypeTranslationTest, self).setUp()
```
samj1912: Why py2 style supers?

test/test_browser.py
```diff
@@ -13,9 +13,10 @@
 LOCAL_PORT = "8000"
 
 
-class BrowserLookupTest(unittest.TestCase):
+class BrowserLookupTest(PicardTestCase):
 
     def setUp(self):
+        super(BrowserLookupTest, self).setUp()
```
samj1912: Python 2 style super?

samj1912: Python 2 style super?

test/test_acoustid.py
```diff
@@ -17,19 +17,20 @@
 }
 
 
-class AcoustIDTest(unittest.TestCase):
+class AcoustIDTest(PicardTestCase):
+    def setUp(self):
+        super(AcoustIDTest, self).setUp()
```
samj1912: Python 2 style super?

test/test_coverart_utils.py
phw: Agreed, those should be fixed before merging

PR #1014, opened 2018-10-22
PICARD-1383: Select release with matching barcode on disc ID lookup

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When a disc ID lookup contains multiple results picking the correct one can be tricky. We can assist the user by making use of the barcode stored on the disc itself.

* JIRA ticket (_optional_): [PICARD-1383](https://tickets.metabrainz.org/browse/PICARD-1383)

# Solution
Read the MCN field (containing the EAN as per spec) from the disc when generating the disc ID. Use this barcode to pre-select the release that matches the barcode.

# Action

### DIFF
--- a/picard/ui/cdlookup.py
+++ b/picard/ui/cdlookup.py
@@ -31,7 +31,10 @@ from picard.mbjson import (
     label_info_from_node,
     release_dates_and_countries_from_node,
 )
-from picard.util import restore_method
+from picard.util import (
+    compare_barcodes,
+    restore_method,
+)
 
 from picard.ui import PicardDialog
 from picard.ui.ui_cdlookup import Ui_Dialog
@@ -61,11 +64,14 @@ class CDLookupDialog(PicardDialog):
             def myjoin(l):
                 return "\n".join(l)
 
+            selected = None
             for release in self.releases:
                 labels, catalog_numbers = label_info_from_node(release['label-info'])
                 dates, countries = release_dates_and_countries_from_node(release)
                 barcode = release['barcode'] if "barcode" in release else ""
                 item = QtWidgets.QTreeWidgetItem(self.ui.release_list)
+                if disc.mcn and compare_barcodes(barcode, disc.mcn):
+                    selected = item
                 item.setText(0, release['title'])
                 item.setText(1, artist_credit_from_node(release['artist-credit'])[0])
                 item.setText(2, myjoin(dates))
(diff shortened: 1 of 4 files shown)


picard/ui/cdlookup.py
```diff
@@ -61,11 +64,15 @@ def __init__(self, releases, disc, parent=None):
             def myjoin(l):
                 return "\n".join(l)
 
+            selected = None
             for release in self.releases:
                 labels, catalog_numbers = label_info_from_node(release['label-info'])
                 dates, countries = release_dates_and_countries_from_node(release)
                 barcode = release['barcode'] if "barcode" in release else ""
                 item = QtWidgets.QTreeWidgetItem(self.ui.release_list)
+                print(disc.mcn, barcode, barcode == disc.mcn)
```
zas: Debug ?

picard/util/__init__.py
```diff
@@ -566,3 +566,21 @@ def reconnect(signal, newhandler=None, oldhandler=None):
...
+    """
+    if barcode1 == barcode2:
+        return True
+    if not barcode1 or not barcode2:
+        return False
+    if len(barcode1) == 12:
+        barcode1 = '0' + barcode1
+    if len(barcode2) == 12:
+        barcode2 = '0' + barcode2
+    return barcode1 == barcode2
```
zas: You may simplify by 0 padding both strings (to 13, for EAN13) using:

```python
return barcode1.zfill(13) == barcode2.zfill(13)
```

Then you can drop `if len...` tests.

phw: Much better. Thanks, fixed

test/test_utils.py
```diff
@@ -224,3 +224,18 @@ def test_invalid_png_data(self):
...
                           imageinfo.identify, data)
+
+class CompareBarcodesTest(unittest.TestCase):
+
+    def test_same(self):
+        self.assertTrue(util.compare_barcodes('0727361379704', '0727361379704'))
+        self.assertTrue(util.compare_barcodes('727361379704', '727361379704'))
+        self.assertTrue(util.compare_barcodes('727361379704', '0727361379704'))
+        self.assertTrue(util.compare_barcodes('0727361379704', '727361379704'))
+        self.assertTrue(util.compare_barcodes(None, None))
```
zas: I'd add tests for empty strings too:

```python
self.assertTrue(util.compare_barcodes('', ''))
```

test/test_utils.py
```diff
@@ -224,3 +224,18 @@ def test_invalid_png_data(self):
...
+
+    def test_not_same(self):
+        self.assertFalse(util.compare_barcodes('0727361379704', '0727361379705'))
+        self.assertFalse(util.compare_barcodes('727361379704', '1727361379704'))
+        self.assertFalse(util.compare_barcodes('0727361379704', None))
+        self.assertFalse(util.compare_barcodes(None, '0727361379704'))
```
zas: I'd add tests for empty strings too:

```python
self.assertFalse(util.compare_barcodes(None, ''))
```

phw: Is this actually what we would expect, or shouldn't `None` and `''` be considered equal?

zas: I guess it'd make sense.

phw: Yes, I think it makes most sense. I implemented it this way, which needed some update to the compare_barcodes function itself.

PR #1020, opened 2018-10-27
PICARD-1384: Create org.musicbrainz.Picard.appdata.xml

abhi-ohri:
Add an app description to MusicBrainz Picard. AppStream adds standardized descriptions of applications, that can be used by different application stores and installers.

# Summary
Add appdata xml file to the project.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [X] Other

# Problem
There is no standard metadata file available to help package distributors display a description of Picard.

* JIRA ticket (_optional_): [PICARD-1384](https://tickets.metabrainz.org/browse/PICARD-1384)

# Solution
Add Appdata  XML file

# Action


samj1912:
I do not see any merit in just creating this file without any links to it in the installer. @phw any comments?

phw:
Thanks, this is already looking good. Apart from the screenshots, which you already said you will add when ready, we need two changes to Picard to make this work as intended:

1. As mentioned in the task the appdata files needs to be installed to `/usr/share/metainfo`. You can achieve this by adding the following lines to `args['data_files']` in [setup.py#L656](https://github.com/metabrainz/picard/blob/master/setup.py#L656)
    ```
    ('share/metainfo', ['org.musicbrainz.Picard.appdata.xml']),
    ```

2. We should take the chance and rename the existing `picard.desktop` file to `org.musicbrainz.Picard.desktop`. For that you have to rename the file and also update the install line in [setup.py#L656](https://github.com/metabrainz/picard/blob/master/setup.py#L656)

The reason for the second change is that this file name is used in the appdata file as `<launchable type="desktop-id">`, and the naming with a reverse domain name like `org.musicbrainz.Picard` is the now recommended way of naming these files according to the [Freedesktop Desktop Entry Specification](https://standards.freedesktop.org/desktop-entry-spec/latest/ar01s02.html). Also this is required by e.g. Flatpak.

phw:
@abhi-ohri One minor thing: Please change the title of this pull request to be prefixed with `PICARD-1384: `. This will automatically link it to the ticket on JIRA.

bochecha:
> We should take the chance and rename the existing picard.desktop file to org.musicbrainz.Picard.desktop. For that you have to rename the file and also update the install line in setup.py#L656

Indeed.

That's why I opened https://tickets.metabrainz.org/browse/PICARD-1393, before knowing @abhi-ohri would send his appdata file upstream, because I knew this would cause problems.

I recommend first renaming the resource files, then adding the appdata.

As it currently is, the appdata can't work.

abhi-ohri:
@phw I have made the required changes.

setup.py
```diff
@@ -654,14 +654,15 @@ def find_file_in_path(filename):
...
 
 args['data_files'] = [
+    ('share/metainfo', ['org.musicbrainz.Picard.appdata.xml']),
     ('share/icons/hicolor/16x16/apps', ['resources/images/16x16/picard.png']),
     ('share/icons/hicolor/24x24/apps', ['resources/images/24x24/picard.png']),
     ('share/icons/hicolor/32x32/apps', ['resources/images/32x32/picard.png']),
     ('share/icons/hicolor/48x48/apps', ['resources/images/48x48/picard.png']),
     ('share/icons/hicolor/128x128/apps', ['resources/images/128x128/picard.png']),
     ('share/icons/hicolor/256x256/apps', ['resources/images/256x256/picard.png']),
     ('share/icons/hicolor/scalable/apps', ['resources/img-src/picard.svg']),
```
bochecha: All these icons also need to be renamed to `org.musicbrainz.Picard.*`.

Which means you must also modify the `Icon=` key in the desktop file. :wink:

abhi-ohri: Should I also rename all icon png files in resources manually or they are auto-generated and I do not need to change their name. @bochecha

bochecha: I see no indication that they are auto-generated, and they are even committed in Git.

Also, I just found that if you rename the image files, you also need to rename them there: https://github.com/metabrainz/picard/blob/master/resources/picard.qrc

To identify which one you need to rename, it's all the ones that get installed by `setup.py` in `${prefix}/share/icons/hicolor/${SIZE}/apps`.

org.musicbrainz.Picard.appdata.xml
```diff
@@ -0,0 +1,73 @@
...
+    <content_attribute id="sex-appearance">none</content_attribute>
+    <content_attribute id="language-profanity">none</content_attribute>
+    <content_attribute id="language-humor">none</content_attribute>
+    <content_attribute id="language-discrimination">none</content_attribute>
+    <content_attribute id="social-chat">none</content_attribute>
+    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-audio">none</content_attribute>
+    <content_attribute id="social-location">none</content_attribute>
+    <content_attribute id="social-contacts">none</content_attribute>
+    <content_attribute id="money-purchasing">none</content_attribute>
```
Freso: This seems inaccurate. Is Picard no longer asking people to donate to MetaBrainz?

org.musicbrainz.Picard.appdata.xml
```diff
@@ -0,0 +1,73 @@
...
+  <launchable type="desktop-id">org.musicbrainz.Picard.desktop</launchable>
+
+  <screenshots>
+    <screenshot type="default">https://www.daitauha.fr/static/screenshots/picard/main.png</screenshot>
+  </screenshots>
+
+  <url type="homepage">https://picard.musicbrainz.org/</url>
+  <url type="bugtracker">https://tickets.metabrainz.org/projects/PICARD/issues/</url>
+  <url type="faq">https://picard.musicbrainz.org/docs/faq/</url>
+  <url type="help">https://community.metabrainz.org/c/picard</help>
```
Freso: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-url suggests that the "help" type should be for "an online user's reference, a software manual or help page." I.e., documentation of some kind, not a support forum. Is it de facto used to link to forums (ie., did you see other AppStream files doing this) and the spec just doesn't reflect this (yet)?

org.musicbrainz.Picard.appdata.xml
```diff
@@ -0,0 +1,73 @@
...
+    <content_attribute id="language-profanity">none</content_attribute>
+    <content_attribute id="language-humor">none</content_attribute>
+    <content_attribute id="language-discrimination">none</content_attribute>
+    <content_attribute id="social-chat">none</content_attribute>
+    <content_attribute id="social-info">none</content_attribute>
+    <content_attribute id="social-audio">none</content_attribute>
+    <content_attribute id="social-location">none</content_attribute>
+    <content_attribute id="social-contacts">none</content_attribute>
+    <content_attribute id="money-purchasing">none</content_attribute>
+    <content_attribute id="money-gambling">none</content_attribute>
```
Freso: AFAICT, [according to the spec](https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-content_rating), you only need to specify the ones that are *not* "none".

abhi-ohri: I did not alter that because it was present in the base file and also in some appdata files I saw in other projects. Should I remove the unneeded ones?

org.musicbrainz.Picard.appdata.xml
abhi-ohri: I agree with you. Its a mistake I made. I am correcting it.
```suggestion
  <url type="help">https://picard.musicbrainz.org/docs/</url>
```

org.musicbrainz.Picard.appdata.xml
abhi-ohri: I checked with OARS asking for money donation is not same as money purchasing. Hence, I marked it as none.

setup.py
abhi-ohri: The images need to be refactored [here](https://github.com/metabrainz/picard/blob/5241c32c49e50aae49c2653f29713aea21ef6399/picard/ui/mainwindow.py) too.

phw:
@abhi-ohri: After further checking the only thing breaking the builds right now is the rename of picard.ico. Since that one is used only for Windows it should not be renamed.

Since you already did this renaming here completely we can also keep it in if you fix just the .icon name. But I will add a separate ticket for this to track it in our changelog.

Thanks for the changes btw., good work, and sorry for the confusion. I just would have preferred to keep this change separate,but now that we already have it here let's not complicate things further :)

abhi-ohri:
@phw Reverted the icon name and builds are passing. Any further instruction?

org.musicbrainz.Picard.appdata.xml
phw: Not sure really, is it ok to remove all of them if they are all none? @bochecha, since you added this initially any thoughts on it? I am not too familiar with OARS data inside appstream.

bochecha: If you remove all of them, then you end up not having any OARS metadata at all, which software centers interpret as "not rated" rather than "rated safe".

bochecha:
> Since you already did this renaming here completely we can also keep it in if you fix just the .icon name. But I will add a separate ticket for this to track it in our changelog.

That's what I opened https://tickets.metabrainz.org/browse/PICARD-1393 for.

org.musicbrainz.Picard.appdata.xml
Freso: @bochecha That's contrary to the spec though. :/ It specifically says:
> In case the `<content_rating/>` tag is empty (no `<content_attribute/>` is present), it is assumed that the component was checked for age ratings and no age restrictions apply.

However, it that's the de facto standard, then that's how it is.

org.musicbrainz.Picard.appdata.xml
Freso: Alright. The generator at https://hughsie.github.io/oars/generate.html has "Users are encouraged to donate real money, e.g. using Patreon" (which Picard *does* do) as `money-purchasing` = `mild`. But if you've asked OARS people and they say differently, then that's probably correct. :)

org.musicbrainz.Picard.appdata.xml
```diff
@@ -0,0 +1,73 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<component type="desktop-application">
+  <id>org.musicbrainz.Picard</id>
+  <project_license>GPL-2.0+</project_license>
```
Freso: @bochecha Hi :) https://spdx.org/licenses/GPL-2.0+ says it's deprecated. It doesn't say what it's deprecated in favour of, but https://spdx.org/licenses/ has "GPL v2.0 or later" as [`GPL-2.0-or-later`](https://spdx.org/licenses/GPL-2.0-or-later.html). Should the license be updated/changed here too?

org.musicbrainz.Picard.appdata.xml
abhi-ohri: @Freso Sorry for the error on my side. It indeed needs to be mild. It was a miscommunication on my part. I will correct it. I didn't know the donation column was in In-App Purchases.
```suggestion
    <content_attribute id="money-purchasing">mild</content_attribute>
```

org.musicbrainz.Picard.appdata.xml
abhi-ohri: @Freso @bochecha Since now we need to set content attribute money-purchasing to _mild_. I guess we could omit the unneeded ones.

abhi-ohri:
@Freso @phw Should I link [PICARD-1393](https://tickets.metabrainz.org/browse/PICARD-1393) with the PR too?

org.musicbrainz.Picard.appdata.xml
bochecha: Does the encouragement to donate appear inside the app?

org.musicbrainz.Picard.appdata.xml
bochecha: > In case the <content_rating/> tag is empty

Oh, maybe I'm mixing up between an empty `<content_rating/>` and no `<content_rating/>` at all?

I'll try an empty `<content_rating/>` in the Flathub build, and see what GNOME Software has to say about it. :slightly_smiling_face:

org.musicbrainz.Picard.appdata.xml
bochecha: The problem is SPDX has multiple versions, and the appstream spec doesn't say which version it relies on.

However the examples use the deprecated form: https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-project_license

When in doubt, the best is to use `appstream-util validate-relax` to verify. Both seem to be accepted, so yes, let's move to the newer SPDX identifier. :slightly_smiling_face:

org.musicbrainz.Picard.appdata.xml
abhi-ohri: We can do that once it is resolved whether to use _money-purchasing_ to _none_ or _mild_ .

org.musicbrainz.Picard.appdata.xml
abhi-ohri: Yes it does on the About Page in the application.

abhi-ohri:
I have added the screenshots as well.

org.musicbrainz.Picard.appdata.xml
phw: We have a note in the about page and an actual "Donate..." menu entry, so I think setting this to mild would be just fair. Let's do this.

org.musicbrainz.Picard.appdata.xml
phw: Agreed with @abhi-ohri . So let's set money-purchasing to mild and remove everything else.

org.musicbrainz.Picard.appdata.xml
Freso: @abhi-ohri :point_up_2: :)

phw: Let's got with the newer syntax "GPL-2.0-or-later" then

org.musicbrainz.Picard.appdata.xml
abhi-ohri: SInce the `content_rating` now has a content-attribute `money-purchasing` we can remove the unneeded tags. @Freso What's your say?

org.musicbrainz.Picard.appdata.xml
abhi-ohri:
```suggestion
  <project_license>GPL-2.0-or-later</project_license>
```
Done.

picard/ui/mainwindow.py
```diff
@@ -1,4 +1,4 @@
-# -*- coding: utf-8 -*-
+﻿# -*- coding: utf-8 -*-
```
Freso: Did your editor add a [BOM](https://en.wikipedia.org/wiki/Byte_order_mark "byte order mark")? Or what changed on this line? BOMs are generally discouraged, and I don't see anything else that would have changed on this line.

phw: Good catch, indeed. I checked the file, there is a BOM. That should be removed for UTF-8 files.

abhi-ohri:
```suggestion
```

picard/ui/mainwindow.py
```diff
@@ -1,4 +1,3 @@
-# -*- coding: utf-8 -*-
```
zas: Why is this line removed ?

abhi-ohri: @zas It is a BOM(Byte Order Mark) and BOMs are not preferred for UTF-8 files. So, @phw and @Freso  asked me to remove it.

phw: Ah, sorry. That was a misunderstanding derstanding. The BOM are just three bytes at the very beginning, before the `#`. They are not visible here, this makes the discussion a bit difficult. But it go added maybe by your editor automatically. It's maybe even still in the file, despite you removed the first line.

So `# -*- coding: utf-8 -*-` should be added again. And before saving check your text editor if there is some option regarding the BOM. What editor are you using?

And to not clutter this pull request with too many commits I would suggest when doing these changes you ammend the last commit with the BOM message. That way it stays only a single commit. You have to force push the to GitHub, but for a feature branch that's ok (but never do this on a project's master branch, it will cause trouble).

abhi-ohri: Done.

phw:
@abhi-ohri I think somewhat we all got confused here, sorry for that. I actually think you already got this right before your last commit and removed the BOM correctly.

But it looks like @zas got confused by the remove and add lines in e201d65f03a612c7f7932e2ffd4a12094876556e, I got confused by @zas's comment thinking it referred to your last commit and then I confused you with my comment about `#` to do the wrong thing. Sorry for that :( Stupid invisible characters.

To avoid further confusion I took the freedom to fixup and merge the branch directly. All done and merged now, thanks.

PR #1022, opened 2018-10-28
PICARD-862: Improve Test Coverage of Picard

virusMac:
# Summary
* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
# Problem
Helping to increase the test coverage of picard

* JIRA ticket (_optional_): [PICARD-862](https://tickets.metabrainz.org/browse/PICARD-862)
# Solution
Added 3 tests for func_pad, func_replace and func_strip from script.py
# Action

### DIFF
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -180,6 +180,21 @@ class ScriptParserTest(unittest.TestCase):
     def test_cmd_upper(self):
         self.assertScriptResultEquals("$upper(AbeCeDA)", "ABECEDA")
 
+    def test_cmd_pad(self):
+        self.assertScriptResultEquals("$pad(abc de,10,-)", "----abc de")
+        self.assertScriptResultEquals("$pad(abc de,e,-)", "")
+        self.assertScriptResultEquals("$pad(abc de,6,-)", "abc de")
+        self.assertScriptResultEquals("$pad(abc de,3,-)", "abc de")
+        self.assertScriptResultEquals("$pad(abc de,0,-)", "abc de")
+        self.assertScriptResultEquals("$pad(abc de,-3,-)", "abc de")
+
+
+    def test_cmd_replace(self):
+        self.assertScriptResultEquals("$replace(abc ab abd a,ab,test)", "testc test testd a")
+
+    def test_cmd_strip(self):
+        self.assertScriptResultEquals("$strip(  \t abc  de \n f  )", "abc de f")
+
     def test_cmd_rreplace(self):
         self.assertScriptResultEquals(r'''$rreplace(test \(disc 1\),\\s\\\(disc \\d+\\\),)''', "test")
 


test/test_script.py
```diff
@@ -180,6 +180,15 @@ def test_cmd_lower(self):
...
         self.assertScriptResultEquals("$upper(AbeCeDA)", "ABECEDA")
 
+    def test_cmd_pad(self):
+        self.assertScriptResultEquals("$pad(abc de,10,-)", "----abc de")
+
+    def test_cmd_replace(self):
+        self.assertScriptResultEquals("$replace(abc ab abd a,ab,test)", "testc test testd a")
+
+    def test_cmd_strip(self):
+        self.assertScriptResultEquals("$strip(   abc  de f  )", "abc de f")
```
phw: For strip we should make sure it removes other whitespace characters then pnly space. I would additionally test for newlines `\n` and tabs `\t` by modifying the test to include those characters, e.g. `"$strip(  \t abc  de f \n )", "abc de f"`

test/test_script.py
```diff
@@ -180,6 +180,15 @@ def test_cmd_lower(self):
     def test_cmd_upper(self):
         self.assertScriptResultEquals("$upper(AbeCeDA)", "ABECEDA")
 
+    def test_cmd_pad(self):
+        self.assertScriptResultEquals("$pad(abc de,10,-)", "----abc de")
```
phw: One test should be added here. The code of `func_pad` does convert the `length` into an integer with `int(length)`. This can cause a `ValueError` if length is a non-numeric string. The function catches this and returns an empty string.

A test should be added for this code path, e.g. test that `"$pad(abc,x,-)"` equals `""`.

As a side note I am not sure whether this is the best expected behavior for the `pad` function. As a caller of the function I would rather expect it to return the input text without padding. But let's keep it is it is for now and don't risk breaking someones script. But it's a nice example of the kind of issues early unit tests can uncover :D

virusMac: In fact, when I was testing $pad() in Picard itself, it was returning that input text without padding :D

rdswift:
Should there also be a test for a negative integer length, or is that covered with the non-integer test?

phw:
> Should there also be a test for a negative integer length, or is that covered with the non-integer test?

Good point, yes, I think there should be one. And no, looking at the code this is not covered by the non-integer test. The current and expected behavior would be for no padding to happen.

test/test_script.py
virusMac: Which test do you mean? :)
(If you mean that test I was talking about above, I checked _$pad()_ just in Picard exe app and it kept old title - it didn't change anything. Everything in "_python setup.py test_" was OK)

phw: I meant the test of "$pad(abc,x,-)" equals "" I mentioned above.

Theory time: If you test a function and you look for test cases you usually should go for the following tests:

1. Test a normal use case that shows typical usage
2. If there are different code paths through a function, make sure you have tests for all code path. E.g. for the function:

      ```
      def func(a):
        if a > 0:
          return "Hello"
        else:
          return "Good By"
      ```

   This should have both a test for a > 0 and a <= 0.

3. Test the corner cases. In the above that would mean there should be a test for a = 0

In our case we have the following function to test:

```
def func_pad(parser, text, length, char):
    try:
        return char * (int(length) - len(text)) + text
    except ValueError:
        return ""
```

Given my statements we should probably have the following:

1. Standard behavior: You already added a test for this:

    - `$pad(abc de,10,-)` --> `"----abc de"`

2. Different code paths: The try except block causes this to have different code paths. We must check the case where a `ValueError` is thrown, that's the case of length cannot be converted into an integer. A test for this would be:

    - `$pad(abc de,x,-)` --> `""`

3. Corner cases: I missed that one, but we should have some tests for this. Multiple conditions should be checked because of `char * (int(length) - len(text)) + text`:

    - `length == len(text)`, expected result is `text` is returned.
    - `length < len(text)`, expected result is `text` is returned.
    - `length == 0`, expected result is `text` is returned.
    - `length < 0`, expected result is `text` is returned.

The very last condition is what @rdswift mentioned. The above means we should add actually add 5 more tests to properly test this function and ensure it does not break even if implementation gets changed.

Can you add the missing 5 tests?

virusMac: I added 4 tests, because I already had `$pad(abc de,x,-)` test

PR #1024, opened 2018-11-02
PICARD-1325: Allow disabling new version update checking for packagers

virusMac:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

When user uses --disable-autoupdate flag when installing or building Picard, it automatically hide "Update checking" feature from user and disable update checking code itself

# Problem
Recently the ability to automatically and manually check for program updates was added to MusicBrainz Picard. Picard release packagers should be able to completely hide this functionality and configuration settings from the user for the packages they produce.

* JIRA ticket: [PICARD-1325](https://tickets.metabrainz.org/browse/PICARD-1325)
# Solution
- added _if_ statements to determine showing all update-checking tabs and update-checking itself
- added description to flag --disable-autoupdate in _picard_install_ and _picard_build_ (setup.py)
- changed variable name from __autoupdate_ to _autoupdate_enabled_ (tagger.py)


picard/tagger.py
```diff
@@ -146,7 +146,7 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
         )
 
         self._cmdline_files = picard_args.FILE
-        self._autoupdate = autoupdate
+        self.autoupdate_enabled = True
```
rdswift: Is this `True` setting left over from your testing?  Should it be `self.autoupdate_enabled = autoupdate` instead?

picard/tagger.py
```diff
@@ -251,7 +251,8 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
         self.stopping = False
 
         # Load release version information
-        self.updatecheckmanager = UpdateCheckManager(parent=self.window)
+        if QtWidgets.QApplication.instance().autoupdate_enabled:
```
rdswift: Can't you simply use the `self.autoupdate_enabled` object here rather than the `instance()` object?

picard/ui/mainwindow.py
```diff
@@ -189,7 +189,8 @@ def keyPressEvent(self, event):
     def show(self):
         self.restoreWindowState()
         super().show()
-        self.auto_update_check()
+        if QtWidgets.QApplication.instance().autoupdate_enabled:
```
rdswift: I believe that the value is exposed as `self.tagger.autoupdate_enabled` so the `instance()` object shouldn't be required.  Same throughout the file.

picard/ui/options/general.py
```diff
@@ -17,8 +17,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from PyQt5 import QtCore
-from PyQt5.QtWidgets import QInputDialog
+from PyQt5 import QtCore,QtWidgets
```
rdswift: If you use `self.tagger.autoupdate_enabled` this likely isn't required.  If it is required, there should be a space after the comma.

picard/ui/options/general.py
```diff
@@ -74,21 +73,23 @@ def load(self):
         self.ui.server_port.setValue(config.setting["server_port"])
         self.ui.analyze_new_files.setChecked(config.setting["analyze_new_files"])
         self.ui.ignore_file_mbids.setChecked(config.setting["ignore_file_mbids"])
-        self.ui.check_for_updates.setChecked(config.setting["check_for_updates"])
-        self.ui.update_level.clear()
-        for level, description in PROGRAM_UPDATE_LEVELS.items():
-            self.ui.update_level.addItem(_(description['title']), level)
-        self.ui.update_level.setCurrentIndex(self.ui.update_level.findData(config.setting["update_level"]))
-        self.ui.update_check_days.setValue(config.setting["update_check_days"])
+        if QtWidgets.QApplication.instance().autoupdate_enabled:
```
rdswift: I believe that the value is exposed as `self.tagger.autoupdate_enabled` so the `instance()` object shouldn't be required.  Same throughout the file.

picard/ui/ui_options_general.py
```diff
@@ -67,58 +67,61 @@ def setupUi(self, GeneralOptionsPage):
...
-        sizePolicy.setVerticalStretch(0)
-        sizePolicy.setHeightForWidth(self.update_level.sizePolicy().hasHeightForWidth())
-        self.update_level.setSizePolicy(sizePolicy)
-        self.update_level.setEditable(False)
-        self.update_level.setObjectName("update_level")
-        self.gridLayout_2.addWidget(self.update_level, 0, 1, 1, 1)
-        self.verticalLayout_2.addLayout(self.gridLayout_2)
-        self.vboxlayout.addWidget(self.groupBox_3)
+
+        if QtWidgets.QApplication.instance().autoupdate_enabled:
```
rdswift: This file is automatically generated by `python setup.py build_ui` command and should not be edited manually.  I suggest that you simply hide the elements associated with the update checking settings when the ui is displayed in `picard/ui/options/general.py`.

picard/ui/ui_options_general.py
```diff
@@ -67,58 +67,61 @@ def setupUi(self, GeneralOptionsPage):
...
+            self.groupBox_3 = QtWidgets.QGroupBox(GeneralOptionsPage)
```
rdswift: Since we're likely to be referring to this to hide it, the name should probably be changed from `groupbox_3` to something more appropriate (e.g. `update_check_groupbox`).

picard/tagger.py
virusMac: Yeah, it was only for testing. Thank you.

picard/ui/ui_options_general.py
virusMac: Is it ok to add that hiding feature to load() function?
I used setEnabled() - please tell me if it's better to hide it completely

picard/ui/mainwindow.py
virusMac: Thanks, I didn't notice that.

picard/ui/ui_options_general.py
rdswift: Yes, I would add it to the `load()` method.  I think it would be better to hide it completely, that way users of a packaged version don't see something that they can't change.  It's less likely to generate questions that way.  ;-)

picard/ui/options/general.py
```diff
@@ -18,7 +18,6 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from PyQt5 import QtCore
-from PyQt5.QtWidgets import QInputDialog
```
rdswift: `QImportDialog` still needs to be imported because it is used in the `login(self)` method.

virusMac: Misclick :D

rdswift: I'm actually surprised that wasn't caught with the automatic checks.  Then again, I would have expected codacy to have caught it and it seems the codacy checking didn't run.

setup.py
```diff
@@ -181,7 +181,7 @@ class picard_build(build):
     user_options = build.user_options + [
         ('build-locales=', 'd', "build directory for locale files"),
         ('localedir=', None, ''),
-        ('disable-autoupdate', None, ''),
+        ('disable-autoupdate', None, 'disable update checking and hide settings for it'),
```
phw: Well done adding a comment for this!

virusMac: Thank you :)

picard/ui/options/general.py
```diff
@@ -19,7 +19,6 @@
 
 from PyQt5 import QtCore
 from PyQt5.QtWidgets import QInputDialog
-
```
zas: This whiteline should stay, can you just re-add it and squash again ?

virusMac: Done, thanks :)

PR #1027, opened 2018-11-08
PICARD-1375: Handle empty values in metadata sanitation

phw:
# Summary

When the metadata contains a value explicitly set to be empty (with e.g. `metadata.set("sometag", [])`) saving the file will fail in `file._script_to_filename` on the iteration for filename sanitation. The issue is that the metadata gets updated with  `metadata[]`, which deletes empty items.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1375](https://tickets.metabrainz.org/browse/PICARD-1375)

# Solution
Use `metadata.set` to set the values.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -315,8 +315,8 @@ class File(QtCore.QObject, Item):
             metadata.update(file_metadata)
         # make sure every metadata can safely be used in a path name
         for name in metadata.keys():
-            if isinstance(metadata[name], str):
-                metadata[name] = sanitize_filename(metadata[name])
+            values = [sanitize_filename(str(v)) for v in metadata.getall(name)]
+            metadata.set(name, values)
         naming_format = naming_format.replace("\t", "").replace("\n", "")
         filename = ScriptParser().eval(naming_format, metadata, self)
         if settings["ascii_filenames"]:


phw:
@samj1912 , @zas Some points for discussion:

I think from an API side it is actually pretty bad that we have both `metadata[t] = v` and `metadata.set(t, v)` which look so much alike but behave absolutely different. Furthermore the special behavior of `metadata.set`  (namely allowing arbitrary values to be set, including empty, without str conversion) is currently not used by core Picard nor the existing plugins. There are some instances of `metadata.set` calls, but they mainly ensure data is copied 1:1, but themselves don't cause e.g. empty values to be added.

Also the reading methods `metadata.get(t)` and `metadata[t]` behave exactly the same, so its inconsistent to have the setters behave different.

But I was not brave enough to change it. Actually it was my first approach to unify both methods, but it is an API breaking change after all. What are your opinions?

I personally would refactor the `Metadata` class as follows:

1. Make `set` and `__setitem__` be the same
2. Make `add` and `add_unique` also do the str conversion for consistency
3. Optional: If needed introduce some functions `setraw` and `getraw` to really set and get raw values without any conversion

zas:
> I personally would refactor the `Metadata` class as follows:

Agreed. +1

phw:
Ok, then let's get this merged as a bug fix and I will prepare a separate PR for the refactoring.

PR #1029, opened 2018-11-11
PICARD-1043: Write movement tags

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Picard does not support the tags for movements as supported e.g. by iTunes, but also other players like MusicBee.

* JIRA ticket (_optional_): [PICARD-1043](https://tickets.metabrainz.org/browse/PICARD-1043)

# Solution
This adds support for the following tags:

|Internal name|ID3|MP4|
|-|-|-|
|movement|MVNM|©mvn|
|movementnumber|MVIN|©mvi|
|movementtotal|MVIN|©mvc|
|showmovement||shwm|

`showmovement` is a iTunes specific flag. If set, iTunes will show the work and movement instead of the track title for display in the list E.g. the following screenshots shows two entries with the same metadata, but one has `showmovement` set, and it displays the work name + movement number and movement instead of the track title

grafik

This flag can also be toggled in the UI by choosing between "work name" and "song":

grafik

iTunes only supports writing and reading this tag for MP4 files, so I have disabled it for all other formats for now. If we find out another player supports these, too, and also has some way to read it from other formats we can change this.

In opposite to the previous attempt at #695 this PR does not attempt to define how to fill these tags from MusicBrainz data. Rather the goal is to provide support for these tags, so they can be used by users in scripting or plugins. E.g. the popular Classical Extras plugin can be configured to make use of this.

# Action


phw:
Related: https://github.com/metabrainz/picard/pull/1030

vzell:
Yes please... +1 for MusicBee

phw:
I merge this but will look into the MusicBee compatibility in a second PR

PR #1033, opened 2018-11-13
PICARD-1320: Fix text color for modified files for dark themes

phw:
# Summary
With dark theme modified files are shown with black text on dark background.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The text color for modified files was supposed to be theme aware, but generating the color failed. Creating a QColor object from a QPalette ColorRole does not work (maybe it worked in Qt4). The result was always a black color.

* JIRA ticket (_optional_): [PICARD-1320](https://tickets.metabrainz.org/browse/PICARD-1320)

# Solution
Switch the text color to use the color for foreground text. When the "color_modified" option gets initialized we don't have access to the current palette, so I decided to remove the "color_modified" config option entirely, as it defaulted to the standard foreground text color anyway and we don't make use of these options currently anyway.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -144,7 +144,7 @@ class MainPanel(QtWidgets.QSplitter):
         }
         FileItem.file_colors = {
             File.NORMAL: TreeItem.text_color,
-            File.CHANGED: config.setting["color_modified"],
+            File.CHANGED: TreeItem.text_color,
             File.PENDING: config.setting["color_pending"],
             File.ERROR: config.setting["color_error"],
         }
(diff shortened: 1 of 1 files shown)


picard/ui/itemviews.py
```diff
@@ -144,7 +144,7 @@ def __init__(self, window, parent=None):
         }
         FileItem.file_colors = {
             File.NORMAL: TreeItem.text_color,
-            File.CHANGED: config.setting["color_modified"],
+            File.CHANGED: TreeItem.text_color,
```
zas: Do i miss something or this is setting the same color for Normal and Changed files ?

phw: It is, that's another reason why I removed that option. The reason probably is that we render that red background color for changed files anyway. Here is a screenshot how it looks like in Picard:

grafik

phw:
Here is another screenshot with how this renders with the patch on a dark theme:

grafik

Without the patch the text for changed files is black, which can be hard to read:

grafik

PR #1034, opened 2018-11-13
PICARD-1060: Fix collections menu not looking and behaving like a normal menu

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The collection sub menu currently does not display and behave like a normal menu. Mainly the menu items render without margins and do not highlight on activation.

* JIRA ticket (_optional_): [PICARD-1060](https://tickets.metabrainz.org/browse/PICARD-1060)

# Solution
The reason for the differences is that we render those menu items with a custom widget which we need for the tri-state checkboxes.

Implement the necessary changes to render with margin and background highlighting. Capture the mouse over and menu item activated events to properly mark mouse overed items as active and render active items with background.

I tested this on Gnome and KDE (KDE with a light and a dark theme). In both cases the menu rendered exactly as e.g. the "other versions" menu.

Since we strictly set the Fusion theme this should reliable work. I have not tested with any other theme. I have also not yet tested on macOS, but even there it should be at least an improvement to the really broken rendering we currently have. The highlight color or margins might be off, though.

# Action

### DIFF
--- a/picard/ui/collectionmenu.py
+++ b/picard/ui/collectionmenu.py
@@ -2,6 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 # Copyright (C) 2013 Michael Wiencek
+# Copyright (C) 2018 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -36,9 +37,11 @@ class CollectionMenu(QtWidgets.QMenu):
     def __init__(self, albums, *args):
         super().__init__(*args)
         self.ids = set(a.id for a in albums)
+        self._ignore_update = False
         self.update_collections()
 
     def update_collections(self):
+        self._ignore_update = True
         self.clear()
         self.actions = []
         for id_, collection in sorted(user_collections.items(),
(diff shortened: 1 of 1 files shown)


phw:
Needs some more work. Updating the UI needs some lock, otherwise refreshing and UI updates can interfere and cause crashes

zas:
I tested, and it seems to work fine.
Though, i wonder if we need this checkbox at all, afaik it's only used to mark the current release. If we get rid of the checkbox, we could use a normal menu right? What about just write the text in bold or just prepend/append some strings to mark it?

phw:
So I did some additional tweaking and I am now quite satisfied with the result. This now uses the paint event to draw a menuitem control, which gives a more accurate representation of how a menu item gets highlighted. On  Linux and Windows with Fusion theme this now looks exactly like any other menu item. Also on Windows with the Windows style the highlighting would look correct.

On Mac the colors are off a bit. But actually the menus are rendered a bit strange by Qt anyway: On macOS one can specify an accent color, but Qt renders the menus always with the same default blue. Only the custom collections menu gets rendered with a shade of the accent color (but not the actual accent color as used by other macOS apps) now.
 
But it looks way better then before without margins and no highlighting whatsoever. So I think this is fine to merge, macOS tweaking if necessary and possible could be done separately.

phw:
> Though, i wonder if we need this checkbox at all, afaik it's only used to mark the current release.

It's actually not about a checkbox in general, it is specifically for a tri-state checkbox. Having only checkmarks would be no problem, that is supported by the menus by default. But we want to have those checkboxes with a "some" state, so if you select multiple albums and some are in a collection and some aren't it gets marked accordingly. See this:

grafik

As this is working now and looking good on all platforms I would keep it

zas:
> As this is working now and looking good on all platforms I would keep it

Ok for me, thanks for the explanation.

PR #1036, opened 2018-11-14
PICARD-1425: Support all movement tags for APE, Vorbis and MP3

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

This finishes up the work started in https://github.com/metabrainz/picard/pull/1029 and adds full support for the movement tags in ID3, Vorbis and APE. The updated tag mapping is then:

|Internal name|ID3|MP4|Vorbis|APE|
|-|-|-|-|-|
|movement|MVNM|©mvn|MOVEMENTNAME|MOVEMENTNAME|
|movementnumber|MVIN|©mvi|MOVEMENT|MOVEMENT|
|movementtotal|MVIN|©mvc|MOVEMENTTOTAL|MOVEMENTTOTAL|
|showmovement|TXXX:SHOWMOVEMENT|shwm|SHOWMOVEMENT|SHOWMOVEMENT|

This is compatible with how MusicBee and MP3Tag handle those tags. In MusicBee it looks as follows (very similar UI to the iTunes one):

grafik

* JIRA ticket (_optional_): [PICARD-1425](https://tickets.metabrainz.org/browse/PICARD-1425)
* JIRA ticket (_optional_): [PICARD-1043](https://tickets.metabrainz.org/browse/PICARD-1043)

# Solution

# Action

### DIFF
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -62,6 +62,10 @@ class APEv2File(File):
         "Barcode": "barcode",
         "EncodedBy": "encodedby",
         "Language": "language",
+        "MOVEMENT": "movementnumber",
+        "MOVEMENTNAME": "movement",
+        "MOVEMENTTOTAL": "movementtotal",
+        "SHOWMOVEMENT": "showmovement",
         "MUSICBRAINZ_ALBUMSTATUS": "releasestatus",
         "MUSICBRAINZ_ALBUMTYPE": "releasetype",
         "musicbrainz_trackid": "musicbrainz_recordingid",
@@ -212,14 +216,10 @@ class APEv2File(File):
     def supports_tag(cls, name):
         unsupported_tags = {
             'gapless',
-            'movement',
-            'movementnumber',
-            'movementtotal',
             'musicip_fingerprint',
             'podcast',
             'podcasturl',
             'show',
-            'showmovement',
             'showsort',
             }
         return bool(name) and name not in unsupported_tags
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -169,6 +169,7 @@ class ID3File(File):
         'Artists': 'artists',
         # 'Work': 'work',
         'Writer': 'writer',
+        'SHOWMOVEMENT': 'showmovement',
     }
     __rtranslate_freetext = dict([(v, k) for k, v in __translate_freetext.items()])
     __translate_freetext['Work'] = 'work'  # Always read, but writing depends on itunes_compatible_grouping
@@ -527,7 +528,7 @@ class ID3File(File):
 
     @classmethod
     def supports_tag(cls, name):
-        unsupported_tags = {'showmovement'}
+        unsupported_tags = {}
         return ((name and not name.startswith("~") and name not in unsupported_tags)
                 or name in ("~rating", "~length")
                 or name.startswith("~id3"))
--- a/picard/formats/vorbis.py
+++ b/picard/formats/vorbis.py
@@ -55,8 +55,10 @@ class VCommentFile(File):
     _File = None
 
     __translate = {
-        "musicbrainz_trackid": "musicbrainz_recordingid",
+        "movement": "movementnumber",
+        "movementname": "movement",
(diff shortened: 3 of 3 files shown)


phw:
As testing data for those tags where already added in #1029 and this just enables them for other formats I did not add any additional tests here.

zas:
> As testing data for those tags where already added in #1029 and this just enables them for other formats I did not add any additional tests here.

Ok, great :)

dgasaway:
FYI, for Vorbis and FLAC, it has been the practice by some (myself included) to use the PART tag for the movement for many years.  There is also PARTNUMBER, though I chose not to use it myself.

https://www.legroom.net/index.php/2009/05/09/ogg-vorbis-and-flac-comment-field-recommendations

dgasaway:
Also, I should point out that I also regularly use PART for things not properly described as MOVEMENT, such as opera numbers and named parts of popular songs.

dgasaway:
By default?  No, I do not.  I use foobar2000, which has the flexibility to add extensive support for WORK and PART on my own.  Which I did. :)

phw:
I'll do some additional research before merging this.

phw:
After some e-mail exchange with @ijabz I think we should proceed with this as implemented here. If somebody wants to use the PARTS tag this can be done with scripting.

There is usually a difference between how PARTS and MOVEMENTNAME get interpreted:

> There is subtle difference between MOVEMENT and PART, PART includes the movement number as well as name
>
>e.g
>
>PART : I. Allegretto - Allegro non troppo
>
>MOVEMENT : Allegretto - Allegro non troppo

dgasaway:
I don't believe that statement is correct.  Perhaps that is how many people use PART (including myself), but as I pointed out, there is a PARTNUMBER to contain the movement number.  By the way, how do you intend to limit this implementation to only classical works?

phw:
> Perhaps that is how many people use PART (including myself), but as I pointed out, there is a PARTNUMBER to contain the movement number. 

Yes, I think the above statement was mainly for the Room player. But the actual point is that it is easy to use and scripting to set a PART tag as and if required.

> By the way, how do you intend to limit this implementation to only classical works?

For the current change this is out of scope. Picard 2.1 will only provide those tags, filling them with meaningful data is left to the user by using scripting and plugins.

In addition to the existing Classical Extras plugin, which can be used for this, there will be a new Work & Movement plugin. That plugin limits the use of work and movement tags to multi-part works. Based on experience with plugins we might integrate such functionality directly into Picard in some future version.

See https://github.com/metabrainz/picard-plugins/tree/2.0/plugins/workandmovement

dgasaway:
> For the current change this is out of scope. Picard 2.1 will only provide those tags, filling them with meaningful data is left to the user by using scripting and plugins.

I don't follow.  I haven't checked the code, but my understanding is that this will populate these tags in files with APE and vcomment tags.  Now you are saying that the values that they save to file will be meaningless and it will be up to the user to use scripts and plugins to make them meaningful?  That's an odd statement.

dgasaway:
Also, my point RE: classical was that part-of ARs are used for more things than movements in classical works.

phw:
> I don't follow. I haven't checked the code, but my understanding is that this will populate these tags in files with APE and vcomment tags.

The changes here and in #1029 just make Picard aware of the various movement tags. That means Picard knows the tags and how to load and save them to various tagging formats. This allows the user to make use of these tags. But with these changes Picard does not fill those tags with any data, for now this is the job of scripts and / or plugins (or manually of course).

This is actually the prerequisite for plugins to fully support use cases like this. E.g. at the moment the Classical Extras plugin can get you movement information, but you cannot use Picard to save this data in a way e.g. iTunes understands. The documentation of Classical Extras goes into detail how to use other tools to rewrite tags accordingly. But that's less than optimal. With Picard 2.1 this use case will be supported.

> Also, my point RE: classical was that part-of ARs are used for more things than movements in classical works.

MusicBrainz provides attributes for movement and act, which can be used for getting this right.

dgasaway:
Thank you for taking the time to explain.

PR #1040, opened 2018-11-15
PICARD-1418: Setup default Qt translations

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The keyboard shortcuts currently do not show up localized, so e.g. German users also see "Ctrl+S" for saving files, but actually should see it localized "Strg+S". Same is true for default dialogs, e.g. file open dialogs currently always show up with English button labels.

* JIRA ticket (_optional_): [PICARD-1418](https://tickets.metabrainz.org/browse/PICARD-1418)

# Solution
Setup default Qt translations. This makes sure standard dialogs and keyboard shortcut hints show up in the user selected language.

# Action

### DIFF
--- a/picard/i18n.py
+++ b/picard/i18n.py
@@ -23,6 +23,8 @@ import locale
 import os.path
 import sys
 
+from PyQt5.QtCore import QLocale
+
 builtins.__dict__['N_'] = lambda a: a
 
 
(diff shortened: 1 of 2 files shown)


phw:
It doesn't look like that it ever works, interesting we never noticed :(

PR #1041, opened 2018-11-15
Add phw to authors in about

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Me not being listed in About copyright section

# Solution

# Action


zas:
Btw, i think we should improve this somehow, any change (including adding an author, or changing a date) leads to new string, which means translation updates, etc... which is imho not strictly needed.
Perhaps we could split this in different parts, one having copyright dates & authors, etc...
What do you think ?

phw:
Yes, agreed. Especially as changes in such a long field are hard for a translator to spot. In professional translation system text gets segmented, usually by sentence, so this would be similar.

PR #1042, opened 2018-11-16
Split About text in smaller parts to prevent unneeded re-translation.

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

On each new year or when adding a new author, the whole About string has to be translated again.

# Solution

Move variable bits outside of it.

# Action

### DIFF
--- a/picard/ui/options/about.py
+++ b/picard/ui/options/about.py
@@ -53,7 +53,7 @@ class AboutOptionsPage(OptionsPage):
         # TR: Replace this with your name to have it appear in the "About" dialog.
         args["translator-credits"] = _("translator-credits")
         if args["translator-credits"] != "translator-credits":
-            # TR: Replace LANG with language you are translatig to.
+            # TR: Replace LANG with language you are translating to.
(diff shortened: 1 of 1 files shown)


zas:
> Looks good to me, I would only change the naming of `flaticon-credits`, this seems to be too specific. I would call it just `icon-credits` or maybe `artwork-credits`.

Actually it was intentional, those credits are specific to flaticon, if, in the future, we have to add more icons credits they are likely to be in another set.

zas:
> I thought the credits for samj1912 do not refer to flaticons, or did I get this wrong? Otherwise I'm fine with this.

You may be right. @samj1912 ??

samj1912:
Yeah, the ones I made are not flat icon ones.

zas:
> Yeah, the ones I made are not flat icon ones.

Ok, i changed to `icons` to be less specific.
I'm not really happy though, i'd like us to be able to output credits to console, but the current stuff is very UI-oriented, with urls, and emails, i may rework this.
About emails, yours isn't displayed because of `<` and `>` around it, was it intentionnal ?

PR #1047, opened 2018-11-23
PICARD-1395: MusicBrainz genres

phw:
# Summary
Picard should support the new genres introduced to MusicBrainz recently. By default this will give much better and cleaner results to users looking for adding genres to their files than using all folksonomy tags.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1395](https://tickets.metabrainz.org/browse/PICARD-1395)

# Solution
This mostly reuses the existing folksonomy tags options but with a few additions:

1. There is a new top option "Use genres from MusicBrainz" (`use_genres`). This enables / disables loading genres from MusicBrainz entirely.
2. The old "Use folksonomy tags as genre" (`folskonomy_tags`) option still is available, but it is a sub option now which triggers only to use all folksonomy tags instead of just genres
3. For existing users there is a config upgrade which enables  `use_genres` if `folskonomy_tags` was enabled before. The default is still that no genres will be used by default
4. I also renamed the other options to have genre instead of tag in the names to avoid future confusion

grafik

# Action
Not part of this change is fixing the fact that currently only genres / tags for a release are considered. Using genres / tags from recordings, release-groups or artists is currently broken due to a shortcoming in the JSON web service, see [PICARD-1288](https://tickets.metabrainz.org/browse/PICARD-1288).


phw:
One question I had no proper answer to was how to deal with the exclusion list. I figured it would still be useful even if just using tags, so I kept it globally (and not just let's say for the folskonomy tags). But the defaults don't make much sense for genres.

In the end I thought it would be best to have it apply for both folskonomy tags and proper genres, but have the default set to be useful for people who set "Use folksonomy tags"

picard/config_upgrade.py
```diff
@@ -246,6 +241,27 @@ def upgrade_to_v2_0_0_dev_3():
             _s[opt] = _CAA_SIZE_COMPAT[value]
 
 
+def upgrade_to_v2_1_0_dev_1():
+    """Upgrade genre related options"""
+    print("upgrade_to_v2_1_0_dev_1")
```
zas: `print` leftover ;)

picard/config_upgrade.py
```diff
@@ -246,6 +241,27 @@ def upgrade_to_v2_0_0_dev_3():
...
+    _s = config.setting
+    if _s["folksonomy_tags"]:
```
zas: Check if `"folksonomy_tags" in _s` ?

zas:
> In the end I thought it would be best to have it apply for both folskonomy tags and proper genres, but have the default set to be useful for people who set "Use folksonomy tags"

Yes, might be the best solution.

picard/config_upgrade.py
phw: Is this needed? We access config.setting['folksonomy_tags'] elsewhere without test. But I could also do a `if _s.get('folksonomy_tags')`

picard/config_upgrade.py
phw: Fixed

picard/config_upgrade.py
phw: I replaced it with `_s.get('folksonomy_tags')`

phw:
> Title in the dialog is currently "Ignore genres:", but none of the list are actually genres...

You see my problem with this :D Suggestion?

zas:
> You see my problem with this :D Suggestion?

Since you want to use this list for both genres and folksonomy tags, what about "Folksonomy tags or genres to exclude (comma-separated list):" ?

phw:
I labeled it "Genres or folksonomy tags to exclude (comma-separated list):"

picard/ui/ui_options_genres.py
```diff
@@ -0,0 +1,118 @@
+# -*- coding: utf-8 -*-
+
+# Automatically generated - don't edit.
+# Use `python setup.py build_ui` to update it.
+
+from PyQt5 import QtCore, QtGui, QtWidgets
```
zas: QtGui is unused (reported by Codacy)

zas: Since it's autogenerated, i guess we can ignore

phw: Yes, w e should ignore this. This is in all files autogenerated from UI. Maybe we can teach codacy to ignore this issue for the affected files.

PR #1049, opened 2018-11-25
Codacy ignore autogenerated ui files

phw:
# Summary
This ignores the autogenerated ui_*.py files. See discussion on https://github.com/metabrainz/picard/pull/1047

**NOTE**: Right now this branch is intentionally on top of #1047 to test its functionality, so please do not merge yet. Once I know it is working as expected I will rebase the changes here and leave only the addition of `.codacy.yml`

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/.codacy.yml
+++ b/.codacy.yml
@@ -1,0 +1,2 @@
+exclude_paths:
+  - 'picard/ui/ui_*.py'


phw:
Works. I also have the suspicion we already have ignored those files in the Codacy admin, because existing ui_*.py files don't show up in analysis, but newly created files, such as `ui_genres.py` gets analysed. We probably should rely on the config file only and remove ignored files from Codacy admin, but I don't have access to it.

mineo:
> I also have the suspicion we already have ignored those files in the Codacy admin

That suspicion is correct! I've removed all the overrides in the Codacy UI.

PR #1050, opened 2018-11-25
PICARD-1371: Do not allow installation on 32-bit Windows

phw:
# Summary
Currently the Windows installer of Picard will happily install on a 32bit Windows, but since we only provide 64 bit builds the installed application fails to launch with some Windows incompatibility warning.

This patch checks if the installer is running on a 64bit system and if not aborts the installation with an appropriate error message.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


phw:
Tested on 64bit Win 10 and 32 bit Win 7 systems.

PR #1053, opened 2018-11-26
PICARD-1398: Fix installing locales when running setup.py bdist_wheel

phw:
# Summary
Builds for Snap package do not contain locale files.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Snap packages use `setup.py bdist_wheel` for generating the package. Currently the way our locale files are built causes them to be not included in the data package of the wheel.

* JIRA ticket (_optional_): [PICARD-1398](https://tickets.metabrainz.org/browse/PICARD-1398)

# Solution
Fix installation of locale files to be in the proper data file used by the setuptools. This ensures not matter how the final packaging works the locale files will be included.

For now only tested locally with normal setup.py install and bdist_wheel commands. Before merging I want to test the CI builds for Win and Mac.

# Action

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -160,8 +160,9 @@ class picard_install(install):
     def finalize_options(self):
         install.finalize_options(self)
         if self.install_locales is None:
-            self.install_locales = '$base/share/locale'
-            self._expand_attrs(['install_locales'])
+            self.install_locales = os.path.join(self.install_data, 'share', 'locale')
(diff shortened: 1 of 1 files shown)


phw:
Both the Windows and Mac builds work and properly include the locales

PR #1059, opened 2018-11-29
PICARD-1219: Do not save empty TIPL/TMCL/IPLS frames to ID3

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Picard saves empty frames to ID3, even if no data is set for them:

v2.3: IPLS  
v2.4: TIPL, TMCL

* JIRA ticket (_optional_): [PICARD-1219](https://tickets.metabrainz.org/browse/PICARD-1219)

# Solution

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -461,8 +461,10 @@ class ID3File(File):
             elif not name.startswith("~") and name not in self.__other_supported_tags:
                 tags.add(self.build_TXXX(encoding, name, values))
 
-        tags.add(tmcl)
-        tags.add(tipl)
+        if tmcl.people:
+            tags.add(tmcl)
+        if tipl.people:
+            tags.add(tipl)
 
         self._remove_deleted_tags(metadata, tags)
 
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -377,7 +377,11 @@ class ID3File(File):
             elif name.startswith('performer:'):
                 role = name.split(':', 1)[1]
                 for value in values:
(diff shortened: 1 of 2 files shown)


phw:
The changes here caused a regression, the issue here was originally introduced when fixing [PICARD-748](https://tickets.metabrainz.org/browse/PICARD-748) in https://github.com/metabrainz/picard/commit/cb75542511cfb848f503f0be94ff8c966e6d9843

The original issue was that ID3 v2.3 compatibility saves both `TMCL` and `TIPL` as `IPLS`, but on loading loads it all back to `TIPL`. That effectively converted existing `IPLS` frames in the v2.3 file to `TIPL`, but our saving code duplicated them as `TMCL`.

I think I found a nice solution by saving only `TIPL` and never `TMCL` when in v2.3 compatibility. That way the mutagen internal conversion only has to convert `TIPL` to `IPLS` and we get no duplication.

phw:
A positive side effect of this is that it will also actually preserve performer tags. Previously Picard would remove existing performer tags, even if the UI showed them as being unchanged and the new metadata did not provide any tags.

With this PR this issue is fixed as well and performer tags will be kept unchanged and behave like any other tag. I added tests to check this.

phw:
I added https://tickets.metabrainz.org/browse/PICARD-1431 for the existing tag preservation bug. I will submit a follow up PR once this one got merged with additional fix for the `TCMP` frame.

PR #1061, opened 2018-11-30
Pep8 fixes

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


zas:
> Great. How good is our PEP8 compliance with this? Could we already add PEP8 validation to our CI?

Well, we aren't 100% compliant since we have our own way to format few things (like using no space around plus in (i+1), or lists formatted with whitespaces to form columns), but i guess we could move forward, and enable most checks but few. Let's do that after 2.1.0 release though

PR #1064, opened 2018-12-06
PICARD-1434: Allow acoustid_id to be deleted by scripting

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Explicitly deleting the `acoustid_id` via e.g. `$delete(acoustid_id)` does not work. But this is a valid use case, as a user might want to clear this tag from files.

* JIRA ticket (_optional_): [PICARD-1434](https://tickets.metabrainz.org/browse/PICARD-1434)

# Solution

# Action
Just noting that doing `$set(acoustid_id,somevalue)` is not enabled by this patch. I think this makes sense, as a script has no way to set this to a meaningful value and usage of this tag in general should be restricted to AcoustId.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -178,7 +178,7 @@ class File(QtCore.QObject, Item):
         for tag, values in saved_metadata.items():
             self.metadata.set(tag, values)
 
-        if acoustid:
+        if acoustid and not "acoustid_id" in metadata.deleted_tags:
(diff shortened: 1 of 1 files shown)


zas:
> @zas We need a 2.1.0 release on JIRA, can you create one?

done

PR #1065, opened 2018-12-06
PICARD-1436 Text extraction of "title" and "label" for translation

rdswift:
# Summary

Some text items are being improperly extracted for translation.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: See the discussion at [https://community.metabrainz.org/t/strings-label-and-title-where-are-they/408487](https://community.metabrainz.org/t/strings-label-and-title-where-are-they/408487) for more information.

# Problem

Some text items are being improperly extracted for translation.

* JIRA ticket: [PICARD-1436](https://tickets.metabrainz.org/browse/PICARD-1436) Text extraction of "title" and "label" for translation

# Solution

Remove text extraction code.

# Action

Please review carefully (as always) because I still don't completely understand the i18n stuff.

### DIFF
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -218,7 +218,7 @@ class InterfaceOptionsPage(OptionsPage):
         list_item = ToolbarListItem(action)
         list_item.setToolTip(_('Drag and Drop to re-order'))
         if action in self.TOOLBAR_BUTTONS:
-            list_item.setText(_(self.TOOLBAR_BUTTONS[action]['label']))
+            list_item.setText(self.TOOLBAR_BUTTONS[action]['label'])
             list_item.setIcon(icontheme.lookup(self._get_icon_from_name(action), icontheme.ICON_SIZE_MENU))
         else:
             list_item.setText(self.SEPARATOR)
@@ -287,7 +287,7 @@ class AddActionDialog(QtWidgets.QDialog):
 
         layout = QtWidgets.QVBoxLayout(self)
 
-        self.action_list = sorted([[_(self.parent().TOOLBAR_BUTTONS[action]['label']), action]
+        self.action_list = sorted([[self.parent().TOOLBAR_BUTTONS[action]['label'], action]
                                    for action in action_list])
 
         self.combo_box = QtWidgets.QComboBox(self)
--- a/picard/util/checkupdate.py
+++ b/picard/util/checkupdate.py
@@ -149,7 +149,7 @@ class UpdateCheckManager(QtCore.QObject):
                     _("Picard Update"),
                     _("There is no update currently available for your subscribed update level: {update_level}\n\n"
                       "Your version: {picard_old_version}\n").format(
-                        update_level=_(PROGRAM_UPDATE_LEVELS[self._update_level]['title']) if self._update_level in PROGRAM_UPDATE_LEVELS else _('unknown'),
+                        update_level=PROGRAM_UPDATE_LEVELS[self._update_level]['title'] if self._update_level in PROGRAM_UPDATE_LEVELS else _('unknown'),
                         picard_old_version=PICARD_FANCY_VERSION_STR,
                     ),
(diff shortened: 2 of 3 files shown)


phw:
Thanks for looking into this. But unfortunately this is not the proper fix for this, the content of the variables needs to be put through the gettext function in order to be shown in proper translation. E.g. the content of PROGRAM_UPDATE_LEVELS[self._update_level]['title'] is marked for translation with N_()

What I would try is moving the complex variable access out of _(). The extraction seem to get confused with the dictionary access, so having something like this

    _(d["something"])

Turned into

    label = d["something"]
    _(label)

Should work

zas:
That's not the proper fix, because it's a bug in Babel: see https://github.com/python-babel/babel/issues/415 

Since it's a workaround, i'd name temporary variables so we can easily remove them when babel will be fixed.

What's about `babel_415_workaround` ? ;)

rdswift:
I kept the original code (commented out) and marked all the temporary workarounds with TODO comments.  Also changed the temporary variable names as suggested.   Somehow I expect Codacy to complain.  :wink:

picard/util/checkupdate.py
```diff
@@ -144,12 +144,15 @@ def _display_results(self):
                 webbrowser2.open(self._available_versions[key]['urls']['download'])
         else:
             if self._show_always:
+                # TODO: Remove temporary workaround once https://github.com/python-babel/babel/issues/415 has been resolved.
+                babel_415_workaround = PROGRAM_UPDATE_LEVELS[self._update_level]['title'] if self._update_level in PROGRAM_UPDATE_LEVELS else 'unknown'
```
phw: Currently this will loose us the string `'unknown'`, we have to mark it as extractable by using `N_('unknown')`

Actually this is the one place here where I would not treat this as a workaround but actually clean it up. This one line conditional with deep variable access is a bit awkward to read. I personally would clean it up to be:

```
if self._update_level in PROGRAM_UPDATE_LEVELS:
    update_level = PROGRAM_UPDATE_LEVELS[self._update_level]['title']
else:
    update_level = N_('unknown')
```

Way easier to read IMHO

zas: Thanks for noticing this, i missed it :(

picard/ui/options/interface.py
```diff
@@ -218,7 +218,10 @@ def _insert_item(self, action, index=None):
         list_item = ToolbarListItem(action)
         list_item.setToolTip(_('Drag and Drop to re-order'))
         if action in self.TOOLBAR_BUTTONS:
-            list_item.setText(_(self.TOOLBAR_BUTTONS[action]['label']))
+            # TODO: Remove temporary workaround once https://github.com/python-babel/babel/issues/415 has been resolved.
+            babel_415_workaround = self.TOOLBAR_BUTTONS[action]['label']
+            list_item.setText(_(babel_415_workaround))
+            #list_item.setText(self.TOOLBAR_BUTTONS[action]['label'])
```
zas: I'd rather remove this commented out code

picard/ui/options/interface.py
```diff
@@ -287,8 +290,14 @@ def __init__(self, action_list, *args, **kwargs):
...
 
-        self.action_list = sorted([[_(self.parent().TOOLBAR_BUTTONS[action]['label']), action]
-                                   for action in action_list])
+        # TODO: Remove temporary workaround once https://github.com/python-babel/babel/issues/415 has been resolved.
+        babel_415_workaround_list = []
+        for action in action_list:
+            babel_415_workaround = self.parent().TOOLBAR_BUTTONS[action]['label']
+            temp_list.append([_(babel_415_workaround), action])
+        self.action_list = sorted(babel_415_workaround_list)
+        #self.action_list = sorted([[self.parent().TOOLBAR_BUTTONS[action]['label'], action]
```
zas: I'd rather remove this commented out code

PR #1072, opened 2018-12-14
PICARD-1443: Sort secondary release type sliders alphabetically

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

The secondary release types in Options > Metadata > Preferred Releases are currently unsorted, which makes it hard to find the correct type. In the release editor on MB.org they are sorted alphabetically, we should do the same.

I would still keep the primary types at the start of the list in their current order. This is again also how the release editor deals with it.

* JIRA ticket (_optional_): [PICARD-1443](https://tickets.metabrainz.org/browse/PICARD-1443)

# Solution

# Action


phw:
The current unsorted look:

grafik

With sorting:

grafik

phw:
This is non-sense, we have to sort the localized values

PR #1073, opened 2018-12-14
Disable the tooltips for sliders on Windows

phw:
# Summary
The behavior of the preferred release sliders is buggy on Windows. The tooltips are stealing the focus, causing the drag operation to stop. Disabled the tooltips on Windows until we find a better solution.

See the forums: https://community.metabrainz.org/t/please-test-new-picard-windows-builds/400717/42

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-518](https://tickets.metabrainz.org/browse/PICARD-518)

# Solution

# Action

### DIFF
--- a/picard/ui/options/releases.py
+++ b/picard/ui/options/releases.py
@@ -19,6 +19,7 @@
 
 from locale import strxfrm
 from operator import itemgetter
+import sys
 
 from PyQt5 import (
     QtCore,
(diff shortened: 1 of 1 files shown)


phw:
@zas A proper solution for this would be to implement our own tooltip as a QWidget. That is what other projects facing this issue seem to do, see e.g. https://mailman.videolan.org/pipermail/vlc-commits/2016-May/035745.html

But I am unsure if we should invest the time or if we rather look into a proper rework of the UI here. The forums discussion has some nice ideas.

What shall we do with https://tickets.metabrainz.org/browse/PICARD-518 ? I tend to re-open it.

zas:
> But I am unsure if we should invest the time or if we rather look into a proper rework of the UI here. The forums discussion has some nice ideas.

I'd just disable it for the release, this part of the UI is a source of complaints since a long time, hard to know what it does for the newbie.



> I tend to re-open it.

I'd rather open a new issue: redesign release preferences UI ;)

phw:
> I'd rather open a new issue: redesign release preferences UI ;)

I did just that: https://tickets.metabrainz.org/browse/PICARD-1444

@zas Some side note: My SSD in my laptop died yesterday (two weeks after support expired of course). I'm not sure how fast I can get a replacement, but I currently have no proper setup for Picard development :(

PR #1077, opened 2018-12-30
PICARD-1450: Fix language labels for zh_CN and zh_TW

zas:
zh_CN (Chinese (China)) should read "中文（中国大陆）" not "中文".
zh_TW (Chinese (Taiwan)) should read "中文（台灣）" not “傳統中文".

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1450](https://tickets.metabrainz.org/browse/PICARD-1450)

# Solution

# Action

### DIFF
--- a/picard/const/languages.py
+++ b/picard/const/languages.py
@@ -71,6 +71,6 @@ UI_LANGUAGES = [
     #(u'ta', u'தமிழ்', N_(u'Tamil')),
     (u'tr', u'Türkçe', N_(u'Turkish')),
     (u'uk', u'Украї́нська мо́ва', N_(u'Ukrainian')),
-    (u'zh_CN', u'中文', N_(u'Chinese')),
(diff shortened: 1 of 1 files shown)


phw:
Ok for this change. But probably the English labels should also be updated. Either change to "Chinese (simplified)" and "Chinese (traditional)" (because the primary difference here is the script) or probably better to "Chinese (China)" and "Chinese (Taiwan)" as suggested in the ticket.

zas:
> Ok for this change. But probably the English labels should also be updated. Either change to "Chinese (simplified)" and "Chinese (traditional)" (because the primary difference here is the script) or probably better to "Chinese (China)" and "Chinese (Taiwan)" as suggested in the ticket.

Makes sense, done ;)

PR #1080, opened 2019-01-11
PICARD-1453: Fix drop events on coverartbox

antlarr:
It seems at some point it has become neccessary to accept
dragMoveEvents as well as dragEnterEvents in order to
let Qt know that we want to receive drop events.

# Summary

Fix drop events on coverartbox which were previously ignored by what appears to be a behaviour change in Qt.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

When dragging  a file from a file manager (or an image from the browser) and dropping it over the CoverArtBox, ActiveLabel.dragEnterEvent is correctly called and it's accepting the event, but ActiveLabel.dropEvent is not called at all because some logic looks to be preventing it.

* JIRA ticket: [PICARD-1453](https://tickets.metabrainz.org/browse/PICARD-1453)

# Solution

It seems at some point it has become necessary to also accept the dragMoveEvent in order to Qt to call the dropEvent method, which is what this pull request does.

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -66,10 +66,12 @@ class ActiveLabel(QtWidgets.QLabel):
         if self.active and event.button() == QtCore.Qt.LeftButton:
             self.clicked.emit()
 
-    def dragEnterEvent(self, event):
+    @staticmethod
+    def dragEnterEvent(event):
         event.acceptProposedAction()
 
-    def dragMoveEvent(self, event):
+    @staticmethod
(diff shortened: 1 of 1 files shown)


antlarr:
@phw: Thanks for merging the pr. In fact, I'm also curious, do you mean you dragged a jpg/png file to the cover art box and the image changed? What Qt version do you have in your system? (it's 5.12 here).

PR #1083, opened 2019-01-29
PICARD-1382: Fail macOS packaging scripts on error

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

The packaging scripts for macOS currently do not fail on error but continue running and return with success.

* JIRA ticket (_optional_): [PICARD-1382](https://tickets.metabrainz.org/browse/PICARD-1382)

# Solution
Use `set -e` to have the script terminate if any command fails.

# Action
Note: Expected result of this PR is that the build fails due to https://tickets.metabrainz.org/browse/PICARD-1456 . Actually a failing build indicates this PR is a success. If the build succeeds this PR is probably not working as expected ;)

### DIFF
--- a/scripts/package-osx.sh
+++ b/scripts/package-osx.sh
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+set -e
+
 if [ -z "$TRAVIS_TAG" ]
 then
     python3 setup.py patch_version --platform=osx_$TRAVIS_OSX_IMAGE
--- a/.travis.yml
+++ b/.travis.yml
@@ -49,11 +49,11 @@ before_install:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then source bin/activate; fi
 - "$INSTALL_DEPS"
 - "$PYTEST"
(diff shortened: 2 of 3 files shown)


phw:
Builds failed successfully :D

scripts/package-osx.sh
```diff
@@ -1,4 +1,6 @@
 #!/usr/bin/env bash
+set -e
```
mineo: Is the `curl --upload-file "$dmg" https://transfer.sh/` code still used? If so, then this is not safe because curl returns an exit code of 0 for a lot of HTTP status codes a human might consider errors (401, 403, maybe even 502). It's possible to work around that by using `--fail`, but even that's not guaranteed to result in a non-zero exit code. To make that uploads with curl robust, you need `curl --write-out '%{http_code}' and check whatever that writes.

phw: https://github.com/metabrainz/picard/pull/1085

PR #1086, opened 2019-01-31
PICARD-1460: Fix installer not detecting running instance

phw:
# Summary

With Qt5 the window class changed, hence the installer no longer detects already running instances on upgrade.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1460](https://tickets.metabrainz.org/browse/PICARD-1460)

# Solution

# Action

### DIFF
--- a/installer/picard-setup.nsi.in
+++ b/installer/picard-setup.nsi.in
@@ -172,7 +172,7 @@ Section Uninstall
 SectionEnd
 
 ; Checks whether program is running.
-!define WNDCLASS "QWidget"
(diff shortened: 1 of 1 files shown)


phw:
Just wanted to note that this will then cause the installer not to detect a running instance of Picard 1.4. We could add tests for both new and old WNDCLASS.

@zas @samj1912 Do we need this?

PR #1088, opened 2019-02-01
PICARD-1463: Build with Python 3.7.2 on Windows and macOS

phw:
# Summary
Fixes crashes when loading some locales (e.g. sv_SE) on Windows.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1463](https://tickets.metabrainz.org/browse/PICARD-1463)

# Solution

# Action


phw:
Unfortunately this does not work yet with Appveyor, they only support Python 3.7.1. See https://www.appveyor.com/docs/windows-images-software/#python

I pushed a change, but I'll have to test the resulting build before merging.

phw:
I misunderstood how this works. Actually looking at the table closer it gets clear that you cannot choose between Python 3.7.0 and 3.7.1 on one machine. It depends on the used image: "Visual Studio 2015" has Python 3.7.0 and Visual Studio 2017 has Python 3.7.2.

Actually the PYTHON_VERSION tag we use in the Appveyor build does not seem to have any effect, we should probably drop it. Currently it is misleading, e.g. we claim to use Python 3.5.6 but we actually build with 3.5.4.

I'm not totally sure what consequences building with "Visual Studio 2017" will have in regards to backward compatibility. I'll try a build and see how it runs on my Windows 10 and Windows 7 VMs. But maybe we should instead aim for using Visual Studio 2015 still and install the desired Python version ourselves.

phw:
I think I got this finally working that we update Python to the version we expect. But then during the last hours they updated the appveyor images and we are now of latest 3.6.8 and 3.7.2, see https://github.com/appveyor/website/commit/5cb03b62d3a1a821fe1312c6d731318d5a26f88d#diff-9473c64cdd931c7835e8aca459dcdf16

That also means I'm no longer quite sure if my update routine works or not. But this branch needs some cleanup anyway, I'll think about whether I keep the update routine in or whether we just stick with the versions Appveyor provides us now.

phw:
@zas @samj1912 I think this is good to merge now. Given the Appveyor images have been updated and include the latest Python versions again I made a step back and decided to stay with existing versions. But the difference is that you no longer set any version variables inside the appveyor.yml, since this was horrible misleading. Instead the Python version is read from the actually used Python.

I cleaned up the commit history a bit, but I would like to keep the commits in between that introduced the updating. They got reverted again at the end, but I think it's useful to see how we got there.

PR #1092, opened 2019-02-12
PICARD-323: Fix discid for multidisc releases

phw:
# Summary
On disc lookup set the disc ID only for the tracks on the disc being looked up.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The feature of Picard to set the disc ID used for a disc lookup on the loaded metadata was from a time when releases had only a single medium. It set the disc ID for all tracks, regardless on whether they are on the looked up disc or not.

* JIRA ticket (_optional_): [PICARD-323](https://tickets.metabrainz.org/browse/PICARD-323)

# Solution
Handle multiple disc IDs and set them only for tracks on the looked up disc. The following changes have been made:

1. Album handles a list of disc IDs
2. A new hidden variable `~musicbrainz_discids` is set on the tracks, holding all disc IDs MusicBrainz has stored for the corresponding medium
3. On lookup the tag `musicbrainz_discid` is set only for the affected tracks
4. Looking up a disc for an already loaded album adds the disc ID to the loaded album. That allows the user to also insert e.g. disc 2 and add it's disc ID to the tags after they have loaded the album using disc 1
5. The changes also fix the bug that re-loading an album lost the disc ID.

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -79,7 +79,7 @@ class Album(DataObject, Item):
 
     release_group_loaded = QtCore.pyqtSignal()
 
-    def __init__(self, album_id):
+    def __init__(self, album_id, discid=None):
         DataObject.__init__(self, album_id)
         self.metadata = Metadata()
         self.orig_metadata = Metadata()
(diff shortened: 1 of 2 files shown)


phw:
addressed the codacy warning

PR #1094, opened 2019-02-12
JSON WS changes

mwiencek:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
Changes the way Picard handles certain WS JSON, so that we can later fix the output on the MBS side in the distant future.

# Problem

See the associated commit messages.

# Solution

See the associated commit messages.

# Action

No action needed.

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -427,9 +427,9 @@ def artist_to_metadata(node, m):
                 ended = value['ended']
                 if ended and "end" in value:
                     m["enddate"] = value['end']
-        elif key == "begin_area":
+        elif key == "begin-area":
             m["beginarea"] = value['name']
-        elif key == "end_area":
+        elif key == "end-area":
             m["endarea"] = value['name']
 
 
--- a/picard/album.py
+++ b/picard/album.py
@@ -70,6 +70,17 @@ from picard.ui.item import Item
 register_album_metadata_processor(coverart)
 
 
+def _create_artist_node_dict(source_node):
+    return {x['artist']['id']: x['artist'] for x in source_node['artist-credit']}
+
+
+def _copy_artist_nodes(source, target_node):
+    for credit in target_node['artist-credit']:
+        artist_node = source.get(credit['artist']['id'])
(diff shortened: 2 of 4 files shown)


picard/mbjson.py
```diff
@@ -427,9 +427,9 @@ def artist_to_metadata(node, m):
                 ended = value['ended']
                 if ended and "end" in value:
                     m["enddate"] = value['end']
-        elif key == "begin_area":
+        elif key == "begin_area" or key == "begin-area":
             m["beginarea"] = value['name']
-        elif key == "end_area":
+        elif key == "end_area" or key == "end-area":
```
phw: Do we need to keep `begin_area` and `end_area` here? Actually this is a straight bugfix, because the begin and end area currently don't show up in artist searches in Picard since a [search query](https://musicbrainz.org/ws/2/artist/?query=beach+boys&fmt=json) will already return `begin-area` / `end-area`. With this small change it works again in Picard.

I added https://tickets.metabrainz.org/browse/PICARD-1471 to track this for the changelog.

mwiencek: The search server outputs them correctly, but musicbrainz-server doesn't, e.g. see https://musicbrainz.org/ws/2/artist/ebfc1398-8d96-47e3-82c3-f782abcdb13d?fmt=json

If Picard doesn't actually make any toplevel artist queries, great! We can remove the old properties. I don't believe they would show up in any release lookups.

mwiencek:
The commit to share artist data was inspired by MBS-9400. I don't know if Picard currently uses tags/genres from track/recording artists at all, but this would allow it to access them, since we'll only be outputting them on the release artists once. I'd like us to be able to change the JSON web service in the future to do something similar for aliases, i.e. not duplicate them in track/recording artists when they already appear in the release artist nodes.

picard/mbjson.py
phw: Yes, currently it is only used for searches, so I think this is fine to check only this case.

PR #1095, opened 2019-02-14
PICARD-1242: Consider AcoustId source count

phw:
# Summary
Take the number of sources for a AcoustId into account when selecting the best match. This gives better matches in cases where an AcoustId is also attached to the wrong recording.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

AcoustId tracks the number of sources for each linked MB recording. A higher number usually gives a higher confidence that this AcoustId is linked to the correct release, but Picard does currently not consider this.

* JIRA ticket (_optional_): [PICARD-1242](https://tickets.metabrainz.org/browse/PICARD-1242)

# Solution

Calculate a score for recordings based on the sources relative to the recording with the most sources. E.g. if there are two recordings linked, one with 30 and one with 10 sources, the first one will get a score of 100 (percent), the second one of 33.

This score will be used when comparing recordings for the best match.

As this scoring roughly is the same as the search scoring MB provides for search results, this PR also makes use of this and utilizes MB search scores the same way.

# Action

This should be properly tested as it affects search results as well (hopefully making them a bit better).

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -80,10 +80,12 @@ class AcoustIDClient(QtCore.QObject):
                     if results:
                         result = results[0]
                         file.metadata['acoustid_id'] = result['id']
-                        if 'recordings' in result:
+                        if 'recordings' in result and result['recordings']:
+                            max_sources = max([r['sources'] for r in result['recordings']])
                             for recording in result['recordings']:
                                 parsed_recording = parse_recording(recording)
                                 if parsed_recording is not None:
+                                    parsed_recording['score'] = recording['sources'] / max_sources * 100
                                     recording_list.append(parsed_recording)
                             log.debug("AcoustID: Lookup successful for '%s'", file.filename)
                 else:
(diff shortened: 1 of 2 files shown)


picard/acoustid/__init__.py
```diff
@@ -80,10 +80,12 @@ def _on_lookup_finished(self, next_func, file, document, http, error):
                     if results:
                         result = results[0]
                         file.metadata['acoustid_id'] = result['id']
-                        if 'recordings' in result:
+                        if 'recordings' in result and result['recordings']:
+                            max_sources = max([r['sources'] for r in result['recordings']])
                             for recording in result['recordings']:
                                 parsed_recording = parse_recording(recording)
                                 if parsed_recording is not None:
+                                    parsed_recording['score'] = recording['sources'] / max_sources * 100
```
zas: Just to be sure, `max_sources` is never zero, right?

phw: Good point. Actually it should never be 0, always at least 1. But given that this comes from an external service, and we might get unexpected results, we should probably ensure it is set to 1

zas: Yes, better safe than sorry. We had such issue in the past (division by zero exception because the ws wasn't returning what was expected, or just because it was the normal behavior but not taken in account in the code).

phw: I made the code more error tolerant, also dealing with the case that this field is not specified at all. We can still fail with a `TypeError` here if `score` would contain something that is not an int, but this at least would be catched by the outer exception handling and Picard would not crash.

PR #1096, opened 2019-02-14
PICARD-1447: Generate appdata release list on build

phw:
# Summary
Generates the release list in appdata XML from NEWS.txt

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The appdata release list needs to be updated before every release. This adds another step to the release process that easily gets forgotten.

* JIRA ticket (_optional_): [PICARD-1447](https://tickets.metabrainz.org/browse/PICARD-1447)

# Solution
Generate the release list in appdata XML on build time using the data from NEWS.txt.

# Action

### DIFF
--- a/org.musicbrainz.Picard.appdata.xml.in
+++ b/org.musicbrainz.Picard.appdata.xml.in
@@ -93,16 +93,6 @@
   </content_rating>
 
   <releases>
-    <release date="2018-01-29" version="2.1.2"/>
-    <release date="2018-01-29" version="2.1.1"/>
-    <release date="2018-12-20" version="2.1.0"/>
-    <release date="2018-09-05" version="2.0.4"/>
-    <release date="2018-08-10" version="2.0.3"/>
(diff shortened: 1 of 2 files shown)


phw:
> The only issue i see here is relying on `NEWS.txt`, but i don't see a better way.

The amazing part is that we actually followed a strict format for the versions and date througout the entire history of the file, so I'm not too concerned about it. My first idea was using Git history, but this really causes trouble because you don't always have access to this in all build environments.

> You may want to rebase though

I still really disagree with this workflow, it is annoying really and I actually deliberately ignore it for the most time. The pull requests we have are largely small and not related to one another, rebasing invalidates the reviews so we merge "unreviewed" PRs and the merge requests created by "Update branch" clutter the commit history. Just my 2 cents, but there is little benefit and too much trouble.

For the few cases where changes to the master branch might interfere with the PR we can still request rebases in reviews.

PR #1097, opened 2019-02-14
Updated dependencies for Windows and macOS builds

phw:
# Summary
Updated to latest PyQt 5.12 and mutagen 1.42

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

Update the dependencies to latest Mutagen and PyQt for macOS and Windows. Mainly to stay recent, but also with some hope to fix [PICARD-1472](https://tickets.metabrainz.org/browse/PICARD-1472)

Using both on Linux without issues, also did my latest Windows testing with those versions. But we should test the resulting binaries from Appveyor first on both platforms.

* JIRA ticket (_optional_): [PICARD-1472](https://tickets.metabrainz.org/browse/PICARD-1472)

# Solution

# Action


phw:
@zas I had no luck using newer PyQt version with macOS xcode6.3 and xcode7.4 images, but I now also downgraded the xcode7.4 build to the version we use for our current releases. My hope is this will fix the code signing issues, which are related to some SIP changes. But we currently can only test this on the master branch after merge.

To make good for the loss I included a macOS build that always builds the latest and greatest (also using the newest xcode image Travis provides).

phw:
I merged it, let's see if the builds with signing enabled run through again. If not I will follow up in this.

phw:
I disabled the new 10.1 image build again in https://github.com/metabrainz/picard/commit/3d39875447922d8d6aa55d1b1c99ccb9bcb157a3 . This one hangs during code signing until the build times out.

But on the positive side builds with xcode6.4 and xcode7.3 work again. We can tackle issues with newer versions again if we need to update.

PR #1099, opened 2019-02-15
PICARD-1422: Windows uninstall previous install

phw:
# Summary
On Windows uninstall previous Picard install before installing the new version. Also make sure to never default to installing into the "Program Files (x86)" folder, which is reserved for 32 bit applications.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

1. Picard just overwrote files in the existing install location. This could potentially lead to left over libraries interfering with the current install. We had reports of crashes that were fixed by clearing the old installation location before a fresh install. So this might be a potential fix for PICARD-1343 and PICARD-1469
2. When updating and choosing a new installation location, Picard would end up being installed twice. An uninstall would only remove the latest install
3. When 32 bit Picard was installed previously an update to Picard 2 would default to installing to "Program Files (x86)" folder, which is reserved for 32 bit applications.

* JIRA ticket (_optional_): [PICARD-1422](https://tickets.metabrainz.org/browse/PICARD-1422)
* JIRA ticket (_optional_): [PICARD-1343](https://tickets.metabrainz.org/browse/PICARD-1343)
* JIRA ticket (_optional_): [PICARD-1469](https://tickets.metabrainz.org/browse/PICARD-1469)

# Solution
Always run uninstaller. Also check the stored installation path and if it is "Program Files (x86)" reset it to the default.

# Action

### DIFF
--- a/installer/picard-setup.nsi.in
+++ b/installer/picard-setup.nsi.in
@@ -211,6 +211,13 @@ Function .onInit
   ClearErrors
   Exec $R0
 
+  ; Check if previous install location was inside $PROGRAMFILES32
+  ReadRegStr $R0 HKLM "Software\MusicBrainz\${PRODUCT_NAME}" "InstallDir"
+  StrLen $R1 $PROGRAMFILES32
+  StrCpy $R2 $R0 $R1
+  StrCmp $R2 $PROGRAMFILES32 0 skipUninstall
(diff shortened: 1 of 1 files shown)


phw:
I'm not 100% happy with some of the strings used currently. But I'm working on localization of the installer anyway and will update the strings in that PR.

PR #1101, opened 2019-02-17
PICARD-1400: Remove Amazon cover art provider

phw:
# Summary
Remove the Amazon cover art provider from core Picard and provide it as a plugin.

This one might be controversial, so let's discuss :)

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The Amazon cover art provider does not really fit in with the rest of Picard core. With CAA we now have an excellent cover art service, and usually Picard focuses on core functionality provided by MB and adds external services as plugins.

The Amazon plugin also has some general issues, as it cannot load all Amazon cover art the MB server finds (see [PICARD-959](https://tickets.metabrainz.org/browse/PICARD-959)).

* JIRA ticket (_optional_): [PICARD-1400](https://tickets.metabrainz.org/browse/PICARD-1400)

# Solution
Remove the cover art provider from core and add it as a plugin instead.

# Action


phw:
Plugin at https://github.com/metabrainz/picard-plugins/pull/207

phw:
This change would probably mean we should call the next version 2.2 :)

zas:
> This change would probably mean we should call the next version 2.2 :)

I think we should perhaps release a 2.1.x version first, including recent fixes/improvements you merged.

phw:
Yes, that's also fine with me. In this case let's delay merging this and the corresponding plugin.

zas:
@phw: bump plugin API version in this PR ?

phw:
I rebased and bumped the API version

PR #1102, opened 2019-02-17
PICARD-1476: Introduce a priority thread pool

phw:
# Summary
Make sure the metadata boxes get updated quickly when the selection changes even if the thread pool is busy.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

When Picard's default thread pool is kept busy this can prevent update of the metadata box on selection changes. Reason is that calculating the updated tags also runs on the thread pool.

This e.g. happens with the BPM plugin, see https://github.com/metabrainz/picard-plugins/issues/200 and discussion  on https://github.com/metabrainz/picard-plugins/issues/200

* JIRA ticket (_optional_): [PICARD-1476](https://tickets.metabrainz.org/browse/PICARD-1476)

# Solution
Introduce a priority thread pool

This pool is meant to be used for threads where the user expects a response as fast as possible.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -160,6 +160,12 @@ class Tagger(QtWidgets.QApplication):
         # It's a valid reference, but its start() method doesn't work.
         self.thread_pool = QtCore.QThreadPool(self)
 
+        # Provide a separate thread pool for operations that should not be
+        # delayed by longer background processing tasks, e.g. because the user
+        # expects instant feedback instead of waiting for a long list of
+        # operations to finish.
+        self.priority_thread_pool = QtCore.QThreadPool(self)
+
         # Use a separate thread pool for file saving, with a thread count of 1,
         # to avoid race conditions in File._save_and_rename.
         self.save_thread_pool = QtCore.QThreadPool(self)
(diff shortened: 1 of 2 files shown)


picard/tagger.py
```diff
@@ -341,6 +347,7 @@ def exit(self):
         self._acoustid.done()
         self.thread_pool.waitForDone()
         self.save_thread_pool.waitForDone()
+        self.priority_thread_pool.waitForDone()
```
zas: Since we have more thread pools now, i'd rather use a dict to store them:

```python
self.threads = ('main', 'priority', 'save')
self.thread_pools = dict()
for thread in self.threads:
    self.thread_pools[thread] = QtCore.QThreadPool(self)
...

for thread in self.threads:
    self.thread_pools[thread].waitForDone()
```

And perhaps change `run_task()` in https://github.com/metabrainz/picard/blob/master/picard/util/thread.py to use the thread name instead.

Of course we could use a class instead of a dict to make it cleaner.

phw: I am actually not too sure I like the dict approach better than the three attributes we have right now. Also we should also probably take care that this does not run out of control and we end up managing a pool of pools :)

PR #1085, opened 2019-01-29
PICARD-1382: Fail macOS build on curl errors

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Minor / simple change (like a typo)
    * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem
Address possibility of curl failing during macOS packaging, see https://github.com/metabrainz/picard/pull/1083#pullrequestreview-197696069

* JIRA ticket (_optional_): [PICARD-1382](https://tickets.metabrainz.org/browse/PICARD-1382)

# Solution

# Action

PR #1103, opened 2019-02-18
CI: Do not fail the build if macOS upload to transfer.sh fails

phw:
# Summary
Do not fail the build if macOS upload to transfer.sh fails.

As discussed with @zas on IRC we do not really care in most cases if this fails, but it currently quite often fails with server errors from transfer.sh. We still want this functionality though, as sometimes you need to access the build binaries for pull requests. If it failed and we find out we need it restarting the build is usually a quick solution.

This basically reverts https://github.com/metabrainz/picard/pull/1085

The general approach that errors in package-osx.sh should fail the build still is true

### DIFF
--- a/scripts/package-osx.sh
+++ b/scripts/package-osx.sh
@@ -45,7 +45,7 @@ hdiutil create -volname "MusicBrainz Picard $VERSION" -srcfolder 'MusicBrainz Pi
 [ "$codesign" = '1' ] && codesign --keychain picard.keychain --verify --verbose --sign 'Developer ID Application: MetaBrainz Foundation Inc.' "$dmg"
 if [ -n "$UPLOAD_OSX" ]
 then
-    curl -v --fail --upload-file "$dmg" https://transfer.sh/
+    curl -v --upload-file "$dmg" https://transfer.sh/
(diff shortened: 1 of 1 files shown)


scripts/package-osx.sh
```diff
@@ -45,7 +45,7 @@ hdiutil create -volname "MusicBrainz Picard $VERSION" -srcfolder 'MusicBrainz Pi
 [ "$codesign" = '1' ] && codesign --keychain picard.keychain --verify --verbose --sign 'Developer ID Application: MetaBrainz Foundation Inc.' "$dmg"
 if [ -n "$UPLOAD_OSX" ]
 then
-    curl -v --fail --upload-file "$dmg" https://transfer.sh/
+    curl -v --upload-file "$dmg" https://transfer.sh/
```
zas: What about making curl command more resistant to temporary issues using `--retry`, `--retry-delay`, `--retry-max-time` options ?

phw: Good idea, I added `--retry 6 --retry-delay 10`. What do you think about the values? I want to give this enough time so the receiving server eventually recovers, but still don't want to delay the test run for too long.

PR #1105, opened 2019-02-18
PICARD-1474: Fix version information in Windows installer exe

phw:
# Summary
Show proper version string in file properties of Windows install .exe file.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Looking at the file properties of the Windows installer show outdated version string, since that was hard coded.

* JIRA ticket (_optional_): [PICARD-1474](https://tickets.metabrainz.org/browse/PICARD-1474)

# Solution
Set the version accordingly.

Before and after:

grafik

Some notes:

1. `VIProductVersion` actually is shown as "File version" ("Dateiversion" in the screenshot) in the properties dialog. This must be defined.
2. `VIProductVersion` must be in the format `0.0.0.0`, only digits allowed
3. You need to specify both `FileVersion` and `ProductVersion` in the version metadata. These are strings. `ProductVersion` is shown in the dialog as "Product version", I don't know how `FileVersion` is used, but it needs to be present and is not shown as `File Version` in the dialog
4. There is some kind of encoding issue in the Copyright field. We can fix that by putting that string into a separate file and enabling Unicode support.. I am working on a translation of the installer, which will take care of this (but don't expect this too soon).

# Action

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -57,7 +57,7 @@ build_script:
 
     pyinstaller --noconfirm picard.spec
 
-    makensis.exe installer\picard-setup.nsi
+    makensis.exe /INPUTCHARSET UTF8 installer\picard-setup.nsi
 
 artifacts:
 - path: installer/*.exe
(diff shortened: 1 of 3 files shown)


phw:
I just realized we can actually fix the encoding issue, it's just a parameter needed on the makensis call

phw:
Appveyor builds look good now

PR #1106, opened 2019-02-20
PICARD-1288: Fix artist genre fallback

phw:
# Summary
Fix artist genre fallback if "Fall back on album's artists genres" is active

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The artists tags where not read and hence could not be used as a fallback if no tags are set on recording, release or release-group level

* JIRA ticket (_optional_): [PICARD-1288](https://tickets.metabrainz.org/browse/PICARD-1288)

# Solution
Together with [MBS-9400](https://tickets.metabrainz.org/browse/MBS-9400) this fixes the issue.

# Action

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -367,7 +367,7 @@ def recording_to_metadata(node, m, track=None):
             if track:
                 for artist in value:
                     artist_obj = track.append_track_artist(artist['artist']['id'])
-                    add_genres_from_artist_credits(artist, artist_obj)
+                    add_genres_from_node(artist['artist'], artist_obj)
         elif key == 'relations':
             _relations_to_metadata(value, m)
         elif key in ('genres', 'tags') and track:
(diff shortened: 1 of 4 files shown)


picard/mbjson.py
```diff
@@ -366,7 +366,8 @@ def recording_to_metadata(node, m, track=None):
             # set tags from artists
             if track:
                 for artist in value:
-                    track.append_track_artist(artist['artist']['id'])
+                    artist_obj = track.append_track_artist(artist['artist']['id'])
+                    add_genres_from_node(artist['artist'], artist_obj)
```
zas: I think it'd be more readable to use another intermediate variable, something like:

```python
                for artist in value:
                    art = artist['artist']
                    artist_obj = track.append_track_artist(art['id'])
                    add_genres_from_node(art, artist_obj)
```

picard/mbjson.py
```diff
@@ -448,7 +449,8 @@ def release_to_metadata(node, m, album=None):
             # set tags from artists
             if album is not None:
                 for artist in value:
-                    album.append_album_artist(artist['artist']['id'])
```
zas: Same here

zas:
Tests are failing though

picard/mbjson.py
phw: Done, also added unit test

phw:
> Tests are failing though

Yes, wrong indentation of one line. Also fixed

PR #1108, opened 2019-02-20
PICARD-729: NAT error handling

phw:
# Summary
Fix handling of non-album tracks in error case and make it generally more usable.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

If loading of a NAT fails it is stuck with the "[loading track information]" message. It does never display any error. Also it is not possible to remove any NAT, you can only remove the entire "non-album tracks" pseudo album.

* JIRA ticket (_optional_): [PICARD-729](https://tickets.metabrainz.org/browse/PICARD-729)

# Solution

The core changes to fix PICARD-729 are:
- Show an actual error state with error icon, error details in tooltip and message "could not load recording"
- Allow removing single NATs

I also did some general cleanup that improve the usability of this feature:
- The NATs pseudo album is loaded always expanded, otherwise newly added NATs are not visible to the user
- Change the message "[loading track information]" to "[loading recording information]"
- Made the "[loading recording information]" localizable

The following animation demonstrates the changes:
peek 2019-02-20 21-47

# Action
Do we need a ticket for the auto expanded NATs and the localized error message? I considered these changes rather minor

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -329,6 +329,7 @@ class Tagger(QtWidgets.QApplication):
             self.nats = NatAlbum()
             self.albums["NATS"] = self.nats
             self.album_added.emit(self.nats)
+            self.nats.item.setExpanded(True)
         return self.nats
 
     def move_file_to_nat(self, file, recordingid, node=None):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -664,6 +664,16 @@ class Tagger(QtWidgets.QApplication):
             self.nats = None
         self.album_removed.emit(album)
 
+    def remove_nat(self, track):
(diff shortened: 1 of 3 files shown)


phw:
One thing to note here is that this includes string changes. On the other hand one of those strings already is visible to user, but not translatable.

zas:
> One thing to note here is that this includes string changes. On the other hand one of those strings already is visible to user, but not translatable.

That's a non-issue, since those were not translated anyway before.

Though there is at least one another string that changed (Check for updates) recently, so i think it'd be a good idea to ask for translations before the next release.

PR #1109, opened 2019-02-21
PICARD-1178: Support comparing cover images with only one type

phw:
# Summary
Fixes issues where files where shown as modified when loading again and comparing against cover art with more than one type.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Cover art images that have both "front" and another type will cause files been shown as modified if they are saved and reloaded. Reason ist that only the front type gets stored, but comparison with the original image with more types than fails.

* JIRA ticket (_optional_): [PICARD-1178](https://tickets.metabrainz.org/browse/PICARD-1178)

# Solution

The following cases can happen:

- Format does not support cover art (this is currently only the case for formats that don't support tagging at all, which means currently WAV and MIDI)
- Format supports cover art, but no types (MP4, APE)
- Format supports cover art with exactly one type (ID3, ASF/WMA, FLAC, Vorbis)
- Format supports cover art with multiple types (currently none)

Case 1. is a non-issue, case 2. is already supported, since we indicate whether loaded cover art supports types or not. Case 3. is currently broken and case 4. is hypothetical.

Implemented a new attribute `CoverArtImage.support_multi_types` which indicates whether an image supports more than one type. Off by default (except for CAA). Only compare all types if both compared images have `support_multi_types == True`, else compare only the `maintype`.

This way we both fix the issue and are prepared for the multi-type case (which currently would only apply when comparing images from cover art sources that support multiple types).

Added tests for `CoverArtImage`, extended tests for formats to consider this case and tested manually with all supported formats.

# Action

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -121,6 +121,8 @@ class CoverArtImage:
     # formats may have types associated with cover art, but some other sources
     # don't provide such information
     support_types = False
+    # Indicates that the source supports multiple types per image.
+    support_multi_types = False
     # `is_front` has to be explicitly set, it is used to handle CAA is_front
     # indicator
     is_front = None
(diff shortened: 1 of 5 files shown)


phw:
Codacy complains about a code duplication in `CoverArtImageFromFile`. This is not really something new this PR introduces, but the PR touches the code so codacy complains.

I still think we should refactor `CoverArtImageFromFile` a bit and have prepared code to do so, but will submit this in a separate PR.

PR #1115, opened 2019-03-02
File cleanup

zas:
- replace list + sort with a heap, faster
- rename  tagger._acoustid to tagger.acoustidclient (it's a AcoustidClient object)
- do not use File._lookup_finished() outside file.py
- use class constants (integers) instead of strings for File._lookup_finished() lookuptype parameter

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -597,10 +597,7 @@ class File(QtCore.QObject, Item):
             log.error("Network error encountered during the lookup for %s. Error code: %s",
                       self.filename, error)
         try:
-            if lookuptype == "metadata":
-                tracks = document['recordings']
-            elif lookuptype == "acoustid":
-                tracks = document['recordings']
+            tracks = document['recordings']
         except (KeyError, TypeError):
             tracks = None
 
--- a/picard/file.py
+++ b/picard/file.py
@@ -65,6 +65,9 @@ class File(QtCore.QObject, Item):
     ERROR = 3
     REMOVED = 4
 
+    LOOKUP_METADATA = 1
+    LOOKUP_ACOUSTID = 2
+
     comparison_weights = {
         "title": 13,
         "artist": 4,
@@ -619,7 +622,7 @@ class File(QtCore.QObject, Item):
                 best_match = match
                 best_sim = best_match[0]
         sim, rg, release, track = best_match
-        if lookuptype != 'acoustid' and sim < config.setting['file_lookup_threshold']:
+        if lookuptype != File.LOOKUP_ACOUSTID and sim < config.setting['file_lookup_threshold']:
             self.tagger.window.set_statusbar_message(
                 N_("No matching tracks above the threshold for file '%(filename)s'"),
                 {'filename': self.filename},
@@ -636,7 +639,7 @@ class File(QtCore.QObject, Item):
 
         self.clear_pending()
 
-        if lookuptype == 'acoustid':
+        if lookuptype == File.LOOKUP_ACOUSTID:
             self.tagger.acoustidmanager.add(self, track['id'])
         if release:
             self.tagger.get_release_group_by_id(rg['id']).loaded_albums.add(release['id'])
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -464,6 +467,12 @@ def _acoustid_update(self):
             recording_id = self.metadata['musicbrainz_recordingid']
         self.tagger.acoustidmanager.update(self, recording_id)
 
+    def analyze(self, use_acoustid):
+        self.set_pending()
+        if use_acoustid:
+            callback = partial(self._lookup_finished, File.LOOKUP_ACOUSTID)
+            self.tagger.acoustidclient.analyze(self, callback)
```
phw: While in theory I liked the idea of making this a method of `File`, I don't like the use of `self.tagger.acoustidmanager` here. IMHO we should reduce the access of properties of the tagger object. In this sense this code makes more sense inside tagger itself.

Side question: What happens to set_pending if `use_acoustid == False`? Does the pending state get reset?

picard/tagger.py
```diff
@@ -745,10 +745,9 @@ def use_acoustid(self):
     def analyze(self, objs):
         """Analyze the file(s)."""
         files = self.get_files_from_objects(objs)
+        use_acoustid = self.use_acoustid
```
phw: This looks to me a bit like over-optimization, could just as well use `self.use_acoustid` at the one line it is used in the loop.

picard/file.py
zas: While i agree about the general idea to reduce the access of properties of the tagger object, that quite unrealistic with the current code, here are the number of occurences of `self.tagger.` among source files in master:

```
     36 picard/ui/mainwindow.py
     23 picard/file.py
     18 picard/album.py
     12 picard/ui/itemviews.py
     12 picard/cluster.py
     11 picard/ui/searchdialog/track.py
      9 picard/ui/searchdialog/album.py
      7 picard/acoustid/__init__.py
      5 picard/ui/ui_options_script.py
      5 picard/ui/options/general.py
      5 picard/browser/browser.py
      5 picard/acoustid/manager.py
      4 picard/collection.py
      3 picard/ui/options/network.py
      3 picard/ui/coverartbox.py
      3 picard/disc.py
      2 picard/ui/searchdialog/artist.py
      2 picard/ui/scriptsmenu.py
      2 picard/ui/options/tags.py
      2 picard/ui/options/renaming.py
      2 picard/ui/options/plugins.py
      2 picard/ui/options/metadata.py
      2 picard/ui/cdlookup.py
      2 picard/track.py
      2 picard/releasegroup.py
      2 picard/plugin.py
      1 test/picardtestcase.py
      1 picard/webservice/__init__.py
      1 picard/util/checkupdate.py
      1 picard/ui/util.py
      1 picard/ui/searchdialog/__init__.py
      1 picard/ui/ratingwidget.py
      1 picard/ui/options/releases.py
      1 picard/ui/options/interface.py
      1 picard/ui/metadatabox.py
```

Imho File class needs access to acoustid stuff (after all, acoustid applies to files).
But i agree this move isn't the best, we need to improve that somehow.

Do you have ideas about how we could improve the situation ??

> Side question: What happens to set_pending if `use_acoustid == False`? Does the pending state get reset?

Good question, it doesn't seem so, but the code is damn messy.

picard/tagger.py
zas: Last time i measured performance regarding use of `config` i was surprised how slow it was, so i tend to avoid it in loops etc... i could spend time to measure this again though.

But i disagree on "over-optimization" here, the fact is this code is very "under-optimized" by design, doing an invariant test in a loop and hiding the complexity behind fake properties.

That's one drawback of Python "features": `self.use_acoustid` is not a variable but a method doing a test calling `config` which is itself a bunch of methods with a disk backend... 

Though i agree resulting code is "weird", i'd rather write it:
`use_acoustid = self.use_tons_of_stuff_you_d_better_cache()` to start with...

picard/file.py
zas: > Side question: What happens to set_pending if `use_acoustid == False`? Does the pending state get reset?

Imho the original code is buggy, what's the point of calling `set_pending()` if nothing is called behind ?

picard/tagger.py
zas: I removed changes concerning acoustid stuff from this PR

phw: Ok, but you made valid points here. Would be an argument making this a function instead of a property as you suggest.

picard/file.py
zas: The pending files issue was addressed by c948e014004806d9d7a24c1a519d734aaef7a9b1

PR #1116, opened 2019-03-02
Rework File._move_additional_files()

zas:
Rework File._move_additional_files()

- add tests for hidden files and patterns starting with dot trick
- use faster os.scandir() (since python 3.5)
- change the logic so we don't build list of all files but only list of files to move
- handle OSError, a file or directory could be removed during the process, or have permissions changed
- log such errors

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -400,35 +400,47 @@ class File(QtCore.QObject, Item):
             image.save(dirname, metadata, counters)
 
     def _move_additional_files(self, old_filename, new_filename):
-        """Move extra files, like playlists..."""
-        old_path = os.path.dirname(old_filename)
-        new_path = os.path.dirname(new_filename)
-        try:
-            names = set(os.listdir(old_path))
-        except os.error:
-            log.error("Error: {} directory not found".naming_format(old_path))
-            return
-        filtered_names = {name for name in names if name[0] != "."}
-        for pattern in config.setting["move_additional_files_pattern"].split():
+        """Move extra files, like images, playlists..."""
+        patterns = config.setting["move_additional_files_pattern"]
+        pattern_regexes = set()
+        for pattern in patterns.split():
             pattern = pattern.strip()
             if not pattern:
                 continue
             pattern_regex = re.compile(fnmatch.translate(pattern), re.IGNORECASE)
-            file_names = names
-            if pattern[0] != '.':
-                file_names = filtered_names
-            for old_file in set(file_names):
-                if pattern_regex.match(old_file):
-                    names.discard(old_file)
-                    filtered_names.discard(old_file)
-                    new_file = os.path.join(new_path, old_file)
-                    old_file = os.path.join(old_path, old_file)
-                    # FIXME we shouldn't do this from a thread!
-                    if self.tagger.files.get(decode_filename(old_file)):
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -400,35 +400,47 @@ def _save_images(self, dirname, metadata):
...
-                        log.debug("File loaded in the tagger, not moving %r", old_file)
+            match_hidden = pattern.startswith('.')
+            pattern_regexes.add((pattern_regex, match_hidden))
+        if not pattern_regexes:
+            return
+        new_path = os.path.dirname(new_filename)
+        old_path = os.path.dirname(old_filename)
+        moves = set()
+        try:
+            # TODO: use with statement with python 3.6+
```
phw: We could look whether we want to increase the minimum version. I think it depends a bit which distros we want to support. But e.g. the oldest Ubuntu we currently can build for is Ubuntu 18.04 Bionic, and that has Python 3.6 (in theory we could also build on 17.10, but that is no longer supported).

zas:
> Much clearer logic.

Yes, it's much better but in fact the whole logic is bad from the start (and i don't know how to "fix" it):
- let's say we have 10k audio files and 5 images in one directory 
- Picard processes one audio file, and move additional files (the directory now contains 10k - 6 files)
- it processes next audio file and look for additional files in 10k - 6 files, applying multiples regex matches etc... for nothing

This move improves things a lot using `os.scandir()` to iterate and, instead of building a long set of files, it just builds a set of files to move, BUT it still has to iterate all files in directory more than once, because each "File" is processed separately.

That's a general issue with Picard, bulk processing is more like a bulk of individual file processing, so it's often very inefficient. I have no quick solution for that, this move is still a good one imho.

It should be noted that if the number of files in one directory is reasonable, the issue might not be obvious, the original behavior was very likely to impact a lot users:
- having a lot of files in one directory
- having more than usual patterns in move additional files options
- having move additional files option enabled
- running windows (listdir is much slower on windows than other platforms)
- and, even worse, having files on a network filesystem

I think it matches some slowness reports we had, but i didn't measure the actual gain on windows (as it don't run this system at all). On linux, it is already much faster, but i just did quick tests, no real measurements yet.

picard/file.py
```diff
@@ -400,35 +400,47 @@ def _save_images(self, dirname, metadata):
...
+                    if pattern_regex.match(entry.name):
+                        new_file_path = os.path.join(new_path, entry.name)
+                        moves.add((entry.path, new_file_path))
+                        break  # we are done with this file
+        except OSError as why:
+            log.error("Failed to scan %r: %s", old_path, why)
+            return
+
+        for old_file_path, new_file_path in moves:
+            # FIXME we shouldn't do this from a thread!
```
zas: @phw : any idea on how to "fix" this FIXME ?

phw: I also asked this myself when looking over this PR. If I understand this correctly it's about `self.tagger.files.get(decode_filename(old_file_path))` not being thread-safe. If that's the issue we probably have to make access to the file list thread safe by providing access methods that ensure only a single thread can access the loaded file list.

PR #1118, opened 2019-03-04
Mbid validate fix

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

`mbid_validate()` was validating a string starting with a valid mbid but too long as valid.
The reason is the use of `re.match()` and no `$` in the regex, so it was matching only start of string, not full string.
Also the documentation explicitely says it should return None if no match, that's not a problem here, but better test against None.
Tests were lacking.

This issue was very unlikely to be a real life one, since the buggy case shouldn't happen with current code, but still... 

# Solution

Add tests and refactor / fix

# Action

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -243,3 +243,19 @@ class CompareBarcodesTest(unittest.TestCase):
         self.assertFalse(util.compare_barcodes('727361379704', '1727361379704'))
         self.assertFalse(util.compare_barcodes('0727361379704', None))
         self.assertFalse(util.compare_barcodes(None, '0727361379704'))
+
+
+class MbidValidateTest(unittest.TestCase):
+
+    def test_ok(self):
+        self.assertTrue(util.mbid_validate('2944824d-4c26-476f-a981-be849081942f'))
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -243,3 +243,19 @@ def test_not_same(self):
         self.assertFalse(util.compare_barcodes('727361379704', '1727361379704'))
         self.assertFalse(util.compare_barcodes('0727361379704', None))
         self.assertFalse(util.compare_barcodes(None, '0727361379704'))
+
+
+class MbidValidateTest(unittest.TestCase):
+
+    def test_ok(self):
+        self.assertEqual(util.mbid_validate('2944824d-4c26-476f-a981-be849081942f'), True)
```
phw: You can directly use `assertTrue` / `assertFalse` here

zas: Done in  5af33683921cbf5e12ca795a01f893cf26414119 (rebased)

PR #1122, opened 2019-03-05
Introduce scantree() and tests for it

zas:


zas:
@phw tests with hidden files are passing on windows :dancer: 
I'll add more tests (symlinks and weird cases) soon

PR #1126, opened 2019-03-09
Better handling of missing libdiscid, catching OSError

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Picard cannot run without this patch and missing `libdiscid` and `python-libdiscid`.

It falls back on `discid` module, but when this one is trying to load `libdiscid` it fails with:
`OSError: libdiscid.so.0: cannot open shared object file: No such file or directory`

# Solution

According to https://python-discid.readthedocs.io/en/latest/install/#dependencies

*The module discid cannot be imported without Libdiscid >= 0.2.2 installed.
If you want to use it as optional dependency, import the module only when
needed or catch the OSError when libdiscid is not found.*

So catch OSError and define discid to None.
After that, Picard runs but emits a warning:
`W: 10:54:37,425 picard/ui/mainwindow.create_actions:465: CDROM: discid library not found - Lookup CD functionality disabled`

# Action

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -4,28 +4,27 @@ MusicBrainz Picard Installation
 Dependencies
 ------------
 
-Before installing Picard, you need to have these:
+Before installing Picard, you need to check you have following dependencies installed.
+
+Required:
 
 * [Python 3.5 or newer](http://python.org/download)
-
 * [PyQt 5.7.1 or newer](http://www.riverbankcomputing.co.uk/software/pyqt/download)
-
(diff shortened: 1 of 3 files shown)


zas:
> Have you checked how the Python `libdiscid` behaves in this case? As this is directly compiled against libdiscid it probably either crashes or just causes `ImportErrorr`, I hope the latter.

It raises ImportError which is handled (it falls back on discid in this case).

PR #1127, opened 2019-03-09
_move_additional_files(): return asap if same directory

zas:
If old file and new file are in the same directory, there's no need to move any additionnal files

Apparently, the code wasn't really taking care of this,
since its introduction in https://github.com/metabrainz/picard/commit/6d4dc078aeff5a57f6fb9eaf568e96fb9a2eda20

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -407,6 +407,11 @@ class File(QtCore.QObject, Item):
 
     def _move_additional_files(self, old_filename, new_filename):
         """Move extra files, like images, playlists..."""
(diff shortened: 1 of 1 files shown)


zas:
> For the user this is not really visible, right?

Nope, and it's just an optimization (i think it was silently failing anyway, copying files over themselves).

PR #1134, opened 2019-03-14
Throttle File.update()

zas:
Since File.update() also updates UI, and sometimes without any change, throttle it

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -26,6 +26,7 @@ import os
 import os.path
 import re
 import shutil
+from time import time
 import unicodedata
 
 from PyQt5 import QtCore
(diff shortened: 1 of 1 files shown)


zas:
@phw can you test this on windows with a lot of files ? This patch is trying to reduce the number of updates (especially when moving File between clusters).

picard/file.py
```diff
@@ -541,6 +546,21 @@ def is_saved(self):
         return self.similarity == 1.0 and self.state == File.NORMAL
 
     def update(self, signal=True):
+        now = time()
```
phw: Couldn't we use the `@throttle(250)` decorator from `picard.util` for this?

zas: Nope, i tried, but properties are changing, so it doesn't work (i don't remember exactly what was the issue, but i ended with this solution, because it didn't work).
If you can, do a test with a lot of files on windows, i'm not even sure it is worth.

PR #1135, opened 2019-03-15
Cleanup File state, tracknumber, discnumber properties

zas:
- use `@property` and `@property.setter` decorators
- make those more obvious
- drop force_update / update parameters as they were hiding `File.update()` calls
- remove redundant `File.update()` calls
- state property: only emit `tagger_stats_changed` signal when `num_pending_files` actually changed
- simplify tests in `File.update()`, introduced in https://github.com/metabrainz/picard/commit/62c705623666175f81eb18248190d08794675780

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -559,8 +559,7 @@ class File(QtCore.QObject, Item):
                     self.state = File.NORMAL
         if signal:
             log.debug("Updating file %r", self)
-            if self.item:
-                self.item.update()
+            self.update_item()
 
     def can_save(self):
         """Return if this object can be saved."""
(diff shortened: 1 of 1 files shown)


picard/file.py
```diff
@@ -685,28 +688,28 @@ def clear_lookup_task(self):
     def set_pending(self):
         if self.state != File.REMOVED:
             self.state = File.PENDING
-            self.update()
+            self.update()  # FIXME: only call self.item.update() ??
```
phw: Yes, likely only `self.item.update()` is necessary for pending state changes. I'm pretty sure this update call is only abot updating the icon. I don't see what other change should happen that requires a call to `update()` that not already handles this itself.

zas: Done in https://github.com/metabrainz/picard/pull/1135/commits/1accac01df4dd7988b11014c1a7edb31e479a929

PR #1137, opened 2019-03-17
Metadata class: inherit from MutableMapping instead of dict

zas:
Inheriting from `dict` is rather messy, so let's inherit from `collections.abc.MutableMapping` instead which is meant for that.

The conversion isn't that easy, and some tests were dependent on `dict` methods.

But the move has a lot of advantages:
- much cleaner
- `del metadata['tag']` is working
- `Metadata([('tag', 'value')])` can be used to create the object from an iterable
- `Metadata([('tag', 'value'), ('ignoreme': '')]` or `Metadata(tag='value', ignoreme='')` are equivalent to:
```python
m = Metadata()
m['tag'] = 'value'
```
- `metadata.keys()` and `metadata.values()` are working as expected
- `len(metadata)` is returning number of tags plus number of images (as ` __bool__` somehow was doing)
- a method `getraw()` was added, because none of `__getitem__()`, `get()`, `getall()` were actually raising a `KeyError` (as they all have default value), which forced to use `dict.get()` method in some tests.
- `metadata.update()` accepts dict-like syntax:
```python
m = Metadata(tag1='a', tag2='b')
m.update(tag1='c')
m.update(tag2='')
```
will lead to:
```
store: {'tag1': ['c']}
deleted: {'tag2'}
```
It will prolly need more work and testing, but it could help to simplify a lot of code.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -284,8 +284,10 @@ class Metadata(MutableMapping):
         self.clear()
         self.update(other)
 
-    def update(self, other):
-        if isinstance(other, self.__class__):
+    def update(self, *args, **kwargs):
+        if len(args) == 1 and isinstance(args[0], self.__class__):
+            # update from Metadata object
+            other = args[0]
             for k, v in other._store.items():
                 self._store[k] = v[:]
             if other.images:
@@ -296,8 +298,13 @@ class Metadata(MutableMapping):
             # Remove deleted tags from UI on save
             for tag in other.deleted_tags:
                 del self[tag]
-        elif isinstance(other, dict):
-            for k, v in other.items():
+        elif len(args) == 1 and isinstance(args[0], MutableMapping):
+            # update from MutableMapping (ie. dict)
+            for k, v in args[0].items():
+                self[k] = v
+        else:
+            # update from a dict-like constructor parameters
+            for k, v in dict(*args, **kwargs).items():
                 self[k] = v
 
     def clear(self):
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -305,10 +305,13 @@ class Metadata(MutableMapping):
             # update from MutableMapping (ie. dict)
             for k, v in args[0].items():
                 self[k] = v
-        else:
+        elif args or kwargs:
             # update from a dict-like constructor parameters
             for k, v in dict(*args, **kwargs).items():
                 self[k] = v
+        else:
+            # no argument, raise TypeError to mimic dict.update()
+            raise TypeError("descriptor 'update' of '%s' object needs an argument" % self.__class__.__name__)
 
     def clear(self):
         self._store.clear()
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -285,20 +285,23 @@ class Metadata(MutableMapping):
         self.update(other)
 
     def update(self, *args, **kwargs):
-        if len(args) == 1 and isinstance(args[0], self.__class__):
+        one_arg = len(args) == 1
+        if one_arg and isinstance(args[0], self.__class__):
             # update from Metadata object
             other = args[0]
-            for k, v in other._store.items():
-                self._store[k] = v[:]
+
+            for k, v in other.rawitems():
+                self.set(k, v[:])
+
+            for tag in other.deleted_tags:
+                del self[tag]
+
             if other.images:
                 self.images = other.images[:]
             if other.length:
                 self.length = other.length
 
-            # Remove deleted tags from UI on save
-            for tag in other.deleted_tags:
-                del self[tag]
-        elif len(args) == 1 and isinstance(args[0], MutableMapping):
+        elif one_arg and isinstance(args[0], MutableMapping):
             # update from MutableMapping (ie. dict)
             for k, v in args[0].items():
                 self[k] = v
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -17,6 +17,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301,
 # USA.
+from collections.abc import MutableMapping
+
 from PyQt5.QtCore import QObject
 
 from picard import config
@@ -45,7 +47,7 @@ MULTI_VALUED_JOINER = '; '
 LENGTH_SCORE_THRES_MS = 30000
 
 
-class Metadata(dict):
+class Metadata(MutableMapping):
 
     """List of metadata items with dict-like access."""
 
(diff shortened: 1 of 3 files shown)


zas:
Note: Codacy is finding too much repeated code, i'll fix it soon.
Travis doesn't seem to build draft PRs, i didn't find why.

test/test_metadata.py
```diff
@@ -299,3 +299,16 @@ def test_metadata_mapping_update_kw(self):
...
         self.assertIn('tag2', m.deleted_tags)
+
+    def test_metadata_mapping_update_kw_del(self):
+        m = Metadata(tag1='a', tag2='b')
+        del m['tag1']
+
+        m2 = Metadata(tag1='c', tag2='d')
+        del m2['tag2']
+
+        m.update(m2)
```
zas: In fact, i'm not sure what we expect here, the patch fixes the fact a tag was marked deleted but still in store, but the result isn't consistent with dict:

dict:

```
>>> m = dict(tag1='a', tag2='b')
>>> del m['tag1']
>>> m2 = dict(tag1='c', tag2='d')
>>> del m2['tag2']
>>> m.update(m2)
>>> print(m)
{'tag2': 'b', 'tag1': 'c'}
```

Metadata:
```
>>> m = Metadata(tag1='a', tag2='b')
>>> del m['tag1']
>>> m2 = Metadata(tag1='c', tag2='d')
>>> del m2['tag2']
>>> m.update(m2)
>>> print(m)
store: {'tag1': ['c']}
deleted: {'tag2'}
images: []
length: 0
```


Here, since `m2` `tag2` was deleted by `del m2['tag2']`, we end with only `tag1` in result: `{'tag1': 'c'}`
I think it's the expected behavior, and all tests pass (including this one), but it would be easier if we actually document `Metadata` expected results & operations, especially when it comes to `deleted_tags`.
BTW, i think this case is buggy in current master.

test/test_metadata.py
```diff
@@ -187,3 +187,146 @@ def test_compare_deleted(self):
...
+        self.assertEqual(m1.images[0], image1)
+
+        m1.append_image(image2)
+        self.assertEqual(m1.images[1], image2)
+
+        m1.remove_image(0)
+        self.assertEqual(m1.images[0], image2)
+
+        m2 = Metadata(a='c', length=4567, images=[image1])
+        m1.update(m2)
```
zas: In this case, current (and previous code) removes `image2` since it isn't in `m2` and `Metadata.update()` just copies the list of images of `m2` to `m1`.

test/test_metadata.py
phw: I agree, this behavior makes sense. The old behavior is not very logical. For a operation like copy the expected behavior really should be that you end up with a real copy, that means both store and deleted are the same.

We have cases where we want to preserve the deleted tags for files, but we already handle those in `file.copy_metadata`

picard/metadata.py
```diff
@@ -290,49 +320,62 @@ def clear_deleted(self):
...
     def set(self, name, values):
-        super().__setitem__(name, values)
-        if name in self.deleted_tags:
-            self.deleted_tags.remove(name)
+        self._store[name] = values
+        self.deleted_tags.discard(name)
 
     def __setitem__(self, name, values):
         if not isinstance(values, list):
             values = [values]
```
phw: Maybe we should accept any iterable here (so it can be converted to a list with `list(values)`, I recently had a case where it would have been convenient to pass a set.

picard/metadata.py
```diff
@@ -290,49 +320,62 @@ def clear_deleted(self):
...
         values = [str(value) for value in values if value]
```
phw: If we already touch this we should probably take care of the inconsistency that `__setitem__` only stores string values and `add` stores the values as is. E.g. 

```python
m["sometag"] = 1  # Stored as string "1"
m.add("sometag", 1)  # Stored as integer 1
```

Actually having the tags treated as strings consistently would allow us to eliminate quite a few type checks and string conversions currently done when setting and reading tags. There might come probably some edge cases where we would like to store non-string values, but those should be rare. As soon as tags get handled by tagger script they get converted into strings anyway. If there will be some need we could still it some `setraw` method or similar.

zas: Yes, for now, i tried to preserve the current behavior, but, clearly, the goal is to make things much more documented and consistent.
And some stuff can be removed (like `Metadata.delete()` which is redundant with `Metatada.__delitem__()` which, as bonus, provide `del m['tag']` syntax, which is more natural imho)

I don't want to start any modification yet, the idea is to first merge this, than start to clean up code actually using `Metadata` (i was forced to change some tests because of `dict.get()` hacky stuff, but this PR has no other changes unrelated to this refactoring).

picard/metadata.py
zas: Yes, i thought about it, but decided to not modify this yet. I will do once we merged this.
Accepting any iterable would be much more logical (and convenient).

zas:
>  I pointed out a few things I would like to see changed in this class, but we could as well do that in separate PRs.

Yes, I'd prefer separarate PRs, this PR is complex enough, even though it doesn't (shouldn't) break compatibility, getting rid of this `dict` pseudo inheritance was very needed.
The fact it was using `dict` as parent, while not providing much of a `dict` behavior is a non-sense we can finally get rid of (thanks to `MutableMapping`)

picard/metadata.py
```diff
@@ -59,15 +61,30 @@ class Metadata(dict):
...
-    def __init__(self):
-        super().__init__()
-        self.images = ImageList()
-        self.has_common_images = True
+    def __init__(self, *args, deleted_tags=None, images=None, length=None, **kwargs):
+        self._store = dict()
         self.deleted_tags = set()
         self.length = 0
+        self.images = ImageList()
+        self.has_common_images = True
```
zas: `has_common_images` isn't considered in `update()`, not sure about this.
Not even sure what it does.
@phw we'll have to take care of this, after this PR is merged.

test/test_metadata.py
```diff
@@ -187,3 +187,150 @@ def test_compare_deleted(self):
...
+        m = self.metadata_d2
+
+        m.update(c=3, d=['u', 'w'], x='')
+        self._check_mapping_update(m)
+
+    def test_metadata_mapping_update_noparam(self):
+        # update without parameter
+        m = self.metadata_d2
+
+        m.update()
```
zas: Actually this isn't really great, as it differs from `dict` behavior:

```
>>> d = dict()
>>> dict.update()
TypeError: descriptor 'update' of 'dict' object needs an argument
```

@phw do you think we should raise a TypeError too ?

zas: @phw: i did it in https://github.com/metabrainz/picard/pull/1137/commits/f6e3aedf7a9fa6cd7134fd79e18e974ead429e19
It makes more sense

PR #1143, opened 2019-03-21
Cleanup missing clear existing tags setting fix

zas:
A bug was introduced recently in https://github.com/metabrainz/picard/commit/04dcd4691bd57f2e24f2054e47e666ab843d891d
It was fixed in https://github.com/metabrainz/picard/commit/4990cef589496cabf0613bee4b6b2bac6a11797c

This PR is a cleanup of this fix

### DIFF
--- a/picard/ui/options/renaming.py
+++ b/picard/ui/options/renaming.py
@@ -33,6 +33,7 @@ from picard.script import (
     ScriptError,
     ScriptParser,
 )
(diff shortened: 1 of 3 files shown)


zas:
@phw : i pushed the actual fix to master, because it was breaking options badly, this PR is just a more advanced version of the fix

PR #1144, opened 2019-03-22
Clean up various matching iterations, unify a bit

zas:
Try to unify the style a bit, and make the code easier to understand.

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -18,8 +18,10 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from operator import itemgetter
-from collections import defaultdict
+from collections import (
+    defaultdict,
+    namedtuple,
+)
 import traceback
 
 from PyQt5 import (
@@ -56,6 +58,7 @@ from picard.script import (
 )
 from picard.track import Track
 from picard.util import (
+    find_best_match,
     format_time,
     mbid_validate,
 )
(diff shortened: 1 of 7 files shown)


zas:
@phw : i'm finally happy with this stuff, please review

picard/album.py
```diff
@@ -469,7 +472,7 @@ def _remove_file(self, track, file):
         file.metadata_images_changed.disconnect(self.update_metadata_images)
         remove_metadata_images(self, [file])
 
-    def _match_files(self, files, recordingid=None, threshold=0):
+    def _match_files(self, files, recordingid=None, threshold=0, debug=True):
```
phw: Why the `debug` parameter here? I don't see it used anywhere, and it is just guarding a call to `log.debug` below.

picard/album.py
```diff
@@ -500,16 +505,24 @@ def build_tracks_cache():
                 if track:
                     yield (file, track)
                     continue
+
             # try to match by similarity
-            # TODO: find a way to speed up this part, it needs to iterate all
-            # tracks for each file, and metadata.compare() is rather complex
```
phw: The todo comment still applies for the new implementation. The complexity is now somewhat hidden in the calls to `candidates()` and `find_best_match()`, but the basic algorithm of iterating over all tracks, applying `metadata.compare()` on each and finally use the one with highest score stays unchanged. To be honest I don't know what we can do against that other then make sure `metadata.compare()` is as fast as it can be, though.

picard/album.py
zas: Because i want to make this testable, and log.debug() is just for output, perhaps not the correct approach though.

picard/album.py
zas: Yes, but i concluded the same thing: we can't avoid to iterate over each file / track so the TODO doesn't make much sense, so i prefer to remove it.

picard/album.py
zas: Finally i remove the `debug` parameter and associated debug output, they aren't strictly needed.
See https://github.com/metabrainz/picard/pull/1144/commits/3bd05efadac0c8776737b94c9c1be100eacc7607

PR #1145, opened 2019-03-24
Oauth cleanup: use properties

zas:
Use properties to get rid of direct access to config.persist

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/oauth.py
+++ b/picard/oauth.py
@@ -98,11 +98,19 @@ class OAuthManager(object):
         else:
             config.persist["oauth_access_token_expires"] = value
 
+    @property
+    def username(self):
+        return config.persist["oauth_username"]
+
+    @username.setter
+    def username(self, value):
+        config.persist["oauth_username"] = value
+
     def is_authorized(self):
         return bool(self.refresh_token and self.refresh_token_scopes)
 
     def is_logged_in(self):
-        return self.is_authorized() and bool(config.persist["oauth_username"])
+        return self.is_authorized() and bool(self.username)
 
     def revoke_tokens(self):
         # TODO actually revoke the tokens on MB (I think it's not implementented there)
@@ -144,10 +152,6 @@ class OAuthManager(object):
         log.debug("OAuth: got access_token %s that expires in %s seconds", access_token, expires_in)
         self.access_token = access_token
         self.access_token_expires = int(time.time() + expires_in - 60)
-
-    def set_username(self, username):
-        log.debug("OAuth: got username %s", username)
-        config.persist["oauth_username"] = username
 
     def refresh_access_token(self, callback):
         log.debug("OAuth: refreshing access_token with a refresh_token %s", self.refresh_token)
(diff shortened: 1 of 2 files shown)


picard/oauth.py
```diff
@@ -44,41 +44,99 @@ class OAuthManager(object):
 
     def __init__(self, webservice):
         self.webservice = webservice
+        self.persist= config.persist
```
phw: Missing space

picard/oauth.py
```diff
@@ -44,41 +44,99 @@ class OAuthManager(object):
...
                 self.forget_access_token()
                 self.refresh_access_token(callback)
 
+    @staticmethod
+    def host_port():
+        return config.setting['server_host'], config.setting['server_port']
+
     def get_authorization_url(self, scopes):
-        host, port = config.setting['server_host'], config.setting['server_port']
+        host, port = self.host_port()
```
phw: Why not make these properties as you've done for other settings?

picard/oauth.py
```diff
@@ -88,27 +146,26 @@ def get_authorization_url(self, scopes):
...
         log.debug("OAuth: got access_token %s that expires in %s seconds", access_token, expires_in)
-        config.persist["oauth_access_token"] = access_token
-        config.persist["oauth_access_token_expires"] = int(time.time() + expires_in - 60)
+        self.access_token = access_token
+        self.access_token_expires = int(time.time() + expires_in - 60)
 
     def set_username(self, username):
         log.debug("OAuth: got username %s", username)
-        config.persist["oauth_username"] = username
+        self.username = username
```
phw: We don't really need `set_username` if it is only a wrapper around the property.

picard/oauth.py
zas: Fixed in dc5d383f3cbef49fb0605a67b4832b6f5d8d6c52

picard/oauth.py
zas: You're right, fixed in 4269fb8e8ea74cd2acdb603dde4bc7254557b826

picard/oauth.py
zas: I kept it for now, because it is used for logging.

About logging, i think it should be disabled by default for OAuth, i'm not confortable with OAuth-related stuff being published along debug log. What do you think ?

phw: Agreed. We should probably not entirely disable logging, but stop logging details like tokens and probably even username.

zas:
I reworked this PR a bit applying suggested changes, and fixing errors, so rebased.

picard/tagger.py
```diff
@@ -292,7 +292,7 @@ def mb_login(self, callback, parent=None):
         else:
             callback(False)
 
-    def on_mb_authorization_finished(self, callback, successful):
+    def on_mb_authorization_finished(self, callback, successful=False):
```
phw: Why the default value here?

zas: To make it a kw arg, so it's possible to add other arguments without breaking code calling this callback. Not strictly necessary, but since i changed calling code to use `callback(successful=value)` it makes more sense imho,  where `callback` is a `partial`.
See https://github.com/metabrainz/picard/blob/8608bf16e8f1fc66df1f1dfd64fd4ebe6b75539b/picard/tagger.py#L289-L291

PR #1146, opened 2019-03-24
Pylint errors/warnings cleanup

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -249,6 +249,18 @@ class Option(QtCore.QObject):
         return cls.registry.get((section, name))
 
 
+def _convert_to_bool(value):
+    # The QSettings IniFormat saves boolean values as the strings "true"
+    # and "false". Thus, explicit boolean and string comparisons are used
+    # to determine the value. NOTE: In PyQt >= 4.8.3, QSettings.value has
+    # an optional "type" parameter that avoids this. But we still support
+    # PyQt >= 4.5, so that is not used.
+    return value is True or value == "true"
+
+def _convert_to_int_list(values):
+    return list(map(int, values))
+
+
 class TextOption(Option):
 
     convert = str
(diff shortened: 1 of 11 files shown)


picard/config.py
```diff
@@ -249,21 +249,26 @@ def get(cls, section, name):
         return cls.registry.get((section, name))
 
 
+def _convert_to_bool(value):
+    # The QSettings IniFormat saves boolean values as the strings "true"
+    # and "false". Thus, explicit boolean and string comparisons are used
+    # to determine the value. NOTE: In PyQt >= 4.8.3, QSettings.value has
+    # an optional "type" parameter that avoids this. But we still support
+    # PyQt >= 4.5, so that is not used.
```
samj1912: This comment is not valid anymore?

picard/util/webbrowser2.py
```diff
@@ -26,10 +26,8 @@
 from picard.const import PICARD_URLS
 
 
-"""
-A webbrowser extension for Picard. It handles and displays errors
-in PyQt and also adds a utility function for opening Picard URLS.
-"""
+# A webbrowser extension for Picard. It handles and displays errors
```
samj1912: This is no longer a valid docstring?

picard/config.py
zas: I just address the pylint warning here, imho the comment is still valid (we don't use QSettings.value "type"), but we also don't use PyQt4 anymore, so dunno.
Is  there an issue here i didn't see ?

picard/util/webbrowser2.py
zas: In fact, it never was, since it's not at the start of file.
I'll convert it in a valid docstring, but since we only have few in code, module docstring or comment is the same.

zas: Fixed in https://github.com/metabrainz/picard/pull/1146/commits/8fe3ee23d82944ad244bfc8941c277eb3c93fb90

picard/config.py
phw: I agree, the comment is still valid. Since we are now on newer Qt versions we should look into this and see how we can simplify bool handling here, but that's for a separate PR.

PR #1148, opened 2019-03-25
PICARD-1496: display count of Other versions in menu and refactoring

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-PICARD-1496](https://tickets.metabrainz.org/browse/PICARD-PICARD-1496)

# Solution
picard_other_versions_with_count

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -18,6 +18,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from functools import partial
+from heapq import heappush, heappop
 import os
 import re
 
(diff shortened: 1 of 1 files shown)


zas:
@phw : i wonder about the count vs label.

Other versions in fact includes the current version (so 7 = current + 6 other versions)
At the same time, i couldn't find a better label. Technically that's all releases in the same release group, limited to 100 (by query).
And the menu displays all releases with the current one checked.

Ideas, thoughts ? (btw, it doesn't have to be in this PR, just raising the "issue")

phw:
> Other versions in fact includes the current version (so 7 = current + 6 other versions)

Since the displayed list makes this clear that the current version is included I think this is fine.

PR #1149, opened 2019-03-25
Fix ASFFile._save modifies metadata

phw:
# Summary
`ASFFile._save` did modify the passed `Metadata` object. Fixed this and added a test to make sure this does not happen.

Also simplified `test_formats.py` by using the new ability to pass tags directly to the `Metadata()` constructor.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/formats/asf.py
+++ b/picard/formats/asf.py
@@ -221,9 +221,9 @@ class ASFFile(File):
             if name.startswith('lyrics:'):
                 name = 'lyrics'
             elif name == '~rating':
-                values[0] = int(values[0]) * 99 // (config.setting['rating_steps'] - 1)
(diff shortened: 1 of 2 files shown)


zas:
Perhaps create a ticket for it, since it fixes something.

phw:
> Perhaps create a ticket for it, since it fixes something.

I thought about it, but I don't think it fixes a current real world problem. In the actual code `File._save()` is only called on a copy of the actual metadata, so this has no impact. I only found this while writing tests for another story. I still think this is unexpected behavior and a potential source for future bugs, but so far this is more about internal code cleanup then about fixing an actual issue.

PR #1150, opened 2019-03-26
PICARD-1492: Fix saving Vorbis ratings with invalid chars in user e-mail

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When trying to save Vorbis tags with a configured rating e-mail that contains characters that are not valid for Vorbis keys an exception is thrown.

* JIRA ticket (_optional_): [PICARD-1492](https://tickets.metabrainz.org/browse/PICARD-1492)

# Solution
Remove invalid characters from configured user email before using it as part of the Vorbis rating key. Make sure to do this sanitizing on both save and load to ensure consistent behavior.

# Action

### DIFF
--- a/picard/formats/vorbis.py
+++ b/picard/formats/vorbis.py
@@ -49,6 +49,13 @@ from picard.util import (
 )
 
 
+INVALID_CHARS = re.compile('([^\x20-}]|=)')
+
+
+def sanitize_key(key):
+    return INVALID_CHARS.sub('', key)
+
+
 class VCommentFile(File):
 
     """Generic VComment-based file."""
@@ -94,7 +101,7 @@ class VCommentFile(File):
                         name, email = name.split(':', 1)
                     except ValueError:
                         email = ''
-                    if email != config.setting['rating_user_email']:
+                    if email != sanitize_key(config.setting['rating_user_email']):
                         continue
                     name = '~rating'
                     try:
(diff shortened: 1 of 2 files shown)


picard/formats/vorbis.py
```diff
@@ -49,6 +49,13 @@
 )
 
 
+INVALID_CHARS = re.compile('([^\x20-}]|=)')
```
zas: What is the source for this list of invalid characters ?
Also please add specific tests for it (in addition of format's one i mean).

phw: This is from the Vorbis specs at https://www.xiph.org/vorbis/doc/v-comment.html#vectorformat and is the reverse of what is tested at https://github.com/quodlibet/mutagen/blob/master/mutagen/_vorbis.py#L26

I added some docstring and a test. Regarding the test I think `test_formats.py` is getting too crowded. I think we should split it into separate files per format (with the common test cases kept at a central file), but that's something for another PR.

phw: See commit https://github.com/metabrainz/picard/pull/1150/commits/cf9528bd86d3cd49eecc6c7ffcb49c63776ade52

zas: >  I think we should split it into separate files per format (with the common test cases kept at a central file), but that's something for another PR.

Yes good idea !

PR #1151, opened 2019-03-26
PICARD-1468: Localize NSIS installer

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The NSIS installer for Windows currently is only shipped in English, but NSIS allows localization.

* JIRA ticket (_optional_): [PICARD-1468](https://tickets.metabrainz.org/browse/PICARD-1468)

# Solution
Extract all the localizable strings into separate language files.

# Action
My initial requirements for this included support for Transifex, but I actually think we don't need this for a first version.

- Support for Transifex would require a script to convert to and from a format that is supported by Transifex, which adds uite some complexity
- Strings in the installer don't change frequently
- Translations can only be done completely, partial translated files are not supported. There is not automatic fallback to e.g. English, which means an empty string in one of the translations will lead to incomplete UI. Of course you can just copy over the English original to the translation to avoid this. But overall I think it is better to stick to manually submitted translations for the installer for now.


phw:
I'm waiting on feedback by Roberto, who contacted me about implementing this.

bovirus:
For me is OK.
Please merge it and I will translate the file for Italian language. 
Thanks.

zas:
> My initial requirements for this included support for Transifex, but I actually think we don't need this for a first version.

Agreed, better than nothing.
But on the long-term translations have to be managed with transifex somehow.

> 
>     * Support for Transifex would require a script to convert to and from a format that is supported by Transifex, which adds uite some complexity
> 
>     * Strings in the installer don't change frequently
> 
>     * Translations can only be done completely, partial translated files are not supported. There is not automatic fallback to e.g. English, which means an empty string in one of the translations will lead to incomplete UI. Of course you can just copy over the English original to the translation to avoid this. But overall I think it is better to stick to manually submitted translations for the installer for now.

Ok, better than nothing, but it would still be great to be able to fit this with transifex.
Found a very small discussion about it: https://nsis-dev.github.io/NSIS-Forums/html/t-321905.html
And https://github.com/owncloud/client/tree/master/admin/win/nsi/l10n have a script doing tx -> nsh

phw:
@bovirus Thanks for the feedback. I have merged this so you can do the Italian translation.

PR #1152, opened 2019-03-26
Coverart stuff cleanup

zas:
The code (mostly mine) is very messy, so time to clean up.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/coverart/providers/__init__.py
+++ b/picard/coverart/providers/__init__.py
@@ -95,15 +95,6 @@ def cover_art_providers():
         yield ProviderTuple(name=p.name, title=p.title, enabled=order[p.name].enabled, cls=p)
 
 
-def is_provider_enabled(provider_name):
-    """Test if provider with name `provider_name` was enabled
-    by user through options"""
-    for name, checked in config.setting['ca_providers']:
-        if name == provider_name:
-            return checked
-    return False
-
-
 class CoverArtProviderMetaClass(type):
     """Provide default properties name & title for CoverArtProvider
     It is recommended to use those in place of NAME and TITLE that might not be defined
@@ -146,14 +137,7 @@ class CoverArtProvider(metaclass=CoverArtProviderMetaClass):
         self.album = coverart.album
 
     def enabled(self):
-        """By default, return True if user enabled the provider
-        through options. It is used when iterating through providers
-        to decide to skip or process one.
-        It can be subclassed to add conditions."""
-        enabled = is_provider_enabled(self.name)
-        if not enabled:
-            log.debug("%s disabled by user" % self.name)
-        return enabled
+        return not self.coverart.front_image_found
 
     def queue_images(self):
         # this method has to return CoverArtProvider.FINISHED or
--- a/picard/coverart/providers/__init__.py
+++ b/picard/coverart/providers/__init__.py
@@ -63,8 +63,8 @@ class ProviderOptions(OptionsPage):
 def register_cover_art_provider(provider):
     _cover_art_providers.register(provider.__module__, provider)
     if hasattr(provider, 'OPTIONS') and provider.OPTIONS:
-        provider.OPTIONS.NAME = provider.NAME
-        provider.OPTIONS.TITLE = provider.TITLE or provider.NAME
+        provider.OPTIONS.NAME = provider.name
+        provider.OPTIONS.TITLE = provider.title
         register_options_page(provider.OPTIONS)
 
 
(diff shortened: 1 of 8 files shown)


zas:
@phw it will break amazon cover art plugin, but fix is easy

picard/coverart/providers/__init__.py
```diff
@@ -62,37 +63,52 @@ def __init__(self, parent=None):
...
+    # (yet) unknown providers are placed at the end, disabled
+    ordered_providers = sorted(_cover_art_providers, key=lambda p: order[p.name].position)
+
+    log.debug("CA Providers order: %s", ' > '.join([p.name for p in ordered_providers]))
+
+    for p in ordered_providers:
+        yield ProviderTuple(name=p.name, title=p.title, enabled=order[p.name].enabled, cls=p)
+
+
+class CoverArtProviderMetaClass(type):
```
phw: I don't see the use of a meta class here, why not just add those two properties to the `CoverArtProvider` base class?

picard/coverart/providers/__init__.py
```diff
@@ -62,37 +63,52 @@ def __init__(self, parent=None):
 def register_cover_art_provider(provider):
     _cover_art_providers.register(provider.__module__, provider)
     if hasattr(provider, 'OPTIONS') and provider.OPTIONS:
-        provider.OPTIONS.NAME = provider.NAME
-        provider.OPTIONS.TITLE = provider.TITLE or provider.NAME
+        provider.OPTIONS.NAME = provider.name
+        provider.OPTIONS.TITLE = provider.title
```
phw: Why do we actually need `provider.OPTIONS`? I checked the existing plugins, and they don't use this but register an options page separately. Not sure what advantage using provider.OPTIONS would bring.

picard/coverart/providers/__init__.py
```diff
@@ -120,15 +136,11 @@ def __init__(self, coverart):
...
-        It can be subclassed to add conditions."""
-        enabled = is_provider_enabled(self.NAME)
-        if not enabled:
-            log.debug("%s disabled by user" % self.NAME)
-        return enabled
+    @staticmethod
+    def enabled():
+        # kept for compatibility
+        # deprecated, and always returns True
+        return True
```
phw: The comment seems to be wrong, as this is still in use in derived classes. There are a few cases where this still can be useful that a provider can decide to be skipped, e.g. the Fanart.tv plugin makes use of this to disable itself if no API key is available. Also the CAA provider has some additional checks.

On the other hand all current implementations check something like `return not self.coverart.front_image_found`, this seems unnecessary. That's something the `CoverArt` class could already take care of.

phw: One use case for not checking `self.coverart.front_image_found` would be if the plugin is meant to download additional images other than front, and you might want to do this always. E.g. Fanart.tv and TheAudioDb plugins currently can also download media (CD) images. They don't make use of this yet, but one scenario could be that they can always download the CD image even if a front image has already been found.

So I would actually suggest to change the default implementation here to 

```
def enabled(self):
        return not self.coverart.front_image_found
```

And only override this in sub classes if necessary (and usually call `super().enabled()`). That way implementations can make easy use of the default, hut if necessary still can override the check for the front image.

picard/coverart/providers/__init__.py
phw: For test I tried to use `OPTIONS` for the Fanart.tv plugin. I think using `ProviderOptions` provides very little benefit. Basically you replace a custom three line constructor necessary to set `self.ui` and calling `self.ui.setupUi()` with a single class property `_options_ui`, but that's something we could even do in the base `OptionsPage` class. Apart from that using `ProviderOptions` only spares you a single call to `register_options_page` and you don't have to specify a parent. The visible effect for the user is that all cover art options show up under the cover art section and not also under plugins, but that's something we could easily change in the plugins (and I already though about doing this).

phw: This is the diff for Fanarttv: https://gist.github.com/phw/08ba5a69d076a6c89c219da3821d1274

zas:
> I tested this with existing plugins and it does not seem to break anything.

Yes, because meanwhile i fixed it ;) but it was too late and forgot to remove the comment.

picard/coverart/providers/__init__.py
zas: Using `CoverArtProvider` properties didn't work for me,  i tried but they return `property object` instead of `str`.

I made a little [test script](https://gist.github.com/zas/3e461215fe963f6f7e6b33eca67cc124) to show the behavior.
Perhaps there's another way to achieve the expected behavior but i found using a metaclass was doing what i expected.
Ideas / improvements are still welcome.

picard/coverart/providers/__init__.py
zas: Done in 56466b9433f15e8

picard/coverart/providers/__init__.py
zas: I think your diff is reversed, but i get the point and i agree we can get rid of this  and directly use OPTIONS. It has to be in a different PR though IMHO.

picard/coverart/providers/__init__.py
phw: Yes, this looks  even cleaner now.

PR #1156, opened 2019-03-27
Installer Italian translation

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Italian translation by @bovirus

# Solution

# Action


phw:
@bovirus Could you test the installer at https://ci.appveyor.com/project/MetaBrainz/picard/builds/23398869/job/pv5h22vm0cprhpfp/artifacts ?

bovirus:
I tested the installer.
Seems it works except for module/component selection. The relative list is empty.

Please change in the installer Italian language

LangString MsgAlreadyInstalled ${LANG_ENGLISH} '${PRODUCT_NAME} è già installato.$\n$\nFai clic su "OK" per disinstallare la versione installata o su "Annulla" per annullare l'aggiornamento.'

in

LangString MsgAlreadyInstalled ${LANG_ENGLISH} "${PRODUCT_NAME} è già installato.$\n$\nFai clic su 'OK' per disinstallare la versione installata o su 'Annulla' per annullare l'aggiornamento."

I swapped the use of "" and ''.
This because all strings are under "xxxx"

It seems that NSIS Italian base language is not updated.
What's the version of NSIS that are you using?

phw:
> Seems it works except for module/component selection. The relative list is empty.

Can you explain this in detail, maybe show a screenshot? Then I can check

> It seems that NSIS Italian base language is not updated.
> What's the version of NSIS that are you using?

That's NSIS 3.03, we are limited to what is available on https://www.appveyor.com/docs/windows-images-software/

> I swapped the use of "" and ''.

That doesn't really matter, we should use the quotes which would be expected in a Italian text. It doesn't matter how the string in the file is finally enclose, you just have to escape the quotation marks accordingly with `$\`.

E.g. you an write either `'...clic su "OK" per...'`, but you can also sourround the string with double quotes and escape them inside, e.g. `"...clic su $\"OK$\" per..."`.

So ignoring the technical details, just from a translation point of view, would you prefer double or single quotes in the Italian text?

bovirus:
I prefer as in the file that I uploaded.

"...clic su 'OK' per..."

Start and end with" and use 'xx' for button name.

bovirus:
MBP-02
MBP-01

Please check screenshot

phw:
@bovirus I updated the file, updated build is at https://ci.appveyor.com/project/MetaBrainz/picard/builds/23400892/job/4e0i5ql1khrkru6f/artifacts

bovirus:
Two small problems

MBP-01

MBP-02

phw:
I made the uninstaller option text translatable.

Not too sure how to label the main data. How do other installers label this? I had seen this labeled "Required" before, but maybe it's a bit unclear how this is meant. Just "Program" I find also a bit confusing. What about being more verbose and write "Program data (required)"?

phw:
I looked at NSIS itself, it labels this section "NSIS Core Files (required)". We could also just go for "MusicBrainz Picard (required)"?

bovirus:
"Required" not explain what module is.
The best could be "Program files (required)" but I'm not sure if it's too long for the small width of component list window.
If it's tool long could be used "Program (required)"

phw:
I changed it to "Program files (required)". I also changed the layout for the component selection screen, so the descriptions are now at the bottom and there is more room for longer component names. The screenshot below shows the German version, where the string is even a bit longer than the Italian one. So this should be fine.

grafik

phw:
I went ahead and merged this. Please let me know if there are any changes needed to the localization.

bovirus:
Where can I get latest version of the installer for the final check?

phw:
The currently latest build is at https://ci.appveyor.com/project/MetaBrainz/picard/build/job/4so84rg832oq0ib9/artifacts

PR #1160, opened 2019-03-28
Fix loading BoolOption from config

phw:
@zas Since commit 2c457b8dcd7c77f78557dfd02a4111862c0eb31c we have a strange bug, where `BoolOption` values seem to get reset to their default values.

I don't fully understand it yet. In both implementations the `not hasattr(self, "convert")` call at https://github.com/metabrainz/picard/blob/master/picard/config.py#L243 is `False`, so it does not get overridden by `type(default)`. But with the changes from 2c457b8dcd7c77f78557dfd02a4111862c0eb31c the convert function never gets called. I assume `IntListOption` is affected as well, but we don't use that.

Any ideas. Just reverting your change of course seemingly fixes the issue, but this does not seem to be the solution. Something is wrong here, and the change just uncovered it.

### DIFF
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -1,0 +1,133 @@
+from test.picardtestcase import PicardTestCase
+
+from picard import config
+
+
+class CommonTests:
+
+    class CommonOptionTest(PicardTestCase):
+        opt_type = config.Option
+        default_value = 'somevalue'
+
+        def setUp(self):
+            self.opt_type.registry = {}
+
+        def test_constructor(self):
+            opt = self.opt_type('test', 'option1', self.default_value)
+            self.assertEqual('test', opt.section)
+            self.assertEqual('option1', opt.name)
+            self.assertEqual(self.default_value, opt.default)
+
+        def test_registry(self):
+            opt = self.opt_type('test', 'option1', self.default_value)
+            self.assertEqual(opt, self.opt_type.get('test', 'option1'))
+
+
+class OptionTest(CommonTests.CommonOptionTest):
+
+    def test_default_convert(self):
+        for default in ['somevalue', True, [], tuple(), 42]:
+            opt = config.Option('test', 'option1', default)
+            self.assertEqual(type(default), opt.convert)
+            self.assertEqual(default, opt.convert(default))
+
+
+class TextOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.TextOption
+    default_value = 'test'
+
+    def _test_convert(self, obj):
+        self.assertEqual('', obj.convert(''))
+        self.assertEqual('test', obj.convert('test'))
+        self.assertEqual('42', obj.convert(42))
+
+    def test_convert_instance(self):
+        opt = self.opt_type('test', 'option1', self.default_value)
+        self._test_convert(opt)
+
+    def test_convert_static(self):
+        self._test_convert(self.opt_type)
+
+
+class BoolOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.BoolOption
+    default_value = False
+
+    def _test_convert(self, obj):
+        self.assertTrue(obj.convert(True))
+        self.assertTrue(obj.convert('true'))
+        self.assertFalse(obj.convert(None))
+        self.assertFalse(obj.convert('unknown'))
+        self.assertFalse(obj.convert('false'))
+
+    def test_convert_instance(self):
+        opt = self.opt_type('test', 'option1', self.default_value)
+        self._test_convert(opt)
+
+    def test_convert_static(self):
+        self._test_convert(self.opt_type)
+
+
+class IntOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.IntOption
+    default_value = 42
+
+    def _test_convert(self, obj):
+        self.assertEqual(42, obj.convert(42))
+        self.assertEqual(42, obj.convert('42'))
+
+    def test_convert_instance(self):
+        opt = self.opt_type('test', 'option1', self.default_value)
+        self._test_convert(opt)
+
+    def test_convert_static(self):
+        self._test_convert(self.opt_type)
+
+
+class FloatOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.FloatOption
+    default_value = 42.5
+
+    def _test_convert(self, obj):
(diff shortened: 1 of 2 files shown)


phw:
Ok, I found out something, see commit 45b4bde0871be43154c05b017c01e899cd44037a.

I know set `@staticmethod` on the conversion functions. The error was invisible because we hide an exception, but actually these functions are treated like instance methods. It seems to behave different for builtins like `bool`. Example:

```python

def _somefunc():
  pass

class T:
  somefunc = _somefunc
  otherfunc = bool

i = T()
i.somefunc()  # Raises TypeError: _somefunc() takes 0 positional arguments but 1 was given
i.otherfunc()  # Works, just calls bool() without passing `self`
```

phw:
@zas I added unit tests to catch issues like this

test/test_config.py
```diff
@@ -0,0 +1,133 @@
...
+
+
+class BoolOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.BoolOption
+    default_value = False
+
+    def _test_convert(self, obj):
+        self.assertTrue(obj.convert(True))
+        self.assertTrue(obj.convert('true'))
+        self.assertFalse(obj.convert(None))
```
zas: May be add:

```python
self.assertFalse(obj.convert(False))
```

It is a bit redundant with this None test, but it makes sense since you have `self.assertTrue(obj.convert(True))` two lines above.

test/test_config.py
```diff
@@ -0,0 +1,133 @@
...
+        self._test_convert(self.opt_type)
+
+
+class FloatOptionTest(CommonTests.CommonOptionTest):
+    opt_type = config.FloatOption
+    default_value = 42.5
+
+    def _test_convert(self, obj):
+        self.assertEqual(42.5, obj.convert(42.5))
+        self.assertEqual(42.5, obj.convert('42.5'))
```
zas: Is conversion from `False/True` or 'xxx' tested somewhere ?

phw:
> Forcing a test to failed i fet 2 times the same error, are tests run twice ??

Kind of, it tests invoking the function both on an instance (`Option().convert(v)`) and on the class (`Option.convert(v)`). This is specifically to verify the convert function is properly setup as a static method.

test/test_config.py
phw: Added tests in d0563fa5c4eb084ddb0b2ae4a7d65a12a0275223

test/test_config.py
phw: Added tests in d0563fa5c4eb084ddb0b2ae4a7d65a12a0275223

zas:
> > 
> 
> Kind of, it tests invoking the function both on an instance (`Option().convert(v)`) and on the class (`Option.convert(v)`). This is specifically to verify the convert function is properly setup as a static method.

Ok, got it.
I think we should have only one test failure at a time, if you group both tests in the same `test_*()` the first one failing will raise the exception and the other will not be run.

```python
class FloatOptionTest(CommonTests.CommonOptionTest):
    opt_type = config.FloatOption
    default_value = 42.5

    def _test_convert(self, obj):
        self.assertEqual(42.5, obj.convert(42.5))
        self.assertEqual(42.5, obj.convert('42.5'))

    def _convert_instance(self):
        opt = self.opt_type('test', 'option1', self.default_value)
        self._test_convert(opt)

    def _convert_static(self):
        self._test_convert(self.opt_type)

    def test_convert(self):
        self._convert_static()
        self._convert_instance()
```

zas:
@phw : i found an issue related to `ListOption`:

Here is the (with extra debug) output:

```
E: 23:42:28,027 /home/zas/src/picard/picard/config.value:109: Error reading option value for key setting/list_of_scripts
  File "/home/zas/src/picard/picard/config.py", line 106, in value
    return option_type.convert(self.__config[key])
TypeError: 'NoneType' object is not iterable
E: 23:42:28,028 /home/zas/src/picard/picard/config.value:115: Called from: '/home/zas/src/picard/picard/config.py' 70 __getitem__ ['        return self.value(name, opt, opt.default)\n'] 0
E: 23:42:28,028 /home/zas/src/picard/picard/config.value:115: Called from: '/home/zas/src/picard/picard/ui/options/scripting.py' 409 load ['        self.list_of_scripts = config.setting["list_of_scripts"]\n'] 0
E: 23:42:28,028 /home/zas/src/picard/picard/config.value:115: Called from: '/home/zas/src/picard/picard/ui/options/dialog.py' 137 __init__ ['            page.load()\n'] 0
```

Picard.ini is containing:

```
list_of_scripts=@Invalid()
```

To trigger the issue, remove any config, open Options, Make It So! (saving options once), close Picard, open it again, open Options again

The exception is raised because `list()` is called with `None` as argument:

```
>>> list(None)
TypeError: 'NoneType' object is not iterable
```

I wonder if those were hidden by silent exception too.
Also not sure about how to fix that.

zas:
@phw : the issue is related to https://www.riverbankcomputing.com/static/Docs/PyQt5/pyqt_qsettings.html `type` parameter, we don't have one, and i guess the empty list is converted to None on read because of that.
See also https://stackoverflow.com/questions/37066961/pyqt4-saving-list-to-qsettings:
 
> NB: for the special case of a list of strings, it's best to use this:
>
>   settings.value('finishcount', [], 'QStringList')
>
> rather than specifying the type with str. Otherwise, empty lists don't get treated properly.

AFAIK, we don't do anything like that.

phw:
> I wonder if those were hidden by silent exception too.

Definitely, this is exactly the kind of issues that were hidden, with Picard's options just falling back to their default. We definitely should get some defined handling here. I personally also would simplify the current handling and actually define a proper `convert` method that can be overridden instead of this `if not hasattr(self, "convert")` check, that somewhat hides what's going on here.

> `type` parameter, we don't have one, and i guess the empty list is converted to None on read because of that.

This goes together with the todo comment for `BoolOption`, which would also benefit from the type parameter. We definitely should look into this and refactor things.

zas:
> We definitely should look into this and refactor things.

I started something on https://github.com/metabrainz/picard/pull/1169
I'm far from happy with it though it fixes the issue, and get rid of useless stuff.
I added tests for each option, and i want to add tests for non-interactive upgrade hooks
Also tests have log ouput, not sure how we should handle that (disable logging i guess).

PR #1168, opened 2019-03-29
Format NEWS as Markdown

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

Currently the NEWS.txt uses some rather arbitrary formatting. Converting it to markdown would give it some better defined structure and would allow us to use a proper Markdown parser in places where we make use of this data, e.g. for https://picard.musicbrainz.org/changelog/

If we want to go that way we also need to update picard-website. I can work on that if we decide to do the Markdown conversion.

# Solution

# Action


zas:
I'm ok with the move to markdown, it will need few changes around, but historic NEWS.txt isn't good anymore for sure.
Though i'd like actual links in markdown instead of just [PICARD-XXX]

phw:
@zas: I linked the ticket numbers. It of course makes the plain file a bit hareder to read, but the positive part is the rendered version is much nicer.

Also adding this to the homepage will simplify things a lot. Instead of our current regex approach we can just use a Markdown parser.

NEWS.md
```diff
@@ -0,0 +1,1163 @@
...
+- [PICARD-1012](https://tickets.metabrainz.org/browse/PICARD-1012) - Buttons on the "User Interface" and "Scripting" pages are smaller than buttons in other places
+- [PICARD-1016](https://tickets.metabrainz.org/browse/PICARD-1016) - Multiple images in related tracks confusing
+- [PICARD-1021](https://tickets.metabrainz.org/browse/PICARD-1021) - Picard loads all pending files before quitting
+- [PICARD-1024](https://tickets.metabrainz.org/browse/PICARD-1024) - Allow specifying a configuration file path
+- [PICARD-1030](https://tickets.metabrainz.org/browse/PICARD-1030) - Allow to add/replace cover art images and keep existing cover art
+
+
+# Version 1.4 - 2017-02-14
+## Bug-fix
+- AcoustID submission fails with code 299 ([PICARD-82](https://tickets.metabrainz.org/browse/PICARD-82))
```
zas: It would be great to have those formatted as new entries

phw: Good point, did some more regexing and updated the old entries.

phw:
I think it's looking good now: https://github.com/metabrainz/picard/blob/9aaa38eec1d5a7c30b69771b4ac462a5163ea887/NEWS.md

PR #1169, opened 2019-03-31
Rework config/settings/options

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


zas:
@phw: i'll need you on this, still a lot of cleanup to do

- qt type issue is solved (but we still have convert)
- Option tests, per type
- upgrade functions tests
- test file is (too) big
- ConfigSection was simplified:
  - removed useless ConfigSection.__config
  - matches against option name instead of the full key, as the prefix is fixed (section name)
  - better logging (really showing the caller, much more useful than trace we had)

picard/config.py
```diff
@@ -288,11 +270,18 @@ class FloatOption(Option):
...
+
+@staticmethod
+def _convert_to_int_list(values):
+    return list(map(int, values))
 
 
 class IntListOption(Option):
 
     convert = _convert_to_int_list
+    qtype = 'QVariantList'
```
phw: Just wanting to note that we don't seem to use `IntListOption` anywhere. I already thought about removing it in my previous PR, but was unsure whether we want to keep it around for plugins. Thoughts?

test/test_config.py
```diff
@@ -1,138 +1,412 @@
...
+
+    def test_text_opt_set_read_back(self):
+        TextOption("setting", "text_option", "abc")
+
+        # set option to "def", and read back
+        self.config.setting["text_option"] = "def"
+        self.assertEqual(self.config.setting["text_option"], "def")
+        self.assertEqual(type(self.config.setting["text_option"]), str)
+
+    def test_text_opt_set_None(self):
```
phw: I think we should stick to lower cased method names as we do elsewhere. Same applies for similar methods below.

test/test_config.py
```diff
@@ -1,138 +1,412 @@
...
+        self.assertEqual(self.config.setting["text_option"], "")
+
+    def test_text_opt_set_empty(self):
+        TextOption("setting", "text_option", "abc")
+
+        # set option to ""
+        self.config.setting["text_option"] = ""
+        self.assertEqual(self.config.setting["text_option"], "")
+
+    def test_text_opt_xxx(self):
```
phw: Don't know what `xxx` means, I would name it `test_text_opt_set_invalid_value` or something like that.

test/test_config.py
```diff
@@ -1,138 +1,412 @@
...
+    def test_text_opt_convert(self):
+        opt = TextOption("setting", "text_option", "abc")
+        self.assertEqual(opt.convert(123), "123")
+
+    def test_text_opt_no_config(self):
+        TextOption("setting", "text_option", "abc")
+
+        # test default, nothing in config yet
+        self.assertEqual(self.config.setting["text_option"], "abc")
+        self.assertEqual(type(self.config.setting["text_option"]), str)
```
phw: You can use `self.assertIsInstance(self.config.setting["text_option"], str)` here.

test/test_config.py
```diff
@@ -1,138 +1,412 @@
...
+        self.config.setValue('setting/text_option', object)
+        self.assertEqual(self.config.setting["text_option"], 'abc')
+
+
+class TestPicardConfigBoolOption(TestPicardConfigCommon):
+
+    ### BoolOption
+    def test_bool_opt_convert(self):
+        opt = BoolOption("setting", "bool_option", False)
+        self.assertEqual(opt.convert(1), True)
```
phw: You can use `self.assertTrue(value)` or `self.assertFalse(value)` for these boolean checks.

zas:
> But I actually think we should have the config upgrade changes in a separate PR, this seems to be like an unrelated change. And its kind of a delicate area of the code to touch.
> 
> Also the file renaming changes seem to be unrelated as well.

Yes, this is why this PR is a draft, i'll split those changes in different PRs at the end.
I just need to have the whole figure.

picard/config.py
zas: I'd rather get rid of it, i'm not aware of any plugin using it.
And imho ListOption is enough

test/test_config.py
zas: Hmmm, yes, i capitalized the N because that's the None type, but i agree, i'll fix that

test/test_config.py
zas: Nor i :) i'll fix it

test/test_config.py
zas: But here i want to test the exact type.
https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertIsInstance suggests: `assertIs(type(obj), cls)`

test/test_config.py
zas: Those aren't equivalent:

```
value = "a"
self.assertTrue(value)        # <---- pass
self.assertEqual(value, True)  # <---- not pass
```

https://docs.python.org/3/library/unittest.html#unittest.TestCase.assertTrue says:
> This method should also be **avoided** when more specific methods are available (e.g. assertEqual(a, b) instead of assertTrue(a == b)), because they provide a better error message in case of failure.

zas:
Part of those changes are now in https://github.com/metabrainz/picard/pull/1172

zas:
Constants move is now in https://github.com/metabrainz/picard/pull/1173

zas:
Config upgrade tests moved to https://github.com/metabrainz/picard/pull/1174

zas:
Closing since it was splitted in smaller PRs

PR #1172, opened 2019-04-02
Rework Config / ConfigSection

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -29,7 +29,9 @@ from picard import (
     version_from_string,
     version_to_string,
 )
-from picard.util import LockableObject
+from picard.util import (
+    LockableObject,
+)
 
 
 class ConfigUpgradeError(Exception):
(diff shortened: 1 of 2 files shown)


picard/config.py
```diff
@@ -70,43 +65,42 @@ def __getitem__(self, name):
...
-                return option_type.convert(self.raw_value(name))
+            if name in self:
+                value = self.raw_value(name, qtype=option_type.qtype)
+                return option_type.convert(value)
             return default
-        except Exception:
-            log.error('Error reading option value', exc_info=True)
+        except Exception as why:
+            log.error('Cannot read %s value: %s', self.key(name), why)
+            callers(log.error, offset=2, length=2)
```
phw: Why `callers` instead of `exc_info`?

zas: Because `callers` lists callers... `exc_info` was too verbose, and wasn't really giving the useful info we need in this case.

phw: Ok, tested it, this somewhat makes sense. My main concern would be that we start cluttering the code with two calls, one for `log.error` and one for `callers`. At least when we start to use this elsewhere we should consider integrating this into the existing `picard.log` module somehow. For this PR I would be ok with merging this as is, though.

But can we at least log the exception itself with `%r`? Otherwise only the exception message is shown, but actually I think the type of exception is more interesting in many cases.

Also why limit the length to 2?

picard/util/__init__.py
```diff
@@ -609,3 +609,11 @@ def find_best_match(candidates, no_match):
     else:
         result = no_match
     return BestMatch(similarity=result.similarity, result=result, num_results=len(sorted_results))
+
+
+def callers(func, offset=3, length=1):
```
phw: Why an offset of 3 and length of 1? `offset=2` like in the call above seems to make sense. And when calling callers without specifying `length` my expectation would be to get all callers (or at least more than one).

picard/config.py
zas: I removed `callers` change from this PR, it can be done afterwards.

PR #1177, opened 2019-04-03
Run isort over all files

phw:
# Summary

Run isort again over all files, fixes some minor issues that sneaked in.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


test/test_coverart_image.py
```diff
@@ -1,12 +1,12 @@
 #!/usr/bin/env python
 # coding: utf-8
 import os.path
-import unittest
-
 from test.picardtestcase import (
     PicardTestCase,
     create_fake_png,
 )
+import unittest
```
zas: I don't think this should be moved after `from test.picardtestcase import`, i think we should add test.* as a section, but something like (untested) :

```
known_test=test.*
sections=FUTURE,STDLIB,THIRDPARTY,QT,TEST,PICARD,LOCALFOLDER,RESOURCES,PICARD_UI
```

picard/ui/util.py
```diff
@@ -27,9 +27,7 @@
     IS_MACOS,
     IS_WIN,
 )
-from picard.util import (
-    find_existing_path,
-)
+from picard.util import find_existing_path
```
zas: I'm not fond of those back and forth changes. They will happen often, a solution would be to always wrap, or to use force single line, neither are great. Prolly have to live with it.

picard/util/__init__.py
```diff
@@ -39,7 +39,6 @@
     IS_WIN,
 )
 
-
```
zas: The number of whilelines is rather confusing, that's 2, but only for functions or classes. So not in this case.

phw: Not sure what you mean. Are you suggesting to always use two lines, even for cases like this? Can this be configured with isor5?

phw:
@zas what do you think of running `isort -c` as part of CI?

Locally I now also set it up as a pre-commit hook for me.

picard/util/__init__.py
zas: Not sure if desirable, and whether it's compatible with PEP8 auto fixers. But
https://github.com/timothycrosley/isort/wiki/isort-Settings

> lines_after_imports: Forces a certain number of lines after the imports and before the first line of functional code. By default this is 2 lines if the first line of code is a class or function. Otherwise it's 1.

zas:
> @zas what do you think of running `isort -c` as part of CI?
> 
> Locally I now also set it up as a pre-commit hook for me.

I'm ok for including in CI, it will help to maintain consistency.
Please document pre-commit hooks you're using it may help new devs joining the project.

test/test_coverart_image.py
phw: Done in dd4da2aa474892ca858163cb8ebc57a14fdc13ce, I like this better. So far we had no clear structure used for these files.

picard/ui/util.py
phw: I don't think it's so bad. It only happens if you change the imports, and it is a very clear change anyway. You can't fully prevent formatting changes when you touch code.

picard/util/__init__.py
phw: I added this in  dc0ec84401150d54a121ff69bd6d21de035d2e54, I think it makes sense. At least flake8 is happy with both variations.

phw:
@zas I documented the pre-commit hook in `CONTRUBUTING.md` and added the isort check to Travis CI.

.travis.yml
```diff
@@ -60,7 +60,7 @@ before_install:
 - if [[ "$TRAVIS_OS_NAME" == "osx" ]]; then bash scripts/package-osx.sh; fi
 script:
 - py.test -v --cov=picard --cov-report xml:coverage.xml
-- isort --recursive --check-only .
+- isort --recursive --check-only $(git ls-tree -r --name-only $(git rev-parse HEAD) | grep "\\.py$")
```
zas: The check is failing on https://travis-ci.org/metabrainz/picard/builds/515581198
It is supposed  to pass on this PR, but log shows:
> ERROR: /home/travis/build/metabrainz/picard/picard/resources.py Imports are incorrectly sorted.

This file (and few others) should be skipped according to .isort.cfg configuration, may be this command doesn't find `.isort.cfg`, `--settings-path` could help perhaps.

PR #1179, opened 2019-04-04
Plugin refactor: explode in smaller bits

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


zas:
Please ignore Codacy for now

zas:
@phw, @samj1912 : i really need your help on this, especially to list plugin "formats" we have

So far a plugin can be:
- a single file
- a directory
- a zipped file (?)
- a zipped directory
- a zipped packaged plugin (.picard.zip with a manifest.json inside)
- ???

Can we clarify each case, i want to create specific tests for all.
My goal is to make this code actually readable and testable.

zas:
I wonder if we couldn't simplify things a lot using : https://docs.python.org/3.5/library/pkgutil.html#module-pkgutil

samj1912:
I would also suggest looking into the new importlib module of python3.7 and its resources submodule which is a light weight alternative to pkgutil. We can also override `__dir__` and `__getattr__` on a module level now which leads to some really nice dynamic imports.

samj1912:
https://www.python.org/dev/peps/pep-0562/ and https://docs.python.org/3/library/importlib.html#importlib.abc.ResourceReader

PR #1185, opened 2019-04-08
Use PICARD_APP_ID constant instead of hardcoded strings

phw:
# Summary
Use `PICARD_APP_ID` constant instead of hardocding "org.muiscbrainz.Picard" everywhere.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -23,6 +23,8 @@ from setuptools import (
 from setuptools.dist import Distribution
 
 from picard import (
+    PICARD_APP_ID,
+    PICARD_DESKTOP_NAME,
     PICARD_VERSION,
     __version__,
 )
@@ -36,6 +38,8 @@ if sys.version_info < (3, 5):
     sys.exit("ERROR: You need Python 3.5 or higher to use Picard.")
 
 PACKAGE_NAME = "picard"
+APPDATA_FILE = PICARD_APP_ID + '.appdata.xml'
+APPDATA_FILE_TEMPLATE = APPDATA_FILE + '.in'
 
 ext_modules = [
     Extension('picard.util._astrcmp', sources=['picard/util/_astrcmp.c']),
@@ -360,11 +364,10 @@ class picard_build_appdata(Command):
         pass
 
     def run(self):
-        with tempfile.NamedTemporaryFile(prefix='org.musicbrainz.Picard.appdata.',
-                                         suffix='.xml') as tmp_file:
+        with tempfile.NamedTemporaryFile(suffix=APPDATA_FILE) as tmp_file:
             self.spawn([
                 'msgfmt', '--xml',
-                '--template=org.musicbrainz.Picard.appdata.xml.in',
+                '--template=%s' % APPDATA_FILE_TEMPLATE,
                 '-d', 'po/appstream',
                 '-o', tmp_file.name,
             ])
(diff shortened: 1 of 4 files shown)


setup.py
```diff
@@ -360,11 +363,11 @@ def finalize_options(self):
         pass
 
     def run(self):
-        with tempfile.NamedTemporaryFile(prefix='org.musicbrainz.Picard.appdata.',
+        with tempfile.NamedTemporaryFile(prefix='%s.appdata.' % PICARD_APP_ID,
```
zas: `with tempfile.NamedTemporaryFile(suffix=APPDATA_FILE)` might work

setup.py
```diff
@@ -729,17 +736,17 @@ def find_file_in_path(filename):
...
+    )
+    for size in (16, 24, 32, 48, 128, 256)
 ]
 
+args['data_files'].append(('share/icons/hicolor/scalable/apps', ['resources/img-src/%s.svg' % PICARD_APP_ID]))
+args['data_files'].append(('share/applications', [PICARD_DESKTOP_NAME]))
+
 if IS_LINUX:
-    args['data_files'].append(('share/metainfo', ['org.musicbrainz.Picard.appdata.xml']))
+    args['data_files'].append(('share/metainfo', APPDATA_FILE))
```
zas: `APPDATA_FILE` is a string, `['org.musicbrainz.Picard.appdata.xml']` a list

setup.py
```diff
@@ -360,11 +363,11 @@ def finalize_options(self):
         pass
 
     def run(self):
-        with tempfile.NamedTemporaryFile(prefix='org.musicbrainz.Picard.appdata.',
+        with tempfile.NamedTemporaryFile(prefix='%s.appdata.' % PICARD_APP_ID,
                                          suffix='.xml') as tmp_file:
             self.spawn([
                 'msgfmt', '--xml',
-                '--template=org.musicbrainz.Picard.appdata.xml.in',
+                '--template=%s.in' % APPDATA_FILE,
```
zas: Since we have it here and below, it might make sense to have:

```python
PACKAGE_NAME = "picard"
APPDATA_FILE = PICARD_APP_ID + '.appdata.xml'
APPDATA_FILE_TEMPLATE = APPDATA_FILE + '.in'
```

And use:

```python
'--template=%s' % APPDATA_FILE_TEMPLATE,
```
And below:

```python
            '--language=appdata',
            APPDATA_FILE_TEMPLATE,
```

phw:
@zas I think I addressed your comments. I squashed the commits to keep this rather simple change simple in the commit history.

phw:
@zas in adec42868cc4c78f2aec975b3bb79128a337d7fb I updated the spec file to also use `PICARD_APP_ID` for the `CFBundleIdentifier` of macOS.

This might be something to discuss, because it changes the ID from `org.musicbrainz.picard` to `org.musicbrainz.Picard`. But it would provide us with a consistent ID across platforms. If we want this maybe best do this before we start e.g. to the app store.

But of course I could change it to lowercase the ID again for staying consistent with past releases.

zas:
> @zas in [adec428](https://github.com/metabrainz/picard/commit/adec42868cc4c78f2aec975b3bb79128a337d7fb) I updated the spec file to also use `PICARD_APP_ID` for the `CFBundleIdentifier` of macOS.
> 
> This might be something to discuss, because it changes the ID from `org.musicbrainz.picard` to `org.musicbrainz.Picard`. But it would provide us with a consistent ID across platforms. If we want this maybe best do this before we start e.g. deploy to the app store.
> 
> But of course I could change it to lowercase the ID again for staying consistent with past releases.

I'm all in for the consistency. What does it break ?

phw:
> I'm all in for the consistency. What does it break ?

I think currently nothing really. If we would deploy via app store we can't change the app ID without breaking updates (it would be a new application).

PR #1186, opened 2019-04-13
PICARD-1130: Add a post save hook

zas:
- register_file_post_save_processor()
- run_file_post_save_processors() is called in File._save_and_rename() just before returning

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1130](https://tickets.metabrainz.org/browse/PICARD-1130)

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -43,6 +43,10 @@ from picard.metadata import (
     Metadata,
     SimMatchTrack,
 )
+from picard.plugin import (
+    PluginFunctions,
+    PluginPriority,
+)
 from picard.util import (
     decode_filename,
     find_best_match,
@@ -330,6 +334,9 @@ class File(QtCore.QObject, Item):
             self.clear_pending()
             self._add_path_to_metadata(self.orig_metadata)
             self.metadata_images_changed.emit()
+
+            # run post save hook
+            run_file_post_save_processors(self)
 
         # Force update to ensure file status icon changes immediately after save
         self.update()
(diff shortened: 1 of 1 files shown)


picard/file.py
```diff
@@ -285,6 +289,7 @@ def _save_and_rename(self, old_filename, metadata):
         # Save cover art images
         if config.setting["save_images_to_files"]:
             self._save_images(os.path.dirname(new_filename), metadata)
+        run_file_post_save_processors(self, new_filename)
```
mineo: I'm not sure if this is the correct place for this - https://github.com/metabrainz/picard/blob/a1a79658bf58d150c927e7ff03b96c8e938d0ae6/picard/file.py#L213 calls `_saving_finished` *after* `_save_and_rename` and `_saving_finished` still does a lot of work in the non-error case, including updating `self.filename` and some metadata. If we want to merge this in its current form, the documentation needs to clearly list which updates are guaranteed to have happened at the point the post-save-processor runs and which are still not done. In particular the not-yet-updated filename of the `File` object in both the `filename` attribute and the metadata is super problematic for the automatic-submission-to-acousticbrainz usecase.

zas: The file is actually saved in `_save_and_rename`, but i agree it makes more sense to call the hook in `_saving_finished`, and just pass file object.

zas: I did that in 5e93e1529fb09e, does it look better ?

picard/file.py
```diff
@@ -768,3 +775,22 @@ def discnumber(self):
...
+
+
+_file_post_save_processors = PluginFunctions(label='file_post_save_processors')
+
+
+def register_file_post_save_processor(function, priority=PluginPriority.NORMAL):
+    """Registers file saved processor.
+
+    Args:
+        function: function to call after save, it will be passed file object and new filename
```
mineo: Only a file object.

PR #1187, opened 2019-04-19
Improve wrong number of arguments script error message

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -385,49 +385,29 @@ class ScriptParserTest(PicardTestCase):
         self.assertEqual(result, u'artist/title')
 
     def test_cmd_with_not_arguments(self):
-        try:
-            self.parser.eval("$noargstest()")
-        except ScriptError:
-            self.fail("Function noargs raised ScriptError unexpectedly.")
+        self.parser.eval("$noargstest()")
 
     def test_cmd_with_wrong_argcount_or(self):
         # $or() requires at least 2 arguments
-        with self.assertRaises(ScriptError):
+        areg = r"^Wrong number of arguments for \$or: Expected at least 2, "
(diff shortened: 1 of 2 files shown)


test/test_script.py
```diff
@@ -390,6 +390,45 @@ def test_cmd_with_not_arguments(self):
         except ScriptError:
             self.fail("Function noargs raised ScriptError unexpectedly.")
 
+    def test_cmd_with_wrong_argcount_or(self):
+        # $or() requires at least 2 arguments
+        with self.assertRaises(ScriptError):
+            self.parser.eval('$or(0)')
+        try:
+            self.parser.eval('$or(0,1)')
```
mineo: The positive tests are already covered by `test_cmd_or`, `test_cmd_eq`, … (if there's something missing in them, like testing with more than 2 arguments, it should be added there). I don't see any reason to repeat them here. That would also make this a test that only tests for the case of error, not for the case of success.

test/test_script.py
```diff
@@ -390,6 +390,45 @@ def test_cmd_with_not_arguments(self):
         except ScriptError:
             self.fail("Function noargs raised ScriptError unexpectedly.")
 
+    def test_cmd_with_wrong_argcount_or(self):
+        # $or() requires at least 2 arguments
+        with self.assertRaises(ScriptError):
```
mineo: Well, since you're changing the message, it would be great if you would use [assertRaisesRegEx](https://docs.python.org/3/library/unittest.html?highlight=unittest#unittest.TestCase.assertRaisesRegex) to verify the message as well.

PR #1188, opened 2019-04-19
LRUCache: inherit from MutableMapping + tests

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/util/lrucache.py
+++ b/picard/util/lrucache.py
@@ -17,8 +17,10 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from collections.abc import MutableMapping
 
-class LRUCache(dict):
+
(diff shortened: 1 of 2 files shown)


samj1912:
Do we even need this class anymore now that functools.lru_cache is available ?

IIRC this was added when picard was py2.

zas:
> Do we even need this class anymore now that functools.lru_cache is available ?

I thought about it, but it's a decorator to apply on a function, which is quite different from what it is done here (basically a modified dict). My goal here was to add tests while removing direct inheritance from dict, not sure if we could use functools.lru_cache instead, plus it is likely to be python too.

samj1912:
AFAICT from the usage we use a hash function to compute the key and store the pixmap. We could easily have a function that returns the pixmap and implement __hash__ for the objects the way the current key for the lru cache is used.

PR #1193, opened 2019-05-26
PICARD-1514: improve genres/tags filtering

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Option `ignore_genres` and associate UI are rather limited.

* JIRA ticket (_optional_): [PICARD-1514](https://tickets.metabrainz.org/browse/PICARD-1514)

# Solution

Replace `ignore_genres` with new option `genres_filter` which supports including/excluding tags using strict, wildcard or regex matching.

genres_filter_ui

# Action


picard/track.py
```diff
@@ -64,6 +66,52 @@
 }
 
 
+class TagGenreFilter:
+
+    def __init__(self, setting=None):
```
phw: I would suggest to change this to not pass the settings but instead `setting['genres_filter']`. This way this class could be easily re-used by other genre plugins with own filter list (or still the default one if preferred depending on the plugin).

picard/track.py
```diff
@@ -64,6 +66,52 @@
...
+                    # FIXME?: only support '*' (not '?' or '[abc]')
+                    # replace multiple '*' by one
+                    star = re.escape('*')
+                    remain = re.sub(star + '+', '*', remain)
+                    regex = '.*'.join([re.escape(x) for x in remain.split('*')])
+                    regex_search = re.compile('^' + regex + '$', re.IGNORECASE)
+                if regex_search:
+                    self.match_regexes[_list].append(regex_search)
+
+    def skip(self, tag):
```
phw: Why not provide a `filter` method which accepts a list and returns only valid genres? Again this is with using this class in plugins in mind, it would make a nicer interface.

picard/track.py
zas: Done in 1aee66ccff044f

picard/track.py
zas: Done in 1b08c55de545b7

PR #1194, opened 2019-05-28
PICARD-1488: Internal music player feature added

timur-enikeev:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* Internal music player has been added

# Problem
There was not internal music player in this application. It can be useful if user wants to know what is the track that he edits.

# Solution

I have added the music player feature based on QtMultimedia.

# Action
Please check on Windows (maybe I will check by myself later) and MacOS (I can't check by myself)

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -25,6 +25,7 @@ import os.path
 from PyQt5 import (
     QtCore,
     QtGui,
+    QtMultimedia,
     QtWidgets,
 )
 
@@ -104,6 +105,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         config.BoolOption("persist", "view_toolbar", True),
         config.BoolOption("persist", "view_file_browser", False),
         config.TextOption("persist", "current_directory", ""),
+        config.IntOption("persist", "mediaplayer_volume", "50"),
     ]
 
     def __init__(self, parent=None):
@@ -111,6 +113,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.selected_objects = []
         self.ignore_selection_changes = False
         self.toolbar = None
+        self.player = QtMultimedia.QMediaPlayer(self)
+        self.player.error.connect(self._on_player_error)
         self.setupUi()
 
     def setupUi(self):
@@ -198,6 +202,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         if config.setting["quit_confirmation"] and not self.show_quit_confirmation():
             event.ignore()
             return
+        if self.player.availability() != QtMultimedia.QMultimedia.ServiceMissing:
+            config.persist['mediaplayer_volume'] = self.player.volume()
         self.saveWindowState()
         event.accept()
 
@@ -564,6 +570,20 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             self.check_update_action.setMenuRole(QtWidgets.QAction.ApplicationSpecificRole)
             self.check_update_action.triggered.connect(self.do_update_check)
 
+        self.internal_play_action = QtWidgets.QAction(self.style().standardIcon(QtWidgets.QStyle.SP_MediaPlay), _("Play"), self)
+        self.internal_play_action.setStatusTip(_("Play selected files in an internal player"))
+        self.internal_play_action.setEnabled(False)
+        self.internal_play_action.triggered.connect(self.internal_play)
+
+        self.pause_action = QtWidgets.QAction(self.style().standardIcon(QtWidgets.QStyle.SP_MediaPause), _("Pause"), self)
+        self.pause_action.setToolTip(_("Pause/resume"))
+        self.pause_action.setStatusTip(_("Pause or resume playing with an internal player"))
+        self.pause_action.setCheckable(True)
+        self.pause_action.setChecked(False)
+        self.pause_action.setEnabled(False)
+        self.pause_action.triggered.connect(self.pause)
+        self.player.stateChanged.connect(self.pause_action.setEnabled)
+
     def toggle_rename_files(self, checked):
         config.setting["rename_files"] = checked
 
@@ -614,6 +634,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         menu.addSeparator()
         menu.addAction(self.show_toolbar_action)
         menu.addAction(self.search_toolbar_toggle_action)
+        menu.addAction(self.player_toolbar_toggle_action)
         menu = self.menuBar().addMenu(_("&Options"))
         menu.addAction(self.enable_renaming_action)
         menu.addAction(self.enable_moving_action)
@@ -652,11 +673,16 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
     def update_toolbar_style(self):
         if config.setting["toolbar_show_labels"]:
             self.toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonTextUnderIcon)
+            self.player_toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonTextUnderIcon)
+            self.volume_label.show()
         else:
             self.toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonIconOnly)
+            self.player_toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonIconOnly)
+            self.volume_label.hide()
 
     def create_toolbar(self):
         self.create_search_toolbar()
+        self.create_player_toolbar()
         self.create_action_toolbar()
 
     def create_action_toolbar(self):
@@ -689,6 +715,33 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
                 except AttributeError:
                     log.warning('Warning: Unknown action name "%r" found in config. Ignored.', action)
         self.show_toolbar()
+
+    def create_player_toolbar(self):
+        """"Create a toolbar with internal player control elements"""
+        self.player_toolbar = toolbar = QtWidgets.QToolBar(_("Player"))
+        toolbar.setObjectName("player_toolbar")
+        self.insertToolBar(self.search_toolbar, self.player_toolbar)
+        self.player_toolbar_toggle_action = self.player_toolbar.toggleViewAction()
+        def add_toolbar_action(action):
+            toolbar.addAction(action)
+            widget = toolbar.widgetForAction(action)
+            widget.setFocusPolicy(QtCore.Qt.TabFocus)
+            widget.setAttribute(QtCore.Qt.WA_MacShowFocusRect)
+        add_toolbar_action(self.internal_play_action)
+        add_toolbar_action(self.pause_action)
+        self.volume_slider = QtWidgets.QSlider(toolbar)
+        self.volume_slider.setOrientation(QtCore.Qt.Horizontal)
+        self.volume_slider.setMinimum(0)
+        self.volume_slider.setMaximum(100)
+        self.volume_slider.valueChanged.connect(self.set_volume)
+        self.volume_slider.setValue(self.get_logarithmic_volume(int(config.persist["mediaplayer_volume"])))
+        self.volume_label = QtWidgets.QLabel(_("Volume"), toolbar)
+        self.volume_widget = QtWidgets.QWidget(toolbar)
+        vbox = QtWidgets.QVBoxLayout(self.volume_widget)
+        vbox.addWidget(self.volume_slider)
+        vbox.addWidget(self.volume_label)
+        toolbar.addWidget(self.volume_widget)
+        toolbar.hide() #Hide by default
 
     def create_search_toolbar(self):
         self.search_toolbar = toolbar = self.addToolBar(_("Search"))
@@ -900,6 +953,42 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         for file in files:
             QtGui.QDesktopServices.openUrl(self._openUrl(file.filename))
 
+    def internal_play(self):
+        """Play selected tracks with an internal player"""
+        self.player.stop()
+        playlist = QtMultimedia.QMediaPlaylist(self)
+        playlist.setPlaybackMode(QtMultimedia.QMediaPlaylist.Sequential)
+        playlist.addMedia([QtMultimedia.QMediaContent(self._openUrl(file.filename)) \
+                                                      for file in self.tagger.get_files_from_objects(self.selected_objects)])
+        self.player.setPlaylist(playlist)
+        self.player.play()
+        self.pause_action.setChecked(False)
+
+    def _on_player_error(self, error):
+        self.set_statusbar_message(self.player.errorString(), echo=log.warning)
+
+    def pause(self, is_paused):
+        """Toggle pause of an internal player"""
+        if is_paused:
+            self.player.pause()
+        else:
+            self.player.play()
+
+    def set_volume(self, slider_value):
+        """Convert to linear scale and set"""
+        linear_volume = QtMultimedia.QAudio.convertVolume(slider_value/100.,
+                                                          QtMultimedia.QAudio.LogarithmicVolumeScale,
+                                                          QtMultimedia.QAudio.LinearVolumeScale)
+        self.player.setVolume(QtCore.qRound(linear_volume * 100))
+
+    @staticmethod
+    def get_logarithmic_volume(player_value):
+        """Return logarithmic scale volume to set slider position"""
+        logarithmic_volume = QtMultimedia.QAudio.convertVolume(player_value/100.,
+                                                               QtMultimedia.QAudio.LinearVolumeScale,
+                                                               QtMultimedia.QAudio.LogarithmicVolumeScale)
+        return QtCore.qRound(logarithmic_volume * 100)
+
     def open_folder(self):
         files = self.tagger.get_files_from_objects(self.selected_objects)
         folders = set([os.path.dirname(f.filename) for f in files])
@@ -990,6 +1079,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.autotag_action.setEnabled(can_autotag)
         self.browser_lookup_action.setEnabled(can_browser_lookup)
         self.play_file_action.setEnabled(have_files)
+        self.internal_play_action.setEnabled(have_files)
         self.open_folder_action.setEnabled(have_files)
         self.cut_action.setEnabled(have_objects)
         files = self.get_selected_or_unmatched_files()


timur-enikeev:
@phw 
> Once I clicked "Play" there was no real indication that playback has started and that it can be stopped by clicking on "Play" again. I think the Play button should turn into a Stop button

We can discuss the interface more detailed. Now if you select "play" while playing, the player will play other file, that is selected now. And my suggestion is not to change the icon but use a checkable action for playing (like for pause).

>I missed an indicator for the playback position.
>Probably optional, but some indication which file is currently playing would be nice.

I wanted to do it in my new pull request. I can start to do it now. It can be in this or following pull request.

>What file formats can we play with this? 

All formats that are supported by multimedia framework that is used by QtMultimedia. I will check behaviour for unsupported formats.

And one more question. Now the behaviour when many files are selected is "play all selected files consecutively". I can change the behaviour to "play the first the of selected files".

phw:
When I just tried again I could not get any playback happen. Clicking on play did nothing, clicking it a second time froze the UI for a while. Same files as before.

> We can discuss the interface more detailed. Now if you select "play" while playing, the player will play other file, that is selected now. And my suggestion is not to change the icon but use a checkable action for playing (like for pause).

That's true, I agree that the button should stay available to start playing another file. Turns out having an intuitive for a simple player isn't that easy after all :)

Maybe having a clear playback indication where you can see playback progress and file being played would solve the issue. The main issue for me as a user was that I wasn't sure if playback already started or not, and I hit the button a second time. In the end it was just the song starting with silence, but it still was confusing.

> And one more question. Now the behaviour when many files are selected is "play all selected files consecutively". I can change the behaviour to "play the first the of selected files".

I think the current implementation makes sense. Other opinions?

Sophist-UK:
1. I assume that we are calling QT to play the file on the main thread. As far as I can tell this is necessary, but there are mentions of QSound causing blocking of the main thread (and  thus the UI), so it seems that this might be an unavoidable side-effect.

2. It would be nice to have options for playback speed and volume. If I am simply listening to a track to help with tagging - as opposed to having a full blown media player built in - then I might want to play back at (say) 1.2x normal, and have it play at 50% of normal volume.

picard/ui/mainwindow.py
```diff
@@ -652,11 +670,17 @@ def create_menus(self):
     def update_toolbar_style(self):
         if config.setting["toolbar_show_labels"]:
             self.toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonTextUnderIcon)
+            self.player_toolbar.setToolButtonStyle(QtCore.Qt.ToolButtonTextUnderIcon)
+            self.volume_label.show()
+
```
phw: This empty line makes Codacy unhappy :)

timur-enikeev: How to send new fixed version?

timur-enikeev: @phw

phw: You can do any changes to your local branch, commit them and then push them back to this branch on Github. This will update the pull request automatically and trigger the test builds and Codacy check again.

timur-enikeev: @phw 
I have fixed the problem. Can we merge it now?

picard/ui/mainwindow.py
```diff
@@ -104,13 +105,15 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         config.BoolOption("persist", "view_toolbar", True),
         config.BoolOption("persist", "view_file_browser", False),
         config.TextOption("persist", "current_directory", ""),
+        config.IntOption("persist", "volume", "50"),
```
zas: I'd make this option more explicit: "mediaplayer_volume"
Reason: if we add more options related to the mediaplayer, we can "group" them using this common prefix.

Do not store the value of the slider position, but the value of the volume (both are the same for now, so the issue doesn't clearly appear).
If we change the way the volume is controlled we don't want to change the actual volume set.

Instead convert to/from to adjust UI control position, and read https://doc.qt.io/qt-5/qaudio.html#convertVolume

picard/ui/mainwindow.py
```diff
@@ -900,6 +949,23 @@ def play_file(self):
         for file in files:
             QtGui.QDesktopServices.openUrl(self._openUrl(file.filename))
 
+    def internal_play(self):
+        """Play selected tracks with an internal player"""
+        self.player.stop()
+        playlist = QtMultimedia.QMediaPlaylist(self)
+        playlist.setPlaybackMode(QtMultimedia.QMediaPlaylist.Sequential)
+        playlist.addMedia([QtMultimedia.QMediaContent(self._openUrl(file.filename)) for file in self.tagger.get_files_from_objects(self.selected_objects)])
```
zas: What does happen if the mediaplayer doesn't support the type of audio files ?

picard/ui/mainwindow.py
Sophist-UK: Can we have an option mediaplayer_speed, default 1.0, which allows you to slow down or speed up the playing speed as some people (like me) like to preview music at a faster (or slower) speed than normal?

picard/ui/mainwindow.py
Sophist-UK: As far as I can tell, the media formats supported depends on the O/S and on the codecs actually installed - and that there is no direct way to query QT to determine what formats are supported on the particular system.

It may well be possible to query the O/S directly to determine this, but adding extensive O/S specific code to achieve this doesn't sound like a great idea to me.

It may be possible to determine after the event that the audio didn't play by e.g. getting the QMediaPlayer Error signal with QMediaPlayer::FormatError passed as the parameter - and it would be possible once you had received this to maintain a list of supported or unsupported media types. Or indeed you could try to play the 1st milisecond of each type of media file you encounter to produce such a list - and you could have a list of default media types that each O/S supports to avoid having to do this for types that are known to be supported. Or you could embed a tiny silent example of each type of supported media in Picard and try to play each when Picard starts. **BUT** whatever way you try to achieve this, it seems to me it will be messy.

So personally not sure what the best approach is - perhaps needs some discussion.

picard/ui/mainwindow.py
timur-enikeev: @Sophist-UK I think it can be done later, in next pull requests. I want firstly add basic functional to the master branch.

picard/ui/mainwindow.py
phw: Currently it will just not play it. On my test there was an error logged to stderr, something like:

> Warning: "No decoder available for type 'application/x-ape'."

But actually this probably comes from gstreamer. I think some basic error handling should be added as part of this PR to log errors and show them in the status bar by connecting to `self.player.error.connect(self._on_player_error)` and having `_on_player_error` doing something like this:

```python
def _on_player_error(self, error):
    self.set_statusbar_message(self.player.errorString(), echo=log.warning)
```

timur-enikeev:
@phw @zas 
please review now.

zas:
> From my side this is good to merge for an initial version. It is cleanly implemented and I find it quite useful. But further improvements as discussed above should be implemented in follow-up PRs.
> 
> @zas What do you say?

- I'd like the second point i raised in https://github.com/metabrainz/picard/pull/1194#discussion_r288492053 to be addressed. Linear volume control (like current slider) doesn't work well, as it is explained at https://doc.qt.io/qt-5/qaudio.html#convertVolume

- it adds a dependency:

> Got keys from plugin meta data ("gstreamermediaplayer")
> QFactoryLoader::QFactoryLoader() checking directory path "/home/zas/.virtualenvs/picard3/bin/mediaservice" ...
> Cannot load library /home/zas/.virtualenvs/picard3/lib/python3.6/site-packages/PyQt5/Qt/plugins/mediaservice/libgstmediaplayer.so: (libQt5MultimediaGstTools.so.5: cannot open shared object file: No such file or directory)
> QLibraryPrivate::loadPlugin failed on "/home/zas/.virtualenvs/picard3/lib/python3.6/site-packages/PyQt5/Qt/plugins/mediaservice/libgstmediaplayer.so" : "Cannot load library /home/zas/.virtualenvs/picard3/lib/python3.6/site-packages/PyQt5/Qt/plugins/mediaservice/libgstmediaplayer.so: (libQt5MultimediaGstTools.so.5: cannot open shared object file: No such file or directory)"
> defaultServiceProvider::requestService(): no service found for - "org.qt-project.qt.mediaplayer"

On Ubuntu-like, that's `libqt5multimedia5-plugins` package, not sure if it can even work with pip-only install, and not sure about other platforms.
See also https://forums.linuxmint.com/viewtopic.php?f=47&t=280944

- I think we have too many buttons in toolbar already, and this player UI cannot be configured through Options > User Interface > Customize Action Toolbar.
- I'm pretty sure we'll have endless complaints about unsupported file formats, because we have no guarantee an audio format supported by Picard (through mutagen) will be supported by QMediaPlayer and its backends.

- this feature should be optional since it has extra dependencies that may not be fulfilled and may not interest everyone. In current state, buttons are shown even if there's no chance for the media player to work, since missing dependencies aren't detected by Picard. Pressing Play just displays "The QMediaPlayer object does not have a valid service" (additionnally, not translated). Not very user-friendly, and such missing dependency really have to be detected at start and the whole feature disabled. 

So, while i agree it can be useful, i'm not really sure it is a good idea, especially since we don't have people to do the "after sale" support.

timur-enikeev:
@zas 
what do you offer now? I think that linear volume control is not a problem now. In future we can fix it.

About unsupported formats:

Is there any problem with launching of the application or termination if there are some unresolved dependences?

I can make the player panel hidden by default and mark it as experimental.

timur-enikeev:
@zas 
If you want, I can do "after sale" support for users that have problems with the player on Linux and Windows.

zas:
> @zas
> what do you offer now? I think that linear volume control is not a problem now. In future we can fix it.
> 
> About unsupported formats:
> 
> Is there any problem with launching of the application or termination if there are some unresolved dependences?
> 
> I can make the player panel hidden by default and mark it as experimental.

The thing is that I'm reluctant to merge an experimental feature now, because we plan to do a stable release before summer, which means feature freeze soon.

Added dependency will need to be reviewed, not sure how it impacts packaging (or even if it does). 

If you can make this feature totally optional (`mediaplayer_enabled` in options), and improve it within next days according to our suggestions, we may include it.

About after sale support, i mean i can already imagine user reports about X format not being supported on Y platform (or player not working at all), and we can't do much about it since it depends on external stuff (Qt and its plugins), so that will be mostly time to research and answer to users.

@phw : did you test on macosx yet ? what do you think ?

timur-enikeev:
@zas 

> About after sale support, i mean i can already imagine user reports about X format not being supported on Y platform (or player not working at all), and we can't do much about it since it depends on external stuff (Qt and its plugins), so that will be mostly time to research and answer to users.

Qt on windows uses system multimedia framework, and, if Windows Media player plays this format, our player should play it too. Linux users are mostly experienced, and most of them should know what packages they should install.

Sophist-UK:
Yes - it is reasonable to assume that in the real world, the vast majority of users tagging particular formats of music files with Picard would have the appropriate codecs installed on their system to play those files.

But occasionally this will not be the case, and Picard should handle such situations gracefully - even if that means displaying an error message in the status bar.

phw:
>     * I'd like the second point i raised in [#1194 (comment)](https://github.com/metabrainz/picard/pull/1194#discussion_r288492053) to be addressed. Linear volume control (like current slider) doesn't work well, as it is explained at https://doc.qt.io/qt-5/qaudio.html#convertVolume

@timur-enikeev I agree with @zas here. Qt already provides all that is necessary and the change isn't really complex, it should be included here. 

>     * I'm pretty sure we'll have endless complaints about unsupported file formats, because we have no guarantee an audio format supported by Picard (through mutagen) will be supported by QMediaPlayer and its backends.

If we make a player feature dependent on if it supports **all** formats (and in return limit addition of new formats to whether they are supported by the player) we probably can't ever include playback functionality. I think we have to deal with the fact that there will be formats the player can't support. Just like e.g. fingerprinting also does not support all formats.

>     * I think we have too many buttons in toolbar already, and this player UI cannot be configured through Options > User Interface > Customize Action Toolbar.

The search toolbar also can't. I think customizing the player becomes more interesting as soon as we have more options, like e.g. progress view and playback speed. Then I could imagine we extend the toolbar customization to also allow the player to be customized. But I think having it as a separate toolbar that can be toggled is a good thing.

>     * since missing dependencies aren't detected by Picard. Pressing Play just displays "The QMediaPlayer object does not have a valid service" (additionnally, not translated). Not very user-friendly, and such missing dependency really have to be detected at start and the whole feature disabled.

@timur-enikeev Can you change this so the player is disabled until explicitly enabled by the user?

Agreed. Can we detect whether QMediaPlayer is available or not in advance? I think currently this will mostly affect GStreamer based platforms, but we have to ensure of course we ship all necessary Qt extensions on macOS and Windows as well (I can look into this part).

> If we can make this feature totally optional (`mediaplayer_enabled` in options), and improve it within next days according to our suggestions, we may include it.

There is already the "View > Player" checkbox, I think setting this as "disabled" by default would be sufficient. Not sure if we also need it in options, but we could of course turn this into a proper option and duplicate this option in Options > User Interface.

> @phw : did you test on macosx yet ? what do you think ?

Not yet, but will do. I think on macOS situation will be like on Windows were you won't have any problems with dependency, because it uses the existing infrastructure of the OS. But of course playback formats are also limited.

> But occasionally this will not be the case, and Picard should handle such situations gracefully - even if that means displaying an error message in the status bar.

The current patch already does this.

phw:
I looked a bit at how to make this completely optional:

1. There can be the case that `PyQt5.QtMultimedia` is not installed at all (on Arch Linux if you don't have the `qt5-multimedia` package installed. This currently will cause an exception on startup since the import fails. This `ImportError` should be handled and player functionality completely disabled if import fails.

2. There is a soft error if only the necessary platform plugins are missing. This can be handled on runtime by calling `self.player.availability()` (see https://doc.qt.io/qt-5/qmediaobject.html#availability). If this is not `QtMultimedia.QMultimedia.Available` the player should also be completely hidden and a warning written to the logs.

phw:
I tested on macOS, it works as expected. Both running from source with pip-installed PyQt5 and running the .dmg package works, we are not missing any dependencies. I haven't tested full audio support, but FLAC, AIFF, MP3 and (unsurprisingly) MP4 work, while WMA and APE don't.

phw:
If a format is supported can be found out using the [QMediaServiceSupportedFormatsInterface](https://doc-snapshots.qt.io/qt5-5.12/qmediaservicesupportedformatsinterface.html).

This is the second time a patch for supporting a built-in player is proposed, let's make it happen this time :) I really like the clean patch provided here as a start, and the minimal dependencies this needs is making this very attractive. What about the following plan:

1. Finalize this PR with:
  a) Use logarithmic scale for audio volume
  b) Disable the player by default
2. Do a follow-up PR to make the player be available only if dependencies are satisfied (as I outlined above)
3. Improve interface by providing playback speed and playback progress controls
4. Handle unsupported formats better (detect if format is supported, and e.g. disable the "Play" button, show a hint to the user). We need to put some more thought into this.

I think after step 2. we could theoretically ship it as an experimental feature. If you really want to be strict about it we could also hide it behind a build or command line flag.

@zas, @timur-enikeev What do you say?

samj1912:
Why not just make this a plugin?

phw:
> Why not just make this a plugin? Picard core is feature packed as is and it would be better if features like these are included as plugins instead.

In general I like this approach, but this needs extending the plugin system first to provide proper interfaces for an extension like this. There has been some talk about this, but nothing concrete yet. But with what we currently have it is not possible to implement this functionality as a plugin in any satisfying way.

zas:
> @zas, @timur-enikeev What do you say?

I agree with your plan.

timur-enikeev:
@phw 
Should I create a new pull request for checking dependencies?

zas:
> Should I create a new pull request for checking dependencies?

As described by @phw : 

> * Finalize this PR with:
>   a) Use logarithmic scale for audio volume
>   b) Disable the player by default

^^ First, in this PR

> 
> * Do a follow-up PR to make the player be available only if dependencies are satisfied (as I outlined above)
> 

^^ Another PR

timur-enikeev:
@zas @phw 
Should we store player's volume or volume slider's position? Storing slider's position is easier. If we want to store player's volume, we should write additional function to convert it to slider's position.

phw:
I think we should store player volume. It's what gets actually set, and having this value makes it easy to change the UI representation at any point, we don't rely on UI implementation details.

timur-enikeev:
@phw @zas 
please review now.

picard/ui/mainwindow.py
```diff
@@ -198,6 +202,7 @@ def closeEvent(self, event):
         if config.setting["quit_confirmation"] and not self.show_quit_confirmation():
             event.ignore()
             return
+        config.persist['mediaplayer_volume'] = self.player.volume()
```
zas: When player doesn't work (no service), this returns 0, may we should only save when player actually works.

picard/ui/mainwindow.py
```diff
@@ -104,13 +105,16 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
...
         config.TextOption("persist", "current_directory", ""),
+        config.IntOption("persist", "mediaplayer_volume", "50"),
     ]
 
     def __init__(self, parent=None):
         super().__init__(parent)
         self.selected_objects = []
         self.ignore_selection_changes = False
         self.toolbar = None
+        self.player = QtMultimedia.QMediaPlayer(self)
```
zas: IMHO, we should test availability using https://doc.qt.io/qt-5/qmediaplayer.html#availability and check if it is `QMultimedia.Available`.
If not available, all associated code should be skipped (and buttons disabled / hidden)

picard/ui/mainwindow.py
```diff
@@ -900,6 +951,37 @@ def play_file(self):
...
+        self.player.stop()
+        playlist = QtMultimedia.QMediaPlaylist(self)
+        playlist.setPlaybackMode(QtMultimedia.QMediaPlaylist.Sequential)
+        playlist.addMedia([QtMultimedia.QMediaContent(self._openUrl(file.filename)) for file in self.tagger.get_files_from_objects(self.selected_objects)])
+        self.player.setPlaylist(playlist)
+        self.player.play()
+        self.pause_action.setChecked(False)
+
+    def _on_player_error(self, error):
+        self.set_statusbar_message(self.player.errorString(), echo=log.warning)
```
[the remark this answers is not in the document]
timur-enikeev: @zas, it will be translated. Qt contains translations for the error string.

picard/ui/mainwindow.py
timur-enikeev: @zas, I thought that it should be in the next pull request (with checking if the QtMultimedia module available).

zas: Ok

picard/ui/mainwindow.py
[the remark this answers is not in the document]
zas: Hmmm then it is translated in debug log... no better.
Well, keep code as is, we'll advise later.

picard/ui/mainwindow.py
```diff
@@ -26,6 +26,7 @@
     QtCore,
     QtGui,
     QtWidgets,
+    QtMultimedia
```
zas: Please keep order and add a comma (read https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#coding-style about `isort`)

timur-enikeev:
@zas @phw please check now.

timur-enikeev:
@zas @phw 
I want to talk about checking of QtMultimedia module. Can we talk here?

My plan is to add

```
try:
    from PyQt5 import QtMultimedia
except:
    qt_multimedia_available = False
else:
    qt_multimedia_available = True
```

to the head of the file, and then check if qt_multimedia_available is true. Will it be OK?

phw:
@timur-enikeev Not sure if this is the best place to discuss, maybe better use the ticket instead of a closed PR?

Anyway, that's exactly how I would handle this as well. Only thing I would add is to log a info message that QtMultimedia is unavailable in the exception case.

phw:
@timur-enikeev Another way to discuss this would be to open a early pull request and mark it as work in progress. I usually do this if I want some early feedback on an implementation to have some place for discussion.

phw:
Ui changes with progress display and playback speed UI are in https://github.com/metabrainz/picard/pull/1231

PR #1198, opened 2019-05-30
PICARD-1506: Provide empty config for tests

phw:
# Summary
Currently Picard's test cases can be affected by existing user configuration and might fail.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1506](https://tickets.metabrainz.org/browse/PICARD-1506)

# Solution
For tests provide an empty config file.

I am not entirely sure what is happening and I could not find anything in the Qt docs at https://doc.qt.io/qt-5/qsettings.html#QSettings-3

But if there is no empty config file Qt seems to initialize the specified config file path with a copy of the application's config. I can easily reproduce what @Freso reported on PICARD-1506, and providing an empty file solves the issue for me.

# Action

### DIFF
--- a/test/data/test.ini
+++ b/test/data/test.ini
@@ -1,0 +1,4 @@
+[application]
+version=2.0.0
+
+[setting]
--- a/test/test_config.py
+++ b/test/test_config.py
@@ -41,6 +41,7 @@ class TestPicardConfigCommon(PicardTestCase):
         super().setUp()
         self.tmp_directory = mkdtemp()
         self.configpath = os.path.join(self.tmp_directory, 'test.ini')
+        shutil.copy(os.path.join('test', 'data', 'test.ini'), self.configpath)
         self.config = Config.from_file(None, self.configpath)
         self.config.application["version"] = "testing"
         logging.disable(logging.ERROR)


zas:
> But if there is no empty config file Qt seems to initialize the specified config file path with a copy of the application's config

I think the culprit code is https://github.com/metabrainz/picard/blob/e263c801178ea19eab60059598286b5c0dcf5df7/picard/config.py#L116-L122 as suggested by Mineo in https://tickets.metabrainz.org/browse/PICARD-1506?focusedCommentId=49499&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-49499

Can you reproduce the issue if this code is disabled ?

phw:
Yes, indeed, this code is responsible for the old config being copied. Also not having any old Picard.conf in `~/.config/MusicBrainz` file prevents the issue from happening.

Reading mineo's comments this only fixes part of the entire issue. I still think we should provide a clean base config file for tests, though. This I think fixes point 1. mineo mentions (tests are reading files which they did not create).

Not sure about 

> 2. Building an Option object only registers it on Option.registry, but not in ConfigSection.__qt_config, leading to an inconsistent view of the available options.

and what side effects fixing this would have. But we should probably tackle separate things separately here.

PR #1202, opened 2019-06-17
PICARD-1529: Avoid loaded NATs getting assigned cover art from other NATs

phw:
# Summary

When loading already tagged NAT files into Picard which contain embedded cover art, the first loaded file loads normally, but every additional loaded file gets the cover art of all previously loaded files attached.

See discussion at https://community.metabrainz.org/t/album-modified-and-complete-on-every-load-but-no-modifications/432044

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1529](https://tickets.metabrainz.org/browse/PICARD-1529)

# Solution
Preserve the file images instead of applying images from `NatAlbum`.

A few considerations:

1. I thought about removing `self.metadata.copy(self.album.metadata)` completely, as by default this does not provide any additional metadata. But having the ability to set tags for all NATs on the `NatAlbum` level has some use cases for people who want specific tags for all NATs.

2. I also considered removing display and handling of cover art on the `NatAlbum`, but I think having the `NatAlbum` display the different cover art has some value in the UI as you can easily see whether your NATs have cover art or not.
 

# Action

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -257,27 +257,14 @@ class Metadata(MutableMapping):
                 result = SimMatchTrack(similarity=sim, releasegroup=rg, release=release, track=track)
         return result
 
-    def copy(self, other):
+    def copy(self, other, copy_images=True):
         self.clear()
-        self.update(other)
+        self._update_from_metadata(other, copy_images)
 
     def update(self, *args, **kwargs):
         one_arg = len(args) == 1
         if one_arg and isinstance(args[0], self.__class__):
-            # update from Metadata object
-            other = args[0]
-
-            for k, v in other.rawitems():
-                self.set(k, v[:])
-
-            for tag in other.deleted_tags:
-                del self[tag]
-
-            if other.images:
-                self.images = other.images.copy()
-            if other.length:
-                self.length = other.length
-
+            self._update_from_metadata(args[0])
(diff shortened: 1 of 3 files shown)


phw:
I ended up adding an optional parameter `copy_images` to `Metadata.copy`. I think this makes the usage pretty clear. Please review.

phw:
> Codacy is complaining about code duplication in tests, please make it happy before merging.

I must say I am not a big fan of codacy's code duplication complaints in tests. A certain amount of duplication often makes tests more readable, as each tests stands on it's own. IMHO too much abstraction can be harmful for test readability. I would actually consider disabling or loosening the duplication checks for test files if possible.

Anyway, I simplified the comparison of the metadata here and that makes both Codacy and me happy :) In case someone is wondering, I intentionally compare the `_store` attributes here:

```python
self.assertEqual(self.metadata._store, m._store)
```

Currently the following would have the same effect:

```python
self.assertEqual(self.metadata, m)
```

`Metadata.__eq__` is inherited from `MutableMapping`, but we might actually consider implementing `Metadata.__eq__` to compare all attributes, not just the store dict. This would change the semantics of `__eq__`, though (and might affect some parts in the existing code that assume `metadata == {}` works).

PR #1203, opened 2019-06-19
QtMultimedia availability check added

timur-enikeev:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

After internal music player has been added, picard crashes if PyQt5.QtMultimedia module is not available. I have added availability check for QtMultimedia module.

# Solution

I have added handling of ImportError exception when we try to import PyQt5.QtMultimedia. If the import fails, qt_multimedia_available variable will be False. If the import was successful, this variable will be True.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -25,7 +25,6 @@ import os.path
 from PyQt5 import (
     QtCore,
     QtGui,
-    QtMultimedia,
     QtWidgets,
 )
 
@@ -84,6 +83,15 @@ from picard.ui.util import (
 )
 
 
+try:
+    from PyQt5 import QtMultimedia
+except Exception as e:
+    qt_multimedia_available = False
+    log.warning("{} - Internal player disabled".format(e.msg))
+else:
+    qt_multimedia_available = True
+
+
 ui_init = ExtensionPoint(label='ui_init')
 
 
@@ -113,8 +121,9 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.selected_objects = []
         self.ignore_selection_changes = False
         self.toolbar = None
-        self.player = QtMultimedia.QMediaPlayer(self)
-        self.player.error.connect(self._on_player_error)
+        if qt_multimedia_available:
+            self.player = QtMultimedia.QMediaPlayer(self)
+            self.player.error.connect(self._on_player_error)
         self.setupUi()
 
     def setupUi(self):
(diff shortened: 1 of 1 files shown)


timur-enikeev:
@samj1912 
> This should be an except ImportError instead of a broad except.

But what if there will be another exception type? I don't know entrails of pyqt. Maybe it can raise another exception type.

picard/ui/mainwindow.py
```diff
@@ -84,6 +83,15 @@
 )
 
 
+try:
+    from PyQt5 import QtMultimedia
+except Exception as e:
+    qt_multimedia_available = False
+    log.warning("{} - Internal player disabled".format(e.msg))
```
zas: No need to pre-format with `log.warning()` and we usually place the detailed reason after the main message.
Also `import` will raise [ImportError](https://docs.python.org/3/library/exceptions.html#ImportError)

```python
try:
    from PyQt5 import QtMultimedia
except ImportError as e:
    qt_multimedia_available = False
    log.warning('Internal player disabled: %s', e.msg)
else:
    qt_multimedia_available = True
```

picard/ui/mainwindow.py
```diff
@@ -113,8 +121,9 @@ def __init__(self, parent=None):
         self.selected_objects = []
         self.ignore_selection_changes = False
         self.toolbar = None
-        self.player = QtMultimedia.QMediaPlayer(self)
-        self.player.error.connect(self._on_player_error)
+        if qt_multimedia_available:
```
zas: We could also try to import directly here and only set player property if a player is actually available.

```python
self.player = None
try:
    from PyQt5 import QtMultimedia
    player = QtMultimedia.QMediaPlayer(self)
    if player.availability() != QtMultimedia.QMultimedia.ServiceMissing:
        self.player = player
        self.player.error.connect(self._on_player_error)
except ImportError as e:
    log.warning('Internal player disabled: %s', e.msg)
```

It would simplify tests:

```python
if self.player:
   do_something_with(self.player)
```

And ensure an exception is raised if `self.player` is used without being really available (like using `self.player.volume()`).

@timur-enikeev @phw : what do you think ?

zas:
> But what if there will be another exception type? I don't know entrails of pyqt. Maybe it can raise another exception type.

If another exception is raised, then it shouldn't be converted to a warning, because it'd be unexpected.
@samj1912 is right here, we should only catch `ImportError`.

But testing lack of exception on module import isn't sufficient to ensure the player is available, see  https://github.com/metabrainz/picard/pull/1203#discussion_r295741302

We have 3 cases:
- module cannot be imported at all
- module can be imported but player isn't available because of missing dependencies (typically gstreamer on linux)
- module can be imported and player is available

picard/ui/mainwindow.py
phw: I am not a big fan of importing inline, I think this is clearer if done at the top. Apart from that I agree, so I would change your example to:

```python
self.player = None
player = QtMultimedia.QMediaPlayer(self)
if qt_multimedia_available and player.availability() != QtMultimedia.QMultimedia.ServiceMissing:
    self.player = player
    self.player.error.connect(self._on_player_error)
```

zas: I implemented this, with extra error logging in https://github.com/metabrainz/picard/pull/1207

zas:
Closing this pull request as https://github.com/metabrainz/picard/pull/1207 includes it.

PR #1211, opened 2019-07-06
Add test *.aac files to MANIFEST.in 

kmosiejczuk:
Add test *.aac files to MANIFEST.in so that source tarball for PyPI has all the data files for the tests.

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

A subset of the regression tests fail when I run them on OpenBSD. Most of them fail due to the test-apev2.aac is missing from the test/data/ directory from the PyPI tarball. Looks like aac files were forgotten from the MANIFEST.in file. Adding them puts them in the tarball generated by python3 setup.py sdist.

# Solution

This simple change to MANIFEST.in should include those files for future source tarballs.

# Action

No other actions.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,6 +5,7 @@ recursive-include scripts *.in
 graft ui
 recursive-include resources *.py
 
+recursive-include test *.aac
 recursive-include test *.aiff
 recursive-include test *.ape
 recursive-include test *.asf


phw:
This was still missing some files, completely added in https://github.com/metabrainz/picard/pull/1282

PR #1212, opened 2019-07-07
PICARD-1549: Fixed pip install on Windows

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
As reported https://stackoverflow.com/questions/56737424/how-to-fix-error-option-single-version-externally-managed-not-recognized/56740829 trying to install Picard via pip on Windows currently fails.

There are actually multiple independent issues:

1. `setup.py` install tries to generate the installer config from `installer/picard-setup.nsi.in`, but it's not included in the source
2. `setup.py` (indirectly) depends on PyQt5
3. Our `setup.py` does not support `--single-version-externally-managed`, this is already fixed in #1210

* JIRA ticket (_optional_): [PICARD-1549](https://tickets.metabrainz.org/browse/PICARD-1549)

# Solution

1. Include the installer files in the source distribution. I think that makes sense to allow the full functionality Picard's setup.py provides, including building the Windows installer if wanted. @mineo What is your opinion on that?

2. Remove the dependency on PyQt5 from `setup.py`. This was indirectly by importing from `picard.const.sys`, because `picard.const` depends on PyQt5.

@zas What is your opinion on that second part? IMHO `picard.const` should not depend on PyQt5, but this is difficult to change now. I tried moving the `USER_DIR` and `USER_PLUGIN_DIR` constants elsewhere, but this could break plugins (at least classical extras does import `USER_DIR`). In the end I figured those simple `IS_WIN` / `IS_LINUX` constants as a simplification of a simple `sys.platform` comparison are not worse the trouble and there is more benefit in having `setup.py` not rely on Picard's internal abstractions like this.

With the above changes I was able to run a successful `pip install .` from Picard source on Windows.

# Action

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -38,3 +38,10 @@ include *.md
 include *.txt
 
 recursive-include resources/images *.png
+
+recursive-include installer *.in
+recursive-include installer *.ini
+recursive-include installer/images *.bmp
+recursive-include installer/images *.ico
+recursive-include installer/images *.svg
+recursive-include installer/languages *.nsh
--- a/setup.py
+++ b/setup.py
@@ -28,10 +28,6 @@ from picard import (
     PICARD_VERSION,
     __version__,
 )
-from picard.const.sys import (
-    IS_LINUX,
-    IS_WIN,
-)
 
 
 if sys.version_info < (3, 5):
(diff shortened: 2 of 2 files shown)


phw:
One thing I noticed: There is some risk that we accidentally introduce a dependency on an external library, e.g. PyQt5, in `picard/__init__py`. This would break clean pip source installs, since it would make `setup.py` get this dependency again. Once we have introduced such a dependency it could be hard to remove again without breaking something in a backward incompatible way.

We should probably add clean `pip install .` runs to our CI for all build platforms (at least Linux + Windows).

zas:
>     2\. This was indirectly by importing from `picard.const.sys`, because `picard.const` depends on PyQt5.

Your solution is fine, I fear such indirect dependency may reappear in the future though, can we test against this somehow?

I'd add a comment about this in `setup.py` code at least.

phw:
> I fear such indirect dependency may reappear in the future though, can we test against this somehow?

That's also my biggest fear here. My proposal would be to run a CI build for a clean pip source install. That way we would notice if we introduce some build breaking dependencies. I can look into how to add this to travis and appveyor builds.

PR #1216, opened 2019-07-10
Build with PyQt 5.13

phw:
# Summary
For now this is just to test builds with latest Qt versions on  macOS and Windows. Specifically I want to test if this makes any difference performance wise on macOS, see [PICARD-1554](https://tickets.metabrainz.org/browse/PICARD-1554)

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1554](https://tickets.metabrainz.org/browse/PICARD-1554)
* JIRA ticket (_optional_): [PICARD-1527](https://tickets.metabrainz.org/browse/PICARD-1527)
* JIRA ticket (_optional_): [PICARD-1282](https://tickets.metabrainz.org/browse/PICARD-1282)

# Solution

# Action

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -27,7 +27,7 @@ env:
 matrix:
   include:
   - os: osx
-    osx_image: xcode6.4
+    osx_image: xcode8.3
     language: generic
     env:
     - DISCID_VERSION=0.6.2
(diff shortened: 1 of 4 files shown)


phw:
Needs https://github.com/metabrainz/picard/pull/1222 to avoid failing builds.

phw:
This should now build. I won't be able to access my mac the next two weeks (or do much anything on Picard btw.), but I will look into this again once I'm back and see whether I can see some performance changes.

Also I will try to find out what versions of macOS we can support with which build. On that it depends which build we should use for release I think.

phw:
I tested the builds, and I think we should go forward with this. I could not totally confirm it fixes the performance issues reported on [PICARD-1554](https://tickets.metabrainz.org/browse/PICARD-1554), but it fixes issues with resizing the options window reported at [PICARD-1527](https://tickets.metabrainz.org/browse/PICARD-1527).

One thing to keep in mind is that this raises the minimum macOS version to 10.12 (macOS Sierra released in 2016). Previously we supported 10.10 (Yosemite releases in 2014). This is the minimum version supported by Qt5 5.13. But I think it is reasonable that we have to limit support for older versions at some point. We can keep 2.1.3 around for macOS 10.10 and 10.11.

We probably need to redesign https://picard.musicbrainz.org/downloads/ a bit, though. It becomes a bit troublesome to browse with the different macOS versions we provide. Maybe just list one macOS box with latest release and an expandable list and a link to open it ("Older macOS versions")

phw:
Another issue fixed by this is [PICARD-1282](https://tickets.metabrainz.org/browse/PICARD-1282)

phw:
Looks like some macOS code signing issue remains, the command for code signing seems to hang (not sure, maybe it is waiting for user input or something). I vaguely remember having read about this issue somewhere, will look into it later.

phw:
This also resolves [PICARD-1353](https://tickets.metabrainz.org/browse/PICARD-1353)

PR #1222, opened 2019-07-22
PICARD-1561: On macOS 10.14 test preserve time only with second accuracy

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
On macOS 10.14 setting the file times with `os.utime` only sets full seconds. This makes our tests fail that test for subsecond accuracy.

* JIRA ticket (_optional_): [PICARD-1561](https://tickets.metabrainz.org/browse/PICARD-1561)

# Solution
Round values to full seconds before comparing. Only do this on macOS 10.14, earlier versions are not affected and later versions we don't know.

I don't really like this check here, but currently don't see a better way to handle this. I currently can't tell whether this is a bug or feature of macOS 10.14, maybe we can get rid of this again at some point.

# Action

### DIFF
--- a/test/test_file.py
+++ b/test/test_file.py
@@ -4,7 +4,15 @@ from tempfile import mkdtemp
 
 from test.picardtestcase import PicardTestCase
 
+from picard.const.sys import IS_MACOS
 from picard.file import File
+
+
+def is_macos_10_14():
+    if IS_MACOS:
+        import platform
+        return platform.mac_ver()[0].startswith('10.14')
+    return False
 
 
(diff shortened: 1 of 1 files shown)


phw:
@zas Regarding your comment of using `AlmostEqual`, I intentionally did not. This is not the case where a value is inaccurate and maybe fluctuates a bit. The values are always correct, the accuracy is just limited to full seconds. So the code falls now back to only comparing seconds for this specific macOS version.

PR #1227, opened 2019-08-05
PICARD-1559: Create directories from script when moving files without renaming

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When moving files without renaming currently the renaming script is not used but files are moved to `move_files_to` only. This is probably unexpected, since the other way round, renaming without moving, uses the script for the filename part only. Hence it makes sense to have move files also use the directory part of the script.

* JIRA ticket (_optional_): [PICARD-1559](https://tickets.metabrainz.org/browse/PICARD-1559)

# Solution
When only move_files without rename_files is active, consider the renaming script but only use the directory part and keep the original filename.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -374,7 +374,6 @@ class File(QtCore.QObject, Item):
         return new_filename, ext
 
     def _format_filename(self, new_dirname, new_filename, metadata, settings):
-        # TODO: tests !!
         old_filename = new_filename
         new_filename, ext = self._fixed_splitext(new_filename)
         ext = ext.lower()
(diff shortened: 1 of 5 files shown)


phw:
@zas Not sure about how to "warn the user" about these changes (as you suggested on Jira). Do you think having this in the changelog and release announcements is enough?

zas:
> Do you think having this in the changelog and release announcements is enough?

Perhaps, as "breaking change" ;)

phw:
Fixed the tests on windows

PR #1231, opened 2019-08-12
PICARD-1488: Improved media player controls

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

As discussed in https://github.com/metabrainz/picard/pull/1194 the media player toolbar should get additional functionality such as playback speed control and playback progress.

# Solution

This improves the media player by adding / changing the following functionality:

1. When opened the first time the toolbar is now positioned at the bottom of the main window. Previously it got squeezed to the top together with the main and the search toolbar, causing many buttons to get hidden in all toolbars. I think having this at the bottom by default makes this more user friendly, but the user can of course move the toolbar to the top if they choose (or let it float).

    For now placement at the side is not possible, I have intentionally disabled this. I think for now the UI does not work too well with a vertical toolbar, and I did not want to go into implementing the changes necessary for this here. I have some ideas how it should look like, but I would like to postpone this.

2. This ads a progress display as a slider together with current and total playing time. It also shows the filename of the current playback below.

3. The volume control has been changed to a popover style volume selection. It displays the volume as percentage, let's you change volume with mouse wheel and if you click on it you get a volume slider.

4. There is a playback speed control. Currently it allows you to adjust playback speed between 0.5 and 1.5 of the original speed in 0.1 steps. The UI works pretty similar to the volume control, including the popover which shows a slider and the ability to change playback speed with the mouse wheel without opening the popover.

5. I have added custom icons from flaticons.com which somewhat fit the look and feel if existing icons. If you find something more suitable let me know.

Here is a screenshot:

grafik

And a video showing it in action:

https://keybase.pub/phwolfer/picard-player-toolbar.webm

# Action

@Sophist-UK: I would appreciate your feedback very much, especially on the playback speed thing. This was added just for you :D

@timur-enikeev: You did the first iteration, what do you think about the changes?


picard/ui/playertoolbar.py
```diff
@@ -99,6 +99,8 @@ def create_toolbar(self):
 
     def set_objects(self, objects):
         self._selected_objects = objects
+        player_enabled = len(objects) > 0
```
zas: Not strictly needed.

```python
self._toolbar.play_action.setEnabled(self._selected_objects)
```

picard/ui/playertoolbar.py
```diff
@@ -357,38 +346,50 @@ def event(self, event):
...
+        super().showEvent(event)
+        self.setMinimumWidth(self.width())
+        self.set_volume(self.volume)
+
+    def event(self, event):
+        if event.type() == QtCore.QEvent.Wheel:
+            delta = event.angleDelta().y()
+            volume = self.volume
+            if delta > 0:
+                volume += 3
```
zas: I'd store 3 in a variable (step?).

picard/ui/playertoolbar.py
```diff
@@ -312,6 +321,9 @@ class PlaybackRateButton(QtWidgets.QToolButton):
 
     def __init__(self, parent, playback_rate):
         super().__init__(parent)
+        button_margin = self.style().pixelMetric(QtWidgets.QStyle.PM_ButtonMargin)
+        min_width = get_text_width(self.font(), _('%1.1f ×') % 8.8)
```
zas: I'd store '%1.1f ×' in a property:

```python
self.rate_fmt = N_('%1.1f ×')
```

```python
label  = _(self.rate_fmt) % rate
```

picard/ui/playertoolbar.py
```diff
@@ -370,7 +377,10 @@ def __init__(self, parent, volume):
         icon = self.style().standardIcon(QtWidgets.QStyle.SP_MediaVolume)
         self.setIcon(icon)
         self.set_volume(volume)
-        self.setText('100%')  # Only for initial min width calculation
+        margins = self.getContentsMargins()
+        button_margin = self.style().pixelMetric(QtWidgets.QStyle.PM_ButtonMargin)
+        min_width = get_text_width(self.font(), '888%')
```
zas: You should use same format on display and width calculation: _('%d%%') rather than hardcoded '888%'

phw:
I addressed the crash directly on master in https://github.com/metabrainz/picard/commit/f57104223ef449e64e9b7a750fd1db7911646f3a and rebased this branch on top of it.

picard/ui/playertoolbar.py
phw: Fixed as part of https://github.com/metabrainz/picard/pull/1231/commits/164c32ff0131de30f5a97634f1a97e5de157f46c

picard/ui/playertoolbar.py
phw: Fixed as part of https://github.com/metabrainz/picard/pull/1231/commits/164c32ff0131de30f5a97634f1a97e5de157f46c

picard/ui/playertoolbar.py
phw: Fixed as part of https://github.com/metabrainz/picard/pull/1231/commits/164c32ff0131de30f5a97634f1a97e5de157f46c . The 3 was chosen because it is the default `pageStep` for `QSlider`. I now made this explicit by setting the variable and also using this in `slider.setPageStep(...)`

picard/ui/playertoolbar.py
phw: Fixed as part of https://github.com/metabrainz/picard/pull/1231/commits/164c32ff0131de30f5a97634f1a97e5de157f46c

phw:
I fixed an issue where the displayed volume would change after restarting Picard. This was caused by rounding issues when converting between logarithmic and linear scale for the volume. I fixed this by now storing the logarithmic scale, so the value stored in config is the same as what is shown in UI. Linear scale thus becomes a purely internal implementation detail.

PR #1237, opened 2019-08-15
PICARD-1575: Fixes for cover art drag and drop from browser

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1575](https://tickets.metabrainz.org/browse/PICARD-1575)

# Solution
This addresses two issues with drag and dropping images from browser to the coverartbox

1. Sometimes servers don't respond with the proper image mime type but only send `application/octet-stream` or `binary/data` (probably in an attempt to trigger the download dialog in the browser). The code now tries to detect the mime type using our imageinfo class.

2. Drag and drop from Bing image search result page works. Same solution as for Google where the real image URL is inside the URL query parameters, this just uses a different parameter name.

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -399,6 +399,8 @@ class CoverArtBox(QtWidgets.QGroupBox):
     def on_remote_image_fetched(self, url, data, reply, error, fallback_data=None):
         data = bytes(data)
         mime = reply.header(QtNetwork.QNetworkRequest.ContentTypeHeader)
+        # Some sites return a mime type with encoding like "image/jpeg; charset=UTF-8"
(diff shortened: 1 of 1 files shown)


phw:
Found another issue: Some sites actually return mime type with charset even for images, like e.g. `image/jpeg; charset=UTF-8`. One example URL for such behavior is https://archive.org/download/Cruisin-AHistoryOfRockNRollRadio/Cruisin-AHistoryOfRockNRollRadio.jpg

Hence the code now only considers the first part of the mime type when checking for support

PR #1247, opened 2019-08-21
PICARD-1580: Fixed crash when closing options window

phw:
# Summary
Picard crashes when closing the option window via window controls (the close X in the window's title bar) while the Preferred Releases page is open.

If the options window gets closed via window controls regsitered signals are already disconnected before the hideEvent gets called.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1580](https://tickets.metabrainz.org/browse/PICARD-1580)

# Solution

# Action

### DIFF
--- a/picard/ui/options/releases.py
+++ b/picard/ui/options/releases.py
@@ -74,7 +74,10 @@ class TipSlider(QtWidgets.QSlider):
     def hideEvent(self, event):
         super().hideEvent(event)
         if not IS_WIN:
-            self.valueChanged.disconnect(self.show_tip)
+            try:
+                self.valueChanged.disconnect(self.show_tip)
+            except TypeError:
+                pass
 
     def show_tip(self, value):
         self.round_value(value)


picard/ui/options/releases.py
```diff
@@ -74,7 +74,10 @@ def showEvent(self, event):
     def hideEvent(self, event):
         super().hideEvent(event)
         if not IS_WIN:
-            self.valueChanged.disconnect(self.show_tip)
+            try:
+                self.valueChanged.disconnect(self.show_tip)
+            except TypeError:
+                pass
```
zas: You may want to use https://github.com/metabrainz/picard/blob/master/picard/util/__init__.py#L560
It is meant to handle those cases (plus reconnects).

phw: Actually this does not really fit here. We want only the disconnect here, not reconnecting. This is just to ensure `self.show_tip` does not get triggered while the view is hidden.

PR #1252, opened 2019-08-29
PICARD-1090: Sharp match indicator icons

phw:
# Summary
The match indicator icons in Picard 2 were a bit blurry.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1090](https://tickets.metabrainz.org/browse/PICARD-1090)

# Solution
This PR fixes this by aligning the SVGs on a clear 64x64 pixel raster, resulting in pixel aligned output in both 16x16 and 64x64 sizes.

Also the pending icons were redone based on the non-pending icons without the border (which wasn't visible in the 16x16 icons anyway, and blurry in the 64x64 icons).

# Action


zas:
> Lgtm. Should we mark svg files as binary files in gitattributes? Same for resources. The diff shouldn't be displayed?

You can display them, switching to rich diff.

image

I wonder about these differences though ^^

Default rendered size is much smaller in new version, and it seems to me the black border is missing.

If you look carefully at screenshots in https://tickets.metabrainz.org/browse/PICARD-1090, you'll notice 1.4.2 version has black border, which disappeared in recent versions.

1.4.2:


2.x:


This PR:
pr1252

phw:
> Default rendered size is much smaller in new version, and it seems to me the black border is missing.

The size is because Inkscape now explicitly set the sizes to what is configured as the document dimensions, which is now 64x64 pixel (previously 58x58 pixel).

Regarding the border: The newer colored versions of the icons @samj1912 added for 2.0 removed those borders, and I think this is a fine choice. It better fits with the rest of the icons. The icons also have a background, so the "fill status" should be clear without any extra borders.

But the gray icons still had borders, at least in the SVG. But in the 16px version it was not visible, and in the 64px version it was blurry at best and probably also not really visible on a retina type screen anyway. So I opted for removing the border also there and fill the box with a background instead. I find it visually more pleasing as well.

We could of course choose to re-add the border again. But on the 16px version it needs to be one pixel in width, which would be 4 pixel on the 64px version where probably a smaller border would be nicer. This would mean we will need to versions of the icons again. 

> Should we mark svg files as binary files in gitattributes? Same for resources. The diff shouldn't be displayed?

Not sure we need this, I sometimes like the SVG text diffs. They are mostly useless if you are doing larger changes with a tool, but sometimes you just do minor changes (e.g. removing `width` / `height` attribute). And resources.py doesn't display a diff by default anyway.

samj1912:
> > Lgtm. Should we mark svg files as binary files in gitattributes? Same for resources. The diff shouldn't be displayed?
> You can display them, switching to rich diff.
Even if they are binaries you can display them. 
> And resources.py doesn't display a diff by default anyway.

The point is that git is tracking the diffs unnecessarily, this is esp. applicable for resources.py where git is keeping a track of these large diffs which we can avoid if we switch them to binary files in git attribute.

It is also an indication that you are not supposed to be focusing on them during PR reviews.

Anyway, it was just a suggestion.

phw:
Good point. I'm all for setting the binary flag for resources.py in this case

PR #1254, opened 2019-08-30
PICARD-1547: Mark WAV unsupported tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
While Picard does not yet support tagging WAV files, the metadatabox still displayed all tags for WAV. This raised the expectation those tags can be edited and saved.

* JIRA ticket (_optional_): [PICARD-1547](https://tickets.metabrainz.org/browse/PICARD-1547)

# Solution
Implement `WAVFile.supports_tag` to always return `False`. This is the same that is already done in the `MIDIFile` format class.

# Action

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -501,9 +501,9 @@ class MetadataBox(QtWidgets.QTableWidget):
                 tag_diff.objects += 1
 
         all_tags = set(list(orig_tags.keys()) + list(new_tags.keys()))
-        tag_names = COMMON_TAGS + \
-            sorted(all_tags.difference(COMMON_TAGS),
-                   key=lambda x: display_tag_name(x).lower())
+        common_tags = [tag for tag in COMMON_TAGS if file.supports_tag(tag)]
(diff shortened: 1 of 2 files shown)


picard/ui/metadatabox.py
```diff
@@ -501,8 +501,9 @@ def _update_tags(self):
                 tag_diff.objects += 1
 
         all_tags = set(list(orig_tags.keys()) + list(new_tags.keys()))
-        tag_names = COMMON_TAGS + \
-            sorted(all_tags.difference(COMMON_TAGS),
+        common_tags = [tag for tag in COMMON_TAGS if file.supports_tag(tag)]
+        tag_names = common_tags + \
```
samj1912: Use parentheses for line continuation?

phw: What about this?

```python
tag_names = common_tags + sorted(all_tags.difference(common_tags),
                                 key=lambda x: display_tag_name(x).lower())
```

That's probably even nicer.

PR #1258, opened 2019-08-30
Treat picard/resources.py as binary file

phw:
This file is changed only automatically, diffing and merging it is usually pointless.

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

# Solution

# Action

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,0 +1,1 @@
+picard/resources.py binary


phw:
@samj1912 I looked over the project, but I think this is the only file where I would say the diff is truly useless. Or do you have anything else?

PR #1263, opened 2019-09-04
PICARD-1586: ReplayGain 2.0 tags

phw:
# Summary
This adds support for the tags as defined in the [ReplayGain 2.0 Spec](https://wiki.hydrogenaud.io/index.php?title=ReplayGain_2.0_specification)

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1586](https://tickets.metabrainz.org/browse/PICARD-1586)

# Solution
Add support to read and write those tags. One thing that made this a bit more complicated is the need to support case insensitive reading of those tags (see discussion on the ticket). This needed some adjustments to ID3, MP4 and ASF formats. But we can reuse those for having more tags being read case-insensitive (e.g. there has been discussion for the `artists` tag and probably others would benefit). But for now this implementation focuses on the ReplayGain tags.

I tried to add extensive tests for this, as there are quite a few things that can go wrong here.

# Action
Not sure whether we should merge this for a 2.2 release or wait till after the release.


picard/formats/id3.py
```diff
@@ -177,6 +180,18 @@ class ID3File(File):
     __rtranslate_freetext = dict([(v, k) for k, v in __translate_freetext.items()])
     __translate_freetext['writer'] = 'writer'  # For backward compatibility of case
 
+    # Freetext fields that are loaded case-insensitive
+    __rtranslate_freetext_ci = {
```
zas: Isn't this redundant with https://github.com/metabrainz/picard/pull/1263/files#diff-0d55c0ba51fcdcf1cab8dfe3acac98b0R166 ?

phw: Not really, one is ASF the other ID3. So while in this specific case it is the same list of tags this is just coincidence and quickly can be different when we add other tags to the list. E.g. we could decide to also read `MusicBrainz Artist Id` case insensitive for ID3 (because this has been discussed before). And the mapping is different for ASF.

What we could look into is to unify the way we specify the mappings here. E.g. ASF by default always has the mapping specified as "picard name => format name", while ID3 and MP4 use "format name => picard name". The first was easier to work with in this case. Would probably be easier if this would be the same across all formats, at least when possible. Also the naming of the mapping constants could be unified (`__TRANS` vs. `__translate_freetext` vs. `__freeform_tags` vs `__translate`).

zas: Thanks for clarification. Keep as is for now.

Moonbase59:
Just had a quick glance, looks good. Will probably try the next days, if time allows.

Just one question? Would Picard leave the casing (lower, upper, mixed) of the existing TG tags the same upon writing the tags? This is mandatory because a) loudgain allows to select if lower- or uppercase tags should been written and b) unfortunately there are still a zillion players out there that do _not_ read case-insensitively:

* KODI insists on lowercase tags
* VLC insists on uppercase tags
* IDJC (currently) insists on lowercase tags.
* foobar2000 reads case-insensitively

I couldn’t test too many players, but there are surely many that only take _one_ variant.
In a typical workflow, replaygaing would happen _before_ tagging with Picard, so it shouldn’t change the casing.

phw:
> Just one question: Would Picard leave the casing (lower, upper, mixed) of the existing RG tags _the same_ upon writing the tags?

This patch allows for case-insensitive reading, but will always write uppercase tags. If preserving case is important this would need to be added by remembering the original case. I can look into this. I think this affects only ID3, MP4 and ASF. Vorbis comments are case insensitive by definition, and APE tags are case insensitive for reading.

phw:
Preserving case was added in https://github.com/metabrainz/picard/pull/1263/commits/29cfbdf1b506669be204ea50c4f89e57caf70b22

I think case preservation is something we want to do for all tags we read case insensitive. Currently those are only the ReplayGain tags, but I see no reason why this should not apply to all case-insensitive tags. Hence this implementation is generic.

Moonbase59:
Looks good. Loudgain actually also writes only uppercase to Vorbis Comments and APE.

Another question: Would these changes also simplify the tag exclusion list? I currently need to replicate all tags in lower- and uppercase like
```
bpm, xfade, XFade, replaygain_track_gain, REPLAYGAIN_TRACK_GAIN, replaygain_track_peak, REPLAYGAIN_TRACK_PEAK, replaygain_album_gain, REPLAYGAIN_ALBUM_GAIN, replaygain_album_peak, REPLAYGAIN_ALBUM_PEAK, replaygain_reference_loudness, REPLAYGAIN_REFERENCE_LOUDNESS, replaygain_album_range, REPLAYGAIN_ALBUM_RANGE, replaygain_track_range, REPLAYGAIN_TRACK_RANGE, r128_track_gain, R128_TRACK_GAIN, r128_album_gain, R128_ALBUM_GAIN
```

phw:
> Another question: Would these changes also simplify the tag exclusion list? I currently need to replicate all tags in lower- and uppercase like

Yes, absolutely. You had to do this because Picard did not know anything about those tags. Hence it only read them as they were in the file. With the new implementation Picard knows about those tags and there are defined names for them (all lowercase `replaygain_*`). So you only need to exclude the lowercase variant now.

Moonbase59:
Just … amazing! I _know_ why I’ve been loving Picard forever and a day, and recommending it to everyone I know. :-)

test/formats/common.py
```diff
@@ -52,7 +55,17 @@ def save_and_load_metadata(filename, metadata):
 
 
 def load_raw(filename):
-    return mutagen.File(filename)
+    file = mutagen.File(filename)
```
samj1912: Can we rename this? File is a keyword.

test/formats/test_id3.py
```diff
@@ -191,6 +196,40 @@ def test_preserve_unchanged_tags_v23(self):
...
+            # Ensure values are not duplicated on repeated save and are saved
+            # case preserving.
+            tags = mutagen.id3.ID3Tags()
+            tags.add(mutagen.id3.TXXX(desc='Replaygain_Album_Peak', text='0.978475'))
+            save_raw(self.filename, tags)
+            loaded_metadata = load_metadata(self.filename)
+            loaded_metadata['replaygain_album_peak'] = '1.0'
+            save_metadata(self.filename, loaded_metadata)
+            raw_metadata = load_raw(self.filename)
+            # print(raw_metadata)
```
samj1912: Remove?

test/formats/common.py
phw: Good point. I addressed this for `load_raw` / `save_raw` in this file. But this is actually something we should look into over the entire source. A few places use `file_` or just `f`, but usage of `file = ...` is quite common in the codebase.

test/formats/test_id3.py
phw: Fixed

phw:
> While you’re at it, could you please verify that Picard actually reads Vorbis Comments and APE case-insensitively?

Yes for Vorbis. It is a bit of a mixed story for APE: Picard will read the tags for which it has a defined mapping case sensitive. This is e.g `Album Artist`. For tags where it has no  explicitly defined mapping it will read case insensitive. This is for example true for `Artist` (which is implicitly mapped to the internal name `artist` in Picard). Also writing does not preserve case here and will always rewrite the tags according to the mapping rules. That is explicitly mapped tags get written as specified (`Album Artist`), unmapped tags get written with title case conversion (`artist` => `Artist`). mutagen actually has a case preserving logic for APE, but this does not apply to the way Picard sets the tags.

I added a ticket for this: https://tickets.metabrainz.org/browse/PICARD-1592
In general the APE tags in Picard are lacking some love.

Moonbase59:
Fair enough for now, I guess, since they are _supposed_ to be read case-insensitively. Everything I tested seems to respect that. Just hacked up a little test case (writing lowercase APE tags and checking how these got handled in a handful of players and even Kid3-qt), all good.

So I should probably not worry if Picard remaps my uppercase APE RG tags to "whatever" case. :-)

picard/formats/asf.py
```diff
@@ -195,7 +210,15 @@ def _load(self, filename):
...
                     metadata["totaldiscs"] = disc[1]
                     values[0] = disc[0]
-            name = self.__RTRANS[name]
+            name_lower = name.lower()
+            if name in self.__RTRANS:
+                name = self.__RTRANS[name]
+            elif name_lower in self.__RTRANS_CI:
+                orig_name = name
+                name = self.__RTRANS_CI[name_lower]
+                self.__casemap[name] = orig_name
```
zas: May be a stupid question but is is possible to have the same tag with case variation more than once in original tags?

picard/formats/id3.py
```diff
@@ -177,6 +180,18 @@ class ID3File(File):
...
+        'replaygain_album_gain': 'REPLAYGAIN_ALBUM_GAIN',
+        'replaygain_album_peak': 'REPLAYGAIN_ALBUM_PEAK',
+        'replaygain_album_range': 'REPLAYGAIN_ALBUM_RANGE',
+        'replaygain_track_gain': 'REPLAYGAIN_TRACK_GAIN',
+        'replaygain_track_peak': 'REPLAYGAIN_TRACK_PEAK',
+        'replaygain_track_range': 'REPLAYGAIN_TRACK_RANGE',
+        'replaygain_reference_loudness': 'REPLAYGAIN_REFERENCE_LOUDNESS',
+    }
+    __translate_freetext_ci = dict([(b.lower(), a) for a, b in __rtranslate_freetext_ci.items()])
```
zas: Since all keys are lowercased, is `__translate_freetext_ci` still needed?

zas: May be introduce a small utility function to do that, as it is repeated at https://github.com/metabrainz/picard/pull/1263/files#diff-4909d6dcdd5d880b988729ebd5d5586dR138

picard/formats/asf.py
phw: Yes, but as they are treated case-insensitive they are treated as the same tag. In this case last casing wins. For replaygain the specs say there must be only one tag anyway.

picard/formats/mp4.py
```diff
@@ -124,15 +125,32 @@ class MP4File(File):
...
     __other_supported_tags = ("discnumber", "tracknumber",
                               "totaldiscs", "totaltracks")
 
+    def __init__(self, filename):
+        super().__init__(filename)
+        self.__casemap = {}
+
     def _load(self, filename):
         log.debug("Loading file %r", filename)
         file = MP4(encode_filename(filename))
```
zas: In https://github.com/metabrainz/picard/pull/1263/files#diff-91b2e53602ea30ecaaf16fe53ceeaa94R236 you initialize `self.__casemap` but not here, intended?

picard/formats/id3.py
phw: I'd rather rework all the formats and unify the tag mapping handle in a separate PR then to blow up this PR. In this process we could extract utility functions for dict reversal.

picard/formats/mp4.py
phw: No, that's an oversight. I think it makes sense to initialize this freshly on load, so in case we would have repeated loads old mappings don't stay around. Fixed this for both ASF and MP4

picard/formats/asf.py
Moonbase59: It _can_ happen, though it’s considered bad style.

Moonbase59: N.B.: For tag formats that support having "the same" tag in different case variations, loudgain currently checks for both (pure) lower- and uppercase variants and removes these before writing new tags. I’ll make a note to add handling for real case insensitivity in loudgain’s tag removal code.

Moonbase59: We can run into curious situations otherwise: Tagger A writes lowercase, tagger B writes uppercase, player A only reads lowercase, player B reads only uppercase. If values are different, the user might get surprising loudness variations and never know why. So I opted for removal of _both_ and rewrite in the case specified on the commandline (-L option).

Moonbase59: https://github.com/Moonbase59/loudgain/issues/10

Moonbase59:
Real case-insensitive RG tag removal done in https://github.com/Moonbase59/loudgain/commit/1cda93072536072ce3e20004b819b8166267cce9

PR #1264, opened 2019-09-05
PICARD-1590: Fixed thread locking when saving options

phw:
# Summary
Avoids some kind of thread lock when saving options.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When saving the options while a file is selected and the `clear_existing_tags` was toggled, Picard could completely freeze up (at least on Linux, I could not reproduce on Windows). The basic cause is the call to `MetadataBox.update`.

To reproduce:

1. Open current master with e.g. `./tagger.py test/data/test.mp3`
2. Select the loaded file (so that the metadatabox shows the metadata of this file)
3. Open Options, toggle Tags > Clear existing tags
4. Save Options

Often the app is now unresponsive. If this does not work, just repeat steps 3. and 4. Usually after not more than 5 tries I can reproduce this. But often it freezes on first try for me.

* JIRA ticket (_optional_): [PICARD-1590](https://tickets.metabrainz.org/browse/PICARD-1590)

# Solution
To be honest I am not entirely sure about the final cause and the solution. But calling `self.tagger.window.metadata_box.update` on the main thread seems to avoid the issue.

I think this is somewhat related that the QDialog itself running code in separate threads. But I really would appreciate some critical review and some better ideas on what is happening. Also if someone can both reproduce the issue

# Action

### DIFF
--- a/picard/ui/options/tags.py
+++ b/picard/ui/options/tags.py
@@ -26,6 +26,7 @@ from PyQt5 import (
 
 from picard import config
 from picard.util.tags import TAG_NAMES
+from picard.util.thread import to_main
 
 from picard.ui.options import (
     OptionsPage,
@@ -97,7 +98,7 @@ class TagsOptionsPage(OptionsPage):
         clear_existing_tags = self.ui.clear_existing_tags.isChecked()
         if clear_existing_tags != config.setting["clear_existing_tags"]:
             config.setting["clear_existing_tags"] = clear_existing_tags
-            self.tagger.window.metadata_box.update()
+            to_main(self.tagger.window.metadata_box.update)
         config.setting["write_id3v1"] = self.ui.write_id3v1.isChecked()
         config.setting["write_id3v23"] = self.ui.write_id3v23.isChecked()
         config.setting["id3v23_join_with"] = self.ui.id3v23_join_with.currentText()


phw:
See also my comments on https://tickets.metabrainz.org/browse/PICARD-1590

phw:
@zas I have a similar hanging happening when using "Add files" dialog, could you check this?

Basically in this case Picard hangs after opening "Add files" dialog, selecting some files and pressing ok. `self.tagger.add_files` still gets called, but then it freezes again somewhere inside this method.

The problem: this happens far less frequent than the issue fixed here, so it is hard to verify if any change remove the issue. Also I am not sure if "Add directory" is affected as well. I guess in theory it is similar enough to have the same issue, but I cannot reproduce this.

> For me this patch is very required for a stable release, I'm not sure the fix is the best one, but it works (tm). I have the feeling

Fully agree, this *must* be fixed for a release. It annoys me a bit that I don't fully understand what is happening, but since this patch prevents the bug I'll take it unless someone comes up with something better :)

zas:
> Basically in this case Picard hangs after opening "Add files" dialog, selecting some files and pressing ok. `self.tagger.add_files` still gets called, but then it freezes again somewhere inside this method.

I can't reproduce such issue.

phw:
> I can't reproduce such issue.

Ok, I merge this one for now. I will try to look into the other hanging a bit more.

PR #1269, opened 2019-09-10
PICARD-1594: Add $is_video() and $is_audio() scripting functions

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
This adds the `$is_audio()` and `$is_video()` functions that are currently provided by the videotools plugin.

* JIRA ticket (_optional_): [PICARD-1594](https://tickets.metabrainz.org/browse/PICARD-1594)

# Solution

# Action
See https://tickets.metabrainz.org/browse/PICARD-1594 and https://tickets.metabrainz.org/browse/PICARD-1537 for my reasons to move this from the plugin.


picard/script.py
```diff
@@ -894,6 +894,24 @@ def iswbound(char):
...
 
+def func_is_audio(parser):
+    """Returns true, if the file processed is an audio file."""
+    if func_is_video(parser) == "1":
+        return ""
+    else:
+        return "1"
+
+
+VIDEO_EXTENSIONS = ['m4v', 'wmv', 'ogv', 'oggtheora']
```
zas: Since it is exported (at least to tests), I'd rather move this constant at top of file and since it isn't meant to be modified (is it?) I think we can use a tuple instead of a list.

phw: Actually I think we should handle this differently and files should set the `~video` tag themselves. This moves the responsibility of handling this to the specific formats. This constant was a workaround needed by the plugin. I'll look into this.

picard/formats/asf.py
```diff
@@ -287,3 +287,11 @@ def _get_tag_name(self, name):
             return self.__TRANS[name]
         else:
             return None
+
+    def _info(self, metadata, file):
+        super()._info(metadata, file)
+        filename = file.filename
+        if isinstance(filename, bytes):
+            filename = filename.decode()
```
zas: An exception is possible here (UnicodeError), unsure if it is something we need to worry about.

phw: Not sure. Usually this filename should already have been valid or we would have failed earlier. After all we pass a Python string to mutagen. I think if this goes wrong here we have a bug, and the exception will make this bug visible (the user will see a loading error). I don't know what we could sanely do here in capturing this exception other than throwing again.

PR #1274, opened 2019-09-11
PICARD-622: Do not reset horizontal scroll position in main window panes on selection change

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The horizontal scroll position in all three main window panes (file browser, untagged files, tagged files) resets to 0 when selection changes. This is default behavior of `QTreeView.scrollTo`.

* JIRA ticket (_optional_): [PICARD-622](https://tickets.metabrainz.org/browse/PICARD-622)

# Solution
Reimplement `scrollTo` to better handle horizontal scroll position.

- For the file browser, set the horizontal scroll position to the indentation level of the parent. This allows intuitive and easy navigation through deep hierarchies without the need of manually handling the scrolling.
- For the untagged / tagged files panes we don't have deep hierarchies, but potentially large data columns. Just keep the current horizontal scrolling position. Usually if the user needs to scroll horizontally here it means the panes are very narrow but the user probably focuses on a certain column. This behavior ensures the focus on the column is not lost.

# Action
@zas Please try this locally. The behavior is hard to explain, one has to actually use it to see if it "feels good".

### DIFF
--- a/picard/ui/filebrowser.py
+++ b/picard/ui/filebrowser.py
@@ -102,6 +102,18 @@ class FileBrowser(QtWidgets.QTreeView):
             self.scrollTo(self.currentIndex())
         QtCore.QTimer.singleShot(0, scroll)
 
+    def scrollTo(self, index, scrolltype=QtWidgets.QAbstractItemView.EnsureVisible):
+        # QTreeView.scrollTo resets the horizontal scroll position to 0.
+        # Reimplemented to instead scroll to horizontal parent position.
+        level = -1
+        super().scrollTo(index, scrolltype)
+        parent = self.currentIndex().parent()
+        while parent.isValid():
+            parent = parent.parent()
+            level += 1
+        pos_x = max(self.indentation() * level, 0)
+        self.horizontalScrollBar().setValue(pos_x)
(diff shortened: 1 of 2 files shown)


zas:
> @zas Please try this locally. The behavior is hard to explain, one has to actually use it to see if it "feels good".

The patch improves the situation, but the horizontal scroll position is still modified for file browser, but it works as expected for item view, according to my tests.

phw:
> The patch improves the situation, but the horizontal scroll position is still modified for file browser, but it works as expected for item view, according to my tests.

Please see above, the file browser does modify the horizontal position to make the selected item visible. The original behavior was to always reset completely to the left. This is not useful view:

screenshot-2019-09-11-14-56-20

If we just maintain the scroll position as we do for the other panes one might instead get this:

grafik

This still is no useful view, this is nearly unusable with keyboard alone. You always have to correct horizontal scroll position using the scrollbar. Hence this patch scrolls to the indentation of the selected file (or rather the selected's parent):

grafik

PR #1280, opened 2019-09-13
PICARD-456: Do not remove empty standard directories

phw:
This protects standard directories like e.g. ~/Desktop being deleted if empty.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Folders considered "special", such as the Desktop, default Music or other media folders or even configuration folders, could get deleted by the "Remove empty directories" option. This PR fixes this.

* JIRA ticket (_optional_): [PICARD-456](https://tickets.metabrainz.org/browse/PICARD-456)

# Solution

Exclude directories listed in `QStandardPaths`. When comparing also resolve symlinks.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -79,6 +79,10 @@ for location in get_qt_enum(QStandardPaths, QStandardPaths.StandardLocation):
         PROTECTED_DIRECTORIES.add(os.path.realpath(path))
 
 
+class SkipRemoveDir(Exception):
+    pass
+
+
 class File(QObject, Item):
 
     metadata_images_changed = pyqtSignal()
@@ -297,13 +301,12 @@ class File(QObject, Item):
                 if not tail:
                     head, tail = os.path.split(head)
                 while head and tail:
-                    try:
-                        self._rmdir(head)
-                    except BaseException:
-                        break
+                    self._rmdir(head)
                     head, tail = os.path.split(head)
-            except OSError:
-                pass
+            except OSError as why:
+                log.warning("Error removing directory: %s", why)
+            except SkipRemoveDir as why:
+                log.debug("Not removing empty directory: %s", why)
         # Save cover art images
         if config.setting["save_images_to_files"]:
             self._save_images(os.path.dirname(new_filename), metadata)
@@ -311,11 +314,12 @@ class File(QObject, Item):
 
     @staticmethod
     def _rmdir(path):
-        if (os.path.realpath(path) not in PROTECTED_DIRECTORIES
-            and not set(os.listdir(path)) - JUNK_FILES):
-            shutil.rmtree(path, False)
+        if os.path.realpath(path) in PROTECTED_DIRECTORIES:
+            raise SkipRemoveDir('%s is a protected directory' % path)
+        elif set(os.listdir(path)) - JUNK_FILES:
+            raise SkipRemoveDir('%s is not empty' % path)
         else:
-            raise OSError
+            shutil.rmtree(path, False)
 
     def _saving_finished(self, result=None, error=None):
         # Handle file removed before save
--- a/picard/file.py
+++ b/picard/file.py
@@ -27,11 +27,7 @@ import re
 import shutil
 import unicodedata
 
-from PyQt5.QtCore import (
-    QObject,
-    QStandardPaths,
-    pyqtSignal,
-)
+from PyQt5 import QtCore
 
 from picard import (
     PICARD_APP_NAME,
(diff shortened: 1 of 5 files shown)


picard/file.py
```diff
@@ -285,7 +302,7 @@ def _save_and_rename(self, old_filename, metadata):
                     except BaseException:
                         break
                     head, tail = os.path.split(head)
-            except EnvironmentError:
```
zas: I wonder about loops/exception handling here.
The try/except block around this second call to `_rmdir()` matches BaseException.
Dunno but doesn't look clean to me, perhaps we should use a new exception class to differentiate "skip" and "failure".

phw: Yes, this is a good point. Also gives us the opportunity to do better logging here, so if we get reports of failed removals of empty dirs we can actually find out why.

https://github.com/metabrainz/picard/pull/1280/commits/6f6fb28414dec7cd266809d7f03945b4b04fba26

picard/util/__init__.py
```diff
@@ -611,3 +611,12 @@ def find_best_match(candidates, no_match):
     else:
         result = no_match
     return BestMatch(similarity=result.similarity, result=result, num_results=len(sorted_results))
+
+
+def get_qt_enum(cls, enum):
```
zas: We should document this function, it's simple, but its usage isn't obvious. An example could do.

picard/file.py
```diff
@@ -280,25 +301,25 @@ def _save_and_rename(self, old_filename, metadata):
...
+                log.warning("Error removing directory: %s", why)
+            except SkipRemoveDir as why:
+                log.debug("Not removing empty directory: %s", why)
         # Save cover art images
         if config.setting["save_images_to_files"]:
             self._save_images(os.path.dirname(new_filename), metadata)
         return new_filename
 
     @staticmethod
     def _rmdir(path):
```
zas: May be we should add tests for it.

picard/file.py
```diff
@@ -280,25 +301,25 @@ def _save_and_rename(self, old_filename, metadata):
...
-        junk_files = (".DS_Store", "desktop.ini", "Desktop.ini", "Thumbs.db")
-        if not set(os.listdir(path)) - set(junk_files):
-            shutil.rmtree(path, False)
+        if os.path.realpath(path) in PROTECTED_DIRECTORIES:
+            raise SkipRemoveDir('%s is a protected directory' % path)
+        elif set(os.listdir(path)) - JUNK_FILES:
+            raise SkipRemoveDir('%s is not empty' % path)
         else:
-            raise OSError
+            shutil.rmtree(path, False)
```
zas: According to https://docs.python.org/3/library/shutil.html#shutil.rmtree second argument is False by default, and so `shutil.rmtree(path)` should be enough.

picard/file.py
```diff
@@ -280,25 +301,25 @@ def _save_and_rename(self, old_filename, metadata):
...
         return new_filename
 
     @staticmethod
     def _rmdir(path):
-        junk_files = (".DS_Store", "desktop.ini", "Desktop.ini", "Thumbs.db")
-        if not set(os.listdir(path)) - set(junk_files):
-            shutil.rmtree(path, False)
+        if os.path.realpath(path) in PROTECTED_DIRECTORIES:
+            raise SkipRemoveDir('%s is a protected directory' % path)
+        elif set(os.listdir(path)) - JUNK_FILES:
```
zas: I wonder about another approach:

- try to os.rmdir(path): if it fails, either it's because of permissions or the like, then bail out, else it may fail because it's non-empty, in this case remove junk files from it, and try again, if it fails again, bail out

This way we don't have to calll listdir() at all, and it's faster when no junk files exist.

What do you think?

phw: >     * in this case remove junk files from it, and try again, if it fails again, bail out

We must not do this, the "junk" files are not necessarily there to be removed at will, we should only remove them if we indeed remove the directory. Example is the `desktop.ini` files on windows: Those control how the directory is displayed in Windows Explorer. If we just remove them in the hope they are the only file in the dir we might break user configurations. They just don't block the directory from being considered empty.

So in case the `os.rmdir` would fail we then would still need to call `os.listdir` to figure out whether the junk files are really all that is there, which wouldn't help us much.

But I agree that this needs better testing, I'm going to factor out this code and add proper tests.

zas: Ok, make sense

picard/util/__init__.py
phw: Done

picard/file.py
phw: Done

picard/file.py
phw: I moved the empty dir detection and removal into a separate module and addes tests, please take a look.

PR #1282, opened 2019-09-13
PICARD-1601: Ensure we package for PyPI with all test dependencies

phw:
Ensure we package for PyPI with all test dependencies.

# Summary
Fully test packaging sdist, extracting it and running the tests inside it. Fixes that we miss packaging test dependencies.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1601](https://tickets.metabrainz.org/browse/PICARD-1601)

# Solution

# Action

### DIFF
--- a/scripts/run-sdist-test.sh
+++ b/scripts/run-sdist-test.sh
@@ -1,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Build the sdist archive, extract it and run the tests.
+
+set -e
+
+rm -rf dist
+python3 setup.py sdist
+cd dist
+SDIST_ARCHIVE=$(echo picard-*.tar.gz)
+tar xvf $SDIST_ARCHIVE
+cd ${SDIST_ARCHIVE%.tar.gz}
+python3 setup.py test
--- a/.travis.yml
+++ b/.travis.yml
@@ -61,7 +61,10 @@ matrix:
     - INSTALL_DEPS="$PIP_INSTALL -r requirements-macos.txt"
   - os: linux
     before_install: skip
-    script: pip3 install .
+    install: skip
+    script:
+    - pip3 install .
+    - scripts/run-sdist-test.sh
   fast_finish: true
 
 before_cache:
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -5,31 +5,38 @@ recursive-include scripts *.in
 graft ui
 recursive-include resources *.py
 
-recursive-include test *.aac
-recursive-include test *.aiff
-recursive-include test *.ape
-recursive-include test *.asf
-recursive-include test *.dat
-recursive-include test *.dsf
-recursive-include test *.flac
-recursive-include test *.gif
-recursive-include test *.jpg
-recursive-include test *.json
-recursive-include test *.m4a
-recursive-include test *.mp3
-recursive-include test *.mpc
-recursive-include test *.ofr
-recursive-include test *.ofs
-recursive-include test *.ogg
-recursive-include test *.opus
-recursive-include test *.png
 recursive-include test *.py
-recursive-include test *.spx
-recursive-include test *.tak
-recursive-include test *.tta
-recursive-include test *.wav
-recursive-include test *.wma
-recursive-include test *.wv
+recursive-include test/data *.aac
+recursive-include test/data *.aiff
+recursive-include test/data *.ape
+recursive-include test/data *.asf
+recursive-include test/data *.dat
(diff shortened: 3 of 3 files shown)


phw:
This is just a first commit to see the tests failing. I have another commit waiting which actually ads missing files to the manifest.

phw:
I have no idea why tests are failing. Locally this is running just fine, but on Travis extracting the archive and running setup.py test (or py.test) on it just fails with module not found issues like:

```
___________________ ERROR collecting test/test_filesystem.py ___________________
ImportError while importing test module '/home/travis/build/metabrainz/picard/dist/picard-2.2.0.dev4/test/test_filesystem.py'.
Hint: make sure your test modules/packages have valid Python names.
Traceback:
test/test_filesystem.py:7: in <module>
    from test.picardtestcase import PicardTestCase
E   ModuleNotFoundError: No module named 'test.picardtestcase'
```

I don't get it. This is in the correct directory, the files in test are there. I don't see the difference to my local setup.

zas:
> I have no idea why tests are failing. Locally this is running just fine, but on Travis extracting the archive and running setup.py test (or py.test) on it just fails with module not found issues like:
> 
> ```
> ___________________ ERROR collecting test/test_filesystem.py ___________________
> ImportError while importing test module '/home/travis/build/metabrainz/picard/dist/picard-2.2.0.dev4/test/test_filesystem.py'.
> Hint: make sure your test modules/packages have valid Python names.
> Traceback:
> test/test_filesystem.py:7: in <module>
>     from test.picardtestcase import PicardTestCase
> E   ModuleNotFoundError: No module named 'test.picardtestcase'
> ```
> 
> I don't get it. This is in the correct directory, the files in test are there. I don't see the difference to my local setup.

I don't see `test/__init__.py` in copied files, not sure if it is intended.

phw:
> I don't see `test/__init__.py` in copied files, not sure if it is intended.

Good catch. It is here locally, but missing on the travis package. Confusing, but I'll look into it.

phw:
@zas Found the issue. I accidentally changed:

    recursive-include test *.py

to 

    recursive-include test/data *.py

Not sure why this works locally, I tried several times with completely removing the dist dir. But it is definitely what broke it on travis. Thanks for finding this, it is fixed now.

samj1912:
I would suggest putting this on as an extra target. https://setuptools.readthedocs.io/en/latest/setuptools.html#declaring-extras-optional-features-with-their-own-dependencies

phw:
We had issues reported for the source distribution not being able to run the tests, and I think running the tests before installing something from source is a valid use case. Would the extras solve this?

scripts/run-sdist-test.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
+
+# Build the sdist archive, extract it and run the tests.
+
+set -e
+
+rm -rf dist
+python3 setup.py sdist
+cd dist
+SDIST_ARCHIVE=$(ls picard-*.tar.gz)
```
zas: This use of `ls` looks suspicious to me. Using `echo` builtin is prolly safer.

samj1912:
@phw I think this belongs in extras. I have never seen a pip package that includes tests. If they do, it's always as an extra target.

samj1912:
The python package authority also recommends excluding tests in their sample project. https://github.com/pypa/sampleproject/blob/master/setup.py#L133

samj1912:
And they have an extra target here to install test dependencies https://github.com/pypa/sampleproject/blob/master/setup.py#L160

phw:
@samj1912 Ok, might make sense then. Can you add a ticket for this? My main concern right now is fixing the existing package (which includes tests) since we are about to make a release, we can optimize it later.

On a quick test on my system tests are always included, no matter if they are set in the manifest or not. Probably also the reason why the same build succeeded for me while failing on Travis. 

Also where we really could look  into to not include everything is resources. Actually most of the images there are not needed in the sdist as they are build into the Python code.

samj1912:
https://tickets.metabrainz.org/browse/PICARD-1602

phw:
Thanks a lot @samj1912 

See also my notes on https://tickets.metabrainz.org/browse/PICARD-1602 , things are a bit strange. @mineo  what are your ideas here?

For now I'm going to merge this PR so we at least have a working sdist, even if it contains more than necessary.

PR #1288, opened 2019-09-18
PICARD-1602: Minify sdist, removed tests and other files

phw:
# Summary
The sdist package includes data not needed for installation. This minifies the sdist to include only what is needed for installation.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1602](https://tickets.metabrainz.org/browse/PICARD-1602)

# Solution
- Explicitly exlcude `test/*`. Turns out including this independent of `MANIFEST.in` is a feature of setuptools, see https://setuptools.readthedocs.io/en/latest/setuptools.html#generating-source-distributions
- Do not include `ui/*`, those should not be needed
- Do not include `resources/*` except the files installed by data_files (and those are automatically included)
- Exclude files generated from `*.in`:`scripts/picard` and `org.musicbrainz.Picard.appdata.xml`. Both files are part of the installation and are therefor included by default by setuptools, but they are generated during build.

# Action


phw:
Thanks a lot to you all for taking a look.

PR #1290, opened 2019-09-18
travisci: Upload macOS artifacts to S3

phw:
# Summary

This uploads the macOS build artifacts to a S3 bucket instead of using transfer.sh

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

`transfer.sh` is frequently blocking builds and is very unreliable. Also we don't control how long artifacts are available and we have no single place to check for the latest artifacts.

# Solution
Setup uploading artifacts to a S3 bucket.

For this to work we need to set the following variables in Travis CI:

- `AWS_ARTIFACTS_BUCKET`
- `AWS_ACCESS_KEY_ID`
- `AWS_SECRET_ACCESS_KEY`

I have already setup a bucket for this and I would be fine with keeping this, but it could make senseto use a bucket created via a MetaBrainz AWS account. It just should be using credentials that have write access to the `AWS_ARTIFACTS_BUCKET` bucket and this bucket only. Also the bucket must be world readable for everyone to be able to download the builds.

For now the dmg images will be placed in `https://s3.eu-central-1.amazonaws.com/artifacts.picard.uploadedlobster.com/metabrainz/picard/${TRAVIS_BUILD_NUMBER}/`

# Action

### DIFF
--- a/scripts/package-osx.sh
+++ b/scripts/package-osx.sh
@@ -45,13 +45,19 @@ mv 'MusicBrainz Picard.tmp' 'MusicBrainz Picard.app'
 dmg="MusicBrainz Picard $VERSION.dmg"
 hdiutil create -volname "MusicBrainz Picard $VERSION" -srcfolder 'MusicBrainz Picard.app' -ov -format UDBZ "$dmg"
 [ "$CODESIGN" = '1' ] && codesign --keychain $KEYCHAIN_PATH --verify --verbose --sign "$CERTIFICATE_NAME" "$dmg"
-if [ -n "$UPLOAD_OSX" ]
-then
-    set +e
+md5 -r "$dmg"
+if [ -n "$UPLOAD_OSX" ]; then
     # make upload failures non fatal
-    curl -v --retry 6 --retry-delay 10 --max-time 180 --upload-file "$dmg" https://transfer.sh/
+    set +e
+    # Set $AWS_ARTIFACTS_BUCKET, $AWS_ACCESS_KEY_ID and $AWS_SECRET_ACCESS_KEY for AWS S3 upload to work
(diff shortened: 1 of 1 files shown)


phw:
Currently its failing to upload, the same setup worked on my fork at phw/picard. Not sure, have to look into this. Credentials should be fine, I just tested locally.

phw:
> Currently its failing to upload

Everything is fine, Travis CI just doesn't make secret variables available to pull requests from external repositories, which makes sense if you think of it (external attackers could easily modify the CI script, create a PR and access our secret variables this way). But we will fall back to transfer.sh in this case.

Running a final set of test builds to see if everything works.

phw:
So this is working now, latest builds for build no. 4982 are available here:

- https://s3.eu-central-1.amazonaws.com/artifacts.picard.uploadedlobster.com/metabrainz/picard/4982/MusicBrainz+Picard+2.2.0.final0_osx_xcode11_20190919061303.dmg
- https://s3.eu-central-1.amazonaws.com/artifacts.picard.uploadedlobster.com/metabrainz/picard/4982/MusicBrainz+Picard+2.2.0.final0_osx_xcode9.4_20190919060554.dmg
- https://s3.eu-central-1.amazonaws.com/artifacts.picard.uploadedlobster.com/metabrainz/picard/4982/MusicBrainz+Picard+2.2.0.final0_osx_xcode9_20190919060606.dmg

Now we just have to do something in regards of discoverability of those builds. At minimum print something noticeable in the builds logs, but I would even prefer some small download page that always shows latest builds.

Also noteworthy: Currently the builds will be cleaned up after 90 days. I think this should be more than enough, could even be that I lower this. Also we do not need to fear that we don't have any builds available, we have monthly builds setup on Travis.

PR #1292, opened 2019-09-19
PICARD-1609: Show actual runtime version of Qt

phw:
# Summary
Use `qVersion()` instead of `QT_VERSION_STR`
    
  This shows the proper runtime version of Qt being used instead of the Qt version PyQt was built against.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1609](https://tickets.metabrainz.org/browse/PICARD-1609)

# Solution

# Action

### DIFF
--- a/picard/util/versions.py
+++ b/picard/util/versions.py
@@ -42,7 +42,7 @@ _versions = OrderedDict([
     ("mutagen-version", mutagen_version),
     ("discid-version", discid_version),
     ("astrcmp", astrcmp_implementation),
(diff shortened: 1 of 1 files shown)


phw:
> Nice catch! (and fix)

Thanks. In most cases I deal with the version is the same, I just found it after I got fooled by it with a pip install with a different version then the system libraries.

PR #1298, opened 2019-10-01
PICARD-1618: Include astrcmp C in macOS and Windows packages

phw:
# Summary
The current packages for Windows and macOS contain only the Python version of astrcmp, not the much faster C module.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The missing C module was caused by update to Pyinstaller >= 3.4. Before that version the compiled `_astrcmp` module got added as a single file `picard.util._astrcmp.pyd`. Later versions actually create the folder hierarchy `picard/util/` which then contains the compiled module.

This has two implications:

1. The `picard/` subfolder was not included in the Windows package
2. The `picard/` subfolder clashed with the name of the main executable `picard` on platforms other then Windows (on Windows it is always `picard.exe`). This made pyinstaller fail if `python setup.py build_ext -i` was run before, but that is required to have the _astrcmp module included.

* JIRA ticket (_optional_): [PICARD-1618](https://tickets.metabrainz.org/browse/PICARD-1618)

# Solution

- For Windows, make sure to include the `picard` subfolder in the installer
- For other platforms change the executable name to `picard-run` so it is different from the `picard` subfolder. This does not apply to Windows, I think it is fine to keep `picard.exe` there.

# Action


phw:
Tested built package on Windows, astrcmp C is installed and used:

```
PS C:\Users\Developer> & 'C:\Program Files\MusicBrainz Picard\picard.exe' -V                                                  PS C:\Users\Developer> Picard 2.2.2.dev1 (python3.7.4_20191001094100), Python 3.7.4, PyQt 5.13.1, Qt 5.13.1, Mutagen 1.42.0, Discid discid 1.1.1, libdiscid 0.6.2, astrcmp C, SSL OpenSSL 1.1.1b  26 Feb 2019
```

Also tested that conditional installation of locales still works.

phw:
Finally also tested on macOS, astrcmp C is now used there as well.

PR #1301, opened 2019-10-02
PICARD-1606: Handle OptionsPage exceptions

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Recently with Picard 2.2 the Classical Extras plugin caused crashes for many users, since it was doing some direct access to internal Picard config implementation which was no longer available after some refactoring.

This provided pretty bad user experience:

- For the user it was not clear what was causing the issues, Picard was just closing
- The user had not really a chance recovering from this situation without deeper knowledge of how to manually change plugins.

* JIRA ticket (_optional_): [PICARD-1606](https://tickets.metabrainz.org/browse/PICARD-1606)

# Solution
Avoids Picard crashing if e.g. a plugin provides a broken OptionsPage implementation by handling exceptions. This gives the user better feedback and eventually even allows user to recover from the situation.

# Action

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -22,7 +22,10 @@ from PyQt5 import (
     QtWidgets,
 )
 
-from picard import config
+from picard import (
+    config,
+    log,
+)
 from picard.const import PICARD_URLS
 from picard.util import (
     restore_method,
@@ -115,8 +118,11 @@ class OptionsDialog(PicardDialog):
 
         self.pages = []
         for Page in page_classes:
-            page = Page(self.ui.pages_stack)
-            self.pages.append(page)
+            try:
+                page = Page(self.ui.pages_stack)
+                self.pages.append(page)
+            except Exception:
+                log.exception('Failed initializing options page %r', page)
         self.item_to_page = {}
         self.page_to_item = {}
         self.default_item = None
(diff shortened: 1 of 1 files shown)


phw:
For testing I manipulated one of my installed plugins at various stages to throw exceptions:

1. Exception in `__init__`: Error is logged, the plugin's option page is not visible
2. Exception in `load`: Error is logged, the plugin's option page is deactivated and cannot be accessed
3. Exception in `check`: Error is logged, saving fails and an error dialog "Unknown error" with exception text is shown for the affected options page
4. Exception in `save`: Error is logged, saving fails and an error dialog "Unknown error" with exception text is shown for the affected options page

I could recover from both cases by removing the plugin and reinstalling it via the UI

picard/ui/options/dialog.py
```diff
@@ -137,7 +143,11 @@ def __init__(self, default_page=None, parent=None):
         self.finished.connect(self.saveWindowState)
 
         for page in self.pages:
-            page.load()
+            try:
+                page.load()
+            except Exception:
+                log.error('Failed loading options page %r', page, exc_info=True)
```
samj1912: This can just be log.exception and you don't need to pass the exc_info

picard/ui/options/dialog.py
```diff
@@ -115,8 +118,11 @@ def __init__(self, default_page=None, parent=None):
 
         self.pages = []
         for Page in page_classes:
-            page = Page(self.ui.pages_stack)
-            self.pages.append(page)
+            try:
+                page = Page(self.ui.pages_stack)
+                self.pages.append(page)
+            except Exception:
+                log.error('Failed initializing options page %r', page, exc_info=True)
```
samj1912: Same here

picard/ui/options/dialog.py
phw: Indeed, I changed it

PR #1306, opened 2019-10-06
PICARD-125: Support for raw AC3 / EAC3 files

phw:
# Summary
Adds support for reading (Enhanced) AC3 files and for tagging them with APEv2 tags.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-125](https://tickets.metabrainz.org/browse/PICARD-125)

# Solution

## Implementation
Add a parser for AC3 and E-AC3 file headers. Support tagging (E-)AC3 files with APEv2 metadata.

Since not all software might play well with the APEv2 header at the end of AC3 files saving the tags is optional. The user can choose to:

- Save APEv2 tags
- Not to save APEv2 tags, but leave existing tags in place (this is the default)
- Not to save APEv2 tags and remove existing APEv2 tags

No matter how the above is set, existing tags will always be read. For compatibility reasons (see tests below) the default is to not add APEv2 tags to AC3 files.

## Testing
- This implementation is two-way compatible with foobar2000 using the foo_ac3 AC3 plugin. EAC3 files are not supported by foo_ac3
-  This implementation is two-way compatible with MusicBee, which natively supports AC3 files with APEv2 tags. EAC3 is not supported.
- Windows Media Player, iTunes and Media Monkey do not support playing raw AC3 files

Applications, which support AC3 playback but not the APEv2 tags don't deal well with the unexpected tags. The following applications play the file, but show a wrong duration and therefor also abort playback early. In my test file of 2 minutes the added tags (including cover art) increased the displayed playback length by 11 seconds:

- GStreamer based applications (Totem, Lollypop)
- VLC
- "Groove Music" on Windows
- ffmpeg

TODO: Further testing with other players 

# Action
1. I am not really happy with the extended interface for Options > Tags. This gets a bit crowded now. So far the majority of options there had dealt with ID3 specifics, now we get additional options for another format (APEv2 / AC3). I think this page should be redesigned to group options by tagging formats.
 
    But I actually expect we might get additional options here in the Picard 2.3 cycle (e.g. freeform tags for MP4 might be optional). So I think we should defer any redesign until we better know what else might go here.
2. It's worth discussing whether we want this functionality in Picard or whether it should be a plugin. I'm fine with both. Only issue with plugins currently is that they don't provide the full testing framework of Picard.


phw:
Testing revealed that we probably should also change the behavior for AAC files to the same logic as here: Not save APEv2 by default, allow removing APEv2 tags optionally. Incompatible players again behave similar when seeing an AAC file with APEv2: Duration is displayed wrongly* and playback ends for the player unexpectedly (when reaching the real end of the audio stream).

* Duration calculation seems to be real tricky with AAC anyway. The estimate players show is often very wrong. VLC even constantly updates the duration estimate during playback, getting more and more accurate the less is left to play.

picard/formats/mutagenext/ac3.py
```diff
@@ -0,0 +1,296 @@
...
+}
+
+AC3_SAMPLE_RATES = [48000, 44100, 32000]
+
+AC3_BITRATES = [
+    32, 40, 48, 56, 64, 80, 96, 112, 128,
+    160, 192, 224, 256, 320, 384, 448, 512, 576, 640
+]
+
+EAC3_FRAME_TYPE_INDEPENDENT = 0
```
samj1912: Use enum here?

phw:
> By any chance do you have plans of contributing this upstream to mutagen?

Yes, absolutely. But my plan is to provide just the AC3 parsing / info part and add APEv2 handling on top in Picard (just like as we do it with AAC).

picard/formats/mutagenext/ac3.py
```diff
@@ -0,0 +1,296 @@
...
+        sample_rate (`int`): audio sampling rate in Hz
+        bitrate (`int`): audio bitrate, in bits per second
+        type (`str`): AC3 or EAC3
+    """
+
+    channels = 0
+    length = 0
+    sample_rate = 0
+    bitrate = 0
+    type = 'AC3'
```
phw: `type` is a reserved keyword.

picard/formats/ac3.py
```diff
@@ -0,0 +1,59 @@
...
+
+
+class AC3File(APEv2File):
+    EXTENSIONS = [".ac3", ".eac3"]
+    NAME = "AC3"
+    _File = ac3.AC3
+
+    def _info(self, metadata, file):
+        super()._info(metadata, file)
+        type = file.info.type or self.NAME
```
zas: `type` is a reserved word

phw:
Patch for AC3 support submitted to mutagen upstream: https://github.com/quodlibet/mutagen/pull/400

phw:
With https://github.com/metabrainz/picard/pull/1306/commits/2ec7cd4a03a0eea8eb5cd8334d7a4ea1f3bf0e88 the proposed implementation in mutagen is used. There is still a very simplified implementation as a fallback, but it does not allow to read the technical variables from AC3.

This way we could support tagging AC3 for older mutagen versions without the burden to carry all the implementation details. Users of future mutagen versions would benefit from full technical details.

phw:
The changes to mutagen have been merged (https://github.com/quodlibet/mutagen/pull/400), this is ready for review.

PR #1310, opened 2019-10-08
Log more information when a plugin failed to load

mineo:
# Summary

* This is a…
    * [ ] Bug fix
    * [X] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Plugin loading errors included nearly no information about which plugin from where was loaded at the point of failure and/or swallowed stack traces. Fix that.

# Solution

Log all the things! The individual commits have more details, including log entries before & after the change.

This also makes `plugin_error` behave more like the functions in `picard.log` in that it just takes an unformatted string with placeholders for formatting and a bunch of strings and does the formatting, instead of leaving that up to its callers.

# Action

No.

### DIFF
--- a/picard/pluginmanager.py
+++ b/picard/pluginmanager.py
@@ -220,8 +220,8 @@ class PluginManager(QtCore.QObject):
         for name in sorted(names):
             try:
                 self._load_plugin_from_directory(name, plugindir)
-            except Exception as e:
-                log.error("Unable to load plugin '%s': %s", name, e)
+            except Exception:
+                self.plugin_error(name, _("Unable to load plugin '%s'"), name, log_func=log.exception)
 
     def _get_plugin_index_by_name(self, name):
         for index, plugin in enumerate(self.plugins):
--- a/picard/pluginmanager.py
+++ b/picard/pluginmanager.py
@@ -169,9 +169,16 @@ class PluginManager(QtCore.QObject):
     def available_plugins(self):
         return self._available_plugins
 
-    def plugin_error(self, name, error, log_func=None):
-        if log_func is None:
-            log_func = log.error
+    def plugin_error(self, name, error, *args, **kwargs):
+        """Log a plugin loading error for the plugin `name` and signal the
+        error via the `plugin_errored` signal.
+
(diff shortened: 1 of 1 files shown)


picard/pluginmanager.py
```diff
@@ -278,11 +285,11 @@ def _load_plugin_from_directory(self, name, plugindir):
                           "version of Picard.") % (plugin.name, plugin.file)
                 self.plugin_error(plugin.name, error, log_func=log.warning)
         except VersionError as e:
-            error = _("Plugin %r has an invalid API version string : %s") % (name, e)
-            self.plugin_error(name, error)
+            error = _("Plugin %r has an invalid API version string : %s")
+            self.plugin_error(name, error, name, e)
```
phw: Why pass `e` here instead of using `log.exception`?

picard/pluginmanager.py
```diff
@@ -238,8 +245,8 @@ def _load_plugin_from_directory(self, name, plugindir):
                 module_file = info[0]
                 module_pathname = info[1]
             except ImportError:
-                error = _("Failed loading plugin %r") % name
-                self.plugin_error(name, error)
+                error = _("Failed loading plugin %r in %r")
+                self.plugin_error(name, error, name, [plugindir])
```
phw: Is passing a list as last argument intentional?

picard/pluginmanager.py
phw: No, you are right. No need to clutter the output with a traceback here, the exception and message are clear enougj

picard/pluginmanager.py
phw: Yup, makes sense

phw:
Looking at this again from a translation point of view I think we should use positional arguments for the formatting. Something like 

```
"Failed loading plugin %r in %r" % (name, plugindir)
```

might well need `name` and `plugindir` switched around in translations.

We should therefor change this to:


```
"Failed loading plugin {0!r} in {1!r}".format(name, plugindir)
```

But we likely have even more placed in the code where this applies.

PR #1312, opened 2019-10-08
PICARD-1632: Genre ui improvements

mineo:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [X] Minor / simple change (like a typo)
    * [ ] Other

# Problem

1. The tooltips on the genre options are quite long and not trivial to understand for people not
already familiar with them. Hiding them after just 5 seconds is way too soon

2. One of the examples claimed `+word*` means `word*` is excluded. This is wrong.

# Solution

1. To not get into bikeshedding over how long the tooltip should be shown, just don't hide it.

2. Fix the description of the example.

# Action

None

### DIFF
--- a/picard/ui/options/genres.py
+++ b/picard/ui/options/genres.py
@@ -63,8 +63,8 @@ TOOLTIP_TEST_GENRES_FILTER = N_("""<html><head/><body>
 This playground will not be preserved on Options exit.
 </p>
 <p>
-Red background means the tag will be skipped<br/>
-Green background means the tag will be kept
+Red background means the tag will be skipped.<br/>
+Green background means the tag will be kept.
 </p>
 </body></html>""")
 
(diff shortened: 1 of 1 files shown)


phw:
Any idea why PyQt 5.8 tests are suddenly failing on Travis? This seems to be unrelated to any code changes.

It fails with sip import errors, but sip gets installed. And it was working fine earlier today.

phw:
Found the issue: pip 5.0.0 got released today, obviously not working well with PyQt 5.8. I think we just need to pin an older version for the 5.8 builds.

phw:
@mineo: Can you add a ticket for the tooltip duration change? I think this is worth a changelog entry.

phw:
Added a ticket: https://tickets.metabrainz.org/browse/PICARD-1632

Fixed the builds in https://github.com/metabrainz/picard/commit/056bde87e611123e0418a56f8325664a041a91da

PR #1314, opened 2019-10-09
PICARD-1633: fix system locale detection on macOS

phw:
# Summary
On macOS the system locale was not detected, thus setting language to "System default" did not work.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1633](https://tickets.metabrainz.org/browse/PICARD-1633)

# Solution
This failed due to:
- macOS returning language / country separated by "-" instead of "_" (en-UK instead of en_UK)
- required pyobjc modules not being packaged

# Action


phw:
I tested the build from Travis CI on my Mac, language detection now worked as expected.

PR #1317, opened 2019-10-09
PICARD-1635: Fixed crashes on UI update for deleted files

phw:
# Summary
When files are already have state = REMOVED but the itemviews UI updates Picard could crash since there are no colors defined for this state.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1635](https://tickets.metabrainz.org/browse/PICARD-1635)

# Solution
Define the color dicts as defaultdict.

Questions to all reviewers: Do you think we should remove the definitions that are now same as default from the dict? I left it in for now so the code is more explicit, but technically it is not needed.

So just:

```
FileItem.file_colors = defaultdict(lambda: TreeItem.text_color, {
    File.PENDING: interface_colors.get_qcolor('entity_pending'),
    File.ERROR: interface_colors.get_qcolor('entity_error'),
})
```

instead of 

```
FileItem.file_colors = defaultdict(lambda: TreeItem.text_color, {
    File.NORMAL: TreeItem.text_color,
    File.CHANGED: TreeItem.text_color,
    File.PENDING: interface_colors.get_qcolor('entity_pending'),
    File.ERROR: interface_colors.get_qcolor('entity_error'),
})
```

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -17,6 +17,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from collections import defaultdict
 from functools import partial
 from heapq import (
     heappop,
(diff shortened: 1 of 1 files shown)


zas:
Keeping it doesn't hurt, but you could get of it keeping a comment or even the code as comment. No strong opinion about this.

PR #1322, opened 2019-10-13
PICARD-1630: Always save FLAC picture blocks after Vorbis comments

phw:
# Summary
If FLAC picture blocks are located before the Vorbis comments, Windows cannot read the tags. By ensuring the pictures are always located after the Vorbis comments the FLAC metadata will be visible and editable in Windows Explorer.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1630](https://tickets.metabrainz.org/browse/PICARD-1630)

# Solution
Always sort the FLAC metadata blocks so picture blocks are located after the Vorbis comments block. This does not affect the order of any other blocks, also the order of the picture blocks is maintained.

# Action

### DIFF
--- a/picard/formats/vorbis.py
+++ b/picard/formats/vorbis.py
@@ -69,6 +69,32 @@ def is_valid_key(key):
     return INVALID_CHARS.search(key) is None
 
 
+def flac_sort_pics_after_tags(metadata_blocks):
+    """
+    Reorder the metadata_blocks so that all picture blocks are located after
+    the first Vorbis comment block.
+
+    Windows fails to read FLAC tags if the picture blocks are located before
+    the Vorbis comments. Reordering the blocks fixes this.
+    """
+    # First remember all picture blocks that are located before the tag block.
+    tagindex = 0
+    picblocks = []
+    for block in metadata_blocks:
+        if block.code == mutagen.flac.VCFLACDict.code:
+            tagindex = metadata_blocks.index(block)
+            break
+        elif block.code == mutagen.flac.Picture.code:
+            picblocks.append(block)
+    else:
+        return  # No tags found, nothing to sort
+
+    # Now move those picture block after the tag block, maintaining their order.
+    for pic in picblocks:
+        metadata_blocks.remove(pic)
+        metadata_blocks.insert(tagindex, pic)
+
+
 class VCommentFile(File):
 
     """Generic VComment-based file."""
@@ -244,7 +270,7 @@ class VCommentFile(File):
             picture.width = image.width
             picture.height = image.height
             picture.type = image_type_as_id3_num(image.maintype)
-            if self._File == mutagen.flac.FLAC:
+            if is_flac:
                 file.add_picture(picture)
             else:
                 tags.setdefault("METADATA_BLOCK_PICTURE", []).append(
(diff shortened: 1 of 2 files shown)


picard/formats/vorbis.py
```diff
@@ -69,6 +69,27 @@ def is_valid_key(key):
...
+        if b.code == mutagen.flac.VCFLACDict.code:
+            tagblock = b
+            break
+        elif b.code == mutagen.flac.Picture.code:
+            picblocks.append(b)
+    else:
+        # No tags found, nothing to sort
+        return
+
+    # Now move those picture block after the tag block, maintaining their order.
```
zas: > maintaining their order ?

Perhaps I'm missing something here.
List of pictures is built from previous loop, using `append()`, but `insert(index, value)` will actually revere the order:

```python
>>> l = ['a', 'b', 'c']
>>> l
['a', 'b', 'c']
>>> for p in ['pic1', 'pic2', 'pic3']:
...     l.insert(1, p)
... 
>>> l
['a', 'pic3', 'pic2', 'pic1', 'b', 'c']
```

picard/formats/vorbis.py
```diff
@@ -69,6 +69,27 @@ def is_valid_key(key):
     return INVALID_CHARS.search(key) is None
 
 
+def flac_sort_pics_after_tags(metadata_blocks):
+    # First remember all picture blocks that are located before the tag block.
+    tagblock = None
+    picblocks = []
+    for b in metadata_blocks:
+        if b.code == mutagen.flac.VCFLACDict.code:
+            tagblock = b
```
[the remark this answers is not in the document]
phw: Oh yes, that's left over code from previous iteration. Will fix it

picard/formats/vorbis.py
phw: Ha, my first thinking was also that I would need to insert at tagindex + 1 and then increment. But turns out the insert position is stable at tagindex, because one by one the blocks are removedat the start and then inserted again. The removal automatically moves all following elements, making room for the insertion.

phw:
@zas I addressed your comments, please check.

> I'm not fond of fixing windows... but ok.

Me neither, but the majority of our user base is there. Also having (often multiple MB large) picture blocks after the tags isn't the worst of design decisions, this makes the metadata itself available earlier. Even foobar2000 decided of going this route, and they otherwise seem rather reluctant to implement anything too specific to other applications.

picard/formats/vorbis.py
phw: Just to demonstrate what I mean with your example:

```python
>>> l = ['pic1', 'pic2', 'pic3', 'a', 'b', 'c']
>>> l
['pic1', 'pic2', 'pic3', 'a', 'b', 'c']
>>> for p in ['pic1', 'pic2', 'pic3']:
...     l.remove(p)
...     l.insert(3, p)
...
>>> l
['a', 'pic1', 'pic2', 'pic3', 'b', 'c']

```

zas: I got it, missed this remove part ;)

phw:
@zas I added some documentation to `flac_sort_pics_after_tags` to make the purpose of this clear. Otherwise this is unchanged.

PR #1327, opened 2019-10-15
Move Windows AttachConsole handling to PyInstaller launch hook

phw:
# Summary
On Windows we try to attach the GUI app to console output. Move this code to a PyInstaller hook.

Reasoning:

- This is only needed on Windows when running as GUI app, not when running directly from console. 
- Attaching to console should happen as early as possible before anything else gets initialized. E.g. with this hook the console can be attached before other PyInstaller hooks start

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -5,8 +5,8 @@ import os
 import platform
 import sys
 
-# Get the version
-# and build a CFBundleVersion compatible version of it according to Apple dev documentation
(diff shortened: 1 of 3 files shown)


phw:
Tested with PyInstaller on Win 10

PR #1329, opened 2019-10-16
PICARD-1644: Fixed gettext install with Python 3.8

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1644](https://tickets.metabrainz.org/browse/PICARD-1644)

# Solution
Fixed the call to `translation.install()`. We passed an invalid parameter to translation.install() , but this had no effect in earlier Python versions.

See docs at https://docs.python.org/3/library/gettext.html#gettext.NullTranslations.install

# Action

### DIFF
--- a/picard/i18n.py
+++ b/picard/i18n.py
@@ -79,33 +79,19 @@ def setup_gettext(localedir, ui_language=None, logger=None):
     try:
         logger("Loading gettext translation, localedir=%r", localedir)
         trans = gettext.translation("picard", localedir)
-        trans.install()
-        _ngettext = trans.ngettext
         logger("Loading gettext translation (picard-countries), localedir=%r", localedir)
         trans_countries = gettext.translation("picard-countries", localedir)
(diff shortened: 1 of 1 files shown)


zas:
Is there any build/test running on python 3.8 yet?

phw:
> Is there any build/test running on python 3.8 yet?

I had https://github.com/metabrainz/picard/pull/1321 , but I decided to wait for final 3.8 release to update this. Looks like older PyQt versions are not available for Python 3.8, we need to adapt the CI config for this.

Also this specific case was not even captured by CI. I think we should have some tests for i18n.py added and also maybe in all CI runs do a simple `python tagger.py --long-version` call just to verify basic startup is working.

phw:
https://github.com/metabrainz/picard/pull/1330 for the python 3.8 tests

PR #1334, opened 2019-10-25
PICARD-1592: APEv2 tag case insensitive reading

phw:
# Summary
According to specification APEv2 tags should be read case insensitive. Previously Picard only read tags for which there was no explicit mapping case insensitive, but treated all explicitly mapped tags case sensitive.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-1592](https://tickets.metabrainz.org/browse/PICARD-1592)

# Solution
Apply the same case insensitive reading logic as it was previously added for some ID3 and ASF tags, but for APE this applies to all tags.

# Action

### DIFF
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -80,34 +80,34 @@ class APEv2File(File):
     _File = None
 
     __translate = {
-        "Album Artist": "albumartist",
-        "MixArtist": "remixer",
-        "Weblink": "website",
-        "DiscSubtitle": "discsubtitle",
-        "BPM": "bpm",
-        "ISRC": "isrc",
-        "CatalogNumber": "catalognumber",
-        "Barcode": "barcode",
-        "EncodedBy": "encodedby",
-        "Language": "language",
-        "MOVEMENT": "movementnumber",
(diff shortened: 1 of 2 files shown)


picard/formats/apev2.py
```diff
@@ -80,42 +80,43 @@ class APEv2File(File):
...
 
     def _load(self, filename):
         log.debug("Loading file %r", filename)
         file = self._File(encode_filename(filename))
         metadata = Metadata()
         if file.tags:
             for origname, values in file.tags.items():
-                if origname.lower().startswith("cover art") and values.kind == mutagen.apev2.BINARY:
+                origname = origname.lower()
+                if origname.startswith("cover art") and values.kind == mutagen.apev2.BINARY:
```
zas: I'd switch tests order here (but I'm not sure if Python optimizes this or not).

phw: ou are right, that's nicer. I switched that around, also same in the _save method.

PR #1338, opened 2019-10-28
PICARD-1628: Make APEv2 tags for AAC optional

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Only some software supports APEv2 tags in AAC files, for some software it even causes issues with length calculation or playback.

See also discussion at https://github.com/metabrainz/picard/pull/1306

* JIRA ticket (_optional_): PICARD-1628

# Solution
Let the user configure whether to save APEv2 to AAC, not save APEv2 to AAC or completely remove APEv2 from AAC.

The default remains to save APEv2 tags to AAC.

# Action


picard/ui/options/tags.py
```diff
@@ -89,6 +92,12 @@ def load(self):
         self.ui.remove_id3_from_flac.setChecked(config.setting["remove_id3_from_flac"])
         self.ui.itunes_compatible_grouping.setChecked(config.setting["itunes_compatible_grouping"])
         self.ui.preserved_tags.setText(config.setting["preserved_tags"])
+        if config.setting["aac_save_ape"]:
+            self.ui.aac_save_ape.setChecked(True)
+        else:
+            self.ui.aac_no_tags.setChecked(True)
+        self.ui.remove_ape_from_aac.setChecked(config.setting["remove_ape_from_aac"])
+        self.ui.remove_ape_from_aac.setEnabled(not config.setting["aac_save_ape"])
```
zas: Since there's already a test for `aac_save_ape` few lines above, I'd either move this in both branchs of the `if` statement (at the price of slight code duplication), or cache it in a local variable.

phw: Yes, valid point. But I keep it for now, this is pretty common all over the options pages, maybe we should clean them all up. In the specific case of the options code it is really just a visual cleanup, though, it could make the code to look less noisy.

PR #1340, opened 2019-10-30
PICARD-1653: Enable macOS "Hardened Runtime"

phw:
# Summary
Enable macOS "Hardened Runtime", which is a requirement for notarizing 

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1653, PICARD-1631

# Solution
Enable macOS "Hardened Runtime", which is a flag applied during runtime and which places certain restrictions on the signed binaries (enforced by the OS).

From the codesign man page:

> On macOS versions >= 10.14.0, opts signed processes into a hardened runtime environment which includes runtime code signing enforcement, library validation, hard, kill, and debugging restrictions. These restrictions can be selectively relaxed via entitlements. Note: macOS versions older than 10.14.0 ignore the presence of this flag in the code signature.

# Action

### DIFF
--- a/scripts/package-osx.sh
+++ b/scripts/package-osx.sh
@@ -42,7 +42,7 @@ cd dist
 ditto -rsrc --arch x86_64 'MusicBrainz Picard.app' 'MusicBrainz Picard.tmp'
 rm -r 'MusicBrainz Picard.app'
 mv 'MusicBrainz Picard.tmp' 'MusicBrainz Picard.app'
-[ "$CODESIGN" = '1' ] && codesign --keychain $KEYCHAIN_PATH --verify --verbose --deep --sign "$CERTIFICATE_NAME" 'MusicBrainz Picard.app'
+[ "$CODESIGN" = '1' ] && codesign --keychain $KEYCHAIN_PATH --verify --verbose --options runtime --deep --sign "$CERTIFICATE_NAME" "MusicBrainz Picard.app"
 
 # Verify Picard executable works and required dependencies are bundled
 VERSIONS=$("MusicBrainz Picard.app/Contents/MacOS/picard-run" --long-version)


phw:
I reverted this for now, it doesn't work:

1. This requires at least macOS 10.14, on older version codesign won't accept the "runtime" option 
2. On our macOS 10.14 build this also fails, I think when trying to run `picard-run --long-version`:

```
MusicBrainz Picard.app: signed app bundle with Mach-O thin (x86_64) [org.musicbrainz.Picard]

[12790] Failed to execute script pyiboot01_bootstrap

  File "site-packages/PyInstaller/loader/pyiboot01_bootstrap.py", line 127, in <module>

  File "<frozen importlib._bootstrap>", line 983, in _find_and_load

  File "<frozen importlib._bootstrap>", line 967, in _find_and_load_unlocked

  File "<frozen importlib._bootstrap>", line 677, in _load_unlocked

  File "/Users/travis/build/metabrainz/picard/lib/python3.7/site-packages/PyInstaller/loader/pyimod03_importers.py", line 627, in exec_module

    exec(bytecode, module.__dict__)

  File "ctypes/__init__.py", line 551, in <module>

  File "ctypes/__init__.py", line 273, in _reset_cache

MemoryError
```

micahflee:
FYI, I solved this problem for a different PyInstaller app. If you still want to enable hardened runtime, you can see the solution here: https://github.com/pyinstaller/pyinstaller/issues/4629#issuecomment-574375331

phw:
@micahflee Thanks a lot for helping out. But actually we had got this running for a while now ;) We actually did not need `com.apple.security.cs.allow-jit`, it's running fine with just

https://github.com/metabrainz/picard/blob/master/scripts/package/entitlements.plist

micahflee:
Ha, good to know!

PR #1345, opened 2019-11-03
PICARD-1098: Support custom tags for MP4 

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Svaing custom tags to MP4 is not supported.

* JIRA ticket (_optional_): PICARD-1098

# Solution
Save tags to `----:com.apple.iTunes:{tagname}`, where `{tagname}` is treated case insensitive, but case preserving. This is compatible with most software supporting custom MP4 tags.

# Action

### DIFF
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -36,6 +36,11 @@ from picard.metadata import Metadata
 from picard.util import encode_filename
 
 
+def _add_text_values_to_metadata(metadata, name, values):
+    for value in values:
+        metadata.add(name, value.decode("utf-8", "replace").strip("\x00"))
+
+
 class MP4File(File):
     EXTENSIONS = [".m4a", ".m4b", ".m4p", ".m4v", ".mp4"]
     NAME = "MPEG-4 Audio"
@@ -125,7 +130,8 @@ class MP4File(File):
     }
     __r_freeform_tags = dict([(v, k) for k, v in __freeform_tags.items()])
 
-    # Tags to load case insensitive
+    # Tags to load case insensitive. Case is preserved, but the specified case
+    # is written if it is unset.
     __r_freeform_tags_ci = {
         "replaygain_album_gain": "----:com.apple.iTunes:REPLAYGAIN_ALBUM_GAIN",
         "replaygain_album_peak": "----:com.apple.iTunes:REPLAYGAIN_ALBUM_PEAK",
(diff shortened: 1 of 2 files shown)


picard/formats/mp4.py
```diff
@@ -36,6 +38,22 @@
 from picard.util import encode_filename
 
 
+def _add_text_values_to_metadata(metadata, name, values):
+    for value in values:
+        value = value.decode("utf-8", "replace").strip("\x00")
+        metadata.add(name, value)
```
zas: IMHO, it could be written:

```python
metadata.add(name, value.decode("utf-8", "replace").strip("\x00")
```

without having too much readibility impact.

picard/formats/mp4.py
```diff
@@ -197,6 +213,17 @@ def _load(self, filename):
...
                         metadata.images.append(coverartimage)
+            # Read other freeform tags always case insensitive
+            elif name.startswith('----:com.apple.iTunes:'):
+                tag_name = name_lower[22:]
+                self.__casemap[tag_name] = name[22:]
+                if (name not in self.__r_text_tags
+                    and name not in self.__r_bool_tags
+                    and name not in self.__r_int_tags
+                    and name not in self.__r_freeform_tags
+                    and name.lower() not in self.__r_freeform_tags_ci
```
zas: Use `name_lower` instead of `name.lower()` here?

picard/formats/mp4.py
```diff
@@ -274,14 +306,13 @@ def _remove_deleted_tags(self, metadata, tags):
...
-        return (name in cls.__r_text_tags
-                or name in cls.__r_bool_tags
-                or name in cls.__r_freeform_tags
-                or name in cls.__r_freeform_tags_ci
-                or name in cls.__r_int_tags
-                or name in cls.__other_supported_tags
-                or name.startswith('lyrics:')
-                or name in ('~length', 'musicip_fingerprint'))
+        unsupported_tags = ['r128_album_gain', 'r128_track_gain']
+        return ((name and not name.startswith("~")
```
zas: I'd wrap the second part, keeping one condition per line logic:

```python
return ((name
         and not name.startswith("~")
         and name not in unsupported_tags
...
```

phw:
The suggestions all make sense, I fixed it up

PR #1346, opened 2019-11-05
PICARD-1658: Ensure bundled libdiscid used in macOS app

phw:
# Summary
Make sure to use the bundled libdiscid.0.dylib in macOS app.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
On macOS a system wide install of libdiscid could be used instead of the one included with the app. Especially with the changes for using the hardened runtime in PICARD-1653 this can cause the app fail to load the discid support, since apps using the hardened runtime can only load signed libraries, and the system wide install (e.g. installed with brew) probably is not signed.

* JIRA ticket (_optional_): PICARD-1653

# Solution
On macOS the libraries located via `ctypes.util.find_library` are ultimately searched for with `ctypes.macholib.dyld.dyld_find`. 

Set `DYLD_FALLBACK_LIBRARY_PATH` to override the search location. See https://macholib.readthedocs.io/en/latest/dyld.html

# Action

### DIFF
--- a/picard/disc.py
+++ b/picard/disc.py
@@ -4,6 +4,7 @@
 # Copyright (C) 2007 Lukáš Lalinský
 # Copyright (C) 2006 Matthias Friedrich
 # Copyright (C) 2013 Laurent Monin
+# Copyright (C) 2018, 2019 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


picard/disc.py
```diff
@@ -19,6 +20,23 @@
...
+from PyQt5 import QtCore
+
+from picard import log
+from picard.const.sys import IS_MACOS
+
+from picard.ui.cdlookup import CDLookupDialog
+
+
+# Ensure libdiscid.dylib gets loaded from app bundle
+if IS_MACOS and getattr(sys, 'frozen', False):
```
zas: You can also use `IS_FROZEN` from https://github.com/metabrainz/picard/blob/master/picard/const/sys.py#L30

phw: Thanks, changed it

picard/disc.py
```diff
@@ -19,6 +20,26 @@
...
+from picard.const.sys import (
+    IS_FROZEN,
+    IS_MACOS,
+)
+
+from picard.ui.cdlookup import CDLookupDialog
+
+
+# Ensure libdiscid.dylib gets loaded from app bundle
+if IS_MACOS and IS_FROZEN:
```
samj1912: Should this be a runtime hook?

phw: I guess we could move it to a runtime hook, yes.

samj1912: It would be great if we could do that. Move as much freezing logic to hooks. That way we can keep the module logic relatively unpolluted with pyinstaller logic.

phw: Here we go: https://github.com/metabrainz/picard/pull/1348

PR #1350, opened 2019-11-08
PICARD-1661: Improve preserved tags auto completion

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The "Preserved tags" input in Options > Tags has an autocomplete feature, suggesting tags.

But this does not work too well:

- Completion only triggers, if you don't have spaces after the comma. So writing "artist, t" will not trigger suggestions starting with "t", only "artist,t" does
- Once a completion has been selected it is entered into the edit field, but the user has to manually add the comma to enter another one. It would make sense to enter the comma for them so it is easy to enter multiple tags

* JIRA ticket (_optional_): PICARD-1661

# Solution
- Ignore leading / trailing whitespace in autocomplete prefix
- Add ", " to end of input when inserting suggestion

# Action

### DIFF
--- a/picard/ui/options/tags.py
+++ b/picard/ui/options/tags.py
@@ -150,11 +150,16 @@ class TagsOptionsPage(OptionsPage):
     def completer_activated(self, text):
         input_field = self.ui.preserved_tags
         current = input_field.text()
-        i = input_field.cursorPosition()
-        p = len(self.completer.completionPrefix())
-        start_text = current[:i - p].rstrip()
-        input_field.setText("%s %s, %s" % (start_text, text, current[i:].lstrip()))
-        input_field.setCursorPosition(len(start_text) + len(text) + 3)
+        cursor_pos = input_field.cursorPosition()
+        prefix_len = len(self.completer.completionPrefix())
+        leading_text = current[:cursor_pos - prefix_len].rstrip()
+        trailing_text = current[cursor_pos:].lstrip()
+        # Replace the autocompletion prefix with the autocompleted text,
+        # append a comma so the user can easily enter the next entry
+        replacement = ("%s %s, " % (leading_text, text)).lstrip()
+        input_field.setText(replacement + trailing_text)
+        # Set cursor position to end of autocompleted input
+        input_field.setCursorPosition(len(replacement))
 
 
 register_options_page(TagsOptionsPage)
--- a/picard/ui/options/tags.py
+++ b/picard/ui/options/tags.py
@@ -18,6 +18,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from functools import partial
+import re
 
 from PyQt5 import (
     QtCore,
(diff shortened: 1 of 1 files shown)


phw:
This is how it works:

Peek 2019-11-08 15-33

Feels much nicer and more usable.

picard/ui/options/tags.py
```diff
@@ -119,7 +120,7 @@ def save(self):
         config.setting["remove_ape_from_mp3"] = self.ui.remove_ape_from_mp3.isChecked()
         config.setting["remove_id3_from_flac"] = self.ui.remove_id3_from_flac.isChecked()
         config.setting["itunes_compatible_grouping"] = self.ui.itunes_compatible_grouping.isChecked()
-        config.setting["preserved_tags"] = self.ui.preserved_tags.text()
+        config.setting["preserved_tags"] = re.sub(r"[,\s]+$", "", self.ui.preserved_tags.text())
```
samj1912: re.compile somewhere outside?

picard/ui/options/tags.py
```diff
@@ -151,8 +152,9 @@ def completer_activated(self, text):
         current = input_field.text()
         i = input_field.cursorPosition()
         p = len(self.completer.completionPrefix())
-        input_field.setText("%s%s %s" % (current[:i - p], text, current[i:]))
-        input_field.setCursorPosition(i - p + len(text) + 1)
+        start_text = current[:i - p].rstrip()
+        input_field.setText("%s %s, %s" % (start_text, text, current[i:].lstrip()))
+        input_field.setCursorPosition(len(start_text) + len(text) + 3)
```
samj1912: Why +3? Maybe add a comment explaining this so that this doesn't go out of date in the future.

picard/ui/options/tags.py
phw: I thought about it, but don't think so. This is a simple regex that doesn't get run frequently and then only once. And I'd rather keep the code here in one place in this case.

picard/ui/options/tags.py
```diff
@@ -151,8 +152,9 @@ def completer_activated(self, text):
         current = input_field.text()
         i = input_field.cursorPosition()
         p = len(self.completer.completionPrefix())
-        input_field.setText("%s%s %s" % (current[:i - p], text, current[i:]))
-        input_field.setCursorPosition(i - p + len(text) + 1)
+        start_text = current[:i - p].rstrip()
```
samj1912: I know that this isn't your code, but would it be possible to rename these (i, p) to something that explains what's happening here?

picard/ui/options/tags.py
phw: The 3 are the extra characters added: It's the length of the existing text before autocomplete + length of the autocompleted text + the three characters (space before the autocomplete, comma and space after).

I tried to make this whole code more readable in https://github.com/metabrainz/picard/pull/1350/commits/feca8d80b38c2ac9cc7173bfc7f54ea36d9a39c6 . This also gets rid of the magic constant.

PR #1351, opened 2019-11-08
PICARD-1631: Notarize macOS app

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Starting with macOS 10.15 Catalina Apple requires apps to be submitted to their malware scan service aka notarization.

* JIRA ticket (_optional_): PICARD-1631

# Solution
Add a script `macos-notarize-app.sh`, which submits the app for notarization, polls the notarization status and finally "staples" the notarization ticket to the app.

The last stapling part is necessary so that the app can be verified without internet connection. If the ticket is not stapled macOS will try to query the notarization status online.

# Action
Notarizing the app does only work for apps using the Hardened Runtime. The developer tools to sign with hardened runtime and to submit for notarization are only available with newer xcode, which means we need to build with macOS 10.14 on Travis CI. We need to check whether the build app will also run on older versions (10.12 or 10.13). If not, we might need to consider providing two builds (one for 10.12 or later and a notarized one for 10.14 or later).


phw:
Successfuly notarized app: https://s3.eu-central-1.amazonaws.com/artifacts.picard.uploadedlobster.com/metabrainz/picard/5235/MusicBrainz%20Picard%202.3.0.dev1_osx_xcode11.2_20191108234016%20macOS%2010.14.dmg

Tested on macOS 10.15, passes the security check without issues.

scripts/package/macos-package-app.sh
```diff
@@ -41,21 +42,29 @@ if [ -f scripts/package/appledev.p12 ] && [ -n "$appledev_p12_password" ]; then
     CODESIGN=1
 fi
 
+# Submit app for notarization on macOS >= 10.14
+if [ $MACOS_VERSION_MAJOR -ge 10 ] && [ $MACOS_VERSION_MINOR -ge 14 ]; then
```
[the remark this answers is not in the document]
mineo: Let's not worry about that now - comparing version numbers on macos is not easy and Apple has stuck with 10.x for nearly 20 years now ;-)

scripts/package/macos-notarize-app.sh
```diff
@@ -0,0 +1,74 @@
...
+  -itc_provider MetaBrainzFoundationInc \
+  --output-format xml)
+
+if [ $? -ne 0 ]; then
+  echo "Submitting $APP_BUNDLE failed:"
+  echo $RESULT
+  exit 1
+fi
+
+REQUEST_UUID=$(echo $RESULT | xpath \
```
mineo: Is xpath preinstalled on macs?

scripts/package/macos-package-app.sh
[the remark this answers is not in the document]
phw: Agree with both of you: We shouldn't worry about a potential macOS 11.x yet (which probably would have changes that make this assumption in the script wrong anyway), but at the same time the condition that way is not correct. I fixed it by comparing for equality, so now it is 10.14 >= current version < 11.0

scripts/package/macos-notarize-app.sh
phw: I don't know if it is installed by default on any macOS or if it is part of the development tools getting installed with xcode. But I haven't installed it manually, and it is available on my install (which is a pretty bare bone macOS 10.15 with xcode installed on top) and also on the CI servers. That's all that really matters :D

scripts/package/macos-notarize-app.sh
```diff
@@ -0,0 +1,74 @@
...
+ditto -c -k --rsrc --keepParent "$APP_BUNDLE" "${APP_BUNDLE}.zip"
+
+# Submit for notarization
+echo "Submitting $APP_BUNDLE for notarization..."
+RESULT=$(xcrun altool --notarize-app --type osx \
+  --file "${APP_BUNDLE}.zip" \
+  --primary-bundle-id org.musicbrainz.Picard \
+  --username $APPLE_ID_USER \
+  --password @env:APPLE_ID_PASSWORD \
+  -itc_provider MetaBrainzFoundationInc \
```
zas: Other options are using double dash, but not this one, expected?

scripts/package/macos-notarize-app.sh
```diff
@@ -0,0 +1,74 @@
...
+
+if [ -z "$APPLE_ID_USER" ] || [ -z "$APPLE_ID_PASSWORD" ]; then
+    echo "You need to set your Apple ID credentials with \$APPLE_ID_USER and \$APPLE_ID_PASSWORD."
+    exit 1
+fi
+
+APP_BUNDLE=$(basename "$1")
+APP_BUNDLE_DIR=$(dirname "$1")
+
+cd "$APP_BUNDLE_DIR"
```
zas: Perhaps append ` || exit 1` in case `cd` fails

scripts/package/macos-package-app.sh
```diff
@@ -41,21 +42,29 @@ if [ -f scripts/package/appledev.p12 ] && [ -n "$appledev_p12_password" ]; then
     CODESIGN=1
 fi
 
+# Submit app for notarization on macOS >= 10.14
+if [ $MACOS_VERSION_MAJOR -eq 10 ] && [ $MACOS_VERSION_MINOR -ge 14 ]; then
```
zas: Is it intentional it doesn't match macos >= 11.0 ?

scripts/package/macos-notarize-app.sh
phw: Yes, that's how this is spelled. Don't ask why. Unlike the other options this option also doesn't show up if you run `xcrun altool --help` 🤷‍♂️

I had to Google a bit to find it. The reason why we need it is that my Apple Developer account is associated with more than one subscription. This parameter selects the proper subscription.

scripts/package/macos-package-app.sh
phw: Yes, see discussion with samj and mineo above. Basically there is no real need to complicate the check here.

scripts/package/macos-notarize-app.sh
phw: Did that

phw:
I merged this now to get this into use. Not quite clear to me is the compatibility with macOS 10.12 / 10.13, we need to verify this. I asked for feedback on https://community.metabrainz.org/t/looking-for-testers-of-picard-build-for-macos-10-15-catalina/450650/

Should the current notarized builds not be compatible with macOS < 10.14 we either need to:

1. Use the builds done on 10.12, then maybe manually (re-)sign (for the hardened runtime) and notarize them locally
2. Provide two builds, one for 10.12 and later (would run on 10.15, but with the security dialog) and one for 10.14 and later

But let's see first, my favorite outcome would be that we can use the 10.14 builds for all cases.

PR #1352, opened 2019-11-11
PICARD-1406: Make setting data in Metadata consistently use strings

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
The methods of `Metadata` which set data are inconsistent. `__setitem__` and `set` behave differently and only `__setitem__` did conversion to string. This is a confusing API. Also because tag values are treated as strings everywhere this lead to a lot of explicit `str()` conversions when using `add` or `add_unique`.

* JIRA ticket (_optional_): PICARD-1406

# Solution
- All Metadata methods setting data (`__setitem__`, `set`, `add`, `add_unique`) will convert values to strings
- `__setitem__` and `set` are equivalent
- get rid of explicit `str()` conversions when calling `add` / `add_unique`

I searched both the Picard and picard-plugins code for uses of `set`, `add` and `add_unique`. In all cases code was either passing already an string or did str conversion. So while this is a potentially breaking change I don't think it should break anything official. And if some plugin really stores some non-string value in Metadata this is already very hackish and easily breaks when combined with scripting.

# Action

### DIFF
--- a/test/test_metadata.py
+++ b/test/test_metadata.py
@@ -53,6 +53,20 @@ class MetadataTest(PicardTestCase):
         self.assertEqual(self.multi3, self.metadata.getraw("multi3"))
         self.assertEqual(["hidden-value"], self.metadata.getraw("~hidden"))
 
+    def test_metadata_set_all_values_as_string(self):
+        for val in (0, 2, True):
+            str_val = str(val)
+            self.metadata.set('val1', val)
+            self.assertEqual([str_val], self.metadata.getraw("val1"))
+            self.metadata['val2'] = val
+            self.assertEqual([str_val], self.metadata.getraw("val2"))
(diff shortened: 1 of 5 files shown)


test/test_metadata.py
```diff
@@ -87,13 +87,6 @@ def test_metadata_implicit_delete(self):
         self.assertNotIn("unknown", self.metadata)
         self.assertNotIn("unknown", self.metadata.deleted_tags)
 
-    def test_metadata_set_explicit_empty(self):
```
phw: I could not find any usage of this in real code, so test gets dropped. Interestingly all other tests continue to work without issue, which kind of shows how the "values are strings" assumption is really inherit to the code.

PR #1354, opened 2019-11-12
PICARD-1626: Translate and sort languages for metadata translation dropdown

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The language dropdown in Options > Metadata is not translated. It also gets sorted purley by ISO language code, which leads to confusing ordering.

* JIRA ticket (_optional_): PICARD-1626

# Solution
Translate the languages and sort the entries by translated name.

# Action


picard/ui/options/metadata.py
```diff
@@ -58,14 +75,14 @@ def load(self):
...
-        locales = sorted(ALIAS_LOCALES.keys())
-        for i, loc in enumerate(locales):
-            name = ALIAS_LOCALES[loc]
-            if "_" in loc:
-                name = "    " + name
-            combo_box.addItem(name, loc)
-            if loc == config.setting["artist_locale"]:
+        current_locale = config.setting["artist_locale"]
+        i = 0
+        for locale, name, level in iter_sorted_locales(ALIAS_LOCALES):
```
zas: We could perhaps use `enumerate()` here:

```python
for i, (locale, name, level) in enumerate(iter_sorted_locales(ALIAS_LOCALES)):
```

PR #1356, opened 2019-11-12
PICARD-1668: Added checks for actual value presence in "trkn" and "disk" mp4 tags

reinaldorauch:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Added checks to the actual presence of values in mp4 tags "trkn" and "disk". If not present, just ignore and don't set it into parsed metadata

# Problem

When importing a file with these tags set but with empty values picard breaks with an exception and marks as errored.

Steps to reproduce:

1. Make a m4a/mp4/acc file with these tags present but with empty values
2. Try to import them in picard

* JIRA ticket : PICARD-1668

# Solution

Add checks of actual value presence in them

# Action

~~I will add tests as I just found them for formats/mp4.~~
Ready to merge

### DIFF
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -191,15 +191,17 @@ class MP4File(File):
                     if value.startswith("MusicMagic Fingerprint"):
                         metadata.add("musicip_fingerprint", value[22:])
             elif name == "trkn":
-                if (len(values) >= 1 and len(values[0] >= 1)):
+                try:
                     metadata["tracknumber"] = values[0][0]
-                    if (len(values[0])>=2):
-                        metadata["totaltracks"] = values[0][1]
+                    metadata["totaltracks"] = values[0][1]
+                except:
+                    pass
             elif name == "disk":
-                if (len(values) >= 1 and len(values[0] >= 1)):
+                try:
                     metadata["discnumber"] = values[0][0]
-                    if (len(values[0])>=2):
-                        metadata["totaldiscs"] = values[0][1]
+                    metadata["totaldiscs"] = values[0][1]
+                except:
+                    pass
             elif name == "covr":
                 for value in values:
                     if value.imageformat not in (value.FORMAT_JPEG,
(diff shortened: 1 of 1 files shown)


picard/formats/mp4.py
```diff
@@ -191,11 +191,15 @@ def _load(self, filename):
                     if value.startswith("MusicMagic Fingerprint"):
                         metadata.add("musicip_fingerprint", value[22:])
             elif name == "trkn":
-                metadata["tracknumber"] = values[0][0]
-                metadata["totaltracks"] = values[0][1]
+                if (len(values) >= 1 and len(values[0] >= 1)):
+                    metadata["tracknumber"] = values[0][0]
+                    if (len(values[0])>=2):
+                        metadata["totaltracks"] = values[0][1]
```
phw: Two things:
1. Check the placing of parantheses in `len(values[0] >= 1)`
2. Please add spacing around `>=`

I also find this part of the code a bit verbose with the many `len`. What about rewriting it to something like:

```python
elif name == "trkn" and len(values):
    count = len(values[0])
    if count > 0:
        metadata["tracknumber"] = values[0][0]
        if count > 1:
            metadata["totaltracks"] = values[0][1]
```

Same things apply to disk.

zas: It would be more pythonic to just catch the raised exception.

reinaldorauch: @zas I tryied that first but i thought that it would be better to use ifs. But I will return to that as I think it will be more easy to read

reinaldorauch:
I will not be able to provide tests because I could not make a file that could reproduce the error

picard/formats/mp4.py
reinaldorauch: Done.

reinaldorauch:
I saw the error in Codacy, I will fix that more later today

PR #1362, opened 2019-11-17
PICARD-1662: Do not save too large images to FLAC

phw:
# Summary
Save FLAC files even if images are too large to embed.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
FLAC has a maximum block size of 16 MB. Currently trying to save FLAC files with images that are larger than this gives a saving error. It is impossible to save the files at all.

* JIRA ticket (_optional_): PICARD-1662

# Solution
Check the expected block size on save and only save images if they have a valid size. Log errors otherwise.

This is a fix for the immediate saving issue reported in PICARD-1662. This is not supposed to be a full solution. As noted in PICARD-1662 it would be preferable if Picard could detect this condition earlier and provide builtin resizing capabilities. This is tracked in PICARD-1663 and related tickets.

# Action

### DIFF
--- a/picard/formats/vorbis.py
+++ b/picard/formats/vorbis.py
@@ -49,6 +49,7 @@ from picard.util import (
 )
 
 
+FLAC_MAX_BLOCK_SIZE = 2 ** 24 - 1  # FLAC block size is limited to a 24 bit integer
 INVALID_CHARS = re.compile('([^\x20-\x7d]|=)')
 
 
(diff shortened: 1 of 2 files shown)


phw:
I added a test case

Sophist-UK:
This is a good catch, but - assuming that it is possible - it would be even better to recognise an album has a track which is a FLAC file and to download an image size which was compatible (in addition to the image size in settings.

phw:
@sophist-uk see related tickets

Sophist-UK:
Ah sorry - should have read the description first.

PR #1369, opened 2019-11-20
PICARD-1674: Linux desktop progress indicator using Unity.LauncherEntry DBus interface

phw:
# Summary
Same functionality as implented on Windows in #1368, but for Linux desktops supporting the DBus based [Unity Launcher API](https://wiki.ubuntu.com/Unity/LauncherAPI).

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1674

# Solution
Expose the [`com.canonical.Unity.LauncherEntry` DBus interface](https://wiki.ubuntu.com/Unity/LauncherAPI#Low_level_DBus_API:_com.canonical.Unity.LauncherEntry). The DBus interface is initiated always except for platforms where it does not make sense. If DBus is not available this is handled by QDBus, the `bus.registerObject` returns False in this case and the `UnityLauncherEntryStatusIndicator` skips any updates.

This was tested on:
- Ubuntu with default Ubuntu GNOME desktop (which is using Dash to Dock)
- KDE Plasma
- Arch Linux with GNOME and Dash to Dock installed
- Plank dock (also installed on GNOME)

For screenshots see https://community.metabrainz.org/t/wishlist-progress-shown-in-windows-taskbar/447056/6

# Action

### DIFF
--- a/picard/ui/statusindicator.py
+++ b/picard/ui/statusindicator.py
@@ -41,9 +41,9 @@ class AbstractProgressStatusIndicator:
         if total_pending == self._last_pending:
             return  # No changes, avoid update
 
+        previous_done = self._max_pending - self._last_pending
+        self._max_pending = max(self._max_pending, previous_done + total_pending)
         self._last_pending = total_pending
-        if total_pending > self._max_pending:
-            self._max_pending = total_pending
 
         if total_pending == 0 or self._max_pending <= 1:  # No need to show progress for single item
             self._max_pending = 0
--- a/picard/ui/statusindicator.py
+++ b/picard/ui/statusindicator.py
@@ -17,37 +17,165 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from picard.const.sys import IS_WIN
+from picard.const.sys import (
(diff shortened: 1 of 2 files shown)


picard/ui/statusindicator.py
```diff
@@ -17,37 +17,167 @@
...
+
+        total_pending = pending_files + pending_requests
+
+        if total_pending == self._last_pending:
+            return  # No changes, avoid update
+
+        previous_done = self._max_pending - self._last_pending
+        new_max = previous_done + total_pending
+        if new_max > self._max_pending:
+            self._max_pending = new_max
```
zas: `previous_done` and `new_max` intermediate variables are only used here, and while they make the code perhaps easier to understand, imho this code is simple enough to allow to replace those 4 lines by:

```python
self._max_pending = max(self._max_pending, self._max_pending - self._last_pending + total_pending)
```

phw: Good point. But getting rid of previous_done harms readability quite a bit, I changed it to:

```
previous_done = self._max_pending - self._last_pending
self._max_pending = max(self._max_pending, previous_done + total_pending)
```

I think that is quite readable.

PR #1370, opened 2019-11-21
PICARD-1675: Disable autoupdate works for pyinstaller and local builds

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Disabling the auto update during build with "python setup.py build --disable-autoupdate" currently has no affect to Windows or macOS builds.

The reason is that those builds use tagger.py as the start script, and not scripts/picard.

* JIRA ticket (_optional_): PICARD-1675

# Solution
Generate tagger.py from tagger.py.in during build and apply the auto update according to build config.

# Action

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -28,6 +28,7 @@ picard.egg-info
 po/attributes/attributes.pot
 po/countries/countries.pot
 scripts/picard
+tagger.py
 win-version-info.txt
 
 # Other files
--- a/.travis.yml
+++ b/.travis.yml
@@ -91,7 +91,8 @@ before_install:
 
 script:
 - py.test -v --cov=picard --cov-report xml:coverage.xml test
-- python3 setup.py build_locales build_ext -i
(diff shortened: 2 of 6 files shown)


zas:
@phw After this patch, `tagger.py` isn't available in source tree by default, and [INSTALL](https://github.com/metabrainz/picard/blob/master/INSTALL.md#running-from-the-source-tree) wasn't updated to reflect those changes.

phw:
@zas Yes, needs to be updated to run setup.py build before

phw:
> After this patch, `tagger.py` isn't available in source tree by default, and [INSTALL](https://github.com/metabrainz/picard/blob/master/INSTALL.md#running-from-the-source-tree) wasn't updated to reflect those changes.

https://github.com/metabrainz/picard/pull/1375

PR #1371, opened 2019-11-25
PICARD-714: User defined top tags

phw:
# Summary
Make the list of tags, which are displayed always on top in the metadata view, configurable.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Picard maintains a hard coded list of tags that are always displayed on the top, make this user configurable.

* JIRA ticket (_optional_): PICARD-714

# Solution
Add an option `metadatabox_top_tags` to hold a list of tags to always be displayed on top, defaulting to the previously hardcoded tag list. Add interface Options > User Interface > Top Tags to configure this list.

Most of the work was to create a custom `EditableTagListView` based on `QListView`. The view

- Allows to edit a list of tags
- Tags are displayed using their display name if available
- Auto completion for known tags
- The list can be sorted by drag and drop or with arrow buttons
- Existing items can be edited by double clicking or using F2 button (or whatever Qt considers the default edit key on the platform)
- A new entry can be added by button or by double clicking below existing tags
- Tags can be deleted by button or with delete key

Screenshot:

grafik

Here is a short screencast:

https://phwolfer.keybase.pub/musicbrainz/picard-tag-list-editor.mp4

I am not totally happy with buttons placement yet, feedback and suggestions welcome. But we could reuse this component for other places were we need to handle an editable list of tags, e.g. "preserve these tags" or the proposed PICARD-1677

# Action


picard/ui/taglistview.py
```diff
@@ -0,0 +1,223 @@
...
+                | QtCore.Qt.ItemIsDragEnabled
+                | QtCore.Qt.ItemIsEnabled
+                | QtCore.Qt.ItemNeverHasChildren)
+        else:
+            return QtCore.Qt.ItemIsDropEnabled
+
+    def insertRows(self, row, count, parent=QtCore.QModelIndex()):
+        super().beginInsertRows(parent, row, row + count - 1)
+        for i in range(count):
+            self._tags.insert(row, ("", ""))
```
zas: I'm bothered by this loop.
What about, something like:

```python
self._tags = self._tags[:row] + [("", "")] * count + self._tags[row:]
```
I didn't test, nor did any speed comparaison, but it gets rid of unused loop variable, and match the style of what you did in `removeRows()` just below.

picard/ui/taglistview.py
```diff
@@ -0,0 +1,223 @@
...
+        for row in sorted(rows, reverse=direction > 0):
+            new_index = model.index(row + direction, 0)
+            model.move_row(row, new_index.row())
+            selection.select(new_index, QtCore.QItemSelectionModel.Select)
+            if row == current_index.row():
+                selection.setCurrentIndex(new_index, QtCore.QItemSelectionModel.Current)
+
+
+class TagListModel(QtCore.QAbstractListModel):
+    def __init__(self, tags=[], parent=None):
```
zas: https://docs.python-guide.org/writing/gotchas/#mutable-default-arguments

picard/ui/taglistview.py
phw: Not a big fan of this to be honest, I find this list construction not very readable. And if this is about performance: In practice this loop won't run more than once since we don't have any way to insert multiple tags. It's just the interface that requires this implementation.

picard/ui/taglistview.py
```diff
@@ -0,0 +1,223 @@
...
+    QtWidgets,
+)
+
+from picard.util.tags import (
+    TAG_NAMES,
+    display_tag_name,
+)
+
+
+class EditableTagListView(QtWidgets.QListView):
```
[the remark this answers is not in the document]
phw: Actually al the tag specifics are in the model, yes. So if we need this somewhere else we can easily adapt this.

phw:
> I'm bothered by the use of tag "human" names in list, but actual tag name in completion. Perhaps we should display both.

Not sure, don't like that very much. We don't do this elsewhere either. I think the current behavior basically matches how the user interacts with the metadata editor: Display names are displayed, internal names are shown when changing.

picard/ui/taglistview.py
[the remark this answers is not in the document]
phw: Currently I mostly see this for other tags lists we have. My plan actually would be to refactor the entire component including buttons to be reusable, but I would like to do this when we actually use it again in another PR.

PR #1373, opened 2019-11-26
Use GitHub Actions for CI

phw:
# Summary
Run tests and packaging / deployment with Github Actions

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

# Solution
This performs all our test, build and deployment tasks we have running on travis / appveyor into Github actions.

Advantages:
- All tasks in one place
- Unified and simpler configuration. E.g. the majority of code in `run-tests.yml` is shared between operating systems (there is only a minor difference for installing gettext on Linux and macOS)
- Packaging of the macOS and Windows apps is pretty similar
- We get artifacts storage for all builds (previously only for appveyor)
- It is more flexible, since Github Actions is more than just for CI. E.g. we can react to changes on releases
- It runs faster (full run with all builds and tests is finished in under 10 minutes)

Disadvantages:
- Only newer versions of OS images are available. E.g. there is only a brand new Catalina 10.15. As far as I have tested we should be fine with current builds (macOS builds should run on our current target of macOS >= 10.12, Windows builds work on Windows 7 still). But currently it is not quite clear how Github will deal with updates in the future (e.g. how long they will keep existing versions of images).

# Action
Once we merge this I would like to disable the triggers for the appveyor and travisci builds. If needed we could keep the configs around for a while. But reactivating the old builds would be no big deal either way.

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -1,0 +1,32 @@
+name: Package for macOS
+
+on: [push]
+
+jobs:
+  package:
+
+    runs-on: macos-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Setup macOS build environment
+      run: |
+        ./scripts/package/macos-setup.sh
+      env:
+        DISCID_VERSION: 0.6.2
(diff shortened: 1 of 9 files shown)


phw:
Closing for now, I move this to a branch on this repository so the builds actually trigger

PR #1374, opened 2019-11-26
Use GitHub Actions for CI

phw:
# Summary
Run tests and packaging / deployment with Github Actions

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

# Solution
This performs all our test, build and deployment tasks we have running on travis / appveyor into Github actions.

Advantages:
- All tasks in one place
- Unified and simpler configuration. E.g. the majority of code in `run-tests.yml` is shared between operating systems (there is only a minor difference for installing gettext on Linux and macOS)
- Packaging of the macOS and Windows apps is pretty similar
- We get artifacts storage for all builds (previously only for appveyor)
- It is more flexible, since Github Actions is more than just for CI. E.g. we can react to changes on releases
- It runs faster (full run with all builds and tests is finished in under 10 minutes)

Disadvantages:
- Only newer versions of OS images are available. E.g. there is only a brand new Catalina 10.15. As far as I have tested we should be fine with current builds (macOS builds should run on our current target of macOS >= 10.12, Windows builds work on Windows 7 still). But currently it is not quite clear how Github will deal with updates in the future (e.g. how long they will keep existing versions of images).

# Action
Once we merge this I would like to disable the triggers for the appveyor and travisci builds. If needed we could keep the configs around for a while. But reactivating the old builds would be no big deal either way.

### DIFF
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -20,6 +20,18 @@ jobs:
       env:
         DISCID_VERSION: 0.6.2
         FPCALC_VERSION: 1.4.3
+    - name: Prepare code signing certificate
+      run: |
+        pip install awscli
+        aws s3 cp "$Env:CODESIGN_PFX_URL" .\codesign.pfx
+        $CertPassword = ConvertTo-SecureString -String $Env:CODESIGN_PFX_PASSWORD -Force -AsPlainText
+        Import-PfxCertificate -CertStoreLocation Cert:\CurrentUser\My -FilePath .\codesign.pfx -Password $CertPassword
+      env:
+        AWS_DEFAULT_REGION: eu-central-1
+        AWS_ACCESS_KEY_ID: ${{ secrets.AWS_ACCESS_KEY_ID }}
+        AWS_SECRET_ACCESS_KEY: ${{ secrets.AWS_SECRET_ACCESS_KEY }}
+        CODESIGN_PFX_URL: ${{ secrets.CODESIGN_PFX_URL }}
+        CODESIGN_PFX_PASSWORD: ${{ secrets.CODESIGN_PFX_PASSWORD }}
     - name: Install dependencies
       run: |
         python -m pip install --upgrade pip
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -1,6 +1,22 @@
 name: Package for Windows
 
-on: [push]
+on:
+  push:
+    paths:
+    - '.github/workflows/package-windows.yml'
+    - 'installer/**'
+    - 'picard/**'
+    - 'po/**.po'
+    - 'resources/win10/**'
+    - 'scripts/package/*.ps1'
+    - 'scripts/pyinstaller/*'
+    - 'appxmanifest.xml.in'
+    - 'picard.ico'
+    - 'picard.spec'
+    - 'requirements*.txt'
+    - 'setup.py'
+    - 'tagger.py.in'
+    - 'win.version-info.txt.in'
 
 jobs:
   package:
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -1,0 +1,43 @@
+name: Package for Windows
+
+on: [push]
+
+jobs:
+  package:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
+    - name: Setup Windows build environment
+      run: |
+        & .\scripts\package\win-setup.ps1 -DiscidVersion $Env:DISCID_VERSION -FpcalVersion $Env:FPCALC_VERSION
+      env:
+        DISCID_VERSION: 0.6.2
+        FPCALC_VERSION: 1.4.3
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements-build.txt
+        pip install -r requirements-win.txt
+    - name: Build Windows 10 app package
+      run: |
+        $Env:PATH += ";C:\Program Files (x86)\Windows Kits\10\bin\10.0.18362.0\x64"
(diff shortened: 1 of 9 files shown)


samj1912:
I am not sure if we want to build on every push. I would rather test on branch pushes and build artifacts only if the tests pass.

samj1912:
Can we also add action ignores for any documentation changes?

samj1912:
I would also cache the dependencies.

phw:
> I would also cache the dependencies.

I added caching for pip (see https://github.com/phw/picard/commit/57043c9877518aff00b9d5a120394e689f6d1919), but I reverted it for this PR. For the packages we install there seems to be no real difference, and the caching overhead is at least as large as the marginal savings that might happen. Compare these two builds:

No caching: https://github.com/phw/picard/runs/322548902
Caching with hit: https://github.com/phw/picard/runs/322595415

> Can we also add action ignores for any documentation changes?

I added a list of paths for the workflows. For the macOS / Windows packaging workflows I was strict and added whitelists, this avoid getting those large builds run on unrelated changes. For the test run I added a couple of paths to be ignored. See https://github.com/metabrainz/picard/pull/1374/commits/a4c123bc2593f0a44b7caf4936e3fe0a6468aea3

> I am not sure if we want to build on every push. I would rather test on branch pushes and build artifacts only if the tests pass.

I originally wondered why you had set up the macOS / Windows builds on Appveyor / Travis to happen before the tests. But it turns out this actually was useful in the past, and I actually want to have the builds happen independent on whether tests pass. For one this shows issues with the builds early, so it's easier to address issues in both tests and packaging (instead of having to go through another CI loop). And also being able to run the binaries often is useful even if parts of the test suite fail.

I think with the whitelist approach we limit the number of times the package builds happen, that reduces the overhead for cases where no changes occurred. It makes it a bit more difficult to find the last build packages in some cases (you have to look through previous build commits until you find one which had packages build), but I can live with this.

If the concern is about build times I think we have some options for the macOS builds to become a bit faster. E.g. the brew update takes rather long and does not seem to be necessary. Also notarization polling could happen a bit more frequently. I will experiment with this.

phw:
> For the test run I added a couple of paths to be ignored. See [a4c123b](https://github.com/metabrainz/picard/commit/a4c123bc2593f0a44b7caf4936e3fe0a6468aea3)

Maybe we should not add any restrictions when to run the standard test suite. If we want to have the tests required before merging a PR merging will be block if the test have not been run. Also those runs don't take too long.

.github/workflows/package-windows.yml
```diff
@@ -0,0 +1,99 @@
...
+  package:
+
+    runs-on: windows-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - name: Set up Python 3.7
+      uses: actions/setup-python@v1
+      with:
+        python-version: 3.7
```
samj1912: Any reason we are using 3.7 for packaging?

.github/workflows/package-windows.yml
```diff
@@ -0,0 +1,99 @@
...
+      run: |
+        $CertPassword = ConvertTo-SecureString -String $Env:CODESIGN_PFX_PASSWORD -Force -AsPlainText
+        $Certificate = Get-PfxCertificate -FilePath .\codesign.pfx -Password $CertPassword
+        & .\scripts\package\win-package-portable.ps1 -BuildNumber $(git rev-list --count HEAD) -Certificate $Certificate
+        Move-Item .\dist\*.exe .\artifacts
+      env:
+        CODESIGN_PFX_PASSWORD: ${{ secrets.CODESIGN_PFX_PASSWORD }}
+    - name: Archive production artifacts
+      uses: actions/upload-artifact@v1
+      if: always()
```
samj1912: Should this be always? Or only on success?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+      run: |
+        python -m pip install --upgrade pip
+        pip install -r requirements.txt
+    - name: Check coding style
+      run: |
+        pip install flake8 isort
+        flake8 picard --count --show-source --statistics
+        isort --check-only --diff --recursive picard test
+    - name: Test with pytest
+      if: always()
```
samj1912: Again, this should only run if pip install succeeds. Why is it always run?

.github/workflows/run-tests.yml
```diff
@@ -0,0 +1,65 @@
...
+      run: sudo apt-get install gettext
+    - name: Install gettext (macOS)
+      if: runner.os == 'macOS'
+      run: |
+        brew install gettext
+        brew link gettext --force
+    - name: Run pip install .
+      run: |
+        python -m pip install --upgrade pip
+        pip install .
```
samj1912: After installing should we print out the version string just as a test?

.github/workflows/package-windows.yml
phw: 3.7 is the latest the works with current PyInstaller. They have mostly done the work, see https://github.com/pyinstaller/pyinstaller/issues/4311. So with the next PyInstaller release we should be able to try Python 3.7 for the macOS / Windows builds.

.github/workflows/package-windows.yml
phw: Always. We build 3 different packages above, even if the job actually failed I want the created packages to get uploaded.

.github/workflows/run-tests.yml
phw: That's intentional, I want to see both the results of the syntax checks and the actual test results. If the syntax checks fails the entire job will still be considered failed, but the tests run anyway. Otherwise you get the issues with your commit reported peu à peu.

Actually this option is something I really like about Github Actions, other CI systems often make achieving this behavior difficult.

phw:
I added the version run after pip install and merged this PR so we can get this running for new code changes. We can do improvements to this once we see how well it works.

I also disabled (but not removed) the webhooks for travis and appveyor. But let's keep this configured for a while so we can easily re-enable it if needed.

PR #1382, opened 2019-12-01
Update Github release from changelog

phw:
# Summary
This ads a Github Action that updates a new tagged release with proper title and changelog.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

# Solution
Implement a Github Action that runs only on release-* tags. It updates the title and changelog entry. For extracting the changelog a new script `changelog-for-version.py` is added.

As an test https://github.com/phw/picard/releases/tag/release-test was created using this action. This was the corresponding workflow run: https://github.com/phw/picard/commit/b2de297439eba89f4d0dbb0eef4331be4f676261/checks?check_suite_id=336456928

# Action

### DIFF
--- a/.github/workflows/github-release.yml
+++ b/.github/workflows/github-release.yml
@@ -1,0 +1,29 @@
+name: Prepare release
+
+on:
+  push:
+    tags:
+      - 'release-*'
+
+jobs:
+  prepare-release:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Prepare changelog
+      id: changelog
+      run: |
+        PICARD_VERSION=$(python -c "import picard; print(picard.__version__)")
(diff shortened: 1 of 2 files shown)


phw:
Codacy is having some trouble with Python 3 :(

zas:
> Codacy is having some trouble with Python 3 :(

Does it actually use Python 3 for this test? At worse, you can disable the check in the source code.
Apart that, the patch looks good to me.

phw:
> Does it actually use Python 3 for this test? At worse, you can disable the check in the source code.
> Apart that, the patch looks good to me.

I saw this already previously for e.g. `raise Error from` (as used in file.py), the prospector check on Codacy reports this as a syntax error. However, in general Prospector can check this. So I assume codacy is using the Python 2 version here, yes. But I haven't found an option to change this :( Unfortunately disabling Prospector looses a whole lot of useful checks.

We should investigate how to handle this, but for now I'm gonna just ignore it here. Actually Codacy has otherwise been behaving quite well and fast recently.

PR #1395, opened 2019-12-05
Add custom up/down/left/right icons

phw:
# Summary
This adds custom icons for buttons to move something left/right/top/down.

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
The new list view added for sorting tags added up/down buttons, which uses the system themes appropriate icons. Those theme icons are not available for macOS / Windows, leading to empty buttons.

# Solution
Added custom icons for up / down navigation. For consistency also added corresponding icons for left/right navigation and adapted the preferred releases options and the cover art type selection dialog to use those new icons.

Some screens:

grafik

grafik

grafik

# Action


phw:
Side note: I would like to refactor the CAA type selection dialog to be built with Qt Designer. The UI is rather complex and managing the little details is definitely easier using Qt Designer. As part of this I already refactored it to rely less on fixed sizes and placement, but it still is a bit difficult to maintain. Also it would need some tooltips or at least accessibility labels for the buttons.

But that's for another PR. This PR already got a bit more complex then what I usually would like to have for a simple "replace the icons" PR, but without the changes to this dialog's layout things looked off when changing just the buttons.

PR #1396, opened 2019-12-05
PICARD-1349: Allow resetting a column to unsorted state

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Currently once a column is selected for sorting the treeview stay sorted. One can toggle between ascending and descending order or sort a different column, but the treeview cannot be set to an unsorted state.

* JIRA ticket (_optional_): PICARD-1349

# Solution
Implement a custom `TristateSortHeaderView`, which toggles between the three states "ascending", "descending" and unsorted.

The implementation has to work around that a `mouseReleaseEvent` occurs both when a click happened and when a move or resize option ended. Only when it was not a move or resize it was an actual click that changes sort order. Unfortunately the default implementation does not indicate a indication of this. To avoid reimplementing all of `mousePressEvent`, `mouseMoveEvent` and `mouseReleaseEvent` and duplicating the original code this implementation instead tries to detect if a move or resize event happened.

Two things to note:
1. Resetting to unsorted does not change the current order of items. We don't remember any "unsorted" order separately. It just means that newly added items won't automatically get sorted into the list but always end up at the end.
2. There is one minor case not covered: If a column is sorted descending and one drags and drops a separate column on top of it without actually causing a column move then this will be counted as a click on the target column and the sort order is reset. I don't see a way to detect this, but it is a minor edge case and probably the user does probably not even notice when it happens.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -62,6 +62,7 @@ from picard.ui.collectionmenu import CollectionMenu
 from picard.ui.colors import interface_colors
 from picard.ui.ratingwidget import RatingWidget
 from picard.ui.scriptsmenu import ScriptsMenu
+from picard.ui.widgets.tristatesortheaderview import TristateSortHeaderView
 
 
 class BaseAction(QtWidgets.QAction):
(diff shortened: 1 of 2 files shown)


phw:
The issue of not being able to reset the sort order became a bit more urgent to fix with https://github.com/metabrainz/picard/pull/1389 (customizable main panel columns). Since this PR the sort order is persisted, so not even restarting Picard can help to get rid of unwanted sorting.

PR #1401, opened 2019-12-08
PICARD-1521: Add TOFN to ID3 metadata

zrdaley:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

This PR adds the `TOFN Original filename` ID3 tag. 

# Problem

Currently this tag is not supported.

* JIRA ticket (_optional_): [PICARD-1521](https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-1521?filter=allopenissues)

# Solution

The `TOFN` tag has been added as a `__translate` key in the `ID3File` class. This will allow it to be [recognized in the `__load` function](https://github.com/metabrainz/picard/blob/master/picard/formats/id3.py#L247).

cc. @awiggs @mehrler12 @Wyllan @raunaqg @omazhary


test/formats/test_id3.py
```diff
@@ -74,6 +74,13 @@ def test_id3_multiple_freeform_delete(self):
             self.assertNotIn('Bar', new_metadata)
             self.assertIn('FooBar', new_metadata)
 
+        @skipUnlessTestfile
+        def test_id3_metadata_tofn(self):
+            metadata = Metadata(self.tags)
+            metadata = save_and_load_metadata(self.filename, metadata)
+
+            self.assertIn('originalfilename', metadata)
```
phw: Please also check for the expected value (`Foo`) in addition to the presence of the tag

zrdaley:
Thank you for the review @phw. I made the changes you requested 😄

PR #1403, opened 2019-12-09
PICARD-1218: Fix tagger script list layout

phw:
# Summary
Fixes issues with the display of the tagger script list in options.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Fixes issues with 

* JIRA ticket: PICARD-1281, PICARD-965

# Solution
- Added label implementation that elides text appropriately if it does not fit into the available width.
- Reworked the layout, moved the layout definition into Qt Designer. The new layout is simpler and behaves better when scaled. Functionality and look are identical.

Tested on Windows, macOS and Linux. On Linux using both the default Fusion theme as well as overriding it with Qt-Adwaita and Breeze.

# Action


phw:
Please do not merge yet, there are some layout issues on macOS still. Will work on this later.

PR #1409, opened 2019-12-11
Show file data in track columns

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
If a main view column is supposed to show data that is only available for a file not the track, this column will stay empty if there is only a single file matched to a track.

# Solution
If only a single file is linked to a track and the data requested for a column is not available for the track, show the file column data instead.

This by default currently won't have a big effect, but it will help if we offer more selectable columns for the main tree views and it helps now for users who add custom columns via plugins.

# Action


17500mph:
:)
Screen Shot 2019-12-12 at 4 22 32 AM

PR #1411, opened 2019-12-12
Added generic SingletonDialog class for managing a single dialog instance

phw:
# Summary
Replace the specific implementation in MainWindow to manage a single instance of OptionsDialog by a generic implementation.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution
Add a `SingletonDialog` class that can be subclassed by dialogs to enable management of a single instance open at a time. This single instance can still be closed and destroyed, but as long as the `SingletonDialog` functions are used to create the dialog no more than one instance exist.

This is currently only used by OptionsDialog but could now be easily reused by dialogs with similar requirements.

# Action

### DIFF
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -57,6 +57,28 @@ class PreserveGeometry:
         config.persist[self.opt_name()] = self.saveGeometry()
 
 
+class SingletonDialog:
+    _instance = None
+
+    @classmethod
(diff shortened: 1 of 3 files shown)


phw:
This code was some result at my attempt of fixing PICARD-1638. While PICARD-1638 itself is harder to fix I guess we can at least include this part of my refactoring :)

PR #1412, opened 2019-12-12
PICARD-321: Support backslash in filenames on non-Windows OS

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Allow backslashes to be used in filenames on systems other than Windows. Only if Windows compatibility is activated backslashes will be replaced as before.

* JIRA ticket (_optional_): PICARD-321

# Solution
The core change is in `util.sanitize_filename`, where there is now no longer a hardcoded replacement of `\` and `/` but instead `os.sep` and `os.altsep` are used an a `win_compat` parameter is accepted in addition. Ensure functions in `textencoding`  and `script_to_filename` set the `win_compat` parameter accordingly.

Tested on Linux and Windows 10

# Action

### DIFF
--- a/picard/util/scripttofilename.py
+++ b/picard/util/scripttofilename.py
@@ -34,15 +34,17 @@ def script_to_filename(naming_format, metadata, file=None, settings=None):
     if settings is None:
         settings = config.setting
     # make sure every metadata can safely be used in a path name
+    win_compat = IS_WIN or settings["windows_compatibility"]
     meta = Metadata()
     for name in metadata:
-        meta[name] = [sanitize_filename(str(v)) for v in metadata.getall(name)]
+        meta[name] = [sanitize_filename(str(v), win_compat=win_compat)
+                      for v in metadata.getall(name)]
     naming_format = naming_format.replace("\t", "").replace("\n", "")
     filename = ScriptParser().eval(naming_format, meta, file)
     if settings["ascii_filenames"]:
-        filename = replace_non_ascii(filename, pathsave=True)
+        filename = replace_non_ascii(filename, pathsave=True, win_compat=win_compat)
     # replace incompatible characters
-    if settings["windows_compatibility"] or IS_WIN:
+    if win_compat:
         filename = replace_win32_incompat(filename)
     # remove null characters
     filename = filename.replace("\x00", "")
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -163,9 +163,13 @@ def strip_non_alnum(string):  # noqa: E302
     return _re_non_alphanum.sub(" ", string).strip()
 
 
-_re_slashes = re.compile(r'[\\/]', re.UNICODE)
-def sanitize_filename(string, repl="_"):  # noqa: E302
-    return _re_slashes.sub(repl, string)
+def sanitize_filename(string, repl="_", win_compat=False):
+    string = string.replace(os.sep, repl)
+    if os.altsep:
+        string = string.replace(os.altsep, repl)
+    if win_compat and os.altsep != '\\':
+        string = string.replace('\\', repl)
+    return string
 
 
 def _reverse_sortname(sortname):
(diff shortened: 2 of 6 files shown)


picard/util/scripttofilename.py
```diff
@@ -34,15 +34,17 @@ def script_to_filename(naming_format, metadata, file=None, settings=None):
     if settings is None:
         settings = config.setting
     # make sure every metadata can safely be used in a path name
+    win_compat = settings["windows_compatibility"] or IS_WIN
```
zas: I'd move `IS_WIN` first.

phw:
I was a bit concerned about performance, because we are calling the textencoding and sanitize_filename functions a lot and the textencoding functions previously where more static with pre-processed tables. Now the pathsave and win_compat parameters are evaluated at runtime. I did some measurements and additional optimizations (everything measured with Python 3.8 on Linux):

**`sanitize_filename`:**

Test script:
```
import timeit
from picard.util import sanitize_filename

print(
    "# win_compat=False",
    timeit.timeit(lambda: sanitize_filename(r"C:\Abc\xyz/abc/xyz/", win_compat=False), number=10000000)
)

print(
    "# win_compat=True",
    timeit.timeit(lambda: sanitize_filename(r"C:\Abc\xyz/abc/xyz/", win_compat=True), number=10000000)
)
```

Results for original regex based implementation:

```
# win_compat=False 9.4286299219998
# win_compat=True 9.725203413000145
```

Results for new implementation:

```
# win_compat=False 2.8284279720000995
# win_compat=True 4.310084980000283
```

More runtime checks, but no regex. Much faster, we are good here.

**`unicode_simplify_combinations`:**

All measurements are done with the optimized `sanitize_filename`.

Test script:

```
import timeit
from picard.util.textencoding import (
    _simplify_combinations,
    unicode_simplify_combinations,
)

string = "".join(_simplify_combinations.keys())

print(
    "# pathsave=False",
    timeit.timeit(lambda: unicode_simplify_combinations(string, pathsave=False), number=100000)
)

print(
    "# pathsave=True, win_compat=True",
    timeit.timeit(lambda: unicode_simplify_combinations(string, pathsave=True, win_compat=True), number=100000)
)

print(
    "# pathsave=True, win_compat=False",
    timeit.timeit(lambda: unicode_simplify_combinations(string, pathsave=True, win_compat=False), number=100000)
)
```

Original implementation:

```
# pathsave=False 8.769120420000036
# pathsave=True 8.867318464000164
```

As expected there is no runtime difference between pathsave and non-pathsave, as the mapping for pathsave is also precalculated.

Original new implementation from this PR:

```
# pathsave=False 8.94882393700027
# pathsave=True, win_compat=True 17.196504073999677
# pathsave=True, win_compat=False 15.297117036999225
```

The pathsave variants now take significantly more time, not good. But maybe we can get rid of the  regex by using something like this:

```
def _replace_unicode_simplify_combinations(char, pathsave, win_compat):
    result = _simplify_combinations.get(char)
    if not result:
        return char
    elif not pathsave:
        return result
    else:
        return sanitize_filename(result, win_compat=win_compat)

def unicode_simplify_combinations(string, pathsave=False, win_compat=False):
    return ''.join(
        _replace_unicode_simplify_combinations(c, pathsave, win_compat) for c in string)
```

This halfs the execution time of the non-pathsave case and at least brings us back into a comparable area for the pathsave case.

```
# pathsave=False 4.798246170000311
# pathsave=True, win_compat=True 12.832013631000336
# pathsave=True, win_compat=False 10.53033615700042
```

I added the changes in https://github.com/metabrainz/picard/pull/1412/commits/15635e6ed82048523ad9342d2106a0c75d604566

picard/util/scripttofilename.py
phw: Did this

PR #1413, opened 2019-12-12
PICARD-1596: Use Github Actions to publish to PyPI

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem
Publish release tags to PyPI using Github Actions.

* JIRA ticket (_optional_): PICARD-1596

# Solution
Added Github Actions workflow to publish to PyPI using Twine.

I have tested the commands in this order locally minus the actual twine upload, but I could not fully test this workflow. I think it is fine, but please look at it carefully. Once this is merged we probably should do a dev release soon to check our deployment process in entirety.

There are some [Github Actions](https://github.com/marketplace?utf8=%E2%9C%93&type=actions&query=pypi) specifically for this, but I think the process using twine directly is so straight forward that adding basically another dependency and passing the credentials around to another external dependency is just adding an unnecessary complexity.

# Action
We should consider also building bdist packages, at least for Windows and maybe macOS.

### DIFF
--- a/.github/workflows/pypi-release.yml
+++ b/.github/workflows/pypi-release.yml
@@ -1,0 +1,33 @@
+name: Publish to PyPI
+
+on:
+  push:
+    tags:
+      - 'release-*'
+
+jobs:
+  pypi-sdist:
+    runs-on: ubuntu-latest
+
+    steps:
+    - uses: actions/checkout@v1
+    - uses: actions/setup-python@v1
+      with:
+        python-version: 3.8
+    - name: Install dependencies
+      run: |
+        python -m pip install --upgrade pip
+        pip install --upgrade -r requirements.txt
+    - name: Run tests
+      run: |
+        python setup.py test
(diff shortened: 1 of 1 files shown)


.github/workflows/pypi-release.yml
```diff
@@ -0,0 +1,33 @@
+name: Prepare release
```
mineo: Maybe rename this as https://github.com/metabrainz/picard/blob/master/.github/workflows/github-release.yml#L1 already defines a workflow with the same name.

phw: Thanks, I forgot to rename it. Fixed

phw:
> Should we be running a `clean` first, just in case?

It does not clean anything right now in this case, but it kind of makes sense and does no harm. Provides some safety for the future should we generate something during test that needs to be cleaned. I added it.

PR #1414, opened 2019-12-15
Allow pre-release version numbers (alpha, beta, rc)

phw:
# Summary
Re-add ability to mark alpha, beta and rc pre-release versions

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem
For Picard versions support the version identifiers 'alpha', 'beta' and 'rc' again (in addition to 'final' and 'dev'). This used to be supported, but got lost at some point
    
Long version strings will be e.g. 2.3.0beta2. If the shortened string is generated the idenfitiers for alpha and beta will be shortened to 'a' and 'b' (e.g. 2.3.0b2). This confirms to pre-release versioning is defined in PEP 440.

The background for this is that I really want to do a proper beta release for Picard 2.3 this time, as there are some changes where I would like to see a release that clearly indicates it is a pre-release meant to gather user feedback. Also before this I would like to tag an alpha release of current development status primarily to test our deployment process with new CI and automation.

Both of this requires proper tagging of pre-releases and ideally a version that is compatible with [PEP 440 pre-releases](https://www.python.org/dev/peps/pep-0440/#pre-releases).

# Solution

# Action
I am not entirely sure why we lost the ability to tag `alpha`, `beta` and `rc` versions. Was it intentional to limit this to `final` and `dev`?

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -41,8 +41,8 @@ class VersionError(Exception):
 def version_to_string(version, short=False):
     if len(version) != 5:
         raise VersionError("Length != 5")
-    if version[3] not in ('final', 'dev'):
-        raise VersionError("Should be either 'final' or 'dev'")
+    if version[3] not in ('final', 'dev', 'alpha', 'beta', 'rc'):
+        raise VersionError("Should be either 'final', 'dev', 'alpha', 'beta' or 'rc'")
     _version = []
     for p in version:
         try:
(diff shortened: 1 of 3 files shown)


zas:
> I am not entirely sure why we lost the ability to tag `alpha`, `beta` and `rc` versions.

Mainly because we don't use them.

samj1912:
I don't recall the exact reason, but I would check the update manager/plugin code to see if it is compatible.

phw:
> but I would check the update manager/plugin code to see if it is compatible.

There can't be anything incompatible now. The changes do not restrict the version further, all the versions that previously where considered valid still are. The only change is that additional identifiers are allowed. But yes, we should document that the version is expected to follow Picard's standards.

Actually the only place where I think this was critical before is the use of `version_from_string` in `config.py`. This happily assumed that versions could be read from config, but `0.15.beta1` is a totally valid version string that was written to config files in this specific version. So before this patch reading such a config file would fail, now it works again.

PR #1415, opened 2019-12-15
Add Github Actions status badge to README.md

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem
Add a Github Actions status badge and replace the currently unused CI badges with it.

# Solution

# Action

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,7 @@
 MusicBrainz Picard
 ==================
-[![Travis Status](https://img.shields.io/travis/metabrainz/picard/master.svg?style=flat-square&label=Travis)](https://travis-ci.org/metabrainz/picard)
+![Github Actions Status](https://github.com/metabrainz/picard/workflows/Run%20tests/badge.svg)
 [![Codacy Grade](https://img.shields.io/codacy/grade/53a33607234a4c18a11a6207d1173c0c/master.svg?style=flat-square&label=Codacy)](https://www.codacy.com/app/MetaBrainz/picard)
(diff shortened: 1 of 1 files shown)


phw:
@zas, @samj1912 : What do you think we should do with the existing CI configs? I am personally very happy with Github Actions. Having one place for all builds makes it much easier then the previous Travis / Appveyor mix.

Circle CI would be the only contender, but I find GHA much easier to setup. Circle CI requires you to come with a custom docker image for many use cases. So my vote would go for removing the configs for all of Appveyor, Travis and Circle.

samj1912:
I am fine with GHA. Seems like they are faster. My only concern is that GH doesn't seem to have old versions of MacOS and as a result the latest versions of picard will have limited support. Maybe we want to keep Travis build on new releases?

phw:
> Maybe we want to keep Travis build on new releases?

This was also my concern, but for now the build versions are compatible with macOS 10.12, which is the oldest we support anyway. Hence I'm fine with dropping Travis.

Actually the backward support for us is currently primarily dependent on the version of Qt we use. E.g. we must be careful with updating to Qt 5.14 on macOS, because they set the minimum target platform to macOS 12.13 (see https://doc.qt.io/qt-5.14/supported-platforms.html).

So in general if we do update any of Python itself, (Py)Qt or PyInstaller we must check which is the oldest macOS these builds still run with.

Things of course can change if Github updates their macOS images to newer versions. My hope would be that they provide some support for older versions and don't just upgrade immediately, but we can't tell yet. But I would worry about this once we hit the issue. It can hit us the same on any CI provider unless we control the build machine ourselves.

zas:
I'm fine with GHA if it fulfills our goals.

PR #1416, opened 2019-12-16
GHA: Run tests with older dependencies

phw:
# Summary
Run tests on Github Actions with older sip dependencies (mainly older PyQt5 versions).

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

# Solution
Run tests with older sip dependencies. To not escalate this too much I limited this to run on Linux only. For macOS and Windows I would prefer to only officially support the versions we have defined in the `requirements-{mac|win}.txt` files, as used by the packaging, and the actual latest versions.

I also consider to run these extended tests only on the master branch and not for all PRs, opinions?

# Action
I had an alternative implementation where I ran tests with different dependencies inside the current jobs, see https://github.com/phw/picard/commit/5291de1f56cc610cbee2dff44c62f14ac57ff9cb

But I dislike how this increases the build time to multiple minutes per job.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -3,7 +3,7 @@ name: Run tests
 on: [push, pull_request]
 
 jobs:
-  build:
+  test-latest:
     runs-on: ${{ matrix.os }}
     strategy:
       matrix:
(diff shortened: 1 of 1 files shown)


zas:
> I also consider to run these extended tests only on the master branch and not for all PRs, opinions?

Yes, it seems enough to me, at least it will help to detect incompatibilities without being too heavy.

PR #1418, opened 2019-12-17
PICARD-1695: Rework script editor

phw:
# Summary
Clean up the script editor UI, make it more consistent with other parts of the UI and enable re-ordering scripts by drag and drop.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
The scripting option page currently looks different than other parts of the UI where lists of items can be edited (e.g. cover art providers, toolbar buttons, top tags)

Some differences:

- The "Add script" button is above the list, whereas it is below the list on similar views
- Buttons to re-order elements are displayed inline as opposed to be below the listview
- Same goes for edit and remove buttons

Some visual oddities:

- The actual editor is nested in two QGroupBox containers
- There is strange padding on top and bottom

* JIRA ticket (_optional_): PICARD-1695

# Solution
Updated UI:

grafik

Also added the ability to move scripts by drag and drop by using `MoveableListView`.

This also simplifies the code. We had previously 4 different models per script to manage and 3 separate lists with different representation of the script list to keep in sync.

# Action


phw:
I like your mockup. Also to have the script's name edited on the left and add a context menu. This could also duplicate the remove action in addition to the rename action.

If we go for the icon buttons we probably should apply this to other similar views as well.

I will look into both things once we have merged this one.

PR #1422, opened 2020-01-02
Implement proper version comparison for dev, alpha, beta and rc releases

phw:
# Summary
The version identifier was previously compared alphabetically, which fails for e.g. "dev" < "alpha".

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
With the recent addition of (again) supporting alpha, beta and rc releases the comparison of the version identifier broke. This would lead to warnings such as "Warning: config file 2.3.0.dev1 was created by a more recent version of Picard (current is 2.3.0.alpha1)".

# Solution
Implement proper version comparison by introducing a Version class (inherited from tuple), which implements proper comparison operators. Also moved the logic fro to and from string conversion there.

`picard.version_from_string` is now considered obsolete, but I did not yet remove it as it might be used by plugin code. `version_to_string` is still useful when one has only a version tuple.

# Action

### DIFF
--- a/picard/version.py
+++ b/picard/version.py
@@ -37,40 +37,45 @@ class Version(tuple):
         'final': 4
     }
 
-    def __new__(cls, major, minor, patch, identifier='final', revision=0):
+    def __new__(cls, major, minor, patch=0, identifier='final', revision=0):
         if identifier not in cls.valid_identifiers():
             raise VersionError("Should be either 'final', 'dev', 'alpha', 'beta' or 'rc'")
         identifier = {'a': 'alpha', 'b': 'beta'}.get(identifier, identifier)
+        try:
+            major = int(major)
+            minor = int(minor)
+            patch = int(patch)
+            revision = int(revision)
+        except (TypeError, ValueError):
+            raise VersionError("major, minor, patch and revision must be integer values")
         return super(Version, cls).__new__(cls, (major, minor, patch, identifier, revision))
 
     @classmethod
     def from_string(cls, version_str):
-        m = cls._version_re.search(version_str)
-        if m:
-            g = m.groups()
-            if g[2] is None:
-                return Version(int(g[0]), int(g[1]), 0, 'final', 0)
-            if g[3] is None:
-                return Version(int(g[0]), int(g[1]), int(g[2]), 'final', 0)
-            return Version(int(g[0]), int(g[1]), int(g[2]), g[3], int(g[4]))
+        match = cls._version_re.search(version_str)
+        if match:
+            (major, minor, patch, identifier, revision) = match.groups()
+            major = int(major)
+            minor = int(minor)
+            if patch is None:
+                return Version(major, minor)
+            patch = int(patch)
+            if identifier is None:
+                return Version(major, minor, patch)
+            revision = int(revision)
+            return Version(major, minor, patch, identifier, revision)
         raise VersionError("String '%s' does not match regex '%s'" % (version_str,
                                                                       cls._version_re.pattern))
 
     @classmethod
     def valid_identifiers(cls):
-        return cls._identifiers.keys()
+        return set(cls._identifiers.keys())
 
     def to_string(self, short=False):
-        _version = []
-        for p in self:
-            try:
-                n = int(p)
-            except ValueError:
-                n = p
-            _version.append(n)
-        if short and _version[3] in ('alpha', 'beta'):
-            _version[3] = _version[3][:1]
-        version = tuple(_version)
+        if short and self[3] in ('alpha', 'beta'):
+            version = (self[0], self[1], self[2], self[3][:1], self[4])
+        else:
+            version = self
         if short and version[3] == 'final':
             if version[2] == 0:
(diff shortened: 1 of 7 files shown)


picard/version.py
```diff
@@ -0,0 +1,123 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+import re
+
+
+class VersionError(Exception):
+    pass
+
+
+class Version(tuple):
```
samj1912: Why not use total_ordering to get the comparators?

samj1912: Why not use a named tuple? From typing.

picard/version.py
```diff
@@ -0,0 +1,123 @@
...
+        'dev': 0,
+        'alpha': 1,
+        'a': 1,
+        'beta': 2,
+        'b': 2,
+        'rc': 3,
+        'final': 4
+    }
+
+    def __new__(cls, major, minor, patch, identifier='final', revision=0):
```
samj1912: This could be init if named tuple is used?

picard/version.py
```diff
@@ -0,0 +1,123 @@
...
+            if g[3] is None:
+                return Version(int(g[0]), int(g[1]), int(g[2]), 'final', 0)
+            identifier = {'a': 'alpha', 'b': 'beta'}.get(g[3], g[3])
+            return Version(int(g[0]), int(g[1]), int(g[2]), identifier, int(g[4]))
+        raise VersionError("String '%s' does not match regex '%s'" % (version_str,
+                                                                      cls._version_re.pattern))
+
+    @classmethod
+    def valid_identifiers(cls):
+        return cls._identifiers.keys()
```
samj1912: This will return dictkeys, it might be better returning a list or set or a tuple?

picard/version.py
```diff
@@ -0,0 +1,123 @@
...
+
+    def __new__(cls, major, minor, patch, identifier='final', revision=0):
+        if identifier not in cls.valid_identifiers():
+            raise VersionError("Should be either 'final', 'dev', 'alpha', 'beta' or 'rc'")
+        return super(Version, cls).__new__(cls, (major, minor, patch, identifier, revision))
+
+    @classmethod
+    def from_string(cls, version_str):
+        m = cls._version_re.search(version_str)
+        if m:
```
samj1912: Can we have better names than m and G?

picard/version.py
```diff
@@ -0,0 +1,123 @@
...
+            g = m.groups()
+            if g[2] is None:
```
samj1912: If named tuple is used, you can have defaults here.

picard/version.py
phw: > Why not use total_ordering to get the comparators?

This would be best, but doesn't work since all comparators are already implemented (inherited from `tuple`) and `total_ordering` only fills in the missing ones. So we have to override them.

picard/version.py
phw: No, actually not. init does not work with tuples at all, since they are immutable after creation. So even with nameduple we would need to use `__new__` (but only for the identifier validation). The code would change to something like:

```
class Version(namedtuple('V', 'major minor patch identifier revision')):

    def __new__(cls, *args, **kwargs):
        instance = super(Version, cls).__new__(cls, *args, **kwargs)
        if instance.identifier not in cls.valid_identifiers():
            raise VersionError("Should be either 'final', 'dev', 'alpha', 'beta' or 'rc'")
        return instance

```

I currently don't see much value in using namedtuple here. It introduces another class in between without bringing much benefit. Having the keywords available could make the code more readable at some places (e.g. `version.minor` instead of `version[1]`). I'll experiment with it and sees if is worth the trouble.

picard/version.py
phw: Yes, or make this thing private or remove it entirely. I'll look into it.

phw:
@samj1912 I addressed your comments, including the use of namedtuple. We don't benefit much of the named fields right now, but it allows the code in `Version.to_string` to be a bit more readable.

test/test_versions.py
```diff
@@ -87,3 +98,54 @@ def test_api_versions_tuple_1(self):
...
+    def test_sortkey(self):
+        self.assertEqual((2, 1, 3, 4, 2), Version(2, 1, 3, 'final', 2).sortkey)
+        self.assertEqual((2, 0, 0, 1, 0), Version(2, 0, 0, 'a', 0).sortkey)
+        self.assertEqual((2, 0, 0, 1, 0), Version(2, 0, 0, 'alpha', 0).sortkey)
+
+    def test_lt(self):
+        v1 = Version(2, 3, 0, 'dev', 1)
+        v2 = Version(2, 3, 0, 'alpha', 1)
+        self.assertLess(v1, v2)
+        self.assertFalse(v2 < v2)
```
zas: What's the goal of this test?

phw: Verifying the __lt__ implementation compares properly

PR #1423, opened 2020-01-06
PICARD-991: AcoustId fingerprint generation

phw:
# Summary
This ads an explicit action to generate the AcoustId fingerprint for selected files to the context menu and Tools menu.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Picard currently does not have an explicit way to generate a AcoustId fingerprint. The only way is using scan, which generates the fingerprint. But actually "Scan" is one of the lookup / search / matching tools Picard offers and its main purpose is to try to match the files. As such it behaves much like the "Lookup" action and is only available for unmatched files. That it also generates the fingerprints is just part of this.

But many users want to use Picard also to contribute back to AcoustId and submit fingerprints, but do the actual matching by other means. The only current workaround is to use "Scan". For already matched files this involves dragging all files to the left, use Scan, wait for the results loading, the  dragging all files back to the actually matched tracks and finally submit.

* JIRA ticket (_optional_): PICARD-991

# Solution
Add an explicit "Generate AcoustID fingerprints" action to the tools menu and context menu. This action really just generates the fingerprint without doing any lookup. It is available both to matched and unmatched files. Also modified the code to not run fpcalc again for Scan if it already ran for a file. That way the user could even use the "Generate AcoustID fingerprints" first, and later use "Scan" without causing it to run fpcalc again.

# Action
I think implementing this also makes implementing some type of indicator, whether the fingerprint was generated for a file or not (PICARD-34), more important. I actually started an implementation of this (using a separate column with an fingerprint icon), but I'm not happy yet with the implementation. I think it's best to keep both issues separate and first merge this part.


phw:
Just for starting the discussion: My current implementation for a fingerprint indicator (PICARD-34) is at https://github.com/phw/picard/tree/PICARD-34-acoustid-fingerprint-indicator . It would add a separate column the user can activate (making use of the new configurable columns):

grafik

I'm not yet fully satisfied with this and would appreciate opinions / alternative proposals.

zas:
> I'm not yet fully satisfied with this and would appreciate opinions / alternative proposals.

It seems good enough to me, a visual indication of which file is missing acoustid would be a great improvement over this patch.

picard/ui/mainwindow.py
```diff
@@ -946,6 +953,14 @@ def analyze(self):
                 return
         return self.tagger.analyze(self.selected_objects)
 
+    def generate_fingerprints(self):
+        if not config.setting['fingerprinting_system']:
+            if self.show_analyze_settings_info():
+                self.show_options("fingerprinting")
+            if not config.setting['fingerprinting_system']:
```
zas: Is `config.setting['fingerprinting_system']` modified in between `if` and here ? It looks weird to me.

phw: No, that needs to be changed. I think during development the outer if originally had some other condition or'ed in and I removed that. I'll look into it.

phw: Actually the code used to be fine, but I think we broke it at some point when `MainWindow.show_options()` no longer used exec (https://github.com/metabrainz/picard/commit/f8eb75596e8090d2f08e9290608dbe7a9f4964d4). The original idea was to allow the user to enable fingerprinting, and Dialog.exec would halt the code execution until the user closed options again. In this case `config.setting['fingerprinting_system']` indeed could change.

Unfortunately this is no longer happening since PICARD-1571 and the code does not have the desired effect. I added PICARD-1699 to address this. I'll submit a fix in a separate PR.

zas: Ok, thanks for explanation, a separate PR will do.

PR #1424, opened 2020-01-07
PICARD-1700: Fix crash on PyQt >= 5.14 on cover art context menu

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
In PyQt 5.14 the exclusive parameter for QActionGroup is no longer available. Also QActionGroup is exclusive by default in Qt5 (it even was in Qt4), so there is no need to set this (see https://doc.qt.io/qt-5/qactiongroup.html#QActionGroup).

* JIRA ticket (_optional_): PICARD-1700

# Solution

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -538,7 +538,7 @@ class CoverArtBox(QtWidgets.QGroupBox):
         if not menu.isEmpty():
             menu.addSeparator()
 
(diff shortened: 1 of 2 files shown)


phw:
@samj1912 Sorry, just added another commit :) This is to avoid a similar issue in the future with `QAction(..., checkable=True)`

PR #1428, opened 2020-01-09
PICARD-1054: Allow Qt to handle default Qt and X command line arguments 

phw:
# Summary
Fixes Picard's handling of argv and allows passing the default Qt command line arguments (e.g. `-style [Stylename]`)

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The current handling of argv caused some issues. The QApplication classes expect a full argv to be passed to them, but we only passed on the remaining args, which caused a few issues.

- argv[0] is supposed to be the path to the executed binary, but it wasn't for us. This breaks e.g. `QCoreApplication.applicationFilePath` and also was what caused us to need to include this `StartupWMClass` workaround
- Command line arguments for Qt and/or X which require an additional keyword (e.g. `-style` or `-desktop`) where broken. E.g. if passing `-style Windows` then "Windows" would be interpreted as a filename, and Qt would only receive `-style` without parameter.

* JIRA ticket (_optional_): PICARD-1054

# Solution
- Properly pass complete argv to `QApplication`
- Add specific support for parameterized default Qt/X command line arguments, so they are handled correctly and not interpreted as filenames. These parameters are hidden from help for now, but are available for users knowing about standard parameters.
- Remove StartupWMClass workaround

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -888,6 +888,14 @@ def process_picard_args():
     parser = argparse.ArgumentParser(
         epilog="If one of the filenames begins with a hyphen, use -- to separate the options from the filenames."
     )
+    # Qt default arguments. Parse them so Picard does not interpret the
+    # arguments as file names to load.
+    parser.add_argument("-style", nargs=1, help=argparse.SUPPRESS)
+    parser.add_argument("-stylesheet", nargs=1, help=argparse.SUPPRESS)
+    # Same for default X arguments
+    parser.add_argument("-display", nargs=1, help=argparse.SUPPRESS)
+
+    # Picard specific arguments
     parser.add_argument("-c", "--config-file", action='store',
                         default=None,
                         help="location of the configuration file")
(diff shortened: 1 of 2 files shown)


picard/tagger.py
```diff
@@ -143,9 +143,7 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
         if not IS_MACOS and not IS_HAIKU:
             self.setStyle('Fusion')
 
-        # Set the WM_CLASS to 'MusicBrainz-Picard' so desktop environments
```
samj1912: Is this comment no longer applicable?

phw: No, the relevant code was removed. Turns out this workaround was actually only necessary because we did not provide the full argv (with argv[0] being set to the application path) to Qt

phw:
> According to my testing, it doesn't introduce any regression (related to https://tickets.metabrainz.org/browse/PICARD-761)

After merging I will test this on some more systems to be sure.

PR #1429, opened 2020-01-10
PICARD-1703: Upgrade to PyInstaller 3.6

phw:
# Summary
Allows us to remove some Windows specific workarounds, fixes a CVE related to the portable onefile install.

See https://pyinstaller.readthedocs.io/en/v3.6/CHANGES.html for details

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

* JIRA ticket (_optional_): PICARD-1703

# Solution

# Action

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -50,18 +50,8 @@ data_files = get_locale_messages()
 
 fpcalc_name = 'fpcalc'
 if os_name == 'Windows':
-    from PyQt5.QtCore import QLibraryInfo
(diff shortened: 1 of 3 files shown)


phw:
> Did we manually check if these are working correctly? iirc the openssl dlls were required by windows 7

Yes, I ran those on Windows. They are also required by Windows 10, without them you don't have SSL support. But the package content is still the same, it is just that this now properly works in PyInstaller. Previous versions failed to copy the DLLs.

PR #1431, opened 2020-01-12
PICARD-34: AcoustID fingerprint indicator

phw:
# Summary
Add a column with a fingerprint indicator.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
If the fingerprint has been calculated for a file or whether is is ready for submission or already submitted is currently very intransparent in Picard. This became especially obvious after implementation of manual fingerprint generation (PICARD-991).

* JIRA ticket (_optional_): PICARD-34

# Solution

Add a new column (for now hidden by default) which shows the status of fingerprinting with a fingerprint item + tooltip. Possible values are:

- No fingerprint (no icon)
- Submitted fingerprint (gray icon)
- Fingerpint submittable (red icon)

grafik

# Action


phw:
The macOS build failing is unrelated to this PR. This suddenly happens without change from our side, see PICARD-1705. Need to figure out what changed on the build setup since the last successful build.

picard/ui/itemviews.py
```diff
@@ -256,6 +262,35 @@ def remove(self, objects):
             self.update_current_view()
 
 
+def paint_fingerprint_icon(painter, rect, icon):
+    if not icon:
+        return
+    size = 16
```
zas: Perhaps use a constant for this, it may be used elsewhere (?)

phw:
I fixed the builds and added a constant for the icon size as suggested. Before merging I need to test this properly on macOS if it works for the layout.

picard/acoustid/manager.py
```diff
@@ -62,29 +62,36 @@ def remove(self, file):
...
+        submissions = list(self._unsubmitted())
+        if not submissions:
             self._check_unsubmitted()
             return
         log.debug("AcoustID: submitting ...")
         self.tagger.window.set_statusbar_message(
             N_('Submitting AcoustIDs ...'),
             echo=None
         )
+        fingerprints = [fingerprint for _, fingerprint in submissions]
```
zas: Even though it is a standard way to declare an unused variable in Python, since it is also used for gettext stuff, I would not use `_` at all.

phw: That's true. It is not really an issue here since this is scoped inside the list comprehension, but you are right we should avoid this.

On a similar note this specific source file makes heavy use of variables named `file`, overwriting the Python `file`. In most places I think we use `file_` for this, but we have a couple of more occurrences of just `file` I think. I will address both in a separate commit.

zas: Yup, the `file` one is really boring, Python guys chose a bad name for their module...

PR #1437, opened 2020-01-22
PICARD-1589: Support language for ID3 comments.

phw:
# Summary
Adds syntax `comment:{language}:{description}` in addition to existing `comment:{language}:{description}` for comment tag names.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
ID3 allows saving a language for comment tags. Currently Picard is hard coded to always write "eng", and most of the files seen in the wild seem to default to "eng".

An exception is MediaMonkey, which allows the user to write "custom" tags. Those custom tags get stored to COMM frames with a description on "Songs-DB_Custom1", "Songs-DB_Custom2" etc. and language set to "XXX".

Picard not supporting this currently has two affects:
- Trying to create such a tag with Picard fails. $set(comment:Songs-DB_Custom1,test) will write a COMM tag with proper description, but language set to "eng".
- Loading such a file will basically duplicate the tag. The existing tag with lang=XXX stays and a new tag with lang=eng gets written.

* JIRA ticket (_optional_): PICARD-1589

# Solution
This adds proper support for the syntax `comment:{language}:{description}` for ID3. The language will be set accordingly, but still default to `eng`.

For Vorbis and Ape tags this works transparently, as the additional language part will just be treated as an extended description part. So e.g. the tag `comment:deu:foo` will be saved as:

```
COMMENT=The comment (deu:foo)
```

Loading this back is fine.

For ASF and MP4 this does currently also not change anything in behavior, as these do neither support the `comment:desc` syntax currently nor do they properly save more than one comment. This needs to be fixed separately, see PICARD-587.

# Action

### DIFF
--- a/test/test_util_tags.py
+++ b/test/test_util_tags.py
@@ -1,0 +1,18 @@
+# -*- coding: utf-8 -*-
+from test.picardtestcase import PicardTestCase
+
+from picard.util.tags import (
+    display_tag_name,
+    parse_comment_tag,
+)
+
+
+class UtilTagsTest(PicardTestCase):
+    def test_display_tag_name(self):
+        self.assertEqual('Artist', display_tag_name('artist'))
+        self.assertEqual('Lyrics', display_tag_name('lyrics:'))
+        self.assertEqual('Comment [Foo]', display_tag_name('comment:Foo'))
+
+    def test_parse_comment_tag(self):
+        self.assertEqual(('XXX', 'foo'), parse_comment_tag('comment:XXX:foo'))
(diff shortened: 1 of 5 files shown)


test/test_util_tags.py
```diff
@@ -0,0 +1,20 @@
+# -*- coding: utf-8 -*-
+import unittest
```
zas: This is unused, according to Codacy

phw: I fixed that

PR #1440, opened 2020-01-25
PICARD-1712: use QtGui.QKeySequence.Delete for remove albums/files shortcut

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

There is no keyboard shortcut for Remove albums/files action, and Ctrl+D is already used for Add Folder.

* JIRA ticket (_optional_): PICARD-1712

# Solution

Use Ctrl+E as shortcut, and `e` as accelerator for Add Folder
Use Del/Ctrl+D for Remove action

See https://doc.qt.io/qt-5/qkeysequence.html

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -560,7 +560,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.view_log_action = QtWidgets.QAction(_("View &Error/Debug Log"), self)
         self.view_log_action.triggered.connect(self.show_log)
         # TR: Keyboard shortcut for "View Error/Debug Log"
-        self.view_log_action.setShortcut(QtGui.QKeySequence(_("Ctrl+E")))
+        self.view_log_action.setShortcut(QtGui.QKeySequence(_("Ctrl+G")))
 
         self.view_history_action = QtWidgets.QAction(_("View Activity &History"), self)
         self.view_history_action.triggered.connect(self.show_history)
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -407,10 +407,10 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.add_files_action.setShortcut(QtGui.QKeySequence.Open)
         self.add_files_action.triggered.connect(self.add_files)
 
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -407,10 +407,10 @@ def create_actions(self):
         self.add_files_action.setShortcut(QtGui.QKeySequence.Open)
         self.add_files_action.triggered.connect(self.add_files)
 
-        self.add_directory_action = QtWidgets.QAction(icontheme.lookup('folder'), _("A&dd Folder..."), self)
+        self.add_directory_action = QtWidgets.QAction(icontheme.lookup('folder'), _("Add Fold&er..."), self)
```
phw: Do we really need to change the accelerator here? This will be different for different languages anyway, and it does not necessarily correspond to the shortcut.

zas:
> I'm just not to sure about the `Ctrl+E`, that looks a bit arbitrary. But I also don't know what to use instead. `Ctrl+F` comes to mind, but that is usually associated with "Find" (and btw. it would probably make sense to bind this and focus the search bar).

Yup, totally arbitrary, but I tried to find one which isn't too comon (like Ctrl+F), I also thought of Ctrl+I (usually used for italic, but since this letter doesn't even appear in Add Folder, I preferred `e` so i could used same for accelerator and shortcut).
I don't like to change shortcuts, but in this case I think using Del/Ctrl+D from Remove is quite expected.

picard/ui/mainwindow.py
zas: I just wanted to use common letter `e` to match Ctrl+E, in many languages it is possible to keep it. It isn't mandatory, and of course it will depend on translations, but since we are at it, I thought it would be more consistent.

phw: Ok, makes sense. The annoying part here is that it breaks translations for no good reason :) But that's something we can handle. I'll go through existing translations and approve this string manually.

phw:
What about using `Ctrl+E` and remap the log window to `Ctrl+G`? The latter is no yet used.

zas:
> What about using `Ctrl+E` and remap the log window to `Ctrl+G`? The latter is no yet used.

I missed that....

Done in https://github.com/metabrainz/picard/pull/1440/commits/e3c7e1882975da7a0d18fb65ef63007a4b7395a3

PR #1441, opened 2020-01-25
PICARD-239: Handle tags like "comment" and "comment:" the same

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Some of Picard's tag names allow an additional description separated by a colon, e.g. "comment:desc". In case the description part is empty and the tag ends on a colon this is now treated the same as without any colon. So "lyrics" and "lyrics:" are the same tag.

* JIRA ticket (_optional_): PICARD-239

# Solution
Change `Metadata` to treat tags ending on a colon behave the same as the same without the colon.

Handling this in Metadata makes this change immediately available in scripting and to all formats. Script using both forms will still work. Only minor changes where necessary to the formats themselves.

# Action

### DIFF
--- a/test/test_metadata.py
+++ b/test/test_metadata.py
@@ -131,6 +131,22 @@ class MetadataTest(PicardTestCase):
         self.assertIn("single1", self.metadata)
         self.assertNotIn("single1", self.metadata.deleted_tags)
 
+    def test_normalize_tag(self):
+        self.assertEqual('sometag', Metadata.normalize_tag('sometag'))
+        self.assertEqual('sometag', Metadata.normalize_tag('sometag:'))
+        self.assertEqual('sometag', Metadata.normalize_tag('sometag::'))
+        self.assertEqual('sometag:desc', Metadata.normalize_tag('sometag:desc'))
+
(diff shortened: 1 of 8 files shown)


test/test_metadata.py
```diff
@@ -131,6 +131,21 @@ def test_metadata_undelete(self):
         self.assertIn("single1", self.metadata)
         self.assertNotIn("single1", self.metadata.deleted_tags)
 
+    def test_normalize_tag(self):
+        self.assertEqual('sometag', Metadata.normalize_tag('sometag'))
+        self.assertEqual('sometag', Metadata.normalize_tag('sometag:'))
```
zas: Since `Metadata.normalize_tag()` is using `str.rstrip()`, it will strip all occurences of `:` from the end.

```python
>>> 'comment:::'.rstrip(':')
'comment'
```

This case isn't tested here. Not sure whether it should or if removing only one occurence is the way to go.

phw: I think it should remove all of them, especiall since we introduced the `comment:{language}:{description}` syntax. I'll add a test

PR #1448, opened 2020-01-30
PICARD-1716: Add script functions for strings, multi-value variables, and loops

rdswift:
# Summary
Add new functions to the scripting library.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Add new functions as:**
    * 'find' substring within string
    * string 'reverse'
    * 'substr' to extract a substring from a string
    * 'get_multi_item' to get an element from a multi-value variable
    * 'foreach' to loop through each item in a multi-value variable
    * 'for' loop
    * 'while' loop

# Problem

This additional functionality will allow users to easily access individual elements of multi-value variables, and perform repetitive code within loops that can be dynamically constructed.  Additional string functions allow searching for and extracting substrings.  Reverse can be used to find the last occurrence of a substring rather than just the first.

* JIRA ticket (_optional_): PICARD-1716

# Solution

Suggested changes to the Scripting page of the Picard Website include:

**$find(haystack,needle)**

Find the location of the first occurrence of one string within another. Returns the zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.  Examples:

```
$findstr(abcdef,cde) returns 2
$findstr(abcdef,) returns 0
$findstr(abcdef,cdf) returns -1
```

**$reverse(text)**

Returns text in reverse order.  Example:

```
$reverse(abcde) returns 'edcba'
```

**$substr(text,start_index,end_index)**

Extract a specified portion of a string.  Returns the substring beginning with the character at the start index, up to (but not including) the character at the end index.  The first character is at index number 0.  If the start index is left blank, it defaults to the first character in the string.  If the end index is left blank, it defaults to the number of characters in the string.  If either index is negative, it is subtracted from the total number of characters in the string to provide the index used.  Examples:

```
$substr(abcde,1,2) returns 'b'
$substr(abcde,,2) returns 'ab'
$substr(abcde,1,) returns 'bcde'
$substr(abcde,,) returns 'abcde'
$substr(abcde,-2,) returns 'de'
$substr(abcde,-4,-2) returns 'bc'
```

**$get_multi_item(multi,item_index)**

Returns the value of the item at the specified index in the multi-value variable.  Index values are zero-based.  Example:

```
$get_multi_item(a; b; c; d; e,2) returns 'c'
```

**$foreach(multi,loop_code,_loop_value="_loop_value",_loop_count="_loop_count")**

Iterates over each element found in the specified multi-value variable, executing the specified code. For each loop, the element value is first stored in the tag specified by _loop_value and the count is stored in the tag specified by _loop_count.  This allows the element or count value to be accessed within the code script.  Example:

```
$foreach(isrc; year; comment; artists,$unset(%_loop_value%))
```

**$for(name,start,end,increment,loop_code)**

Standard 'for' loop.  Sets the value of name to the start value (integer) and loops through the specified code until the value of name is greater than or equal to the end value (integer).  At the end of each loop, the value of name is incremented by the specified increment value (integer).  Example:

```
$set(testing,Testing)
$for(_count,1,4,1,$set(testing,%testing% %_count%...))
```

**$while_loop(condition,loop_code)**

Standard 'while' loop.  Loops through the loop_code until the condition is no longer True.  Example:

```
$set(testing,Echo)
$while($lt($len(%testing%,20)),$set(testing,%testing% echo...))
```

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -914,11 +914,11 @@ def func_find_str(parser, haystack, needle):
     """Find the location of the first occurrence of one string within another.
 
     Arguments:
-        haystack {str} -- The string to search
-        needle {str} -- The substring to find
+        haystack: The string to search
+        needle: The substring to find
 
     Returns:
-        str -- Returns the zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.
+        The zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.
     """
     return str(haystack.find(needle))
 
@@ -927,10 +927,10 @@ def func_reverse_str(parser, text):
     """Returns 'text' in reverse order.
 
     Arguments:
-        text {str} -- Text to be processed
+        text: Text to be processed
 
     Returns:
-        str -- Text in reverse order.
+        Text in reverse order.
     """
     return text[::-1]
 
@@ -939,112 +939,128 @@ def func_substr(parser, text, start_index, end_index):
     """Extract a specified portion of a string.
 
     Arguments:
-        text {str} -- The string from which the extract will be made.
-        start_index {int} -- Index of the first character to extract.
-        end_index {int} -- Index of the first character that will not be extracted.
+        text: The string from which the extract will be made.
+        start_index: Index of the first character to extract.
+        end_index: Index of the first character that will not be extracted.
 
     Returns:
-        str -- Returns the substring beginning with the character at the start index,
-               up to (but not including) the character at the end index.  The first
-               character is at index number 0.  If the start index is left blank, it
-               defaults to the first character in the string.  If the end index is
-               left blank, it defaults to the number of characters in the string.
-               If either index is negative, it is subtracted from the total number of
-               characters in the string to provide the index used.
+        Returns the substring beginning with the character at the start index,
+        up to (but not including) the character at the end index.  The first
+        character is at index number 0.  If the start index is left blank, it
+        defaults to the first character in the string.  If the end index is
+        left blank, it defaults to the number of characters in the string.
+        If either index is negative, it is subtracted from the total number of
+        characters in the string to provide the index used.
     """
-    _start = int(start_index) if start_index else None
-    _end = int(end_index) if end_index else None
-    return text[_start:_end]
+    try:
+        start = int(start_index) if start_index else None
+    except ValueError:
+        start = None
+    try:
+        end = int(end_index) if end_index else None
+    except ValueError:
+        end = None
+    return text[start:end]
 
 
-def func_get_multi_item(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
+def func_get_multi(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
     """Returns value of the item at the specified index in the multi-value variable.  Index values are zero-based.
 
     Arguments:
-        multi {multi-value} -- The multi-value from which the item is to be retrieved.
-        item_index {int} -- The zero-based index of the item to be retrieved.
+        multi: The multi-value from which the item is to be retrieved.
+        item_index: The zero-based index of the item to be retrieved.
 
     Returns:
-        str -- Returns value of the item at the specified index in the multi-value variable.  Index values are zero-based.
+        Returns the value of the item at the specified index in the multi-value variable.
     """
     if item_index:
-        _index = int(item_index.eval(parser))
+        try:
+            index = int(item_index.eval(parser))
+        except ValueError:
+            return ''
     else:
         return ''
-    _multi = _get_multi_values(parser, multi, separator)
+    multi_var = _get_multi_values(parser, multi, separator)
     try:
-        return str(_multi[_index])
+        return str(multi_var[index])
     except IndexError:
         return ''
 
 
-def func_foreach_multi_item(parser, multi, loop_code, _loop_value="_loop_value", _loop_count="_loop_count", separator=MULTI_VALUED_JOINER):
-    """Iterates over each element found in the specified multi-value variable, executing the specified code.
+def func_foreach_multi(parser,
+                       multi,
+                       loop_code,
+                       separator=MULTI_VALUED_JOINER):
+    """Iterates over each element found in the specified multi-value variable.
+
+    Iterates over each element found in the specified multi-value variable, executing the specified code.
     For each loop, the element value is first stored in the tag specified by _loop_value and the count is
     stored in the tag specified by _loop_count.  This allows the element or count value to be accessed within
     the code script.
 
     Arguments:
-        multi {multi-value} -- The multi-value to be iterated.
-        loop_code {str} -- Script code to be processed on each iteration.
-
-    Keyword Arguments:
-        _loop_value {str} -- Tag name to use for the element for the current loop (default: {"_loop_value"})
-        _loop_count {str} -- Tag name to use for the count of the current loop (default: {"_loop_count"})
+        multi: The multi-value to be iterated.
+        loop_code: Script code to be processed on each iteration.
     """
     loop_count = 0
-    _multi = _get_multi_values(parser, multi, separator)
-    for value in _multi:
+    multi_value = _get_multi_values(parser, multi, separator)
+    for value in multi_value:
         loop_count += 1
-        func_set(parser, _loop_count, str(loop_count))
-        func_set(parser, _loop_value, str(value))
+        func_set(parser, '_loop_count', str(loop_count))
+        func_set(parser, '_loop_value', str(value))
         loop_code.eval(parser)
-    func_unset(parser, _loop_count)
-    func_unset(parser, _loop_value)
-    return ''
-
-
-def func_for_loop(parser, name, start, end, increment, loop_code):
-    """Standard Python 'for' loop.
-
-    Sets the value of name to the start value (integer) and loops through the specified code until the value
-    of name is greater than or equal to the end value (integer).  At the end of each loop, the value of name is
-    incremented by the specified increment value (integer).
-
-    Arguments:
-        name {str} -- Tag name to use for the loop count.  This can be accessed within the code block.
-        start {int} -- The initial value for the loop counter.
-        end {int} -- The value at which the loop exits.
-        increment {int} -- The value by which the loop count variable is incremented for each loop.
-        loop_code {str} -- Script code to be processed on each iteration.
-    """
-    if name and start and end and increment and loop_code:
-        _count = int(start.eval(parser))
-        _end = int(end.eval(parser))
-        _inc = int(increment.eval(parser))
-        _name = str(name.eval(parser))
-        if _inc > 0:
-            while _count < _end:
-                func_set(parser, _name, str(_count))
-                loop_code.eval(parser)
-                _count += _inc
-            func_unset(parser, _name)
+    func_unset(parser, '_loop_count')
+    func_unset(parser, '_loop_value')
     return ''
 
 
 def func_while_loop(parser, condition, loop_code):
-    """Standard Python 'while' loop.  Also includes a runaway check to limit the maximum number of iterations.
+    """Standard 'while' loop.  Also includes a runaway check to limit the maximum number of iterations.
 
     Arguments:
-        condition {str} -- Script code to check before each iteration through the loop.
-        loop_code {str} -- Script code to be processed on each iteration.
+        condition: Script code to check before each iteration through the loop.
+        loop_code: Script code to be processed on each iteration.
     """
     runaway_check = 1000
+    loop_count = 0
     if condition and loop_code:
-        while condition.eval(parser) and runaway_check:
+        while condition.eval(parser) and loop_count < runaway_check:
+            loop_count += 1
+            func_set(parser, '_loop_count', str(loop_count))
             loop_code.eval(parser)
-            runaway_check -= 1
+        func_unset(parser, '_loop_count')
     return ''
+
+
+def func_map_multi(parser,
+                   multi,
+                   loop_code,
+                   separator=MULTI_VALUED_JOINER):
+    """Iterates over each element found in the specified multi-value variable and updates the value.
+
+    Iterates over each element found in the specified multi-value variable and updates the value of the
+    element to the value returned by the specified code. For each loop, the element value is first stored in
+    the tag specified by _loop_value and the count is stored in the tag specified by _loop_count.  This
+    allows the element or count value to be accessed within the code script.
+
+    Arguments:
+        multi: The multi-value to be iterated.
+        loop_code: Script code to be processed on each iteration.
+
+    Returns the updated multi-value variable.
+    """
+    loop_count = 0
+    multi_value = _get_multi_values(parser, multi, separator)
+    for value in multi_value:
+        loop_count += 1
+        func_set(parser, '_loop_count', str(loop_count))
+        func_set(parser, '_loop_value', str(value))
+        multi_value[loop_count - 1] = str(loop_code.eval(parser))
+    func_unset(parser, '_loop_count')
+    func_unset(parser, '_loop_value')
+    if type(separator) is not str:
+        separator = separator.eval(parser)
+    return separator.join(multi_value)
 
 
 register_script_function(func_if, "if", eval_args=False)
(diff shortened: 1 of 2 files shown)


picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+        str -- Returns the substring beginning with the character at the start index,
+               up to (but not including) the character at the end index.  The first
+               character is at index number 0.  If the start index is left blank, it
+               defaults to the first character in the string.  If the end index is
+               left blank, it defaults to the number of characters in the string.
+               If either index is negative, it is subtracted from the total number of
+               characters in the string to provide the index used.
+    """
+    _start = int(start_index) if start_index else None
+    _end = int(end_index) if end_index else None
```
phw: This must catch `ValueError` in case the int conversion fails. I'd say we treat the value throwing the exception should just be considered `None`.

picard/script.py
```diff
@@ -967,3 +1104,10 @@ def func_is_video(parser):
 register_script_function(func_title, "title")
 register_script_function(func_is_audio, "is_audio")
 register_script_function(func_is_video, "is_video")
+register_script_function(func_find_str, "find")
+register_script_function(func_reverse_str, "reverse")
+register_script_function(func_substr, "substr")
+register_script_function(func_get_multi_item, "get_multi_item", eval_args=False)
```
phw: Could we just call it `get_multi` or maybe `get_item`?

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+
+    Arguments:
+        multi {multi-value} -- The multi-value from which the item is to be retrieved.
+        item_index {int} -- The zero-based index of the item to be retrieved.
+
+    Returns:
+        str -- Returns value of the item at the specified index in the multi-value variable.  Index values are zero-based.
+    """
+    if item_index:
+        _index = int(item_index.eval(parser))
```
phw: Again needs to fetch `ValueError`

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+        name {str} -- Tag name to use for the loop count.  This can be accessed within the code block.
+        start {int} -- The initial value for the loop counter.
+        end {int} -- The value at which the loop exits.
+        increment {int} -- The value by which the loop count variable is incremented for each loop.
+        loop_code {str} -- Script code to be processed on each iteration.
+    """
+    if name and start and end and increment and loop_code:
+        _count = int(start.eval(parser))
+        _end = int(end.eval(parser))
+        _inc = int(increment.eval(parser))
```
phw: `ValueError` ;)

picard/script.py
zas: `get_multi` has my preference, and we can rename `func_*_multi_item` to `func_*_multi` as well.

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+    else:
+        return ''
+    _multi = _get_multi_values(parser, multi, separator)
+    try:
+        return str(_multi[_index])
+    except IndexError:
+        return ''
+
+
+def func_foreach_multi_item(parser, multi, loop_code, _loop_value="_loop_value", _loop_count="_loop_count", separator=MULTI_VALUED_JOINER):
```
zas: Wrap arguments?

Sophist-UK:
@rdswift Bob - just a reminder to update the script function reference on the Picard Web Site repo once this has merged.

phw:
We should also have test cases for the new functions in `test_script.py`

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+    _end = int(end_index) if end_index else None
+    return text[_start:_end]
+
+
+def func_get_multi_item(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
+    """Returns value of the item at the specified index in the multi-value variable.  Index values are zero-based.
+
+    Arguments:
+        multi {multi-value} -- The multi-value from which the item is to be retrieved.
+        item_index {int} -- The zero-based index of the item to be retrieved.
```
samj1912: How does an `int` type have an `eval` method?

samj1912: Let's change the type from `int` to the right Python type. (ScriptVariable/Function)

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+
+    Arguments:
+        name {str} -- Tag name to use for the loop count.  This can be accessed within the code block.
+        start {int} -- The initial value for the loop counter.
+        end {int} -- The value at which the loop exits.
+        increment {int} -- The value by which the loop count variable is incremented for each loop.
+        loop_code {str} -- Script code to be processed on each iteration.
+    """
+    if name and start and end and increment and loop_code:
+        _count = int(start.eval(parser))
```
samj1912: nitpick: Can we remove the `_` prepend from the variables?

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+
+    Keyword Arguments:
+        _loop_value {str} -- Tag name to use for the element for the current loop (default: {"_loop_value"})
+        _loop_count {str} -- Tag name to use for the count of the current loop (default: {"_loop_count"})
+    """
+    loop_count = 0
+    _multi = _get_multi_values(parser, multi, separator)
+    for value in _multi:
+        loop_count += 1
+        func_set(parser, _loop_count, str(loop_count))
```
zas: `loop_count` and `_loop_count` ?
Using '_' prefix is mainly for private functions, or unused variables (though that's just a convention).
Also you may want to use `enumerate()`.

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+    of name is greater than or equal to the end value (integer).  At the end of each loop, the value of name is
+    incremented by the specified increment value (integer).
+
+    Arguments:
+        name {str} -- Tag name to use for the loop count.  This can be accessed within the code block.
+        start {int} -- The initial value for the loop counter.
+        end {int} -- The value at which the loop exits.
+        increment {int} -- The value by which the loop count variable is incremented for each loop.
+        loop_code {str} -- Script code to be processed on each iteration.
+    """
```
zas: This function may not be that useful, it looks over complicated.

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+
+
+def func_while_loop(parser, condition, loop_code):
+    """Standard Python 'while' loop.  Also includes a runaway check to limit the maximum number of iterations.
+
+    Arguments:
+        condition {str} -- Script code to check before each iteration through the loop.
+        loop_code {str} -- Script code to be processed on each iteration.
+    """
+    runaway_check = 1000
```
zas: You should move this inside `if` block.

picard/script.py
rdswift: Updated to include the ValueError check.

picard/script.py
rdswift: Changed to $get_multi and func_get_multi.

picard/script.py
rdswift: Check added.

picard/script.py
rdswift: Check added.

picard/script.py
rdswift: I removed the option for different loop tag names (because I couldn't think of a case where alternate loop tag names would be necessary), and wrapped the rest.

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
         return ""
 
 
+def func_find_str(parser, haystack, needle):
+    """Find the location of the first occurrence of one string within another.
+
+    Arguments:
+        haystack {str} -- The string to search
```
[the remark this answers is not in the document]
rdswift: Docstrings updated.  Tried to follow the guide that you linked.

picard/script.py
```diff
@@ -910,6 +910,143 @@ def func_is_video(parser):
...
+    Returns:
+        str -- Returns the substring beginning with the character at the start index,
+               up to (but not including) the character at the end index.  The first
+               character is at index number 0.  If the start index is left blank, it
+               defaults to the first character in the string.  If the end index is
+               left blank, it defaults to the number of characters in the string.
+               If either index is negative, it is subtracted from the total number of
+               characters in the string to provide the index used.
+    """
+    _start = int(start_index) if start_index else None
```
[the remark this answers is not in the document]
rdswift: No reason at all.  They should all have the leading underscores removed now.

picard/script.py
rdswift: I actually removed the type designations altogether, because they didn't seem to be broken out in the example in the style guide you linked.

picard/script.py
rdswift: Done.

picard/script.py
rdswift: Underscores removed.  I'll look at `enumerate()` (didn't notice that comment until after I submitted my update)..

picard/script.py
rdswift: Yeah, I only included it because there is a `for()` loop in Python.  I agree that everything that the `for()` loop provides can be done using the `while()` loop so I've removed the `for()` loop from the code.

picard/script.py
rdswift: Right.  Done.

picard/script.py
rdswift: Switched to use `enumerate()`.

rdswift:
I've removed the redundant and complicated `for()` loop, added tests for all of the new functions, and added in a first draft of a `map()` function.  Thanks for all the helpful feedback.

picard/script.py
rdswift: > Yeah, I only included it because there is a for() loop in Python.

Wrong.  I've been doing a lot of work in PHP lately, so that must have been what I was thinking of.  In any event, it's gone now.

phw: I guess Python not having this is what made me frown on  the for loop :D

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+    stored in the tag specified by _loop_count.  This allows the element or count value to be accessed within
+    the code script.
+
+    Arguments:
+        multi: The multi-value to be iterated.
+        loop_code: Script code to be processed on each iteration.
+    """
+    loop_count = 0
+    multi_value = _get_multi_values(parser, multi, separator)
+    for index, value in enumerate(multi_value):
```
zas: You can use `enumerate()` optional argument:

```python
 for loop_count, value in enumerate(multi_value, 1):
```

and get rid of `loop_count = 0` and `loop_count = index + 1`

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+    """Iterates over each element found in the specified multi-value variable and updates the value.
+
+    Iterates over each element found in the specified multi-value variable and updates the value of the
+    element to the value returned by the specified code. For each loop, the element value is first stored in
+    the tag specified by _loop_value and the count is stored in the tag specified by _loop_count.  This
+    allows the element or count value to be accessed within the code script.
+
+    Arguments:
+        multi: The multi-value to be iterated.
+        loop_code: Script code to be processed on each iteration.
```
zas: It has 4 arguments, but only 2 are documented here.

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+
+    Returns the updated multi-value variable.
+    """
+    loop_count = 0
+    multi_value = _get_multi_values(parser, multi, separator)
+    for index, value in enumerate(multi_value):
+        loop_count = index + 1
```
zas: Same remark: `enumerate(multi_value, 1)` and `multi_value[loop_count - 1] = ...` will simplify code a bit.

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+    func_unset(parser, '_loop_count')
+    func_unset(parser, '_loop_value')
+    return ''
+
+
+def func_while_loop(parser, condition, loop_code):
+    """Standard 'while' loop.  Also includes a runaway check to limit the maximum number of iterations.
+
+    Arguments:
+        condition: Script code to check before each iteration through the loop.
```
zas: First argument isn't documented

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+    """Iterates over each element found in the specified multi-value variable.
+
+    Iterates over each element found in the specified multi-value variable, executing the specified code.
+    For each loop, the element value is first stored in the tag specified by _loop_value and the count is
+    stored in the tag specified by _loop_count.  This allows the element or count value to be accessed within
+    the code script.
+
+    Arguments:
+        multi: The multi-value to be iterated.
+        loop_code: Script code to be processed on each iteration.
```
zas: Missing documentation for 2 arguments: `parser` and `separator`

picard/script.py
```diff
@@ -910,6 +910,159 @@ def func_is_video(parser):
...
+            index = int(item_index.eval(parser))
+        except ValueError:
+            return ''
+    else:
+        return ''
+    multi_var = _get_multi_values(parser, multi, separator)
+    try:
+        return str(multi_var[index])
+    except IndexError:
+        return ''
```
zas: I think it could be rewritten to a simpler form:

```python
    if not item_index:
        return ''
    try:
        index = int(item_index.eval(parser))
        multi_var = _get_multi_values(parser, multi, separator)
        return str(multi_var[index])
    except (ValueError, IndexError):
        return ''
```

picard/script.py
rdswift: That makes perfect sense.  Sorry I missed that.  I'll change it.

picard/script.py
rdswift: Should I remove all the arguments from the docstrings instead?  I see that none of the other functions include the arguments list.

picard/script.py
rdswift: Will change.

picard/script.py
rdswift: Much cleaner!  I'll make the change.

rdswift:
A few more questions...

Now that we've removed the `for()` function, should we rename the new `foreach()` to just `for()` to keep it in line with `for()` in Python?

When returning the new multi-value in the `map()` function, it it is called with a separator override should it use the standard separator in the return string or should it use the override separator?

Is there any value in adding two more functions:  `$join(multi,join_phrase,separator='; ')` to return a string of the multi elements joined by the join phrase; and `$sub_multi(multi,start,end,separator='; ')` to return a slice of the original multi?  This would pretty much round out the functionality set.

Thanks.

picard/script.py
zas: Nope, better have more documentation

picard/script.py
```diff
@@ -910,6 +910,163 @@ def func_is_video(parser):
...
+    Arguments:
+        parser: The ScriptParser object used to parse the script.
+        multi: The multi-value to be iterated.
+        loop_code: String of script code to be processed on each iteration that yields the new value for
+            the multi-value element.
+        separator: String used to separate the elements in the multi-value.
+
+    Returns the updated multi-value variable.
+    """
+    loop_count = 0
```
zas: You don't need to initialize `loop_count` here since it is done in the for loop.

rdswift: Crap!  I missed that.  Thanks.

zas:
> Now that we've removed the `for()` function, should we rename the new `foreach()` to just `for()` to keep it in line with `for()` in Python?

You mean `foreach_multi` vs `for_multi` ?
I'd say keep it as is (`foreach_multi`).

> 
> When returning the new multi-value in the `map()` function, if it is called with a separator override should it use the standard separator in the return string or should it use the override separator?
> 

Code currently returns string using  override separator, AFAIK, it seems correct to me.


> Is there any value in adding two more functions: `$join(multi,join_phrase,separator='; ')` to return a string of the multi elements joined by the join phrase; and `$sub_multi(multi,start,end,separator='; ')` to return a slice of the original multi? This would pretty much round out the functionality set.

If you think they can be useful, why not. Though I'd do it in another PR, and perhaps name them using suffix `_multi`: `join_multi()` and `slice_multi()` (`sub` is misleading imho).

PR #1449, opened 2020-01-30
PICARD-1717: Backup config on upgrade

phw:
# Summary
Make an update of existing config file on upgrade.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Picard's configuration will get updated if necessary on updates, but there is no support for downgrade. That means older versions of Picard might not be able to run with config files created by newer versions and an update might make the config unusable by older versions.

* JIRA ticket (_optional_): PICARD-1717

# Solution
1. When running a version that is newer then the loaded config file, make a copy of the current config with a name of `Picard-x.y.z.ini`, where `x.y.z` is the old version. This ensures old config gets saved before it is altered, giving the user the opportunity to restore the old config if needed.
2. When starting Picard and there is an existing `Picard-x.y.z.ini` file matching the current `PICARD_VERSION`, use this instead of the default `Picard.ini`. This ensures the user can easily roll back to an older Picard version or use an old and new Picard side by side.

Both of the above only is applied when using the default app config. One can still launch with the `-c` parameter to load a specific file.

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -18,6 +18,8 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from operator import itemgetter
+import os
+import shutil
 
 from PyQt5 import QtCore
 
@@ -146,6 +148,7 @@ class Config(QtCore.QSettings):
             this.sync()
 
         this.__initialize()
+        this._backup_settings()
         return this
 
     @classmethod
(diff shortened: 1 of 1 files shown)


picard/config.py
```diff
@@ -234,10 +242,25 @@ def run_upgrade_hooks(self, outputfunc=None):
...
             self._write_version()
 
+    def _backup_settings(self):
+        if Version(0, 0, 0) < self._version < PICARD_VERSION:
+            backup_path = self._versioned_config_filename()
+            log.info('Backing up config file to %s', backup_path)
+            try:
+                shutil.copyfile(self.fileName(), backup_path)
+            except OSError:
+                log.error('Failed backing up config file to %s', backup_path)
```
zas: @phw : I'd also catch and ignore other exceptions, to ensure this never prevents Picard to start.
History of `shutil.copyfile()` shows evolution of raised exceptions, and we can't be sure new ones will not appear in future Python versions.
What do you think?

phw: Let's keep it more explicit for now. I will submit a separate PR to handle unexpected exceptions better and provide better user feedback for those cases.

PR #1453, opened 2020-02-01
Unify new and old script function names

zas:
Before introduction of new functions, we had a pattern: "name" script function was defined by "func_name"
Recent patches changed this, introducing inconsistencies.
Also we have inmulti and lenmulti, so stick to this scheme for getmulti and others.

func_find_str -> func_find
find_str -> find
func_reverse_str -> func_reverse
reverse_str -> reverse
func_get_multi -> func_getmulti
func_foreach_multi -> func_foreach
func_while_loop -> func_while
func_map_multi -> func_map

A new decorator named `script_function()` is introduced to simplify registration of script functions.

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -910,7 +910,7 @@ def func_is_video(parser):
         return ""
 
 
-def func_find_str(parser, haystack, needle):
+def func_find(parser, haystack, needle):
     """Find the location of the first occurrence of one string within another.
 
     Arguments:
@@ -924,7 +924,7 @@ def func_find_str(parser, haystack, needle):
     return str(haystack.find(needle))
 
 
-def func_reverse_str(parser, text):
+def func_reverse(parser, text):
     """Returns 'text' in reverse order.
 
     Arguments:
@@ -966,7 +966,7 @@ def func_substr(parser, text, start_index, end_index):
     return text[start:end]
 
 
-def func_get_multi(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
+def func_getmulti(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
     """Returns value of the item at the specified index in the multi-value variable.  Index values are zero-based.
 
     Arguments:
@@ -988,10 +988,7 @@ def func_get_multi(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
         return ''
 
 
-def func_foreach_multi(parser,
-                       multi,
-                       loop_code,
-                       separator=MULTI_VALUED_JOINER):
+def func_foreach(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
     """Iterates over each element found in the specified multi-value variable.
 
     Iterates over each element found in the specified multi-value variable, executing the specified code.
@@ -1015,7 +1012,8 @@ def func_foreach_multi(parser,
     return ''
 
 
-def func_while_loop(parser, condition, loop_code):
+@script_function(eval_args=False)
+def func_while(parser, condition, loop_code):
     """Standard 'while' loop.  Also includes a runaway check to limit the maximum number of iterations.
 
     Arguments:
@@ -1034,10 +1032,7 @@ def func_while_loop(parser, condition, loop_code):
     return ''
 
 
-def func_map_multi(parser,
-                   multi,
-                   loop_code,
-                   separator=MULTI_VALUED_JOINER):
+def func_map(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
     """Iterates over each element found in the specified multi-value variable and updates the value.
 
     Iterates over each element found in the specified multi-value variable and updates the value of the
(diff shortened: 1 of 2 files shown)


Sophist-UK:
I may have misread the code, but it seems to me that any scripts with e.g. find_str in them are going to fail after this PR.

If this is so, do we either need to maintain backward compatibility or provide an upgrade function to modify any existing scripts?

Sophist-UK:
P.S. I love the idea of using a decorator instead of having a separate call to register the function.

zas:
> I may have misread the code, but it seems to me that any scripts with e.g. find_str in them are going to fail after this PR.

Those script functions were added in https://github.com/metabrainz/picard/commit/2bd872dc89694aa8069c3c4fa8c0d3115f90a752  3 days ago, so that's unlikely it is widely used yet.

This is why this has to be fixed now.

phw:
> If this is so, do we either need to maintain backward compatibility or provide an upgrade function to modify any existing scripts?

We don't provide backward compatibility for development branch and those script functions have not been released yet.

picard/script.py
```diff
@@ -988,7 +1076,8 @@ def func_get_multi(parser, multi, item_index, separator=MULTI_VALUED_JOINER):
         return ''
 
 
-def func_foreach_multi(parser,
+@script_function(eval_args=False)
+def func_foreachmulti(parser,
```
phw: I think `foreach`, `while` and `map` are fine. `foreachmulti` looks strange and the `multi` is not really needed here.

picard/script.py
```diff
@@ -339,6 +339,33 @@ def register_script_function(function, name=None, eval_args=True,
...
+                    By default, ``func_foo`` will create ``foo`` script function
+        ``debug``: if set to True, print passed parameters
+
+    Exemple:
+        @script_function(eval_args=False)
+        def func_myscriptfunc():
+            ...
+    """
+    def script_function_decorator(func):
+        nonlocal name, eval_args, check_argcount, prefix, debug
```
phw: I haven't run the code, but I don't see a need for `nonlocal` here.

Sophist-UK:
Fair enough!!

picard/script.py
zas: Nor I, but python complain with UnboundLocalError without

picard/script.py
zas: Well, since those only apply to "multi", I just kept the style of `inmulti` and `lenmulti`.

picard/script.py
```diff
@@ -339,6 +339,33 @@ def register_script_function(function, name=None, eval_args=True,
...
+def script_function(name=None, eval_args=True, check_argcount=True, prefix='func_', debug=False):
+    """Decorator helper to register script functions
+
+    It calls ``register_script_function()`` and share same arguments
+    Extra optional arguments:
+        ``prefix``: define the prefix to be removed from defined function to name script function
+                    By default, ``func_foo`` will create ``foo`` script function
+        ``debug``: if set to True, print passed parameters
+
+    Exemple:
```
phw: Type "Exemple" = "Example"

picard/script.py
phw: It's needed for `name`, because that gets manipulated after having been read. So `nonlocal name` should be sufficient.

picard/script.py
phw: Thanks for the change, I think that reads better.

IMHO those names are also awkward :D But were needed for distinguishing from existing `$in` and `$len`. But maybe we should have used actually different names, e.g. `contains` and `count`. Not sure about the first one, but `count` sounds actually much better to me then `lenmulti`. Maybe we should consider to add an alias for this (and deprecate `$lenmulti`), what do you think?

zas: Yup, we could create aliases and deprecate, but imho we'll have to keep `lenmulti` and `inmulti` for a long time... not sure it is worth the hassle.

picard/script.py
zas: Yup, I simplified, and removed debug parameter.

picard/script.py
zas: That's not a typo, that's French ;) fixed & rebased

rdswift:
> Before introduction of new functions, we had a pattern: "name" script function was defined by "func_name"
Recent patches changed this, introducing inconsistencies.

Sorry for causing a problem with the naming.  I wonder if it's worth including a short note in the module explaining the pattern to be used so that others don't accidentally make the same mistakes I did because they didn't know?

BTW, I :heart:  the decorator idea.

zas:
>  I wonder if it's worth including a short note in the module explaining the pattern to be used so that others don't accidentally make the same mistakes I did because they didn't know?

Yes, that's a good idea.
BTW, we reviewed and approved without noticing, so "not your fault" ;)

> BTW, I heart the decorator idea.

Thanks.

PR #1455, opened 2020-02-04
PICARD-1726: Fix crash when closing options while loading plugins

phw:
# Summary
Fix crash when closing options while loading plugins

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Closing options while the web request for plugins is running causes Picard to crash with

    RuntimeError: wrapped C/C++ object of type QTreeWidget has been deleted

That's because the C++ object has already been deleted while the callback still has a reference to the Python wrapper objects.

See also http://enki-editor.org/2014/08/23/Pyqt_mem_mgmt.html

* JIRA ticket (_optional_): PICARD-1726

# Solution
Work around Qt C++ objects getting deleted before request callback gets called by connecting to the destroyed signal and setting a deleted state.

This feels kind of hacky, but I could not find a clean way to otherwise detect if an object is deleted. This fixes the issue and I think is an acceptable workaround. But I would welcome any more elegant solution :)

# Action

### DIFF
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -235,6 +235,16 @@ class PluginsOptionsPage(OptionsPage):
         self.ui.folder_open.clicked.connect(self.open_plugin_dir)
         self.ui.reload_list_of_plugins.clicked.connect(self.reload_list_of_plugins)
 
+        # Keep track whether the object has been destroyed to avoid trying to update
+        # deleted UI widgets after plugin list refresh.
+        self._deleted = False
+
+        # The on destroyed cannot be created as a method on this class or it will never get called.
+        # See https://stackoverflow.com/questions/16842955/widgets-destroyed-signal-is-not-fired-pyqt
+        def on_destroyed(obj=None):
+            self._deleted = True
+        self.destroyed.connect(on_destroyed)
+
         self.manager = self.tagger.pluginmanager
         self.manager.plugin_installed.connect(self.plugin_installed)
         self.manager.plugin_updated.connect(self.plugin_updated)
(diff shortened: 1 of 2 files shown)


picard/ui/options/plugins.py
```diff
@@ -235,6 +235,16 @@ def __init__(self, parent=None):
...
+        # Keep track whether the object has been destroyed to avoid trying to update
+        # deleted UI widgets after plugin list refresh.
+        self._deleted = False
+
+        # The on destroyed cannot be created as a method on this class or it will never get called.
+        # See https://stackoverflow.com/questions/16842955/widgets-destroyed-signal-is-not-fired-pyqt
+        def on_destroyed(obj=None):
+            self._deleted = True
+        self.destroyed.connect(on_destroyed)
+
```
zas: I wonder if it wouldn't make sense to have this at `OptionsPage` class level.
After all, this issue isn't specific to plugins options pages, it has impact on it because it has running queries at the time it is closed, but this could be the case for any other options pages. In this case, `_delete` could be renamed `delete` and marked as property.

phw: You are right in general. But I checked the code and currently this is the only place where we call some threaded background code with a callback. And as I think this is kind of ugly I would prefer to contain this to the place where it is needed unless we need it somewhere else as well.

One argument against this would be if we want to make this available to plugins. But this would mean making it a somewhat official API. 

So not sure really. If you think it's better to have it centralized I can live with that as well :)

zas: Well, we can move it at any time, though making it available for any options pages (including those from plugins) would help in similar case (where callback's calls are delayed and page closed), so that's not purely a plugins options page thing. It seems more logical to me it is a feature of the parent class.
Even though I don't know any other options pages that might be concerned by this issue, we are pretty sure some plugins are actually doing queries using web service, and those are very likely to be concerned by this issue.

phw: Ok, convinced :) e29315c2fd61ad140f798eb57ce741c81e189b7c

PR #1458, opened 2020-02-04
PICARD-1729: Add $datetime() function and associated tests

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Add a scripting function to return the current date and time.  This would be useful to record the last time the metadata was updated in a file.

# Problem

There is currently no way to retrieve the current date and time to a tag.  This could be useful for including the timestamp of the last metadata update to a file.

* JIRA ticket: PICARD-1729

# Solution

Added the `$datetime()` scripting function to return the current date and time.  This function accepts an optional parameter to set the format string (using the strftime formatting).

# Action

Another pull request to update the `Scripting` documentation on Picard Website will be prepared if this addition is accepted.

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -21,6 +21,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from collections import namedtuple
+import datetime
 from functools import reduce
 from inspect import getfullargspec
 import operator
@@ -1201,3 +1202,28 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
         return separator.join(multi_var[start:end])
     except IndexError:
         return ''
+
+
+@script_function()
+def func_datetime(parser, format=None):
+    """Return the current date and time as a string.
+
+    Arguments:
+        parser: The ScriptParser object used to parse the script.
+        format: A string or the ScriptVariable/Function that evaluates to the
+            string used to format the output.  Default is '%Y-%m-%d %H:%M:%S'
+            if blank.  Uses strftime() format.
+
+    Returns:
+        Returns the current date and time as a string.
+    """
+    # local_tz required for Python 3.5 which does not allow setting astimezone()
+    # on a naive datetime.datetime object.  This provides timezone information to
+    # allow the use of %Z and %z in the output format.
+    local_tz = datetime.datetime.now(datetime.timezone.utc).astimezone().tzinfo
+
+    # Handle case where format evaluates to ''
+    if not format:
+        format = '%Y-%m-%d %H:%M:%S'
+
+    return datetime.datetime.now(tz=local_tz).strftime(format)
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1,3 +1,4 @@
+import datetime
 from unittest.mock import MagicMock
 
 from test.picardtestcase import PicardTestCase
@@ -14,6 +15,38 @@ from picard.script import (
     ScriptUnknownFunction,
     register_script_function,
 )
+
+
+class _TestTimezone(datetime.tzinfo):
+    def utcoffset(self, dt):
+        # Set to GMT+2
+        return datetime.timedelta(hours=2) + self.dst(dt)
+
+    def dst(self, dt):
+        d = datetime.datetime(dt.year, 4, 1)
+        self.dston = d - datetime.timedelta(days=d.weekday() + 1)
+        d = datetime.datetime(dt.year, 11, 1)
+        self.dstoff = d - datetime.timedelta(days=d.weekday() + 1)
+        if self.dston <=  dt.replace(tzinfo=None) < self.dstoff:
+            return datetime.timedelta(hours=1)
+        else:
+            return datetime.timedelta(0)
+
+    def tzname(self,dt):
+        return "TZ Test"
+
+
+class _DateTime(datetime.datetime):
+    @classmethod
+    def now(cls, tz=None):
+        return datetime.datetime(2020, 1, 2, hour=12, minute=34, second=56, microsecond=789, tzinfo=tz)
+
+    def astimezone(self, tz=None):
+        # Ignore tz passed to the method and force use of test timezone.
+        tz = _TestTimezone()
+        utc = (self - self.utcoffset()).replace(tzinfo=tz)
+        # Convert from UTC to tz's local time.
+        return tz.fromutc(utc)
 
 
 class ScriptParserTest(PicardTestCase):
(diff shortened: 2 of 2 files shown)


picard/script.py
```diff
@@ -1201,3 +1201,22 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
...
+            string used to format the output.  Uses strftime() format.
+
+    Returns:
+        Returns he current date and time as a string.
+    """
+    import datetime
+    # Handle case where format evaluates to ''
+    if not format:
+        format = '%Y-%m-%d %H:%M:%S'
+    return datetime.datetime.now().strftime(format)
```
zas: I'd use `datetime.now().astimezone().strftime(format)`:

```python
>>> datetime.now().astimezone().strftime("%z")
'+0100'
>>> datetime.now().strftime("%z")
''
```

Not sure it works with Python3.5 though (@phw ?)

test/test_script.py
```diff
@@ -943,3 +943,23 @@ def test_cmd_slice(self):
             self.parser.eval("$slice(abc; def,0)")
         with self.assertRaisesRegex(ScriptError, areg):
             self.parser.eval("$slice(abc; def),0,1,:,extra")
+
+    def test_cmd_datetime(self):
+        context = Metadata()
+        context["foo"] = "%Y%m"
+        # Tests with context
+        self.assertScriptResultEquals("$len($datetime(%foo%))", "6", context)
```
zas: I wonder if it would make sense to have an optional parameter to pass a defined date & time to `$datetime()` so it wouldn't use `now()`. We could then test the exact returned value (a string representing a known time) and get rid of those `$len()`.

@phw what do you think?

picard/script.py
```diff
@@ -1201,3 +1201,22 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
         return separator.join(multi_var[start:end])
     except IndexError:
         return ''
+
+
+@script_function()
+def func_datetime(parser, format='%Y-%m-%d %H:%M:%S'):
```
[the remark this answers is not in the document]
rdswift: Makes sense.  Wish I would have seen that.

picard/script.py
```diff
@@ -1201,3 +1201,22 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
...
+    """Return the current date and time as a string.
+
+    Arguments:
+        parser: The ScriptParser object used to parse the script.
+        format: A string or the ScriptVariable/Function that evaluates to the
+            string used to format the output.  Uses strftime() format.
```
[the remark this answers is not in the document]
rdswift: Added something.  Is this what you were thinking?

picard/script.py
```diff
@@ -1201,3 +1201,22 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
...
+
+    Returns:
+        Returns he current date and time as a string.
+    """
+    import datetime
```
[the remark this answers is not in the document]
rdswift: Moved to the top of the file.

picard/script.py
rdswift: Looking at the python docs it looks like it should.  I added it and we'll see if it passes the tests.

test/test_script.py
rdswift: Added something.  For now I limited the input to the '%Y-%m-%d %H:%M:%S' format.  I thought about using the override format if supplied but opted not.  Can change this if you think it would be more consistent that way.

rdswift:
It looks like the `astimezone()` processing on naive instances wasn't implemented until Python 3.6.  I'll have to do some experimenting to see if I can somehow force it to use the local timezone information.

picard/script.py
```diff
@@ -1201,3 +1202,32 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
...
+    # Handle case where format evaluates to ''
+    if not format:
+        format = '%Y-%m-%d %H:%M:%S'
+    if preset:
+        try:
+            time_now = datetime.datetime.strptime(preset, '%Y-%m-%d %H:%M:%S')
+        except ValueError:
+            time_now = datetime.datetime.now()
+    else:
+        time_now = datetime.datetime.now()
```
zas: I'm not sure to understand this new code.
Also this is repeated, it should be possible to factorise.
The string `'%Y-%m-%d %H:%M:%S'` is still repeated.

Does this code actually provide timezone information (`%z` and '%Z') ?

test/test_script.py
zas: What I was thinking about is to pass a fixed timestamp and ensure it results in something consistent.

picard/script.py
rdswift: The short answer is no, this does not provide the timezone information.  I am still working on trying to do that, but wanted to leave the code in a (semi) working state while I looked for solutions.  The problem is that Python 3.5 does not allow applying the `astimezone()` method to a niave `datetime` object, and I have yet to come up with a reliable means of getting the local `tzinfo` to use when creating the `datetime` object using `strptime`.

rdswift: I found a way to get the local timezone to set it when creating the `datetime` objects.  This should now provide the timezone information to allow the `%Z` and `%z` format specifiers to work properly.

test/test_script.py
rdswift: I think I may be misunderstanding.  What I've added was the ability to set the date/time used in the function output as an optional input to the function.  I'm now thinking that you might have been looking for a way to pass a fixed time not to the function itself, but to the test somehow.  If that's it, then I need to go away and think about how that might be done.

rdswift: I think I understand now.  Have changed the testing to use a fixed date/time.  Is this what you were thinking?

rdswift:
I think that's got it.  I seem to have turned that into a lot more work than was necessary through my misunderstanding.  Sorry for the extra work it caused in reviewing.

test/test_script.py
```diff
@@ -943,3 +950,31 @@ def test_cmd_slice(self):
...
             self.parser.eval("$slice(abc; def),0,1,:,extra")
+
+    def test_cmd_datetime(self):
+        # Save origninal datetime object and substitute one returning
+        # a fixed now() value for testing.
+        original_datetime = datetime.datetime
+        datetime.datetime = _DateTime
+
+        context = Metadata()
+        context["foo"] = "%Y%m%d%H%M%S.%f"
```
zas: If your intent is to restore original `datetime.datetime` whatever happens, those two lines should also be in the try/finally block.

test/test_script.py
```diff
@@ -943,3 +950,31 @@ def test_cmd_slice(self):
...
+            # Tests with static input
+            self.assertScriptResultEquals("$datetime(\%Y\%m\%d\%H\%M\%S.\%f)", "20200102123456.000789", context)
+            # Tests with missing input
+            self.assertScriptResultEquals("$datetime()", "2020-01-02 12:34:56", context)
+            # Tests with invalid format
+            self.assertScriptResultEquals("$datetime(xxx)", "xxx", context)
+            # Tests with invalid number of arguments
+            areg = r"^Wrong number of arguments for \$datetime: Expected between 0 and 1, "
+            with self.assertRaisesRegex(ScriptError, areg):
+                self.parser.eval("$datetime(abc,def)")
```
zas: I don't see any test for timezone correctness in there, do I miss something?

test/test_script.py
rdswift: Good point.  I was thinking to only include the lines doing the testing, but I see that an exception thrown here would cause it not to be restored.

test/test_script.py
rdswift: Nope.  I'll have to add that in.

PR #1465, opened 2020-02-10
PICARD-1688: Submit AcoustID fingerprints in batches

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Avoids issues with exceeding the body size limit for submission requests. Also this makes the entire submission process more robust, as the submission of many fingerprints can now succeed partially instead of failing in its entirety.

* JIRA ticket (_optional_): PICARD-1688

# Solution
Submit fingerprints in batches of 240.

The upper limit is around 250 fingerprints, which is equivalent to roughly 1 MiB of data. But use a batch count of actually 240 fingerprints per submission to have some leeway for varying data.

# Action

### DIFF
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -37,6 +37,9 @@ class Submission(object):
 
 class AcoustIDManager(QtCore.QObject):
 
+    # AcoustID has a post limit of around 1 MB. With the data submitted by
+    # Picard this is roughly around 250 fingerprints. Submit a few less to have
+    # some leeway.
     BATCH_SUBMIT_COUNT = 240
 
     def __init__(self):
@@ -98,18 +101,19 @@ class AcoustIDManager(QtCore.QObject):
             self._check_unsubmitted()
             return
         submission_batch = submissions[:self.BATCH_SUBMIT_COUNT]
-        log.debug("AcoustID: submitting batch of %d fingerprints...", len(submission_batch))
+        submissions = submissions[self.BATCH_SUBMIT_COUNT:]
+        fingerprints = [fingerprint for file_, fingerprint in submission_batch]
+        log.debug("AcoustID: submitting batch of %d fingerprints (%d remaining)...",
+            len(submission_batch), len(submissions))
         self.tagger.window.set_statusbar_message(
             N_('Submitting AcoustIDs ...'),
             echo=None
         )
-        submissions = submissions[self.BATCH_SUBMIT_COUNT:]
-        fingerprints = [fingerprint for file_, fingerprint in submission_batch]
         next_func = partial(self._batch_submit, submissions)
         self.tagger.acoustid_api.submit_acoustid_fingerprints(fingerprints,
             partial(self._batch_submit_finished, submission_batch, next_func))
 
-    def _batch_submit_finished(self, submissions, next, document, http, error):
+    def _batch_submit_finished(self, submissions, next_func, document, http, error):
         if error:
             try:
                 error = load_json(document)
@@ -135,4 +139,4 @@ class AcoustIDManager(QtCore.QObject):
                 submission.orig_recordingid = submission.recordingid
                 file.update()
             self._check_unsubmitted()
-        next()
+        next_func()
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -37,6 +37,8 @@ class Submission(object):
 
 class AcoustIDManager(QtCore.QObject):
 
+    BATCH_SUBMIT_COUNT = 240
+
     def __init__(self):
         super().__init__()
         self._fingerprints = {}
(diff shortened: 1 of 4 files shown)


picard/acoustid/manager.py
```diff
@@ -82,16 +84,32 @@ def submit(self):
...
-        fingerprints = [fingerprint for file_, fingerprint in submissions]
+        submissions = submissions[self.BATCH_SUBMIT_COUNT:]
+        fingerprints = [fingerprint for file_, fingerprint in submission_batch]
+        next_func = partial(self._batch_submit, submissions)
         self.tagger.acoustid_api.submit_acoustid_fingerprints(fingerprints,
-            partial(self.__fingerprint_submission_finished, submissions))
+            partial(self._batch_submit_finished, submission_batch, next_func))
 
-    def __fingerprint_submission_finished(self, submissions, document, http, error):
+    def _batch_submit_finished(self, submissions, next, document, http, error):
```
zas: I'd prefer `next_func` instead of `next` since [`next()`](https://docs.python.org/3/library/functions.html#next) is a builtin.

picard/acoustid/manager.py
```diff
@@ -82,16 +84,32 @@ def submit(self):
...
+            log.debug("AcoustID: submitted all fingerprints")
+            self.tagger.window.set_statusbar_message(
+                N_('AcoustIDs successfully submitted.'),
+                echo=None,
+                timeout=3000
+            )
+            self._check_unsubmitted()
+            return
+        submission_batch = submissions[:self.BATCH_SUBMIT_COUNT]
+        log.debug("AcoustID: submitting batch of %d fingerprints...", len(submission_batch))
```
zas: I'd move those lines below and change debug message to show remaining fingerprints too.

```python
submission_batch = submissions[:self.BATCH_SUBMIT_COUNT]
submissions = submissions[self.BATCH_SUBMIT_COUNT:]
log.debug("AcoustID: submitting batch of %d fingerprints (%d remaining)...", len(submission_batch), len(submissions))
```

picard/acoustid/manager.py
```diff
@@ -37,6 +37,8 @@ def __init__(self, fingerprint, duration, orig_recordingid=None, recordingid=Non
 
 class AcoustIDManager(QtCore.QObject):
 
+    BATCH_SUBMIT_COUNT = 240
```
zas: Please explain (in code comment) why the value of 240 was chosen (is it recommended somewhere or just a "good" value?)

picard/acoustid/manager.py
```diff
@@ -112,13 +130,9 @@ def __fingerprint_submission_finished(self, submissions, document, http, error):
...
-                N_('AcoustIDs successfully submitted.'),
-                echo=None,
-                timeout=3000
-            )
+            log.debug('AcoustID: %d fingerprints successfully submitted', len(submissions))
             for file, submission in submissions:
                 submission.orig_recordingid = submission.recordingid
                 file.update()
             self._check_unsubmitted()
+        next()
```
zas: In case of error during the submission of one batch, it will try to submit next batch, right?
What about removing stuff from `submissions` only on success and try again few times?

phw: Fingerprints are only marked as submitted on success. The user can submit again if there are issues. But most likely those issues are permanent.

zas: I understand the user can retry manually, what I mean is we could retry few times in case of temporary errors (ie. network) but since items to be submitted are removed from the list before success/failure is checked, it's not possible.
My idea was to invert the logic: try to submit, and on success, remove from `submissions` list. On failure, either try again after a small delay, or just give up immediatly (removing from `submissions` too = current behavior).

For this, you need to pass the whole `submissions` list, and call next with list without batch successfully submitted or call next with current batch again with a small delay/retry.
That's more a change to the code logic, with the idea of implementing delay/retries if needed later.

phw:
@zas I added retry logic. I did it a bit different than suggested: If a submission fails for one batch it is not immediately retried. Instead all the failings are collected and at the very end if it failed it will start again submitting the failed fingerprints. It will try a total of 3 times to submit (first attempt and 2 retries).

picard/acoustid/manager.py
```diff
@@ -82,16 +87,42 @@ def submit(self):
...
+        log.debug("AcoustID: submitting total of %d fingerprints...", len(submissions))
+        self._batch_submit(submissions)
+
+    def _batch_submit(self, submissions, retries=2, unsuccessful=None):
+        if not submissions:  # All fingerprints submitted
+            if unsuccessful:
+                if retries > 0:  # Retry unsuccessful submissions
+                    log.info("AcoustID: failed submitting %d fingerprints, retrying %d times",
+                        len(unsuccessful), retries)
+                    self._batch_submit(unsuccessful, retries - 1)
```
zas: What about using a small delay between retries? I think we could use something like:

```python
DELAY_MSEC = 1000
QtCore.QTimer.singleShot(DELAY_MSEC, partial(self._batch_submit, unsuccessful, retries - 1))
```
```

Comment (phw) [reply picard/acoustid/manager.py]: There is already a natural delay between submissions. Otherwise I don't think adding a small delay would help much. If this would be a server application running in the background I would say we try again after a few minutes, but this does not really help.

I get the feeling we are trying to solve an issue here we don't really have. The webservice module already does do retries on some kind of errors. Picard is doing a lot of web requests. Every one could fail, we could add retries for all of them. But I don't think it is good client behavior to just rerun requests for no good reason if the server responded with a failure. I'm also not sure what makes AcoustID special that we should try extra hard to get this submitted. If this fails it is either a temporary network issue somewhere in between, but this might also take some time to get resolved, or a more permanent server side issue, which likely won't be solved a second later nor even 10 minutes later.

Thinking of it I'm actually tempted to revert the retry logic here.

Comment (zas) [reply picard/acoustid/manager.py]: Ok, thanks for explanations, if reverting retry logic makes things simpler (and retry isn't really needed), I'm for it.

File: picard/webservice/__init__.py [hunk 6]
```diff
@@ -391,6 +391,7 @@ def _handle_reply(self, reply, request):
             url = reply.request().url().toString(QUrl.RemoveUserInfo)
             log.error("Network request error for %s: %s (QT code %d, HTTP code %d)",
                       url, errstr, error, code)
+
```
zas: Unrelated change

test/test_acoustidmanager.py
```diff
@@ -1,18 +1,50 @@
-from unittest.mock import MagicMock
+from unittest.mock import (
+    MagicMock,
+    Mock,
+)
 
 from test.picardtestcase import PicardTestCase
 
 from picard.acoustid.manager import AcoustIDManager
+from picard import config
```
zas: Import order

PR #1472, opened 2020-02-13
PICARD-1741: Fix $slice return value on index error

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Change the return value from the `$slice` function to an empty list rather than an empty string on an index error.

# Problem

When the `$slice` function encountered an index error, it returned an empty string rather than an empty list.  This could cause argument errors in other parts of the script that expected a multi-value (i.e. list) as input.

* JIRA ticket: PICARD-1741

# Solution

Changed `$slice` function to return an empty list on an index error.

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -1201,7 +1201,7 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
             separator = separator.eval(parser)
         return separator.join(multi_var[start:end])
     except IndexError:
-        return ''
+        return []
 
 
 @script_function()


rdswift:
I was putting together an example script and testing it in the `test_script.py` file.  It fails when `$slice` returns the empty string, but succeeds when I patch the function in `script.py` to return `[]` on an index error.  My test is:

```
    def test_zzz(self):
        context = Metadata()
        context["foo"] = ['A', 'B', 'C', 'D', 'E', 'F', 'G', 'H', 'I', 'J', 'K', 'L', 'M', 'N', 'O', 'P', 'Q', 'R', 'S', 'T', 'U', 'V', 'W', 'X', 'Y', 'Z']
        _test = """
        $setmulti(bar,$slice(%foo%,,6))
        $if($eq($lenmulti(%bar%),1),
            $set(my_out,$getmulti(%bar%,0)),
            $if($eq($lenmulti(%bar%),2),
                $set(my_out,$getmulti(%bar%,0) feat. $getmulti(%bar%,1)),
                $if($lt($lenmulti(%bar%),6),
                    $set(_temp,$join($slice(%bar%,1,-1),\, ))
                    $set(my_out,$getmulti(%bar%,0) feat. %_temp% & $getmulti(%bar%,-1)),

                    $set(_count,$sub($lenmulti(%bar%),4))
                    $set(_temp,$join($slice(%bar%,1,4),\, ))
                    $set(my_out,$getmulti(%bar%,0) feat. %_temp%\, plus %_count% others)
                )
            )
        )
        """
        self.parser.eval(_test, context)
        self.assertScriptResultEquals("%my_out%", "A feat. B, C, D, plus 2 others", context)
```

rdswift:
In looking at the code for the other multi-value functions, I believe that you're right.  There must be a different problem that's causing the error.

phw:
Your example works for me. Actually it does not even run into the except clause.

Looking at the code I'm also not sure under which circumstances this even should throw an index error:

```python
try:
    multi_var = _get_multi_values(parser, multi, separator)
    if not isinstance(separator, str):
        separator = separator.eval(parser)
    return separator.join(multi_var[start:end])
except IndexError:
    return ''
```

The `multi_var[start:end]` is the actual candidate here, but AFAIK and as far as I have tested the slicing notation will not throw an IndexError but return an empty list.

So clearly something else must be off here.

rdswift:
FFS, now I can't get it to fail either.  I'll close this for now, and reopen if necessary.  Sorry to bother you about this.

zas:
BTW, your change to empty list should have triggered a test failure, and it seems it didn't, can you complete the tests?

phw:
> BTW, your change to empty list should have triggered a test failure, and it seems it didn't, can you complete the tests?

@zas I'm not sure this is even possible, I can't see a way on how to enter the except clause here :D Until I miss something we could remove the try-except IndexError here.

rdswift:
> I can't see a way on how to enter the except clause here :D Until I miss something we could remove the try-except IndexError here.

It should trigger an index error if the end index is less than the start index, e.g. `$slice(First:A; Second:B; Third:C,1,0)` (and I do test for this).

rdswift:
But it seems that the indexerror is not triggered.  You're right in that I should remove it.

phw:
> It should trigger an index error if the end index is less than the start index, e.g. `$slice(First:A; Second:B; Third:C,1,0)` (and I do test for this).

Actually no, Pythons slice operator does not throw IndexError at all. In your example which is roughly equivalent to `['a', 'b', 'c'][1:0]` Python would just return an empty list which then gets turned into an empty string by the join. Actually you can throw all kind of crazy integer indexes at the `list[n:m]` notation and the worst thing happening is getting `[]` as result.

rdswift:
I have no idea what I was thinking of when I added the try block.  It should be removed.  I assume I should create a new ticket and pull request to keep the record clean?

phw:
>  I assume I should create a new ticket and pull request to keep the record clean?

I guess it makes sense, yes :)

zas:
> @zas I'm not sure this is even possible, I can't see a way on how to enter the except clause here :D Unless I miss something we could remove the try-except IndexError here.

Yup, just got it ;)

PR #1473, opened 2020-02-13
Remove unnecessary 'try' block from $slice function

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**:  Removes the unused `try` block from the `$slice` scripting function.

# Problem

The `$slice` scripting function contains a `try/except` block where the exception is never triggered.  This should be removed.

# Solution

Remove the unnecessary `try/except` around the code block.

# Action


phw:
Thanks a lot

PR #1474, opened 2020-02-14
PICARD-1743: Add multi-value sort and reverse functions

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Minor / simple change (like a typo)
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds scripting `$sortmulti` and `$reversemulti` functions.

# Problem

The ability to sort a multi-value variable could be handy for example to sort the `artists` list alphabetically before saving to a tag.  In addition, the ability to reverse the order of a multi-value variable could be used to provide a descending sort, and may be useful in other instances.

* JIRA ticket: PICARD-1743

# Solution

Added scripting `$sortmulti` and `$reversemulti` functions.

# Action

If these additions are approved, the Picard Website documentation will need to be updated.

PR #1475, opened 2020-02-14
Introduce MultiValue class to simplify script functions

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -1172,9 +1172,7 @@ def func_map(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
         multi_value[loop_count - 1] = str(loop_code.eval(parser))
     func_unset(parser, '_loop_count')
     func_unset(parser, '_loop_value')
-    if not isinstance(separator, str):
-        separator = separator.eval(parser)
-    return separator.join(multi_value)
+    return multi_value.separator.join(multi_value)
 
 
 @script_function(eval_args=False)
@@ -1224,9 +1222,7 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
     except ValueError:
         end = None
     multi_var = MultiValue(parser, multi, separator)
-    if not isinstance(separator, str):
-        separator = separator.eval(parser)
-    return separator.join(multi_var[start:end])
+    return multi_var.separator.join(multi_var[start:end])
 
 
 @script_function()
--- a/picard/script.py
+++ b/picard/script.py
@@ -347,6 +347,9 @@ class MultiValue(MutableSequence):
     def __repr__(self):
         return '%s(%r, %r, %r)' % (self.__class__.__name__, self.parser, self._multi, self.separator)
 
+    def __str__(self):
+        return self.separator.join(self)
+
 
 def enabled_tagger_scripts_texts():
     """Returns an iterator over the enabled tagger scripts.
@@ -1172,7 +1175,7 @@ def func_map(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
         multi_value[loop_count - 1] = str(loop_code.eval(parser))
     func_unset(parser, '_loop_count')
     func_unset(parser, '_loop_value')
-    return multi_value.separator.join(multi_value)
+    return str(multi_value)
 
 
 @script_function(eval_args=False)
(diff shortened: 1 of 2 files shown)


zas:
> zas, any objections of merging this and #1474 (after it has been adapted to the refactoring here) only after the release?

Yes, that's was my plan.

picard/script.py
```diff
@@ -372,27 +421,6 @@ def _compute_logic(operation, *args):
     return operation(args)
 
 
-def _get_multi_values(parser, multi, separator):
-    if isinstance(separator, ScriptExpression):
-        separator = separator.eval(parser)
-
-    if separator == MULTI_VALUED_JOINER:
```
zas: @phw Is this correct? I mean if separator is evaluated to `'; '` (value of `MULTI_VALUED_JOINER`), do we still want go through this branch?

picard/script.py
```diff
@@ -1195,10 +1221,8 @@ def func_slice(parser, multi, start_index, end_index, separator=MULTI_VALUED_JOI
         end = int(end_index.eval(parser)) if end_index else None
     except ValueError:
         end = None
-    multi_var = _get_multi_values(parser, multi, separator)
-    if not isinstance(separator, str):
```
zas: In `_get_multi_values()`, `separator` is tested against `ScriptExpression` but here we check if it'snt an instance of `str`.
Imho it is much more correct to test if `isinstance(separator, ScriptExpression)`.
Does it deserve a patch before next release?

phw: I agree it is more correct, yes. But I don't think it warrants a patch. In theory this should have no other type other than `str` or `ScriptExpression`, so both statements do the same. Unless we both miss something essential. Let's play it safe and keep the code as is, don't want to introduce any last minute regression :)

picard/script.py
phw: The intention seems to be yes. But to be honest I don't really get the code here. Why do we have special handling if the separator is the default?

zas: I don't get the code either, hence the question. I'll have a deeper look after the release and before we merge this PR.

picard/script.py
```diff
@@ -299,6 +300,53 @@ def eval(self, script, context=None, file=None):
...
+            # Convert ScriptExpression containing only a single variable into variable
+            return self.parser.context.getall(normalize_tagname(multi[0].name))
+
+        # Fall-back to converting to a string and splitting if haystack is an expression
+        # or user has overridden the separator character.
+        multi = multi.eval(self.parser)
+        if self.separator:
+            return multi.split(self.separator)
+        else:
+            return [multi]
```
phw: I found the `return multi.split(separator) if separator else [multi]` one liner in this case even a bit more readable, but no strong preference here.

phw:
@zas About the codacy warning: Shall we disable this check? I think we now that the asserts are not run in pyc code, but that does not mean there is no use of asserts in cases like this.

zas:
> @zas About the codacy warning: Shall we disable this check? I think we now that the asserts are not run in pyc code, but that does not mean there is no use of asserts in cases like this.

Not sure, in this specific case, it will never be triggered but on bad coding practice, and I hope tests will catch this.

picard/script.py
zas: > Only thing that should be added is tests, we should use the chance and introduce some tests for `MultiValue`.

Yes, I'll do that

zas: > I found the `return multi.split(separator) if separator else [multi]` one liner in this case even a bit more readable, but no strong preference here.

Since, I used longer variable names. So I prefer to keep it as is.

picard/script.py
```diff
@@ -299,6 +300,49 @@ def eval(self, script, context=None, file=None):
...
+            # Convert ScriptExpression containing only a single variable into variable
+            self._multi = self.parser.context.getall(normalize_tagname(multi[0].name))
+        else:
+            # Fall-back to converting to a string and splitting if haystack is an expression
+            # or user has overridden the separator character.
+            evaluated_multi = multi.eval(self.parser)
+            if self.separator:
+                self._multi = evaluated_multi.split(self.separator)
+            else:
+                self._multi = [evaluated_multi]
```
phw: This part reads so much clearer then the original code. It is much more clear that there are two cases to consider her.

PR #1479, opened 2020-02-18
PICARD-1748: limit number of countries displayed to 10

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Long list of countries make display of alternative releases far too wide

* JIRA ticket (_optional_): PICARD-1748

# Solution

Limit number of countries displayed to 10

Capture du 2020-02-19 00-46-48

# Action

### DIFF
--- a/picard/releasegroup.py
+++ b/picard/releasegroup.py
@@ -2,6 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 # Copyright (C) 2012 Michael Wiencek
+# Copyright (C) 2020 Laurent Monin
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -30,7 +31,10 @@ from picard.mbjson import (
     media_formats_from_node,
 )
 from picard.metadata import Metadata
-from picard.util import uniqify
+from picard.util import (
+    limited_join,
+    uniqify,
+)
 
 
 class ReleaseGroup(DataObject):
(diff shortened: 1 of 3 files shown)


picard/releasegroup.py
```diff
@@ -86,8 +90,8 @@ def _parse_versions(self, document):
             release = {
                 "id":      node['id'],
                 "year":    node['date'][:4] if "date" in node else "????",
-                "country": "+".join(countries) if countries
-                           else node.get('country', '') or "??",
+                "country": limited_join(countries, 10, '+', '…') if countries
```
phw: Is 10 fine, or even lower this? Also I like the `limited_join` function and how it drops in the middle, but is it actually useful for this case? Does it help to see the last entries or would rather only showing the first 10 + "..." be better?

picard/releasegroup.py
```diff
@@ -86,8 +90,8 @@ def _parse_versions(self, document):
...
+                else node.get('country', '') or "??",
                 "format":  media_formats_from_node(node['media']),
                 "label":  ", ".join([' '.join(x.split(' ')[:2]) for x in set(labels)]),
                 "catnum": ", ".join(set(catnums)),
```
phw: Should we handle limiting label and catnum as well?

picard/releasegroup.py
zas: Actually, I was thinking about adding another parameter, to place the "..." at the end instead.
But since we have no idea what will be or not in the list, it doesn't change much

picard/releasegroup.py
zas: Yes, very likely, but that's not this ticket.

PR #1482, opened 2020-02-19
PICARD-1750: Set AcoustID fingerprint to AcoustIDManager

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

If a file already contains an acoustid_fingerprint tag and "Ignore existing AcoustID fingerprints" is turned off and if scan is used on this file, then:

    fpcalc does not get called (expected)
    the fingerprint does not get added to AcoustId manager (fail)

The result is that even after matching the file to a proper recording it is not possible to submit the AcoustID.

Also ideall Picard would already indicate the presence of the fingerprint after loading the file.

* JIRA ticket (_optional_): PICARD-1750

# Solution
- Refactor the code so that `File` actually manages the acoustid_* attributes. This makes these always available and centralizes the calls to update AcoustIDManager.
- Already set `File.acoustid_*` attributes on file load, if the file contains the `acoustid_fingerprint` tag **and** `ignore_existing_acoustid_fingerprints` is not set. This ensures the fingerprint is available from the start and can be used for submissions.
- When using the fingerprint from metadata on `AcoustIDClient.analyze` also  set `File.acoustid_*` attributes. This covers the case where the user had the `ignore_existing_acoustid_fingerprints` on load but later analyzes with this option disabled.

# Action

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -462,6 +462,7 @@ def release_to_metadata(node, m, album=None):
                 m['~releaselanguage'] = value['language']
             if 'script' in value:
                 m['script'] = value['script']
+    m['~releasecountries'] = countries_from_node(node)
     add_genres_from_node(node, album)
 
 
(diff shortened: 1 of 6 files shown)


phw:
Sorry, wrong branch :/

PR #1484, opened 2020-02-20
PICARD-1754: Removed QKeySequence.Delete shortcut for remove action

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Pressing the delete shortcut currently always removes files or albums, even if the metadatabox is focused.

* JIRA ticket (_optional_): PICARD-1754

# Solution
This reverts 84d8341. QKeySequence.Delete is still used for item removal on the itemview and metadatabox panes (see MainWindow.keyPressEvent at https://github.com/metabrainz/picard/blob/master/picard/ui/mainwindow.py#L192). But if this action is registered globally item removal interfers with removing elements from MetadataBox.

# Action


phw:
> That was a bad move, good catch.

Not really mine :D Just bisected it

PR #1485, opened 2020-02-20
PICARD-1755: Specify available languages for Windows Store.

phw:
# Summary
Update the languages displayed on the Windows Store.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Windows Store currently only lists English as available language

* JIRA ticket (_optional_): PICARD-1755

# Solution
For now limit this to the languages for which there is also a localized store page (which is based on completion of translating the appstream data).

This needs testing, I don't know whether this works as expected.

# Action

### DIFF
--- a/appxmanifest.xml.in
+++ b/appxmanifest.xml.in
@@ -15,7 +15,16 @@
     <Logo>Square70x70Logo.png</Logo>
   </Properties>
   <Resources>
-    <Resource Language="en-us"/>
+    <Resource Language="en"/>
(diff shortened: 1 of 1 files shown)


phw:
This seems to work actually. I can't fully test this until the next time we actually update the package on the store, but uploading it shows the added languages afterwards in the store management UI. So I'm positive this will show correctly on the store page itself.

PR #1487, opened 2020-02-20
PICARD-1756: Fix plugin update version comparison

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Plugin update check is assuming version 1.10 < 1.9 due to comparing the version segments as strings.

* JIRA ticket (_optional_): PICARD-1756

# Solution
Use `Version` class to implement version comparison. This also simplifies version handling at some points.

# Action

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -38,6 +38,7 @@ PICARD_BUILD_VERSION_STR = ""
 
 
 def version_to_string(version, short=False):
+    """Deprecated: Use picard.version.Version.to_string instead"""
     if len(version) != 5:
         raise VersionError("Length != 5")
     if not isinstance(version, Version):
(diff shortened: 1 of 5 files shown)


picard/__init__.py
```diff
@@ -38,6 +38,7 @@
 
 
 def version_to_string(version, short=False):
+    """Deprecated: Use picard.version.Version.to_string instead"""
```
zas: Should we use https://docs.python.org/3/library/warnings.html#warnings.warn for that?

phw: Yes, I think we should do that. Probably with some decorator. We have a few other places to use it

PR #1492, opened 2020-02-21
PICARD-1760: Prevent duplicate in top tags editor

phw:
# Summary
Prevent entering duplicate values in top tags options.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
It was possible to enter duplicate values in the top tags editor. Those duplicated tracks would also show up duplicated in the metadatabox. This behavior is at least confusing.

* JIRA ticket (_optional_): PICARD-1760

# Solution
Implement a `UniqueEditableListView` and use it for the `TagListEditor`. If a value is inserted or an existing row is edited all other occurrences of the same value will be removed.

# Action

### DIFF
--- a/picard/ui/widgets/editablelistview.py
+++ b/picard/ui/widgets/editablelistview.py
@@ -137,26 +137,41 @@ class EditableListView(QtWidgets.QListView):
 class UniqueEditableListView(EditableListView):
     def __init__(self, parent=None):
         super().__init__(parent)
+        self._is_drag_drop = False
 
     def setModel(self, model):
+        current_model = self.model()
+        if current_model:
+            current_model.dataChanged.disconnect(self.on_data_changed)
         super().setModel(model)
         model.dataChanged.connect(self.on_data_changed)
 
+    def dropEvent(self, event):
+        self._is_drag_drop = True
+        super().dropEvent(event)
+        self._is_drag_drop = False
+
     def on_data_changed(self, top_left, bottom_right, roles):
-        role = roles[0]
+        # Do not drop duplicates during drag and drop as there is always
+        # a duplicate in the list for a short time in this case.
+        if self._is_drag_drop:
+            return
         model = self.model()
-        if role == QtCore.Qt.EditRole:
-            value = model.data(top_left, role)
+        if QtCore.Qt.EditRole in roles:
+            value = model.data(top_left, QtCore.Qt.EditRole)
             if not value:
                 return
             # Remove duplicate entries from the model
-            unique_list = []
-            for row, item in enumerate(model.items):
-                if item != value or row == top_left.row():
-                    unique_list.append(item)
-            if len(unique_list) != model.rowCount():
-                self.update(unique_list)
-                self.select_key(value)
+            changed_row = top_left.row()
+            row = 0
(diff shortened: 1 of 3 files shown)


picard/ui/widgets/editablelistview.py
```diff
@@ -134,6 +134,47 @@ def _move_rows_relative(self, rows, direction):
...
+        self._is_drag_drop = True
+        super().dropEvent(event)
+        self._is_drag_drop = False
+
+    def on_data_changed(self, top_left, bottom_right, roles):
+        # Do not drop duplicates during drag and drop as there is always
+        # a duplicate in the list for a short time in this case.
+        if self._is_drag_drop:
+            return
+        role = roles[0]
```
zas: Not sure if it is an issue or not, but https://doc.qt.io/qt-5/qabstractitemmodel.html#dataChanged says:
> An empty vector in the roles argument means that all roles should be considered modified. The order of elements in the roles argument does not have any relevance.

So it seems more correct to do:

```python
if QtCore.Qt.EditRole in roles:
   value = model.data(top_left, QtCore.Qt.EditRole)
```

picard/ui/widgets/editablelistview.py
```diff
@@ -134,6 +134,47 @@ def _move_rows_relative(self, rows, direction):
...
+        if role == QtCore.Qt.EditRole:
+            value = model.data(top_left, role)
+            if not value:
+                return
+            # Remove duplicate entries from the model
+            changed_row = top_left.row()
+            row = 0
+            for item in (model.items):
+                if item == value and row != changed_row:
+                    self.model().removeRow(row)
```
zas: Is there a reason to not use `model` in place of `self.model()` ?

picard/ui/widgets/editablelistview.py
```diff
@@ -134,6 +134,47 @@ def _move_rows_relative(self, rows, direction):
...
+        role = roles[0]
+        model = self.model()
+        if role == QtCore.Qt.EditRole:
+            value = model.data(top_left, role)
+            if not value:
+                return
+            # Remove duplicate entries from the model
+            changed_row = top_left.row()
+            row = 0
+            for item in (model.items):
```
zas: Are parenthesis needed?

picard/ui/widgets/editablelistview.py
phw: It's not really an issue because we control the model. But I change it anyway.

picard/ui/widgets/editablelistview.py
phw: Fixed

picard/ui/widgets/editablelistview.py
phw: They are a left over, fixed

phw:
> This is a bit hacky (but code is clean), it seems we have no much choice here.

I would not really call it hacky. It is just that the underlying interfaces are very generic (supporting all kind of data displays two dimensional tables and data organized in tree hierarchy). So for our simple case of just managing a list of strings a seemingly simple task like removing duplicates needs more verbose code then probably would be expected.

PR #1493, opened 2020-02-22
Coverage & cleanup

zas:
Improve test coverage and general cleanup with the help of coverage tools.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -391,13 +391,10 @@ def album_artist_from_path(filename, album, artist):
     """
     if not album:
         dirs = os.path.dirname(filename).replace('\\', '/').lstrip('/').split('/')
-        if len(dirs) == 0:
-            return album, artist
         # Strip disc subdirectory from list
-        if len(dirs) > 0:
-            if re.search(r'(^|\s)(CD|DVD|Disc)\s*\d+(\s|$)', dirs[-1], re.I):
-                del dirs[-1]
-        if len(dirs) > 0:
+        if re.search(r'(^|\s)(CD|DVD|Disc)\s*\d+(\s|$)', dirs[-1], re.I):
+            del dirs[-1]
+        if dirs:
             # For clustering assume %artist%/%album%/file or %artist% - %album%/file
             album = dirs[-1]
             if ' - ' in album:
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -392,7 +392,7 @@ def album_artist_from_path(filename, album, artist):
     if not album:
         dirs = os.path.dirname(filename).replace('\\', '/').lstrip('/').split('/')
         # Strip disc subdirectory from list
-        if re.search(r'(^|\s)(CD|DVD|Disc)\s*\d+(\s|$)', dirs[-1], re.I):
+        if re.search(r'\b(?:CD|DVD|Disc)\s*\d+\b', dirs[-1], re.I):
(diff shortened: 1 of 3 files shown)


picard/util/__init__.py
```diff
@@ -391,13 +391,10 @@ def album_artist_from_path(filename, album, artist):
...
     if not album:
         dirs = os.path.dirname(filename).replace('\\', '/').lstrip('/').split('/')
-        if len(dirs) == 0:
-            return album, artist
         # Strip disc subdirectory from list
-        if len(dirs) > 0:
-            if re.search(r'(^|\s)(CD|DVD|Disc)\s*\d+(\s|$)', dirs[-1], re.I):
-                del dirs[-1]
-        if len(dirs) > 0:
+        if re.search(r'\b(?:CD|DVD|Disc)\s*\d+\b', dirs[-1], re.I):
```
phw: Why was the check for `len(dirs) == 0` from above removed? `dirs[-1]` will fail in this case.

picard/util/imageinfo.py
```diff
@@ -106,7 +106,7 @@ def identify(data):
             pass
 
     # PDF
-    elif data[:4] == '%PDF':
+    elif data[:4] == b'%PDF':
```
phw: This is actually a bug fix, right? If I see this correctly recognizing PDF was broken. If so I think we should add a ticket for the changelog.

picard/util/__init__.py
```diff
@@ -512,10 +509,6 @@ def load_json(data):
     return json.loads(__convert_to_string(data))
 
 
-def parse_json(reply):
-    return load_json(reply.readAll())
```
phw: As discussed on IRC this is a potentially plugin breaking API, but:

- No plugin is currently accessing this (and probably shouldn't but use `load_json` instead)
- It is very specific to the webservice (accepting a response object)
- Plugin can be easily adapted if needed

picard/util/__init__.py
zas: As said in the commit message:
> len(dirs) == 0 cannot happen, because `whatever.split('/')` is always a list with at least 1 element.

So `dirs` contains at least one element, the empty string'

picard/util/imageinfo.py
zas: Yes, but I wonder when it actually happens, I didn't yet look in details.

picard/util/__init__.py
zas: I'll remove those commits.

zas: Done, and rebased

picard/util/imageinfo.py
zas: Found it, it happens at least in one case:
when one drops a pdf on cover art image box
Without the patch, it fails with:
D: 21:07:10,873 ui.coverartbox.dropEvent:99: Dropped file:///home/zas/Téléchargements/t.pdf url (with 0 bytes of data)
W: 21:07:10,873 ui.coverartbox.load_remote_image:458: Can't load image: Unrecognized image data

zas: Fix was moved to https://github.com/metabrainz/picard/pull/1494

picard/util/__init__.py
zas: This is actually tested in https://github.com/metabrainz/picard/pull/1493/files#diff-85c7da7d97a696e8017da5742a876475R211

PR #1495, opened 2020-02-23
PICARD-1771: Scoring change to ignore undesired release types

RaysDev:
# Summary
Scoring change to ignore undesired release types, and add some weight for tracks with non-null release date.

* This is a…
    * [] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

1. Some results have no release date, which are less desirable to match to.

2. While user can leverage the sliders in the preferences screen, to lower the priority of certain "release types," they have no way to completely exclude it. since final result is based on several factors, these undesirable types still get chosen by picard. I found this particularly troublesome with compilations. I want to tag my mp3 with the original album, not a recent compilation.

PICARD-1771

# Solution

1. Higher priority should be given to results that have a non-null release date.

2. If the slider in preferences screen is set to zero (far left) for a given Release Type, then Picard should never have its final pick be of that release type. If user just wants to de-prioritize it, they can move slider slightly right of the far left.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -81,7 +81,7 @@ def weights_from_release_type_scores(parts, release, release_type_scores,
             type_score = type_scores.get(release_type, other_score)
             if type_score == 0:
                 skip_it = True
-            score += type_scores.get(release_type, other_score)
+            score += type_score
         score /= len(types_found)
 
     if skip_it:
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -67,6 +67,9 @@ def weights_from_release_type_scores(parts, release, release_type_scores,
     # If no types are found, it is set to the score of the 'Other' type or 0.5 if 'Other' doesnt exist
     # It appends (score, weight_release_type) to passed parts list
 
+    # if our preference is zero for the release_type, force to never return this recording. By using large zero weight. This means it only gets picked if there are no others at all
+    skip_it = False
+
     type_scores = dict(release_type_scores)
     score = 0.0
     if 'release-group' in release and 'primary-type' in release['release-group']:
@@ -75,9 +78,16 @@ def weights_from_release_type_scores(parts, release, release_type_scores,
             types_found += release['release-group']['secondary-types']
         other_score = type_scores.get('Other', 0.5)
         for release_type in types_found:
+            type_score = type_scores.get(release_type, other_score)
+            if type_score == 0:
+                skip_it = True
             score += type_scores.get(release_type, other_score)
         score /= len(types_found)
-    parts.append((score, weight_release_type))
+
+    if skip_it:
+        parts.append((0, 9999))
+    else:
+        parts.append((score, weight_release_type))
 
 
 def weights_from_preferred_countries(parts, release,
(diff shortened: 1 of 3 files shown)


picard/metadata.py
```diff
@@ -75,9 +78,16 @@ def weights_from_release_type_scores(parts, release, release_type_scores,
             types_found += release['release-group']['secondary-types']
         other_score = type_scores.get('Other', 0.5)
         for release_type in types_found:
+            type_score = type_scores.get(release_type, other_score)
+            if type_score == 0:
+                skip_it = True
             score += type_scores.get(release_type, other_score)
```
phw: Better use `score += type_score`.

Sophist-UK:
> Thanks. For ignoring the release types of score 0 I think this is great, only a small code improvement.
> 
> I am not 100% sure on the idea of preferring the data just because it exists. Wouldn't it be better to also check this against a possible date in the existing tags?
> 
> E.g. if I have a file that already has a date set to "2017" should it really prefer the release from MB with date set to 1990? Maybe that one release with no date set yet actually fits the 2017 file better?
> 
> And if we compare dates we have to be aware that this is often not really reliable information on badly tagged files. Also this needs to consider partial dates.
> 
> @zas what's your take on this?



> Thanks. For ignoring the release types of score 0 I think this is great, only a small code improvement.
> 
> I am not 100% sure on the idea of preferring the data just because it exists. Wouldn't it be better to also check this against a possible date in the existing tags?
> 
> E.g. if I have a file that already has a date set to "2017" should it really prefer the release from MB with date set to 1990? Maybe that one release with no date set yet actually fits the 2017 file better?
> 
> And if we compare dates we have to be aware that this is often not really reliable information on badly tagged files. Also this needs to consider partial dates.

Either the existing YEAR is correct or it isn't, and being "one year out" (for those who are in the UK and listen to Steve Wright on BBC Radio 2) is really the same as being wildly out.

So, my personal view (which could be a bad one) is that a release should be preferred only if the YEAR matches exactly.

picard/metadata.py
RaysDev: Done

RaysDev:
I'm reviewing your suggestion to match on date.  My goal was just that when multiple results are available, and we are trying to get to best one, it should not be one that does not have a date.  Your idea is a step better, and your right the date comparisons could get complex.  So for now, lets stick with the "exists" logic, and I'll see if I can enhance it to go a step further and compare the file value.  Should be doable, atleast at year level.

RaysDev:
Thats fine on date logic to separate PR.  I assume you will just skip those lines when you merge.  If you want me to remove it and commit that, let me know.

phw:
@RaysDev I merged the release type change in https://github.com/metabrainz/picard/commit/3c3a4c2c8e39aef5c5b7030626a71ed317d6d1ea . As discussed please open another PR for the date change against the devel branch.

Another note for your other PRs: We recently updated the copyright notices in the file headers for all files and now plan to keep it up-to-date. In your other pull requests, can you add a copyright line to the files you changed?

PR #1496, opened 2020-02-23
Parse additional fields from the AcoustId response

RaysDev:
#Summary
Parse additional fields from the AcoustId response

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
AcoustId returns many data elements, but not all are parsed out and made available within Picard.

# Solution
Changed acoustid/json_helpers.py to parse out additional data elements.

### DIFF
--- a/picard/acoustid/json_helpers.py
+++ b/picard/acoustid/json_helpers.py
@@ -4,6 +4,7 @@
 #
 # Copyright (C) 2017 Sambhav Kothari
 # Copyright (C) 2018-2020 Philipp Wolfer
+# Copyright (C) 2020 Ray Bouchard
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -18,6 +19,12 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""
+The idea here is to bring the data returned by the AcoustID service into the
+same format as the JSON result from the MB web service. Below methods help us
+to do that conversion process.
+"""
 
 
 def _make_releases_node(recording):
@@ -36,16 +43,51 @@ def _make_releases_node(recording):
                 release_mb['title'] = release['title']
             else:
                 release_mb['title'] = release_group['title']
+
             if 'country' in release:
                 release_mb['country'] = release['country']
+
+            if 'date' in release:
+                release_mb['date'] = release['date']
+
+            if 'medium_count' in release:
+                release_mb['medium-count'] = release['medium_count']
+
+            if 'track_count' in release:
+                release_mb['track-count'] = release['track_count']
+
             release_mb['media'] = []
             for medium in release['mediums']:
-                medium_mb = {}
+                media_mb = {}
                 if 'format' in medium:
-                    medium_mb['format'] = medium['format']
-                medium_mb['track-count'] = medium['track_count']
-                release_mb['media'].append(medium_mb)
-            release_list.append(release_mb)
+                    media_mb['format'] = medium['format']
+
+                if 'track_count' in medium:
+                    media_mb['track-count'] = medium['track_count']
+
+                if 'position' in medium:
+                    media_mb['position'] = medium['position']
+
+                if 'tracks' in medium:
+                    media_mb['track'] = medium['tracks']
+                    for track_mb in media_mb['track']:
+                        track_mb['number'] = track_mb['position']
+
+                release_mb['media'].append(media_mb)
+
+            # AcoustId service is returning country/date as attrib of the release, but really, according to MusicBrainz database schema definition,
+            # https://musicbrainz.org/doc/MusicBrainz_Database/Schema
+            # Its a one-to-many sub attribute in release events.
+            # They do return the releaseevents, but seem to be copying the first one on to the release.
+            # So if we have releaseevents, then use them to create multiple records, and ignore what is coming on the release
+            if 'releaseevents' in release:
+                for releaseevent in release['releaseevents']:
+                    release_mb['country'] = releaseevent.get('country', '')
+                    release_mb['date'] = releaseevent.get('date', '')
+                    release_list.append(release_mb)
+            else:
+                release_list.append(release_mb)
+
     return release_list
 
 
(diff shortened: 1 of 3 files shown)


picard/acoustid/json_helpers.py
```diff
@@ -34,16 +34,48 @@ def _make_releases_node(recording):
...
                 release_mb['title'] = release_group['title']
+
             if 'country' in release:
                 release_mb['country'] = release['country']
+
+            if 'date' in release:
+                release_mb['date'] = release['date']
+
+            if 'medium_count' in release:
+                release_mb['medium_count'] = release['medium_count']
```
phw: The idea here is to bring the data returned by the AcoustID service into the same format as the JSON result from the MB web service. But there is actually no `medium_count` attribute on the MB web service. If this is needed it can be taken from the length of the `media` list.

I would vote for not including this attribute here. But if we do it should be renamed `medium-count`.

picard/acoustid/json_helpers.py
```diff
@@ -34,16 +34,48 @@ def _make_releases_node(recording):
...
                 release_mb['media'].append(medium_mb)
-            release_list.append(release_mb)
+
+            # AcoustId service is returning country/date as attrib of the release, but really, according to MusicBrainz schema definition,
+            # its a one-many sub attribute in release events.
+            # They do return the releaseevents, but seem to be copying the first one on to the release.
+            # So if we have releaseevents, then use them to create multiple records, and ignore what is coming on the release
+            if 'releaseevents' in release:
+                for releaseevent in release['releaseevents']:
+                    release_mb['country'] = releaseevent.get('country', '')
```
phw: There is actually not country attribute in the MB response. Instead there is an area attribute looking something like this:

```
"area": {
    "iso-3166-1-codes": [
        "DE"
    ],
    "id": "85752fda-13c4-31a3-bee5-0e5cb1f51dad",
    "sort-name": "Germany",
    "disambiguation": "",
    "name": "Germany"
}
```

Overall I actually don't think we have much value of extracting this attribute here.

picard/acoustid/json_helpers.py
```diff
@@ -34,16 +34,48 @@ def _make_releases_node(recording):
...
-                medium_mb['track-count'] = medium['track_count']
+
+                if 'track_count' in medium:
+                    medium_mb['track-count'] = medium['track_count']
+
+                if 'position' in medium:
+                    medium_mb['position'] = medium['position']
+
+                if 'tracks' in medium:
+                    medium_mb['tracks'] = medium['tracks'][0]
```
phw: What exactl is `medium['tracks'][0]` here? Isn't this just the first track? Would it have the correct format?

picard/acoustid/json_helpers.py
RaysDev: Done.  renamed to media-count, and used the length of the list as suggested.

picard/acoustid/json_helpers.py
RaysDev: I do see country in the mb response. It looks like this

"date": "1996-06-18",
**"country": "US",**
"release-events": [
	{
		"date": "1996-06-18",
		"area": {
			"id": "489ce91b-6658-3307-9877-795b68554c98",
			"name": "United States",
			"sort-name": "United States",
			"iso-3166-1-codes": [
				"US"
			]
		}
	}
],

picard/acoustid/json_helpers.py
RaysDev: element zero because the json structure supports a list, but we did a search on single track, so would never have more than one Track element, so no looping needed.  I'll add comment to code.

picard/acoustid/json_helpers.py
phw: You are right. Must have been asleep, actually I just updated code surrounding this in https://github.com/metabrainz/picard/commit/800f61e6b6e239ac4cada0740ae68e04296a4154

picard/acoustid/json_helpers.py
phw: But `medium_mb['tracks']` should have the same format in both cases. If it is the list in one case and a single object in the other it makes it difficult to use.

There is also the risk that code uses `len(medium_mb['tracks'])` to get the track count, but I don't think we have this.

RaysDev: ok, I'm guessing from one of your other recent comment, that by "both cases" you mean musicbrainz vs acoustId, so now I understand that "json_helpers" is a converter, and result should look like MB.  I will make them have same structure for tracks, and also add a comment to code about this.

phw: Ah, yes. Sorry if this is not clear. Basically all the code run against this result later (mostly the matching in `Metadata`) is using the structure as returned by the MB web service. So in order to use the same logic for the AcoustID results the `parse_recording` function is supposed to rewrite the data from AcoustID into the same structure.

You can consider the result of this function the internal representation of the data as Picard uses it. But to simplify things this matches the structure as it is returned by MB web service. AcoustID data is similar enough so it can be rewritten.

RaysDev:
I believe all requested changes on this PR are completed, including requested change for medium_mb['tracks'].  Let me know otherwise.

picard/metadata.py
```diff
@@ -300,6 +300,8 @@ def compare_to_track(self, track, weights):
 
         result = SimMatchTrack(similarity=-1, releasegroup=None, release=None, track=None)
         for release in releases:
+            if 'medium-count' not in release and 'media' in release:  # generate this field for MusicBrainz, it is already part of acoustid
+                release['medium-count'] = len(release['media'])
```
phw: Do I miss something here? `release['medium-count']` is not used in the code here, why is it set here? The problem I see with this is that it changes the data passed to `compare_to_track`, which is an unwanted side effect. If we want to make sure `release['medium-count']` is set it should be done elsewhere (e.g. after loading the data), but not here.

RaysDev: ok to skip these 2 lines, we can re-visit in future if needed.

RaysDev: And in thinking about it, yes I agree, the medium-count should not be added as part of compare_to_track().  as thats not its purpose.  If we want to add this later, I'm open to suggestions on where to put it.

phw: Ok, I merged it without those lines. I need to look at where we best can fit this if needed.

PR #1476, opened 2020-02-15
Match insight

RaysDev:
* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Minor / simple change (like a typo)
    * [ ] Other

**#Summary**
Minor adjustment to scoring logic, and when matching file to AcoustId/musicbrainz results, provides insight to user on the scoring/final selection process.

**# Problem**
AcoustID/Musicbrainz can return multiple possibilities in their response, and we need to match the input file to just one of them.
1. some results have no release date, which are less desirable to match to.

2. while user can leverage the sliders in the preferences screen, to lower the priority of certain "release types," they have no way to completely exclude it.  since final result is based on several factors, these undesirable types still get chosen by picard.  I found this particularly troublesome with compilations.  I want to tag my mp3 with the original album, not a recent compilation.

3. I've observed that sometimes acoustid will return over hundred results for Picard to pick from.  While picard has a great algorithm for picking one.  I found it often did not chose the one I preferred (ie the original album for the specific recording).  I was left wondering why, and what could I tweak in my mp3 file's tags to get Picard to make the right match.  There was no information available.

**# Solution**
1. higher priority should be given to results that have a non-null release date.

2. if the slider in preferences screen is set to zero (far left) for a given Release Type, then Picard should never have its final pick be of that release type.  If user just wants to de-prioritize it, they can move slider slightly right of the far left.

3.  First I added detailed logging of the matching (ie scoring) process so I could see why it was picking a particular result.  But few users would want to dig through logs, and it was not time efficient when you have a lot of tracks.  Further, having the information is nice, but ultimately if we did not like Picard's selection, then we want to be able to select an alternative.  So I moved the info from the logs into a new dialog.  Available by right clicking on the track in Right side pane, and choosing "Match Target".  While there is already a "Other Versions" options, its very limited.  This new dialog, not only shows the possible results, but also has filtering/sorting capability from the column header.  So user can narrow down the results, and then click on one to apply it to the track.

PR #1497, opened 2020-02-23
New Match Dialog for picking alternative result

RaysDev:
# Summary
New Match Dialog to show all matches from acoustid and musicbrainz, and allow user to choose alternative track.  Particuarly useful when a file/recording is included in multiple "release groups"

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
AcoustID and MusicBrainz lookups are features to automate the tagging of files, so users don't have to type all the tag info in manually.  However, the lookups provide multiple potential matches.  While Picard scores them and picks one, the user has no way to pick a different one.

# Solution
Provide a dialog box that shows all the possible matches.  Provide tools (sorting, grid filtering, export to tsv) to allow user to select alternative result.


picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+""" NOTE: THIS FILE IS HIGHLY BASED ON same __init__.py in searchdialog """
```
zas: This comment seems to say: `duplicated code`....
And a quick check shows it is the case, which is obviously bad. I guess it is a reminder for something you want to improve/

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,395 @@
...
+                        if item.text() not in self.keywords[j]:
+                            columnsShow[i] = False
+            for key, value in columnsShow.items():
+                self.setRowHidden(key, not value)
+
+        except Exception:
+            log.error(traceback.format_exc())
+
+
+class SortableTableWidgetItem(QtWidgets.QTableWidgetItem):
```
zas: Duplicate of https://github.com/metabrainz/picard/blob/170b8e0aa70ee4a4c49e89ab42e26404e62e90dc/picard/ui/searchdialog/__init__.py#L172-L179

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+        self.columns = [
+            ('nbr',                  "Log\nSeq"),
+            ('match_dtl_entry_type', "Entry Type"),
+            ('is_new',               "Better?"),
+            ('similarity',           "Similarity"),
+            ('acoust_id',            "AcoustId ID"),
+            ('rec_id',               "Recording ID"),
+            ('rec_title',            "Recording Title"),
+            ('rec_len',              "Recording\nLength"),
+            ('rec_src_cnt',          "Recording\nSource\nCount"),
```
zas: Those are column key (or short non-translatable unique name) and English names, right?
What about using an OrderedDict?
Also those `\n` inside labels look like an attempt to format strings, but it doesn't work when it comes to i18n (those strings should be marked for translation, `_(Recording Length)`.
Wrapping should be done by Qt somehow (Qt.TextWordWrap?)

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+
+            fingerprint_lkp_url = ("https://" + ACOUSTID_HOST + "/v2/lookup"
+                                   + "?meta=recordings%20releasegroups%20releases%20tracks%20compress%20sources"
+                                   + "&fingerprint=" + fp
+                                   + "&duration=" + str(dur)
+                                   + "&client=" + ACOUSTID_KEY
+                                   + "&clientversion=" + PICARD_VERSION_STR
+                                   + "&format=json") if fp is not None else ""
+
+            json = match_details_log.get_file_result_json(file_.filename) if (file_ is not None) else ""
```
zas: `if file_ is not None`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            if json is not None:
+                self.set_json_edit(json)
+            if file_ is not None:
```
zas: Test already done few lines above, may be code can be reordered to do it once only.

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+                self.set_table_item(row, 'med_trk_cnt',          entry.med_trk_cnt)
+                self.set_table_item(row, 'rel_tot_trk_cnt',      entry.rel_tot_trk_cnt)
+                self.set_table_item(row, 'parts',                str(entry.parts))
+                self.set_table_item(row, 'release_parts',        str(entry.release_parts))
+                self.set_table_item(row, 'recording_base_score', entry.recording_base_score)
+
+            self.show_table(sort_column='nbr', sort_order=QtCore.Qt.AscendingOrder)
+
+        except TypeError as e:
+            log.warning("MATCH:No log entries for " + file_.filenamei + " due to " + e)
```
zas: `log.warning("MATCH:No log entries for %s due to %s" , file_.filenamei, e)`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            #filename = tempfile.gettempdir() + os.path.sep + "picard_match_details.tsv"
+
+            with open(filename, "w") as text_file:
+                print(match_details_log.match_dtl_entry.get_hdr_tsv(), file=text_file)
+
+                for row, entry in enumerate(self.entries):
+                    output_str = str(row) + "\t" + entry.fmt_as_tsv() + os.linesep
+                    text_file.write(output_str)
+            log.debug("Wrote file with match details to file:" + text_file.name)
+        except Exception:  # Ignore any errors as this output file is non-essential
```
zas: Which exceptions are supposed to be raised?

picard/webservice/__init__.py
```diff
@@ -425,7 +427,8 @@ def _handle_reply(self, reply, request):
                     self._handle_redirect(reply, request, redirect)
                 elif request.response_parser:
                     try:
-                        document = request.response_parser(reply)
+                        WebService.last_raw_reply = reply.readAll()
```
zas: Really unsure about this.

picard/webservice/__init__.py
```diff
@@ -572,6 +576,22 @@ def get_response_parser(cls, response_type):
...
+        """ get last raw reply - the http body
+
+        This class passes the http body to a parsing function, and returns the result of that as the document.
+        Note that reply.readAll() is only able to be called once, as it consumes the actual http body.
+        Therefore the unparsed reply is not available to anything else.
+        Now we pre-saved the the http body and make it available via this getter.
+        Ideally this would not be static.  However, Webservice class is abstracted with layers of classes that make it unavailable from any class that would want to get the raw reply.
+        So making it static allows us to expose this value without re-architecting the whole program.
+        This should be re-factored in the future.
+        WARNING: This is not thread-safe, so if multiple threads are making Webservice calls concurrently, they could get the wrong results here.
```
zas: ^^

phw:
First off: I like the general idea of allowing the user to select a different result if multiple matches are found. But the current implementation as it is now is IMHO suitable as a debugging tool, but not really for a tool to select the proper recording for the user.

I think in general using Picard should neither require understanding of Picard's source code nor the web service responses of MB or AcoustID. If I use this dialog on one file I get this:

grafik

So I get:
- As first element a link to the web service response
- A Python binary string for the result of former web service
- A table showing numbers and IDs that don't help me in any way selecting the proper recording
- Further inspection shows the currently selected entry is shown three times with different "entry type"

I think this should be changed so that it shows all the results in decreasing order of relevance, with the most important information shown that allows one to make a decisions. Entries should only be listed once with the currently selected shown on top. Hence I propose the following changes:

1. Remove the display of the raw data at the top. If this is really wanted the link to the lookup result is sufficient, but it should be moved below the table
2. In the table completely remove the columns loq seq, better?, track ID, entry type, scoring parts, scoring release parts and recording base score. The latter three could be output as debug messages during the lookup (if they not already are, didn't check)
3. Combine corresponding ID and title into a single field with the title linked to the ID
4. Link the AcoustID
5. Maybe combine track /medium count and total into a single field with [count] / [total]
6. Highlight the currently selected entry
7. Round by descending similarity
8. Round the similarity to 2 or 3 digits

I think my example from the screenshot should look something like this:

| Track name | Length | Artist | Release | Track | Medium | Date | Country | Type | Format | AcoustID | AcoustID source count | Similarity |
|-----------------|-----------|---------|------------|---------|------------|--------|-------------|--------|----------|-------------|--|-|
| **[Essence](https://musicbrainz.org/recording/dda7708a-5559-4e04-b11a-268268941dc7)** | **04:07** | **[Absurd Minds](https://musicbrainz.org/artist/979ce33b-4d8e-4e44-8c7e-f9d3054f4386)** | **[The Focus](https://musicbrainz.org/release/86baa9d0-b610-4721-80ee-bc70f5c0680a)** | **1/13** | **1/1** | **2003-10-13** | **DE** | **Album** | **CD** | **[b5123596-4a8d-43f3-98b7-3cdf8ba2ec39](https://acoustid.org/track/b5123596-4a8d-43f3-98b7-3cdf8ba2ec39)** | 38 | **0.907** |
| [Essence](https://musicbrainz.org/recording/dda7708a-5559-4e04-b11a-268268941dc7) | 04:07 | [Absurd Minds](https://musicbrainz.org/artist/979ce33b-4d8e-4e44-8c7e-f9d3054f4386) | [The Focus](https://musicbrainz.org/release/9e777cfd-f342-3615-926f-7702928e38b5) | 1/13 | 1/2 | 2003-10-13 | DE | Album | CD | **[b5123596-4a8d-43f3-98b7-3cdf8ba2ec39](https://acoustid.org/track/b5123596-4a8d-43f3-98b7-3cdf8ba2ec39)** | 38 | 0.862 |
| [Essence](https://musicbrainz.org/recording/dda7708a-5559-4e04-b11a-268268941dc7) | 04:07 | [Absurd Minds](https://musicbrainz.org/artist/979ce33b-4d8e-4e44-8c7e-f9d3054f4386) | [The Focus](https://musicbrainz.org/release/1e6efde9-e4fe-4c91-9153-86030a601cc6) | 1/13 | 1/1 | 2003-10-13 | DE | Album | Digital Media | **[b5123596-4a8d-43f3-98b7-3cdf8ba2ec39](https://acoustid.org/track/b5123596-4a8d-43f3-98b7-3cdf8ba2ec39)** | 38 | 0.823 |

I have modeled the above roughly following the recording search dialog. As the results are similar I think they should use a similar display, including consistent wording.

A few other observations:

- Do not copy the result to the clipboard by default. This is unexpected. If needed there could be a button for it.
- "Export grid" is a nice idea, but probably should be renamed to "Export results". It also should indicate the file format (e.g. use a .csv file ending). It also looks like it defaults to exporting to the working directory of Picard, which usually is just where the executable is located. Default to the user's home instead and maybe remember last saving position.

Sophist-UK:
> First off: I like the general idea of allowing the user to select a different result if multiple matches are found. 

Me too.
> I think in general using Picard should neither require understanding of Picard's source code nor the web service responses of MB or AcoustID. 

I'm afraid that I agree with Philip here.
> I think this should be changed so that it shows all the results in decreasing order of relevance, with the most important information shown that allows one to make a decisions. Entries should only be listed once with the currently selected shown on top. 

I agree.
> SNIP
> I have modelled the above roughly following the recording search dialog. As the results are similar I think they should use a similar display, including consistent wording.

I would personally like to see the similarity score as the first column - but this is only a personal preference.
> A few other observations:
> * Do not copy the result to the clipboard by default. This is unexpected. If needed there could be a button for it.

This change is IMO essential. What if the user has something important on the clipboard?

phw:
> I would personally like to see the similarity score as the first column - but this is only a personal preference.

I have really no strong opinion here. But I have a slight issue with exposing this value too much here. I see how this is useful to list it here as it has some relevance. But in the end it's also an implementation details that one should not put too much weight on the actual value. It's more important to see in relation to other results. I think two things are important here:

- The actual value can change between Picard releases if the weights get changed or new criteria are added. So something like "Picard 2.3 gave a score of 0.92, but Picard 2.4 only gives 0.87" is not a valid bug report.
- You usually will seldom get a 1.0 score unless you have pushed the release type sliders to the outer right.

But of course in this context it is useful to show. Also it makes it visible to advanced users how the results relate to the thresholds configured in Advanced > Matching. Thinking about it it could even be visualized here in some way if a result got excluded if it was below threshold. Do these result currently show up in this list?

picard/ui/mainwindow.py
```diff
@@ -1080,6 +1096,15 @@ def update_actions(self):
...
         self.track_search_action.setEnabled(have_objects)
 
+        try:
+            enable_match_dialog = False
+            if single.linked_files[0].filename in match_details_log.get_all_match_dtls():
+                enable_match_dialog = have_objects
+        except (IndexError, AttributeError):
+            enable_match_dialog = False
+
+        self.track_match_details_action.setEnabled(enable_match_dialog)
```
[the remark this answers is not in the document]
RaysDev: I'm little confused on this one.  On original PR, you told me to code it this way.  I've made some additional changes as best I can figure what you have in mind. check out next commit.
>image

picard/ui/matchdetailsdialog/__init__.py
RaysDev: My comment was mostly to guide you on the review of a large new file.

Doing a diff, they have similar structure, but there is little duplicated logic.

I think it is beyond scope of this PR to re-write SearchDialog to use a common base class with MatchDetailsDialog, beyond the PicardDialog they already share.  Maybe this is best for a separate refactoring PR.

picard/ui/matchdetailsdialog/__init__.py
RaysDev: Ideally all this logic would have already been in a super class that SearchDialog inherited from, then I could have used it in MatchDetailsDialog.  I just don't see messing with SearchDialog, to be good to do as part of this PR.  Let me know otherwise.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: \n does work, the Qt.TextWordWrap unfortunately does not.  I spent alot of time googling and trying different things but it never had any affect.  Even if it did, it lacks control over where the breaks occur.

for the column key+labels, I followed same style as what you already have in SearchDialog
>image

for the i18n, I must have misunderstood your previous comment.  I thought you told me to remove it.
>image

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: Made more sense to me when I had the catchall, so now I removed it.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: Changed to IOError only.

picard/webservice/__init__.py
RaysDev: Agreed, I had documented it and explained it best I could.  not much other options without drastic code changes.  Its a moot point now, as phw has asked to remove the display of json, there is no need to capture it anymore.  I'll be removing this code.

RaysDev:
Per PHW comment above.

> I think this should be changed so that it shows all the results in decreasing order of relevance, with the most important information shown that allows one to make a decisions. Entries should only be listed once with the currently selected shown on top. Hence I propose the following changes:
> 
> 1. Remove the display of the raw data at the top. If this is really wanted the link to the lookup result is sufficient, but it should be moved below the table

Done

> 2. In the table completely remove the columns loq seq, better?, track ID, entry type, scoring parts, scoring release parts and recording base score. The latter three could be output as debug messages during the lookup (if they not already are, didn't check)

Done, but I added and Advanced|Matching option (default off) to show the extra columns

> 3. Combine corresponding ID and title into a single field with the title linked to the ID
> 4. Link the AcoustID

All the Ids are now gone. However, leaving the linking for future effort. Also not sure where you wanted it to link to?  MusicBrainz I would guess?, but where on the Acoust Result Id?

> 5. Maybe combine track /medium count and total into a single field with [count] / [total]

Skipping this one, since it would prevent sorting/filitering on the value in a useful way.

> 6. Highlight the currently selected entry

Skipping for now.  I think it would be hard/time consuming to code, and seems like a "nice to have", so probably in future PR.  On first time into the dialog, selection is always top row,  so its only if you come back in a second time.

> 7. Round by descending similarity

Done - Assume you meant "sort" vs "Round" 

> 8. Round the similarity to 2 or 3 digits

Done

 
> * Do not copy the result to the clipboard by default. This is unexpected. If needed there could be a button for it.

Done

> * "Export grid" is a nice idea, but probably should be renamed to "Export results". It also should indicate the file format (e.g. use a .csv file ending). It also looks like it defaults to exporting to the working directory of Picard, which usually is just where the executable is located. Default to the user's home instead and maybe remember last saving position.

Done - Renamed; changed default directory; add ".tsv" as file extension since tab delimited

RaysDev:
I have made all the changes requested by zas and phw above.  A few of them I provided specific feedback back on inline above, otherwise I made all the changes.  I need a little more time to test before committing...

RaysDev:
>Highlight the currently selected entry
>>Skipping for now. I think it would be hard/time consuming to code, and seems like a "nice to have", so probably in future PR. On first time into the dialog, selection is always top row, so its only if you come back in a second time.

Done. I was able to get this working

picard/ui/matchdetailsdialog/__init__.py
zas: > Ideally all this logic would have already been in a super class that SearchDialog inherited from

Well, until this PR, there was no reason to do so. Though we usually try to reduce code redundancy, and when such situation arises, we extract common stuff and make a preliminary PR introducing classes to be re-used in subsequent PR.

I did a quick comparaison, there are a lot of lines in common. You even introduced small changes that artificially decrease code duplication, like in `save_table_header_state()`, `restore_table_header_state()`, where the only change is the way you format debug message, one using sprintf-style the other format(), but with exact same message and parameter.

This is definitively something we want to avoid. So let's just make a preliminary PR, that moves common code you need in your new classes to separate classes you'll inherit from.

@phw what do you think?

picard/match_details_log.py
```diff
@@ -0,0 +1,240 @@
...
+
+import numbers
+import traceback
+
+from picard import log
+
+
+__all_file_result_selected_entry_nbr = {}  # dictionary of song file name and selected_entry
+__all_file_match_dtls = {}                 # dictionary of a song file and match detail entries associated with it.
+__cur_file = None
```
zas: I'm not sure to understand comments there.
What's a "song file"? What is the exact purpose of those dictionnaries?

I noticed you shorten variable and class names without good reason:
`dtls` ->  `details`
`lbl` -> `label`
`cur` -> `current`
`nbr` -> `number`
etc.

picard/match_details_log.py
```diff
@@ -0,0 +1,240 @@
...
+
+
+def add_match_dtl_entry(match_dtl_entry):
+
+    # Only log if we are being called for a file comparison (AcoustId Scan or Lookup CD) not track comparison (Search Similiar)
+    global __cur_file
+    if __cur_file is None:
+        return
+
+    log.debug("MATCH:" + str(match_dtl_entry))
```
zas: Prefer:
`log.debug("MATCH: %s", match_dtl_entry)`

phw:
> So let's just make a preliminary PR, that moves common code you need in your new classes to separate classes you'll inherit from.

Yes, agreed on this

picard/match_details_log.py
```diff
@@ -0,0 +1,240 @@
...
+               "{m.trk_id:s}\t"
+               "{m.trk_pos:2d}\t"
+               "{m.med_trk_cnt:2d}\t"
+               "{m.rel_tot_trk_cnt:2d}\t"
+               "{parts:s}\t"
+               "{release_parts:s}\t"
+               "{m.recording_base_score:5.3f}")
+
+        fmtd_str = fmt.format(m=self,
+                              rg_sec_type=str(self.rg_sec_type),
```
zas: Those explicit conversion to `str` should not be needed.

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,386 @@
...
+
+    @property
+    def table_headers(self):
+        return list(self.columns.values())
+
+    def colpos(self, colname):
+        return self.__colkeys.index(colname)
+
+    def set_acoustid_lookup_url_edit(self, p_acoustid_lookup_url):
+        self.acoustid_lookup_url.setText("<a href='" + p_acoustid_lookup_url + "'>" + _("Click to see AcoustId Fingerprint Lookup in browser") + "</a> ")
```
zas: Please use formatting instead of string concatenation:

```python
self.acoustid_lookup_url.setText(_('<a href="{url:s}">Click to see AcoustId Fingerprint Lookup in browser</a>').format(url=p_acoustid_lookup_url))
```

It gives context to translators, and they can properly escape string knowing that's html.

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,386 @@
...
+        self.input_filename_lbl.setFocusPolicy(QtCore.Qt.StrongFocus)
+        myFont = QtGui.QFont()
+        myFont.setBold(True)
+        self.input_filename_lbl.setFont(myFont)
+        self.input_filename_lbl.setTextInteractionFlags(QtCore.Qt.TextSelectableByMouse)
+        self.verticalLayout.addWidget(self.input_filename_lbl)
+
+        self.tip_widget = QtWidgets.QWidget(self)
+        self.tip_lbl = QtWidgets.QLabel(self.tip_widget)
+        self.tip_lbl.setText(_("\nTip: Column header left click to sort, right click to filter"))
```
zas: Is this `\n` strictly needed?

picard/ui/matchdetailsdialog/matchdetails.py
zas: >  I must have misunderstood your previous comment. I thought you told me to remove it.

What I meant is that strings like `Rec Len` aren't easy to translate, translators have to understand what they have to translate, so `Recording Length` is much better. But yes, you have to mark those strings for translation.

zas:
@RaysDev please do not rebase each time you make changes it makes those hard to follow, such big set of changes shouldn't be 1 commit. Thanks ;)

RaysDev:
>I tried to test the current state again, but received some exceptions for various cases:.

 all set now.  sorry, I noticed it after committing, but it was late, so fixed it today.

picard/match_details_log.py
RaysDev: "song file"  :) I couldn't think of a more descriptive name. its what you drag from left pane to the middle pane.  I could use just "file"?   These dictionaries are where we hold all the match data, and selections for all the files that have been "scanned" or "looked up"

Right, its common where I come from to use abbreviations.  Do you not do that?  Only guide I found was  Pep-0008 "use English words wherever feasible (in many cases, abbreviations and technical terms are used "  so seems its ok to use abbreviations.

picard/match_details_log.py
RaysDev: seems with out it I get TypeError: unsupported format string passed to list.__format__  
I tried some other variations and I get KeyError.  Please share if you know how to do it.

picard/ui/matchdetailsdialog/__init__.py
RaysDev: No, I think its left over from some of the json info I was previously showing.  I removed it.

picard/ui/matchdetailsdialog/__init__.py
RaysDev: > please do not rebase each time you make changes it makes those hard to follow, such big set of changes shouldn't be 1 commit.

**Noted going forward.**  Sorry, I thought in the beginning it was suggested that way.  I was "git squashing" it down each time.

picard/match_details_log.py
zas: Ok, so "song file" is an audio file,  File in Picard (see https://github.com/metabrainz/picard/blob/3c3a4c2c8e39aef5c5b7030626a71ed317d6d1ea/picard/file.py#L88)

I see `__all_file_result_selected_entry_nbr[p_filename] = p_selected_entry_nbr` in the code, so keys for this dictionnary are file paths, right?
Since we already have those infos in `File` wouldn't make sense to store details in those objects?

> Right, its common where I come from to use abbreviations.

When it comes to Picard code, it is a bad practice, having an explicit name for a class, function or variable is (in most cases) better. Of course, they are exceptions, but using `dtls` instead `details` is hardly useful, and server absolutly no purpose nowadays (a long time ago length of variable names was limited by language, or available memory, but that's not the case anymore since a long long time).

picard/match_details_log.py
zas: Likely because `self.rg_sec_type` isn't a string, but a list. But I think the conversion should be done once when setting the property.

https://github.com/metabrainz/picard/pull/1497/files#diff-ce85bb5328861d2a80f6a7abefcf2338R119

picard/match_details_log.py
RaysDev: I updated the comment to reflect File instead of "song file".

I'll avoid the abbreviations going forward.

Yes we could track the selected_entry_nbr in the File class, as its at the same level of granularity.  but I think the solution is more modular this way.  I think its better to encapsulate the logic and state information for the match details in one module.  This variable holds state info for the match details rather than describing a File.  Without the Match Details Dialog, it would have no context/purpose.

picard/match_details_log.py
RaysDev: Done, conversion done once when creating object.  Also took concept further that we should not have to execute format() more than once either.  so doing it once now when object is created.

RaysDev:
I repointed PR to devel branch as requested.

RaysDev:
I made changes to Match Dialog to leverage new superclass.  I believe at this point I have completed all requested changes.

RaysDev:
@phw just saw your note on the copyright notice.  I will work on that next...

RaysDev:
> @phw just saw your note on the copyright notice. I will work on that next...

I did the copyrights on all the PRs.  Please be aware that test_acoustid.py did not have a notice, so I added it, but it should probably have some other names besides mine.

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+    def __init__(self, parent):
+        """ Initialize the dialog box object with information about itself, including the columns in the table/grid."""
+        super().__init__(parent)
+        self.file_ = None
+        self.setWindowTitle(_("Match Details"))
+        self.columns = [
+            ('nbr',                  "_Log\nSeq"),
+            ('match_dtl_entry_type', "_Entry Type"),
+            ('is_new',               "_Better?"),
+            ('rec_id',               "_Recording ID"),
```
zas: Why are label strings starting with an underscore?

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            ('acoust_id',            "AcoustId ID"),
+            ('rec_src_cnt',          "AcoustId\nRecording\nSource\nCount"),
+            ('similarity',           "Similarity")
+        ]
+
+    def load_match_details(self, track, file_):
+        """ Retrieves data and loads it to the screen/dialog box """
+        if file_ is not None:
+            # NOTE: some of the values, like meta, are not exposed as constants
+            fp = file_.acoustid_fingerprint if hasattr(file_, 'acoustid_fingerprint') else ""
```
zas: You can use [`getattr()`](https://docs.python.org/3/library/functions.html#getattr) optional `default` parameter:

```python
fp = getattr(file_, 'acoustid_fingerprint', '')
```

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            dur = file_.acoustid_length if hasattr(file_, 'acoustid_length') else ""
```
zas: `dur` -> `duration`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+
+            fingerprint_lkp_url = ("https://" + ACOUSTID_HOST + "/v2/lookup"
+                                   + "?meta=recordings%20releasegroups%20releases%20tracks%20compress%20sources"
+                                   + "&fingerprint=" + fp
+                                   + "&duration=" + str(dur)
+                                   + "&client=" + ACOUSTID_KEY
+                                   + "&clientversion=" + PICARD_VERSION_STR
+                                   + "&format=json") if fp is not None else ""
```
zas: I'd make this test more obvious. If I understand correctly you set `fingerprint_lkp_url` to empty string if `fp` is None. Since the expression is already very long, I almost missed it.

But in this case, you still call https://github.com/metabrainz/picard/pull/1497/files#diff-a470bf968b12bbf5662c8ba4c21f00f7R172 which builds a link with an empty url.

1. I'd make the test `fp is None` more obvious
2. Get rid of `set_acoustid_lookup_url_edit` which is only used here
3. Better handling of the `fp is None` when it comes to `acoustid_lookup_url` label (hide?)

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            self.set_input_filename_lbl(file_.filename)
+            self.set_acoustid_lookup_url_edit(fingerprint_lkp_url)
+
+        self.current_file = file_
+        self.prepare_table()
+        self.entries = match_details_log.get_match_dtls(file_.filename)
+        display_row_nbr = -1
+        selected_row_nbr = -1
+        for entry_nbr, entry in enumerate(self.entries):
+            if entry.match_dtl_entry_type == "Recording candidate release score":
```
zas: Perhaps we should use a numeric type rather than the label directly.
It has various major advantages:
- if you want to change the text, you change it at one place
- it speeds up this comparaison
- it shows clearly what are all possible types
- those numeric types can be whatever you want (class attributes, module level "constants", etc.)

The overall idea is something like:

```python

class MatchDetailsEntryType:
    yourtype1 = 1
    yourtype2 = 2
    ...

type2label = {
    MatchDetailsEntryType.yourtype1: N_('your label'),
     ...
}

if entry.match_dtl_entry_type == MatchDetailsEntryType.yourtype1:
    label = type2label[MatchDetailsEntryType.yourtype1]
    # you may mark the string for translation in the type2label with N_() and translate with _(label)
    do_something_with_label(_(label))
    ...


match_details_log.MatchDtlEntry(MatchDetailsEntryType.yourtype1, ...)
```

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+    def load_selection(self, entry_nbr):
+        """Load the album corresponding to the selected track.
+
+        Allow user to pick from our new screen and use that track instead of one its currently assigned to.
+        Based on logic from searchdialog/track.py line 186
+        """
+
+        entry = self.entries[entry_nbr]
+
+        if entry.rel_id != "":
```
zas: `if entry.rel_id:`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+                if album._files == 0:
+                    # Remove album if it has no more files associated
+                    self.tagger.remove_album(album)
+            else:
+                self.tagger.move_file_to_track(self.current_file, entry.rel_id, entry.rec_id)
+        else:
+            if self.current_file and getattr(self.current_file.parent, 'album', None):
+                album = self.current_file.parent.album
+                self.tagger.move_file_to_nat(self.current_file, entry.rec_id, None)
+                if album._files == 0:
```
zas: `if not album._files:`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+
+        log.debug("using alternative release id:" + entry.rel_id)
+
+    def export_match_dtl_entries_to_file(self):
+        """ Export entries to a file that can be viewed in external program
+        """
+        try:
+            # Prompt user for file location to save to
+            dir = str(Path.home())
+            suggested_file_and_path = dir + os.sep + "picard_match_results.tsv"
```
zas: Use `os.path.join()` or `PurePath.joinpath()`

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+            save_file = QFileDialog.getSaveFileName(None, 'Save File', suggested_file_and_path)
+            if save_file[0] == '':
```
zas: Is `save_file` possibly empty?

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,211 @@
...
+                filename = save_file[0]
+
+            # Alternatively write to temp dir:
+            #filename = tempfile.gettempdir() + os.path.sep + "picard_match_details.tsv"
+
+            with open(filename, "w") as text_file:
+                print(match_details_log.MatchDtlEntry.hdr_tsv, file=text_file)
+
+                for row, entry in enumerate(self.entries):
+                    output_str = str(row) + "\t" + entry.fmtd_as_tsv + os.linesep
```
zas: Not sure we want the file format to be system-dependent.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: These columns are considered private and are not shown in the dialog, unless the Advanced|Matching option is set.  I believe the leading underscore is common convention of indicating private vars in python.  Ideally would have added an is_hidden attribute to "columns", but that could break lots of code, and there was a specific comment already in Picard stating  "# self.columns has to be an ordered dict, with column name as keys, and matching label as values"  so I did not want to modify it.  This logic is documented in matchdetailsdialog.py line 229 in show_table() where its used.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: Done. all cleaned up, `fp is None` is more prominent, and the clickable link in the UI is only shown if the fingerprint is available.  also expanded the fp abbreviation :)

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: Done.  Yes you are absolutely right.  I had previously considered making it an enum, and now I've made the change.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: not possible to be empty.  you can't hit "save" with an empty file name, and if you hit "cancel" you get an array with one element ''.

picard/ui/matchdetailsdialog/matchdetails.py
RaysDev: do you mean the line separators?  I can't think of a reason why you would want to hardcode it, as it would be wrong for some os.

RaysDev:
all new changes requested are done and committed.

RaysDev:
just checking in on this PR? is it in progress?

phw:
@RaysDev sorry for the lack of feedback fr our side. But yes, in general I still would like to see this finished. I am not available 5his weekend, but I would like to see your other PR getting reviewed and hopefully merged next week. Then we can focus on getting this here updated. Currently I think there still needs to be some updates, but I need to look at it in detail. I'll see to get you some feedback early next week.

phw:
So we are finally getting somewhere with the related stories merged :) @RaysDev Could you rebase this branch against current devel?

RaysDev:
I've done the rebase against current devel branch.  It was a little confusing with the conflicts has some of my other PRs were also in here, like the superclass for search dialog.  I believe i got everything.  I did a compare, and ran a few tests, and its looking good.  Really I'd like to do a little more compare/testing, but its very late here, and so will have to follow up on that later.  Thanks.

picard/match_details_log.py
```diff
@@ -0,0 +1,243 @@
...
+    global __cur_file
+    __cur_file = p_file
+
+
+def set_file_result_selected_entry_nbr(p_filename, p_selected_entry_nbr):
+    """ Sets the result selected_entry_nbr associated with this file
+
+    This is the entry user last chose from the Match Dialog
+    """
+    if (p_selected_entry_nbr is not None):
```
zas, line 47: Remove useless parenthesis here.

picard/match_details_log.py
```diff
@@ -0,0 +1,243 @@
...
+    return __all_file_match_dtls.get(filename)
+
+
+def get_all_match_dtls():
+    return __all_file_match_dtls
+
+
+def can_get_match_dtls(filename):
+    can_get_match_dtls = filename in __all_file_match_dtls
+    return can_get_match_dtls
```
zas, line 66: Since it is simple enough: `return filename in __all_file_match_dtls`

picard/match_details_log.py
```diff
@@ -0,0 +1,243 @@
...
+    # Only log if we are being called for a file comparison (AcoustId Scan or Lookup CD) not track comparison (Search Similiar)
+    global __cur_file
+    if __cur_file is None:
+        return
+
+    log.debug("MATCH: %s", str(match_dtl_entry))
+
+    match_dtl_entries = __all_file_match_dtls.get(__cur_file.filename)
+    if match_dtl_entries is None:
+        match_dtl_entries = []
```
zas, line 80: Is it possible to drop this statement using `get()` default value?

```python
match_dtl_entries = __all_file_match_dtls.get(__cur_file.filename, [])
```

RaysDev:
I've completed the feedback items you provided @zas and commited.

zas:
> I've completed the feedback items you provided @zas and commited.

Thanks.
Can you also address Codacy check? https://app.codacy.com/manual/MetaBrainz/picard/pullRequest?prid=5052949

RaysDev:
I had reviewed it, but its invalid. This is an implementation of an abstract method from superclass, so even though this subclass's logic does not reference any variables, It can't be converted to a function.

RaysDev:
hi, just checking in on this? please let me know

phw:
@RaysDev Sorry for the delay, but I also want to have this move forward. Last weeks have been rather busy on my side, though.

With your other changes merged this PR looks much better now. We just need to rebase it one more time. I have already done this in https://github.com/phw/picard/tree/match_dialog , you can either take over this branch or rebase yourself.

I think we'll have a few comments on the code, but overall. A few comments already from running the code:

grafik

- Should it really show the "Release Group Title"? I'd rather change it to use the "Release Title". Opinions?
- I think with the heade being two lines anyway we could write out "Primary Type" and "Secondary Type". The way this is abbreviated here doesn't work well with translations anyway. And I think we could get away without mentioning "Release Group" here, so just "Primary Type", "Secondary Type"
- Can we have "Secondary Type" values a comma spearates list, instead of JSON array?
- Instead of "no sec type" / "no pri type" we could just leave it empty
- "AcoustID Recording Source Count" is really a long label, but not sure how to shorten it. Does "AcoustID Source Count" work? A bit shorter at least

phw:
Just detected the filters, those are pretty cool. Just a few notes:

- The layout looks a bit cramped, should have at least some padding around
- Right clicking the column header always resets the filter

grafik

RaysDev:
> * Should it really show the "Release Group Title"? I'd rather change it to use the "Release Title". Opinions?

I based my design on studying AcoustId response json, and in that case ReleaseGroup contain releases, and hence I would prefer to stick with ReleaseGroupTitle.  However, in looking at the MusicBrainz response json, its surprisingly reversed.  Meaning Releases contain ReleaseGroups, and hence I can understand your suggestion.  Since the Title values are the same in every case I looked at, and its already coded/tested, lets leave it as is.  Let me know otherwise.

> * I think with the heade being two lines anyway we could write out "Primary Type" and "Secondary Type". The way this is abbreviated here doesn't work well with translations anyway. And I think we could get away without mentioning "Release Group" here, so just "Primary Type", "Secondary Type"

Done for writing out the full words, but would please like to keep "Release Group" as I think "type" becomes vague without that context


> * Can we have "Secondary Type" values a comma spearates list, instead of JSON array?

Done

> * Instead of "no sec type" / "no pri type" we could just leave it empty or use something generic ("none", "n/a")

Done

> * "AcoustID Recording Source Count" is really a long label, but not sure how to shorten it. Does "AcoustID Source Count" work? A bit shorter at least

Done

RaysDev:
> *The layout looks a bit cramped, should have at least some padding around
I'll look into this.

> * Right clicking the column header always resets the filter

This wasn't a defect, just a lack of robust functionality.  I made some enhancements and now works the way you would expect it.  retaining selected values, etc...

RaysDev:
As mentioned above I made the requested changes,  give me little time to review the new git branch you created vs rebase, and I'll get it committed.

RaysDev:
> We just need to rebase it one more time. I have already done this in https://github.com/phw/picard/tree/match_dialog , you can either take over this branch or rebase yourself.

I cloned your repo and made changes to the branch you mention above.  I don't have permission to commit there.  Is there something you need to grant?  I thought this is what you meant by "take over this branch" but if you meant something else, let me know.

phw:
> I cloned your repo and made changes to the branch you mention above. I don't have permission to commit there. Is there something you need to grant? I thought this is what you meant by "take over this branch" but if you meant something else, let me know.

Sorry if I added confusion. I actually meant you could push this branch to your branch here to update this PR

RaysDev:
I rebased my match_dialog branch on top of upstream/devel, and so my code changes are on top of that.  The original feature and the ui changes you requested week or so ago are squashed, but they are in there.  I struggled for a while with GIT, but I believe it is now the way you wanted it.

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,278 @@
...
+
+        # sort
+        self.table.setSortingEnabled(self.sorting_enabled)
+        if self.sorting_enabled and sort_column:
+            self.table.sortItems(self.colpos(sort_column), sort_order)
+
+        # Hide debug cols. using first char of label to mark it private
+        should_hide_debug_cols = not config.setting['show_debug_col_in_match_dialog']
+        for i in range(self.table.columnCount()):
+            is_debug_col = self.table.horizontalHeaderItem(i).text().startswith('_')
```
zas, line 263: Just a plain No to this.

Remove '_' from labels, add `set()` for shown (or hidden) columns.

```patch
diff --git a/picard/ui/matchdetailsdialog/__init__.py b/picard/ui/matchdetailsdialog/__init__.py
index 62f473f9..30e11e6b 100644
--- a/picard/ui/matchdetailsdialog/__init__.py
+++ b/picard/ui/matchdetailsdialog/__init__.py
@@ -260,7 +260,7 @@ class MatchDetailsDialogBase(TableBasedDialog):
         # Hide debug cols. using first char of label to mark it private
         should_hide_debug_cols = not config.setting['show_debug_col_in_match_dialog']
         for i in range(self.table.columnCount()):
-            is_debug_col = self.table.horizontalHeaderItem(i).text().startswith('_')
+            is_debug_col = self.colname(i) in self.debug_only_columns
             hide_this_col = is_debug_col and should_hide_debug_cols
             self.table.setColumnHidden(i, hide_this_col)
 
diff --git a/picard/ui/matchdetailsdialog/matchdetails.py b/picard/ui/matchdetailsdialog/matchdetails.py
index c23f3558..b51e6617 100644
--- a/picard/ui/matchdetailsdialog/matchdetails.py
+++ b/picard/ui/matchdetailsdialog/matchdetails.py
@@ -56,34 +56,46 @@ class MatchDetailsDialog(MatchDetailsDialogBase):
         self.file_ = None
         self.setWindowTitle(_("Match Details"))
         self.columns = [
-            ('nbr',                  "_Log\nSeq"),
-            ('match_dtl_entry_type', "_Entry Type"),
-            ('is_new',               "_Better?"),
-            ('rec_id',               "_Recording ID"),
+            ('nbr',                  "Log\nSeq"),
+            ('match_dtl_entry_type', "Entry Type"),
+            ('is_new',               "Better?"),
+            ('rec_id',               "Recording ID"),
             ('rec_title',            "Recording Title"),
             ('rec_len',              "Recording\nLength"),
             ('rec_artist',           "Recording\nArtist"),
-            ('rg_id',                "_Release Group ID"),
+            ('rg_id',                "Release Group ID"),
             ('rg_title',             "Release Group\nTitle"),
             ('rg_pri_type',          "Release Group\nPrimary Type"),
             ('rg_sec_type',          "Release Group\nSecondary Types"),
-            ('rel_id',               "_Release ID"),
+            ('rel_id',               "Release ID"),
             ('rel_ctry',             "Release\nCountry"),
             ('rel_dt',               "Release\nDate"),
             ('med_fmt',              "Medium\nFormat"),
             ('med_pos',              "Medium\nPosition"),
             ('med_tot_cnt',          "Medium\nCount"),
-            ('trk_id',               "_Track ID"),
+            ('trk_id',               "Track ID"),
             ('trk_pos',              "Track\nPosition"),
             ('med_trk_cnt',          "Medium\nTrack\nCount"),
             ('rel_tot_trk_cnt',      "Release\nTotal Track\nCount"),
-            ('parts',                "_Scoring Parts"),
-            ('release_parts',        "_Scoring Release Parts"),
-            ('recording_base_score', "_Recording\nBase Score\n(Sources Ratio)"),
+            ('parts',                "Scoring Parts"),
+            ('release_parts',        "Scoring Release Parts"),
+            ('recording_base_score', "Recording\nBase Score\n(Sources Ratio)"),
             ('acoust_id',            "AcoustId ID"),
             ('rec_src_cnt',          "AcoustId\nSource\nCount"),
             ('similarity',           "Similarity")
         ]
+        self.debug_only_columns = {
+            'is_new',
+            'match_dtl_entry_type',
+            'nbr',
+            'parts',
+            'rec_id',
+            'recording_base_score',
+            'release_parts',
+            'rel_id',
+            'rg_id',
+            'trk_id',
+        }
 
     def load_match_details(self, track, file_):
         """ Retrieves data and loads it to the screen/dialog box """
diff --git a/picard/ui/tablebaseddialog.py b/picard/ui/tablebaseddialog.py
index 3381f854..da35d2bd 100644
--- a/picard/ui/tablebaseddialog.py
+++ b/picard/ui/tablebaseddialog.py
@@ -115,6 +115,9 @@ class TableBasedDialog(PicardDialog):
     def colpos(self, colname):
         return self.__colkeys.index(colname)
 
+    def colname(self, pos):
+        return self.__colkeys[pos]
+
     @abstractmethod
     def get_value_for_row_id(self, row, value):
         pass
```

picard/ui/matchdetailsdialog/matchdetails.py
zas: image

See my other comment about this.

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,226 @@
...
+                                       + "&format=json")
+            self.set_acoustid_lookup_url_edit(fingerprint_lkp_url)
+
+        self.current_file = file_
+        self.prepare_table()
+        self.entries = match_details_log.get_match_dtls(file_.filename)
+        display_row_nbr = -1
+        selected_row_nbr = -1
+
+        regex = re.compile(r"[\[\]]")
```
zas, line 115: What's the purpose of this regex?

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,226 @@
...
+                self.set_table_item_val(display_row_nbr, 'med_fmt', entry.med_fmt)
+                self.set_table_item_val(display_row_nbr, 'med_pos', entry.med_pos)
+                self.set_table_item_val(display_row_nbr, 'med_tot_cnt', entry.med_tot_cnt)
+                self.set_table_item_val(display_row_nbr, 'trk_id', entry.trk_id)
+                self.set_table_item_val(display_row_nbr, 'trk_pos', entry.trk_pos)
+                self.set_table_item_val(display_row_nbr, 'med_trk_cnt', entry.med_trk_cnt)
+                self.set_table_item_val(display_row_nbr, 'rel_tot_trk_cnt', entry.rel_tot_trk_cnt)
+                self.set_table_item_val(display_row_nbr, 'parts', str(entry.parts))
+                self.set_table_item_val(display_row_nbr, 'release_parts', str(entry.release_parts))
+                self.set_table_item_val(display_row_nbr, 'recording_base_score', entry.recording_base_score)
```
zas, line 153: It'd be more readable to iterate a list imho.

Something like:

```python
column_values = [
    ('nbr', entry_nbr),
    ('match_dtl_entry_type', entry.match_dtl_entry_type),
    ('is_new', entry.is_new_str),
    ('similarity', "%5.3f" % entry.similarity),
...
]

self.table.insertRow(display_row_nbr)
for colname, value in column _values:
    self.set_table_item_val(display_row_nbr, colname, value)

```

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,278 @@
...
+        self.accept_button.setEnabled(False)
+        self.buttonBox.addButton(self.accept_button, QtWidgets.QDialogButtonBox.AcceptRole)
+
+        self.buttonBox.addButton(StandardButton(StandardButton.CANCEL), QtWidgets.QDialogButtonBox.RejectRole)
+        self.buttonBox.accepted.connect(self.accept)
+        self.buttonBox.rejected.connect(self.reject)
+        self.verticalLayout.addWidget(self.buttonBox)
+
+    # Override
+    def show_table(self, selected_row_nbr=-1, sort_column=None, sort_order=QtCore.Qt.DescendingOrder):
```
zas, line 248: You override with a different prototype, Codacy will complain.

Perhaps define overriden method as:
```python
def show_table(self, sort_column=None, sort_order=QtCore.Qt.DescendingOrder, **kwargs):
```

and this one as:

```python
def show_table(self, sort_column=None, sort_order=QtCore.Qt.DescendingOrder, selected_row_nbr=-1, **kwargs):
```

Not sure.

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,226 @@
...
+                self.tagger.move_file_to_nat(self.current_file, entry.rec_id, None)
+
+        log.debug("using alternative release id:" + entry.rel_id)
+
+    def export_match_dtl_entries_to_file(self):
+        """ Export entries to a file that can be viewed in external program
+        """
+        try:
+            # Prompt user for file location to save to
+            dir = str(Path.home())
```
zas, line 204: `dir` is a reserved keyword, and in this case rather useless intermediate variable.
This should work as well:

```python
os.path.join(Path.home(), "picard_match_results.tsv")
```

picard/ui/matchdetailsdialog/matchdetails.py
```diff
@@ -0,0 +1,226 @@
...
+            suggested_file_and_path = os.path.join(dir, "picard_match_results.tsv")
+            save_file = QFileDialog.getSaveFileName(None, 'Save File', suggested_file_and_path)
+            if save_file[0] == '':
+                return
+            else:
+                filename = save_file[0]
+
+            # Alternatively write to temp dir:
+            # filename = tempfile.gettempdir() + os.path.sep + "picard_match_details.tsv"
```
zas, line 213: This comment should be removed.

picard/ui/matchdetailsdialog/matchdetails.py
zas: What I mean is that you could use `\n` instead of variable `os.linesep` so the format is the same on all platforms.
But if we don't care, you may as well use:

```python
print('%d\t%s' % (row, entry.fmtd_as_tsv), file=text_file)
```

And don't bother with `os.linesep`

picard/ui/options/matching.py
```diff
@@ -43,6 +44,7 @@ class MatchingOptionsPage(OptionsPage):
         config.FloatOption("setting", "file_lookup_threshold", 0.7),
         config.FloatOption("setting", "cluster_lookup_threshold", 0.7),
         config.FloatOption("setting", "track_matching_threshold", 0.4),
+        config.BoolOption("setting", "show_debug_col_in_match_dialog", False),
```
zas, line 47: Instead of a generic option, it could be a way to toggle those in the dialog itself, because it's rather hard to find as is.

picard/ui/matchdetailsdialog/__init__.py
```diff
@@ -0,0 +1,278 @@
...
+            for i in range(self.rowCount()):
+                for j in range(self.columnCount()):
+                    item = self.item(i, j)
+                    if j in self.unique_visible_values_by_col:
+                        if item.text() not in self.unique_visible_values_by_col[j]:
+                            columnsShow[i] = False
+            for key, value in columnsShow.items():
+                self.setRowHidden(key, not value)
+
+        except Exception:
```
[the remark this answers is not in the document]
Sophist-UK: Also, if we are going to have a try/except, I think that the except clause needs to handle specific expected exceptions and not a generic ANY Exception.

RaysDev:
@Phw @zas 

> @RaysDev Sorry for the delay, but I also want to have this move forward. Last weeks have been rather busy on my side, though.
> 
> With your other changes merged this PR looks much better now. We just need to rebase it one more time. I have already done this in https://github.com/phw/picard/tree/match_dialog , you can either take over this branch or rebase yourself.


Please review above history as well as preceding PR #1476  (combined 152 comments), **@PWH said what was in his personal branch was ready to go after I rebased again?**    Are these new change requests really necessary before we can realize some benefit to all this work?

At this point I'll request you make the final tweaks that you feel are required before making this capability available to folks that use Picard.  Its been 5 months since I submitted the PR, and I unfortunately can't work on this anymore.  I've already split from 1 to 5 PRs, made hundreds of requested changes, rebased to devel branch, rebased again as you have made changes on your branch....  It would be really unfortunate if this gets thrown away at this point.

phw:
I'm going to close this. This PR turned out to try to much at the same time, and also adding a dialog with too much technical complexity. We might investigate other approaches of providing easy selection of alternatives.

PR #1499, opened 2020-02-24
Adding / removing preserved tags must normalize the tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
`PreservedTags` only lower cased the tags on loading, not when adding or removing tags. This makes behavior inconsistent when upercased tags are added to the list.

* JIRA ticket (_optional_): PICARD-1765

# Solution

# Action

### DIFF
--- a/picard/util/preservedtags.py
+++ b/picard/util/preservedtags.py
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 # Copyright (C) 2018 Laurent Monin
-# Copyright (C) 2019 Philipp Wolfer
+# Copyright (C) 2019-2020 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -40,12 +40,12 @@ class PreservedTags:
         return tag.strip().lower()
 
     def add(self, name):
-        self._tags.add(name)
+        self._tags.add(self._normalize_tag(name))
         self._to_config()
 
     def discard(self, name):
-        self._tags.discard(name)
+        self._tags.discard(self._normalize_tag(name))
         self._to_config()
 
     def __contains__(self, key):
-        return key.lower() in self._tags
+        return self._normalize_tag(key) in self._tags
--- a/test/test_util_preservedtags.py
+++ b/test/test_util_preservedtags.py
@@ -1,0 +1,55 @@
+# -*- coding: utf-8 -*-
+from test.picardtestcase import PicardTestCase
+
+from picard import config
+from picard.util.preservedtags import PreservedTags
+
+
+class PreservedTagsTest(PicardTestCase):
+    def setUp(self):
+        super().setUp()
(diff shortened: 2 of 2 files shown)


test/test_util_preservedtags.py
```diff
@@ -0,0 +1,46 @@
...
+from test.picardtestcase import PicardTestCase
+
+from picard import config
+from picard.util.preservedtags import PreservedTags
+
+
+class PreservedTagsTest(PicardTestCase):
+    def setUp(self):
+        super().setUp()
+        config.setting['preserved_tags'] = "tag1, tag2"
```
zas: I'd use `PreservedTags.opt_name` instead of `preserved_tags`.

picard/util/preservedtags.py
```diff
@@ -40,11 +40,11 @@ def _normalize_tag(tag):
...
-        self._tags.add(name)
+        self._tags.add(self._normalize_tag(name))
         self._to_config()
 
     def discard(self, name):
-        self._tags.discard(name)
+        self._tags.discard(self._normalize_tag(name))
         self._to_config()
 
     def __contains__(self, key):
```
zas: Imho, we should use `self._normalize_tag(key)` instead of `key.lower()` here too

test/test_util_preservedtags.py
```diff
@@ -0,0 +1,46 @@
...
+        self.assertNotIn("tag1", preserved)
+        # Discard must persists the change
+        self.assertNotIn("tag1", PreservedTags())
+
+    def test_discard_case_insensitive(self):
+        preserved = PreservedTags()
+        self.assertIn("tag1", preserved)
+        preserved.discard("TAG1")
+        self.assertNotIn("tag1", preserved)
+
```
zas: I'd add another test to verify tag order and actual config setting value:

def test_order(self):
    preserved = PreservedTags()
    preserved.add('tag3')
    preserved.add('tag2')
    preserved.add('tag1')
    preserved.discard('tag2')
    self.assertEqual(config.setting[PreservedTags.opt_name], 'tag1, tag3')

picard/util/preservedtags.py
phw: Agreed. I changed this and added a test check for this

picard/util/preservedtags.py
```diff
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 # Copyright (C) 2018 Laurent Monin
-# Copyright (C) 2019 Philipp Wolfer
+# Copyright (C) 20192-2020 Philipp Wolfer
```
zas: ^^typo

phw:
It's a bit difficult to trigger an actual issue caused by this, so maybe nobody actually experienced this yet. But it is possible to experience a small glitch in the behavior of the context menu entry in the metadatabox for adding and removing tags to preserved tags.

I added PICARD-1765 to track this.

PR #1502, opened 2020-02-24
CoverArtImage: improve tests & coverage

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action


test/test_coverart_image.py
```diff
@@ -126,6 +145,483 @@ def test_set_data(self):
...
+        cases = {
+            'options': {
+                'caa_image_type_as_filename': True,
+                'cover_image_filename': 'cover',
+                'save_images_overwrite': False,
+            },
+            'expected': {
+                1: {},
+                2: {'front.png': '1'},
+                3: {'cover.png': '2', 'front.png': '1'},  # FIXME: is this what we expect?
```
phw: I guess this is expected if explicitly marking an image not as `is_front`. But this should be a theoretical case for now (unless we allow users to modify the type, which could happen). But this is another example why I find `caa_image_type_as_filename` problematic.

test/test_coverart_image.py
```diff
@@ -126,6 +145,483 @@ def test_set_data(self):
...
+        cases = {
+            'options': {
+                'caa_image_type_as_filename': False,
+                'cover_image_filename': 'folder',
+                'save_images_overwrite': True,
+            },
+            'expected': {
+                1: {},
+                2: {'folder.png': '1'},
+                3: {'folder.png': '2'},
```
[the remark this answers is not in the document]
zas: Yes, that's the main issue: how do we know images are related? This isn't properly implemented as it, and what to do with pre-existing files ?

test/test_coverart_image.py
zas: Yes, `caa_image_type_as_filename` is problematic.

test/test_coverart_image.py
```diff
@@ -126,6 +145,483 @@ def test_set_data(self):
         self.assertEqual(filesize, len(imgdata))
         self.assertEqual(coverartimage.data, imgdata)
 
+    def test_delete_file_removed(self):
+        payload = b'atest'
+        image_ = create_image(payload, comment='1')
+        image = create_image(payload, comment='2')
```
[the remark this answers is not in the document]
zas: I tried to make things easier to understand in 1aa46ff

zas:
@phw not sure what to do with this PR

test/test_coverart_image.py
phw: After recent discussion that we also had for the cover art options refactoring this would indeed be expected behaviour. If `image_type_as_filename` is set it will use the type as filename for all images except for the cover. The cover will use the `cover_image_filename`.

This example is intentionally confusing I guess, but the image with type "back" is marked as "Is_front", while the image with type "front" is not. Hence the back type image will get the "cover" name and the front type image gets named by type, which is "front".

In practice this could happen only with a badly behaving cover art provider. Current providers would mark type "front" as "is_front".

PR #1505, opened 2020-02-27
PICARD-1774: Return "0" for $lenmulti if 'name' is not entered or is blank

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Return "0" for `$lenmulti` if `name` is not entered or is blank.

# Problem

* JIRA ticket: PICARD-1774

When calling the `$lenmulti` function with no name specified (e.g. `$lenmulti(,)` or `$lenmulti(,; )`) the function returns a count of "1".  This should probably return "0".

# Solution

Return "0" for `$lenmulti` if `name` is not entered or is blank.

# Action

When updating the tests, I noticed an unexpected return value if the function was passed a variable that evaluated to an empty string and a separator override was set.  The function returned "1" when I would expect it to return "0".  Possibly a bug in the multi-var parser?  I noted this on the specific tests.

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -805,6 +805,8 @@ def func_len(parser, text=""):
 
 @script_function(eval_args=False)
 def func_lenmulti(parser, multi, separator=MULTI_VALUED_JOINER):
+    if not multi:
+        return "0"
     return str(len(MultiValue(parser, multi, separator)))
 
 
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -732,6 +732,17 @@ class ScriptParserTest(PicardTestCase):
         # Test no separator
         self.assertScriptResultEquals("$lenmulti(%foo%,)", "1", context)
         self.assertScriptResultEquals("$lenmulti(%bar%,)", "1", context)
+        # Test blank name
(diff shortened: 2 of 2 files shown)


rdswift:
Thinking about it some more, I think this PR is simply addressing a symptom, and won't be required if there is a bug in the multi-var processing that is returning [""] rather than [] when the name is (or evaluates to) "".

phw:
> Thinking about it some more, I think this PR is simply addressing a symptom, and won't be required if there is a bug in the multi-var processing that is returning [""] rather than [] when the name is (or evaluates to) "".

I think this is right. Can you look over https://github.com/metabrainz/picard/pull/1506 ?

This also changes some assertion in tests for $map and $foreach, because those currently also loop 1 time over empty values.

phw:
Fixed by merging https://github.com/metabrainz/picard/pull/1506

rdswift:
Super!  Thanks for fixing the source of the problem.  That makes much more sense that my addressing a symptom.

PR #1508, opened 2020-02-28
PICARD-1776: Trap ValueError exception in $datetime function

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Trap the `ValueError` in `$datetime`.

# Problem

* JIRA ticket: PICARD-1776

When specifying a platform-specific format code (e.g. "%-d" on Windows 7), `$datetime` crashes with a ValueError.

# Solution

Trap the `ValueError` in `$datetime` and ~~return "Invalid Format" message~~ raise a `ScriptRuntimeError` exception.

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -41,6 +41,7 @@ import datetime
 from functools import reduce
 from inspect import getfullargspec
 import operator
+from queue import LifoQueue
 import re
 import unicodedata
 
@@ -150,10 +151,10 @@ class ScriptFunction(object):
             args = [arg.eval(parser) for arg in self.args]
         else:
             args = self.args
-        parser._function_stack.append(self.name)
+        parser._function_stack.put(self.name)
         # Save return value to allow removing function from the stack on successful completion
         return_value = function(parser, *args)
-        parser._function_stack = parser._function_stack[:-1]
+        parser._function_stack.get()
         return return_value
 
 
@@ -185,7 +186,7 @@ Grammar:
     _cache = {}
 
     def __init__(self):
-        self._function_stack = []
+        self._function_stack = LifoQueue()
 
     def __raise_eof(self):
         raise ScriptEndOfFile("Unexpected end of script at position %d, line %d" % (self._x, self._y))
@@ -1281,7 +1282,7 @@ def func_datetime(parser, format=None):
     except ValueError:
         raise ScriptRuntimeError(
             "Unsupported format code",
-            parser._function_stack[-1],
+            parser._function_stack.get(),
             parser._x,
             parser._y
         )
--- a/picard/script.py
+++ b/picard/script.py
@@ -73,6 +73,10 @@ class ScriptUnknownFunction(ScriptError):
     pass
 
 
+class ScriptRuntimeError(ScriptError):
+    pass
+
+
 class ScriptText(str):
 
     def eval(self, state):
@@ -1267,7 +1271,10 @@ def func_datetime(parser, format=None):
     try:
         return datetime.datetime.now(tz=local_tz).strftime(format)
     except ValueError:
-        return ''
+        raise ScriptRuntimeError(
+            "Unsupported format code in $%s at position %i, line %i"
+            % ('datetime', parser._x, parser._y)
+        )
 
 
 @script_function(eval_args=False)
--- a/picard/script.py
+++ b/picard/script.py
@@ -1264,8 +1264,10 @@ def func_datetime(parser, format=None):
     # Handle case where format evaluates to ''
     if not format:
         format = '%Y-%m-%d %H:%M:%S'
-
-    return datetime.datetime.now(tz=local_tz).strftime(format)
+    try:
+        return datetime.datetime.now(tz=local_tz).strftime(format)
+    except ValueError:
+        return ''
 
 
 @script_function(eval_args=False)
--- a/picard/script.py
+++ b/picard/script.py
@@ -74,7 +74,8 @@ class ScriptUnknownFunction(ScriptError):
 
 
 class ScriptRuntimeError(ScriptError):
-    pass
+    def __init__(self, message, name, line, position):
+        super().__init__("${1:s}:{2:d}:{3:d}: {0:s}".format(message, name, line, position))
 
 
 class ScriptText(str):
@@ -149,7 +150,11 @@ class ScriptFunction(object):
             args = [arg.eval(parser) for arg in self.args]
         else:
             args = self.args
-        return function(parser, *args)
+        parser._function_stack.append(self.name)
+        # Save return value to allow removing function from the stack on successful completion
+        return_value = function(parser, *args)
+        parser._function_stack = parser._function_stack[:-1]
+        return return_value
 
 
 class ScriptExpression(list):
@@ -178,6 +183,9 @@ Grammar:
 
     _function_registry = ExtensionPoint(label='function_registry')
     _cache = {}
+
+    def __init__(self):
+        self._function_stack = []
 
     def __raise_eof(self):
         raise ScriptEndOfFile("Unexpected end of script at position %d, line %d" % (self._x, self._y))
@@ -1272,8 +1280,10 @@ def func_datetime(parser, format=None):
         return datetime.datetime.now(tz=local_tz).strftime(format)
     except ValueError:
         raise ScriptRuntimeError(
-            "Unsupported format code in $%s at position %i, line %i"
-            % ('datetime', parser._x, parser._y)
+            "Unsupported format code",
+            parser._function_stack[-1],
+            parser._x,
+            parser._y
         )
 
 
--- a/picard/script.py
+++ b/picard/script.py
@@ -1283,8 +1283,8 @@ def func_datetime(parser, format=None):
         raise ScriptRuntimeError(
             "Unsupported format code",
             parser._function_stack.get(),
-            parser._x,
-            parser._y
+            parser._y,
+            parser._x
         )
 
 
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -32,10 +32,7 @@
 import copy
 import datetime
 import os
-from unittest import (
-    skipIf,
-    skipUnless,
-)
+from unittest import skipUnless
 from unittest.mock import MagicMock
 
 from test.picardtestcase import PicardTestCase
@@ -52,6 +49,7 @@ from picard.script import (
     ScriptExpression,
     ScriptFunction,
     ScriptParser,
+    ScriptRuntimeError,
     ScriptSyntaxError,
     ScriptUnknownFunction,
     register_script_function,
@@ -1194,25 +1192,12 @@ class ScriptParserTest(PicardTestCase):
 
         try:
             context = Metadata()
+            # Tests with Windows-specific format
+            self.assertScriptResultEquals(r"$datetime(\%Y-\%#m-\%#d)", "2020-1-2", context)
             # Tests with invalid format
-            self.assertScriptResultEquals(r"$datetime(\%B \%-d\, \%Y)", "", context)
-        except (AssertionError, ValueError, IndexError) as err:
-            raise err
-        finally:
-            # Restore original datetime object
-            datetime.datetime = original_datetime
-
-    @skipIf(IS_WIN, "non-windows test")
-    def test_cmd_datetime_non_windows(self):
-        # Save origninal datetime object and substitute one returning
-        # a fixed now() value for testing.
-        original_datetime = datetime.datetime
-        datetime.datetime = _DateTime
-
-        try:
-            context = Metadata()
-            # Tests with invalid format
-            self.assertScriptResultEquals(r"$datetime(\%Y-\%-m-\%-d)", "2020-1-2", context)
+            areg = r"^Unsupported format code in \$datetime"
+            with self.assertRaisesRegex(ScriptRuntimeError, areg):
+                self.parser.eval(r"$datetime(\%B \%-d\, \%Y)")
         except (AssertionError, ValueError, IndexError) as err:
             raise err
         finally:
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -31,6 +31,11 @@
 
 import copy
 import datetime
+import os
+from unittest import (
+    skipIf,
+    skipUnless,
+)
 from unittest.mock import MagicMock
 
 from test.picardtestcase import PicardTestCase
@@ -52,6 +57,9 @@ from picard.script import (
     register_script_function,
     script_function,
 )
+
+
+IS_WIN = os.name == 'nt'
 
 
 class _TestTimezone(datetime.tzinfo):
@@ -1170,6 +1178,41 @@ class ScriptParserTest(PicardTestCase):
             areg = r"^Wrong number of arguments for \$datetime: Expected between 0 and 1, "
             with self.assertRaisesRegex(ScriptError, areg):
                 self.parser.eval("$datetime(abc,def)")
+        except (AssertionError, ValueError, IndexError) as err:
+            raise err
+        finally:
+            # Restore original datetime object
+            datetime.datetime = original_datetime
+
+    @skipUnless(IS_WIN, "windows test")
+    def test_cmd_datetime_windows(self):
+        ''' Test failure specific to Windows only. '''
+        # Save origninal datetime object and substitute one returning
+        # a fixed now() value for testing.
+        original_datetime = datetime.datetime
+        datetime.datetime = _DateTime
+
+        try:
+            context = Metadata()
+            # Tests with invalid format
+            self.assertScriptResultEquals(r"$datetime(\%B \%-d\, \%Y)", "", context)
+        except (AssertionError, ValueError, IndexError) as err:
+            raise err
+        finally:
+            # Restore original datetime object
+            datetime.datetime = original_datetime
+
+    @skipIf(IS_WIN, "non-windows test")
+    def test_cmd_datetime_non_windows(self):
+        # Save origninal datetime object and substitute one returning
+        # a fixed now() value for testing.
+        original_datetime = datetime.datetime
+        datetime.datetime = _DateTime
+
+        try:
+            context = Metadata()
+            # Tests with invalid format
+            self.assertScriptResultEquals(r"$datetime(\%Y-\%-m-\%-d)", "2020-1-2", context)
         except (AssertionError, ValueError, IndexError) as err:
             raise err
         finally:
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1202,7 +1202,7 @@ class ScriptParserTest(PicardTestCase):
 
         try:
             context = Metadata()
-            areg = r"^\$datetime:\d+:\d+: Unsupported format code"
+            areg = r"^\$datetime:1:\d+: Unsupported format code"
             # Tests with invalid format code (platform dependent tests)
             for test_case in tests_to_run:
                 with self.assertRaisesRegex(ScriptRuntimeError, areg):
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -31,7 +31,6 @@
 
 import copy
 import datetime
-from unittest import skipUnless
 from unittest.mock import MagicMock
 
 from test.picardtestcase import PicardTestCase
@@ -39,7 +38,6 @@ from test.picardtestcase import PicardTestCase
 from picard import config
 from picard.cluster import Cluster
 from picard.const import DEFAULT_FILE_NAMING_FORMAT
-from picard.const.sys import IS_WIN
 from picard.metadata import Metadata
 from picard.script import (
     MULTI_VALUED_JOINER,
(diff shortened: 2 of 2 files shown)


picard/script.py
```diff
@@ -1258,8 +1258,10 @@ def func_datetime(parser, format=None):
     # Handle case where format evaluates to ''
     if not format:
         format = '%Y-%m-%d %H:%M:%S'
-
-    return datetime.datetime.now(tz=local_tz).strftime(format)
+    try:
+        return datetime.datetime.now(tz=local_tz).strftime(format)
+    except ValueError:
+        return 'Invalid Format'
```
samj1912: It looks like all the other script functions return a blank string on error or a default value which makes sense. Invalid Format seems to be neither of those. Should we just raise the error here or return a blank string instead?

samj1912:
As much as I appreciate these PRs, with all these changes, we have switched from a behavior where we were noisy (but hopefully not crashing) and showing errors in the script to the user, to silently switching things to defaults. Since scripts affect file naming, using defaults might mean we override multiple files to the same name. This might mean potential data loss. I hope we are aware of this subtle change in behavior with these changes and that we document them so that the users know the dangers of having invalid arguments in scripts.

zas:
> As much as I appreciate these PRs, with all these changes, we have switched from a behavior where we were noisy (but hopefully not crashing) and showing errors in the script to the user, to silently switching things to defaults. Since scripts affect file naming, using defaults might mean we override multiple files to the same name. This might mean potential data loss. I hope we are aware of this subtle change in behavior with these changes and that we document them so that the users know the dangers of having invalid arguments in scripts.

You're right about everything, but on `hopefully not crashing` (we fixed a lot of cases leading to Picard crashes regarding scripting).

Can you suggest which changes you'd make to limit the risk of data loss, while maintaining compatibility (as a draft PR, ideally)?

phw:
> As much as I appreciate these PRs, with all these changes, we have switched from a behavior where we were noisy (but hopefully not crashing) and showing errors in the script to the user, to silently switching things to defaults.

I would agree with you, if it would actually be true. But that's not what is happening here. We are not switching from informing  the user about errors to silent defaults, unless you count crashing Picard as informing the user about errors.

The scripting functions in Picard actually never have been raising runtime errors. All the existing functions have always been returning defaults of some kind in case of error. The only actual errors raised are related to parsing the file. Any other unexpected exception during runtime would crash Picard and was considered a bug.

That also means that getting into a situation where a script raises a `ScriptError` while saving a file are really rare, since the kind of errors checked for are already handled when saving the options. One way to get into such a situation is removing a plugin which provides a function used in scripts.

I am actually not opposed to introduce runtime errors and raise them during e.g. loading of releases or saving files (and showing the action to have failed accordingly). But we have to be aware that this would be the "subtle change in behavior", not the other way around.

If this is what we want to do we should probably introduce a `ScriptRuntimeError` derived from `ScriptError` and throw it here in the error condition. I would agree that it could make sense here, as I am also not sure whether returning `"Invalid Format"` or `""` here is better or makes sense at all.

picard/script.py
rdswift: Updated to return a blank string to be consistent with exception returns from other functions.

rdswift:
> Can we get a test for this? If this is only reproducible on Windows the test can be run only on Windows with a `@unittest.skipUnless(IS_WIN, "windows test")` annotation on the test function.

Tests added.  Actually I added both Windows and non-Windows tests.

rdswift:
I added the `ScriptRuntimeError` class as suggested.  In thinking about it, I agree that it makes more sense to catch the error during the script development if possible, rather than simply returning something unexpected.

picard/script.py
```diff
@@ -1258,8 +1262,13 @@ def func_datetime(parser, format=None):
...
     if not format:
         format = '%Y-%m-%d %H:%M:%S'
-
-    return datetime.datetime.now(tz=local_tz).strftime(format)
+    try:
+        return datetime.datetime.now(tz=local_tz).strftime(format)
+    except ValueError:
+        raise ScriptRuntimeError(
+            "Unsupported format code in $%s at position %i, line %i"
+            % ('datetime', parser._x, parser._y)
```
zas: We could perhaps store concerned function, position and line in the `ScriptRuntimeError', so we can write:
Actual script function name can prolly be obtained from registry.

```python
raise ScriptRuntimeError("message", function, line, position)
```

rdswift: I'll have a look at this.  I think I know what you mean, but I'll need to go through the code to find out how to get the function name from the registry.  I'm not very familiar with that.

Also, should the `ScriptRuntimeError` exception be handled in `ScriptParser` or in the modules calling `ScriptParser` such as `picard.scripttofilename.script_to_filename_with_metadata()` (or even further upstream)?  Also, is there any agreement on how the exception should be handled?  For now, I suspect that it might just cause Picard to crash.  Or should the `ScriptRuntimeError` exception handling be dealt with in separate PRs?

phw:
> I suspect that it might just cause Picard to crash

No, this is exactly what does not happen since it is derived from ScriptError. Code running code is supposed to handle this ( and already does).

rdswift:
> Code running code is supposed to handle this ( and already does).

Okay.  That's good to know.  One thing that I wanted to try here but I'm currently experiencing network issues and can't connect with the MusicBrainz website, is the following in a tagging script (on a Windows system):

```
$set(_datetime_test,$if($gt(%totaltracks%,1),\%-d,))
$set(_current_datetime,$datetime(%_datetime_test))
```

When entered, the script should process `$datetime` with an empty argument, which is valid, allowing the script to be saved.  When a release with 2 or more tracks is loaded, `$datetime` will be processed with "\%-d" as an argument, which should trigger a `ScriptRuntimeError` exception.  That would demonstrate (to me) how it is handled.

picard/script.py
rdswift: Is this sort of what you had in mind, or am I misunderstanding that you wanted the actual `def` function name passed to `ScriptRuntimeError` rather than the script function name?

```
    except ValueError:
        name = 'datetime'
        raise ScriptRuntimeError(
            "Unsupported format code in $%s at position %i, line %i"
            % (name, parser._x, parser._y),
            parser.functions[name].function.__name__,
            parser._x,
            parser._y
        )
```

zas: Yes, and no.
The idea would be to pass the short message: "Unsupported format code" along name, line and position.
And build the final message from it, something like: `"{scriptfuncname:s}:{line:d}:{position:d}: {message:s}".format(message, name, line, position)`
Perhaps we can just use hardcoded string for the script function name (simpler), but ideally it should be obtained automatically (from stack? -> python function, loop up in registry, find corresponding script file name)

About catching those: it is obvious we don't want Picard to crash, but instead display exception message to user, if it happens when the user runs the script, how/when is the question.

@phw what do you think?

phw:
> The idea would be to pass the short message: "Unsupported format code" along name, line and position. And build the final message from it, ...

If we get this working this would be best, yes. Then runtime errors can easily be used.

> About catching those: it is obvious we don't want Picard to crash, but instead display exception message to user, if it happens when the user runs the script, how/when is the question.

Yes, exactly. Status quo is that Picard does not crash on `ScriptError` because it is already handled. This is already the expected way to communicate errors during script execution. So I would say in the scope of this PR we are fine.

But actually better handling these errors and giving feedback to the user is currently really lacking. I think errors in the naming script will simply show the saving error (and log the exception), while errors in the loading scripts will only get logged, but the releases will still get loaded. The reason this is not handled in a nicer way is probably because in the current code it is extremely rare that these errors happen during actual script execution, as the majority of cases is already handled when trying to save the scripts. You have to create a situation where a script cannot be run anymore (easiest is to remove a plugin providing a used function).

So now if we actually raise typical runtime errors we have a paradigm shift, as these errors can happen depending on the data. But I fully agree with @samj1912 that this is for the better. . So we should look into making the error cause more visible.

I think we should change (in a separate PR):

- Errors in loading scripts should make the loading fail. Currently the error would then be shown in the info dialog of the release.
- Errors during saving should in general be shown somehow on the failed file (not only in the log). Maybe do something similar and show the error in the info dialog of the file (or track, if the file is mapped to a track)?

rdswift:
> The idea would be to pass the short message: "Unsupported format code" along name, line and position.  And build the final message from it

Okay, I was wondering if that was what you meant.  It makes sense to me.  Done.

> Perhaps we can just use hardcoded string for the script function name (simpler), but ideally it should be obtained automatically (from stack? -> python function, loop up in registry, find corresponding script file name)

I think I found a simpler way by maintaining a `_function_stack` list in the parser.  The current function name is placed on the stack just before the function is executed, and removed upon successful completion.  That way the name is automatically available within the actual function code as `parser._function_stack[-1]`.

test/test_script.py
```diff
@@ -47,13 +49,17 @@
...
     ScriptParser,
+    ScriptRuntimeError,
     ScriptSyntaxError,
     ScriptUnknownFunction,
     register_script_function,
     script_function,
 )
 
 
+IS_WIN = os.name == 'nt'
```
zas: Is there a specific reason to not use `IS_WIN` from `picard.const.sys`?

test/test_script.py
```diff
@@ -1174,6 +1180,30 @@ def test_cmd_datetime(self):
...
+        original_datetime = datetime.datetime
+        datetime.datetime = _DateTime
+
+        try:
+            context = Metadata()
+            # Tests with Windows-specific format
+            self.assertScriptResultEquals(r"$datetime(\%Y-\%#m-\%#d)", "2020-1-2", context)
+            # Tests with invalid format codes (Windows does not support -d or -m codes)
+            areg = r"^\$datetime:\d+:\d+: Unsupported format code"
+            with self.assertRaisesRegex(ScriptRuntimeError, areg):
```
zas: What about testing `ScriptRuntimeError` on non-win systems?

picard/script.py
```diff
@@ -175,6 +184,9 @@ class ScriptParser(object):
     _function_registry = ExtensionPoint(label='function_registry')
     _cache = {}
 
+    def __init__(self):
+        self._function_stack = []
```
zas: I'm not sure implementing it as a simple list is future-proof, because `list` isn't thread-safe.

There are altenatives that are thread-safe, and could prolly work:

```python
from queue impot LifoQueue

stack = LifoQueue()
stack.put('bar')
stack.put('foo')
...
if not stack.empty():
  last = stack.get()
```

Since we do `stack = stack[:-1]` but on failure where we only read last value, it could make sense to use lifo queues and be thread-safe as bonus.

test/test_script.py
rdswift: No reason other than I forgot there was one.  Fixed now.  Thanks.

test/test_script.py
rdswift: I was going to, but I haven't yet found anything that would cause it to trigger.  Linux and Mac seem to be much more tolerant about the `strftime` format string.  Are you aware of something?

zas: I just tested on linux with Python 3.6.2: when `strftime()` is passed an unsupported format string (like `%E` or `%Q`) it is just returned as is (this behavior differs on Windows, where unsupported format raises exception). So you're right, at least on this implementation, there's no way to trigger this error this way.

I tested all ascii chars from 1 to 255:

```python
for i in range(1, 255):
    print("%s %s" % (chr(i), datetime.datetime.now(tz=local_tz).strftime("%" + chr(i))))
```
None of those leads to ValueError.

**BUT there's one case: if the format string ends with '%'**

```python
>>> datetime.datetime.now(tz=local_tz).strftime("%")
ValueError: strftime format ends with raw %
```

https://docs.python.org/3/library/datetime.html#strftime-and-strptime-behavior

rdswift:
> BUT there's one case: if the format string ends with '%'

PERFECT.  I'll add a test for that on all platforms.

rdswift:
ARGH!  It looks like Python 3.7 and higher isn't raising the ValueError exception on the format string ending with '%'.  Python 3.4 and 3.5 do raise the exception.  Python 3.8 raises the exception here on my Windows system.  Is there a way to limit the test to a certain version of Python?  Until then, I'll comment out that test.

rdswift:
Oops, the last commit message should be 'Disable test because it does not apply to Python > 3.6 on Ubuntu`

test/test_script.py
```diff
@@ -1171,10 +1171,11 @@ def test_cmd_datetime(self):
...
                 self.parser.eval("$datetime(abc,def)")
-            # Tests with invalid format code (ends with % failes on all platforms)
-            areg = r"^\$datetime:\d+:\d+: Unsupported format code"
-            with self.assertRaisesRegex(ScriptRuntimeError, areg):
-                self.parser.eval(r"$datetime(\%)")
+            # The following is disabled because it only raises the exception for Python < 3.7 on Ubuntu
+            # # Tests with invalid format code (ends with % failes on all platforms)
+            # areg = r"^\$datetime:\d+:\d+: Unsupported format code"
+            # with self.assertRaisesRegex(ScriptRuntimeError, areg):
+            #     self.parser.eval(r"$datetime(\%)")
```
zas: For reference:
https://bugs.python.org/issue35066
https://github.com/python/cpython/pull/11550

zas: You could split this to another test and check if ending '%' causes ValueError or not to skip it or not.

Something like (untested):

```python

    def test_cmd_datetime_ending_percent(self):
        try:
            datetime.datetime.now().strftime('%')
        except ValueError:
            pass
        else:
            self.skipTest('datetime module supports trailing %')
        # Save original datetime object and substitute one returning
        # a fixed now() value for testing.
        original_datetime = datetime.datetime
        datetime.datetime = _DateTime

        try:
            context = Metadata()
            context["foo"] = "%Y%m%d%H%M%S.%f"
            # Tests with invalid format code (ends with % failes on all platforms)
            areg = r"^\$datetime:\d+:\d+: Unsupported format code"
            with self.assertRaisesRegex(ScriptRuntimeError, areg):
                self.parser.eval(r"$datetime(\%)")
        except (AssertionError, ValueError, IndexError) as err:
            raise err
        finally:
            # Restore original datetime object
            datetime.datetime = original_datetime
```
```

File: test/test_script.py [hunk 7]
```diff
@@ -1168,6 +1171,37 @@ def test_cmd_datetime(self):
...
+        except (AssertionError, ValueError, IndexError) as err:
+            raise err
+        finally:
+            # Restore original datetime object
+            datetime.datetime = original_datetime
+
+    @skipUnless(IS_WIN, "windows test")
```
zas: Instead of skipping on Windows, we should rather skip if underlying datetime format is not supported on the platform we test on.

Basically: execute `datetime.strftime(whatever)` and if it raises an exception, expect `$datetime(whatever)` raises `ScriptRuntimeError` as well.

zas:
This shows something: using `$datetime()` in a script makes it non-portable between platforms :(

picard/script.py
```diff
@@ -314,6 +327,11 @@ def eval(self, script, context=None, file=None):
             ScriptParser._cache[key] = self.parse(script, True)
         return ScriptParser._cache[key].eval(self)
 
+    def get_current_function(self):
+        if not self._function_stack.empty():
+            return self._function_stack.get()
+        return ''
```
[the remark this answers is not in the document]
rdswift: Good point.  I'll make the change.

test/test_script.py
rdswift: I was thinking about checking in `func_datetime` and then forcing it to raise a `ValueError` but that just wouldn't be right.  I really like your solution, and will see about incorporating it.

test/test_script.py
rdswift: Similar to your other test suggestion above.  I like it.

rdswift:
> This shows something: using `$datetime()` in a script makes it non-portable between platforms :(

Yeah, that is bad.  The only way I can immediately think of that we could force it to be consistent is to check for the offending condition(s) in the function itself and raise a ScriptRuntimeError (or perhaps a new ScriptValueError), but that would probably cause problems in the future when all supported platforms actually allowed the offending argument.

zas:
> Yeah, that is bad. The only way I can immediately think of that we could force it to be consistent is to check for the offending condition(s) in the function itself and raise a ScriptRuntimeError (or perhaps a new ScriptValueError), but that would probably cause problems in the future when all supported platforms actually allowed the offending argument.

Well, it shouldn't be much an issue, but since we plan to ease scripts sharing through script files support, script writes will have to be aware of this, so perhaps making it clear in the function documentation is needed.

test/test_script.py
```diff
@@ -1168,6 +1169,47 @@ def test_cmd_datetime(self):
             areg = r"^Wrong number of arguments for \$datetime: Expected between 0 and 1, "
             with self.assertRaisesRegex(ScriptError, areg):
                 self.parser.eval("$datetime(abc,def)")
+            # The following is disabled because it only raises the exception for Python < 3.7 on Ubuntu
+            # # Tests with invalid format code (ends with % failes on all platforms)
+            # areg = r"^\$datetime:\d+:\d+: Unsupported format code"
+            # with self.assertRaisesRegex(ScriptRuntimeError, areg):
+            #     self.parser.eval(r"$datetime(\%)")
```
zas: @rdswift you can now remove this comment I guess

There's a typo in top comment "origninal"

rdswift:
> Well, it shouldn't be much an issue, but since we plan to ease scripts sharing through script files support, script writers will have to be aware of this, so perhaps making it clear in the function documentation is needed.

So we should document the formatting arguments that should be avoided because they are not supported across all platforms.  Will do.  Should we also check for them in the function code to enforce that they are not being used?

test/test_script.py
```diff
@@ -1168,6 +1169,47 @@ def test_cmd_datetime(self):
...
+            ('%-m', r'$datetime(\%-m)'),    # Non zero-padded month
+            ('%3Y', r'$datetime(\%3Y)'),    # Length specifier shorter than string
+        ]
+        tests_to_run = []
+        # Get list of tests for unsupported format codes
+        for (test_arg, test_case) in possible_tests:
+            try:
+                datetime.datetime.now().strftime(test_arg)
+            except ValueError:
+                tests_to_run.append(test_case)
```
zas: `test_case` can be built from `test_arg`, so I think we can simplify `possible_tests` (and make it immutable while at it):

```python
possible_tests = (
  '%',  # Hanging % at end of format
...

tests_to_run.append(r"$datetime(\%s)" % test_arg)
```

zas:
> Should we also check for them in the function code to enforce that they are not being used?

I don't think so, it would make things more complicated for nothing.

test/test_script.py
rdswift: Good eye.  I'll have to look into a spell checker for VS Code. :wink:

test/test_script.py
rdswift: I'll make the change.  I thought about that originally, but thought it might be an issue if the argument didn't need to start with an escape character.  As it turns out, everything we're (currently) testing is escaped so it's not a problem.

test/test_script.py
phw: @rdswift I use the [Code Spell Checker](https://marketplace.visualstudio.com/items?itemName=streetsidesoftware.code-spell-checker) extension for vscode, which I think is pretty nice. It does not only check strings and comments but also variable names. So if you have e.g. `metedata = Metadata()` it would highlight that typo.

rdswift:
> I don't think so, it would make things more complicated for nothing.

Whew.  Thanks for confirming.

When I started on this, I had no idea that what seemed like such a simple scripting function would become so complicated to implement.  Platform inconsistencies are such a PITA.

test/test_script.py
rdswift: Just installed it.  It is a sweet add-on.  Thanks.

picard/script.py
```diff
@@ -73,6 +74,11 @@ class ScriptUnknownFunction(ScriptError):
     pass
 
 
+class ScriptRuntimeError(ScriptError):
+    def __init__(self, message, name, line, position):
+        super().__init__("${1:s}:{2:d}:{3:d}: {0:s}".format(message, name, line, position))
```
zas: `line` & `position` parameters are in the wrong order.

https://github.com/metabrainz/picard/pull/1508/files#diff-49991922d5e1bb94a02cf28bf4619a1eR1280-R1281

test/test_script.py
```diff
@@ -1174,6 +1175,42 @@ def test_cmd_datetime(self):
...
+
+        try:
+            context = Metadata()
+            areg = r"^\$datetime:1:\d+: Unsupported format code"
+            # Tests with invalid format code (platform dependent tests)
+            for test_case in tests_to_run:
+                with self.assertRaisesRegex(ScriptRuntimeError, areg):
+                    self.parser.eval(r'$datetime(\{0})'.format(test_case))
+        except (AssertionError, ValueError, IndexError) as err:
+            raise err
```
zas: @phw @rdswift I'm not sure this `except ... as err` / `raise err` is needed. `try` / `finally` might be enough

rdswift: Of course it's not necessary.  In fact, it's dumb the way I wrote it.  What was I thinking?  :blush:

rdswift:
Sorry.  A quick addition to the tests to confirm that the function stack is returning the correct name (by nesting functions in the test) and to confirm that the proper line and position numbers are passed.

PR #1509, opened 2020-02-28
PICARD-1796: Date scoring logic

RaysDev:
#  Summary
New logic to include the release date in scoring process.

(Separated out from PR 1771 as you requested)

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Seems some of the results from AcoustId don't have a release date, and we need a way to de-prioritize those records with lower quality data from the match process.

PICARD-1796

# Solution

Enhanced the scoring process to consider the release date.  It will give highest Part Score to results that have a date that match our own tag, next to records that have a date where year is +/-2 years, next to records that have a non-blank date, next a minimal score if the date is missing, and the lowest Part Score if the date is there but does not match.

# Action
Sorry need little guidance on creating the JIRA you asked to have associated with this.

Please ignore the _config.yml for theme.  Very bizzare that showed up when I first forked, and I definitely did not commit it???

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -260,7 +260,7 @@ class File(QtCore.QObject, Item):
             # https://docs.python.org/3/library/os.html#os.utime
             # Since Python 3.3, ns parameter is available
             # The best way to preserve exact times is to use the st_atime_ns and st_mtime_ns
-            #  fields from the os.stat() result object with the ns parameter to utime.
+            # fields from the os.stat() result object with the ns parameter to utime.
             st = os.stat(filename)
         except OSError as why:
             errmsg = "Couldn't read timestamps from %r: %s" % (filename, why)
@@ -489,7 +489,7 @@ class File(QtCore.QObject, Item):
         new_path = os.path.dirname(new_filename)
         old_path = os.path.dirname(old_filename)
         if new_path == old_path:
-            #  skip, same directory, nothing to move
+            # skip, same directory, nothing to move
             return
         patterns = config.setting["move_additional_files_pattern"]
         pattern_regexes = set()
--- a/picard/file.py
+++ b/picard/file.py
@@ -23,6 +23,7 @@
 # Copyright (C) 2016-2018 Sambhav Kothari
 # Copyright (C) 2017-2018 Antonio Larrosa
 # Copyright (C) 2019 Joel Lintunen
+# Copyright (C) 2020 Ray Bouchard
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -111,6 +112,7 @@ class File(QtCore.QObject, Item):
         "releasecountry": 2,
         "format": 2,
         "isvideo": 2,
+        "date": 4,
     }
 
     class PreserveTimesStatError(Exception):
@@ -258,7 +260,7 @@ class File(QtCore.QObject, Item):
             # https://docs.python.org/3/library/os.html#os.utime
             # Since Python 3.3, ns parameter is available
             # The best way to preserve exact times is to use the st_atime_ns and st_mtime_ns
-            # fields from the os.stat() result object with the ns parameter to utime.
+            #  fields from the os.stat() result object with the ns parameter to utime.
             st = os.stat(filename)
         except OSError as why:
             errmsg = "Couldn't read timestamps from %r: %s" % (filename, why)
@@ -487,7 +489,7 @@ class File(QtCore.QObject, Item):
         new_path = os.path.dirname(new_filename)
         old_path = os.path.dirname(old_filename)
         if new_path == old_path:
-            # skip, same directory, nothing to move
+            #  skip, same directory, nothing to move
             return
         patterns = config.setting["move_additional_files_pattern"]
         pattern_regexes = set()
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -37,7 +37,8 @@ from collections.abc import (
     Iterable,
     MutableMapping,
 )
-import re
+
+from dateutil.parser import parse
 
 from PyQt5.QtCore import QObject
 
@@ -228,24 +229,15 @@ class Metadata(MutableMapping):
 
     @staticmethod
     def extract_year_from_date(dt):
-        """ Tries to handle various date formats, not exhaustive but common, and extract the year"""
+        """ Extracts year from  passed in date either dict or string """
 
-        if isinstance(dt, dict):
-            year = dt.get('year')
-        elif re.match(r"^[0-9]{8}$", dt):  # 20201231
-            year = dt[0:4]
-        elif re.match(r"^[0-9]{6}$", dt):  # 201231
-            year = dt[0:4]
-        elif re.match(r"^[0-9]{4}$", dt):  # 2012
-            year = dt[0:4]
-        elif re.match(r"^[0-9]{4}[-/]", dt):  # 2020-12-31 or 2020/12/31
-            year = dt[0:4]
-        elif re.match(r"^[0-9]{2}/[0-9]{2}/[0-9]{4}", dt):  # 12/31/2020
-            year = dt[6:10]
+        if isinstance(dt, Mapping):
+            year = int(dt.get('year'))
         else:
-            year = "0"
+            parsed_dt = parse(dt)
+            year = parsed_dt.year
 
-        return int(year)
+        return year
 
     def compare_to_release_parts(self, release, weights):
         parts = []
@@ -279,15 +271,16 @@ class Metadata(MutableMapping):
             release_year = Metadata.extract_year_from_date(release_date)
             if "date" in self:
                 metadata_date = self['date']
-                metadata_year = Metadata.extract_year_from_date(metadata_date)
                 if release_date == metadata_date:
                     factor = 1.00
-                elif release_year == metadata_year:
-                    factor = 0.95
-                elif abs(release_year - metadata_year) <= 2:
-                    factor = 0.85
                 else:
-                    factor = 0.00
+                    metadata_year = Metadata.extract_year_from_date(metadata_date)
+                    if release_year == metadata_year:
+                        factor = 0.95
+                    elif abs(release_year - metadata_year) <= 2:
+                        factor = 0.85
+                    else:
+                        factor = 0.00
             else:
                 factor = 0.65
         else:
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -38,8 +38,6 @@ from collections.abc import (
     MutableMapping,
 )
 
-from dateutil.parser import parse
-
 from PyQt5.QtCore import QObject
 
 from picard import config
@@ -52,7 +50,10 @@ from picard.plugin import (
     PluginPriority,
 )
 from picard.similarity import similarity2
-from picard.util import linear_combination_of_weights
+from picard.util import (
+    extract_year_from_date,
+    linear_combination_of_weights,
+)
 from picard.util.imagelist import ImageList
 from picard.util.tags import PRESERVED_TAGS
 
@@ -227,18 +228,6 @@ class Metadata(MutableMapping):
         sim = linear_combination_of_weights(parts) * get_score(release)
         return SimMatchRelease(similarity=sim, release=release)
 
-    @staticmethod
-    def extract_year_from_date(dt):
-        """ Extracts year from  passed in date either dict or string """
-
-        if isinstance(dt, Mapping):
-            year = int(dt.get('year'))
-        else:
-            parsed_dt = parse(dt)
-            year = parsed_dt.year
-
-        return year
-
     def compare_to_release_parts(self, release, weights):
         parts = []
         if "album" in self:
@@ -268,13 +257,13 @@ class Metadata(MutableMapping):
         factor = 0.0
         if "date" in release:
             release_date = release['date']
-            release_year = Metadata.extract_year_from_date(release_date)
+            release_year = extract_year_from_date(release_date)
             if "date" in self:
                 metadata_date = self['date']
                 if release_date == metadata_date:
                     factor = 1.00
                 else:
-                    metadata_year = Metadata.extract_year_from_date(metadata_date)
+                    metadata_year = extract_year_from_date(metadata_date)
                     if release_year == metadata_year:
                         factor = 0.95
                     elif abs(release_year - metadata_year) <= 2:
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -37,6 +37,7 @@ from collections.abc import (
     Iterable,
     MutableMapping,
 )
+import re
 
 from PyQt5.QtCore import QObject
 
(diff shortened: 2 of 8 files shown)


picard/metadata.py
```diff
@@ -199,6 +200,27 @@ def compare_to_release(self, release, weights):
...
+            year = dt[0:4]
+        elif re.match(r"^[0-9]{4}[-/]", dt):  # 2020-12-31 or 2020/12/31
+            year = dt[0:4]
+        elif re.match(r"^[0-9]{2}/[0-9]{2}/[0-9]{4}", dt):  # 12/31/2020
+            year = dt[6:10]
+        else:
+            year = "0"
+
+        return int(year)
+
```
zas: We can do something like:

```python
from collections import Mapping
import re

YEAR_REGEX_YMD = re.compile(r'^(?P<year>\d{4})[-/]?(?:\d{2}[-/]?)?(?:\d{2}[-/]?)?$')
YEAR_REGEX_DMY = re.compile(r'^(?:\d{2}[-/])?(?:\d{2}[-/])?(?P<year>\d{4})$')


def extract_year_from_date(dt):
    """ Tries to handle various date formats, not exhaustive but common, and extract the year"""

    if isinstance(dt, Mapping):
        dt = dt.get('year', '')

    match = YEAR_REGEX_DMY.match(dt)
    if match:
        return int(match.group('year'))

    match = YEAR_REGEX_YMD.match(dt)
    if match:
        return int(match.group('year'))

    return 0


cases = (
    '20201231',
    '202031',
    '20203',
    '2020',
    '2020-12-31',
    '2020/12/31',
    '12/31/2020',
    '12-31-2020',
    '12-2020',
    '01-01-2020',
    'x',
    {'year': '2020'},
    {'year': 'x'},
    {'year': '2020/12/31'}
)

for dt in cases:
    print("%s -> %d" % (dt, extract_year_from_date(dt)))
```


Output:

```
20201231 -> 2020
202031 -> 2020
20203 -> 0
2020 -> 2020
2020-12-31 -> 2020
2020/12/31 -> 2020
12/31/2020 -> 2020
12-31-2020 -> 2020
12-2020 -> 2020
01-01-2020 -> 2020
x -> 0
{'year': '2020'} -> 2020
{'year': 'x'} -> 0
{'year': '2020/12/31'} -> 2020
```


Also, for such function, it is mandatory to write test code (in `test/`).
Also, this function should go in `picard/util` as it's a static method, not specific to `Metadata`.

_config.yml
```diff
@@ -0,0 +1 @@
+theme: jekyll-theme-midnight
```
zas: What's this file?

picard/file.py
```diff
@@ -229,7 +230,7 @@ def _preserve_times(self, filename, func):
             # https://docs.python.org/3/library/os.html#os.utime
             # Since Python 3.3, ns parameter is available
             # The best way to preserve exact times is to use the st_atime_ns and st_mtime_ns
-            # fields from the os.stat() result object with the ns parameter to utime.
+            #  fields from the os.stat() result object with the ns parameter to utime.
```
zas: This change is unrelated.

picard/file.py
```diff
@@ -458,7 +459,7 @@ def _move_additional_files(self, old_filename, new_filename):
         new_path = os.path.dirname(new_filename)
         old_path = os.path.dirname(old_filename)
         if new_path == old_path:
-            # skip, same directory, nothing to move
+            #  skip, same directory, nothing to move
```
zas: Unrelated change.

picard/metadata.py
```diff
@@ -218,6 +240,34 @@ def compare_to_release_parts(self, release, weights):
...
+        # release has a date but it does not match ours(all else equal, its better to have an unknown date than a wrong date. since the unknown could actually be correct)
+        factor = 0.0
+        if "date" in release:
+            release_date = release['date']
+            release_year = Metadata.extract_year_from_date(release_date)
+            if "date" in self:
+                metadata_date = self['date']
+                metadata_year = Metadata.extract_year_from_date(metadata_date)
+                if release_date == metadata_date:
+                    factor = 1.00
```
zas: You don't need to call `extract_year_from_date()` twice in this case

picard/metadata.py
```diff
@@ -199,6 +200,27 @@ def compare_to_release(self, release, weights):
         sim = linear_combination_of_weights(parts) * get_score(release)
         return SimMatchRelease(similarity=sim, release=release)
 
+    @staticmethod
+    def extract_year_from_date(dt):
```
samj1912: You can just use dateutil parser and then get the year.

zas: Oh, this too ;)

samj1912: https://dateutil.readthedocs.io/en/stable/parser.html#module-dateutil.parser

I would much rather have this (which is thoroughly tested and supported) than trying to implement a bunch of regexes.

picard/metadata.py
phw: Agreed, especially putting this to util. Also some existing code in Picard assumes dates have the format `yyyy-mm-dd` and year can be extracted as the first four characters. Having this logic in utils allows us to review the cases and reuse the year extraction.

_config.yml
RaysDev: Let me share in case you run into this again.  I was really stumped on this because I would remember creating a file and committing it, but I didn't.  Apparently while looking for Dark Mode on github.com, I ran into theme selector for "GitHub Pages".   There are icons of various themes, and a button that says "Select Theme."  This button actually creates the _config.yml file in your repo and commits it, without such indication.  Anyway I'll remove the file, but I thought it might save you some trouble in future.

picard/file.py
RaysDev: ugh, I turned on some of the auto formatting in my editor, and it got a little over zealous.  Same for next few.

picard/metadata.py
RaysDev: Done, but I'm surprised we prefer to add a dependency on a third party library for such a small use.

zas: > Done, but I'm surprised we prefer to add a dependency on a third party library for such a small use.

It was a suggestion, and I see nothing wrong in using regexes (regexes engines are well-tested too).
But Python has tons of good modules available, and it is good practice to re-use existing stuff whenever possible, but I'm reluctant to rely on third-party libs for another reason: **can anyone there confirm no licensing conflict?**

phw: > **can anyone there confirm no licensing conflict?**

We can use this under the https://opensource.org/licenses/BSD-3-Clause, which we can use in our GPL code.

phw: I must say I'm also a bit concerned about adding another dependency just for this, so far we had managed to keep the external dependencies to a minimum for Picard. On the other hand we really should review other uses of date handling in our code, I think Picard currently relies on the assumption date is always `YYYY-MM-DD` or any shorter form of this (without day or without month and day). But this might not be true. So having a stable library for date handling might make sense. Two things though:

- `extract_year_from_date` probably should be moved to utils
- `dateutil` must be added to requirements*.txt

RaysDev:
ok, I moved extract_year_from_date() to utils and updated requirements.txt

picard/file.py
```diff
@@ -258,7 +260,7 @@ def _preserve_times(self, filename, func):
             # https://docs.python.org/3/library/os.html#os.utime
             # Since Python 3.3, ns parameter is available
             # The best way to preserve exact times is to use the st_atime_ns and st_mtime_ns
-            # fields from the os.stat() result object with the ns parameter to utime.
+            # fields from the os.stat() result object with the ns parameter to utime.
```
phw: If someone else is wondering: This is changing from a non-breaking space to a regular space. It's strictly speaking an unrelated change, but I don't mind including this minor cleanup in a commit like this. But if someone has a strong opinion on this we could move it to a separate commit.

phw:
@samj1912 @zas This is looking good to me, but I would like to get your final review on this. Especially your opinion on the added dependency.

picard/metadata.py
```diff
@@ -244,6 +247,35 @@ def compare_to_release_parts(self, release, weights):
         except (ValueError, KeyError):
             pass
 
+        # Date Logic - in order of preference
```
samj1912: It would be nice to move these comments near the actual else cases as it allows us to understand what's happening.

picard/metadata.py
```diff
@@ -244,6 +247,35 @@ def compare_to_release_parts(self, release, weights):
...
+            release_date = release['date']
+            release_year = extract_year_from_date(release_date)
+            if "date" in self:
+                metadata_date = self['date']
+                if release_date == metadata_date:
+                    factor = 1.00
+                else:
+                    metadata_year = extract_year_from_date(metadata_date)
+                    if release_year == metadata_year:
+                        factor = 0.95
```
samj1912: Brownie points if you make these constants with understandable names!

picard/util/__init__.py
```diff
@@ -645,3 +649,15 @@ def limited_join(a_list, limit, join_string='+', middle_string='…'):
     start = a_list[:half]
     end = a_list[-half:]
     return join_string.join(start + [middle_string] + end)
+
+
+def extract_year_from_date(dt):
+    """ Extracts year from  passed in date either dict or string """
+
+    if isinstance(dt, Mapping):
```
samj1912: I would check for str here instead of mapping. Then return the parsed year. Makes the intention clear.

picard/metadata.py
phw: @RaysDev  For the requirements we actually need to update not only the requirements.txt but also requirements-win.txt and requirements-macos.txt. We use the later ones to freeze the dependencies for those operating systems as we often encounter issues with specific versions. But for general development we use the requirements.txt as we try to go with newer versions whenever possible.

zas:
> @samj1912 @zas This is looking good to me, but I would like to get your final review on this. Especially your opinion on the added dependency.

Well, I'm fine with extra dependency, if we find out it causes any problem we'll just find an alternative.
For the rest, it looks good to me, but I'd like few minor comments be addressed before merging.

RaysDev:
> For the rest, it looks good to me, but I'd like few minor comments be addressed before merging.

Sorry, I'm not clear at this point which items you would like addressed.  is it the moving/spreading of the comments in metadata.py line 250 "# Date Logic - in order of preference"   ?

zas:
> > For the rest, it looks good to me, but I'd like few minor comments be addressed before merging.
> 
> @zas Sorry, I'm not clear at this point which items you would like addressed. is it the moving/spreading of the comments in metadata.py line 250 "# Date Logic - in order of preference" ?

https://github.com/metabrainz/picard/pull/1509#discussion_r394206521
https://github.com/metabrainz/picard/pull/1509#discussion_r394206258
https://github.com/metabrainz/picard/pull/1509/files#r385592761

picard/metadata.py
RaysDev: done

picard/metadata.py
RaysDev: done

picard/util/__init__.py
RaysDev: In previous review above I was asked to change it to check for Mapping.

RaysDev:
All requested changes are completed.

zas:
> All requested changes are completed.

Can you please just squash last two commits (b9868b6 f962c46) in one? It'd be cleaner.

picard/metadata.py
```diff
@@ -153,6 +156,15 @@ class Metadata(MutableMapping):
         ('totaldiscs', 4),
     ]
 
+    __date_match_factors = dict([
```
samj1912: Please use normal dict declaration - 

```
VALUES = {
    "key": "value",
    ...
}

**Comment by RaysDev** [comment]
squash completed as requested

File: picard/metadata.py (hunk 11 above)
Comment (RaysDev) [reply picard/metadata.py]: I made the change as requested, and I have no concern with this,  but what I don't understand, is that just one line above, the pre-existing syntax used for __weights uses the syntax I had.  so now just one line apart we have two ways of doing the same thing.  I would think consistency would be more important here.

Comment (zas) [reply picard/metadata.py]: > the pre-existing syntax used for __weights

`__weights` is a list while `__date_match_factors` is a dict.

That said, the syntax you used is perfectly acceptable, but you create a dictionnary by passing a list of tuples to constructor. Using dict literal syntax `{}` is known as being faster (even though the time difference was significantly reduced with Python3 and even more with recent 3.x versions).

See: https://stackoverflow.com/questions/6610606/is-there-a-difference-between-using-a-dict-literal-and-a-dict-constructor

File: picard/metadata.py [hunk 12]
```diff
@@ -244,6 +256,35 @@ def compare_to_release_parts(self, release, weights):
...
+                    elif abs(release_year - metadata_year) <= 2:
+                        date_match_factor = self.__date_match_factors['close_year']
+                    # release has a date but it does not match ours(all else equal, its better to have an unknown date than a wrong date. since the unknown could actually be correct)
+                    else:
+                        date_match_factor = self.__date_match_factors['differed']
+            # release has a date but we don't have one (all else equal, we prefer tracks that have non-blank date values)
+            else:
+                date_match_factor = self.__date_match_factors['exists_vs_null']
+        # release has a no date (all else equal, we don't prefer this release since its date is missing)
+        else:
```
zas: I just noticed you placed comments outside blocks they belong to.

I think  it is much better to keep the comment in the code block (and wrap long comments):

```python
       # Date Logic
        date_match_factor = 0.0
        if "date" in release and release['date'] != '':
            release_date = release['date']
            release_year = extract_year_from_date(release_date)
            if "date" in self:
                metadata_date = self['date']
                if release_date == metadata_date:
                    # release has a date and it matches what our metadata had exactly.
                    date_match_factor = self.__date_match_factors['exact']
                else:
                    metadata_year = extract_year_from_date(metadata_date)
                    if release_year == metadata_year:
                        # release has a date and it matches what our metadata had for year exactly.
                        date_match_factor = self.__date_match_factors['year']
                    elif abs(release_year - metadata_year) <= 2:
                        # release has a date and it matches what our metadata had closely (year +/- 2).
                        date_match_factor = self.__date_match_factors['close_year']
                    else:
                        # release has a date but it does not match ours (all else equal,
                        # its better to have an unknown date than a wrong date, since
                        # the unknown could actually be correct)
                        date_match_factor = self.__date_match_factors['differed']
            else:
                # release has a date but we don't have one (all else equal, we prefer
                # tracks that have non-blank date values)
                date_match_factor = self.__date_match_factors['exists_vs_null']
        else:
            # release has a no date (all else equal, we don't prefer this
            # release since its date is missing)
            date_match_factor = self.__date_match_factors['no_release_date']
```

@samj1912 what do you think?

RaysDev: done and committed.  thanks for typing it out to show me what you meant.

picard/metadata.py
RaysDev: I can see the new syntax is better, I just wasn't getting the difference from __weights.  but in tracing through the code, I can see weights is used differently in this class.  Even differently than the `comparison_weights` are used in file.py. which actually is using the same syntax you are suggesting here :)

picard/metadata.py
zas: Thanks.

Perhaps you can just squash last 4 commits

RaysDev: sure, I did the squash

phw:
I added PICARD-1796 to track this. To me this looks good, recent changes improved readability.

PR #1510, opened 2020-02-28
Dialog superclass

RaysDev:
# Summary
This creates a superclass with the common code pulled out of SerachDialog so that I can use it in Match Dialog, per your suggestion in PR-1497

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem
PR 1497 created new dialog box which resulted in duplicated code with SearchDialog 

# Solution
Refactored out the duplicated common code to a new class that can be shared. and changed SearchDialog to use it.

### DIFF
--- a/picard/ui/tablebaseddialog.py
+++ b/picard/ui/tablebaseddialog.py
@@ -1,0 +1,200 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+# Copyright (C) 2020 Ray Bouchard
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""
+This can be used for basic dialogs that mostly contain a table as its core feature
+"""
+
+from abc import abstractmethod
+from collections import OrderedDict
+
+from PyQt5 import (
+    QtCore,
+    QtWidgets,
+)
+from PyQt5.QtCore import pyqtSignal
+
+from picard import (
+    config,
+    log,
+)
+from picard.util import (
+    natsort,
+    restore_method,
+    throttle,
+)
+
+from picard.ui import PicardDialog
+
+
+class ResultTable(QtWidgets.QTableWidget):
+
+    def __init__(self, parent):
+        super().__init__(parent)
+        self.setSelectionMode(QtWidgets.QAbstractItemView.ExtendedSelection)
+        self.setSelectionBehavior(QtWidgets.QAbstractItemView.SelectRows)
+        self.setEditTriggers(QtWidgets.QAbstractItemView.NoEditTriggers)
+        self.horizontalHeader().setStretchLastSection(True)
+        self.horizontalHeader().setSectionResizeMode(QtWidgets.QHeaderView.Stretch)
+        self.horizontalHeader().setSectionResizeMode(QtWidgets.QHeaderView.Interactive)
+
+        @throttle(1000)  # only emit scrolled signal once per second
+        def emit_scrolled(x):
+            parent.scrolled.emit()
+        self.horizontalScrollBar().valueChanged.connect(emit_scrolled)
+        self.verticalScrollBar().valueChanged.connect(emit_scrolled)
+        self.setHorizontalScrollMode(QtWidgets.QAbstractItemView.ScrollPerPixel)
+
+    def prepare(self, headers):
+        self.clear()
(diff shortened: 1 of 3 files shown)


phw:
I think it is missing the actual tablebaseddialog.py ;)

Also please rebase against current master to solve the merge conflict and don't commit the `_config.yml `. I guess it is a config specific to the editor you use, but it looks like a specific user setting as well. So please don't commit it, but you can add it to the .gitignore instead.

picard/ui/tablebaseddialog.py
```diff
@@ -0,0 +1,200 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+# Copyright (C) 2020 Ray Bouchard
```
phw: This should have the copyrights from the original copied over, since a lot of this code got just copied as well.

picard/ui/tablebaseddialog.py
```diff
@@ -0,0 +1,200 @@
...
+            self.accept_button.setEnabled(True)
+        self.table.itemSelectionChanged.connect(enable_accept_button)
+
+    def prepare_table(self):
+        self.table.prepare(self.table_headers)
+        self.restore_table_header_state()
+
+    @abstractmethod
+    def show_table(self, sort_column=None, sort_order=QtCore.Qt.DescendingOrder):
+        pass
```
phw: I wonder if we could set the current implementation as default. It looks generic enough to me, and if it does not fit a subclass can still overwrite it. What do you think?

picard/ui/tablebaseddialog.py
```diff
@@ -0,0 +1,200 @@
...
+
+    @abstractmethod
+    def accept(self):
+        super().accept()
```
phw: This is not really needed to implement here if it just calls `super().accept()`, since `QDialog` already provides this. So this implementation currently does not provide any value. But again I wonder whether we should include the current default implementation from searchdialog, e.g.:

```python
def accept(self):
    if self.table:
        selected_rows = []
        for idx in self.table.selectionModel().selectedRows():
            row = self.table.itemFromIndex(idx).data(QtCore.Qt.UserRole)
            selected_rows.append(row)
        self.accept_event(selected_rows)
    super().accept()

@abstractmethod
def accept_event():
    pass
```

This looks useful for me for a table based result dialog and your new match details dialog could then also just implement `accept_event`

samj1912:
Not a big fan of sharing code via inheritance. Especially for UI components which are easily composable. I would like you to reevaluate your decision after reading the pros and cons of each. (Do a simple Qt composition vs inheritance search) if the pragmatic conclusion after being aware of the pros and cons of each is to go forward with inheritance, I'm fine with the decision. But inheritance for code sharing promotes white box code re-use, which for a large code base like Picard just leads to a load of pain.

@phw @zas please weigh in on your thoughts as well.

zas:
> Not a big fan of sharing code via inheritance. Especially for UI components which are easily composable. I would like you to reevaluate your decision after reading the pros and cons of each. (Do a simple Qt composition vs inheritance search) if the pragmatic conclusion after being aware of the pros and cons of each is to go forward with inheritance, I'm fine with the decision. But inheritance for code sharing promotes white box code re-use, which for a large code base like Picard just leads to a load of pain.
> 
> @phw @zas please weigh in on your thoughts as well.

I'm not even sure what you suggest.
Can you give an example of what you would do in a PR?

I have no objection on anything if it improves long-term maintainability.

phw:
I have to agree with zas, I'm not sure what is really suggested for this case. Can you be more concrete on how you would use composition instead of inheritance for this use case?

We basically want to have the search result dialogs, that will be reused in slightly different contexts. The dialog always shows a table of result data, allows the selection of one or multiple results and closes and performs an action when triggering the accept button. The overall behavior and layout should be identical. Roughly something like this:

picard-search-result

For concrete implementations the actual content, the exact table layout and the action to perform will differ. Also additional widgets might be shown, such as in case of the actual search dialogs showing the search text input, or additional buttons.

picard/ui/tablebaseddialog.py
RaysDev: I like that idea.  I was struggling because the order of operations is important inside this method. so super() doesn't work well, but override does.

picard/ui/tablebaseddialog.py
RaysDev: the super.accept() was required in the superclass, otherwise the dialog did not close.

I made the change, but in the overridden MatchDetailsDialog.accept() I had to call QDialog.accept(self) to get the dialog to close.

RaysDev:
> Also please rebase against current master to solve the merge conflict 

Could you please give me a few tips on how to do this.

picard/ui/tablebaseddialog.py
phw: That's strange, should not be needed. But I think the proper way would be to not reimplement `accept` in `MatchDetailsDialog` at all, but only `accept_event` and have it something like this:

```
    def accept_event(self, rows):
        if not rows:
            return
        selected_entry = rows[0]
        match_details_log.set_file_result_selected_entry(self.current_file.filename, selected_entry)
        self.load_selection(selected_entry)
```

That way the base class is generic and you only implement the special handling needed for `MatchDetailsDialog`. Yes, the generic implementation handles multiple rows, but it is easy to handle only a single selection in the subclass so this is fine.

RaysDev: Done.  good idea!

samj1912:
@phw / @zas sorry I misunderstood something. Carry on!

PR #1511, opened 2020-03-02
PICARD-1781: Have $find return '' rather than '-1' on not found

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Have `$find` return "" rather than "-1" on not found

# Problem

* JIRA ticket: PICARD-1781

The `$find` scripting function currently returns "-1" if `needle` is not found in `haystack`.  This was based on the return value of the Python `find` method.  This is not consistent with other Picard scripting functions, which return "" for a `False` (not found?) condition.  For consistency, consider changing the `$find` function to return "" on a "not found" condition.

# Solution

Have $find return "" rather than "-1" on not found.

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -1045,7 +1045,7 @@ def func_find(parser, haystack, needle):
         needle: The substring to find.
 
     Returns:
-        The zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.
+        The zero-based index of the first occurrence of needle in haystack, or "" if needle was not found.
     """
     index = haystack.find(needle)
     if index < 0:
--- a/picard/script.py
+++ b/picard/script.py
@@ -1047,7 +1047,10 @@ def func_find(parser, haystack, needle):
     Returns:
         The zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.
     """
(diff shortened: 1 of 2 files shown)


picard/script.py
```diff
@@ -1047,7 +1047,10 @@ def func_find(parser, haystack, needle):
     Returns:
         The zero-based index of the first occurrance of needle in haystack, or -1 if needle was not found.
```
zas: s/occurance/occurence/?

phw: The docstring needs to be updated to reflect the change

rdswift:
Will the next release be 2.3.2 so I can make a note in the docs about the response change taking effect at a certain version?  Otherwise, I'll make a note about it being "-1" up to Picard 2.3.1 and "" after.  Will there be a target branch set up in picard-website to capture the changes in documentation until the new version is released?  Thanks.

phw:
@rdswift I guess we'll have a 2.3.2 and afterwards concentrate on the 2.4. So I have setup a https://github.com/metabrainz/picard-website/tree/picard-2.3.2 .

PR #1512, opened 2020-03-03
PICARD-1783: Fix deletion of comment, performer and lyrics tags

phw:
# Summary
This adds tests for deleting comment, performer and lyrics tags and fixes various issues with this in ID3, MP4, Vorbis and APEv2.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Adding tests for saving the tags which can have optional description (lyrcis:, performer:, comment:) reveals various issues with deleting those tags.

* JIRA ticket (_optional_): PICARD-1783

# Solution

# Action

### DIFF
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -169,7 +169,7 @@ class APEv2File(File):
                         if len(disc) > 1:
                             metadata["totaldiscs"] = disc[1]
                             value = disc[0]
-                    elif name == 'performer' or name == 'comment':
+                    elif name in ('performer', 'comment'):
                         if value.endswith(')'):
                             start = value.rfind(' (')
                             if start > 0:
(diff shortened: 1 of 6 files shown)


picard/formats/apev2.py
```diff
@@ -236,14 +236,17 @@ def _remove_deleted_tags(self, metadata, tags):
...
-                tag_type = re.compile(r"\(%s\)$" % tag.split(':', 1)[1])
-                existing_tags = tags.get(real_name)
-                if existing_tags:
-                    for item in existing_tags:
-                        if tag_type.search(item):
-                            tags.get(real_name).remove(item)
+            if real_name in ('Lyrics', 'Comment', 'Performer'):
+                parts = tag.split(':', 1)
+                if len(parts) == 2:
+                    tag_type_regex = r"\(%s\)$" % parts[1]
```
zas: It is safer to to [escape](https://docs.python.org/3/library/re.html#re.escape) `parts[1]` imho.

picard/formats/vorbis.py
```diff
@@ -316,10 +316,16 @@ def _remove_deleted_tags(self, metadata, tags):
...
+                    parts = tag.split(':', 1)
+                    if len(parts) == 2:
+                        tag_type_regex = r"\(%s\)$" % parts[1]
+                    else:
+                        tag_type_regex = r"[^)]$"
+                    existing_tags = tags.get(real_name)
+                    for item in existing_tags:
+                        if re.search(tag_type_regex, item):
+                            existing_tags.remove(item)
+                    tags[real_name] = existing_tags
```
zas: The only difference I see between this code and the one in apev2.py is `tags.get(real_name)` and `tags.get(real_name, [])`.
Is this intentional? Can we reduce code redundancy somehow?

test/formats/common.py
```diff
@@ -330,14 +331,51 @@ def test_delete_simple_tags(self):
...
+                del metadata[key]
+                new_metadata = save_and_load_metadata(self.filename, metadata)
+                self.assertNotIn(name, new_metadata)
+                # Ensure the names with description did not get deleted
+                if name_with_description in original_metadata:
+                    self.assertIn(name_with_description, new_metadata)
+
+        @skipUnlessTestfile
+        def test_delete_tags_with_description(self):
+            for key in ('comment:foo', 'comment:de:foo', 'performer:foo', 'lyrics:foo'):
```
zas: I would add few more tests with descriptions containing unexpected characters or the like, to ensure we have a good behavior in unexpected cases too.

picard/formats/apev2.py
phw: Indeed, that was another bug here. I fixed this and added tests

picard/formats/vorbis.py
phw: Yes, this is intentional. I guess we could also write `tags.get(real_name, [])` for Vorbis, but it is not needed there.

We could somehow move those 10 lines into a separate function, but this is really one of those cases where I would keep the code separate. The tag handling is very format specific. The interfaces are similar, but not identical and I am not fully convinced that this is generic enough that we can safely assume the code will be identical for all cases.

zas: Ok, thanks for the explanation.

PR #1513, opened 2020-03-04
PICARD-1777: Properly support fractional scaling on Windows 10

phw:
# Summary
With Qt 5.14 setting HighDpiScaleFactorRoundingPolicy allows proper scaling to e.g. 150% on Windows 10. Otherwise Qt5 would just scale to the next full factor, e.g. 200%.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1777

# Solution
Since Qt 5.14 is is possible to set `HighDpiScaleFactorRoundingPolicy.PassThrough`. This will configure Qt to use the scaling factor as provided by the operating system, e.g. 1.5 on Windows if the scaling is set to 150%. By default it is set to `Round`, which results in 1.5 getting rounded up to 2.

I tested this on Windows and Linux:

- Windows with Qt 5.14 will now correctly consider the scaling factor. With Qt 5.13 this is not supported and a system setting of 150% will effectively scale to 200%.
- Linux GNOME 3.34 (Wayland) Qt 5.14 fractional scaling is experimental and a mess, looks awful and blurry.
- ~~Linux GNOME 3.34 in general using just the default settings the scaling is only respected for the font sizes, not the graphics. You can manually tweak this for Qt by setting appropriate environment variables. But the proposed change makes no difference here.~~ That's no true, I had `QT_AUTO_SCREEN_SCALE_FACTOR=0` set on my environment due to issues I had with this in the past. But by default it works well.
- Linux KDE Plasma 5.18 and Qt 5.14 scaling of Qt applications works as expected, including using fractional scaling. This is true with and without the proposed change here.
- On macOS I can't test as I don't have a retina display. I suppose it is fine as this platform had the best support for scaling.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -955,6 +955,11 @@ def main(localedir=None, autoupdate=True):
     # Allow High DPI Support
     QtWidgets.QApplication.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps)
     QtWidgets.QApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
+    # HighDpiScaleFactorRoundingPolicy is available since Qt 5.14. This is
+    # required to support fractional scaling properly.
+    if hasattr(QtGui.QGuiApplication, 'setHighDpiScaleFactorRoundingPolicy'):
+        QtGui.QGuiApplication.setHighDpiScaleFactorRoundingPolicy(
+            QtCore.Qt.HighDpiScaleFactorRoundingPolicy.PassThrough)
 
     # Enable mnemonics on all platforms, even macOS
     QtGui.qt_set_sequence_auto_mnemonic(True)


phw:
> though I think we should document how to override it (using env var `QT_SCALE_FACTOR_ROUNDING_POLICY`).

I am not sure this is the most useful environment variable to document for end users. Let's say the user has configured the system for 1.5 scaling but just for Picard wants to use a scaling of 1.0 or 2.0 they are probably better off setting `QT_AUTO_SCREEN_SCALE_FACTOR=0` (will ignore the scaling) or `QT_SCALE_FACTOR=2` instead of trying to figure out the proper value for `QT_SCALE_FACTOR_ROUNDING_POLICY`. But so far all the complaints we got was about the system configured scaling not being applied.

PR #1514, opened 2020-03-04
Fix function line, column in ScriptRuntimeError.  Add tests.

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Have function `ScriptRuntimeError` show the proper start position of the function that raised the exception.  Add tests.

# Problem

The column number currently displayed for a `ScriptRuntimeError` exception is one character past the closing bracket for the function that raised the exception.  This is confusing.

# Solution

Update the code to pass the starting column number of the function (actually the $ at the start of the function name) to the exception.  Add tests.

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -59,7 +59,14 @@ class ScriptError(Exception):
 
 
 class ScriptParseError(ScriptError):
-    pass
+    def __init__(self, message, stackitem):
+        super().__init__(
+            "{0} at position {1}, line {2}".format(
+                message,
+                stackitem.line,
+                stackitem.column
+            )
+        )
 
 
 class ScriptEndOfFile(ScriptParseError):
@@ -70,13 +77,32 @@ class ScriptSyntaxError(ScriptParseError):
     pass
 
 
-class ScriptUnknownFunction(ScriptError):
+class ScriptUnknownFunction(ScriptParseError):
     pass
 
 
 class ScriptRuntimeError(ScriptError):
-    def __init__(self, message, name, line, position):
-        super().__init__("${1:s}:{2:d}:{3:d}: {0:s}".format(message, name, line, position))
+    def __init__(self, message, functionstackitem):
+        super().__init__(
+            "{0}:{1}:${2}: {3}".format(
+                functionstackitem.line,
+                functionstackitem.column,
+                functionstackitem.name,
+                message
+            )
+        )
+
+
+class StackItem:
+    def __init__(self, line, column):
+        self.line = line
+        self.column = column
+
+
+class FunctionStackItem(StackItem):
+    def __init__(self, line, column, name):
+        super().__init__(line, column)
+        self.name = name
 
 
 class ScriptText(str):
@@ -107,12 +133,13 @@ FunctionRegistryItem = namedtuple("FunctionRegistryItem",
                                   ["function", "eval_args",
                                    "argcount"])
 Bound = namedtuple("Bound", ["lower", "upper"])
-FunctionStackItem = namedtuple("FunctionStackItem", ["function", "column", "line"])
 
 
 class ScriptFunction(object):
 
     def __init__(self, name, args, parser, column=0, line=0):
+        self.stackitem = StackItem(line, column)
+        self.functionstackitem = FunctionStackItem(line, column, name)
         try:
             argnum_bound = parser.functions[name].argcount
             argcount = len(args)
@@ -129,16 +156,16 @@ class ScriptFunction(object):
                     too_many_args = False
 
                 if too_few_args or too_many_args:
-                    raise ScriptError(
-                        "Wrong number of arguments for $%s: Expected %s, got %i at position %i, line %i"
-                        % (name, expected, argcount, parser._x, parser._y)
+                    raise ScriptSyntaxError(
+                        "Wrong number of arguments for $%s: Expected %s, got %i"
+                        % (name, expected, argcount),
+                        StackItem(parser._y, parser._x)
                     )
         except KeyError:
-            raise ScriptUnknownFunction("Unknown function '%s'" % name)
+            raise ScriptUnknownFunction("Unknown function '%s'" % name, self.stackitem)
 
         self.name = name
         self.args = args
-        self.function = FunctionStackItem(name, column, line)
 
     def __repr__(self):
         return "<ScriptFunction $%s(%r)>" % (self.name, self.args)
@@ -147,13 +174,13 @@ class ScriptFunction(object):
         try:
             function, eval_args, num_args = parser.functions[self.name]
         except KeyError:
-            raise ScriptUnknownFunction("Unknown function '%s'" % self.name)
+            raise ScriptUnknownFunction("Unknown function '%s'" % self.name, self.stackitem)
 
         if eval_args:
             args = [arg.eval(parser) for arg in self.args]
         else:
             args = self.args
-        parser._function_stack.put(self.function)
+        parser._function_stack.put(self.functionstackitem)
         # Save return value to allow removing function from the stack on successful completion
         return_value = function(parser, *args)
         parser._function_stack.get()
@@ -191,13 +218,13 @@ Grammar:
         self._function_stack = LifoQueue()
 
     def __raise_eof(self):
-        raise ScriptEndOfFile("Unexpected end of script at position %d, line %d" % (self._x, self._y))
+        raise ScriptEndOfFile("Unexpected end of script", StackItem(self._y, self._x))
 
     def __raise_char(self, ch):
         #line = self._text[self._line:].split("\n", 1)[0]
         #cursor = " " * (self._pos - self._line - 1) + "^"
         #raise ScriptSyntaxError("Unexpected character '%s' at position %d, line %d\n%s\n%s" % (ch, self._x, self._y, line, cursor))
-        raise ScriptSyntaxError("Unexpected character '%s' at position %d, line %d" % (ch, self._x, self._y))
+        raise ScriptSyntaxError("Unexpected character '%s'", StackItem(self._y, self._x))
 
     def read(self):
         try:
@@ -242,7 +269,7 @@ Grammar:
             if ch == '(':
                 name = self._text[start:self._pos-1]
                 if name not in self.functions:
-                    raise ScriptUnknownFunction("Unknown function '%s'" % name)
+                    raise ScriptUnknownFunction("Unknown function '%s'" % name, StackItem(line, column))
                 return ScriptFunction(name, self.parse_arguments(), self, column, line)
             elif ch is None:
                 self.__raise_eof()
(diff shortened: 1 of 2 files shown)


picard/script.py
```diff
@@ -1280,11 +1284,12 @@ def func_datetime(parser, format=None):
...
     except ValueError:
+        function = parser._function_stack.get()
         raise ScriptRuntimeError(
             "Unsupported format code",
-            parser._function_stack.get(),
-            parser._y,
-            parser._x
+            function.function,
+            function.line,
+            function.column
```
zas: `function` is the tuple `FunctionStackItem`, right?
So we could just pass it to `ScriptRuntimeError`:

```python
stackitem = = parser._function_stack.get()
raise ScriptRuntimeError(
     "Unsupported format code",
     stackitem
)
```

Which makes me think we could perhaps make FunctionStackItem a class rather than a named tuple:

```python
class StackItem:
   def __init__(self, line, column):
       self.line = line
       self.column = column

class FunctionStackItem(StackItem):
     def __init__(self, line, column, name):
        super().__init__(line, column)
        self.name = name
```

So we could have something more consistent between all Script exceptions:

```python
raise ScriptRuntimeError("message", functionstackitem)
raise ScriptUnknownFunction("message", stackitem)
raise ScriptEndOfFile("message", stackitem)
raise ScriptSyntaxError("message", stackitem)
```

rdswift:
@zas 

I like all the changes you noted and will implement them all.  I suspect that I should also change the name of the PR because it will extend far beyond just fixing the line and column numbers for the ScriptRuntimeError.

picard/script.py
```diff
@@ -70,13 +77,32 @@ class ScriptSyntaxError(ScriptParseError):
...
+class ScriptUnknownFunction(ScriptParseError):
     pass
 
 
 class ScriptRuntimeError(ScriptError):
-    def __init__(self, message, name, line, position):
-        super().__init__("${1:s}:{2:d}:{3:d}: {0:s}".format(message, name, line, position))
+    def __init__(self, message, functionstackitem):
+        super().__init__(
+            "{0}:{1}:${2}: {3}".format(
```
zas: Please use named arguments:

```python
"{line}:{column}:${name}: {message}".format(
    line=functionstackitem.line,
    column=functionstackitem.column,
    name=functionstackitem.name,
    message=message
)
```

Also is there a special reason to not have `$` in the name of the script function? Documentation about scripting says:
`Functions start with a dollar sign and end with an argument list enclosed in parentheses (e.g. $lower(...)).`
So `$` is definitively a part of the function name.

So we could perhaps do:

```python
class FunctionStackItem(StackItem):
    def __init__(self, line, column, name):
        super().__init__(line, column)
        self.name = '$' + name
```

and get rid of `$` in related format strings. What do you think?

picard/script.py
```diff
@@ -145,13 +174,13 @@ def eval(self, parser):
         try:
             function, eval_args, num_args = parser.functions[self.name]
         except KeyError:
-            raise ScriptUnknownFunction("Unknown function '%s'" % self.name)
+            raise ScriptUnknownFunction("Unknown function '%s'" % self.name, self.stackitem)
```
zas: I wonder if we could just pass a special stackitem:

`UnknownFunctionStackItem(self.stackitem.line, self.stackitem.column, self.name)`

or even:

`UnknownFunctionStackItem(self.stackitem, self.name)`

 and format the message in `ScriptUnknownFunction()`.

Here what I'm thinking about:

```python
class StackItem:
    def __init__(self, line, column):
        self.line = line
        self.column = column

    def __str__(self):
        return '{line:d}:{column:d}'.format(
            line=self.line,
            column=self.column
        )


class FunctionStackItem(StackItem):
    def __init__(self, line, column, name):
        super().__init__(line, column)
        self.name = '$' + name

    def __str__(self):
        return '{prefix:s}:{name:s}'.format(
            prefix=super().__str__(),
            name=self.name
        )


class UnknownFunctionStackItem(FunctionStackItem):
    def __init__(self, *args):
        if isinstance(args[0], StackItem):
            stackitem = args[0]
            line = stackitem.line
            column = stackitem.column
            name = args[1]
        else:
            line, column, name = args
        super().__init__(line, column, name)

    def __str__(self):
        return '{prefix:s} Unknown function'.format(
            prefix=super().__str__()
        )
```

```python
stackitem = StackItem(4, 8)
print(str(UnknownFunctionStackItem(stackitem, "xxx")))
print(str(UnknownFunctionStackItem(9, 54, "yyy")))
```

outputs:

```
4:8:$xxx Unknown function
9:54:$yyy Unknown function
```


Good or bad idea?

zas: Actually we don't need different classes (suggested by @phw).

```python
class StackItem:
    def __init__(self, line, column, name=None):
        self.line = line
        self.column = column
        self.name = name

    def __str__(self):
        if name is None:
            return '{line:d}:{column:d}'.format(
                line=self.line,
                column=self.column
            )
        else:
            return '{line:d}:{column:d}:${name}'.format(
                line=self.line,
                column=self.column,
                name=self.name
            )
```

picard/script.py
```diff
@@ -59,7 +59,14 @@ class ScriptError(Exception):
 
 
 class ScriptParseError(ScriptError):
-    pass
+    def __init__(self, message, stackitem):
+        super().__init__(
+            "{0} at position {1}, line {2}".format(
+                message,
+                stackitem.line,
+                stackitem.column
```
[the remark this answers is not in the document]
rdswift: Yeah, that's a habit I need to develop.  I've gone through my code and tried to change all occurrences to use named arguments.

picard/script.py
rdswift: I couldn't think of any reason to not have '$'  as part of the function name in the class, so I've added it as suggested.

picard/script.py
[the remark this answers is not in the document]
rdswift: > So I would unify to:
`"{line}:{column}: {message}"` and `"{line}:{column}:{name}: {message}"`

I can do that, but that would mean that the error message shown when entering a script would change from (for example) "**Unexpected character at line 10, column 27**" to "**10:27: Unexpected character**".

I'm okay with that because I understand what the "10:27" means, but I wonder if other users will find this confusing.

For now, I've set up all exceptions to use the `"{line}:{column}: {message}"` and `"{line}:{column}:{name}: {message}"` format, including those displayed to the user during script development / input.

Also moved the message text for the `ScriptEndOfFile` and `ScriptUnknownFunction` exceptions to the exceptions themselves so that is only appears in one place, since there is a standard message for these two exceptions.

picard/script.py
rdswift: I used the combined class as suggested.  (Great idea.)

picard/script.py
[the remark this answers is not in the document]
zas: > I'm okay with that because I understand what the "10:27" means, but I wonder if other users will find this confusing.

The idea behind this is to make messages consistent, and ease translation, the final format of the message shouldn't prevent us to improve this.
Of course "Unexpected character at line 10, column 27" is clearer than "10:27: Unexpected character", and if you prefer keep it, anything til formats of all message are consistent.

picard/script.py
```diff
@@ -59,24 +59,66 @@ class ScriptError(Exception):
...
+            )
+        )
+
+
+class StackItem:
+    def __init__(self, line, column, name=None):
+        self.line = line
+        self.column = column
+        if name is None:
+            self.name = name
```
zas:
```python
self.name = None
```

picard/script.py
```diff
@@ -189,13 +233,13 @@ def __init__(self):
...
     def __raise_eof(self):
-        raise ScriptEndOfFile("Unexpected end of script at position %d, line %d" % (self._x, self._y))
+        raise ScriptEndOfFile(StackItem(line=self._y, column=self._x))
 
     def __raise_char(self, ch):
         #line = self._text[self._line:].split("\n", 1)[0]
         #cursor = " " * (self._pos - self._line - 1) + "^"
         #raise ScriptSyntaxError("Unexpected character '%s' at position %d, line %d\n%s\n%s" % (ch, self._x, self._y, line, cursor))
-        raise ScriptSyntaxError("Unexpected character '%s' at position %d, line %d" % (ch, self._x, self._y))
+        raise ScriptSyntaxError(StackItem(line=self._y, column=self._x), "Unexpected character '%s'")
```
zas:
```python
raise ScriptSyntaxError(StackItem(line=self._y, column=self._x), "Unexpected character '%s'" % ch)
```

picard/script.py
```diff
@@ -189,13 +233,13 @@ def __init__(self):
         self._function_stack = LifoQueue()
 
     def __raise_eof(self):
-        raise ScriptEndOfFile("Unexpected end of script at position %d, line %d" % (self._x, self._y))
+        raise ScriptEndOfFile(StackItem(line=self._y, column=self._x))
 
     def __raise_char(self, ch):
         #line = self._text[self._line:].split("\n", 1)[0]
         #cursor = " " * (self._pos - self._line - 1) + "^"
         #raise ScriptSyntaxError("Unexpected character '%s' at position %d, line %d\n%s\n%s" % (ch, self._x, self._y, line, cursor))
```
zas: Those commented lines can be removed

picard/script.py
zas: This didn't trigger a test error, one should be added.

picard/script.py
```diff
@@ -1280,11 +1326,10 @@ def func_datetime(parser, format=None):
...
         return datetime.datetime.now(tz=local_tz).strftime(format)
     except ValueError:
+        stackitem = parser._function_stack.get()
         raise ScriptRuntimeError(
-            "Unsupported format code",
-            parser._function_stack.get(),
-            parser._y,
-            parser._x
+            stackitem,
+            "Unsupported format code"
```
zas: Since it is shorter now, it can be on one line.

```python
raise ScriptRuntimeError(stackitem, "Unsupported format code")
```

picard/script.py
rdswift: Changed as requested.

picard/script.py
rdswift: Test updated, and it would now catch that error (if it still existed).  I tested it before fixing the error.

picard/script.py
rdswift: Removed.  Sorry, I missed that in my earlier cleanup.

picard/script.py
rdswift: All on one line now.

picard/script.py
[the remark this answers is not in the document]
rdswift: Left as  `"{line}:{column}: {message}"` and `"{line}:{column}:{name}: {message}"`.  We can always change it later if enough people notice and complain.

PR #1516, opened 2020-03-05
Improve handling of CAA thumbnail sizes, add tests

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -80,7 +80,12 @@ _CAA_THUMBNAIL_SIZE_MAP = OrderedDict([
     (1200, CaaSizeItem('1200', N_('1200 px'))),
     (-1, CaaSizeItem(None, N_('Full size'))),
 ])
-
+_CAA_THUMBNAIL_SIZE_ALIASES = {
+    'large': '500',
+    'small': '250',
+    '500': 'large',
+    '250': 'small',
+}
 _CAA_IMAGE_SIZE_DEFAULT = 500
 
 _CAA_IMAGE_TYPE_DEFAULT_INCLUDE = ['front']
@@ -103,8 +108,13 @@ def caa_url_fallback_list(desired_size, thumbnails):
     for item_id, item in reversed_map.items():
         if item_id == -1 or item_id > desired_size:
             continue
-        if item.thumbnail in thumbnails:
-            urls.append(thumbnails[item.thumbnail])
+        url = thumbnails.get(item.thumbnail, None)
+        if url is None:
+            size_alias = _CAA_THUMBNAIL_SIZE_ALIASES.get(item.thumbnail, None)
+            if size_alias is not None:
+                url = thumbnails.get(size_alias, None)
+        if url is not None:
+            urls.append(url)
     return urls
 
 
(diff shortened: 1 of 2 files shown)


picard/coverart/providers/caa.py
```diff
@@ -75,12 +75,17 @@
...
+    (500, CaaSizeItem('500', N_('500 px'))),
     (1200, CaaSizeItem('1200', N_('1200 px'))),
     (-1, CaaSizeItem(None, N_('Full size'))),
 ])
-
+_CAA_THUMBNAIL_SIZE_ALIASES = {
+    'large': '500',
+    'small': '250',
+    '500': 'large',
+    '250': 'small',
```
phw: Why does it match in two directions? As I understand it this is only needed for translating the pixel sizes to the name used in the JSON response, so we don't have a need for the reverse.

Also I wonder if we really need this. Having the name a part of `CaaSizeItem` would be totally sufficient I think and better keep things together. If we also want to have the actual size as part of `CaaSizeItem` we should probably add both?

picard/coverart/providers/caa.py
```diff
@@ -103,8 +108,13 @@ def caa_url_fallback_list(desired_size, thumbnails):
     for item_id, item in reversed_map.items():
         if item_id == -1 or item_id > desired_size:
             continue
-        if item.thumbnail in thumbnails:
-            urls.append(thumbnails[item.thumbnail])
+        url = thumbnails.get(item.thumbnail, None)
+        if url is None:
+            size_alias = _CAA_THUMBNAIL_SIZE_ALIASES.get(item.thumbnail, None)
+            if size_alias is not None:
+                url = thumbnails.get(size_alias, None)
```
phw: Does it actually happen that we have `"large"` in the response but not `"500"`? Because as I understand it this is the use case here.

zas: Yes, that's possible :(
This is why I want to add "compatibility" with old ("large" & "small") and new ("250", "500"), even newer ("1200") cases, because `index.json` files aren't always properly updated.

https://ia802706.us.archive.org/18/items/mbid-ad7a42c2-8baa-4f6c-8d71-1ae2f7cf3afa/index.json

```json
{
  "images": [
    {
      "types": [
        "Front"
      ],
      "front": true,
      "back": false,
      "edit": 19432737,
      "image": "http://coverartarchive.org/release/ad7a42c2-8baa-4f6c-8d71-1ae2f7cf3afa/2366957454.jpg",
      "comment": "",
      "approved": true,
      "thumbnails": {
        "large": "http://coverartarchive.org/release/ad7a42c2-8baa-4f6c-8d71-1ae2f7cf3afa/2366957454-500.jpg",
        "small": "http://coverartarchive.org/release/ad7a42c2-8baa-4f6c-8d71-1ae2f7cf3afa/2366957454-250.jpg"
      },
      "id": "2366957454"
    }
  ],
  "release": "http://musicbrainz.org/release/ad7a42c2-8baa-4f6c-8d71-1ae2f7cf3afa"
}
```

picard/coverart/providers/caa.py
zas: `CaaSizeItem` is meant for display in interface (size selector in CAA options), and should have unique items, for size selection. Perhaps we can merge and simplify, but it is likely to imply changes in UI code, which I wanted to avoid in this PR.

I added both ("large" & "500" + "small" & "250") during tests devel.
And yes, now, "large" and "small" keys are useless, removed in ff006f1.

The "500"/"250" without "large"/"small" shouldn't happen, but the reverse does (even though it is rare), see my comment below for an example.

PR #1522, opened 2020-03-25
PICARD-1800: Simplify Unicode 'Bullet' to ASCII 'Hyphen-minus'

Undearius:
Minor change to add U+2022 "•" to be simplified to a period "." for ASCII compliant file naming. This stops it from appearing as the default underscore _ character.


phw:
Thanks a lot for the pull request. Makes sense to have a replacement for this. I just wonder if a period is the proper one. How are the bullets usually used? I think in our context maybe often as a delimiter between different parts of a track title or such, so maybe it would make more sense to use a `-` instead?

@zas what do you think?

Also we must not forget to add a ticket for this at https://tickets.metabrainz.org/projects/PICARD/issues/ to track this for the changelog

zas:
> Thanks a lot for the pull request. Makes sense to have a replacement for this. I just wonder if a period is the proper one. How are the bullets usually used? I think in our context maybe often as a delimiter between different parts of a track title or such, so maybe it would make more sense to use a `-` instead?

Yes, you may be right, bullets are usually more like a `-` than a dot, because they are used as separator.

zas:
Apart the discussion about the choice of dot or whatever, the patch looks good to me.

I wonder if we could improve things in this field, making things user-editable somehow, but that's not the subject here.

Undearius:
I did struggle with settling on a decision of which character to use, so I'm open to discuss it. Go with functionality and use a hyphen or stylistically and use a period.

The specific example that caused me to open this was for 54-40's album with the same name, which the bullet functions as a hyphen. Off the top of my head, I also know about Bee Gees album ESP but when I looked it up just now, it's stylized as "E.S.P".

It ultimately comes down to either function or style. As this doesn't affect the tags themselves, they'll still visually show bullet separators, it'll be how the user interacts with the file or directory.

phw:
Thanks a lot, merged and ready for next release :) I also added a ticket PICARD-1800

PR #1523, opened 2020-03-28
PICARD-1795: Allow deleting iTunNORM and iTunPGAP tags from ID3

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The tags `comment:iTunNORM` and `comment:iTunPGAP` in ID3 tags cannot be deleted.

* JIRA ticket (_optional_): PICARD-1795

# Solution
There was a special code preventing the deletion of those tags added in https://github.com/metabrainz/picard/commit/bff60dab8a69674bfe14e6b5f02b5b597bdcb4ec

I don't see any harm being done allowing those tags to be deleted on request. There is some special handling for storing the `iTunNORM` tag as iTunes expected it, but this is independent of the issue of allow the deletion.

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -554,11 +554,10 @@ class ID3File(File):
                     _remove_people_with_role(tags, ['TMCL', 'TIPL', 'IPLS'], role)
                 elif name.startswith('comment:') or name == 'comment':
                     (lang, desc) = parse_comment_tag(name)
-                    if desc.lower()[:4] != 'itun':
-                        for key, frame in list(tags.items()):
-                            if (frame.FrameID == 'COMM' and frame.desc == desc
-                                and frame.lang == lang):
-                                del tags[key]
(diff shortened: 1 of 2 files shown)


phw:
@samithaj Can you please take a look at this? You added the original code and I might be missing something

samithaj:
> @samithaj Can you please take a look at this? You added the original code and I might be missing something
I think it's @samj1912

phw:
> I think it's @samj1912

Yes of course . Sorry for bothering you, I wasn't careful with the autocompletion :(

samj1912:
This has been a while. I don't exactly remember this, but it had something to do with the fact that iTunes needs case sensitive tags or something and mutagen id3 didn't support that. Anyway, lgtm. I can't find any resources as of right now backing up the above.

PR #1524, opened 2020-04-17
PICARD-1803: Fixed decamelcasing relationship attributes

phw:
# Summary
EWI must not become E W I

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The _decamelcase function in mbjson turns "EWI" into "E W I".

* JIRA ticket (_optional_): PICARD-1803

# Solution
Fix the decamelcase function, to not enter spaces before a upper case character if it is preceded by another upper case or whitespace character.

# Action
@zas This fixes the issue of added whitespaces in EWI, but I am not sure about two things:

1. Do we need the decamelcase at all? For which attributes is it needed? I have a faint memory of this, but it probably is for some legacy reason. If it is only for some specific attributes we maybe could at least call decamelcase only when needed.

2. With this patch "EWI" still gets turned to lowercase "ewi". Is this expected? Should we change it?

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -100,7 +100,7 @@ _RELEASE_GROUP_TO_METADATA = {
 
 
 def _decamelcase(text):
-    return re.sub(r'([A-Z])', r' \1', text).strip()
+    return re.sub(r'(?<![A-Z\s])([A-Z])', r' \1', text).strip()
 
 
 _REPLACE_MAP = {}
--- a/test/data/ws_data/recording_credits.json
+++ b/test/data/ws_data/recording_credits.json
@@ -164,6 +164,30 @@
     "attributes": ["solo"],
     "type": "vocal",
     "type-id": "0fdbe3c6-7700-4a31-ae54-b53f06ae1cfa"
+  }, {
+    "attribute-credits": {},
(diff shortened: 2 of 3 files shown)


zas:
> * Do we need the decamelcase at all? For which attributes is it needed? I have a faint memory of this, but it probably is for some legacy reason. If it is only for some specific attributes we maybe could at least call decamelcase only when needed.

I didn't find the historic reason of this, this is why I said I wasn't sure how to fix it.
What's the purpose of this?

> 
> * With this patch "EWI" still gets turned to lowercase "ewi". Is this expected? Should we change it?

I'd avoid to change capitalization without good reason.

If we don't have any good reason to change capitalization I'd rather remove this code, but I guess there was a reason this code was first added, until we are sure what we are doing I'm fine with this patch.

phw:
> If we don't have any good reason to change capitalization I'd rather remove this code, but I guess there was a reason this code was first added, until we are sure what we are doing I'm fine with this patch.

Agreed. Let's merge this to fix the actual bug, but I'll try to find out why this was added in the first place and maybe we can get rid of it.

PR #1529, opened 2020-04-30
PICARD-1809: Load based on extensions, fallback to guessing if it fails

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Prioritize loading based on file extension and fallback to format guessing if it fails.

# Problem

The current logic opens one file at a time, reads the file header, try to guess the file format based on the header and then proceeds to the loading phase. This results in a huge overhead of opening every single file twice, which in case of large libraries can be even worse as the file doesn't stay in memory cache.

* JIRA ticket (_optional_): PICARD-1809

# Solution

Assume that a file extension is correct if one exists. Try to load the file based on the file extension and fallback to format guessing based on the header if it fails. Finally, retry loading with guessed format. There are 3 file opening operations in case of failures, but the file stays cached in memory.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -165,7 +165,16 @@ class File(QtCore.QObject, Item):
         if self.tagger.stopping:
             log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
-        return self._load(filename)
+        try:
+            # Try loading based on extension first
+            return self._load(filename)
+        except Exception:
+            from picard.formats import guess_format
+            # If it fails, force format guessing and try loading again
+            file_format = guess_format(filename)
+            if not file_format and type(file_format) == type(self):
+                raise
+            return file_format._load(filename)
 
     def _load(self, filename):
         """Load metadata from the file."""
--- a/picard/formats/__init__.py
+++ b/picard/formats/__init__.py
@@ -79,15 +79,16 @@ def guess_format(filename, options=_formats):
 def open_(filename):
     """Open the specified file and return a File instance with the appropriate format handler, or None."""
     try:
-        # First try to guess the format on the basis of file headers
-        audio_file = guess_format(filename)
-        if not audio_file:
-            i = filename.rfind(".")
-            if i < 0:
-                return None
+        # Use extension based opening as default
+        i = filename.rfind(".")
+        if i >= 0:
             ext = filename[i+1:].lower()
-            # Switch to extension based opening if guess_format fails
             audio_file = _extensions[ext](filename)
+        else:
+            # If there is no extension, try to guess the format based on file headers
+            audio_file = guess_format(filename)
+        if not audio_file:
+            return None
         return audio_file
     except KeyError:
         # None is returned if both the methods fail


picard/file.py
```diff
@@ -167,7 +167,15 @@ def _load_check(self, filename):
...
             log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
-        return self._load(filename)
+        try:
+            # Try loading based on extension first
+            ret = self._load(filename)
+        except Exception:
+            from picard.formats import guess_format
+            # If it fails, force guess and try loading again
+            self = guess_format(filename)
```
phw, line 176: A small improvement could be that it just raises if the newly detected format is the same as before. Also guess_format might return `None`. So maybe something like this:

```
format = guess_format(filename)
if not format or format == self:
    raise
ret = format._load(filename)
```

picard/file.py
```diff
@@ -167,7 +167,15 @@ def _load_check(self, filename):
         if self.tagger.stopping:
             log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
-        return self._load(filename)
+        try:
+            # Try loading based on extension first
+            ret = self._load(filename)
+        except Exception:
```
phw, line 173: This is one of the few cases where I think catching a generic exception is ok. Fetching a `MutagenError` might be more specific, but there had been some cases where mutagen throws all kinds of exception (even though there have been a couple of upstream fixes for this recently).

picard/file.py
Sophist-UK: Do we need more unit tests for the guessing functionality?

Gabrielcarvfer: Hmm, the score function checks for the extension. I'm removing that.

phw: That's likely because mutagen itself also considers the extension in its scoring logic, so it is not a pure content based detection. Our own extension based check still is a performance shortcut :)

phw: I don't think you need a workaround. Let's just leave it as is for now. I don't think it does much harm that mutagen additionally also considers the extension. If this is something to change it should be done on mutagen.

Also it's mostly unrelated to this change, if we want to update the scoring logic from the Picard side we should do this in a separate PR.

phw: Now it's calling guess_format twice ;) You should store it in a variable. Also you can do the comparison with self then. No need to try loading with the same format twice (e.g. an broken MP3 file that gets correctly detected by extension and content but which will not load due to a different error, no need to open it twice)

Gabrielcarvfer: Saving it to a variable and then assigning throws an exception, but doing it directly to self doesn't. I don't think it's a huge problem for three reasons: you don't really expect files with wrong extensions that often (e.g. VLC completely freaks out), files without extensions were already guessed and the files are already cached in memory and will have their metadata loaded in the near future. It is quite different from guessing every single file format, and then trying to load each of them, having the file cache purged in the meantime.

Gabrielcarvfer: Measured the loading times: fast path ~1ms per file, slow path ~2ms per file. The difference between using one or two guess_format is statistically irrelevant.

zas: > Saving it to a variable and then assigning throws an exception, but doing it directly to self doesn't.

> The difference between using one or two guess_format is statistically irrelevant, as expected.

I don't get why calling `guess_format()` twice is needed. Even if performance isn't much impacted, what prevent us to get rid of a useless call?

picard/file.py
```diff
@@ -165,7 +165,17 @@ def _load_check(self, filename):
         if self.tagger.stopping:
             log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
-        return self._load(filename)
+        try:
+            # Try loading based on extension first
+            ret = self._load(filename)
```
zas, line 170: After those changes, I guess we could just `return self._load(filename)` and get rid of `ret`.

picard/file.py
Gabrielcarvfer: In this case, my mistake. It works fine if the guessed file is correct and fails when it doesn't (e.g. rename an .aac as .mp3, due to how mutagen scores are calculated it will be recognized as an MP3 file, and will fail to load. Removing the duplicated call.

phw:
I gave this a test run, found a small annoyance: If loading fails for some reason with an exception and the detection finds the same format again and tries loading again the same exception is logged twice.

So could we actually not run loading again if it failed for the same format before, like I originally suggested? E.g. this will do:

```
format = guess_format(filename)
if not format or type(format) == type(self):
    raise
return format._load(filename)
```

Also no need for assigning to `self`. `self`is just a local variable in this case, assigning to it does nothing useful but maybe leads someone reading the code it is doing something special.

Gabrielcarvfer:
> So could we actually not run loading again if it failed for the same format before, like I originally suggested? E.g. this will do:
> 
> ```
> format = guess_format(filename)
> if not format or type(format) == type(self):
>     raise
> return format._load(filename)
> ```
OK

> Also no need for assigning to `self`. `self`is just a local variable in this case, assigning to it does nothing useful but maybe leads someone reading the code it is doing something special.
I am not really sure how these would interact with the UI, so it made sense to me.

Gabrielcarvfer:
Tried force pushing to rerun the checks, but MacOS with Python 3.7 seems to be bugged. =/

PR #1534, opened 2020-05-09
PICARD-1812: Save RIFF INFO tags for WAVE

phw:
# Summary
This adds support for saving RIFF INFO tags for WAVE files in addition to the ID3 tags.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1812

# Solution
RIFF INFO tags are treated a bit similar to how we handle ID3v1. It is basically a very limited legacy format with limited tag support and no proper Unicode support. I provided an encoding option, so one can have Unicode support in RIFF INFO with some ASCII backward compatibility, but I don't even know what tools would support this. Otherwise the default is ISO-8859-1.

Maybe we could either exchange ISO-8859-1 this for Windows-1252 or provide this as an additional option. Not sure what is more compatible, but Windows seems to default to Windows-1252 here while some software might use ISO-88591-1. Both are very similar, but Windows-1252 supports some additional characters. I need to test this.  

For getting the tags I looked a lot in https://exiftool.org/TagNames/RIFF.html and experimented with what Windows supports and writes itself (if you edit WAVE tags with Windows Explorer).

Apart from the encoding question above I consider this PR ready.

# Action
We could try to move parts of this code upstream to mutagen, but I'm not sure about the interface yet and what is really useful there.


phw:
Here is a screenshot of the UI:

grafik


It's getting a bit crowded here. I thought about moving the file format specific options for AAC, AC3 and WAVE into a tab list. But I would like to do this in a separate PR and concentrate on the RIFF INFO issue here.

phw:
I changed the encoding to default to Windows-1252. This provides best compatibility with Windows and supports a couple of additional characters not supported by ISO-8859-1, being otherwise compatible.

User can still choose to use UTF-8 instead, which e.g. will be read successfully by foobar2000 (but which will prefer the ID3 tags anyway if present).

phw:
I guess we should have some tests for the RIFF saving and loading. I'll add some.

phw:
@Gabrielcarvfer  Did another test run with my 12900 files across ~990 albums, and I think we are very close now :) Loading behaved well, took 22 minutes overall.

But one issue remains: After loading everything Picard still shows ~6000 files as pending. Somewhere it fails to update the status for the files sometimes.

phw:
@zas I added tests. For now those do not run on CI because we do not test against mutagen master there. But I ran them locally

PR #1535, opened 2020-05-11
PICARD-1818: Make PyQt5.QtDBus optional

jcaesar:
# Summary
* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other: Minor enhancement?
* Make PyQt5.QtDBus optional / do not fail if QtDBus cannot be imported

# Problem

It is possible to install PyQt5 without support for dbus ([e.g. under gentoo](https://gitweb.gentoo.org/repo/gentoo.git/tree/dev-python/PyQt5/PyQt5-5.14.2.ebuild?id=52d0279f1ae255004dbf5b3e5b8525d6ff1e4912#n125)) resulting in Picard failing to start up with a message like
```
  File "/usr/bin/picard", line 2, in <module>
    from picard.tagger import main
  File "/usr/lib/python3.7/site-packages/picard/tagger.py", line 61, in <module>
    from PyQt5.QtDBus import QDBusConnection
ModuleNotFoundError: No module named 'PyQt5.QtDBus'
```

PICARD-1818

# Solution

Wrap imports of `PyQt5.QtDBus` with `try`.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -58,7 +58,6 @@ from PyQt5 import (
     QtGui,
     QtWidgets,
 )
-from PyQt5.QtDBus import QDBusConnection
 
 from picard import (
     PICARD_APP_ID,
@@ -972,9 +971,12 @@ def main(localedir=None, autoupdate=True):
     if picard_args.long_version:
         return longversion()
 
-    if not (IS_WIN or IS_MACOS or IS_HAIKU):
+    try:
+        from PyQt5.QtDBus import QDBusConnection
         dbus = QDBusConnection.sessionBus()
         dbus.registerService(PICARD_APP_ID)
+    except ImportError:
+        pass
 
     tagger = Tagger(picard_args, unparsed_args, localedir, autoupdate)
 
--- a/picard/ui/statusindicator.py
+++ b/picard/ui/statusindicator.py
@@ -90,94 +90,102 @@ if IS_WIN:
     DesktopStatusIndicator = WindowsTaskbarStatusIndicator
 
 elif not (IS_MACOS or IS_HAIKU):
-    from PyQt5.QtCore import (
-        Q_CLASSINFO,
-        QObject,
-        pyqtSlot,
-    )
-    from PyQt5.QtDBus import (
-        QDBusAbstractAdaptor,
-        QDBusConnection,
-        QDBusMessage,
-    )
+    QDBusConnection = None
 
-    from picard import PICARD_DESKTOP_NAME
+    try:
+        from PyQt5.QtCore import (
+            Q_CLASSINFO,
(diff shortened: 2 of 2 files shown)


picard/tagger.py
```diff
@@ -973,8 +972,13 @@ def main(localedir=None, autoupdate=True):
...
     if not (IS_WIN or IS_MACOS or IS_HAIKU):
-        dbus = QDBusConnection.sessionBus()
-        dbus.registerService(PICARD_APP_ID)
+        try:
+            from PyQt5.QtDBus import QDBusConnection
+        except ImportError:
+            pass
+        else:
+            dbus = QDBusConnection.sessionBus()
+            dbus.registerService(PICARD_APP_ID)
```
zas, line 981: What about getting rid of platform checks?

```python
        try:
            from PyQt5.QtDBus import QDBusConnection
            dbus = QDBusConnection.sessionBus()
            dbus.registerService(PICARD_APP_ID)
        except ImportError:
             dbus = None
```       

And later check with `if dbus:`

@phw : any drawback with this approach?

jcaesar: (Was wondering about that, too.)

phw: I just wanted to avoid that we register the service on DBus on platforms where we definitely know we will not need it. It's likely very uncommon to find this, but dbus is potentially available for Windows. Similar for Haiku. No idea about macOS.

But yeah, we could probably remove the check here. I would however not remove the check from statusindicator.py. There is really no need to even try to setup UnityLauncherEntryStatusIndicator on those platforms.

phw:
@jcaesar Looks like you are the victim of flake8 having released a 3.8 version ;) I guess we should either pin the used flake8 to 3.7.9 for now or actually decide how to deal with the reported issues.

phw:
This is all about not using `I`, `O` or `l` as variable names because they can be confused (https://www.flake8rules.com/rules/E741.html). Not so much an issue with the font I use here.

It's just a handful of cases where we use `l` (short for list, length or language). We could fix those.

@zas What's your take? Adapt the rule and update the code?

phw:
I guess the rule isn't too bad, decided to fix it: #1537

phw:
We need an issue for the changelog, I added PICARD-1818

jcaesar:
Ah, I see. I'll wait for you to merge #1537 then.

phw:
@jcaesar Could you rebase against master?

phw:
@jcaesar Let's spare the rebase action :D I just merged this PR, it should run through on master. Thanks a lot for this fix!

jcaesar:
Wow, speedy. Thank you.

PR #1536, opened 2020-05-11
PICARD-1704: Dark mode for Windows

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
Adds dark mode for Picard on Windows.

# Problem

Picard doesn't have a dark theme on Windows.

* JIRA ticket (_optional_): PICARD-1704

# Solution

Scan the registry for the user settings regarding the dark theme use on Windows and change the Picard color palette accordingly during the startup, using [qtmodern](https://github.com/gmarull/qtmodern). It doesn't repaint during execution, but I doubt people change the theme settings frequently.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -25,6 +25,7 @@
 # Copyright (C) 2018 Bob Swift
 # Copyright (C) 2018 virusMac
 # Copyright (C) 2019 Joel Lintunen
+# Copyright (C) 2020 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -159,7 +160,7 @@ class Tagger(QtWidgets.QApplication):
     _debug = False
     _no_restore = False
 
-    def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
+    def __init__(self, picard_args, unparsed_args, localedir, autoupdate, dark_theme=False):
 
         # Use the new fusion style from PyQt5 for a modern and consistent look
         # across all OSes.
@@ -173,6 +174,10 @@ class Tagger(QtWidgets.QApplication):
         super().setStyleSheet(
             'QGroupBox::title { /* PICARD-1206, Qt bug workaround */ }'
         )
+
+        if dark_theme:
+            import qtmodern.styles
+            qtmodern.styles.dark(self)
 
         self._cmdline_files = picard_args.FILE
         self.autoupdate_enabled = autoupdate
@@ -404,8 +409,8 @@ class Tagger(QtWidgets.QApplication):
     def run(self):
         if config.setting["browser_integration"]:
             self.browser_integration.start()
-        self.window.show()
         QtCore.QTimer.singleShot(0, self._run_init)
+        self.window.show()
         res = self.exec_()
         self.exit()
         return res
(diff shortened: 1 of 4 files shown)


phw:
I did some tests with this a while back, and I was actually not too happy with qtmodern. Adding a dependency for what is bascially just setting a different color palette looks a bit too much to me, especially as the colors being used arw not even fully matching the Windows default palette and the chose highlight color is completely ignored. 

I tried it a bit different with actually setting the palette inside Picard, see https://github.com/phw/picard/tree/PICARD-1704-windows-10-dark-mode

This allows to also honor the configured accent color and we can actually easier tweak the colors in case of issue reports. Now I was also not fully satisifed with the result, that's why I haven't submitted this yet. See below for a comparisson (left custom palette, right qtmodern).

grafik

I'm not opposed to more got with the qtmodern palette, though, as it is a bit lighter. But with honoring the Windows accent color. And I would really go with the palette shipped as part of Picard, it's really not much code overall, see e.g. https://github.com/phw/picard/commit/9ecfe3f0d3996785750e866de81eb8bebd19a064

The one thing I thought really should be added is an adpated theme for the script syntax highlighting. I think the default doesn't work too well with a dark theme.

Gabrielcarvfer:
> 
> I did some tests with this a while back, and I was actually not too happy with qtmodern. Adding a dependency for what is bascially just setting a different color palette looks a bit too much to me, especially as the colors being used arw not even fully matching the Windows default palette and the chose highlight color is completely ignored.

OK.

> 
> I tried it a bit different with actually setting the palette inside Picard, see https://github.com/phw/picard/tree/PICARD-1704-windows-10-dark-mode

Looks amazing, especially the purple in dark gray.

> This allows to also honor the configured accent color and we can actually easier tweak the colors in case of issue reports. Now I was also not fully satisifed with the result, that's why I haven't submitted this yet. See below for a comparisson (left custom palette, right qtmodern).

The accent is a gimmick and just make things super hard in terms of maintaining proper contrast.

> I'm not opposed to more got with the qtmodern palette, though, as it is a bit lighter. But with honoring the Windows accent color. And I would really go with the palette shipped as part of Picard, it's really not much code overall, see e.g. [phw@9ecfe3f](https://github.com/phw/picard/commit/9ecfe3f0d3996785750e866de81eb8bebd19a064)

Your's looks better anyways.

> The one thing I thought really should be added is an adpated theme for the script syntax highlighting. I think the default doesn't work too well with a dark theme.

How about something like Darcula?

/Dark%20Dracula%20Theme%20for%20Sublime%20text.PNG)

phw:
> The accent is a gimmick and just make things super hard in terms of maintaining proper contrast.

True in general, but it seemed to work rather well. I had to play a bit with the accent text color, but as Windows limits the allowed colors it worked pretty well IMHO.

> How about something like Darcula?

Yes, could work 👍 Or maybe Tango.

In my branch I lost myself a bit when I started to move the theming into a separate file:
https://github.com/phw/picard/commit/945bb9762b95f46dd3d581fb20c702200750fb8e . I think I started over engineering things here ;) That was another reason I put it on ice. But I think the two commits before this were actually fine.

Gabrielcarvfer:
I'd pick the purple (or maybe the orange). Looks great, great contrast, fits the branding.

Few results. All of them look kinda wrong, but I think due to the color distribution. Whatever is inside %% takes 80% (exaggeration)  of the screen. From these, I would say pastel tones are better, cyan/green/magenta are too bright, but toning them down reduces the contrast.

<details>
    <summary>
        Based on Monokai palette: color order in scripting.py (magenta, orange, pink, green, cyan)
    </summary>

image
</details>

<details>
    <summary>
        Based on Darcula palette: color order (magenta, orange, green, yellow, cyan)
   </summary>

image
</details>

<details>
    <summary>
        Based on Tango fusion palette: (orange, (0,190,255,250), green, magenta, (167,66,245,220))
   </summary>

image
</details>

picard/tagger.py
```diff
@@ -159,7 +160,7 @@ class Tagger(QtWidgets.QApplication):
     _debug = False
     _no_restore = False
 
-    def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
+    def __init__(self, picard_args, unparsed_args, localedir, autoupdate, dark_theme=False):
```
zas, line 163: Just a minor thing, instead of passing a boolean, it would be more future-proof to pass a numeric constant.
Something like:

```python
from enum import IntEnum, unique

@unique
class Themes(IntEnum):
    default = 0
    dark = 1
    sometheme = 2

def __init__(self, picard_args, unparsed_args, localedir, autoupdate, theme=Themes.default):

...
if theme == Themes.dark:
    ...

```

It will also allow us to implement a theme chooser, and to save/load user prefs.

Gabrielcarvfer: I'm going to close this PR. Philipp's branch is way ahead on that. Trying palettes for the script highlighting.



<details>
   <summary>
   My favorite one so far
   </summary>

image

image

</details>

phw:
@Gabrielcarvfer Allright, thanks for starting this discussion here. That means I need to submit a PR finally :D I will open  a draft PR with my current code for discussion and we can improve upon that.

PR #1539, opened 2020-05-13
PICARD-1704: Support Windows 10 dark mode

phw:
# Summary
Enable a dark layout for Picard on Windows if Windows has dark mode enabled.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
This is an alternative suggestion to #1536. I had this started as an experiment, but now putting it up for further discussion.

The colors are using the actual colors used by the core Windows apps (Settings, Explorer).

One detail that cannot be changed that way is the color of the Windows title bar, as this is done by Windows itself. And for some reason Windows either renders this for the active Window in the configured accent color, but inactive Windows always get a light gray even in dark mode. Only UWP apps get a dark titlebar. The only way around for a native app is rendering its own titlebar, but I honestly think we should not do this.

* JIRA ticket (_optional_): PICARD-1704

# Solution

# Action
Todo:

- Set an editor theme for the script editor in dark mode

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -175,6 +175,48 @@ class Tagger(QtWidgets.QApplication):
             'QGroupBox::title { /* PICARD-1206, Qt bug workaround */ }'
         )
 
+        # Set a dark palette on Windows 10 if dark app mode is enabled
+        if IS_WIN:
+            import winreg
+
+            dark_theme = False
+            with winreg.OpenKey(winreg.HKEY_CURRENT_USER, r"Software\Microsoft\Windows\CurrentVersion\Themes\Personalize") as key:
+                try:
+                    dark_theme = winreg.QueryValueEx(key, "AppsUseLightTheme")[0] == 0
+                except OSError:
+                    log.warning('Failed reading AppsUseLightTheme from registry')
+
+            if dark_theme:
+                with winreg.OpenKey(winreg.HKEY_CURRENT_USER, r"Software\Microsoft\Windows\DWM") as key:
+                    try:
+                        accent_color_dword = winreg.QueryValueEx(key, "ColorizationColor")[0]
+                        accent_color_hex = '#{:06x}'.format(accent_color_dword & 0xffffff)
+                        accent_color = QtGui.QColor(accent_color_hex)
+                    except OSError:
+                        log.warning('Failed reading ColorizationColor from registry')
+                        # accent_color = QtGui.QColor(0, 120, 215)  # Windows 10 default blue
+                        accent_color = QtGui.QColor(119, 27, 133)  # MusicBrainz Picard purple
+
+                palette = QtGui.QPalette(self.palette())
+                palette.setColor(QtGui.QPalette.Window, QtGui.QColor(51, 51, 51))
+                palette.setColor(QtGui.QPalette.WindowText, QtCore.Qt.white)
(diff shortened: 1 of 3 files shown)


phw:
@Gabrielcarvfer Please take a look. I submitted it as is for now, but we can change anything here. In general I think moving the style handling out of tagger.py is a good thing. Not too sure about the actual code. When I wrote this I remember I wasn't that happy with the result, but its not too bad either :D

Gabrielcarvfer:
I think it is more than fine the way it is.

I've posted a few palettes for script syntax highlighting in the other PR. Tried a bunch of different combinations with the color wheel and, so far, my favorite is this one that is a mix between Tango fusion and Darcula. Nothing too bright nor too dark, pastel tones, reasonable contrast between different classes. 

image

image

phw:
I cleaned up the code a bit and added switching the syntax theme if a dark theme is used:

grafik

Don't bother with the editor font looking a bit pixelated, on my Windows there is currently some issue with Qt not loading the proper fonts when run from source. Does not happen with installed Picard. I'll look into this, but it is unrelated to the changes here :)

For future enhancements we could also apply this separate syntax theme on other platforms should we detect a dark theme being used. But this PR focuses on Windows 10.

phw:
@Gabrielcarvfer Could you take a look at this?

Gabrielcarvfer:
Everything looks great. The only thing people may complain is that changing the color accent isn't passed to all components. I would rather pick either MB purple or orange.
image

phw:
Not quite sure how you mean. If some UI elements are not picking up the accent color, does it make a difference if we use the accent color the user configured or we hard code one? In both cases either some elements don't use that color or we have to workaround it. What do you have in mind where the accent color is not used but should be?

Gabrielcarvfer:
> Not quite sure how you mean. If some UI elements are not picking up the accent color, does it make a difference if we use the accent color the user configured or we hard code one? 

I'm not sure either. I bet most people just choose one and stick with it, so it won't be a problem, but I bet someone will report this as a bug. =x

phw:
I guess we will get a bug report anyway :D If we would select a fixed color I bet one of the first things we get as a ticket would be "support Windows accent color". So I'd prefer doing this by default for now and see what feedback we get.

phw:
I played around with this a bit on other platforms, and the general setup here will really also be useful for other platforms. Especially having `BaseTheme.is_dark_theme` is very useful for selecting a better syntax theme also on other platforms.

I tested this with a special implementation on macOS, and there it is definitely one thing missing for proper dark mode support. I have a implementation ready for this.

Also on all other platforms. If you run on Linux Picard might already use a dark appearance, especially if you are using KDE and have configured it to use dark mode. But the syntax theme does not fit. Reimplementing `BaseTheme.is_dark_theme` would solve this. Right now I have not looked into this and have no idea how to detected the use of a dark appearance on Linux.

Gabrielcarvfer:
>Right now I have not looked into this and have no idea how to detected the use of a dark appearance on Linux.

On distros with GNOME, running the command `gsettings get org.gnome.desktop.interface gtk-theme` returns the theme, which have the `:dark` suffix in case of dark themes.

phw:
Merged to gather wider feedback in 2.4 beta release.

PR #1542, opened 2020-05-15
PICARD-1823: Sort genre tags lexicographically

atj:
This commit sorts genres lexicographically so that the order is
deterministic, preventing tag updates due to changes in hash functions
etc.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

[PICARD-1823](https://tickets.metabrainz.org/browse/PICARD-1823)

When opening releases on different platforms I've noticed that the order of genre tags is non-deterministic (probably due to differences in the Python hash function), which causes the genre tag to be needlessly rewritten. 

# Solution

This trivial commit calls sort() on the genre array, ensuring that the order of genres is deterministic.

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -315,6 +315,7 @@ class Track(DataObject, Item):
                 break
             name = _TRANSLATE_TAGS.get(name, name.title())
             genre.append(name)
+        genre.sort()
         join_genres = config.setting['join_genres']
         if join_genres:
             genre = [join_genres.join(genre)]


zas:
Can you please create a ticket (if none exists for this) and link it (PICARD-XXXX: as title's prefix)?

atj:
> Can you please create a ticket (if none exists for this) and link it (PICARD-XXXX: as title's prefix)?

Done.

Sophist-UK:
Personally I don't think a comment is necessary.

PR #1543, opened 2020-05-17
PICARD-1843: Improves load and clustering performance

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Speeds up loading by scanning the entire directory before feeding threads instead of concurrently, preventing starvation of worker threads.
Speeds up clustering and file moving by disabling panel sorting during operations. 

# Problem

Load and clustering are slow, especially with a large library. More details can be found [here](https://github.com/metabrainz/picard/pull/1531).

Partially addresses PICARD-840, PICARD-905, PICARD-1432 and PICARD-1780, but none of them in particular.

* JIRA ticket (_optional_): PICARD-1843

# Solution

1 - Scan entire directory before feeding worker threads, to prevent starvation (around 3x/8x speedup without/with the following changes).
2 - Postpone inclusion of loaded files to the unclustered_files cluster to the tagger.file_loaded section, where it can be included directly to the correct album cluster instead of being moved later, saving up a bunch of time for partially processed libraries.
3 - During movements between cluster/panels, temporarily disable treeview sorting, to speed up inclusion of new entries (100x speedup for 19k files clustering).
Testing with a 19k file library:
Loading 50m->6m (avg. 6 songs/sec -> 53 songs/sec)
Clustering 5h17m->3m (avg. 1 songs/sec -> 106 songs/sec)
Total(loading+clustering) 6h7m->9m

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -187,7 +187,7 @@ class Tagger(QtWidgets.QApplication):
         # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
         # It's a valid reference, but its start() method doesn't work.
         self.thread_pool = QtCore.QThreadPool(self)
-        self.thread_pool.setMaxThreadCount(4)
+        self.thread_pool.setMaxThreadCount(6)
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -187,12 +187,14 @@ class Tagger(QtWidgets.QApplication):
         # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
         # It's a valid reference, but its start() method doesn't work.
         self.thread_pool = QtCore.QThreadPool(self)
+        self.thread_pool.setMaxThreadCount(4)
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
         # expects instant feedback instead of waiting for a long list of
         # operations to finish.
         self.priority_thread_pool = QtCore.QThreadPool(self)
+        self.priority_thread_pool.setMaxThreadCount(1)
 
         # Use a separate thread pool for file saving, with a thread count of 1,
         # to avoid race conditions in File._save_and_rename.
@@ -522,75 +524,48 @@ class Tagger(QtWidgets.QApplication):
             for file in new_files:
                 file.load(partial(self._file_loaded, target=target))
 
-    def add_directory(self, path):
-        if config.setting['recursively_add_files']:
-            self._add_directory_recursive(path)
-        else:
-            self._add_directory_non_recursive(path)
-
-    def _add_directory_recursive(self, path):
-        ignore_hidden = config.setting["ignore_hidden_files"]
-        walk = os.walk(path)
-
-        def get_files():
-            try:
-                root, dirs, files = next(walk)
-                if ignore_hidden:
-                    dirs[:] = [d for d in dirs if not is_hidden(os.path.join(root, d))]
-            except StopIteration:
-                return None
-            else:
-                number_of_files = len(files)
-                if number_of_files:
-                    mparms = {
-                        'count': number_of_files,
-                        'directory': root,
-                    }
-                    log.debug("Adding %(count)d files from '%(directory)r'" %
-                              mparms)
-                    self.window.set_statusbar_message(
-                        ngettext(
-                            "Adding %(count)d file from '%(directory)s' ...",
-                            "Adding %(count)d files from '%(directory)s' ...",
-                            number_of_files),
-                        mparms,
-                        translate=None,
-                        echo=None
-                    )
-                return (os.path.join(root, f) for f in files)
-
-        def process(result=None, error=None):
-            if result:
-                if error is None:
-                    self.add_files(result)
-                thread.run_task(get_files, process)
-
-        process(True, False)
-
-    def _add_directory_non_recursive(self, path):
+    def _scan_dir(self, folders, recursive, ignore_hidden):
         files = []
-        for f in os.listdir(path):
-            listing = os.path.join(path, f)
-            if os.path.isfile(listing):
-                files.append(listing)
-        number_of_files = len(files)
-        if number_of_files:
-            mparms = {
-                'count': number_of_files,
-                'directory': path,
-            }
-            log.debug("Adding %(count)d files from '%(directory)r'" %
-                      mparms)
-            self.window.set_statusbar_message(
-                ngettext(
-                    "Adding %(count)d file from '%(directory)s' ...",
-                    "Adding %(count)d files from '%(directory)s' ...",
-                    number_of_files),
-                mparms,
-                translate=None,
-                echo=None
-            )
-            # Function call only if files exist
+        local_folders = list(folders)
+        while local_folders:
+            current_folder = local_folders.pop(0)
+            current_folder_files = []
+
+            try:
+                for entry in os.scandir(current_folder):
+                    if ignore_hidden and is_hidden(entry.path):
+                        continue
+                    if recursive and entry.is_dir():
+                        local_folders.append(entry.path)
+                    else:
+                        current_folder_files.append(entry.path)
+            except FileNotFoundError:
+                pass
+
+            number_of_files = len(current_folder_files)
+            if number_of_files:
+                mparms = {
+                    'count': number_of_files,
+                    'directory': current_folder,
+                }
+                log.debug("Adding %(count)d files from '%(directory)r'" %
+                          mparms)
+                self.window.set_statusbar_message(
+                    ngettext(
+                        "Adding %(count)d file from '%(directory)s' ...",
+                        "Adding %(count)d files from '%(directory)s' ...",
+                        number_of_files),
+                    mparms,
+                    translate=None,
+                    echo=None
+                )
+                files.extend(current_folder_files)
+        return files
+
+    def add_directory(self, path):
+            files = self._scan_dir([path],
+                                   config.setting['recursively_add_files'],
+                                   config.setting["ignore_hidden_files"])
             self.add_files(files)
 
     def get_file_lookup(self):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -842,12 +842,15 @@ class Tagger(QtWidgets.QApplication):
             files = list(self.unclustered_files.files)
         else:
             files = self.get_files_from_objects(objs)
+
+        self.window.panel.set_sorting(False)
         for name, artist, files in Cluster.cluster(files, 1.0):
             QtCore.QCoreApplication.processEvents()
             cluster = self.load_cluster(name, artist)
             for file in sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')):
                 file.move(cluster)
                 QtCore.QCoreApplication.processEvents()
+        self.window.panel.set_sorting(True)
 
     def load_cluster(self, name, artist):
         for cluster in self.clusters:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -431,7 +431,7 @@ class Tagger(QtWidgets.QApplication):
     def _file_loaded(self, file, target=None):
         self._pending_files_count -= 1
         if self._pending_files_count == 0:
-            self.window.panel.set_sorting(True)
+            self.window.set_sorting(True)
 
         if file is None:
             return
@@ -526,7 +526,7 @@ class Tagger(QtWidgets.QApplication):
         if new_files:
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
-            self.window.panel.set_sorting(False)
+            self.window.set_sorting(False)
             if target is None or target is self.unclustered_files:
                 target = None
             for file in new_files:
@@ -843,14 +843,14 @@ class Tagger(QtWidgets.QApplication):
         else:
             files = self.get_files_from_objects(objs)
 
-        self.window.panel.set_sorting(False)
+        self.window.set_sorting(False)
         for name, artist, files in Cluster.cluster(files, 1.0):
             QtCore.QCoreApplication.processEvents()
             cluster = self.load_cluster(name, artist)
             for file in sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')):
                 file.move(cluster)
                 QtCore.QCoreApplication.processEvents()
-        self.window.panel.set_sorting(True)
+        self.window.set_sorting(True)
 
     def load_cluster(self, name, artist):
         for cluster in self.clusters:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -361,7 +361,7 @@ class Tagger(QtWidgets.QApplication):
         """Move `file` to recording `recordingid` on album `albumid`."""
         self.window.setUpdatesEnabled(False)
         album = self.load_album(albumid)
-        file.move(album.unmatched_files)
+        album.run_when_loaded(partial(file.move, album.unmatched_files))
         album.run_when_loaded(partial(album.match_files, [file],
                                       recordingid=recordingid))
         self.window.setUpdatesEnabled(True)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -343,6 +343,7 @@ class Tagger(QtWidgets.QApplication):
 
     def move_files_to_album(self, files, albumid=None, album=None):
         """Move `files` to tracks on album `albumid`."""
+        self.window.setUpdatesEnabled(False)
         if album is None:
             album = self.load_album(albumid)
         if album.loaded:
@@ -350,6 +351,7 @@ class Tagger(QtWidgets.QApplication):
         else:
             for file in list(files):
                 file.move(album.unmatched_files)
+        self.window.setUpdatesEnabled(True)
 
     def move_file_to_album(self, file, albumid):
         """Move `file` to a track on album `albumid`."""
@@ -357,10 +359,12 @@ class Tagger(QtWidgets.QApplication):
 
     def move_file_to_track(self, file, albumid, recordingid):
         """Move `file` to recording `recordingid` on album `albumid`."""
+        self.window.setUpdatesEnabled(False)
         album = self.load_album(albumid)
         file.move(album.unmatched_files)
         album.run_when_loaded(partial(album.match_files, [file],
                                       recordingid=recordingid))
+        self.window.setUpdatesEnabled(True)
 
     def create_nats(self):
         if self.nats is None:
@@ -371,12 +375,14 @@ class Tagger(QtWidgets.QApplication):
         return self.nats
 
     def move_file_to_nat(self, file, recordingid, node=None):
+        self.window.setUpdatesEnabled(False)
         self.create_nats()
         file.move(self.nats.unmatched_files)
         nat = self.load_nat(recordingid, node=node)
         nat.run_when_loaded(partial(file.move, nat))
         if nat.loaded:
             self.nats.update()
+        self.window.setUpdatesEnabled(True)
 
     def exit(self):
         if self.stopping:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -502,7 +502,7 @@ class Tagger(QtWidgets.QApplication):
     def add_files(self, filenames, target=None, result=None):
         """Add files to the tagger."""
         if result:
-            filenames = result
+            filenames = result   # Handles add_directory task results coming from a worker thread
         ignoreregex = None
         pattern = config.setting['ignore_regex']
         if pattern:
@@ -533,11 +533,9 @@ class Tagger(QtWidgets.QApplication):
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
             self.window.set_sorting(False)
-            if target is None or target is self.unclustered_files:
-                target = None
+            self._pending_files_count += len(new_files)
             for file in new_files:
                 file.load(partial(self._file_loaded, target=target))
-            self._pending_files_count += len(new_files)
 
     def _scan_dir(self, folders, recursive, ignore_hidden):
         files = []
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -278,6 +278,7 @@ class Tagger(QtWidgets.QApplication):
 
         self.browser_integration = BrowserIntegration()
 
+        self._pending_files_count = 0
         self.files = {}
         self.clusters = ClusterList()
         self.albums = {}
@@ -428,6 +429,10 @@ class Tagger(QtWidgets.QApplication):
         return super().event(event)
 
     def _file_loaded(self, file, target=None):
+        self._pending_files_count -= 1
+        if self._pending_files_count == 0:
+            self.window.panel.set_sorting(True)
+
         if file is None:
             return
         
@@ -488,8 +493,10 @@ class Tagger(QtWidgets.QApplication):
         elif isinstance(target, ClusterList):
             self.cluster(files)
 
-    def add_files(self, filenames, target=None):
+    def add_files(self, filenames, target=None, result=None):
         """Add files to the tagger."""
+        if result:
+            filenames = result
         ignoreregex = None
         pattern = config.setting['ignore_regex']
         if pattern:
@@ -519,10 +526,12 @@ class Tagger(QtWidgets.QApplication):
         if new_files:
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
+            self.window.panel.set_sorting(False)
             if target is None or target is self.unclustered_files:
                 target = None
             for file in new_files:
                 file.load(partial(self._file_loaded, target=target))
+            self._pending_files_count += len(new_files)
 
     def _scan_dir(self, folders, recursive, ignore_hidden):
         files = []
@@ -563,10 +572,12 @@ class Tagger(QtWidgets.QApplication):
         return files
 
     def add_directory(self, path):
-            files = self._scan_dir([path],
-                                   config.setting['recursively_add_files'],
-                                   config.setting["ignore_hidden_files"])
-            self.add_files(files)
+            thread.run_task(
+                partial(self._scan_dir, [path],
+                        config.setting['recursively_add_files'],
+                        config.setting["ignore_hidden_files"]),
+                partial(self.add_files, None),
+                traceback=self._debug)
 
     def get_file_lookup(self):
         """Return a FileLookup object."""
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -25,6 +25,7 @@
 # Copyright (C) 2018 Bob Swift
 # Copyright (C) 2018 virusMac
 # Copyright (C) 2019 Joel Lintunen
+# Copyright (C) 2020 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -425,7 +426,11 @@ class Tagger(QtWidgets.QApplication):
         return super().event(event)
 
     def _file_loaded(self, file, target=None):
-        if file is None or file.has_error():
+        if file is None:
+            return
+        
+        if file.has_error():
+            self.unclustered_files.add_file(file)
             return
 
         if target is not None:
@@ -456,6 +461,8 @@ class Tagger(QtWidgets.QApplication):
                           file, recordingid)
                 self.move_file_to_nat(file, recordingid)
                 return
+
+        self.unclustered_files.add_file(file)
 
         # fallback on analyze if nothing else worked
         if config.setting['analyze_new_files'] and file.can_analyze():
@@ -511,7 +518,6 @@ class Tagger(QtWidgets.QApplication):
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
             if target is None or target is self.unclustered_files:
-                self.unclustered_files.add_files(new_files)
                 target = None
             for file in new_files:
                 file.load(partial(self._file_loaded, target=target))
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -435,7 +435,7 @@ class Tagger(QtWidgets.QApplication):
 
         if file is None:
             return
-        
+
         if file.has_error():
             self.unclustered_files.add_file(file)
             return
@@ -572,12 +572,11 @@ class Tagger(QtWidgets.QApplication):
         return files
 
     def add_directory(self, path):
-            thread.run_task(
-                partial(self._scan_dir, [path],
-                        config.setting['recursively_add_files'],
-                        config.setting["ignore_hidden_files"]),
-                partial(self.add_files, None),
-                traceback=self._debug)
+        thread.run_task(partial(self._scan_dir, [path],
+                            config.setting['recursively_add_files'],
+                            config.setting["ignore_hidden_files"]),
+                        partial(self.add_files, None),
+                        traceback=self._debug)
 
     def get_file_lookup(self):
         """Return a FileLookup object."""
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -171,8 +171,8 @@ class MainPanel(QtWidgets.QSplitter):
         super().__init__(parent)
         self.window = window
         self.create_icons()
-        self.views = [FileTreeView(window, self), AlbumTreeView(window, self)]
-        self._selected_view = self.views[0]
+        self._views = [FileTreeView(window, self), AlbumTreeView(window, self)]
+        self._selected_view = self._views[0]
         self._ignore_selection_changes = False
 
         def _view_update_selection(view):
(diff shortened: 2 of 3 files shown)


Sophist-UK:
Just to note that this is a replacement for #1531 where there is a relevant discussion in comments about approach.

zas:
> Testing with a 19k file library:
> Loading 50m->6m (avg. 6 songs/sec -> 53 songs/sec)
> Clustering 5h17m->3m (avg. 1 songs/sec -> 106 songs/sec)
> Total(loading+clustering) 6h7m->9m

Whoaaa. From hours to minutes? You rule. :+1:

picard/cluster.py
```diff
@@ -263,12 +264,15 @@ def clear_lookup_task(self):
...
-    def cluster(files, threshold):
+    def cluster(files, threshold, tagger=None):
         win_compat = config.setting["windows_compatibility"] or IS_WIN
         artist_dict = ClusterDict()
         album_dict = ClusterDict()
         tracks = []
+
+        i = 0
         for file in files:
+            i += 1
```
zas, line 275: You can use `enumerate()`;

```python
for i, file in enumerate(files):
```

picard/cluster.py
```diff
@@ -296,7 +307,9 @@ def cluster(files, threshold):
 
         # Now determine the most prominent names in the cluster and build the
         # final cluster list
+        i = 0
         for album_id, album in albums.items():
```
zas, line 311: `enumerate()` too

picard/cluster.py
```diff
@@ -318,6 +331,13 @@ def cluster(files, threshold):
 
             yield album_name, artist_name, (files[i] for i in album)
```
zas, line 332: `i` is already used in the same function, perhaps rename one of those to `j`, even though it has no impact on execution, it'd be more readable.

picard/tagger.py
```diff
@@ -511,81 +526,60 @@ def add_files(self, filenames, target=None):
...
-        else:
-            self._add_directory_non_recursive(path)
+        Thread(target=self._add_directory, args=[path, config.setting['recursively_add_files'], config.setting["ignore_hidden_files"]]).start()
 
-    def _add_directory_recursive(self, path):
-        ignore_hidden = config.setting["ignore_hidden_files"]
-        walk = os.walk(path)
+    def _add_directory(self, path, recursive, ignore_hidden):
+        if not os.path.exists(path) or not os.path.isdir(path):
+            return
```
zas, line 576: Instead of doing those tests, I'd rather let `_scandir()` raises exceptions, and catch them below. File can be modified between this line and first `scandir()`` call, so it may happen anyway (unlikely, but possible).

picard/tagger.py
```diff
@@ -844,18 +838,45 @@ def autotag(self, objects):
...
+        self.window.panel.set_sorting(False)
+
+        cluster_files = {}
+        for name, artist, files in Cluster.cluster(files, 1.0, self):
             cluster = self.load_cluster(name, artist)
             for file in sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')):
-                file.move(cluster)
-                QtCore.QCoreApplication.processEvents()
+                if cluster not in cluster_files:
+                    cluster_files[cluster] = [file]
```
zas, line 859: You can use a defaultdict here.

```python
from collections import defaultdict

cluster_files = defauldict(list)
for name, artist, files in Cluster.cluster(files, 1.0, self):
    cluster = self.load_cluster(name, artist)
    cluster_files[cluster].extend(sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')))
```

picard/tagger.py
```diff
@@ -844,18 +838,45 @@ def autotag(self, objects):
...
+                else:
+                    cluster_files[cluster].append(file)
+
+        pending_events = []
```
zas, line 863: We could use a `deque` instead of a list for this purpose: https://docs.python.org/3/library/collections.html#collections.deque

picard/ui/itemviews.py
```diff
@@ -284,6 +285,10 @@ def remove(self, objects):
         else:
             self.update_current_view()
 
+    def set_sorting(self, sort=True):
+        self.views[0].setSortingEnabled(sort)
+        self.views[1].setSortingEnabled(sort)
```
zas, line 290: What are `0` and `1` here?

picard/tagger.py
```diff
@@ -511,81 +526,60 @@ def add_files(self, filenames, target=None):
...
-                    "Adding %(count)d file from '%(directory)s' ...",
-                    "Adding %(count)d files from '%(directory)s' ...",
-                    number_of_files),
-                mparms,
-                translate=None,
-                echo=None
-            )
-            # Function call only if files exist
-            self.add_files(files)
+        files = self._scan_dir(ignore_hidden, [path], recursive)
```
zas, line 578: Nitpicking.
Order of arguments in `_scan_dir(ignore_hidden, [path], recursive)` is different from order of arguments in  `_add_directory(self, path, recursive, ignore_hidden)`.

We could make it more consistent with few changes.

picard/ui/itemviews.py
Gabrielcarvfer: Left (clusters) and right (albums) panels

picard/tagger.py
Gabrielcarvfer: OK

picard/cluster.py
Gabrielcarvfer: OK

picard/cluster.py
Gabrielcarvfer: OK

picard/cluster.py
Gabrielcarvfer: OK

picard/tagger.py
Gabrielcarvfer: Is there a good reason to do that? The order doesn't really matter, as the panels will sort them later anyways. I've used it just to separate processing from UI updates.

picard/tagger.py
Gabrielcarvfer: OK

picard/tagger.py
Gabrielcarvfer: OK

picard/tagger.py
Gabrielcarvfer: Oh, now I see. Didn't notice the removal of the inner for. xD

picard/ui/itemviews.py
zas: We should name 2 constants for those, and use them instead of numeric values.

Gabrielcarvfer: How about the following?
```
class Panels(Enum):
    CLUSTERS = 0
    ALBUMS = 1

[...]

def set_sorting(self, sort=True):
    self.views[Panels.CLUSTERS].setSortingEnabled(sort)
    self.views[Panels.ALBUMS].setSortingEnabled(sort)
```

picard/tagger.py
```diff
@@ -844,18 +836,41 @@ def autotag(self, objects):
...
+            for file in files:
+
+                for event in list(pending_events):
+                    if event.isSet():
+                        pending_events.pop(0)
+
+                event = thread.to_main(file.move, cluster)
+                pending_events.append(event)
+
+                if len(pending_events) > 30:
```
zas, line 869: This hardcoded value could be stored in a constant (at top of the file).

Gabrielcarvfer:
I've made a tiny change in `_add_directory` to add files in batches of 100s of files. Completely eliminates the freezing Phillip reported on the previous PR without the different thread setup with queues. :)

picard/tagger.py
```diff
@@ -466,6 +476,10 @@ def move_files(self, files, target):
         if target is None:
             log.debug("Aborting move since target is invalid")
             return
+
+        # Temporarily disable left and right panel sorting
+        self.window.panel.set_sorting(False)
```
zas, line 481: I was wondering about those, they are good candidates for self-made with statement.

I was thinking about be able to write `with SortingDisabled(self.window.panel):`, here is a small test script implementing the idea:

```python
class Obj:
    def __init__(self):
        self.sorting = True

    def set_sorting(self, value):
        self.sorting = value

    def __str__(self):
        return "Sorting: %r" % self.sorting


class SortingDisabled():
    def __init__(self, obj):
        self.obj = obj

    def __enter__(self):
        self.saved_value = self.obj.sorting
        self.obj.set_sorting(False)
        return self.obj

    def __exit__(self, type, value, traceback):
        self.obj.set_sorting(self.saved_value)
        return False

print("basic with statement")
obj = Obj()
print(obj)
with SortingDisabled(obj):
    print(obj)
print(obj)

print("with exception in with statement")
print(obj)
try:
    with SortingDisabled(obj):
        print(obj)
        raise Exception("Exception test raised")
except Exception as e:
    print(e)

print(obj)

print("with original value being False")
obj = Obj()
obj.sorting = False
print(obj)
with SortingDisabled(obj):
    print(obj)
print(obj)
```

What do you think? The idea behind is to ensure sorting value is restored whatever happens in between.

picard/tagger.py
```diff
@@ -511,81 +528,58 @@ def add_files(self, filenames, target=None):
...
-                mparms,
-                translate=None,
-                echo=None
-            )
-            # Function call only if files exist
-            self.add_files(files)
+        # Add files in batches to prevent freezes
+        while files:
+            thread.to_main(self.add_files, files[:100]).wait()
+            files = files[100:]
```
zas, line 582: Perhaps move `100` to constant `NUM_FILES_PER_BATCH` ?

Gabrielcarvfer: OK

picard/tagger.py
```diff
@@ -142,6 +145,7 @@ def _patched_shutil_copystat(src, dst, *, follow_symlinks=True):
 
 _orig_shutil_copystat = shutil.copystat
 shutil.copystat = _patched_shutil_copystat
+max_concurrent_events = 30
```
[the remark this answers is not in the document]
Gabrielcarvfer: Sure.

picard/cluster.py
```diff
@@ -496,7 +512,12 @@ def cluster(self, threshold):
                     c = similarity(token_x, token_y)
                     if c >= threshold:
                         heappush(heap, ((1.0 - c), [x, y]))
-            QtCore.QCoreApplication.processEvents()
+            if tagger and clustering_type:
+                tagger.window.set_statusbar_message(
+                    N_("Clustering - step %d: clustering by %s (%d/%d)" % ({"artist": 2, "album": 3}[clustering_type], clustering_type, y, self.cluster_dict.get_size())),
```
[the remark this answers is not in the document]
Gabrielcarvfer: It surely does.

picard/tagger.py
Gabrielcarvfer: Not really sure on this one. On the current implementation, due to concurrency, the worst thing that can happen is the sorting be enabled during the operation and things get slower (T1 T2 T1). On this one, the worst thing that can happen is the sorting to get disabled and not be re-enabled in the end or future calls. (T1 T2 T1 T2)

picard/ui/itemviews.py
```diff
@@ -256,16 +263,16 @@ def update_selection(self, i, j):
...
-            self.update_selection(0, 1)
+            self.update_selection(Panels.CLUSTERS.value, Panels.ALBUMS.value)
             self._ignore_selection_changes = False
 
-    def update_selection_1(self):
+    def update_selection_albums(self):
         if not self._ignore_selection_changes:
             self._ignore_selection_changes = True
-            self.update_selection(1, 0)
+            self.update_selection(Panels.ALBUMS.value, Panels.CLUSTERS.value)
```
zas, line 275: Finally I'm not very happy with Panels enum, we have a lot of code redundancy in there.
I suggested few changes in https://github.com/Gabrielcarvfer/picard/pull/1

picard/tagger.py
Gabrielcarvfer: If the other thread finishes before the first (e.g. move stuff while clustering is running).
It can be handled with either a semaphore or a lock+counter combo inside set_sorting. I assumed it wouldn't be a problem due to how fast things are, where there is no reason not to wait stuff finish processing before moving/editing/etc. 

I didn't touch on set_sorting during add_files and lookup/scanning for that exact reason, as they're bounded by IO and API, being far slower and having way higher probability of manual intervention (moving).

picard/ui/itemviews.py
```diff
@@ -842,12 +853,6 @@ def __init__(self, obj, sortable, *args):
         self.setTextAlignment(7, QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
         self.setTextAlignment(8, QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
 
-    def __lt__(self, other):
-        if not self.sortable:
-            return False
-        column = self.treeWidget().sortColumn()
-        return self.sortkey(column) < other.sortkey(column)
-
```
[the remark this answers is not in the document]
Gabrielcarvfer: My mistake. I thought sorting was working just fine without it.

picard/tagger.py
Gabrielcarvfer: How about the following?
```
        # Add 1 if sorting is disabled, subtract 1 if sorting is enabled
        with self._sorting_lock:
            self._sorting_count += -1 if sort else 1
            
        # Guarantee counter is >= 0
        self._sorting_count = 0 if self._sorting_count < 0 else self._sorting_count
        
        # If counter == 0, re-enable sorting
        sort = True if self._sorting_count == 0 else False    
        self.views[Panels.CLUSTERS.value].setSortingEnabled(sort)
        self.views[Panels.ALBUMS.value].setSortingEnabled(sort)
```

Should solve all cases. Once sorting is disabled, will remain disabled until all DisableEnable pairs have been called.

picard/ui/itemviews.py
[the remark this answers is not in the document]
Gabrielcarvfer: image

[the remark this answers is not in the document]
Gabrielcarvfer: I've prepared a workaround for this, disabling sorting on tagger._add_directory and re-enabling during tagger._file_loaded. Requires an additional lock, counter for failed/aborted loads, and a dictionary to keep track of pending files being loaded. Would be impossible to do that with the `with` clause alone, as the code that should be enclosed within the `with` block is split across multiple calls (an add_files for each batch + one file_loaded for each file).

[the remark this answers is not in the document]
Gabrielcarvfer: Tested again without __lt__, sorting seems to work fine and way faster. I think my previous test was problematic due to unfinished work on the lock+counter scheme for the sorting.

picard/tagger.py
zas: OK, basically we switch from a boolean to an integer counter, each time a thread disables sorting it stays disabled until the last thread actually re-enables it.

I can't say I'm very enthusiastic, not because of the solution (it should work), but because of the problem.

Enabling/disabling sorting looks messy from start, and it becomes even uglier when concurrency comes into play.

@phw what do you think about this? Can we get rid of sorting on/off or should we improve current stuff so it plays well with concurrency?

picard/ui/itemviews.py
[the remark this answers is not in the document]
zas: Interesting, keep up the good job, it seems promising.

picard/ui/itemviews.py
```diff
@@ -250,40 +269,44 @@ def create_icons(self):
...
+    def set_sorting(self, sort=True):
+        # Add 1 if sorting is disabled, subtract 1 if sorting is enabled
+        with self._sorting_lock:
+            self._sorting_count += -1 if sort else 1
+
+        # Guarantee counter is >= 0
+        self._sorting_count = 0 if self._sorting_count < 0 else self._sorting_count
+
+        # If counter == 0, re-enable sorting
+        sort = True if self._sorting_count == 0 else False
```
zas, line 305: This should be equivalent to:

```python
sort = not self._sorting_count
```

picard/ui/itemviews.py
```diff
@@ -250,40 +269,44 @@ def create_icons(self):
...
         else:
             self.update_current_view()
 
+    def set_sorting(self, sort=True):
+        # Add 1 if sorting is disabled, subtract 1 if sorting is enabled
+        with self._sorting_lock:
+            self._sorting_count += -1 if sort else 1
+
+        # Guarantee counter is >= 0
+        self._sorting_count = 0 if self._sorting_count < 0 else self._sorting_count
```
zas, line 302: This can be replaced by:

```python
self._sorting_count = max(0, self._sorting_count)
```

picard/ui/itemviews.py
```diff
@@ -250,40 +269,44 @@ def create_icons(self):
...
+
+        # If counter == 0, re-enable sorting
+        sort = True if self._sorting_count == 0 else False
+        if self._views[0].isSortingEnabled() != sort:
```
zas, line 306: This test looks weird to me. Why is it needed?

Gabrielcarvfer: It isn't. Just skipping the other calls if the panel sorting matches the sort flag. No need to disable something that already is disabled, or enable something that already is enabled. Didn't really test if it triggers sorting on the second case.

picard/ui/itemviews.py
Gabrielcarvfer: OK

picard/ui/itemviews.py
Gabrielcarvfer: OK

picard/ui/itemviews.py
zas: I'd rather simplify, and get rid of the test.

```python
            for view in self._views:
                view.setSortingEnabled(sort)
```

picard/file.py
```diff
@@ -557,7 +557,6 @@ def move(self, parent):
                 self.parent.remove_file(self)
             self.parent = parent
             self.parent.add_file(self)
-            self.acoustid_update()
```
zas, line 560: Where is this handled now?

Gabrielcarvfer: It isn't. Not really sure if it affected anything. If the analysis was pending, it already got cancelled above. If the analysis already finished, it wouldn't be impacted. On my testing everything worked fine and ended up ignoring it.

picard/ui/itemviews.py
Gabrielcarvfer: It does trigger a re-sorting. Measured the times:

1. Load
- disable empty panel: 0ms
- enable panel with 19k files: 185ms

1. Clustering
- disable panel with 19k files: 0ms
- enable panel with 19k files + 1.9k clusters: 70ms

picard/file.py
Gabrielcarvfer: Faster solution: calling acoustid_update to copy the metadata but skipping update_item.

Gabrielcarvfer:
The latest commit shaves another +-3 minutes (total 9m->6m) by disabling panel focus (preventing user selection) and collapsing clustered/unclustered files by default when set_sorting is called. 

Keeping clusters collapsed naturally reduces draws and speeds things up.

Disabling panel focus still allows people to expand/collapse clusters and click on songs to see the metadata/play/etc, but they can't select/move stuff. This mitigates the evil update_item, which calls isSelected + window.update_selection, etc (which is what causes the slowing down/freezing when a ton of stuff is selected).

Gabrielcarvfer:
Kind of overkill to shave another 2-3m from the original 9m (9m->6-7m), but the latest commit temporarily disables the focusing ability (which prevents selection and mitigates `update_item` calls such as `isSelected` and `window.update_selection`) and collapses the clusters (reducing drawing time) while processing.

Expanding clusters, clicking on songs to display their metadata and play them still works while focus is disabled.

I don't think you're going to like that, but just showing that is possible. The difference between loading times of the same library with and without filesystem cache fell to mere 1-2m.

zas:
With current code Picard consistently fails when I drop a directory with 863 already tagged files (mbid embedded) to cluster pane (and those are automatically moved to album pane).

```
D: 13:48:44,942 file.move:552: Moving <FLACFile '01 Night.flac'> from <Cluster 66fe1907-dfac-405a-98f4-dc50c4d13174 'Unending/Unmatched Files'> to <Track 4d7067a5-b925-4925-b9e7-5e106613cb38 'Night'>
D: 13:48:44,945 file.move:552: Moving <FLACFile '03 Blackwing.flac'> from <Cluster 66fe1907-dfac-405a-98f4-dc50c4d13174 'Unending/Unmatched Files'> to <Track 6b893845-0a0b-485e-9ba4-5ab7c49173eb 'Blackwing'>
  File "./picard/coverart/__init__.py", line 144, in next_in_queue
    provider = next(self.providers)
StopIteration

During handling of the above exception, another exception occurred:

  File "./picard/webservice/__init__.py", line 462, in _process_reply
    self._handle_reply(reply, request)
  File "./picard/webservice/__init__.py", line 451, in _handle_reply
    handler(reply.readAll(), reply, error)
  File "./picard/coverart/__init__.py", line 123, in _coverart_downloaded
    self.next_in_queue()
  File "./picard/coverart/__init__.py", line 160, in next_in_queue
    self.next_in_queue()
  File "./picard/coverart/__init__.py", line 160, in next_in_queue
    self.next_in_queue()
  File "./picard/coverart/__init__.py", line 160, in next_in_queue
    self.next_in_queue()
  File "./picard/coverart/__init__.py", line 164, in next_in_queue
    self.album._finalize_loading(None)
  File "./picard/album.py", line 382, in _finalize_loading
    self.match_files(self.unmatched_files.files)
  File "./picard/album.py", line 558, in match_files
    file.move(target)
  File "./picard/file.py", line 557, in move
    self.parent.remove_file(self)
  File "./picard/cluster.py", line 143, in remove_file
    self._update_related_album(removed_files=[file])
  File "./picard/cluster.py", line 114, in _update_related_album
    self.related_album.update()
  File "./picard/album.py", line 487, in update
    self.item.update(update_tracks)
  File "./picard/ui/itemviews.py", line 948, in update
    item.update(update_album=False)
TypeError: update() got an unexpected keyword argument 'update_album'
Abandon (core dumped)
```

Sophist-UK:
It appears to me that we may be moving away from a stable solution again as minor tweaks are added. I would recommend that we think about regressing those back to a stable solution (if that is possible) and then use that as a base for another PR to tweak it further.

Gabrielcarvfer:
> 
> 
> With current code Picard consistently fails when I drop a directory with 863 already tagged files (mbid embedded) to cluster pane (and those are automatically moved to album pane).
> 
> ```
> D: 13:48:44,942 file.move:552: Moving <FLACFile '01 Night.flac'> from <Cluster 66fe1907-dfac-405a-98f4-dc50c4d13174 'Unending/Unmatched Files'> to <Track 4d7067a5-b925-4925-b9e7-5e106613cb38 'Night'>
> D: 13:48:44,945 file.move:552: Moving <FLACFile '03 Blackwing.flac'> from <Cluster 66fe1907-dfac-405a-98f4-dc50c4d13174 'Unending/Unmatched Files'> to <Track 6b893845-0a0b-485e-9ba4-5ab7c49173eb 'Blackwing'>
>   File "./picard/coverart/__init__.py", line 144, in next_in_queue
>     provider = next(self.providers)
> StopIteration
> 
> During handling of the above exception, another exception occurred:
> 
>   File "./picard/webservice/__init__.py", line 462, in _process_reply
>     self._handle_reply(reply, request)
>   File "./picard/webservice/__init__.py", line 451, in _handle_reply
>     handler(reply.readAll(), reply, error)
>   File "./picard/coverart/__init__.py", line 123, in _coverart_downloaded
>     self.next_in_queue()
>   File "./picard/coverart/__init__.py", line 160, in next_in_queue
>     self.next_in_queue()
>   File "./picard/coverart/__init__.py", line 160, in next_in_queue
>     self.next_in_queue()
>   File "./picard/coverart/__init__.py", line 160, in next_in_queue
>     self.next_in_queue()
>   File "./picard/coverart/__init__.py", line 164, in next_in_queue
>     self.album._finalize_loading(None)
>   File "./picard/album.py", line 382, in _finalize_loading
>     self.match_files(self.unmatched_files.files)
>   File "./picard/album.py", line 558, in match_files
>     file.move(target)
>   File "./picard/file.py", line 557, in move
>     self.parent.remove_file(self)
>   File "./picard/cluster.py", line 143, in remove_file
>     self._update_related_album(removed_files=[file])
>   File "./picard/cluster.py", line 114, in _update_related_album
>     self.related_album.update()
>   File "./picard/album.py", line 487, in update
>     self.item.update(update_tracks)
>   File "./picard/ui/itemviews.py", line 948, in update
>     item.update(update_album=False)
> TypeError: update() got an unexpected keyword argument 'update_album'
> Abandon (core dumped)
> ```

Can you please try it again? I didn't manage to reproduce the issue on Windows nor in Ubuntu. Undid a few changes that could be causing the issue (the ones I've commented previously).

Gabrielcarvfer:
> 
> 
> It appears to me that we may be moving away from a stable solution again as minor tweaks are added. I would recommend that we think about regressing those back to a stable solution (if that is possible) and then use that as a base for another PR to tweak it further.

*Feature creep intensifies* 

Kidding, I'm going to prepare another PR with optimizations for the selection part. There is a ton of stuff to do there (getting selected items list from the panels themselves instead of checking in each individual file and call for a draw to paint the selection of the item). This alone will probably make update_item and metadata box updates with `(different across X items)` way faster.

phw:
Performance indeed is looking very good. Just a few comments from trying out the code (on Linux):

- Closing the application while loading is still in progress the main process does not exit and gets stuck, needs to be killed
- This breaks sorting of number based columns (length, track no. etc.) and in general turns off "natural sorting". If it turns out to be too expensive we should at least have proper sorting of numeric columns implemented. Also we must be aware that removing the natural sorting regresses on PICARD-1260

Also the changes to clustering look unrelated here. Is that extra logging needed?

zas:
> Could you please try it again?

Yup, same, I found out it happens with one specific file (I may share it for you to test, but it's big, do not fit in emails). I'll investigate the issue in details tomorrow.

Sophist-UK:
I have not had experience of turning sorting on and off in Picard, but I have done similar in Excel for the same performance reasons. In case it helps, I think that there may be two things that are needed:

a. Make sure that whenever you turn off automatic sorting, you have an event that turns it back on again. In Excel that is typically late in the synchronous execution of the macro, but in Picard it may need to be asynchronous.

b. When you turn off sorting, you need to remember the existing sort order, and when you turn on automatic sorting you may also need to execute an explicit sort (immediately before or immediately after turning automatic sorting back on) to ensure that the list appears as it should.

Gabrielcarvfer:
>
>     * Closing the application while loading is still in progress the main process does not exit and gets stuck, needs to be killed

Testing again. Thank you for the heads up.

>     * This breaks sorting of number based columns (length, track no. etc.) and in general turns off "natural sorting". If it turns out to be too expensive we should at least have proper sorting of numeric columns implemented. Also we must be aware that removing the natural sorting regresses on [PICARD-1260](https://tickets.metabrainz.org/browse/PICARD-1260)
> 

It is pretty expensive, but it's fine if it had a purpose. I just didn't see the difference while testing. Restoring it then.

> Also the changes to clustering look unrelated here. Is that extra logging needed?

Unrelated? Do you mean offloading it to a thread? I can revert it back, but it freezes the UI during processing. Or maybe the statusbar updates, that I've included it to know how things were going. 2 minutes waiting for processing without any indication seems like nothing is happening. ._.

Gabrielcarvfer:
> 
> 
> > Could you please try it again?
> 
> Yup, same, I found out it happens with one specific file (I may share it for you to test, but it's big, do not fit in emails). I'll investigate the issue in details tomorrow.

I would appreciate if you could send the file. Reproducing the issue would help a lot. :)

Sophist-UK:
> It is pretty expensive, but it's fine if it had a purpose. I just didn't see the difference while testing. Restoring it then.

See my comment above - I am starting to wonder whether you can disable and re-enable automatic sorting when it is an asynchronous task - or if you do whether you need to do a repeating timer-initiated manual sort to simulate automatic sorting with reduced frequency and performance impact.

Gabrielcarvfer:
> 
> See my comment above - I am starting to wonder whether you can disable and re-enable automatic sorting when it is an asynchronous task - or if you do whether you need to do a repeating timer-initiated manual sort to simulate automatic sorting with reduced frequency and performance impact.

My bad. Technically, you can disable. The Qt documentation suggests that the program itself sorts the items and include/update them in a sorted manner, letting sorting disabled by default. Re-enabling works just fine and triggers a sort operation automatically. Triggering the sort periodically isn't really necessary, unless the code stops relying on Qt for sorting.

Gabrielcarvfer:
Loading of files that have already been tagged with their recordingid still is pretty slow, and slows down everything else. The problem is in the call chain `tagger._file_loaded` -> `tagger._move_file_to_track` -> `tagger.load_album`, which usually takes 10s of milliseconds. Still working on it.

Gabrielcarvfer:
Simplified the PR removing event handling, python threads and queues, clustering progress statusbar updates, etc. Didn't manage to speed up more the albums part. I think the network request to get the album metadata/id is what locks the main thread and slows everything down.

Gabrielcarvfer:
Disabling window updates while moving files to albums/clusters (there are a bunch of item updates for each album/cluster update) is very effective, 26s shaved from the original 1m11s. Now the `match_files` part, or resulting updates, seem to be the issue.

picard/tagger.py
```diff
@@ -425,7 +435,15 @@ def event(self, event):
         return super().event(event)
 
     def _file_loaded(self, file, target=None):
-        if file is None or file.has_error():
+        self._pending_files_count -= 1
```
zas, line 438: Was this decrement missing?

Gabrielcarvfer: I've created a new variable to keep track of files being added. Not really sure if File.num_pending_files can replace this.

zas: Ok, no problem.

picard/tagger.py
```diff
@@ -510,82 +532,57 @@ def add_files(self, filenames, target=None):
         if new_files:
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
+            self.window.set_sorting(False)
             if target is None or target is self.unclustered_files:
-                self.unclustered_files.add_files(new_files)
                 target = None
             for file in new_files:
                 file.load(partial(self._file_loaded, target=target))
+            self._pending_files_count += len(new_files)
```
zas, line 540: Should'nt `self._pending_files_count` set before the previous loop? Or even incremented in the loop:

```python
for file in new_files:
    self._pending_files_count += 1
    file.load(partial(self._file_loaded, target=target))
```

picard/tagger.py
```diff
@@ -479,8 +499,10 @@ def move_files(self, files, target):
         elif isinstance(target, ClusterList):
             self.cluster(files)
 
-    def add_files(self, filenames, target=None):
+    def add_files(self, filenames, target=None, result=None):
         """Add files to the tagger."""
+        if result:
+            filenames = result
```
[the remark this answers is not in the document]
Gabrielcarvfer: That's due to `thread.run_task` that returns results of offloaded tasks as `result`. 
Related to 
```def add_directory(self, path):
        thread.run_task(partial(self._scan_dir, [path],
                            config.setting['recursively_add_files'],
                            config.setting["ignore_hidden_files"]),
                        partial(self.add_files, None),
                        traceback=self._debug)```

File: picard/tagger.py (hunk 21 above)
Comment (Gabrielcarvfer) [reply picard/tagger.py:540]: Yup. Fixing.

File: picard/tagger.py [hunk 23]
```diff
@@ -510,82 +532,57 @@ def add_files(self, filenames, target=None):
         if new_files:
             log.debug("Adding files %r", new_files)
             new_files.sort(key=lambda x: x.filename)
+            self.window.set_sorting(False)
             if target is None or target is self.unclustered_files:
```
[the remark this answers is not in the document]
Gabrielcarvfer: Seems to be the case. Both lines can be removed. Unless the target is none in file_loaded, it gets moved to unclustered_files anyway.

Gabrielcarvfer:
Fun fact: the long freeze during loading of files that are matched to album tracks is caused by the album load followed by file moves. Postponing the moves until the album is loaded, `album.when_loaded(file.move, album.unmatched_files)`, virtually eliminates the freeze .

Sophist-UK:
This PR is (IMO) truly excellent work on diagnosing performance issues. However, I do recommend that it is brought to a close soon, once we are confident that the result is stable, with a new PR for further enhancements.

Gabrielcarvfer:
Still trying to squeeze a few more seconds on album loads. 

There's a pretty extreme case reported on the forums of a jazz collection with 3k tracks that currently (tested on this branch), that takes 18 seconds to load (and freezes the screen). 

±8s are spent on asciipunct, ±7s on moves + match files, ±1s on stripping spaces. 

Most of move/match files time is spent changing columns of itemviews to display in the panels. Hard to shave anything else from it.

zas:
@phw: before merging, perhaps create a meta ticket for the changelog

Gabrielcarvfer:
Ticket created and PR updated :)

phw:
Thanks a lot, great work again

Sophist-UK:
Yes - great work by @Gabrielcarvfer !! Looking forward to trying it!! Will wait until next full release.

PR #1544, opened 2020-05-17
PICARD-1830: Clear last status bar message after 3 seconds

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

https://community.metabrainz.org/t/small-change-to-picard-version-2-3-2-when-adding-files/477514

* JIRA ticket (_optional_): PICARD-1830

# Solution

Clear last status bar message after few seconds.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -370,7 +370,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         # _ is defined using builtins.__dict__, so setting it as default named argument
         # value doesn't work as expected
         translate = kwargs.get('translate', _)
-        timeout = kwargs.get('timeout', 0)
+        timeout = kwargs.get('timeout', 3000)
         history = kwargs.get('history', log.history_info)
         if len(args) == 1 and isdict(args[0]):
             # named place holders


picard/ui/mainwindow.py
```diff
@@ -370,7 +370,7 @@ def isdict(obj):
         # _ is defined using builtins.__dict__, so setting it as default named argument
         # value doesn't work as expected
         translate = kwargs.get('translate', _)
-        timeout = kwargs.get('timeout', 0)
+        timeout = kwargs.get('timeout', 3000)
```
mineo, line 373: Please add the new default value to the docstring.

zas:
> IMHO this is not a great solution for multiple reasons:
> 

Any better solution to suggest? This patch is just a quick hack, to start something.

mineo:
The best option would be 2. from my list: add a message "Added ... files from ..." with a timeout when done.

Sophist-UK:
I agree with @mineo that error messages should not be removed - indeed an error message overwritten by another status message should be reinstated when the newer status message is removed by the timer.

For what it is worth, my personal view is that messages should disappear from the status bar after an extended period of (say) 30s or more, and should be replaced by a previous error message, or if no error message then a message that says "Working..." if there are outstanding activities or "Ready." if all background activities are finished.

If I leave Picard open for several hours whilst I am away from the computer, with loads or saves etc. in progress when I walk away, I really only want to know if things are still going or not or if there was some sort of error whilst I was away.

phw:
I have to agree with the commenters above. And I would also prefer mineo's suggestion of having a pair of messages here, with the "finished" message taking care of overwriting the status.

zas:
Not a proper solution, so closing.

PR #1545, opened 2020-05-17
Prevent data loss in case of failed tag saving

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
In case file saving rewrites file tags, a copy of the original file is used. If the saving succeeds, the original file is replaced with the new one. If it fails, the temporary file is deleted.

# Problem

Files can get corrupted during tag saving, due to mutagen/filesystem/network issues or the unexpected interruption of Picard during the saving procedure.

# Solution

Creates a copy of the original file and save the updated tags to it. If the saving succeeds, the new file replaces the original. If the saving fails, the corrupted file is deleted. If Picard suddenly crashes during saving, the original file is kept.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -159,10 +159,10 @@ class File(QtCore.QObject, Item):
             partial(self._loading_finished, callback),
             priority=1)
 
-    def _load_check(self, filename):
+    def _load_check(self, filename, force=False):
         # Check that file has not been removed since thread was queued
         # Don't load if we are stopping.
-        if self.state != File.PENDING:
+        if self.state != File.PENDING and not force:
             log.debug("File not loaded because it was removed: %r", self.filename)
             return None
         if self.tagger.stopping:
@@ -175,7 +175,10 @@ class File(QtCore.QObject, Item):
             from picard.formats.util import guess_format
             # If it fails, force format guessing and try loading again
             file_format = guess_format(filename)
-            if not file_format or type(file_format) == type(self):
+            if force:
+                file_format.clear_pending()
+
+            if not file_format and type(file_format) == type(self):
                 raise
             return file_format._load(filename)
 
@@ -294,8 +297,21 @@ class File(QtCore.QObject, Item):
         if self.tagger.stopping:
             log.debug("File not saved because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
+
         new_filename = old_filename
-        if not config.setting["dont_write_tags"]:
+        # Determine if metadata changed before trying to save tags
+        changed = self.metadata != self.orig_metadata
+        if not config.setting["dont_write_tags"] and changed:
+            # Copy file to memory
+            try:
+                num_bytes = os.stat(old_filename).st_size
+                with open(old_filename, 'rb') as f:
+                    temporary_memory_file = f.read(num_bytes)
+            except FileNotFoundError as e:
+                log.debug("Failed to backup file before saving the new tags: %s", old_filename)
+                raise e
+
+            # Rewrite file tags
             save = partial(self._save, old_filename, metadata)
             if config.setting["preserve_timestamps"]:
                 try:
@@ -304,6 +320,31 @@ class File(QtCore.QObject, Item):
                     log.warning(why)
             else:
                 save()
+
+            # Open saved file to check if it works
+            try:
+                saved_file = File(new_filename)
+                saved_file.clear_pending()
+                saved_metadata = saved_file._load_check(new_filename, True)
+                saved_file.remove()
+                if not saved_metadata:
+                    raise ValueError("Saved file returned empty metadata: %s" % new_filename)
+            except Exception as e:
+                log.debug("Saving file failed with exception '%s'. Restoring original file: '%s'", e, old_filename)
+                # Restore original file
+                with open(new_filename, 'wb') as f:
+                    restore = partial(f.write, temporary_memory_file)
+                    if config.setting["preserve_timestamps"]:
+                        try:
+                            self._preserve_times(new_filename, restore)
+                        except self.PreserveTimesUtimeError as why:
+                            log.warning(why)
+                    else:
+                        restore()
+            finally:
+                # Free cached file memory
+                del temporary_memory_file
+
         # Rename files
         if config.setting["rename_files"] or config.setting["move_files"]:
             new_filename = self._rename(old_filename, metadata)


picard/file.py
```diff
@@ -296,14 +299,39 @@ def _save_and_rename(self, old_filename, metadata):
             return None
         new_filename = old_filename
         if not config.setting["dont_write_tags"]:
-            save = partial(self._save, old_filename, metadata)
+            old_filename_temp = old_filename+".temp"
+
+            # Make a copy to modify
+            shutil.copy(old_filename, old_filename_temp)
```
zas, line 305: So the idea is to work on a copy, which is safer for sure.
But it has drawbacks: it copies the whole file before, and if metadata to be written requires a rewrite, that will be 2 times, it can be awfully slow on network file systems. Alone it deserves an option for the user to choose between safety & performance.

Also `shutil.copy()` will not preserve owner/group/permissions.
Temporary file has to be created with a dedicated method (we can't assume `somefile.temp` doesn't exist).

This code has to be tested on all platforms and various situations: code creating/moving/copying is very likely to have different behaviors in different envs.

phw: Yes, I'm not too happy with this for the performance reasons alone. This for sure can make a huge difference. E.g. in many situations Picard only has to rewrite the tag section. If this is the case it safes a lot of IO, especially for large files such as FLAC. If we make creating copies mandatory this involves a lot more data writing.

Gabrielcarvfer: Isn't `shutil.copy2()` that doesn't carry permissions? Other metadata (dates, etc) can be copied with `shutil.copystat`.
I don't see why we can't assume .temp doesn't exist. A temporary file is a temporary file.

I would never trade data integrity for performance, but I see the point. Better keep the copy in memory, perform the metadata writes in memory, check if works in memory and then dump to the disk as the .temp, and use shutil.move to replace the original. I'm going to work on that.

zas: > I don't see why we can't assume .temp doesn't exist. A temporary file is a temporary file.

Don't assume, ensure it doesn't exist.
Here you just name a file with suffix `.temp` based on old file name, and such file might exist already, this is why all temporary files creation methods are ensuring no such file exists (and try to return a file descriptor to avoid race conditions), and use names to reduce the risk of conflict.



> Isn't `shutil.copy2()` that doesn't carry permissions? Other metadata (dates, etc) can be copied with `shutil.copystat`.

Nope, `shutil.copy2()` will actually copy more "metadata", but none of those are copying user & group information for example (that's the normal behavior on *nix for copies).

So if we want to preserve those you have to copy them over from original file, and hope the program can actually set them (it may not).

https://docs.python.org/3/library/shutil.html#shutil.copy

Gabrielcarvfer: Changed the logic to keep the contents of the file in memory while performing the saving. If the saving fails, rewrite the original contents, and in case it succeeds the memory is freed.

It's pretty hard to further reduce IO/syscalls due to the interface between Picard and mutagen (path instead of buffer/file handler/etc).

picard/file.py
```diff
@@ -296,6 +299,16 @@ def _save_and_rename(self, old_filename, metadata):
             return None
         new_filename = old_filename
         if not config.setting["dont_write_tags"]:
+            # Copy file to memory
+            try:
+                num_bytes = os.stat(old_filename).st_size
+                source = open(old_filename, 'rb')
+                temporary_memory_file = source.read(num_bytes)
+            except FileNotFoundError as e:
+                log.debug("Failed to backup %s before saving the new tags.", old_filename)
```
mineo, line 308: I'm aware Picard's code is not consistent in this, but
```suggestion
                log.debug("Failed to backup '%s' before saving the new tags.", old_filename)
```

picard/file.py
```diff
@@ -296,6 +299,16 @@ def _save_and_rename(self, old_filename, metadata):
             return None
         new_filename = old_filename
         if not config.setting["dont_write_tags"]:
+            # Copy file to memory
+            try:
+                num_bytes = os.stat(old_filename).st_size
```
mineo, line 304: IMHO this would be a bit easier to read (requires less context) by using `new_filename` here and in the rest of this changeset.

picard/file.py
Gabrielcarvfer: I've changed them to be more in line with the other log formats: "something happened with file : $filename".

picard/file.py
Gabrielcarvfer: Made a few changes and should be more clear. :)

picard/file.py
phw: Actually you can pass a file descriptor to mutagen, that works just fine. I even experimented with this in https://github.com/phw/picard/commit/28b36a779a14da8df3a04511c006d21eff13d7ff . The code is a bit hackish at places and I never found the time to finish it, but it works.

Gabrielcarvfer:
I've made a few additional changes, including a check to skip saving tags if current metadata is equal to the old. This should speed up file saving when a bunch of unmodified files are selected along with modified ones.
```
        # Determine if metadata changed before trying to save tags
        changed = self.metadata != self.orig_metadata
        if not config.setting["dont_write_tags"] and changed:
        [...]
```

mineo:
>         changed = self.metadata != self.orig_metadata

That's a bit dangerous since the Metadata class doesn't define an `__eq__` method at the moment, so you can create metadata objects that compare equal, but maybe should not:

```
>>> a = Metadata(images=[], deleted_tags=["c"], a=2, b=3)
>>> b = Metadata(images=[], deleted_tags=["d"], a=2, b=3)
>>> a == b
True
>>> b = Metadata(images=[1], deleted_tags=["d"], a=2, b=3)
>>> a == b
True
>>> b = Metadata(images=[1], deleted_tags=["d"], length=2, a=2, b=3)
>>> a == b
True
```

I'm not sure if that's a problem at the moment (images are saved separately, length is not saved as a tag at all, if a deleted tag clears a tag `==` will no longer be True), but it would be great to make this a bit more explicit by having `Metadata.__eq__` just call `MutableMapping.__eq__` .

Gabrielcarvfer:
I'm surprised with those results. Tried out a ton of different changes and all of them seemed to work just fine.

picard/file.py
zas: > Changed the logic to keep the contents of the file in memory while performing the saving. If the saving fails, rewrite the original contents, and in case it succeeds the memory is freed.

The only really method is to write new file to a temporary file, then do an "atomic" rename. If you store in memory, and the program crashes you'll not be able to rewrite originals.

So to implement this properly:

- create an uniquely named temporary file in the same directory of the target file, keep it open, use its file descriptor
- copy permissions and times from original file before writing any data
- write data to temp file, ensure it is ok (no error + optionally, checksum)
- rename temp file to file "atomically" (if the system allows it, POSIX systems do)

If the program or system or machine crashes during write, a temp file will be left over.
If write ops fail, old file will not be overwritten.
If permissions cannot be set properly, it will error.

Gabrielcarvfer: Absolutely correct on that. Still working on that, but I had to slow down a bit due to work and just wanted to commit the checking thing before I forgot about it.

mineo:
> I'm surprised with those results. Tried out a ton of different changes and all of them seemed to work just fine.

I was not trying to say something breaks with your changeset :-) It just took me a while to figure out what actually happens when Metadata objects are compared for equality - I was expecting the images and length attributes to be taken into account, but at the moment only the key-value pairs from the internal dict are. This is OK (at least I couldn't find a situation where this breaks), but I would prefer that to be made explicit by letting `Metadata.__eq__` call `MutableMapping.__eq__`.

phw:
I think this would solve PICARD-450

Gabrielcarvfer:
Closing this for now. Will work on it when I get some free time again. :)

PR #1548, opened 2020-05-24
PICARD-1831: Mitigate performance impacts of file selection and UI updates during processing

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Mitigate code that forces unnecessary draws during item updates, letting Qt handle the changes. Prevent exponential checks caused by `MainPanel._view_update_selection`, as detailed in https://github.com/metabrainz/picard/pull/1543#issuecomment-633148605 during the removal of an item of the selected list (all moves cause this). 

# Problem

* JIRA ticket (_optional_): PICARD-1831

# Solution

The solution implemented in this PR bypasses the the selection checks by disabling them during processing (e.g. moving things around either manually or during clustering). As soon as the move finishes, the selection can be changed. 

The other change to reduce draws works by removing `if self.isSelected(): [...]`, which updated the screen on the spot if changes were detected and the file is selected. Removing the calls delay the screen update, but drastically reduces draws. This results in speed up of operations such as saving.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -21,6 +21,7 @@
 # Copyright (C) 2016 Suhas
 # Copyright (C) 2016-2017 Sambhav Kothari
 # Copyright (C) 2018 Vishal Choudhary
+# Copyright (C) 2020 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -193,6 +194,9 @@ class MainPanel(QtWidgets.QSplitter):
             File.PENDING: interface_colors.get_qcolor('entity_pending'),
             File.ERROR: interface_colors.get_qcolor('entity_error'),
         })
+
+    def set_processing(self, processing=True):
+        self._ignore_selection_changes = processing
 
     def save_state(self):
         config.persist["splitter_state"] = self.saveState()
@@ -866,8 +870,6 @@ class ClusterItem(TreeItem):
         album = self.obj.related_album
         if self.obj.special and album and album.loaded:
             album.item.update(update_tracks=False)
-        if self.isSelected():
-            TreeItem.window.update_selection()
 
     def add_file(self, file):
         self.add_files([file])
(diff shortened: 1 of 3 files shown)


picard/ui/itemviews.py
```diff
@@ -259,13 +264,15 @@ def update_selection(self, i, j):
     def update_selection_0(self):
         if not self._ignore_selection_changes:
             self._ignore_selection_changes = True
-            self.update_selection(0, 1)
+            if not self._processing:
+                self.update_selection(0, 1)
```
zas, line 268: What about disconnecting signals instead?

Something like:

```python
    def __init__(self, window, parent=None):
        super().__init__(parent)
        self.window = window
        self.create_icons()
        self.views = [FileTreeView(window, self), AlbumTreeView(window, self)]
        self._selected_view = 0
        self._ignore_selection_changes = False
        self.set_processing(False)
```

```python
    def set_processing(self, processing=True):
        self._processing = processing
        if processing:
            self.views[0].itemSelectionChanged.disconnect()
            self.views[1].itemSelectionChanged.disconnect()
        else:
            self.views[0].itemSelectionChanged.connect(self.update_selection_0)
            self.views[1].itemSelectionChanged.connect(self.update_selection_1)
```

```python
    def update_selection_0(self):
        if not self._ignore_selection_changes:
            self._ignore_selection_changes = True
            self.update_selection(0, 1)
            self._ignore_selection_changes = False

    def update_selection_1(self):
        if not self._ignore_selection_changes:
            self._ignore_selection_changes = True
            self.update_selection(1, 0)
            self._ignore_selection_changes = False
```

Gabrielcarvfer: Trying it out, but it does seems to be a better idea (unless the connect/disconnect costs are higher than checking the event and skipping the update_selection).

Gabrielcarvfer: Just realized that I'm dumb. It's being used the same way as self._ignore_selection_changes. Replacing self._processing references to _ignore_selection_changes.

zas: OK :)

phw:
Clean and effective. Let's merge this :)

PR #1552, opened 2020-06-05
Remove frames from compatid3 supported by mutagen >= 1.37

phw:
# Summary

This cleans up compatid3 by removing frames that are natively supported by mutagen >= 1.37 (which is our current minimum supported version). In 1.38 even `GRP1` would be supported, but I think it is ok to stick with 1.37 for now.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -65,9 +65,6 @@ from picard.util.tags import parse_comment_tag
 
 
 id3.GRP1 = compatid3.GRP1
-id3.TCMP = compatid3.TCMP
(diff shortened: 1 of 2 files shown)


rdswift:
Does https://picard.musicbrainz.org/docs/mappings/ need to be updated?

phw:
@rdswift No, this is no functional change. Just removes some code no longer needed because mutagen supports this itself.

PR #1555, opened 2020-06-14
PICARD-1844: Further improve loading and clustering performance

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Clustering is slow due to a multitude of factors, but especially due to screen updates and things blocking the main thread, which handles these updates. Offloading clustering to a worker thread mitigates this problem.

# Problem

Some of the changes meant to https://github.com/metabrainz/picard/pull/1543 ended up being lost while I was reworking the code to remove the Python threads code. This PR contains those changes.

* JIRA ticket (_optional_): PICARD-1844

# Solution

This PR contains the following changes to address the slow clustering issue:

Updating statusbar with the clustering progress, as it takes some time to complete
~Return a Python event when sending callbacks for the main thread, allowing worker threads to wait for the main thread, giving it time to work on the UI~
~Allow for worker thread tasks without callbacks for the main thread~
~Run clustering on a worker thread and slowly send move commands (that update the UI) to the main thread~
Moving files to clusters in batches
Size hints for icon updates and panel height (with uniform row height)

For comparison, the clustering of my 19k file library that originally took ~5h (expanded clusters), was reduced to 10m on the previous PR (collapsed clusters). This PR further reduces the clustering time of the same library to ~3m (collapsed clusters) and ~5m (expanded clusters, 60x speedup over the original). 

# Action

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -35,6 +35,7 @@
 
 
 from collections import defaultdict
+from enum import IntEnum
 from heapq import (
     heappop,
     heappush,
@@ -62,6 +63,7 @@ from picard.util.imagelist import (
     remove_metadata_images,
     update_metadata_images,
 )
+from picard.util.progresscheckpoints import ProgressCheckpoints
 
 from picard.ui.item import Item
 
@@ -266,12 +268,17 @@ class Cluster(QtCore.QObject, Item):
             self.lookup_task = None
 
     @staticmethod
-    def cluster(files, threshold):
+    def cluster(files, threshold, tagger=None):
         win_compat = config.setting["windows_compatibility"] or IS_WIN
         artist_dict = ClusterDict()
         album_dict = ClusterDict()
         tracks = []
-        for file in files:
+        num_files = len(files)
+
+        # 10 evenly spaced indexes of files being clustered, used as checkpoints for every 10% progress
+        status_update_steps = ProgressCheckpoints(num_files, 10)
+
+        for i, file in enumerate(files):
             artist = file.metadata["albumartist"] or file.metadata["artist"]
             album = file.metadata["album"]
             # Improve clustering from directory structure if no existing tags
@@ -284,11 +291,21 @@ class Cluster(QtCore.QObject, Item):
             # For each track, record the index of the artist and album within the clusters
             tracks.append((artist_dict.add(artist), album_dict.add(album)))
 
-        artist_cluster_engine = ClusterEngine(artist_dict)
-        artist_cluster_engine.cluster(threshold)
+            if tagger and status_update_steps.is_checkpoint(i):
+                statusmsg = "Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)"
+                mparams = {
+                    'step': ClusterType.METADATA.value,
+                    'cluster_type': _(ClusterEngine.cluster_type_label(ClusterType.METADATA)),
+                    'update': status_update_steps.progress(i),
+                }
+                tagger.window.set_statusbar_message(statusmsg, mparams)
+                QtCore.QCoreApplication.processEvents()
 
-        album_cluster_engine = ClusterEngine(album_dict)
-        album_cluster_engine.cluster(threshold)
+        artist_cluster_engine = ClusterEngine(artist_dict, ClusterType.ARTIST)
+        artist_cluster_engine.cluster(threshold, tagger)
+
+        album_cluster_engine = ClusterEngine(album_dict, ClusterType.ALBUM)
+        album_cluster_engine.cluster(threshold, tagger)
 
         # Arrange tracks into albums
         albums = {}
@@ -456,9 +473,20 @@ class ClusterDict(object):
         return word, count
 
 
-class ClusterEngine(object):
+class ClusterType(IntEnum):
+    METADATA = 1
+    ARTIST = 2
+    ALBUM = 3
 
-    def __init__(self, cluster_dict):
+
+class ClusterEngine(object):
+    CLUSTER_TYPE_LABELS = {
+        ClusterType.METADATA: N_('Metadata Extraction'),
+        ClusterType.ARTIST: N_('Artist'),
+        ClusterType.ALBUM: N_('Album'),
+    }
+
+    def __init__(self, cluster_dict, cluster_type):
         # the cluster dictionary we're using
         self.cluster_dict = cluster_dict
         # keeps track of unique cluster index
@@ -467,6 +495,14 @@ class ClusterEngine(object):
         self.cluster_bins = {}
         # Index the word ids -> clusters
         self.index_id_cluster = {}
+        self.cluster_type = cluster_type
+
+    @staticmethod
+    def cluster_type_label(cluster_type):
+        return ClusterEngine.CLUSTER_TYPE_LABELS[cluster_type]
+
+    def _cluster_type_label(self):
+        return ClusterEngine.cluster_type_label(self.cluster_type)
 
     def get_cluster_from_id(self, clusterid):
         return self.index_id_cluster.get(clusterid)
@@ -486,12 +522,15 @@ class ClusterEngine(object):
 
         return maxWord
 
-    def cluster(self, threshold):
-
+    def cluster(self, threshold, tagger=None):
         # Keep the matches sorted in a heap
         heap = []
+        num_files = self.cluster_dict.get_size()
 
-        for y in range(self.cluster_dict.get_size()):
+        # 20 evenly spaced indexes of files being clustered, used as checkpoints for every 5% progress
+        status_update_steps = ProgressCheckpoints(num_files, 20)
+
+        for y in range(num_files):
             token_y = self.cluster_dict.get_token(y).lower()
             for x in range(y):
                 if x != y:
@@ -499,14 +538,22 @@ class ClusterEngine(object):
                     c = similarity(token_x, token_y)
                     if c >= threshold:
                         heappush(heap, ((1.0 - c), [x, y]))
-            QtCore.QCoreApplication.processEvents()
 
-        for i in range(self.cluster_dict.get_size()):
-            word, count = self.cluster_dict.get_word_and_count(i)
+            word, count = self.cluster_dict.get_word_and_count(y)
             if word and count > 1:
-                self.cluster_bins[self.cluster_count] = [i]
-                self.index_id_cluster[i] = self.cluster_count
+                self.cluster_bins[self.cluster_count] = [y]
+                self.index_id_cluster[y] = self.cluster_count
                 self.cluster_count = self.cluster_count + 1
+
+            if tagger and status_update_steps.is_checkpoint(y):
+                statusmsg = "Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)"
+                mparams = {
+                    'step': self.cluster_type.value,
+                    'cluster_type': _(self._cluster_type_label()),
+                    'update': status_update_steps.progress(y),
+                }
+                tagger.window.set_statusbar_message(statusmsg, mparams)
+                QtCore.QCoreApplication.processEvents()
 
         for i in range(len(heap)):
             c, pair = heappop(heap)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -43,6 +43,7 @@
 
 
 import argparse
+from collections import defaultdict
 from functools import partial
 from itertools import chain
 import logging
@@ -842,12 +843,13 @@ class Tagger(QtWidgets.QApplication):
             files = self.get_files_from_objects(objs)
 
         self.window.set_sorting(False)
+        cluster_files = defaultdict(list)
         for name, artist, files in Cluster.cluster(files, 1.0, self):
-            QtCore.QCoreApplication.processEvents()
             cluster = self.load_cluster(name, artist)
-            for file in sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')):
-                file.move(cluster)
-                QtCore.QCoreApplication.processEvents()
+            cluster_files[cluster].extend(sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')))
+        for cluster, files in cluster_files.items():
+            cluster.add_files(files)
+            QtCore.QCoreApplication.processEvents()
         self.window.set_sorting(True)
 
     def load_cluster(self, name, artist):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -342,7 +342,6 @@ class Tagger(QtWidgets.QApplication):
 
     def move_files_to_album(self, files, albumid=None, album=None):
         """Move `files` to tracks on album `albumid`."""
-        self.window.setUpdatesEnabled(False)
         if album is None:
             album = self.load_album(albumid)
         if album.loaded:
@@ -350,7 +349,6 @@ class Tagger(QtWidgets.QApplication):
         else:
             for file in list(files):
                 file.move(album.unmatched_files)
-        self.window.setUpdatesEnabled(True)
 
     def move_file_to_album(self, file, albumid):
         """Move `file` to a track on album `albumid`."""
@@ -358,12 +356,10 @@ class Tagger(QtWidgets.QApplication):
 
     def move_file_to_track(self, file, albumid, recordingid):
         """Move `file` to recording `recordingid` on album `albumid`."""
-        self.window.setUpdatesEnabled(False)
         album = self.load_album(albumid)
         album.run_when_loaded(partial(file.move, album.unmatched_files))
         album.run_when_loaded(partial(album.match_files, [file],
                                       recordingid=recordingid))
-        self.window.setUpdatesEnabled(True)
 
     def create_nats(self):
         if self.nats is None:
@@ -374,14 +370,12 @@ class Tagger(QtWidgets.QApplication):
         return self.nats
 
     def move_file_to_nat(self, file, recordingid, node=None):
-        self.window.setUpdatesEnabled(False)
         self.create_nats()
         file.move(self.nats.unmatched_files)
         nat = self.load_nat(recordingid, node=node)
         nat.run_when_loaded(partial(file.move, nat))
         if nat.loaded:
             self.nats.update()
-        self.window.setUpdatesEnabled(True)
 
     def exit(self):
         if self.stopping:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -849,7 +849,7 @@ class Tagger(QtWidgets.QApplication):
             files = self.get_files_from_objects(objs)
 
         self.window.set_sorting(False)
-        for name, artist, files in Cluster.cluster(files, 1.0):
+        for name, artist, files in Cluster.cluster(files, 1.0, self):
             QtCore.QCoreApplication.processEvents()
             cluster = self.load_cluster(name, artist)
             for file in sorted(files, key=attrgetter('discnumber', 'tracknumber', 'base_filename')):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -482,6 +482,7 @@ class Tagger(QtWidgets.QApplication):
         if target is None:
             log.debug("Aborting move since target is invalid")
             return
+        self.window.set_sorting(False)
         if isinstance(target, (Track, Cluster)):
             for file in files:
                 file.move(target)
@@ -494,6 +495,7 @@ class Tagger(QtWidgets.QApplication):
             self.move_files_to_album(files, album=target)
         elif isinstance(target, ClusterList):
             self.cluster(files)
+        self.window.set_sorting(True)
 
     def add_files(self, filenames, target=None, result=None):
         """Add files to the tagger."""
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -187,7 +187,6 @@ class Tagger(QtWidgets.QApplication):
         # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
         # It's a valid reference, but its start() method doesn't work.
         self.thread_pool = QtCore.QThreadPool(self)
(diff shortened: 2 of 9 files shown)


Gabrielcarvfer:
Additional 1.42x load speedup (from 10m down to 7m on my test library) lost in the rework of the previous PR.

Gabrielcarvfer:
> Actually on my Linux system the clustering becomes slower with this patch, having 4090 files loaded. The clustering algorithm itself runs fast (so fast that the status bar messages don't really matter), but then moving the files to the clusters is slow. On current master the entire process is finished after 59 s, but with this branch it takes 107 s. Probably something strange again with the threading.

That's pretty weird. Booting into Ubuntu to test it again, on WSL it is working as expected.

> Talking about the status messages: I don't think those detailed messages are a good idea. They basically spam the activity log. In my case the clustering algorithm takes 500 ms, but in this time spews out 4585 messages. If it is some use during debugging I would rather have them as debug messages instead. But even in this case it probably would be sufficient to only log the steps with the total number of processings being done.

Changing to percentages. 10%/5% steps for clustering steps 1 and 2/3, maxing at 50 messages.
Processing alone usually takes +-2 mins for 18k files, it's pretty weird not having any king of feedback. Maybe disabling the clustering button while processing would suffice.

> During clustering I also get this message twice:
> 
> ```
>   File "./picard/tagger.py", line 430, in event
>     func()
>   File "./picard/tagger.py", line 500, in move_files
>     file.move(target)
>   File "./picard/file.py", line 559, in move
>     self.parent.add_file(self)
>   File "./picard/cluster.py", line 131, in add_file
>     self.add_files([file])
>   File "./picard/cluster.py", line 125, in add_files
>     self.item.add_files(files)
> AttributeError: 'NoneType' object has no attribute 'add_files'
> ```

Hmm, didn't see this during debugging on Ubuntu nor on Windows, but I suspect this was the reason for `QtCore.QCoreApplication.processEvents` during clustering.

picard/cluster.py
```diff
@@ -467,6 +485,10 @@ def __init__(self, cluster_dict):
         self.cluster_bins = {}
         # Index the word ids -> clusters
         self.index_id_cluster = {}
+        self.cluster_type = cluster_type
+
+    def cluster_type_label(self):
+        return ClusterEngine.CLUSTER_TYPE_LABELS[self.cluster_type]
```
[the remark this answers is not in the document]
Gabrielcarvfer: Not really sure how to deal with that. I've tried adding the new entries in picard.pot, translated entries into pt_BR.po, build it, but didn't work.

Gabrielcarvfer:
Super weird, but checking perf data the slow part related to setting icons is that Qt tries to guess the row width/height to fit the image every single update. Setting up a size hint reduced that overhead. 

```
COLUMN_ICON_SIZE = 16
COLUMN_ICON_BORDER = 2
IconSize = QtCore.QSize(COLUMN_ICON_SIZE+COLUMN_ICON_BORDER, 
                        COLUMN_ICON_SIZE+COLUMN_ICON_BORDER)
[...]
self.setSizeHint(MainPanel.TITLE_COLUMN, IconSize)
```

Disabling the screen updates like I suggested on the previous commit doesn't make much more difference. Instead, I've removed the limit of worker threads.

Load times in both Windows/Ubuntu dropped another 1m-1m30s. (5m30s on Windows without cached files vs 2m40s with cached files, and a lot of freezes)

phw:
> but I suspect this was the reason for QtCore.QCoreApplication.processEvents during clustering.

Should not be. The reason for this call was responsiveness. If you have some kind of longer processing you can either run it on a separate thread or run it on the main thread but call QCoreApplication.processEvents regularly to allow the UI to update. For some cases this is better. If you run in a thread you should not call it.

The errors being thrown seems to be related to certain files or something other specific case. I'll try to debug this here.

picard/cluster.py
```diff
@@ -456,9 +465,19 @@ def get_word_and_count(self, index):
...
+class ClusterType(IntEnum):
+    ARTIST = 2
+    ALBUM = 3
+
+
 class ClusterEngine(object):
 
-    def __init__(self, cluster_dict):
+    CLUSTER_TYPE_LABELS = {
+        ClusterType.ARTIST: 'artist',
```
zas, line 476: You need to mark strings here for translation:

```python
ClusterType.ARTIST: N_('artist'),
ClusterType.ALBUM: N_('album'),
```

picard/cluster.py
```diff
@@ -499,7 +522,15 @@ def cluster(self, threshold):
...
-            QtCore.QCoreApplication.processEvents()
+            if tagger and (y % status_update_steps) == 0:
+                tagger.window.set_statusbar_message(
+                    N_("Clustering - step %d/3: clustering by %s (%d/%d)" % (self.cluster_type.value,
+                                                                             self.cluster_type_label(),
+                                                                             5*y/status_update_steps,
+                                                                             100)),
+                    {},
+                    timeout=0.01
+                )
```
zas, line 533: This isn't used properly. Not sure I get it correctly but it should be something like:

```python
status_msg = N_("Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)")
mparms = {
    'step': self.cluster_type.value,
    'cluster_type': _(self.cluster_type_label()),
    'update': 5*y/status_update_steps,
}
tagger.window.set_statusbar_message(status_msg, mparms)
```

Explanation: by default, `set_statusbar_message` will call `_()` on passed message format, `mparms` is a dict used to fill named fields (better to use those for translators).
`100` is hardcoded, and doesn't need any translation, since we have `/100` I guess it's a percentage, hence `%%` which is the escaped form of `%`.
I removed 'clustering by' because it would complexify translation (you need to translate full strings, you can't just concatenate, because not all languages work this way), so this simpler form avoids this (but if you insist, I can explain further).

picard/tagger.py
```diff
@@ -489,11 +483,11 @@ def move_files(self, files, target):
         if isinstance(target, (Track, Cluster)):
             for file in files:
                 file.move(target)
-                QtCore.QCoreApplication.processEvents()
+            QtCore.QCoreApplication.processEvents()
         elif isinstance(target, File):
             for file in files:
                 file.move(target.parent)
-                QtCore.QCoreApplication.processEvents()
+            QtCore.QCoreApplication.processEvents()
```
zas, line 490: Why isn't this called for others `if` branches? I mean since it isn't in the loops, it could be called once after the last `elif`

picard/ui/itemviews.py
```diff
@@ -859,6 +862,7 @@ def __init__(self, obj, sortable, *args):
         self.setTextAlignment(1, QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
         self.setTextAlignment(7, QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
         self.setTextAlignment(8, QtCore.Qt.AlignRight | QtCore.Qt.AlignVCenter)
+        self.setSizeHint(MainPanel.TITLE_COLUMN, ICON_SIZE)
```
zas, line 865: How does this related to current patch?

phw:
So the latest version of the patch is just amazing from a performance perspective. The clustering now took only 5s (compared to the 59s on master, see above. Also the exceptions I mentioned above no longer happen.

There are some other things not quite right now, though:

- Moving files back to "Unclustered" now completely blocks the UI. As this is terrible slow (also on current master, takes several minutes for my ~4k files) the lack of UI update is really an issue. We don't need to have this optimized in this PR, but maybe some calls to `QCoreApplication.processEvents` would help allowing the UI to update during this.
- Once you move files back to unclustered you cannot cluster them again.

picard/cluster.py
Gabrielcarvfer: Yup, it was supposed to be a percentage but I didn't manage to print it. 
'Clustering by' removal is fine, it was duplicated with the first word anyway. 
I've tried your suggestion, but replaced lowercased type labels with the already existing 'Artist' and 'Album'.

Gabrielcarvfer:
Undid a bunch of changes, especially around the clustering part. The size hint alone plus the moves in batches during clustering did the trick. Clustering with expanded clusters fell to 2m14s (150x speedup over the original). 

Not sure if the translation part is done correctly, but it worked here.

Working on moving clustered files back to unclustered files.

picard/ui/itemviews.py
Gabrielcarvfer: Technically this thing is the key to responsiveness and performance. 35% of CPU time was wasted guessing the icon size during each file update. This single line prevents that.
https://github.com/metabrainz/picard/pull/1555#issuecomment-644475419

picard/tagger.py
Gabrielcarvfer: I didn't really knew why it was there, so I thought it was fine. Phillipp made me realize that it is a good idea to keep it where it was.

picard/ui/itemviews.py
Sophist-UK: Is there anywhere else in Picard where we need to do this?

Gabrielcarvfer: Nope. Just a single line. The hinted size is used in all updates of the items.

Other place that called for size hints was the rowheight to determine the vertical scrollbar size. Mitigated by setting uniform row height.

The last place that can use this are the horizontal scrollbars, that try to guess the width based on column contents. Way harder to mitigate that.

Gabrielcarvfer:
The weird edge case with the ultimate jazz collection (3k tracks) is reasonably hard to solve. Computing the 168 discs in parallel won't work as the data is tied to the UI, only ways to solve is either delaying things or doing them faster. Do you guys have anything against cython? :)

picard/ui/itemviews.py
Gabrielcarvfer: On the _clone side of things (QtThreads), most of the time is spent calculating the coverart hash (10% of CPU time) and decoding the file metadata (22% of CPU time).
perf3

Gabrielcarvfer: Tried replacing the md5 with a fletcher variant (Addler32), the hash column is gone but didn't seem to affect much. Ubuntu seems to be bottlenecked by the disk drivers (ntfs-3g). Tried loading from a sata SSD on Windows, but nothing changed. 

I assume the bottleneck has moved not to the CPU time, nor to the disk, but wait due to syscalls.

phw:
Tested the latest code, and it's looking great. Moving back to unclustered now is responsive and much faster then in the original code. Clustering remains to be very fast.

I haven't looked at the code in detail yet, will do this later.

> Do you guys have anything against cython? :)

Nope, absolutely not. I mean we already have for example the C implementation for astrcmp, so I have no objections to also compile other CPU intensive bits. And I personally would prefer using cython over writing directly in C. And the unicode conversions are likely a good candidate.

picard/cluster.py
```diff
@@ -499,7 +531,15 @@ def cluster(self, threshold):
                     c = similarity(token_x, token_y)
                     if c >= threshold:
                         heappush(heap, ((1.0 - c), [x, y]))
-            QtCore.QCoreApplication.processEvents()
+            if tagger and (y % status_update_steps) == 0:
+                statusmsg = _("Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)")
+                mparams = {
+                    'step': self.cluster_type.value,
+                    'cluster_type': _(self._cluster_type_label()),
+                    'update': 5 * y // status_update_steps,
```
zas, line 539: A minor issue, got 110% :)

```
2020-06-17 10:59:48,137 - Clustering - step 1/3: Metadata Extraction (110%)
```

picard/util/textencoding.py
```diff
@@ -181,19 +181,17 @@ def unicode_simplify_compatibility(string):
...
 def unicode_simplify_punctuation(string, pathsave=False, win_compat=False):
-    return ''.join(
-        _replace_unicode_simplify_punctuation(c, pathsave, win_compat) for c in string)
+    temp = []
+    for c in string:
+        try:
+            result = _simplify_punctuation[c]
+        except KeyError:
+            result = c
+        if c != result and pathsave:
```
zas, line 191: If `KeyError` was raised, we are 100% `c == result` so I think you can move this in the try part.

```python
        try:
            result = _simplify_punctuation[c]
            if c != result and pathsave:
                result = sanitize_filename(result, win_compat=win_compat)
        except KeyError:
            result = c
```

Sophist-UK:
@Gabrielcarvfer If there are no more performance improvements to be made, it looks like your work is done here. And a big WELL DONE!!!!!

That said, and I haven't researched this at all...

> On the _clone side of things (QtThreads), most of the time is spent calculating the coverart hash (10% of CPU time, I think md5 can be replaced with fletcher + value check, just like in ZFS) and decoding the file metadata (22% of CPU time).

1. A hash is a hash - faster hashing would be good.

2. We should definitely try to find a C-based JSON decoder, ideally one which doesn't lock the GIL during execution.

Gabrielcarvfer:
It's basically done at this point, but I'm still puzzled on why moving to a SSD didn't change anything.

> 1. A hash is a hash - faster hashing would be good.

But different hashes are used for different purposes depending on what you want to accomplish and the application threshold for collision rates. Fletcher have very high collision rates when compared to cryptographic hashes.

Even then, if wall clock times doesn't even budge, I think maintaining the original makes moresense.

> 2. We should definitely try to find a C-based JSON decoder, ideally one which doesn't lock the GIL during execution.

Hmm, based on the instrumentation, decoding requests is virtually irrelevant on the big picture. The network requests themselves are not, and this is supposedly handled by Qt code. Probably related to the caching stuff, or maybe it keeps opening/closing ports like there's no tomorrow, etc. Not sure yet.

Gabrielcarvfer:
> Nope, absolutely not. I mean we already have for example the C implementation for astrcmp, so I have no objections to also compile other CPU intensive bits. And I personally would prefer using cython over writing directly in C. And the unicode conversions are likely a good candidate.

I will create a new branch with the changes for you to take a look. It didn't really affect performance (wall clock wise), but did create debugging symbols which are invaluable when using perf/Performance Recorder on Linux/Windows.

There are two big limitations: 
- to cythonize an entire package, there can't be code in _init_ files. So I ended up creating a library per file.
- any file with the translation built-ins can't be properly translated and compiled, so no cluster/tagger/album/file.py.

picard/util/textencoding.py
Gabrielcarvfer: OK

picard/cluster.py
Gabrielcarvfer: Haha, that is unexpected. I've seen 105%, but thought it was due to rounding error

Gabrielcarvfer: Oh, I think I get it. Must be due to status_update_steps == 1 when the number of files < 10 or 5.

Gabrielcarvfer: It isn't the most elegant solution, nor guarantee the 100%, but 110% is no more.

Gabrielcarvfer:
Hmm, tested the version with Cython and Addler32 on Windows again. Seems live I've forgotten to compile it. Load times with file cache in memory actually dropped from ~2m30s to ~1m13s. Without caches remained the same.

Gabrielcarvfer:
Another proof the bottleneck is elsewhere: cprofile says config.get_item by itself is super expensive.  
image

Cprofile indicate huge savings by using memoization
image

Looking at loading finished times, this looks like a major change
image

At wall clock, it is the exact same thing.

Also, removing the `value` lock greatly improves the responsiveness on Windows.

picard/cluster.py
```diff
@@ -499,7 +531,15 @@ def cluster(self, threshold):
                     c = similarity(token_x, token_y)
                     if c >= threshold:
                         heappush(heap, ((1.0 - c), [x, y]))
-            QtCore.QCoreApplication.processEvents()
+            if tagger and y in status_update_steps:
+                statusmsg = _("Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)")
+                mparams = {
+                    'step': self.cluster_type.value,
+                    'cluster_type': _(self._cluster_type_label()),
+                    'update': 100*(y+1)//self.cluster_dict.get_size(),
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

picard/cluster.py
```diff
@@ -499,7 +531,15 @@ def cluster(self, threshold):
                     c = similarity(token_x, token_y)
                     if c >= threshold:
                         heappush(heap, ((1.0 - c), [x, y]))
-            QtCore.QCoreApplication.processEvents()
+            if tagger and y in status_update_steps:
+                statusmsg = _("Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)")
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

picard/cluster.py
```diff
@@ -486,10 +518,10 @@ def get_cluster_title(self, cluster):
 
         return maxWord
 
-    def cluster(self, threshold):
-
+    def cluster(self, threshold, tagger=None):
         # Keep the matches sorted in a heap
         heap = []
+        status_update_steps = set([int(i*self.cluster_dict.get_size()/20) for i in range(1, 21)])
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done. Looks pretty weird, but it is harder than I thought to calculate unique and evenly spaced integer points.

picard/cluster.py
```diff
@@ -284,11 +287,21 @@ def cluster(files, threshold):
             # For each track, record the index of the artist and album within the clusters
             tracks.append((artist_dict.add(artist), album_dict.add(album)))
 
-        artist_cluster_engine = ClusterEngine(artist_dict)
-        artist_cluster_engine.cluster(threshold)
+            if tagger and i in status_update_steps:
+                statusmsg = _("Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)")
```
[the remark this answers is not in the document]
Gabrielcarvfer: OK.

picard/cluster.py
```diff
@@ -266,12 +267,14 @@ def clear_lookup_task(self):
...
 
     @staticmethod
-    def cluster(files, threshold):
+    def cluster(files, threshold, tagger=None):
         win_compat = config.setting["windows_compatibility"] or IS_WIN
         artist_dict = ClusterDict()
         album_dict = ClusterDict()
         tracks = []
-        for file in files:
+        status_update_steps = set([int(i*len(files)/10) for i in range(1, 11)])
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

Gabrielcarvfer:
Trying to reduce the scope of the lock. Assuming that config variables are mostly constants and only changed before/after processing, I would remove that lock. I've changed the behavior to use the cached version, and if not available or out-of-date, use the standard raw_value path.

picard/cluster.py
```diff
@@ -486,22 +521,33 @@ def get_cluster_title(self, cluster):
...
 
-    def cluster(self, threshold):
-
+    def cluster(self, threshold, tagger=None):
         # Keep the matches sorted in a heap
         heap = []
+        num_files = self.cluster_dict.get_size()
+
+        # 20 evenly spaced indexes of files being clustered, used as checkpoints for every 5% progress
+        status_update_steps = set([int(i*num_files/20) for i in range(1, 21)])
```
zas, line 530: Since it is used twice in same file, make it a small function ?
Something like:

```python
def progress_checkpoints(num_files, each_n_percent=10):
    """"Create a set of evenly spaced indexes of files being clustered, used as checkpoints for progress""""
    num_checkpoints = int(100/each_n_percent)
    return set([int(i*num_files/num_checkpoints) for i in range(1, num_checkpoints+1)])


    status_update_steps = progress_checkpoints(num_files, each_n_percent=5)
```

Or even a small class:

```python
class ProgressCheckpoints:

    def __init__(self, num_files, each_n_percent=10):
        """"Create a set of evenly spaced indexes of files being clustered, used as checkpoints for progress""""
        num_checkpoints = int(100/each_n_percent)
        self._checkpoints = set([int(i*num_files/num_checkpoints) for i in range(1, num_checkpoints+1)])

    def is_checkpoint(self, index):
      return index in self._checkpoints

```

```python
status_update_steps = ProgressCheckpoints(num_files, each_n_percent=20)

if status_update_steps.is_checkpoint(y): ...

```

It can then moved to utils/ or smt to be reused elsewhere

picard/cluster.py
```diff
@@ -486,22 +521,33 @@ def get_cluster_title(self, cluster):
...
+                statusmsg = "Clustering - step %(step)d/3: %(cluster_type)s (%(update)d%%)"
+                mparams = {
+                    'step': self.cluster_type.value,
+                    'cluster_type': _(self._cluster_type_label()),
+                    'update': 100*(y+1)//num_files,
+                }
+                tagger.window.set_statusbar_message(statusmsg, mparams)
+                QtCore.QCoreApplication.processEvents()
+
+        for i in range(num_files):
```
zas, line 550: Cannot this loop merged with previous one?

Gabrielcarvfer: Yup. Just didn't want to change how the code was organized. I didn't even look into what those loops do.

picard/cluster.py
Gabrielcarvfer: OK

Gabrielcarvfer: Ended up improving the logic. :D

Sophist-UK:
@Gabrielcarvfer You have done a great job - you have scientifically investigated the performance issues shown by profiling and doggedly produced improvements to the point that no further improvements can be made - and effectively worked a miracle.

All kudos to you!!!!

Gabrielcarvfer:
Technically you all have made the heavy lifting, I just made a few adjustments here and there. Thank you for Picard and MB. :)

Sophist-UK:
That my friend is the magic of open source!!!!

Sophist-UK:
These changes are IMO so significant as to justify a special @Gabrielcarvfer release of Picard.

PR #1559, opened 2020-06-19
PICARD-1840: Fix possible crash when unsetting lyricist

phw:
# Summary
This is a bugfix for https://github.com/metabrainz/picard/pull/1556, which would crash if there is no lyricist set for a work marked as instrumental.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1840

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -390,7 +390,10 @@ def recording_to_metadata(node, m, track=None):
     if m.length:
         m['~length'] = format_time(m.length)
     if 'instrumental' in m.getall('~performance_attributes'):
-        m.unset('lyricist')
+        try:
+            m.unset('lyricist')
+        except KeyError:
+            pass
         m['language'] = 'zxx'
 
 


phw:
@zas Maybe up for discussion: When writing this I actually did not expect unset could fail. We could think about changing Metadata.unset for this. Are there any real downsides? Other operations here are also very forgiving. E.g. both `metadata.get('unknown')`  and `metadata.delete('unknown')` will both work, but `metadata.unset('unknown')` will raise.

I'm kind of unsure of this, but I tend to drop this PR again and change Metadata.unset instead.

Sophist-UK:
@phw I think you should fix `unset` - IMO an exception should only be raised if the result cannot be determined. In this case `unset` of a known or unknown key should have the same result - no stored value, so this exception is not of any value.

phw:
Yep, agreed. Dropping this PR in favor of https://github.com/metabrainz/picard/pull/1560

PR #1562, opened 2020-06-19
PICARD-1847: Move tag compatibility options to separate views

phw:
# Summary
Restructure the options page Tags > Tags Compatibility into multiple option pages.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1847

# Solution
picard-options-tags-multi

# Action


zas:
image

Should those 2 options which relates to RIFF and ID3 moved to format-specific pages?

phw:
I had thought about this as well. But those are rather options about relating to two different formats. If we would have multiple FLAC options we could add that option removing ID3 there, I think that would fit better. But as it is now I would keep those options on the main tags options page

PR #1564, opened 2020-06-20
PICARD-1753: Fixed script font not scaling on Windows

phw:
# Summary
Fix using the proper monospaced font on Windows in correct size.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The script editor font on Windows is rendered smaller then elsewhere. Also sometimes it picks a font which renders pixelated. Same was true for the log view.

* JIRA ticket (_optional_): PICARD-1753

# Solution

Explicitly define a font based on operating system. The generic approach did not work well on Windows and chose a less optimal font. Also explicitly setting the font overwrote the font size on Windows to be way too small.

picard-editor-font-fixed

I also tried other ways getting a decent default font, but it did not work reliable. E.g. ` const QFont fixedFont = QFontDatabase::systemFont(QFontDatabase::FixedFont)`   gave a very bad result on Linux, and about the same mess as before on Windows.

On macOS the situation is unchanged and ok.

# Action

### DIFF
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -42,7 +42,7 @@ from picard.util import restore_method
 
 if IS_MACOS:
     FONT_FAMILY_MONOSPACE = 'Menlo'
-if IS_WIN:
+elif IS_WIN:
     FONT_FAMILY_MONOSPACE = 'Courier'
 else:
     FONT_FAMILY_MONOSPACE = 'Monospace'
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -33,7 +33,19 @@ from PyQt5 import (
 )
 
 from picard import config
+from picard.const.sys import (
+    IS_MACOS,
+    IS_WIN,
+)
 from picard.util import restore_method
+
+
+if IS_MACOS:
+    FONT_FAMILY_MONOSPACE = 'Menlo'
+if IS_WIN:
+    FONT_FAMILY_MONOSPACE = 'Courier'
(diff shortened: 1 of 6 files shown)


picard/ui/__init__.py
```diff
@@ -33,9 +33,21 @@
...
+from picard.const.sys import (
+    IS_MACOS,
+    IS_WIN,
+)
 from picard.util import restore_method
 
 
+if IS_MACOS:
+    FONT_FAMILY_MONOSPACE = 'Menlo'
+if IS_WIN:
```
zas, line 45: Shouldn't be `elif` ? Because on MacOS, `FONT_FAMILY_MONOSPACE` will be set to `Monospace` in `else` part. Or I miss something.

Sophist-UK: Looks like it should be `elif` to me too.

phw: Yes, absolutely. I'll change that. 

If we have this set here explicitly, we also could remove the workaround of setting a font substitution on macOS with `QtGui.QFont.insertSubstitution('Monospace', 'Menlo')` (see https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L231). We only need one or the other. I tend to remove the font substitution and rely on the explicit font definition, what do you think?

Btw, I also tried the font substitution trick on Windows and it did not deliver good results. The font getting used looked horrible. Hence the explicit definition.

PR #1565, opened 2020-06-21
PICARD-1849: add album_save_post_processors to plug-in API.

thomasvs:
# Fixes https://tickets.metabrainz.org/browse/PICARD-1849

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

I wanted to extend the M3U playlist plugin to automatically generate an .m3u8
when Albums get saved.

To be able to do so, I needed to add album_save_post_processors so that I could
get access to an album on save.

* JIRA ticket (_optional_): PICARD-1849

# Solution

File save is done in a thread, so I needed to implement a way to wait for those
threads to finish by using the file_save_post_processors hook.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -99,7 +99,10 @@ from picard.const.sys import (
 )
 from picard.dataobj import DataObject
 from picard.disc import Disc
-from picard.file import File, register_file_post_save_processor
+from picard.file import (
+    File,
+    register_file_post_save_processor,
+)
 from picard.formats import open_ as open_file
 from picard.i18n import setup_gettext
 from picard.pluginmanager import PluginManager
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -76,6 +76,7 @@ from picard.album import (
     Album,
     NatAlbum,
     run_album_post_removal_processors,
+    run_album_post_save_processors,
 )
 from picard.browser.browser import BrowserIntegration
 from picard.browser.filelookup import FileLookup
@@ -98,7 +99,7 @@ from picard.const.sys import (
 )
 from picard.dataobj import DataObject
 from picard.disc import Disc
-from picard.file import File
+from picard.file import File, register_file_post_save_processor
 from picard.formats import open_ as open_file
 from picard.i18n import setup_gettext
 from picard.pluginmanager import PluginManager
@@ -293,6 +294,10 @@ class Tagger(QtWidgets.QApplication):
         # Load release version information
         if self.autoupdate_enabled:
             self.updatecheckmanager = UpdateCheckManager(parent=self.window)
+
+        # track all file save completions for album save
+        self._album_saves = []
+        register_file_post_save_processor(self._file_post_save)
 
     def register_cleanup(self, func):
         self.exit_cleanup.append(func)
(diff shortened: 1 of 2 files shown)


picard/tagger.py
```diff
@@ -658,9 +666,14 @@ def get_files_from_objects(self, objects, save=False):
 
     def save(self, objects):
         """Save the specified objects."""
-        files = self.get_files_from_objects(objects, save=True)
-        for file in files:
-            file.save()
+        for o in objects:
```
zas, line 669: Please use `obj` instead of `o`

picard/tagger.py
```diff
@@ -773,6 +786,17 @@ def remove(self, objects):
         if files:
             self.remove_files(files)
 
+    # run album_save hooks when the last file is saved
+    def _file_post_save(self, file):
+        for album, files in self._album_saves:
+            if file in files:
```
zas, line 792: Since `file` will not be found in empty `files`:

```python
if not files:
    ...
elif file in files:
   ...
```

picard/tagger.py
```diff
@@ -658,9 +666,14 @@ def get_files_from_objects(self, objects, save=False):
 
     def save(self, objects):
         """Save the specified objects."""
-        files = self.get_files_from_objects(objects, save=True)
-        for file in files:
-            file.save()
+        for o in objects:
+            log.debug('save object %r', o)
+            files = self.get_files_from_objects([o, ], save=True)
```
zas, line 671: This defeats uniquification (?) implemented in https://github.com/metabrainz/picard/blob/7b41e1283b4fd740e1fe4fb3933961c72c4d489f/picard/tagger.py#L663-L665

picard/tagger.py
```diff
@@ -773,6 +786,17 @@ def remove(self, objects):
...
+    # run album_save hooks when the last file is saved
+    def _file_post_save(self, file):
+        for album, files in self._album_saves:
+            if file in files:
+                files.remove(file)
+            if not files:
+                log.debug("Album %r saved, running post_save_processors", album)
+                run_album_post_save_processors(album)
+                self._album_saves.remove((album, files))
+                return
```
zas, line 798: I'm not sure to understand why we return immediatly here

picard/tagger.py
phw: Yes, this can potentially lead to the same file saved twice: You can have both the album and the files selected for saving, but saving should only happen once.

Also we need to be clear what is expected to happen:

1. If I select the album and save, should the album_save_post_processor be run? (I assume this is a clear YES)
2. If I select all files of an album, but not the album itself, should the album_save_post_processor be run?
3. If I select only some files of an album, but not the album itself, should the album_save_post_processor be run?

I tend to say yes to all of the above. But for case 3 it might be useful if the registered processor gets some information, that only a subset has been saved. Optimal would be to pass a list of actually saved files to the processor, so it could figure out what to do with this based on the use case.

phw:
I'm closing this now, as the plugin system was rewritten for version 3. Hence this needs to be updated to version 3, ideally in a new PR. Also this PR has pending changes requested without update for a long time.

The main problem with such a hook is still that there is not really the concept of saving the album. And saving could potentially be started multiple times by the user.

PR #1567, opened 2020-06-23
PICARD-1852: Move about info into standalone dialog

phw:
# Summary
Make the about info a standalone dialog, separate from options, as discussed in #1564

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The about info is not really part of the options, and it interferes a bit with the "remember last options page" functionality added in #1564

* JIRA ticket (_optional_): PICARD-1852

# Solution
grafik

# Action


zas:
image

On my system dialog is too small vertically, so a scroll bar is displayed. Not much an issue, but we could perhaps shorten it by displaying libs list as a comma-separated list.

phw:
I made the dialog a bit larger by default and made the component version be separated by comma. Now it also fits on my Linux setup (which looked before like on your screenshot). With my changes:

grafik

I intentionally left the dialog to show scrollbars if the content doesn't fit, so it doesn't break or blow up the dialog to unusable size in translated versions. E.g. here is a screenshot of the German dialog with a long list of translator credits (which should probably be cleaned up a bit visually):

grafik

picard/ui/aboutdialog.py
```diff
@@ -67,7 +66,7 @@ def load(self):
         else:
             args["translator_credits"] = ""
 
-        args['third_parties_versions'] = '<br />'.join(["%s %s" %
+        args['third_parties_versions'] = ', '.join(["%s %s" %
```
zas, line 69: Actually we could replace spaces in version strings by non-breaking spaces so it wraps more nicely.

```python
        args['third_parties_versions'] = ', '.join(
            map(lambda s: s.replace(' ', '&nbsp;'),
                ["%s %s" % (versions.version_name(name), value) for name, value in versions.as_dict(i18n=True).items()
                 if name != 'version']
                )
        )
```

With approximately the same width, without:

image

With this change:

image

phw: Good idea. I coded it a bit differently and also replace `-` with a non-breaking hyphen. Otherwise python-libdiscid could break like

> Discid discid 1.1.0 (compat layer from python‑
> libdiscid 1.1)

picard/ui/aboutdialog.py
```diff
@@ -87,8 +89,7 @@ def load(self):
 <a href="https://www.flaticon.com/authors/smashicons">Smashicons</a>
 from <a href="https://www.flaticon.com">www.flaticon.com</a>""")
 
-        text = _("""<p align="center"><span style="font-size:15px;font-weight:bold;">MusicBrainz Picard</span><br/>
-Version %(version)s</p>
+        text = _("""<p align="center">Version %(version)s</p>
```
zas, line 92: @phw I think we should explode this string in smaller bits, since everytime a little change happens it needs to be re-translated.
See https://github.com/phw/picard/pull/4

phw: Yes, this is a big improvement, thanks a lot.

PR #1568, opened 2020-06-23
PICARD-1853: Fix natsort crashing on strings with '\0'

phw:
# Summary
Fix crash when loading files where tag values contain null characters, if those tags are used for column sorting.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
`natsort.natkey` will crash on strings which contain `\0` chars.

```
(env) ~\devel\musicbrainz\picard [PICARD-1852-about-dialog]> python .\tagger.py
W: 16:20:30,547 ui.mainwindow.create_actions:529: CDROM: discid library not found - Lookup CD functionality disabled
  File ".\picard\ui\itemviews.py", line 872, in __lt__
    return self.sortkey(column) < other.sortkey(column)
  File ".\picard\ui\itemviews.py", line 877, in sortkey
    return natsort.natkey(self.text(column).lower())
  File ".\picard\util\natsort.py", line 36, in natkey
    return [int(s) if s.isdigit() else strxfrm(s)
  File ".\picard\util\natsort.py", line 36, in <listcomp>
    return [int(s) if s.isdigit() else strxfrm(s)
ValueError: embedded null character
```

* JIRA ticket (_optional_): PICARD-1853

# Solution
Remove null chars from strings in natkey

# Action

### DIFF
--- a/picard/util/natsort.py
+++ b/picard/util/natsort.py
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 #
-# Copyright (C) 2019 Philipp Wolfer
+# Copyright (C) 2019-2020 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


phw:
Notably this crash happens only on Windows, which is interesting as the basic behavior of strxfrm raising if a null char is in the string is the same. But for some reason reading the files seems to be different.

PR #1569, opened 2020-06-23
PICARD-1854: Improve sort performance

phw:
# Summary
Improve performance of column sorting in main view.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1854

# Solution
- Calculate the sort key only once
- Do not use natsort for all columns

In current master having loaded around 27k files changing the sort order (either sorting by a new column or switching from ascending to descending order) takes slightly above 2s. With this patch only the initial sorting is slow (but still a bit faster, as it optimizes that the sortkey needs to be calculated twice for each comparison), later sort order changes are 0.5 to clearly under 1s.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -175,6 +175,15 @@ class MainPanel(QtWidgets.QSplitter):
     LENGTH_COLUMN = _column_indexes['~length']
     FINGERPRINT_COLUMN = _column_indexes['~fingerprint']
 
+    NAT_SORT_COLUMNS = [
+        _column_indexes['title'],
+        _column_indexes['album'],
+        _column_indexes['discsubtitle'],
+        _column_indexes['tracknumber'],
+        _column_indexes['discnumber'],
+        _column_indexes['catalognumber'],
+    ]
+
     def __init__(self, window, parent=None):
         super().__init__(parent)
         self.window = window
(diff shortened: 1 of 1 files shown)


picard/ui/itemviews.py
```diff
@@ -860,21 +874,38 @@ def __init__(self, obj, sortable, *args):
...
+    def setText(self, column, text):
+        self._sortkeys[column] = None
+        return super().setText(column, text)
+
     def __lt__(self, other):
         if not self.sortable:
             return False
         column = self.treeWidget().sortColumn()
-        return self.sortkey(column) < other.sortkey(column)
+        return self.sortkey(column) > other.sortkey(column)
```
zas, line 894: Intentional change?

phw: It was, but I have to recheck this again and I probably should not have mixed this in here. On my Windows system the sort order was not matching the indication (assuming arrow down means ascending, arrow up descending order). So I suspected this was an oversight or regression. But now on my Linux laptop all looks fine. I'll revert this here and investigate the sort order difference separately. Maybe it is intentional and just how it works on Windows.

phw: I rmeoved this change from the PR

PR #1571, opened 2020-06-24
Script doc fixes

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/script.py
+++ b/picard/script.py
@@ -1296,7 +1296,8 @@ def func_swapprefix(parser, text, *prefixes):
 @script_function(check_argcount=False, documentation=N_(
     """`$delprefix(text,*prefixes="a","the")`
 
-Deletes the specified prefixes from the beginning of text. If no prefix is specified 'A' and 'The' are used by default.
+Deletes the specified `prefixes` from the beginning of `text`.
+If no prefix is specified 'A' and 'The' are used by default.
 
 _Since Picard 1.3_"""
 ))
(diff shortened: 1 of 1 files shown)


picard/script.py
```diff
@@ -1165,7 +1182,8 @@ def func_is_complete(parser):
 @script_function(documentation=N_(
     """`$firstalphachar(text,nonalpha="#")`
 
-Returns the first character of `text`. If `text` is not an alphabetic character `nonalpha` is returned instead.
+Returns the first character of `text`.
+If `text` is not an alphabetic character `nonalpha` is returned instead, if not passed `#`.
```
phw, line 1186: I find the part ", if not passed #" hard to understand. What about "..., which defaults to #" or maybe more explicit "If nonalpha is not passed, # is used."?

picard/script.py
```diff
@@ -1375,16 +1388,14 @@ def func_ne_any(parser, x, *args):
 
 Returns `text` in title case (first character in every word capitalized).
 
+Example:
+
+    $set(album,$title(%album%))
 _Since Picard 2.1_"""
-))
+), debug_doc=True)
```
phw, line 1395: Is debug_doc set by purpose?

zas: Nope, leftover from experiment, I'll clean this.

picard/script.py
zas: I addressed this and other issues, rebased.

zas:
I just removed useles backquotes on an example

rdswift:
I see that you've changed 'len' to 'length' for clarity.  Is it worthwhile to also change 'num' to 'number'?

zas:
> I see that you've changed 'len' to 'length' for clarity. Is it worthwhile to also change 'num' to 'number'?

Yes, good suggestion, done in c7477c1

PR #1575, opened 2020-07-02
PICARD-1864: Reorder move to target

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Postpone move to target to make sure MBIDs can be checked first.

# Problem

* JIRA ticket (_optional_): PICARD-1864

During refactoring to reduce moves between clusters, I didn't pay enough attention to target!=none case of file_loaded (which doesn't happen unless you're moving things around from file browser), skipping necessary checks.

# Solution

Reorder the move to target and remove return to make sure all files are tested for 'analyze_new_files' and 'ignore_file_mbids'.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -433,10 +433,6 @@ class Tagger(QtWidgets.QApplication):
             self.unclustered_files.add_file(file)
             return
 
-        if target is not None:
-            self.move_files([file], target)
-            return
-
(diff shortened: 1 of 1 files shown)


phw:
Thanks a lot for the fix

PR #1576, opened 2020-07-03
PICARD-1867: Fix guess_format fallback on file load error

phw:
# Summary
Properly handle trying an alternative file format on file load errors by injecting the guessed file format into the file load queue. Avoids issues with the loaded file object being the wrong type and having two file objects causing the pending files to be displayed wrongly.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The implementation of trying alternatve file format in File._load_check has some flaws.

For one, it just returns the result of a different format, but the original `File` object stays the same. So you e.g. get a MP3 file called "mp3.wav" properly loaded because `WAVFile._load_check()`  will return the result of `MP3File._load()`, but the original file object will still be of type `WAVFile`.

Also the creation of the additional file object will increase the pending files count, but this never gets released.

* JIRA ticket (_optional_): PICARD-1867

# Solution
Handle retrying to load later. Instead of trying to resolve this in place, remove the currently loaded file object and inject the alternative format as a separate file into the load queue.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -183,7 +183,12 @@ class File(QtCore.QObject, Item):
 
             # If loading failed, force format guessing and try loading again
             from picard.formats.util import guess_format
-            alternative_file = guess_format(self.base_filename)
+            try:
+                alternative_file = guess_format(self.filename)
+            except (FileNotFoundError, OSError):
+                log.error("Guessing format of %s failed", self.filename, exc_info=True)
+                alternative_file = None
+
             if alternative_file:
                 # Do not retry reloading exactly the same file format
                 if type(alternative_file) != type(self):  # pylint: disable=unidiomatic-typecheck
--- a/picard/file.py
+++ b/picard/file.py
@@ -168,16 +168,7 @@ class File(QtCore.QObject, Item):
         if self.tagger.stopping:
             log.debug("File not loaded because %s is stopping: %r", PICARD_APP_NAME, self.filename)
             return None
-        try:
-            # Try loading based on extension first
-            return self._load(filename)
-        except Exception:
-            from picard.formats.util import guess_format
-            # If it fails, force format guessing and try loading again
-            file_format = guess_format(filename)
-            if not file_format or type(file_format) == type(self):
-                raise
-            return file_format._load(filename)
+        return self._load(filename)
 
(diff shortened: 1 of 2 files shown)


phw:
@Gabrielcarvfer Could you also take a look at this and review the code?

picard/file.py
```diff
@@ -189,6 +180,19 @@ def _loading_finished(self, callback, result=None, error=None):
         if error is not None:
             self.error = str(error)
             self.state = self.ERROR
+
+            # If loading failed, force format guessing and try loading again
+            from picard.formats.util import guess_format
+            alternative_file = guess_format(self.base_filename)
```
Gabrielcarvfer, line 186: I think 'self.filename' is the correct field (absolute path).

picard/tagger.py
```diff
@@ -527,6 +527,10 @@ def add_files(self, filenames, target=None, result=None):
                 file.load(partial(self._file_loaded, target=target))
                 QtCore.QCoreApplication.processEvents()
 
+    def add_file(self, file_, callback):
+        self._pending_files_count += 1
```
Gabrielcarvfer, line 531: If this is only used in case the extension load fails, than there is no need to increment the pending files count.

phw: Indeed. Looks like we might have a bug here anyway, because in case of error the callback in tagger.py might never get called and the counter never decreases. I'll take a look.

phw: See changes in https://github.com/metabrainz/picard/pull/1576/commits/656ac2d097c4d4c5d9a01e7ffcd7ec751afbfd9b, https://github.com/metabrainz/picard/pull/1576/commits/df3c4c941d186fd62d93f7b008369169be2050cf and https://github.com/metabrainz/picard/pull/1576/commits/c303343892105804ff672e3ac6a10a6a1d15c9da

Gabrielcarvfer: Seems fine. Going to test it.

Gabrielcarvfer: Didn't catch anything wrong while testing. :)

phw:
Great, I merged it. Thanks a lot to both of you for review and testing.

PR #1577, opened 2020-07-03
PICARD-1868: Translate the cover art type "Unknown"

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The cover art type selection dialog does not translate "Unknown".

* JIRA ticket (_optional_): PICARD-1868

### DIFF
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -365,7 +365,7 @@ class CAATypesSelectorDialog(PicardDialog):
         self.list_ignore.clear()
         for caa_type in CAA_TYPES:
             name = caa_type['name']
-            title = translate_caa_type(caa_type['title'])
+            title = translate_caa_type(name)
             item = QtWidgets.QListWidgetItem(title)
             item.setData(QtCore.Qt.UserRole, name)
             if name in includes:


picard/coverart/utils.py
```diff
@@ -41,8 +41,10 @@
 
 
 def translate_caa_type(name):
-    if name == 'unknown':
-        return _(CAA_TYPES_TR[name])
+    print("translate_caa_type", name, CAA_TYPES_TR.get(name))
+    name_lower = name.lower()
```
zas, line 45: This shouldn't be needed...

https://github.com/metabrainz/picard/commit/259e176144c10fd82e662591849ec6ba698ad4b1#diff-4246654415c5ccfdafff4d7925531509R360

I think this last change was incorrect, it should be:

```python
title = translate_caa_type(name)
```

`translate_caa_type()` takes the name of the CAA type, especially to avoid this. But I guess we overlooked the change in rdswift's patch.

`CAA_TYPES_TR` is a dict matching CAA name (always in English, and lower-case) to the English CAA title.

phw: Indeed, you are right. It's the call being wrong. I updated the PR.

PR #1579, opened 2020-07-03
PICARD-1866: Restore cheaper update_selection for itemview updates

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Restore itemview calls to MainWindow.update_selection, fixing coverart updates on the metadatabox. 

# Problem

While trying to improve performance with large libraries, I ended up removing MainWindow.update_selection calls, which are exponential. Further analysis shown that the expensive part of the call is the comparison of tag values of selected files and respective updates of the metadatabox.

* JIRA ticket (_optional_): PICARD-1866

# Solution

Calls to MainWindow.update_selection are restored in itemview updates, but using a flag to indicate the selection remains the same to skip tag comparison during updates. This solves part of the reported issues, while still requiring the user to manually change the file selection to trigger the expensive updates.

This deviates from the previous behavior but ensures that performance is kept basically constant throughout processing even when a ton of files are selected.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -920,6 +920,8 @@ class ClusterItem(TreeItem):
         album = self.obj.related_album
         if self.obj.special and album and album.loaded:
             album.item.update(update_tracks=False)
+        if self.isSelected():
+            TreeItem.window.update_selection(new_selection=False)
 
     def add_file(self, file):
         self.add_files([file])
@@ -994,6 +996,8 @@ class AlbumItem(TreeItem):
                 self.setToolTip(MainPanel.TITLE_COLUMN, _("Album unchanged"))
         for i, column in enumerate(MainPanel.columns):
             self.setText(i, album.column(column[1]))
+        if selection_changed:
+            TreeItem.window.update_selection(new_selection=False)
         # Workaround for PICARD-1446: Expand/collapse indicator for the release
         # is briefly missing on Windows
         self.emitDataChanged()
@@ -1077,6 +1081,8 @@ class TrackItem(TreeItem):
             self.setText(i, track.column(column[1]))
             self.setForeground(i, color)
             self.setBackground(i, bgcolor)
+        if self.isSelected():
+            TreeItem.window.update_selection(new_selection=False)
         if update_album:
             self.parent().update(update_tracks=False)
 
@@ -1098,7 +1104,8 @@ class FileItem(TreeItem):
             if not tree_widget.itemWidget(self, MainPanel.FINGERPRINT_COLUMN):
                 tree_widget.setItemWidget(self, MainPanel.FINGERPRINT_COLUMN,
                     FingerprintColumnWidget(file=file))
-
+        if self.isSelected():
+            TreeItem.window.update_selection(new_selection=False)
         parent = self.parent()
         if isinstance(parent, TrackItem) and update_track:
             parent.update(update_files=False)
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1112,7 +1112,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.tags_from_filenames_action.setEnabled(bool(files))
         self.track_search_action.setEnabled(have_objects)
 
-    def update_selection(self, objects=None):
+    def update_selection(self, objects=None, new_selection=True):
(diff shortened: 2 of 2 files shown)


Sophist-UK:
As I have pointed out before the issue with the metadata summary code is that it iterates over all selected albums and tracks every time you change the selection - and that is VERY CPU intensive.

IMO it needs primarily needs an algorithm change - to maintain a metadata summary per album and to use those when the entire album (all tracks) are part of the selection rather than iterate over all the tracks in the album every time. This should have c. an order of magnitude reduction in CPU time. (On reflection, caching these values might be better so that you invalidate the cache on a metadata change in the album rather than recalculate the summary every time you make a metadata change.)

Once that has been done, it might be possible to move the summarisation into a worker thread in order to ensure that the GUI stays responsive.

Gabrielcarvfer:
> IMO it needs primarily needs an algorithm change - to maintain a metadata summary per album and to use those when the entire album (all tracks) are part of the selection rather than iterate over all the tracks in the album every time. This should have c. an order of magnitude reduction in CPU time. (On reflection, caching these values might be better so that you invalidate the cache on a metadata change in the album rather than recalculate the summary every time you make a metadata change.)

The hard part is updating only the bare essentials. Update_selection doesn't know exactly what changed since the last call.
The metadatabox counters cache can be invalidated with the new flag. Caching per album/cluster seems a good idea, but not sure if it really is necessary. Does anyone really expect a selection to keep being updated as processing progresses? What do they gain from that? The coverart or fields not being updated is problematic.

> Once that has been done, it might be possible to move the summarisation into a worker thread in order to ensure that the GUI stays responsive.

It is done in a worker thread, but that keep restarting the accounting as this gets called frequently (single song update triggers the album update, which trigger updates of all files inside the album, all of them calls update_selection if selected). It's a huge chain reaction.

Sophist-UK:
> Caching per album/cluster seems a good idea, but not sure if it really is necessary. Does anyone really expect a selection to keep being updated as processing progresses? What do they gain from that? The coverart or fields not being updated is problematic.
The big issue comes when you select hundreds or thousands of albums e.g. to do a mass Save, But typically you have selected albums individually before, so they may already be summarised.

I am not sure how easy it would be to program, but perhaps we could just have a delayed timer after any metadata change to summarise an album.

> It is done in a worker thread, but that keep restarting the accounting as this gets called frequently (single song update triggers the album update, which trigger updates of all files inside the album, all of them calls update_selection if selected). It's a huge chain reaction.
Ah - last time I looked at this - which was in 1.4 so several years ago - it was still on the main thread.

phw:
So this works of course for the immediate issue reported (the cover art view not updating after save). But the cover art is only a specific kind of metadata, and the rest of the view not updating to reflect the saved state is still irritating.

I'd say we go with this for now as an improvement and merge it, but looks like we still need to figure out how to optimize the metadata aggregation.

I actually wonder how useful the metadata aggregation is for some cases. It is for sure when selecting a release or multiple files of a single release. And maybe if you have specifically selected certain files, because you want to compare them. When you have selected hundreds or thousands of files it most often is probably not useful (just showing "different for x items" messages). There are exceptions of course, e.g. you want to add a certain tag to your entire collection, but these are specific cases.

But maybe we should limit the automatic aggregation and update to specific cases, maybe a certain number of selected files. If you have selected more we show a message in the metadatabox together with a button ("Edit tags of selected items") to actually aggregate and show the tags.

The common use case is that people select all files to save them, but not to edit the tags. The aggregation is most of the time not useful here, but caused a performance issue.

Another thing: Even the updating of only the cover art as implemented here might get problematic if we actually implement PICARD-1828 (update cover art for an arbitrary number of selected files). So probably the entire metadatabox, including the cover art, would benefit from the above behavior of disabling automated update in some cases.

Gabrielcarvfer:
How about also updating if the selection is a single album/track/cluster/file? Should solve the issue for most cases.

phw:
@Gabrielcarvfer Yes, that should help in many cases.

Gabrielcarvfer:
Question: how exactly are tag changes forwarded from items to their clusters? Tracks/albums are working fine with the patch in the same branch, but clusters are still bugged. =x

Gabrielcarvfer:
Nevermind, it was just that 'Album artist' field changes the 'Artist' field of clusters.

PR #1580, opened 2020-07-06
PICARD-1866: Update metadatabox if a single file/cluster/track/album metadata changes.

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Update metadatabox more frequently to reflect user changes after editing a single file/cluster/track/album.

# Problem

* JIRA ticket (_optional_): PICARD-1866

In PR https://github.com/metabrainz/picard/pull/1579 part of the issue in PICARD-1866 was solved, but it still required users to click on other items and then click again on the edited item to update its tags. 

# Solution

This PR lets Picard continue updating the metadata box in case a single file/cluster/track/album is selected, which should solve the issue for what is assumed to be the common use case. If more than a cluster/album are selected, the metadatabox keeps its original value during updates to prevent reprocessing the tags. If the file/track selection changes, the metadatabox is updated.

# Action

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -453,12 +453,14 @@ class MetadataBox(QtWidgets.QTableWidget):
     def update(self):
         if self.editing:
             return
+        new_selection = False
         if self.selection_dirty:
+            new_selection = True
             self._update_selection()
-        thread.run_task(self._update_tags, self._update_items,
+        thread.run_task(partial(self._update_tags, new_selection), self._update_items,
             thread_pool=self.tagger.priority_thread_pool)
 
-    def _update_tags(self):
+    def _update_tags(self, new_selection=True):
         self.selection_mutex.lock()
         files = self.files
         tracks = self.tracks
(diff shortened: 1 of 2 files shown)


picard/ui/metadatabox.py
```diff
@@ -467,6 +469,21 @@ def _update_tags(self):
...
+            single_file = len(files) == 1
+            single_track = len(tracks) == 1
+            if single_file or single_track:
+                break
+
+            # Or if we are dealing with a single cluster/album
+            single_file_album = len(set([file.metadata["album"] for file in files])) == 1
+            single_track_album = len(set([track.metadata["album"] for track in tracks])) == 1
+            if single_file_album or single_track_album:
+                break
```
zas, line 484: I think we can avoid some calculations rewriting it:

```python
        while not new_selection:  # Just an if with multiple exit points
            # If we are dealing with the same selection
            # skip updates unless it we are dealing with a single file/track
            if len(files) == 1:
                break
            if len(tracks) == 1:
                break
            # Or if we are dealing with a single cluster/album
            single_file_album = len(set([file.metadata["album"] for file in files])) == 1
            if single_file_album:
                break
            single_track_album = len(set([track.metadata["album"] for track in tracks])) == 1
            if single_track_album:
                break
            return self.tag_diff
```

I didn't profile though.

Gabrielcarvfer: We sure can. I'm going to change it.

Gabrielcarvfer:
Still having weird results for clusters, but not completely sure if expected.

<details>
<summary>
Before changing artist of a cluster
</summary>



</details>

<details>
<summary>
After changing artist of a cluster
</summary>



</details>

<details>
<summary>
After changing album artist of a cluster
</summary>

image
</details>

picard/ui/metadatabox.py
Gabrielcarvfer: I've made a few changes to cache these results and it is a bit faster. It looks kinda bad to me style-wise, but not sure how you, @phw or @Sophist-UK would do it.

PR #1581, opened 2020-07-10
Increase contrast between focused and unfocused highlighted items while using accent colors on Windows

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Change color of inactive highlighted items to differentiate them from active highlighted items.

# Problem

Active and inactive highlighted items currently use the same color, confusing users as reported in [link](https://community.metabrainz.org/t/picard-2-4-beta-2-selection-focus-issue/484898).

# Solution

Differentiate active and inactive color for highlighted items, deriving the color from the selected accent.
```
            if dark_theme:
                palette.setColor(QtGui.QPalette.Link, accent_color.lighter())
                palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.Highlight, accent_color.lighter())
                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, accent_color)
            else:
                palette.setColor(QtGui.QPalette.Link, accent_color)
                palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.Highlight, accent_color)
                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, accent_color.lighter())
```

# Action

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -81,16 +81,25 @@ class BaseTheme:
     def update_palette(self, palette, dark_theme, accent_color):
         if accent_color:
             accent_text_color = QtCore.Qt.white if accent_color.lightness() < 160 else QtCore.Qt.black
+            palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.Highlight, accent_color)
             palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.HighlightedText, accent_text_color)
-            palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.HighlightedText, accent_text_color)
-            if dark_theme:
-                palette.setColor(QtGui.QPalette.Link, accent_color.lighter())
-                palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.Highlight, accent_color.lighter())
-                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, accent_color)
-            else:
-                palette.setColor(QtGui.QPalette.Link, accent_color)
-                palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.Highlight, accent_color)
-                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, accent_color.lighter())
+
+            link_color = accent_color
+            while True:
+                if dark_theme:
+                    if link_color.lightness() < 160:
+                        link_color = link_color.lighter()
+                        continue
+                else:
+                    if link_color.lightness() > 160:
+                        link_color = link_color.darker()
+                        continue
+                break
+            palette.setColor(QtGui.QPalette.Link, link_color)
+
+        if dark_theme:
+            palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, QtGui.QColor(51, 51, 51))
+            palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.HighlightedText, QtCore.Qt.white)
 
 
 if IS_WIN:
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -81,12 +81,16 @@ class BaseTheme:
     def update_palette(self, palette, dark_theme, accent_color):
         if accent_color:
             accent_text_color = QtCore.Qt.white if accent_color.lightness() < 160 else QtCore.Qt.black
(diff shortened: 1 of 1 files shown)


picard/ui/theme.py
```diff
@@ -81,12 +81,24 @@ def get_syntax_theme(self):
...
-            palette.setColor(QtGui.QPalette.HighlightedText, accent_text_color)
+            palette.setColor(QtGui.QPalette.Active, QtGui.QPalette.HighlightedText, accent_text_color)
+
+            if accent_text_color == QtCore.Qt.white:
+                contrasting_accent_color = accent_color.lighter()
+            else:
+                contrasting_accent_color = accent_color.darker()
+
+            contrasting_accent_text_color = QtCore.Qt.white if contrasting_accent_color.lightness() < 160 else QtCore.Qt.black
+            palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.HighlightedText, contrasting_accent_text_color)
```
phw, line 92: I think we should set the accent color only for the active state here and don't set anything for inactive. For the white theme on Windows it will already use a gray color for the inactive state, so it matches previous behavior (colored active highlight, grey inactive highlight).

When using a dark theme we should instead set proper colors for the inactive highlight in `WindowsTheme.update_palette`, something like this would do:

```
        def update_palette(self, palette, dark_theme, accent_color):
            # Adapt to Windows 10 color scheme (dark / light theme and accent color)
            super().update_palette(palette, dark_theme, accent_color)
            if dark_theme:
                # ...
                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.Highlight, QtGui.QColor(51, 51, 51))
                palette.setColor(QtGui.QPalette.Inactive, QtGui.QPalette.HighlightedText, QtCore.Qt.white)
```

Gabrielcarvfer: Not sure on this one. My default accent is dark grey (63, 63, 63) and contrast is really poor by assuming (51, 51, 51). First two columns are selecting the album on the right panel and the artist field assuming (51,51,51). Last two columns are using the code in the PR, which results in active highlight (94,94,94) while inactive is kept at (63,63,63). I told you that the accent thing was a problem =x
Untitled

phw: > Maybe do the same thing, get the nearest gray and use it instead of assuming 51,51,51?

Not sure, what would be the nearest grey for your (94,94,94)? If we choose the nearest grey for the selected accent color and  make that lighter it could work.

I chose (51,51,51) because this matches what Windows uses. And Windows seems to use the same grey for inactive independent of the accent color. Not sure why we should make things more complicated then Windows is doing it. E.g. in Explorer active / inactive selection look like on your screenshot if I choose a dark accent color.

grafik

Gabrielcarvfer: Getting the nearest gray to the "contrasting" color. It isn't as good as I thought.
image

image

Gabrielcarvfer: Using the perceived brightness. Far less aggressive.

image


image

Gabrielcarvfer: > Not sure, what would be the nearest grey for your (94,94,94)? If we choose the nearest grey for the selected accent color and make that lighter it could work.

The nearest gray for a gray color is itself. For other colors it is more complicated.

> I chose (51,51,51) because this matches what Windows uses. And Windows seems to use the same grey for inactive independent of the accent color. Not sure why we should make things more complicated then Windows is doing it. E.g. in Explorer active / inactive selection look like on your screenshot if I choose a dark accent color.

Hmm, weird. I don't know then.

phw: I think we could go with the perceived brightness variant. I any case we should keep it grey, that's matching both the Windows behavior and the behavior without overriding the accent color.

But IMHO using a fixed gray for inactive would be ok. It is what Windows does, and it is also what happens on macOS. There Qt automatically uses the system configured accent color for active selections, and a fixed gray for inactive selections:

grafik

Mimicking on Windows what Qt is doing automatically on macOS was kind of my idea why I chose to use the accent color.

Gabrielcarvfer: Well, reverted some changes to use the fixed color. The perceived brightness version is on a different branch. If anyone complain, it is there. :)

phw: Ok, let's merge this. If needed we can tweak it, but I think this is now working well enough for most use cases.

PR #1586, opened 2020-07-13
PICARD-1881: Use scripting syntax function color for script docs

phw:
Instead of having a separate configurable color use the color from the syntax theme.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
When a dark background is used for the scripting documentation window the functions are not readable. Change the code to use the function color for the syntax theme, which can  be made aware of the color.

* JIRA ticket (_optional_): PICARD-1881

# Solution
Take the function color from the syntax theme.

# Action
Right now this fixes the issue for Windows, not macOS and KDE. For macOS our release builds do currently not support dark theme anyway. For KDE we need to detect usage of dark UI.


phw:
With this change:

grafik

Without:

grafik

phw:
The next improvement would of course be to have actual syntax highlighting for example code in the docs ;)

PR #1590, opened 2020-07-16
PICARD-1887: Detect use of macOS dark mode

phw:
# Summary
If Qt uses a dark UI on macOS (because dark mode is enabled in macOS preferences) Picard should detect this and use the new dark syntax theme.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
If macOS is configured to use a dark UI Qt5 will automatically use the dark colors. But then the syntax theme becomes hardly readable.

* JIRA ticket (_optional_): PICARD-1887

# Solution
Detect the use of dark mode on macOS. This allows setting the proper dark syntax theme.

Note that this does not mean dark mode will work in our current release version. Having the release builds using dark mode is done in PICARD-1357. Dark mode works when running from source, it's an issue with I think mainly PyInstaller being built against older SDK. Fixing this requires some care and probably releasing two different versions to still support older macOS.

# Action

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -146,6 +146,25 @@ if IS_WIN:
 
     theme = WindowsTheme()
 
+elif IS_MACOS:
+    try:
+        import AppKit
+    except ImportError:
+        AppKit = None
+
+    class MacTheme(BaseTheme):
+        def is_dark_theme(self):
+            if not AppKit:
+                return False
+            appearance = AppKit.NSAppearance.currentAppearance()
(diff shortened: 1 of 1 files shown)


phw:
I tested two scenarios:

1. Running from source. In this case Qt detects the dark mode correctly and uses a light / dark layout accordingly. The detection code here works as expected, we get the proper syntax theme
2. Running the release build. In this case Qt will always use the light mode independent of system settings (see my comment above). Again the detection code implemented in this PR works correctly and always detects the light layout getting used independent o0f system setting. That means we get the proper syntax theme.

PR #1593, opened 2020-07-19
PICARD-1890: Fix spurious Restart Picard message about color changes

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1890

- detection of color changes was broken, user was getting warn message even when there were no changes.

# Solution

- rework save to config part
- code cleanup
- added tests

# Action

### DIFF
--- a/picard/ui/colors.py
+++ b/picard/ui/colors.py
@@ -25,10 +25,15 @@ from PyQt5 import QtGui
 from picard import config
 
 
+class UnknownColorException(Exception):
+    pass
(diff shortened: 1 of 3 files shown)


phw:
Good catch. If I see this correctly the error message started to happen after we recently removed one of the color configs again. LGTM

PR #1596, opened 2020-07-20
PICARD-1891: handle incorrect date parsing exceptions

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1891

Incorrect year data could cause unhandled exceptions.

# Solution

- catch exceptions in extract_year_from_date() and log them to debug
- return None if year cannot be extracted
- check for None in metadata.compare_to_release_parts() before trying to compare years

This isn't perfect, the function is called a lot of time and it spams debug log (if enabled)

Note: I removed a very old comment in imports that doesn't make any sense (it was introduced 6 years ago in 323d12892c7f975b1e7286f57d72b60274a096e7) and it was making isort unhappy.

# Action

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -655,15 +655,10 @@ def limited_join(a_list, limit, join_string='+', middle_string='…'):
 def extract_year_from_date(dt):
     """ Extracts year from  passed in date either dict or string """
 
-    if isinstance(dt, Mapping):
-        try:
+    try:
+        if isinstance(dt, Mapping):
             return int(dt.get('year'))
-        except (TypeError, ValueError) as e:
-            log.debug(e)
-    else:
-        try:
+        else:
             return parse(dt).year
-        except (ParserError, TypeError) as e:
-            log.debug(e)
-
-    return None
+    except (ParserError, TypeError, ValueError):
+        return None
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -50,11 +50,14 @@ import sys
 from time import time
 import unicodedata
 
-from dateutil.parser import parse
+from dateutil.parser import (
+    ParserError,
+    parse,
+)
 
 from PyQt5 import QtCore
 
-# Required for compatibility with lastfmplus which imports this from here rather than loading it direct.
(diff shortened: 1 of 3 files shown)


picard/util/__init__.py
```diff
@@ -655,9 +656,14 @@ def extract_year_from_date(dt):
     """ Extracts year from  passed in date either dict or string """
 
     if isinstance(dt, Mapping):
-        year = int(dt.get('year'))
+        try:
+            return int(dt.get('year'))
+        except (TypeError, ValueError) as e:
+            log.debug(e)
```
phw, line 662: I wonder if we really need to log this here. It's kind of expected functionality here that it might fail, and the function reacts to this by returning None. IMHO a `pass` or `return None` would be ok here and below.

test/test_utils.py
```diff
@@ -77,6 +78,23 @@ def test_incorrect(self):
...
 
 
+class ExtractYearTest(PicardTestCase):
+    def test_string(self):
+        self.assertEqual(extract_year_from_date(""), None)
+        self.assertEqual(extract_year_from_date(2020), None)
+        self.assertEqual(extract_year_from_date("2020"), 2020)
+        self.assertEqual(extract_year_from_date('2020-02-28'), 2020)
+        self.assertEqual(extract_year_from_date('2015.02'), 2015)
+        self.assertEqual(extract_year_from_date('2015; 2015'), None)
```
phw, line 88: I'd add a test for the format as supported by ID3 [1]: `yyyy-MM-ddTHH:mm:ss`

```
self.assertEqual(extract_year_from_date('2020-07-21T13:00:00'), 2020)
```

[1] https://id3.org/id3v2.4.0-structure

picard/util/__init__.py
zas: Done in 4f592a8

test/test_utils.py
zas: Done in 5e5535a

PR #1603, opened 2020-08-12
PICARD-1909: Update metadatabox on tag modifying actions

phw:
# Summary
Ensure the metadatabox does get updated after removal, pasting from clipboard or "use original values".

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
The metadatabox did not trigge rthe updates on important actions.

* JIRA ticket (_optional_): PICARD-1909

# Solution
Run `MetadataBox.update()` after those actions. I tried to make sure to run this update only after the entire action, so not for every call to e.g. `set_tag_values()`.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1112,7 +1112,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.tags_from_filenames_action.setEnabled(bool(files))
         self.track_search_action.setEnabled(have_objects)
 
-    def update_selection(self, objects=None, new_selection=True):
+    def update_selection(self, objects=None, new_selection=True, drop_album_caches=False):
         if self.ignore_selection_changes:
             return
 
@@ -1178,7 +1178,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
         if new_selection:
             self.metadata_box.selection_dirty = True
-        self.metadata_box.update()
+        self.metadata_box.update(drop_album_caches=drop_album_caches)
         self.cover_art_box.set_metadata(metadata, orig_metadata, obj)
         self.selection_updated.emit(objects)
 
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -294,6 +294,7 @@ class MetadataBox(QtWidgets.QTableWidget):
                     self.clipboard = list(self.tag_diff.new[tag])
             elif e.key() == QtCore.Qt.Key_V and column == 2 and tag != "~length":
                 self.set_tag_values(tag, list(self.clipboard))
+                self.update()
         return super().event(e)
 
     def closeEditor(self, editor, hint):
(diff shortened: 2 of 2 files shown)


phw:
@Gabrielcarvfer Could you review this? You were deeper into the event handling, I don't want to cause a bigger regression here. But I think the update() should only run for some explicit actions and should not again cause any general slowdowns.

Gabrielcarvfer:
> 
> 
> @Gabrielcarvfer Could you review this? You were deeper into the event handling, I don't want to cause a bigger regression here. But I think the update() should only run for some explicit actions and should not again cause any general slowdowns.

I was thinking on just including a forced update with same selection after editing tags. One liner.

```
    def set_tag_values(self, tag, values, objects=None):
        if objects is None:
            objects = self.objects
        self.parent.ignore_selection_changes = True
        if values == [""]:
            values = []
        if not values and self.tag_is_removable(tag):
            for obj in objects:
                del obj.metadata[tag]
        elif values:
            for obj in objects:
                obj.metadata[tag] = values
                obj.update()
        self.parent.ignore_selection_changes = False
        self.parent.update_selection(new_selection=False) # Previous updates are ignored, this forces the update
```
https://github.com/metabrainz/picard/blob/master/picard/ui/metadatabox.py#L405

Gabrielcarvfer:
I was testing, in fact I suggest a two liner:
```
self.parent.update_selection(new_selection=False)
self.update(drop_album_caches=True)
```

The reason behind this is part of the bug I've found based on the post in the forums.
https://community.metabrainz.org/t/picard-2-4-released/489666/16

I've made a few additional changes to fix the issue and expand the "Use original values" also for tracks without attached/linked files and related albums. :)

https://github.com/Gabrielcarvfer/picard/commit/026eed1a866ecb943cd44d1cec7f859cbf54a504

phw:
The reason I did not run the update in set_tag_values was to run the update only once if you remove or reset many tags. Not sure how big the impact is. Maybe we shod try on an album level with one of those huge releases.

Otherwise your approach is of course cleaner and ensures all uses of set_tag_value trigger an update. So if there is no significant performance difference I'd favor your patch.

I need to test this odd original values getting changed for tracks thing.

phw:
I measured this a bit, there is no huge difference. I loaded the release https://musicbrainz.org/release/356bcf92-7260-4a95-b5b8-334df0872593, selected the album entry, then selected all tags in the metadatabox and first removed all tags, then selected all tags again and used the "Use Original Values" action. Did this several times. There is a small runtime difference, but it is not too big. The numbers on my system:

This patch (running the update only once):

|                                              | remove | use origs |
|-------------------------------|----------|-------------|
| update only once                 | 0,010004282 s | 16,68726269 s |
| update on each tag change | 0,02965696653 s | 17,03120899 |

So it's not too big a deal in absolute values (this mostly shows that use original values is a really slow operation and should be optimized). Not sure, if the small gain in absolute runtime is worth my patch. Gabriel's patch might also avoid some other update issues in the future. What do others think?

@Gabrielcarvfer About this line:

    self.parent.update_selection(new_selection=False)

Which case does this capture? In my patch I have not used it

phw:
Allright, I now see how `MainWindow.update_selection` is needed to update the tree views. I addressed this in  5a8a716b886ef829fff3b0028d95f5f5ca17989b. This also makes sure to run metadatabox.update only once.

We have a separate issue, but this one seems to be older: When deleting tags in the metadatabox this change is never reflected in the tree views. I need to look into this in more detail, but I added PICARD-1911 to not forget it.

@Gabrielcarvfer About the original values on tracks issues, could you create another issue and submit it in a separate PR? I tested your changes, so far it seems to work, but we should not mix the different issues up.

phw:
Fix for PICARD-1911 was rather trivial: The changed objects where updated on tag changes, but not tag removal. Fixed in https://github.com/metabrainz/picard/pull/1603/commits/584e57d0fe021e56a3a6f8d9cc3e196920a0193c

Gabrielcarvfer:
> 
> @Gabrielcarvfer About the original values on tracks issues, could you create another issue and submit it in a separate PR? I tested your changes, so far it seems to work, but we should not mix the different issues up.

Sure, it seems related to PICARD-254.

>  
> We have a separate issue, but this one seems to be older: When deleting tags in the metadatabox this change is never reflected in the tree views. I need to look into this in more detail, but I added [PICARD-1911](https://tickets.metabrainz.org/browse/PICARD-1911) to not forget it.

PICARD-629?

PR #1604, opened 2020-08-13
PICARD-1913: Expand "Use original values" to tracks and albums

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Use original values is only available for tracks with files attached. Now tracks and albums can also be resetted. 

# Problem

Try selecting an album with a few files matched to tracks and others with no files. Now change a tag (e.g. album name), then right-click to restoring the original values. 

If the tag was saved to the album, it won't be recovered as the orig_metadata field was empty (L378 of album) and the context menu didn't create the calls to set the tags to their original values (L366-370 of metadatabox).

If the tag was saved to an unmatched track, it won't be recovered as the context menu didn't create the calls to set the tags to their original values (L371-374 of metadatabox). Even if it was restored, you wouldn't see as the TagDiff only showed the current value as the original (L530 of OG metadatabox).

* JIRA ticket (_optional_): PICARD-1913

# Solution

Prepare functions to restore OG tag values of unmatched tracks and albums, fix tagdiff source and copy the metadata of albums during loading finalization.

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -20,6 +20,7 @@
 # Copyright (C) 2017 Antonio Larrosa
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2019 Joel Lintunen
+# Copyright (C) 2020 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


phw:
Looks good to me and works as expected. I added PICARD-1913 for this. PICARD-254 describes a different behavior of not being able to use "Use original value" at all for multiple selected tags. Also it is a past issue solved in 1.1

PR #1606, opened 2020-08-14
PICARD-1915: Update metadata view after selected album finished loading

phw:
# Summary
When a album finished loading **and** is selected, update the metadata view.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1915, PICARD-1914

# Solution
Update the metadatabox.

# Action
@zas @Gabrielcarvfer This might have quite serious performance impact if the user is loading a lot of albums **and** has all of them selected. In this case it is run after every album load and a tag diff over all albums is created. Should we limit this update, e.g. only run the update when 1 or a few albums are selected?

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -397,8 +397,7 @@ class Album(DataObject, Item):
                 func()
             self._after_load_callbacks = []
             if self.item.isSelected():
-                self.tagger.window.metadata_box.selection_dirty = True
-                self.tagger.window.metadata_box.update()
+                self.tagger.window.refresh_metadatabox()
 
     def _finalize_loading_track(self, track_node, metadata, artists, va, absolutetracknumber, discpregap):
         # As noted in `_parse_release` above, the release artist nodes
--- a/picard/album.py
+++ b/picard/album.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Hendrik van Antwerpen
 # Copyright (C) 2008 ojnkpjg
-# Copyright (C) 2008-2011, 2014, 2018-2019 Philipp Wolfer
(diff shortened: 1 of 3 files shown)


zas:
> Should we limit this update, e.g. only run the update when 1 or a few albums are selected?

I'd say yes, and make this limit user configurable (from never update, update only if < n albums, always update), and default to 10 or so. That said, only IF we can't avoid this, I'd prefer a solution which doesn't involve such limit.

phw:
I measured the loading time of ~340 releases with all releases selected during loading with and without this patch and could not really measure a bug difference. Complete time to process and load all releases was ~347 s to 351 s for each run, whether the metadatabox update ran or not.

So I'd definitely merge this as is. It's rather an edge case, I doubt people will usually select all releases during loading and sit there, and even then they probably need much more releases then my test case. In  theory this metadatabox update begins to have a bigger impact the more releases are selected of course, but it doesn't seem to bad for at least a few hundred.

> and make this limit user configurable (from never update, update only if < n albums, always update), and default to 10 or so. That said, only IF we can't avoid this, I'd prefer a solution which doesn't involve such limit.

I guess you mean this more generic for the metadatabox in general? I definitely would not add an option for this specific case only. But we could have a limit for the metadatabox showing the tag diff only until a certain threshold of selected items. If more are selected we could show a button to update and show the tag diff. But given my testing above I assume we could even set it rather high by default (500? 1000?).

The update of the metadatabox can still be quite slow for larger amounts of items. For a single album it is merely ~0.02 s on my system, but for all 340 albums selected it is over 0.7 s. It just does not have such an big impact on the loading since it is run asynchronously and loading is mostly waiting for network responses.

phw:
I also included the fix for PICARD-1914 in this PR, as it is very similar in solution.

PR #1609, opened 2020-08-17
PICARD-1899: Link to new documentation pages

phw:
# Summary
Update all documentation links to point directly to picard-docs.musicbrainz.org. 

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1899

# Solution
Update the URLs. Also allow explicit definition of help URLs for each option page by specifying a `HELP_URL`. `HELP_URL` might be an absolute URL with protocol (e.g. a plugin could directly link to https://github.com/rdswift/picard-plugins/blob/2.0/plugins/format_performer_tags/docs/README.md or something nicer) or start with a slash `/config/options_filerenaming.html`. The first case is primarily meant for plugins which want to link to their own documentation, the later case will always be linked to picard-docs.musicbrainz.org. 

# Action
Language handling is excluded for now, but can be easily added later. As currently the docs are only available in English and not yet setup for translation this should be sufficient. Later Picard releases could whitelist languages to allow, with a fallback to use English.

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -38,6 +38,7 @@ from PyQt5.QtCore import QStandardPaths
 from picard import (
     PICARD_APP_NAME,
     PICARD_ORG_NAME,
+    PICARD_VERSION,
 )
 from picard.const.attributes import MB_ATTRIBUTES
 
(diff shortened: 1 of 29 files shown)


picard/const/__init__.py
```diff
@@ -68,20 +69,27 @@
 CAA_HOST = "coverartarchive.org"
 CAA_PORT = 443
 
+# Prepare documentation URLs
+if PICARD_VERSION.identifier == 'final':
+    DOCS_VERSION = "v{}".format(PICARD_VERSION.to_string(short=True))
+else:
+    DOCS_VERSION = ""  # points to latest version
+DOCS_BASE_URL = "https://picard-docs.musicbrainz.org/" + DOCS_VERSION + '/en'
```
rdswift, line 77: For latest version, this will set `DOCS_BAS_URL` to "https://picard-docs.musicbrainz.org//en" (note the double slash).  Not sure if this is a problem.  Perhaps something like:

```
if PICARD_VERSION.identifier == 'final':
    DOCS_VERSION = "v{}/".format(PICARD_VERSION.to_string(short=True))
else:
    DOCS_VERSION = ""  # points to latest version
DOCS_BASE_URL = "https://picard-docs.musicbrainz.org/" + DOCS_VERSION + 'en'
```

picard/const/__init__.py
```diff
@@ -68,20 +69,27 @@
...
-    'plugins':          "https://picard.musicbrainz.org/plugins/",
-    'forum':            "https://community.metabrainz.org/c/picard",
-    'donate':           "https://metabrainz.org/donate",
-    'chromaprint':      "https://acoustid.org/chromaprint#download",
-    'acoustid_apikey':  "https://acoustid.org/api-key",
+    'home':                "https://picard.musicbrainz.org/",
+    'documentation':       DOCS_BASE_URL + '/',
+    'troubleshooting':     DOCS_BASE_URL + '/en/troubleshooting/troubleshooting.html',
+    'doc_options':         DOCS_BASE_URL + '/en/config/configuration.html',
+    'doc_scripting':       DOCS_BASE_URL + '/en/scripting.html',
```
rdswift, line 85: Isn't the "en" portion already included in `DOCS_BASE_URL`?

rdswift:
> ...The first case is primarily meant for plugins which want to link to their own documentation...

Very nice!

> Language handling is excluded for now, but can be easily added later. As currently the docs are only available in English and not yet setup for translation this should be sufficient. Later Picard releases could whitelist languages to allow, with a fallback to use English.

If you point to a top level file (either `/index.html` or `/{version}/index.html`) they will automatically try to match a supported language to one of the user's preferred languages (as provided by the browser) with a fallback to English if there is no match.  I'm still trying to find a way to do this with the topic-specific pages.  For now they will display in the specified language, but selecting a new language from the options settings (at the bottom of the left column) will switch to the new language and stay on the same topic page.

picard/const/__init__.py
zas: Yes, scripting doc points to `https://picard-docs.musicbrainz.org//en/en/scripting.html̀` which is a 404

rdswift:
> Scripting help button opens `https://picard-docs.musicbrainz.org//en/config/options_scripting.html` (double slash in path)

I thought I mentioned that (as well as possibly adding and extra "en" in the URL) in my review comments.  I guess I should have marked them as changes requested rather than just comments.

picard/const/__init__.py
phw: Yes, I moved the language handling out of the options to const and this is the result. I fixed this.

PR #1613, opened 2020-08-19
PICARD-1918: Fix issue saving cover art without front image

phw:
# Summary
Trying to save cover art as separate files would fail, if there is no image with type "front" and caa_save_single_front_image is enabled.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1918

# Solution
Check if a front image was found before updating the list of images to save. This way we don't end up with `images == [None]` 

# Action
This fixed the exception, but we have a more general problem with this.

The caa_save_single_front_image option is really odd, and I am not fully sure what it is supposed to do. First off it is presented as a CAA option, but it applies to all cover art.

Also the code as it was and is now suggests that if there is no front image all images will be saved anyway. That looks wrong to me. Either it should save nothing or maybe fall back to the first image found.

I think we should move this option to the general Options > Cover Art page, rename it to save_only_one_front_image and implement it exactly like embed_only_one_front_image. @zas ?

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -497,7 +497,9 @@ class File(QtCore.QObject, Item):
         counters = defaultdict(lambda: 0)
         images = []
         if config.setting["caa_save_single_front_image"]:
-            images = [metadata.images.get_front_image()]
+            front = metadata.images.get_front_image()
+            if front:
+                images.append(front)
         if not images:
             images = metadata.images
         for image in images:


zas:
> I think we should move this option to the general Options > Cover Art page, rename it to save_only_one_front_image and implement it exactly like embed_only_one_front_image. @zas ?

Yes, it would make more sense. Cover art options are rather messy, and as shown in https://github.com/metabrainz/picard/pull/1502 all this stuff is messy, and lack of testing. Frankly, this is something we should clearly rework from scratch, as it was build along time more like successive hacks rather than designed. So any change giving more consistency and sense to this stuff is very welcome.

It should be noted that most of the complexity comes from the very high number of cases produced by the number of cover art options cross the number of image "types" (CAA + other providers)

zas (approved):

phw:
Ok, so let's merge this for the fix and I'll prepare a refactoring for this option. Having a bit more consistency here hopefully makes it also easier to rework this later and hopefully completing your work on testing this stuff.

PR #1618, opened 2020-09-05
PICARD-1929: Translate NSIS with Transifex

phw:
# Summary
This allows us to handle the translations of the NSIS Windows installer using Transifex.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1929

# Solution
- Instead of keeping the translations in NSIS format, use JSON files
- Have scripts to convert from JSON to NSIS format and vice versa
- Integrate generation of the NSIS translation files into the build process
- Setup configuration for Transifex

# Action
I could not finally test the Transifex setup nor can I do the Transifex configuration of the new `musicbrainz.picard_installer` resource.

@zas could you set this up? I think it needs a new resource which then should regularly pull from `https://raw.githubusercontent.com/metabrainz/picard/master/installer/i18n/sources/en.json`


zas:
> @zas could you set this up? I think it needs a new resource which then should regularly pull from

I did. Now when doing `python setup.py get_po_files`, it also retrieves picard_installer translations. See https://www.transifex.com/musicbrainz/musicbrainz/picard_installer/

There are warnings about escaping, I let you check that.

Imho we should rename `get_po_files` to `get_translations` now it has a much wider use.

phw:
Thanks a lot for  setting this up. The escaping warnings are a bit strange, I'll look into this. I don't think it is a real issue, but having those displayed would at least be confusing.

> Imho we should rename `get_po_files` to `get_translations` now it has a much wider use.

Good point, I'll rename that. Maybe even better to `pull_translations`, makes it a bit clearer we are getting translations from external.

phw:
> There are warnings about escaping, I let you check that.

Found the issue. It did not really cause us problems, but Transifex has problems if the JSON files had non-ASCII characters encoded (e.g. `\u00fc` instead of `ü`). But since the JSON files are UTF-8 encoded anyway and it is better for inspecting those files to show the proper characters I have changed this in f9dd69dcc8eec2a9694545a061f12e7768a5d2b6

I have also resaved the affected translations on Transifex, which immediately resolved the warnings. And we already have a successful test, the installer was already translated into Spanish and I pulled this in.

setup.py
```diff
@@ -764,7 +767,7 @@ def _picard_packages():
         'install': picard_install,
         'install_locales': picard_install_locales,
         'update_constants': picard_update_constants,
-        'get_po_files': picard_get_po_files,
+        'pull_translations': picard_pull_translations,
```
zas, line 770: I wonder if we should keep `get_po_files` as an alias for a while.

zas (approved):

setup.py
phw: There are not many users of this I think, and adapting to it is a quick change. I guess if you get an error message it is easy to update any script or whatever one might have.

PR #1619, opened 2020-09-05
PICARD-1936: Rename the "Whitelist" cover art provider to "Allowed Cover Art URLs"

phw:
# Summary
This PR makes changes to avoid the terms blacklist or whitelist in the sources and UI.

Most notable it renames the "Whitelist" cover art provider to "Allowed Cover Art URLs", which also helps in clarifying what this provider is actually about.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1936

# Solution

# Action

### DIFF
--- a/picard/ui/options/cover.py
+++ b/picard/ui/options/cover.py
@@ -54,8 +54,7 @@ class CoverOptionsPage(OptionsPage):
         config.BoolOption("setting", "save_images_overwrite", False),
         config.ListOption("setting", "ca_providers", [
             ('Cover Art Archive', True),
-            ('Amazon', True),
-            ('Whitelist', True),
+            ('UrlRelationships', True),
             ('CaaReleaseGroup', False),
             ('Local', False),
         ]),
--- a/.pylintrc
+++ b/.pylintrc
@@ -5,11 +5,11 @@
 # run arbitrary code.
 extension-pkg-whitelist=
 
-# Add files or directories to the blacklist. They should be base names, not
+# Add files or directories to the ignore list. They should be base names, not
 # paths.
 ignore=CVS
 
-# Add files or directories matching the regex patterns to the blacklist. The
+# Add files or directories matching the regex patterns to the ignore list. The
 # regex matches against base names, not paths.
 ignore-patterns=ui_.*\.py
 
--- a/picard/config_upgrade.py
+++ b/picard/config_upgrade.py
@@ -303,6 +303,15 @@ def upgrade_to_v2_4_0_beta_3(config):
     _s[opt] = [t.strip() for t in _s.raw_value(opt, qtype='QString').split(',')]
 
 
+def upgrade_to_v2_5_0_dev_1(config):
+    """Rename whitelist cover art provider"""
+    _s = config.setting
+    _s['ca_providers'] = [
+        ('UrlRelationships' if n == 'Whitelist' else n, s)
+        for n, s in _s['ca_providers']
+    ]
+
+
 def rename_option(config, old_opt, new_opt, option_type, default):
     _s = config.setting
     if old_opt in _s:
@@ -327,4 +336,5 @@ def upgrade_config(config):
     cfg.register_upgrade_hook(upgrade_to_v2_1_0_dev_1)
     cfg.register_upgrade_hook(upgrade_to_v2_2_0_dev_3)
     cfg.register_upgrade_hook(upgrade_to_v2_4_0_beta_3)
+    cfg.register_upgrade_hook(upgrade_to_v2_5_0_dev_1)
     cfg.run_upgrade_hooks(log.debug)
--- a/picard/coverart/providers/__init__.py
+++ b/picard/coverart/providers/__init__.py
@@ -42,7 +42,7 @@ from picard.coverart.providers.provider import (  # noqa: F401 # pylint: disable
     CoverArtProvider,
     ProviderOptions,
 )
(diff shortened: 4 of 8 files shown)


hiccup2:
Why should this word be removed?
It's a very common term that effectively explains what it does, and it has been doing that for a very long time.
It originates back to 16th century England, and I believe it became a popular term in the US in relation to conflicts regarding workers, employers and trade unions.

picard/ui/options/cover.py
```diff
@@ -54,8 +54,7 @@ class CoverOptionsPage(OptionsPage):
         config.BoolOption("setting", "save_images_overwrite", False),
         config.ListOption("setting", "ca_providers", [
             ('Cover Art Archive', True),
-            ('Amazon', True),
```
zas, line 57: Was this forgotten in https://github.com/metabrainz/picard/pull/1101?

zas (approved):
LGTM

zas:
> Why should this word, and the word blacklist be removed?
> It's a very common term that effectively explains what it does, and it has been doing that for a very long time.
> The word blacklist originates back to 16th century England, and I believe it became a popular term in the US in relation to conflicts regarding workers, employers and trade unions.

It is more about making things easier to understand for end users. Whitelist for cover art doesn't tell anything about the actual source, using "Url Relationships" is actually what it does.

picard/ui/options/cover.py
phw: Yes, that's a left over from this change.

phw:
> Why should this word, and the word blacklist be removed?

As a cover art source it is not really obvious what "whitelist" means. You need a lot of explaining to tell people what kind of cover art this is actually fetching.

> It's a very common term that effectively explains what it does [...]

Actually not really. To my experience this term is badly understood outside IT, I can't count the times I have been explaining the meaning of blacklist/whitelist to people. The terms themselves also don't effectively explain themselves, their meaning is opaque to someone not already familiar with the term.

Making sense of those terms also suggests a mindset of white equals good and black equals bad. And while to some extend this is for sure for various reasons (like e.g. we are diurnal animals), this connotation in general has become somewhat unpopular.

Sophist-UK:
Playing devil's advocate about word history, I am not sure that words having a "history" is a reason not to change them. The "N" word has a history dating back to the 1700s, and is a pejorative derivative of a different word dating buck many hundreds of years that describes a colour that is still in common usage today in some European-based languages. But a 200-300 years age doesn't make it an appropriate word to use. The choice of words to use in an App, in prose or poetry, in news reports etc. can only be made based on context, whether use of the word is necessary, what alternatives are available and how it will be taken by some people. The N word is generally agreed to be unacceptable these days regardless of history or that it was previously in common usage.

That said, I have never like political correctness like "Chair Person" rather than "Chairman" or "Madam Chairman" - if we took this to its logical extreme and wanted to be certain that were not discriminating against anything, we would have to use "Seat Being" so as not to discriminate against sofas/stools and parrots/extra-terrestrials.

In this particular case, according to [Wikipedia - Blacklisting](https://en.wikipedia.org/wiki/Blacklisting) the word "blacklisting" dates back to the 1600s and has nothing to do with skin colour. That said, according to [Wikipedia - Whitelisting](https://en.wikipedia.org/wiki/Whitelisting) the IT industry in general seems to be moving away from these terms, presumably under pressure of political-correctness, and if general usage in this context is changing, then perhaps we should change with it.

But, all of this seems moot anyway. :-)

phw:
> But, all of this seems moot anyway. :-)

Agreed on this. Politics aside, the word "whitelist" especially in this context is just not very clear.

hiccup2:
It's a shame that words that have nothing to do with race or skin colour have become a target of debate and alleged offense.

But I do appreciate you guys far too much to make this into some senseless discussion. (and that's also why I didn't raise it on the forum)
And while I don't agree with the first reason that was given: "to avoid the terms blacklist or whitelist", the second reason and explanation makes full sense.

I hope that I will live long enough to see that the majority of intelligent people regain the sense to recognise and distinguish words that have no offending history or intention, and that the words that are generally considered offensive—such as the 'N' word that sophist referred to—are not to be avoided _only_ by people of one colour and perfectly o.k. to use for people of another colour, but are avoided by the whole community.

Weird times.

Sophist-UK:
@hiccup2 Think of this as a Venn diagram of the world's population with two overlapping circles in it:

1. A circle representing intelligent people who are resistant to political correctness. This by definition excludes anyone that is unable to make reasoned and reasonable judgements about politicians, and so excludes the majority of British and US adults.
2. A circle representing non-racist people.

The overlap represents the people who understand that e.g. whitelist is not a racist term. Imagine just how small a proportion of the population that represents. Sad, isn't it.

hiccup2:
Obviously there will always be people with a lower education or intelligence than others.
That's not sad, it's not fair, it's nature.
The sad part is when the more educated, smarter, and good intended people stoop down.

Sophist-UK:
What is sad is when the more educated, smarter people sit in an ivory tower and don't try to reach out and communicate with everyone else in terms they can relate to and which can help move them away from the brainwashing that they don't realise they are being fed by morally-corrupt power-mad greedy self-centred politicians out to win at any cost, particularly when they are out to win by destroying democracy.

But this is not really the forum for political discourse.

PR #1620, opened 2020-09-07
PICARD-1937: local coverart context menu

phw:
# Summary
Adds a context menu entry to the cover art box to choose a local file using the file browser.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1937

# Solution
Add the context menu option.

I also added some refactoring to move the `ActiveLabel` class to picard.ui.widgets, and by this also moving the drag and drop handlers to the proper class.

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -552,7 +552,7 @@ class CoverArtBox(QtWidgets.QGroupBox):
             menu.addSeparator()
 
         load_image_behavior_group = QtWidgets.QActionGroup(self.parent)
-        action = QtWidgets.QAction(_('Replace front cover art on drop'), self.parent)
+        action = QtWidgets.QAction(_('Replace front cover art'), self.parent)
         action.setCheckable(True)
         action.triggered.connect(partial(self.set_load_image_behavior, behavior='replace'))
         load_image_behavior_group.addAction(action)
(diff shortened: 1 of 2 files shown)


picard/ui/coverartbox.py
```diff
@@ -550,6 +542,12 @@ def contextMenuEvent(self, event):
             use_orig_value_action.triggered.connect(self.item.keep_original_images)
             menu.addAction(use_orig_value_action)
 
+        if self.item and self.item.can_show_coverart:
+            name = _('Choose local file ...')
+            choose_local_file_action = QtWidgets.QAction(name, self.parent)
+            choose_local_file_action.triggered.connect(self.choose_local_file)
+            menu.addAction(choose_local_file_action)
+
```
zas, line 550: The options (defined below) to Replace or Append "on drop" should be renamed since it seems they also apply to the file set through this dialog (and that's not a "drop").

zas (changes requested):
LGTM, apart the comment about Replace/Append

picard/ui/coverartbox.py
phw: I just removed the "on drop" part, so now it just reads "Replace front cover art" and "Append front cover art". I think this is clear enough for a menu action, everything else requires a longer explanation text.

zas (approved):

PR #1623, opened 2020-09-08
PICARD-1938: Add "Load selected files" action to file browser

phw:
# Summary
Extend the context menu of the file browser with a "Load selected files" action

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1938

# Solution
Extend the context menu with the "Load selected files" action. Also switch to a manually created context menu to include the separator

# Action

### DIFF
--- a/picard/ui/filebrowser.py
+++ b/picard/ui/filebrowser.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2006-2008 Lukáš Lalinský
 # Copyright (C) 2008 Hendrik van Antwerpen
-# Copyright (C) 2008-2009, 2019 Philipp Wolfer
+# Copyright (C) 2008-2009, 2019-2020 Philipp Wolfer
 # Copyright (C) 2011 Andrew Barnert
 # Copyright (C) 2012-2013 Michael Wiencek
 # Copyright (C) 2013 Wieland Hoffmann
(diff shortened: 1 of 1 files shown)


picard/ui/filebrowser.py
```diff
@@ -168,6 +180,18 @@ def _get_destination_from_path(self, path):
...
+        indexes = self.selectedIndexes()
+        if not indexes:
+            return
+        paths = set(self.model.filePath(index) for index in indexes)
+        dirs = set(p for p in paths if os.path.isdir(p))
+        files = paths - dirs
+        tagger = QtCore.QObject.tagger
+        for d in dirs:
+            tagger.add_directory(d)
+        tagger.add_files(files)
```
zas, line 193: Not really related to this PR, but this code makes me things about it:

That's a long time I'm bothered by this separation between files and directories in the code and in the UI (file selection).

At this point, it would probably make sense to unify `add_directory()` (which calls `add_files()`) and `add_files()` in one function `add_paths()` taking a list of paths, testing them for validity and handle files and directories.

It would reduce code redundancy (here introduced, because we already test for dirs elsewhere in order to call either `add_directory()` or `add_files()`.

What do you think?

zas (approved):

picard/ui/filebrowser.py
phw: Yes, agreed. I actually thought about this, too, while writing the code. I'll look at this in a separate PR. I'm not sure yet if it will get a combined function, because we benefit from handling it in a single call function, or if it will just sit above the existing add_files / add_directories functions.

PR #1624, opened 2020-09-08
Added tag-release.sh script to create the release tags

phw:
# Summary
Added a script to automate the creation of the release tags. That makes it easier to have consistent release tags and avoid mistakes.

I had actually used this the last couple of releases instead of trying to manually get everything right (and mistyping 

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

### DIFF
--- a/scripts/tools/tag-release.sh
+++ b/scripts/tools/tag-release.sh
@@ -1,0 +1,8 @@
+#!/bin/bash
+cd "$(dirname "$0")/../../" || exit
+
+PICARD_VERSION=$(python -c "import picard; print(picard.__version__)")
+TAG="release-$PICARD_VERSION"
+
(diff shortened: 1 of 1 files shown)


zas (changes requested):
Most Codacy complaints make sense:
- add a shebang to define shell
- use double quotes around variables (`"$TAG"`)
- use double quotes around echo message: `echo "Tagging $TAG..."`
Also I think we need to be able to pass options to `git tag`, mainly to let user selects key to use (see https://git-scm.com/docs/git-tag, `-u`)

phw:
@zas I addressed the issues and changed the script to pass command line parameters on to the git tag command

zas (approved):

PR #1626, opened 2020-09-10
PICARD-1942: Display tracklist in Info Dialog Infos tab for albums too

zas:
On not-yet-matched clusters, context menu shows Infos item and user can see a copiable tracklist in the Info tab.
Capture du 2020-09-10 12-39-08

On loaded albums the context menu will be there only when album has cover art (and no errors).

This patch unifies things a bit without adding much code, Infos dialog is accessible as soon an album is loaded, and Infos tab displays a text tracklist for it, which can be convenient for the user as it is copiable.

Capture du 2020-09-10 12-40-04

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1942

# Solution

# Action

### DIFF
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -290,6 +290,48 @@ def format_file_info(file_):
                              htmlescape(i[1])), info))
 
 
+def format_tracklist(cluster):
+    info = []
+    info.append("<b>%s</b> %s" % (_('Album:'),
+                                  htmlescape(cluster.metadata["album"])))
+    info.append("<b>%s</b> %s" % (_('Artist:'),
+                                  htmlescape(cluster.metadata["albumartist"])))
+    info.append("")
+    TrackListItem = namedtuple('TrackListItem', 'tracknumber, title, artist, length')
+    tracklists = defaultdict(list)
+    if isinstance(cluster, Album):
+        objlist = cluster.tracks
+    else:
+        objlist = cluster.iterfiles(False)
+    for obj_ in objlist:
+        m = obj_.metadata
+        artist = m["artist"] or m["albumartist"] or cluster.metadata["albumartist"]
+        track = TrackListItem(m["tracknumber"], m["title"], artist,
+                              m["~length"])
+        tracklists[obj_.discnumber].append(track)
+
+    def sorttracknum(item):
+        try:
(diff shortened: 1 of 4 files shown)


picard/ui/infodialog.py
```diff
@@ -290,6 +290,48 @@ def format_file_info(file_):
...
+    for obj_ in objlist:
+        m = obj_.metadata
+        artist = m["artist"] or m["albumartist"] or cluster.metadata["albumartist"]
+        track = TrackListItem(m["tracknumber"], m["title"], artist,
+                              m["~length"])
+        tracklists[obj_.discnumber].append(track)
+
+    def sorttracknum(item):
+        try:
+            return int(item.tracknumber)
```
phw, line 315: The int conversion is not needed here, the `tracknumber` property always returns an int. Hence we also never go into the exception here, tracknumber will be 0 instead. But this also means it fails in the "3/10" case. Do we have a bug here? I see the following options:

1. Keep the conversion as is, but use `m["tracknumber"]` directly here
2. Integrate the regex parsing into `Item.tracknumber` property (is this wanted at other places?)
3. Use `item.tracknumber`, but compare it to 0 and if it is do the regex parsing
4. Drop parsing the string

I tend to prefer 2, but depends on how else this property gets used.

phw (commented):
This makes a lot of sense, it's really inconsistent currently. Could you add a ticket for this?

picard/ui/infodialog.py
zas: I think you missed something:
- `item.tracknumber` in this context is from metadata
- `item` is a TrackListItem named tuple in https://github.com/metabrainz/picard/blob/8491e3f92b078bc7a40f109ddc263da9c3ed0b00/picard/ui/infodialog.py#L378
- `discnumber` comes from `File` or `Track` `discnumber` property though

phw: Got it, thanks :) From my side I think this is good to merge, definitely makes the UI more consistent

zas:
> Could you add a ticket for this?

https://tickets.metabrainz.org/browse/PICARD-1942

phw (approved):

PR #1630, opened 2020-09-18
PICARD-1879: Sequential matching of files on drop

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

When dragging and dropping multiple files onto a track, assign the first file to this track and the following files to the following tracks in order.

# Problem

* JIRA ticket (_optional_): PICARD-1879

Currently when dragging and dropping multiple files onto a track all the dropped files get assigned to this track. This is probably seldom what you want unless you explicitly have multiple files of the same track.

This patch changes the default behavior so multiple files get matched in sequence to multiple tracks, starting with the first file getting matched to the track being dropped on and the following files to the next tracks.

For the rare circumstances someone really wants to add multiple files to a single track the patch allows dragging with the Alt modifier pressed. I consider this an advanced feature we could document and advanced users could use. But in practice often this is not really necessary, as usually even if you have multiple files for the same track you probably have only 2 or so (e.g. lossy and lossless format, as I usually tag).

# Solution

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -170,6 +170,13 @@ class Album(DataObject, Item):
                     file.metadata['musicbrainz_discid'] = track_discids
                     file.update()
 
+    def get_next_track(self, track):
+        try:
+            index = self.tracks.index(track)
+            return self.tracks[index + 1]
+        except (IndexError, ValueError):
+            return None
+
     def get_album_artists(self):
         """Returns the list of album artists (as AlbumArtist objects)"""
         return self._album_artists
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -468,9 +468,15 @@ class Tagger(QtWidgets.QApplication):
             log.debug("Aborting move since target is invalid")
             return
         self.window.set_sorting(False)
-        if isinstance(target, (Track, Cluster)):
+        if isinstance(target, Cluster):
             for file in files:
                 file.move(target)
+                QtCore.QCoreApplication.processEvents()
+        elif isinstance(target, Track):
+            album = target.album
+            for file in files:
+                file.move(target)
+                target = album.get_next_track(target) or album.unmatched_files
                 QtCore.QCoreApplication.processEvents()
         elif isinstance(target, File):
             for file in files:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -463,7 +463,7 @@ class Tagger(QtWidgets.QApplication):
             log.debug("Trying to analyze %r ...", file)
             self.analyze([file])
 
-    def move_files(self, files, target):
+    def move_files(self, files, target, move_to_multi_tracks=True):
         if target is None:
             log.debug("Aborting move since target is invalid")
             return
@@ -476,7 +476,8 @@ class Tagger(QtWidgets.QApplication):
             album = target.album
             for file in files:
                 file.move(target)
-                target = album.get_next_track(target) or album.unmatched_files
+                if move_to_multi_tracks:  # Assign next file to following track
+                    target = album.get_next_track(target) or album.unmatched_files
                 QtCore.QCoreApplication.processEvents()
         elif isinstance(target, File):
             for file in files:
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -436,7 +436,8 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         self.setHeader(ConfigurableColumnsHeader(self))
         self.window = window
         self.panel = parent
-
+        # Should multiple files dropped be assigned to tracks sequentially?
+        self._move_to_multi_tracks = True
         self.setHeaderLabels([_(h) if n != '~fingerprint' else ''
                               for h, n in MainPanel.columns])
         self.restore_state()
@@ -719,7 +720,7 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         hscrollbar.setValue(xpos)
 
     @staticmethod
-    def drop_urls(urls, target):
+    def drop_urls(urls, target, move_to_multi_tracks=True):
         files = []
         new_paths = []
         tagger = QtCore.QObject.tagger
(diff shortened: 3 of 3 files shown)


Nikhil1O1:
Hello, I'm new here as well as in the opensource community. How can I help? Is there any issue a newbie can tackle?
@phw

zas:
Few remarks:
- How's defined the order of dropped files?
- Isn't there a risk to link wrong track to wrong file, because file is badly named for example?
- Default behavior could be user-defined in Preferences (instead of setting it to True, use an option defaulting to True).

Apart that, looks good to me.

phw:
>     * How's defined the order of dropped files?

In normal selection it is based on the order you see on the selected files on the left. If you select files one by one with Ctrl-click the order is the order of selection.

>     * Isn't there a risk to link wrong track to wrong file, because file is badly named for example?

Not sure how you mean this. This is all about manual matching, so there is always the risk the user matches a file to the "wrong" track. Often this is used where automatic matching fails, so e.g. the metadata comparison might show a bad score e.g. because of bad tags, but the user still chooses to match the track manually.

>     * Default behavior could be user-defined in Preferences (instead of setting it to True, use an option defaulting to True).

I'm hesitant introducing too many options. Picard already has many options which often makes it hard to figure out why a user gets a certain behavior. If we make this a fixed behavior with both options available at all times this is much easier to document and explain. And while I voiced a strong opinion on the ticket that there is a use case for occasionally dropping more than one file to a single track, I also must agree with Scott that this is seldom a useful behavior.

phw:
> Hello, I'm new here as well as in the opensource community. How can I help? Is there any issue a newbie can tackle?
> @phw

@Nikhil1O1 Great that you want to help, we can always use more . This pull request is not really the  best place to discuss this, though. If you want to help in development take a look at open issues at https://tickets.metabrainz.org/projects/PICARD/issues/ or maybe try to improve something that you personally want to get changed in Picard. It is always a good idea to get in touch and start some discussion before you start with development, especially if f it is something more complex. For this you can start a discussion on the forums (https://community.metabrainz.org/c/picard/) or also ask on IRC (https://musicbrainz.org/doc/Communication/IRC).

Sophist-UK:
IMO (for what it is worth) here is a brain dump of the possible algorithms that might be used when you drop multiple files onto a track (rather than an album):

A. Multiple files are all allocated to the single track;
B. Multiple files are allocated to all tracks, starting at the track they were dropped on, one per track in sequence that the files were originally selected;
C. Multiple files are allocated to **empty** tracks (i.e. without existing files matched), starting at the track they were dropped on, one per track in sequence that the files were originally selected;
D. Multiple files are allocated to empty tracks based on best match against the empty tracks; or
E. Multiple files are allocated to all tracks based on best match against the tracks.

I am not sure whether we need to distinguish between 2 use cases of dropping multiple files onto a track:
1) When track already has one or more files matched; AND
2) When track has no files matched

I am also not absolutely sure what **currently** happens when you drop (multiple) files on an album, but hopefully it is one of the above.

I am also unclear whether one of te above option is heaps better than all others and we only need to offer that algorithm in all circumstances of dropping multiple files onto a track, or whether we need to provide the user with Settings to allow them to define a choice, or whether we need to offer a pop-up when there is ambiguity about this?.

phw:
I thought about making the dropping "more intelligent" by doing some logic as you have outlined. But in the end I think for an action like this there should be an easy to understand and predictable outcome. If the result is something like "the first file was matched to the second track, because you dropped it there. The second would have gone to the fourth because the third track already has a file matched, but metadata matched so badly it was matched to track 6 instead. The third file was then matched to fourth track instead because this became free(!?)" then more often than not this will not do at all what the user wanted and they can hardly predict the outcome without studying a rule book first.

> I am also not absolutely sure what **currently** happens when you drop (multiple) files on an album, but hopefully it is one of the above.

It matches the files to the tracks, and in case matches are below threshold to unmatched tracks. So that's option E already being available.

zas (approved):
LGTM

Sophist-UK:
What we want to maximise is files being allocated to the right tracks. What we want to avoid are unpredictability, and if not put exactly where the user dropped them we want to avoid allocating them to the wrong tracks (any more than dropping them onto an album would get them wrong). 

My gut reaction is:

1. If you are dropping one file onto a track, put it exactly on that track.
2. If you are dropping multiple files onto a track, do as you would do if they had been dropped on the Album.
3. Have a setting to enable this approach as opposed to current behaviour of putting all files onto the track they are dropped on.

phw:
>     2\. If you are dropping multiple files onto a track, do as you would do if they had been dropped on the Album.

I don't think this is a good idea, for two reasons.

First, it would not have solved my issue at hand when I wrote this patch. The matching to the album functionality is already there, but it was no use to me. What I wanted to do was dropping some 60+ files to a release and get them matched. Existing tags have been not helpful, but they were in order. Dropping on the album did not do anything useful, everything went to unmatched files.

The second reason is exactly because of the unpredictability of this. Dropping files on a specific track is a very explicit action done by the user. See the following scenarios:

1. As a user I drop 60 files onto a single track. All files get attached to this track. Actually this is a predictable outcome, and as a user I can understand why this happens. Problem with this is it is not useful.
2. As a user I drop 60 files onto a single track. The first file gets attached to the track I dropped to, the rest of the files in order to the following tracks. This is also predictable behavior. Even if I did not know this would happen, once I have seen it I can easily understand it and use it to my advantage in the future.
3. As a user I drop 60 files onto a single track. The files get matched to seemingly random tracks or get all moved to unmatched files. None ends up on the track I dropped it to. I try again and drop on a different track, but they get matched exactly the same way, it doesn't matter where I drop them.

If I'd encounter the third case as a user I would consider this a bug. And as I wrote above I would be careful with introducing another option for this. And if it is just for the choice between a most times useless and a surprise action I'd rather do something useful by default.

I also don't really get the argument of the possible mismatches. If I drop a file on a track I want the tool to obey my wish. If it tries to be clever and basically tells me:"I don't think this is a good choice, I made a better choice for you" this quickly becomes annoying. Use cases for manually dragging files usually are about handling something manually because automatic matches did not work.

Note that dropping on an album is a different story. In this case I only drop it on the album, that the software then guesses the best tracks seems sensible.

PR #1631, opened 2020-09-21
PICARD-259: Execute tagger script when file gets moved to track

phw:
# Summary
This allows using existing file metadata and and specific file variables (such as e.g. %_bitrate%) to be used in tagger script.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-259, PICARD-23

# Solution
When a file gets moved to a track run the scripts on a combined metadata object consisting of the file's metadata + the track's metadata.

A similar patch had been proposed earlier in #170, but it got reverted. This patch is conceptually the same, but a bit simpler due to:

- we now have `track.orig_metadata` already available
- the script is not run against the track's metadata again, which also means we do not need to handle the file removal as restoring file's original metadata is already done. I don't think we need to re-run the script on the track, as the file's data is used for saving and even display is already using the file metadata in case a single file is matched. In the multi-file case it does not make sense to apply any file's metadata to the track anyway
- The previous patch caused issues with the preserved metadata (PICARD-1247). This is avoided here, since `file.copy_metadata` (which takes care of preserved tags) is run after running the scripts.

I did some measurements on the overall runtime effect of loading ~4k files in ~340 releases. Loading took ~350 seconds on my system, without a measurable difference even with a heavy script (https://gist.github.com/phw/2345486ff89538c3ee1ca621ce23959e). The reason seems to be that we have the one second delay in network requests, which is much longer than script execution takes in between.

I would appreciate if you give this some proper testing, we must avoid this having any unwanted side effect. But in my tests this worked well.

# Action

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -415,9 +415,9 @@ class NonAlbumTrack(Track):
     def _parse_recording(self, recording):
         m = self.metadata
         recording_to_metadata(recording, m, self)
-        self.orig_metadata.copy(m)
         self._customize_metadata()
         run_track_metadata_processors(self.album, m, recording)
+        self.orig_metadata.copy(m)
         self.run_scripts(m)
         self.loaded = True
         self.status = None
--- a/picard/track.py
+++ b/picard/track.py
@@ -147,6 +147,7 @@ class Track(DataObject, Item):
         self.num_linked_files = 0
         self.metadata = Metadata()
         self.orig_metadata = Metadata()
+        self.scripted_metadata = Metadata()
         self.error = None
         self._track_artists = []
 
@@ -175,8 +176,10 @@ class Track(DataObject, Item):
         metadata = Metadata(file.metadata)
         metadata.update(self.orig_metadata)
         self.run_scripts(metadata)
+        # Apply changes to the track's metadata done manually after the scripts ran
+        meta_diff = self.metadata.diff(self.scripted_metadata)
+        metadata.update(meta_diff)
         file.copy_metadata(metadata)
-        file.metadata['~extension'] = file.orig_metadata['~extension']
         file.update(signal=False)
         self.update()
 
--- a/picard/track.py
+++ b/picard/track.py
@@ -170,7 +170,12 @@ class Track(DataObject, Item):
     def update_file_metadata(self, file):
         if file not in self.linked_files:
             return
-        file.copy_metadata(self.metadata)
+        # Run the scripts for the file to allow usage of
+        # file specific metadata and variables
+        metadata = Metadata(file.metadata)
+        metadata.update(self.orig_metadata)
+        self.run_scripts(metadata)
+        file.copy_metadata(metadata)
         file.metadata['~extension'] = file.orig_metadata['~extension']
         file.update(signal=False)
         self.update()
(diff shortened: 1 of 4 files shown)


picard/track.py
```diff
@@ -170,7 +170,12 @@ def add_file(self, file):
     def update_file_metadata(self, file):
         if file not in self.linked_files:
             return
-        file.copy_metadata(self.metadata)
+        # Run the scripts for the file to allow usage of
+        # file specific metadata and variables
+        metadata = Metadata(file.metadata)
+        metadata.update(self.orig_metadata)
+        self.run_scripts(metadata)
+        file.copy_metadata(metadata)
```
kepstin, line 178: In my local version of this patch I've been using the following sequence:
```python
        if file not in self.linked_files:
            return
        file.copy_metadata(self.orig_metadata)
        file.metadata['~extension'] = file.orig_metadata['~extension']
        self.run_scripts(file.metadata)
        file.update(signal=False)
        self.update()

```
which avoids making an extra metadata object. I think this might also work better in case of a file getting unmatched and then rematched on a different track?

kepstin: I'm not 100% sure that my version of the code is actually doing the right thing here, It's been a long time since I've gone through how the loading sequence works.

phw:
This is not yet a good solution, need to rethink it. The main issue is that this always relies on the track.orig_metadata and applies the scripts again. This totally breaks manual changes to the track prior to matching files.

But using track.metadata instead would break some scripts which do give different results when run multiple times on the same data.

picard/track.py
phw: I had some reason to run `run_scripts` before `file.copy_metadata`, but I don't remember right now. For rematching it would not make a difference, same end result.

phw: Now I remember: Your code breaks the preserve tags functionality under some circumstances. If the script changes a tag this will be written to the file even if you have this tag on your preserve list. The preserving is handled by file.copy_metadata. Hence the additional work to prepare the scripted metadata first before applying it.

phw:
I think I have a working solution in 21d4f040298bf28c66bc88fb4c5b24966425f696. This will now track the manual changes to the tracks metadata and reapply them to the file.

The basic theory is in the loading and modification process of a track's metadata I outlined at https://raw.githubusercontent.com/phw/picard-diagrams/master/picard-load-metadata.svg

There are basically two main parts: The loading process and the user controlled changes.

The loading is of course loading the metadata from MB, what Picard internally does, but it also includes the changes done by plugins. After loading we store this state as `track.orig_metadata`.

This is followed by what I consider the user controlled changes. This again has two basic steps. The first is automatic and is applying the tagger scripts.  Afterwards the user can do any number of manual changes (either by manually editing tags in the metadata box or by manually running scripts).

If we want to cleanly apply the metadata to a file, we have to replay these steps. The base is the file's metadata updated by `track.orig_metadata`. That way we get the track's tags after loading. Now we have to apply the user controlled changes. First is easy, we replay the scripts on this combined metadata.

But then we also need to apply the changes manually done by the user to the track prior to matching the file. For this we remember the track metadata after the scripts have been run (`track.scripted_metadata`) and take a diff between the tracks current metadata (`track.metadata`) and `track.scripted_metadata`. These are the changes done manually. We apply them to the file as well.

Note that still file metadata and track metadata of course now can differ, that's actually the point of this PR. For one the file might contain additional metadata not in the track. Also the scripts can now make use of file variables and metadata, thus giving different results for tag values.

Is this a sound approach?

zas (approved):

picard/track.py
kepstin: This was intentional - I wanted to be able to use scripts to modify the values in "preserved" tags.

phw:
Just as information: zas and I discussed some refactorings to make handling of multiple metadata objects per e.g. track easier. I'll submit a PR for this separately.

PR #1634, opened 2020-09-25
PICARD-1951: Prevent complete hiding of metadata box or file panes

phw:
# Summary
Prevent complete hiding of metadata box or file panes

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
By completely dragging the splitter between the file panes and the metadata box up or down the user can completely hide any of those parts. This results in a UI that appears broken, and without any hint to the user that there is a missing part and that they can make it visible again.

See the screenshots in PICARD-1951

* JIRA ticket (_optional_): PICARD-1951

# Solution
Set the parts of the splitter to be not-collapsible.

Note: According to Qt5 documentation one should be able to do this with a single call to `splitter.setChildrenCollapsible(False)`, see https://doc.qt.io/qt-5/qsplitter.html#childrenCollapsible-prop

But when I tried this (with PyQt 5.15.1 on Windows 10) it had no effect. Only setting `setCollapsible` per element worked.

# Action

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -41,7 +41,7 @@ PICARD_APP_NAME = "Picard"
 PICARD_DISPLAY_NAME = "MusicBrainz Picard"
 PICARD_APP_ID = "org.musicbrainz.Picard"
 PICARD_DESKTOP_NAME = PICARD_APP_ID + ".desktop"
-PICARD_VERSION = Version(2, 5, 0, 'dev', 1)
+PICARD_VERSION = Version(2, 5, 0, 'dev', 2)
 
 
 # optional build version
--- a/picard/config_upgrade.py
+++ b/picard/config_upgrade.py
@@ -312,6 +312,12 @@ def upgrade_to_v2_5_0_dev_1(config):
     ]
 
 
+def upgrade_to_v2_5_0_dev_2(config):
+    """Reset main view splitter states"""
+    config.persist["splitter_state"] = b''
+    config.persist["bottom_splitter_state"] = b''
+
+
 def rename_option(config, old_opt, new_opt, option_type, default):
     _s = config.setting
     if old_opt in _s:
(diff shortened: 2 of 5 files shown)


zas:
I tested `mainLayout.setChildrenCollapsible(False)` and it works but ONLY after removing existing related persist settings (try after removing config file). So the issue is not with the function itself, but rather with persistence of previous settings for the splitter.

This is exactly what is reported on https://bugreports.qt.io/browse/QTBUG-35311 (not a bug).

Obviously it would be preferrable to use  `mainLayout.setChildrenCollapsible(False)` since it's cleaner, but not sure how to fix the persist issue.

phw:
Now I see, thanks for investigating. What about just clearing the peristet splitter config with a config migration?

And we probably should do the same for the file pane splitter, I think this is also collapsible which results in a seemingly broken UI.

zas:
> What about just clearing the peristet splitter config with a config migration?

Yes, it would be a solution.

phw:
I have updated the PR to solve this more compete. I also undid setting the splitter handle in between file views and metadata box to only 1 pixel. That is a small visual change, as it makes the handle actually visible. IMHO it makes it clearer that the metadata box can be resized (which is a very useful feature). Some might say it looks better without this handle in between, but I actually think it is a very minor visual difference.

phw:
Why does codacy complain for the two Option lines that they are not assigned, but does not care about all the other options defined the same way in that file?

zas:
> Why does codacy complain for the two Option lines that they are not assigned, but does not care about all the other options defined the same way in that file?

Likely because of trailing commans on those lines ;)

phw:
Missed that. Now Codacy is happy :)

zas (approved):

PR #1635, opened 2020-09-26
PICARD-1952: Use native theme in Linux

guoyunhe:
# Summary

Most Linux distributions have good Qt themes. Using native themes on Linux gives users best experience.

# Problem

Picard has a different appearance than other apps in Linux.

PICARD-1952

# Solution

Make Picard use native styles in Linux.

# Action

No

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -27,6 +27,7 @@ from PyQt5 import (
 from picard import log
 from picard.const.sys import (
     IS_HAIKU,
+    IS_LINUX,
     IS_MACOS,
     IS_WIN,
 )
@@ -55,7 +56,7 @@ class BaseTheme:
     def setup(self, app):
         # Use the new fusion style from PyQt5 for a modern and consistent look
         # across all OSes.
-        if not IS_MACOS and not IS_HAIKU:
+        if not IS_MACOS and not IS_HAIKU and not IS_LINUX:
             app.setStyle('Fusion')
 
         app.setStyleSheet(


zas:
Can you please create a matching ticket (if none exists) on https://tickets.metabrainz.org/projects/PICARD (and use PICARD-XXXX in PR title and in description)? We use Jira to generate changelogs.

guoyunhe:
Created https://tickets.metabrainz.org/browse/PICARD-1952

phw:
I think we already had a ticket for this, but can't search right now.

Anyway, this is just something we have to eecide upon. I'm personally fine with this change.

I was not involved in the original decision, but I think the idea in setting a fixed theme I think was having a unified experience and less trouble with support and theme related issues.

However, we have been relaxing this anyway, with first reverting this change for macOS, then Haiku, now allowing dark theme on Windows. And support for dark appearances is probably the biggest challenge to support anyway, and even without allowing Qt themes we face this already on Linux and other platforms.

Only thing I would then change is making the condition a `if IS_WIN:`

guoyunhe:
Using native styles is a big advantage for user experience in Linux. I assume it is also true for BSD and Windows. The "Fusion" theme looks quite out-dated. It makes people think the software is not actively developed.

phw:
Found the issue I mentioned, it's PICARD-1297. Actually in this ticket I proposed an own patch with an option to explicitly opt-in to use the system theme. But I also argumented against adding this option to not have to officially support all kind of themes. Instead users really wanting this should use one of the available alternative ways such as setting `QT_STYLE_OVERRIDE` or using the `-style` command line parameter.

I must say I'm a bit undecided on the issue :) On the one hand I would like to see Picard integrate as well as possible with the users Desktop. Being a Linux user myself I also prefer applications to use my selected theme. On the other hand I don't want to fix issues or even research issues caused by some random theme.

@zas What's your opinion on this?

If we do this, I see two ways:

1. Either we completely allow it, as proposed here. With a strong exception for Windows (see below)
2. We add a setting for this, as I had proposed, defaulting to use the standard theme (Fusion). This setting will only be available on platforms other than Windows (see below), macOS (has a very OS specific default theme) and Haiku (similar to macOS). So basically that is for Linux and BSD users with their wild growth of different desktop environments and looks.

> I assume it is also true for [...] Windows.

I think the Fusion theme is a pretty good choice on Windows, we should stick with it. On Windows the OS does not offer that kind of theming support as you have it on Linux. And only if we stick to a fixed theme we can have e.g. the dark theme changes we recently implemented for Windows.

If we go the route of changing the theme on all other OS, I think the cleanest solution would be to reimplement the `setup` method in the `WindowsTheme` class and only do the `app.setStyle('Fusion')` there.

> The "Fusion" theme looks quite out-dated.

Mmh, I don't agree :)

zas:
> @zas What's your opinion on this?

No strong opinion, I'm fine with Fusion on Linux and you get a point about "On the other hand I don't want to fix issues or even research issues caused by some random theme." ;)

rdswift:
Personally, I like the consistency that fusion provides across platforms, especially when designing settings screens for plugins and such.  I've found that layouts can vary a bit otherwise.

phw:
Ok, I updated my code changes from last year to have an option for this and put them up as an alternative proposal: #1638

I also noticed that overriding the theme as I described above no longer works since the 2.4 release.

phw:
@guoyunhe I am closing this as we have merged #1638 instead. So the user will be able to opt-in to use the system configured Qt theme. Thanks a lot for getting this issue rolling again!

PR #1637, opened 2020-09-28
PICARD-1908: Show release search dialog for release-group URLs

phw:
# Summary
This allows dragging release-group URLs into Picard or entering them in the search field. The release search dialog will be shown filtered by the release-group MBID.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1908

# Solution
Show the AlbumSearch dialog with an advanced query to filter by release group. Extend the search dialogs to force using advanced query syntax without changing the users preferences.

# Action
We could extend this to other entities (e.g. work URL -> search for matching recordings, artist URL -> search for matching releases). But release group was the most obvious to me, especially as I have seen users trying to load release group URLs instead of release URLs more than once. Thoughts?

### DIFF
--- a/picard/browser/filelookup.py
+++ b/picard/browser/filelookup.py
@@ -44,6 +44,8 @@ from picard.util import (
     webbrowser2,
 )
 
+from picard.ui.searchdialog.album import AlbumSearchDialog
+
 
 class FileLookup(object):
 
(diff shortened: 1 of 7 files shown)


zas (approved):
Tested, nice improvement!

PR #1638, opened 2020-09-28
PICARD-1952: Add option to use system theme on systems other than macOS and Windows

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

In #1635 it is proposed to use the system Qt theme on Linux. This is an alternative proposal to still default to the Fusion theme but add an option to opt-in using the system theme.

# Problem

* JIRA ticket (_optional_): PICARD-1952

Picard uses the Fusion themes on most platforms, with the notable exception of macOS where Qt's "macintosh" theme is used.

On Linux it is common to configure different Qt themes, so some users want Picard to obey this. On the other hand we as the devlopers can't guarantee that the UI remains pretty and usable when an arbitrary theme is being used.

# Solution

Still default to the Fusion theme, but add an option "Use system theme" in Options > User interface . When selected, Picard will use the system's default theme. When this is being activated a warning is shown to the user:

> Please note that using the system theme might cause the user interface to be not shown correctly. If this is the case disable the "Use system theme" option to use Picard's default theme again.

**Note:** There are already ways for the user on Linux to affect the look of Picard.

- On KDE Qt5 respects the color palette settings. So if  the user has configured KDE to use the dark variant of the Breeze theme, Picard will still use the Fusion theme but it will be dark colored.
- One can override the theme by setting `QT_STYLE_OVERRIDE`, e.g. one could start Picard with:

      QT_STYLE_OVERRIDE=Breeze picard
- One can override the theme by using the `-style` command line parameter, e.g. one could start Picard with:

      picard -style=Breeze

We could also rely on those mechanisms only.

**Update:** I just noticed that the override methods above stopped working with the 2.4 release. It works in 2.3.2 and earlier. In this case we probably should restore support for this.

# Action

Make a decision!

### DIFF
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -220,6 +220,11 @@ class InterfaceOptionsPage(OptionsPage):
         if new_theme_setting != config.setting["use_system_theme"]:
             restart_warning_title = _('Theme changed')
             restart_warning = _('You have changed the application theme. You have to restart Picard in order for the change to take effect.')
+            if new_theme_setting:
+                restart_warning += '\n\n' + _(
+                    'Please note that using the system theme might cause the user interface to be not shown correctly. '
+                    'If this is the case disable the "Use system theme" option to use Picard\'s default theme again.'
+                )
         elif new_language != config.setting["ui_language"]:
             restart_warning_title = _('Language changed')
             restart_warning = _('You have changed the interface language. You have to restart Picard in order for the change to take effect.')
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -40,6 +40,11 @@ from PyQt5.QtCore import QStandardPaths
 
 from picard import config
 from picard.const import UI_LANGUAGES
+from picard.const.sys import (
+    IS_HAIKU,
+    IS_MACOS,
+    IS_WIN,
+)
 from picard.util import icontheme
 
 from picard.ui import PicardDialog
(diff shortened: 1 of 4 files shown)


phw:
@zas @samj1912 @mineo @rdswift @Sophist-UK 

I'd like to hear some opinions on this. Let's assume that using Macintosh theme on macOS and Fusion theme on Windows are given and we don't change that. Let's consider how to deal with other platforms. I see the following options:

1. Strictly stick to the "Fusion" theme (current status quo in Picard 2.4)
2. Use the "Fusion" theme, but allow the override via `QT_STYLE_OVERRIDE` or `-style` argument (status quo up to Picard 2.3.2)
3. Use the "Fusion" theme by default, but have an option to opt-in to use the system theme (in addition to allowing the override) (this PR)
4. Always use the system theme (#1638)

zas:
> Use the "Fusion" theme by default, but have an option to opt-in to use the system theme (in addition to allowing the override) (this PR)

Imho, the best solution.

rdswift:
I agree with @zas on using the fusion theme by default but allow using the system theme and override.

mineo (dismissed):
In my experience, while theming on Linux is possible, the results are usually not great and look horrible in random applications :( Therefore making this opt-in makes sense to me.

zas (dismissed):

samj1912 (approved):

phw:
Thanks a lot for all the feedback. So let's go with the opt-in solution.

I resolved the merge conflicts and merged the changes.

PR #1639, opened 2020-09-28
PICARD-1955: Use built-in search by default

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Make the built-in search enabled by default.

# Problem

* JIRA ticket (_optional_): PICARD-1955

The built-in search is easier to use for new users and does not rely on the browser integration (which causes frequent trouble for new users and needs quite some technical explaining if it does not work as expected). Also the "Lookup in browser" functionality is still available separately, even from within the search dialogs.

We keep the option. The browser based search can offer more flexibility, as it allows using all the search functionality the MB website has to offer. Hence advanced users might want to turn off the built-in search.

Existing users are not affected, their config remains untouched.

### DIFF
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -122,7 +122,7 @@ class InterfaceOptionsPage(OptionsPage):
     options = [
         config.BoolOption("setting", "toolbar_show_labels", True),
         config.BoolOption("setting", "toolbar_multiselect", False),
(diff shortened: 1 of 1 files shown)


zas (approved):
It makes sense.

PR #1640, opened 2020-09-28
PICARD-1956: Prevent the same file being added to a cluster more than once

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1956

If one selects an album and files inside this album and drags them onto the cluster item, the created cluster contains the selected files twice. This leads to further confusion of the UI, as moving the cluster files to an album will still leave the cluster with a reference to the duplicated files.

# Solution
Prevent adding the same file twice to a cluster.

# Action

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2006-2008, 2011 Lukáš Lalinský
 # Copyright (C) 2008 Hendrik van Antwerpen
 # Copyright (C) 2008 Will
-# Copyright (C) 2010-2011, 2014, 2018-2019 Philipp Wolfer
+# Copyright (C) 2010-2011, 2014, 2018-2020 Philipp Wolfer
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
 # Copyright (C) 2012 Wieland Hoffmann
@@ -115,18 +115,22 @@ class Cluster(QtCore.QObject, Item):
             self.related_album.update()
 
     def add_files(self, files):
+        added_files = []
         for file in files:
+            if file in self.files:
+                continue
+            added_files.append(file)
             self.metadata.length += file.metadata.length
             file._move(self)
             file.update(signal=False)
             if self.can_show_coverart:
                 file.metadata_images_changed.connect(self.update_metadata_images)
-        self.files.extend(files)
+        self.files.extend(added_files)
         self.metadata['totaltracks'] = len(self.files)
-        self.item.add_files(files)
+        self.item.add_files(added_files)
         if self.can_show_coverart:
-            add_metadata_images(self, files)
-        self._update_related_album(added_files=files)
+            add_metadata_images(self, added_files)
+        self._update_related_album(added_files=added_files)
 
     def add_file(self, file):
         self.add_files([file])
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -41,6 +41,7 @@ from heapq import (
     heappush,
 )
 import ntpath
+from operator import attrgetter
 import re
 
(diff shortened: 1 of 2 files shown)


picard/cluster.py
```diff
@@ -115,18 +115,22 @@ def _update_related_album(self, added_files=None, removed_files=None):
             self.related_album.update()
 
     def add_files(self, files):
+        added_files = []
```
zas, line 118: It would be better to use a set for this, it is significantly faster for doing `file in self.files`. Does order matter?

zas (changes requested):

picard/cluster.py
phw: I assumed order would matter for a cluster, but maybe it doesn't. I need to check

phw: Cluster files are sorted in https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L823

So even without column sorting they show up in order. I think generally changing `Cluster.files` to a set could also have some side effects or break someone's plugin. Otherwise it would work in general, I tested it.

Making `added_files` a set would not make much difference I think, we don't search in it. It's just a list that gets appended to.

We could in this method do a `existing_files= set(self.files)` call, and then use existing_files in `if file in existing_files`:

```
def add_files(self, files):
    added_files = []
    existing_files= set(self.files)
    for file in files:
        if file in existing_files:
            continue
        added_files.append(file)
        ...
```

But I doubt there is much difference.

zas: Profile it ;) So we can decide.

phw:
So I benchmarked it and did some changes in 8bd6e500d52dcccc34ce5a10ab0ccf0096f85fd7.

First about using `list.append` vs. `set.add` there is no real difference. Both are O(1), in my benchmarking `list.append` was about 2% faster on my system, but there was quite some variance so it's not really conclusive.

Now about the interesting part. I took two lists with 1 million elements, which overlap in half:

```python
N = 1000 * 1000

existing = list(range(0, N))
new = list(range(N//2, N+N//2))
```

The goal is to get everything which is in `new` but not in `existing`. First attempt is with converting `existing` to a set and using `in` check:

```python
l = []
existing = set(existing)

for i in new:
  if i in existing:
    l.append(i)
```

100 runs of this averaged to 0.2504 seconds. Now the same

Basically the same with `set.difference` gives better results:

```python
existing = set(existing)
new = set(new)
l = new - existing
```

100 runs took only 0,1032 seconds.

Not doing the set conversion and just use `in` check on the list is horrible slow:

```python
l = []
for i in new:
  if i in existing:
    l.append(i)
```

This has I think a O(N^2) complexity. With `N=10000` it took 0.9830 seconds, with `N=100000` it took 95.6751 seconds and I did not dare to test it with 1 million :D

The differences shrink away of course with small `N`, but still with `N=10` conversion to a set first is still 25% faster then keeping the list for a for loop with in check. If you use set arithmetic it is even 65% faster.

So lesson learned: Whenever you think to do checking if elements are in a list inside a loop, convert the list to a set first. And consider using set arithmetic if it makes sense.

In 8bd6e500d52dcccc34ce5a10ab0ccf0096f85fd7 I adapted the code accordingly. I kept the sorting, but moved it from tagger.py to inside the `Cluster.add_files` method, which has the advantage the it will run on actually added files only. The sorting is needed for the `self.item.add_files(added_files)` call. That way the cluster's item in the tree view show up in proper order in the unsorted case.

zas:
> This has I think a O(N^2) complexity.

Yes.
Your last changes look good to me.

zas (approved):

PR #1641, opened 2020-09-29
PICARD-1888: Cluster files attached to tracks with original metadata

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

When files currently attached to tracks are dragged to the Cluster list, they get clustered with their new metadata from the tracks. Ensure they get their original data reset prior to doing the clustering.

# Problem

* JIRA ticket (_optional_): PICARD-1888

# Solution
Remove the files from the track prior to clustering them.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -486,6 +486,10 @@ class Tagger(QtWidgets.QApplication):
         elif isinstance(target, Album):
             self.move_files_to_album(files, album=target)
         elif isinstance(target, ClusterList):
+            for file in files:
+                if isinstance(file.parent, Track):
+                    file.parent.remove_file(file)
+                    QtCore.QCoreApplication.processEvents()
             self.cluster(files)
         self.window.set_sorting(True)
 


phw:
I had been struggling a bit with the proper solution, but I think this is the cleanest ways. Alternatives I considered:

- Moving them to unclustered first. But this ads quite some overhead and slows things down. It is also visually odd.
- Instead of `file.parent.remove_file` just do the orig_metadata restore: `file.copy_metadata(file.orig_metadata, preserve_deleted=False)`. At this stage it would be faster, but it duplicates internal logic of `Track.remove_file` and `Track.remove_file` gets called anyway later on.

zas (approved):
It makes sense (after reading the discussion in the bug report). LGTM.

phw:
Yes, it is a bit tricky.

Btw, looking into this makes another issue visible: If you load a file to the unmatched tracks, make manual tag changes, move it over to a track and back again your manual changes get lost. The original data gets restored. I think from a users perspective it would be good to also have the metadata directly before the file gets matched to the track. So we might want to store this metadata as well. Kind of gets back to our recent discussion about better handling multiple versions of the same metadata.

PR #1643, opened 2020-09-29
PICARD-1782: Allow locking itemview table headers

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Allow locking the table headers in the main view.

Locked headers cannot be changed. This prevents accidental resorting and changes to the list views. The lock state of each table header is persisted.

# Problem

* JIRA ticket (_optional_): PICARD-1782

# Solution

Implement locking in `TristateSortHeaderView` and add an action to the header's context menu to toggle this.

grafik

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -394,20 +394,29 @@ class ConfigurableColumnsHeader(TristateSortHeaderView):
 
     def contextMenuEvent(self, event):
         menu = QtWidgets.QMenu(self)
+        parent = self.parent()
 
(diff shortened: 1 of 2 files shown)


zas (approved):
Tested, LGTM.
I wonder if we could display a tooltip when mousing over locked headers to remind user those are locked and how to unlock, but patch is mergeable as is imho.

phw:
Good idea with the tooltip

PR #1646, opened 2020-10-02
macOS: Handle OS version 11.0 in package script

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

In macOS packaging script support detection of version 11.0 for notarization support.

# Problem

There are actual two issues with the version detection for macOS 11 Big Sur:

1. Have notarization enabled for version 11. This was discussed in https://github.com/metabrainz/picard/pull/1351, but not added since we did not know anything about major version 11 at that point.
2. Support versions with only two segments. The versioning code assumed the version always had 3 segments (0.0.0), but Big Sur reports itself as being 11.0 only. This caused the MACOS_VERSION_MAJOR variable to be 11.0. The changed script will both properly detect the major version for 11.0.0 and 11.0.

# Solution

# Action

### DIFF
--- a/scripts/package/macos-package-app.sh
+++ b/scripts/package/macos-package-app.sh
@@ -44,7 +44,7 @@ if [ -f scripts/package/appledev.p12 ] && [ -n "$appledev_p12_password" ]; then
 fi
 
 # Submit app for notarization on macOS >= 10.14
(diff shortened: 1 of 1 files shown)


phw:
Note that this PR does not enable us to actually build on Big Sur. Building on Big Sur breaks because of PyInstaller not working there yet. But this PR prepares the proper handling of the versions.

zas (approved):

PR #1647, opened 2020-10-02
PICARD-1882: Generic detection of dark color scheme in use

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Try to automatically switch the syntax highighting color on a dark color scheme.

# Problem

* JIRA ticket (_optional_): PICARD-1882

Picard might use a dark color palette. This can happen e.g. on KDE (which sets a color palette based on the selected theme), or  by the user actually switching the Qt style once we have #1638 merged. Dark background colors makes reading the default syntax highlighting theme difficult. We have an optimized version for dark colors and should use this.

# Solution
This PR tries to dynamically detect dark background colors being used by actually looking at the lightness of the base color set in the palette.

Note: If the color is somewhere in between both themes might actually look bad. But this would be more an issue with the theme itself, the issue is not limited to our use case. Actual themes avoid this as it provides bad contrast with both light and dark fonts.

### DIFF
--- a/picard/ui/options/scripting.py
+++ b/picard/ui/options/scripting.py
@@ -120,10 +120,9 @@ class ScriptingDocumentationDialog(PicardDialog):
             postprocessor=process_html,
         )
 
-        syntax_theme = theme.get_syntax_theme()
         html = DOCUMENTATION_HTML_TEMPLATE % {
             'html': "<dl>%s</dl>" % funcdoc,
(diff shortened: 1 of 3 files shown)


phw:
The following screenshot shows the scripting options with dark color palette on KDE with this patch:

grafik

And here in current Picard without this patch:

grafik

zas (approved):

PR #1651, opened 2020-10-04
PICARD-259: Fix clear_existing_tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Restore the "clear existing tags" functionality, which broke when we enabled scripting to use the file's metadata in #1631

# Problem

* JIRA ticket (_optional_): PICARD-259

# Solution
If "clear existing tags" is enabled, do not use the files metadata in the scripting. Still allow use of the technical variables (~bitrate, ~channels etc.).

Question/Discussion: This also means you can't use the file's existing variables in your script if you have clear existing tags enabled, but this kind of makes sense to me. If we actually would want to allow access to metadata for the script without actually making it part of the file's metadata I would have a solution for this ready: A metadata proxy object that holds the main metadata for read write, but also gives access to a secondary readonly metadata object. We could run the scripting on this like:

```
metadata = Metadata(self.orig_metadata)
metadata_proxy = MetadataProxy(metadata, readonly=file.metadata)
self.run_scripts(metadata_proxy)
# ...
file.copy_metadata(metadata)
```

That means the metadata object would only contain track.orig_metadata and what the script then changes in it, but everything else that is in file.metadata is not part of it but available for the script to read from.

What is expected behavior?

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -87,6 +87,9 @@ from picard.util.tags import PRESERVED_TAGS
 from picard.ui.item import Item
 
 
+FILE_INFO_TAGS = ('~bitrate', '~sample_rate', '~channels', '~bits_per_sample', '~format')
+
+
 class File(QtCore.QObject, Item):
 
     metadata_images_changed = QtCore.pyqtSignal()
@@ -354,8 +357,7 @@ class File(QtCore.QObject, Item):
             self.base_filename = os.path.basename(new_filename)
             length = self.orig_metadata.length
             temp_info = {}
-            for info in ('~bitrate', '~sample_rate', '~channels',
-                         '~bits_per_sample', '~format'):
+            for info in FILE_INFO_TAGS:
                 temp_info[info] = self.orig_metadata[info]
             # Data is copied from New to Original because New may be
             # a subclass to handle id3v23
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -37,6 +37,7 @@ from collections.abc import (
     Iterable,
     MutableMapping,
 )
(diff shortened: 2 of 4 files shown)


zas:
> This also means you can't use the file's existing variables in your script if you have clear existing tags enabled

I think most users expect to be able to use existing file metadata in their scripts, and clear existing tags should happen at write time (so all metadata, whatever is the source, is usable during the whole process).



> A metadata proxy object that holds the main metadata for read write, but also gives access to a secondary readonly metadata object.

We already discussed that for another issue, and I think we definitively need such feature. Good time to introduced it?

phw:
> I think most users expect to be able to use existing file metadata in their scripts, and clear existing tags should happen at write time

I'm still not fully sure about this. This is a screenshot of a file that would have two tags getting removed by clear existing tags:

grafik

Would the user really expect to use comment or lyrics in this case in the script? Maybe if you think of it in terms of the script using the original data to come up with the final tag results. Maybe not if you expect the script to run on the data now available.

But writing this I can see that option 1 maybe really more useful. I'll adapt the code to it.

zas (dismissed):

phw:
@zas Can you look over https://github.com/metabrainz/picard/pull/1651/commits/a7dac890fbcf65d4d2e57c85f0e962fa9d7d3cb4 ? This fixes some incompatibility between Metadata and MultiMetadataProxy. With this I'm quite happy with the functionality and would merge it.

zas (approved):

PR #1653, opened 2020-10-06
PICARD-1963: Avoid stack overflow when loading large amount of files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Avoid a possible crash due to a stack overflow when loading files.

# Problem
When loading a large list of files this could result in a stack overflow,
since tagger.move_files calls QCoreApplication.processEvents, which in
return might process an event to call tagger.move_files again, resulting
in a deep recursion. See PICARD-1963 for a more detailed explanation.

* JIRA ticket (_optional_): PICARD-1963

# Solution

Refactor the special case of moving a single file into a separate method.

Note that instead removing the processEvents calls from tagger.move_files is not really an option. This method is basically a large UI update task, so needs to run in the main thread. And not processing events in between makes Picard unresponsive in other cases.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -453,7 +453,7 @@ class Tagger(QtWidgets.QApplication):
                 return
 
         if target:
-            self.move_files([file], target)
+            self.move_file(file, target)
         else:
             self.unclustered_files.add_file(file)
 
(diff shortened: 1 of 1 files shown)


zas (approved):
Good catch! LGTM

PR #1654, opened 2020-10-06
Add process_events_iter to allow iteration while processing events

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Refactor iterations using QCoreApplication.processEvents to not run this on every iteration.

# Problem

# Solution
EventProcessingIterator acts as a wrapper around another iterable to allow running QCoreApplication.processEvents.
Instead of running processEvent on every iteration it is done in steps.

This allows us to weight responsiveness against total processing time.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -107,12 +107,12 @@ from picard.track import (
     Track,
 )
 from picard.util import (
-    EventProcessingIterator,
     check_io_encoding,
     decode_filename,
     encode_filename,
     is_hidden,
     mbid_validate,
+    process_events_iter,
     thread,
     uniqify,
     versions,
@@ -469,21 +469,21 @@ class Tagger(QtWidgets.QApplication):
             return
         self.window.set_sorting(False)
         if isinstance(target, Cluster):
-            for file in EventProcessingIterator(files):
+            for file in process_events_iter(files):
                 file.move(target)
         elif isinstance(target, Track):
             album = target.album
-            for file in EventProcessingIterator(files):
+            for file in process_events_iter(files):
                 file.move(target)
                 if move_to_multi_tracks:  # Assign next file to following track
                     target = album.get_next_track(target) or album.unmatched_files
         elif isinstance(target, File):
-            for file in EventProcessingIterator(files):
+            for file in process_events_iter(files):
                 file.move(target.parent)
         elif isinstance(target, Album):
             self.move_files_to_album(files, album=target)
         elif isinstance(target, ClusterList):
-            for file in EventProcessingIterator(files):
+            for file in process_events_iter(files):
                 if isinstance(file.parent, Track):
                     file.parent.remove_file(file)
             self.cluster(files)
@@ -523,7 +523,7 @@ class Tagger(QtWidgets.QApplication):
             new_files.sort(key=lambda x: x.filename)
             self.window.set_sorting(False)
             self._pending_files_count += len(new_files)
-            for file in EventProcessingIterator(new_files):
+            for file in process_events_iter(new_files):
                 file.load(partial(self._file_loaded, target=target))
 
     @staticmethod
@@ -820,7 +820,7 @@ class Tagger(QtWidgets.QApplication):
         for name, artist, files in Cluster.cluster(files, 1.0, self):
             cluster = self.load_cluster(name, artist)
             cluster_files[cluster].extend(files)
-        for cluster, files in EventProcessingIterator(cluster_files.items(), steps=2):
+        for cluster, files in process_events_iter(cluster_files.items(), steps=2):
             cluster.add_files(files)
         self.window.set_sorting(True)
 
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -107,6 +107,7 @@ from picard.track import (
     Track,
 )
 from picard.util import (
+    EventProcessingIterator,
     check_io_encoding,
     decode_filename,
     encode_filename,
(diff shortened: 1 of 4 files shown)


phw:
@zas This is especially to address your concerns in the past :) And indeed it does some good things to processing time.

In my tests I loaded 4k files, and clustered them later. Files where loaded with ignoring MBIDs, so no album loading happening.

I tested this with several step sizes. The times below are after all file loading has finished (pending file is 0) and for clustering after the cluster calculation has finished and after files have been added to new clusters. These times were taken without user interaction, so processEvent was not busy.

```
steps=1
files loaded 48.11406111717224
clusters calculated 0.7927291393280029
clusters created 16.309009075164795


steps=2
files loaded 35.83295226097107
clusters calculated 0.7678284645080566
clusters created 13.657451629638672


steps=4
files loaded 29.2277774810791
clusters calculated 0.7673459053039551
clusters created 12.670256853103638


steps=8
files loaded 25.040839195251465
clusters calculated 0.787574052810669
clusters created 12.554292917251587


steps=10
files loaded 23.15113139152527
clusters calculated 0.7411868572235107
clusters created 11.94653582572937


steps=100
files loaded 19.863458156585693
clusters calculated 0.9040839672088623
clusters created 11.061842441558838


steps=1000
files loaded 23.454566955566406
clusters calculated 0.6948745250701904
clusters created 11.503607273101807
```

As one can see the processing times shrink noticeable with slight increases in step size, but at some point there is no further benefit.

For the final setup I went with steps=10 by default, because this still felt good to use, with maybe a slightly more noticeable delay at some points. We could go with 8 maybe. With larger steps the UI lagged clearly. But for adding the files to clusters I chose a step of 2 (https://github.com/metabrainz/picard/pull/1654/files#diff-b28e51608eae88e428219f416397b86dR824), because everything above made the UI laggy (a single cluster.add_files call is rather heavy).

```
files loaded 22.853195667266846
clusters calculated 0.7508537769317627
clusters created 13.734225273132324
```

picard/util/__init__.py
```diff
@@ -88,6 +88,29 @@ def unlock(self):
         self.__lock.unlock()
 
 
+class EventProcessingIterator:
+    """
+    Wrapper around an existing iterator, that calls QCoreApplication.processEvents()
+    in certain iteration steps.
+    """
```
Sophist-UK, line 95: I think you need a note here that this enumerator should only be used in code that runs on the main thread and not on code which runs in a worker thread.

Sophist-UK:
Nice refactor!!

picard/util/__init__.py
phw: Good point, addressed this in https://github.com/metabrainz/picard/pull/1654/commits/4dbe09445cabde0007867f166a8a5b7a31c6779e . Also rewrote the iterator using a generator.

picard/util/__init__.py
```diff
@@ -88,6 +88,19 @@ def unlock(self):
...
+    Creates an iterator over iterable that calls QCoreApplication.processEvents()
+    in certain iteration steps.
+
+    This must only be used in the main thread.
+    """
+    for i, item in enumerate(iterable):
+        if (i + 1) % steps == 0:
+            QtCore.QCoreApplication.processEvents()
+        yield item
```
zas, line 101: Using `steps` has low overhead, but it's hard to estimate the correct value, as each step duration is not really predictable (as it depends on many things, like code executed and hardware the code runs on).

I wonder about acceptable overhead and if we could use a timer.
Something like:

```python
from time import monotonic, sleep


def process_events_iter(iterable, interval=0.1):
    if interval:
        start = monotonic()
    for item in iterable:
        if interval:
            now = monotonic()
            delta = now - start
            if delta > interval:
                start = now
                print("call QtCore.QCoreApplication.processEvents()")
        yield item


for i in process_events_iter(range(0, 20), interval=0.15):
    print(i)
    sleep(0.05)

for i in process_events_iter(range(0, 20), interval=0.4):
    print(i)
    sleep(0.05)
```

Since code processed in each step is likely to be much slower than call to `monotonic()` it could be acceptabe. What do you think?

phw: Let's try it

phw: I don't get satisfying results with the timer approach. When loading the files the UI stotters a lot initially. I could not find any interval that works well. I think we should stick with the counter.

zas: > I don't get satisfying results with the timer approach. When loading the files the UI stotters a lot initially. I could not find any interval that works well. I think we should stick with the counter.

That's weird.
I did some tests, and it seems to me the reason is because:
- when iteration is very short, processEvents() isn't called (because delta > interval doesn't happen)

Here is a modified test code:

```python
def process_events_iter(iterable, steps=None, interval=0.1):
    """
    Creates an iterator over iterable that calls QCoreApplication.processEvents()
    in certain iteration steps.

    This must only be used in the main thread.
    """
    if interval:
        start = monotonic()
    called = False
    _id = id(iterable)
    print("start id=%d len=%d interval=%f" % (_id, len(iterable), interval))

    def process_events():
        print("===== call QtCore.QCoreApplication.processEvents() ====")

    for item in iterable:
        if interval:
            now = monotonic()
            delta = now - start
            print("id=%d delta=%f" % (_id, delta))
            if delta > interval:
                start = now
                process_events()
                called = True
        yield item
    if not called:
        process_events()
```

It calls `process_events()` at least once at end of iteration, or every 0.1 seconds.
I tested with 1370 files moved to clusters.

Sophist-UK: I would not bother keeping track of whether processEvents() was ever called, but instead to simply call it regardless at the end of the loop.

picard/tagger.py
```diff
@@ -824,9 +827,8 @@ def cluster(self, objs):
         for name, artist, files in Cluster.cluster(files, 1.0, self):
             cluster = self.load_cluster(name, artist)
             cluster_files[cluster].extend(files)
-        for cluster, files in cluster_files.items():
+        for cluster, files in process_events_iter(cluster_files.items(), steps=2):
```
zas, line 830: `steps` parameter should be removed with new code

zas (changes requested):

picard/tagger.py
phw: Fixed that

zas (approved):
LGTM

PR #1656, opened 2020-10-09
PICARD-1964: Support right-to-left text flow in scripting docs

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fixed rendering of scripting documentation for scripts running from right to left.

# Problem

* JIRA ticket (_optional_): PICARD-1964

# Solution
Set the general text direction to right-to-left. Also insert a left order mark into code blocks, to fix rendering of code snippets.

The latter is necessary, because in a code block like "$set(...)" the leading $ would still be rendered in RTL mode, that means it would visually turn into "get(a)$". Take e.g. the following without the left-order-mark:

<span dir="rtl">עִבְרִית $get(a) עִבְרִית</span>

And with left-order-mark:

<span dir="rtl">עִבְרִית &lrm;$get(a) עִבְרִית</span>

Example of how it looks like:

grafik

# Action

### DIFF
--- a/picard/ui/options/scripting.py
+++ b/picard/ui/options/scripting.py
@@ -67,7 +67,8 @@ dt {
     color: %(script_function_fg)s
 }
 dd {
-    padding: 50px;
+    /* Qt does not support margin-inline-start, use margin-left/margin-right instead */
+    margin-%(inline_start)s: 50px;
     margin-bottom: 50px;
 }
 code {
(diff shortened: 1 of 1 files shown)


zas (approved):
A bit hacky thanks to Qt, but LGTM.

PR #1658, opened 2020-10-10
PICARD-1390: Configurable network transfer timeout

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Apply a network timeout of 30 seconds by default to avoid network requests hanging indefinitely.

# Problem

* JIRA ticket (_optional_): PICARD-1390, PICARD-1969

Currently if a network requests hangs and does not respond it can block Picard indefinitely. Picard becomes unusable and requires a restart.

# Solution
Apply a network transfer timeout. The default is 30 seconds, but this can be configured by the user in Options > Advanced > Network. The timeout aborts the request if no data has been transferred for the given period. That means a large download request does not get aborted if it takes 30 seconds, as long as it continuously delivers data. 

If possible this uses the native transfer timeout implementation available since Qt 5.15. But there is a fallback implementation for older Qt versions. As far as I can tell from my testing their behavior is identical.

In https://github.com/metabrainz/picard/commit/e73c33e3751d461e11ebb354f3cd39a72b3fb33f I also fixed a separate issue of the networking options where it would update the browser_integration_port even if settings don't get saved (see PICARD-1969). This is technically an unrelated change, but the fix is so minor I did not want to open a separate but conflicting PR for it. Let's keep our lives easier ;)


picard/webservice/__init__.py
```diff
@@ -338,9 +338,39 @@ def _send_request(self, request, access_token=None):
...
+        timer.setTimerType(QtCore.Qt.PreciseTimer)
+        timer.timeout.connect(partial(self._timeout_request, reply))
+        reply.finished.connect(timer.stop)
+        reply.uploadProgress.connect(partial(self._reset_transfer_timeout, timer))
+        reply.downloadProgress.connect(partial(self._reset_transfer_timeout, timer))
+        timer.start(config.setting['network_transfer_timeout'])
+
+    @staticmethod
+    def _reset_transfer_timeout(timer, bytesTransferred, bytesTotal):
+        timer.start(config.setting['network_transfer_timeout'])
```
zas, line 367: We could store `config.setting['network_transfer_timeout']` as a property of `WsRequest` instead. It will be set for the starting request and then only change for new requests but I don't think it is much an issue.

```diff
diff --git a/picard/webservice/__init__.py b/picard/webservice/__init__.py
index 2f8d56ba..a1784904 100644
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -150,6 +150,7 @@ class WSRequest(QNetworkRequest):
         self.priority = priority
         self.important = important
         self._high_prio_no_cache = False
+        self.timeout = config.setting['network_transfer_timeout']
 
         self.access_token = None
         self._init_headers()
@@ -340,16 +341,17 @@ class WebService(QtCore.QObject):
         data = request.data
         if False and hasattr(self.manager, 'setTransferTimeout'):  # Available since Qt 5.15
             native_timeout = True
-            request.setTransferTimeout(config.setting['network_transfer_timeout'])
+            request.setTransferTimeout(request.timeout)
         else:
             native_timeout = False
         reply = send(request, data.encode('utf-8')) if data is not None else send(request)
+        self._active_requests[reply] = request
         if not native_timeout:
             self._start_transfer_timeout(reply)
-        self._active_requests[reply] = request
 
     def _start_transfer_timeout(self, reply):
-        if not config.setting['network_transfer_timeout']:
+        timeout = self._active_requests[reply].timeout
+        if not timeout:
             return
         # Fallback implementation of a transfer timeout for Qt < 5.15.
         # Aborts a request if no data gets transferred for TRANSFER_TIMEOUT milliseconds.
@@ -358,13 +360,14 @@ class WebService(QtCore.QObject):
         timer.setTimerType(QtCore.Qt.PreciseTimer)
         timer.timeout.connect(partial(self._timeout_request, reply))
         reply.finished.connect(timer.stop)
-        reply.uploadProgress.connect(partial(self._reset_transfer_timeout, timer))
-        reply.downloadProgress.connect(partial(self._reset_transfer_timeout, timer))
-        timer.start(config.setting['network_transfer_timeout'])
+        reset_func = partial(self._reset_transfer_timeout, timer, timeout)
+        reply.uploadProgress.connect(reset_func)
+        reply.downloadProgress.connect(reset_func)
+        timer.start(timeout)
 
     @staticmethod
-    def _reset_transfer_timeout(timer, bytesTransferred, bytesTotal):
-        timer.start(config.setting['network_transfer_timeout'])
+    def _reset_transfer_timeout(timer, timeout, bytesTransferred, bytesTotal):
+        timer.start(timeout)
 
     @staticmethod
     def _timeout_request(reply):
```

ui/options_network.ui
```diff
@@ -228,6 +228,47 @@
      </layout>
     </widget>
    </item>
+   <item>
+    <layout class="QHBoxLayout" name="horizontalLayout_3">
+     <item>
+      <widget class="QLabel" name="label_3">
+       <property name="text">
+        <string>Request timeout:</string>
```
zas, line 236: For translations, I think it would be better to have `Request timeout in seconds:` instead of 2 labels.

zas (changes requested):

ui/options_network.ui
phw: Changed that

picard/webservice/__init__.py
phw: I first thought we might have multiple WebService instances running, but actially we should have only one, so I added a set_transfer_timeout method and update it on config change. We do the same for proxy settings. It should be fine, because even if a plugin should create other short lived instances it would work.

Mid-term we should think about having an even based setting system, so objects can subscribe to config changes. Would avoid dependencies on option UI on other business logic objects.

picard/webservice/__init__.py
```diff
@@ -331,6 +332,14 @@ def setup_proxy(self):
                 proxy.setPassword(config.setting["proxy_password"])
         self.manager.setProxy(proxy)
 
+    def set_transfer_timeout(self, timeout):
+        timeout_ms = timeout * 1000
+        if hasattr(self.manager, 'setTransferTimeout'):  # Available since Qt 5.15
+            self.manager.setTransferTimeout(timeout_ms)
+            self._transfer_timeout = 0
+        else:  # Use fallback imßplementation
```
zas, line 340: `imßplementation`, so german ;)

zas (changes requested):
Apart the typo in comment, looks good to me

picard/webservice/__init__.py
phw: lol, fixed :D

zas (approved):

PR #1659, opened 2020-10-16
PICARD-1972: When matching files to tracks preserve track length

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fixes track times getting reset to 0 if files get matched to tracks and clear_existing_tags is active.

# Problem

* JIRA ticket (_optional_): PICARD-1972

# Solution
Use the full track metadata object to copy all data and don't treat it as a dict.

This just copies the metadata tags as a dict:

```
m = Metadata(old_metadata)
```

This does a full copy:

```
m = Metadata()
m.update(old_metadata)
```

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -175,9 +175,8 @@ class Metadata(MutableMapping):
         self.images = ImageList()
         self.has_common_images = True
 
-        d = dict(*args, **kwargs)
-        for k, v in d.items():
-            self[k] = v
+        if args or kwargs:
+            self.update(*args, **kwargs)
         if images is not None:
             for image in images:
                 self.images.append(image)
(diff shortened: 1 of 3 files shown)


zas (changes requested):
Actually this is very counter-intuitive. Instead we should fix constructor to act as `update()` if possible and keep current syntax. What do you think?

phw:
> Instead we should fix constructor to act as `update()` if possible and keep current syntax. What do you think?

Agreed, I'll look into this. But we need to review current usage of this. But probably most of the times it is currently used for dict like objects other then metadata, so would be fine.

phw:
@zas: Now `Metadata(other_metadata)`  behaves as expected. This is really better, avoids bugs like this.

zas (approved):

PR #1660, opened 2020-10-16
PICARD-1974: Fix Tags From Filenames handling same tag multiple times

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Fix tags from filenames crashing if same tag is used more than once in match expression.#

Also allow parsing hidden tags.

# Problem

* JIRA ticket (_optional_): PICARD-1974, PICARD-1975

# Solution

Refactor tags from filenames to have a `TagMatchExpression` class to handle parsing the expression and matching files. This allows testing.

Fix the issue with multiple groups of same name by separating tag names from group names and ensure group names are unique. Also fix setting hidden tags with an underscore expression like `%_foo%` (PICARD-1975).

### DIFF
--- a/picard/ui/tagsfromfilenames.py
+++ b/picard/ui/tagsfromfilenames.py
@@ -26,6 +26,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
+from collections import OrderedDict
 import os.path
 import re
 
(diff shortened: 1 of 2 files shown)


picard/ui/tagsfromfilenames.py
```diff
@@ -26,19 +26,72 @@
...
 
 
+class TagMatchExpression:
+    def __init__(self, expression, replace_underscores=False):
+        self.replace_underscores = replace_underscores
+        self._tag_re = re.compile(r"(%\w+%)")
+        self.numeric_tags = ('tracknumber', 'totaltracks', 'discnumber', 'totaldiscs')
+        self.parse(expression)
+
+    def parse(self, expression):
```
zas, line 51: Since `parse()` is only used in `__init__()` I would get rid of it, or rename it `_parse()` (as it isn't meant to be called directly).

picard/ui/tagsfromfilenames.py
```diff
@@ -26,19 +26,72 @@
...
 from picard.ui import PicardDialog
 from picard.ui.ui_tagsfromfilenames import Ui_TagsFromFileNamesDialog
 from picard.ui.util import StandardButton
 
 
+class TagMatchExpression:
+    def __init__(self, expression, replace_underscores=False):
+        self.replace_underscores = replace_underscores
+        self._tag_re = re.compile(r"(%\w+%)")
+        self.numeric_tags = ('tracknumber', 'totaltracks', 'discnumber', 'totaldiscs')
```
zas, line 48: Perhaps rename to class `_numeric_tags` as it isn't meant to be modified (as class attribute, one could subclass and modify it).

picard/ui/tagsfromfilenames.py
```diff
@@ -26,19 +26,72 @@
...
+        self.parse(expression)
+
+    def parse(self, expression):
+        self._group_map = group_map = OrderedDict()
```
zas, line 52: Since `group_map` is used only once in following loop, I think we can just use `self._group_map`

phw:
This now handles the same tag used multiple times by generating multiple matches.

E.g. if you match a filename such as `The Artist - Great Song feat. Somebody` against 

    %performer% - %title% feat. %performer%

you will get the performer tag with multi-value `['The Artist', 'Somebody']`.

picard/ui/tagsfromfilenames.py
```diff
@@ -26,19 +26,72 @@
...
+        format_re = ['(?:^|/)']
+        for i, part in enumerate(self._tag_re.split(expression)):
+            if part.startswith('%') and part.endswith('%'):
+                name = part[1:-1]
+                group = '%s_%i' % (name, i)
+                tag = normalize_tagname(name)
+                group_map[group] = tag
+                if tag in self.numeric_tags:
+                    format_re.append('(?P<' + group + r'>\d+)')
```
zas, line 61: For the sake of consistency, use `r` prefix everywhere?

```python
format_re.append(r'(?P<' + group + r'>\d+)')
```

picard/ui/tagsfromfilenames.py
```diff
@@ -26,19 +26,72 @@
...
+            else:
+                format_re.append(re.escape(part))
+        # Optional extension
+        format_re.append(r'(?:\.\w+)?$')
+        self._format_re = re.compile("".join(format_re))
+
+    @property
+    def matched_tags(self):
+        # Return unique values, but preserve order
+        return list(OrderedDict.fromkeys(self._group_map.values()).keys())
```
zas, line 75: `.keys()` shouldn't be needed (as converting dict to list always use keys).

zas (changes requested):
Very good fix for an issue that was here since ages. Just few stylistic comments.

phw:
@zas Good points, I incorporated your suggestions and rebased the branch.

zas (approved):
LGTM

PR #1665, opened 2020-10-21
 PICARD-214: Fix Picard adding numbers on file rename on case insensitive file systems

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
If a file got renamed and the only change was in casing Picard would append a number on case insensitive file systems (except on Windows).

On Windows this worked because `os.path.normcase`, which was used by `picard.util.pathcmp`, would do the proper thing and normalize the casing. But on other OS it would leave the paths as is, even if the underlying filesystem is case insensitive.

* JIRA ticket (_optional_): PICARD-214

# Solution
Use `os.path.samefile` to properly check if the old and new path refer to the same file. This is fully supported even on Windows since Python 3.2.

Note that `os.path.samefile` requires a stat on both paths in order to compare inodes. But I don't see a easy way around that.

# Action
Needs more testing.

So far I tested this on:

- Windows: As expected behavior is unchanged. Case-only changes do not result in numbers being added to files and the casing is changed.
- On Linux with ext4 and case insensitive directory: In this setup I could reproduce the original issue. The change to `os.path.samefile` fixes it (1) 

But this needs additional testing on:

- macOS with case insensitive file system (both AFPS and HFS)
- Mounted FAT32 or NTFS drives on platforms other than Windows
- Network file systems. Not sure whether `os.path.samefile` can handle those.

(1) One thing to note with ext4 case-insensitive is that the file's casing will not be changed on saving. But actually this seems to be an issue with the ext4 case-insensitive implementation in general. In my tests I could also not rename such a file. I think POSIX does not even allow for such a thing.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -154,10 +154,6 @@ def decode_filename(filename):
         return filename.decode(_io_encoding)
 
 
-def pathcmp(a, b):
-    return os.path.normcase(a) == os.path.normcase(b)
-
-
 def format_time(ms, display_zero=False):
     """Formats time in milliseconds to a string representation."""
     ms = float(ms)
@@ -372,16 +368,6 @@ def tracknum_from_filename(base_filename):
     if numbers:
         return numbers[0]
     return None
-
-
-# Provide os.path.samefile equivalent which is missing in Python under Windows
-if IS_WIN:
-    def os_path_samefile(p1, p2):
-        ap1 = os.path.abspath(p1)
-        ap2 = os.path.abspath(p2)
-        return ap1 == ap2
-else:
-    os_path_samefile = os.path.samefile
 
 
 def is_hidden(filepath):
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -154,6 +154,21 @@ def decode_filename(filename):
         return filename.decode(_io_encoding)
 
 
+def samepath(path1, path2):
+    return os.path.normcase(os.path.normpath(path1)) == os.path.normcase(os.path.normpath(path2))
+
+
+def samefile(path1, path2):
+    """Returns True, if both `path1` and `path2` refer to the same file.
+
+    Behaves similar to os.path.samefile, but first checks identical paths including
+    case insensitive comparison on Windows using os.path.normcase. This fixes issues on
+    some network drives (e.g. VirtualBox mounts) where two paths different only in case
+    are considered separate files by os.path.samefile.
+    """
+    return samepath(path1, path2) or os.path.samefile(path1, path2)
+
+
 def format_time(ms, display_zero=False):
     """Formats time in milliseconds to a string representation."""
     ms = float(ms)
--- a/picard/file.py
+++ b/picard/file.py
@@ -75,6 +75,7 @@ from picard.util import (
     emptydir,
     find_best_match,
     format_time,
+    samefile,
     thread,
     tracknum_from_filename,
 )
@@ -485,7 +486,7 @@ class File(QtCore.QObject, Item):
         tmp_filename = new_filename
         i = 1
         while (os.path.exists(new_filename + ext)
-               and not os.path.samefile(old_filename, new_filename + ext)):
+               and not samefile(old_filename, new_filename + ext)):
             new_filename = "%s (%d)" % (tmp_filename, i)
             i += 1
         new_filename = new_filename + ext
--- a/picard/file.py
+++ b/picard/file.py
@@ -75,7 +75,6 @@ from picard.util import (
     emptydir,
     find_best_match,
     format_time,
-    pathcmp,
     thread,
     tracknum_from_filename,
 )
@@ -485,8 +484,8 @@ class File(QtCore.QObject, Item):
             os.makedirs(new_dirname)
         tmp_filename = new_filename
         i = 1
-        while (not pathcmp(old_filename, new_filename + ext)
-               and os.path.exists(new_filename + ext)):
+        while (os.path.exists(new_filename + ext)
+               and not os.path.samefile(old_filename, new_filename + ext)):
             new_filename = "%s (%d)" % (tmp_filename, i)
             i += 1
         new_filename = new_filename + ext


zas (approved):
This solution looks clean to me, and since `os.path.samefile()` documentation specifies the same implementation is used on all systems (since Python 3.4) it should work without much surprise.
The cost of 2 `stat()` shouldn't have much impact anyway.
Lgtm.

picard/util/__init__.py
```diff
@@ -154,6 +154,19 @@ def decode_filename(filename):
...
+def samefile(path1, path2):
+    """Returns True, if both `path1` and `path2` refer to the same file.
+
+    Behaves similar to os.path.samefile, but first checks identical paths including
+    case insensitive comparison on Windows using os.path.normcase. This fixes issues on
+    some network drives (e.g. VirtualBox mounts) where two paths different only in case
+    are considered separate files by os.path.samefile.
+    """
+    return (
+        os.path.normcase(os.path.normpath(path1)) == os.path.normcase(os.path.normpath(path2))
```
zas, line 166: Perhaps just define a small function to shorten lines:

```python
def samepath(path1, path2):
    return os.path.normcase(os.path.normpath(path1)) == os.path.normcase(os.path.normpath(path2))

def samefile(path1, path2):
....
    return samepath(path1, path2) or os.path.samefile(path1, path2))

```

zas (changes requested):

phw:
I tested the following cases of renaming files on case insensitive file systems with changes only in casing (`Test.mp3` -> `test.mp3`):

| Case | With patch | Current master |
|------|-------------|-----------------|
| Windows 10, local NTFS fs | Ok | Ok |
| Windows, FAT32 | Ok (1) | Ok (1) |
| Windows 10, SMB share on Windows | Ok | Ok |
| Windows 10, VirtualBox shared folder | Ok | Ok |
| Linux, ext4 case insensitive | Ok (2) | Appends numbers |
| macOS, FAT32 | Ok (2) | Appends numbers |
| Linux, SMB share on Windows | Appends numbers (3) | Appends numbers |
| macOS, AFPS case insensitive | Ok | Appends numbers |
| macOS, FAT32 | Ok | Appends numbers |
| macOS, SMB share on Windows | Ok | Appends numbers |

---
1) Works without numbers getting appended (paths are detected to be the same file) but renames changing only the casing are not applied. But this is also not possible with Explorer.
2) Works without numbers getting appended (paths are detected to be the same file) but renames changing only the casing are not applied. This seems to be a general issue on Linux when dealing with case insensitive files. I also could not change the casing with Nautilus or on console with mv (the latter will actually fail as both paths refer to the same file).
3) On Linux accessing a SMB share differences in file name casing are considered separate files, even though they access the same file. `os.stat` reports different inodes for e.g. `Test.mp3`, `TEST.mp3` and `test.mp3` even if they all refer to the same file.

zas (approved):

Sophist-UK:
> 1. Works without numbers getting appended (paths are detected to be the same file) but renames changing only the casing are not applied. But this is also not possible with Explorer.

This is also my experience with explorer - to change the case alone, you actually have to change something else (I typically append a letter) and then remove the addition in a second rename.

I know that this has now been merged, but I do wonder whether it is better for Picard unexpectedly (from the user's perspective) to fail to set the case correctly or whether to make the results conform to user expectations Picard should have special-case code to detect cases where a simple change in case will not be applied and to follow the above double-rename approach to ensure that it happens,

phw:
> Picard should have special-case code to detect cases where a simple change in case will not be applied and to follow the above double-rename approach to ensure that it happens,

Maybe, but this is a separate issue. Also this is an issue only applying in certain cases, see the table above. From the tested cases this only affects Windows with FAT32 and Linux when dealing with case insensitive file systems.

elomatreb:
>     2\. I also could not change the casing with Nautilus or on console with mv (the latter will actually fail as both paths refer to the same file).

One (hacky) way to change the name in situations like these is using two `mv`s, by changing it to an intermediary name that isn't equivalent according to the case folding rules.

phw:
Actually the situation on Windows is more complex, and a bit confusing. I have two Windows 10 installs to test with, both the same Windows 10 version. On the one I tested the cases above except the Fat32 case the renaming of case only changes works. On the other Windows 10 install case-only changes are not applied, independent of file system.

I found the root cause, and will submit a fix, but I don't understand yet why the difference in behavior. The problem is that the new filename is filtered through `os.path.realpath`. On the system where renaming does not work this returns the actual casing of the file name as it is on disk. Let's make an example:

We have a file `TEST.mp3` which gets loaded as `TEST.mp3`. The renaming script would assign it the name `test.mp3` But `File.make_filename` runs this name through `os.path.realpath` (see [file.py#L470](https://github.com/metabrainz/picard/blob/master/picard/file.py#L470)). On the system where the renaming does not work and the file exists on disk it returns again `TEST.mp3`. Result is Picard considers the filename unchanged and does not apply the naming.

On the other system however `os.path.realpath` returns the name as given, that means `test.mp3`. Thus the renaming works.

> One (hacky) way to change the name in situations like these is using two `mv`s, by changing it to an intermediary name that isn't equivalent according to the case folding rules.

Yes, that would be the idea. In practice it would require 3 renames, first the attempt to rename, then checking if casing was unchanged (which indicates the file system is affected by the limitation), then the double rename. The tricky part probably is to detect

phw:
Found the cause for the differen behavior. I was running from source, on the one system with Python 3.7 (as we currently use it for releases), the other one with Python 3.9. Actually this breaks for us with Python 3.8, as there was a change in [`os.path.realpath`](https://docs.python.org/3/library/os.path.html#os.path.realpath):

> Changed in version 3.8: Symbolic links and junctions are now resolved on Windows.

I added PICARD-1990 to track this.

phw:
So I moved my test table to https://gist.github.com/phw/d30ffb94171c5ea3248691f951310d54 and added tickets for the individual remaining issues.

rdswift:
> > One (hacky) way to change the name in situations like these is using two `mv`s, by changing it to an intermediary name that isn't equivalent according to the case folding rules.
> 
> Yes, that would be the idea. In practice it would require 3 renames, first the attempt to rename, then checking if casing was unchanged (which indicates the file system is affected by the limitation), then the double rename. The tricky part probably is to detect

I'm not necessarily promoting the "hacky solution", but it could work if you always did the intermediate name first and then rename to the final name.  No need to check if the casing was unchanged from the (prior code) renaming.

It's academic now because it looks like your latest changes have resolved this.  :+1:

phw:
> ... but it could work if you always did the intermediate name first and then rename to the final name.

As this is about edge cases always doing this sounds a bit too much to me, I'd like to avoid this. But this means there must be a simple way to check the casing stored on disk.

The alternative would be to know that we are on a case insensitive file system, but this is also not too easy. To get this fully right you need to check per directory, and even checking this seems to be problematic. Due to lack of APIs it probably would involve probing this by testing with an actual file. 

> It's academic now because it looks like your latest changes have resolved this. +1

No, the problem persists for some specific file systems, see my table. On Windows it affects FAT32 file systems (e.g. on an external USB or mounted audio player).  It is also present on Linux when dealing with case insensitive file systems. The problem is that a simple rename "TEST" -> "test" does not do any changes for those setups. This is not specific to Picard or Python, e.g. both on Windows and Linux standard tools also cannot do this without the intermediate step.

The question is whether it is worth the trouble, as it affects only some cases. The big cases work, that is Windows on file systems other than FAT32 and since 2.5 now also macOS. On Linux dealing with case insensitive file systems is not too common. The one big case on Linux is probably SMB shares, but this is in this regard otherwise broken.

elomatreb:
> As this is about edge cases always doing this sounds a bit too much to me, I'd like to avoid this.

One compromise would be to offer a case-insensitivity compatibility option that enables this intermediary step always in the settings, for people who know they have slightly strange configurations that are not addressed by the other fixes.

Sophist-UK:
My own vote would be to add special case code to ensure that case is as expected, but if others feel differently I would understand.

elomatreb:
The problem with doing it always is that a move can be a pretty time-intensive operation when working on slow drives or filesystems, it could lead to significant slowdowns for people in cases where a single move would suffice.

phw:
Isn't a simple rename on the same fs rather cheap? I would still prefer to do it only if necessary, as in the majority of cases it is not an issue (both because it only affects a few setup and because only a subset of all file savings will be about case changes only). But for those cases it should not be a big impact.

The tricky part is actually to get the current casing of the file. For Windows we can actually win API calls for this, I have a crude patch ready for this. But for Linux I have not found a proper way to do it. The hammer method is to list the files in the directory and search for a case insensitive match, then check whether it is the same file. But especially if there are many files this is definitely more expensive then a double rename.

Maybe it is indeed better to check first if a rename would only change the casing. If it is, then do a double rename in all cases without checking further (we could exclude macOS from this behavior, as it seems by the issue). I think this is what Sophist is basically suggesting.

Sophist-UK:
> The problem with doing it always is that a move can be a pretty time-intensive operation when working on slow drives or filesystems, it could lead to significant slowdowns for people in cases where a single move would suffice.

Moves are only time-consuming between drives where the data needs to be moved. "Moves" within the same file system (on both Windows & Linux) are typically done only by adjusting the directory entries - so e.g. two additional moves within the same directory to correct situations where the case is incorrect should have only a tiny overhead. 

(If Picard were a transactional app, then two renames in series being non-atomic might have been an issue i.e. Picard crashing after the first rename and before the second. However Picard is not transactional in this sense, and such a problem would be corrected by a subsequent save.)

elomatreb:
Even a simple rename can take arbitrary amounts of time, for example when interacting with network drives over spotty wireless connections, over connections with high latency, or even just with protocols that inherently tend to be slow (e.g. WebDAV).

Sophist-UK:
> Even a simple rename can take arbitrary amounts of time, for example when interacting with network drives over spotty wireless connections, over connections with high latency, or even just with protocols that inherently tend to be slow (e.g. WebDAV).

@elomatreb I guess that is a reason NOT to do two renames for EVERY file to ensure that the case is correct every time, but instead to handle case issues with two further renames on an exception basis. The difficulty with exceptions is that you can only do it if the APIs allow you to detect it (see @phw 's comment above).

phw:
@Sophist-UK agreed. I think if we do this only in the situation where a rename changes only the case we already limit the impact, case only changes are not the norm. Of course it can for specific cases also mean it is done for every file. We can do some exceptions:

 - For macOS we don't need any workaround at all
- On Windows we could do a check via GetLongPathNameW after the rename. If case changing was not successful follow up by a double rename. That way we add a mandatory filename lookup for all case only changes, but do additional moves only when needed. It's a compromise: Prevent the double moves for cases where it works (NTFS, Network shares), but do a triple move where it doesn't (FAT32).

What bothers me a bit is the Linux situation. Basically we would force a double move for all case changing renames, even though the majority of file systems there is case sensitive and not affected. But I don't see a better solution right now.

Sophist-UK:
> What bothers me a bit is the Linux situation. Basically we would force a double move for all case changing renames, even though the majority of file systems there is case sensitive and not affected. But I don't see a better solution right now.

I assume that you are referring to `Linux, SMB share on Windows` rather than `Linux, ext4 case insensitive` or `Linux, FAT32`.

The discussion below only pertains to situations of case changes on `Linux, SMB share on Windows`...

I am not sure whether on Linux there is Python functionality that allows you to access the inode, and if there is whether there is additional Python functionality allowing you to get the actual filename from an inode. But if these both exist, then presumably we can compare and do a double rename to fix in necessary.

If you have the neither of these functions, I am not sure that there is any technical solution.

But if you have the first functionality (to get the inode) but not the second then ...

We have an existing file and we can get the inode for it (inode1). After the rename, we can also request inodes against both the old and new names (inodes2/3 respectively). So the question is whether having the original inode and the two new ones allows us to make any determination of whether the case change has worked.

So the

phw:
> I assume that you are referring to Linux, SMB share on Windows rather than Linux, ext4 case insensitive or Linux, FAT32.

Actually the other was around. The SAMBA case is even a step further away from getting solved. There we currently can't even detect that two case-different paths actually refer to the same file. I don't know yet how to solve this. A os.stat will report different inodes, even if it is the same file in the end. If someone has an idea how to fix this that would be great.

But for the other Linux cases I now know how to handle it. Basically we need to do this:

```
if only_filename_case_differs(oldpath, newpath) and same file(oldpath.lower(), oldpath.upper()):
  # case-only change on case insensitive file system
  double_rename(oldpath, newpath)
else:
  # A normal rename
  rename(oldpath, newpath)
```

Thanks a lot for the brainstorming. Patch is basically ready, just need to write it down ;)

Sophist-UK:
> Thanks a lot for the brainstorming. Patch is basically ready, just need to write it down ;)

Hurrah!! (This sort of synergy is what Open Source is all about.)

phw:
@Sophist-UK @elomatreb https://github.com/metabrainz/picard/pull/1674

PR #1667, opened 2020-10-24
PICARD-1990: Fix file renaming not performed on case-only changes

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fix renaming files on Windows if only casing is changed and Python >= 3.8 is in use.

# Problem

* JIRA ticket (_optional_): PICARD-1990

This affects only Windows with Python >= 3.8, as `os.path.realpath` will
return the path as it is on disk.

# Solution
Do not make `File.make_filename` return `os.path.realpath`. This should be fine, as it already use `os.path.realpath` on the directory. Still use `os.path.realpath` on the result of `File._rename` to ensure we use a canonical name for the reloaded file.

See https://github.com/metabrainz/picard/pull/1665#issuecomment-715899992 for in-depth explanation.

Also https://gist.github.com/phw/d30ffb94171c5ea3248691f951310d54 for an overview of case-insensitive behavior and issues.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -461,17 +461,18 @@ class File(QtCore.QObject, Item):
                 new_dirname = os.path.normpath(os.path.join(os.path.dirname(filename), new_dirname))
         else:
             new_dirname = os.path.dirname(filename)
+        try:
+            new_dirname = os.path.realpath(new_dirname)
+        except FileNotFoundError:
+            # os.path.realpath can fail if cwd does not exist and path is relative
+            pass
         new_filename = os.path.basename(filename)
 
         if settings["rename_files"] or settings["move_files"]:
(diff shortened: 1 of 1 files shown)


phw:
Needs some more work. It works on testing, but the tests failing indicate it should behave a bit different.

phw:
Also needs to check we do not regress on on whatever https://github.com/metabrainz/picard/commit/0cb228b876d38db25d4522a77b08ef60d2edf157 tried to fix and PICARD-1306

phw:
The updated version fixes the issue while `File.make_filename` still behaves the same as before.

zas (approved):

PR #1668, opened 2020-10-25
PICARD-1987: Fix ~multiartist, ~pregap and ~datatrack variables not set for files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1987

After the refactoring for running scripts on file metadata the three special hidden variables `~multiartist`, `~pregap` and `~datatrack`, as those did get set after running `File._finalize_loading_track` and are hence not set in the `File.orig_metadata`.

# Solution

Set the variables `~pregap` and `~datatrack` as part of `File._finalize_loading_track`. `~multiartist` is still set afterwards, but ensure it is also set to `File.orig_metadata`.

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -335,21 +335,25 @@ class Album(DataObject, Item):
                 if "pregap" in medium_node:
                     discpregap = True
                     absolutetracknumber += 1
-                    track = self._finalize_loading_track(medium_node['pregap'], mm, artists, va, absolutetracknumber, discpregap)
-                    track.metadata['~pregap'] = "1"
+                    extra = {
+                        '~pregap': '1'
+                    }
+                    self._finalize_loading_track(medium_node['pregap'], mm, artists, va, absolutetracknumber, discpregap, extra)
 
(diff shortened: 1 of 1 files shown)


picard/album.py
```diff
@@ -413,7 +427,7 @@ def _finalize_loading(self, error):
             if self.item.isSelected():
                 self.tagger.window.refresh_metadatabox()
 
-    def _finalize_loading_track(self, track_node, metadata, artists, va, absolutetracknumber, discpregap):
+    def _finalize_loading_track(self, track_node, metadata, artists, extrametadata=None):
```
zas, line 430: Just a minor thing, but `extra_metadata` is more readable than `extrametadata`

zas (changes requested):

picard/album.py
phw: Right, I renamed it

zas (approved):

PR #1671, opened 2020-10-28
PICARD-1997: Reduce performance impact of fingerprint column

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1997, PICARD-1989

The fingerprint column has a huge impact on rendering performance. The column content gets repainted a lot and can reduce loading performance significantly.

# Solution
Use the default icon implementation of `QTreeWidget` instead of custom implementation.

In my tests this improved performance a lot. If the fingerprint column is visible this reduced loading time for 4k files by 80%. Even if the column is hidden loading time is still reduced by 32%. See the following results for loading 4k files (with ignore MBIDs):

|                        |  master | this patch | reduction |
|--------------------|------------|---------------|------------|
| column visible | 78s       | 16s           |  79%     |
| column hidden | 22s       | 15s           |  32%     |

As an alternative I also tried setting the `sizeHint` on the custom `FingerprintColumnWidget` and even caching the painted column, but this had no big impact. In the end I could reduce the loading time of the 4k files to 72s this way, which is not really a gain compared to the 16s with this patch.

This is visually not the same. Instead of being neatly centered the icon is now left aligned, and I can't get the column smaller. This is how it is in current releases:

grafik

This is the patched version:

grafik

The first one is clearly nicer and the reason why I did the implementation as it is on master. But this is for sure not worth the huge performance hit.

In https://github.com/metabrainz/picard/commit/ab8b8effe5abc5cac31a9d0c556af1d05d4dc2e3 I also removed the size hint for the title column, which fixed PICARD-1989. In my tests this had no really measurable impact on loading time, maybe it shoved of 0.5 to 1.0 seconds. But this is definitely balanced out by the fingerprint column gains.

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -323,29 +323,10 @@ def paint_fingerprint_icon(painter, rect, icon):
     if not icon:
         return
     size = COLUMN_ICON_SIZE
-    padding_h = (rect.width() - size) / 2
+    padding_h = COLUMN_ICON_BORDER
     padding_v = (rect.height() - size) / 2
     target_rect = QtCore.QRect(rect.x() + padding_h, rect.y() + padding_v, size, size)
     painter.drawPixmap(target_rect, icon.pixmap(size, size))
-
-
-class FingerprintColumnWidget(QtWidgets.QWidget):
-    def __init__(self, file, parent=None):
-        super().__init__(parent=parent)
-        self._file = file
-
-    def paintEvent(self, event=None):
-        painter = QtGui.QPainter(self)
-        paint_fingerprint_icon(painter, self.rect(), self.decide_icon())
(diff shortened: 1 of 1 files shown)


zas (approved):
This is a trade between performance and fancier look, but regarding numbers it is definitively a good move.
LGTM.

phw:
> This is a trade between performance and fancier look, but regarding numbers it is definitively a good move.

Yes, indeed. I spent a lot of time on the fancier look and really wanted it to be like this, but with this penalty it is just not worth sticking to it. I guess some users will report the new look as an issue.

PR #1673, opened 2020-10-29
PICARD-1828: Allow setting cover art for multiple selected files

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1828

# Solution
Add a new class `FileList`, which acts kind of like a temporary ad-hoc cluster. In fact I have made it the parent of the `Cluster` class, as it shares some basic code.

The cover art box only can handle a single object. When multiple files are selected create a `FileList` object and use this for the cover art box.

This PR also optimized the event handling and avoids unnecessary signalling. Previously the coverart box caused many selection changes. For "use original coverart" it even relied on a roundtrip to update cover art on items -> items signal updates -> main window updates selection -> cover art box updates view. This is not needed, the cover art box can update right away. Indeed with > 100 files selected this made updating nearly impossible. With the changed signalling I had no issue updating the cover art of all 4k loaded files.

# Action
I think the approach to let cover art box only handle a single item and have an item that can handle multi-selection is the right approach. We probably should extend this to the metadatabox, I can imagine this would simplify the handling there. Currently much of the complexity comes from the need to update the selection and gather data on many files. If we instead pass a new `FileList` item to metadatabox we could avoid much of the threading complexity involved in updating the file list.

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -70,7 +70,46 @@ from picard.util.progresscheckpoints import ProgressCheckpoints
 from picard.ui.item import Item
 
 
-class Cluster(QtCore.QObject, Item):
+class FileList(QtCore.QObject, Item):
+
+    def __init__(self, files=None):
+        QtCore.QObject.__init__(self)
+        self.metadata = Metadata()
+        self.orig_metadata = Metadata()
+        self.files = files or []
+        self.update_metadata_images_enabled = True
+        for file in self.files:
+            if self.can_show_coverart:
+                file.metadata_images_changed.connect(self.update_metadata_images)
+        update_metadata_images(self)
+
+    def iterfiles(self, save=False):
+        for file in self.files:
+            yield file
+
+    def update(self):
+        pass
+
+    @property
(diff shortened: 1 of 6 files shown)


picard/cluster.py
```diff
@@ -70,7 +70,46 @@
...
+        self.metadata = Metadata()
+        self.orig_metadata = Metadata()
+        self.files = files or []
+        self.update_metadata_images_enabled = True
+        for file in self.files:
+            if self.can_show_coverart:
+                file.metadata_images_changed.connect(self.update_metadata_images)
+        update_metadata_images(self)
+
+    def iterfiles(self, save=False):
```
zas, line 86: `save` parameter isn't used, is there a reason to keep it?

phw: Yes, this is for API compatibility with `Album` and `ClusterList`.

zas (approved):

PR #1674, opened 2020-10-31
PICARD-1992: Fix case only changes when renaming files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1991, PICARD-1992

If a rename results only in the casing of file names changing this is not applied by a simple `shutil.move` or `os.rename` in some situations on case insensitive file systems. Specifically this never works for Linux when using case insensitive file systems (e.g. FAT32, ext4 with case-insensitivity enabled, CIFS share) and on Windows 10 for FAT32 and extFAT file systems.

See also the comparison on https://gist.github.com/phw/d30ffb94171c5ea3248691f951310d54

# Solution
IF a case-only change happens (file stays in same directory, file name changes only in casing) on a case insensitive file system (case sensitive file systems are not affected by design) and the OS is affected by this (macOS handles this correctly already on OS level)

THEN apply a workaround by renaming the file twice, first to a temporary name then to the final name.

The workaround for Windows and Linux is basically the same. For Windows I tried to optimize this, because it already works there for NTFS and nework shares. To avoid double renames especially on network shares the renaming is first attempted normally. Then the actual case of the file is checked and only if the renaming was unsuccessful (i.e. on FAT32 or extFAT) another double rename is performed.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -80,8 +80,8 @@ from picard.util import (
     tracknum_from_filename,
 )
 from picard.util.filenaming import (
-    fix_filename_casing,
     make_short_filename,
+    move_ensure_casing,
 )
 from picard.util.preservedtags import PreservedTags
 from picard.util.scripttofilename import script_to_filename_with_metadata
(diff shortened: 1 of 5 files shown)


zas (approved):
LGTM

PR #1675, opened 2020-11-01
PICARD-2001: Ensure dropped files are moved to a valid target

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fix dropped files not showing up in UI if getting dropped on other files.

# Problem

* JIRA ticket (_optional_): PICARD-2001

It is possible to drag and drop files from the file browser onto invalid drop targets. If those drop targets do not allow adding files the files will be added to the tagger but won't show up in the UI.

# Solution
Always move added files to a valid target in the UI. If no drop target is given or the drop target does not support adding files always drop to unclustered files.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -462,18 +462,20 @@ class Tagger(QtWidgets.QApplication):
 
     def move_file(self, file, target):
         if target is None:
-            self.unclustered_files.add_file(file)
-        elif isinstance(target, Album):
+            if not file.parent:
+                target = self.unclustered_files
+            else:
+                log.debug("Aborting move since target is invalid and file already has a parent")
+        if isinstance(target, Album):
             self.move_files_to_album([file], album=target)
-        elif isinstance(target, ClusterList):
-            if isinstance(file.parent, Track):
-                file.parent.remove_file(file)
-            self.cluster([file])
+        elif isinstance(target, File) and target.parent:
+            file.move(target.parent)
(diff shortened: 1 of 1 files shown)


phw:
Fixed another way to get the files lost: Dropping directly on "Custers" would have also moved the file to nowhere. I should have called this PR "PICARD-2001: A file odyssey".

picard/tagger.py
```diff
@@ -453,29 +453,26 @@ def _file_loaded(self, file, target=None, remove_file=False):
...
+        self.move_file(file, target)
 
         # fallback on analyze if nothing else worked
         if config.setting['analyze_new_files'] and file.can_analyze():
             log.debug("Trying to analyze %r ...", file)
             self.analyze([file])
 
     def move_file(self, file, target):
-        if target is None:
-            log.debug("Aborting move since target is invalid")
```
zas, line 468: What if target is None and file.parent? Shouldn't we (debug) log this case or is it handled somewhere?

phw: In current code it actually should never happen that file has a parent as the `move_file` method is only used for new files. But yes, I intentionally made this function so it could be used more generally without breaking. Target None and file having a parent would mean the file stays where it is. Could make sense to log this for debugging, I'll add it.

zas (approved):
LGTM

PR #1677, opened 2020-11-04
PICARD-2005: Fix tags being shown as unchanged if multiple tracks are selected

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2005

IF multiple tracks have been selected where some have changes in existing tags but the first track being looked at has no changes because incoming metadata is empty,  
THEN the tag is being shown as unchanged.

If you take the example from PICARD-2005 and look at the code at https://github.com/metabrainz/picard/blob/master/picard/ui/metadatabox.py#L538 the following happens:

If the first file being considered is the one of track 2, the tag "lyrics_language" will have:

orig values `['Japanese']`  
new values `[]`

It then gets new values set to orig values, and the tag is added to `existing_tags`. This results in no change (both ori values and new values are the same).

The next track being looked at is track 1, then it will have

orig values `['Japanese']`  
new values `[jpn]`

This tag is clearly changed. But because the tag name is in `existing_tags` the if clause also triggers and it also gets new values set to orig values. Now there is no difference anymore and this counts as unchanged.

# Solution

Get rid of the `existing_tags` set.

@zas I suppose this was added for a reason, but looking at that I cannot imagine any use case this tried to solve. The condition seems clearly wrong. But we might miss something and introduce a different regression. Any ideas what this was for?

# Action

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -519,9 +519,6 @@ class MetadataBox(QtWidgets.QTableWidget):
         tag_diff = TagDiff(max_length_diff=config.setting["ignore_track_duration_difference_under"])
         orig_tags = tag_diff.orig
         new_tags = tag_diff.new
-        # existing_tags are orig_tags that would not be overwritten by
-        # any new_tags, assuming clear_existing_tags is disabled.
-        existing_tags = set()
         tag_diff.objects = len(files)
 
         clear_existing_tags = config.setting["clear_existing_tags"]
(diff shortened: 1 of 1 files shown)


zas:
> @zas I suppose this was added for a reason, but looking at that I cannot imagine any use case this tried to solve. The condition seems clearly wrong. But we might miss something and introduce a different regression. Any ideas what this was for?

This test was added in 5a2d849324b68276d70d9902c7148c43b6e9647f 8 years ago, so hard to tell.

zas (approved):
Not sure about possible regression, but the fix looks better.

phw:
Actually the existing_tags check was added in 

> https://github.com/metabrainz/picard/commit/eb321969cf86348dcb53133ea35b5015ee86659e

But probably this was just carried over in later refactorings. I'll take a closer look at this and do some additional testing tomorrow. But if I don't find anything misbehaving let's go indeed forward and merge this.

PR #1679, opened 2020-11-10
PICARD-2006: Optimize moving files to albums on load

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2006, PICARD-2012

When files gets loaded that are already matched to a release, the files are not added to the release until after the album is fully loaded. This was changed in https://github.com/metabrainz/picard/commit/fac2b37d8cf8dad8e821de541b9c40ee2ae0a0b8 .

This causes various issues:

- The files are not available to plugins or cover art providers that run during the loading of the release. See PICARD-2006
- In case of a MBID redirect, where the redirect target is already loaded, Picard attempts to move unmatched files to the already loaded release. This broke with the above commit, causing the files to being still loaded without being visible in the UI. PICARD-2012
- The files are not visible in the UI until after loading has finished.

# Solution
Basically revert https://github.com/metabrainz/picard/commit/fac2b37d8cf8dad8e821de541b9c40ee2ae0a0b8 and add files to albums right away. This ensures files of an album are available to plugins and cover art
providers during loading.

But refactor `Album.match_files` to optimize this a bit. If the album is already loaded `Album.match_files`  will match directly as before. If the album is still loading it will add the files to `unmatched_files` instead. After loading has finished album already calls `self.match_files(self.unmatched_files.files)`, so this works.

This optimizes two cases:
- If a file gets moved to an already loaded release the unnecessary move to unmatched_files before matching is skipped
- If a file gets moved to a release during loading, `Album.match_files` is only called once after loading has finished. In the old implementation it would be called once in `Album._finalize_loading`  and then again once per pre-matched file as a after_load_callback.

During testing I did not observe any performance issues. https://github.com/metabrainz/picard/commit/fac2b37d8cf8dad8e821de541b9c40ee2ae0a0b8  was originally done as a performance improvement, but probably other optimizations have a bigger impact. My test of loading 4k files took the same amount of time (both just loading files and overall loading times) with this branch and master.

# Action
`Album.run_when_loaded`  is with this change currently unused in Picard. But I kept it, I think it is a useful API for plugins.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -526,7 +526,7 @@ class Album(DataObject, Item):
         file.metadata_images_changed.disconnect(self.update_metadata_images)
         remove_metadata_images(self, [file])
 
-    def _match_files(self, files, recordingid=None, threshold=0):
+    def _match_files(self, files, threshold=0):
         """Match files to tracks on this album, based on metadata similarity or recordingid."""
         tracks_cache = defaultdict(lambda: None)
 
@@ -547,7 +547,7 @@ class Album(DataObject, Item):
             if file.state == File.REMOVED:
                 continue
             # if we have a recordingid to match against, use that in priority
-            recid = recordingid or file.metadata['musicbrainz_recordingid']
+            recid = file.match_recordingid or file.metadata['musicbrainz_recordingid']
             if recid and mbid_validate(recid):
                 if not tracks_cache:
                     build_tracks_cache()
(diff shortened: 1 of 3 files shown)


phw:
Just for completeness: I had also experimented with keeping the current behavior of not moving files to `Album.unmatched_files`  but instead keeping them in a separate list of files to process after loading. This again was an attempt to avoid the UI updates. But I dropped this approach. It made things again more complex (you need to consider these files at various places where unmatched_files are used), they must be returned as part of `Album.iterfiles`. This ads to probable bugs, does not solve the issue that the user cannot see the files during loading and in the end does not seem to give any real performance gain.

zas (approved):
LGTM

PR #1680, opened 2020-11-14
PICARD-1948: Limit fractional scaling support to Windows

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1948

This was originally introduced in PICARD-1777 to improve scaling support on Windows.
It causes UI regressions on Linux were it sometimes results in wrong scaled
fonts if no scaling factor is configured.

grafik

# Solution

Limit this to Windows for now.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -944,8 +944,9 @@ def main(localedir=None, autoupdate=True):
     QtWidgets.QApplication.setAttribute(QtCore.Qt.AA_UseHighDpiPixmaps)
     QtWidgets.QApplication.setAttribute(QtCore.Qt.AA_EnableHighDpiScaling)
     # HighDpiScaleFactorRoundingPolicy is available since Qt 5.14. This is
-    # required to support fractional scaling properly.
-    if hasattr(QtGui.QGuiApplication, 'setHighDpiScaleFactorRoundingPolicy'):
+    # required to support fractional scaling on Windows properly.
+    # It causes issues without scaling on Linux, see https://tickets.metabrainz.org/browse/PICARD-1948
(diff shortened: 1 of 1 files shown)


zas:
I really wonder why we have this at all.
Basically it forces "Don't round" policy when the default is "Round".

_"The default value is Qt::HighDpiScaleFactorRoundingPolicy::Round. On Qt for Android the default is Qt::HighDpiScaleFactorRoundingPolicy::PassThrough, which preserves historical behavior from earlier Qt versions."_

Doing so forces Picard to ignore value set through env variable `QT_SCALE_FACTOR_ROUNDING_POLICY` (right?)

https://doc.qt.io/qt-5/qt.html#HighDpiScaleFactorRoundingPolicy-enum
https://doc.qt.io/qt-5/qguiapplication.html#setHighDpiScaleFactorRoundingPolicy

phw:
> Basically it forces "Don't round" policy when the default is "Round".

That's the entire point. If you round and you set UI scaling to 150% (factor 1.5) the actual scaling will be by factor 2.0. Basically it will only scale by whole numbers.

Setting the rounding policy to pass through will actually scale by 1.5. This is also known as fractional scaling.

zas (approved):

phw:
Interestingly fractional scaling is with or without setting this option working well on my system with GNOME. I guess the Qt documentation does not tell the full story here, things seem to be more platform specific.

Overall scaling as far as I know works pretty well on macOS, on Linux at least with GNOME it's also working for me. But on Windows it is really problematic. Hence I think limiting this change to Windows is fine.

PR #1684, opened 2020-11-21
Generate appdata for non-Linux platforms

BSDKaffee:
# Summary
Generate appdata for non-Linux platforms.

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem
Setup attempts to install the appdata for all platforms not darwin, haiku1, or win32, yet it is only generated for linux. This was introduced in [edfa273](https://github.com/metabrainz/picard/commit/edfa2733535b328ab7deeff6c7cf619642fd1319) and causes installation errors on platforms such as FreeBSD since the appdata was not previously generated. 

# Solution
Generate the appdata for the platforms that setup tries to install it on.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -292,7 +292,7 @@ class picard_build(build):
                 'publisher': os.environ.get('PICARD_APPX_PUBLISHER', default_publisher),
                 'version': '.'.join([str(v) for v in store_version]),
             })
-        elif sys.platform == 'linux':
+        elif sys.platform not in ['darwin', 'haiku1', 'win32']:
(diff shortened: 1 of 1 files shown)


phw (approved):
Thanks a lot!

Actually this wasn't introduced just recently, but has been this way since appdata is being built, see https://github.com/metabrainz/picard/commit/0de126a3cd6c9f7594d53c7bfd7c6b447df053bb :D

phw:
I added PICARD-2023 to track this for the changelog

PR #1685, opened 2020-11-24
PICARD-2020: Handle multi-value musicbrainz_albumid/musicbrainz_recordingid on load

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2020

If already tagged files are loaded into Picard, but the existing album or recording IDs are stored as multi-value (can be the same ID stored multiple times), then Picard will fail to detect the IDs and hence not load the album or NAT.

# Solution
if musicbrainz_albumid or musicbrainz_recordingid contain multiple values, use the first one to load the proper album / recording for already tagged files.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -428,10 +428,12 @@ class Tagger(QtWidgets.QApplication):
             return
 
         if not config.setting["ignore_file_mbids"]:
-            recordingid = file.metadata['musicbrainz_recordingid']
+            recordingid = file.metadata.getall('musicbrainz_recordingid')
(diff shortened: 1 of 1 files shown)


zas (approved):

phw:
@zas Side topic, but do you have any idea what is up with Codacy? The checks for the branches are triggered, they also run. Github integration is also still setup.

PR #1687, opened 2020-11-25
PICARD-2022: Fix crash when loading network share without access rights on Windows

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fix crash when trying to load an inaccessible network share on Windows.

# Problem

* JIRA ticket (_optional_): PICARD-2022

os.path.realpath can raise an exception if the path is not accessible. In the concrete case I had a folder shared publicly from my Linux machine via samba, but had not granted public permissions (on Linux file system). Windows showed the shared folder, but could not access it. Trying to load the folder into Picard lead to the crash

# Solution
Handle os.path.realpath with more care, introduce a picard.util.normpath for the common combination of os.path.normpath(os.path.realpath(...))

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -112,6 +112,7 @@ from picard.util import (
     encode_filename,
     is_hidden,
     mbid_validate,
+    normpath,
     process_events_iter,
     thread,
     uniqify,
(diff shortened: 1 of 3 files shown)


zas (approved):
Lgtm

PR #1689, opened 2020-11-28
PICARD-2028: Ignore selection changes when clustering or removing files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2028

Removing or clustering files calls MainWindow.update_selection a lot. This can cause a significant slow down if many files are selected.

# Solution
Call MainWindow.update_selection only once after the deletion / clustering is completed.

Also refactored MainWindow.ignore_selection_changes to use a context manager, so one can do:

```python
with window.ignore_selection_changes:
   # perform some action
   # window.update_selection will automatically be called when the context manager exits
```

# Action

The main cause why the many calls to update_selection caused such a significant slowdown in PICARD-2028 is actually the cover art for multiple selected files feature. While this performs well for the case that the user selects a huge amount of files it is still significantly slow that repated update_selection calls will greatly harm the performance.

I am in parallel working on optimizations for this part only, as there might be other places in code where this has a negative effect. But the fixes here already have the effect that the performance of removing all selected files is basically equivalent to how it was in Picard 2.5.1

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -743,6 +743,7 @@ class Tagger(QtWidgets.QApplication):
     def remove(self, objects):
         """Remove the specified objects."""
         files = []
+        self.window.ignore_selection_changes = True
         for obj in objects:
             if isinstance(obj, File):
                 files.append(obj)
@@ -766,6 +767,8 @@ class Tagger(QtWidgets.QApplication):
                 self.remove_cluster(obj)
         if files:
             self.remove_files(files)
+        self.window.ignore_selection_changes = False
+        self.window.update_selection()
 
     def _lookup_disc(self, disc, result=None, error=None):
         self.restore_cursor()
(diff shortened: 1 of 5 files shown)


Sophist-UK:
@phw I do like the way you have implemented this.

phw:
@zas As discussed I added tests for `IgnoreSelectionContext`. Also this class of course could be used more generically and I thought about moving it to a more general place. But actually if in doubt and something like this is not actually reused I prefer keeping it local to the place where it is used until an actual use case comes up for it.

phw:
No idea why the CI is failing :( Any ideas? pytest runs just fine here locally

zas:
> No idea why the CI is failing :( Any ideas? pytest runs just fine here locally

No idea either :-(

phw:
Yes, it seems to be related to pyobjc 7. But other branches build well with this. A rebase did not work, so I now just wen ahead and merged it. If it still fails on master I'll downgrade pyobjc to < 7

PR #1690, opened 2020-11-29
PICARD-2035: Improve the tooltip on Album errors

Sophist-UK:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Improves the tooltip when a processing error (like a CAA retrieval failure) causes the Album icon to indicate an error.

# Problem

I was getting albums with an error icon. I tracked this down to a failure to retrieve CAA covers.
 
* JIRA ticket (_optional_): PICARD-2033, PICARD-2035

# Solution

Provide a better tooltip.

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -976,7 +976,7 @@ class AlbumItem(TreeItem):
                     item.update(update_album=False)
         if album.errors:
             self.setIcon(MainPanel.TITLE_COLUMN, AlbumItem.icon_error)
-            self.setToolTip(MainPanel.TITLE_COLUMN, _("Error"))
+            self.setToolTip(MainPanel.TITLE_COLUMN, _("Processing error(s): See the Errors tab in the Album Info popup"))
         elif album.is_complete():
             if album.is_modified():
                 self.setIcon(MainPanel.TITLE_COLUMN, AlbumItem.icon_cd_saved_modified)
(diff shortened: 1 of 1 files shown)


phw (dismissed):
Yes, actually we talked about something similar just recently on IRC.

Actually we should also review this for individual files and non-album tracks and unify the behavior. Currently error handling is slightly different there.

I added PICARD-2035 for this instead of reusing PICARD-2033

picard/ui/itemviews.py
```diff
@@ -976,7 +976,7 @@ def update(self, update_tracks=True):
                     item.update(update_album=False)
         if album.errors:
             self.setIcon(MainPanel.TITLE_COLUMN, AlbumItem.icon_error)
-            self.setToolTip(MainPanel.TITLE_COLUMN, _("Error"))
+            self.setToolTip(MainPanel.TITLE_COLUMN, _("Processing error(s): See the Errors tab in the Album Info popup"))
```
mineo, line 979: If we want to stick with what the online docs use for all dialogs:

```suggestion
            self.setToolTip(MainPanel.TITLE_COLUMN, _("Processing error(s): See the Errors tab in the Album Info dialog"))
```

mineo (approved):

mineo (approved):

phw (approved):

PR #1691, opened 2020-11-29
PICARD-2037: Make Info/Errors tab more readable

Sophist-UK:
# Summary

On the Errors tab in the Info popup, text is dark red on a dark grey background which is not very readable. Switching to bright red.

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

On the Errors tab in the Info popup, text is dark red on a dark grey background which is not very readable. 

* JIRA ticket (_optional_): PICARD-2037

# Solution

Switching to log-error colour defined in Options/UI/Colours.

### DIFF
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -357,7 +357,7 @@ class AlbumInfoDialog(InfoDialog):
         tab_index = tabWidget.indexOf(tab)
         if album.errors:
             tabWidget.setTabText(tab_index, _("&Errors"))
-            text = '<br />'.join(map(lambda s: '<font color="darkred">%s</font>' %
+            text = '<br />'.join(map(lambda s: '<font color="red">%s</font>' %
                                      '<br />'.join(htmlescape(str(s))
                                                    .replace('\t', ' ')
                                                    .replace(' ', '&nbsp;')
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -364,7 +364,7 @@ class AlbumInfoDialog(InfoDialog):
                                                    .replace('\t', ' ')
                                                    .replace(' ', '&nbsp;')
                                                    .splitlines()
-                                                   )), 
+                                                   )),
                                      album.errors)
                                  )
             self.ui.info.setText(text + '<hr />')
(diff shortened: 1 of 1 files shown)


picard/ui/infodialog.py
```diff
@@ -357,7 +357,7 @@ def _display_info_tab(self):
         tab_index = tabWidget.indexOf(tab)
         if album.errors:
             tabWidget.setTabText(tab_index, _("&Errors"))
-            text = '<br />'.join(map(lambda s: '<font color="darkred">%s</font>' %
```
phw, line 360: Does this look good on  light theme? We probably need to distinguish here. Same is actually true for the log window, maybe this could be fixed similar.

Dark layout can be checked with:

```
from picard.ui.theme import theme

theme.is_dark_theme
```
That's also what we use for switching syntax highlighting colors.

phw (commented):
Makes sense, we should just check that it works with light gray background also and if needed use different colors, see my comment.

Also we need a separate ticket for this as well.

picard/ui/infodialog.py
zas: What about making it configurable? Using https://github.com/metabrainz/picard/blob/master/picard/ui/colors.py

Sophist-UK: Actually, I was just  thinking that. (I replaced a hard coded colour with a different hard coded colour.)

Leave this with me.

P.S. We could also do with an option to select system, light or dark theme.

Sophist-UK: I am not going to fix any theme issues in this PR - just change the text colour.

Sophist-UK:
Well - code style error just says it is indentation, but I have checked and I can't work out what the indentation should be. HELP!!!

phw (approved):

phw:
Yes, that looks like a fine solution. Regarding the coding: I merged this now locally and refactored the code a bit to make it more readable and easier to format:

https://github.com/metabrainz/picard/commit/46d15f5e15b6efe056360c7f28605f38a93daa76

flake8 should now be happy

Sophist-UK:
Thanks.

PR #1692, opened 2020-11-29
PICARD-2034: Add copy and paste context menu entries in metadata view

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2034

The metadata view supports copy and pasting values using the usualy Ctrl+C, Ctrl+V shortcuts, but no indication that it is supported.

# Solution

Add copy and paste actions to context menu of the metadata box. Also use the system specific shortcut instead of hard coded Ctrl+C/Ctrl+V.

# Action


picard/ui/metadatabox.py
```diff
@@ -282,20 +283,32 @@ def edit(self, index, trigger, event):
...
+                self.clipboard = list(self.tag_diff.orig[tag])
+            elif column == 2:
+                self.clipboard = list(self.tag_diff.new[tag])
+
+    def paste_value(self):
+        item = self.currentItem()
+        if item:
+            column = item.column()
+            tag = self.tag_diff.tag_names[item.row()]
+            if column == 2 and tag != "~length":
```
zas, line 309: I think we should replace hardcoded column numbers by constants, and this tag match for length should prolly replaced by a function call (`can_be_pasted(tag)` or whatever), but that's unrelated to this PR.

zas (approved):

picard/ui/metadatabox.py
phw: Actually I would like to see tags marked as readonly and checked for that. Then we could unify the checks here a bit.

PR #1698, opened 2020-12-01
PICARD-2028: Optimize and speedup coverart updates

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2028

Cover art updates are rather expensive. This can be even more problematic then before since support for setting cover art for multiple selected files in #1673.

This PR aims to optimize this by unifying the cover art handling across different object types and avoiding updates if they are not necessary.

# Solution

- Unify handling of cover art between the different object types `Album`, `Track`, `Cluster` and `File`. The API of those objects is now similar enough to remove special handling at many places
- Avoid calling updates to cover art box unless selection changed. The cover art box instead relies on signalling for the rare case where cover art changes for a loaded object
- Avoid signalling in cover art box when cover art is changed. This was already partly done in #1673, but is now more complete. Basically the cover art box knows it updated the album, it does not need signalling back from changed items to update its view
- In general avoid emitting `metadata_images_changed` unless necessary
- Fix how keep_existing_tags was handled. In current Picard setting keep_existing_tags on all files of an album does not properly update the track and hence also not the album

# Action
This is quite a list of changes. But it's all interconnected optimizations. If wanted I could split this up. The individual commits are already kind of separate changesets, with a working system after each commit. But they build upon one another.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -403,6 +403,7 @@ class Album(DataObject, Item):
                     file.move(self.unmatched_files)
             self.metadata = self._new_metadata
             self.orig_metadata.copy(self.metadata)
+            self.orig_metadata.images.clear()
             self.tracks = self._new_tracks
             del self._new_metadata
             del self._new_tracks
--- a/picard/album.py
+++ b/picard/album.py
@@ -718,9 +718,9 @@ class Album(DataObject, Item):
         if not self.update_metadata_images_enabled:
             return
 
-        update_metadata_images(self)
-        self.update(False)
-        self.metadata_images_changed.emit()
+        if update_metadata_images(self):
+            self.update(False)
+            self.metadata_images_changed.emit()
 
     def keep_original_images(self):
         self.enable_update_metadata_images(False)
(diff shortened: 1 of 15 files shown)


zas (approved):
That's a log of changes at once, but I didn't find any issue after reviewing and testing.
Overall it's a significative improvement.

phw:
Yes, this is a surprisingly complex thing to test. A lot of fine details how it behaves. From my testing this should now behave as in Picard 2.5.1 or earlier, even a bit better. It fixes some oddity with album art display on album level in Picard 2.5.2, where original artwork would always contain both the data from files as well as the new image from cover art providers. It also fixes an older oddity that "use original images" applied to all tracks would not reflect on the album itself.

Otherwise I tested with tracks, individual files, clusters. Adding / removing files from tracks.

I'll go ahead and merge it, but keep testing this before next release.

PR #1701, opened 2020-12-03
Add a tagger.iter_files_from_objects method

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

`tagger.get_files_from_objects` is used in many places, usually for iterating over the selected files. `tagger.get_files_from_objects` however first creates a list of objects, chains their `iterfiles()` methods and again converts the result into a list.

# Solution

Add a `picard.util.iter_files_from_objects` function which yields all files for the selected objects, allowing efficient iteration over all selected files. This also reduces dependency on the global tagger object.

Keep `get_files_from_objects` returning a list for cases where a list is actually wanted and for backward compatibility.

# Action

### DIFF
--- a/picard/ui/scriptsmenu.py
+++ b/picard/ui/scriptsmenu.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2018 Laurent Monin
-# Copyright (C) 2018 Philipp Wolfer
+# Copyright (C) 2018, 2020 Philipp Wolfer
 # Copyright (C) 2018 Yvan Rivière
 #
 # This program is free software; you can redistribute it and/or
@@ -36,7 +36,7 @@ from picard.script import (
     ScriptParser,
 )
 from picard.track import Track
-from picard.util import uniqify
+from picard.util import iter_unique
 
 
 class ScriptsMenu(QtWidgets.QMenu):
@@ -53,7 +53,7 @@ class ScriptsMenu(QtWidgets.QMenu):
         s_text = script[3]
         parser = ScriptParser()
 
-        for obj in self._get_unique_metadata_objects():
+        for obj in self._iter_unique_metadata_objects():
             try:
                 parser.eval(s_text, obj.metadata)
                 obj.update()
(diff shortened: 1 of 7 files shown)


picard/ui/scriptsmenu.py
```diff
@@ -66,20 +66,19 @@ def _run_script(self, script):
...
+        return iter_unique(self._iter_metadata_objects(self.tagger.window.selected_objects))
 
-    def _get_metadata_objects(self, objs):
+    def _iter_metadata_objects(self, objs):
         for obj in objs:
             if hasattr(obj, 'metadata'):
                 yield obj
             if isinstance(obj, Cluster):
-                yield from self._get_metadata_objects(obj.files)
+                yield from self._iter_metadata_objects(obj.iterfiles())
```
zas, line 77: Wouldn't it be more correct to have `elif` instead of `if` here?

zas (dismissed):
LGTM

picard/ui/scriptsmenu.py
phw: I changed it to elif for the checks below. Note that the `hasattr(obj, 'metadata'):` is separate in addition to the rest (yield the object itself, yield it's children).

picard/util/__init__.py
```diff
@@ -369,6 +375,13 @@ def uniqify(seq):
     return [x for x in seq if x not in seen and not add_seen(x)]
 
 
+def iter_unique(seq):
+    """Creates an iterator only returning unique values from seq"""
+    seen = set()
+    add_seen = seen.add
```
zas, line 381: I don't get why `add_seen` is needed here.

zas (changes requested):

picard/util/__init__.py
phw: Me neither :D https://github.com/metabrainz/picard/pull/1701/commits/77bf10c0609a95ce838c9ec1795661b1ba1628d3

zas: It was added in https://github.com/metabrainz/picard/commit/28b1a0a05b557aaff271771960da51bc09e53921
Not sure why ;)

zas (approved):
LGTM

PR #1706, opened 2020-12-11
PICARD-2046: improve AcoustID submission process, better handling of maximum payload

zas:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

In certain cases, AcoustID submission could fail with Payload Too Large error.
Batch size was limited to a number of items, but do not take in account actual size of items.

* JIRA ticket (_optional_): PICARD-2046

# Solution

Rework AcoustID fingerprints submission to handle payload limits

- approximate payload of each submission
  Calculating exact payload would add extra complexity, so we do an approximation with safety margin
- limit batch to a maximum payload
- handle AcoustID server HTTP 413 (Payload too large), retrying with a smaller batch
- limit the number of attempts for each submission to 5
- improve logging

# Action

### DIFF
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -50,36 +50,36 @@ class AcoustIDManager(QtCore.QObject):
 
     def __init__(self, acoustid_api):
         super().__init__()
-        self._fingerprints = {}
+        self._submissions = {}
         self._acoustid_api = acoustid_api
 
     def add(self, file, recordingid):
         if not file.acoustid_fingerprint or not file.acoustid_length:
             return
         puid = file.metadata['musicip_puid']
-        self._fingerprints[file] = Submission(file.acoustid_fingerprint, file.acoustid_length, recordingid, recordingid, puid)
+        self._submissions[file] = Submission(file.acoustid_fingerprint, file.acoustid_length, recordingid, recordingid, puid)
         self._check_unsubmitted()
 
     def update(self, file, recordingid):
-        submission = self._fingerprints.get(file)
+        submission = self._submissions.get(file)
         if submission is None:
             return
         submission.recordingid = recordingid
         self._check_unsubmitted()
 
     def remove(self, file):
-        if file in self._fingerprints:
-            del self._fingerprints[file]
+        if file in self._submissions:
+            del self._submissions[file]
         self._check_unsubmitted()
 
     def is_submitted(self, file):
-        submission = self._fingerprints.get(file)
+        submission = self._submissions.get(file)
         if submission:
             return not submission.recordingid or submission.orig_recordingid == submission.recordingid
         return True
 
     def _unsubmitted(self):
-        for file, submission in self._fingerprints.items():
+        for file, submission in self._submissions.items():
             if submission.recordingid and submission.orig_recordingid != submission.recordingid:
                 yield (file, submission)
 
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -40,6 +40,9 @@ class Submission(object):
         self.orig_recordingid = orig_recordingid
         self.recordingid = recordingid
 
+    def is_submitted(self):
+        return not self.recordingid or self.orig_recordingid == self.recordingid
+
 
 class AcoustIDManager(QtCore.QObject):
 
@@ -75,12 +78,12 @@ class AcoustIDManager(QtCore.QObject):
     def is_submitted(self, file):
         submission = self._submissions.get(file)
         if submission:
-            return not submission.recordingid or submission.orig_recordingid == submission.recordingid
+            return submission.is_submitted()
         return True
 
     def _unsubmitted(self):
         for file, submission in self._submissions.items():
-            if submission.recordingid and submission.orig_recordingid != submission.recordingid:
+            if not submission.is_submitted():
                 yield (file, submission)
 
     def _check_unsubmitted(self):
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -93,9 +93,11 @@ class AcoustIDManager(QtCore.QObject):
             return submission.is_submitted()
         return True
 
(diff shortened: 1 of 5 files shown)


picard/acoustid/manager.py
```diff
@@ -39,55 +39,83 @@ def __init__(self, fingerprint, duration, orig_recordingid=None, recordingid=Non
...
         self.recordingid = recordingid
+        self.attempts = 0
+
+    def __len__(self):
+        # payload approximation
+        # it is based on actual measures, as an example:
+        # with no puid, 2 fingerprints total size of 7501 bytes, post body was 7719 bytes (including all fields)
+        # so that's an overhead of ~3%
+        # we use 10% here, to be safe
+        return int((len(self.fingerprint) + len(self.puid)) * 1.1)
```
phw, line 50: Should we include `len(self.recording_id) + len(str(self.duration))` here? We have the data available, so helps us to be more accurate.

picard/acoustid/manager.py
```diff
@@ -39,55 +39,83 @@ def __init__(self, fingerprint, duration, orig_recordingid=None, recordingid=Non
...
+    def _batch(self, submissions):
+        size = 0
+        for file, submission in submissions:
+            if submission.attempts > self.MAX_ATTEMPTS:
+                continue
+            size += len(submission)
+            if size > self.max_batch_size:
+                log.debug("AcoustID: Max batch size attained: %d > %d", size, self.max_batch_size)
+                break
+            submission.attempts += 1
```
phw, line 110: Nice detail here is that retry is per fingerprint submission. But submission.attempts should get reset I think if one hits the submit action again.

picard/acoustid/manager.py
```diff
@@ -106,37 +134,54 @@ def _batch_submit(self, submissions, errors=None):
...
-            }
-            previous_errors.append(mparms)
-            log_msg = N_("AcoustID submission failed with error '%(error)s': %(message)s")
-            log.error(log_msg, mparms)
-            self.tagger.window.set_statusbar_message(
-                log_msg, mparms, echo=None, timeout=3000)
+            response_code = self._acoustid_api.webservice.http_response_code(http)
+            if response_code == 413:
+                self.max_batch_size = int(self.max_batch_size * self.BATCH_SIZE_REDUCTION_FACTOR)
+                log.warn("AcoustID: payload too large, batch size reduced to %d", self.max_batch_size)
```
phw, line 169: I guess here goes a call to something like 

```
        self._acoustid_api.submit_acoustid_fingerprints(
            [submission for file_, submission in submissions],
            partial(self._batch_submit_finished, submissions, previous_errors, next_func)
        )
```

picard/acoustid/manager.py
zas: With 10% overhead it isn't needed, but yes if we want to increase precision (not sure if really needed at this point).

picard/acoustid/manager.py
zas: Not needed, see last iteration

picard/acoustid/manager.py
```diff
@@ -39,62 +39,92 @@ def __init__(self, fingerprint, duration, orig_recordingid=None, recordingid=Non
...
+        remaining = []
+        batch_size = 0
+        for file, submission in submissions:
+            if submission.attempts < self.MAX_ATTEMPTS:
+                batch_size += len(submission)
+                if batch_size < self.max_batch_size:
+                    submission.attempts += 1
+                    batch.append((file, submission))
+                    continue
+            remaining.append((file, submission))
```
phw, line 125: It is probably not necessary to iterate over all submissions every time. After batch size hits max it is only appending the remaining, so we don't really need to check the submission attempts and length again. I would instead of the continue do a break and then add all the not yet processed items in submissions to the remaining list (this probably needs keeping count of the already processed ones).

What I don't quite get is how we should handle the submissions which reached their submission attempt limit. As I understand it they stay in the manager forever, getting iterated over and over again. Maybe we should reset the submission attempts after one submission cycle finished?

phw (commented):
This is great. Just a small comment for discussion on the failed attempts, that part is not yet fully clear to me how this should work best.

picard/acoustid/manager.py
zas: > What I don't quite get is how we should handle the submissions which reached their submission attempt limit. As I understand it they stay in the manager forever, getting iterated over and over again. Maybe we should reset the submission attempts after one submission cycle finished?

Well,if a set of submissions reaches max attempts, those weren't submitted, keeping it in the manager let the user attempts a (manual) retry, in case cause of failure was temporary.
Really not sure what else we can do.

zas: > It is probably not necessary to iterate over all submissions every time. After batch size hits max it is only appending the remaining, so we don't really need to check the submission attempts and length again. I would instead of the continue do a break and then add all the not yet processed items in submissions to the remaining list (this probably needs keeping count of the already processed ones).

What do you think about 8cff4372559a51af70e7682cf478b5dddacec310 to address this?

phw: Looks good to me.

But I still don't get the submission.attempts handling :)

Later manual retry is fine for me. Causes for reached max. attempts might be interent connectivity or server issues, and retrying later can help.

But as I see it the attempt count for a submission never gets reset. So even on a later retry to submit the fingerprints the submission count would be ready at .sc attempts. Or do I miss something?

zas: > But as I see it the attempt count for a submission never gets reset.

Ah, you're right.
What about resetting them in https://github.com/metabrainz/picard/blob/8cff4372559a51af70e7682cf478b5dddacec310/picard/acoustid/manager.py#L105-L112

phw (approved):
Great improvement, LGTM

PR #1708, opened 2020-12-11
PICARD-2058: Allow selection of files with upper case extension in add file dialog on Linux

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
On case-sensitive file systems the Add File dialog does not show files with matching extension, if it is uppercase (e.g. test.MP3).

# Problem

* JIRA ticket (_optional_): PICARD-2058

# Solution
Add both lower and upper case extensions to the filters.

Do this only on OS other than Win or macOS. macOS and Windows usually support case sensitive file names. Furthermore on both systems the file dialog filters list all extensions we provide, which becomes a bit long when we give the full list twice.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -66,7 +66,10 @@ from picard.cluster import (
     FileList,
 )
 from picard.const import PROGRAM_UPDATE_LEVELS
-from picard.const.sys import IS_MACOS
+from picard.const.sys import (
+    IS_MACOS,
+    IS_WIN,
+)
 from picard.file import File
 from picard.formats import supported_formats
 from picard.plugin import ExtensionPoint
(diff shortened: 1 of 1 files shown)


zas (approved):
This approach is very limited, since there will still be an issue with mixed-case filenames (`.Mp3`, `.fLaC`).
But I guess we don't have much choices if we stick to `QtWidgets.QFileDialog.getOpenFileNames()`.

phw:
> if we stick to `QtWidgets.QFileDialog.getOpenFileNames()`

It's not really specific to `getOpenFileNames`, it is `QFileDialog` in general. And in the end a result of sticking to the functionality common across OS.

zas:
> > if we stick to `QtWidgets.QFileDialog.getOpenFileNames()`
> 
> It's not really specific to `getOpenFileNames`, it is `QFileDialog` in general. And in the end a result of sticking to the functionality common across OS.

Yes, but with `QFileDialog` we have another option: https://doc.qt.io/qt-5/qfiledialog.html#setMimeTypeFilters

phw:
> Yes, but with `QFileDialog` we have another option: https://doc.qt.io/qt-5/qfiledialog.html#setMimeTypeFilters

It's not really another option unfortunately. It is basically just an alternative way to specify file extensions. If you specify a  mime type Qt sets the file extensions for your (lower case only). See also https://stackoverflow.com/questions/34858220/qt-how-to-set-a-case-insensitive-filter-on-qfiledialog

I originally wanted to use the mime types instead, but given the implementation there is not really a point to it. Chances are good it does not know about some of the less common file types we support and that we have the better extension to type mapping already.

PR #1720, opened 2020-12-22
PICARD-1926: Use for per-thread config instances

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1926, PICARD-1689, PICARD-2013, ~PICARD-1590~

QSettings is not thread safe, but instead each thread must use its own instance of QSettings. Otherwise this leads to deadlocks when reading settings shortly after writing settings in different threads.

# Solution
Prepare per-thread config instances by adding config.get_config() and using it to fix deadlocks experienced in the metadatabox.

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -125,11 +125,16 @@ class Config(QtCore.QSettings):
     Use `get_config()` to obtain a Config instance for the current thread.
     Changes to one Config instances are automatically available to all other instances.
 
+    Use `Config.from_app` or `Config.from_file` to obtain a new `Config` instance.
+
     See: https://doc.qt.io/qt-5/qsettings.html#accessing-settings-from-multiple-threads-or-processes-simultaneously
     """
 
     def __init__(self):
-        self.__known_keys = set()
+        # Do not call `QSettings.__init__` here. The proper overloaded `QSettings.__init__`
+        # gets called in `from_app` or `from_config`. Only those class methods must be used
+        # to create a new instance of `Config`.
+        pass
 
     def __initialize(self):
         """Common initializer method for :meth:`from_app` and
@@ -144,12 +149,6 @@ class Config(QtCore.QSettings):
         TextOption("application", "version", '0.0.0dev0')
         self._version = Version.from_string(self.application["version"])
         self._upgrade_hooks = dict()
-        # Save known config names for faster access and to prevent
-        # strange cases of Qt locking up when accessing QSettings.allKeys()
-        # or QSettings.contains() shortly after having written settings
-        # inside threads.
-        # See https://tickets.metabrainz.org/browse/PICARD-1590
-        self.__known_keys = set(self.allKeys())
 
     @classmethod
     def from_app(cls, parent):
@@ -186,19 +185,6 @@ class Config(QtCore.QSettings):
                                   parent)
         this.__initialize()
         return this
-
-    def setValue(self, key, value):
-        super().setValue(key, value)
-        self.__known_keys.add(key)
-
-    def remove(self, key):
-        super().remove(key)
-        self.__known_keys.discard(key)
-
-    def contains(self, key):
-        # Overwritten due to https://tickets.metabrainz.org/browse/PICARD-1590
-        # See comment above in __initialize
-        return key in self.__known_keys or super().contains(key)
 
     def switchProfile(self, profilename):
         """Sets the current profile."""
--- a/picard/config.py
+++ b/picard/config.py
@@ -322,7 +322,6 @@ config = None
 setting = None
 persist = None
 
-PURGE_INTERVAL_MILLISECONDS = 60000
 _thread_configs = {}
 _thread_config_lock = threading.RLock()
 
@@ -336,7 +335,7 @@ def setup_config(app, filename=None):
     _thread_configs[threading.get_ident()] = config
     setting = config.setting
     persist = config.persist
-    QtCore.QTimer.singleShot(PURGE_INTERVAL_MILLISECONDS, _purge_config_instances)
+    _init_purge_config_timer()
 
 
 def get_config():
@@ -358,7 +357,18 @@ def get_config():
     return thread_config
 
 
-def _purge_config_instances():
+def _init_purge_config_timer(purge_interval_milliseconds=60000):
+    def run_purge_config_timer():
+        purge_config_instances()
+        start_purge_config_timer()
+
+    def start_purge_config_timer():
+        QtCore.QTimer.singleShot(purge_interval_milliseconds, run_purge_config_timer)
+
+    start_purge_config_timer()
+
+
+def purge_config_instances():
     """Removes cached config instances for no longer active threads."""
     _thread_config_lock.acquire()
     try:
@@ -369,4 +379,3 @@ def _purge_config_instances():
             del _thread_configs[thread_id]
     finally:
         _thread_config_lock.release()
-        QtCore.QTimer.singleShot(PURGE_INTERVAL_MILLISECONDS, _purge_config_instances)
--- a/picard/config.py
+++ b/picard/config.py
@@ -30,6 +30,7 @@
 from operator import itemgetter
 import os
 import shutil
+import threading
 
 from PyQt5 import QtCore
 
@@ -57,7 +58,6 @@ class ConfigSection(LockableObject):
         self.__name = name
         self.__prefix = self.__name + '/'
         self.__prefix_len = len(self.__prefix)
-        self._memoization = {}
 
     def key(self, name):
         return self.__prefix + name
@@ -78,8 +78,6 @@ class ConfigSection(LockableObject):
         try:
             key = self.key(name)
             self.__qt_config.setValue(key, value)
-            if key in self._memoization:
-                self._memoization[key][0] = False
         finally:
             self.unlock()
 
@@ -92,8 +90,6 @@ class ConfigSection(LockableObject):
             if name in self:
                 key = self.key(name)
                 self.__qt_config.remove(key)
-                if key in self._memoization:
-                    del self._memoization[key]
         finally:
             self.unlock()
 
@@ -109,21 +105,14 @@ class ConfigSection(LockableObject):
     def value(self, name, option_type, default=None):
         """Return an option value converted to the given Option type."""
         if name in self:
-            key = self.key(name)
+            self.lock_for_read()
             try:
-                valid, value = self._memoization[key]
-            except KeyError:
-                valid = False
-
-            if not valid:
-                self.lock_for_read()
-                try:
-                    value = self.raw_value(name, qtype=option_type.qtype)
-                    value = option_type.convert(value)
-                    self._memoization[key] = [True, value]
-                except Exception as why:
-                    log.error('Cannot read %s value: %s', self.key(name), why, exc_info=True)
-                    value = default
+                value = self.raw_value(name, qtype=option_type.qtype)
+                value = option_type.convert(value)
+            except Exception as why:
+                log.error('Cannot read %s value: %s', self.key(name), why, exc_info=True)
+                value = default
+            finally:
                 self.unlock()
             return value
         return default
@@ -131,7 +120,13 @@ class ConfigSection(LockableObject):
 
 class Config(QtCore.QSettings):
 
-    """Configuration."""
+    """Configuration.
+    QSettings is not thread safe, each thread must use its own instance of this class.
+    Use `get_config()` to obtain a Config instance for the current thread.
+    Changes to one Config instances are automatically available to all other instances.
+
+    See: https://doc.qt.io/qt-5/qsettings.html#accessing-settings-from-multiple-threads-or-processes-simultaneously
+    """
 
     def __init__(self):
         self.__known_keys = set()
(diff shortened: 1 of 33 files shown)


phw:
@zas This is WIP and tests need fixing, but I would appreciate your early feedback on this. This would need to get rolled out over the entire code base.

The main threads config instance stays available as `config.config` (and `config.settings` or `config.persist`) for backward compatibility, especially plugins. For the most parts we don't run into issues as we only read from config. The issue really only arises if reading happens shortly after writing from different threads.

We tried to patch around this a lot: We had a fix for similar PICARD-1590 which we likely can remove then. We have the locks in `ConfigSection` which try to make this thread  safe, but fail because QSettings does it's work delayed and asynchronous. Probably these locks also can be removed if we make use of `get_config`.

picard/config.py
```diff
@@ -341,12 +336,51 @@ class ListOption(Option):
...
     setting = config.setting
     persist = config.persist
+    QtCore.QTimer.singleShot(PURGE_INTERVAL, purge_config_instances)
+
+
+def get_config():
+    """Returns a config object for the current thread.
+
+    Config objects for threads are
+    """
```
zas, line 360: Overall the patch looks good, but I think we should totally get rid of global `config` (also `setting` and `persist`), because keeping those along `get_config()` is rather weird and prolly prone to errors.
It would be great if `get_config()` could work in any context (threads or not).

phw: get_config works in any context, you totally should use it also in the main thread. The idea to keep the main thread's config still available as config.config is really for backward compatibility. If we remove it we break a lot of plugins.

zas: Ah ok, it makes sense.

phw: And it also allows me to convert the code base step by step while still having a working system. I hope that makes it easier to update some parts of the code and test it. For starters I'll focus on those parts where we actually run threaded.

picard/config.py
```diff
@@ -341,12 +319,51 @@ class ListOption(Option):
...
+
+def purge_config_instances():
+    """Removes cached config instances for no longer active threads."""
+    _thread_config_lock.acquire()
+    try:
+        all_threads = set([thread.ident for thread in threading.enumerate()])
+        for thread_id in list(_thread_configs.keys()):
+            if thread_id not in all_threads:
+                log.debug('Purging config instance for thread %s.', thread_id)
+                del _thread_configs[thread_id]
```
zas, line 366: This can perhaps be rewriiten so it uses only sets:

```python
        all_threads = set([thread.ident for thread in threading.enumerate()])
        threads_config = set(_thread_configs)
        for thread_id in threads_config.difference(all_threads):
            log.debug('Purging config instance for thread %s.', thread_id)
            del _thread_configs[thread_id]
```

picard/config.py
```diff
@@ -341,12 +319,51 @@ class ListOption(Option):
...
+            config_file = config.fileName()
+            log.debug('Instantiating Config for thread %s using %s.', thread_id, config_file)
+            thread_config = Config.from_file(None, config_file)
+            _thread_configs[thread_id] = thread_config
+        finally:
+            _thread_config_lock.release()
+    return thread_config
+
+
+def purge_config_instances():
```
zas, line 358: Since this method is only called from `setup_config()` it should perhaps be renamed `_purge_config_instances()`

picard/config.py
```diff
@@ -341,12 +319,51 @@ class ListOption(Option):
 setting = None
 persist = None
 
+PURGE_INTERVAL = 60000
```
zas, line 322: Since `PURGE_INTERVAL` is a number of milliseconds (according to https://doc.qt.io/qt-5/qtimer.html) it would be cleared to include unit in the name: `PURGE_INTERVAL_MILLISECONDS`.

picard/config.py
```diff
@@ -341,12 +319,51 @@ class ListOption(Option):
...
+def get_config():
+    """Returns a config object for the current thread.
+
+    Config objects for threads are
+    """
+    thread_id = threading.get_ident()
+    thread_config = _thread_configs.get(thread_id)
+    if not thread_config:
+        _thread_config_lock.acquire()
+        try:
```
zas, line 348: Is there a specific reason to use try/finally here? It may hide exceptions.

Is it possible to just use:

```python
with _thread_config_lock:
    do_stuff
```

and if exceptions are expected:

```python
with _thread_config_lock:
    try:
        do_stuff
    except Exception as e:
        log_this
```

phw: The locks from threading do not implement a context manager, so you cannot use `with`. But

```python
_thread_config_lock.acquire()
try:
    do_stuff
finally:
    _thread_config_lock.release()
```

is functional equivalent to

```python
with _thread_config_lock:
    do_stuff
```

It doesn't hide exceptions, as it doesn't catch any. But it ensures the lock is released. It shouldn't matter here. No exceptions are expected here, and should one arise the main thread would exit anyway, so no real need to release the lock. But I think it is good practice to follow this pattern when dealing with locks.

picard/config.py
phw: Not sure, I don't really think that it is easier to read.

phw:
I addressed the comments. Currently the state of this should be fine to fix the deadlock issues we had, as far as I can tell all uses of the config instance inside threads now use a thread specific `Config` instance. Things to do:

1. Tests are still failing sometimes, and I don't yet know why. This needs to be fixed before merging.

2. For consistency we should also convert other uses and getting rid of direct access to `config.config` in Picard's codebase (but as discussed above keep it for compatibility with plugins). We could do the rest of the conversion in a separate PR in order to not blow this one up too much.

3. I will review existing plugins. If a plugin accesses config inside a thread I'll refactor it in a way that keeps backward compatibility. Either move the config access outside of the thread or try to import `get_config` with a fallback that will just do
    ```python
    def get_config():
       return config.config
    ```

phw:
One thing I clearly noticed during my tests is that instantiating a `QSettings` instance is really slow. When originally adapting the tests to the new config handling I actually used a real `QSettings` instance (with a temporary file) per test. This made the tests run 3 times slower. So we really should avoid creating the `Config` object too often. But this patch ensures this by having only one instance per thread and keeping it around at least as long as the thread is active.

One thing is that I measured a slow down of file loading performance. Instead of 4k files getting loaded in ~13.6 seconds it now took ~15.4 seconds. I am not sure yet on the cause. It is not the instantiation of `Config` objects. This only happens 8 times (`QThreadPool` provides 8 threads on my system due to available CPU cores). So it is probably the repeated calls to `get_config`.

picard/config.py
zas: Damn, I shouldn't review stuff without enough coffee in my body...

picard/config.py
```diff
@@ -109,32 +105,27 @@ def raw_value(self, name, qtype=None):
...
-                except Exception as why:
-                    log.error('Cannot read %s value: %s', self.key(name), why, exc_info=True)
-                    value = default
-                self.unlock()
+                value = self.raw_value(name, qtype=option_type.qtype)
+                value = option_type.convert(value)
+            except Exception as why:
+                log.error('Cannot read %s value: %s', self.key(name), why, exc_info=True)
+                value = default
+            self.unlock()
```
zas, line 115: Use `finally` here to ensure `unlock()` is called whatever happens?

picard/ui/metadatabox.py
```diff
@@ -132,21 +136,24 @@ def __tag_ne(self, tag, orig, new):
...
-    def add(self, tag, orig_values, new_values, removable, removed=False, readonly=False):
+    def add(self, tag, orig_values, new_values, removable, removed=False, readonly=False, top_tags=None):
         if orig_values:
             self.orig.add(tag, orig_values)
 
         if new_values:
             self.new.add(tag, new_values)
 
+        if not top_tags:
+            top_tags = []
```
zas, line 147: Instead of a list, since it is only used below as `tag in top_tags`, and never modified, we could use (and pass a set).

picard/ui/metadatabox.py
```diff
@@ -566,12 +576,14 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
...
         }
 
+        config = get_config()
         tag_diff = TagDiff(max_length_diff=config.setting["ignore_track_duration_difference_under"])
         orig_tags = tag_diff.orig
         new_tags = tag_diff.new
         tag_diff.objects = len(files)
 
         clear_existing_tags = config.setting["clear_existing_tags"]
+        top_tags = config.setting['metadatabox_top_tags']
```
zas, line 586: `top_tags` -> set?

zas (commented):
Overall, I like this move

zas:
Weird, tests are segfaulting with 384f2cf

picard/config.py
zas: Yup, not easier to read for sure, but likely faster. You can still rewrite `list(_thread_configs.keys())` to `list(_thread_configs))`.

phw:
> Weird, tests are segfaulting with [384f2cf](https://github.com/metabrainz/picard/commit/384f2cfb0ecc25731c8c2bfded2243d30c41e51f)

The reason is that it calls the parent constructor in __init__. Without this it doesn't crash. Actually before https://github.com/metabrainz/picard/commit/ebd98c9758d074804c2dd83059827f31fbdb13fd#diff-74efa3849396e3135d3796b5965e96c02fcc9dc1036c4b28bba71506193303d6L109 we had only an empty `__init__` method.

I am not sure what consequences it has to omit the QSettings constructor, but it seems to work.

Sophist-UK:
{Heresy alert}

I know this is a bit off the wall (only a bit?) but this seems to be a way complicated solution to a thread issue with Qt/PyQt.

Can we not find a native thread aware solution in Python that is way easier?

phw:
It just uses QSettings as it is intended to be used, I don't see it as overly complicated. Do you have any suggestion of another solution that is simpler and both syncs between threads and processes?

picard/config.py
```diff
@@ -341,12 +322,51 @@ class ListOption(Option):
...
+    _thread_config_lock.acquire()
+    try:
+        all_threads = set([thread.ident for thread in threading.enumerate()])
+        threads_config = set(_thread_configs)
+        for thread_id in threads_config.difference(all_threads):
+            log.debug('Purging config instance for thread %s.', thread_id)
+            del _thread_configs[thread_id]
+    finally:
+        _thread_config_lock.release()
+        QtCore.QTimer.singleShot(PURGE_INTERVAL_MILLISECONDS, _purge_config_instances)
```
zas, line 372: What about something like:

```python
def _purge_configs_timer_start(purge_interval_milliseconds=60000):
    def _purge_config_instances():
        """Removes cached config instances for no longer active threads."""
        _thread_config_lock.acquire()
        try:
            all_threads = set([thread.ident for thread in threading.enumerate()])
            threads_config = set(_thread_configs)
            for thread_id in threads_config.difference(all_threads):
                log.debug('Purging config instance for thread %s.', thread_id)
                del _thread_configs[thread_id]
        finally:
            _thread_config_lock.release()
            _purge_configs_timer_start(purge_interval_milliseconds)

    QtCore.QTimer.singleShot(purge_interval_milliseconds, _purge_config_instances)
```

phw: Instead of defining a new `_purge_config_instances` per call I'd rather put this on module level, at which point `_purge_configs_timer_start` is just a call to `QTimer.singleShot`. Maybe if you want to have the interval encapsulated something like:

```
def _init_purge_config_timer(purge_interval_milliseconds=60000):
    def purge_config_instances():
        """Removes cached config instances for no longer active threads."""
        _thread_config_lock.acquire()
        try:
            all_threads = set([thread.ident for thread in threading.enumerate()])
            threads_config = set(_thread_configs)
            for thread_id in threads_config.difference(all_threads):
                log.debug('Purging config instance for thread %s.', thread_id)
                del _thread_configs[thread_id]
        finally:
            _thread_config_lock.release()
            start_purge_config_timer()

    def start_purge_config_timer():
        QtCore.QTimer.singleShot(purge_interval_milliseconds, purge_config_instances)

    start_purge_config_timer()
```

phw: What do you think about 8053e59 ?

zas (approved):

phw:
Tested this again with the docker-image, where I could reliable reproduce the freezes. They are completely gone with this fix.

RobertKrawitz:
I think this fix is still incomplete.  To reproduce: select all items in an album with multiple tracks along with the album itself, do Add tag.  When I then click Save, it locks up in the same way.

(I wish Picard let you save the current state of processing rather than just save the files; I'm going to have to redo all my work as a result.)

phw:
@RobertKrawitz Are you testing the current master branch?

RobertKrawitz:
Sorry, missed the notification email on this -- I filed a jira ticket a few days ago that was on the latest (at the time) master branch, post-this merge

phw:
@RobertKrawitz All good, we have been discussing there already. phw == outsidecontext :D

RobertKrawitz:
To further update the comment: it appears that adding a tag to a missing track (with a pair of 8th notes next to it) is most likely to trigger a hang right now.

PR #1722, opened 2020-12-26
PICARD-2084: Set the AcoustID port to 443, to use TLS

RomFouq:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Set the AcoustID port to 443, to use TLS.

# Problem

Requests to `api.acoustid.org` are sent in plain, unencrypted HTTP, because the port is set to be `80`, while the server does support the port `443` with TLS.

# Solution

Set the AcoustID port to 443 so that requests are sent using HTTPS.

# Action


phw (approved):
Makes sense. We need a ticket to track this for the changelog, I added PICARD-2084

phw:
Thanks a lot for this!

PR #1723, opened 2020-12-26
PICARD-1950: Provide updated builds for macOS 10.12+ and 10.14+

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [x] Other

# Problem

* JIRA ticket (_optional_): PICARD-1950, maybe PICARD-1943, maybe PICARD-1747

- Builds for macOS are limited to PyQt/Qt 5.13.1 to maintain compatibility with macOS 10.12
- PyQt 5.13.2 and most later versions have a bug causing the app to hang unresponsive on launch on macOS 11. This is fixed in 5.15.2
- Builds were failing with updated Python versions due to signing issues

# Solution

- Upgrade to Python 3.9.1 for all builds. Signing issues are resolved by force resigning all files and upgrading to PyInstaller 4.1.
- Provide separate builds for macOS 10.12+ and 10.14+.
  - The 10.12+ are similar to our current builds and use PyQt/Qt 5.13.1, but with updated Python and PyInstaller. They still run on later versions, including 11.0.
  - The 10.14+ versions runs on 10.14 and later and provides the latest PyQt/Qt 5.15.2

# Notes

- The builds have been tested on macOS 10.12 and 11.0
- The update of PyInstaller to 4.1 also works with Windows. The PICARD-2067 disaster was not caused by this (it was related to Python upgrade to 3.8 on Windows, a separate issue we still need to look into)
- We will need to provide two current mac downloads on the website. But I think we need to maintain macOS 10.12 / 10.13 support but have some value in providing a more up-to-date version for later macOS.
- These changes might fix PICARD-1943, but I can't say as I cannot reproduce the issue and hence cannot verify if it happens with this build
- The 10.14+ build with updated Qt might also fix the tearing issues reported at PICARD-1747. But again I cannot say as I cannot reproduce the reported issue.


phw:
Builds failing is strange, the very same commit build fine on https://github.com/phw/picard/actions/runs/445624056 . Will look into it.

phw:
Found the problem: My github fork actually is set up for complete code signing, whereas pull request builds don't have access to the secrets, hence no code signing. Hence we get into signature issues when running the app for test. I have disabled the execution of the app if code signing is disabled.

zas (approved):
LGTM

PR #1724, opened 2020-12-27
PICARD-2083: Allow empty fpcalc path, auto detect path if empty

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Allow having an empty path for fpcalc configured and auto detect the path in this case.

# Problem

* JIRA ticket (_optional_): PICARD-2083

Currently Picard will use the path to fpcalc as configured in options. In some cases the path can change with each launch. E.g. when packaged with PyInstaller we usually want to use the bundled fpcalc, and the path to it can change (especially for portable Windows build, which gets extracted to a temporary folder with unique name). Same is true for snap installs on Linux, where each new version installed gets a different path under `/snap/picard/`.

As a workaround for the PyInstaller case Picard on launch always resets the fpcalc path to the autodetected value. This has the side effect that the user cannot configure an alternative path in Options > Fingerprinting (it will get reset on every launch).

# Solution

Change the paradigm a bit. The majority of users likely runs with auto detected path and never has to configure this at all. So make this the default:

- The `acoustid_fpcalc` option can now be empty, which is the default. If empty the path gets auto detected **without** explicitly overwriting the `acoustid_fpcalc` with the detected path.
- Options > Fingerprinting will show the auto detected path as a placeholder hint in the text input and also use it for validating the executable. Only if auto detection does not work the user needs to actually select something
- If the user has selected a specific path this will get used. This is now also true on PyInstaller builds. This means on all platforms can the user explicitly configure a different executable
- Add a config upgrade to reset the path on empty for PyInstaller builds and for snap installs, thus triggering to use auto detection in those cases by default

# Action

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -41,7 +41,7 @@ PICARD_APP_NAME = "Picard"
 PICARD_DISPLAY_NAME = "MusicBrainz Picard"
 PICARD_APP_ID = "org.musicbrainz.Picard"
 PICARD_DESKTOP_NAME = PICARD_APP_ID + ".desktop"
-PICARD_VERSION = Version(2, 5, 6, 'dev', 1)
+PICARD_VERSION = Version(2, 5, 6, 'dev', 2)
 
 
 # optional build version
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -34,8 +34,8 @@ from picard import (
     log,
 )
 from picard.acoustid.json_helpers import parse_recording
+from picard.config import get_config
 from picard.const import FPCALC_NAMES
-from picard.const.sys import IS_FROZEN
 from picard.util import find_executable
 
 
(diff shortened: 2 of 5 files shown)


zas (approved):
Nice improvement. LGTM.

rdswift:
@phw Have you started a new picard-docs branch locally to capture these sort of changes in the documentation?  If not, I'll start a new branch `next_version` on the repository and add the update for this change.

rdswift:
I started a new branch called `next_version_2.6` in the `rdswift/picard-docs` repository and updated the description for the fingerprinting configuration section.  Please feel free to make changes as required.  This new branch can also be used to collect any other documentation changes for the next (v2.6) release.

phw:
> I started a new branch called `next_version_2.6` in the `rdswift/picard-docs` repository and updated the description

Thanks a lot for this, I had not yet had the time to consider the docs :)

PR #1725, opened 2020-12-31
PICARD-2087: Fix crash with Qt < 5.10 when adding new tag

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2087

Picard crashes when adding a new tag.

`QListWidget.isPersistentEditorOpen` is only available in Qt 5.10 or later.

# Solution

Check for support of `QListWidget.isPersistentEditorOpen`. This missing doesn't have a huge impact on functionality. We will just start editing again, which makes the current editor close and open again. This gives some flickering when clicking on edit tag with editing already active, but no big deal really.

# Action

### DIFF
--- a/picard/ui/edittagdialog.py
+++ b/picard/ui/edittagdialog.py
@@ -138,7 +138,11 @@ class EditTagDialog(PicardDialog):
 
     def edit_value(self):
         item = self.value_list.currentItem()
-        if item and not self.value_list.isPersistentEditorOpen(item):
+        if item:
+            # Do not initialize editing if editor is already active. Avoids flickering of the edit field
+            # when already in edit mode. `isPersistentEditorOpen` is only supported in Qt 5.10 and later.
+            if hasattr(self.value_list, 'isPersistentEditorOpen') and self.value_list.isPersistentEditorOpen(item):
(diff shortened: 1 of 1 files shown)


picard/ui/edittagdialog.py
```diff
@@ -138,7 +138,11 @@ def tag_selected(self, index):
 
     def edit_value(self):
         item = self.value_list.currentItem()
-        if item and not self.value_list.isPersistentEditorOpen(item):
+        if item:
+            # Do not initialize editing if editor is already active. Avoids flickering of the edit field
+            # when already in edit mode. `isPersistentEditorOpen` is only supported in Qt 5.10 and later.
+            if hasattr(self.value_list, 'isPersistentEditorOpen') and self.value_list.isPersistentEditorOpen(item):
+                return
```
phw, line 145: Just in case this would come up: I intentionally put this into a separate `if`. I think it makes it easier to comment this workaround and read what's going on.

zas (approved):

PR #1727, opened 2021-01-01
PICARD-1926: finish config changes

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1926, PICARD-2088

This finishes #1720 and ensures `get_config` is used instead of accessing the global config object directly.

# Solution

Use `get_config` to access config. Remove now obsolete locking mechanism of `ConfigSection`. In my tests this resolved PICARD-2088.

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -346,6 +346,8 @@ def get_config():
     thread_id = threading.get_ident()
     thread_config = _thread_configs.get(thread_id)
     if not thread_config:
+        if not config:
+            return None  # Not yet initialized
         _thread_config_lock.acquire()
         try:
             config_file = config.fileName()
--- a/picard/config.py
+++ b/picard/config.py
@@ -40,7 +40,6 @@ from picard import (
     PICARD_VERSION,
     log,
 )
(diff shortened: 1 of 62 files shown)


picard/config.py
```diff
@@ -74,24 +73,16 @@ def __getitem__(self, name):
...
     def remove(self, name):
-        self.lock_for_write()
-        try:
-            if name in self:
-                key = self.key(name)
-                self.__qt_config.remove(key)
-        finally:
-            self.unlock()
+        if name in self:
+            key = self.key(name)
```
zas, line 84: Since `name in self` is using `__contains__` which is also calling `self.key(name)`, we could make it more efficient (and easier to understand imho):

```python
key = self.key(name)
config = self.__qt_config
if config.contains(key):
    config.remove(key)
```

phw: Good point, fixed that

zas (approved):

PR #1728, opened 2021-01-02
PICARD-2089: Support WebP for cover art images

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2089

Adds support for loading WebP images for cover art.

# Solution
Add parsing the sizes from WebP to `picard.util.imageinfo`.

The implementation from `picard.util.bitreader` is taken from mutagen, with small restructuring. We could import it from mutagen, but it is private API there (and the important `LSBBitReader` even lives in the TAK implementation, as it is only used there).

# Action


zas (approved):
LGTM

PR #1730, opened 2021-01-03
Use classes for image type identification code

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/util/imageinfo.py
+++ b/picard/util/imageinfo.py
@@ -83,13 +83,21 @@ class IdentifyJPEG(IdentifyImageType):
         jpeg.read(2)
         b = jpeg.read(1)
         try:
+            # https://en.wikibooks.org/wiki/JPEG_-_Idea_and_Practice/The_header_part
+            # https://www.disktuna.com/list-of-jpeg-markers/
+            # https://de.wikipedia.org/wiki/JPEG_File_Interchange_Format
+            SOF_markers = {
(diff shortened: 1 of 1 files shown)


picard/util/imageinfo.py
```diff
@@ -43,189 +43,236 @@ class UnexpectedError(IdentificationError):
...
+        def _tiff_read_value(type, order, data):
+            if type == TIFF_TYPE_LONG:
+                value = data[:4]
+                format = order + 'I'
+            elif type == TIFF_TYPE_SHORT:
+                value = data[:2]
+                format = order + 'H'
+            else:
+                raise UnexpectedError('TIFF: unexpected field type %s' % type)
+            return struct.unpack(format, value)[0]
```
phw, line 221: I would not define this inside the function but on the class itself (`self._read_value`, the tiff prefix can also be dropped).

Same probably for the constants.

phw (changes requested):
Well done. I would restructure the TIFF class a bit to not define everything inside the read function, but otherwise LGTM

picard/util/imageinfo.py
zas: Agreed, done in 7bf8e06

phw (approved):
Cool. Much clearer to extend for new formats in the future.

PR #1732, opened 2021-01-04
PICARD-204: Support for recording-level original release date

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Use first-release-date of recordings if available.

# Problem

* JIRA ticket (_optional_): PICARD-204

# Solution

Use the `first-release-date` field on recordings if available and fill `originaldate` and `originalyear` with this information. If the data is not set on a recording (should not happen actually once this is fully rolled out) it would still use the data from the release group.

For more flexibility this makes available two additional variables `~recordingoriginaldate` and `~releaseoriginaldate` for scripting. This e.g. allows to always use the original date for the entire release instead of using the per recording date.

# Action

At this moment this is partially rolled out on musicbrainz.org. This patch already works for getting originaldate when querying standalone-recordings. For full support also for releases with different per-recording dates https://github.com/metabrainz/musicbrainz-server/pull/1847 needs to be merged and rolled out.

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -384,9 +384,11 @@ def recording_to_metadata(node, m, track=None):
             add_isrcs_to_metadata(value, m)
         elif key == 'video' and value:
             m['~video'] = '1'
+    config = get_config()
     if m['title']:
         m['~recordingtitle'] = m['title']
-    if m['~recordingoriginaldate']:
+    if m['~recordingoriginaldate'] and (
+        config.setting["originaldate_use_recording"] or not m['originaldate']):
         m['originaldate'] = m['~recordingoriginaldate']
         m['originalyear'] = m['originaldate'][:4]
     if m.length:
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -387,8 +387,7 @@ def recording_to_metadata(node, m, track=None):
     config = get_config()
     if m['title']:
         m['~recordingtitle'] = m['title']
-    if m['~recordingoriginaldate'] and (
-        config.setting["originaldate_use_recording"] or not m['originaldate']):
+    if m['~recordingoriginaldate'] and config.setting["originaldate_use_recording"]:
         m['originaldate'] = m['~recordingoriginaldate']
         m['originalyear'] = m['originaldate'][:4]
     if m.length:
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -5,7 +5,7 @@
 # Copyright (C) 2017 David Mandelberg
 # Copyright (C) 2017-2018 Sambhav Kothari
 # Copyright (C) 2017-2019 Laurent Monin
-# Copyright (C) 2018-2020 Philipp Wolfer
+# Copyright (C) 2018-2021 Philipp Wolfer
 # Copyright (C) 2019 Michael Wiencek
 #
 # This program is free software; you can redistribute it and/or
@@ -73,6 +73,7 @@ _MEDIUM_TO_METADATA = {
 
 _RECORDING_TO_METADATA = {
     'disambiguation': '~recordingcomment',
+    'first-release-date': '~recordingoriginaldate',
     'title': 'title',
 }
 
@@ -93,7 +94,7 @@ _ARTIST_TO_METADATA = {
 
 _RELEASE_GROUP_TO_METADATA = {
     'disambiguation': '~releasegroupcomment',
-    'first-release-date': 'originaldate',
+    'first-release-date': '~releaseoriginaldate',
     'title': '~releasegroup',
 }
 
@@ -385,6 +386,9 @@ def recording_to_metadata(node, m, track=None):
             m['~video'] = '1'
     if m['title']:
         m['~recordingtitle'] = m['title']
+    if m['~recordingoriginaldate']:
+        m['originaldate'] = m['~recordingoriginaldate']
+        m['originalyear'] = m['originaldate'][:4]
     if m.length:
         m['~length'] = format_time(m.length)
     if 'instrumental' in m.getall('~performance_attributes'):
@@ -494,7 +498,8 @@ def release_group_to_metadata(node, m, release_group=None):
         elif key == 'secondary-types':
             add_secondary_release_types(value, m)
     add_genres_from_node(node, release_group)
-    if m['originaldate']:
+    if m['~releaseoriginaldate']:
+        m['originaldate'] = m['~releaseoriginaldate']
         m['originalyear'] = m['originaldate'][:4]
     m['releasetype'] = m.getall('~primaryreleasetype') + m.getall('~secondaryreleasetype')
 
--- a/picard/ui/options/metadata.py
+++ b/picard/ui/options/metadata.py
@@ -75,7 +75,7 @@ class MetadataOptionsPage(OptionsPage):
         BoolOption("setting", "convert_punctuation", True),
         BoolOption("setting", "standardize_artists", False),
         BoolOption("setting", "standardize_instruments", True),
-        BoolOption("setting", "originaldate_use_recording", True),
+        BoolOption("setting", "originaldate_use_recording", False),
     ]
 
     def __init__(self, parent=None):
--- a/picard/ui/options/metadata.py
+++ b/picard/ui/options/metadata.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006-2008, 2011 Lukáš Lalinský
-# Copyright (C) 2008-2009, 2018-2019 Philipp Wolfer
+# Copyright (C) 2008-2009, 2018-2021 Philipp Wolfer
 # Copyright (C) 2011 Johannes Weißl
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2013, 2018 Laurent Monin
@@ -75,6 +75,7 @@ class MetadataOptionsPage(OptionsPage):
         BoolOption("setting", "convert_punctuation", True),
         BoolOption("setting", "standardize_artists", False),
         BoolOption("setting", "standardize_instruments", True),
+        BoolOption("setting", "originaldate_use_recording", True),
     ]
 
     def __init__(self, parent=None):
@@ -103,6 +104,10 @@ class MetadataOptionsPage(OptionsPage):
         self.ui.nat_name.setText(config.setting["nat_name"])
         self.ui.standardize_artists.setChecked(config.setting["standardize_artists"])
         self.ui.standardize_instruments.setChecked(config.setting["standardize_instruments"])
+        if config.setting["originaldate_use_recording"]:
+            self.ui.originaldate_use_recording.setChecked(True)
+        else:
+            self.ui.originaldate_use_releasegroup.setChecked(True)
 
     def save(self):
         config = get_config()
@@ -119,6 +124,7 @@ class MetadataOptionsPage(OptionsPage):
                 self.tagger.nats.update()
         config.setting["standardize_artists"] = self.ui.standardize_artists.isChecked()
         config.setting["standardize_instruments"] = self.ui.standardize_instruments.isChecked()
+        config.setting["originaldate_use_recording"] = self.ui.originaldate_use_recording.isChecked()
 
     def set_va_name_default(self):
         self.ui.va_name.setText(self.options[0].default)
--- a/test/test_config_upgrade.py
+++ b/test/test_config_upgrade.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2019 Laurent Monin
-# Copyright (C) 2019-2020 Philipp Wolfer
+# Copyright (C) 2019-2021 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -52,6 +52,7 @@ from picard.config_upgrade import (
     upgrade_to_v2_5_0_dev_1,
     upgrade_to_v2_5_0_dev_2,
     upgrade_to_v2_6_0_dev_1,
+    upgrade_to_v2_6_0_dev_2,
 )
 from picard.const import (
     DEFAULT_FILE_NAMING_FORMAT,
@@ -317,3 +318,9 @@ class TestPicardConfigUpgrades(TestPicardConfigCommon):
         upgrade_to_v2_6_0_dev_1(self.config)
         picard.config_upgrade.IS_FROZEN = False
         self.assertEqual("", self.config.setting["acoustid_fpcalc"])
+
+    def test_upgrade_to_v2_6_0_dev_2(self):
+        BoolOption("setting", "originaldate_use_recording", True)
+        self.assertTrue(self.config.setting["originaldate_use_recording"])
+        upgrade_to_v2_6_0_dev_2(self.config)
+        self.assertFalse(self.config.setting["originaldate_use_recording"])
--- a/test/test_config_upgrade.py
+++ b/test/test_config_upgrade.py
@@ -52,7 +52,6 @@ from picard.config_upgrade import (
     upgrade_to_v2_5_0_dev_1,
     upgrade_to_v2_5_0_dev_2,
     upgrade_to_v2_6_0_dev_1,
-    upgrade_to_v2_6_0_dev_2,
 )
 from picard.const import (
     DEFAULT_FILE_NAMING_FORMAT,
@@ -318,9 +317,3 @@ class TestPicardConfigUpgrades(TestPicardConfigCommon):
         upgrade_to_v2_6_0_dev_1(self.config)
         picard.config_upgrade.IS_FROZEN = False
         self.assertEqual("", self.config.setting["acoustid_fpcalc"])
-
-    def test_upgrade_to_v2_6_0_dev_2(self):
-        BoolOption("setting", "originaldate_use_recording", True)
-        self.assertTrue(self.config.setting["originaldate_use_recording"])
-        upgrade_to_v2_6_0_dev_2(self.config)
-        self.assertFalse(self.config.setting["originaldate_use_recording"])
--- a/test/test_mbjson.py
+++ b/test/test_mbjson.py
@@ -150,55 +150,69 @@ class ReleaseTest(MBJSONTest):
         formats = media_formats_from_node(self.json_doc['media'])
         self.assertEqual(formats, '12" Vinyl')
 
-    def test_originaldate_recording(self):
-        config.setting["originaldate_use_recording"] = True
+    def test_originaldate_for_release(self):
         metadata = Metadata()
         release_group_to_metadata(self.json_doc['release-group'], metadata)
         release_to_metadata(self.json_doc, metadata)
+        self.assertEqual('1973-03-24', metadata['~releaseoriginaldate'])
         self.assertEqual('1973-03-24', metadata['originaldate'])
+        self.assertEqual('1973', metadata['originalyear'])
 
-        # Track 1 has a custom first release date
+    def test_originaldate_recording_1(self):
+        # Track 1 has a custom first release date, it should overwrite the existing date
         track1 = self.json_doc['media'][0]['tracks'][0]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track1['recording'], recording_metadata)
-        self.assertEqual('1972-02-23', recording_metadata['originaldate'])
-        self.assertEqual('1972', recording_metadata['originalyear'])
+        metadata = Metadata()
+        metadata['originaldate'] = '1973-03-24'
+        metadata['originalyear'] = '1973'
+        config.setting["originaldate_use_recording"] = True
+        recording_to_metadata(track1['recording'], metadata)
+        self.assertEqual('1972-02-23', metadata['~recordingoriginaldate'])
+        self.assertEqual('1972-02-23', metadata['originaldate'])
+        self.assertEqual('1972', metadata['originalyear'])
 
+    def test_originaldate_recording_2(self):
         # Track 2 has the same first release date set as the release
         track2 = self.json_doc['media'][0]['tracks'][1]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track2['recording'], recording_metadata)
-        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
-        self.assertEqual('1973', recording_metadata['originalyear'])
-
-        # Track 3 has no first release date set
-        track2 = self.json_doc['media'][0]['tracks'][2]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track2['recording'], recording_metadata)
-        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
-        self.assertEqual('1973', recording_metadata['originalyear'])
-
-    def test_originaldate_release_group(self):
-        config.setting["originaldate_use_recording"] = False
         metadata = Metadata()
-        release_group_to_metadata(self.json_doc['release-group'], metadata)
-        release_to_metadata(self.json_doc, metadata)
+        metadata['originaldate'] = '1973-03-24'
+        metadata['originalyear'] = '1973'
+        recording_to_metadata(track2['recording'], metadata)
+        self.assertEqual('1973-03-24', metadata['~recordingoriginaldate'])
         self.assertEqual('1973-03-24', metadata['originaldate'])
-        # Track 1 has a custom first release date
-        track1 = self.json_doc['media'][0]['tracks'][0]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track1['recording'], recording_metadata)
-        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
-        self.assertEqual('1973', recording_metadata['originalyear'])
+        self.assertEqual('1973', metadata['originalyear'])
 
-    def test_originaldate_release_group_fallback_to_recording(self):
+    def test_originaldate_recording_3(self):
+        # Track 3 has no first release date set, existing data is kept
+        track2 = self.json_doc['media'][0]['tracks'][2]
+        metadata = Metadata()
+        metadata['originaldate'] = '1973-03-24'
+        metadata['originalyear'] = '1973'
+        recording_to_metadata(track2['recording'], metadata)
+        self.assertEqual('', metadata['~recordingoriginaldate'])
+        self.assertEqual('1973-03-24', metadata['originaldate'])
+        self.assertEqual('1973', metadata['originalyear'])
+
+    def test_originaldate_recording_use_release_group(self):
+        # Track 1 has a custom first release date, but the existing one should be used
+        track1 = self.json_doc['media'][0]['tracks'][0]
+        metadata = Metadata()
+        metadata['originaldate'] = '1973-03-24'
+        metadata['originalyear'] = '1973'
         config.setting["originaldate_use_recording"] = False
+        recording_to_metadata(track1['recording'], metadata)
+        self.assertEqual('1972-02-23', metadata['~recordingoriginaldate'])
+        self.assertEqual('1973-03-24', metadata['originaldate'])
+        self.assertEqual('1973', metadata['originalyear'])
+
+    def test_originaldate_recording_use_release_group_fallback_to_recording(self):
         # Track 1 has a custom first release date and not date is set already in the metadata
         track1 = self.json_doc['media'][0]['tracks'][0]
-        recording_metadata = Metadata()
-        recording_to_metadata(track1['recording'], recording_metadata)
-        self.assertEqual('1972-02-23', recording_metadata['originaldate'])
-        self.assertEqual('1972', recording_metadata['originalyear'])
+        metadata = Metadata()
+        config.setting["originaldate_use_recording"] = False
+        recording_to_metadata(track1['recording'], metadata)
+        self.assertEqual('1972-02-23', metadata['~recordingoriginaldate'])
+        self.assertEqual('1972-02-23', metadata['originaldate'])
+        self.assertEqual('1972', metadata['originalyear'])
 
 
 class NullReleaseTest(MBJSONTest):
--- a/test/test_mbjson.py
+++ b/test/test_mbjson.py
@@ -54,7 +54,8 @@ settings = {
     "standardize_instruments": True,
     "release_ars": True,
     "preferred_release_countries": [],
-    "artist_locale": 'en'
+    "artist_locale": 'en',
+    "originaldate_use_recording": True
 }
 
 
@@ -149,7 +150,36 @@ class ReleaseTest(MBJSONTest):
         formats = media_formats_from_node(self.json_doc['media'])
         self.assertEqual(formats, '12" Vinyl')
 
-    def test_originaldate(self):
+    def test_originaldate_recording(self):
+        config.setting["originaldate_use_recording"] = True
+        metadata = Metadata()
+        release_group_to_metadata(self.json_doc['release-group'], metadata)
+        release_to_metadata(self.json_doc, metadata)
+        self.assertEqual('1973-03-24', metadata['originaldate'])
+
+        # Track 1 has a custom first release date
+        track1 = self.json_doc['media'][0]['tracks'][0]
+        recording_metadata = Metadata(metadata)
+        recording_to_metadata(track1['recording'], recording_metadata)
+        self.assertEqual('1972-02-23', recording_metadata['originaldate'])
+        self.assertEqual('1972', recording_metadata['originalyear'])
+
+        # Track 2 has the same first release date set as the release
+        track2 = self.json_doc['media'][0]['tracks'][1]
+        recording_metadata = Metadata(metadata)
+        recording_to_metadata(track2['recording'], recording_metadata)
+        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
+        self.assertEqual('1973', recording_metadata['originalyear'])
+
+        # Track 3 has no first release date set
+        track2 = self.json_doc['media'][0]['tracks'][2]
+        recording_metadata = Metadata(metadata)
+        recording_to_metadata(track2['recording'], recording_metadata)
+        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
+        self.assertEqual('1973', recording_metadata['originalyear'])
+
+    def test_originaldate_release_group(self):
+        config.setting["originaldate_use_recording"] = False
         metadata = Metadata()
         release_group_to_metadata(self.json_doc['release-group'], metadata)
         release_to_metadata(self.json_doc, metadata)
@@ -158,20 +188,17 @@ class ReleaseTest(MBJSONTest):
         track1 = self.json_doc['media'][0]['tracks'][0]
         recording_metadata = Metadata(metadata)
         recording_to_metadata(track1['recording'], recording_metadata)
+        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
+        self.assertEqual('1973', recording_metadata['originalyear'])
+
+    def test_originaldate_release_group_fallback_to_recording(self):
+        config.setting["originaldate_use_recording"] = False
+        # Track 1 has a custom first release date and not date is set already in the metadata
+        track1 = self.json_doc['media'][0]['tracks'][0]
+        recording_metadata = Metadata()
+        recording_to_metadata(track1['recording'], recording_metadata)
         self.assertEqual('1972-02-23', recording_metadata['originaldate'])
         self.assertEqual('1972', recording_metadata['originalyear'])
-        # Track 2 has the same first release date set as the release
-        track2 = self.json_doc['media'][0]['tracks'][1]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track2['recording'], recording_metadata)
-        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
-        self.assertEqual('1973', recording_metadata['originalyear'])
-        # Track 3 has no first release date set
-        track2 = self.json_doc['media'][0]['tracks'][2]
-        recording_metadata = Metadata(metadata)
-        recording_to_metadata(track2['recording'], recording_metadata)
-        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
-        self.assertEqual('1973', recording_metadata['originalyear'])
 
 
 class NullReleaseTest(MBJSONTest):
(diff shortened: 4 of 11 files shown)


zas (dismissed):
Nice improvement (once MB part will be totally deployed)

Sophist-UK:
Hmm - whilst I like the idea, I am really not sure that this is semantically correct i.e. the earliest release date of a release within a release group is **not** the same as the earliest date that the recording was released. Two examples to illustrate this:

1. Tracks on a "Best of" Release will most often be based on earlier recordings, say off the original album. So each track on a "Best of" release might have a different first-release-date - whereas at present the original release date is the same for all tracks on the album because it is based on the (common) Release Group.
2. But even if we look at the original album, the earliest release date of the original Album (i.e. from the Release Group) may not be the earliest release date of the track e.g. if it was previously released as a single.

It seems to me that we need either to:

1. Provide an Options... / Metadata setting for the user to choose which method they want to use... ; or
2. Simply set the hidden variables and let the user override the release group earliest release date with the recording earliest release date in a script if they wish to do so.

Sophist-UK:
I have a feeling (haven''t had the time to check) that you only get recording level metadata from MB if you set Options... / Metadata / Use Track Relationships - and not everyone will have enabled that.

If this option is disabled, but Release Relationships is enabled, what will the original release date be set to?

If neither is enabled?

phw:
> the earliest release date of a release within a release group is **not** the same as the earliest date that the recording was released.

Yes, of course it's different. That's actually the point of this :) Getting the actual earliest release date per recording is one of the most often asked for features.

> * Provide an Options... / Metadata setting for the user to choose which method they want to use... ; or
> 
> * Simply set hidden variables and let the user override the release group earliest release date with a different earliest release date in a script if they wish to do so.

I thought about adding an option, but for now decided to go with 2: set hidden variables, which can be used in a script. This is already part of this PR.

> If this option is disabled, but Release Relationships is enabled, what will the original release date be set to?

This functionality is unaffected by this. Recordings are always included in the release responses obtained by Picard. The track / release relationship options are only about actual relationships (the `*-rels` inc options of the webservice), because those are more expensive to query.

Sophist-UK:
> Yes, of course it's different. That's actually the point of this :) Getting the actual earliest release date per recording is one of the most often asked for features.

I am not for a moment suggesting that this is not a useful function, but rather that it changes the functionality substantially and creates a significant backwards incompatibility.

> I thought about adding an option, but for now decided to go with 2: set hidden variables, which can be used in a script. This is already part of this PR.

Yes - that bit seems fine to me, it is the bit where we change the definition of an established tag to something quite different that concerns me.

phw:
> Yes - that bit seems fine to me, it is the bit where we change the definition of an established tag to something quite different that concerns me.

I think the recording first release date is the better data. We did use the data from the release group because this was the best we could get at the time. Before we had the dates on the release group it even was empty. Back then you could use e.g. the lastfmplus plugin to fill it (per track).

phw:
We could of course insert a script snipped on update for existing users, to revert to previous functionality. Or if we make it an option we could actually distinguish and make recording level dates the default, but on upgrade stick with release group.

Sophist-UK:
> We could of course insert a script snipped on update for existing users to revert to previous functionality. 

I think this would be confusing to users without a popup explanation. But the script update could create such a popup.

> Or if we make it an option we could actually distinguish and make recording level dates the default, but on upgrade stick with release group.

This sounds OK. However I see this as a significant change in functionality - because previously all tracks would have the same original date / year and after this they could all be different - and if you use e.g. %original year% in a file naming script, then you will get the files for an album spread around across several directories. IMO changing how existing variables are defined and breaking backward compatibility is a **BAD IDEA**.

rdswift:
I tend to agree with @Sophist-UK on this one in that this change will cause existing scripts to break (by producing different results without anything being changed by the user).  If we stick with that, we will need to make it **VERY** clear to the users when it is released.

Also, is this intended to be included in the next 2.5.x release, or is it planned for the 2.6 release?  I'll start working on the changes to the documentation, but need to know which version to update.  Thanks.

phw:
Ok, convinced, I'll add an option to switch this.

@rdswift This change will be part of Picard 2.6. We'll do a beta release for that. We'll also likely have a 2.5.6 release with bugfixes only.

phw:
I have added options to switch between using the first release date of recording or release group as discussed above:

grafik

Together with the new variables this should give enough flexibility to use this as preferred.

picard/config.py
```diff
@@ -195,6 +195,10 @@ def register_upgrade_hook(self, func, *args):
 
     def run_upgrade_hooks(self, outputfunc=None):
         """Executes registered functions to upgrade config version to the latest"""
+        if self._version == Version(0, 0, 0, 'dev', 0):
+            # This is a freshly created config
+            self._version = PICARD_VERSION
+            self._write_version()
```
zas, line 201: Perhaps this should go in a separated PR, it seems unrelated (though it makes sense)

test/test_config_upgrade.py
```diff
@@ -317,3 +318,9 @@ def test_upgrade_to_v2_6_0_dev_1_frozen(self):
         upgrade_to_v2_6_0_dev_1(self.config)
         picard.config_upgrade.IS_FROZEN = False
         self.assertEqual("", self.config.setting["acoustid_fpcalc"])
+
+    def test_upgrade_to_v2_6_0_dev_2(self):
+        BoolOption("setting", "originaldate_use_recording", True),
```
zas, line 323: Extra comma at the end of line

test/test_mbjson.py
```diff
@@ -149,6 +150,53 @@ def test_media_formats_from_node(self):
...
+        recording_to_metadata(track2['recording'], recording_metadata)
+        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
+        self.assertEqual('1973', recording_metadata['originalyear'])
+        # Track 3 has no first release date set
+        track2 = self.json_doc['media'][0]['tracks'][2]
+        recording_metadata = Metadata(metadata)
+        recording_to_metadata(track2['recording'], recording_metadata)
+        self.assertEqual('1973-03-24', recording_metadata['originaldate'])
+        self.assertEqual('1973', recording_metadata['originalyear'])
+
```
zas, line 177: Hard to read, is it possible to split in smaller functions? or, at least, add few blank lines

zas (changes requested):
Clearly for 2.6

picard/config.py
phw: True, I added https://github.com/metabrainz/picard/pull/1733 . I will rebase this PR once that got merged.

test/test_mbjson.py
phw: I split the tests into smaller tests, more limited to testing the actual code.

picard/ui/options/metadata.py
```diff
@@ -75,6 +75,8 @@ class MetadataOptionsPage(OptionsPage):
         BoolOption("setting", "convert_punctuation", True),
         BoolOption("setting", "standardize_artists", False),
         BoolOption("setting", "standardize_instruments", True),
+        BoolOption("setting", "standardize_instruments", True),
```
zas, line 78: Duplicated line

picard/mbjson.py
```diff
@@ -383,8 +384,13 @@ def recording_to_metadata(node, m, track=None):
             add_isrcs_to_metadata(value, m)
         elif key == 'video' and value:
             m['~video'] = '1'
+    config = get_config()
     if m['title']:
         m['~recordingtitle'] = m['title']
+    if m['~recordingoriginaldate'] and (
+        config.setting["originaldate_use_recording"] or not m['originaldate']):
```
zas, line 391: Unrelated to this PR but this use-case of `get_config()` makes me think we perhaps need a shortcut:

```python
def get_config_setting(name):
   config = get_config()
   return config.setting[name]
```

Not sure about that, because it may lead to write:

```python
   if get_config_setting('x'):
        ....
   if get_config_setting('y'):
        ....
   if get_config_setting('z'):
        ....
```

which is likely less efficient than:

```python
   config = get_config()
   if config.setting['x']:
        ....
   if config.setting['y']:
        ....
   if config.setting['z']:
        ....
```

Thoughts?

zas (changes requested):

picard/mbjson.py
phw: We could have that in addition for the simple cases, but IMHO there is not much difference between these two:

```python
get_config_setting('x')
get_config().setting['x']
```

That would also be slightly more efficient in this case, but not as readable I think.

And yes, if you need to access it multiple times it's better to get the config object once and reuse it. Or even not get it at all. We should sometimes see if we can avoid accessing config directly from helper functions but instead use specific parameters. That could also be an option here, leaving it to the caller of function to extract the proper value from config.

So e.g. we could have in this case a:

```
def recording_to_metadata(node, m, track=None, set_originaldate=True):
    ...
```

Of course the risk is that we get too many parameters or we have to route them through too many levels.

picard/ui/options/metadata.py
phw: Fixed

zas (dismissed):

rdswift:
> I have added options to switch between using the first release date of recording or release group as discussed above:

Don't get me wrong, I appreciate what you're doing here, but I (personally) have an aversion to using a setting to select between two different contents for the same tag.  I would much prefer both options be available as separate tags/variables, and let the user decide in a script how they want to use them because there may be occasions when having both options available in the same script are required (desired).  This is the very reason that I developed the Additional Artists Variables plugin -- because I needed both the standardized and as-credited artists' names in the same script and the **Use standardized names** setting was limiting me to having only one available.  As long as all options are available, I can live with your original proposal to change the content of the default tag (because I agree that it is better information), even though that means I will need to make a one-time correction to my file naming script.

Thanks for confirming that the changes will be part of the 2.6 release.  I'll start drafting the changes so we have something to discuss / edit.

Sophist-UK:
@phw Philipp - We are agreed that these values should be put in variables - but when Bob and I argue that we should NOT be changing the definition of an existing tag you repeat the same argument that this is OK because the values are in variables. Bob and I have both argued that this is NOT a justification and we are both open to persuasion, yet you continue to repeat the same argument without providing any additional justification.

I can't help but feel that this is NOT how open source collaboration is supposed to work - the purpose of reviews and comments are to debate and reach consensus, not for one member to push through a change when others disagree with its suitability,

Can we please put the merge of this PR on hold until this is resolved?

Sophist-UK:
P.S. I am still concerned that this change will have unintended consequences - and the examples I have raised do not seem to have been addressed.

Sophist-UK:
I think I would be happier if there was code that would change variables in scripts to keep the functionality the same whenever the new option was changed i.e. scripts should be modified so that:

* if you change to "first release date of the recording", then all uses of `%originaldate%` are changed to `%_releaseoriginaldate%` (as well as `$set(originaldate,` to `$set(_releaseoriginaldate,` etc. and the same for `originalyear`
* if you change to "first release date of release group", then you do the same but use `%_recordingoriginaldate%` instead.

picard/mbjson.py
```diff
@@ -383,8 +384,13 @@ def recording_to_metadata(node, m, track=None):
...
+        m['originaldate'] = m['~recordingoriginaldate']
+        m['originalyear'] = m['originaldate'][:4]
     if m.length:
```
Sophist-UK, line 394: a. You can still have original date set from recordings (i.e. with different dates across the album - potentially leading to file path issues) even if you have selected Release Group as the source. This is simply BAD!!! and will confuse people.

b. We probably need `~releaseoriginalyear` and `~recordingoriginalyear` if we implement the script updating functionality I have suggested in my global comment.

ui/options_metadata.ui
```diff
@@ -77,6 +77,27 @@
...
+       </widget>
+      </item>
+      <item>
+       <widget class="QRadioButton" name="originaldate_use_releasegroup">
+        <property name="text">
+         <string>Use the first release date of the release group</string>
+        </property>
+       </widget>
+      </item>
      </layout>
```
Sophist-UK, line 101: I believe that this has so much opportunity to confuse people when file naming scripts go wrong that the UI needs something more to warn / help people.

Sophist-UK (changes requested):
Please find some specific issues highlighted.

rdswift:
@phw I just had a thought about this...  What if we created a new variable containing the version number of the currently running Picard, perhaps something like `_picardversion`?  That way we could start letting people know now of the upcoming change in the `originaldate` and `originalyear` tags, and they could start updating their scripts accordingly by simply checking if the variable is not empty and is greater than or equal to "2.6".  We could promote it heavily in the Community Forum, and even include an appropriate code snippet that the users could use in their scripts.  That way we could help them "version proof" their scripts now, so that when the change happens it's a non-event as far as the scripts go.

phw (commented):
> I can't help but feel that this is NOT how open source collaboration is supposed to work - the purpose of reviews and comments are to debate and reach consensus, not for one member to push through a change when others disagree with its suitability,

I'm seriously confused by this. You brought up the debate above that this change might have unexpected consequences especially on renaming scripts and suggested to have an option to choose the behavior. I added such an option and now it is me "push through a change when others disagree with its suitability".

I am not trying to push something through, but to implement a feature that has quite often be requested and discussed by users. The response in the past has always been that this would be great to have, but technical not really feasible with the current web service. Now we have the web service extended to provide the data and I think we should make use of it.

As I understood your initial comment you are in general in favor of the change but wanted the backward compatibility addressed. That's what I tried to do. From your reaction it is clear to me that you are unhappy with the proposed solution. But it is not quite clear to me how you imagine this to work instead.

I am quite happy to rework this PR, but at the current point of debate I would like to have some concrete suggestion on how to implement this. Should this be done in an entirely different way? Is it just the default value that concerns you or something else?

> I think I would be happier if there was code that would change variables in scripts to keep the functionality the same whenever the new option was changed [...]

I don't think we should do that. For one I think changing user's scripts when flipping options to negate the effect of this option change is definitely not something expected. It also can go easily wrong. E.g. I have the script:

```
$if($eq(%date%,%originaldate%),$delete(originaldate))
```

Now what do we switch? Only reading this tag or also writing? If we just change it for reading tags we might get:

```
$if($eq(%date%,%_releaseoriginaldate%),$delete(originaldate))
```

That's still a working script that has some effect, but not the one I intended. If it also changes the writing part to

```
$if($eq(%date%,%_releaseoriginaldate%),$delete(_releaseoriginaldate))
```

It does nothing useful anymore.

zas:
> I can't help but feel that this is NOT how open source collaboration is supposed to work - the purpose of reviews and comments are to debate and reach consensus, not for one member to push through a change when others disagree with its suitability,

Accusing @phw to "push" changes against community is very unfair. He just defends his point of view, which is equally important as yours or anyone's. Note that nothing was merged yet.
And as him, I fail to understand which solution you are suggesting.


@Sophist-UK @rdswift : can you explain clearly how you think this issue should be fixed?
@phw: move the discussion to forums

Sophist-UK:
> Accusing @phw to "push" changes against community is very unfair. He just defends his point of view, which is equally important as yours or anyone's. Note that nothing was merged yet.
> And as him, I fail to understand which solution you are suggesting.

The reason I feel unhappy is that unfortunately (and very unusually - for both Philipp and the MB Community generally) Philipp has NOT defended his point of view - I have raised specific examples which potentially create a problem, and these examples have not been addressed, and repeating a statement that variables have been created does not address these examples.

The solution I really believe in (as an advanced user) is to simply create new variables and leave it to the user to change scripts if they want to.

However, I realise that this doesn't address what Philipp (and possibly others in the Community) are saying is a real need i.e. to use recording first release dates as original date and to make it easy to do so (i.e. without scripts) - and I can genuinely see the benefit of that - in which case we need an option as Philipp has created.

BUT, the points I have summarised have **not** been addressed.

1. The risk of file naming scripts scattering albums, and warning users that recording first release dates can be different across tracks and it shouldn't be used for file paths.
2. Recording first release dates being potentially wrong - we have no analysis of how frequently different recordings of the same work by the same artist will give rise to incorrect first release dates for a track.

> @Sophist-UK @rdswift : can you explain clearly how you think this issue should be fixed?

How to address these:

1. IMO there needs to be a brief warning on the Options / Metadata page about switching to recording first release dates with a link to a documentation page that gives explicit warnings and examples.

2. For Picard we could see whether it is possible to get first release date for all recordings of a work (by this artist or by any artist). I suspect that this would mean turning on Recording relationships by default, in which case we would need to assess the impact of server load. 

(For Musicbrainz, we should get a feel for how bad the duplicate recordings issue is and for how often different recordings will give a wrong first release date, and perhaps look at whether we can clean up duplicate recordings (and improve the release creation to avoid creating new duplicates.)

picard/mbjson.py
Sophist-UK: Yes - fallback seems fine for NAT tracks.

We only need `~releaseoriginalyear` and `~recordingoriginalyear` if we implement automatic script renaming - and as you have pointed out there is always a counter example where you will break something if you do this.

Sophist-UK:
> @phw I just had a thought about this... What if we created a new variable containing the version number of the currently running Picard, perhaps something like `_picardversion`? That way we could start letting people know now of the upcoming change in the `originaldate` and `originalyear` tags, and they could start updating their scripts accordingly by simply checking if the variable is not empty and is greater than or equal to "2.6". 

a. Many users don't read the forums. Many will use Picard out of the box and expect it to work the same after an upgrade. Others like me a script users but don't monitor the forums.

b. I am not really sure that what is effectively a transfer of responsibility for breaking backwards compatibility from developers to users is the correct way forwards.

Sophist-UK:
> I don't think we should do that. For one I think changing user's scripts when flipping options to negate the effect of this option change is definitely not something expected. It also can go easily wrong. E.g. I have the script:
> 
> ```
> $if($eq(%date%,%originaldate%),$delete(originaldate))
> ```
> 
> Now what do we switch? Only reading this tag or also writing? If we just change it for reading tags we might get:
> 
> ```
> $if($eq(%date%,%_releaseoriginaldate%),$delete(originaldate))
> ```
> 
> That's still a working script that has some effect, but not the one I intended. If it also changes the writing part to
> 
> ```
> $if($eq(%date%,%_releaseoriginaldate%),$delete(_releaseoriginaldate))
> ```
> 
> It does nothing useful anymore.

I agree - there is almost certainly a counter-example of how we will break a script by such changes. Forget that idea.

phw:
I have started a [discussion in the forums](https://community.metabrainz.org/t/original-release-date-community-opinion-on-how-to-handle-the-originaldate-tag-first-release-date-of-release-group-and-or-recording/512588) and also added https://github.com/metabrainz/picard/pull/1734 which would only add variables `%_recordingoriginaldate%` and `%_releaseoriginaldate%` and set `originaldate` / `originalyear` for non-album tracks.

> and these examples have not been addressed

I have not addressed the specific examples. But what they have in common is that they are examples where the first release date per recording differs from the first release date of the RG (often earlier in case of compilations or albums with previous single releases, maybe later on re-releases which include additional tracks etc.). But these are not unintended side effects, these examples are the very reason for why someone wants to have the per recording date.

> The risk of file naming scripts scattering albums, and warning users that recording first release dates can be different across tracks and it shouldn't be used for file paths.

Of course this was addressed, by adding the option and defaulting it to the release group for upgrades by default. Sure we can discuss if this change is sufficient, but saying I ignored this when I explicitly added code to address this is not fair.

> Recording first release dates being potentially wrong

That's also true for the release groups date and basically all the data we have.

> IMO there needs to be a brief warning on the Options / Metadata page about switching to recording first release dates with a link to a documentation page that gives explicit warnings and examples.

That's also true for basically all the options in Options > Metadata, Options > Tags and Options > File Naming. I really don't see the original date as much different. But I agree that we should make it clearer what impact an option has. If we go with having an actual option to change this behavior it could maybe warn in the specific case that you a) switch to using the recording date and b) `%originaldate%` and/or `%originalyear%` is being used in the file naming script.


>     For Picard we could see whether it is possible to get first release date for all recordings of a work (by this artist or by any artist).

For this I would argue the same as before we had the `first-release-date` per recording: Querying this data might significantly slow down recording time. So I would see this rather inside a plugin instead of core functionality. It might also not always be what is wanted. I had written some thoughts about implementing this at https://community.metabrainz.org/t/how-to-get-picard-writing-the-original-release-date-year-of-a-recording/355388/28

If MB web service would provide a similar date field on the work (date of first released recording) this would be a different story and we could provide this as another alternative. And yes, it would require getting work data also, but that would be Ok. If you want to have many other of the advanced fields available you need to have track level relationships enabled already.

picard/mbjson.py
phw: I changed the behavior and moved setting the originaldate / originalyear only for non-album tracks unconditionally. For releases it will not change the dates to the recording dates unless explicitly enabled.

I also changed the default to stay with the release group date in any case. So if we decide to have an actual option inside the UI I think this would address the most severe concerns. Plus having a warning of some kind maybe.

Sophist-UK:
> I have not addressed the specific examples. But what they have in common is that they are examples where the first release date per recording differs from the first release date of the RG (often earlier in case of compilations or albums with previous single releases, maybe later on re-releases which include additional tracks etc.). But these are not unintended side effects, these examples are the very reason for why someone wants to have the per recording date.

The specific examples may be a reason to make an improvement, but my examples were given explicitly to highlight issues - and they needed to be addressed explicitly. We should not make improvements which are generally wanted without considering and addresses any side-effects that they might have.

> > The risk of file naming scripts scattering albums, and warning users that recording first release dates can be different across tracks and it shouldn't be used for file paths.

If you responded previously to my comments about this and I missed them, then I apologise unreservedly.

> Of course this was addressed, by adding the option and defaulting it to the release group for upgrades by default. Sure we can discuss if this change is sufficient, but saying I ignored this when I explicitly added code to address this is not fair.

Yes - you did make it default to disabled for existing users. But I spent the time giving a detailed explanation of the various issues, and the code change didn't address all of them, and there wasn't any comment discussion of the points I raised either.

> > Recording first release dates being potentially wrong
>
> That's also true for the release groups date and basically all the data we have.

No - this is different. 

Release Group first release date is generally going to be correct because generally MB data in this area is good. And mistakes are either missing or genuinely incorrect data.

Issues with recording first release date are both more endemic and a consequence of a release editing approach which encouraged duplicate recordings rather than encouraged recording reuse unless known to be different. The reality of the music world is that recordings are reused (either as-is, or with minor edits or with remixing or even remastering) much more often than tracks are re-recorded. 

> > IMO there needs to be a brief warning on the Options / Metadata page about switching to recording first release dates with a link to a documentation page that gives explicit warnings and examples.
> 
> That's also true for basically all the options in Options > Metadata, Options > Tags and Options > File Naming. I really don't see the original date as much different. But I agree that we should make it clearer what impact an option has. If we go with having an actual option to change this behavior it could maybe warn in the specific case that you a) switch to using the recording date and b) `%originaldate%` and/or `%originalyear%` is being used in the file naming script.
>

No - this is different. We are changing the definition of an existing tag so that it works differently. If a user doesn't realise this and saves his album collection, they will have a big shock and potentially have a huge task to collect everything back together again (if they don't realise that Picard can collect them together when they fix the file naming script).
 
> > For Picard we could see whether it is possible to get first release date for all recordings of a work (by this artist or by any artist).
> 
> For this I would argue the same as before we had the `first-release-date` per recording: Querying this data might significantly slow down recording time. So I would see this rather inside a plugin instead of core functionality. It might also not always be what is wanted. I had written some thoughts about implementing this at https://community.metabrainz.org/t/how-to-get-picard-writing-the-original-release-date-year-of-a-recording/355388/28
>
> If MB web service would provide a similar date field on the work (date of first released recording) this would be a different story and we could provide this as another alternative. And yes, it would require getting work data also, but that would be Ok. If you want to have many other of the advanced fields available you need to have track level relationships enabled already.

And as I have pointed out in my example, using the recording first release date may be wrong a significant proportion of the time. And again, if you did respond explicitly to the issues raised in my comment, I apoloigise for having missed them,

I haven't had time to check the contents of the WS response with the Use-Rels options enabled to see if you get details of other recordings of the same work, but I am guessing by reading between the lines of the above comment that it isnt and that you would have to do one or more separate WS queries to get this data. If this is true, then I would agree that until we get an earliest release date on a work (see comments to your suggestion about this below), we should lave this to a plugin rather than do it in Picard native.

But if perchance it is provided when you enabled the appropriate Use-Rels then perhaps we can have additional choice(s) for this new option that give the earliest first release date of all recordings for this work by the same artist (or same primary artist).

HOWEVER, adding a first release date to a work is more complex. A quick braindump suggests that we might want:

a. A manually set first release date for e.g. the date that Beethoven's Fifth Symphony was first performed (though I suspect that any recording of this has been lost in the mists of time... ;-) )

b. A first release date being the earliest of all recordings of this work;

c. A first release date for each primary artist who recorded it;

d. Different first release dates for studio and live recordings by each primary artist.

BUT the need for this is exacerbated by the proliferation of duplicate recordings.

Sophist-UK:
P.S. Are there situations where there is no recording first release date but there is a release group first release date? (Seems to me to be unlikely, just not sure whether it is impossible or not.)

phw:
> P.S. Are there situations where there is no recording first release date but there is a release group first release date? (Seems to me to be unlikely, just not sure whether it is impossible or not.)

As I understand it this could not happen. The recording would be part of a release, let's say only of exactly this one. Either this release has a release event, at which point also the recording would have a first release date and the release group definitely also. Or it has none. In this case the recording also has no first release date. The release group could or could not have a first release date, depending on whether there are other releases with release event.

> if you did respond explicitly to the issues raised in my comment, I apologise for having missed them,

Actually no, I think did not respond to the specific issues, as I don't even know which ones we are currently talking about. When I wrote "I have not addressed the specific examples" I meant these here:

> 1. Tracks on a "Best of" Release will most often be based on earlier recordings, say off the original album. So each track on a "Best of" release might have a different first-release-date - whereas at present the original release date is the same for all tracks on the album because it is based on the (common) Release Group.
> 2. But even if we look at the original album, the earliest release date of the original Album (i.e. from the Release Group) may not be the earliest release date of the track e.g. if it was previously released as a single.
> 3. It is quite possible that a remastered version of the "Best of" album uses a new recording - in which case you might even get an recording earliest release date which is actually later than the Release Group earliest release date.

I don't know really what to answer here specifically. These examples are all correct. Those are cases where the recordings first release date differs from the one of the RG. And yes, that means it can (and likely will) be different across tracks. The only issue I see with this is when `originaldate` gets used as a directory name. And that IMHO is covered by keeping the RG date the default for `originaldate` and having `%_releaseoriginaldate%` available.

> Issues with recording first release date are both more endemic and a consequence of a release editing approach which encouraged duplicate recordings rather than encouraged recording reuse unless known to be different.

Yes, that's a fair point. It's also easier to correct a wrong earliest date for a release group. In the worst case you only now the date and add a stub release without track list. For fixing a recording's date much more research and editing work is necessary.

> No - this is different. We are changing the definition of an existing tag so that it works differently.

I would not really call it a change of definition. This is still the earliest release date. We are changing the data source. Which one to use is — and I think there is agreement on this part — a matter of preference. And again, other options also change the value of tags. Standardize and translate artist quite significantly alters the artist names (also making them more or less suited for folder names), toggling release or track relationships makes entire tags be available or not, enabling folksonomy tags for genres can give quite different results. We have those options **because** they alter the behavior of the application in a significant way, not despite of this.

As I wrote it above I find it an interesting idea to add specific warnings if e.g. the `originaldate` tag is used in naming scripts. That definitely makes sense.

Ok, anyway. Given the discussion my personal preferred solution at this moment would actually be to just add the variable, thus go with https://github.com/metabrainz/picard/pull/1734/ . This does provide the tools for users wanting to use this data, while not triggering the problematic cases of unintentional changes of existing scripts or misinterpreted options in the UI. Is this something we could actually get an agreement on?

Sophist-UK:
> > P.S. Are there situations where there is no recording first release date but there is a release group first release date? (Seems to me to be unlikely, just not sure whether it is impossible or not.)
> 
> As I understand it this could not happen. The recording would be part of a release, let's say only of exactly this one. Either this release has a release event, at which point also the recording would have a first release date and the release group definitely also. Or it has none. In this case the recording also has no first release date. The release group could or could not have a first release date, depending on whether there are other releases with release event.

That's what I thought likely bust just wanted to check that we didn't need to handle this.
 
> > if you did respond explicitly to the issues raised in my comment, I apologise for having missed them,
> 
> Actually no, I think did not respond to the specific issues, as I don't even know which ones we are currently talking about. When I wrote "I have not addressed the specific examples" I meant these here:
> 
> > 1. Tracks on a "Best of" Release will most often be based on earlier recordings, say off the original album. So each track on a "Best of" release might have a different first-release-date - whereas at present the original release date is the same for all tracks on the album because it is based on the (common) Release Group.
> > 2. But even if we look at the original album, the earliest release date of the original Album (i.e. from the Release Group) may not be the earliest release date of the track e.g. if it was previously released as a single.
> > 3. It is quite possible that a remastered version of the "Best of" album uses a new recording - in which case you might even get an recording earliest release date which is actually later than the Release Group earliest release date.
> 
> I don't know really what to answer here specifically. These examples are all correct. Those are cases where the recordings first release date differs from the one of the RG. And yes, that means it can (and likely will) be different across tracks. The only issue I see with this is when `originaldate` gets used as a directory name. And that IMHO is covered by keeping the RG date the default for `originaldate` and having `%_releaseoriginaldate%` available.

My concern was NOT that these would cause differing dates across tracks, but rather than the data might actually be pseudo random in nature. You know the old adage - which is a better clock, one which is right twice a day or one which is right only once a month? In this case, I am concerned that the date returned may or may not be correct - it ciould be wildly out - and it is pot luck whether you get the right date or not. If users start getting weird and inconsistent dates, it will generate a lot of support traffic and potentially harm the MusicBrainz / Picard reputation.

> > Issues with recording first release date are both more endemic and a consequence of a release editing approach which encouraged duplicate recordings rather than encouraged recording reuse unless known to be different.
> 
> Yes, that's a fair point. It's also easier to correct a wrong earliest date for a release group. In the worst case you only know the date and add a stub release without track list. For fixing a recording's date much more research and editing work is necessary.

> > No - this is different. We are changing the definition of an existing tag so that it works differently.
> 
> I would not really call it a change of definition. This is still the earliest release date. We are changing the data source. Which one to use is — and I think there is agreement on this part — a matter of preference. And again, other options also change the value of tags. Standardize and translate artist quite significantly alters the artist names (also making them more or less suited for folder names), toggling release or track relationships makes entire tags be available or not, enabling folksonomy tags for genres can give quite different results. We have those options **because** they alter the behavior of the application in a significant way, not despite of this.

Whether we call it a change of definition or not, the consequences of switching to recording first release dates if you have a file naming script that uses %originalyear% could be significant. Whilst standardise & translate artist might change the directory name, it won't scatter your files to the four winds - nor will turning on Use-Rels options. This is significantly different in this respect and if we don't warn users we may generate significant support traffic and harm the Picard reputation.

> As I wrote it above I find it an interesting idea to add specific warnings if e.g. the `originaldate` tag is used in naming scripts. That definitely makes sense.

If I am listing my ideal solution, there would be a general warning on Options / Metadata, and a specific warning if using %originaldate/year% directly in a file naming script (but too complicated to track use to set a variable in a tagging script which is then used in a file naming script).

> Ok, anyway. Given the discussion my personal preferred solution at this moment would actually be to just add the variable, thus go with #1734 . This does provide the tools for users wanting to use this data, while not triggering the problematic cases of unintentional changes of existing scripts or misinterpreted options in the UI. Is this something we could actually get an agreement on?

I believe we are starting to reach consensus on this PR, but equally happy for #1734 to be merged in the meantime as an interim solution.

rdswift:
> @rdswift : can you explain clearly how you think this issue should be fixed?

I've added my thoughts to the discussion in the [Community Forum](https://community.metabrainz.org/t/original-release-date-community-opinion-on-how-to-handle-the-originaldate-tag-first-release-date-of-release-group-and-or-recording/512588/7).

After considering the whole thing a bit further, I believe that the default content of the `originaldate` and `originalyear` tags (as opposed to variables) should be the "best" information available -- in this case based on the recording dates.  This can easily be changed in scripts if desired using the new `_releaseoriginaldate` and `_recordingoriginaldate` variables.

Because all of the information is available for scripting use, either through tags or variables, I see no need for yet another UI setting.  Another setting means another complication or item to check when trying to troubleshoot an issue.  How many times have issues been brought up as a result of the "Use standardized artists" setting -- either from the users not knowing it existed or not understanding how it impacted the information provided?  I believe that having both sets of information available as variables removes the need for such a setting.

Similarly, I do not believe that any sort of UI popup or information dialog is appropriate (or required) because of a change (improvement) to the code.  This can (and should) be handled in the documentation such as the User Guide, release announcement (change list), and Community Forum.  One thing that I might suggest is to include a link to the release announcement in the "update available" dialog box so that the user can quickly and easily see what has changed.

Sophist-UK:
hmmm - just read the new community thread. Perhaps we should keep the new option after all because of the data quality issues.

rdswift:
> Does this change of viewpoint help or just annoy further?

[Chuckling]  I think this helps.  Seems like we're getting closer to consensus.

First off, I don't meant to try to diminish or dismiss your concerns regarding backward compatibility and possibly breaking current scripts.  This is a real concern and should be considered.  If there is one absolutely correct way of addressing it, I certainly don't know what it is.  Some developers note in output to standard error or log messages that something is going to be deprecated (or changed) as of a certain release.  I don't think that would work here.  A one-time popup during startup (or every time until you check a "don't show this message on startup" box) might work, and could be reusable by changing the message and reset the startup warning setting on each new release.  It could provide a means to notify the user of something to watch out for with the release.  If not required, the message could be empty or null, causing the popup to not be displayed.  Because this would not be a one-shot effort and the functionality could be used ***if required*** on any future release, I could support the implementation of such functionality.  Further, the "release notes" popup could be made available on request as an item in the "Help" menu.

My only comment regarding the notification about using recording-level information in naming scripts is that it may be perfectly valid (or even desired) if used in the file name portion (as opposed to the file path portion).  For that reason, if something like this warning is added (which I agree could be helpful to the users), it should note that it ***may*** result in scattered files.  The alternative would be to try to parse the script to determine if the information is used in the name portion or the path portion, which seems to me to be more work than its worth.  This way the test could be performed by examining the entire naming script when it is updated (saved?) to see if it contained a reference to specified tags or variables (provided as a list to the test?) checking for both forms of access: either `%variabletocheck%` or `$get(variabletocheck)`.

I think this has been a useful discussion, once we got past the positioning and personal preferences and started focusing on the issues,  Some interesting (and valuable) insights.

Sophist-UK:
> First off, I don't meant to try to diminish or dismiss your concerns regarding backward compatibility and possibly breaking current scripts.

> I think this has been a useful discussion, once we got past the positioning and personal preferences and started focusing on the issues, Some interesting (and valuable) insights.

Clearly we have not yet got past the personal comments. ;-)

> A one-time popup during startup (or every time until you check a "don't show this message on startup" box) might work, and could be reusable by changing the message and reset the startup warning setting on each new release. It could provide a means to notify the user of something to watch out for with the release.

Yes - a "confirm I have read about the release changes" box would be better than a one off pop-up. 

> My only comment regarding the notification about using recording-level information in naming scripts is that it may be perfectly valid (or even desired) if used in the file name portion (as opposed to the file path portion). 

It is easy to check the path part of the script (everything to the left of the last "/" - even if it is used in a `$if`).

Sophist-UK:
A quick question: has anyone actually tested the new original-date functionality against a reasonable range of pop, jazz, classical, OST, singles, EPs etc. to confirm that the vast majority of results are reasonable?

phw:
Ok, I was pretty busy and too overloaded the last days to comment on this again. But after thinking about this the last days I currently more for keeping the default as is and just make the release and recording original dates available as variables, not having a UI for now. I think this kind of is what Sophist's original thinking was. It looks like we successfully convinced each other of our point of views so successfully, that we have different opinions on this again :D

Ok, here is my reasoning:

**Keep the default as is**: My original idea of changing the default was that the default to the date of the recording was that I considered it the better data. This I now think is wrong for two reasons:
  1. Just conceptually it is up to your individual preferences and interpretation what you want to see there. Yes, the ID3 spec says for the TDOR frame "timestamp describing when the original recording of the audio was released", and the ASF metadata specs says for WM/OriginalReleaseTime "contains the date on which the content was originally released". But if you want to consider the first album release as the first release of a song this is fine.
  2. The first recording date has its limitations, and it is up to interpretation. For one it requires proper reuse of recordings. Also it is specifically about specific recordings, but as Sophist pointed out above other valid interpretations are first release date of any recording of that work, however you want it either by only the same artist or any

So not changing the default and keep it as is, but give the user the ability to make use of the now available first-release-date for recordings makes sense to me. It was also a concern raised in the comments in the forums.

**Provide scripting variables:** I think this is clear (and I also think this was never debated above): We have learned from the past that having different values available for scripting allows for a lot of flexibility. We should apply this pattern in more cases where currently options change how a tag gets populated, with the user having to choose which data will be available. Bob's artist variables plugin is a great example of how useful this is.

The discussions in the forums also indicates that people would like to have it as scripting variables. In any way it does no harm to those not using scripting.

**Do not provide options:** My thinking currently is as follows:

- Quoting Sophist: "Every option we create results in more complex code which needs to be supported forever more."
- The data quality issue raised in the forums: As I said above the quality of the recording date is probably not always great. Some had the concern that this might trigger users to do stupid edits. Making this less prominently available and more for advanced usage might make this less of a problem (I understand that Sophist argues the other way around for this).
- We can still introduce such an option at any later time, it is additional to the rest of the implementation. So should we find out that the values are fine to use, the stupid edits don't happen and users prefer an option we can do this.
- I find this especially problematic if it conflicts with scripting. E.g. you might have a script `$set(originaldate,%date%` (or anything similar, setting originaldate). Now it doesn't matter how you set this option, it has no effect. That might be confusing. It is of course true for other options as well, but I think the effect for originaldate is bigger.

Argument pro option: Some users said they don't do scripting. But instructing a user to use `$set(originaldate,%_recordingoriginaldate%)` in a script is now not that difficult.

**Change the variable names:** I think the choice of `%_recordingoriginaldate%` and `%_releaseoriginaldate%` was not so great. On the forums it was also mentioned that one would like to use the actual recording date of a recording for this, and I see a name conflict coming up if we try to add a variable for that also :D So my new suggestion is to stick with something similar as the web service does, which calls those fields `first-release-date`:

`%_recording_first_release_date%`, `%_release_group_first_release_date%`

I also would argue for using the underscores. My original suggestion was trying to follow our current naming, where we just write everything in one word: `%_primraryreleasetype`, `%_releasegorupcomment%`. Consistent with this it would be `%_recordingfirstreleasedate`, `%_releasegroupfirstreleasedate`, but that really becomes hard to read. We probably should start getting better named variables.

Sophist-UK:
I am fine with the except for one concern - _that_the_new_variables_are_now_very_long_names_indeed!!

How about `%_recording_firstreleasedate%` and `%_releasegroup_firstreleasedate%`?

phw:
Those names look better, I'll update the other PR.

rdswift:
> Those names look better, I'll update the other PR.

And I'll update the documentation changes.

phw:
Thanks for the feedback. Closing this for now in favor of https://github.com/metabrainz/picard/pull/1734 . Let's continue any discussion there.

rdswift:
@phw I think the ***Doc update*** flag can be removed from this PR, which has been closed and superseded by PR #1734.  Also, the documentation has been updated for the new PR.

PR #1734, opened 2021-01-06
PICARD-204: Support for recording-level original release date in scripting

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-204

This is an alternative implementation of https://github.com/metabrainz/picard/pull/1732

# Solution

Provide variables `%_recording_firstreleasedate%` and `%_releasegroup_firstreleasedate%` and set `originaldate` / `originalyear` for non-album tracks. For releases the tag variables for `originaldate` and `originalyear` will still be set from the release group's first release date, but the user can choose to use the recording date instead.

E.g. things like this are possible:

```
$set(originaldate,$if2(%_recording_firstreleasedate%,%_releasegroup_firstreleasedate%))
```

```
$set(originaldate,$if($inmulti(%releasetype%,compilation),%_recording_firstreleasedate%,%_releasegroup_firstreleasedate%))
```

# Action

### DIFF
--- a/test/test_mbjson.py
+++ b/test/test_mbjson.py
@@ -190,7 +190,7 @@ class RecordingTest(MBJSONTest):
         self.assertEqual(m['~artists_sort'], 'Sheeran, Ed')
         self.assertEqual(m['~length'], '4:41')
         self.assertEqual(m['~recordingtitle'], 'Thinking Out Loud')
-        self.assertEqual(m['~recordingoriginaldate'], '2014-06-20')
+        self.assertEqual(m['~recording_firstreleasedate'], '2014-06-20')
         self.assertNotIn('originaldate', m)
         self.assertNotIn('originalyear', m)
         self.assertEqual(t.genres, {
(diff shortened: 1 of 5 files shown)


phw:
See discussion at https://github.com/metabrainz/picard/pull/1732

test/test_mbjson.py
```diff
@@ -54,7 +54,8 @@
     "standardize_instruments": True,
     "release_ars": True,
     "preferred_release_countries": [],
-    "artist_locale": 'en'
+    "artist_locale": 'en',
+    "originaldate_use_recording": True
```
zas, line 58: Since Python allows it, we should keep a comma at the end of the last line, it makes diffs cleaner.

zas (changes requested):
Apart minor comment about comma, LGTM.

test/test_mbjson.py
phw: Actually this option was removed, so this line is not needed at all. I added the comma anyway as a small syntax change, we usually keep trailing commas in those cases. I would be even ok if we would test for this in our linting

PR #1735, opened 2021-01-13
PICARD-2092: non annoying script autocompletion

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2092

The script autocompleter can sometimes be annoying:

- It triggers on all kind of key presses, e.g. simply hitting Shift shows the autocompletion
- Accepting a choice only works with Tab. I personally always try to press return to accept the choice, causing a linebreak and the suggestion to abort
- It only shows fully matching suggestions

# Solution
- Only trigger on key presses that cause actual text input or removal.
- Allow explicitly opening the completion poput with Ctrl+Space at any time.
- Allow accepting choice with enter.-
- Use UnfilteredPopupCompletion completion mode ("All possible completions are displayed in a popup window with the most likely suggestion indicated as current.").
- Improved replacement of starting `(` for functions and ending `%` for variables. The `(` for functions is now automatically inserted if the inserted function ins not already followed by a `(` and the cursor placed afterwards.
- Extended variable completion by explicitly adding other known hidden variables.
- Use only the text up to the current cursor position for the suggestion. E.g. if you have a text like `$eq_all`  and you have placed the cursor directly after `$eq` the auto completion will suggest `$eq`, `$eq_all` and `$eq_any` as next choices.

# Action
@Sophist-UK Could you run this code and see if this makes the autocompletion not annoy you?

### DIFF
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -44,7 +44,7 @@ from picard.ui import FONT_FAMILY_MONOSPACE
 from picard.ui.theme import theme
 
 
-EXTRA_VARIABLES = [
+EXTRA_VARIABLES = (
     '~absolutetracknumber',
     '~albumartists_sort',
     '~albumartists',
@@ -70,7 +70,7 @@ EXTRA_VARIABLES = [
     '~silence',
     '~totalalbumtracks',
     '~video',
-]
+)
 
 
 class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
@@ -151,17 +151,21 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
 
 class ScriptCompleter(QCompleter):
     def __init__(self, parent=None):
-        choices = ['$' + name for name in script_function_names()]
-        choices += self.all_tags
-        super().__init__(choices, parent)
+        super().__init__(self.choices, parent)
         self.setCompletionMode(QCompleter.UnfilteredPopupCompletion)
         self.highlighted.connect(self.set_highlighted)
         self.last_selected = ''
 
     @property
+    def choices(self):
+        yield from {'$' + name for name in script_function_names()}
+        yield from {'%' + name.replace('~', '_') + '%' for name in self.all_tags}
+
+    @property
     def all_tags(self):
-        tags = list(TAG_NAMES.keys()) + list(PRESERVED_TAGS) + EXTRA_VARIABLES
-        return ['%' + name.replace('~', '_') + '%' for name in tags]
+        yield from TAG_NAMES.keys()
+        yield from PRESERVED_TAGS
+        yield from EXTRA_VARIABLES
 
     def set_highlighted(self, text):
         self.last_selected = text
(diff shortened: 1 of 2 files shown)


zas:
After some testing I still find it doesn't work that well.
For example, just pressing Enter open completion dialog, even if nothing was entered yet (no prefix).
And when deleting, it stays around.

phw:
I did a couple of tweaks. This should be now much more restricted and really only trigger on relevant input and closes otherwise. I have also added all built-in hidden variables to the completion list.

Regarding variable completion we could of course do much better, e.g. by parsing the scripts for set variables or allow plugins to specify extra variables they support. We could even try to automatically gather tags getting set by plugins or manually and offer them in the completion as well. But I think this would be it's own ticket.

zas (commented):
While testing completion changes, I noticed Qt errors:

```
QTextCursor::setPosition: Position '6' out of range
QTextCursor::setPosition: Position '28' out of range
```

`setPosition()` prolly tries to move outside document or the like.

phw:
Thanks for testing. This warning happened when inserting the completion at the very end of the script. Fixed with https://github.com/metabrainz/picard/pull/1735/commits/4189fce65fa4928eb10bacfb35fb283a43ae5048

picard/ui/widgets/scripttextedit.py
```diff
@@ -145,50 +186,87 @@ def enable_completer(self):
         self.popup_shown = False
 
     def insert_completion(self, completion):
+        if not completion:
+            return
         tc = self.cursor_select_word()
+        if completion.startswith('$'):
+            completion += '('
         tc.insertText(completion)
+        pos = tc.position()
```
zas, line 195: `pos` can be moved inside `if` branch below.

picard/ui/widgets/scripttextedit.py
```diff
@@ -145,50 +186,87 @@ def enable_completer(self):
...
+    def cursor_select_word(self, full_word=True):
         tc = self.textCursor()
         current_position = tc.position()
         tc.select(QTextCursor.WordUnderCursor)
+        if not full_word:
+            tc.setPosition(current_position, QTextCursor.KeepAnchor)
         selected_text = tc.selectedText()
+        # Check for start of function or end of variable
+        if selected_text and selected_text[0] in ('(', '%'):
+            current_position -= 1
```
zas, line 221: May be check `current_position` isn't 0 before? Useful?

zas (changes requested):
It works much before than before. Good job.

picard/ui/widgets/scripttextedit.py
```diff
@@ -117,13 +151,18 @@ def highlightBlock(self, text):
...
-        self.setCompletionMode(QCompleter.PopupCompletion)
+        self.setCompletionMode(QCompleter.UnfilteredPopupCompletion)
         self.highlighted.connect(self.set_highlighted)
         self.last_selected = ''
 
+    @property
+    def all_tags(self):
+        tags = list(TAG_NAMES.keys()) + list(PRESERVED_TAGS) + EXTRA_VARIABLES
+        return ['%' + name.replace('~', '_') + '%' for name in tags]
+
```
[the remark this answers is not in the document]
phw: I rewrote it a bit different.

picard/ui/widgets/scripttextedit.py
phw: I think it is at least better to check this. I could not reproduce a case for this, but we would rely on `tc.selectedText()` never returning something at position 0 which matches the condition. So better to check for > 0

zas (approved):
LGTM

PR #1738, opened 2021-01-15
PICARD-1998: Add "director" (for videos) tag

phw:
# Summary
This adds a tag "director" which is filled with the "video director" AR from MB by default.

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1998

# Solution

Add the `director` tag and fill it with the `video director` AR from MB.

Map the tag as follows:

 * MP4: ©dir (from iTunes Metadata Format Specification)
 * ASF: [WM/Director](https://docs.microsoft.com/en-us/previous-versions/ms867702(v=msdn.10)?redirectedfrom=MSDN#wmdirector)
 * ID3: TXXX/DIRECTOR
 * APE: Director
 * Vorbis: DIRECTOR

For ID3, APE and Vorbis this is probably not too relevant, as few people might use this for audio files only. But the mapping follows our usual naming. For MP4 and ASF there are explicit definitions in the respective specs. For Matroska there would be as well, but we don't support that yet.

# Action

Needs doc update to add the tag and tag mapping.

### DIFF
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -92,6 +92,7 @@ class APEv2File(File):
     __translate = {
         "albumartist": "Album Artist",
         "remixer": "MixArtist",
+        "director": "Director",
         "website": "Weblink",
         "discsubtitle": "DiscSubtitle",
         "bpm": "BPM",
--- a/picard/formats/asf.py
+++ b/picard/formats/asf.py
@@ -184,6 +184,7 @@ class ASFFile(File):
         'djmixer': 'WM/DJMixer',
         'mixer': 'WM/Mixer',
         'artists': 'WM/ARTISTS',
+        'director': 'WM/Director',
         'work': 'WM/Work',
         'website': 'WM/AuthorURL',
     }
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -193,6 +193,7 @@ class ID3File(File):
         'ASIN': 'asin',
         'MusicMagic Fingerprint': 'musicip_fingerprint',
         'ARTISTS': 'artists',
+        'DIRECTOR': 'director',
         'WORK': 'work',
         'Writer': 'writer',
         'SHOWMOVEMENT': 'showmovement',
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -80,6 +80,7 @@ class MP4File(File):
         "\xa9lyr": "lyrics",
         "\xa9cmt": "comment",
         "\xa9too": "encodedby",
+        "\xa9dir": "director",
         "cprt": "copyright",
         "soal": "albumsort",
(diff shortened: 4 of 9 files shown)


phw:
@zas @Sophist-UK I did a full mapping for this for all tagging formats, as I think this is more what we currently aim for. The alternative would be to map this only for ASF and MP4 and keep it as unsupported for other formats. Thoughts?

zas:
> @zas @Sophist-UK I did a full mapping for this for all tagging formats, as I think this is more what we currently aim for. The alternative would be to map this only for ASF and MP4 and keep it as unsupported for other formats. Thoughts?

I'm fine with full mapping.

Sophist-UK:
> > @zas @Sophist-UK I did a full mapping for this for all tagging formats, as I think this is more what we currently aim for. The alternative would be to map this only for ASF and MP4 and keep it as unsupported for other formats. Thoughts?
> 
> I'm fine with full mapping.

I agree - we should definitely strive to support all tags on all formats, and if there is no de-jure or de-facto standard, then we should try to create a de-facto standard in a way that is compatible with the underlying tagging standard i.e. use freeform comments wherever explicit tags are not defined.

phw:
Ok, thanks for the feedback. This case is probably a bit special as it is a tag only relevant for videos, so it is clear why this is not commonly used for ID3 or APE. But yes, at least in this case I think it is straight forward and maybe it is useful for the one or other use case.

Sophist-UK:
I think it is better to have superfluous tags (e.g. director for audio only tagging formats) than have inconsistencies in the tag coverage.

phw:
Ok, thanks for the feedback. This case is probably a bit special as it is a tag only relevant for videos, so it is clear why this is not commonly used for ID3 or APE. But yes, at least in this case I think it is straight forward and maybe it is useful for the one or other use case.

phw:
> I think it is better to have superfluous tags (e.g. director for audio only tagging formats) than have inconsistencies in the tag coverage.

I tend to agree, even though I am still a bit torn between positions here. The original philosophy, which Lukas followed when setting up the tags, was to only have tags mapped if there was at least some precedence of using this tag elsewhere. The advantage is that we don't make anything up and later discover that we need to change it because others in the future decide to use a different tag. On the other hand that leads to surprises for users who expect all tags to be saved. Also a result of this is e.g. that for the most part APEv2 tags where neglected, which lead to a pretty inconsistent way how tags are spelled there.

Currently I think we should aim for mapping tags to all formats if we think it makes sense, but there might be cases where we probably should not force it.

rdswift:
I believe that the ***Doc update*** tag can be removed because the documentation updates have been made in https://github.com/rdswift/picard-docs/pull/82.

PR #1739, opened 2021-01-18
PICARD-2102: After cluster action select the Clusters item

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
After clustering completed, if there is no selection or only Unclustered files is selected without any files, automatically select the Clusters item. This allows for easier workflow as the user can now directly use actions like Lookup on the new clusters.

# Problem

* JIRA ticket (_optional_): PICARD-2102

# Solution
Update the selection after clustering.

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2007 Robert Kaye
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Hendrik van Antwerpen
-# Copyright (C) 2008-2011, 2014-2015, 2018-2020 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014-2015, 2018-2021 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2009 Nikolai Prokoschenko
 # Copyright (C) 2011 Tim Blechmann
@@ -323,6 +323,14 @@ class MainPanel(QtWidgets.QSplitter):
             self._views[0].collapseAll()
         else:
             self._views[0].expandAll()
+
+    def select_object(self, obj):
+        item = obj.item
+        for view in self._views:
+            if view.indexFromItem(item).isValid():
+                view.setCurrentItem(item)
+                self._update_selection(view)
+                break
 
 
 def paint_fingerprint_icon(painter, rect, icon):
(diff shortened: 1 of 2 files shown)


phw:
@zas @Sophist-UK Could you test run this and tell me what you think about this behavior?

For Windows https://github.com/metabrainz/picard/suites/1851250518/artifacts/35901011 would be the portable build.

Personally I think it makes sense. The behaviour is helpful, but unobtrusive. E.g. if you just cluster and everything gets clustered you end up with the "Clusters" item selected and can quickly continue. But if clustering takes longer and you have selected something else in the meantime, maybe focusing on checking some tags on matched files or something, it does not steal your focus.

Sophist-UK:
Tested. It is better than existing but if I am being very fussy:

1. If nothing is selected and I click cluster, then the Clusters folder is selected even if there are no clusters in the folder to lookup.
2. If I have the Unclustered files folder or one or more unclustered files selected, then the Clusters folder is not selected even though perhaps it should be.

phw:
>     If nothing is selected and I click cluster, then the Clusters folder is selected even if there are no clusters in the folder to lookup.

We could try to handle this specific case where nothing was clustered, but I don't think it is worth the effort. The behavior does no harm and is consistent. It's also not that common. It only happens if you have nothing selected but there are unclustered files that cannot be clustered.

>     If I have the Unclustered files folder or one or more unclustered files selected, then the Clusters folder is not selected even though perhaps it should be.

This only happens if there are files left which cannot be clustered. If that's the case you still have an active selection with the left over files, and I think we should not forcefully remove existing selections. Doing so can be extremely annoying in an asynchronous application.

See also my comment on https://tickets.metabrainz.org/browse/PICARD-2102?focusedCommentId=57243&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-57243

Sophist-UK:
Let's approve this then.

phw:
Btw, related to the behavior on clustering when no or only a single selection is made is PICARD-212 and PICARD-1257

I think the last time @zas and I discussed this we agreed that it would make sense to allow single file clusters. At least if you explicitly create them. But if you have selected many files which it cannot cluster it likely should not end up creating many 1 file clusters, that also seems to be strange. Not sure we have fully thought this through, but current behavior is kind of unexpected / surprising.

zas (approved):
LGTM

PR #1740, opened 2021-01-19
PICARD-2105: Improve script function popup descriptions

Sophist-UK:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Improve descriptions by making some function definitions with optional arguments less technical, make functions with variable arguments more consistent and add a couple of examples to clarify.

# Problem
Some script functions do not have examples and description is technical in nature e.g. $delprefix says `$delprefix(text,*prefix="a","the")`. For the non-technical it suggests that you need to use a *prefix and put quotes around the text items, neither of which is true.

This PR just improves some of these descriptions.

* JIRA ticket: PICARD-2105

# Solution
Changes to description text strings only

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -870,10 +870,11 @@ def func_truncate(parser, text, length):
 
 
 @script_function(check_argcount=False, documentation=N_(
-    """`$swapprefix(text,prefix,...)`
+    """`$swapprefix(text,prefix1,prefix2,...)`
 
-Moves the specified `prefixes` from the beginning to the end of `text`.
-If no prefix is specified 'A' and 'The' are used by default.
+Moves the specified prefixes from the beginning to the end of `text`. Multiple
+prefixes can be specified as separate parameters. If no prefix is specified 'A'
+and 'The' are used by default.
 
 Example:
 
@@ -891,10 +892,11 @@ def func_swapprefix(parser, text, *prefixes):
 
 
 @script_function(check_argcount=False, documentation=N_(
-    """`$delprefix(text,prefix,...)`
+    """`$delprefix(text,prefix1,prefix2,...)`
 
-Deletes the specified `prefixes` from the beginning of `text`.
-If no prefix is specified 'A' and 'The' are used by default.
+Deletes the specified prefixes`from the beginning of `text`. Multiple
+prefixes can be specified as separate parameters.  If no prefix is specified 'A'
+and 'The' are used by default.
 
 Example:
 
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -191,7 +191,7 @@ def func_if(parser, _if, _then, _else=None):
 
 
 @script_function(eval_args=False, documentation=N_(
-    """`$if2(a1,a2,a3,...)`
+    """`$if2(a1,a2,a3...)`
 
 Returns first non empty argument."""
 ))
@@ -492,7 +492,7 @@ def func_trim(parser, text, char=None):
 
 
 @script_function(documentation=N_(
-    """`$add(x,y,*args)`
+    """`$add(x,y...)`
 
 Add `y` to `x`.
 Can be used with an arbitrary number of arguments.
@@ -510,7 +510,7 @@ def func_add(parser, x, y, *args):
 
 
 @script_function(documentation=N_(
-    """`$sub(x,y,*args)`
+    """`$sub(x,y...)`
 
 Subtracts `y` from `x`.
 Can be used with an arbitrary number of arguments.
@@ -528,7 +528,7 @@ def func_sub(parser, x, y, *args):
 
 
 @script_function(documentation=N_(
-    """`$div(x,y,*args)`
+    """`$div(x,y...)`
 
 Divides `x` by `y`.
 Can be used with an arbitrary number of arguments.
@@ -548,7 +548,7 @@ def func_div(parser, x, y, *args):
 
 
 @script_function(documentation=N_(
-    """`$mod(x,y,*args)`
+    """`$mod(x,y...)`
 
 Returns the remainder of `x` divided by `y`.
 Can be used with an arbitrary number of arguments.
@@ -566,7 +566,7 @@ def func_mod(parser, x, y, *args):
 
 
 @script_function(documentation=N_(
-    """`$mul(x,y,*args)`
+    """`$mul(x,y...)`
 
 Multiplies `x` by `y`.
 Can be used with an arbitrary number of arguments.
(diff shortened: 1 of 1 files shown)


phw:
Good improvement, especially getting rid of the *args and default value parameters.

But I must say I find `a1,a2,a3,...` (with the last comma) clearer to read.

phw (dismissed):
Right. At the very least this needs to be consistent. No strong opinion either way. I'm fine with this PR as it is, but if someone wants to have the trailing comma added to all the docs with variable parameters that's also ok for me.

picard/script/functions.py
```diff
@@ -492,7 +492,7 @@ def func_trim(parser, text, char=None):
 
 
 @script_function(documentation=N_(
-    """`$add(x,y,*args)`
+    """`$add(x,y...)`
```
zas, line 495: In those functions, two arguments are mandatory, so `$add(x,y,...)` would be more correct.

`$add(x,y,...)` but `if2(a1,...)`

picard/script/functions.py
```diff
@@ -870,14 +870,14 @@ def func_truncate(parser, text, length):
 
 
 @script_function(check_argcount=False, documentation=N_(
-    """`$swapprefix(text,*prefixes="a","the")`
+    """`$swapprefix(text,prefix...)`
 
 Moves the specified `prefixes` from the beginning to the end of `text`.
```
zas, line 875: Since function description is modified, text should be modified accordingly (`prefixes` vs `prefix`)

zas (changes requested):

picard/script/functions.py
Sophist-UK: If we say `$add(x,y,...)` a naive user might put `$add(x,y,)`.

`$if2` should probably be `$if2(a1,a2...)`

picard/script/functions.py
Sophist-UK: IMO no. I know this has dual use - user documentation and technical documentation - but the code is simple and `prefixes` is the correct variable name, and I did not want to make code changes in this PR.

But if the consensus is that we should do this, I can make this change.

picard/script/functions.py
phw: I'd prefer the trailing comma, it just looks more "correct" to me, at least that is how I am used to see such notations. But no really strong feelings about it, either way is an improvement.

picard/script/functions.py
phw: I see how `prefix` makes sense here. I would add a sentence like "Multiple prefixes can be specified." or such to make this even more clear.

phw (commented):
If we can get a consensus on the notation for repeatable parameters I would like to see this changed. I put my weak vote towards trailing comma.

rdswift:
> I put my weak vote towards trailing comma.

That would be my preference as well.

As a thought, for optional arguments such as the `prefix` argument discussed above, would it be clearer if they were shown enclosed within square brackets as we've done in the documentation?  For example, https://picard-docs.musicbrainz.org/v2.5/en/functions/func_swapprefix.html (which I should also change to use ***prefix*** rather than ***prefixes***).

Sophist-UK:
I suggest that you all chip in with changes to my PR and we then merge it.

picard/script/functions.py
Sophist-UK: Or put `$swapprefix(text,prefix1,...)`

picard/script/functions.py
Sophist-UK: Actually I sort of agree - it looks better when it is `$add(x,y...)` - I just worry that non-technical folks who are not used to programming will end up including a trailing comma. I suspect that `$mul(2,3)` and `$mul(2,3,)` may have different results as currently coded - perhaps we should correct this so that empty arguments are ignored and not treated as

Sophist-UK: Just tested:
```
$set(_test1,a$mul(2,3))
$set(_test2,a$mul(2,3,))
_test1 = a6
_test2 = a
```
I think that mathematical functions should:
1. Ignore (trailing?) empty arguments; and
2. Return `0` rather than nothing.

phw:
> I suggest that you all chip in with changes to my PR and we then merge it.

Ok, so far everyone here said they prefer the trailing comma, so let's change this. I don't see more potential in misunderstanding `$foo(a,b,...)` then `$foo(a,b...)`. Actually if I wouldn't know I likely would be more confused by the latter, unsure how to enter multiple.

Also a small sentence to say that multiple prefixes are allowed for the `prefix` case.

> I think that mathematical functions should:
> 
>     1. Ignore (trailing?) empty arguments; and
>     2. Return `0` rather than nothing.

I think I'm fine with 1), but would be careful with 2). Currently it will return empty if there is no defined result. I think returning empty for an undefined expression such as `$mul(2,a)` is better than returning 0. `$mul(2,0)` should return 0, as this is the actual result. Also changing the return value for the undefined case might break stuff.

In anyway, that would be a separate PR.

Sophist-UK:
> > I suggest that you all chip in with changes to my PR and we then merge it.
> 
> Ok, so far everyone here said they prefer the trailing comma, so let's change this. I don't see more potential in misunderstanding `$foo(a,b,...)` then `$foo(a,b...)`. Actually if I wouldn't know I likely would be more confused by the latter, unsure how to enter multiple.
 
I don't have time to redo this at present - higher priorities. So if anyone else wants to make the change to my PR or by submitting an alternative, then please go ahead.

> Also a small sentence to say that multiple prefixes are allowed for the `prefix` case.
> 
> > I think that mathematical functions should:
> > ```
> > 1. Ignore (trailing?) empty arguments; and
> > 2. Return `0` rather than nothing.
> > ```
> 
> I think I'm fine with 1), but would be careful with 2). Currently it will return empty if there is no defined result. I think returning empty for an undefined expression such as `$mul(2,a)` is better than returning 0. `$mul(2,0)` should return 0, as this is the actual result. Also changing the return value for the undefined case might break stuff.
> 
> In anyway, that would be a separate PR.

Yes - a separate PR, and I would agree that undefined expressions should return the empty string rather than "0". So `$mul(2,0)` would return "0" and `$mul(2,a)` or `$div(1,0)` would return "". This also enables you to see the difference between a zero result and an error result.

zas:
> I don't have time to redo this at present - higher priorities. So if anyone else wants to make the change to my PR or by submitting an alternative, then please go ahead.

Those changes aren't that urgent anyway, this PR will wait. Finish it when you'll be available.

phw:
@Sophist-UK @zas I did the above discussed changes of adding the trailing comma and tried to clarify the documentation for ` $swapprefix` and ` $delprefix`. Could you both take a final look? Then we could merge this and get this into 2.6 beta 1, which would give some time to adapt translations for 2.6 final.

zas (approved):
LGTM

PR #1742, opened 2021-01-21
Patch 1

Tche333:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Add a debug log when receiving a response from WebService which is useful in case of AcousticId submission, specifically when the submissions are blocked in pending state on server in order to track the status.

# Problem

The AcouticID doesn't seem to be able to process new submission. From Picard point of view, submissions were successfully submitted. However, submissions might be blocked in 'pending' status.

# Solution

This debug log statement allows to track those submissions status.

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -492,6 +492,7 @@ class WebService(QtCore.QObject):
                 elif request.response_parser:
                     try:
                         document = request.response_parser(reply)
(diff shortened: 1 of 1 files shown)


phw (approved):
Makes sense, yes. Thanks!

PR #1743, opened 2021-02-06
PICARD-2109: Make CaaReleaseGroup enabled by default

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Most casual users want to tag their releases with (usually the highest res digital square image).
This is best fulfilled by the CaaReleaseGroup cover art provider, so make it the default first option.

# Problem

* JIRA ticket (_optional_): PICARD-2109

# Solution

Change default config for new installs.

### DIFF
--- a/picard/ui/options/cover.py
+++ b/picard/ui/options/cover.py
@@ -58,9 +58,9 @@ class CoverOptionsPage(OptionsPage):
         TextOption("setting", "cover_image_filename", "cover"),
         BoolOption("setting", "save_images_overwrite", False),
         ListOption("setting", "ca_providers", [
+            ('CaaReleaseGroup', True),
             ('Cover Art Archive', True),
             ('UrlRelationships', True),
-            ('CaaReleaseGroup', False),
             ('Local', False),
         ]),
     ]


phw:
I very much agree with the argumentation Simon provided on PICARD-2109

Sophist-UK:
Whilst I would personally prefer the CA for the Special Deluxe Edition Release rather than the generic RG CA, I do agree that this is probably the best default to use as a starting point for new users.

phw:
@Sophist-UK Yes, exactly my thinking. Easy to be disappointed by "bad quality" CA images if what you expect is a digitally cleaned up square image. On the other hand those of us who deeply care about the correct image matching the physical release spent more time configuring this anyway.

picard/ui/options/cover.py
```diff
@@ -58,9 +58,9 @@ class CoverOptionsPage(OptionsPage):
         TextOption("setting", "cover_image_filename", "cover"),
         BoolOption("setting", "save_images_overwrite", False),
         ListOption("setting", "ca_providers", [
+            ('CaaReleaseGroup', True),
```
zas, line 61: Instead I would just set it to `True` but keep its position. This way cover art specific to the release will be used if it exists else release group's one will be used.

If release group is moved to top, users setting a specific release might be surprised to not have matching cover art...

Sophist-UK: > Instead I would just set it to `True` but keep its position. This way cover art specific to the release will be used if it exists else release group's one will be used.
> 
> If release group is moved to top, users setting a specific release might be surprised to not have matching cover art...

I am not sure why it is currently not second - but rather than set it to true and keep it where it, if you are going to leave the Release CAA at the top then I think RG CAA should come immediately after it.

zas: This cover art source was initially a hack to get an image when all other sources failed (typically for a release having no cover art on CAA, but one for the release group).
It wasn't meant to become the main source.

zas (approved):
Initially not convinced, but I guess new users will find what they expect.

Sophist-UK:
Perhaps we can cover this with some additional documentation?

phw:
> This cover art source was initially a hack to get an image when all other sources failed (typically for a release having no cover art on CAA, but one for the release group).
> It wasn't meant to become the main source.

It always was a source that was commonly recommended for users who want to have a good looking square image, even if it does not represent the exact edition they own.

I am not 100% sure about this either, it really depends on what we want to provide as default. I know discussions like this happen all the time, and some users naivly try to change the cover art for a release with a generic one because the original one is not "good" looking. If this change will solve this issue I can of course not say. But this change depends on what we expect the common user to expect by default: Do they look for the cover most accurately representing their physical copy, with all its flaws, or do they want best visual fidelity for their media center?

I kind of agree with the bug reporter that the casual user probably is more after the latter, while the MB enthusiast will configure Picard's cover art anyway. But actually if we are in doubt, and according to the points made above we seem to are, it's probably better to not change it.

> I am not sure why it is currently not second - but rather than set it to true and keep it where it, if you are going to leave the Release CAA at the top then I think RG CAA should come immediately after it.

It makes sense like this if you consider what the cover art sources do:

- CAA: This is the offical and recommended way to server cover art, it returns cover art for the specific release
- Allowed Cover Art URLs: This is deprecated and for the most part irrelevant (there might be a few leftovers in the database), but again it delivers cover art specific to a release.
- CAA Release Group: Gives a generic image for the "album"

So if the goal is to return a cover specific for the release this order makes sense.

> Perhaps we can cover this with some additional documentation?

Additional documentation IMHO does not solve this, and https://picard-docs.musicbrainz.org/en/config/options_cover_art_providers.html already explains the different sources quite well.

zas:
@phw I think we could enable CAA release group by default (set it to True), but keep the current order so it's still a fallback: if a cover art specific to the release exists, it will be used, if not the one from the release group will be used. What do you think?

phw:
Closing this in favor of the weaker #1751

I also added this as a separate ticket PICARD-2125 and will leave PICARD-2109 open for future discussion.

PR #1747, opened 2021-02-15
PICARD-2088: Override QSettings config file syncing

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Syncing the config file can trigger a deadlock between QSettings internal mutex and
the Python GIL in PyQt up to 5.15.2. Workaround this by handling this ourselves
with custom file locking.

# Problem

* JIRA ticket (_optional_): PICARD-2088

# Solution
As the default implementation for synchronizing the config file can lead to application freezes override it.
Ensure the config file is not updated by more than one process by locking the file using fasteners.

This patch disables file locking of QSettings, intercepts the UpdateRequest event and does implement its own file locking when actually syncing the settings file.

# Action
As side effect is that this raises the requirements to Qt >= 5.10, as [`QSettings.setAtomicSyncRequired`](https://doc.qt.io/qt-5/qsettings.html#setAtomicSyncRequired) is not available in earlier versions.

I thought about making this conditional, but I think we shouldn't do that. If we don't disable the file locking in QSettings this workaround is non-functional and running with earlier versions of Qt will suffer from this bug. It is not safe to claim compatibility in this case. But it would break compatibility with e.g. Ubuntu 18.04 Bionic.

The alternative would be to just not apply the workaround for earlier versions, accepting that this bug exists but is for most users not triggered that frequently. We don't really have numbers here- We got regular reports of users encountering this. Some only occasional, some more frequently. But for most it seems to work, otherwise there would have been a bigger outcry.


phw:
From all what I can tell theoretically and practically this should indeed fix the issue.

On the ticket this was also tested by two others who could before reproduce the issue, and for me locally for the cases where I could reproduce the issue before I can no longer do so.

zas (approved):
Thanks for the work you did on those issues.
This workaround looks the only way to solve it on our side, and it isn't too intrusive.
Bumping PyQt version and the new dependency over fasteners don't look a problem to me.

LGTM.

phw:
Just to have this mentioned: Before trying fasteners I tried using QFileLock (just to avoid an additional dependency). But actually this very reliable caused a deadlock, which might be the actual core issue we are dealing with here. Hence using fasteners as a proper Python alternative

PR #1748, opened 2021-02-16
Build with PyInstaller 4.2

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

Build with latest PyInstaller 4.2

Apart from various bugfixes those two changes are interesting for us:

- Fixes with invalid code signatures with newer Python versions on macOS (but we already work around this by forcefully codesigning the binaries with our own signature at 293c2c7)
- On Windows the app manifest gets set the longPathAware attribute. This is a part of the puzzle to fully support long path names on Windows 10, see PICARD-2076 

See https://pyinstaller.readthedocs.io/en/stable/CHANGES.html

### DIFF
--- a/requirements-build.txt
+++ b/requirements-build.txt
@@ -1,3 +1,3 @@
 Babel==2.6
-PyInstaller==4.1
+PyInstaller==4.2
 setuptools<45.0.0


phw:
Tested on Windows 7, Windows 10, macOS 10.12 and macOS 11.0

PR #1749, opened 2021-02-17
PICARD-1528: Fix crash on "search similar albums"

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

In case both clusters and other items where selected this could fail. The assumption that the first selected object is always a cluster was wrong. 

# Problem

* JIRA ticket (_optional_): PICARD-1528

# Solution
Get the first actual cluster from the selected objects

# Action


zas (approved):
LGTM

PR #1752, opened 2021-02-25
PICARD-2126: Send Access-Control-Allow-Origin in browser integration

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2126

Trying to access the browser integration via JavaScript on musicbrainz.org gives a cross origin error:

Access to XMLHttpRequest at 'http://127.0.0.1:8000/' from origin 'https://musicbrainz.org' has been blocked by CORS policy: No 'Access-Control-Allow-Origin' header is present on the requested resource.

This can be resolved by the browser integration by setting a Access-Control-Allow-Origin response header.

While the current integration does not require this (the browser integration is just a link), having this supported would allow more flexibility for the future. E.g. the website could check if Picard is available, as I have [prototyped with a user script](https://community.metabrainz.org/t/bring-me-a-green-tagger-button/478155/16).

Or there could be actual feedback where clicking on the tagger button does a scripted request and depending on the response shows a notification to the user.

# Solution

Check for a `Origin` request header and use its value for setting a `Access-Control-Allow-Origin` response header. The origin hostname is validated, it must be either for `*.musicbrainz.org` or must match the configured MB server hostname in options.

This was tested using MusicBrainz Auto Tagger Button user script in Microsoft Edge.

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -5,7 +5,7 @@
 # Copyright (C) 2006-2007, 2011 Lukáš Lalinský
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
-# Copyright (C) 2012-2013, 2018 Philipp Wolfer
+# Copyright (C) 2012-2013, 2018, 2021 Philipp Wolfer
 # Copyright (C) 2013, 2018 Laurent Monin
 # Copyright (C) 2016 Suhas
 # Copyright (C) 2016-2017 Sambhav Kothari
@@ -26,6 +26,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
+import re
 from urllib.parse import (
     parse_qs,
     urlparse,
@@ -38,18 +39,52 @@ from picard.config import get_config
 from picard.util import mbid_validate
 
 
-def response(code):
+def response(code, origin=None):
     if code == 200:
         resp = '200 OK'
     elif code == 400:
         resp = '400 Bad Request'
     else:
         resp = '500 Internal Server Error'
+    if origin:
+        allowOrigin = 'Access-Control-Allow-Origin: {}\r\n'.format(origin)
+    else:
+        allowOrigin = ''
     return bytearray(
         'HTTP/1.1 {}\r\n'
         'Cache-Control: max-age=0\r\n'
+        '{}'
         '\r\n'
-        'Nothing to see here.\r\n'.format(resp), 'ascii')
+        'Nothing to see here.\r\n'.format(resp, allowOrigin), 'ascii')
+
+
+def _read_origin(conn):
+    while True:
+        line = conn.readLine().data()
+        if not line or line == b'\r\n':
+            break
+        if line.startswith(b'Origin:'):
+            header, origin = line.decode().split()
+            log.debug('Browser sent origin %s', origin)
+            if _is_valid_origin(origin):
+                return origin
+            else:
+                return None
+    return None
+
+
+RE_VALID_ORIGINS = re.compile(r'(.*\.)?musicbrainz.org')
+
+
+def _is_valid_origin(origin):
+    url = urlparse(origin)
+    hostname = url.hostname
+    if not hostname:
+        return False
+    if RE_VALID_ORIGINS.match(hostname):
+        return True
+    config = get_config()
+    return config.setting['server_host'] == hostname
 
 
 class BrowserIntegration(QtNetwork.QTcpServer):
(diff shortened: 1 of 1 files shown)


picard/browser/browser.py
```diff
@@ -38,18 +39,52 @@
...
+            header, origin = line.decode().split()
+            log.debug('Browser sent origin %s', origin)
+            if _is_valid_origin(origin):
+                return origin
+            else:
+                return None
+    return None
+
+
+RE_VALID_ORIGINS = re.compile(r'(.*\.)?musicbrainz.org')
```
zas, line 76:
```suggestion
RE_VALID_ORIGINS = re.compile(r'^(?:[^\.]+\.)*musicbrainz\.org$')
```

This regex is incorrect (when used with `re.match()`):

```
>>> RE_VALID_ORIGINS = re.compile(r'(.*\.)?musicbrainz.org')
>>> RE_VALID_ORIGINS.match('a.b.musicbrainzzorg')
<_sre.SRE_Match object; span=(0, 19), match='a.b.musicbrainzzorg'>
>>> RE_VALID_ORIGINS.match('a.b.musicbrainz.org')
<_sre.SRE_Match object; span=(0, 19), match='a.b.musicbrainz.org'>
>>> RE_VALID_ORIGINS.match('not.musicbrainzborg.com')
<_sre.SRE_Match object; span=(0, 19), match='not.musicbrainzborg'>
>>> RE_VALID_ORIGINS.match('....musicbrainz.org')
<_sre.SRE_Match object; span=(0, 19), match='....musicbrainz.org'>
>>> RE_VALID_ORIGINS.match('musicbrainz.org')
<_sre.SRE_Match object; span=(0, 15), match='musicbrainz.org'>
```

A safer regexp would be: 

```
>>> RE_VALID_ORIGINS = re.compile(r'^(?:[^\.]+\.)*musicbrainz\.org$')
>>> RE_VALID_ORIGINS.match('a.b.musicbrainzzorg')
>>> RE_VALID_ORIGINS.match('a.b.musicbrainz.org')
<_sre.SRE_Match object; span=(0, 19), match='a.b.musicbrainz.org'>
>>> RE_VALID_ORIGINS.match('not.musicbrainzborg.com')
>>> RE_VALID_ORIGINS.match('....musicbrainz.org')
>>> RE_VALID_ORIGINS.match('musicbrainz.org')
<_sre.SRE_Match object; span=(0, 15), match='musicbrainz.org'>
```

It allows subdomains, escape last dot, matches start & end of the string, disallows string with dot at start or multiple dots. Also there's no need to capture first part (`(reg) -> (?:reg)`)

picard/browser/browser.py
```diff
@@ -38,18 +39,52 @@
...
+
+
+def _is_valid_origin(origin):
+    url = urlparse(origin)
```
zas, line 80: `urlparse()` may raise `ValueError` in some cases:

```>>> url = urlparse('http://a.b.c[/path')
  File "/usr/lib/python3.6/urllib/parse.py", line 368, in urlparse
    splitresult = urlsplit(url, scheme, allow_fragments)
  File "/usr/lib/python3.6/urllib/parse.py", line 436, in urlsplit
    raise ValueError("Invalid IPv6 URL")
ValueError: Invalid IPv6 URL
```

To be safe, we should prolly catch it and return `False`.

picard/browser/browser.py
```diff
@@ -38,18 +39,52 @@
...
+    hostname = url.hostname
```
zas, line 81: Since we don't use `url`:

```python
try:
    hostname = urlparse(origin).hostname
except ValueError:
    return False
```

zas (changes requested):

picard/browser/browser.py
```diff
@@ -38,18 +39,52 @@
...
-        'Nothing to see here.\r\n'.format(resp), 'ascii')
+        'Nothing to see here.\r\n'.format(resp, allowOrigin), 'ascii')
+
+
+def _read_origin(conn):
+    while True:
+        line = conn.readLine().data()
+        if not line or line == b'\r\n':
+            break
+        if line.startswith(b'Origin:'):
```
zas, line 66: Since that's an http header, the match is too strict.

https://tools.ietf.org/html/rfc7230#section-3.2

> Each header field consists of a case-insensitive field name followed
>    by a colon (":"), optional leading whitespace, the field value, and
>    optional trailing whitespace.

Perhaps we should use some already existing functions for that, perhaps https://docs.python.org/3/library/http.client.html#http.client.parse_headers (dunno if appropriate in this context)

zas (changes requested):

phw:
Closing this in favor of #1754

PR #1754, opened 2021-02-26
Refactor browser integration to use http.server and implement CORS support

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other

Refactor browser integration to use `http.server` instead of the custom HTTP implementation and implement CORS support to fix PICARD-715 and PICARD-2126

The integration has been tested with Firefox, Google Chrome and MS Edge. It has also been tested with [MB auto tagger user script](https://github.com/phw/musicbrainz-auto-tagger-button).

# Problem

* JIRA ticket (_optional_): PICARD-2126, PICARD-715
* #1752

For future proof browser integration Picard should handle CORS requests. The current custom HTTP implementation makes this harder to do and maintain as we start to implement more HTTP features.

# Solution

Replace the custom HTTP implementation with `http.server` and add CORS handling on top of it.

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -116,10 +116,13 @@ class RequestHandler(BaseHTTPRequestHandler):
     def _handle_get(self):
         parsed = urlparse(self.path)
         args = parse_qs(parsed.query)
+        action = parsed.path
 
-        if 'id' in args and args['id']:
+        if action == '/':
+            self._response(200, SERVER_VERSION)
+        elif 'id' in args and args['id']:
             mbid = args['id'][0]
-            if self._load_mbid(parsed.path, mbid):
+            if self._load_mbid(action, mbid):
                 self._response(200, _('MBID "%s" loaded') % mbid)
             else:
                 self._response(400, _('Could not load MBID "%s"') % mbid)
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -123,6 +123,21 @@ class BrowserIntegration(QtCore.QObject):
 
 class RequestHandler(BaseHTTPRequestHandler):
 
+    def do_OPTIONS(self):
+        origin = self.headers['origin']
+        if _is_valid_origin(origin):
+            self.send_response(204)
+            self.send_header('Access-Control-Allow-Origin', origin)
+            self.send_header('Access-Control-Allow-Methods', 'GET')
+            self.send_header('Access-Control-Allow-Credentials', 'false')
+            self.send_header('Access-Control-Allow-Private-Network', 'true')
+            self.send_header('Access-Control-Max-Age', 3600)
+            self.send_header('Vary', 'Origin')
+            self.end_headers()
+        else:
+            self.send_response(401)
+            self.end_headers()
+
     def do_GET(self):
         try:
             self._handle_get()
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -5,7 +5,7 @@
 # Copyright (C) 2006-2007, 2011 Lukáš Lalinský
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
-# Copyright (C) 2012-2013, 2018 Philipp Wolfer
+# Copyright (C) 2012-2013, 2018, 2021 Philipp Wolfer
 # Copyright (C) 2013, 2018 Laurent Monin
 # Copyright (C) 2016 Suhas
 # Copyright (C) 2016-2017 Sambhav Kothari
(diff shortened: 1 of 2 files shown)


picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+            self.end_headers()
 
+    def do_GET(self):
         try:
-            line = rawline.decode()
-            if parse_line(line):
-                conn.write(response(200))
+            self._handle_get()
+        except Exception as e:
+            self._response(500, str(e))
```
zas, line 145: While it is likely safer to catch exceptions here, I think we shouldn't return exception message in response, but rather log it.

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+
+    def _handle_get(self):
+        parsed = urlparse(self.path)
```
zas, line 148: Do we need to reparse the path?

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
-            conn.write(response(500))
-            log.error(e)
-            return
-        finally:
-            conn.disconnectFromHost()
-
-    def _accept_connection(self):
-        conn = self.nextPendingConnection()
-        conn.readyRead.connect(self._process_request)
+                self._response(400, _('Could not load MBID "%s"') % mbid)
```
zas, line 159: I would avoid returning a non-validated string here, just something like "id isn't a valid mbid".
Also it makes no sense to translate those strings imho (client may use a different language).

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+        else:
+            self._response(400, _('Missing parameter "id".'))
+
+    @staticmethod
+    def _load_mbid(action, mbid):
+        if not mbid_validate(mbid):
+            log.error("Browser integration failed: bad mbid %r", mbid)
```
zas, line 166: Logging input from client is usually a bad idea, especially since it doesn't validate.

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+        if self.server:
             self.stop()
 
         config = get_config()
         if config.setting["browser_integration_localhost_only"]:
-            self.host_address = QtNetwork.QHostAddress(QtNetwork.QHostAddress.LocalHost)
+            host_address = '127.0.0.1'
         else:
-            self.host_address = QtNetwork.QHostAddress(QtNetwork.QHostAddress.Any)
+            host_address = '0.0.0.0'  # nosec
```
zas, line 98: It makes me think we could prolly add an option to specify listening address instead of listening on any. Something like if `browser_integration_localhost_only` is False, use `browser_integration_ip_address`, which defaults to `0.0.0.0`.
It would make sense for machines connected to multiple local networks.

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+            try:
+                self.server = HTTPServer((host_address, port), RequestHandler)
+            except OSError:
+                continue
+            log.info("Starting the browser integration (%s:%d)", host_address, port)
+            self.listen_port_changed.emit(port)
+            threading.Thread(target=self.server.serve_forever).start()
+            break
+        else:
+            log.error("Failed finding an available port for the browser integration.")
```
zas, line 110: Perhaps add tried range in this message, and while at it, store 65535 in a variable:

```python
min_port = config.setting["browser_integration_port"]
max_port = 65535
for port in range(min_port, max_port):
    ...
else:
    log.error("Failed finding an available port in range %d:%d for the browser integration.", min_port, max_port)
```

We don't need to check `min_port` <= `max_port`, right?

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
-            return
-        finally:
-            conn.disconnectFromHost()
-
-    def _accept_connection(self):
-        conn = self.nextPendingConnection()
-        conn.readyRead.connect(self._process_request)
+                self._response(400, _('Could not load MBID "%s"') % mbid)
+        else:
+            self._response(400, _('Missing parameter "id".'))
```
zas, line 161: Same here, no need to translate response messages.

zas (changes requested):

picard/browser/browser.py
phw: No. Codewise it would just be an empty list and run into the else. Practically this is avoided as we restrict the values ` browser_integration_port` can be set to.

picard/browser/browser.py
phw: Yes, good idea. But for sure rather an exception for the majority of users (but so is the use case of enabling external networks at all)

picard/browser/browser.py
phw: Yes, to get the query string. The path here is the full path in the HTTP header, e.g. `/openalbum?id=...`. It's the same as in our old implementation, only that we don't need the custom code to get to the path.

picard/browser/browser.py
phw: Fixed it. And yes, handling the exceptions here is kind of mandatory. The behavior without it is really not nice, the request would just not receive any answer

phw (commented):
@zas Good points, I reworked the response content and logging parts

picard/browser/browser.py
```diff
@@ -25,115 +25,169 @@
...
+            self._response(500, 'Unexpected request error')
+
+    def _handle_get(self):
+        parsed = urlparse(self.path)
+        args = parse_qs(parsed.query)
+        action = parsed.path
+
+        if action == '/':
+            self._response(200, SERVER_VERSION)
+        elif 'id' in args and args['id']:
```
zas, line 154: We should perhaps invert the logic here, checking valid actions first, and call `_load_mbid()` with the method to use, something like:

```python
        if action == '/':
            self._response(200, SERVER_VERSION)
        elif action == '/openalbum':
            self._load_mbid(tagger.load_album, args)
        elif action == ''/opennat':
             self._load_mbid(tagger.nat_album, args)
        else:
             self._response(400, 'Unknown action')
```

And do the parsing of `id` in `_load_mbid()`:

```python
def _load_mbid(self, loading_func, args):
    if 'id' in args and args['id']:
       mbid = args['id'][0]
       if not mbid_validate(mbid):
           self._response(400, '"id" is not a valid MBID.')
       else:
           tagger = QtCore.QCoreApplication.instance()
           def load_it(loader):
               tagger.bring_tagger_front()
               loader(mbid)
           to_main(load_it, loading_func)
           self._response(200, 'MBID "%s" loaded' % mbid)
    else:
        self._response(400, 'Missing parameter "id".')
```
 (indentation may be incorrect)

zas (changes requested):

zas (approved):

PR #1757, opened 2021-03-03
PICARD-2130: Refactor cover art options

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2130

The cover art options are currently not quite clear and can be confusing:

- The option "Use the first image type as the filename." is placed on CAA options, but applies globally
- Same for the "Save only one front image as separate file"
- It is not clear how images not affected by "Use the first image type as the filename." are saved, as the image name configuration is done elsewhere
- Some cover options apply to saving images as files, some for embedding images, but it is not quite clear what does what
- Inconsistent wording between "Save only one front image as separate file" and "Only embed a front image", even though both are conceptually very similar (one for saving separate files, one for embedding)
- Add the placeholder text "cover" to the `cover_image_filename` input. If this input is empty Picard will fall back to use just "cover" as filename, and having this placeholder text helps

# Solution

- Rename the option to `caa_image_type_as_filename` to `image_type_as_filename` and move it to the global Cover options
- Rename the option to `caa_save_single_front_image` to `save_only_one_front_image` and move it to the global Cover options. This is consistent with the similar `embed_only_one_front_image` option.
- Separate Cover options into those for embedding and those for saving as files. This makes it clear which option actually applies. Also unavailable options are inactive.
- If `save_only_one_front_image` is enabled deactivate the `image_type_as_filename` checkbox, as it no longer applies (no non-front images will be saved)
- Consistent wording: "Embed only a single front image" and "Save only a single front image as separate file"
- Add the placeholder text "cover" to the `cover_image_filename` input. If this input is empty Picard will fall back to use just "cover" as filename, and having this placeholder text helps

grafik


phw:
And of course the "Cover Art Archive" option page is now much simpler and really about options related to CAA downloading:

grafik

zas (approved):
LGTM

PR #1758, opened 2021-03-03
PICARD-2131: Workaround for slow reaction of tagger button in Firefox

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2131

On Firefox the tagger button request blocks for > 4 seconds before succeeding since recent refactoring of the browser listener code. Closer investigation reveals that before the actual request there is kind of a ghost request received on the socket, but which blocks on reading. The exact cause is not clear to me, it does not happen on other browsers or when invoking the Picard URL in any other way from within Firefox.

# Solution
Apply a timeout to the request socket. That makes the initial ghost request fail quickly and not block the actual processing.

An alternative approach that also worked for me:

```
def handle(self):
    self.request.setblocking(False)
    super().handle()
```

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -153,6 +153,12 @@ class RequestHandler(BaseHTTPRequestHandler):
             log.error('Browser integration failed handling request', exc_info=True)
             self._response(500, 'Unexpected request error')
 
+    def log_error(self, format, *args):
+        log.error(format, *args)
+
+    def log_message(self, format, *args):
+        log.info(format, *args)
+
     def _handle_get(self):
         parsed = urlparse(self.path)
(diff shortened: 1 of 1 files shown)


phw:
@zas This worked really well. I actually left only your code changes in this commit. In my tests the reduced timeout was not necessary. Even when trying to trigger a lot of requests I could not get it to have any bad effects. Adding the timeout would cause some errors to get logged that might be confusing.

But I also integrated the http server with our normal logging.

zas (approved):

PR #1761, opened 2021-03-12
PICARD-2095: UI theme selection

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
Let users change the color theme used by the UI.

# Problem

Some people say they want a dark theme, others want to keep the light theme even though they use the system dark theme, or vice-versa.
* JIRA ticket (_optional_): PICARD-2095, PICARD-1357

# Solution

Dropdown menu replaces the hidden checkbox and lets user choose either to keep the theme in sync with the system (current behavior) or force either the light or the dark theme.

# Action


Gabrielcarvfer:
Just to keep track of progress. Working on:
- [x] Windows
- [ ] MacOS
- [ ] Linux
- [ ] HaikuOS

Gabrielcarvfer:
Changing stuff on the MacOS is pretty hard. If you initialize the tagger, you can't change the style and break any further change made.

<details>
<summary>
Dark theme + system light theme. Here forcing the dark theme before initializing the Tagger.
</summary>

image

</details>

<details>
<summary>
Dark theme + system light theme. Here without forcing the dark theme during Tagger initialization.
</summary>

image

</details>

Gabrielcarvfer:
Just noticed I've screwed up by writing stuff to the ui_options_interface.py (which is automatically generated)...

picard/ui/options/interface.py
```diff
@@ -169,6 +168,12 @@ def __init__(self, parent=None):
         super().__init__(parent)
         self.ui = Ui_InterfaceOptionsPage()
         self.ui.setupUi(self)
+
+        self.ui.ui_theme.clear()
+        for theme in ["System", "Light", "Dark"] if IS_WIN or IS_MACOS else ["System"]:
+            self.ui.ui_theme.addItem(theme)
```
phw, line 174: This currently breaks the theme options on Linux. With this change the only option on Linux is using the Qt theme configured on the system, something we explicitly did not want to have. Instead the default should be that the "Fusion" theme is used and the user can opt-in to use the system theme. For that I think it would make sense to rename the current "System" to "Default" and have a separate "System theme" option for Linux.

picard/ui/theme.py
```diff
@@ -159,18 +167,29 @@ def update_palette(self, palette, dark_theme, accent_color):
...
-                return False
+                return dark_theme
+            if self._loaded_config_theme is not None and self._loaded_config_theme != "System":
+                dark_theme = self._loaded_config_theme == "Dark"
+                # MacOS uses an NSAppearance object to change the current application appearance
+                if dark_theme:
+                    appearance = AppKit.NSAppearance._darkAquaAppearance()
+                else:
+                    appearance = AppKit.NSAppearance._aquaAppearance()
+                AppKit.NSApplication.sharedApplication().setAppearance_(appearance)
```
phw, line 180: I have not yet tested on macOS, but the most important thing to check here is to verify whether this works with the packaged app or not. Because as it is currently the dark theme selection does not work for the app package.

phw (commented):
Thanks for working on this. As I wrote in the comments this currently breaks things a bit on Linux, this needs to be sorted out. Also it needs to be verified that explicit dark mode selection is actually working for macOS with the app bundle. If not it might make sense to limit this option to Windows only.

phw:
I just tested this on macOS. Indeed this works also in the app bundle! That's great, you're awsome :dagger: 

It does not fix the issue that automatically setting the color mode does not work, as macOS for some reason insists that Picard would not support this. But your forcing the color mode works.

I would be for renaming the current "System" to "Default". Then it does not imply so much that it uses the system setting. And for Linux (or really any OS except macOS, Windows and Haiku) I'd add a "System theme" entry to the dropdown that does the  not force the Fusion theme (exactly same behavior as we now get with the checkbox).

There should also be an option migration to convert the old "use_system_theme" to "ui_theme". See https://github.com/metabrainz/picard/blob/master/picard/config_upgrade.py . You basically would just bump the version number by one and add an upgrade function for this version.

phw:
Maybe needs to check the macOS version also. I think the dark mode is available since macOS 10.14, so we should not offer this option for earlier versions. Which means we can hide the dropdown completely.

zas (changes requested):

po/pt_BR.po
```diff
@@ -5951,3 +5951,15 @@ msgid ""
 "\n"
 "%s"
 msgstr "Erro ao abrir o navegador web:\n\n%s"
+
+#: picard/ui/theme.py:39
+msgid "Light"
+msgstr "Claro"
```
zas, line 5957: We use transifex, so I don't think those changes should be included.
We'll update `picard.pot` after the merge.

zas (changes requested):
The patch looks much better :+1:

po/pt_BR.po
Gabrielcarvfer: Sure, I was just testing if it worked correctly (failed miserably in using the translations in the local build). I'm going to squash and fix the commits. :)

Gabrielcarvfer:
> Maybe needs to check the macOS version also. I think the dark mode is available since macOS 10.14, so we should not offer this option for earlier versions. Which means we can hide the dropdown completely.

Something like?
```
        self.update_buttons = self.move_view.update_buttons

        if IS_MACOS:
            import platform
            macos_version = platform.mac_ver[0]
            if macos_version < "10.14":
                self.ui.labelTheme.hide()
                self.ui.ui_theme.hide()
```

po/pt_BR.po
Gabrielcarvfer: I don't really remember. Should I update something with the new entries to be translated?

Gabrielcarvfer:
> It does not fix the issue that automatically setting the color mode does not work, as macOS for some reason insists that Picard would not support this. But your forcing the color mode works.

Didn't really try to change it on the spot. Maybe it works... 

> I would be for renaming the current "System" to "Default". Then it does not imply so much that it uses the system setting. And for Linux (or really any OS except macOS, Windows and Haiku) I'd add a "System theme" entry to the dropdown that does the not force the Fusion theme (exactly same behavior as we now get with the checkbox).

Done. Tried with adwaita-dark, and it looks pretty bad (lack of contrast of the accent colors), but it works. 
Something important that should probably be in the docs: pyqt installed with pip did not recognize the system themes. 
Installing the package manager version made it work. 

> There should also be an option migration to convert the old "use_system_theme" to "ui_theme". See https://github.com/metabrainz/picard/blob/master/picard/config_upgrade.py . You basically would just bump the version number by one and add an upgrade function for this version.

Something like the following?
```
config.setting["ui_theme"] = UI_THEME_SYSTEM if config.setting["use_system_theme"] else UI_THEME_DEFAULT
del config.setting["use_system_theme"]
```

Gabrielcarvfer:
> Didn't really try to change it on the spot. Maybe it works...

It does work, but didn't manage to trigger the update without raising an exception (which draws a new window).

po/pt_BR.po
zas: To test translations locally, check https://github.com/metabrainz/picard/blob/master/INSTALL.md#running-from-the-source-tree
Basically, after updating one of `locale/*/*.po`, just run `python setup.py build_locales -i` so it compiles po files.

zas: What we'll do after the patch is merged:
- run `python setup.py regen_pot_file` and commit changes
- transifex will pick up this file within 6 hours (4 times a day)
- new translatable strings will be available on transifex
- before next release, we'll run `python setup.py pull_translations` to resync repo `.po` files with transifex

picard/config_upgrade.py
```diff
@@ -331,6 +331,16 @@ def upgrade_to_v2_6_0_beta_2(config):
...
 
 
+def upgrade_to_v2_6_0_beta_3(config):
+    """Replace use_system_theme with ui_theme options"""
+    from picard.ui.theme import (
+        UI_THEME_DEFAULT,
+        UI_THEME_SYSTEM
+    )
+    config.setting["ui_theme"] = UI_THEME_SYSTEM if config.setting["use_system_theme"] else UI_THEME_DEFAULT
+    del config.setting["use_system_theme"]
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done. Thank God you reminded of the test. I thought calling _s['ui_theme'] would create the new entry like in a dictionary, but it didn't. :)

picard/ui/options/interface.py
```diff
@@ -227,16 +237,16 @@ def save(self):
...
         restart_warning = None
-        if new_theme_setting != config.setting["use_system_theme"]:
+        if new_theme_setting != config.setting["ui_theme"]:
             restart_warning_title = _('Theme changed')
             restart_warning = _('You have changed the application theme. You have to restart Picard in order for the change to take effect.')
             if new_theme_setting:
                 restart_warning += '\n\n' + _(
                     'Please note that using the system theme might cause the user interface to be not shown correctly. '
-                    'If this is the case disable the "Use system theme" option to use Picard\'s default theme again.'
+                    'If this is the case select the "Light" theme option to use Picard\'s default theme again.'
```
zas, line 249: Under Linux, one only has choice between `System` or `Default`, so this message doesn't have much sense, as there's no `Light` them option to select:
image

zas (changes requested):

picard/ui/options/interface.py
Gabrielcarvfer: 🧐
Should I change the Windows/Mac DEFAULT_THEME option to SYSTEM_THEME, make it the default value and rename the DEFAULT_THEME to FUSION_THEME?

After that I would change the "Light" in this message to "Fusion" (or "Default").

phw: I think we are mixing up things here. There are both Qt5 themes and color palettes. So what we have is the following:

- Windows: We always use the "Fusion" theme, no exception here. By default we auto detect if Windows is configured for light or dark UI. In light we use Fusion's default color palette, in dark we use a custom one. We need the user be able to select between three states: "System", "Dark" and "Light".
- macOS: We always use Qt's macOS theme, no exception. In theory and when running from source we can detect whether light or dark UI is used. In the current app package this does not work. But we can explicitly force dark or light UI. We need the user be able to select between three states: "Default", "Dark" and "Light"
- Linux and most other OS: By default we use the "Fusion" theme, but we allow users to opt-in to use the system configured Qt5 theme. In both cases we can end up with light and dark UI (e.g. on KDE with dark UI configured it will set a dark color palette for Qt5 independent of the theme). We detect this via a heuristic and adapt some UI colors accordingly. We need two options: "Default" (using the Fusion theme) and "System"
- Haiku: We have some small special case here. We always use the default Haiku theme. There is no explicit dark or light UI option in Haiku. The user might still have configured dark UI colors. We detect this via the heuristic. We need no option for the user.

phw: > The Windows/Mac "Default" theme changes with the system theme (like the Linux "System" theme).

Just to clarify: This is not exactly the same. We always use Fusion on Windows and we always use the mac theme macOS, because those platforms do not have by default any way for the user to change their Qt5 theme. Linux on the other hand does, but by default we force the Fusion theme.

picard/ui/theme.py
```diff
@@ -75,6 +108,8 @@ def setup(self, app):
 
     @property
     def is_dark_theme(self):
+        if self._loaded_config_theme != UI_THEME_DEFAULT:
+            return self._loaded_config_theme == UI_THEME_DARK
```
phw, line 112: This should be

```
@property
    def is_dark_theme(self):
        if self._loaded_config_theme == UI_THEME_DARK:
            return True
        elif self._loaded_config_theme == UI_THEME_LIGHT:
            return  False
        else:
            return self._dark_theme
```

Otherwise it breaks the heuristic to detect whether the theme in use has dark background or not.

phw (changes requested):
> Tried with adwaita-dark, and it looks pretty bad (lack of contrast of the accent colors), but it works.

That's because the default implementation of `is_dark_theme` property is not correct, see my comment :) It would return always `False` if the ui theme was set to "System".

rdswift:
I get the feeling that once this is finalized it's going to be a nightmare to try to explain (with screen shots) in the documentation.  Hopefully I can build something based on @phw's explanation above.

phw:
@rdswift I don't think it will be that bad actually. One can describe the case where the user can explicitly choose between dark ad light theme, which is available on macOS and Windows. And then there is the option to use the Qt5 system configured in the desktop environment for Linux and similar OS.

rdswift:
Whew!  Sounds like most of the complication is behind the scenes then (in the code).  You guys make things so easy for the rest of us.  Thanks.

Gabrielcarvfer:
Don't know how about you, but when I read "Default" I assume the usual Picard look (Fusion+Light).

That's why I originally used "System".

phw:
> That's why I originally used "System".

Sure, you could also use "System" on Windows. Maybe that's better, it just makes it a bit more complicated in the implementation (System/Light/Dark on Win, Default/Light/Dark on Mac, Default/System on Linux and similar)

zas (changes requested):

picard/ui/theme.py
```diff
@@ -28,11 +30,39 @@
...
     IS_MACOS,
     IS_WIN,
 )
 
 
+OS_SUPPORTS_THEMES = True
+if IS_MACOS:
+    import platform
+    macos_version = platform.mac_ver()[0]
+    if macos_version < "10.14":  # Dark theme support was introduced in Mojave
```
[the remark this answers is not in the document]
phw: Should we instead convert it to a tuple like `(10, 14)` and compare that? Would probably be better than string comparison. Should not make a difference on the systems currently supported, but still.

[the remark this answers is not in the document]
zas: Yes, I agree, it would be safer.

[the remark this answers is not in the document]
Gabrielcarvfer: Something like this?
```
if IS_MACOS:
    def check_platform_version():
        import platform
        major, minor = tuple(map(int, platform.mac_ver()[0].split(".")))
        mohave_major, mohave_minor = (10, 14)  # Dark theme support was introduced in Mojave
        if major <= mohave_major and minor < mohave_minor:
            return False
        return True
    OS_SUPPORTS_THEMES = check_platform_version()
    del check_platform_version
```

picard/ui/theme.py
```diff
@@ -28,11 +30,39 @@
...
+UI_THEME_SYSTEM = "System"
+# Those are labels for display
+# example to display translated "System" theme name: _(UI_THEMES[UI_THEME_DEFAULT])
+# here strings are just marked for translation
+UI_THEMES = {
+    UI_THEME_DEFAULT: N_("Default"),
+    UI_THEME_LIGHT: N_("Light"),
+    UI_THEME_DARK: N_("Dark"),
+    UI_THEME_SYSTEM: N_("System")
+}
```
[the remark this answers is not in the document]
Gabrielcarvfer: Not sure. I don't think it makes a huge difference. All strings fit in a 64 bit word, so it literally could be done with a single instruction.

picard/ui/theme.py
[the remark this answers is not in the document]
Gabrielcarvfer: Uh, they include a last digit (sometimes).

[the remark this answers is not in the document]
Gabrielcarvfer: It should be fine now. Get the entire tuple but only compare the first two fields and assume OS XI (if it ever comes out) will support dark theme.

picard/ui/theme.py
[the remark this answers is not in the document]
phw: I agree, shouldn't make a big difference. And having the strings in config is I think an advantage for debugging and readability.

picard/ui/theme.py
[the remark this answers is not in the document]
phw: Calling `int`  can throw a `ValueError`. We should check this, don't want Picard to crash on start if a future macOS returns strange version here.

You also don't really need to compare the individual elements. A tuple comparison like `(10, 13, 2) < (10, 14)`  works as expected. But you could limit the split version already to two elements, e.g. `platform.mac_ver()[0].split(".")[:2]`. That would prevent unexpected results, especially in combination with potential `ValueError`. E.g. should for some reason mac_ver() return something like `"11.2.beta"`.

phw:
@Gabrielcarvfer I did a small change in https://github.com/metabrainz/picard/pull/1761/commits/06c60530098597e1a8fe07adb0edddc597c26791 to not have theme selection on Haiku and treat all other OS like Linux (which e.g. would be FreeBSD).

phw:
Overall this looks very good now to me. The macOS version code needs some minor tweaking, but apart from that I would suggest cleaning up the commit history a bit and this would be good to merge. It works as expected on the platforms I tested (Windows 10, macOS 11 and Ubuntu 20.10).

My KDE Neon and Haiku VMs are currently broken, so I could not check there (especially on KDE Neon with dark color palette). But given the code it should be fine.

picard/config_upgrade.py
```diff
@@ -331,6 +331,18 @@ def upgrade_to_v2_6_0_beta_2(config):
     rename_option(config, "caa_save_single_front_image", "save_only_one_front_image", BoolOption, False)
 
 
+def upgrade_to_v2_6_0_beta_3(config):
+    """Replace use_system_theme with ui_theme options"""
+    from picard.ui.theme import (
+        UiTheme,
+    )
```
zas, line 338: This one could be on one line.

test/test_config_upgrade.py
```diff
@@ -330,3 +331,14 @@ def test_upgrade_to_v2_6_0_beta_2(self):
         self.assertTrue(self.config.setting['image_type_as_filename'])
         self.assertNotIn('caa_save_single_front_image', self.config.setting)
         self.assertTrue(self.config.setting['save_only_one_front_image'])
+
+    def test_upgrade_to_v2_6_0_beta_3(self):
+        from picard.ui.theme import (
+            UiTheme,
+        )
```
zas, line 338: Import could be on one line here.

zas (changes requested):

picard/ui/theme.py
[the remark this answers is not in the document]
phw: Yes, I agree with you here. Let's stick to the enum, but save (lowercase) strings to the config

picard/ui/theme.py
```diff
@@ -159,18 +222,29 @@ def update_palette(self, palette, dark_theme, accent_color):
...
-            try:
-                basic_appearance = appearance.bestMatchFromAppearancesWithNames_([
-                    AppKit.NSAppearanceNameAqua,
-                    AppKit.NSAppearanceNameDarkAqua
-                ])
-                return basic_appearance == AppKit.NSAppearanceNameDarkAqua
-            except AttributeError:
-                return False
+                return dark_theme
+            if self._loaded_config_theme is not None and self._loaded_config_theme != UiTheme.DEFAULT:
```
zas, line 228: Perhaps initialize to `UiTheme.DEFAULT` in `BaseTheme.__init__()` and simplify  to:

```python
if self._loaded_config_theme != UiTheme.DEFAULT:
```

zas (changes requested):

picard/ui/theme.py
Gabrielcarvfer: I vaguely remember choosing None for a reason, probably some weird behavior on the Mac while debugging (screenshots), but seems to work fine with the default now.

picard/ui/theme.py
```diff
@@ -28,11 +34,55 @@
...
 )
 
 
+OS_SUPPORTS_THEMES = True
+if IS_MACOS:
+    def check_platform_version():
+        import platform
+        current_version = tuple(map(int, platform.mac_ver()[0].split(".")))
+        mojave_version = (10, 14)  # Dark theme support was introduced in Mojave
+        if current_version[0] <= mojave_version[0] and current_version[1] < mojave_version[1]:
```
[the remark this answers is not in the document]
Gabrielcarvfer: I inverted the operators *without noticing* while testing and got mesmerized it didn't work... xD

picard/ui/theme.py
```diff
@@ -28,11 +34,55 @@
...
     IS_HAIKU,
+    IS_LINUX,
     IS_MACOS,
     IS_WIN,
 )
 
 
+OS_SUPPORTS_THEMES = True
+if IS_MACOS:
+    def check_platform_version():
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

zas (changes requested):

picard/ui/theme.py
```diff
@@ -33,6 +39,76 @@
...
+# here strings are just marked for translation
+_UI_THEME_LABELS = {
+    UiTheme.DEFAULT: N_('Default'),
+    UiTheme.DARK: N_('Dark'),
+    UiTheme.LIGHT: N_('Light'),
+    UiTheme.SYSTEM: N_('System'),
+}
+
+
+def ui_theme_label(theme):
```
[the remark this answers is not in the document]
Gabrielcarvfer: Sure. Ended up changing the other two too. enum_to_name and name_to_enum is better than to/from string. :)

picard/ui/options/interface.py
```diff
@@ -227,16 +240,16 @@ def save(self):
...
-        new_theme_setting = self.ui.use_system_theme.isChecked()
+        new_theme_setting = theme_enum_to_name(self.ui.ui_theme.itemData(self.ui.ui_theme.currentIndex()))
         new_language = self.ui.ui_language.itemData(self.ui.ui_language.currentIndex())
         restart_warning = None
-        if new_theme_setting != config.setting["use_system_theme"]:
+        if new_theme_setting != config.setting["ui_theme"]:
             restart_warning_title = _('Theme changed')
             restart_warning = _('You have changed the application theme. You have to restart Picard in order for the change to take effect.')
-            if new_theme_setting:
+            if new_theme_setting == UiTheme.SYSTEM:
```
zas, line 249: Here we compare result of `theme_enum_to_name()` and enum value (int).

zas (changes requested):

picard/ui/options/interface.py
Gabrielcarvfer: You're reviewing this way too fast @__@

picard/ui/theme.py
```diff
@@ -33,6 +39,74 @@
...
+
+
+@unique
+# Those are values stored in config file:
+class UiTheme(IntEnum):
+    DEFAULT = 1
+    DARK = 2
+    LIGHT = 3
+    SYSTEM = 4
+
```
zas, line 68: Since we now use names in config file, we don't have much need of `IntEnum`, using something like:

```python
class UiTheme(Enum):

    DEFAULT = 'default'
    DARK = 'dark'
    LIGHT = 'light'
    SYSTEM = 'system'

    def __str__(self):
        return self.value
```

would allow us to get rid of int/name conversion.

`str(UiTheme.DARK)` being `dark`

Gabrielcarvfer: We've gone full circle. ._.

Gabrielcarvfer: Hmm, here was I thinking this would work. =x
```
_UI_THEME_LABELS = {}
_UI_THEME_LABELS.update([(theme_entry, N_(theme_entry.value.capitalize())) for theme_entry in list(UiTheme)])
```
Capitalizing the theme in UiTheme and moving the N_ stuff there would save a few more lines.

zas: Nope, you can't do that.

`N_()` is a noop function to mark strings for translation. It should never (ever) be called with anything but a string.

zas (dismissed):
The patch looks good now, sorry for the long circle, but at least we have something pretty.
Good work :)

Gabrielcarvfer:
@zas should I squash the commits?

picard/ui/theme.py
```diff
@@ -33,6 +36,57 @@
...
+        try:
+            current_version = tuple(map(int, platform.mac_ver()[0].split(".")))[:2]
+        except ValueError:
+            log.warning("Error while converting the MacOS version string into a tuple: %s" % platform.mac_ver()[0])
+            return False
+
+        mojave_version = (10, 14)  # Dark theme support was introduced in Mojave
+        return current_version >= mojave_version
+    OS_SUPPORTS_THEMES = is_dark_theme_supported()
+    del is_dark_theme_supported
```
zas, line 52: Not sure it is worth deleting the function, that's quite unusual, and prolly cost more than not doing it.

zas:
> @zas should I squash the commits?

Perhaps not all, but at least those needed to make commit history cleaner.

picard/ui/options/interface.py
```diff
@@ -209,14 +219,15 @@ def load(self):
...
         self.ui.quit_confirmation.setChecked(config.setting["quit_confirmation"])
-        self.ui.use_system_theme.setChecked(config.setting["use_system_theme"])
         current_ui_language = config.setting["ui_language"]
         self.ui.ui_language.setCurrentIndex(self.ui.ui_language.findData(current_ui_language))
         self.ui.filebrowser_horizontal_autoscroll.setChecked(config.setting["filebrowser_horizontal_autoscroll"])
         self.ui.starting_directory.setChecked(config.setting["starting_directory"])
         self.ui.starting_directory_path.setText(config.setting["starting_directory_path"])
         self.populate_action_list()
         self.ui.toolbar_layout_list.setCurrentRow(0)
+        current_theme = UiTheme(config.setting["ui_theme"])
```
zas, line 229: Since we removed previous code (with defaults), we need to ensure value read from config file is a valid `UiTheme`, else we'll have an exception raised.

I think the simplest would be to add a default value to the enum, use Python 3.6+ `Enum._missing_`:

```python
class UiTheme(Enum):
    DEFAULT = 'default'
    DARK = 'dark'
    LIGHT = 'light'
    SYSTEM = 'system'

    def __str__(self):
        return self.value

    @classmethod
    def _missing_(cls, value):
        return cls.DEFAULT
```

To test: stop Picard, change `ui_theme` value in config file to something invalid, start Picard

zas (changes requested):
The circle isn't close yet, sorry ;)

picard/ui/options/interface.py
Gabrielcarvfer: <details>
<summary>
Best circle is a closed one xD
</summary>
https://www.youtube.com/watch?v=ZvdXNx-dtQk
</details>

zas (dismissed):

picard/ui/theme.py
phw: It looks like an unnecessary and in the end pointless small optimization to me. I personally would not do this. It's no big deal either way, but it won't have any real effect, neither positively nor negatively.

Gabrielcarvfer:
Python 3.5 got deprecated and killed the CI ._.

phw:
It fails because of `Enum._missing_`, that's only supported in 3.6 or later. Do we strictly need this? We shouldn't access non existent enum members anyway.

We will need to ditch 3.5 support at some point, but not sure this is a good reason.

zas:
> We will need to ditch 3.5 support at some point, but not sure this is a good reason.

Well, a good reason will be that 3.5 reached end of life few months ago: https://devguide.python.org/devcycle/#end-of-life-branches
And since we are about to release a stable 2.6, that's perhaps the right time to do it.


> It fails because of `Enum._missing_`, that's only supported in 3.6 or later. Do we strictly need this? We shouldn't access non existent enum members anyway.

In fact, we shouldn't, but since config file can be modified we have to check if read values are valid. We could ofc catch exception and handle it.

Gabrielcarvfer:
> In fact, we shouldn't, but since config file can be modified we have to check if read values are valid. We could ofc catch exception and handle it.

I think we can get away with it by catching the exception exclusively in the BaseTheme.setup(). It is the the first time the setting is touched, coming straight from the file.

phw:
> And since we are about to release a stable 2.6, that's perhaps the right time to do it.

I think you are right about that. We also now require Qt >= 5.10, and that breaks support for Ubuntu 18.04 (one can still install via Flatpak or Snap instead, though, or stay with 2.5.x). I don't see a way around that, but even Ubuntu 18.04 did have Python 3.6 already. So yes, let's make 3.6 the new minimum supported.

picard/ui/options/interface.py
```diff
@@ -199,8 +207,10 @@ def fcmp(x):
         self.move_view = MoveableListView(self.ui.toolbar_layout_list, self.ui.up_button,
                                           self.ui.down_button, self.update_action_buttons)
         self.update_buttons = self.move_view.update_buttons
-        if IS_MACOS or IS_WIN or IS_HAIKU:
-            self.ui.use_system_theme.hide()
+
+        if not OS_SUPPORTS_THEMES:
+            self.ui.ui_theme.hide()
+            self.ui.label_theme.hide()
```
phw, line 213: Minor UI issue: The ui_theme combo box is inside a layout container (currently called `horizontalLayout_2`) together with an additional spacer. Hiding just the combobox still leaves this to use space because of this spacer.

I would suggest changing this by adding a plain empty Widget, name it something like `ui_theme_container` and make the current label and combobox + spacer childs of this container widget. The container should have layout margins all set to 0 to not introduce additional spacing around it. Then you can hide just this container here:

    self.ui.ui_theme_container.hide()

phw (changes requested):
This is great and we should get this merged! Thanks a lot. Just two minor things:

Some layout quirks when having theme selection disabled, see my comment.

Apart from that I had the options throw an exception the first time I ran the code because of invalid config value in my settings. While this is for one only caused by me using a broken config due to implementation changes while iterating and also could be solved with additional exception handling, let's go back to use `Enum._missing_`. That makes this more robust and simplifies the code.

I have already disabled the Python 3.5 tests as being required here on Github and also a proposed PR to raise the minimum version to 3.6.

Sorry for the back and forth here, but you had been so quick to do the changes ;)

picard/ui/theme.py
zas: @Gabrielcarvfer I guess you can remove this `del`

zas (changes requested):

picard/ui/options/interface.py
Gabrielcarvfer: No idea how to do that with no margins and still look right. It refused to align with the other stuff in the same layout.

phw: Yeah, Qt can be tricky here ;) I pushed https://github.com/metabrainz/picard/pull/1761/commits/bc908c41ec1310529dfda90e179bc0ac2aaed519 to your branch.  You can just squash my commit into the previous one.

The trick is to not add a layout as a child to the container widget, but have the widget itself have a vertical layout. In Qt designer if you right click on a widget that has childs you can select an explicit layout for this container. This only works if there is at least one child already, that can sometimes be annoying.

picard/ui/theme.py
```diff
@@ -33,6 +36,60 @@
...
+    DEFAULT = 'default'
+    DARK = 'dark'
+    LIGHT = 'light'
+    SYSTEM = 'system'
+
+    def __str__(self):
+        return self.value
+
+    @classmethod
+    def __missing__(cls, value):
```
phw, line 68: It should be `_missing_` (with single underscores).

zas: Good catch, I missed it.

zas, line 68:
```suggestion
    def _missing_(cls, value):
```

zas (changes requested):

picard/ui/options/interface.py
```diff
@@ -169,6 +171,12 @@ def __init__(self, parent=None):
         super().__init__(parent)
         self.ui = Ui_InterfaceOptionsPage()
         self.ui.setupUi(self)
+
+        self.ui.ui_theme.clear()
+        for theme in AVAILABLE_UI_THEMES:
+            self.ui.ui_theme.addItem(_(theme_enum_to_label(theme)), theme)
+        self.ui.ui_theme.setCurrentIndex(0)
```
zas, line 178: It shouldn't change much, but better select the actual default theme, rather than the first entry.

```suggestion
        self.ui.ui_theme.setCurrentIndex(self.ui.ui_theme.findData(UiTheme.DEFAULT))
```

zas (changes requested):

zas (approved):
@Gabrielcarvfer I directly made 2 minor fixes (see last commits), to me, this PR is ready to merge

phw (approved):

Gabrielcarvfer:
Sorry for being absent today, a bunch of work piled up. Thank you for being awesome to work with. :)

rdswift:
For the documentation changes, is something like the following okay?

-----

**User interface color theme**

This option allows the user to select the color scheme used by Picard.  The available choices are:

* Default - The default color scheme based on the operating system display settings.
* Light - A light display scheme.
* Dark - A dark display theme.

If the light or dark theme options are selected, these will override the operating system display theme and use the selected color theme for Picard.

-----

phw:
@rdswift That basically fits it, but I would distinguish between macOS / Windows and other OS. For macOS and Windows your original wording is fine:

- Default - The default color scheme based on the operating system display settings.
- Light - A light display scheme.
- Dark - A dark display theme.

The default handling on macOS will work properly once https://github.com/metabrainz/picard/pull/1764 has been merged (which I'm very happy to have finally got working).

For Linux and simular OS:

- Default - The default theme used by Picard.
- System -The Qt5 theme configured in the desktop environment.

zas:
@phw I wrote a small patch to add those descriptions to combobox items, see https://github.com/metabrainz/picard/pull/1765

rdswift:
I've updated the description based on the comments by @phw and the wording in the patch by @zas, and created https://github.com/rdswift/picard-docs/pull/89.  I also added a note about not honoring the colors set in the "Colors" section of the option settings when anything but the Default value is selected.   Please note any comments or suggestions on that PR.  Thanks.

PR #1763, opened 2021-03-16
Python 3.6 is now the minimum supported version

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Some of the features that might be interesting for us:

- [PEP 498](https://www.python.org/dev/peps/pep-0498/) f-strings (be careful and don't use for translations, though)
- [PEP 519](https://www.python.org/dev/peps/pep-0519/) file system path protocol

In addition:
- Python 3.6 would also have been beneficial in https://github.com/metabrainz/picard/pull/1761 for `Enum._missing_`.
- The bdist packages we build for Windows and macOS are already for 3.6 and later only (because of some issue with the required build tools not supporting 3.5, I currently don't remember the details)
- Even Ubuntu 18.04 already had Python 3.6, but it does not have the required Qt 5.10. Older Linux versions still on 3.5 likely don't fulfill the Qt requirement.
- python-libdiscid recently did not build on 3.5 anymore. It has been fixed upstream, but the days for 3.5 support are clearly counted.

# Solution

Make 3.6 the minimum supported version, remove running tests for older versions.

# Action

In the release announcement we should have a section for packagers mentioning the changes (PyQt + Qt >= 5.10, Python >= 3.6)

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -8,7 +8,7 @@ jobs:
     strategy:
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
-        python-version: [3.5, 3.6, 3.7, 3.8, 3.9]
+        python-version: [3.6, 3.7, 3.8, 3.9]
     env:
       CODACY_PROJECT_TOKEN: ${{ secrets.CODACY_PROJECT_TOKEN }}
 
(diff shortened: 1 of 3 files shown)


zas (approved):

phw:
Totally forgot about it, but mutagen also dropped Python 3.5 support: https://github.com/quodlibet/mutagen/issues/486

PR #1764, opened 2021-03-18
PICARD-2144: Fix automatic dark mode detection on macOS

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

Fix macOS app automatically detecting dark mode.

# Problem

* JIRA ticket (_optional_): PICARD-2144

While automatic dark mode detection was running in theory, because Qt5 supports it, it was not working for our binary builds. The reason is that macOS only activates this for an app if it was build for macOS 10.14 or later, but the default PyInstaller bootloader is compatible with older versions.

# Solution
Use a custom build of PyInstaller. For the resulting binary macOS now assumes that it does support dark mode.

# Action


phw:
See also https://github.com/pyinstaller/pyinstaller/issues/4886#issuecomment-631876110

zas (approved):

PR #1766, opened 2021-03-23
PICARD-2157: Reduce comparison overhead of logging window

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Calculate the starting position of log entries to print in the log window instead of comparing individual positions.

# Problem

* JIRA ticket (_optional_): PICARD-2157

The logging window drastically slows down Picard on Windows (at least on debug mode). The cause of the slowdown is the comparison of positions of entries that should be printed in the queue log. 

# Solution

Instead of comparing each item, calculate the starting position and copy everything available after that.

# Action

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -3,12 +3,13 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2007, 2011 Lukáš Lalinský
-# Copyright (C) 2008-2010, 2019 Philipp Wolfer
+# Copyright (C) 2008-2010, 2019, 2021 Philipp Wolfer
 # Copyright (C) 2012-2013 Michael Wiencek
 # Copyright (C) 2013, 2015, 2018-2019 Laurent Monin
 # Copyright (C) 2016-2018 Sambhav Kothari
 # Copyright (C) 2017 Sophist-UK
 # Copyright (C) 2018 Wieland Hoffmann
+# Copyright (C) 2021 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -89,6 +90,26 @@ class TailLogHandler(logging.Handler):
         self.tail_logger.updated.emit()
 
 
+def _calculate_bounds(previous_position, first_position, last_position, queue_length):
+    # If first item of the queue is bigger than prev, use first item position - 1 as prev
+    # e.g. queue = [8, 9, 10] , prev = 6, new_prev = 8-1 = 7
+    if previous_position < first_position:
+        previous_position = first_position-1
+
+    # The offset of the first item in the queue is
+    # equal to the length of the queue, minus the length to be printed
+    offset = queue_length - (last_position - previous_position)
+
+    # If prev > last_position, offset will be bigger than queue length offset > queue_length
+    # This will force an empty list
+    if offset > queue_length:
+        offset = queue_length
+
+    # If offset < 1, there is a discontinuity in the queue positions
+    # Setting queue_length to 0 informs the caller that something is wrong and the slow path should be taken
+    return offset, queue_length
+
+
 class TailLogger(QtCore.QObject):
     updated = QtCore.pyqtSignal()
 
@@ -100,8 +121,17 @@ class TailLogger(QtCore.QObject):
 
     def contents(self, prev=-1):
         with self._queue_lock:
-            contents = [x for x in self._log_queue if x.pos > prev]
-        return contents
+            # If log queue is empty, return
+            if not self._log_queue:
+                return []
+            offset, length = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))
+
+            if offset >= 0:
+                return (self._log_queue[i] for i in range(offset, length))
+                # If offset < 0, there is a discontinuity in the queue positions
+                # Use a slower approach to get the new content.
+            else:
+                return (x for x in self._log_queue if x.pos > prev)
 
     def clear(self):
         with self._queue_lock:
--- a/picard/ui/logview.py
+++ b/picard/ui/logview.py
@@ -9,6 +9,7 @@
 # Copyright (C) 2014 Sophist-UK
 # Copyright (C) 2016, 2018 Sambhav Kothari
 # Copyright (C) 2018 Wieland Hoffmann
+# Copyright (C) 2021 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -209,6 +210,8 @@ class LogView(LogViewCommon):
         self.hbox.addWidget(self.save_log_as_button)
         self.save_log_as_button.clicked.connect(self._save_log_as_do)
 
+        self._prev_logitem_level = log.VERBOSITY_DEFAULT
+
     def _clear_highlight_do(self):
         self.highlight_text.setText('')
         self.highlight_button.setEnabled(False)
@@ -294,10 +297,10 @@ class LogView(LogViewCommon):
     def _add_entry(self, logitem):
         if not self.is_shown(logitem):
             return
-        fmt = self.textCursor.blockCharFormat()
-        self.textCursor.setBlockCharFormat(self._format(logitem.level))
+        if self._prev_logitem_level != logitem.level:
+            self.textCursor.setBlockCharFormat(self._format(logitem.level))
+            self._prev_logitem_level = logitem.level
         super()._add_entry(logitem)
-        self.textCursor.setBlockCharFormat(fmt)
 
     def _set_verbosity(self, level):
         self.verbosity = level
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -1,0 +1,112 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2021 Gabriel Ferreira
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from collections import deque
+
+from test.picardtestcase import PicardTestCase
+
+from picard.log import _calculate_bounds
+
+
+class TestLogItem:
+    def __init__(self, pos=0):
+        self.pos = pos
+
+
+class TestLogItemQueue:
+    def __init__(self):
+        self._log_queue = deque(maxlen=10)
+
+    def contents(self, prev=-1):
+        if not self._log_queue:
+            return []
+        offset, length = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))
+
+        if offset >= 0:
+            return (self._log_queue[i] for i in range(offset, length))
+            # If offset < 0, there is a discontinuity in the queue positions
+            # Use a slower approach to get the new content.
+        else:
+            return (x for x in self._log_queue if x.pos > prev)
+
+    def push(self, item):
+        self._log_queue.append(item)
+
+
+class LogQueueCommonTest(PicardTestCase):
+    def setUp(self):
+        super().setUp()
+        self.item_queue = TestLogItemQueue()
+
+
+class LogQueueBoundsTestCase(LogQueueCommonTest):
+    def test_1(self):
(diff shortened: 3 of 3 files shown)


Gabrielcarvfer:
Should I include a test? 
<details>
<summary>
I've tested with this:
</summary>

```
from collections import deque

class Item():
    _log_queue = deque(maxlen=10)

    def __init__(self, pos=0):
        self.pos = pos

    @staticmethod
    def contents(prev=-1):
        # If last item position > deque maxlen, we are not using absolute numbers anymore
        last_pos = Item._log_queue[-1].pos
        if last_pos >= Item._log_queue.maxlen:
            offset = (prev + last_pos + 2) % Item._log_queue.maxlen
        else:
            offset = prev+1
        contents = [Item._log_queue[i] for i in range(offset, len(Item._log_queue))]
        return contents

    @staticmethod
    def push(item):
        Item._log_queue.append(item)

for i in range(8):
    Item.push(Item(i))
list = Item.contents()
if list[0].pos == 0 and list[-1].pos == 7:
    print("OK")
else:
    exit(-1)

for i in range(8, 15):
    Item.push(Item(i))
list2 = Item.contents(list[-1].pos)
if list2[0].pos == 8 and list2[-1].pos == 14:
    print("OK")
else:
    exit(-2)

for i in range(15, 20):
    Item.push(Item(i))
list3 = Item.contents(list2[-1].pos)
if list3[0].pos == 15 and list3[-1].pos == 19:
    print("OK")
else:
    exit(-3)
```
</details>

Gabrielcarvfer:
Do not merge. Just found a weird edge case. Reworking it. :)

phw:
Good one, that part of code clearly needs some refactoring. Looking forward to your updated patch.

> Should I include a test?

Yes, having tests for this would be really good. The logic is tricky and can easily break on future changes.

Gabrielcarvfer:
> Yes, having tests for this would be really good. The logic is tricky and can easily break on future changes.
Done. It is clearer, faster and rock solid (assuming the positions are monotonically increasing).
Not sure why the tests didn't work here, worked just fine locally. Looking into it. ._.

Gabrielcarvfer:
OK, never reuse variables between tests.

Gabrielcarvfer:
Still thinking on the discontinuation issue. 

If there is a missing position in the deque (e.g. [5, 10, 11]), the offset calculation would return a negative number (prev=4, offset=len-(last-prev)=3-[11-4]=-4), which is completely wrong as 4 is prev of 5, which is in offset=0).

Maybe putting something similar to the old code as a slow path in case offset < 0 (which indicates the discontinuation).

picard/log.py
```diff
@@ -100,7 +112,11 @@ def __init__(self, maxlen):
 
     def contents(self, prev=-1):
         with self._queue_lock:
-            contents = [x for x in self._log_queue if x.pos > prev]
+            # If log queue is empty, return
+            if len(self._log_queue) == 0:
+                return []
+            bounds = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))
+            contents = [self._log_queue[i] for i in range(*bounds)]
```
phw, line 119: I think we could make this an iterator:

    (self._log_queue[i] for i in range(*bounds))

`contents` only gets used to iterate over it afterwards, so we don't really need to create a list here.

test/test_log.py
```diff
@@ -0,0 +1,99 @@
...
+    def clear():
+        Item._log_queue.clear()
+
+
+class LogQueueBoundsTestCase(PicardTestCase):
+    def test_1(self):
+        # Common case where the item positions are within the max size of the queue
+        # [0,1,2,3,4,5,6,7], len = 8, maxlen = 10, offset = 0
+        for i in range(8):
+            Item.push(Item(i))
```
phw, line 57: The tests all use a single global `deque`. I'd rather have a separate instance created in `setup()` for this to initialize a fresh instance for each test. That also removes the need for each test to run `Item.clear()`.

phw (changes requested):
This works great and looks good to me. Only minor comments.

> If there is a missing position in the deque (e.g. [5, 10, 11]), the offset calculation would return a negative number

That actually should never happen, am I right? As I understand we have a single deque with upcounting positions, and nothing can be removed in between anyway.

But if we want the code to be more robust even in case of changes to the rest of the code a check like you suggested could be added.

phw:
I added PICARD-2157

@zas you implemented most of the current logic, what is your opinion of the continuation thing? Should it be handled?

zas:
> I added [PICARD-2157](https://tickets.metabrainz.org/browse/PICARD-2157)
> 
> @zas you implemented most of the current logic, what is your opinion of the continuation thing? Should it be handled?

Well, as said, it shouldn't happen. It would be safer to handle it, if we can find a simple enough way to do it.

Gabrielcarvfer:
<details>
<summary>
Looks terrible.
</summary>

```
def _get_log_contents(previous_position, log_queue):
    # If log queue is empty, return empty list
    if len(log_queue) == 0:
        return []

    first_position = log_queue[0].pos

    # If first item of the queue is bigger than prev, use first item position - 1 as prev
    # e.g. queue = [8, 9, 10] , prev = 6, new_prev = 8-1 = 7
    if previous_position < first_position:
        previous_position = first_position - 1

    last_position = log_queue[-1].pos

    # If the last position is smaller than the previous position, nothing useful in the queue
    if last_position < previous_position:
        return []

    queue_length = len(log_queue)

    # The offset of the first item in the queue is
    # equal to the length of the queue, minus the length to be printed
    offset = queue_length - (last_position - previous_position)

    # If offset is negative, there is a discontinuation in the log queue and we take the slow path
    if offset < 0:
        i = 0
        for queue_item in log_queue:
            if queue_item.pos > previous_position:
                offset = i-1
                break
    return (log_queue[i] for i in range(offset, queue_length))
```
</details>

<details>
<summary>
I much prefer doing something like this.
</summary>

```
def _calculate_bounds(previous_position, first_position, last_position, queue_length):
    # If first item of the queue is bigger than prev, use first item position - 1 as prev
    # e.g. queue = [8, 9, 10] , prev = 6, new_prev = 8-1 = 7
    if previous_position < first_position:
        previous_position = first_position-1

    # The offset of the first item in the queue is
    # equal to the length of the queue, minus the length to be printed
    offset = queue_length - (last_position - previous_position)

    # If prev > last_position, offset will be bigger than queue length offset > queue_length
    # This will force an empty list
    if offset > queue_length:
        offset = queue_length

    # If offset < 1, there is a discontinuity in the queue positions
    # Setting queue_length to 0 informs the caller that something is wrong and the slow path should be taken
    if offset < 0:
        queue_length = 0
    return offset, queue_length


    [...]

    def contents(self, prev=-1):
        with self._queue_lock:
            # If log queue is empty, return
            if len(self._log_queue) == 0:
                return []
            bounds = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))

            if not bounds[1]:
                # Take the slow path
                bounds = [0, len(self._log_queue)]
                for i, queue_item in enumerate(self._log_queue):
                    if queue_item.pos > prev:
                        bounds[0] = i-1
                        break
            contents = (self._log_queue[i] for i in range(*bounds))
        return contents
```
</details>

phw:
What about

```python
    def contents(self, prev=-1):
        with self._queue_lock:
            if len(self._log_queue) == 0:
                return []
            offset, length = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))
            if offset >= 0:
                contents = (self._log_queue[i] for i in range(offset, length))
            # If offset < 1, there is a discontinuity in the queue positions
            # Use a slower approach to get the new content.
            else:
                contents = (x for x in self._log_queue if x.pos > prev)
        return contents
```

Leaving `_calculate_bounds`  as is and handling the special case as we did before. In both cases it's an iterator.

Gabrielcarvfer:
Changing variables name is fine, and makes clearer. I was thinking more if using the length as a flag would be ok, or you would prefer an exception.

The problem is in the slow path. I've seen 17k items in the queue, and the OG code (also used in your suggestion) would check item by item, usually to get only the a few items in the end of the queue. 

I was thinking of using binary search  to make the slow path faster (assuming items are at the very least sorted).

phw:
As this is for a case we expect to never happen I would not overcomplicate it. If this was some other code I would actually say that it should fail and log an error.

Since we are in error logging code here it probably is a good idea to still continue, as error logging is something you especially want in case of a bug. But IMHO it is more useful to keep the code real simple for the unexpected case.

Gabrielcarvfer:
For the most part, it should be quite a bit faster.
I was looking at related stuff, and found this
```
        fmt = self.textCursor.blockCharFormat()
        self.textCursor.setBlockCharFormat(self._format(logitem.level))
        super()._add_entry(logitem)
        self.textCursor.setBlockCharFormat(fmt)  # is this format really necessary? didn't seem to change anything when commented out, at least on Windows
        ```

**Comment by zas** [comment]
> For the most part, I think this is done. However I started looking at related stuff, and found this
> 
> ```
> fmt = self.textCursor.blockCharFormat()
> self.textCursor.setBlockCharFormat(self._format(logitem.level))
> super()._add_entry(logitem)
> self.textCursor.setBlockCharFormat(fmt)  # is this format really necessary? didn't seem to change anything when commented out, at least on Windows
> ```

I think it is needed to revert to normal format after the change to `self._format(logitem.level)`

**Comment by Gabrielcarvfer** [comment]
> I think it is needed to revert to normal format after the change to `self._format(logitem.level)`

But doesn't it get updated per logitem.level? So if the next item has a different level, it will switch to the correct one and print, then the next logitem will do the same, etc. Should be fine if all logitems have their own format.

**Comment by zas** [comment]
Yes, it could probably be optimized, we could even skip changing format if not strictly needed.

**Comment by Gabrielcarvfer** [comment]
<details>
<summary>
With double SetCharBlockFormat
</summary>

pre_format_changes
</details>

<details>
<summary>
Only call SetCharBlockFormat if logitem.level is different from the previous item
</summary>

post_format_changes
post_format_change_setBlockCharFormat
</details>

14k calls down to 11 seems like a win to me, even though I have no idea how to measure responsiveness of anything. Nothing seems to have changed.

**Review by phw — APPROVED** [review]
Looks great to me. Good catch with the format change. This reduction of calls definitely makes sense.

File: picard/ui/logview.py [hunk 3]
```diff
@@ -294,10 +297,10 @@ def is_shown(self, logitem):
     def _add_entry(self, logitem):
         if not self.is_shown(logitem):
             return
-        fmt = self.textCursor.blockCharFormat()
-        self.textCursor.setBlockCharFormat(self._format(logitem.level))
+        if self._prev_logitem_level != logitem.level:
+            self.textCursor.setBlockCharFormat(self._format(logitem.level))
         super()._add_entry(logitem)
-        self.textCursor.setBlockCharFormat(fmt)
+        self._prev_logitem_level = logitem.level
```
zas, line 303: What about:

```python
        if self._prev_logitem_level != logitem.level:
            self.textCursor.setBlockCharFormat(self._format(logitem.level))
            self._prev_logitem_level = logitem.level
        super()._add_entry(logitem)
```

test/test_log.py
```diff
@@ -0,0 +1,113 @@
...
+    def __init__(self, pos=0):
+        self.pos = pos
+
+
+class ItemQueue:
+    def __init__(self):
+        self._log_queue = deque(maxlen=10)
+
+    def contents(self, prev=-1):
+        if len(self._log_queue) == 0:
```
zas, line 38: It could be written in a more pythonic form:

```python
if not self._log_queue:
    return []
```

test/test_log.py
```diff
@@ -0,0 +1,113 @@
...
+            return []
+        offset, length = _calculate_bounds(prev, self._log_queue[0].pos, self._log_queue[-1].pos, len(self._log_queue))
+
+        if offset >= 0:
+            contents = (self._log_queue[i] for i in range(offset, length))
+            # If offset < 0, there is a discontinuity in the queue positions
+            # Use a slower approach to get the new content.
+        else:
+            contents = (x for x in self._log_queue if x.pos > prev)
+        return contents
```
zas, line 48: Since both branches set `contents`, we could just `return` directly (dropping variable)

test/test_log.py
```diff
@@ -0,0 +1,113 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from collections import deque
+
+from test.picardtestcase import PicardTestCase
+
+from picard.log import _calculate_bounds
+
+
+class Item:
```
zas, line 28: We already have a class named `Item` [elsewhere](https://github.com/metabrainz/picard/blob/4e13a1835d1f97cd381204f11eb33c16f241e558/picard/ui/item.py#L30), perhaps use another name.

zas (changes requested):
Looks good to me, just minor comments.

test/test_log.py
Gabrielcarvfer: Changed to TestLogItem. Should be better. :)

test/test_log.py
Gabrielcarvfer: Sure. I usually try to keep a single point of exit to simplify measurements and debugging.

test/test_log.py
Gabrielcarvfer: Done.

zas (approved):

phw (approved):
We'll merge this for 2.6.1 after the 2.6 release

PR #1768, opened 2021-03-24
PICARD-2158: Use Scripting Documentation dialog in renaming options

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2158, PICARD-2159

In Options > Scripting there is a new Scirpting Documentation dialog, but this dialog is not available in Options > File Naming. Instead there is only a link to online documentation

# Solution

Make it consistent and have the same Scripting Documentation dialog at both places. Also make `ScriptingDocumentationDialog` a `SingletonDialog` instead of custom implementation to ensure only one instance.

grafik


phw:
Found a bug on macOS where the scripting docs dialog gets opened behind the options dialog on macOS and is not focusable. See PICARD-2159

The issue is independent of the changes for reusing the dialog, but since it touches the same code I include the fix in this PR, see https://github.com/metabrainz/picard/pull/1768/commits/c50b6d863c2c941df6441bb76604587e8e5175ac

zas (approved):

PR #1769, opened 2021-03-27
PICARD-2161: Add function to remove duplicate entries from multi-value variable

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  This adds a new scripting function `$uniquemulti()` to remove duplicate items from a multi-value variable.

# Problem

Duplicate items can be created in a multi-value variable when combining lists such as genres.  There is no easy way to remove duplicate entries.

* JIRA ticket (_optional_): PICARD-2161

# Solution

Add a new scripting function `$uniquemulti()` to remove duplicate entries.  Include an option to do case-sensitive comparisons.

# Action

Note to update the Picard documentation to include this new function.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1299,26 +1299,26 @@ def func_reversemulti(parser, multi, separator=MULTI_VALUED_JOINER):
 
 
 @script_function(eval_args=False, documentation=N_(
-    """`$uniquemulti(name,case_sensitive="",separator="; ")`
+    """`$unique(name,case_sensitive="",separator="; ")`
 
 Returns a copy of the multi-value tag `name` with no duplicate elements.
     By default, a case-insensitive comparison of the elements is performed.
 
 Example 1:
 
-    $uniquemulti(a; A; B; b; cd; Cd; cD; CD; a; A; b)
+    $unique(a; A; B; b; cd; Cd; cD; CD; a; A; b)
 
 Result: A; CD; b
 
 Example 2:
 
-    $uniquemulti(a; A; B; b; a; b; A; B, cd,True)
+    $unique(a; A; B; b; a; b; A; B, cd,True)
 
 Result: A; B; a; b; cd
 """
 ))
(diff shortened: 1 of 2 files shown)


rdswift:
Would this be better if I used the `picard.util.uniquify()` function, and dropped the ability for case-insensitive comparisons?

zas:
> Would this be better if I used the `picard.util.uniquify()` function, and dropped the ability for case-insensitive comparisons?

Nope, code looks good as is to me.

picard/script/functions.py
```diff
@@ -1296,3 +1296,29 @@ def func_sortmulti(parser, multi, separator=MULTI_VALUED_JOINER):
...
+    $uniquemulti(a; A; B; b; a; b; A; B, cd,True)
+
+Result: A; B; a; b; cd
+"""
+))
+def func_uniquemulti(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER):
+    multi_value = MultiValue(parser, multi, separator)
+    if not case_sensitive:
+        multi_value._multi = list({v.lower(): v for v in multi_value}.values())
+    return multi_value.separator.join(sorted(list(set(multi_value))))
```
zas, line 1324: `sorted(list(set(...)))` -> `sorted(set(...))`

zas (changes requested):

phw:
Could we get away with just $unique instead of $uniquemulti or would that be unclear?

For me I would always expect a function $unique to work on a list of things.

rdswift:
> Could we get away with just $unique instead of $uniquemulti or would that be unclear?
> 
> For me I would always expect a function $unique to work on a list of things.

I'm okay with changing it to just `$unique`.  That actually makes more sense.

zas (approved):

phw (approved):

PR #1770, opened 2021-03-27
PICARD-2162: Improve performance of coverart box when selecting multiple files

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Make selection updates faster by comparing less stuff.

# Problem

When the selected items change, the metadatabox triggers an update of the coverart box, which compares and selects unique images to display.

* JIRA ticket (_optional_): PICARD-2162

# Solution

Replace the comparison of sets of images by a dictionary update. If the dictionary grows, new images were found and the 'changed' flag is set. If size remains the same, nothing changes.

# Action

### DIFF
--- a/picard/util/imagelist.py
+++ b/picard/util/imagelist.py
@@ -117,9 +117,7 @@ class ImageListState:
         self.update_orig_metadata = False
 
 
-def _process_images(state, src_obj):
-    from picard.track import Track
-
+def _process_images(state, src_obj, Track):
     # Check new images
     if state.update_new_metadata:
         src_dict = src_obj.metadata.images.hash_dict()
@@ -144,9 +142,11 @@ def _process_images(state, src_obj):
 
 
 def _update_state(obj, state):
+    from picard.track import Track
+
     changed = False
     for src_obj in state.sources:
-        _process_images(state, src_obj)
+        _process_images(state, src_obj, Track)
 
     if state.update_new_metadata:
         updated_images = ImageList(state.new_images.values())
--- a/picard/util/imagelist.py
+++ b/picard/util/imagelist.py
@@ -6,6 +6,7 @@
 # Copyright (C) 2017 Sambhav Kothari
 # Copyright (C) 2018, 2020 Philipp Wolfer
 # Copyright (C) 2019 Laurent Monin
+# Copyright (C) 2021 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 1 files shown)


Gabrielcarvfer:
With 18k files selected, the time spent on picard.util.imagelist._update_state fell from around 1.5 s (pretty long freeze) to 90 ms (16x faster, which still is noticeable).

Gabrielcarvfer:
Original
pre_changes
Switch from sets to dict
post_changes
Moving the Track import to the caller and passing it by argument to the callee (~30x speedup :D)
post_changes_2

picard/util/imagelist.py
```diff
@@ -104,42 +119,46 @@ def __init__(self):
...
 
-def _process_images(state, src_obj):
-    from picard.track import Track
-
+def _process_images(state, src_obj, Track):
     # Check new images
     if state.update_new_metadata:
-        if state.new_images != set(src_obj.metadata.images):
+        src_dict = src_obj.metadata.images.hash_dict()
+        og_len = len(state.new_images)
```
zas, line 126: s/og_len/orig_len/?

zas (changes requested):

picard/util/imagelist.py
Gabrielcarvfer: Changed for the `s`. I didn't want to use `orig_len` because there are `metadata` and `orig_metadata`.

zas: Actually `s` is even less explicit (and `s` is used much more often for strings than lengths, if we were to keep one-char variable name, `l` would be more logicial).
What's `len(state.new_images)` ?

picard/util/imagelist.py
```diff
@@ -104,42 +119,46 @@ def __init__(self):
...
 
-def _process_images(state, src_obj):
-    from picard.track import Track
-
+def _process_images(state, src_obj, Track):
     # Check new images
     if state.update_new_metadata:
-        if state.new_images != set(src_obj.metadata.images):
+        src_dict = src_obj.metadata.images.hash_dict()
+        s = len(state.new_images)
```
zas, line 126: What about `pre_update_length` or `prev_len`?

zas (changes requested):

picard/util/imagelist.py
Gabrielcarvfer: Way better.

zas (approved):

phw (approved):

PR #1771, opened 2021-03-28
Cythonize picard

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
Make performance profiling easier by having debugging symbols that can be traced by perf/AMD uProf/Intel VTune/etc.

# Problem

Profiling Picard is quite hard as most of Python profilers can't see methods invoked by Qt/C. All of them end up mashed together in the exec_ or self (for objects inheriting Qt classes).

# Solution

Automatically find Picard sources that can be compiled and use their compiled versions if Cython is available.

# Action

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -29,6 +29,8 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+import builtins
+from gettext import ngettext
 
 from picard.version import (
     Version,
@@ -48,6 +50,9 @@ PICARD_VERSION = Version(2, 6, 0, 'beta', 3)
 # it should be in the form '<platform>_<YYMMDDHHMMSS>'
 # ie. win32_20140415091256
 PICARD_BUILD_VERSION_STR = ""
+
+builtins.__dict__['N_'] = lambda a: a
+builtins.__dict__['ngettext'] = ngettext
 
 
 def version_to_string(version, short=False):
--- a/picard/i18n.py
+++ b/picard/i18n.py
@@ -35,9 +35,6 @@ from picard.const.sys import (
 )
 
 
-builtins.__dict__['N_'] = lambda a: a
-
-
 def setup_gettext(localedir, ui_language=None, logger=None):
     """Setup locales, load translations, install gettext functions."""
     if not logger:
--- a/setup.py
+++ b/setup.py
@@ -21,6 +21,7 @@
 # Copyright (C) 2018 Kartik Ohri
 # Copyright (C) 2018 virusMac
 # Copyright (C) 2019 Kurt Mosiejczuk
+# Copyright (C) 2021 Gabriel Ferreira
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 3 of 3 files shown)


Gabrielcarvfer:
The "cythonization" of tagger, file and metadatabox (second commit) is just an example of changes that needed to be made to compile the sources. 

Examples of interesting stuff that you can see by doing this:

<details>
<summary>CPU load of format (a.k.a. decoding mp3 with mutagen)</summary>

image
</details>

<details>
<summary>CPU load of coverart processing</summary>

image
</details>

<details>
<summary>Branches and cache misses during selection of multiple files (how I found the issue in #1770)</summary>

image

image
</details>

Sophist-UK:
This is an interesting PR. But as far as I can see, it will cythonize if it possibly can.

Given the significance of this, can I suggest that you run it only if a command line parameter is added - "-c" perhaps?

Sophist-UK:
This is an interesting PR. But as far as I can see, it will cythonize if it possibly can.

Given the significance of this, can I suggest that you run it only if a command line parameter is added - "-c" perhaps?

Gabrielcarvfer:
Sure, I was thinking about that. Not sure how many people run it from source and build it though.

phw:
What's the reason to factor out all those _() calls?  This'll get horrible if we need to apply it to more of the code base.

Can we find a way to make _() and related messages to the cythonized modules instead?

Gabrielcarvfer:
> What's the reason to factor out all those _() calls? This'll get horrible if we need to apply it to more of the code base.
> 


image

> Can we find a way to make _() and related functions available to the cythonized modules instead?

Yup, looking into it.

Gabrielcarvfer:
Putting the following in `picard/__init__.py` and excluding a few specific sources do the trick. 
```
import builtins
from gettext import ngettext
builtins.__dict__['N_'] = lambda a: a
builtins.__dict__['ngettext'] = ngettext
```

Excluded files:
```
./picard/i18n.py => pgettext_attributes
./picard/ui/options/releases.py => gettext_countries
./picard/coverart/utils.py => pgettext_attributes
```

Gabrielcarvfer:
Hmm, very weird. Now Picard can only have a single window (clicking to open the settings window actually show up on the main window).

Gabrielcarvfer:
Hmm, very interesting stuff. It turns out the lock in DataHash.__init__ is held during the md5 calculation (which can take some time) + time to write the temporary file (which can take a very long time, since we're also reading files at the same time). 

After a quick look, I think there's no need to lock before the md5 nor to keep the lock while writing the file. Taking the MD5 out of the lock block and splitting it into two locks reduces mean time to process from 150ms to 40ms per cover.

Gabrielcarvfer:
> My concern with this patch was performance. And indeed the individual modules became much faster. The overall impact on Picard however was not that noticeable. 

We will never see gains like those from 1843 and 1844 again. Overall impact is basically null because locking and IO prevent the CPU from processing, and if the CPU process indiscriminately, then the interface freezes (this is why I didn't not open a PR for the double locking I've shown).

> Anyway, as I understand your concern with this PR is more about the profiling and getting actual symbols to help identifying the relevant code paths.
> 
> For me the question is:
> 
> * Do we have any actual downside of cythonizing everything (if possible)?

Build size (~500MB). The mainwindow.py binary is the biggest with 40MB. Itemviews.py comes second with 13MB. 
Now that the N_ issue got solved, I think it is possible to compile every source into a single big chungus library. I'm going to test if it can trim some of that size down.

> * Do we have any benefit of cythonizing everything (if possible)?

Debugging symbols.

> * Should we cythonize everything in official builds?

No opinions on this. It takes quite some time to compile everything, and cython is pretty verbose. At the same time, some modules have speedups of 2-4x without any additional changes.

> In any case I think it should be optional during build. Both your and mine patch did this by actually only cythonize if cython is available. But actually I would be in favor of a flag as proposed by @Sophist-UK .

I've included a '-cythonize' option. '-c' is reserved.

phw:
After a short discussion with zas I'm going to close this for now. This was valuable to explore the results, but the minor gains are not worth the extended maintenance and support requirements.

If we would do something like this in the future I would be more in favor of a more minimal approach of only cythonizing selected modules. But even that showed little performance benefits overall. See also my comment above.

PR #1773, opened 2021-03-28
PICARD-2160: Dark Mode switching does not change list elements leading to mix of light and dark mode.

Gabrielcarvfer:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
Seems like I've forgotten to force a theme once loaded when the Default option was selected, allowing for on-the-fly theme changes. However, some palettes get initialized before the theme itself, leading to the reported issue.

# Problem

* JIRA ticket (_optional_): PICARD-2160

# Solution

Enforce current Mac Appearance during the MacTheme setup, locking it to either the Light or the Dark theme.

# Action

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -209,31 +209,43 @@ elif IS_MACOS:
     except ImportError:
         AppKit = None
 
+    dark_appearance = False
+    if AppKit:
+        # Default procedure to identify the current appearance (theme)
+        appearance = AppKit.NSAppearance.currentAppearance()
+        try:
+            basic_appearance = appearance.bestMatchFromAppearancesWithNames_([
+                AppKit.NSAppearanceNameAqua,
+                AppKit.NSAppearanceNameDarkAqua
+            ])
+            dark_appearance = basic_appearance == AppKit.NSAppearanceNameDarkAqua
+        except AttributeError:
+            pass
+
     class MacTheme(BaseTheme):
+        def setup(self, app):
+            super().setup(app)
+
+            if self._loaded_config_theme != UiTheme.DEFAULT:
+                dark_theme = self._loaded_config_theme == UiTheme.DARK
+            else:
+                dark_theme = dark_appearance
+
+            # MacOS uses a NSAppearance object to change the current application appearance
+            # We call this even if UiTheme is the default, preventing MacOS from switching on-the-fly
+            if dark_theme:
+                appearance = AppKit.NSAppearance._darkAquaAppearance()
+            else:
+                appearance = AppKit.NSAppearance._aquaAppearance()
+            AppKit.NSApplication.sharedApplication().setAppearance_(appearance)
+
         @property
         def is_dark_theme(self):
             dark_theme = False
-            if not AppKit:
-                return dark_theme
-            if self._loaded_config_theme is not None and self._loaded_config_theme != UiTheme.DEFAULT:
-                dark_theme = self._loaded_config_theme == UiTheme.DARK
-                # MacOS uses a NSAppearance object to change the current application appearance
-                if dark_theme:
-                    appearance = AppKit.NSAppearance._darkAquaAppearance()
-                else:
-                    appearance = AppKit.NSAppearance._aquaAppearance()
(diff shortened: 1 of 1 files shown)


picard/ui/theme.py
```diff
@@ -209,31 +209,43 @@ def update_palette(self, palette, dark_theme, accent_color):
...
+            super().setup(app)
+
+            if self._loaded_config_theme != UiTheme.DEFAULT:
+                dark_theme = self._loaded_config_theme == UiTheme.DARK
+            else:
+                dark_theme = dark_appearance
+
+            # MacOS uses a NSAppearance object to change the current application appearance
+            # We call this even if UiTheme is the default, preventing MacOS from switching on-the-fly
+            if dark_theme:
```
phw, line 236: As AppKit is optional you need to check for it here with `if AppKit` as well.

phw (changes requested):
Thanks, tested and works. I think we should merge this as a fix, otherwise macOS users will be hit by this.

Just noting that the core issue runs a bit deeper and is not limited to macOS and not even directly related to recent theming work, rather it is basically as old as Picard itself. The same effect happens on all platforms if Qt dynamically switches the color palette. This e.g. can also be triggered by switching the color palette in KDE Neon.

I once had a partial patch for this, but it was before we had the theme module and related helpers. The fix would be to listen for QGuiApplication.paletteChanged and adjust colors accordingly. My patch did this isolated in itemviews to reset the background colors in the list. But with the new theme module the solution would probably be to listen to palette changes there, determine whether this has an impact on dark / light theme changes and emit a "theme_changed" signal. Then places using colors that depend on dark / light mode can use this signal to update (this might actually the tricky part, quite a few cases to consider.

A different workaround would be to disable palette changes altogether after application has been initialized.

But let's merge this first as is as it provides theme handling on macOS consistent with how it currently works on Windows.

phw:
@Gabrielcarvfer Please see https://github.com/Gabrielcarvfer/picard/pull/2 . I tested without AppKit and this has some fixes. Support running without AppKit has no high priority, but still let's make it work better in this case as well.

phw (approved):

picard/ui/theme.py
```diff
@@ -204,37 +199,49 @@ def update_palette(self, palette, dark_theme, accent_color):
     theme = WindowsTheme()
 
 elif IS_MACOS:
-    try:
-        import AppKit
-    except ImportError:
-        AppKit = None
+    dark_appearance = False
+    if OS_SUPPORTS_THEMES:
```
zas, line 203: I think it would be better to test for `AppKit` even though at the moment `OS_SUPPORTS_THEMES` is always `False` when `AppKit` is unavailable.
It would be much cleaner and future-proof to enclose code depending on `AppKit` in a `if AppKit:`

To me, `OS_SUPPORTS_THEMES` is another thing.

zas (changes requested):

zas (dismissed):

picard/ui/theme.py
phw: One reason to check for OS_SUPPORTS_THEMES is that the following code is not needed if that is false, independent of whether AppKit would be available or not. The fully correct check would probably be `if OS_SUPPORTS_THEMES  and AppKit`, but as OS_SUPPORTS_THEMES  is always false if no AppKit is present I skipped that.

zas: > OS_SUPPORTS_THEMES is always false if no AppKit is present

Actually that was my point, because if the expression defining `OS_SUPPORTS_THEMES` changes, it might not be true in the future, so I was thinking it would be safer to have explicit test for Appkit.

phw: I have changed the checks to `OS_SUPPORTS_THEMES and AppKit` for code that makes direct use of AppKit. You are right, that could prevent bugs on future refactorings. `OS_SUPPORTS_THEMES ` is not necessarily tied to AppKit, that's just how it turns out to work currently.

phw (approved):

zas (approved):

PR #1774, opened 2021-03-30
PICARD-2166: ‘Use original values’ for a tag on a large selection takes very long with a high CPU usage

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
Batch tag processing (use originals or remove tags) triggers a ton of mainwindow.update_selection (caused by individual file updates inside the current selection), which can be very expensive. This change stops mainwindow.update_selection temporarily during file changes, letting the following explicitly call to mainwindow.update_selection do its thing just a single time.

# Problem

Each file.update() called for each changed/removed tag of each modified file in metadatabox._apply_update_funcs triggers a mainwindow.update_selection call, which tries to update the metadatabox, coverartbox and actions available in Picard.

For a small number of files, this translates into a short freeze that might not be perceptible, but for a big selection of files this can take quite some time.

* JIRA ticket (_optional_): PICARD-2166

# Solution

Prevent mainwindow.update_selection during bulk changes and let the explicit call deal with the changes during metadatabox._apply_update_funcs.

# Action

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -448,8 +448,9 @@ class MetadataBox(QtWidgets.QTableWidget):
         event.accept()
 
     def _apply_update_funcs(self, funcs):
-        for f in funcs:
-            f()
+        with self.parent.ignore_selection_changes:
+            for f in funcs:
+                f()
         self.parent.update_selection(new_selection=False, drop_album_caches=True)
 
     def edit_tag(self, tag):


Gabrielcarvfer:
Side effect, you won't be able to change selection while this is processing. Hopefully it will be so fast that it won't matter. =x

Test results considering changes to 4 text tags in 41 files without linked albums and rolling them back to the originals.
image

phw:
Thanks a lot again! Need to look into this in more detail if there are any side effects we might miss, but from a first look this looks like the right thing to do.

> Side effect, you won't be able to change selection while this is processing. Hopefully it will be so fast that it won't matter. =x

If it makes this operation fast enought I think that's acceptable.

Gabrielcarvfer:
Don't merge this yet. The same issue happens when trying to edit a tag (I was in the middle of testing with 18k files... not responding). xD

Gabrielcarvfer:
You're welcome. :)

phw (approved):

phw:
Merged with https://github.com/metabrainz/picard/commit/c5cf29c1c2c84dca7f1da6c55a0cf0a82a68edcb

PR #1776, opened 2021-03-31
PICARD-2178: Select random sample of loaded files to use as renaming examples

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
Instead of using the two provided file examples for renaming, pick up to 10 random entries from the loaded files and use them to show before/after the renaming script. Also includes a button to refresh the up to 10 entries.

# Problem

* JIRA ticket (_optional_): PICARD-2178

Writing a naming script is hard by itself.

# Solution

Make it a bit easier to check the script by selecting loaded files as examples and showing the before and after.

# Action


Sophist-UK:
Screenshots please.

Gabrielcarvfer:
@Sophist-UK 

<details>
<summary>
Without loading song files (use the default examples)
</summary>

empty
</details>

<details>
<summary>
With song files loaded and set not to move files
</summary>

with_stuff

</details>

<details>
<summary>
With song files loaded and set to move files
</summary>

with_path
</details>

picard/ui/options/renaming.py
```diff
@@ -173,22 +176,42 @@ def _example_to_filename(self, file):
...
+            return "", ""
+
+    def _sample_example_files(self):
+        # Get a new sample of randomly selected files to use as renaming examples
+        import random
+        keys = list(self.tagger.files.keys())
+        files_len = len(keys)
+        indexes = set([random.randrange(0, files_len) for i in range(min(files_len, 10))])  # pick up to 10 samples
+        self._sampled_example_files = [self.tagger.files[keys[index]] for index in indexes]
+        self.update_examples()
```
zas, line 196: A much simpler (and reliable) way to do it:

```python
    def _sample_example_files(self):
        # Get a new sample of randomly selected files to use as renaming examples
        import random
        files = self.tagger.files
        length = min(len(files), 10)
        self._sampled_example_files = [files[key] for key in random.sample(files.keys(), k=length)]
        self.update_examples()
```

zas (changes requested):
I like this approach.
I wonder though about the layout, because with long paths, and small options window, having 2 columns doesn't help.

But anyway, that's a good move!

picard/ui/options/renaming.py
Gabrielcarvfer: The C programmer in me always like to access stuff with numerical indexes. xD

zas:
I don't think horizontal scrollbars should be synced, but vertical ones (which only appear with small dialog height), yes.
About the button, perhaps move it on the same line as the ones to show documentation and default, not ideal, but it will give a bit more vertical space for the display.

zas:
I think this PR can be linked to PICARD-270

Gabrielcarvfer:
> I don't think horizontal scrollbars should be synced, but vertical ones (which only appear with small dialog height), yes.

It works with 
```
self.ui.example_filename_before.verticalScrollBar().valueChanged.connect(
    self.ui.example_filename_after.verticalScrollBar().setValue)

self.ui.example_filename_after.verticalScrollBar().valueChanged.connect(
    self.ui.example_filename_before.verticalScrollBar().setValue)
```

But if you move the horizontal bar fast enough, it creates this weirdness where part of the text simply dissapears (it gets fixed by scrolling vertically or waiting a bit).
weird


> About the button, perhaps move it on the same line as the ones to show documentation and default, not ideal, but it will give a bit more vertical space for the display.

Like this? (had to cut down the text)
image


> edit: also perhaps align to right so the actual file name is shown, since it is more likely the start of the full path is common.

Looks disorienting to me
image

zas:
> > I don't think horizontal scrollbars should be synced, but vertical ones (which only appear with small dialog height), yes.
> 
> It works with
> 
> ```
> self.ui.example_filename_before.verticalScrollBar().valueChanged.connect(
>     self.ui.example_filename_after.verticalScrollBar().setValue)
> 
> self.ui.example_filename_after.verticalScrollBar().valueChanged.connect(
>     self.ui.example_filename_before.verticalScrollBar().setValue)
> ```
> 
> But if you move the horizontal bar fast enough, it creates this weirdness where part of the text simply dissapears (it gets fixed by scrolling vertically or waiting a bit).

I think you need to block signals before changing values, see https://stackoverflow.com/questions/50794761/how-to-make-synchronized-scroll-bar-in-qlistwidget-in-pyqt4 for an example.


> 
> > About the button, perhaps move it on the same line as the ones to show documentation and default, not ideal, but it will give a bit more vertical space for the display.
> 
> Like this? (had to cut down the text)

Yes

> 
> Looks disorienting to me

I meant scroll to the right (sorry, not align)

Gabrielcarvfer:
> I think you need to block signals before changing values, see https://stackoverflow.com/questions/50794761/how-to-make-synchronized-scroll-bar-in-qlistwidget-in-pyqt4 for an example.

Yup, you're completely correct on that. 

> I meant scroll to the right (sorry, not align)

Hmm, I tried doing that, but doesn't work as intended. It only scrolls to right when the window is already painted. It doesn't work during initialization because the field width was not determined yet. Tried changing scrolling from left-to-right to right-to-left, but just reverses the scroller.

```
def _scroll_examples_to_right(self):
        # maximum() = 0 when the window just started and is bound by where the user is looking at the view 
       # (depending on the vertical scroll, local horizontal maximum might not be global horizontal maximum)
       # a.k.a. it doesn't completely scroll to right  if the largest entry is not displayed to the user
        self.ui.example_filename_before.horizontalScrollBar().setValue(
            self.ui.example_filename_before.horizontalScrollBar().maximum()) 
        self.ui.example_filename_after.horizontalScrollBar().setValue(
            self.ui.example_filename_after.horizontalScrollBar().maximum())
```

Sophist-UK:
> I wonder though about the layout, because with long paths, and small options window, having 2 columns doesn't help.

I agree - having 2 columns makes displaying long paths difficult on smaller screens that yours.

Could we use "From:" and "To:" instead, perhaps like this

Without "Move" on the same line:
```
From: TICKET_TO_RIDE.mp3 To: ticket_to_ride.mp3
```

With "Move" on separate lines:
```
From: /mnt/dev/Loved_Tracks_part4/Chilly Gonzales/Chilly Gonzales - You Can Dance/0924 Chilli Gonzales - You Can Dance (Edwin Van Cleef remix).mp3
To: /home/gabri/Music/Chilly Gonzales/Chilly Gonzales - You Can Dance/0924 Chilli Gonzales - You Can Dance (Edwin Van Cleef remix).mp3
```

I would also like to propose a change to the selection of tracks when you click the button - in that, if the user has albums or tracks selected in the Album list, then the random tracks should be chosen from those albums / tracks selected - and if none are selected then they should be random from all loaded tracks as currently. The rationale for this is that if you have a significant number of albums loaded and being saved, and you see one album where the naming is problematical, you really don't want to have to click the button repeatedly until you get one of the problem files listed - far easier to select the problem album and click "Pick random examples from selected / loaded files".

zas:
> Hmm, I tried doing that, but doesn't work as intended. It only scrolls to right when the window is already painted. It doesn't work during initialization because the field width was not determined yet. Tried changing scrolling from left-to-right to right-to-left, but just reverses the scroller.

Yes, it seems messy, don't bother with that now.

Gabrielcarvfer:
> I would also like to propose a change to the selection of tracks when you click the button - in that, if the user has albums or tracks selected in the Album list, then the random tracks should be chosen from those albums / tracks selected - and if none are selected then they should be random from all loaded tracks as currently. 

Probably needs more testing, but seems to work fine.

> I agree - having 2 columns makes displaying long paths difficult on smaller screens that yours.
> 
> Could we use "From:" and "To:" instead, perhaps like this

I would prefer something like inline diff with red/green coloring (just like in the GitKraken UI), but I'm a disaster with GUIs...
I really don't know what would be better here. I think it is reasonable the way it is, as the window can always be resized and most people will probably have a 16:9 display anyways.

zas (dismissed):
@Gabrielcarvfer don't bother too much about the UI, the feature is good enough to be merged. We can always improve this later.

phw (changes requested):
I like the idea, but I don't like the UI very much.

That renaming page is already very cramped, especially on screens with limited vertical space. With the changes I have about 4 lines of available space for the renaming script, that makes it difficult to use.

I also don't find the two column layout very clear, especially if it is showing me examples of a single album and all I see is the same long path. With the default examples I have two distinct cases (album vs. compilation), which allows me to tweak my file naming script to those cases. Now if I have a single release loaded I see many mostly identical examples.

I currently have no really good idea, but I think we should redesign this page. Maybe we can somehow put elements into separate views without making it too confusing.

Gabrielcarvfer:
> Triggered an exception when opening the options, not sure why. I had opened it before, closed it again, loaded another release, then it crashed with an endless recursion when I opened options again. Currently can't look into it in more detail, but here is the exception:
> ```

Just found it. Happens when you select an empty cluster and try to open the options.

rdswift:
Rather than the two columns, since the first column is static (for the currently selected records to process), perhaps the source file / path could be displayed as a tooltip when hovering over one of the lines?  Then there could be only one column, displaying the result of the script.

Sophist-UK:
> @Gabrielcarvfer don't bother too much about the UI, the feature is good enough to be merged. We can always improve this later.

@zas I don't have any issues with this idea in principle - so I am not absolutely sure why I am writing this comment - except that it seems to go against the GIT concept of always being release ready. (Presumably one concern being that the later improvements might not come and we might be left with a bad UI in a release.) Or perhaps my concern stems from the idea of merging a PR that several people other people have raised concerns about.

But then again, small PRs which take you from a point of stability to another point of stability can often be better than bigger PRs that are less easy to review.

In the end, you and Philipp are the custodians of the repo, so it is your decision and definitely not my place to question.

Sophist-UK:
> I would prefer something like inline diff with red/green coloring (just like in the GitKraken UI), but I'm a disaster with GUIs...

Yes - I would be happy with using Red/Green colouring to show old/new, but then again I am not Red/Green colour blind. :-)

Sophist-UK:
> Rather than the two columns, since the first column is static (for the currently selected records to process), perhaps the source file / path could be displayed as a tooltip when hovering over one of the lines? Then there could be only one column, displaying the result of the script.

We don't display the source now (because they are theoretical examples without a source), and it works - after all the whole point of file naming scripts is that the new result is what is important.

So - yes - a tooltip might be an excellent way of showing the existing file name / path.

zas:
What I mean is we can always improve the UI, but as is it is already an improvement, and if stable, it can be merged.
Improvements can come in future PRs.
Let's keep things simple and small, merge, improve.

Gabrielcarvfer:
> Rather than the two columns, since the first column is static (for the currently selected records to process), perhaps the source file / path could be displayed as a tooltip when hovering over one of the lines? 

Just tried doing that, the tooltip refused to show up. Searched up for solutions, but nothing seemed to work (increasing time, letting user select the item, changing focus settings, trying to trigger it manually by catching the a click event, etc). ¯\\(°_o)/¯

rdswift:
Darn.  Thanks for looking at it.

Gabrielcarvfer:
I hit gold:
```
Note that by default tooltips are only shown for widgets that are children of the active window. You can change this behavior by setting the attribute Qt::WA_AlwaysShowToolTips on the window, not on the widget with the tooltip.
```

phw:
> except that it seems to go against the GIT concept of always being release ready.

> What I mean is we can always improve the UI, but as is it is already an improvement, and if stable, it can be merged.

I kind of agree with not @Sophist-UK

Gabrielcarvfer:
It looks way more confusing, but here it is.

rdswift:
> It looks way more confusing, but here it is.

Sorry to hear that.  "confusing" != "good", especially with Picard.  It is complicated enough that new users (and some of us old farts) are already confused.  Thanks for trying it.

zas:
> It looks way more confusing, but here it is.

I agree, it's awful.
To be frank, I would have keep UI improvements for later, initial version was working at least.

Gabrielcarvfer:
> Could we agree on leaving the naming example improvements unmerged until we have the 2.6.1 or even 2.6.2 out and target it for 2.7?

I wouldn't mind that. I'm not trying to rush anything. :)

zas:
> Could we agree on leaving the naming example improvements unmerged until we have the 2.6.1 or even 2.6.2 out and target it for 2.7?

 Of course, it doesn't have to be merged for 2.6.1, but initial (2 columns) revised patch was working, what's wrong with it? It's sub-optimal, but that's more due to our Options dialog. Pushing for changes in UI without having a clear idea of what we want is just a waste of time.
So, I'm for keeping this PR small and simple, we all agree the UI isn't perfect, but no one came with a good idea yet. Perhaps users will suggest improvements to this feature.

About plans: 2.6.x should include mainly bug fixes and performance optimizations, we can start a 2.7 branch and merge new features there.

picard/ui/options/renaming.py
```diff
@@ -173,22 +191,49 @@ def _example_to_filename(self, file):
...
-        self.ui.example_filename.setText(example1)
-        self.ui.example_filename_va.setText(example2)
+        self.ui.example_filename_before.clear()
+        self.ui.example_filename_after.clear()
+
+        if not self._sampled_example_files:
+            self._sample_example_files()
+        example_files = self._sampled_example_files
+
+        for example in example_files:
```
zas, line 233: I would just sort the list by _after_ path, so the list isn't shuffled too much when renewing, and the list is still in ordered track number (title01 before ttile02), this sort may not be correct in all cases, but that's an improvement imho:

```python
        examples = [self._example_to_filename(example) for example in example_files]
        for before, after in sorted(examples, key=lambda x: x[1]):
            self.ui.example_filename_before.addItem(before)
            self.ui.example_filename_after.addItem(after)
```

What do you think?

phw: Yes, the random shuffling does not help I think. I think it would be better with the sorting as suggested.

phw (approved):
I created the 2.7.x branch and updated the PR to merge into this.

Otherwise I'm happy with this. I still think we need to improve the UI before having this in any release, but the issues here are more general and not really specific to the PR. Having the examples take up more space just makes the issues more visible.

So far not good idea. I kind of think the editor window + examples need more space, which means they probably should be moved to their own page. But this moves them away from the toggles to enable /disable renaming/moving. Anyway, let's tackle this separately.

picard/ui/options/renaming.py
Gabrielcarvfer: I agree. Also, I was testing this and noticed something weird. %title% doesn't seem to be returning the %title% tag, but the filename (at least on the displayed examples). I'm currently looking into it.

zas (approved):

picard/ui/options/renaming.py
```diff
@@ -126,6 +127,23 @@ def __init__(self, parent=None):
...
+            self.ui.example_filename_before.verticalScrollBar().setValue(value)
+            self.ui.example_filename_after.verticalScrollBar().setValue(value)
+            self.ui.example_filename_before.blockSignals(False)
+            self.ui.example_filename_after.blockSignals(False)
+
+        self.ui.example_filename_before.verticalScrollBar().valueChanged.connect(
+            _sync_toolbar_vert)
+
+        self.ui.example_filename_after.verticalScrollBar().valueChanged.connect(
+            _sync_toolbar_vert)
```
zas, line 146: I think `_sync_toolbar_vert` was meant to be `_sync_scroolbar_vert`.
Also there is no need to block signals and set value of the scrollbar emiting the signal.


We could make it more generic and working for more than 2 scrollbars:

```python
        def sync_vertical_scrollbars(widgets):
            """Sync position of vertical scrollbars for listed widgets"""
            def _sync_scrollbar_vert(widget, value):
                widget.blockSignals(True)
                widget.verticalScrollBar().setValue(value)
                widget.blockSignals(False)

            widgets = set(widgets)
            for widget in widgets:
                for other in widgets - {widget}:
                    widget.verticalScrollBar().valueChanged.connect(
                        partial(_sync_scrollbar_vert, other))

        # Sync example lists vertical scrolling
        sync_vertical_scrollbars((self.ui.example_filename_before, self.ui.example_filename_after))
```

`sync_vertical_scrollbars()` could be then moved to `picard/ui/util.py`

zas (changes requested):

phw:
@Gabrielcarvfer FYI: I rebased this PR to trigger re-analysis by Codacy, as we had to need to setup Codacy again.

zas (approved):

zas:
> I kind of think the editor window + examples need more space, which means they probably should be moved to their own page.

I see another reason to move them away from Options pages: it would be convenient to change the selection and see the result, which isn't possible right now with Options dialog.
While testing, I noticed I would like to have the possibility to change which files are used for examples without closing/re-opening the dialog.

Just few thoughts.

phw:
Ok, let's merge this into the branch for 2.7. And we need to come up with some ideas to restructure the renaming options properly. I added PICARD-2179 to track this for now.

rdswift:
Because of the real estate concerns, perhaps the naming script editor (including the examples) could be a separate sub-page of the file renaming options page, with only a read-only copy of the script available for viewing on the main file renaming options page?  That way the file renaming options are still all together, but only the script editor and examples would be on the script editor sub-page, allowing more room for the editor.

PR #1779, opened 2021-04-02
PICARD-2150: Allow user to hide metadata view

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2150

The metadata view can take quite a lot of space, especially on small screens. While it can be resized sometimes it would be good to have it out of view. In the past users could completely hide it by dragging it down. We disabled this as users accidentally lost access to this view without a proper UI hint how to get it back.

# Solution
Add an action to the View menu to show/hide the metadata view, together with a keyboard shortcut.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -229,13 +229,11 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.panel.insertWidget(0, self.file_browser)
         self.panel.restore_state()
 
-        self.metadata_box = MetadataBox(self)
-        self.cover_art_box = CoverArtBox(self)
-        self.show_cover_art()
-
         self.log_dialog = LogView(self)
         self.history_dialog = HistoryView(self)
 
+        self.metadata_box = MetadataBox(self)
+        self.cover_art_box = CoverArtBox(self)
         metadata_view_layout = QtWidgets.QHBoxLayout()
         metadata_view_layout.setContentsMargins(0, 0, 0, 0)
         metadata_view_layout.setSpacing(0)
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -1296,6 +1307,15 @@ def refresh_metadatabox(self):
         self.tagger.window.metadata_box.selection_dirty = True
         self.tagger.window.metadata_box.update()
 
+    def show_metadata_view(self):
+        """Show/hide the cover art box."""
+        show = self.show_metadata_view_action.isChecked()
+        if show:
+            self.metadata_view.show()
+        else:
+            self.metadata_view.hide()
```
zas, line 1316: I think you can replace the whole `if/else` with:

```python
self.metadata_view.setVisible(show)
```

picard/ui/mainwindow.py
```diff
@@ -1296,6 +1307,15 @@ def refresh_metadatabox(self):
         self.tagger.window.metadata_box.selection_dirty = True
         self.tagger.window.metadata_box.update()
 
+    def show_metadata_view(self):
+        """Show/hide the cover art box."""
```
zas, line 1311: Update the comment as it hides the whole metadata box

zas (changes requested):
A nice improvement :)

Sophist-UK:
I haven't had the chance to check the code, but for performance reasons do make sure that the metadata window is not updated when you change selections if it is hidden, and ensure that the metadata is updated to match selections when you make it visible again.

phw:
Good point. The current patch does not prevent updates, but we could keep updates to both metadata and cover art while the view is hidden. I'll update the PR

phw:
I implemented skipping updating the metadata view and cover art box if they are hidden. This actually would already have been useful previously for the cover art box only. Now it skips updating both the metadata view and the cover art box accordingly.

zas (approved):

PR #1780, opened 2021-04-03
PICARD-2174: Optimize implementation to display format specific metadata

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2174, PICARD-905

# Solution

Replace File.new_metadata with a method File.format_specific_metadata which returns a single key formatted for display in format specific way.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -157,7 +157,7 @@ class File(QtCore.QObject, Item):
         return '<%s %r>' % (type(self).__name__, self.base_filename)
 
     # pylint: disable=no-self-use
-    def format_specific_metadata(self, metadata, tag, settings=None):
+    def format_specific_metadata(self, metadata, tag, settings_cache=None):
         """Can be overridden to customize how a tag is displayed in the UI.
         This is useful if a tag saved to the underlying format will differ from
         the internal representation in a way that would cause data loss. This is e.g.
--- a/picard/file.py
+++ b/picard/file.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2004 Robert Kaye
 # Copyright (C) 2006-2009, 2011-2013, 2017 Lukáš Lalinský
-# Copyright (C) 2007-2011, 2015, 2018-2020 Philipp Wolfer
+# Copyright (C) 2007-2011, 2015, 2018-2021 Philipp Wolfer
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008-2009 Nikolai Prokoschenko
 # Copyright (C) 2009 Carlin Mangar
@@ -157,7 +157,20 @@ class File(QtCore.QObject, Item):
         return '<%s %r>' % (type(self).__name__, self.base_filename)
 
     # pylint: disable=no-self-use
-    def format_specific_metadata(self, metadata, tag):
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        """Can be overridden to customize how a tag is displayed in the UI.
+        This is useful if a tag saved to the underlying format will differ from
+        the internal representation in a way that would cause data loss. This is e.g.
+        the case for some ID3v2.3 tags.
+
+        Args:
+            metadata: The metadata object to read the tag from
+            tag: Name of the tag
+            settings: Dictionary of settings. If not set, config.setting should be used
+
+        Returns:
+            An array of values for the tag
+        """
         return metadata.getall(tag)
 
     def load(self, callback):
--- a/picard/file.py
+++ b/picard/file.py
@@ -16,7 +16,7 @@
 # Copyright (C) 2012-2014 Wieland Hoffmann
 # Copyright (C) 2013 Calvin Walton
 # Copyright (C) 2013-2014 Ionuț Ciocîrlan
-# Copyright (C) 2013-2014, 2017 Sophist-UK
+# Copyright (C) 2013-2014, 2017, 2021 Sophist-UK
 # Copyright (C) 2013-2014, 2017-2019 Laurent Monin
 # Copyright (C) 2016 Rahul Raturi
 # Copyright (C) 2016 Ville Skyttä
@@ -156,9 +156,9 @@ class File(QtCore.QObject, Item):
     def __repr__(self):
         return '<%s %r>' % (type(self).__name__, self.base_filename)
 
-    @property
-    def new_metadata(self):
-        return self.metadata
+    # pylint: disable=no-self-use
+    def format_specific_metadata(self, metadata, tag):
+        return metadata.getall(tag)
 
     def load(self, callback):
         thread.run_task(
@@ -370,14 +370,14 @@ class File(QtCore.QObject, Item):
             temp_info = {}
             for info in FILE_INFO_TAGS:
                 temp_info[info] = self.orig_metadata[info]
-            images_changed = self.orig_metadata.images != self.new_metadata.images
+            images_changed = self.orig_metadata.images != self.metadata.images
             # Data is copied from New to Original because New may be
             # a subclass to handle id3v23
             config = get_config()
             if config.setting["clear_existing_tags"]:
-                self.orig_metadata.copy(self.new_metadata)
+                self.orig_metadata.copy(self.metadata)
             else:
-                self.orig_metadata.update(self.new_metadata)
+                self.orig_metadata.update(self.metadata)
             # After saving deleted tags should no longer be marked deleted
             self.metadata.clear_deleted()
             self.orig_metadata.clear_deleted()
@@ -630,22 +630,21 @@ class File(QtCore.QObject, Item):
         return self.similarity == 1.0 and self.state == File.NORMAL
 
     def update(self, signal=True):
-        new_metadata = self.new_metadata
-        names = set(new_metadata.keys())
-        names.update(self.orig_metadata.keys())
+        metadata = self.metadata
+        names = set(metadata) | set(self.orig_metadata)
         config = get_config()
         clear_existing_tags = config.setting["clear_existing_tags"]
         ignored_tags = config.setting["compare_ignore_tags"]
         for name in names:
             if (not name.startswith('~') and self.supports_tag(name)
                 and name not in ignored_tags):
-                new_values = new_metadata.getall(name)
+                new_values = metadata.getall(name)
                 if not (new_values or clear_existing_tags
-                        or name in new_metadata.deleted_tags):
+                        or name in metadata.deleted_tags):
                     continue
                 orig_values = self.orig_metadata.getall(name)
                 if orig_values != new_values:
-                    self.similarity = self.orig_metadata.compare(new_metadata, ignored_tags)
+                    self.similarity = self.orig_metadata.compare(metadata, ignored_tags)
                     if self.state == File.NORMAL:
                         self.state = File.CHANGED
                     break
--- a/picard/file.py
+++ b/picard/file.py
@@ -157,7 +157,7 @@ class File(QtCore.QObject, Item):
         return '<%s %r>' % (type(self).__name__, self.base_filename)
 
     # pylint: disable=no-self-use
-    def format_specific_metadata(self, metadata, tag, settings_cache=None):
+    def format_specific_metadata(self, metadata, tag, settings=None):
         """Can be overridden to customize how a tag is displayed in the UI.
         This is useful if a tag saved to the underlying format will differ from
         the internal representation in a way that would cause data loss. This is e.g.
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -56,6 +56,7 @@ from picard.formats.mutagenext import (
 )
 from picard.metadata import Metadata
 from picard.util import (
+    cached_settings,
     encode_filename,
     sanitize_date,
 )
@@ -650,16 +651,12 @@ class ID3File(File):
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
 
-    def format_specific_metadata(self, metadata, tag, settings=None):
-        if not settings:
-            config = get_config()
-            settings = {
-                "write_id3v23": config.setting["write_id3v23"],
-                "id3v23_join_with": config.setting["id3v23_join_with"],
-            }
+    def format_specific_metadata(self, metadata, tag, settings_cache=None):
+        if 'write_id3v23' not in settings_cache or 'id3v23_join_with' not in settings_cache:
+            cached_settings(('write_id3v23', 'id3v23_join_with'), settings_cache)
 
-        if not settings["write_id3v23"]:
-            return super().format_specific_metadata(metadata, tag, settings)
+        if not settings_cache["write_id3v23"]:
+            return super().format_specific_metadata(metadata, tag, settings_cache)
 
         values = metadata.getall(tag)
         if not values:
@@ -674,7 +671,7 @@ class ID3File(File):
         # unless it's TIPL or TMCL which can still be multi-valued.
         if (len(values) > 1 and tag not in ID3File._rtipl_roles
                 and not tag.startswith("performer:")):
-            join_with = settings["id3v23_join_with"]
+            join_with = settings_cache["id3v23_join_with"]
             values = [join_with.join(values)]
 
         return values
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006-2009, 2011-2012 Lukáš Lalinský
-# Copyright (C) 2008-2011, 2014, 2018-2020 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014, 2018-2021 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2011-2012 Johannes Weißl
 # Copyright (C) 2011-2014 Michael Wiencek
@@ -650,12 +650,16 @@ class ID3File(File):
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
 
-    def format_specific_metadata(self, metadata, tag):
-        config = get_config()
-        if not config.setting["write_id3v23"]:
-            return super().format_specific_metadata(metadata, tag)
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        if not settings:
+            config = get_config()
+            settings = {
+                "write_id3v23": config.setting["write_id3v23"],
+                "id3v23_join_with": config.setting["id3v23_join_with"],
+            }
 
-        join_with = config.setting["id3v23_join_with"]
+        if not settings["write_id3v23"]:
+            return super().format_specific_metadata(metadata, tag, settings)
 
         values = metadata.getall(tag)
         if not values:
@@ -670,6 +674,7 @@ class ID3File(File):
         # unless it's TIPL or TMCL which can still be multi-valued.
         if (len(values) > 1 and tag not in ID3File._rtipl_roles
                 and not tag.startswith("performer:")):
+            join_with = settings["id3v23_join_with"]
             values = [join_with.join(values)]
 
         return values
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -10,7 +10,7 @@
 # Copyright (C) 2011-2014 Wieland Hoffmann
 # Copyright (C) 2013 Calvin Walton
 # Copyright (C) 2013-2014, 2017-2019 Laurent Monin
-# Copyright (C) 2013-2015, 2017 Sophist-UK
+# Copyright (C) 2013-2015, 2017, 2021 Sophist-UK
 # Copyright (C) 2015 Frederik “Freso” S. Olesen
 # Copyright (C) 2016 Christoph Reiter
 # Copyright (C) 2016-2018 Sambhav Kothari
@@ -650,36 +650,29 @@ class ID3File(File):
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
 
-    @property
-    def new_metadata(self):
+    def format_specific_metadata(self, metadata, tag):
         config = get_config()
         if not config.setting["write_id3v23"]:
-            return self.metadata
-
-        copy = Metadata()
-        copy.copy(self.metadata)
+            return super().format_specific_metadata(metadata, tag)
 
         join_with = config.setting["id3v23_join_with"]
-        copy.multi_valued_joiner = join_with
 
-        for name, values in copy.rawitems():
-            # ID3v23 can only save TDOR dates in YYYY format. Mutagen cannot
-            # handle ID3v23 dates which are YYYY-MM rather than YYYY or
-            # YYYY-MM-DD.
-            if name == "originaldate":
-                values = [v[:4] for v in values]
-            elif name == "date":
-                values = [(v[:4] if len(v) < 10 else v) for v in values]
+        values = metadata.getall(tag)
+        if not values:
+            return values
 
-            # If this is a multi-valued field, then it needs to be flattened,
-            # unless it's TIPL or TMCL which can still be multi-valued.
-            if (len(values) > 1 and name not in ID3File._rtipl_roles
-                    and not name.startswith("performer:")):
-                values = [join_with.join(values)]
+        if tag == "originaldate":
+            values = [v[:4] for v in values]
+        elif tag == "date":
+            values = [(v[:4] if len(v) < 10 else v) for v in values]
 
-            copy[name] = values
+        # If this is a multi-valued field, then it needs to be flattened,
+        # unless it's TIPL or TMCL which can still be multi-valued.
+        if (len(values) > 1 and tag not in ID3File._rtipl_roles
+                and not tag.startswith("performer:")):
+            values = [join_with.join(values)]
 
-        return copy
+        return values
 
 
 class MP3File(ID3File):
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -651,12 +651,12 @@ class ID3File(File):
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
 
-    def format_specific_metadata(self, metadata, tag, settings_cache=None):
-        if 'write_id3v23' not in settings_cache or 'id3v23_join_with' not in settings_cache:
-            cached_settings(('write_id3v23', 'id3v23_join_with'), settings_cache)
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        if not settings:
+            settings = cached_settings(('write_id3v23', 'id3v23_join_with'))
 
-        if not settings_cache["write_id3v23"]:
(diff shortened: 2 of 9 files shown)


Sophist-UK:
@phw Thanks!!! I assume that this means you think my untested code was OK. :-)

phw:
> I assume that this means you think my untested code was OK. :-)

Yes, absolutely. Just did a correction to the functions declaration in id3.py, the rest I just copied over from your freehand code :)

But some odd experience with the metadata box now. Updating it actually takes longer now then before. The good part is is that there is no noticeable difference between ID3 v2.3 and 2.4 as before, but both have a longer delay. I tested with ~4k files, selecting all.

I find this odd, because the ID3 v2.4 case didn't really change much. There is just a single function call in between (which is of course done for each individual tag, but I am surprised this small indirection is so noticeable).

Needs more work, but I need to go now and can't look into the details.

phw:
ok, https://github.com/metabrainz/picard/pull/1780/commits/e55e12f06d9c95277d1620d49c8092f6b332e62c addresses the issue. ` config.setting['foo']` is just slow, should be avoided in code that gets called a lot.

I really need to go now, seriously :) Please someone test this

Sophist-UK:
Before we we calling new_metadata once per file.

Now we are calling format_specific_metadata once per tag per file. So 1-2 orders of magnitude more calls.

Just have to be careful what happens if someone changes from v2.3 to v2.4 without restarting Picard if we are caching these settings without having any safeguards to invalidate the cache if the settings are changed.

picard/ui/metadatabox.py
```diff
@@ -585,14 +585,20 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
         top_tags = config.setting['metadatabox_top_tags']
         top_tags_set = set(top_tags)
 
+        # format_settings = {
+        #     "write_id3v23": config.setting["write_id3v23"],
+        #     "id3v23_join_with": config.setting["id3v23_join_with"],
+        # }
+        format_settings = None
```
Sophist-UK, line 592: I don't think you can replace a call by reference object with a different object and have it passed back. So that you keep the same object, I think you need to set this to an empty dict rather than None and change the code in format_specific_metadata to something like:
```
    def format_specific_metadata(self, metadata, tag, settings={}):
        if not settings:
            config = get_config()
            settings["write_id3v23"] = config.setting["write_id3v23"]
            settings["id3v23_join_with"] = config.setting["id3v23_join_with"]
```

Sophist-UK: I tested this in a sandbox with:
```
variable = None
def change(v):
    v = {"item": "text"}
change(variable)
print(variable)
```
and got `None` printed.

Sophist-UK (changes requested):

picard/ui/metadatabox.py
phw: Not sure I follow you. Actually you should not set the default parameter to an object, as this will create an object only once. You would always manipulate the same object on later calls. Instead creating the object when it is empty makes sure the object does exist only for that call.

Sophist-UK: Try my example and see why it will pass None in every time and get settings every time.

You need to set an empty Dict object in the calling routine and add items to it in the called routine in order that the added items are passed back to the calling routine and then passed into the called routine on the second call - thus avoiding the expensive config settings calls.

phw: Ah, now I see what you mean. No, that is not supposed to pass anything back. That just happens because in the caller the code to set the config is commented out. I only did this for debugging, I'll update that :)

Gabrielcarvfer:
Times for 18k files

<details>
<summary>Current ID3v23 </summary>

idv3
</details>

<details>
<summary>Current ID3v24 </summary>

no_id3v23
</details>

<details>
<summary>This branch so far</summary>

image
</details>

Even shaving all time spent on config.setting, it will still take around 4s, which is double the v24 time.

I think the way to go is either cache the ID3v23 tags while loading (not a huge fan, but it can be offloaded along with mutagen load to a different process - I'm sidetracking again, but I was playing with it and it makes Picard butter smooth). 

Ideally, I would say to convert the v23 tags into v24 during load and then right back to v23 during saving if they change (and write v23 tags is enabled).

picard/formats/id3.py
```diff
@@ -650,36 +650,34 @@ def _save_tags(self, tags, filename):
...
-            copy[name] = values
-
-        return copy
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        if not settings:
+            config = get_config()
+            settings = {
+                "write_id3v23": config.setting["write_id3v23"],
+                "id3v23_join_with": config.setting["id3v23_join_with"],
+            }
```
zas, line 659: Untested but such help function could simplify things if we are going to do that more often:

```python
def cached_settings(keys, config=None):
    if config is None:
        config = get_config()
    return {key: config.setting[key] for key in keys}


settings = cached_settings(["write_id3v23", "id3v23_join_with"])
```

picard/ui/metadatabox.py
Sophist-UK: I don't think the code should be set in the calling routing - it should just pass an empty dict object - the contents of the dict object should be set inside format_specific_mnetadata because: a) that avoids code duplication and b) f_s_m knows what settings it needs to cache.

picard/formats/id3.py
Sophist-UK: @zas I definitely like this idea. This does mean you need to run this code in the calling routine rather than passing an empty dict in the called routine. 

I would prefer to see the following code:
```
def cached_settings(keys, config=None, settings={}):
    if config is None:
        config = get_config()
    for key in keys:
         settings[key] = config.setting[key]
    return settings
```
which can then be used in a calling routine like this:
```
settings = cached_settings(["write_id3v23", "id3v23_join_with"])
```
or in a called routine like this:
```
fsm_settings = {}
x = format_specific_metadata(..., fsm_settings)

def format_specific_metadtata(..., settings={}):
    if not settings:
        settings = cached_settings(["write_id3v23", "id3v23_join_with"], settings)  #or
        cached_settings(["write_id3v23", "id3v23_join_with"], settings)
```

picard/ui/metadatabox.py
```diff
@@ -585,14 +585,19 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
         top_tags = config.setting['metadatabox_top_tags']
         top_tags_set = set(top_tags)
 
+        format_settings = {
```
Sophist-UK, line 588: As per previous comment, since it is f_s_m that knows what settings it needs, and to prevent code duplication, I think this should be `format_settings = {}` or even `format_settings_cache` and this (mutable dict) object should then be populated inside `format_specific_metadata`.

Sophist-UK (changes requested):

phw:
I did some benchmarking. We have three cases:
1. Selecting ID3 files with ID3v2.4 enabled
2. Selecting ID3 files with ID3v2.3 enabled
3. Selecting other file formats (no call to the ID3 functions)

Below are the times in seconds spent in `MetadataBox._update_tags` when selecting 4145 files on my system

  | ID3v2.4 | ID3v2.3 | Other formats
-- | -- | -- | --
master | 0.524680978 | 1.059419109 | 0.4853345462
this PR | 0.611909558 | 0.6093371344 | 0.528998111

The PR brings a clear performance improvement for the ID3v2.3 use case, but  is a bit slower for 2.4 and slightly slower for other formats. The positive side is that 2.3 and 2.4 now basically take the same time (as expected). In my timings above 2.4 is even slightly slower, but that is just fluctuation in the measuring. For other formats it is one additional function call per tag, the effect is not so bad.

IMHO overall this is a win. But probably should be tested with more files.

> Ideally, I would say to convert the v23 tags into v24 during load and then right back to v23 during saving if they change (and write v23 tags is enabled). Not sure if this is possible though.

That is exactly how loading / saving files work at the moment. What this is about is visualizing the tags as they will be written in the metadatabox. I actually think given the results above we should not invest too much effort to cache those few affected tags for the ID3v2.3 case. I think the complication of caching those values and keeping the cache up-to-date is not worth it. Likely we can achieve larger gains by looking at the metadata summarizing more general. E.g. Sophist suggested elsewhere to keep caches of partial results for releases. This also involves tracking metadata changes and updating caches, but if that is solved it likely gives much better results then just caching the few 2.3 formatted tags.

picard/ui/metadatabox.py
phw: I don't like this passing the settings back, that makes IMHO for a pretty confusing interface. I'd usually not expect a function to manipulate the passed settings. It also makes validating the settings more complicated. In the current case the function can just check if settings is not None, then initialize it. In the other case it needs to check for each key. The goal is exactly to keep this settings handling basically out of the function.

For the same reason I would not introduce another function to handle that. We try to reduce the call complexity here.

Sophist-UK: > I don't like this passing the settings back, that makes IMHO for a pretty confusing interface.

I don't think this is any more confusing than requiring someone to know what settings are needed and duplicating the code.

>I'd usually not expect a function to manipulate the passed settings.

I agree that this is less intuitive - but same comment as above.

> It also makes validating the settings more complicated. 

It depends. The code I have suggested is no more complicated than current - but it doesn't validate that the settings include what is needed - because it assumes that an empty dict is passed in. But if you want to validate it would be a one line change:
```
 def format_specific_metadata(self, metadata, tag, settings={}):
        if ("write_id3v23" not in settings or "id3v23_join_with" not in settings):
            config = get_config()
            settings["write_id3v23"] = config.setting["write_id3v23"]
            settings["id3v23_join_with"] = config.setting["id3v23_join_with"]
```
> In the current case the function can just check if settings is not None, then initialize it. In the other case it needs to check for each key. The goal is exactly to keep this settings handling basically out of the function.

In that case, settings should NOT be given a default of None.
 
> For the same reason I would not introduce another function to handle that. We try to reduce the call complexity here.

I agree we should **not** introduce another function call if it gets called once per tag per file. But if it is only called once per metadatabox update, what is the problem of having a helper function? If the new thread-safe config functions have a performance impact, then IMO there needs to be a mitigation of that provided as a helper.

phw: Ok, I gave this a try with passing back the config values back to a cache object. Probably the biggest advantage is indeed avoiding duplication. Also we could have different implementations of `File.format_specific_metadata` that will need different config values.

This PR needs some tests added for both the new `cached_settings()`  and `File.format_specific_metadata`, otherwise this seems to be fine to me.

>  If the new thread-safe config functions

It's not only the thread safe config functions, it is config access in general. `config.settings['foo']`  is comparable slow, that's nothing new. If you have config.setting access in the ID3 format_specific_metadata then the above use case for ID3v2.3 goes up from ~0.6 seconds to ~2.0 seconds on my system. If you additionally have `get_config`  call in it it goes up to ~2.2 seconds. That's big, but the biggest bottleneck is general config access.

Sophist-UK:
> Below are the times in seconds spent in `MetadataBox._update_tags` when selecting 4145 files on my system
> 
>  	ID3v2.4	ID3v2.3	Other formats
> master	0.524680978	1.059419109	0.4853345462
> this PR	0.611909558	0.6093371344	0.528998111
>
> IMHO overall this is a win. But probably should be tested with more files.

This PR is about UI response time - with c. 4k files and differences of << 1s in response time, it doesn't mean much. But if we assume c. 10x as many files i.e. c. 80k files, and measurements scale linearly (which may not be the case) then the table would look as follows:

|  | ID3v2.4 | ID3v2.3 | Other formats |
|---|---|---|---|
| master | 5.2s | 10.6s | 4.6s |
| this PR | 6.1s | 6.1s | 5.3s |

I am not sure how noticeable a <1s increase would be on a 5s response time, but I do think that a >4s reduction on a 10s response time would be noticeable. And definitely more consistent - though I think people typically stick to one format.

> > Ideally, I would say to convert the v23 tags into v24 during load and then right back to v23 during saving if they change (and write v23 tags is enabled). Not sure if this is possible though.

> That is exactly how loading / saving files work at the moment. ...

Ideally I believe that internal storage of tagging information should be format independent so that e.g. scripts work regardless of format - and as far as I know that is the case.

picard/util/__init__.py
```diff
@@ -708,3 +709,13 @@ def extract_year_from_date(dt):
             return parse(dt).year
     except (TypeError, ValueError):
         return None
+
+
+def cached_settings(keys, settings=None, config=None):
```
Sophist-UK, line 714: Do we need some comments that explain the settings and config parameters i.e. settings to update an existing dict, config to use an existing thread-safe Config object?

picard/util/__init__.py
```diff
@@ -708,3 +709,13 @@ def extract_year_from_date(dt):
...
+    if settings is None:
+        settings = {}
```
Sophist-UK, line 716: What is the difference between:
```
def cached_settings(keys, settings=None, config=None):
    if settings is None:
        settings = {}
```
and
```
def cached_settings(keys, settings={}, config=None):
```
except that the latter has 2 fewer lines of code and more clearly defines what type of object should be passed as settings.

phw: In the second case all calls to cached_settings without a settings parameter share the same object (because the {} as default gets created only once when the function is declared):

```python
s1 = cached_settings(['write_id3v23'])
print(s)  # e.g. {'write_id3v23': True}

s2 = cached_settings(['windows_compatibility'])
print(s)  # e.g. {'write_id3v23': True, 'windows_compatibility': False}

s1 == s2  # True
```

That could cause unwanted bugs if unrelated calls to this function share the same object.

picard/ui/metadatabox.py
Sophist-UK: > Ok, I gave this a try with passing back the config values back to a cache object. Probably the biggest advantage is indeed avoiding duplication. Also we could have different implementations of `File.format_specific_metadata` that will need different config values.

Good point - and we need to handle cases where the metadata being summarised includes a variety of different formats. And I think the current implementation does that.

> This PR needs some tests added for both the new `cached_settings()` and `File.format_specific_metadata`, otherwise this seems to be fine to me.

Thanks to @phw Philipp for taking this PR on from my initial code - I really **must** try to find the time to set up a development environment so that Philipp doesn't have to cope with my comments from the sidelines and generating additional tests that I haven't provided.

> > If the new thread-safe config functions
> 
> It's not only the thread safe config functions, it is config access in general. `config.settings['foo']` is comparable slow, that's nothing new. If you have config.setting access in the ID3 format_specific_metadata then the above use case for ID3v2.3 goes up from ~0.6 seconds to ~2.0 seconds on my system. If you additionally have `get_config` call in it it goes up to ~2.2 seconds. That's big, but the biggest bottleneck is general config access.

Hmmmm...

Perhaps what we need is a more general config cache that is applied automatically, with cache entries saved with a time-stamp and ignored/updated if the time-stamp is older than (say) 5 seconds. I cannot remember what exactly the Config issues was that required thread specific Config objects, but if we had a global cache, would that allow us to move back to a single global Config object and just deal with any issue of locking to prevent simultaneous updates between threads and / or multiple Picard instances?

picard/util/__init__.py
Sophist-UK: Ah - I didn't realise that the default values were static!!! I bet that is a common gotcha.

picard/file.py
```diff
@@ -630,22 +630,22 @@ def is_saved(self):
         return self.similarity == 1.0 and self.state == File.NORMAL
 
     def update(self, signal=True):
-        new_metadata = self.new_metadata
-        names = set(new_metadata.keys())
+        metadata = self.metadata
+        names = set(metadata.keys())
```
zas, line 634: I think both lines could be combined:

```python
names = set(metadata) + set(self.orig_metadata)
```

zas (changes requested):
It looks much better now

picard/file.py
phw: Updated that, but needs to use `|` operator not `+`

picard/util/__init__.py
phw: It is indeed. I think our linting on codacy would have complained loud and clear about this. There might be some rare cases where this behavior makes sense, but due to its surprising nature the general recommendation is to avoid this.

phw:
> Good point - and we need to handle cases where the metadata being summarised includes a variety of different formats. And I think the current implementation does that.

Actually the PR would not have worked well if there are different implementations of `format_specific_metadata` that access different keys of the settings_cache. If these implementations have code like:

```
if not settings_cache:
    cached_settings(('write_id3v23', 'id3v23_join_with'), settings_cache)
```

then only the first called implementation does actually initialize the settings. Following solutions come to mind:

1. Instead of `not settings_cache` each implementation needs to check for all used keys, as I have done in https://github.com/metabrainz/picard/pull/1780/commits/c909f6e3ca03000d0a57010da101238a2a61f527 . This is not very nice and cries for a helper function. Another function call on every call is something we try to avoid here.
2. Separate setting caches are maintained based on subclass of `File`. That makes handling in metadatabox.py more complicated, not very keen on this solution either
3. Going back to have the settings dict initializted by the caller instead of passing it back. To be honest I kind of favor this approach. If we are concerned about needing all keys we could create a complete dict with all settings. It would only run once per update.

zas:
>     3\. Going back to have the settings dict initializted by the caller instead of passing it back. To be honest I kind of favor this approach. If we are concerned about needing all keys we could create a complete dict with all settings. It would only run once per update.

I vote for this, at it seems simpler.

Sophist-UK:
> 3. Going back to have the settings dict initializted by the caller instead of passing it back. To be honest I kind of favor this approach. If we are concerned about needing all keys we could create a complete dict with all settings. It would only run once per update.

I am not in favour of this for reasons of code locality and duplication previously given.

I would be in favour of a more global approach to config settings caching that does NOT require any special coding, but I am unclear whether this would actually be possible. In the absence of a global cache, I believe that the current approach is the optimal one even if it seems a bit odd.

I do think that it would benefit from a couple of comment lines explaining the rationale and how to use it.

Gabrielcarvfer:
> That this is about is visualizing the tags as they will be written in the metadatabox. I actually think given the results above we should not invest too much effort to cache those few affected tags for the ID3v2.3 case.

I know nothing about the tags part, so for me it looks weird copying, to then process specific tags for a very specific format and then display it. Do the metadabox have/create fields exclusively for ID3v2.3 tags? (I never really tried tagging anything other than for testing, just blindly trust AcoustID and MB to update my library).

> Ideally I believe that internal storage of tagging information should be format independent so that e.g. scripts work regardless of format - and as far as I know that is the case.

Yup, that what I was thinking and expecting, that is why I don't understand why the special treatment for these v2.3 tags.

Sophist-UK:
> I know nothing about the tags part, so for me it looks weird copying, to then process specific tags for a very specific format and then display it. Do the metadabox have/create fields exclusively for ID3v2.3 tags? (I never really tried tagging anything other than for testing, just blindly trust AcoustID and MB to update my library).

Copying the metadata was (with hindsight) not a good choice of code.
 
> Yup, that what I was thinking and expecting, that is why I don't understand why the special treatment for these v2.3 tags.

ID3v2.3 has some constraints (like original date) - and when you save the file with ID3v2.3 tags, these constraints are actualised in the tags.

So the purpose of this code is to convert the format independent metadata so that for ID3 tagged files - with ID3v2.3 compatibility set - what is displayed in the metadatabox is what will be saved to the file.

Gabrielcarvfer:
> So the purpose of this code is to convert the format independent metadata so that for ID3 tagged files - with ID3v2.3 compatibility set - what is displayed in the metadatabox is what will be saved to the file.

Ipsis litteris (exactly as shown)? And if correct, is that really necessary? I can see why you would want that for testing, but not normal use. ._.

Sophist-UK:
> > So the purpose of this code is to convert the format independent metadata so that for ID3 tagged files - with ID3v2.3 compatibility set - what is displayed in the metadatabox is what will be saved to the file.
> 
> Ipsis litteris (exactly as shown)? And if correct, is that really necessary? I can see why you would want that for testing, but not normal use. ._.

IMO you would want to turn it off for testing (so you see what is held internally) and turn it on in normal use so you see what is saved to files.

phw:
> I am not in favour of this for reasons of code locality and duplication previously given.

I turned it around again, please check. No code duplication, as it passes the entire `config.setting` as a dict. The other way around is just to error prone and puts to much burden on the individual implementations. If we put the responsibility of setting up a settings cache into each implementation of `format_specific_metadata` which wants to handle the settings needs to be very careful to handle that correctly.

> I would be in favour of a more global approach to config settings caching that does NOT require any special coding, but I am unclear whether this would actually be possible.

We should add the patch from https://github.com/metabrainz/picard/commit/18c4e93cf1073e21857e4e39d311eb28d9d8f76f#diff-74efa3849396e3135d3796b5965e96c02fcc9dc1036c4b28bba71506193303d6 again. I had this removed in the refactoring of the settings when I removed the then obsolete locks also. I'll submit a separate PR for this.

I did some measuring again. Test case was again my 4k files with ID3v2.3, then selecting all of them. Remember that ID3 `format_specific_metadata` accesses settings two times.

As in the table above `MetadataBox._update_tags` still takes ~0.6 seconds for this PR (that is with passing a dict to `format_specific_metadata`).

If I change the code to pass `config.setting`  directly to `format_specific_metadata` then the time goes up to ~2 seconds. If I apply the patch from 18c4e93cf1073e21857e4e39d311eb28d9d8f76f I can get this down to ~1.3 seconds. So this caching helps mitigating the performance impact of settings access, but it still is preferable to not access individual settings in a loop, but do that outside the loop. 

> Ipsis litteris (exactly as shown)? And if correct, is that really necessary? I can see why you would want that for testing, but not normal use. ._.

The assumption users make is that they see what Picard will write to a file. If they see a date like `2021-04` but Picard only writes `2021`  this causes confusion. It causes even more confusion when loading the file back. Now the original value shows `2021`, the new value `2021-04`. This could either be displayed as a change in the metadata view (which would currently be the case), but actually it will not change anything when writing. One could also opt for not showing it as changed at all, but that would just likewise give a wrong impression (two columns showing different content, but being displayed as not changed).

picard/ui/metadatabox.py
```diff
@@ -585,14 +585,16 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
         top_tags = config.setting['metadatabox_top_tags']
         top_tags_set = set(top_tags)
 
+        settings = config.setting.as_dict()
+
         for file in files:
-            new_metadata = file.new_metadata
+            new_metadata = file.metadata
             orig_metadata = file.orig_metadata
             tags = set(list(new_metadata.keys()) + list(orig_metadata.keys()))
```
zas, line 593: Isn't this equivalent to:

```suggestion
            tags = set(new_metadata) | set(orig_metadata)
```

zas (changes requested):

picard/ui/metadatabox.py
phw: Thanks for fixing this! Works as expected ;)

zas (approved):

picard/file.py
```diff
@@ -156,9 +156,22 @@ def __init__(self, filename):
...
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        """Can be overridden to customize how a tag is displayed in the UI.
+        This is useful if a tag saved to the underlying format will differ from
+        the internal representation in a way that would cause data loss. This is e.g.
+        the case for some ID3v2.3 tags.
+
+        Args:
+            metadata: The metadata object to read the tag from
+            tag: Name of the tag
+            settings: Dictionary of settings. If not set, config.setting should be used
```
mineo, line 169: It's a bit weird that config.setting is recommended here, but the only overridden implementation uses cached_settings instead.

mineo (approved):

picard/file.py
phw: Which is just a helper function to read certain keys from config.setting into a dictionary. There is no need to use that, it was just wanted for avoiding code duplication for simar use cases.

The point here is that implementations shod be able to rely on the fact that the passed dict provides all the keys of config.setting, and if none is passed they should hence read the keys from config.setting, which the current one mentation does

phw: We don't necessarily need this helper, though. After the discussion basically went full circle I would also be fine with doing this in the ID3 implementation of format_specific_metadata again:

```python
def format_specific_metadata(self, metadata, tag, settings=None):
    if not settings:
        settings = get_config().setting
```

mineo: I'm ok with either of the following options:

1. Use `settings = get_config().setting` in the ID3 implementation. This is the "I read the docstring of File.format_specific_metadata and implemented the fallback verbatim" version :-)
2. Add a small note to the ID3 implementation saying that `cached_settings` is faster for that case than accessing config.setting - if it still is, I kind of lost track of the various pull requests around config value reading :-) But if it's faster for the ID3 implementation that really reads only one - and in a subset of cases, two - values from config.setting, wouldn't that generally be also true for any other implementation that needs to read at least one value?

phw: >     2\. s faster for that case than accessing config.setting - if it still is

It's not faster in that case. Both cases make two reads to config.setting. If several calls to this function are made a dict of settings certainly should be passed. That's also the only use case right now. For single call (should there be one) it is easier to just not pass settings object, in which case there should not really be a performance difference either way.

As this helper function is kind of pointless and maybe ads more confision now I removed it again in add39216bfa3ba58e80b2c9a21205d45e5281212 . @zas Fine with that?

picard/formats/id3.py
```diff
@@ -650,36 +650,30 @@ def _save_tags(self, tags, filename):
...
-    def new_metadata(self):
-        config = get_config()
-        if not config.setting["write_id3v23"]:
-            return self.metadata
-
-        copy = Metadata()
-        copy.copy(self.metadata)
+    def format_specific_metadata(self, metadata, tag, settings=None):
+        if not settings:
+            settings = get_config().setting
```
Sophist-UK, line 655: Is there any material difference between `get_config().setting.as_dict()` which returns a dict and `get_config().setting` which returns a ConfigSection object? Should we return a dict here too?

picard/file.py
zas: Yes, it isn't needed anymore.

picard/formats/id3.py
phw: The difference is in performance, `as_dict` is rather expensive. This is not good:

```python
settings = get_config().setting.as_dict()
if settings['foo']:
   do_something()
```

The config value `foo`  is accessed only once, `as_dict` has to read and copy all values from dict. It makes sense if you have something like this:

```python
settings = get_config().setting.as_dict()
for i in range(0, 10000):
    if settings['foo']:
        do_something()
```

Now the costs of accessing `setting[]` would be way greater than the costs of the dict conversion.

Of course the following would be even better:

```python
settings = get_config().setting
foo = settings['foo']
for i in range(0, 10000):
    if foo:
        do_something()
```

But in this case we don't do this, because we don't know in advance what settings are needed in the loop. It's a compromise, but as_dict is fast enough that it doesn't matter much in the case where it is called only once.

PR #1781, opened 2021-04-05
PICARD-2177: Add script function to replace entries in multi-value variables

atj:
* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

This is a simple patch to add a "$replacemulti" script function. It was originally implemented as a plugin for a user who was asking in IRC if it was possible to change/replace genres:

* [Community post](https://community.metabrainz.org/t/trying-to-standardize-genres-other-tags/527069/13)
* [Plugin gist](https://gist.github.com/atj/2333e8279bfa194b9ab624c9b637df65)

Happy to create a JIRA ticket if required.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -286,6 +286,30 @@ def func_replace(parser, text, old, new):
     return text.replace(old, new)
 
 
+@script_function(eval_args=False, documentation=N_(
+    """`$replacemulti(name,search,replace,separator="; ")`
+
+Replaces occurrences of `search` with `replace` in the multi-value variable `name`.
+
+Example:
+
+    $replacemulti(%genre%,Idm,IDM)
+"""
+))
+def func_replacemulti(parser, multi, search, replace, separator=MULTI_VALUED_JOINER):
+    if not multi or not search or replace is None or not separator:
+        return multi
+
+    search = search.eval(parser)
+    replace = replace.eval(parser)
+    multi_value = MultiValue(parser, multi, separator)
+    for n, value in enumerate(multi_value):
+        if value == search:
+            multi_value[n] = replace
+
+    return str(multi_value)
+
+
 @script_function(documentation=N_(
     """`$in(x,y)`
 
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -529,6 +529,20 @@ class ScriptParserTest(PicardTestCase):
     def test_cmd_replace(self):
         self.assertScriptResultEquals("$replace(abc ab abd a,ab,test)", "testc test testd a")
 
+    def test_cmd_replacemulti(self):
+        context = Metadata()
+        context["genre"] = ["Electronic", "Idm", "Techno"]
+        self.assertScriptResultEquals("$replacemulti(%genre%,Idm,IDM)", "Electronic; IDM; Techno", context)
+
+        context["genre"] = ["Electronic", "Jungle", "Bardcore"]
+        self.assertScriptResultEquals("$replacemulti(%genre%,Bardcore,Hardcore)", "Electronic; Jungle; Hardcore", context)
+
+        context["test"] = ["One", "Two", "Three"]
+        self.assertScriptResultEquals("$replacemulti(%test%,Four,Five)", "One; Two; Three", context)
+
+        context["test"] = ["Four", "Five", "Six"]
+        self.assertScriptResultEquals("$replacemulti(%test%,Five,)", "Four; ; Six", context)
+
     def test_cmd_strip(self):
         self.assertScriptResultEquals("$strip(  \t abc  de \n f  )", "abc de f")
 


picard/script/functions.py
```diff
@@ -286,6 +286,38 @@ def func_replace(parser, text, old, new):
...
+
+    multi = multi.split(separator)
+    changed = False
+    for n, value in enumerate(multi):
+        if value == search:
+            multi[n] = replace
+            changed = True
+
+    if changed:
+        parser.context[name] = separator.join(multi)
```
Sophist-UK, line 316: If I read this code correctly, this actually changes the variable provided.

Would it not be better to have a usage like `$setmulti(genre,$replacemulti(%genre%,Idm,IDM))`?

rdswift:
> Happy to create a JIRA ticket if required.

I suggest that you do just that.  That way it will be added to the list of changes that is generated for the blog posting.

Also, when this is merged, please either enter a PR against https://github.com/rdswift/picard-docs to add this to the documentation, or send me a reminder so that I can do it.  Thanks.

picard/script/functions.py
```diff
@@ -286,6 +286,38 @@ def func_replace(parser, text, old, new):
...
+
+Replaces occurrences of `search` with `replace` in the multi-value tag `name`.
+
+Example:
+
+    $replacemulti(genre,Idm,IDM)
+"""
+))
+def func_replacemulti(parser, name, search, replace, separator=MULTI_VALUED_JOINER):
+    if not name or not separator:
```
rdswift, line 300: Should this also check for a missing `search` parameter, or do you want to allow replacement of empty list elements?

Sophist-UK: I agree - `search` should be a required parameter. Indeed, I think that `replace` should be a required parameter, but it should be allowed to be empty.

i.e. `$replacemulti(%genre%)` and `$replacemulti(%genre%,search)` should both give an error, but `$replacemulti(%genre%,search,)` should not.

Sophist-UK (changes requested):

picard/script/functions.py
atj: Yes, referencing the variable by name was how I intended the function to be used, as that made the most sense to me. I did consider what you have suggested, but I didn't go with it because a) it's more complex/verbose to use and less readable (IMO) and b) I can't really think of a use case where someone would take the value of an existing multi-value variable, replace an element within it and then assign it to a different variable.

I'm not wedded to this approach so if there is a convention I'm missing then I am happy to change it.

picard/script/functions.py
atj: Thanks both, this was an oversight on my part.

atj: What would be expected behaviour for an empty replacement ie. `$replacemulti(%genre%,search,)`? Should the search element be removed or set to the empty string?

picard/script/functions.py
Sophist-UK: Yes - I think that there is a convention.

And although your approach is less verbose, it is a different approach to most functions which return the value. And it is only less verbose if you want to change the existing variable - it is more verbose and even less conventional if you want to assign the changed contents to a different variable:
```
$setmulti(new_genre,$replacemulti(%genre%,Idm,IDM))
vs.
$setmulti(new_genre,%genre%)
$replacemulti(new_genre,Idm,IDM)
```

picard/script/functions.py
rdswift: Good question.  I can see potential uses either way, but if I had to pick one I think I would go with removing them.  Perhaps there's a need for a new function `$cleanmulti()` to remove empty elements, because I'm not sure that there's currently an easy way to do that.  Then I would change my suggestion to leave them as empty strings.

atj: I've pushed some updates to addresses these issues. Currently the code sets an element to the empty string if replace is not specified, and I've added a test to confirm this behaviour.

picard/script/functions.py
atj: Thanks for the comments, I believe the implementation should now work as per your approach.

atj:
> > Happy to create a JIRA ticket if required.
> 
> I suggest that you do just that. That way it will be added to the list of changes that is generated for the blog posting.

Done - PICARD-2177

phw (approved):

picard/script/functions.py
phw: To me that makes sense

phw:
Thanks a lot, looks good to me. Let's get this merged.

atj:
Thanks for the constructive help and reviews @rdswift and @Sophist-UK

test/test_script.py
```diff
@@ -529,6 +529,20 @@ def test_cmd_pad(self):
     def test_cmd_replace(self):
         self.assertScriptResultEquals("$replace(abc ab abd a,ab,test)", "testc test testd a")
 
+    def test_cmd_replacemulti(self):
+        context = Metadata()
+        context["genre"] = ["Electronic", "Idm", "Techno"]
+        self.assertEqual("Electronic; IDM; Techno", self.parser.eval("$replacemulti(%genre%,Idm,IDM)", context))
```
zas, line 535:
```suggestion
        self.assertScriptResultEquals("$replacemulti(%genre%,Idm,IDM)", "Electronic; IDM; Techno", context)
```

test/test_script.py
```diff
@@ -529,6 +529,20 @@ def test_cmd_pad(self):
...
+
+        context["genre"] = ["Electronic", "Jungle", "Bardcore"]
+        self.assertEqual("Electronic; Jungle; Hardcore", self.parser.eval("$replacemulti(%genre%,Bardcore,Hardcore)", context))
```
zas, line 538:
```suggestion
        self.assertScriptResultEquals("$replacemulti(%genre%,Bardcore,Hardcore)", "Electronic; Jungle; Hardcore", context)
```

test/test_script.py
```diff
@@ -529,6 +529,20 @@ def test_cmd_pad(self):
...
+
+        context["test"] = ["One", "Two", "Three"]
+        self.assertEqual("One; Two; Three", self.parser.eval("$replacemulti(%test%,Four,Five)", context))
```
zas, line 541:
```suggestion
        self.assertScriptResultEquals("$replacemulti(%test%,Four,Five)", "One; Two; Three", context)
```

test/test_script.py
```diff
@@ -529,6 +529,20 @@ def test_cmd_pad(self):
...
+
+        context["test"] = ["One", "Two", "Three"]
+        self.assertEqual("One; ; Three", self.parser.eval("$replacemulti(%test%,Two,)", context))
```
zas, line 544:
```suggestion
        self.assertScriptResultEquals("$replacemulti(%test%,Two,)", "One; ; Three", context)
```

zas (changes requested):
Looks good.
In tests, you can use `assertScriptResultEquals()` to shorten lines a bit (see suggestions)

atj:
> Looks good.
> In tests, you can use `assertScriptResultEquals()` to shorten lines a bit (see suggestions)

Thanks, done. I changed the content of the last test case to clarify it's not related to the previous one.

zas (approved):

phw (dismissed):

picard/script/functions.py
```diff
@@ -286,6 +286,30 @@ def func_replace(parser, text, old, new):
     return text.replace(old, new)
 
 
+@script_function(eval_args=False, documentation=N_(
+    """`$replacemulti(value,search,replace,separator="; ")`
```
rdswift, line 290: I suggest changing this to:

```
    """`$replacemulti(name,search,replace,separator="; ")`
```

to make it consistent with the documentation on Line 292.

phw: I took the freedom to quickly fix this naming, will merge this now. Thanks a lot @atj

PR #1782, opened 2021-04-06
PICARD-2176:  Add column to see if cover art is present

petitminion:
modified:   picard/album.py
	modified:   picard/ui/itemviews.py
Ad colomn "Cover" to find album without cover more easily

# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
This add a new column to the mainwindows so you can order the album by coverart. This way its easy to find and change the song wich do not have covers.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -21,6 +21,7 @@
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2019 Joel Lintunen
 # Copyright (C) 2020 Gabriel Ferreira
+# Copyright (C) 2021 Petit Minion
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -666,22 +667,8 @@ class Album(DataObject, Item):
                 unsaved = self.get_num_unsaved_files()
                 if unsaved:
                     text += '; %d*' % (unsaved,)
-                # CoverArt.set_metadata uses the orig_metadata.images if metadata.images is empty
-                # in order to show existing cover art if there's no cover art for a release. So
-                # we do the same here in order to show the number of images consistently.
-                if self.metadata.images:
-                    metadata = self.metadata
-                else:
-                    metadata = self.orig_metadata
-
-                number_of_images = len(metadata.images)
-                if getattr(metadata, 'has_common_images', True):
-                    text += ngettext("; %i image", "; %i images",
-                                     number_of_images) % number_of_images
-                else:
-                    text += ngettext("; %i image not in all tracks", "; %i different images among tracks",
-                                     number_of_images) % number_of_images
-                return text + ')'
+                text += self._cover_art_description(self.metadata.images)
+                return text
             else:
                 return title
         elif column == '~length':
@@ -696,6 +683,8 @@ class Album(DataObject, Item):
             return self.metadata['~totalalbumtracks']
         elif column == 'discnumber':
             return self.metadata['totaldiscs']
+        elif column == 'covercount':
+            return self._cover_art_description(self.metadata.images, False)
         else:
             return self.metadata[column]
 
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -18,6 +18,7 @@
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2020 Ray Bouchard
 # Copyright (C) 2020 Gabriel Ferreira
+# Copyright (C) 2021 Petit Minion
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -240,6 +241,8 @@ class Cluster(FileList):
             return self.metadata['totaltracks']
         elif column == 'discnumber':
             return self.metadata['totaldiscs']
+        elif column == 'covercount':
+            return self._cover_art_description(self.metadata.images)
         return self.metadata[column]
 
     def _lookup_finished(self, document, http, error):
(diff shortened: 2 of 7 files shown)


zas:
@petitminion can you please clean up the commit message? Also there's a template for PR, you should use it.
Anyway, thanks for this contribution, we'll review it soon.

phw (changes requested):
Thanks a lot. That makes sense in general. A few things bother me, but not sure how to deal with all of them.

1. Code duplication: The same code is used twice. That could be easily solved by adding a helper method like this:

    ```python
    def _cover_art_description(self):
        # CoverArt.set_metadata uses the orig_metadata.images if metadata.images is empty
        # in order to show existing cover art if there's no cover art for a release. So
        # we do the same here in order to show the number of images consistently.
        if self.metadata.images:
            metadata = self.metadata
        else:
            metadata = self.orig_metadata

        number_of_images = len(metadata.images)
        text = ''
        if getattr(metadata, 'has_common_images', True):
            text = ngettext("%i image", "%i images",
                                number_of_images) % number_of_images
        else:
            text = ngettext("%i image not in all tracks", "%i different images among tracks",
                                number_of_images) % number_of_images
        return text
    ```

    Then each column showing this description can use helper function.

2. Duplication of information: The very same information is already present in the title column. I think this is not needed. But removing it from the title column completely would be annoying for users relying on this. Two possible ways to solve this:

    - In the title column we detect whether the covercount column is also present, and if yes, the cover art counts are not shown in the title. Checking the config could be somewhat expensive, though
    - Or we make the covercount column simpler and just show more compact information, e.g. really only the count as a numeric column instead of a descriptive string. Maybe this also makes it more useful for sorting.

3. Lack of information for tracks and files: I think if we add this column we should also show useful information for files, tracks and even clusters. This can be done as part of this PR, but I would also be fine with handling this separately.


Also, but this is unrelated to this PR, we should think about refactoring the column handling a bit. The column configuration is not very flexible, as it all depends on the position of the columns in the `MainPanel.columns`  array. That makes it difficult to add, remove or reorder the columns without breaking user configuration. Ideally we should probably store the activated columns as an array of column names.

petitminion:
Hello, 

1. Yep I though about that but wasn't sure 
2. Very new to python and dev so 2.1 seem too complicated for me, 2.2 looks like i can do 
3. Will look if I understand how to do this 

will try to change the commit message too

petitminion:
I have 1 and 2 good to go but for 3 I have a problem : the function i created need to be accessible to the 4 modules (track album cluster and files). Mayby I should move it to some place like picard.util ? but I'm not sure this will be more clean or feasible because we have to pass the metadata attribute of the Track instance...
I can also push ignoring 3

picard/album.py
```diff
@@ -696,6 +696,21 @@ def column(self, column):
...
+        elif column == 'covercount':
+            text = ""
+            if self.metadata.images:
+                metadata = self.metadata
+            else:
+                metadata = self.orig_metadata
+
+            number_of_images = len(metadata.images)
+            if getattr(metadata, 'has_common_images', True):
+                text += ngettext("%i image", "%i images",
```
zas, line 708: Why do you concatenate to empty `text` here? You may just `return ngettext(...)`

zas (changes requested):

petitminion:
classe album, track, file  -> filelistitem
cluster -> filelist -> filelistitem 
--> I should put the function def _cover_art_description(self): in fililisitem to make it work for all 4 class ? gonna try

petitminion:
ok its done but i'm fiting with git, thought i just had to make a new commit and the push, but it says : 
```
! [rejected]          cover_column -> cover_column (non-fast-forward)
error: failed to push some refs to 'https://github.com/petitminion/picard'
hint: Updates were rejected because the tip of your current branch is behind
hint: its remote counterpart. Integrate the remote changes (e.g.
hint: 'git pull ...') before pushing again.
hint: See the 'Note about fast-forwards' in 'git push --help' for details.

```

phw:
Thanks a lot for working on this.

Regarding the git issue: You likely changed the history on your local branch, probably due to rebasing. Your branch on Github contains currently a single commit https://github.com/metabrainz/picard/pull/1782/commits/411673caab3d4eccae13630aeaeab17e7e43e9a4 . Your local branch differs.

If you are sure your local branch contains everything and you don't loose changes you can force push your local branch with ` git push --force`. This will overwrite the branch on Github with your local changes.

In this case it is likely fine. But always double check before doing this, easy to loos work with that. The other reason why this fails can be that additional commits where added to the remote branch. That's especially the case if multiple people are working on that branch. In that case you'd need to run `git pull` first to integrate the remote changes to your local branch.

petitminion:
thx a lot I didn't knew I could force o/ i get this :s 
```
remote: fatal error in commit_refs
To https://github.com/petitminion/picard
 ! [remote rejected]   cover_column -> cover_column (failure)
error: failed to push some refs to 'https://github.com/petitminion/picard'

```

petitminion:
ok it worked but changes don't appears here

petitminion:
This is making my brain blowing, can i open a new merge request ?

phw:
> This is making my brain blowing, can i open a new merge request ?

Yes, sure. That is indeed odd, I see that there are different commits on your Github fork. Usually this pull request should get updated, not sure what happened here. Feel free to submit a new one.

I'm looking over the code already. I'll have a few minor comments, but overall this looks pretty good, thanks a lot.

petitminion:
thx ! done in #1788

zas:
Why are there 2 PRs? Which one is the correct one?

phw:
There was trouble with Github not updating this PR after pushing changes, so @petitminion opened a separate PR. Now this PR also seems to have catched up. No idea really what happened. Let's close this one and use the other

PR #1786, opened 2021-04-09
PICARD-2180: Add new $cleanmulti() function

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Adds a new `$cleanmulti()` function.

# Problem

Through various plugins and functions, it is possible to create empty elements in multi-value variables.  These could cause display problems, and possibly tagging issues, and there is no current way to remove these empty elements.

* JIRA ticket (_optional_): PICARD-2180

# Solution

Create a new `$cleanmulti()` scripting function to remove all empty elements from a multi-value variable.

# Action

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1353,7 +1353,9 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
 @script_function(eval_args=False, documentation=N_(
     """`$cleanmulti(name,separator="; ")`
 
-Removes empty elements from the multi-value variable `name`.
+Removes empty elements from the multi-value variable `name`. A literal value
+representing a multi-value can be substituted for `name`, using the separator
+(or "; " if not passed) to coerce the value into a proper multi-valued tag.
 
 Example:
 
@@ -1365,5 +1367,4 @@ def func_cleanmulti(parser, multi, separator=MULTI_VALUED_JOINER):
         return ''
 
     multi_value = MultiValue(parser, multi, separator)
-    temp = [x for x in multi_value if x]
-    return multi_value.separator.join(temp)
+    return multi_value.separator.join([x for x in multi_value if x])
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1348,3 +1348,22 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
     if not case_sensitive:
         multi_value._multi = list({v.lower(): v for v in multi_value}.values())
     return multi_value.separator.join(sorted(set(multi_value)))
+
+
+@script_function(eval_args=False, documentation=N_(
+    """`$cleanmulti(name,separator="; ")`
+
+Removes empty elements from the multi-value variable `name`.
+
+Example:
+
+    $cleanmulti(%foo%)
+"""
+))
+def func_cleanmulti(parser, multi, separator=MULTI_VALUED_JOINER):
+    if not multi or not separator:
+        return ''
+
+    multi_value = MultiValue(parser, multi, separator)
+    temp = [x for x in multi_value if x]
+    return multi_value.separator.join(temp)
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1538,13 +1538,16 @@ class ScriptParserTest(PicardTestCase):
         context["foo"] = ['a', 'b', 'c', '', 'd']
         context["bar"] = ['a', 'b', 'c', None, 'd']
         context["baz"] = ['a', 'b', 'c', 'd']
-        context['empty'] = []
+        context["empty"] = []
+        context["string"] = "a; b; c; ; d"
         result = "a; b; c; d"
         # Tests with context
         self.assertScriptResultEquals("$cleanmulti(%foo%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%bar%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%baz%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%empty%)", "", context)
+        # Tests with separator override
+        self.assertScriptResultEquals("$cleanmulti(%string%,:)", "a; b; c; ; d", context)
         # Tests with missing inputs
         self.assertScriptResultEquals("$cleanmulti(,)", "", context)
         self.assertScriptResultEquals("$cleanmulti(,; )", "", context)
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1546,8 +1546,12 @@ class ScriptParserTest(PicardTestCase):
         self.assertScriptResultEquals("$cleanmulti(%bar%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%baz%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%empty%)", "", context)
+        self.assertScriptResultEquals("$cleanmulti(%string%)", "a; b; c; ; d", context)
+        # Tests with literal value
+        self.assertScriptResultEquals("$cleanmulti(a; b; c; ; d)", "a; b; c; d", context)
         # Tests with separator override
-        self.assertScriptResultEquals("$cleanmulti(%string%,:)", "a; b; c; ; d", context)
+        self.assertScriptResultEquals("$cleanmulti(a; b; c; ; d,; )", "a; b; c; d", context)
+        self.assertScriptResultEquals("$cleanmulti(a; b; c; ; d,:)", "a; b; c; ; d", context)
         # Tests with missing inputs
         self.assertScriptResultEquals("$cleanmulti(,)", "", context)
         self.assertScriptResultEquals("$cleanmulti(,; )", "", context)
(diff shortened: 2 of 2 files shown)


mineo:
Is there ever any good reason to keep empty values during calls to `setmulti` and the Metadata class's APIs? I can't imagine these values are useful. Maybe we can automatically clean them in the write codepaths automagically.

Sophist-UK:
I haven't got a clue what might be wanted during the processing of a script.

But I can't see an empty value having any meaning in an written tag - and indeed in some formats empty values may actually not be allowed.

rdswift:
I can't think of any reason, but who knows how people process their information.  :wink:  For example, the recent `$replacemulti()` function can return empty elements if the replacement is empty.

I do like your suggestion though.  It would likely be cleaner in the long run.

Sophist-UK:
What should happen if I explicitly do `$setmulti(m,a; ; b)`?

Should the explicitly defined empty middle item be removed? That seems wrong.

But if I can set it manually, then I think you have to allow e.g. $replacemulti to create empty items. in which case we had better have a $cleanmulti function to delete them.

Or does  `$delmulti(%m%,)` do exactly this?

rdswift:
> Or does `$delmulti(%m%,)` do exactly this?

I don't believe that there is a `$delmulti()` function currently within Picard.

picard/script/functions.py
```diff
@@ -1348,3 +1348,22 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
...
+    $cleanmulti(%foo%)
+"""
+))
+def func_cleanmulti(parser, multi, separator=MULTI_VALUED_JOINER):
+    if not multi or not separator:
+        return ''
+
+    multi_value = MultiValue(parser, multi, separator)
+    temp = [x for x in multi_value if x]
+    return multi_value.separator.join(temp)
```
zas, line 1369: I think that's simple and short enough so you can get rid of `temp` and just write:

```python
return multi_value.separator.join([x for x in multi_value if x])
```

picard/script/functions.py
```diff
@@ -1348,3 +1348,22 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
     if not case_sensitive:
         multi_value._multi = list({v.lower(): v for v in multi_value}.values())
     return multi_value.separator.join(sorted(set(multi_value)))
+
+
+@script_function(eval_args=False, documentation=N_(
+    """`$cleanmulti(name,separator="; ")`
+
+Removes empty elements from the multi-value variable `name`.
```
zas, line 1356: It also works with a literal value, so this description isn't totally exact.

picard/script/functions.py
```diff
@@ -1348,3 +1348,22 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
...
+
+Example:
+
+    $cleanmulti(%foo%)
+"""
+))
+def func_cleanmulti(parser, multi, separator=MULTI_VALUED_JOINER):
```
zas, line 1363: Not sure about the function name, but I didn't find better.

picard/script/functions.py
rdswift: Changed as suggested.  Thanks.

picard/script/functions.py
rdswift: Description updated to include a note about using a literal as an input.

picard/script/functions.py
rdswift: I agree that there's probably a better name, but I can't think of one right now.  I'll happily change it if someone comes up with a better suggestion.  The only other one I'm thinking right now is `$removeemptymulti()` but I'm not sure that's any better.

rdswift: Maybe `$delemptymulti()`?

mineo:
> What should happen if I explicitly do $setmulti(m,a; ; b)?

I tried to see what happens in this case with the following script:

```
$setmulti(foo,a; ; b; ; ; )
$join(%foo%,/)
$lenmulti(%foo%)
$setmulti(bar,$map(%foo%,A))
--
$join(%bar%,/)
$lenmulti(%bar%)
--
$getmulti(%foo%,0)
$getmulti(%foo%,1)
```

The result was `a/b2--A/A2--AA`, so either I'm completely misunderstanding something, or the existing scripting functions already drop empty values. The "view script variables" plugin also shows only two values in `foo` and bar.

Curious as to how it's actually possible to add empty values, I tried to modify the new testcases like so:

```diff
diff --git a/test/test_script.py b/test/test_script.py
index 547eeb42..e21e7a3b 100644
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1542,6 +1542,7 @@ class ScriptParserTest(PicardTestCase):
         context["string"] = "a; b; c; ; d"
         result = "a; b; c; d"
         # Tests with context
+        self.assertScriptResultEquals("$lenmulti(%foo%)", "5", context)
         self.assertScriptResultEquals("$cleanmulti(%foo%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%bar%)", result, context)
         self.assertScriptResultEquals("$cleanmulti(%baz%)", result, context)

```

(and `bar`). It turns out, that fails because the length of `foo` is not - as the tests suggests - 5 (a, b, c, d + the empty element), but 4, making the `cleanmulti` calls on `foo` and `bar` effectively no-ops. So the question I have is: how do I get a multi-value with empty values in picard?

mineo:
Ah, it's https://github.com/metabrainz/picard/blob/7faa2037090dbdc000a4d2f25780399c9be77792/picard/metadata.py#L434 that already discards empty values.

Changing the implementation of cleanmulti to

```
diff --git a/picard/script/functions.py b/picard/script/functions.py
index 6e744c94..19c5c094 100644
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1367,4 +1367,4 @@ def func_cleanmulti(parser, multi, separator=MULTI_VALUED_JOINER):
         return ''
 
     multi_value = MultiValue(parser, multi, separator)
-    return multi_value.separator.join([x for x in multi_value if x])
+    return multi_value.separator.join([x for x in multi_value])
```

causes the tests with literal values to fail, but the ones with a value read from a variable still pass.

rdswift:
> So the question I have is: how do I get a multi-value with empty values in picard?

Try the following test case:

```
        # Temporary tests for mineo
        test_in = '$setmulti(foo,a; b; c; d)$map(%foo%,$if($eq(%_loop_count%,2),,%_loop_value%)) / %foo%'
        test_out = 'a; ; c; d / a; ; c; d'
        self.assertScriptResultEquals(test_in, test_out, context)
```

I believe that it's also possible using the new `$replacemulti()` scripting command.

mineo:
> I believe that it's also possible using the new $replacemulti() scripting command.

This looks like a bug in the implementation of replacemulti. It wants to handle a missing/empty replacement character (https://github.com/metabrainz/picard/blob/7faa2037090dbdc000a4d2f25780399c9be77792/picard/script/functions.py#L300, but uses `replace is None` where it should use `not replace`. In fact, the no-op codepath of replacemulti doesn't work at all (https://tickets.metabrainz.org/browse/PICARD-2182).

The map function is interesting. It can in fact be used to generate multi-value with empty values, but only if you never use `$setmulti` to save that value again. To be honest, I find that very confusing for users like me that use the file naming options page to debug a script, they'll see one value in the examples and a different value is written to the tags.

It's a tricky situation that really requires understanding of Picard's internals: glancing at the rest of the multi-value functions, I think all behave differently when a multi-value is passed as a literal value (empty values are kept/empty values can be created), compared to when one is passed from a `setmulti`'d variable (because setmulti discards empty values). We should stick to either behavior, but not both. I vote for always discarding empty values :-)

picard/script/functions.py
zas: Not convinced by `$delemptymulti()` either. @phw @Sophist-UK @mineo any idea?

Sophist-UK:
> Ah, it's
> 
> https://github.com/metabrainz/picard/blob/7faa2037090dbdc000a4d2f25780399c9be77792/picard/metadata.py#L434
> 
> that already discards empty values.

If `metadata` disallows empty values already, then IMO for consistency:

A. `$replacemulti` should delete empty values; and
B. `$cleanmulti` should not be implemented.

The alternative would be to start to allow empty values by removing this restriction from `metadata`, but as I have previously noted this makes no real sense as a tag value, might be explicitly disallowed by some formats and could break players which are not expecting it - and more importantly could then create a backwards compatibility issue that breaks certain scripts.

zas:
> If `metadata` disallows empty values already, then IMO for consistency:
> 
> A. `$replacemulti` should delete empty values; and
> B. `$cleanmulti` should not be implemented.

I tend to agree, if we can avoid to have empty values, we don't need a function to remove them.

rdswift:
> I tend to agree, if we can avoid to have empty values, we don't need a function to remove them.

In that case, I'll cancel this PR.  Should I instead update `$map()` and `$replacemulti()` to automatically remove empty elements?

Sophist-UK:
> In that case, I'll cancel this PR. Should I instead update `$map()` and `$replacemulti()` to automatically remove empty elements?

IMO (for what this is worth) yes we should do that.

rdswift:
PR #1790 opened to address the `$map()` function.

PR #1787, opened 2021-04-10
PICARD-2181: Allow Markdown for plugin descriptions

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2181
* https://github.com/metabrainz/picard-website/pull/229

Many plugins try to come with a more comprehensive description, including some markup. So far those plugins contained HTML tags. Allowing markdown will allow easier and better readable (in source) formatting.

# Solution

Use the already used python-markdown module to render the plugin descriptions. As this module is considered optional provide a fallback that only ensures newlines are preserved.

Rendering with Markdown:

grafik

Without Markdown, using fallback:

grafik

As HTML is allowed in markdown existing descriptions continue to render properly. Likewise pure markdown still remains readable without conversion to HTML.

# Action

For the listing of downloadable plugins even older Picard versions would render the description fine as the description is served by the website after conversion to HTML. We could change that, but I would keep it that way for compatibility for a while.

We should not immediately start converting plugin descriptions to Markdown, though. Older Picard versions would not render the description of installed plugins properly. Instead we should use Markdown only for plugins compatible with Picard API version 2.7 or later. This change is here to start transitioning.

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -83,6 +83,7 @@ api_versions = [
     "2.4",
     "2.5",
     "2.6",
+    "2.7",
 ]
 
 api_versions_tuple = [Version.from_string(v) for v in api_versions]
--- a/picard/plugin.py
+++ b/picard/plugin.py
@@ -45,6 +45,15 @@ from picard.version import (
 )
 
 
+try:
+    from markdown import markdown
+except ImportError:
+    def markdown(text):
+        # Simple fallback, just make sure line breaks are applied
+        if not text:
+            return ''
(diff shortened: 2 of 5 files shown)


Sophist-UK:
@phw - Great idea!!

zas (approved):

phw:
> @phw - Great idea!!

Thanks. The patches for Picard itself and the website have been lying around here for over a year. I wanted to submit this for the "next" version, and then never did :)

PR #1788, opened 2021-04-10
PICARD-2176: Add column to see if cover art is present 

petitminion:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
This add  a cover columns in album, tracks, files and cluster to be able to order by presence or abscence of cover art.

# Problem

* JIRA ticket (_optional_): PICARD-2176

# Solution

added a method in picard.metadata wich is use by the 4 module to retrive cover art information

# Action

Because the ouptu of this method have to be a number, or a string (when it used for the title colum of the album module), I create a argument "forcount". But not shure this is the best way to go. 

this is my first piece of programming in python, hope it not to dirty o/

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -533,7 +533,6 @@ class MultiMetadataProxy:
     metadata to use file specific metadata, without making it actually part
     of the track.
     """
-
     WRITE_METHODS = [
         'add_unique',
         'add',
--- a/picard/track.py
+++ b/picard/track.py
@@ -248,6 +248,8 @@ class Track(DataObject, FileListItem):
         if column == 'title':
             prefix = "%s-" % m['discnumber'] if m['discnumber'] and m['totaldiscs'] != "1" else ""
             return "%s%s  %s" % (prefix, m['tracknumber'].zfill(2), m['title'])
+        if column == 'covercount':
+            return self._cover_art_description(self.metadata.images, False)
         elif column in m:
             return m[column]
         elif self.num_linked_files == 1:
--- a/picard/ui/item.py
+++ b/picard/ui/item.py
@@ -8,6 +8,7 @@
 # Copyright (C) 2012 Chad Wilson
 # Copyright (C) 2013 Laurent Monin
 # Copyright (C) 2014 Sophist-UK
+# Copyright (C) 2021 Petit Minion
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -98,6 +99,28 @@ class Item(object):
 
     def clear_errors(self):
         self._errors = []
+
+    def _cover_art_description(self, images, detailed=True):
+        """Return the number of cover art images in `images` for display in the UI
+
+        Args:
+            images: An instance of picard.util.imagelist.ImageList
+            detailed: If True, return a detailed text about the images and whether they are the same across all tracks.
+                    If False only return the number as a string.
+
+        Returns:
+            A string explaining the cover art image count.
+        """
+        number_of_images = len(images)
+        if detailed:
+            if getattr(self, 'has_common_images', True):
+                return ngettext("; %i image)", "; %i images)",
+                                number_of_images) % number_of_images
+            else:
+                return ngettext("; %i image not in all tracks)", "; %i different images among tracks)",
+                                number_of_images) % number_of_images
+        else:
+            return str(number_of_images)
 
 
 class FileListItem(Item):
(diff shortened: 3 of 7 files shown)


picard/metadata.py
```diff
@@ -520,6 +520,22 @@ def __repr__(self):
     def __str__(self):
         return ("store: %r\ndeleted: %r\nimages: %r\nlength: %r" % (self._store, self.deleted_tags, [str(img) for img in self.images], self.length))
 
+    def _cover_art_description(self, forcount="no"):
```
phw, line 523: I think we should not place the helper function here. It is very UI related, and metadata itself is supposed to be primarily a data holding object.

Not 100% sure where to put this, but my suggestion would be to add it to the `Item` class (see https://github.com/metabrainz/picard/blob/master/picard/ui/item.py#L30), from which all those classes are derived. I would then change the function to accept an `images` parameter, so it can be called in the column methods as :

```
self.cover_art_description(self.metadata.images)
```

Another thing is that instead of `forcount` accepting a string it would be better as a boolean. I'd actually turn it around and call it `detailed`, so your function declaration looks like this:

```
def _cover_art_description(self, images, detailed=True):
```

You can then call it as:

```
self.cover_art_description(self.metadata.images, detailed=True)
```

or

```
self.cover_art_description(self.metadata.images, detailed=False)
```

which reads nicer.

phw (changes requested):
I'm fine with how this works. Some minor comments about the `_cover_art_description` helper method. Maybe @zas has some opinion where to best place it, but I gave my suggestion.

One thing: The commit messages should in general be more descriptive. These changes could be put into a single commit message, maybe something like:

> PICARD-2176: Add column for cover art count

It would be great if you could squash your commits into a single one with such a message. A kind of nice explanation how to do this is at https://www.git-tower.com/learn/git/faq/git-squash/

But if that causes you problems don't worry. I can also take care of that when merging your changes :)

Also please add yourself to the copyright sections at the top of the files you change.

petitminion:
oups forgot to squash, will try we i get back home

picard/ui/item.py
```diff
@@ -99,6 +100,19 @@ def error_append(self, msg):
     def clear_errors(self):
         self._errors = []
 
+    def _cover_art_description(self, images, detailed=True):
+        # Return the number of cover image found in metadata.images. Used I the cover column of the UI.
```
zas, line 104: To describe the method, use Python docstring (see https://www.python.org/dev/peps/pep-0257/) -> replace `# comment` with `""" comment """` (it can be multi-line).

What does `Used I the cover column of the UI.` mean? (tu peux expliquer en français)

zas (changes requested):

picard/ui/item.py
phw: Maybe something like this for a full function description:

```python
"""Return the number of cover art images in `images` for display in the UI

Args:
    images: An instance of picard.util.imagelist.ImageList
    detailed: If True, return a detailed text about the images and whether they are the same across all tracks.
              If False only return the number as a string.

Returns:
    A string explaining the cover art image count.
"""
```

Apart from this description this PR looks good to me. Good work!

petitminion:
sorry for all the errors and thx for your reviews !

phw (approved):

phw:
> sorry for all the errors and thx for your reviews !

You're welcome, glad to help. Thanks for the improvement and for getting through this with us :)

zas (changes requested):

picard/track.py
```diff
@@ -248,6 +248,8 @@ def column(self, column):
         if column == 'title':
             prefix = "%s-" % m['discnumber'] if m['discnumber'] and m['totaldiscs'] != "1" else ""
             return "%s%s  %s" % (prefix, m['tracknumber'].zfill(2), m['title'])
+        if column == 'covercount':
+            return self.metadata._cover_art_description("yes")
```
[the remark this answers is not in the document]
phw: Missed that. Thanks @petitminion for fixing

picard/ui/item.py
```diff
@@ -99,6 +100,28 @@ def error_append(self, msg):
     def clear_errors(self):
         self._errors = []
 
+    def _cover_art_description(self, images, detailed=True):
```
zas, line 103: It would be better to define 2 functions, instead of passing boolean argument `detailed`, which is `True` only once in current code. Common code is rather small in both cases, and it would be clearer.
Also since those methods are used outside their module, I would get rid of `_` prefix:

```python
def cover_art_description(self, images):
    return str(len(images))

def cover_art_description_detailed(self, images):
    number_of_images = len(images)
    if getattr(self, 'has_common_images', True):
       ...
```

zas (changes requested):

zas (approved):

PR #1790, opened 2021-04-11
PICARD-2185: Remove empty elements from $map() output

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

The `$map()` function does not remove empty elements.  This is inconsistent with other multi-value variable processing.

* JIRA ticket (_optional_): PICARD-2185

# Solution

Automatically remove empty elements from `$map()` output.

# Action

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -290,6 +290,7 @@ def func_replace(parser, text, old, new):
     """`$replacemulti(name,search,replace,separator="; ")`
 
 Replaces occurrences of `search` with `replace` in the multi-value variable `name`.
+Empty elements are automatically removed.
 
 Example:
 
@@ -1214,7 +1215,7 @@ def func_map(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
         multi_value[loop_count - 1] = str(loop_code.eval(parser))
     func_unset(parser, '_loop_count')
     func_unset(parser, '_loop_value')
-    return multi_value.separator.join([x for x in multi_value if x])
+    return str(multi_value)
 
 
 @script_function(eval_args=False, documentation=N_(
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1196,6 +1196,8 @@ Iterates over each element found in the multi-value tag `name` and updates the
     `_loop_value` and the count is stored in the tag `_loop_count`. This allows
     the element or count value to be accessed within the `code` script.
 
+Empty elements are automatically removed.
+
 Example:
 
     $map(First:A; Second:B,$upper(%_loop_count%=%_loop_value%))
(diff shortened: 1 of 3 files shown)


test/test_script.py
```diff
@@ -1204,19 +1204,23 @@ def test_cmd_while(self):
...
         # Tests with context
+        context["foo"] = "First:A; Second:B; Third:C"
         self.assertScriptResultEquals("$map(%foo%,$upper(%_loop_count%=%_loop_value%))", foo_output, context)
+        context["bar"] = ["First:A", "Second:B", "Third:C"]
         self.assertScriptResultEquals("$map(%bar%,$upper(%_loop_count%=%_loop_value%))", loop_output, context)
         # Tests with static inputs
         self.assertScriptResultEquals("$map(First:A; Second:B; Third:C,$upper(%_loop_count%=%_loop_value%))", loop_output, context)
+        # Tests for removing empty elements
+        context["baz"] = ["First:A", "Second:B", "", "Third:C"]
+        self.assertScriptResultEquals("$map(%baz%,$upper(%_loop_count%=%_loop_value%))", loop_output, context)
```
mineo, line 1219: This assertion passes, but it does not test what - I believe - the reader thinks it does: the problem is that after `context["baz"] = ....` the empty element is already removed (`$lenmulti(%baz%)` is 3 instead of 4), so whether the map function discards empty elements or not is irrelevant to the outcome of the assertion. The assertion still passes if the remove-empty-elements code in `func_map` is removed.

rdswift: I think I see what you're saying.  Does this latest commit address your concern?

picard/script/functions.py
```diff
@@ -1208,10 +1210,11 @@ def func_map(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
     for loop_count, value in enumerate(multi_value, 1):
         func_set(parser, '_loop_count', str(loop_count))
         func_set(parser, '_loop_value', str(value))
+        # Make changes in-place
         multi_value[loop_count - 1] = str(loop_code.eval(parser))
     func_unset(parser, '_loop_count')
     func_unset(parser, '_loop_value')
-    return str(multi_value)
+    return multi_value.separator.join([x for x in multi_value if x])
```
zas, line 1217: Instead, we could modify `MultiValue.__str__()` to skip empty values (if those aren't expected anyway). It would reduce code redundancy.

```python
    def __str__(self):
        return self.separator.join([x for x in self if x])
```

https://github.com/metabrainz/picard/blob/ba862a14f8ec2df8f363e399f4c0bd64bdcf0bb5/picard/script/parser.py#L400


Not sure if there's any drawbacks. Thoughts?

rdswift: That actually makes a lot of sense to me, assuming that you always want to strip out empty elements (which it sounds like we do).  If we decide to go that way, should I make the change in this PR or create a separate one?

zas: I would do it in this PR, but let's wait feedback from @phw about this

phw: I agree that this is a good idea. I can't see any drawbacks currently

mineo (approved):

zas (approved):

phw (approved):

PR #1791, opened 2021-04-11
PICARD-2184: Allow completely silent Windows installer

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2184

Running the Windows installer with the /S flag will now not show any dialog and perform all default actions, including uninstalling the previous version.

# Solution

Detect whether the installer runs in silent mode. Set defaults for MessageBox and for uninstall of previous version also run the uninstall in silent mode.

See https://nsis.sourceforge.io/Docs/Chapter4.html#silent for technical background.

# Action

### DIFF
--- a/installer/picard-setup.nsi.in
+++ b/installer/picard-setup.nsi.in
@@ -207,7 +207,7 @@ SectionEnd
 Function un.onInit
   FindWindow $0 "${WNDCLASS}" "${WNDTITLE}"
   StrCmp $0 0 continueInstall
-    MessageBox MB_ICONSTOP|MB_OK "$(MsgApplicationRunning)"
+    MessageBox MB_ICONSTOP|MB_OK "$(MsgApplicationRunning)" /SD IDOK
     Abort
   continueInstall:
   !insertmacro INSTALLOPTIONS_EXTRACT "removeSettings.ini"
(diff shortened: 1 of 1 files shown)


zas (commented):
I can't test.

phw:
@rdswift Could you test the installer from https://github.com/metabrainz/picard/actions/runs/738952012 if that works a) when run normally with UI and b) if run with the /S flag?

In case of the /S flag it should not show any UI but perform the uninstall of the previous version automatically.

phw:
Did another fix: The installer did not wait until uninstaller had finished before proceeding. This was even an issue in the GUI case. The user could choose to ignore the uninstaller window  and go though the install process anyway, because the installer opened in the background.

In the silent case this actually seemed to work fine, but I guess in the worst case both would run in parallel and conflict.

Please use the builds from https://github.com/metabrainz/picard/actions/runs/741086756 for testing.

NSIS scripting syntax is awful, I tried my best to keep it somewhat readable.

rdswift:
Tested on Windows 7 system with the following results.

### Test 1:

- Initial installed version: 2.6.0b3
- Install command: "picard-setup-2.7.0.dev1.exe"
- UI Display: Usual installation UI including uninstall information.
- Final installed version: 2.7.0.dev1


### Test 2:

- Initial installed version: 2.7.0.dev1
- Install command: "picard-setup-2.7.0.dev1.exe"
- UI Display: Usual installation UI including uninstall information.
- Final installed version: 2.7.0.dev1


### Test 3:

- Initial installed version: 2.7.0.dev1
- Install command: "picard-setup-2.7.0.dev1.exe /S"
- UI Display: None.
- Final installed version: 2.7.0.dev1


### Test 4:

- Initial installed version: 2.6.0b3
- Install command: "picard-setup-2.7.0.dev1.exe /S"
- UI Display: None.
- Final installed version: 2.7.0.dev1

rdswift:
In the case of Test 2 above, since both the installer and uninstaller are the new 2.7.0.dev1 version would it be possible to call the uninstaller with the `/S` switch so that the install UI is still displayed but the uninstall UI is suppressed?  I think that would be the ideal situation.

phw:
Thanks for testing

> In the case of Test 2 above, since both the installer and uninstaller are the new 2.7.0.dev1 version would it be possible to call the uninstaller with the `/S` switch so that the install UI is still displayed but the uninstall UI is suppressed? I think that would be the ideal situation.

I thought about having the uninstaller run silently in case of GUI installs. But that would mean the user just sees nothing while the uninstaller is running, and it might fail in some unexpected way without the user seeing any dialog. I'd prefer the current way were the user actively goes through the uninstall process.

I also don't see really why the process should be different depending on version.

rdswift:
Okay, I misunderstood the use case.  After re-reading it, it all now makes more sense.  And I agree that there could be a problem if there were errors encountered during a silent uninstall.  Forget I mentioned it.  :grinning:

PR #1792, opened 2021-04-13
PICARD-2179: Add separate window for naming script editing.

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

With the addition of the example files that can be selected by the user when editing the fine naming script, the available space for the script editing field is extremely limited.

# Solution

Add a new sub-window with only the editor and examples shown, and allow resizing the window as well as a splitter between the editing field and the examples to allow further (relative) expansion of the editing portion.

image

image

# Action

Do we want to make additional changes, such as:
- Disable script editing from the main screen.
- Use only the default test cases on the main screen to free up more space for displaying the script.
- Refactor to reduce the duplication of code between the two class modules (e.g.: default test data, example file selection code, example display generation code, etc.)
- Others?


phw:
This goes into the correct direction, but I'm unsure about the details.

Maybe we should indeed just show the naming examples and a button to edit the script, then have the details dialog as you have it.

In later updates we could extend the main file naming screen even more. One idea would be that we define a couple of file naming presets, that we offer as a dropdown, for easy configuration. The script editor button then would allow you to customize the underlying scripts.

Sophist-UK:
I like the idea in principle providing that it is applied to all script editing fields - and it would also provide the basis for more functionality (like displaying context help automatically for functions and variables). But if we keep script editing available on the options page itself, I would also support splitting the File Naming options page into two - one just for scripting, the other for all other options - and this could be a simpler quick fix. To give my personal view on the questions asked:

- Disable script editing from the main screen - **Maybe** - to avoid the complexity of providing the same (but perhaps a subset) of functionality in the options page compared to the popup..
- Use only the default test cases on the main screen to free up more space for displaying the script - **No** - always show examples on the File Naming and Cover Art options pages so you can see what current settings do**
- Refactor to reduce the duplication of code between the two class modules (e.g.: default test data, example file selection code, example display generation code, etc.) - **Yes** - or move all editing code only to the pop-up. If we end up with the pop-up having more functionality than the options pages, then also having editing on the options page may get messy.

Additionally:
- Make the script editing box available for Tagging and Cover Art naming scripts
- With more screen space in the pop-up box, we can add more functionality to help scripters - like auto-display of context sensitive help for functions and variables.

Pros and cons of ***only*** having editing in the popup:
- **-** Usability for simple scripts or iterative editing of scripts - additional actions to perform a simple edit
- **+** Same UI and common code for all script editing
- **+** Screen space and base for improved editing functionality

zas:
I like the idea of having a dedicated dialog for this core feature of Picard.
- dialog shouldn't depend on Options dialog at all
- user should be able to modify selected files while the dialog is shown (in order to update examples)
- user should be able to minimize it while interacting with main window
- user should be able to resize each part (and possibly select different layout)
- file naming script editor shouldn't be duplicated, but we could have one access from Options and one directly from main menu (with keyboard shortcut/accelerator)

rdswift:
Some significant changes with the latest commit:
- script editor available from main screen, options screen, and Ctl-Shift-E shortcut
- script editor dialog does not depend on options dialog
- script editor dialog can be minimized to allow changing option settings, selected files, etc.
- script editor allows resizing of split between editing field and examples, and the split between the two example columns
- script editor examples show error message if file naming and file moving are both disabled (but still allows script editing)
- script editor refreshes examples when receiving keyboard focus to immediately accommodate changes to file selection, option settings, etc.

If this is moving in the right direction, some things remaining to be done include:
- disable naming script editor on options page and make the script display read-only
- refactor to move examples to constants to avoid duplication
- refactor example code to avoid duplication

One thing that bugs me but I haven't figured out how to resolve it yet is that opening the script editor from the options window does not allow you to enter the editor because the options window is modal and the editor isn't.

This also currently only allows editing of the file naming script.

zas:
Good work :)

Few thoughts:

- we display 10 items as examples, but the current limit (and the fact it's random) isn't really obvious, perhaps a tooltip on Renew examples would help, dunno, but that's something we need to improve
- when selecting one entry in Before, I'd expect the corresponding entry in After to be hightlighted
- Dialog should be titled "File naming script editor" (and corresponding menu item too)
- is there a way to format the script so it isn't a loooong line?
- buttons should perhaps be just under the script edit widget
- when changing selection one has to press Renew examples to update examples, perhaps we could just handle a selection change signals and update on the fly when the window is displayed
- when activating the script editor via button in Options the behavior is very weird (here, the dialog starts minimized, and I can move it to front)

That's just thoughts after a quick test, I'll review this more in depth soon.

I think it takes the right direction (but we're still not there yet), good job @rdswift

zas:
> One idea would be that we define a couple of file naming presets, that we offer as a dropdown, for easy configuration

Yes.

rdswift:
> Few thoughts:
> 
> * we display 10 items as examples, but the current limit (and the fact it's random) isn't really obvious, perhaps a tooltip on Renew examples would help, dunno, but that's something we need to improve

Tooltip added.  I agree that we need to improve this, but I'm not sure how to make it clearer / more intuitive at this point.

> * when selecting one entry in Before, I'd expect the corresponding entry in After to be hightlighted

Selecting an item in one list now also selects the corresponding item in the other list.  It would be better if we could force both lists to use the same color for the highlight regardless of whether or not the list is currently the selected widget, but I haven't yet figured out how to do that.  Any suggestions / hints would be appreciated.

> * Dialog should be titled "File naming script editor" (and corresponding menu item too)

Changed in both places.

> * is there a way to format the script so it isn't a loooong line?

Added a "Word wrap" checkbox to the editor.

> * buttons should perhaps be just under the script edit widget

Moved as suggested.

> * when changing selection one has to press Renew examples to update examples, perhaps we could just handle selection change signals and update on the fly when the window is displayed

Updates are now triggered automatically as files are selected / deselected.

> * when activating the script editor via button in Options the behavior is very weird (here, the dialog starts minimized, and I can't move it to front)

Still a problem.  I believe that it's because the Options window is modal, but the script editor window isn't.  That prevents accessing the script editing window while the Options window is open.  I tried a few different combinations with zero success.  I suspect that the only thing that might work is to make the Options window not modal, but I'm not sure that we want to do that.

> One idea would be that we define a couple of file naming presets, that we offer as a dropdown, for easy configuration

Dropdown added with a couple of simple examples.  We can easily edit or add more as required.  Also added tooltips to all of the pushbuttons.

> That's just thoughts after a quick test, I'll review this more in depth soon.

I look forward to it.  :grin:

rdswift:
I found out how to set the colors so that they are the same in both lists.  I tried to select a combination that looks okay in both light and dark themes, but am open to suggestions if there is a better combination.

phw:
I am not fully convinced we should make the dialog available separately as part of this PR. I'd rather do this in a separate step once we are satisfied with how it works when accessed via options, because I think this still must be possible anyway.

Regarding the file naming and scripting dialog I had started to scribble down some ideas. It is just some ideas and not fully finished or thought through, especially the script edit dialog, but because the discussion here is progressing so quickly I put it up for discussion:

scripteditor

renamingoptions

Some core ideas:

- Being a separate dialog we could show the scripting docs inside that dialog. It probably should have some way to hide it in order to get more space for the editor window.
- The file naming page would allow a selection of multiple naming scripts. We could offer 2 or 3 common defaults, but the user can edit and add new ones. This would implement PICARD-524 and PICARD-1127. The scripts have a free name, but we would set it for the defaults to something that gives an idea how the naming ends up.

In the past we had some discussion to instead of having "only" multiple naming scripts we would rather implement option profiles (PICARD-9). But I think the multiple naming scripts would fit well into what we are trying to do here. The presets would simplify the selection of a naming script for new users, currently they sometimes feel overwhelmed by the scripting. Advanced users would have multiple scripts available.

The edit script dialog should, especially if it becomes standalone, have the option to select the script. So the current input field in my mockup for the script name should rather be a editable dropdown, with the ability to add / remove entries. An easy way to change the currently active script for naming should also be available.

> Selecting an item in one list now also selects the corresponding item in the other list. It would be better if we could force both lists to use the same color for the highlight regardless of whether or not the list is currently the selected widget, but I haven't yet figured out how to do that. Any suggestions / hints would be appreciated.

I'd use one list instead with two columns. That's easier than having two lists. Also for the user, e.g. scrolling will move both entries the same, and if we allow resizing the user can choose to have one column larger than the other.

> I suspect that the only thing that might work is to make the Options window not modal, but I'm not sure that we want to do that.

Definitely not as part of this PR. In general that would be an option, but it brings with it a whole lot of complications. I once considered this when fixing window issues on macOS. Basically while options where open it was still allowed to interact with the main menu, which brought with it all kinds of problems :) Otherwise I think it would be very beneficial for quickly changing settings while tagging.

Sophist-UK:
@phw Philipp - Thanks for the wireframes - a picture says a thousand words. 

A. I really like the idea of storing multiple file naming scripts and being able to select one of them!!

B. IMO the documentation panel should be context sensitive - if you type a $ it should bring up the entire function list, if you type $i it should show a subset, if you type % it should show the list of all variables, if you type %_ the list of hidden variables, if you type %artist the subset of variables that match. We would have to work out how this should best display if you e.g. put the cursor after the $ in an existing $setmulti - should it display the full list or just $setmulti? (This could be left for a future enhancement.)

C. I do **not** think we should extend this window to cover Tagging and Cover Art scripts in this PR, but IMO this future goal should be born in mind when coding this.

phw:
In order to go forward and get something merged I think we should try to get to something step by step in separate PRs. I'd keep this PR focused to provide a simple tag editing dialog (basically how it is implemented already). Just for the exampled I'd try how this works with a two column list instead and refactor the examples to be reusable in both the file naming and the script edit dialog (maybe put it into a separate class that can be reused).

For the file naming dialog I would actually remove the script edit completely (after all it is one goal to make more space available here again) and replace it with only the button to open the script editor as an intermediate solution until we implement something more sophisticated.

Then see to get the additional features added, such as:

- Multiple naming scripts (if we decide to do this)
- Documentation inside the dialog
- Standalone dialog
- Using this dialog for cover art naming and maybe tagger scripts

For the standalone dialog things will be a bit more complicated. For one we must make sure the dialog updates whenever naming options get changed (which can happen both via Options or via main menu). And for the problems with it not showing up properly when opened from Options we might have no other choice than either have a workaround to show this dialog as modal on top of Options (in addition to already open script dialog, or close the existing dialog instead?) as well or actually making Options non-modal, which is not a trivial change. The first option is not more than a workaround, though.

If we go a route like this I'm sure we can come up with a properly working solution for a 2.7 release.

> IMO the documentation panel should be context sensitive...

Yes, something like this I had in mind. This definitely needs to be fleshed out more, but I think we could do nice things with it. I'd spare this for a future PR, though.

> C. I do **not** think we should extend this window to cover Tagging and Cover Art scripts in this PR, but IMO this future goal should be born in mind when coding this.

I have my doubts we should do it all in the exactly same dialog instance, but the dialogs definitely should share code and general layout. What is different in all those cases is the examples, and maybe we also would want to provide different documentation

rdswift:
> In order to go forward and get something merged I think we should try to get to something step by step in separate PRs. I'd keep this PR focused to provide a simple tag editing dialog (basically how it is implemented already).

Okay, I've cut it back to just being accessible from the Options screen, and set it such that you can make settings changes on the Options screen and have them reflected in the examples on the editing screen.  Tested with both screens open side-by-side.

> Just for the exampled I'd try how this works with a two column list instead and refactor the examples to be reusable in both the file naming and the script edit dialog (maybe put it into a separate class that can be reused).

The example code has been refactored into a separate class (currently in the `picard.ui.scripteditor` module).

I'm still working on using a two-column list rather than the synchronized single-column lists, but I haven't yet been able to come up with anything that provides as useful a display (e.g.: separate horizontal scroll bars on each column).  I'll keep working on it.

> For the file naming dialog I would actually remove the script edit completely (after all it is one goal to make more space available here again) and replace it with only the button to open the script editor as an intermediate solution until we implement something more sophisticated.

That's next on my list -- perhaps even before the two-column list -- because it should be pretty straightforward.

I hope to have the next version available for review sometime tomorrow.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,377 @@
...
+)
+from picard.ui.options import OptionsPage
+from picard.ui.options.scripting import (
+    ScriptCheckError,
+    ScriptingDocumentationDialog,
+)
+from picard.ui.ui_scripteditor import Ui_ScriptEditor
+
+
+PRESET_SCRIPTS = [
```
zas, line 54: This list will be editable (ideally).
So I think we should get rid of the dummy entry (it should be the current user script, named `your script` or whatever).
Presets should be appended to it (not modifiable).
The combo box could be made editable to let user change titles.

We should perhaps objectify each item:

```python
class FileNamingScript:
   def __init__(self, script='', title=None, readonly=False):
       if title is None:
           title = N_('default filenaming script title')
       self.title = title
       self.script = script
       self.readonly = readonly   # for presets
```

Just a bunch of ideas

picard/ui/ui_scripteditor.py
```diff
@@ -0,0 +1,200 @@
...
+        self.label.setText(_("Preset file naming scripts:"))
+        self.preset_naming_script_select.setToolTip(_("Load the selected preset script into the editor"))
+        self.preset_naming_script_select.setText(_("Select Preset"))
+        self.file_naming_word_wrap.setToolTip(_("Word wrap long lines in the editor"))
+        self.file_naming_word_wrap.setText(_("Word wrap text"))
+        self.example_filename_sample_files_button.setToolTip(_("Up to 10 items chosen at random from files selected in the main window"))
+        self.example_filename_sample_files_button.setText(_("Reload Examples"))
+        self.scripting_documentation_button.setToolTip(_("Display the scripting documentation in a new window"))
+        self.scripting_documentation_button.setText(_("Scripting Documentation"))
+        self.file_naming_format_reload.setToolTip(_("Reload the script currently saved in the configuartion option settings"))
```
zas, line 189:
```suggestion
        self.file_naming_format_reload.setToolTip(_("Reload the script currently saved in the configuration option settings"))
```

zas (changes requested):

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+        self.update_examples()
+
+    def check(self):
+        self.check_format()
+
+    def check_format(self):
+        config = get_config()
+        parser = ScriptParser()
+        try:
+            parser.eval(self.ui.file_naming_format.toPlainText())
```
zas, line 408: Use `get_script()`?

```python
script_text = self.get_script()
try:
    parser.eval(script_text)
except Exception as e:
    raise ScriptCheckError("", str(e))
if not script_text and config.setting["rename_files"]:
    raise ScriptCheckError("", _("The file naming format must not be empty."))
```

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+            log.debug('Importing naming script file: %s' % filename)
+            file_text = ""
+            try:
+                with open(filename, 'r', encoding='utf8') as i_file:
+                    file_text = i_file.read()
+            except OSError as error:
+                self.display_error(error)
+                return
+            if file_text.strip():
+                self.ui.file_naming_format.setPlainText(file_text.strip())
```
zas, line 376:
```suggestion
                self.set_script(file_text)
```

zas (changes requested):

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+
+    def eventFilter(self, object, event):
+        if event.type() == QtCore.QEvent.WindowActivate or event.type() == QtCore.QEvent.FocusIn:
+            self.update_examples()
+        return False
+
+    def select_preset_script(self):
+        selected_script = self.ui.preset_naming_scripts.currentIndex()
+        if selected_script > 0:
+            self.ui.file_naming_format.setPlainText(PRESET_SCRIPTS[selected_script]['script'])
```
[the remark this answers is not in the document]
rdswift: Makes sense.  The `self.set_script()` was a later add-in change, and I never thought to use it internally.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+        script_text = self.get_script()
+        if script_text:
+            options = QtWidgets.QFileDialog.Options()
+            options |= QtWidgets.QFileDialog.DontUseNativeDialog
+            filename, _ = QtWidgets.QFileDialog.getSaveFileName(self, "Export Script File", "picard_naming_script.pts", "All Files (*);;Picard Script Files (*.pts)", options=options)
+            if filename:
+                log.debug('Exporting naming script file: %s' % filename)
+                try:
+                    with open(filename, 'w', encoding='utf8') as o_file:
+                        o_file.write(self.get_script() + '\n')
```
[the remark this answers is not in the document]
rdswift: We can do that, but I was thinking more about someone saving the text file for posting to github (such as https://github.com/rdswift/picard-scripts/blob/master/naming_scripts/Bob_Swift/Bob%20Swift's%20Naming%20Script.pts) or if they wanted to use an external editor.

I assumed that the script name / script text combination would be saved to the user configuration settings, similar to what's done with the tagging scripts (and that this would follow in a later PR).

picard/ui/scripteditor.py
rdswift: Much better.  `self.get_script()` was also a later addition, and I didn't think to use it internally.

rdswift:
Now that you guys are reviewing the actual code, can I assume that we're (sort of) agreed with the general format of the page?  I was hoping to achieve some sort of consensus before any significant refactoring, optimization, and other code clean-up (which I know this needs).  I've also been holding off adding new features, which @phw has suggested should be applied with separate follow-up PRs.

zas:
> Now that you guys are reviewing the actual code, can I assume that we're (sort of) agreed with the general format of the page? I was hoping to achieve some sort of consensus before any significant refactoring, optimization, and other code clean-up (which I know this needs). I've also been holding off adding new features, which @phw has suggested should be applied with separate follow-up PRs.

To me, what you did til now is already an improvement, but I agree we didn't yet (totally) agree with something.

picard/ui/scripteditor.py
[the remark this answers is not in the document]
zas: Ok, to KISS, a simple text file is good for sharing. I misunderstood the purpose.
But I was also thinking about saving scripts "configuration" to separate files, so we could provide those as a service on the website (a repository of filenaming scripts).
That's too early. So let's stick to the simple (and good) approach for now.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+            self.current_row = self.ui.example_filename_after.currentRow()
+            self.ui.example_filename_before.setCurrentRow(self.current_row)
+
+    def save_script(self):
+        self.signal_save.emit()
+
+    def get_script(self):
+        return str(self.ui.file_naming_format.toPlainText()).strip()
+
+    def set_script(self, script_text=None):
```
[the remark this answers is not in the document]
rdswift: Actually, I modified it as per one of your later comments.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,428 @@
...
+            try:
+                with open(filename, 'r', encoding='utf8') as i_file:
+                    file_text = i_file.read()
+            except OSError as error:
+                self.display_error(error)
+                return
+            if file_text.strip():
+                self.ui.file_naming_format.setPlainText(file_text.strip())
+            else:
+                self.ui.file_naming_format.setPlainText("$noop( The import file was empty. )")
```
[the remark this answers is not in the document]
rdswift: In the None case, this now modifies the script with some default text.  Also applies when importing an empty file.

rdswift:
I think I've now reached a logical stopping point, unless you guys would like something changed.

zas:
Last iteration is pretty good for me.
There's something bothering me though:
if user selects a preset, the user script might be lost (well, not really if save isn't pressed and user restarts Picard, the script will be loaded from config, but that's unobvious).
I was thinking about a possible solution:
- replace the current `"- please select an item -"` with something like `script saved in configuration`, and loads it again.
- remove the "Select preset" button and just toggle scripts on combo box change

Apart that, I think that's a much nicer interface than the current prod one, and I'm ok with it.
What do you think @phw @Sophist-UK ?

picard/ui/options/renaming.py
```diff
@@ -147,6 +155,32 @@ def _sync_scrollbar_vert(widget, value):
...
+    def match_after_to_before(self):
+        if self.ui.example_filename_before.currentRow() != self.current_row:
+            self.current_row = self.ui.example_filename_before.currentRow()
+            self.ui.example_filename_after.setCurrentRow(self.current_row)
+
+    def match_before_to_after(self):
+        if self.ui.example_filename_after.currentRow() != self.current_row:
+            self.current_row = self.ui.example_filename_after.currentRow()
+            self.ui.example_filename_before.setCurrentRow(self.current_row)
+
```
zas, line 177: I think we can reduce code redundancy a bit here, also we need to block signals because `setCurrentRow()` will trigger change signal in turn (resulting in useless processing):

```python
    def match_after_before(self, source, target):
        """Matches selected item in target to source"""
        if source.currentRow() != self.current_row:
            self.current_row = source.currentRow()
            target.blockSignals(True)
            target.setCurrentRow(self.current_row)
            target.blockSignals(False)

    def match_after_to_before(self):
        """Sets the selected item in the 'after' list to the corresponding item in the 'before' list.
        """
        self.match_after_before(self.ui.example_filename_before, self.ui.example_filename_after)

    def match_before_to_after(self):
        """Sets the selected item in the 'before' list to the corresponding item in the 'after' list.
        """
        self.match_after_before(self.ui.example_filename_after, self.ui.example_filename_before)
```

@rdswift what do you think?

zas (changes requested):

picard/ui/options/renaming.py
rdswift: I like it!  That piece of code always bothered me because it just didn't look right, but I wasn't seeing an easy way to fix it.  This does it.  Thanks.

rdswift:
Implemented suggested changes:

- Include current configuration naming script in selection list
- Update script loaded into the editor on selection change
- Only update script on "Make it so" if it has been saved from editor
- Optimize synchronization of example lists
- Correct spelling mistakes

zas:
Much better imho, thanks @rdswift 

I think it would be better to keep the Show documentation checkbox at the same place (on the right), because when you check it, it moves to center, and if you re-click without moving the mouse cursor, it hasn't the expected behavior (which is unchecking it) but instead open documentation in the browser.

image
image

rdswift:
> I think it would be better to keep the Show documentation checkbox at the same place (on the right), because when you check it, it moves to center, and if you re-click without moving the mouse cursor, it hasn't the expected behavior (which is unchecking it) but instead open documentation in the browser.

You're right.  It makes more sense to fix the position of the checkbox option.  Updated accordingly, and updated screen shot above.

phw (changes requested):
I finally tried this live, and I like the overall feeling of using this very much. The expandable documentation is great, saving/loading to files is a useful enhancement, the scripting dropdown gives a feeling how this could be in a final version.

Given that the 2.7.x branch is open for development and we expect a longer cycle I am all in favor of merging this. The only real change I would do here is regarding window modality, see my comments. Otherwise I see this going into the right direction. If we have this merged we have a clean state again on which we can iterate.

Those are the things to tackle after this has been merged but before a relaese:

- Make actual use of allowing multiple scripts, instead of the hard coded list
- The file naming preview syncing of selection can behave funny, if you drag and drop. Overall I think we can improve on this
- Some changes to the renaming dialog still. Not 100% sure, but I think once we allow for actual multiple naming scripts being created by the user we can have this dropdown there as in my mockup.
- Window modality (basically making the script editor non-modal in a way that works on all platforms without users loosing the window). If this gets too complicated we could do it later. But it's something I definitely want to see in a separate PR focusing in this issue alone.

Further improvements could be, but maybe at a later time:

- More clever documentation view (context sensitive? searchable?)
- We had the general idea of storing scripts as files on the file system, allowing to edit them both inside Picard and with external editor. 
- Make use of this dialog for cover art naming scripts
- Use a similar dialog for tagger scripts (but this needs more thought)

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,584 @@
...
+
+    def __init__(self, parent=None, examples=None):
+        """Stand-alone file naming script editor.
+
+        Args:
+            parent (QMainWindow or OptionsPage, optional): Parent object. Defaults to None.
+            examples (ScriptEditorExamples, required): Object containing examples to display. Defaults to None.
+        """
+        super().__init__(parent)
+        self.PARENT = parent
```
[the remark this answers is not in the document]
rdswift: Removed from both the `ScriptEditorExamples` and `ScriptEditorPage` classes.  These were leftovers from earlier versions and were no longer used in either class any longer.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,584 @@
...
+        self.examples = examples
+        self.setWindowFlags(QtCore.Qt.Window)
```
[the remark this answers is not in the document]
rdswift: Revised as suggested.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,584 @@
...
+        self.installEventFilter(self)
+
+        self.ui.file_naming_editor_save.clicked.connect(self.save_script)
+
+        self.ui.file_naming_format.setEnabled(True)
+
+        text = '<a href="' + PICARD_URLS['doc_scripting'] + '">' + N_('Open Scripting Documentation in your browser') + '</a>'
+        self.ui.scripting_doc_link.setText(text)
+
+        def process_html(html, function):
```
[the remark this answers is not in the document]
rdswift: Moved to a new custom widget, although I'm not sure if this is exactly what you had in mind.  If this is correct, then we should probably also revise `picard/ui/options/scripting.py` accordingly, and move the new widget into a separate file in `picard/ui/widgets'.  Just let me know and I can make the changes.

rdswift:
> I finally tried this live, and I like the overall feeling of using this very much. The expandable documentation is great, saving/loading to files is a useful enhancement, the scripting dropdown gives a feeling how this could be in a final version.

Thanks.  It has come along nicely, but much of that is due to the reviews and suggestions provided.

> If we have this merged we have a clean state again on which we can iterate.

That would be great.  I already have a start (preliminary working version) of some of the items in a local branch (not yet published to GitHub), including:

* Added a file naming script class.
* Add titles to scripts as shown in the mock-up above by @phw (thanks for that).
* Allow multiple user-defined scripts along with the hard-coded system examples.  Ability to add, copy, and remove items in the list (with `system` scripts unable to be removed).
* Ability to import/export scripts from/to local files in both plain text and JSON formats (which include titles and possible other metadata -- still pretty early stages of this).

I will hold off on submitting anything for consideration until we finish with this PR in order to avoid the dreaded "scope creep".  :wink:

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,630 @@
...
+class ScriptingDocumentationWidget(QtWidgets.QWidget):
+    """Custom widget to display the scripting documentation.
+    """
+    def __init__(self, parent, *args, **kwargs):
+        """Custom widget to display the scripting documentation.
+
+        Args:
+            parent (QWidget): Frame widget in which the documentation is displayed.
+        """
+        super(ScriptingDocumentationWidget, self).__init__(*args, **kwargs)
```
zas, line 556: You can use `super().__init__(*args, **kwargs)` here.

zas (changes requested):

zas:
@rdswift unrelated to the PR itself, but please do not title your commits with "Changes requested by ...", "Implement suggested changes" or the like, just describe what the commit does. Such title is only informative in the context of this PR, but in the commit history it wouldn't help much.
Instead, if you want to credit someone, do it in the commit message, but not in the first line.
Also, when implementing different suggested changes, please still keep multiple small commits (smaller commits help when using git bisect, and also for code review).

For 212af58b026942d994f938 I would expect at least 4 commits, titled:
- remove SingletonDialog from ScriptEditorPage class
- remove unused self.parent and self.PARENT variables
- remove setWindowFlags and add setWindowModality
- create custom ScriptingDocumentationWidget

And in the commit message of each of them "Changes requested by outsidecontext"

Just saying because that's 2 times in this PR, I don't want it to become a (bad) habit ;)

picard/ui/scripteditor.py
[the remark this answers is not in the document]
phw: Yes, that's mostly what I meant. But the structure currently is a bit odd. The Widget does not really act like a widget, but acts as a helper class that ads widgets to the given parent.

Instead the Widget should add the proper layouts to itself and display all the subwidget in itself (so it is the parent of its sub widgets). Then your dialog or the old scripting docs dialog can do something like this:

```python
doc_widget = ScriptingDocumentationWidget(self)
self.some_layout.addWidget(doc_widget)
```

That makes this truely reusable without ScriptingDocumentationWidget relying on some external structure.

rdswift:
> just describe what the commit does. Such title is only informative in the context of this PR, but in the commit history it wouldn't help much.

Sorry about that.  I obviously wasn't thinking beyond this PR.  I'll try to do a better job in the future.

picard/ui/scripteditor.py
```diff
@@ -460,9 +460,12 @@ def toggle_wordwrap(self):
     def import_script(self):
         """Import the current script from an external text file.
         """
+        default_path = os.path.normpath(QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.DocumentsLocation))
```
zas, line 463: Perhaps we should make those properties of the class:

```python
default_script_directory = os.path.normpath(QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.DocumentsLocation))
default_script_filename = "picard_naming_script.pts"
```

So it can be overriden by subclassing, and reduce code redundancy.

zas (changes requested):

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,642 @@
...
+        """Export the current script to an external text file.
+        """
+        script_text = self.get_script()
+        if script_text:
+            default_path = os.path.normpath(os.path.join(self.default_script_directory, self.default_script_filename))
+            dialog_title = N_("Export Script File")
+            dialog_file_types = N_("All Files") + " (*);;" + N_("Picard Script Files") + " (*.pts)"
+            options = QtWidgets.QFileDialog.Options()
+            options |= QtWidgets.QFileDialog.DontUseNativeDialog
+            filename, _ = QtWidgets.QFileDialog.getSaveFileName(self, dialog_title, default_path, dialog_file_types, options=options)
```
zas, line 491: Don't use `_` for unused variables, as it is defined as builtin `_()` for i18n purposes (gettext).

zas (changes requested):

picard/ui/scripteditor.py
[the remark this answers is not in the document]
rdswift: I think this latest commit c4b644d fixes it.  If this is correct, should I move it to a separate file in `picard/ui/widgets`?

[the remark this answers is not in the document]
rdswift: Sorry, I got carried away and made the changes.  Can revert if it was inappropriate.

rdswift:
Sorry for the rather minor commit, but I found out that the PR will not be updated if I push multiple commits to the branch on my repository.  It will only update if there is a single commit in the push.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,549 @@
...
+        # Sync example lists vertical scrolling
+        sync_vertical_scrollbars((self.ui.example_filename_before, self.ui.example_filename_after))
+
+        # Set highlight colors for selected list items
+        example_style = self.ui.example_filename_before.palette()
+        highlight_bg = example_style.color(QPalette.Active, QPalette.Highlight)
+        highlight_fg = example_style.color(QPalette.Active, QPalette.HighlightedText)
+        stylesheet = "QListView::item:selected { color: " + highlight_fg.name() + "; background-color: " + highlight_bg.name() + "; }"
+        self.ui.example_filename_after.setStyleSheet(stylesheet)
+        self.ui.example_filename_before.setStyleSheet(stylesheet)
```
zas, line 337: This code is (partly?) duplicated: https://github.com/metabrainz/picard/pull/1792/files#diff-55ef3fe9f30be58c257e46963eb58be72363347b468d02a15c9485cf4143713dR141-R166

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,549 @@
...
+
+        # Set splitter background color to highlight background color to make splitters more visible (disabled because it looks ugly)
+        # stylesheet = "QSplitter::handle { background: " + highlight_bg.name() + "; }"
+        # self.ui.splitter_between_editor_and_examples.setStyleSheet(stylesheet)
+        # self.ui.splitter_between_editor_and_documentation.setStyleSheet(stylesheet)
+        # self.ui.splitter_between_before_and_after.setStyleSheet(stylesheet)
```
zas, line 343: If code is commented out, better remove it from source (but keep it in commit history)

zas (changes requested):
I tested the code again, and it works very well.
There are still duplicated code though, after some cleanup I'm ok for a merge, a lot of work was done already on this.

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,547 @@
...
+        self._sampled_example_files = []
+        config = get_config()
+        self.settings = config.setting
+        self.example_list = []
+
+    def update_sample_example_files(self):
+        """Get a new sample of randomly selected / loaded files to use as renaming examples.
+        """
+        import random
+        max_samples = 10  # pick up to 10 samples
```
zas, line 101: I would move this to class property and use it instead of hardcoded (10)  in following strings:

```
picard/ui/ui_options_renaming.py:        self.label_2.setText(_("If you select files from the Cluster pane or Album pane prior to opening the Options screen, up to 10 files will be randomly chosen from your selection as file naming examples.  If you have not selected any files, then some default examples will be provided."))
picard/ui/ui_scripteditor.py:        self.example_filename_sample_files_button.setToolTip(_("Up to 10 items chosen at random from files selected in the main window"))
```

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,547 @@
...
+
+        self.ui.preset_naming_scripts.clear()
+        for item in PRESET_SCRIPTS:
+            self.ui.preset_naming_scripts.addItem(item["title"])
+        self.ui.preset_naming_scripts.setCurrentIndex(0)
+        self.ui.preset_naming_scripts.currentIndexChanged.connect(self.select_script)
+
+        self.synchronize_vertical_scrollbars((self.ui.example_filename_before, self.ui.example_filename_after))
+
+        self.wordwrap = QtWidgets.QTextEdit.NoWrap
```
zas, line 316: Is this really needed?

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,547 @@
...
+        self.update_example_listboxes(self.ui.example_filename_before, self.ui.example_filename_after, examples)
+
+        if send_signal:
+            self.signal_update.emit()
+
+    def toggle_wordwrap(self):
+        """Toggles wordwrap in the script editing textbox.
+        """
+        if self.ui.file_naming_word_wrap.isChecked():
+            self.wordwrap = QtWidgets.QTextEdit.WidgetWidth
```
zas, line 449: What's the point of storing the value as a property? It isn't preserved anyway (perhaps it should be though).

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,547 @@
...
+        else:
+            self.wordwrap = QtWidgets.QTextEdit.NoWrap
+        self.ui.file_naming_format.setLineWrapMode(self.wordwrap)
+
+    def import_script(self):
+        """Import the current script from an external text file.
+        """
+        dialog_title = N_("Import Script File")
+        dialog_file_types = N_("All Files") + " (*);;" + N_("Picard Script Files") + " (*.pts)"
```
zas, line 458: Those strings are marked for translation (using `N_()`) but never actually translated (or I miss something).
Also `dialog_file_types` has the same value in `export_script()` so perhaps it could be made common.

zas (changes requested):

picard/ui/scripteditor.py
rdswift: Set as a class property as suggested.  Also set the usage notes and tool tip strings as class properties so that they are consistent and only need to be translated once.

picard/ui/scripteditor.py
rdswift: Nope.  Left over from an earlier version of the code.  Removed.

picard/ui/scripteditor.py
rdswift: Removed for now.  I was thinking that perhaps some of the screen configuration settings should be preserved, but I was leaving that for a future PR.

picard/ui/scripteditor.py
rdswift: The selected file type is returned from the dialog and I had planned to use this on export to determine whether to write the file as plain text or JSON.  Since the JSON option isn't available right now, I'll make these a common property used in both import and export functions.

zas (approved):
To me, ready to merge, thanks for the fantastic job on this @rdswift

zas (approved):

zas (approved):

picard/ui/scripteditor.py
```diff
@@ -0,0 +1,554 @@
...
+
+    @staticmethod
+    def update_example_listboxes(before_listbox, after_listbox, examples):
+        before_listbox.clear()
+        after_listbox.clear()
+        for before, after in sorted(examples, key=lambda x: x[1]):
+            before_listbox.addItem(before)
+            after_listbox.addItem(after)
+
+    def display_examples(self, send_signal=True):
```
zas, line 440: I don't think we ever have `send_signal == False` in current code. Perhaps this option can be removed.

rdswift: Good catch!  That was a leftover from a previous version of the code.  You're right, it's no longer used, so I will remove it.  Thanks.

picard/ui/scripteditor.py
```diff
@@ -437,7 +437,7 @@ def update_example_listboxes(before_listbox, after_listbox, examples):
             before_listbox.addItem(before)
             after_listbox.addItem(after)
 
-    def display_examples(self, send_signal=True):
+    def display_examples(self):
         """Update the display of the before and after file naming examples.  Optionally emits an `update` signal.
 
         Args:
```
zas, line 443: @rdswift remove this part of the docstring too as `send_signal` is gone

rdswift: Done.  Thanks.

picard/ui/options/renaming.py
```diff
@@ -117,35 +118,49 @@ def __init__(self, parent=None):
...
-                for other in widgets - {widget}:
-                    widget.verticalScrollBar().valueChanged.connect(
-                        partial(_sync_scrollbar_vert, other))
-
-        # Sync example lists vertical scrolling
-        sync_vertical_scrollbars((self.ui.example_filename_before, self.ui.example_filename_after))
+
+        self.ui.example_filename_sample_files_button.clicked.connect(self.update_example_files)
+
+        self.examples = ScriptEditorExamples(self, self)
```
zas, line 135: Why `(self, self)` here?

```python
def __init__(self, parent=None, tagger=None):
```

https://github.com/metabrainz/picard/pull/1792/files#diff-b826bd41b92baa93cfac5658a55109f4926829b68a630d63a2ac342c17c0980fR91

rdswift: Actually, the `parent` argument was never used so I removed it.  Used keywords as you suggested and now pass the `tagger` object directly rather than the parent of the `tagger` option.

phw (approved):
This looks and works well, great work.

I just added a commit https://github.com/metabrainz/picard/pull/1792/commits/7d5531367e736ca1f202b020ee291f60df843412 to remove some of the constants from `ScriptEditorPage` that are unused, because this is a separate dialog and not an options page. Regarding the help I think we should add back a help button to this dialog which then can open the appropriate help page in the docs, but I think we can do this in a later iteration. Maybe after we know how docs will be changed to reflect this new UI.

PR #1794, opened 2021-04-13
Cover art number column rework

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/item.py
+++ b/picard/ui/item.py
@@ -127,10 +127,10 @@ class Item(object):
 
         number_of_images = len(self._images)
         if getattr(self, 'has_common_images', True):
-            return ngettext("; %i image)", "; %i images)",
+            return ngettext("%i image", "%i images",
                             number_of_images) % number_of_images
         else:
-            return ngettext("; %i image not in all tracks)", "; %i different images among tracks)",
+            return ngettext("%i image not in all tracks", "%i different images among tracks",
                             number_of_images) % number_of_images
 
 
--- a/picard/ui/item.py
+++ b/picard/ui/item.py
@@ -100,27 +100,38 @@ class Item(object):
     def clear_errors(self):
         self._errors = []
 
-    def _cover_art_description(self, images, detailed=True):
-        """Return the number of cover art images in `images` for display in the UI
+    @property
+    def _images(self):
+        return self.metadata.images
 
-        Args:
-            images: An instance of picard.util.imagelist.ImageList
-            detailed: If True, return a detailed text about the images and whether they are the same across all tracks.
-                    If False only return the number as a string.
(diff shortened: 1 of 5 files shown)


picard/ui/item.py
```diff
@@ -100,27 +100,38 @@ def error_append(self, msg):
     def clear_errors(self):
         self._errors = []
 
-    def _cover_art_description(self, images, detailed=True):
-        """Return the number of cover art images in `images` for display in the UI
+    @property
+    def _images(self):
+        return self.metadata.images
```
phw, line 105: When I suggested passing `self.metadata.images`  to the helper methods I had two issues in mind:

1. I wasn't sure if we should rely upon `Item`  having a `metadata` attribute. Of course classes using those helper methods currently always would have

2. Wanted to keep it flexible so it can also be used for other metadata objects e.g. (self.orig_metadata)

Of course code otherwise becomes a bit nicer this way. Thoughts?

zas: I thought about it too, this is why I introduced `_images`:
- we could check for `metadata.images` attributes there
- or alternatively using `orig_images`

But, for now, I see only one case we may want to do that, when there's no images in metadata but some in orig_metadata, like we do in https://github.com/metabrainz/picard/blob/de45f3cd9d0a5ba5953d47d5c231e37c83a8bf63/picard/ui/infodialog.py#L125-L142

picard/ui/item.py
```diff
@@ -100,27 +100,38 @@ def error_append(self, msg):
...
-            else:
-                return ngettext("; %i image not in all tracks)", "; %i different images among tracks)",
-                                number_of_images) % number_of_images
+        if not self.can_show_coverart:
+            return ''
+
+        number_of_images = len(self._images)
+        if getattr(self, 'has_common_images', True):
+            return ngettext("; %i image)", "; %i images)",
+                            number_of_images) % number_of_images
```
phw, line 131: We don't need to do that in this PR, but we should modify the way this strings gets build with the semicolon and different parts to make it more translation friendly.

phw (approved):

picard/ui/item.py
zas: Actually I'll make it in this PR, since the current way of doing things doesn't make much sense.
Check e720c78 and tell me if it's better.

I'm not sure how to test this, in which case does those strings appear??

phw: Yes, much better!

> I'm not sure how to test this, in which case does those strings appear??

Album title, various cases:

grafik

phw (approved):

PR #1798, opened 2021-04-30
PICARD-2171: Optional guess title and tracknumber

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2171

If tracknumber and title are not present in tags, Picard will guess those from the filename. Not all users might always want this automation, see PICARD-2170

# Solution

Add an option to enable/disable guessing title and tracknumber from filename, enabled by default.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -78,7 +78,7 @@ from picard.util import (
     is_absolute_path,
     samefile,
     thread,
-    tracknum_from_filename,
+    tracknum_and_title_from_filename,
 )
 from picard.util.filenaming import (
     make_short_filename,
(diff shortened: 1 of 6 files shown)


phw:
I hadn't noticed the existing test_tracknum tests. I need to check how to deal with them, I don't think I agree with all of the cases there.

phw:
I removed the part of changing tracknumber parsing (PICARD-1568), I'll work on this separately at https://github.com/metabrainz/picard/pull/1799 .

So this is focused on adding the option only.

picard/file.py
```diff
@@ -225,24 +225,21 @@ def _loading_finished(self, callback, result=None, error=None):
         self.update()
         callback(self)
 
-    def _copy_loaded_metadata(self, metadata):
-        filename, _ = os.path.splitext(self.base_filename)
+    def _copy_loaded_metadata(self, metadata, guess_tracknumber_and_title=True):
```
zas, line 228: Instead of passing this option, I'd pass a list of functions to run on metadata before copying it. It would be much more generic.

Something like:

```python
def _copy_loaded_metadata(self, metadata):
    metadata['~length'] = format_time(metadata.length)
    for processor in self.copy_loaded_metadata_processors:
          processor(metadata)
```

And in `_loading_finished()`:

```python
self.copy_loaded_metadata_processors = list()
if config.setting["guess_tracknumber_and_title"]:
  self.copy_loaded_metadata_processors.append(self._guess_tracknumber_and_title)
```

phw: Good idea, that parameter made me unhappy already. I'll change that.

phw: I did it a bit different in https://github.com/metabrainz/picard/pull/1798/commits/a73e65fcc3594923823081769c4ab22faad8b7d6 . I wanted to avoid using an instance variable to implicitly pass around parameters, instead do this explicitly.

zas (approved):

PR #1799, opened 2021-04-30
PICARD-1568: improved tracknum from filename

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1568

# Solution

Try to improve `util.tracknum_from_filename` by:

1. Remove the check, where all numbers of the track get extracted and the smallest is used. This led to many false positives, such as "Symphony no. 5 in D minor" being detected as track no. 5.
2. Concentrate on well defined cases
3. Consider the typical case of file names starting with both disc no. and track no. (e.g. "2-09 Song.mp3" for track 9 on disc 2)
4. Add more tests, especially also for some known bad cases

This will never be perfect, but I think this works for more cases then previously. Especially I want to avoid to many false positives, especially for those cases that are clearly wrong.

# Action


phw:
I extended the tests and added most of the tests that previously where in the separate `test_tracknum.py` file (which I overlooked, as it was in unusually named).

Some tests changed. E.g. `song-70s 69 comment.mp3` previously counted as track no. 69. But because I removed the logic to use the smallest of all numbers found in the name it is now None. And I think that's fine. This is a really inconconclusive case. code tries to avoid false positives at the cost of some false negatives.

picard/util/__init__.py
```diff
@@ -380,11 +380,17 @@ def iter_unique(seq):
 # order is important
 _tracknum_regexps = (
     # search for explicit track number (prefix "track")
-    r"track[\s_-]*(?:no|nr)?[\s_-]*(\d+)",
-    # search for 2-digit number at start of string
-    r"^(\d{2})\D",
-    # search for 2-digit number at end of string
-    r"\D(\d{2})$",
+    r"track[\s_-]*(?:(?:no|nr)\.?)?[\s_-]*(?P<number>\d+)",
```
zas, line 383: I wonder if it is worth the hassle, but perhaps we could compile regexps at this level:

```python
_tracknum_regexps = (
    # search for explicit track number (prefix "track")
    re.compile(r"track[\s_-]*(?:(?:no|nr)\.?)?[\s_-]*(?P<number>\d+)", re.I),
....

    for r in _tracknum_regexps:
        match = r.search(filename)
        if match:
....
```

zas (dismissed):

picard/util/__init__.py
phw: Actually yes, I already did this, but looks like I forgot to push it :)

phw: Please check https://github.com/metabrainz/picard/pull/1799/commits/b35acf62cf1afc55e8a37c935928171ff8b34b24

zas (approved):

picard/util/__init__.py
zas: Looks good :+1:

PR #1800, opened 2021-04-30
PICARD-1718: crash handler

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1718

This is another submission of https://github.com/metabrainz/picard/pull/1466

Overall I'm actually for now finally merging this. If Picard suddenly crashes we and the users are usually flying in the dark. It is difficult to get users to start Picard from command line to get a proper exception stack trace, and depending on circumstance and setup (e.g. Windows portable or store versions) even impossible.

The patch as is I had now several times applied locally to narrow down a reported crash.

As I mentioned in #1466 there are still cases that might not be caught, but given that it improves the debugging a lot for the majority of cases let's merge this.

I reduced code duplication a bit compared to #1466. This makes this slightly more likely to fail, but picard/__init__.py is luckily rather minimal. We are trying to do something useful in case of a hard crash. Worst thing happening is that we crash without logging or displaying the details, which is the status quo.

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -87,3 +87,48 @@ api_versions = [
 ]
 
 api_versions_tuple = [Version.from_string(v) for v in api_versions]
+
+
+def crash_handler():
+    """Implements minimal handling of an exception crashing the application.
+    This function tries to log the exception to a log file and display
+    a minimal crash dialog to the user.
+    This function is supposed to be called from inside an except blog.
+    """
+    # First try to get traceback information and write it to a log file
+    # with minimum chance to fail.
+    import sys
+    from tempfile import NamedTemporaryFile
+    import traceback
+    trace = traceback.format_exc()
+    logfile = None
+    try:
+        with NamedTemporaryFile(suffix='.log', prefix='picard-crash-', delete=False) as f:
+            f.write(trace.encode(errors="replace"))
+            logfile = f.name
+    except:  # noqa: E722,F722 # pylint: disable=bare-except
+        print("Failed writing log file {0}".format(logfile), file=sys.stderr)
+        logfile = None
+
+    # Display the crash information to the user as a dialog. This requires
+    # importing Qt5 and has some potential to fail if things are broken.
+    from PyQt5.QtCore import Qt, QUrl
+    from PyQt5.QtWidgets import QApplication, QMessageBox
+    # assigning QApplication to a variable is required to keep the object alive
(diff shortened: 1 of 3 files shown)


picard/__init__.py
```diff
@@ -87,3 +87,48 @@ def version_from_string(version_str):
...
+    except:  # noqa: E722,F722 # pylint: disable=bare-except
+        print("Failed writing log file {0}".format(logfile), file=sys.stderr)
+        logfile = None
+
+    # Display the crash information to the user as a dialog. This requires
+    # importing Qt5 and has some potential to fail if things are broken.
+    from PyQt5.QtCore import Qt, QUrl
+    from PyQt5.QtWidgets import QApplication, QMessageBox
+    # assigning QApplication to a variable is required to keep the object alive
+    _unused = QApplication(sys.argv)  # noqa: F841
```
zas, line 118: Overall it seems to work, but I noticed something: when exiting by closing the message box, the application coredumps.
I think that's because QApplication.quit() is never called.

picard/__init__.py
```diff
@@ -87,3 +87,48 @@ def version_from_string(version_str):
...
+    msgbox = QMessageBox()
```
zas, line 119: I think we could make this message resizeable (it isn't).

I tried various things, but this one seems to work:

```python
from PyQt5.QtWidgets import (
    QMessageBox,
    QSizePolicy,
    QTextEdit,
)


class ResizeableMessageBox(QMessageBox):
    """Resizeable QMessageBox"""
    # Based on Serge Gubenko's answer at https://stackoverflow.com/a/2664019

    def __init__(self, *args, **kwargs):
        super().__init__(*args, **kwargs)
        self.setSizeGripEnabled(True)

    def event(self, e):
        result = super().event(e)

        def setsize(obj):
            obj.setMinimumHeight(0)
            obj.setMaximumHeight(16777215)
            obj.setMinimumWidth(0)
            obj.setMaximumWidth(16777215)
            obj.setSizePolicy(QSizePolicy.Expanding, QSizePolicy.Expanding)

        setsize(self)

        textedit = self.findChild(QTextEdit)
        if textedit is not None:
            setsize(textedit)

        return result
```

zas (changes requested):

phw:
So let's go with this. Even if it just would handle 50% of the difficult crashes I'd be happy :D

PR #1801, opened 2021-04-30
Fix linter errors

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Addresses some pylint and bandit warnings.

Also disables the "method could be a function (no-self-use)" pylint warning. While often correct, I find this mostly annoying. I don't think having

```
class Foo:

  def some_method(self):
       do_something()
```

is really bad, even if it does not use self. In my experience this often is the result of

- Base classes defining methods that are supposed to be overridden
- Use of self inside a classe's methods changes during refactorings back and forth

But if we want to keep this warning we should be strict and either make all occurrences of this warning functions / staticmethod or explicitly mark them as wanted (e.g. for the base class case). Currently the no-self-use is responsible for the majority of warnings on codacy and adds a lot of noise.


picard/coverart/__init__.py
```diff
@@ -158,7 +158,7 @@ def next_in_queue(self):
                 finally:
                     if ret != CoverArtProvider.WAIT:
                         self.next_in_queue()
-                    return
+                return
```
zas, line 161: This `return` was in the `finally` block, now it is outside (well, I didn't audit the whole code to see what it changes or not).

picard/cluster.py
```diff
@@ -475,7 +475,7 @@ def add(self, word):
         index, count = self.words[word]
         if index == -1:
             token = self.tokenize(word)
-            if token == '':
+            if token == '':  # nosec
```
zas, line 478: Which warning was it printing??

picard/coverart/__init__.py
phw: In the normal flow this is the same, the return runs after the exception handling. But the former hid the exception completely. It shouldn't matter in this case, as we are catching `BaseException`  anyway. But the difference is visible in this code:

```
>>> def func():
...   try:
...     print("try")
...     raise Exception("foo")
...   finally:
...     print("finally")
...     return
... 
>>> func()
try
finally
```

The raised exception is completely invisible, as if it was catched.

Compare to this:

```
>>> def func2():
...   try:
...     print("try")
...     raise Exception("foo")
...   finally:
...     print("finally")
...   return
... 
>>> func2()
try
finally
  File "<stdin>", line 4, in func2
Exception: foo
```

picard/cluster.py
phw: Bandit B105: Test for use of hard-coded password strings

It checks for variable name "token". We could rename the variable also, but it is fitting in this case. It's kind of a stupid check, really.

zas (approved):

PR #1802, opened 2021-05-03
Replace QThread with Python threads using concurrent.futures

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

This aims to replace QThread with concurrent.futures to reduce dependency on Qt for core functionality outside of actual UI.

# Solution

The implementation uses concurrent.futures, but maintains the existing API interface exposed by `picard.util.threads.run_task`. For normal usage of the threading feature (as this is currently being used by Picard itself and from the plugins we host) this should be backward compatible.

Differences:

- If a caller provides its own value for the `thread_pool` parameter this now needs to be a concurrent.futures.Executor. So if a plugin makes use of this (existing plugins do not) by passing a `QThreadPool` instance it would be incompatible.
- `run_task` now returns an instance of `concurrent.futures.Future`. This can be useful for the caller to get more control, e.g. to add a done callback or cancel the task. Previously this did not return anything, so this should not be a real incompatibility, rather offer new flexibility.
- The `priority` parameter becomes obsolete, as there is no equivalent for this in concurrent.futures.

# Action
We could provide a separate new interface to replace `picard.util.threads.run_task` with something more suited. We can also consider running tasks in separate processes by using `ProcessPoolExecutor` instead of `ThreadPoolExecutor`. But this requires some thought to the code being invoked here.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -44,6 +44,7 @@
 
 import argparse
 from collections import defaultdict
+from concurrent.futures import ThreadPoolExecutor
 from functools import partial
 import logging
 import os.path
@@ -182,21 +183,18 @@ class Tagger(QtWidgets.QApplication):
         if picard_args.debug or "PICARD_DEBUG" in os.environ:
             self.set_log_level(logging.DEBUG)
 
-        # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
-        # It's a valid reference, but its start() method doesn't work.
-        self.thread_pool = QtCore.QThreadPool(self)
+        # Default thread pool
+        self.thread_pool = ThreadPoolExecutor()
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
         # expects instant feedback instead of waiting for a long list of
         # operations to finish.
-        self.priority_thread_pool = QtCore.QThreadPool(self)
-        self.priority_thread_pool.setMaxThreadCount(1)
+        self.priority_thread_pool = ThreadPoolExecutor(max_workers=1)
 
         # Use a separate thread pool for file saving, with a thread count of 1,
         # to avoid race conditions in File._save_and_rename.
-        self.save_thread_pool = QtCore.QThreadPool(self)
-        self.save_thread_pool.setMaxThreadCount(1)
+        self.save_thread_pool = ThreadPoolExecutor(max_workers=1)
 
         if not IS_WIN:
             # Set up signal handling
@@ -375,9 +373,9 @@ class Tagger(QtWidgets.QApplication):
         self.stopping = True
         log.debug("Picard stopping")
         self._acoustid.done()
-        self.thread_pool.waitForDone()
-        self.save_thread_pool.waitForDone()
-        self.priority_thread_pool.waitForDone()
+        self.thread_pool.shutdown()
+        self.save_thread_pool.shutdown()
+        self.priority_thread_pool.shutdown()
         self.browser_integration.stop()
         self.webservice.stop()
         self.run_cleanup()
--- a/picard/util/thread.py
+++ b/picard/util/thread.py
@@ -9,6 +9,7 @@
 # Copyright (C) 2016 Sambhav Kothari
 # Copyright (C) 2017 Sophist-UK
 # Copyright (C) 2018 Vishal Choudhary
+# Copyright (C) 2020 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -25,14 +26,16 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
+from functools import partial
 import sys
 import traceback
 
 from PyQt5.QtCore import (
     QCoreApplication,
     QEvent,
-    QRunnable,
 )
+
+from picard import log
 
 
 class ProxyToMainEvent(QEvent):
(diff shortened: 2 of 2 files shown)


Sophist-UK:
I have some thoughts about running tasks in separate processes - but that is not a discussion for here.

However I am unclear what the purpose of switching from QThreads to Python Threads - you say it is to reduce dependencies on Qt except for UI but I am unclear why this is beneficial.

phw:
> However I am unclear what the purpose of switching from QThreads to Python Threads - you say it is to reduce dependencies on Qt except for UI but I am unclear why this is beneficial.

@zas and I had discussing on several occasions about reducing the dependency on Qt for core, non-GUI functionality in Picard and relying more on Python's standard libraries for this. Beside threading the other two areas we had been discussing are networking and configuration management.

I think the ideas overall are twofold: For one we had been struggling with bugs and workarounds with QSettings and QNetwork, and we think that we can implement Picard's requirements here without relying on Qt. Another issue is also the recently unclear situation around Qt's new policy in regard to LTS releases. 

The other part is that we in general want to move into a direction where Picard's core functionality can be separated from the UI logic, and I don't think such a separated logic does need or benefit from Qt.

This particular change is only a rather minor part of the puzzle. I originally wrote this patch last year when we had discussions about Picard's threading and multi processing. For one I wanted to see if such a change has any impact, negative or positive, on the performance (it doesn't).

The other reason was that `concurrent.futures` offers us a better API we can benefit from in the future. For one we can have a Future object, that can be passed around, potentially be cancelled and callbacks added to it. This will make it much easier to implement features to e.g. cancel individual tasks. In addition we can easily gain the ability to run tasks in separate processes by using `ProcessPoolExecutor`, essentially having the same API to run parallel tasks, independent of whether those run in a thread of the same process or on another process.

zas (approved):
It looks good, I didn't find any issue while testing.

Sophist-UK:
@phw Thanks for the explanation. For what it's worth (which is probably not much) I tend to agree - because I think in the end we will have more flexibility with Python native stuff.

phw:
I commented a bit on https://github.com/phw/picard/pull/8#issuecomment-832145337 what I think is the direction to move this.

This patch really is meant as a way to replace the underlying library while staying as compatible as possible with existing code.

Sophist-UK:
> Still thinking about this PR, not entirely sure yet.
> 
> If we start to break compatibility in `run_task` we should probably also remove the `priority` parameter completely.
> 
> My original thinking was to keep this compatible for the most part (at least to current usage), and then introduce a new module, let's say `util.task`, with an incompatible API. That module would allow both running in threads and processes. In regards to different thread pools I like your approach, but I would go one step further and move the management of those pools completely into this new module. So tagger would not maintain a list of pools (and the module then needs a dependency on tagger), but instead it would ask the task module to create such a pool with something like:
> 
> ```
>  picard.util.tasks.register_thread_pool('default')
>  picard.util.tasks.register_thread_pool('priority', max_workers=1)
>  picard.util.tasks.register_process_pool('background')
> ```
> 
> Just a rough idea, but that's kind of what I originally had in mind.

I think that you are both right - that we should be thinking about how to provide a whole new low-level architecture for worker tasks (which could be threads or processes) and how data / control / signals are transferred between them.

The problem with maintaining a backwards compatible API without creating a Version break is that you will never be able to get rid of it. Someone will still be running an old version of Picard and plugins which use the new API will be broken.

So I suspect that we should really be thinking about Picard V3 where we rearchitect the underlying APIs and update all existing plugins to use them. The worker tasks API is a key one if we want to take advantage of multi-processors, but there may well be other APIs we would like to update as well. For example, we use Qt for web calls - should we be switching to a Python native approach for that? Do we want to go much further in separating the UI from other code? Do we want to extend the plugin concept to make other things work with standard plugins - like genres and lyrics? Or perhaps make the web services data more modular e.g. modules for releases, release groups, works, recordings, artists etc.?

Are there major extensions to scripting that need a version change? For example, suppose we wanted to give people direct access to the webservice metadata so that they can easily access any metadata that is included not just that which is extracted as standard by Picard (and which would otherwise need a plugin)?

Indeed, do we want to make Picard so plugin oriented that you could replace ALL plugins with e.g. a complete new set related to e-books so that you can tag your ebook collection using metadata in bookbrainz?

phw:
Some breaking changes in this specific area are not so bad actually. It is not functionality used very much by plugins, and the few that do can be updated. And keeping thread.py around for a wile in a compatible way is no big deal either. I mean it is 60 lines of code or so that needed very little maintencance in the past, we can keep that around for compatibility for a few releases even if we add another API and use that internally.

I think one big thing that would warant a version 3 release would be the switch to Qt6, because this would break many plugins anyway. And with this switch we could of course do a lot of breaking changes. I actually wouldn't do that change before Qt6 really, and it will be still a bit until it is feasible to switch to Qt6. Until then we can already start refactoring parts in the 2.x line. E.g. QSettings could be replaced while staying compatible. webservice module should get some more serious overhault with different API, but even that could be added in parallel.

Sophist-UK:
@phw Philipp - you make a good point that I hadn't considered which is (paraphrasing / reading between the lines) that it is perfectly feasible to introduce new APIs whilst maintaining a backwards-compatibility shim for existing APIs if there is a future plan to have a new version where the old APIs are withdrawn - and of course gradual step-wise implementation of changes is always better than a grand-strategy for a big-bang new version.

Sophist-UK:
As a further aside for future thinking, we should be aware that the Python team are in the process of implementation sub-interpreters. Which is a bit like multi-processing except it uses threads each running its own interpreter with its own GIL - which enables concurrent multi-threading. I suspect that it will be a bit of a trade-off having some of the pros and cons of multi-threading and some of the pros and cons of multi-processing. For example, having a separate GIL means separate objects (like multi-processing) and separate garbage collection (like multi-processing) - so you need to pass stuff back and forward through queues or similar. BUT, this could give us the multi-CPU benefits and be simpler to implement than multi-processing.

phw:
Ok, let's merge this and see how it works out. If this should have any unintended side effects we should be able to catch that in the 2.7 development cycle, if necessary this is easy to revert. If all works well (and I expect that) the doors are open to make better use of this with an updated API.

> Which is a bit like multi-processing except it uses threads each running its own interpreter with its own GIL - which enables concurrent multi-threading.

I guess this will have the same constraints in passing data between those threads as multi-processing. So probably doesn't make much difference then for us in regards what is needed to make use of it. With concurrent.futures and using `ProcessPoolExecutor` there wouldn't be much difference in running things in separate process. But currently the stuff we run in threads is not directly able to run with ProcessPoolExecutor due to the objects being passed around and accessed in the threaded code.

PR #1804, opened 2021-05-04
Add common functionality to open help to PicardDialog

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

Move code common to open a help page, with keyboard shortcut handling, to PicardDialog. This will make it easier to add similar functionality to other dialogs.

### DIFF
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2014, 2018 Laurent Monin
 # Copyright (C) 2016-2018 Sambhav Kothari
 # Copyright (C) 2018 Vishal Choudhary
-# Copyright (C) 2019-2020 Philipp Wolfer
+# Copyright (C) 2019-2021 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -36,11 +36,15 @@ from picard.config import (
     Option,
     get_config,
 )
+from picard.const import DOCS_BASE_URL
 from picard.const.sys import (
     IS_MACOS,
     IS_WIN,
 )
-from picard.util import restore_method
+from picard.util import (
+    restore_method,
+    webbrowser2,
+)
 
 
 if IS_MACOS:
@@ -105,6 +109,7 @@ class SingletonDialog:
 
 class PicardDialog(QtWidgets.QDialog, PreserveGeometry):
 
+    help_url = None
     flags = QtCore.Qt.WindowSystemMenuHint | QtCore.Qt.WindowTitleHint | QtCore.Qt.WindowCloseButtonHint
 
     def __init__(self, parent=None):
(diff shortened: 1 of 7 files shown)


picard/ui/__init__.py
```diff
@@ -113,9 +118,21 @@ def __init__(self, parent=None):
...
+        elif event.matches(QtGui.QKeySequence.HelpContents) and self.help_url:
+            self.show_help()
         else:
             super().keyPressEvent(event)
 
+    def show_help(self):
+        if self.help_url:
+            url = self.help_url
+            if url.startswith('/'):
+                url = DOCS_BASE_URL + url
```
zas, line 130: I don't think there's any case where `DOCS_BASE_URL` could start with `goto://` so may be:

```python
if url.startswith('/'):
    webbrowser2.open(DOCS_BASE_URL + url)
elif url.startswith('goto://'):
    webbrowser2.goto(url[7:])
else:
    webbrowser2.open(url)
```

picard/ui/__init__.py
```diff
@@ -113,9 +118,21 @@ def __init__(self, parent=None):
...
+            if url.startswith('goto://'):
```
zas, line 131: In fact, we could prolly simplify things here, as none of the keys from `PICARD_URLS` contains a `/`, they will never conflict with URLs, what about just checking if `url in PICARD_URLS` in `webbrowser2.open()` and totally drop `goto` method?

picard/ui/options/dialog.py
```diff
@@ -193,18 +182,17 @@ def disable_page(self, name):
...
         # If URL is empty, use the first non empty parent help URL.
         while current_page.PARENT and not url:
             current_page = self.item_to_page[self.page_to_item[current_page.PARENT]]
             url = current_page.HELP_URL
         if not url:
-            url = DOCS_BASE_URL
-        elif url.startswith('/'):
-            url = DOCS_BASE_URL + url
-        webbrowser2.open(url)
+            url = '/config/configuration.html'
```
zas, line 194: Since the code above, prepend `DOCS_BASE_URL` when url starts with `/`, I think we could remove concatenation in https://github.com/metabrainz/picard/blob/0e30004337a9436b3b57f6b3bd6e8df396fe804a/picard/const/__init__.py#L83-L87

So we could use `PICARD_URLS['doc_options']` here instead of hardcoded `'/config/configuration.html'` or, if we make the change to `open()` I suggested above, just `doc_options`.

zas (changes requested):

picard/ui/__init__.py
```diff
@@ -113,9 +118,18 @@ def __init__(self, parent=None):
...
             self.close()
+        elif event.matches(QtGui.QKeySequence.HelpContents) and self.help_url:
+            self.show_help()
         else:
             super().keyPressEvent(event)
 
+    def show_help(self):
+        if self.help_url:
+            url = self.help_url
+            if url.startswith('/'):
```
zas, line 129: I wonder if we still need this.

phw: yes, for links from option pages. They define only partial URLs (to avoid having to use DOCS_BASE_URL everywhere)

zas (approved):

PR #1806, opened 2021-05-04
PICARD-2204: Support unicode escape sequences in tagger scripts

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2204, PICARD-2200

This adds support for Unicode escape sequences in the form of `\uFFFF` to tagger script.

Rationale:

- This would allow explicitly inserting special characters and makes it easier to define some characters the user cannot write
- Allows making invisible control characters visible (see PICARD-2200)

# Solution

Extend the tagger script parser for support of Unicode escape sequences. The Unicode escape sequences always start with `\u` followed by a four digit hexadecimal number representing the Unicode code point. E.g. to get the character 湖 one can also use `\u6e56`. To explicitly insert a thin space a user can write `\u2009`.

This also extends the script editor for proper syntax highlighting and implements automatic replacement of control characters other then  newlines and tabulators with the corresponding escape sequence. This avoids issues such as accidentally inserted invisible control characters with unwanted side effects as discussed in PICARD-2200.

E.g. if a user copy and pastes from the docs and copies the left-to-right mark we insert into the docs by accident they will get this:

grafik

# Action

This is the first actual syntax extension to tagger script since it's first implementation. Old script stay completely compatible. Scripts using Unicode escape sequences will not parse in older Picard versions, as `\u` is an invalid escape. Older versions will fail cleanly with a syntax error.

### DIFF
--- a/picard/script/parser.py
+++ b/picard/script/parser.py
@@ -249,6 +249,8 @@ Grammar:
 
     def read_multi(self, count):
         text = ch = self.read()
+        if not ch:
+            self.__raise_eof()
         count -= 1
         while ch and count:
             ch = self.read()
@@ -313,11 +315,11 @@ Grammar:
                 elif ch == 't':
                     text.append('\t')
                 elif ch == 'u':
-                    hex = self.read_multi(4)
+                    codepoint = self.read_multi(4)
                     try:
-                        text.append(chr(int(hex, 16)))
-                    except ValueError:
-                        self.__raise_unicode(hex)
+                        text.append(chr(int(codepoint, 16)))
+                    except (TypeError, ValueError):
+                        self.__raise_unicode(codepoint)
                 elif ch not in "$%(),\\":
                     self.__raise_char(ch)
                 else:
--- a/picard/script/parser.py
+++ b/picard/script/parser.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2006-2009, 2012 Lukáš Lalinský
 # Copyright (C) 2007 Javier Kohen
-# Copyright (C) 2008-2011, 2014-2015, 2018-2020 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014-2015, 2018-2021 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2009 Nikolai Prokoschenko
 # Copyright (C) 2011-2012 Michael Wiencek
@@ -67,6 +67,10 @@ class ScriptEndOfFile(ScriptParseError):
 
 
 class ScriptSyntaxError(ScriptParseError):
+    pass
+
+
+class ScriptUnicodeError(ScriptSyntaxError):
     pass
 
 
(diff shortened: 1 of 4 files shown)


picard/script/parser.py
```diff
@@ -293,6 +312,12 @@ def parse_text(self, top):
                     text.append('\n')
                 elif ch == 't':
                     text.append('\t')
+                elif ch == 'u':
+                    hex = self.read_multi(4)
```
zas, line 316: Don't use `hex`, that's a builtin.

picard/script/parser.py
```diff
@@ -239,6 +247,17 @@ def read(self):
                 self._x += 1
         return ch
 
+    def read_multi(self, count):
+        text = ch = self.read()
```
zas, line 251: I guess you need to test result here, because the script could be just `\\u`, it should also raise eof in this case (add a test for it).

picard/ui/widgets/scripttextedit.py
```diff
@@ -175,6 +179,19 @@ def get_selected(self):
...
 
 
+def _clean_text(text):
+    return "".join(_replace_control_chars(text))
+
+
+def _replace_control_chars(text):
+    import unicodedata
+    for ch in text:
+        if ch not in ('\n', '\r', '\t') and unicodedata.category(ch)[0] == "C":
```
zas, line 189: IMHO, `('\n', '\r', '\t')` should be stored in a variable outside the loop, and made a set:

```python
simple_ctrl_chars = {'\n', '\r', '\t'}
for ch in text:
  if ch not in simple_ctrl_chars ...
```

zas (changes requested):

Sophist-UK:
Is it possible to show the actual unicode characters in a tooltip so you can see what the actual character is?

picard/script/parser.py
Sophist-UK, line 316: Do you need to check that 4 characters were read? What about '\uabc'? Should this raise a unicode error or eof?

picard/script/parser.py
phw: Right, this was not handled properly. Fixed and extended the tests.

picard/ui/widgets/scripttextedit.py
phw: Yes, that reads better. Changed it

phw:
> Is it possible to show the actual unicode characters in a tooltip so you can see what the actual character is?

I like this idea. `QTextCharFormat` actually allows setting a tooltip (https://doc.qt.io/qt-5/qtextcharformat.html#setToolTip) and the description looks like this could do what we want, but I was so far unable to make this work. I'll investigate some more.

zas (approved):
It looks good to me. Nice improvement.

phw:
Ok, I merged it. But I will look into script editor tooltips. To my understanding this should work somehow, and we could make use of this in different ways.

phw:
@Sophist-UK the tooltip thing is more complicated then I had hoped. I added https://tickets.metabrainz.org/browse/PICARD-2207 , as I think this would be really need.

Sophist-UK:
Philipp - Good of you to attempt this.

PR #1808, opened 2021-05-05
PICARD-2205: Update syntax highlighting to allow functions starting with _ or 0-9

phw:
The scripting syntax definition and parser implementation allow such functions.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2205

According to the tagger script grammar [1]  and the actual implementation a function name can consist of the characters [a-zA-Z0-9_], without restriction on placement.

Thus the following examples are all valid function names: `$_`, `$123`, `$_a_`, `$_1a`

But the syntax highlighting assumes function names must start with [a-zA-Z].

https://github.com/metabrainz/picard/blob/master/picard/script/parser.py#L202-L212

# Solution

Update the syntax highlighting to properly highlight those functions.

### DIFF
--- a/picard/script/parser.py
+++ b/picard/script/parser.py
@@ -210,8 +210,8 @@ Grammar:
   text        ::= [^$%] | '\$' | '\%' | '\(' | '\)' | '\,' | unicodechar
   argtext     ::= [^$%(),] | '\$' | '\%' | '\(' | '\)' | '\,' | unicodechar
   identifier  ::= [a-zA-Z0-9_]
-  variable    ::= '%' identifier '%'
-  function    ::= '$' identifier '(' (argument (',' argument)*)? ')'
+  variable    ::= '%' (identifier | ':')+ '%'
+  function    ::= '$' (identifier)+ '(' (argument (',' argument)*)? ')'
   expression  ::= (variable | function | text)*
   argument    ::= (variable | function | argtext)*
 """
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -81,7 +81,7 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
     def __init__(self, document):
         super().__init__(document)
         syntax_theme = theme.syntax_theme
-        self.func_re = QtCore.QRegExp(r"\$(?!noop)[a-zA-Z][_a-zA-Z0-9]*\(")
+        self.func_re = QtCore.QRegExp(r"\$(?!noop)[_a-zA-Z0-9]*\(")
         self.func_fmt = QtGui.QTextCharFormat()
         self.func_fmt.setFontWeight(QtGui.QFont.Bold)
         self.func_fmt.setForeground(syntax_theme.func)


zas (approved):

Sophist-UK:
Personally I think that we should handle this with a documentation change rather than enabling identifiers which go against typical naming conventions in other languages.

zas:
> Personally I think that we should handle this with a documentation change rather than enabling identifiers which go against typical naming conventions in other languages.

As described, this patch fixes syntax highlighting according to the defined script grammar (which permits `$_abc`, since a long time).

I fail to see how this relates to "typical naming conventions in other languages", which, to be honest, means nothing.

Sophist-UK:
Sorry - I misread the description and the code. I thought that we currently only allowed function names starting with an alphabetic character whereas this is a change for code highlighting.

If there was not a possibility of creating backwards incompatibility, I would be advocating that we should only support identifiers starting with an alphabetic characters, but we would not want to introduce such an incompatible change.

So please ignore my previous comment.

phw:
> If there was not a possibility of creating backwards incompatibility, I would be advocating that we should only support identifiers starting with an alphabetic characters, but we would not want to introduce such an incompatible change.

Yes, actually I thought about this, but that would introduce an actual syntax change. Current situation is that actually all functions I have ever seen started with [a-z]. Underscore I could actually see as useful, e.g. for custom functions or real short stuff like `$_()`. Mainly for personal use actually, I wouldn't make such functions as part of official source.

Actually I guess in a potential Picard 3 we could introduce some syntactic changes to the language. On my personal wish list are proper comments (something like `/* ... */` where everything inside does not need to follow normal syntax like and nesting rules) and builtin support for regular expressions, to avoid this crazy double escaping which gives us `\\s+\\\(?f\(ea\)?t\\.[^\)]*\\\)?`.

Sophist-UK:
My votes for script editing improvements:

1. Automated indenting
2. Avoiding double escaping - I guess it might be possible to automatically remove double escaping in existing scripts when upgrading between minor versions.
3. Access to existing file metadata $existing(variable).
4. Comments - however I guess it is possible people have strings now which look like comments or will want them in the future, so I am really unsure how to do this in a way that is not going to block other functionality. `$noop` sort of does the job even if a but cumbersome.
5. A new function `$ai(question)` that allows you to ask any question about the track as a parameter and gets the answer back. e.g. `$ai(first recording date?)` or `$ai(composer's date of birth?)` or `$ai(drummer's wife's middle name?)`.

As for identifier syntax changes, I have never seen `$1()` or `$_()` either, and whilst I actually doubt it has happened that doesn't mean that someone somewhere hasn't written a private plugin which implements something like that. That is the problem with backwards compatibility for things that you cannot automatically upgrade away - it's just becomes a real millstone.

P.S. I was just kidding about `$ai()`.

phw:
> As for identifier syntax changes, I have never seen `$1()` or `$_()` either, and whilst I actually doubt it has happened that doesn't mean that someone somewhere hasn't written a private plugin which implements something like that.

From a theoretical point of view the tagger script syntax has no arbitrary restriction like "variable and/or function names cannot start with numbers" because it doesn't need them.

If you have a language where there is a function call `foo()` and an integer `42` it does make sense to not allow a function named `42foo()` to avoid disambiguity and the need for the parser to constantlye unsure about what it is parsing until a later point.

Tagger script does not need this because it has definite symbols for starting functions and variables. `$foo()` is a function call, `%foo%` is a variable. At the same time it also needs those symbols because `foo` itself also is valid tagger script (as is `The Rolling Stones`). Allowing `foo()` for a function would make the same trouble as starting a function name with numbers does in other languages.

You can also see this pattern elsewhere. `42foo` is an invalid variable name in Python, C, JavaScript, etc. But `$42foo` is valid in Perl, PHP and shell script, as are other variable types in Perl such as `@42`.

phw:
> P.S. I was just kidding about `$ai()`.

Now someone will need to write this plugins :D

Sophist-UK:
The issue is that scripts run synchronously in Picard so making a web call when the script runs is going to absolutely kill performance and UI responsiveness.

The obvious solution is to run this as a track metadata plugin and pre-download the answers so that when the script runs `$ai()` does not need to hold everything up until it gets the data.

An obvious optimisation to reduce the amount of data needed to download would be to examine the scripts to see what questions are being asked. Unfortunately, `$set(a1,$ai(question1))$set(a2,$ai(%a1%))` is the counter example - so we will unfortunately need to pre-download the answers to all possible questions. Fortunately this is an AI-based plugin, so we should be able to achieve this given sufficient computing resources.

As for who should write this script, I believe that only @zas, @php or @Sophist-UK have the skills for this - and I am unfortunately too busy to have the time right now - so I guess it is down to one of you other two masters-of-the-Picard-universe to write it. :-)

Sophist-UK:
P.S. We will also need to protect against clever-bar-steward-users who will believe themselves to be cleverer that `$ai()` and ask self-referencing questions or overly philosophical ones or ones which have no finite response (like "what is the final digit in the number Pi) or which have no logical answer.

This problem clearly pre-dates Picard - the Captain not the software - because I clearly recall Spock (who is a clever-bar-steward-user if there ever was one) asking computers to do impossible things in order to screw make them hang.

PR #1810, opened 2021-05-05
PICARD-524: Multiple naming scripts

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: This adds the ability to maintain a list of file naming scripts and select which script to use.

# Problem

Picard currently only allows a single file naming script, and there is no way to easily select from a list of scripts.

* JIRA ticket (_optional_): PICARD-524

# Solution

This PR adds the ability to maintain a list of file naming scripts and select which script Picard uses.  Each script has associated metadata such as title, author, version, license and description, which is saved with the script.  Scripts can be imported from / exported to external files, in either a plain text format for the script only (\*.pts or \*.txt) or as a script package in JSON format including all metadata (\*.pnsp or \*.json).  The extension .pnsp was created to signify "Picard Naming Script Package".  The .pts extension was previously created by @phw when creating his syntax formatting plugins.

The user's list of scripts stores the scripts in the settings file in JSON format rather than storing the FileNamingScript objects directly.  This was done to allow for changes to the FileNamingScript object without affecting the user's stored list of scripts.  When scripts are added to or removed from the list, the changes are recorded immediately without having to use the "Make it so!" button.  This allows the list to be maintained outside of the Options settings if there is a future change to open the script editor directly from the main screen.  The selection of which script is to be used still requires "Make it happen!" before the change is saved.

The preset scripts cannot be edited directly, however there is an option to create a copy of any currently displayed script, and the copy can be edited.  This also allows safely making trial edits to a script by first creating a copy of the script and using the copy as the working version.

A new PicardScript base class was developed to allow for a possible future TaggingScript subclass in addition to the new FileNamingScript subclass.  A new test file has been added with tests associated with the new PicardScript and FileNamingScript classes.

# Action

The initial version of this code assigns an ID to each FileNamingScript object created, but that ID is not currently used.  Should this be removed until such time as there is a use identified?


picard/script/__init__.py
```diff
@@ -101,3 +111,229 @@ def enabled_tagger_scripts_texts():
...
+        return new_object
+
+
+def get_file_naming_script_presets():
+    """Generator of preset example file naming script objects.
+
+    Yields:
+        FileNamingScript: the next example FileNamingScript object
+    """
+    AUTHOR = N_("Picard Development Team")
```
zas, line 294: Note sure authors should be marked for translation.
Also I'd use `MusicBrainz Picard Development Team`.

picard/script/__init__.py
```diff
@@ -101,3 +111,229 @@ def enabled_tagger_scripts_texts():
...
+    LICENSE = "GNU Public License version 2"
+
+    yield FileNamingScript(
+        title=N_("Preset 1: Default file naming script"),
+        script=DEFAULT_FILE_NAMING_FORMAT,
+        readonly=True,
+        deletable=False,
+        author=AUTHOR,
+        description=DESCRIPTION,
+        version="n/a",
```
zas, line 305: Why "n/a" here and "1.0" for other presets?

picard/ui/scripteditor.py
```diff
@@ -354,23 +355,134 @@ def eventFilter(self, object, event):
...
+        item in the script selection combo box.
+
+        Args:
+            save_enabled (bool, optional): Allow updates to be saved to this item. Defaults to True.
+        """
+        selected = self.ui.preset_naming_scripts.currentIndex()
+        if selected < 0:
+            return
+        script_item = self.get_selected_item()
+        self.ui.script_title.setReadOnly(bool(script_item.readonly or selected < 1))
```
zas, line 477: `bool()` shouldn't be needed here.

picard/ui/scripteditor.py
```diff
@@ -354,23 +355,134 @@ def eventFilter(self, object, event):
...
+        self.ui.file_naming_format.setReadOnly(script_item.readonly)
+        self.ui.file_naming_editor_save.setEnabled(bool(save_enabled and not script_item.readonly))
```
zas, line 479: bool() shouldn't be needed here.

picard/ui/scripteditor.py
```diff
@@ -354,23 +355,134 @@ def eventFilter(self, object, event):
...
+        self.ui.file_naming_editor_copy.setEnabled(save_enabled)
+        self.ui.file_naming_editor_select.setEnabled(save_enabled)
+        self.ui.file_naming_editor_delete.setEnabled(bool(script_item.deletable and save_enabled))
```
zas, line 482: bool() shouldn't be needed here.

picard/ui/scripteditor.py
```diff
@@ -517,6 +714,19 @@ def check_format(self):
...
+    def confirmation_dialog(self, message_text):
+        """Display a confirmation dialog with Ok and Cancel buttons.
+
+        Args:
+            message_text (str): Message to display
+
+        Returns:
+            bool: True if Ok, otherwise False
+        """
+        dialog = QtWidgets.QMessageBox(QtWidgets.QMessageBox.Warning, N_('Confirm'), message_text, QtWidgets.QMessageBox.Ok | QtWidgets.QMessageBox.Cancel, self)
```
zas, line 726: Perhaps split this long line.

picard/ui/scripteditor.py
```diff
@@ -540,5 +750,66 @@ def test(self):
...
+
+        self.ui.script_title.setReadOnly(True)
+        self.ui.script_last_updated.setReadOnly(True)
+        self.ui.script_author.setReadOnly(self.readonly)
+        self.ui.script_version.setReadOnly(self.readonly)
+        self.ui.script_license.setReadOnly(self.readonly)
+        self.ui.script_description.setReadOnly(self.readonly)
+        self.ui.buttonBox.button(QtWidgets.QDialogButtonBox.Close).setVisible(bool(self.readonly))
+        self.ui.buttonBox.button(QtWidgets.QDialogButtonBox.Cancel).setVisible(bool(not self.readonly))
+        self.ui.buttonBox.button(QtWidgets.QDialogButtonBox.Save).setVisible(bool(not self.readonly))
```
zas, line 796: bool() shouldn't be needed here.

zas (changes requested):
I didn't test yet, just reviewed the code, but I will soon.
A step to more manageable file naming scripts :+1:

picard/ui/scripteditor.py
```diff
@@ -449,43 +598,91 @@ def toggle_wordwrap(self):
...
             options |= QtWidgets.QFileDialog.DontUseNativeDialog
             filename, file_type = QtWidgets.QFileDialog.getSaveFileName(self, dialog_title, default_path, dialog_file_types, options=options)
             if filename:
+                # Fix issue where Qt may set the extension twice
+                (name, ext) = os.path.splitext(filename)
+                if ext and str(name).endswith('.' + ext):
+                    filename = name
                 log.debug('Exporting naming script file: %s' % filename)
+                if file_type == self.FILE_TYPE_PACKAGE:
+                    script_text = script_item.to_json()
```
zas, line 670: IMHO, we should save it as indented json so it's easier to read when shared.

zas (changes requested):

picard/ui/scripteditor.py
```diff
@@ -449,43 +598,91 @@ def toggle_wordwrap(self):
...
             try:
                 with open(filename, 'r', encoding='utf8') as i_file:
-                    self.set_script(i_file.read())
+                    file_content = i_file.read()
             except OSError as error:
-                error_message = 'Error importing file "{0}": {1}'.format(filename, error.strerror)
-                log.error(error_message)
-                self.display_error(OptionsCheckError(N_("File Error"), error_message))
+                log.error(error_import % (filename, error.strerror))
+                error_message = error_import_tx % (filename, error.strerror)
```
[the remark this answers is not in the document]
rdswift: I thought the standard was to log in English, but display translated pop-ups to the user?  I can change so that these are all translated (there should be 5 or 6 cases like this), if that's preferred.

picard/ui/scripteditor.py
```diff
@@ -449,43 +598,91 @@ def toggle_wordwrap(self):
...
+                error_message = error_import_tx % (filename, _('The file was empty'))
+                self.display_error(ScriptImportError(self.file_error_title, error_message))
+                return
+            if file_type == self.FILE_TYPE_PACKAGE:
+                try:
+                    script_item = FileNamingScript().create_from_json(file_content)
+                except JSONDecodeError as error:
+                    log.error(error_decode % (filename, error.msg))
+                    error_message = error_decode_tx % (filename, error.msg)
+                    self.display_error(ScriptImportError(self.file_error_title, error_message))
```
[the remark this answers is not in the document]
rdswift: Agreed.  I had that at one point, but it became unwieldy when trying to use both translated for the display and untranslated for the logs.  I'll add it back.

picard/ui/scripteditor.py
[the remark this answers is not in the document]
zas: > I thought the standard was to log in English, but display translated pop-ups to the user?

It is, you are missing the point here.
 `N_()` doesn't translate anything, so as is, both are untranslated.
 But the string is the same in both cases, read again the code I suggested (`_(error_import) `)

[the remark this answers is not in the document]
rdswift: Fixed in https://github.com/metabrainz/picard/pull/1810/commits/e2b3020e982c77b05b63050475bb21d8e24547a8

picard/ui/scripteditor.py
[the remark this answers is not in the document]
rdswift: Fixed in https://github.com/metabrainz/picard/pull/1810/commits/e2b3020e982c77b05b63050475bb21d8e24547a8

picard/ui/scripteditor.py
```diff
@@ -449,43 +598,91 @@ def toggle_wordwrap(self):
...
-                        o_file.write(self.get_script() + '\n')
+                        o_file.write(script_text + '\n')
+                    dialog = QtWidgets.QMessageBox(
+                        QtWidgets.QMessageBox.Information,
+                        _("Export Script"),
+                        _("Script successfully exported to ") + filename,
+                        QtWidgets.QMessageBox.Ok,
+                        self
+                    )
+                    dialog.exec_()
```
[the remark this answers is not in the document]
rdswift: Fixed in https://github.com/metabrainz/picard/pull/1810/commits/2e3c82210596d5dad8c18fddee684782550c2b46

picard/script/__init__.py
```diff
@@ -101,3 +114,237 @@ def enabled_tagger_scripts_texts():
...
+                        updated = True
+        # Don't overwrite `last_updated` with a generated value if a last_updated value has been provided.
+        if updated and 'last_updated' not in settings:
+            self.update_last_updated()
+
+    def copy(self):
+        """Create a copy of the current script object with updated title and last updated attributes.
+        """
+        new_object = deepcopy(self)
+        new_object.update_script_setting(title=self.title + N_(" (Copy)"))
```
zas, line 217: From i18n point of view such concatenation will not work well.
It should be something like:

```python
new_object.update_script_setting(title=_("%s (Copy)" % self.title))
```

Reason: in other languages the suffix (here ` (Copy)`) could be placed before the title for example.

picard/script/__init__.py
```diff
@@ -101,3 +114,237 @@ def enabled_tagger_scripts_texts():
...
+        Args:
+            indented (bool): Indent the output.
+
+        Returns:
+            str: The properties of the script object formatted as a JSON string.
+        """
+        items = {key: getattr(self, key) for key in dir(self) if key in self.JSON_OUTPUT}
+        if indented:
+            return json.dumps(items, indent=4, sort_keys=True)
+        return json.dumps(items, sort_keys=True)
```
zas, line 234: Instead of passing `indented` we could just pass `indent` as an integer or `None`, and simplify code to:

```python
def to_json(self, indent=None):
...
    items = {key: getattr(self, key) for key in dir(self) if key in self.JSON_OUTPUT}
    return json.dumps(items, sort_keys=True, indent=indent)
```

And of course change callers to use `indent=4` instead of `indented=True`.

picard/ui/scripteditor.py
```diff
@@ -448,44 +602,92 @@ def toggle_wordwrap(self):
...
 
+    def output_error(self, title, fmt, filename, msg):
+        """Log error and display error message dialog.
+
+        Args:
+            title (str): Title to display on the error dialog box
+            fmt (str): Format for the error type being displayed
+            filename (str): Name of the file being imported or exported
+            msg (str): Error message to display
+        """
```
zas, line 613: Since `title` is always `self.FILE_ERROR_TITLE`, perhaps just write a small wrapper to shorten lines a bit:

```python
def output_file_error(self, fmt, filename, msg):
      self.output_error(self.FILE_ERROR_TITLE, fmt, filename, msg)
```

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
     if not config.setting["enable_tagger_scripts"]:
         return []
     return [(s_name, s_text) for _s_pos, s_name, s_enabled, s_text in config.setting["list_of_scripts"] if s_enabled and s_text]
+
+
+@unique
+class PicardScriptType(Enum):
```
zas, line 120: Why not `IntEnum`?

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
...
+class PicardScript():
+    """Base class for Picard script objects.
+    """
+    # Base class developed to support future tagging script class as possible replacement for currently used tuples in config.setting["list_of_scripts"].
+
+    TYPE = PicardScriptType.BASE
+    JSON_OUTPUT = ['title', 'script']
+
+    # Don't automatically trigger changing the `script_last_updated` property when updating these properties.
+    _last_updated_ignore_list = ['last_updated', 'readonly', 'deletable', 'id']
```
zas, line 137: You can use a set here (but if it is meant to be modified at runtime, which isn't it seems right now):

```suggestion
    _last_updated_ignore_list = {'last_updated', 'readonly', 'deletable', 'id'}
```

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
...
+
+    def update_from_json(self, json_string):
+        """Updates the values of the properties based on the contents of the JSON string provided.
+        Properties in the JSON string that are not found in the script object are ignored.
+
+        Args:
+            json_string (str): JSON string containing the property settings.
+        """
+        settings = json.loads(json_string)
+        self._update_from_dict(settings)
```
zas, line 259: Since that's pretty simple & short, I think we can drop the intermediate variable here:

```python
self._update_from_dict(json.loads(json_string))
```

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
...
+    yield FileNamingScript(
+        title=_("Preset 1: Default file naming script"),
+        script=DEFAULT_FILE_NAMING_FORMAT,
+        readonly=True,
+        deletable=False,
+        author=AUTHOR,
+        description=DESCRIPTION,
+        version="1.0",
+        license=LICENSE,
+        last_updated="2019-08-05",
```
zas, line 317: Just a minor thing, but for consistency I would stick with the format defined at https://github.com/metabrainz/picard/pull/1810/files#diff-22922bde1619b8ae742614141b5fd647e7c9f973fee086f776e8d6c74212e37bR184

zas (changes requested):

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
...
+        self._update_from_dict(settings)
+
+
+class FileNamingScript(PicardScript):
+    """Picard file naming script class
+    """
+    TYPE = PicardScriptType.FILENAMING
+    JSON_OUTPUT = ['title', 'script', 'author', 'description', 'license', 'version', 'last_updated']
+
+    def __init__(self, script='', title='', id=None, readonly=False, deletable=True, author='', description='', license='', version='', last_updated=None):
```
zas, line 268: Wrap parameters here (line too long)

picard/script/__init__.py
```diff
@@ -101,3 +114,236 @@ def enabled_tagger_scripts_texts():
...
+
+    Yields:
+        FileNamingScript: the next example FileNamingScript object
+    """
+    AUTHOR = "MusicBrainz Picard Development Team"
+    DESCRIPTION = _("This preset example file naming script does not require any special settings, tagging scripts or plugins.")
+    LICENSE = "GNU Public License version 2"
+
+    yield FileNamingScript(
+        title=_("Preset 1: Default file naming script"),
```
zas, line 309: If we want to re-order presets having the number in the title will lead to re-translation, so I think we should do something like:

```python
def preset_title(number, title):
    return _("Preset %d: %s) % (number, _(title))

...
yield FileNamingScript(
        title=preset_title(1, N_("Default file naming script")),
```

zas (changes requested):

picard/script/__init__.py
rdswift: Because I screwed up.  :blush:

picard/script/__init__.py
rdswift: I agree, it looks better being consistent.  Is the format okay?  I was wondering if it would make more sense showing local time (with the timezone).  Also not sure if we actually need the seconds precision.

picard/ui/scripteditor.py
```diff
@@ -354,23 +362,141 @@ def eventFilter(self, object, event):
...
+            deletable=False,
+            id="current"
+        )
+        self.set_script(script_item.script)
+        self.ui.preset_naming_scripts.addItem(self.SCRIPT_TITLE_SYSTEM % script_item.title, script_item)
+
+        for script_json in self.naming_scripts:
+            script_item = FileNamingScript().create_from_json(script_json)
+            if script_item.get_value('title'):
+                self.ui.preset_naming_scripts.addItem(self.SCRIPT_TITLE_USER % script_item.title, script_item)
```
zas, line 383: I noticed something:

If I edit the title through Details, it is prefixed with "User: ", but if I edit it through main dialog "Title:" it isn't.

picard/ui/scripteditor.py
```diff
@@ -383,14 +509,44 @@ def synchronize_selected_example_lines(current_row, source, target):
...
+            self.ui.preset_naming_scripts.setCurrentIndex(0)
+            self.ui.preset_naming_scripts.blockSignals(False)
+            self.update_scripts_list()
+            self.select_script()
 
     def save_script(self):
+        """Saves changes to the current script to the script list and combo box item.
+        """
+        selected = self.ui.preset_naming_scripts.currentIndex()
+        if selected == 0:
```
zas, line 535: Why is this index handled differently?

picard/ui/scripteditor.py
```diff
@@ -448,44 +611,102 @@ def toggle_wordwrap(self):
...
+                    script_item = FileNamingScript().create_from_json(file_content)
+                except JSONDecodeError as error:
+                    self.output_file_error(self.FILE_ERROR_DECODE, filename, error.msg)
+                    return
+                if not (script_item.get_value('title') and script_item.get_value('script')):
+                    self.output_file_error(self.FILE_ERROR_DECODE, filename, N_('Invalid script package'))
+                    return
+            else:
+                script_item = FileNamingScript(
+                    title=_("Imported from ") + filename,
```
zas, line 669: Avoid concatenation for i18n.

```suggestion
                    title=_("Imported from %s") % filename,
```

picard/ui/scripteditor.py
```diff
@@ -517,6 +738,23 @@ def check_format(self):
...
+        """Display a confirmation dialog with Ok and Cancel buttons.
+
+        Args:
+            message_text (str): Message to display
+
+        Returns:
+            bool: True if Ok, otherwise False
+        """
+        dialog = QtWidgets.QMessageBox(QtWidgets.QMessageBox.Warning,
+                                       _('Confirm'),
```
zas, line 751: We should move first element to next line (with proper indentation) (as done elsewhere):

```python
dialog = QtWidgets.QMessageBox(
    QtWidgets.QMessageBox.Warning,
```

picard/ui/scripteditor.py
```diff
@@ -383,14 +509,44 @@ def synchronize_selected_example_lines(current_row, source, target):
...
+        self.synchronize_selected_example_lines(self.examples_current_row, self.ui.example_filename_after, self.ui.example_filename_before)
+
+    def delete_script(self):
+        """Removes the currently selected script from the script selection combo box and script list.
+        """
+        if self.confirmation_dialog(_('Are you sure that you want to delete the script?')):
+            idx = self.ui.preset_naming_scripts.currentIndex()
+            self.ui.preset_naming_scripts.blockSignals(True)
+            self.ui.preset_naming_scripts.removeItem(idx)
+            self.ui.preset_naming_scripts.setCurrentIndex(0)
```
zas, line 526: When deleting it is a bit weird it goes to first element after deletion, perhaps it should select the next element instead (or previous one if none). Dunno exactly what's best in this case, but jumping to first element is disturbing.

picard/ui/scripteditor.py
```diff
@@ -448,44 +611,102 @@ def toggle_wordwrap(self):
...
-                    self.set_script(i_file.read())
+                    file_content = i_file.read()
             except OSError as error:
-                error_message = 'Error importing file "{0}": {1}'.format(filename, error.strerror)
-                log.error(error_message)
-                self.display_error(OptionsCheckError(N_("File Error"), error_message))
+                self.output_file_error(self.FILE_ERROR_IMPORT, filename, error.strerror)
+                return
+            if not file_content.strip():
+                self.output_file_error(self.FILE_ERROR_IMPORT, filename, N_('The file was empty'))
```
zas, line 656:
```suggestion
                self.output_file_error(self.FILE_ERROR_IMPORT, filename, _('The file was empty'))
```

Else it will not be translated.

picard/ui/scripteditor.py
```diff
@@ -448,44 +611,102 @@ def toggle_wordwrap(self):
...
+                return
+            if file_type == self.FILE_TYPE_PACKAGE:
+                try:
+                    script_item = FileNamingScript().create_from_json(file_content)
+                except JSONDecodeError as error:
+                    self.output_file_error(self.FILE_ERROR_DECODE, filename, error.msg)
+                    return
+                if not (script_item.get_value('title') and script_item.get_value('script')):
+                    self.output_file_error(self.FILE_ERROR_DECODE, filename, N_('Invalid script package'))
```
zas, line 665:
```suggestion
                    self.output_file_error(self.FILE_ERROR_DECODE, filename, _('Invalid script package'))
```

We want this message to be translated.

picard/ui/scripteditor.py
```diff
@@ -448,44 +611,102 @@ def toggle_wordwrap(self):
...
                 except OSError as error:
-                    error_message = 'Error exporting file "{0}": {1}'.format(filename, error.strerror)
-                    log.error(error_message)
-                    self.display_error(OptionsCheckError(N_("File Error"), error_message))
+                    self.output_file_error(self.FILE_ERROR_EXPORT, filename, error.strerror)
+                else:
+                    dialog = QtWidgets.QMessageBox(
+                        QtWidgets.QMessageBox.Information,
+                        _("Export Script"),
+                        _("Script successfully exported to ") + filename,
```
zas, line 705:
```suggestion
                        _("Script successfully exported to %s") % filename,
```

picard/script/__init__.py
```diff
@@ -101,3 +114,257 @@ def enabled_tagger_scripts_texts():
...
+        self.author = author
+        self.description = description
+        self.license = license
+        self.version = version
+
+    def copy(self):
+        new_object = super().copy()
+        new_object.readonly = False
+        new_object.deletable = True
+        return new_object
```
zas, line 313: I wonder if this is needed as it is apparently only used in `copy_script()`.
Perhaps just use native `FileNamingScript.copy()` and set `readonly` and `deletable` in `copy_script()`.
Overriding this method can play against us if we happen to want to preserve attributes of the copied script later.
Another approach: create a new method `editable_copy()' instead and use that.

picard/ui/scripteditor.py
```diff
@@ -259,16 +246,22 @@ class ScriptEditorPage(PicardDialog):
...
+        ListOption(
+            "setting",
+            "file_naming_scripts",
+            [],
+        ),
     ]
 
+    FILE_ERROR_IMPORT = N_('Error importing "%s". %s.')
+    FILE_ERROR_DECODE = N_('Error decoding "%s". %s.')
+    FILE_ERROR_EXPORT = N_('Error exporting file "%s". %s.')
```
zas, line 258: I think those should be moved near where they are used, to provide better context to translators.
I don't think they should be class-level attributes.

zas (changes requested):

picard/script/__init__.py
zas: Dunno, I'm a bit bothered by this field, because it's complicated to maintain, perhaps it should be just handled automatically by Picard, if it has the specified format, and left untouched if not.
Frankly dunno at this point.
Seconds precision seems too much, but on the other side, that's pretty common datetime format.

picard/script/__init__.py
```diff
@@ -101,3 +114,257 @@ def enabled_tagger_scripts_texts():
...
+            self.last_updated = last_updated
+
+    def _set_new_id(self):
+        """Sets the ID of the script to a new system generated uuid.
+        """
+        self.id = str(uuid.uuid4())
+
+    def get_value(self, setting):
+        """A safe way of getting the value of the specified property setting, because
+        it handles missing properties by returning None rather than raising an exception.
```
phw, line 166: What do you think about maybe making this dict like by implementing `__getitem__`? Or if we want to keep it as attributes implement `__getattr__`?

picard/ui/scripteditor.py
phw: I think it's because this specific script is handled separately. The logic as for now is that we still have the config option `file_naming_format`, which is used elsewhere in the application. This is always index 0 here. Then we have separate scripts.

In the scope of this PR I think that behavior is still ok, but actually we should change it. We should completely get rid of `file_naming_format` and instead just store which of the scripts currently is the active one. We introduce a helper function to get the current script, and that is used wherever we need the actual active naming script.

Because we can keep other parts of Picard unchanged I think it's ok to keep the behavior with `file_naming_format` for this PR. We have enough to discuss here already. But afterwards we should actually change this. I guess we then could even get rid of the special index 0 "System" entry here.

phw (changes requested):
Now with this change this is becoming really great. And works rather well already. I like the ideas of supporting both plain text files and the naming script packages. Not sure if JSON is the best format for that, but I'm ok with that for this iteration. Maybe something like YAML would work better, as it would make editing the script in the exported file in a text editor easier.

What I do not fully grasp is the "select" button. As I understand it should mark the active script, but clicking it doesn't change anything. Also if I read the script selection combobox correctly, the script currently used for file naming would always be what is marked as system, right?

Do we need the select button at all? Maybe the active script should always be what is currently selected. Or otherwise if I click select it should be reflected somehow, e.g. by marking the current script as the active one in some way.

This dialog probably should also get a standard button. Currently as a user I am a bit confused how to deal with this dialog. Is just closing it enough? Do I need to click save (for each script) first?

Maybe there should be at least a "Close button", or maybe separate "Save and close" and "Cancel" buttons.

phw:
> Perhaps just something the Picard version used to generate the script.

Or we actually add a SCRIPT_VERSION. And set it to 1.1 (1.0 would have been current Picard, 1.1 is with the Unicode escape extension).

zas:
> Maybe something like YAML would work better

Yes, I agree YAML would be easier to edit, and it should be fairly easy to support, though it would require an extra dependency (pyyaml), as it isn't supported natively AFAIK.

phw:
> though it would require an extra dependency (pyyaml)

Yes, it concerns me a bit because we are picking up more dependencies all the time, something we avoided for long. But I guess it would be really beneficial here.

picard/ui/scripteditor.py
rdswift: It wasn't using the `self.update_combo_box_item()` helper when saving from main editing screen.  Good catch!

picard/ui/scripteditor.py
rdswift: That was to ensure that changes saved to the "Current file naming script" would be saved to the working version of the config settings for processing when `Make it so!` was triggered.  Changes made to the other user scripts would just be saved to the script object and updated in the user's script list.  To make those changes to the naming script in the config settings, the user needs to use the `Select` button.

I'm not completely happy with having both the `Save` and `Select` buttons because it isn't intuitive how they differ, but I'm struggling to find a better solution.

picard/script/__init__.py
rdswift: All moved into the PicardScript base class and removed from the FileNamingScript subclass.

picard/script/__init__.py
rdswift: Much nicer!  Done.

rdswift:
As for script selection, I've been toying with the idea of simply having a checkbox on the list of scripts to identify and select the script used for the file naming.  Still working through the process, but I think it might be more intuitive.  It also makes it easier to have the ability to select the active script from the main screen (either from the menu or from a small dialog box).

picard/script/__init__.py
rdswift: Actually, that's how I had it originally (readonly and automatically updated by Picard).  I can change it back to that easily enough, and drop the seconds resolution at the same time (although that came in handy during testing :wink: ).

rdswift:
Another thought...  Rather than using signals, what do you think about using callback functions for the save and update stuff?  Somehow that seems cleaner to me if this screen is accessed from different parents.

rdswift:
I'm not sure why the test is failing, but it appears that I've screwed something up by adding pyyaml to the requirements.

rdswift:
Made some revisions to help make the UI more intuitive:

- Save selected script id to a new `selected_file_naming_script_id` setting, and automatically display that script when the editor is opened.
- Revised the script selector combo box label to indicate that the selected (displayed) script is the one that Picard will use.
- Removed the `Select` button and now automatically set the currently selected (displayed) script as the one used by Picard.
- Automatically update the `file_naming_format` setting when script changes are saved in the editor.
- Added a warning dialog to confirm actions which may cause unsaved edits to be lost.
- Added a `Close` button.

One consequence to this is that changes to the file naming scripts and which script is selected do not require the "Make it so!" action, because they are updated directly from the script editor screen.

rdswift:
I think there are way too may buttons below the editor, and it looks overly crowded, so I'm working on only showing the main ones and have the rest of the functionality available from a menu bar.  The initial result looks like the image below.  If we go with the menu bar, I'd also suggest getting rid of the two checkboxes (word wrap and documentation toggles) to free up a bit more space and look a bit cleaner.  Is this worth pursuing?

image

zas:
> I think there are way too may buttons below the editor, and it looks overly crowded, so I'm working on only showing the main ones and have the rest of the functionality available from a menu bar. The initial result looks like the image below. If we go with the menu bar, I'd also suggest getting rid of the two checkboxes (word wrap and documentation toggles) to free up a bit more space and look a bit cleaner. Is this worth pursuing?

IMHO, yes. The main reason: with that many buttons, we'll be quickly out of space (especially in some languages), plus it's hard to have a logical layout, so that's quickly confusing.
On the cons side, I'd say it's perhaps unusual to have such menu bar in dialogs, in Picard.
But I don't see better way to organize things here.

Sophist-UK:
The screen shot looks good to me. I have no issues with rarely used functions being in the Script menu.

My only question is whether the `Reset` button undoes any changes since the last save, and if so whether it should be called `Revert` instead.

phw:
I agree with both @zas  and @Sophist-UK . It is much cleaner with a menu. Also if we switch this to be usable side by side with the main Window it might actually not be so odd anymore. Most other dialogs in Picard are rather minimal, but this has the complexity of a real tool.

> One consequence to this is that changes to the file naming scripts and which script is selected do not require the "Make it so!" action, because they are updated directly from the script editor screen.

If we wanted this to stay a sub dialog of options I'd say we need to change that. E.g. we also had this behavior in the cover art types selection, and it was very confusing.

But again, as we want to make this dialog usable separate from options this behavior would be kind of expected.

zas (changes requested):
The menu is definitively an improvement :)

picard/ui/scripteditor.py
```diff
@@ -328,12 +324,136 @@ def __init__(self, parent=None, examples=None):
...
+        self.docs_browse_action.setIcon(icontheme.lookup('lookup-musicbrainz'))
+        self.docs_browse_action.triggered.connect(self.docs_browser)
+        help_menu.addAction(self.docs_browse_action)
+
+        self.ui.layout_for_menubar.addWidget(main_menu)
+
+    def add_tooltips(self):
+        """Add tooltips to the widgets.
+        """
+        # Done here to avoid duplication and allow changes/corrections without having to recompile the UI file.
```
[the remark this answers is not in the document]
rdswift: Actually, it does make more sense as you suggested.  Changed.

phw (approved):
Great work, thanks a lot for implementing this. The menu is really a big improvement over trying to get all the buttons into the UI.

It works really well, and the dialog offers us much flexibility for future improvements. LGTM

PR #1813, opened 2021-05-08
PICARD-2207: Show script editor tooltips for functions and Unicode escapes

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2207

This shows a function's documentation or the Unicode character name for a Unicode escape sequence as a mouse over tooltip in the script editor.

# Solution

Listen to mouse move events in the script editor. Determine the document position for the mouse cursor and check whether there is a matching entity (function or unicode escape) at that position by backtracking until either `$`  or `\u` is being found. Then handle getting proper tooltip depending on the entity.

Functions will show the documentation:

grafik

For control characters it will show the Unicode name and if it is a printable character also the character itself:

grafik

For control characters only the name is shown:

grafik

# Action

### DIFF
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -186,6 +186,81 @@ class ScriptCompleter(QCompleter):
         return self.last_selected
 
 
+class DocumentedScriptToken:
+
+    allowed_chars = re.compile('[A-Za-z0-9_]')
+
+    def __init__(self, doc, cursor_position):
+        self._doc = doc
+        self._cursor_position = cursor_position
+
+    def is_start_char(self, char):
+        return False
+
+    def is_allowed_char(self, char, position):
+        return self.allowed_chars.match(char)
+
+    def get_tooltip(self, position):
+        return None
+
+    def _read_text(self, position, count):
+        text = ''
+        while count:
+            char = self._doc.characterAt(position)
+            if not char:
+                break
+            text += char
+            count -= 1
+            position += 1
+        return text
+
+
+class FunctionScriptToken(DocumentedScriptToken):
+
+    def is_start_char(self, char):
+        return char == '$'
+
+    def get_tooltip(self, position):
+        doc = self._doc
+        char = doc.characterAt(position)
+        if char != '$':
+            return None
+        function = ''
+        while True:
+            position += 1
+            char = doc.characterAt(position)
+            if not self.allowed_chars.match(char):
+                break
+            function += char
+        try:
+            return script_function_documentation(function, 'html')
+        except ScriptFunctionDocError:
+            return None
+
+
+class UnicodeEscapeScriptToken(DocumentedScriptToken):
+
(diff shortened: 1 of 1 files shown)


phw:
Added tooltip for variables also, showing the localized human readable tag name if available:

grafik

phw:
This actually works quite well for me. I'd appreciate some testing what others think of this. Are the tooltips ok, or is it annoying?

An alternative would be to show this tip only with an explicit action, e.g. a context menu entry or keyboard shortcut, that triggers showing this tooltip on demand.

zas:
>  Are the tooltips ok, or is it annoying?

To me, ok. Once we'll have menu we can add something to toggle those on/off (along with other "display" options, like Line Wrap & Documentation)

Sophist-UK:
I guess tooltips might be annoying to some people esp. if you leave the cursor over text and you are typing.

As an alternative to having yet another option to turn them off, how about only showing tooltips when the cursor has moved in the last (say) 10 seconds. If you leave the cursor over the text, the tooltip will disappear after 10 seconds and not reappear even if the text underneath changes.

phw:
> how about only showing tooltips when the cursor has moved in the last (say) 10 seconds

Tooltips are shown when moving the mouse already.

> If you leave the cursor over the text, the tooltip will disappear after 10 seconds and not reappear even if the text underneath changes.

I considered adding a timeout, but I find these timeouts either annoying or useless most of the time, in this case certainly. You either set this timeout really low, and then users are fighting the timeout when they want to read a longer tooltip. Or you set it rather high, let's say 30 or even 60 seconds, and then I don't really see the point.

phw:
> I noticed some warnings on console when appending characters at the end of the script, but they also happen on master, so likely unrelated to this change (`QTextCursor::setPosition: Position '192' out of range`)

I think you mentioned that before, but I have not yet been able to reproduce. It must be related to autocompletion, because that is where `QTextCursor::setPosition` is being used. Can you try to find out what exactly is needed to reproduce this issue?

zas:
> Can you try to find out what exactly is needed to reproduce this issue?

I tried, it happens when moving text cursor to the end of line, but that's not because of one of ours `TextCursor.setPosition()`, so prolly something internal to Qt. That's just a warning, it is generated by https://code.woboq.org/qt5/qtbase/src/gui/text/qtextcursor.cpp.html#1165
I can reproduce with Python 3.8.5, PyQt 5.15.4, Qt 5.15.2 (just moving the cursor to the end of line, and the warning is generated when it hits it)

So I don't think that's a Picard problem.

phw:
I still can't reproduce it, with same PyQt / Qt versions (tried on both my Ubuntu and Windows machine). I can get such warnings when intentionally calling cursor.setPosition outside the range. But I also reviewed our usages of cursor.setPosition, and as far as I can see we always make sure to be inside the current document, so I guess you are likely right and that warning is triggered by something unrelated to our code.

PR #1815, opened 2021-05-12
PICARD-2208: Allow resizing the "Before" and "After" columns in file naming examples

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Adds a splitter bar to allow resizing the "Before" and "After" columns in the file naming examples.  Also removes two unused layout widgets.

# Problem

The "Before" and "After" file naming example columns are fixed width on the Options --> File Naming screen, but they are sizable on the script editor screen.  This is inconsistent, and doesn't allow the same flexibility.

* JIRA ticket (_optional_): PICARD-2208

# Solution

Add a splitter bar to allow resizing the "Before" and "After" columns in the file naming examples on the Options --> File Naming screen.

# Action

None.


phw (changes requested):
The file naming script combo box behaves strange, I think the QGroupBox it is using is not needed here:

grafik

A label followed by a QComboBox would be sufficient.

Also the Before / After columns look slightly different in the filenaming option pages and the scripting dialog. In the options page there is a separator also between the column headers, in the dialog there isn't. I think it looks a bit nicer in the dialog. I don't know exactly why it is different, but this is the dialog:

grafik

And here is the same in the options:

grafik

Somehow there are more visible frames, that make this display look rather busy. I think the only difference is setting the two frames (left / right) to frameShape=NoFrame, otherwise I think the structure is the same.

Also setting the GroupBox for the filenaming examples to vertically MinimumExpanding gives the entire view a better behavior when resizing I think

rdswift:
@phw 

Agreed on all points and updated accordingly.  Thanks for catching them.

zas (approved):

phw (approved):

phw:
Thanks a lot

PR #1816, opened 2021-05-13
PICARD-2179: Persist state of scripting docs pane

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Persist the state of the scripting docs in the script editor dialog.

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -33,6 +33,7 @@ from PyQt5.QtGui import QPalette
 
 from picard import log
 from picard.config import (
+    BoolOption,
     ListOption,
     TextOption,
     get_config,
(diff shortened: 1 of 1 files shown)


zas (approved):

rdswift (approved):
This looks good to me.

Is there any benefit in also persisting things like window size and location, similar to the Options window?  I'm okay without it, but others might find it odd that this one dialog always starts with the default size.

phw:
Yes, I think it would be beneficial to restore this. Since this dialog derives from PicardDialog a simple

    self.restore_geometry()

in init after setupUi() should already fix this

rdswift:
Along with a corresponding `self.save_geometry()` in the closing event.  Were you planning on adding this or did you want me to add a PR for it?

phw:
> Along with a corresponding self.save_geometry() in the closing event. 

That's handled automatically already.

I'll add that when I merge this PR

rdswift:
> That's handled automatically already.

I'm not so sure about that because `closeEvent()` has been redefined in `scripteditor.py`.  Perhaps the redefined close event should also be calling `super.closeEvent()` ?

phw:
You are right, closeEvent should call super, but only if it is actually closing. This would also ensure other things on the finished signal would run. I implemented saving the window state at https://github.com/metabrainz/picard/pull/1817

PR #1818, opened 2021-05-14
PICARD-2209: Ensure DesktopStatusIndicator is registered only once

phw:
Otherwise this can lead to performance issues if the user minimizes / maximizes the main window multiple times, causing multiple DesktopStatusIndicator getting registered.

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2209

When the user toggles the main window between minimized and maximized Picard will register the DesktopStatusIndicator multiple times. Doing this repeatedly can cause slow downs due to multiple registered indicators. 

# Solution

Remember whether showEvent was already called or not. We cannot move this into e.g. __init__, as for Windows the window must actually exist before you can access the taskbar button's API.

# Action

Would QWidget provide another place to call this, but only once the widget has been fully created?

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -168,6 +168,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
     defaultsize = QtCore.QSize(780, 560)
     selection_updated = QtCore.pyqtSignal(object)
+    ready_for_display = QtCore.pyqtSignal()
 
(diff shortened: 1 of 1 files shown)


phw:
I'll refactor this a bit based on #1819

zas (approved):

PR #1820, opened 2021-05-15
Add automatic save / restore of splitter settings

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

The `PreserveGeometry` helper class does not automatically save and restore splitter information, so this must be coded manually for each dialog.

# Solution

Modify the helper class to automatically save the state of all splitters found when processing the `save_geometry()` method, and to restore the states when the `restore_geometry()` method is executed.  The state of each splitter found is stored in a dictionary for the dialog, with the keys based on the lineage (ancestry) of the splitters.  When restoring, each splitter found in the dialog is checked against the dictionary and its state is updated if a corresponding key is found.

This does not require that a splitter has an assigned `objectName()`, however names should be assigned if possible to help ensure that the lineage string key is unique for each splitter within the dialog.

# Action

If this change is accepted, we will need to find a way to migrate any existing splitter settings before they are removed from the user's persistent settings.


picard/ui/__init__.py
```diff
@@ -61,12 +61,38 @@ class PreserveGeometry:
...
+        Args:
+            widget (QtWidget): Widget to process.
+            lineage (str, optional): Starting string. Defaults to ''. This is used during recursion and should not be specified in the initial call.
+
+        Returns:
+            str: Lineage / ancestry string for the specified widget.
+        """
+        if widget is not None:
+            # If widget doesn't have an assigned objectName() use the widget class name to help create a unique lineage string.
+            name = widget.objectName() if widget.objectName() else widget.__class__.__name__
```
zas, line 86: Since "names should be assigned if possible to help ensure that the lineage string key is unique for each splitter within the dialog", it would make sense to log (debug level) splitters not having an `objectName`.

zas (commented):
> _If this change is accepted, we will need to find a way to migrate any existing splitter settings before they are removed from the user's persistent settings._

Not sure it is worth bothering. It would be better, but I don't think we'll get much complaints if some splitters aren't properly preserved when upgrading to next major release.

I like the patch, it makes things a bit simpler to maintain.

phw:
I like the patch, makes it really easier to implement those dialogs that preserve their layout and ensures behavior is consistent. 

> Not sure it is worth bothering. It would be better, but I don't think we'll get much complaints if some splitters aren't properly preserved when upgrading to next major release.

Yes, I wouldn't bother with that really. I think it is very important for normal preferences, but not really for window state. Window layouts are subject to change on a release. So I think it would be ok to just remove the customer splitter preserve / restore logic from other dialogs.

rdswift:
> Not sure it is worth bothering. It would be better, but I don't think we'll get much complaints if some splitters aren't properly preserved when upgrading to next major release.

It was actually easier than I expected, so I added a configuration upgrade action.

picard/config_upgrade.py
```diff
@@ -342,6 +343,49 @@ def upgrade_to_v2_6_0_beta_3(config):
...
+    )
+
+    # ScriptEditorDialog splitters
+    _check_and_add(
+        new_key="splitters_ScriptEditorDialog",
+        key_map=[
+            ('script_editor_splitter_samples', '.splitter_between_editor_and_examples.ScriptEditor.RenamingOptionsPage.pages_stack.splitter.Dialog.MainWindow'),
+            ('script_editor_splitter_samples_before_after', '.splitter_between_before_and_after.groupBox.splitter_between_editor_and_examples.ScriptEditor.RenamingOptionsPage.pages_stack.splitter.Dialog.MainWindow'),
+            ('script_editor_splitter_documentation', '.splitter_between_editor_and_documentation.frame.frame_4.splitter_between_editor_and_examples.ScriptEditor.RenamingOptionsPage.pages_stack.splitter.Dialog.MainWindow'),
+        ]
```
zas, line 376: I wonder if we really need such "lineage", I see potential issues with that:
- when UI is modified, lineage may too, so saved config is lost, a new one will be create, but old one will stay in config file
- if we want to avoid that, we need to write upgrade functions after each UI change that may concern a splitter, that's not great
- isn't Objectname usable alone for that? If we enforce a unique and mandaroty name for each splitter, we don't need the whole lineage

Is this making sense?

rdswift: > Is this making sense?

Yes.  All good points.  I'll make the appropriate changes.

rdswift: It now uses objectName() value if it is available, otherwise it will use the lineage.  I also added names to the two main screen splitters.

picard/ui/__init__.py
```diff
@@ -61,12 +63,51 @@ class PreserveGeometry:
...
+        """
+        name = widget.objectName()
+        if not name:
+            name = self._get_lineage(widget).strip('.')
+            log.debug("Splitter does not have objectName(): %s" % name)
+        return name
+
+    def _get_splitter_items(self):
+        try:
+            return self.findChildren(QtWidgets.QSplitter)
```
zas, line 107: See https://github.com/rdswift/picard/pull/5/commits/6f65a408a334982c564c7d1c4ef5becbcf127c01

picard/ui/__init__.py
```diff
@@ -75,10 +116,19 @@ def restore_geometry(self):
             self.restoreGeometry(geometry)
         elif self.defaultsize:
             self.resize(self.defaultsize)
+        splitters = config.persist[self.splitters_name()]
+        if splitters:
+            splitter_items = self._get_splitter_items()
+            for splitter in splitter_items:
+                name = self._get_name(splitter)
+                if name in splitters:
+                    splitter.restoreState(splitters[name])
```
zas, line 125: I would also remove unused entries (this can happen after a change to object name)

See https://github.com/rdswift/picard/pull/5/commits/0a129ab520beb8abea369db1bfa05e503721f352

zas (changes requested):

picard/ui/__init__.py
```diff
@@ -61,12 +63,51 @@ class PreserveGeometry:
...
+
+        Args:
+            widget (QtWidget): Widget to process.
+
+        Returns:
+            str: The name of the widget or the lineage if there is no name assigned.
+        """
+        name = widget.objectName()
+        if not name:
+            name = self._get_lineage(widget).strip('.')
```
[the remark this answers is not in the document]
rdswift: Much nicer.  Merged.  Thanks.

picard/ui/mainwindow.py
```diff
@@ -361,11 +361,9 @@ def restoreWindowState(self):
         self.restore_geometry()
         if config.persist["window_maximized"]:
             self.setWindowState(QtCore.Qt.WindowMaximized)
-        bottom_splitter_state = config.persist["bottom_splitter_state"]
-        if bottom_splitter_state.isEmpty():
+        splitters = config.persist["splitters_MainWindow"]
+        if splitters is None or 'main_window_bottom_splitter' not in splitters:
```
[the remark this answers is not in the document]
phw: Just removing setting sizes here would make the bottom metadata view 50% height, which is awkward as a default. The hard coded values here aren't great, either. But at least with empty config it means the lower part has it's minimum height (limited by the cover art box).

picard/ui/itemviews.py
```diff
@@ -191,6 +190,7 @@ class MainPanel(QtWidgets.QSplitter):
     def __init__(self, window, parent=None):
         super().__init__(parent)
         self.setChildrenCollapsible(False)
+        self.setObjectName('main_panel_splitter')
```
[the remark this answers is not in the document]
rdswift: Good point.  Moved splitter object naming to MainWindow class.

picard/ui/itemviews.py
```diff
@@ -235,15 +235,12 @@ def tab_order(self, tab_order, before, after):
...
-        config = get_config()
-        config.persist["splitter_state"] = self.saveState()
         for view in self._views:
             view.save_state()
 
     @restore_method
     def restore_state(self):
-        config = get_config()
-        self.restoreState(config.persist["splitter_state"])
+        pass
```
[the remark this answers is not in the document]
rdswift: Nope.  Removed.

picard/config_upgrade.py
```diff
@@ -371,4 +415,5 @@ def upgrade_config(config):
     cfg.register_upgrade_hook(upgrade_to_v2_6_0_dev_1)
     cfg.register_upgrade_hook(upgrade_to_v2_6_0_beta_2)
     cfg.register_upgrade_hook(upgrade_to_v2_6_0_beta_3)
+    cfg.register_upgrade_hook(upgrade_to_v2_7_0_dev_1)
```
zas, line 418: That's not totally correct, because the hook will not run for people running current dev version (2.7.0.dev1).
We need to bump version to 2.7.0dev2 and name this upgrade hook `upgrade_to_v2_7_0_dev_2`

picard/config_upgrade.py
```diff
@@ -342,6 +343,49 @@ def upgrade_to_v2_6_0_beta_3(config):
     _s.remove("use_system_theme")
 
 
+def upgrade_to_v2_7_0_dev_1(config):
+    """Replace manually set persistent splitter settings with automated system.
+    """
+    def _check_and_add(new_key, key_map):
```
zas, line 349: This name isn't very explicit. I suggest `upgrade_persisted_splitter()` (or something like that?)

Also `new_key` is used in parameters and `newkey` in the loop, that's ambiguous.
What about `new_persist_key` and `old/new_splitter_key`?

picard/ui/options/dialog.py
```diff
@@ -150,7 +149,7 @@ def __init__(self, default_page=None, parent=None):
         # work-around to set optimal option pane width
         self.ui.pages_tree.expandAll()
         max_page_name = self.ui.pages_tree.sizeHintForColumn(0) + 2*self.ui.pages_tree.frameWidth()
-        self.ui.splitter.setSizes([max_page_name,
+        self.ui.dialog_splitter.setSizes([max_page_name,
```
zas, line 152: The indentation of the next line needs to be updated (not sure it needs to be wrapped though).

Unrelated to this PR, but what about:

```python
left = self.ui.pages_tree.sizeHintForColumn(0) + 2*self.ui.pages_tree.frameWidth()
right = self.geometry().width() - left
self.ui.dialog_splitter.setSizes([left, right])
```

And stick with such notation for splitters, (left/right, top/bottom).

zas (changes requested):

picard/config_upgrade.py
rdswift: I was wondering how best to handle that.  Will make the change.  Thanks.

picard/ui/options/dialog.py
rdswift: That makes it a lot easier to understand.  I like it.

zas (approved):
LGTM

phw (approved):
A great improvement. I was sceptical at first, as I was not sure whether this would bring much benefit or could be generalized enough. But the result is really great. This now also means the UI is now more consistent, as some additional splitters that were not preserved before are now preserved.

I think if the user can manually change the size of a widget it is annoying if this resets easily.

PR #1821, opened 2021-05-17
Allow disabling graphical crash handler with --no-crash-dialog

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

This is mainly for running Picard in CI environments and debugging. E.g. on CI we sometimes run `picard --version`  just to verify the build package is working. But if that fails without graphical environment this fails hard when trying to show the error dialog with an error message unrelated to the original issue. See e.g. https://github.com/phw/picard/runs/2594991408:

> qt.qpa.plugin: Could not load the Qt platform plugin "xcb" in "" even though it was found.
> This application failed to start because no Qt platform plugin could be initialized. Reinstalling the application may fix this problem.
>
> Available platform plugins are: eglfs, linuxfb, minimal, minimalegl, offscreen, vnc, wayland-egl, wayland, wayland-xcomposite-egl, wayland-xcomposite-glx, webgl, xcb.
>
> /home/runner/work/_temp/fe09af7f-24b1-48eb-8c5a-def104386bab.sh: line 1:  3488 Aborted                 (core dumped) picard --long-version
> Error: Process completed with exit code 134.

# Solution

# Action

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -110,7 +110,7 @@ jobs:
         pip install .
     - name: Test running installed package
       if: runner.os != 'Windows'
-      run: picard --long-version
+      run: picard --long-version --no-crash-dialog
     - name: Verify sdist package
       if: runner.os != 'Windows'
       run: |
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -99,7 +99,7 @@ def crash_handler():
     import sys
 
     # Allow disabling the graphical crash handler for debugging and CI purposes.
-    if '--no-crash-dialog' in sys.argv:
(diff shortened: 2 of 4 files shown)


zas:
Perhaps we should automatically enable `--no-crash-dialog` when running picard with pure command-line options (like `--version`)

phw:
Yes, that would make sense. I also thought about enabling this automatically, but detecting if Qt cannot run fully automatically is difficult, and we don't seem to be able to handle the crash that is caused by trying, as this segfaults directly.

Do we have any other condition except `--version` / `--long-version` that would apply?

zas:
> Do we have any other condition except `--version` / `--long-version` that would apply?

I don't think so. All other options only make sense in graphical context.

zas (approved):

PR #1822, opened 2021-05-17
PICARD-2213: Use YAML for script package export/import

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2213, PICARD-2179

Use YAML for the export of a script package. This allows for easier manual editing of the file, especially the script itself.

An example file generated can be seen at

https://gist.github.com/phw/095f4c28e5135487090863a1c2afc5a7

# Solution

# Action

I think we should have the pyyaml dependency optional. I'd rather not have two different script package formats, so I'd stick with YAML, but disable package export if that is not available. The export as a plain text of the script only will be always available.

@zas @rdswift Thoughts?


picard/ui/scripteditor.py
```diff
@@ -288,7 +289,7 @@ def __init__(self, parent=None, examples=None):
 
         self.FILE_TYPE_ALL = _("All Files") + " (*)"
         self.FILE_TYPE_SCRIPT = _("Picard Script Files") + " (*.pts *.txt)"
-        self.FILE_TYPE_PACKAGE = _("Picard Naming Script Package") + " (*.pnsp *.json)"
+        self.FILE_TYPE_PACKAGE = _("Picard Naming Script Package") + " (*.pnsp *.yaml)"
```
zas, line 292: Is the intent to replace json or is it in addition?

zas:
> I'd rather not have two different script package formats, so I'd stick with YAML, but disable package export if that is not available.

To me the main issue is the extra dependency (while json support is native).
I wonder if we should support json & yaml, json is more widely supported natively by languages & tools, YAML easier to edit by humans.
But we assume a lot of people will edit by hand (rather than using Picard interface) here. Not really sure about that.

phw:
I intentionally wanted only a single format, as I don't see the benefit of having multiple. But the dependency bothers me as well, and having this optional could mean some Picard installs support importing the YAML files, others don't. Won't apply to our official binary builds, but e.g. on Arch Linux this dependency for sure would be optional.

Overall for me this is about thus difference:

```json
{
    "author": "",
    "description": "",
    "id": "766bb2ce-5170-45f1-900c-02e7f9bd41cb",
    "last_updated": "2021-05-09 15:47:57 UTC",
    "license": "",
    "script": "$if($eq(%_extension%,fla),$set(_extension,flac))\n$if($is_video(),video,$if($is_lossless(),lossless,lossy))/\n$if($is_video(),\n$noop(Video track)\n$if($ne(%album%,[non-album tracks]),\n$if2(%albumartist%,%artist%) - %album%$if(%discsubtitle%, - %discsubtitle%)/%_discandtracknumber% %title%,\nMusic Videos/%artist%/%artist% - %title%),\n$if($eq(%compilation%,1),\n$noop(Various Artist albums)\n$firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \\(%album_edition%\\))//%_discandtracknumber% %artist% - %title%,\n$noop(Single Artist Albums)\n$firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \\(%album_edition%\\))//%_discandtracknumber% %title%\n))",
    "script_language_version": "1.1",
    "title": "Primary file naming script2",
    "version": ""
}
```

vs. 

```yml
author: outsidecontext
description: My default naming script
id: 766bb2ce-5170-45f1-900c-02e7f9bd41cb
last_updated: 2021-05-14 13:16:45 UTC
license: MIT
script: |
  $if($eq(%_extension%,fla),$set(_extension,flac))
  $if($is_video(),video,$if($is_lossless(),lossless,lossy))/
  $if($is_video(),
  $noop(Video track)
  $if($ne(%album%,[non-album tracks]),
  $if2(%albumartist%,%artist%) - %album%$if(%discsubtitle%, - %discsubtitle%)/%_discandtracknumber% %title%,
  Music Videos/%artist%/%artist% - %title%),
  $if($eq(%compilation%,1),
  $noop(Various Artist albums)
  $firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \(%album_edition%\))//%_discandtracknumber% %artist% - %title%,
  $noop(Single Artist Albums)
  $firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \(%album_edition%\))//%_discandtracknumber% %title%
  ))
script_language_version: '1.1'
title: outsidecontext's file naming
version: '1.0'
```

The main data, the script, is well readable in the YAML output, even editable. External editors could even have proper syntax highlighting for this, e.g. I'd try to extend the vscode syntax highlighting for tagger script to cover the YAML format as well.

Hence my patch only supports one. But we could also always support the JSON output, then at least that would be available always. I think we should have distinct file extensions in for both then. Suggestions?

zas:
> ```yaml
> ```yaml
> 
> ```
> 
> 
>     
>       
>     
> 
>       
>     
> 
>     
>   
> The main data, the script, is well readable in the YAML output, even editable.

It is definitively much more readable.
And I tend to agree having 2 formats (in fact 3, since we support plain text) is a bit too much.
So let's go for YAML only support, with the required dependency (in the future, I'm pretty sure a native Python module will support that).

rdswift:
I was actually going to suggest having options for both JSON and YAML, but I can also see that might be a bit too much.  My only concern would be if we allow only YAML, when people try distributing script packages there may be no way to import them on some platforms.  Perhaps we need to create our own text file format for the package files, something like:

```
[metadata]
author=Joe Blow
license=GPL 2.0
title=My Wonderful Script
lastupdated=2021-05-16 10:12:15 UTC
[end metadata]

[description]
This section allows multi-line
description.
[end description]

[script]
$noop(
  This allows multi-line scrips and should be easily editable.
)
[end script] 
```

It should be easy enough to come up with a read / write parser.

zas:
> Perhaps we need to create our own text file format for the package files

I'd rather add an extra dependency rather than adding parsing code for that.

About JSON vs YAML, both have pros & cons. 
It's rather simpler to convert one to the other, so I don't think we need to support both.

One thing that give YAML another advantage: if we get rid of JSON for plugins metadata (when we'll redesign plugin system) it would fit well.
Also we already talked about reworking config system, I think YAML can be a good format if we want the config file to be human-editable again (it can easily replace INI format).
PyYAML is C-based so it is fast, it is used by many important projects (so unlikely to disappear).

phw:
> One thing that give YAML another advantage: if we get rid of JSON for plugins metadata (when we'll redesign plugin system) it would fit well.
> Also we already talked about reworking config system, I think YAML can be a good format if we want the config file to be human-editable again (it can easily replace INI format).

Fully agree on both. YAML was designed for better human readability and more complete type support. It achieves this at the cost of higher complexity, both of the spec and the implementation. JSON is more simple, easier to parse and create.

But I think this makes YAML a really good fit for config files and plugin metadata.

phw:
I think this is overall ready for review. Just one point I am of mixed opinion: For the current functionality pyyaml is optional, not having it disabled the package export. For now that would work, and we handle e.g. Markdown support similar.

But it could mean there will be Picard installs without this support, mostly on Linux and similar. Three paths to go:

1. Keep it as is, and put the responsibility on the package maintainers to properly set pyyaml as an optional or recomended dependency. Also we probably would want to look into setting up `extra_requires` for the pip package.
2. Middle ground: Keep it optional, but document it as required and also make the pip package as requiring it.
3. Make it a hard requirement (meaning Picard will fail to start without this package)

Originally I was more in favour of 1 or 2, but I know think we should go with 3 actually. Experience shows that external package managers don't add new optional dependencies and just stick with the hard requirements. And if we want to avoid support requests by users unable to import the script package someone posted on the forums we better make this functionality non-optional.

Also I'd like to reorder the fields a bit into a more logical order (with e.g. title on top). Maybe like this:

```
title: outsidecontext's file naming
description: My default naming script
author: outsidecontext
license: MIT
id: 766bb2ce-5170-45f1-900c-02e7f9bd41cb
last_updated: 2021-05-14 13:16:45 UTC
version: '1.0'
script_language_version: '1.1'
script: |
  $if($eq(%_extension%,fla),$set(_extension,flac))
  $if($is_video(),video,$if($is_lossless(),lossless,lossy))/
  $if($is_video(),
  $noop(Video track)
  $if($ne(%album%,[non-album tracks]),
  $if2(%albumartist%,%artist%) - %album%$if(%discsubtitle%, - %discsubtitle%)/%_discandtracknumber% %title%,
  Music Videos/%artist%/%artist% - %title%),
  $if($eq(%compilation%,1),
  $noop(Various Artist albums)
  $firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \(%album_edition%\))//%_discandtracknumber% %artist% - %title%,
  $noop(Single Artist Albums)
  $firstalphachar($if2(%albumartistsort%,%artistsort%))/$if2(%albumartist%,%artist%)/%album%$if(%album_edition%, \(%album_edition%\))//%_discandtracknumber% %title%
  ))
```

zas:
> Originally I was more in favour of 1 or 2, but I know think we should go with 3 actually. Experience shows that external package managers don't add new optional dependencies and just stick with the hard requirements. And if we want to avoid support requests by users unable to import the script package someone posted on the forums we better make this functionality non-optional.

I agree.

> 
> Also I'd like to reorder the fields a bit into a more logical order (with e.g. title on top).

Yes, since it's about readibility it makes sense to enforce fields order.

phw:
I did the updates:

- YAML export now has a defined field order
- `description` gets expported as string literal
- pyyaml is now a required dependency again. Picard supporting import/export of scripting package is mandatory
- I moved the classes used for script serialization into separate submodule
- Extended tests to check YAML serialization

picard/ui/scripteditor.py
```diff
@@ -288,7 +290,7 @@ def __init__(self, parent=None, examples=None):
...
```
rdswift, line 293: This may be a dumb question, but is there a possibility in the future of also saving tagging scripts in this same format?  If so, should the description be changed to "Picard Script Package" and the extension to (perhaps) "*.psp"?

phw: Yes, I think definitely, that should be a goal. What about `*.ptsp` ? So we have `.pts` for Picard Tagger Script, and `.ptsp`  for Picard Tagger Script Package?

phw:
Rebased against current master to make use of https://github.com/metabrainz/picard/pull/1824

zas:
Is there any reason to keep export/import to json with this patch?

rdswift:
> Is there any reason to keep export/import to json with this patch?

The list of available scripts in the user's settings is currently stored as JSON.  I suppose this could be changed to YAML.

phw:
Yes, it is currently still used for the internal serialization. It would have made sense to keep it if we made PyYAML optional, but since we made it mandatory I also see no pressing reason to keep it.

phw:
I removed the JSON serialization, all based on YAML now.

Nice detail: Even if you had been using this already with saved scripts, these will still be available after the change. As JSON is a valid YAML subset the currently stored scripts are loaded just fine. They will be converted into the "full" YAML format when saving.

zas (approved):
LGTM

kellnerd:
FYI, I just tested this with the latest commit on `master` using my file naming script. Since my script is formatted using tabs (in addition to linebreaks), the `script` property was exported as an ugly quoted string instead of the raw multiline block which can be seen in the example:

```yaml
title: Primary file naming script
description: ""
author: ''
license: ''
version: ''
last_updated: 2021-07-24 14:56:20 UTC
script_language_version: '1.1'
script: "$noop( File properties )\n$if($ne_any(%_extension%,mp3),\n\t$set(_FileFormat,$upper(%_extension%))\n\
  )\n$if($gt(%_bits_per_sample%,16),\n\t$set(_HighPrecision,%_bits_per_sample%bit)\n\
  )\n$if($gt(%_sample_rate%,44100),\n\t$set(_HighResolution,$div(%_sample_rate%,1000)kHz)\n\
  )\n$set(_FileInfo,$strip(%_FileFormat% $releasetag_avg(_bitrate,0)kbps %_HighPrecision%\
  \ %_HighResolution%))\n$if(%_FileInfo%,\n\t$set(_FileInfo, [%_FileInfo%])\n)\n\n\
  $noop( Generate path from custom variables, fall back to standard tags )\n$set(_path,\n\
  \t$noop( Root directory and artist/series folder )\n\t$if2(%_SpecialReleaseType%,Music)\n\
  \t/\n\t$if2(%_SeriesFolder%,%_ArtistFolder%,%albumartist%,%artist%)\n\t/\n\t$noop(\
  \ Release folder, except for standalone recordings )\n\t$if(%albumartist%,\n\t\t\
  $if2(%_EpisodeFolder%,%_AlbumFolder%,%album%)%_FileInfo%\n\t/)\n\t$noop( Disc and\
  \ track number, if available, except for standalone recordings )\n\t$if($and(%tracknumber%,%albumartist%),\n\
  \t\t$if(%_MultiDisc%,\n\t\t\t$if2(%_PaddedDiscNum%,%discnumber%)\n\t\t-)\n\t\t$if($eq($num(%_musicbrainz_tracknumber%,0),0),$noop(\
  \ e.g. $num(A1,0) => \"0\" )\n\t\t\t$if2(%_musicbrainz_tracknumber%,%_PaddedTrackNum%,%tracknumber%)\n\
  \t\t,$noop( Track number is not alpha-numerical, prefer padded consecutive numbers\
  \ )\n\t\t\t$if2(%_PaddedTrackNum%,%tracknumber%)\n\t\t)\n\t. )\n\t$noop( Artist\
  \ credits for multi-artist releases and all standard releases )\n\t$if($or(%_multiartist%,$not(%_SpecialReleaseType%)),\n\
  \t\t$if2(%_TrackArtist%,%artist%)\n\t - )\n\t$noop( Track/Recording title )\n\t\
  $if2(%_TrackTitle%,%title%)\n)\n\n$noop( Additional custom replacements, e.g. to\
  \ prevent replacement of illegal characters with an underscore )\n$set(_path,$replace(%_path%,:\
  \ , - ))\n$set(_path,$replace(%_path%,?,))\n\
  \n$noop( Output the constructed path, remove line breaks and indentation caused\
  \ by script formatting )\n$rreplace(%_path%,[\\\\t\\\\n]+,)\n"
id: dbbc8640-1131-40ad-8ae1-6c0d44f3343d
```

@phw Do you know whether it is possible to handle tabs in the `MultilineLiteral` YAML representer? Or do I have to abandon my formatting workaround to get nice export results?

PR #1824, opened 2021-05-19
Use requirements.txt for pip package install_requires

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

- We frequently forget to update setup.py if we have changes in requirements, leading to broken builds
- We do not currently specify version restrictions in setup.py, which might cause incompatible depedencies being used by pip
- Redundancy

# Solution

Read requirements.txt and use it for install_requires

# Action
I considered putting some of the dependency into extra_requires. But this is only about discid and markdown, and I think it is preferable to have both installed. Also I think it is an unneeded complication.

### DIFF
--- a/requirements.txt
+++ b/requirements.txt
@@ -1,9 +1,9 @@
-python-dateutil>=2.7.3
-discid
-fasteners
-markdown
-mutagen>=1.37
-pyobjc-core<7.0; sys_platform == 'darwin'
-pyobjc-framework-Cocoa<7.0; sys_platform == 'darwin'
-PyQt5>=5.10
+python-dateutil~=2.7
+discid~=1.0
+fasteners~=0.14
+markdown~=3.2
+mutagen~=1.37
+pyobjc-core~=6.2; sys_platform == 'darwin'
+pyobjc-framework-Cocoa~=6.2; sys_platform == 'darwin'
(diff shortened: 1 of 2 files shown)


phw:
In 8ea9a591 I also updated the version specification. Trying to specify a version for all dependencies and using the compatible release operator to allow only minor versions. E.g. currently PyQt5>=5.10 would also include 6.0, but that would be incompatible. Now it is PyQt5~=5.10, which matches all 5.x releases >= 5.10

Sophist-UK:
Can't you say `PyQt5>=5.10,<6`?

phw:
> Can't you say `PyQt5>=5.10,<6`?

That's equivalent to `PyQt5~=5.10`, see https://www.python.org/dev/peps/pep-0440/#compatible-release

Sophist-UK:
> > Can't you say `PyQt5>=5.10,<6`?
> 
> That's equivalent to `PyQt5~=5.10`, see https://www.python.org/dev/peps/pep-0440/#compatible-release

Ah - yes - so `~=` is actually better. 👍

zas (approved):
Lgtm

PR #1826, opened 2021-05-28
PICARD-2179: Renaming UI tweaks

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

This aims to tweak the new renaming / script editor UI a bit and make it more closely match other / standard dialogs

# Solution

- Move the button to open the script editor next to the script selection dropdown:
     
grafik 

- Use a QDialogButtonBox for the dialog buttons, placing the buttons at usual positions (system dependent)
- Add a help button / menu entry for the dialog. This points to the new URL for the script editor dialog docs as proposed at https://github.com/rdswift/picard-docs/pull/104

grafik

# Action


picard/ui/scripteditor.py
```diff
@@ -345,11 +352,6 @@ def make_menu(self):
         """
         config = get_config()
         main_menu = QtWidgets.QMenuBar()
-        base_font = self.ui.file_naming_editor_close.font()
-        main_menu.setStyleSheet(
-            "QMenuBar { font-family: %s; font-size: %upt; }" %
-            (base_font.family(), base_font.pointSize())
-        )
```
phw, line 352: @rdswift I could not figure out what this was supposed to fix. Do we still need this?

zas:
While you are on UI improvements I noticed the menu in the filenaming script editor is starting far away from left border, can you check if it is also the case for you?

image

phw:
Indeed. The QMenuBar is clearly meant to be used full width I fixed that by removing the margins of the dialog and adding another layout to hold the actual dialog content, which then re-adds the margins.

grafik

I need to test all these layout changes on Windows and macOS as well, so far I only worked on Linux.

phw:
Looks good on Windows. On macOS the menu bar is shown globally anyway, which seems to work just fine.

But the script editor still opens behind the options dialog, I'll look into this.

picard/ui/scripteditor.py
rdswift: I had to include that here because Qt was using a different font face and size for the menu bar here on my Windows system.  It seems that the menu was not inheriting the style properly.  I must have done something wrong, but wasn't sure how to fix it, so that was a quick hack.

rdswift (dismissed):
This all looks really good to me.  So much cleaner.

picard/ui/scripteditor.py
phw: Ok. At least for me the menu shows up with exactly the same font and font size as on the main window.

Was this on Windows 10 or 7?

rdswift: Windows 7.  I suspect that the work that you did may have fixed it, but I haven't checked yet.

phw:
Fixed the issue with window modality on macOS

picard/ui/scripteditor.py
```diff
@@ -260,6 +260,8 @@ class ScriptEditorDialog(PicardDialog):
     TITLE = N_("File naming script editor")
     STYLESHEET_ERROR = OptionsPage.STYLESHEET_ERROR
 
+    help_url = '/config/options_filerenaming_editor.html'
```
zas, line 263: This is returning a 404 for now, expected?

rdswift: I believe so because it hasn't been merged into the master branch yet.  It's still in the branch collecting the changes for the upcoming 2.7 release.

rdswift:
> Fixed the issue with window modality on macOS

There may be an issue with this fix.  On my Windows 7 system, it makes the script editor application modal and no longer allows you to change any of the other file naming settings in the Options dialog (e.g. "Move files when saving", "Destination directory" and "Rename files when saving") and show how they impact the script being edited.  As I recall, that was one of the things that @zas requested in my original development.

Unfortunately, I have nothing to offer as an alternative.  :frowning:

zas:
> On my Windows 7 system, it makes the script editor application modal and no longer allows you to change any of the other file naming settings in the Options dialog (e.g. "Move files when saving", "Destination directory" and "Rename files when saving") and show how they impact the script being edited.

In the current master, you can change the target directory in the underlying dialog, but it has no impact (since it isn't taken in account till "Make It So!"). So, in practice, this modality change is not much an issue.

That said, I think we are missing something here, we can now switch scripts easily, but we can't switch the full profile (ie. dest dir and/or windows compat options).
And I think it would be very useful to be able to do that (it was requested by users since years). If we make fileremaning profiles (storing a set of options), how will they play with script editor UI?

rdswift:
> In the current master, you can change the target directory in the underlying dialog, but it has no impact (since it isn't taken in account till "Make It So!"). So, in practice, this modality change is not much an issue.

Actually, in the current master when changes are made in the underlying dialog, they are reflected immediately in the examples displayed.

> That said, I think we are missing something here, we can now switch scripts easily, but we can't switch the full profile (ie. dest dir and/or windows compat options).  And I think it would be very useful to be able to do that (it was requested by users since years). If we make fileremaning profiles (storing a set of options), how will they play with script editor UI?

I'll have to think about that.  The one complication that comes to mind immediately is if the profile also includes the active plugins and, to a lesser extent the tagging scripts.  Because of the plugins, would changing the profile require a restart of Picard, or could we get by with simply toggling installed plugins as active or inactive?

If we leave the plugins (and possibly tagger scripts) out the profile for now, we should be able to save profiles containing the relevant settings from the File Renaming dialog including the selected naming script, and allow the user to quickly select between the available profiles.  We could probably even have similar management actions such as add, delete, copy, import and export of profiles either on the File Renaming dialog or as a separate sub-dialog.  There may even be some way to re-purpose the Script Editor dialog to also include the profile management, although I think it would be better to handle the profile selection and editing directly on the current File Renaming dialog.

Just thinking out loud here but what if the current File Renaming dialog had a profile selector at the top of the page and all settings shown below were those for the currently selected profile (and were updated when the selected profile changed)?  Changes to a profile's settings would be stored in the dialog's settings and not saved to the user's configuration until "Make It So!".  I think this might be relatively easy to do if each profile object contained the settings:

* Profile ID (str)
* Profile Name (str)
* Move files when saving (bool)
* Destination directory (bool)
* Move additional files (bool)
* Move additional files list (str)
* Delete empty directories (bool)
* Rename files when saving (bool)
* Replace non-ASCII characters (bool)
* Windows compatibility (bool)
* Selected naming script ID (str)

The user's main settings would contain a dictionary of available profiles (with key as Profile ID and value as YAML representation of the profile object) and the selected profile ID, similar to the way we now store a dictionary of the available naming scripts and the currently selected script ID.  In this case, we would no longer need to save the currently selected script ID in the user's settings because that would come from the setting in the currently selected profile.

That would leave the script editor for its intended purpose of editing / managing the collection of file naming scripts.  It would be a clean separation of the different options / functions.  Of course, we'll have to decide how we want to handle things like if a naming script that is used on a profile is deleted, but those are just implementation details.

In any case, I suggest that this be done in a separate PR.

rdswift:
@phw 

This is a bit of a hack, but it seems to work here on my Windows 7 system.  In the `ScriptEditorDialog` module, I tried adding:

```
    def show(self, *args, **kwargs):
        self.setModal(True)
        super().show()
        super().hide()
        self.setModal(False)
        super().show(*args, **kwargs)
```

Unfortunately I'm not able to do the macos testing so I'm not sure if it fixes the window opening behind issue on the mac.

picard/ui/scripteditor.py
phw: Yes, that's currently expected. We could also link it to the general renaming page for now and update the URL later when we have this page actually available, but I thought we would easily forget about that.

phw:
@rdswift
 The workaround would work, but I'm still investigating. For some reason the ScriptingDocumentationDialog (as it is still used in Options > Scripting) works as expected, even on macOS: It opens in front of options, but is non-modal

phw:
Ok, the issue is that the dialog gets created before the options dialog is even shown, that causes some confusion on macOS.

While the workwround above might work, the proper solution probably would be to enable this dialog only on demand. This requires some refactoring.

For now I have cleaned the modal settings a bit (it was both set in .ui files and code, with different values, sometimes overwriting what was set in code) and enforce the modal dialog only on macOS. Let's keep it that way for this PR.

I'll look into refactoring this a bit so it works properly on macOS without hacks.

phw:
> And I think it would be very useful to be able to do that (it was requested by users since years). If we make fileremaning profiles (storing a set of options), how will they play with script editor UI?

Two options come to mind:

1. Keep multiple scripts as is, but also add multiple profiles for the renaming options. As the script is the most complex of those settings this allows a user to have multiple renaming profiles, while either using the exactly same script or different ones.
2. Get rid of multiple scripts, and only have multiple renaming profiles. It's a bit more simple structure, but each renaming profile would have a separate script. That means even if you want to use the same script and only use e.g. different target directories you need to do script changes to all your profile.

I personally would prefer option 1. and that's also what Bob outlined above.

What I wonder is what should all be part of this profile? Currently it would make sense to have all the renaming options, but I can see how users then also want to make the tagging options part of this. E.g. you might want to save to id3v2.4 in one profile, but id3v2.3 in another profile. This could make sense if one uses a profile to save the files for e.g. their hardware player without prope ID3 support.

rdswift:
> What I wonder is what should all be part of this profile? Currently it would make sense to have all the renaming options, but I can see how users then also want to make the tagging options part of this. E.g. you might want to save to id3v2.4 in one profile, but id3v2.3 in another profile. This could make sense if one uses a profile to save the files for e.g. their hardware player without prope ID3 support.

That's an excellent point.  I never thought about the tagging format options.  So we're probably looking at (as a minimum):

* all renaming options, including selected file naming script
* all tagging script settings, or perhaps just which scripts are enabled
* all tagging options

I'm actually wondering if it might be easier to have multiple copies (one for each profile) of the user's entire `settings` and select which one is currently active for use / editing, along with providing some overall profile management functions (e.g. add, remove, copy, import / export).

zas (approved):
LGTM

PR #1828, opened 2021-05-29
PICARD-2215: Add script $countryname() function.

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds a new function to return the country name for a specified country code.

# Problem

Release country information is provided by country code, but there is no way to get the name of the country corresponding to the code.

* JIRA ticket (_optional_): PICARD-2215

# Solution

Add a new function to return the country name for a specified country code.

# Action

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1358,10 +1358,14 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
 
 
 @script_function(documentation=N_(
-    """`$countryname(country_code)`
+    """`$countryname(country_code, translate="")`
 
 Returns the name of the country for the specified country code.  If the country code is invalid an empty string will be returned.
+If translate is not blank, the output will be translated into the current locale language.
 """
 ))
-def func_countryname(parser, country_code):
-    return gettext_countries(RELEASE_COUNTRIES.get(country_code.strip().upper(), ""))
+def func_countryname(parser, country_code, translate=""):
+    name = RELEASE_COUNTRIES.get(country_code.strip().upper(), "")
+    if translate:
+        return gettext_countries(name)
+    return name
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -44,6 +44,7 @@ import operator
 import re
 import unicodedata
 
+from picard.const.countries import RELEASE_COUNTRIES
 from picard.metadata import MULTI_VALUED_JOINER
 from picard.script.parser import (
     MultiValue,
@@ -1354,3 +1355,16 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
     if not case_sensitive:
         multi_value._multi = list({v.lower(): v for v in multi_value}.values())
     return multi_value.separator.join(sorted(set(multi_value)))
+
+
+@script_function(documentation=N_(
+    """`$countryname(country_code)`
+
+Returns the name of the country for the specified country code.  If the country code is invalid an empty string will be returned.
+"""
+))
+def func_countryname(parser, country_code):
+    code = country_code.strip().upper()
+    if code not in RELEASE_COUNTRIES:
+        return ''
+    return RELEASE_COUNTRIES[code]
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1364,7 +1364,4 @@ Returns the name of the country for the specified country code.  If the country 
 """
 ))
 def func_countryname(parser, country_code):
-    code = country_code.strip().upper()
-    if code not in RELEASE_COUNTRIES:
-        return ''
-    return RELEASE_COUNTRIES[code]
+    return RELEASE_COUNTRIES.get(country_code.strip().upper(), "")
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1364,4 +1364,4 @@ Returns the name of the country for the specified country code.  If the country 
 """
 ))
 def func_countryname(parser, country_code):
-    return RELEASE_COUNTRIES.get(country_code.strip().upper(), "")
+    return gettext_countries(RELEASE_COUNTRIES.get(country_code.strip().upper(), ""))
--- a/test/test_script.py
+++ b/test/test_script.py
@@ -1575,9 +1575,11 @@ class ScriptParserTest(PicardTestCase):
         context["foo"] = "ca"
         context["bar"] = ""
         context["baz"] = "INVALID"
-        # Test with Russian locale
-        i18n.setup_gettext('build/locale', ui_language='ru')
-        self.assertScriptResultEquals("$countryname(ca)", "Канада", context)
+
+        # Test with Russian locale (Disabled because test fails on GitHub actions. Perhaps locale information is not available?)
+        # i18n.setup_gettext('build/locale', ui_language='ru')
+        # self.assertScriptResultEquals("$countryname(ca)", "Канада", context)
+
         # Reset locale to English for remaining tests
         i18n.setup_gettext('build/locale', ui_language='en')
         self.assertScriptResultEquals("$countryname(ca)", "Canada", context)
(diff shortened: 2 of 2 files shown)


picard/script/functions.py
```diff
@@ -1354,3 +1355,16 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
...
+    """`$countryname(country_code)`
+
+Returns the name of the country for the specified country code.  If the country code is invalid an empty string will be returned.
+"""
+))
+def func_countryname(parser, country_code):
+    code = country_code.strip().upper()
+    if code not in RELEASE_COUNTRIES:
+        return ''
+    return RELEASE_COUNTRIES[code]
```
mineo, line 1370:
```suggestion
    return RELEASE_COUNTRIES.get(code, "")
```

picard/script/functions.py
```diff
@@ -1354,3 +1355,16 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
     if not case_sensitive:
         multi_value._multi = list({v.lower(): v for v in multi_value}.values())
     return multi_value.separator.join(sorted(set(multi_value)))
+
+
+@script_function(documentation=N_(
+    """`$countryname(country_code)`
+
+Returns the name of the country for the specified country code.  If the country code is invalid an empty string will be returned.
```
mineo, line 1363: IMHO it's important to mention here that the return value is always in english, independent of the language of Picard's UI.

picard/script/functions.py
Sophist-UK: How about:
```
    return _(RELEASE_COUNTRIES.get(code, ""))
```
to allow for translation into local language.

This would require `picard.const.countries.py` to be changed so that each country was e.g.
```
    'AE': N_('United Arab Emirates'),
```
You might want a parameter e.g. `$countryname(code,translate)` where translate is non-blank to enable this translation.

picard/script/functions.py
rdswift: Actually, I'm thinking to wrap the output in `gettext_countries()` to return the translated value (similar to what is done in `picard.ui.options.releases.ReleasesOptionsPage`), but this breaks the tests because it isn't added to the `builtins` dictionary when the tests are executed.  Thoughts?

phw: Not entirely sure. I'd probably would not want the tag values to always change based on the UI language of Picard. A translate boolean flag as suggested by Sophist might be good.

rdswift: I was able to get the tests to work by importing the `picard.i18n` module.

picard/script/functions.py
rdswift: > This would require `picard.const.countries.py` to be changed so that each country was e.g.

Actually the translations are already available through the `picard.i18n` module.

> You might want a parameter e.g. `$countryname(code,translate)` where translate is non-blank to enable this translation.

Good idea.  Optional parameter added.

picard/script/functions.py
rdswift: As suggested, I added an option parameter to determine whether or not the output is translated.

phw:
I'll look into the failing tests. Maybe we just need to build the locales before running the test.

test/test_script.py
```diff
@@ -1568,3 +1568,35 @@ def test_cmd_unique(self):
...
             self.parser.eval("$unique(B:AB; D:C; E:D; A:A; C:X,1,:,extra)")
+
+    def test_cmd_countryname(self):
+        from picard import i18n
+        context = Metadata()
+        context["foo"] = "ca"
+        context["bar"] = ""
+        context["baz"] = "INVALID"
+
+        # Test with Russian locale (Disabled because test fails on GitHub actions. Perhaps locale information is not available?)
```
zas, line 1579: Yes, there's no sync for translation files in the workflow.
That said, it's a bad idea to rely on external files for tests (as values can change, unlikely in this case but still).

Instead you should override underlying `gettext_countries()` to return a constant value:

```python
        def mock_gettext_countries(arg):
            return "Канада"

        from unittest import mock
        with mock.patch('builtins.gettext_countries', mock_gettext_countries):
            self.assertScriptResultEquals("$countryname(ca)", "Canada", context)
            self.assertScriptResultEquals("$countryname(ca,)", "Canada", context)
            self.assertScriptResultEquals("$countryname(ca, )", "Канада", context)
            self.assertScriptResultEquals("$countryname(ca,yes)", "Канада", context)
```

zas (changes requested):

rdswift:
> I'll look into the failing tests. Maybe we just need to build the locales before running the test.

I believe that it's because the `build/locale` directory is in the `.gitignore` file, so the translations are not available.  I used the `mock` method suggested by @zas and all seems to be working well (and will provide consistent test results).

picard/script/functions.py
```diff
@@ -1175,7 +1176,7 @@ def func_foreach(parser, multi, loop_code, separator=MULTI_VALUED_JOINER):
 Standard 'while' loop. Executes `code` repeatedly until `condition` no longer
     evaluates to `True`. For each loop, the count is stored in the tag
     `_loop_count`. This allows the count value to be accessed within the `code`
-    script. The function limites the maximum number of iterations to 1000 as a
+    script. The function limits the maximum number of iterations to 1000 as a
```
zas, line 1179: Even though that's a trivial change it shouldn't be in this PR. I did in d13d89b3433c7791c5e27e50c6532912259d8fa5 (pushed to master branch).

zas (changes requested):

picard/script/functions.py
rdswift: Sorry about that.

test/test_script.py
```diff
@@ -1568,3 +1569,45 @@ def test_cmd_unique(self):
...
+
+    def test_cmd_countryname(self):
+        from picard import i18n
+        context = Metadata()
+        context["foo"] = "ca"
+        context["bar"] = ""
+        context["baz"] = "INVALID"
+
+        # Test with Russian locale
+        i18n.setup_gettext('build/locale', ui_language='ru')
```
phw, line 1581: I think the calls to i18n.setup_gettext can be removed, since this is now mocked

phw (changes requested):
After this has been rebased against master (to address zas' description fix) and i18n.setup_gettext  being removed from the tests (unless I missed something) that looks fine to me to merge.

test/test_script.py
rdswift: I believe that the i18n calls are still required (at least once) to add `gettext_countries` to `builtins`.  I get an `AttributeError` exception if I try running the tests locally with them removed.

phw: Ah, yes. I think that's right. Other tests also need to initialize this. They usually use `setup_gettext(self.localedir, "C")`, but since we are using locale specific stuff here let's keep it the way it is

rdswift: I was able to resolve this by removing the `i18n` calls and manually add `gettext_countries` to `builtins` if it wasn't found.

rdswift: Rebased to remove the typo commit.

phw (approved):

test/test_script.py
```diff
@@ -1568,3 +1570,42 @@ def test_cmd_unique(self):
...
+        context = Metadata()
+        context["foo"] = "ca"
+        context["bar"] = ""
+        context["baz"] = "INVALID"
+
+        def mock_gettext_countries_en(arg):
+            return "Canada"
+
+        def mock_gettext_countries_ru(arg):
+            return "Канада"
```
zas, line 1586: To fully mimic `gettext_countries()` those can be modified to:

```python
        def mock_gettext_countries_en(arg):
            return "Canada" if arg == 'Canada' else arg

        def mock_gettext_countries_ru(arg):
            return "Канада" if arg == 'Canada' else arg
```

If the string has not translation, it is returned as is (this is done after a conversion from country code to country name).

This would permit to test for unknown translation of a correct code:

```python
with mock.patch('builtins.gettext_countries', mock_gettext_countries_ru):
    self.assertScriptResultEquals("$countryname(fr,yes)", "France", context)
    self.assertScriptResultEquals("$countryname(INVALID,yes)", "", context)
```

And since English is the default language, the returned string is always the passed string, so we can simplify:

```python
        def mock_gettext_countries_en(arg):
            return arg
```

zas (changes requested):

test/test_script.py
rdswift: Excellent suggestions.  Thank-you.

zas (approved):
LGTM

mineo (approved):

PR #1829, opened 2021-05-31
PICARD-2218: Allow regular expressions in $performer(pattern)

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2218, PICARD-2217, PICARD-2216

As discussed in PICARD-2216 the `$performer(pattern)` function is not very flexible if you want to access a specific instrument. The pattern will always match anywhere in the string, e.g. `$performer(vocals)` will match both "performer:vocals" and "performer: background vocals". You have no way to match only exactly "performer:vocals". This is partly intentional (`$performer` is supposed to match a substring),  but also not very flexible. Having it support regular expressions would allow for e.g. `$performer(^vocals$)` for exact match.

While looking at that an actual bug was detected: Currently the pattern is matched against the full tag name (e.g. `performer:vocals`). That means any pattern that matches against `performer:`, such as `perf` or `:`, will always match all performance entries. This is for sure not expected behavior.

# Solution
Support regular expressions (PICARD-2218) and fix the bug of matching also the "performer:" part of the tag name (PICARD-2217)

# Action

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -764,15 +764,16 @@ def func_lenmulti(parser, multi, separator=MULTI_VALUED_JOINER):
     """`$performer(pattern="",join=", ")`
 
 Returns the performers where the performance type (e.g. "vocal") matches `pattern`, joined by `join`.
-You can specify a regular expression by surrounding the pattern with `/.../`. For example
-`/^guitars?$/` matches the performance type "guitar" or "guitars", but not e.g. "bass guitar".
+You can specify a regular expression in the format `/pattern/flags`. `flags` are optional. Currently
+the only supported flag is "i" (ignore case). For example `$performer(/^guitars?$/i)` matches the
+performance type "guitar" or "Guitars", but not e.g. "bass guitar".
 
 _Since Picard 0.10_"""
 ))
 def func_performer(parser, pattern="", join=", "):
     values = []
     try:
-        regex = pattern_as_regex(pattern, allow_wildcards=False, flags=re.IGNORECASE)
+        regex = pattern_as_regex(pattern, allow_wildcards=False)
     except re.error:
         return ''
     for name, value in parser.context.items():
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -52,7 +52,10 @@ from picard.script.parser import (
     ScriptRuntimeError,
     normalize_tagname,
 )
-from picard.util import uniqify
+from picard.util import (
+    pattern_as_regex,
+    uniqify,
+)
 
 
 try:
@@ -761,21 +764,21 @@ def func_lenmulti(parser, multi, separator=MULTI_VALUED_JOINER):
     """`$performer(pattern="",join=", ")`
 
 Returns the performers where the performance type (e.g. "vocal") matches `pattern`, joined by `join`.
-`pattern` can be a regular expression.
+You can specify a regular expression by surrounding the pattern with `/.../`. For example
+`/^guitars?$/` matches the performance type "guitar" or "guitars", but not e.g. "bass guitar".
 
 _Since Picard 0.10_"""
 ))
 def func_performer(parser, pattern="", join=", "):
     values = []
+    try:
+        regex = pattern_as_regex(pattern, allow_wildcards=False, flags=re.IGNORECASE)
+    except re.error:
+        return ''
     for name, value in parser.context.items():
         if name.startswith("performer:"):
             name, performance = name.split(':', 2)
-            try:
-                match = bool(re.search(pattern, performance))
-            except re.error:
-                # fall back to simple string matching if regex is invalid
-                match = pattern in performance
-            if match:
+            if regex.search(performance):
                 values.append(value)
     return join.join(values)
 
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -767,8 +767,10 @@ _Since Picard 0.10_"""
 def func_performer(parser, pattern="", join=", "):
     values = []
     for name, value in parser.context.items():
-        if name.startswith("performer:") and pattern in name:
-            values.append(value)
+        if name.startswith("performer:"):
+            name, performance = name.split(':', 2)
+            if pattern in performance:
+                values.append(value)
     return join.join(values)
 
 
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -761,6 +761,7 @@ def func_lenmulti(parser, multi, separator=MULTI_VALUED_JOINER):
     """`$performer(pattern="",join=", ")`
 
 Returns the performers where the performance type (e.g. "vocal") matches `pattern`, joined by `join`.
+`pattern` can be a regular expression.
 
 _Since Picard 0.10_"""
 ))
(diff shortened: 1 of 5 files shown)


picard/script/functions.py
```diff
@@ -760,14 +760,22 @@ def func_lenmulti(parser, multi, separator=MULTI_VALUED_JOINER):
...
 ))
 def func_performer(parser, pattern="", join=", "):
     values = []
     for name, value in parser.context.items():
-        if name.startswith("performer:") and pattern in name:
-            values.append(value)
+        if name.startswith("performer:"):
+            name, performance = name.split(':', 2)
+            try:
+                match = bool(re.search(pattern, performance))
```
zas, line 773: I wonder about case sensitity here. Is `name` guaranteed to be lowercased? Should we do a case-insensitive match?
Do we expect `$performer(^GUITAR\$)` to match?

phw: Yes, I think we should compare it case insensitive. It's definitely not guaranteed to be lowercase. It usually probably is, as it is usually filled by MB data, but even there is no full guarantee that instruments are allways lowercase.

zas: We should convert whole `name` to lowercase and use case-insensitive regex matching (`re.search` flag) and lowercase `pattern` if the fallback section.

Sophist-UK:
Whilst we are thinking about regular expressions for `$performer` are there any other script functions that would benefit from regular expressions?

zas:
I wonder if we should introduce a new function instead (like `$performer_regex()`). Even though in this case it isn't much a problem I can see issues arising from the fallback mechanisn (matching substring `.*` is very different from matching regex `.*`).

picard/script/functions.py
Sophist-UK: Given that the regex is going to be run on every track, should we cache the compiled regex?

Sophist-UK (commented):
> I wonder if we should introduce a new function instead (like `$performer_regex()`). Even though in this case it isn't much a problem I can see issues arising from the fallback mechanisn (matching substring `.*` is very different from matching regex `.*`).

In other software I use which has a scripting language regex patterns are distinguished from non-regex patterns by regex being of the form `/pattern/`. (To be precise, they allow optional prefix and suffix characters too for e.g. case insensitivity, but providing we are going to do case insensitive by default I don't think we need these - which makes recognising regex patterns easy.) Given that it is highly unlikely that users will have existing patterns in this form, I think this would provide backwards compatibility and avoid needing a separate regex version of the function.

zas:
> In other software I use which has a scripting language regex patterns are distinguished from non-regex patterns by regex being of the form `/pattern/`. 

Actually that would be the best approach, and it opens the possibility to handle regex flags ('/pattern/i').

phw:
> In other software I use which has a scripting language regex patterns are distinguished from non-regex patterns by regex being of the form `/pattern/`.

This would make sense if we'd add this on a language level and having regex as an actual syntax element of the language. This would then also allow us to get rid of double escaping, e.g. `/.*(foo)$/` would be a valid regex without the need to escape the parantheses or the $. That's definitely something that should be considered for a tagger script v2, as it also would simplify many functions. But we cannot easily add this to existing parser, it is a real breaking change. Also delimiter would likely not be a slash, as this is more useful as a directory separator. But maybe something like `#regex#flags`.

For a single function I don't think we should suddenly add this syntax. It would be inconsistent with other functions and would require additional escaping of `/` (e.g. current `$performer(tenor horn / alto horn)` would need to become `$performer(/tenor horn \\/ alto horn/)`.

For the existing `$performer` function I really see no big problem here. For one this is already using pattern matching, so it is expected that this matches a part of the performance type. And current usages will be rather simple, like `$performer(bass guitar)`. And `bass guitar` is a completely valid regex. Even such things like `$performer(tenor horn / alto horn)` would still work, `tenor horn / alto horn` is again a completely valid regex which matches e.g. "tenor horn / alto horn".

There is currently no point why someone would use `$performer(^guitar)` right now, as there is not a single instrument containing a `^`.

For this reason I'd also prefer not to add a separate $rperformer function or such. We already have a lot of functions, that we all need to maintain and keep documented.

phw:
Just thinking aloud: If we would want to extend the current scripting language with regex as a language element something awkward like `\eregex\eflags` would work, or maybe `\e/..../flags` for a more regular syntax. That's because escape sequences provide us a backward compatible way for extending the language as we have seen in the recent unicode escape sequence extension. Any `\e` would be a syntax error currently, so we are save to make use of this.

Having said this, I don't like this syntax very much. But at least `\e/.*(foo)$/i` could be safely parsed, without extra escapes inside the regex being needed and the ability to support regex flags.

Sophist-UK:
> For a single function I don't think we should suddenly add this syntax. It would be inconsistent with other functions and would require additional escaping of `/` (e.g. current `$performer(tenor horn / alto horn)` would need to become `$performer(/tenor horn \\/ alto horn/)`.
> 
> For the existing `$performer` function I really see no big problem here. For one this is already using pattern matching, so it is expected that this matches a part of the performance type. And current usages will be rather simple, like `$performer(bass guitar)`. And `bass guitar` is a completely valid regex. Even such things like `$performer(tenor horn / alto horn)` would still work, `tenor horn / alto horn` is again a completely valid regex which matches e.g. "tenor horn / alto horn".

For backwards compatibility as @zas pointed out we cannot take current patterns as regex by default as e.g. `.` means something completely different. We therefore need some way for a user to explicitly state that they are using a regex.

If we recognise regex strings as `/pattern/` then `/tenor horn / alto horn/` wouldn't require the middle `/` to be escaped.

So `$performance(tenor horn / alto horn)` would be normal search as per current functionality, whilst `$performance(/tenor horn / alto horn/)` would use regex (though of course in this example they would be the same).

Note: There are other potential ways of recognising a regex e.g. must start with a `^` but that would mean e.g.  `$performance(^.*alto horn)` which I think is awkward. 

But I do agree that if we are going to make regex part of the language then \e/pattern/ is a better way (\r being theoretically better than \e but already in use).

phw:
Ok, I have extended this as follows:

1. The existing matching is (function wise) completely unchanged: `$performer(drum)` still matches case sensitive and performance type that contains drum
2. You can explicitly define a regular expression in the form `/pattern/flags`. So you can e.g. do something like `$performer(/^guitars?$/i)` to match both "guitar" and "guitars" case insensitive. The only useful and supported flag is "i" (ignore case). Theoretically also "m" (multiline would be supported, but that actually should have no relevance in this context.

The implementation reuses the code that previously was used in the `TagGenreFilter` class to parse patterns as regex. It was extended and extracted into a utility function `picard.util.pattern_as_regex`.

Technically both the actual regex matching and the full string matching uses a regular expression. That makes `pattern_as_regex` easier to implement and more versatile to handle. `pattern_as_regex` can also support wildcard matching with "*" (e.g. `foo*`), but this is turned off for `$performer` to stay truly backward compatible.

> Whilst we are thinking about regular expressions for `$performer` are there any other script functions that would benefit from regular expressions?

`$unset` comes to mind. It already has a hard coded "wildcard" matching, but only very specific. E.g. you can do `$unset(performer:*)`, but not `$unset(performer:drum*)`.

So this could actually benefit from this implementation, which would both allow extended real wildcard matching (e.g. `$unset(performer:drum*)` and also regex matching (e.g. `$unset(/^performer:.*drum.*/i)`. Or using this with completely different tags, e.g. unsetting all tags ending with "sort" by using `$unset(*sort)`.

Sophist-UK:
Personally I think it extremely unlikely that it would be an issue with $performer becoming case insensitive by default. Ditto for the regex version with a flag to make it case sensitive.

However, I think that compatibility with PCRE probably trumps the regex being case insensitive by default, and if regex is case sensitive then I think non-regex should also stay case sensitive.

zas (approved):
Much better with `/regex/flags` support, it will be easier to add this to other functions.
LGTM

PR #1832, opened 2021-06-03
Refactor File._guess_tracknumber_and_title() and add tests

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

Make `tracknum_and_title_from_filename()` returns a named tuple and use Python `sets` features to find out which field is missing.
Add related tests.

# Problem

# Solution

# Action

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -410,6 +410,9 @@ def tracknum_from_filename(base_filename):
     return None
 
 
+GuessedFromFilename = namedtuple('GuessedFromFilename', ('tracknumber', 'title'))
+
+
 def tracknum_and_title_from_filename(base_filename):
     """Guess tracknumber and title from filename.
     Uses `tracknum_from_filename` to guess the tracknumber. The filename is used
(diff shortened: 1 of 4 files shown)


picard/util/__init__.py
```diff
@@ -426,8 +426,8 @@ def tracknum_and_title_from_filename(base_filename):
         tnlen = len(tracknumber)
         if stripped_filename[:tnlen] == tracknumber:
             title = stripped_filename[tnlen:].lstrip()
-
-    return (tracknumber, title)
+    GuessedFromFilename = namedtuple('GuessedFromFilename', ('tracknumber', 'title'))
```
phw, line 429: Should this named toupee rather be defined outside of the function? It seems odd the redefine it on every call

phw (changes requested):
Good refactoring!

picard/util/__init__.py
zas: Yes, you're right, done in dc0aeb8

phw (approved):

PR #1833, opened 2021-06-03
Refactor script editor

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Refactor the Script Editor to correct issue with window display on mac systems.  Also correct issue with logic when saving changes to script metadata.

# Problem

The Script Editor object was being created before the Options > File Naming screen was being displayed, causing the Script Editor window to open behind the Options window on mac systems.  Also, saving script metadata always resulted in an "Unsaved Changes" dialog.

# Solution

Refactor the Script Editor to correct issue with window display on mac systems.

Correct issue with logic when saving changes to script metadata.

Miscellaneous changes to reduce code duplication.

# Action

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -1135,20 +1135,17 @@ class ScriptDetailsEditor(PicardDialog):
             ).exec_()
             return
         if self.has_changed():
-            if self.change_check():
-                if not self.ui.script_last_updated.isModified() or not self.ui.script_last_updated.text().strip():
-                    self.set_last_updated()
-                self.script_item.update_script_setting(
-                    title=self.ui.script_title.text().strip(),
-                    author=self.ui.script_author.text().strip(),
-                    version=self.ui.script_version.text().strip(),
-                    license=self.ui.script_license.text().strip(),
-                    description=self.ui.script_description.toPlainText().strip(),
-                    last_updated=self.ui.script_last_updated.text().strip()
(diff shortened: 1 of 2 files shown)


picard/ui/scripteditor.py
```diff
@@ -254,6 +298,82 @@ def confirmation_dialog(parent, message):
...
+
+    Args:
+        naming_scripts (list): List of available user-defined naming scripts in YAML format
+        selected_script_id (str): ID code for the currently selected script
+        combo_box (QComboBox): Combo box object to populate
+
+    Returns:
+        int: The index of the currently selected script
+    """
+    SCRIPT_TITLE_USER = _("User: %s")
```
zas, line 336: We should get rid of `SCRIPT_TITLE_USER` and `ScriptEditorDialog.SCRIPT_TITLE_USER` and just create a new method at module level:

```python
def script_title_user(title):
    return _("User: %s") % title
```

Then replace all occurences like:

```python
SCRIPT_TITLE_USER % script_item["title"]
```

with:

```python
script_title_user(script_item["title"])
```

zas (changes requested):
Good cleanup.

zas (approved):
Lgtm

phw (approved):

PR #1834, opened 2021-06-04
PICARD-2226: Removed config object per thread again

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2226

Syncing between those instances does not work [as advertised](https://doc.qt.io/qt-5/qsettings.html#accessing-settings-from-multiple-threads-or-processes-simultaneously), which causes missing updates on config changes where a config change is not reflected in tags.

# Solution

Get rid of the multiple config instances, one per thread.

We added this in the hope it would resolve our deadlock issue. But the root cause was different, and is both fixed by recent PyQt5 and as well by our locking workaround at https://github.com/metabrainz/picard/blob/master/picard/config.py#L164-L186 , which makes sure this does not happen even with older versions.

While https://doc.qt.io/qt-5/qsettings.html#accessing-settings-from-multiple-threads-or-processes-simultaneously claims that one should use one tread per instance we otherwise never had issues with this until the deadlock issue came up in 2.0.

# Action

I consider this a hotfix, that we can quickly roll out without too much code changes. Ideally we want to refactor the config module completely and get rid of QSettings, which has been a time sink with all the troubles it was causing.

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -362,9 +362,6 @@ config = None
 setting = None
 persist = None
 
-_thread_configs = {}
-_thread_config_lock = threading.RLock()
-
 
 def setup_config(app, filename=None):
     global config, setting, persist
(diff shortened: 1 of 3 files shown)


phw:
Works well so far in my testing

zas (approved):
It solves the problem for me.

PR #1835, opened 2021-06-05
PICARD-9: Add multiple user profiles

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Adds the ability to have multiple user profiles and switch between them quickly.

# Problem

* JIRA ticket (_optional_): PICARD-9

Users have requested the ability to have multiple user profiles so that they can quickly and easily change multiple settings.

# Solution

This change adds the ability to maintain and select from multiple profiles.  Each profile holds a copy of the `config.setting.as_dict()` output, and switching profiles replaces the current `config.setting` with the stored value.  Any changes made to the configuration settings are only applied to the currently selected profile, which is updated when the settings are updated.

The changes include:

* Addition of a new `UserProfile` class.
* Moving the list of scripts from `config.setting` to `config.persist` to make them available to all profiles.
* Add logic to block deleting a file naming script if it is used in a profile other than the current profile.
* Add "user_profiles" and "selected_user_profile" items in the `options.persist` section to store the available profiles and which one is currently selected.
* Add a profile management dialog that allows adding and removing profiles, editing profile titles and descriptions (notes), and selecting which profile is currently active.  This is available from the main screen under the Options item on the menu bar.
* Refactor the confirmation dialog function into the `picard.ui` module to reduce code duplication in the script editor and profile editor modules.
* Added a placeholder url to the documentation which is still under development.

Note that any testing should be done using a backup copy of your configuration file because the changes made are not backwards compatible.

# Action

Saving a copy of the full settings in each profile may be a bit of overkill and can use a lot of storage, but it is less complicated for the users (in my opinion).  This could be changed to a fixed list of selected settings, or even to a user-selected list of specific settings if that is the preferred method.

### DIFF
--- a/picard/config_upgrade.py
+++ b/picard/config_upgrade.py
@@ -34,6 +34,7 @@ from picard import log
 from picard.config import (
     BoolOption,
     IntOption,
+    ListOption,
     Option,
     TextOption,
 )
@@ -386,6 +387,23 @@ def upgrade_to_v2_7_0_dev_2(config):
     )
 
 
+def upgrade_to_v2_7_0_dev_3(config):
+    from picard.profile import UserProfile
+    _s = config.setting
+    _p = config.persist
+    ListOption("setting", "file_naming_scripts", [])
+    ListOption("persist", "file_naming_scripts", [])
+    Option("persist", "user_profiles", {})
+    TextOption("persist", "selected_user_profile", "")
+    # Migrate naming scripts to new persist key
+    _p["file_naming_scripts"] = _s["file_naming_scripts"]
+    _s.remove("file_naming_scripts")
+    # Create initial user profile
+    profile = UserProfile(title=_("Initial User Profile"))
+    _p["selected_user_profile"] = profile["id"]
+    _p["user_profiles"] = {profile["id"]: profile.to_dict()}
+
+
 def rename_option(config, old_opt, new_opt, option_type, default):
     _s = config.setting
     if old_opt in _s:
@@ -416,4 +434,5 @@ def upgrade_config(config):
     cfg.register_upgrade_hook(upgrade_to_v2_6_0_beta_2)
     cfg.register_upgrade_hook(upgrade_to_v2_6_0_beta_3)
     cfg.register_upgrade_hook(upgrade_to_v2_7_0_dev_2)
+    cfg.register_upgrade_hook(upgrade_to_v2_7_0_dev_3)
     cfg.run_upgrade_hooks(log.debug)
--- a/picard/profile.py
+++ b/picard/profile.py
@@ -1,0 +1,141 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2021 Bob Swift
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+
+from copy import deepcopy
+import uuid
+
+from picard.config import get_config
+
+
+class ProfileImportError(Exception):
+    def __init__(self, *args):
+        super().__init__(*args)
+
+
+class UserProfile():
+    """Base class for Picard user profile objects.
+    """
+
+    OUTPUT_FIELDS = ('title', 'description', 'id', 'settings')
+    DEFAULT_TITLE = N_("User Profile")
+
+    def __init__(
+        self,
+        title='',
+        description='',
+        id=None,
+        settings_dict=None
+    ):
+        """Base class for Picard user profile objects
+
+        Args:
+            title (str): Title of the profile.
+            description (str): Description of the profile.
+            id (str): ID code for the script. Defaults to a system generated uuid.
+            settings_dict (dict): Dictionary of settings as produced using config.setting.as_dict()
+        """
+        self.title = title if title else _(self.DEFAULT_TITLE)
+        self.description = description
+        if not id:
+            self._set_new_id()
+        else:
+            self.id = id
+        self.settings = None
+        self.load_settings(settings_dict)
+
+    def _set_new_id(self):
+        """Sets the ID of the script to a new system generated uuid.
+        """
+        self.id = str(uuid.uuid4())
+
+    def __getitem__(self, setting):
+        if hasattr(self, setting):
+            value = getattr(self, setting)
+            if isinstance(value, str):
+                value = value.strip()
+            return value
+        return None
+
+    def __setitem__(self, setting, value):
+        if hasattr(self, setting):
+            if isinstance(value, str):
+                value = value.strip()
+            setattr(self, setting, value)
+
+    def load_settings(self, settings):
+        """Load settings from a dictionary.  If the dictionary provided is None or empty, uses the current config.setting.as_dict().
+
+        Args:
+            settings (dict): Dictionary of settings as produced using config.setting.as_dict()
+        """
+        if settings and isinstance(settings, dict):
+            self.settings = settings
+        else:
+            config = get_config()
+            self.settings = config.setting.as_dict()
+
+    def settings_to_config(self):
+        """Updates the current config.setting with the profile object's stored settings.
+        """
+        config = get_config()
+        for key, value in self.settings.items():
+            if key in config.setting:
+                config.setting[key] = value
+
+    def copy(self):
+        """Create a copy of the current profile object with updated title.
+        """
+        new_object = deepcopy(self)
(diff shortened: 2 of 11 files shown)


zas (changes requested):
Are you aware of https://github.com/metabrainz/picard/blob/fe5970f25a7baeac7291b53872ebeff5d6b7979d/picard/config.py#L224-L230 ?

It is there since 15 years or more, and unused since, but basically it was made to allow switching profiles ;)

picard/profile.py
```diff
@@ -0,0 +1,141 @@
...
+
+from copy import deepcopy
+import uuid
+
+from picard.config import get_config
+
+
+class ProfileImportError(Exception):
+    def __init__(self, *args):
+        super().__init__(*args)
```
zas, line 30: You may just `pass` here.

```python
class ProfileImportError(Exception):
    pass
```

picard/ui/mainwindow.py
```diff
@@ -737,14 +742,30 @@ def _update_cd_lookup_button(self):
     def toggle_rename_files(self, checked):
         config = get_config()
         config.setting["rename_files"] = checked
+        self.update_settings_in_profile()
 
     def toggle_move_files(self, checked):
         config = get_config()
         config.setting["move_files"] = checked
+        self.update_settings_in_profile()
```
zas, line 750: I'm not fond of this mechanism. Does this mean it has to be called everytime the value is changed to be stored in profile?
Actually not even sure to understand this part.

picard/ui/options/dialog.py
```diff
@@ -208,6 +208,14 @@ def accept(self):
...
                 self._show_page_error(page, e)
                 return
+        config = get_config()
+        selected_user_profile = config.persist["selected_user_profile"]
+        profiles = config.persist["user_profiles"]
+        if selected_user_profile in profiles:
+            profiles[selected_user_profile]['settings'] = config.setting.as_dict()
+            config.persist["user_profiles"] = profiles
+        else:
+            log.error('Unable to update user profile "%s"', selected_user_profile)
```
zas, line 218: A lot of code in common with `update_settings_in_profile()` above.

zas (changes requested):

picard/config_upgrade.py
```diff
@@ -386,6 +387,23 @@ def upgrade_persisted_splitter(new_persist_key, key_map):
     )
 
 
+def upgrade_to_v2_7_0_dev_3(config):
+    from picard.profile import UserProfile
+    _s = config.setting
+    _p = config.persist
+    ListOption("setting", "file_naming_scripts", [])
+    ListOption("persist", "file_naming_scripts", [])
```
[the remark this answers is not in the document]
phw: Yes, any actual user setting should never go into "persist"

phw:
I am not to sure about this really. I actually don't think it is desirable to have all settings being part of the profile. There is a lot about settings where I don't think users are commonly switching them over when tagging different files. If everything is part of the profile, it means also the user needs to update every setting in all of their profiles. I think this ads a complexity that gets hard to handle and makes this feature less useful. E.g. changing UI colors or networking settings. "Scripts" section I also find problematic, as I don't want to duplicate all my scripts for all profiles I might have. But maybe I want to change the scripts that are active per profile!? What most users want from this feature is being able to quickly switch between different configurations depending on what files they tag.

So IMHO this feature should be limited to settings that are actually tagging relevant. Of course now this starts the discussion, and there are many cases where people will disagree. I think a clear candidate are the "File Naming" options. Likely also "Tags". The nice thing about both of them is that they apply mostly on file saving, so they can be easily changed while data is already loaded. Now people might also like to have "Metadata" and "Cover Art" included, but this applies mostly on loading, so switching profiles has no immediate effect on already loaded data. For "Scripts" see my comment above.

If we have only parts of options included in profiles this also needs some restructuring in options dialog I think. Maybe have a tagging profiles section, where one manages the tagging profiles and below it all the sub sections that are actually relevant here.

The second thing is that I think the existing mechanism that was prepared for this, but never actually used and not in a working state, would work rather well if finalized. The basic idea is that in addition to the existing setting and persist sections there is a third config section called profile. Thus one can define an option to be stored in the profile:

```
BoolOption("profile", "move_files", False),
```

That way one can have specific settings that are loaded from the profile, but others that are from setting.

But the actual profile can be changed. By default a "default" profile is used. But additional profiles could be added, and they can be changed with `config.switchProfile` (which should be renamed to `config.switch_profile` btw.).

The switching I think once work, but broke as `ConfigSection` has no public attribute called `name` anymore, but that could be easily fixed. What is totally missing are methods to add, list and remove profiles, these would need to be added.

rdswift:
Thanks for the comments.  It appears painfully obvious to me that I've set out on the wrong track here, so I'll cancel this PR because I don't think that there's much of it that can be reused in the final solution.

Don't get me wrong, I see the merit in everything said.  What I'm having difficulty picturing is how to allow the user to determine  which override settings are stored in the profile, and how to add, remove and edit these selections in each individual profile.  In addition to the obvious candidates such as file naming, tags, metadata and cover art, I believe that it would also be beneficial to allow storing overrides indicating which tagging scripts are active and (possibly) which plugins are enabled.  Right now, I just don't know how that UI would look (or work).

rdswift:
> Are you aware of
> 
> https://github.com/metabrainz/picard/blob/fe5970f25a7baeac7291b53872ebeff5d6b7979d/picard/config.py#L224-L230
> 
> ?
> It is there since 15 years or more, and unused since, but basically it was made to allow switching profiles ;)

I saw that, but I couldn't find where it actually did anything.

zas:
> > Are you aware of
> > https://github.com/metabrainz/picard/blob/fe5970f25a7baeac7291b53872ebeff5d6b7979d/picard/config.py#L224-L230
> > 
> > ?
> > It is there since 15 years or more, and unused since, but basically it was made to allow switching profiles ;)
> 
> I saw that, but I couldn't find where it actually did anything.

Well, it was originally meant to handle profiles, but was never used.
I guess we can remove this dead code, if you don't use it.

rdswift:
I just assumed that I wasn't seeing it, especially when @phw said that he thought it once worked.  It wouldn't be the first time I missed something because I didn't follow the code back far enough.

I do like the concept, although I wonder if it would be easier to store the individual profiles as dictionary items in a single ConfigSection (only because I don't know the implications of adding and removing ConfigSection items from the Config object).  If ConfigSection items can be easily added and removed, then that may very well be the best approach to use.

The other thing that I'm having trouble conceptualizing is how to get the profile-specific overridden values when actually processing, but not when editing the (base) Options settings.  I don't see how the current `config.setting['whatever']` could do it, unless the processing called some sort of helper function instead for retrieving the values (which actually might be pretty easy to implement, other than deciding how to handle file naming examples -- whether or not to use the overrides).

rdswift:
Before I start doing any coding, I want to run this past you to see if this is more along the lines of what you would like to see...

Each profile would be stored in a dictionary containing the human-readable `name` of the profile (used for the combo box for selecting a profile), possibly the unique id for the profile, and a dictionary of the `overrides` that will will be used rather than the values from `config.settings`.  These profile dictionaries could be stored in a `config.profile['profiles']` dictionary with the key being the system-generated uuid identifier for the profile.  Alternately, they could be stored as separate ConfigSection items if someone can help me understand how that's supposed to work.

The profile editor dialog would include a combo box to select which profile is being edited, along with a text box for editing the profile's name (and possibly a text box to provide a description / notes for the profile).  Below that would be a page (or tab) for each Options section whose items could be selected for overrides.  Each item on each page or tab would be preceded by a check box to enable the override for that item only.  Items not checked would be disabled (or read-only) on the page or tab, and would not be saved to the `overrides` dictionary for the profile.  This way the size of the `overrides` dictionary is held to a minimum, with no extra (unused) information stored.

Initially, I'm suggesting that the settings pages or tabs include:
* Metadata (main page items only -- not Preferred Releases, Genres or Ratings)
* Tags (collapsed to a single page or tab)
* Cover Art (main page items only, and perhaps not even the cover art providers section)
* File Naming (including the script selector, but without the script editor or naming examples)

We might also consider including Tagger Scripts, but only to indicate whether they are enabled in general and only which scripts are enabled -- no reordering or editing functionality.

The profile editor would provide buttons to:
1. Add a new blank profile (with all override items unchecked).
2. Create a copy of the currently displayed profile as a new profile.
3. Delete the currently displayed profile (with confirmation).
4. Revert the currently displayed profile to its last saved state (with confirmation).
5. Save the profile.
6. Cancel / close the editor (confirming the loss of any unsaved changes if applicable).
7. Open a help url in the user's browser.

Similar to the script editor, this may be an overwhelming number of buttons so a menu bar may be appropriate.

The "default" profile (with no overrides enabled -- empty `overrides` dictionary) would not be displayed for editing or deletion.  (In fact, we may not even need to save this as a profile.)

There would be an item added to the menu bar on the main screen to open the profile editor (because I think we may want to keep this separate from the Options settings dialog) and an item to select a profile (including an option for "default") to use.  We may want to consider some visible indicator, such as an addition to the main window title bar, to indicate that there is a profile that is being applied.

Finally, there would be a new helper function to provide the settings to use for the actual file processing, automatically applying the override values if available from the selected profile (other than 'default' which would provide the values directly from `config.setting` with no overrides applied).

So what do you think?  Please feel free to pick this apart, because I'm sure that you've been thinking about this for a lot longer than I, and you have a better sense of what the future plans are for Picard.  I just want to get some sense of whether this is more along the right track before jumping into writing any significant amount of code.  Thanks.

zas:
> hould I move it to a comment on the [PICARD-9](https://tickets.metabrainz.org/browse/PICARD-9) ticket instead?

Yes, please move it to ticket ;)

PR #1836, opened 2021-06-09
Fix unsaved changes confirmation check when changing selected script

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Correct the logic for checking for unsaved changes when selecting a new script in the script editor

# Problem

There is an error in the logic that caused the unsaved changes check from being performed when switching the selected script in the combo box.

# Solution

Correct the logic so that the check is properly completed.

# Action


phw (approved):
Thanks for the fix

rdswift:
> Thanks for the fix

It's only fair, since it was my mistake in the first place.  :blush:

PR #1838, opened 2021-06-13
PICARD-2231: Build with Python 3.8 on Windows

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2231, PICARD-2230

Try to build the Windows packages with Python 3.8 again. This caused issues in PICARD-2067, but either those are now resolved or we should investigate deeper.

Not using Python 3.9, as this would drop Win 7 support. For now we can maintain Win 7 support, at least as long as there is a supported Python version and we have no other requirement forcing us to update. We still seem to have quite some users on Win 7.

# Solution

# Action

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -121,10 +121,10 @@ jobs:
       with:
         fetch-depth: 0  # Fetch entire history, needed for setting the build number
     - run: git fetch --depth=1 origin +refs/tags/release-*:refs/tags/release-*
-    - name: Set up Python 3.7
+    - name: Set up Python 3.8
       uses: actions/setup-python@v1
       with:
-        python-version: 3.7
+        python-version: 3.8
     - name: Setup Windows build environment
       run: |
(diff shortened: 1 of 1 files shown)


phw:
Works for me, both the installer and portable build run on Windows 10. The Python version included is 3.8.10, which is exactly what we need.

@rdswift Could you check the build at https://github.com/metabrainz/picard/actions/runs/933615461 and see if this works on your Windows 7?

rdswift:
I tried the portable version and it seems to work just fine on my Windows 7 system.  I assume that the install version should also work, so I haven't tried it.

phw:
Thanks for testing. I merge this and we can get wider testing.

But could you test the installer anyway? I think it should work this time, but it can behave differently then the installed version in regards to startup as it uses a bit different bootstrapping code.

rdswift:
Just tested the install version and everything seemed to work fine, other than some oddities regarding the script editor (but I don't think those have anything to do with your changes).  Well done.

PR #1841, opened 2021-06-14
PICARD-2230: Do not crash if browser integration fails to start or stop

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2230

An unexpected, arbitrary error in browser integration such as reported in PICARD-2230 would cause Picard not to start.

The actual core issue of PICARD-2230 is resolved by updating Python to version 3.8.10 for our Windows builds, which fixes https://bugs.python.org/issue26227.

But the general issue of bugs in browser integration will prevent Picard to start should also be addressed. After all we are dealing with setting up a local web server here, and there is all kind of stuff out of our control that can go wrong here.

# Solution

# Action

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -112,26 +112,32 @@ class BrowserIntegration(QtCore.QObject):
         else:
             host_address = '0.0.0.0'  # nosec
 
-        for port in range(config.setting["browser_integration_port"], 65535):
-            try:
-                self.server = ThreadingHTTPServer((host_address, port), RequestHandler)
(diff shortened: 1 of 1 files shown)


phw:
Or would this exception handling better be placed in https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L399-L404 ?

zas (approved):

PR #1843, opened 2021-06-14
PICARD-2232: Allow add cluster/files and disc ID submission to unofficial servers

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2232, PICARD-1645

The new add cluster / files functionality added in #1839 always submits to main servers, regardless of what server has been configured. Likewise disc ID submission always submits to musicbrainz.org as well, see PICARD-1645.

In the majority of cases that is the wanted behavior. But sometimes they might want to submit to their local mirror, either for testing/debugging or because they actually want the submissions only to have locally.

# Solution

The default behavior is that the main database is being used. As long as the user has configured the official servers (musicbrainz.org or beta.musicbrainz.org) those are used for submission of addrelease and discid as well.

grafik

But if the user enters an unofficial server host a warning is displayed:

grafik

By default it shows the warning and for submission the primary server (musicbrainz.org) will be used. But the can enable "Submit data using the configured server" and addrelease / discid submission will use the custom server.

# Action

This only handles data submission for the addrelease functionality and disc ID. Both of which are done via opening an URL in browser.

I guess we also should support other kind of data, but I'm not sure. Currently there is submission of ratings and managing collections. Theoretically we could extend to submission of tags / genres, ISRC and barcode. Some probably should go to official server. Not sure about personal data like collections and ratings.

Anyway, implementing this requires some more work because of authentication. Basically if the user uses an unofficial server but submission is done to primary server we need to maintain two separate logins, one for primary server for submission and one for querying the custom server.

### DIFF
--- a/picard/util/mbserver.py
+++ b/picard/util/mbserver.py
@@ -18,9 +18,14 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from collections import namedtuple
+
 from picard.config import get_config
 from picard.const import MUSICBRAINZ_SERVERS
 from picard.util import build_qurl
+
+
+ServerTuple = namedtuple('ServerTuple', ('host', 'port'))
 
 
 def is_official_server(host):
@@ -46,12 +51,12 @@ def get_submission_server():
     config = get_config()
     host = config.setting['server_host']
     if is_official_server(host):
-        return (host, 443)
+        return ServerTuple(host, 443)
     elif host and config.setting['use_server_for_submission']:
         port = config.setting['server_port']
-        return (host, port)
+        return ServerTuple(host, port)
     else:
-        return (MUSICBRAINZ_SERVERS[0], 443)
+        return ServerTuple(MUSICBRAINZ_SERVERS[0], 443)
 
 
 def build_submission_url(path=None, query_args=None):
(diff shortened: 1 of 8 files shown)


picard/util/mbserver.py
```diff
@@ -0,0 +1,68 @@
...
+
+    Args:
+        host: the hostname
+
+    Returns: True, if host is an official MusicBrainz server, False otherwise
+    """
+    return host in MUSICBRAINZ_SERVERS
+
+
+def get_submission_host():
```
zas, line 37: Since it returns a tuple (host, port), we could rename it `get_submission_server()` and use an named tuple.

```python
ServerTuple = namedtuple('ServerTuple', ('host', 'port'))
```

```python
        return ServerTuple(MUSICBRAINZ_SERVERS[0], 443)
```

```python
server = get_submission_server()
do_something(server.host, server.port)
```

zas (approved):
LGTM
Just a minor stylistic comment, but it can be done after this is merged anyway.

picard/util/mbserver.py
zas: I prepared a patch doing that: https://github.com/phw/picard/pull/10

zas (approved):
LGTM

PR #1845, opened 2021-06-15
Add naming script selector to main window menu bar

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

There is no quick and easy way to select a file naming script to use from the main menu.

# Solution

Add a new menu item under the "Options" section to select the file naming script to use.  The available scripts are listed as sub-menu items, and the currently selected script is indicated with a check mark.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -83,6 +83,7 @@ from picard.const.sys import (
 from picard.file import File
 from picard.formats import supported_formats
 from picard.plugin import ExtensionPoint
+from picard.script import get_file_naming_script_presets
 from picard.track import Track
 from picard.util import (
     icontheme,
@@ -120,6 +121,7 @@ from picard.ui.passworddialog import (
     PasswordDialog,
     ProxyDialog,
 )
+from picard.ui.scripteditor import user_script_title
 from picard.ui.searchdialog.album import AlbumSearchDialog
 from picard.ui.searchdialog.track import TrackSearchDialog
 from picard.ui.statusindicator import DesktopStatusIndicator
@@ -182,6 +184,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         TextOption("persist", "current_directory", ""),
         FloatOption("persist", "mediaplayer_playback_rate", 1.0),
         IntOption("persist", "mediaplayer_volume", 50),
+        Option("setting", "file_renaming_scripts", {}),
+        TextOption("setting", "selected_file_naming_script_id", ""),
     ]
 
     def __init__(self, parent=None, disable_player=False):
@@ -815,6 +819,12 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         menu.addAction(self.enable_renaming_action)
         menu.addAction(self.enable_moving_action)
         menu.addAction(self.enable_tag_saving_action)
+
+        self.script_quick_selector_menu = QtWidgets.QMenu(_("&Select file naming script"))
+        self.script_quick_selector_menu.setIcon(icontheme.lookup('document-open'))
+        self.make_script_selector_menu()
+
+        menu.addMenu(self.script_quick_selector_menu)
         menu.addSeparator()
         menu.addAction(self.options_action)
         menu = self.menuBar().addMenu(_("&Tools"))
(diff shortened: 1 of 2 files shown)


picard/ui/mainwindow.py
```diff
@@ -182,6 +184,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         TextOption("persist", "current_directory", ""),
         FloatOption("persist", "mediaplayer_playback_rate", 1.0),
         IntOption("persist", "mediaplayer_volume", 50),
+        Option("setting", "file_renaming_scripts", {}),
+        TextOption("setting", "selected_file_naming_script_id", ""),
```
phw, line 188: These options don't need to be redefined here. Actually they should be only defined once. Currently they are also duplicated in options/renaming.py and in scripteditor.py.

The options usually are defined at the place they get set, so usually in the options. I think here maybe the scripteditor.py is the most fitting place. The only requirement is that this file gets imported on start, so that the option is registered.

picard/ui/mainwindow.py
```diff
@@ -1480,6 +1490,32 @@ def check_for_update(self, show_always):
...
+        config = get_config()
+        naming_scripts = config.setting["file_renaming_scripts"]
+        selected_script_id = config.setting["selected_file_naming_script_id"]
+        self.script_quick_selector_menu.clear()
+
+        def _add_menu_item(title, id):
+            script_action = QtWidgets.QAction(title, self.script_quick_selector_menu)
+            script_action.triggered.connect(partial(self.select_new_naming_script, id))
+            if id == selected_script_id:
+                script_action.setIcon(self.style().standardIcon(QtWidgets.QStyle.SP_DialogApplyButton))
```
phw, line 1503: I wonder if it would be better to either do

```python
script_action.setCheckable(True)
script_action.setChecked(id == selected_script_id)
```

Or if we don't want to have every item with an empty checkbox then only:

```python
if id == selected_script_id:
    script_action.setCheckable(True)
    script_action.setChecked(True)
```

The icon works too, I guess. But maybe we need to check how this looks across platforms. I think it can even happen that menu icons get disabled by the platform, then it would not show up at all.
```

File: picard/ui/mainwindow.py [hunk 3]
```diff
@@ -1480,6 +1490,32 @@ def check_for_update(self, show_always):
...
+            self.script_quick_selector_menu.addAction(script_action)
+
+        for (id, naming_script) in sorted(naming_scripts.items(), key=lambda item: item[1]['title']):
+            _add_menu_item(user_script_title(naming_script['title']), id)
+
+        # Add preset scripts not provided in the user-defined scripts list.
+        for script_item in get_file_naming_script_presets():
+            _add_menu_item(user_script_title(script_item['title']), script_item['id'])
```
phw, line 1511: The presets show up for me as "User: Preset: The title". But in the combobox in options it works as expected and they only display as "Preset: The title".

phw (changes requested):
Thanks. Tested, functionality wise this works well, just some smaller code changes.

picard/ui/mainwindow.py
phw: The existing duplicates where cleaned up by zas in https://github.com/metabrainz/picard/pull/1847

So just removing the above two lines from this PR should be enough.

zas (commented):
It would perhaps be convenient to add a direct access to script editor in this menu (after a separator). What do you think?

Apart that (and phw comments), good move.

phw:
> It would perhaps be convenient to add a direct access to script editor in this menu (after a separator). What do you think?

rdswift already suggested this yesterday. So in general we should definitely have this, requirement is that this works well with window focus and doesn't conflict with the way this is handled when opening from options. I'd say we should do this, but definitely in a separate PR focusing just on this.

In regards to the location I would maybe even place it directly in the Options menu, next to the "Select file naming script" entry.

rdswift:
> I'd say we should do this, but definitely in a separate PR focusing just on this.

Agreed.  I've already started looking at it, but will only push the fixes for the other comments on this PR and start a new PR for the rest of the work.

picard/ui/mainwindow.py
rdswift: I rebased on the updated `master` branch just in case.

picard/ui/mainwindow.py
rdswift: I tried it with the check boxes on all lines and actually like that better than any of the other options I've explored.  Good suggestion.

picard/ui/mainwindow.py
rdswift: Definitely a copy'n'paste error.  Oops.  :blush:

picard/ui/mainwindow.py
```diff
@@ -1497,8 +1497,8 @@ def make_script_selector_menu(self):
         def _add_menu_item(title, id):
             script_action = QtWidgets.QAction(title, self.script_quick_selector_menu)
             script_action.triggered.connect(partial(self.select_new_naming_script, id))
-            if id == selected_script_id:
-                script_action.setIcon(self.style().standardIcon(QtWidgets.QStyle.SP_DialogApplyButton))
+            script_action.setCheckable(True)
+            script_action.setChecked(id == selected_script_id)
```
zas, line 1501: Actually since those are exclusive, it should be radio buttons

See https://github.com/rdswift/picard/pull/8

zas (changes requested):

picard/ui/mainwindow.py
rdswift: Much better looking as well.  Thanks.

phw (approved):

zas (approved):

PR #1847, opened 2021-06-16
Remove multiple declarations of options

zas:

### DIFF
--- a/picard/ui/options/general.py
+++ b/picard/ui/options/general.py
@@ -62,7 +62,6 @@ class GeneralOptionsPage(OptionsPage):
         TextOption("setting", "server_host", MUSICBRAINZ_SERVERS[0]),
         IntOption("setting", "server_port", 443),
         BoolOption("setting", "use_server_for_submission", False),
-        TextOption("persist", "oauth_refresh_token", ""),
         BoolOption("setting", "analyze_new_files", False),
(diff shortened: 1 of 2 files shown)


picard/ui/options/general.py
```diff
@@ -62,12 +62,10 @@ class GeneralOptionsPage(OptionsPage):
         TextOption("setting", "server_host", MUSICBRAINZ_SERVERS[0]),
         IntOption("setting", "server_port", 443),
         BoolOption("setting", "use_server_for_submission", False),
-        TextOption("persist", "oauth_refresh_token", ""),
         BoolOption("setting", "analyze_new_files", False),
         BoolOption("setting", "ignore_file_mbids", False),
         TextOption("persist", "oauth_refresh_token", ""),
         TextOption("persist", "oauth_refresh_token_scopes", ""),
-        TextOption("persist", "oauth_access_token", ""),
```
phw, line 70: I don't see a second declaration of oauth_access_token, is it not needed?

phw (changes requested):

picard/ui/options/general.py
zas: > I don't see a second declaration of oauth_access_token, is it not needed?

No idea why this was committed, I guess lack of coffee. Fixed, forced pushed.

phw (approved):

PR #1849, opened 2021-06-16
Show script editor from main screen menu

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Enable opening the file naming script editor from the main window.

# Problem

The file naming script editor could only be opened from the "File Naming" section of the "Options" ==> "Options..." screen.  This should also be available directly from Picard's main screen to accommodate quick changes to go along with the ability to quickly select the file naming script to be used.

# Solution

Provide a new "Options" ==> "Open script editor..." menu item.

# Action

In order to avoid possible issues with multiple copies of the script editor open simultaneously (one from the main screen and a second from the "File Naming" section of the "Options" ==> "Options..." screen), when it is opened from the main screen, the command to open the "Options" ==> "Options..." screen is disabled when the script editor is open.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -29,7 +29,6 @@
 # Copyright (C) 2019 Timur Enikeev
 # Copyright (C) 2020 Gabriel Ferreira
 # Copyright (C) 2021 Petit Minion
-# Copyright (C) 2021 Bob Swift
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -121,7 +120,11 @@ from picard.ui.passworddialog import (
     PasswordDialog,
     ProxyDialog,
 )
-from picard.ui.scripteditor import user_script_title
+from picard.ui.scripteditor import (
+    ScriptEditorDialog,
+    ScriptEditorExamples,
+    user_script_title,
+)
 from picard.ui.searchdialog.album import AlbumSearchDialog
 from picard.ui.searchdialog.track import TrackSearchDialog
 from picard.ui.statusindicator import DesktopStatusIndicator
@@ -202,6 +205,10 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             if player.available:
                 self.player = player
                 self.player.error.connect(self._on_player_error)
+
+        self.script_editor_dialog = None
+        self.examples = None
+
         self.setupUi()
 
     def setupUi(self):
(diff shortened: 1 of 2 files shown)


rdswift:
@zas, I see that opening the script editor from the main screen, closing it and then opening it from the "File Naming" options screen is generating an error message in the logs:

```
E: 12:46:34,369 config.__init__:333: Option persist/splitters_ScriptEditorDialog already declared
at .\picard\ui\__init__.py:66: in __init__
        Option("persist", self.splitters_name(), {})
```

Any suggestions?

Thinking out loud, I wonder if the method that sets this option for the splitters should first check to see that the option doesn't already exist before setting it (since they are set automatically).

zas:
> Any suggestions?

Not really. I saw those, not sure how to fix.
@phw?

phw:
> I wonder if the method that sets this option for the splitters should first check to see that the option doesn't already exist before setting it (since they are set automatically).

Yes, that's because those options are declared dynamically inside a function, so they redeclare whenever the function runs again. I think this is the only place where we have that, usually options are defined statically when a module gets imported.

I think yes, we need to check if the options exist and only declare them if the don't. Probably we should have some helper function to do so.

rdswift:
> I think yes, we need to check if the options exist and only declare them if the don't. Probably we should have some helper function to do so.

I have a change ready.  Will submit a PR.

zas (approved):

phw (approved):
I like the solution of disabling the options. That allows us to gave the editor still non-modal without having to deal with the details on how to make the options non-modal (and all the additional complexity this would bring) for now.

It might be some users stumble over this, though. Maybe we can show a tooltip (is that possible in a menu?) or other kind of notice on the disabled menu. But for now let's go with this.

rdswift:
> It might be some users stumble over this, though. Maybe we can show a tooltip (is that possible in a menu?) or other kind of notice on the disabled menu. But for now let's go with this.

An alternative would be to leave the options enabled but intercept the call to check if the script editor was open, and display a messagebox explaining that they need to close the script editor before they can open the "Options..." settings.  That might be clearer for the user than a possibly elusive tooltip.

phw:
Indeed, that would be an alternative. Or we just leave it as is and try to solve this better directly. Maybe it would be possible to turn this into a singleton dialog and still make sure it display correctly even if one opens the script editor first, then the options. Maybe it needs some tricks like hiding the dialog and showing it again with different parent or such.

Or we actually take the leap and try to make options non-modal, too. But I'm sure there are quite a few stumble blocks on this route.

PR #1850, opened 2021-06-16
Add class method to check if Option exists.

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Add a check for dynamically set Options (window size and splitter settings) to avoid errors when trying to define multiple times.

# Problem

Dynamically set Options (window size and splitter settings) register errors when the same window is opened from different paths.  An example is the File Naming Script Editor.

# Solution

Add an Option class method to check if an option exists.  Use this to check before setting window size and splitter setting Options to avoid duplicate Options.

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -343,6 +343,10 @@ class Option(QtCore.QObject):
     def get(cls, section, name):
         return cls.registry.get((section, name))
 
+    @classmethod
+    def exists(cls, section, name):
+        return (section, name) in cls.registry
+
 
 class TextOption(Option):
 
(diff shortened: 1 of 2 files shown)


rdswift:
If the `Option.exists()` method isn't required for anything else, it could be replaced by an `add_if_missing()` method, thus saving a separate check for each Option in `picard/ui/__init__.py`.

phw (approved):
Well done

PR #1851, opened 2021-06-21
PICARD-9: Add support for user profiles

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Add support for user profiles.

# Problem

* JIRA ticket (_optional_): PICARD-9

There has been a long-standing request to allow for different user settings profiles, with a quick and easy method for switching between profiles.  The plan for implementing user profiles involves maintaining a stack of profiles that would be applied sequentially (for active profiles in the stack).  See the discussion on the JIRA ticket.

# Solution

Implement a new subclass of `ConfigSection` to use for user settings that applies the user profile stack automatically to all `config.setting` assignments and retrievals.  Add new keys to store the user profiles list and settings (overrides) for each profile in a new `config.profiles` section.  Add a new dialog accessed from the main screen menu to manage the user profiles and settings.

# Action

Additional work to be completed when this is finalized:

* [ ] Modify the File Naming Script Editor to block deleting a file naming script that is included in a profile's settings.
* [ ] Update the documentation to explain the use of profiles and the profile editor.  Much of this can be copied from @zas's comments on PICARD-9.


picard/config.py
```diff
@@ -129,6 +131,51 @@ def value(self, name, option_type, default=None):
...
+        super().__init__(config, name)
+        self.__qt_config = config
+        self.__name = name
+        self.__prefix = self.__name + '/'
+        self._memoization = defaultdict(Memovar)
+        Option.add_if_missing(name, 'user_profiles', [])
+
+    def _get_profiles(self):
+        profiles = self['user_profiles']
+        return profiles if profiles is not None else []
```
zas, line 147: We could make this a generator returning only active profiles:

```python
def _get_active_profiles(self):
     profiles = self['user_profiles']
     if profiles:
           for profile in profiles:
               if profile['active']:
                   yield profile
```

It would help to simplify code below.

```python
for profile in self._get_active_profiles():
      value = getattr(profile['settings'], name, None)
      if value is not None:
           return value
```

picard/config.py
```diff
@@ -129,6 +131,51 @@ def value(self, name, option_type, default=None):
...
+        opt = Option.get(self.__name, name)
+        if opt is None:
+            return None
+        return self.value(name, opt, opt.default)
+
+    def __setitem__(self, name, value):
+        # Don't process settings that are not profile-specific
+        if name in UserProfileGroups.get_profile_settings_list():
+            profiles = self._get_profiles()
+            for idx in range(len(profiles)):  # pylint: disable=consider-using-enumerate
```
zas, line 167: I don't get why we need to iterate using indexes here.

picard/profile.py
```diff
@@ -0,0 +1,125 @@
...
+
+    @classmethod
+    def get_profile_settings_list(cls):
+        """Iterable of all settings names in all profile groups.
+
+        Yields:
+            str: Setting name
+        """
+        for key, profile_group in cls.PROFILE_GROUPS.items():
+            for setting in profile_group["settings"]:
```
zas, line 119: What if 2 profile groups contain the same setting?

zas (changes requested):

picard/config.py
rdswift: Good suggestion.  I also found a problem where `getattr` wasn't returning the value, so I modified that a bit as well.

picard/config.py
rdswift: Because I thought there would be a problem modifying the item in the list being iterated.  I just tested and found out that was an incorrect assumption.  Modified to remove the indexes.

picard/profile.py
rdswift: They actually shouldn't, but even if they do the setting would appear twice in the list.  The problem would come if the user had both groups selected in the same profile and then removed one of the groups, where it might be possible that the setting would be removed from that profile.  When the time comes to do the setting removal, the code should take that into account.  Good catch!

zas: This was the whole point of what I called "stack", the idea behind this is to let user (at some point) defines whichever group of settings he wants, and applies first one met.

If we have group 1 (G1) with setting (S1), and group 2 with same setting but different value (S2), along default setting (SD), and all groups are enabled, we iterate from top to bottom (bottom being default):

Enabled groups order:
```
G2 -> G1 -> D
```

```
S2? -> no -> S1? -> no -> use SD
       | yes         | yes
       |- use S2     |-use S1
```

rdswift: I think we're using the same terms for different things.  To clarify my use:

**profile** is an item in the **profile stack**.  Each profile contains zero or more **settings** which are used to override the base settings if the profile is active.  When retrieving a value from `config.setting`, the system looks at each profile in the stack from top to bottom (i.e. first to last in the `config.setting["user_profiles"]` list).  If the profile is "active" and contains a value (which is not `None`) for the requested setting, then that value is returned and the profile review process terminates. If there is no active profile that provides the requested setting, then the value from `config.setting` is returned.

When saving a setting, the system again looks at each profile in the stack from top to bottom (i.e. first to last in the `config.setting["user_profiles"]` list).  If an "active" profile is found, and it contains an item in its "settings" dictionary for the setting being saved, then that profile is updated with the new setting and the process review process terminates, with the updated profile saved to `config.setting["user_profiles"]`.  If there is no active profile that includes the setting in its "settings" dictionary, then the value in `config.setting` is updated.

I think that we share the same view as to how the profile stack and setting assignment and retrieval would work.

**setting groups** are used to logically separate the settings into lists of related settings, and are primarily intended to allow the user to add/remove settings to/from a profile as a group rather than specifying each individual setting.  For example, when the user is creating or editing a profile, they could be presented with a tree view of settings available for selection with the setting groups as the branches and the settings within each setting group as the leaves.  Selection/deselection could be done with checkboxes at the branch and leaf level in the tree view.  When a setting is selected for a profile, that setting would be added to the profile's "settings" dictionary with an initial value of `None`.  When a setting is deselected from a profile, that setting would be removed from the profile's "settings" dictionary.

Depending on how we decide to proceed, potentially only the branch level would be displayed for selection.  In this case, when a setting group is selected for a profile, each of the settings associated with that settings group would be added to the profile's "settings" dictionary with an initial value of `None`.  When a settings group is deselected from a profile, each of the settings associated with that settings group would be removed from the profile's "settings" dictionary.

zas: > setting groups

That's what you named "profile groups" in the code, right? Perhaps it would be clearer if named "setting groups", according to the description above those don't have much to do with profiles (apart the fact they are used by them).

rdswift: That makes a lot of sense, and should help clear up some of the confusion.  I'll make the change.

picard/config.py
```diff
@@ -129,6 +131,53 @@ def value(self, name, option_type, default=None):
...
+                    return profile['settings'][name]
+        opt = Option.get(self.__name, name)
+        if opt is None:
+            return None
+        return self.value(name, opt, opt.default)
+
+    def __setitem__(self, name, value):
+        # Don't process settings that are not profile-specific
+        if name in UserProfileGroups.get_all_settings_list():
+            profiles = self._get_profiles()
```
zas, line 169: We can use `_get_active_profiles()` here too I think.

picard/profile.py
```diff
@@ -0,0 +1,130 @@
...
+    @classmethod
+    def get_all_settings_list(cls):
+        """Iterable of all settings names in all setting groups.
+
+        Yields:
+            str: Setting name
+        """
+        for key, settings_group in cls.SETTINGS_GROUPS.items():
+            for setting in settings_group["settings"]:
+                yield setting
```
zas, line 120: You can use `yield from settings_group["settings"]`.

That said, since we don't care order (right?), I would use an union of sets to return a set of unique values in this case (also we don't need SETTINGS_GROUPS keys).

```
settings = set()
for settings_group in cls.SETTINGS_GROUPS.values():
     settings |= settings_group["settings"]
return settings
```

What do you think?

picard/profile.py
```diff
@@ -0,0 +1,130 @@
...
+
+    @classmethod
+    def get_setting_groups_list(cls):
+        """Iterable of all setting groups keys.
+
+        Yields:
+            str: Key
+        """
+        for key in cls.SETTINGS_GROUPS:
+            yield key
```
zas, line 130: We can replace those 2 lines by: `yield from cls.SETTINGS_GROUPS`

zas (changes requested):

picard/config.py
rdswift: I'm not sure that we can because we need the full list of profiles to write back to the "user_profiles" setting.  I'll do some testing just to make sure.

An alternative **might** be to store the list as an ordered dictionary, or as a dictionary with some sort of sort key.

picard/profile.py
rdswift: I like it.  :wink:

picard/config.py
zas: > **EDIT 2:** Another alternative would be to store the actual settings for each profile in a dictionary keyed on unique profile ID (uuid?) with the id, activated status and title in a list of tuples (ordered by position in the profile stack). That may provide faster access to the settings, and avoid having to re-write the whole list every time a setting is updated.

Yes, this approach is likely better, worth a try imho.

picard/profile.py
```diff
@@ -0,0 +1,130 @@
...
+    @classmethod
+    def get_all_settings_list(cls):
+        """Iterable of all settings names in all setting groups.
+
+        Yields:
+            str: Setting name
+        """
+        settings = set()
+        for settings_group in cls.SETTINGS_GROUPS.values():
+            settings |= set(settings_group["settings"])
```
zas, line 120: We could define settings as sets directly in `SETTINGS_GROUPS` (and sort on display if needed).

picard/config.py
```diff
@@ -129,6 +131,62 @@ def value(self, name, option_type, default=None):
...
+        if profile_settings is None:
+            log.error("Unable to find settings for user profile '%s'", id)
+            return {}
+        return profile_settings
+
+    def __getitem__(self, name):
+        # Don't process settings that are not profile-specific
+        if name in UserProfileGroups.get_all_settings_list():
+            for id in self._get_active_profile_ids():
+                profile_settings = self._get_profile_settings(id)
```
zas, line 165: A generator could replace those 2 lines:

```python
def _get_active_profile_settings(self):
    for id in self._get_active_profile_ids():
        yield self._get_profile_settings(id)
```

```python
for profile_settings in self._get_active_profile_settings():
    ....
```

picard/config.py
```diff
@@ -129,6 +131,62 @@ def value(self, name, option_type, default=None):
...
+        opt = Option.get(self.__name, name)
+        if opt is None:
+            return None
+        return self.value(name, opt, opt.default)
+
+    def __setitem__(self, name, value):
+        # Don't process settings that are not profile-specific
+        if name in UserProfileGroups.get_all_settings_list():
+            for id in self._get_active_profile_ids():
+                settings = self._get_profile_settings(id)
```
zas, line 177: A generator could replace those 2 lines:

```python
def _get_active_profile_settings_with_id(self):
    for id in self._get_active_profile_ids():
        yield id, self._get_profile_settings(id)
```

```python
for id, profile_settings in self._get_active_profile_settings_with_id():
    ....
                self["user_profile_settings"][id] = settings
```

picard/config.py
```diff
@@ -129,6 +131,62 @@ def value(self, name, option_type, default=None):
...
+                if name in settings:
+                    settings[name] = value
+                    profile_settings = self["user_profile_settings"]
+                    profile_settings[id] = settings
+                    name = 'user_profile_settings'
+                    value = profile_settings
```
zas, line 183: Unsure about this `value`, shouldn't be `settings` ?

zas (changes requested):

picard/profile.py
rdswift: I'll keep that in mind, but I'm thinking of changing the structure to include a display name for the profile editor dialog.  Is it okay if we re-visit this later?

zas: Of course, that's early stages anyway.

picard/config.py
rdswift: I'll actually use the `_get_active_profile_settings()` generator (updated as suggested below) and just ignore the `id` part of the tuple returned.  That should avoid some code duplication.

picard/config.py
rdswift: I'll actually use this same generator for the `_get_active_profile_settings()` and just ignore the `id` part of the tuple returned.  That should avoid some code duplication.

picard/config.py
rdswift: That whole part is inefficient.  Revised.

zas (changes requested):

rdswift:
I added a working version of the profile editor dialog, which I've been using for local testing.  Once we firm up structures and such, I'll see about adding some tests as well.

zas:
I did a quick test, it works well.
I'm a bit bothered by the fact profile-modified options are stored in an option, rather than in their own config sub-section.
Also we'll need to handle config upgrades (when an option name is modified, we'll need to propagate the change to profiles).

That's a good improvement already. Good job!

rdswift:
> I'm a bit bothered by the fact profile-modified options are stored in an option, rather than in their own config sub-section.

It should be fairly easy to store the profile settings dictionaries by profile id under a new `config.profiles` section, or were you thinking about having separate `ConfigSection` objects for each profile?

rdswift:
After moving the profile settings to a new `config.profiles` group, it seems that it broke all of the tests that use the `config.setting`, even though it seems to work perfectly when running Picard.  The tests all seem to be failing because `config.setting` can't see `config.profiles` and fail with a KeyError.  I suspect there's something that needs to be fixed in the `PicardTestCase` module, but I'm not finding it.

zas:
> After moving the profile settings to a new `config.profiles` group, it seems that it broke all of the tests that use the `config.setting`, even though it seems to work perfectly when running Picard. The tests all seem to be failing because `config.setting` can't see `config.profiles` and fail with a KeyError. I suspect there's something that needs to be fixed in the `PicardTestCase` module, but I'm not finding it.

I think that's due to the use of global `config` in `SettingConfigSection`.
https://github.com/metabrainz/picard/blob/2ea04a49cb964f13177f142e23942558329b2283/picard/config.py#L142

Use `self.__qt_config` instead of `config`:
https://github.com/metabrainz/picard/blob/2ea04a49cb964f13177f142e23942558329b2283/picard/config.py#L150
https://github.com/metabrainz/picard/blob/2ea04a49cb964f13177f142e23942558329b2283/picard/config.py#L162
https://github.com/metabrainz/picard/blob/2ea04a49cb964f13177f142e23942558329b2283/picard/config.py#L185
https://github.com/metabrainz/picard/blob/2ea04a49cb964f13177f142e23942558329b2283/picard/config.py#L187

rdswift:
> I think that's due to the use of global `config` in `SettingConfigSection`.

That was exactly the problem.  **THANK-YOU!**

picard/config_upgrade.py
```diff
@@ -427,6 +427,17 @@ def rename_option(config, old_opt, new_opt, option_type, default):
         _s[new_opt] = _s.value(old_opt, option_type, default)
         _s.remove(old_opt)
 
+        _p = config.profiles
+        ListOption.add_if_missing("profiles", "user_profiles", [])
+        Option.add_if_missing("profiles", "user_profile_settings", {})
```
zas, line 432: Since those are also used in `SettingConfigSection.__init__()`, we should move them to a class method and call this method.

Something like:

```python
class SettingConfigSection(ConfigSection):
    ....
    @classmethod
    def init_profile_options(cls):
        ListOption.add_if_missing("profiles", cls.PROFILES_KEY, [])
        Option.add_if_missing("profiles", cls.SETTINGS_KEY, {})
```

Untested though.

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,454 @@
...
+class ProfileEditorDialog(SingletonDialog, PicardDialog):
+    """User Profile Editor Page
+    """
+    TITLE = N_("User profile editor")
+    STYLESHEET_ERROR = OptionsPage.STYLESHEET_ERROR
+
+    help_url = PICARD_URLS["doc_profile_edit"]
+
+    PROFILES_KEY = "user_profiles"
+    SETTINGS_KEY = "user_profile_settings"
```
zas, line 58: Use `SettingConfigSection.PROFILES_KEY` and `SettingConfigSection.SETTINGS_KEY` ?

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,454 @@
...
+    POSITION_KEY = "last_selected_profile_pos"
+
+    options = [
+        ListOption.add_if_missing("profiles", PROFILES_KEY, []),
+        Option.add_if_missing("profiles", SETTINGS_KEY, {}),
```
zas, line 63: Are those really needed (as they are declared in `SettingConfigSection.__init__()`) ?

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,454 @@
...
+
+        self.main_window = parent
+        # # Get pointer to the main window regardless of where this dialog was called.
+        # temp = self
+        # while temp:
+        #     name = temp.objectName() if temp.objectName() else temp.__class__.__name__
+        #     if name == 'MainWindow':
+        #         self.main_window = temp
+        #         break
+        #     temp = temp.parent()
```
zas, line 82: Don't forget to remove commented out code

picard/config.py
```diff
@@ -153,9 +215,10 @@ def __initialize(self):
 
         self.setAtomicSyncRequired(False)  # See comment in event()
         self.application = ConfigSection(self, "application")
-        self.setting = ConfigSection(self, "setting")
+        self.profiles = ConfigSection(self, "profiles")
+        self.setting = SettingConfigSection(self, "setting")
         self.persist = ConfigSection(self, "persist")
-        self.profile = ConfigSection(self, "profile/default")
+        # self.profile = ConfigSection(self, "profile/default")
```
zas, line 221: This can be removed I guess.

zas (changes requested):

picard/config.py
rdswift: Okay, but it's such a neat idea.  It's just that I couldn't see how to effectively use it with the stacked profiles solution.

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+
+        Args:
+            id (str): ID of the profile
+
+        Returns:
+            dict: Profile settings
+        """
+        # Add empty settings dictionary if no dictionary found for the profile.
+        if id not in self.profile_settings:
+            self.profile_settings[id] = {}
```
zas, line 170: I'm not fond of a `get` method modifying things.
The comment tells what it does, but doesn't explain why it is needed. Can you clarify?

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+        Args:
+            update_settings (bool, optional): Update settings tree. Defaults to True.
+        """
+        item = self.get_current_selected_item()
+        if item:
+            id = item.profile_id
+            self.current_profile_id = id
+            settings = self.get_settings_for_profile(id)
+            if update_settings:
+                self.make_setting_tree(settings=settings)
```
zas, line 205: Since `settings` is only used in this `if` branch, its declaration should be moved here.

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+            self.ui.settings_status.setText("")
+
+    def profile_settings_changed(self):
+        """Check if the settings selected in the settings tree have changed from the last saved settings.
+
+        Returns:
+            bool: True if settings have changed, otherwise False
+        """
+        checked_items = set(self.get_checked_items_from_tree())
+        settings = set(self.profile_settings[self.current_profile_id].keys())
```
zas, line 293: `.keys()` is useless I think here (since that's a dict-like object, it will return keys anyway on conversion to `set`)

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+                    _("Save Settings"),
+                    _("There are no profile settings changes to save."),
+                    QtWidgets.QMessageBox.Ok,
+                    self
+                )
+                dialog.exec_()
+            return
+
+        checked_items = set(self.get_checked_items_from_tree())
+        settings = set(self.profile_settings[self.current_profile_id].keys())
```
zas, line 315: `.keys()` can prolly be removed.

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+
+        # Add new items to settings
+        for item in checked_items:
+            if item not in settings:
```
zas, line 319: Use `for item in  checked_items.difference(settings)`

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+                self.profile_settings[self.current_profile_id][item] = None
+
+        # Remove unchecked items from settings
+        for item in settings:
+            if item not in checked_items:
```
zas, line 324: Use `for item in settings.difference(checked_items)`

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+                if item.checkState(0) == QtCore.Qt.Checked:
+                    yield item.data(0, QtCore.Qt.UserRole)
+
+    def unsaved_changes_confirmation(self):
+        """Check if there are unsaved changes and ask the user to confirm the action resulting in their loss.
+
+        Returns:
+            bool: True if no unsaved changes or user confirms the action, otherwise False.
+        """
+        if not self.loading and self.current_profile_id and self.profile_settings_changed() and not self.confirmation_dialog(
```
zas, line 433: In `current_item_changed()` there's already a test against `self.loading`, imho it makes more sense to test for it before calling this method anyway. Unsure.

test/test_profiles.py
```diff
@@ -0,0 +1,191 @@
...
+    Option,
+    TextOption,
+)
+from picard.profile import UserProfileGroups
+
+
+class TestPicardProfilesCommon(PicardTestCase):
+
+    PROFILES_KEY = 'user_profiles'
+    SETTINGS_KEY = 'user_profile_settings'
```
zas, line 42: Use `SettingConfigSection.PROFILES_KEY` and `SettingConfigSection.SETTINGS_KEY`

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+            return
+        self.building_tree = True
+        for id, group in UserProfileGroups.SETTINGS_GROUPS.items():
+            title = group["title"]
+            group_settings = group["settings"]
+            widget_item = QtWidgets.QTreeWidgetItem([title])
+            widget_item.setFlags(QtCore.Qt.ItemIsEnabled | QtCore.Qt.ItemIsUserCheckable | QtCore.Qt.ItemIsAutoTristate)
+            widget_item.setCheckState(0, QtCore.Qt.Unchecked)
+            for setting in group_settings:
+                child_item = QtWidgets.QTreeWidgetItem([_(setting[1])])
```
zas, line 231: I don't like numeric indexes.
We could use namedtuple for profile settings, as per https://github.com/rdswift/picard/pull/9

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+                child_item.setData(0, QtCore.Qt.UserRole, setting[0])
```
zas, line 232: What's the first `0`? Please use a module or class constant instead.

zas (changes requested):
Just (few more) stylistic remarks, but the patch is very good ;)

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,455 @@
...
+        """Save any changes to the current profile's settings, save all updated profile
+        information to the user settings, and close the profile editor dialog.
+        """
+        # Save any unsaved changes
+        if self.ui.profile_list.count():
+            self.save_profile(silent=True)
+
+        all_profiles = list(self._all_profiles())
+        all_profile_ids = set(x['id'] for x in all_profiles)
+        keys = set(deepcopy(self.profile_settings).keys())
```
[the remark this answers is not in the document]
rdswift: Right.  That was a holdover from when I was iterating though `self.profile_settings` and was getting an error about the iterator changing whenever I deleted one of the elements.

picard/ui/profileeditor.py
rdswift: It was actually to avoid a KeyError exception if that element was missing.  This happens when a new profile is created.  I'll change it to:

```
return self.profile_settings[id] if id in self.profile_settings else {}
```

and add some code to make sure that the key is created when a new element is created.  To do that I'll return the ID from `ProfileListWidget.add_profile()`.  That actually makes more sense than the way I'm currently doing it.

rdswift: Actually, it's a lot more of an issue to change than I expected because other code is triggered when the `ProfileListWidget.add_profile()` is executed which also relies on this key having been created.  I'll add more of an explanation of what's happening to the `get_settings_for_profile()` method.

picard/ui/profileeditor.py
rdswift: You're right.  It's no longer necessary here.

picard/ui/profileeditor.py
rdswift: That's the column in the `QTreeWidgetItem` that the changes apply to.  I'll set and use a class constant.

zas (dismissed):
LGTM

zas (dismissed):

picard/config.py
```diff
@@ -129,6 +131,70 @@ def value(self, name, option_type, default=None):
...
+
+    def __setitem__(self, name, value):
+        # Don't process settings that are not profile-specific
+        if name in UserProfileGroups.get_all_settings_list():
+            for id, settings in self._get_active_profile_settings():
+                if name in settings:
+                    settings[name] = value
+                    all_settings = self.__qt_config.profiles[self.SETTINGS_KEY]
+                    all_settings[id] = settings
+                    self.__qt_config.profiles[self.SETTINGS_KEY] = all_settings
```
zas, line 191: @rdswift I wonder about those 3 lines:

https://github.com/metabrainz/picard/blob/e01b97844132e01c7f5463eef91653cc064c369e/picard/config.py#L189-L191

Aren't they equivalent to:

```python
self.__qt_config.profiles[self.SETTINGS_KEY][id] = settings
```

(I may miss something here)

rdswift: You're not missing something, I was.  I didn't realize that you could update a subscripted value directly.  I'll make the change.

zas (approved):

picard/ui/profileeditor.py
```diff
@@ -0,0 +1,456 @@
...
+
+        self.delete_profile_button = QtWidgets.QPushButton(_('Delete'))
+        self.delete_profile_button.setToolTip(_("Delete the profile"))
+        self.delete_profile_button.clicked.connect(self.delete_profile)
+        self.ui.buttonbox.addButton(self.delete_profile_button, QtWidgets.QDialogButtonBox.ActionRole)
+
+        self.save_profile_button = QtWidgets.QPushButton(_('Save'))
+        self.save_profile_button.setToolTip(_("Save changes to the profile"))
+        self.save_profile_button.clicked.connect(self.save_profile)
+        self.ui.buttonbox.addButton(self.save_profile_button, QtWidgets.QDialogButtonBox.ActionRole)
```
phw, line 128: This should be `ApplyRole`

phw (changes requested):
Good work. This works really well, and I am especially glad that this was possible with rather small and clean changes to the config module itself.

From a user perspective I find the settings concept rather difficult to grasp, but it is very flexible. Before deep diving into the code itself I tried this functionality from a user's perspective. But I definitely had to read the documentation to fully understand it (even though I had kind of suspected how this works based on our previous discussions).

I have a few notes about the profile editor dialog itself:

- The buttons should be placed in `QDialogButtonBox` and should be added with appropriate roles. This ensures the buttons render consistent with other dialogs and platform dependent.
- I find the separate "Save" and "Make it so!" buttons really confusing and not easy to use. I'd favor having this behave like options themselves: Have a "Cancel" button and a single "Make it so!" button to save and close
- I would consider moving the "New", "Copy" and "Delete" buttons below the widget, next to the ordering buttons. That would make it consistent with other widgets we use in options where the buttons to control a list of things (e.g. scripts) are next to the widget presenting this list. It would also make the main dialog buttons look less cluttered

I would also add a keyboard shortcut to open the profile editor. `Ctrl+Shift+P` maybe? That would allow the user to quickly change profiles without the menu.

For a future refinement to make the behavior of profiles more clear I think it would be beneficial to indicate in the profile editor which settings actually have distinct values. Maybe even show the value itself (which would probably require that we have a useful rendering depending on the option type, e.g. show BoolOption as "Yes" / "No" or something.

But we can do this in a follow up PR.

picard/ui/profileeditor.py
rdswift: I actually removed it completely to simplify the UI.  Changes are now saved locally (within the dialog) as they are made.

rdswift:
> I have a few notes about the profile editor dialog itself:

> * The buttons should be placed in `QDialogButtonBox` and should be added with appropriate roles. This ensures the buttons render consistent with other dialogs and platform dependent.

Most were in the buttonbox, except "Make It So!".  Now are all in buttonboxes with the appropriate roles assigned.

> * I find the separate "Save" and "Make it so!" buttons really confusing and not easy to use. I'd favor having this behave like options themselves: Have a "Cancel" button and a single "Make it so!" button to save and close

I actually simplified it by removing the "Save" button and auto-save the changes locally (within the dialog) as they are made.  Do we still need a "Reset" button?

> * I would consider moving the "New", "Copy" and "Delete" buttons below the widget, next to the ordering buttons. That would make it consistent with other widgets we use in options where the buttons to control a list of things (e.g. scripts) are next to the widget presenting this list. It would also make the main dialog buttons look less cluttered

Done.  I agree that they look better there.

> I would also add a keyboard shortcut to open the profile editor. `Ctrl+Shift+P` maybe? That would allow the user to quickly change profiles without the menu.

Done.  At the same time I also added one for the file naming script editor to be consistent.  I hope that's okay.

picard/ui/profileeditor.py
```diff
@@ -139,7 +135,7 @@ def load(self):
         """Load initial configuration.
         """
         config = get_config()
-        self.profile_settings = config.profiles[self.SETTINGS_KEY]
+        self.profile_settings = deepcopy(config.profiles[self.SETTINGS_KEY])
```
zas, line 138: Why this change?

rdswift: Because changes within the dialog were cascading into the `config.profiles` even without the user clicking "Make It So!"

rdswift: I'll add a comment in the code explaining that.

zas (dismissed):

zas (approved):

zas (approved):

phw (approved):
The dialog now looks much clearer with re-arranged buttons.

LGTM

PR #1852, opened 2021-06-21
Convert script editor dialog to SingletonDialog

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Allow opening Settings dialog while File Naming Script Editor is open.

# Problem

When the file naming script editor is open from the main screen, the user is unable to open the Settings dialog.  This is not intuitive behavior.

# Solution

Convert the file naming script editor to a `SingletonDialog` and switch its parent setting to match the currently active window (main screen or settings dialog).

# Action


phw (changes requested):
Finally found enough time and had the head free to look at the code, and actually also test run it on Linux and macOS.

First off: This patch looks good and the important parts work.

On macOS there is an issue with window position:

- Open the script editor dialog first
- Then open options dialog

Kind of unexpected the script editor now stays behind options.It is in front of options when opening it from options. So there seems to be some issue with the reparenting. Not yet sure what to do about this.

But one thing we actually should change in some way: If you have the script editor open and either then open or close the options dialog you loose all pending changes you had in the dialog, it completely resets.

I understand why this happens when one closes options, because this also closes the dialog and you have code to reopen it. I don't yet fully understand why this happens when options get opened. But in both cases we should ensure somehow that the unsaved state does not get lost.

Also I wonder if the dialog really needs to fully close when closing options. Maybe we can unparent it before the options actually close?

rdswift:
> Kind of unexpected the script editor now stays behind options.It is in front of options when opening it from options. So there seems to be some issue with the reparenting. Not yet sure what to do about this.

Perhaps we need to close it and re-open it from Options.  It's disappointing that the re-parenting doesn't work.

> But one thing we actually should change in some way: If you have the script editor open and either then open or close the options dialog you loose all pending changes you had in the dialog, it completely resets.
>
> I understand why this happens when one closes options, because this also closes the dialog and you have code to reopen it. I don't yet fully understand why this happens when options get opened. But in both cases we should ensure somehow that the unsaved state does not get lost.

I think it's forcing a `load()` after re-parenting when opening Options.  I'll have a look at how to avoid that.  For sure the changes will be lost if it closes and re-opens (as is currently the case when closing Options), so that may require more thought.

> Also I wonder if the dialog really needs to fully close when closing options. Maybe we can unparent it before the options actually close?

I tried that and closing Options still destroys the script editor dialog object (without closing it so that the class still thinks there is an active dialog).  It appears that when the parent is set to Options, the script editor dialog gets added to the list of objects to clean up, but changing the parent doesn't remove it from the list.  That's why I ended up closing it when Options is closed.  I agree that it would be best if we could keep it open, but I haven't found a way to do that.

picard/ui/scripteditor.py
```diff
@@ -408,6 +411,26 @@ class ScriptEditorDialog(PicardDialog):
...
     default_script_filename = "picard_naming_script.ptsp"
 
+    @classmethod
+    def show_instance(cls, *args, **kwargs):
+        if cls._instance:
+            # Accommodate condition where OptionsPage closing may have destroyed the
+            # ScriptEditorDialog instance without resetting the cls._instance property
+            cls._instance.show()
+            if not cls._instance.isVisible():
+                cls._instance = None
```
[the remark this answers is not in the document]
rdswift: That's what I thought should happen, but it seems that closing Options actually destroys the script editor object without first triggering its `_on_dialog_finished()`.  That's the source of most of the issues.

picard/ui/options/renaming.py
```diff
@@ -142,6 +142,14 @@ def __init__(self, parent=None):
 
         self.current_row = -1
 
+        # Get pointer to main window to enable updating menu items
+        test_parent = self
+        while test_parent is not None:
+            if getattr(test_parent.__class__, '__name__', '') == 'MainWindow':
+                self.main_window = test_parent
+                break
+            test_parent = test_parent.parent()
```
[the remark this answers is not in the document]
rdswift: Oops.  That was just some code that I had in while testing and forgot to remove.  :blush:

rdswift:
I made a couple of changes to avoid doing a full reload when changing parents, and everything seems to be working here on my Windows 7 system.  It now retains changes in progress when opening and closing Options.  Unfortunately, I don't think this will resolve the window placement issue on the mac.

picard/ui/scripteditor.py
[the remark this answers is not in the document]
phw: Ok :( It seems to run, but quite likely this is a timing issue and there is a short time where the instance is still there. Let's keep your check then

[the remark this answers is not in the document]
rdswift: Thinking about this a bit more, I believe your suggestion of moving the re-parenting code into the `SingletonDialog` class and handling it before activating the window is what made this work.  The way I had it originally, the window was already activated against the old parent before trying to apply the new parent.

picard/ui/mainwindow.py
```diff
@@ -1098,7 +1099,28 @@ def show_about(self):
...
+        options_dialog = OptionsDialog.show_instance(page, self)
+        options_dialog.finished.connect(self.options_closed)
+        if self.script_editor_is_open:
+            # Disable signal processing to avoid saving changes not processed with "Make It So!"
+            self.script_editor_dialog.signal_save.disconnect()
+            self.script_editor_dialog.signal_selection_changed.disconnect()
+            self.script_editor_dialog.signal_update_scripts_list.disconnect()
+            self.script_editor_dialog.signal_index_changed.disconnect()
+
+            for op_page in options_dialog.pages:
```
phw, line 1111: Instead of the search code here I would probably add a helper to `OptionsDialog`. Actually it can retrieve the page widget with a direct lookup, no need for a loop:

```python
    def get_page(self, name):
        return self.item_to_page[self.page_to_item[name]]
```

Then here it becomes:

```python
options_dialog.get_page('filerenaming').show_script_editing_page()
```

phw (commented):
I tried to get this working on macOS, without success so far. What I notice is that the parent change on the dialog happens before the options dialog is actually visible, and I think that's the culprit.

So to fix this I tried moving the handling to `OptionsDialog` by implementing `openEvent` and also overwriting `open`. But both did not give any better result. The only thing I achieved was that then the script editor also showed up behind options on Windows.

Strangely when I did remove the parent change when opening options the dialog properly came to front when triggered by the button in renaming options.

Currently I am out of ideas what to try :(

rdswift:
> Currently I am out of ideas what to try :(

You may have already tried this, but I just pushed a commit that moves the call to open the script editor from the main window to Options (which seems to work well here).  It may be a case of the origin of the call affecting things.

zas:
I found a reproducible exception while testing:

- open script editor from main menu
- open Options dialog
- press Edit Script in Options (it will make Script Editor dialog to get focus)
- close Script Editor

```
  File "./picard/ui/options/renaming.py", line 203, in script_editor_dialog_close
    if not self.script_editor_dialog.loading:
AttributeError: 'NoneType' object has no attribute 'loading'
```

picard/ui/mainwindow.py
```diff
@@ -1098,7 +1099,23 @@ def show_about(self):
...
     def show_options(self, page=None):
-        return OptionsDialog.show_instance(page, self)
+        options_dialog = OptionsDialog.show_instance(page, self)
+        options_dialog.finished.connect(self.options_closed)
+        if self.script_editor_is_open:
+            # Disable signal processing to avoid saving changes not processed with "Make It So!"
+            self.script_editor_dialog.signal_save.disconnect()
+            self.script_editor_dialog.signal_selection_changed.disconnect()
+            self.script_editor_dialog.signal_update_scripts_list.disconnect()
+            self.script_editor_dialog.signal_index_changed.disconnect()
```
zas, line 1109: I would keep signals connecting/disconnecting as part of `ScriptEditorDialog`.

picard/ui/mainwindow.py
```diff
@@ -1581,7 +1586,8 @@ def script_editor_closed(self):
         """Process "finished" signal from the script editor.
         """
         self.show_script_editor_action.setEnabled(True)
-        self.options_action.setEnabled(True)
+        self.script_editor_is_open = False
+        self.script_editor_dialog = None
```
zas, line 1590: I really wonder if we need `script_editor_is_open` at all, what about just using `script_editor_dialog` ensuring it is `None` when dialog is closed (which should be the case already, right?).

zas (changes requested):

picard/ui/mainwindow.py
rdswift: It's easy enough to get rid of it within the MainWindow dialog, but it is also used:
1. to indicate that the script editor has been opened in Options as in `picard/ui/options/renaming.py:L200` (in which case `self.script_editor_dialog` will still be `None`), and
2. to indicate that the script editor has been closed in Options as in `picard/ui/options/renaming.py:L203` (in which case `self.script_editor_dialog` will still be a `ScriptEditorDialog` instance).

In the second case, I suppose that we could have `picard/ui/options/renaming.py` force the MainWindow `self.script_editor_dialog` to `None` (with the appropriate cleanup actions), but I'm not sure how to handle the first case without potentially introducing more `ScriptEditorDialog` parent issues.  It may be something as simple as setting `self.script_editor_dialog` to `True` and open an instance if it is `not None`.  I'll have to think on that a bit more.

rdswift: I just tried it and it seems to work well.  I'll push an update.

phw:
Just a quick update that this does still not work on macOS. I haven't found the time yet to experiment here in depth.

rdswift:
> Just a quick update that this does still not work on macOS. I haven't found the time yet to experiment here in depth.

When you're testing, you might try changing Line 507 of `picard/ui/scripteditor.py` from `flags = self.windowFlags() | QtCore.Qt.Window` to `flags = self.windowFlags() | QtCore.Qt.Window | QtCore.Qt.Tool` or possibly just `flags = self.windowFlags() | QtCore.Qt.Tool`.  Accrding to the description at https://doc.qt.io/qt-5/qt.html#WindowType-enum 

> ... On macOS, tool windows correspond to the NSPanel class of windows. This means that the window lives on a level above normal windows making it impossible to put a normal window on top of it. By default, tool windows will disappear when the application is inactive. This can be controlled by the Qt::WA_MacAlwaysShowToolWindow attribute.

I'm kind of grasping at straws here, but it sounds like it might be promising.  Sorry I can't test it here, but I don't have access to a macos system.

rdswift:
I tracked the events triggered when opening the OptionsDialog object to try to identify the latest event to use for triggering the re-parenting of the ScriptEditorDialog.  It appears that the `QEvent.Enter` event may be the one to use to delay the re-parenting as long as possible to allow the window to become activate on macos.

The list of triggered events on my Windows 7 system when opening and then closing OptionsDialog is:

```
000001: QEvent.Move (13)
000002: QEvent.PlatformSurface (217)
000003: QEvent.WinIdChange (203)
000004: QEvent.WindowIconChange (34)
000005: QEvent.Polish (75)
000006: QEvent.DynamicPropertyChange (170)
000007: QEvent.ChildPolished (69)
000008: QEvent.ChildPolished (69)
000009: QEvent.Move (13)
000010: QEvent.Resize (14)
000011: QEvent.Show (17)
000012: QEvent.Move (13)
000013: QEvent.Resize (14)
000014: QEvent.CursorChange (183)
000015: QEvent.ShowToParent (26)
000016: QEvent.ZOrderChange (126)
000017: QEvent.WindowActivate (24)
000018: QEvent.ActivationChange (99)
000019: QEvent.Paint (12)
000020: QEvent.PolishRequest (74)
000021: QEvent.LayoutRequest (76)
000022: QEvent.UpdateLater (78)
000023: QEvent.UpdateRequest (77)
000024: QEvent.Paint (12)
000025: QEvent.Enter (10)
000026: QEvent.LayoutRequest (76)
000027: QEvent.LayoutRequest (76)
000028: QEvent.LayoutRequest (76)
000029: QEvent.LayoutRequest (76)
000030: QEvent.CursorChange (183)
000031: QEvent.UpdateRequest (77)
000032: QEvent.Paint (12)
000033: QEvent.CursorChange (183)
000034: QEvent.UpdateRequest (77)
000035: QEvent.Paint (12)
000036: QEvent.UpdateRequest (77)
000037: QEvent.Paint (12)
000038: QEvent.UpdateRequest (77)
000039: QEvent.Paint (12)
000040: QEvent.UpdateRequest (77)
000041: QEvent.UpdateRequest (77)
000042: QEvent.Paint (12)
000043: QEvent.WindowDeactivate (25)
000044: QEvent.ActivationChange (99)
000045: QEvent.Hide (18)
000046: QEvent.HideToParent (27)
000047: QEvent.DeferredDelete (52)
```

phw:
@rdswift I tried again to get the workflow of having the script editor dialog open first, then opening options on macOS running. No matter what I tried, the script editor stays behind options. One can interact with it (at least as far as it is visible), but cannot get it to front.

I also tried making the window modal before re-parenting. Then it was in front, but exclusive and one could not interact with options. Changing modal back after showing the window does not change this. Changing it back before showing has no effect, resulting in the original behavior.

Not really sure what to do about this.

rdswift:
That's disappointing.  Perhaps as a really ugly hack we could close the script editor then reopen it again, but that would likely not allow us to save any unsaved changes first.  I'll do some experimenting to see if I can come up with anything better.

rdswift:
I wonder if the answer might lay in https://doc.qt.io/qt-5/macos-issues.html#using-native-cocoa-panels ?

phw:
@rdswift I had no luck with getting this to work on macOS. I was hopeful about your latest patches, but they did not work. Also it is not solely a timing issue as I had first thought, delaying the opening does not work. I found not way to get the script editor show up in front of the option dialog if the script editor is opened earlier.

I tend toward just accepting this as a known issue on macOS. @zas What do you think?

zas:
> @zas What do you think?

Well, if this patch improves user experience on all platforms but on macos, I'd say it is still an improvement, and should be merged.
We can add a ticket about the macos-specific issue.

phw:
Ok, in this case let's merge, but we should skip the last three commits. I can test this one more time and then merge this locally.

rdswift:
> @rdswift I had no luck with getting this to work on macOS. I was hopeful about your latest patches, but they did not work.

Darn.  I had high hopes for that solution.  I'm all out of suggestions.

> Ok, in this case let's merge, but we should skip the last three commits.

I'll roll back to commit https://github.com/metabrainz/picard/pull/1852/commits/265704a54bfe0afec980cb22ab4504fbb49db52d and rebase to get rid of the conflicts then.

phw (approved):
LGTM

I added PICARD-2249 about the macOS dialog issue.

zas (approved):
Lgtm

PR #1854, opened 2021-06-30
PICARD-1734: Add AcousticBrainz analyzer features

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
Call the AcousticBrainz/Essentia feature extractor from Picard and submit the extracted features.

# Problem

* JIRA ticket (_optional_): PICARD-1734

# Solution

A single button triggers the check for duplicate AcousticBrainz features duplicates (same recordingID + extractor version), extract features of non-duplicates and tries to submit them.

# Action

Part of the AcousticBrainz code comes from the [acoustic-brainz client](https://github.com/MTG/acousticbrainz-client/blob/master/abz/acousticbrainz.py), that has a different license (GPL-v3). Not sure what to do with it.


Gabrielcarvfer:
@phw commit [b1033f1598ddafe8c524ba86968a1046ee19b12f](https://github.com/metabrainz/picard/pull/1854/commits/b1033f1598ddafe8c524ba86968a1046ee19b12f) has the version working with the requests library. Latest commit refuses to work with the webservice post for me (tried mingw python, but msys2 doesn't have pywin32; anaconda python doesn't work; official python doesn't work). Maybe Pycharm is doing something weird.

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+from picard.util import load_json
+from picard.webservice import ratecontrol
+
+
+ACOUSTICBRAINZ_HOST = "acousticbrainz.org"
+ACOUSTICBRAINZ_PORT = 80
+
+ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 1000)
+
+acousticbrainz_extractor = os.path.dirname(sys.argv[0]) + os.sep + "streaming_extractor_music" + ("" if sys.platform != "win32" else ".exe")
```
phw, line 42: You can use `picard.util.find_executable` here.

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+                          priority=True,
+                          important=False,
+                          parse_response_type=None)
+
+
+def extractor(input_path, output_path):
+    # If an a duplicate entry doesn't exist, process the file
+    args = [acousticbrainz_extractor, input_path, output_path]
+    startupinfo = subprocess.STARTUPINFO()
+    startupinfo.dwFlags |= subprocess.STARTF_USESHOWWINDOW
```
phw, line 61: Note that `subprocess.STARTUPINFO` is only available on Windows. What exactly does setting the `STARTF_USESHOWWINDOW` flag do, is that needed?

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,121 @@
...
+
+    if duplicate:
+        # If an entry for the same recording ID already exists
+        # using the same extractor version exists, skip extraction
+        results = (None, 0, "Duplicate")
+        extractor_callback(results)
+    else:
+        # Run extractor on a different thread and call the callback when done
+        future = tagger.thread_pool.submit(extractor, input_path, output_path)
+        future.add_done_callback(extractor_callback)
```
phw, line 97: Found out the issue: The problem is that calling the callback does not run on the main thread, and the callback then calls `submit_features`, which in turn means the webservice.post is not run on the main thread and that fails.

`tagger.thread_pool` is really not supposed to be used directly. Instead of custom handling you should use `picard,util.thread.run_task` like this:

```
run_task(partial(extractor, input_path, output_path), extractor_callback)
```

The callack then needs to accept a result and an error object, something like this:

```
def abz_extractor_callback(file, result, error):
    if error:
        # handle the error

    abz_metadata_file, result, error = result  # Now this conflicts a bit with the naming, that should be fixed
```

Actually the `extractor` function could probably be changed to throw on error, then the error would automatically end up as the error parameter in abz_extractor_callback

picard/acousticbrainz/__init__.py
Gabrielcarvfer: It prevents new shells from popping up. I could use the same thing fpcalc does here.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: It really did fix it. :D


And I found either a bug on the extractor or a borked file. I guess this is why the recordingId check was in place.

Extractor says:
musicbrainz_trackid (supposedly the recordingID) = **'235fa3a9-feb6-434c-8801-34370c63af50'**
acoustid_id = '4519f96a-b95d-4fad-bbbc-14eb6d1913fb'

Picard says:
musicbrainz_recordingid = '3e832e23-d08a-495f-82e2-8a3632947673'
musicbrainz_trackid = 'b9fc737c-db38-491f-a45f-3e987c6a8a30'
acoustid_id = '4519f96a-b95d-4fad-bbbc-14eb6d1913fb'

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Done.

Gabrielcarvfer:
At least the Windows installer version seems to be working. Thinking about how we could show if the file has already been processed.
I've included two fields to the Files (acousticbrainz_is_duplicate and acousticbrainz_features_file).

I was thinking of adding a column with states (color coated?):
Empty = unprocessed (acousticbrainz_features_file = None + is_duplicate = False)
Gray = extracted (acousticbrainz_features_file != None)
Blue(?) = duplicate (is_duplicate=True)
Green(?) = successfully submitted (acousticbrainz_features_file != None + is_duplicate = True)
Red(?) = submission failed (additional flag?)


<details>
<summary>Procedures</summary>

It works like this:
1. You click to process the file, it checks the files fields (tagger.extract_and_submit_acousticbrainz_features)
1.1 If is_duplicate is set, it calls abz_extractor_callback directly
1.2 If it has a features_file but is_duplicate isn't set, it calls abz_extractor_callback directly to submit the results
1.3 If its not a duplicate not had its features extracted yet, proceeds to 2
2. It tries to fetch the features for that file recordingid (acousticbrainz.__init__.run_extractor_entry)
2.1 If it succeeds, is_duplicate is marked as True and calls abz_extractor_callback/extractor_callback directly
2.2 If it fails, proceeds to 3
3. AcousticBrainz extractor is called (acousticbrainz.__init__.run_extractor) and proceeds to abz_extractor_callback/extractor_callback
4. abz_extractor_callback checks the results
4.1 If 1.2 or the extractor reported writing the results to the output json, we assign the extracted features file to the file.acousticbrainz_features_file and proceed to submission 5
4.2 If it was reported as a duplicate (1.1, 2.1), the is_flagged is set as True, this gets logged and submission is skipped
4.3 If it failed unexpectedly, the error gets logged
5. Before submitting, we check if features file trackid matches with the file recordingid (weird errors, but whatever). 
5.1 If they don't match, log error and abort submission
5.2 If they match, submit features and proceed to the callback 6
6. Checks results:
6.1 If submission was OK, mark file as a duplicate and log successful transmission
6.2 If submission failed, log failure.
</details>

picard/acousticbrainz/__init__.py
phw: > musicbrainz_trackid (supposedly the recordingID)

Yes, that's the recording ID. For historic reasons the tag is called musicbrainz_trackid, and that's what the extractor just puts into the JSON file. Picard internally calls this `musicbrainz_recordingid`.

But strange result. Could it be because of the actual recording ID not yet written to the file?

picard/acousticbrainz/__init__.py
phw: Ok, understood. I guess because fpcalc uses some Qt functions instead those take care of that already.

No objection to use subprocess here, especially as we have started to prefer Python standard library over Qt. I just would probably put it into a helper function in picard.util (maybe `run_executable` to follow the naming of `find_executable`) and the startupinfo stuff needs to be done only for Windows. Making it easy to call a command line executable and get back the output seems to be something that is very useful, also for some plugins.

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+    # Check if extracted recording id matches the current file
+    if features['metadata']['tags']['musicbrainz_trackid'] != file.metadata['musicbrainz_recordingid']:
+        tagger.log.debug("AcousticBrainz features recordingId do not match the file metadata: %s" % file.filename)
+        submission_callback(tagger, file, None, None, True)
+        return
+
+    featstr = json.dumps(features)
+
+    tagger.webservice.post(ACOUSTICBRAINZ_HOST,
```
zas, line 122: Reduce indentation of parameters by wrapping just after first parenthesis and just before last one:

```python
tagger.webservice.post(
    ACOUSTICBRAINZ_HOST,
    ACOUSTICBRAINZ_PORT,
    "/%s/low-level" % file.metadata["musicbrainz_recordingid"],
...
)
```

picard/tagger.py
```diff
@@ -185,7 +185,7 @@ def __init__(self, picard_args, unparsed_args, localedir, autoupdate):
             self.set_log_level(logging.DEBUG)
 
         # Default thread pool
-        self.thread_pool = ThreadPoolExecutor()
+        self.thread_pool = ThreadPoolExecutor(max_workers=max(1, os.cpu_count()-2))
```
zas, line 188: Why this change? Why this value? It deserves a comment IMHO.

According to https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.ThreadPoolExecutor

> Default value of max_workers is changed to min(32, os.cpu_count() + 4). This default value preserves at least 5 workers for I/O bound tasks. It utilizes at most 32 CPU cores for CPU bound tasks which release the GIL. And it avoids using very large resources implicitly on many-core machines.

po/picard.pot
```diff
@@ -8,14 +8,14 @@ msgid ""
 msgstr ""
 "Project-Id-Version: picard 2.7.0.dev3\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
-"POT-Creation-Date: 2021-06-29 11:58+0200\n"
+"POT-Creation-Date: 2021-06-30 20:54-0300\n"
```
zas, line 11: Please do not regenerate `picard.pot` within the PR (we'll do it, but once the PR will be merged)

picard/tagger.py
```diff
@@ -842,6 +842,66 @@ def finished(file, result):
             file.set_pending()
             self._acoustid.fingerprint(file, partial(finished, file))
 
+    def extract_and_submit_acousticbrainz_features(self, objs):
+        """Extract AcousticBrainz features and submit them."""
+
+        def abz_extractor_callback(file, result, error):
+            file.clear_pending()
+
+            from concurrent.futures import Future
```
zas, line 851: We usually do imports at top of the file (as recommended by PEP8).

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,130 @@
...
+
+ratecontrol.set_minimum_delay((ACOUSTICBRAINZ_HOST, ACOUSTICBRAINZ_PORT), 1000)
+
+acousticbrainz_extractor = find_executable("streaming_extractor_music")
+acousticbrainz_extractor_sha = precompute_extractor_sha(acousticbrainz_extractor)
+acousticbrainz_extractor_version = check_extractor_version(acousticbrainz_extractor)
+
+
+def run_extractor_entry(tagger, recording_id, input_path, output_path, extractor_callback):
+    tagger.webservice.get(ACOUSTICBRAINZ_HOST,
```
zas, line 51: Wrap after first parenthesis, it will give more width for parameters.

picard/tagger.py
```diff
@@ -842,6 +842,66 @@ def finished(file, result):
...
+
+                # Check if file was either already processed or sent to the AcousticBrainz server
+                if file.acousticbrainz_features_file:
+                    abz_extractor_callback(file, (file.acousticbrainz_features_file, 0, "Writing results"), False)
+                elif file.acousticbrainz_is_duplicate:
+                    abz_extractor_callback(file, (None, 0, "Duplicate"), False)
+                else:
+                    # Launch the acousticbrainz on a separate process
+                    log.debug("Extracting AcousticBrainz features from %s" % file.filename)
+                    run_extractor_entry(self,
```
zas, line 899: Same here, we usually prefer:

```python
verylongfunc(
    p1,
    p2,
    ....
)
```

Over:

```python
verylongfunc(p1,
             p2,
             ...)
```

Main reasons:
- it is easier to re-indent
- change in function name only affects one line
- change to a parameter only affects one line
- more space to write long parameter

zas (changes requested):

picard/acousticbrainz/__init__.py
Gabrielcarvfer: I don't think so. I saved the file before testing, and it happened with different files. I've seen at least 3 cases while testing yesterday, but only tested with a few hundred files.

picard/tagger.py
Gabrielcarvfer: I can leave it as is, but when a bunch of Essentia extractors are running, it drastically slows down the entire computer, not only Picard. Maybe switch to a different thread pool?

po/picard.pot
Gabrielcarvfer: OK o>

picard/acousticbrainz/__init__.py
phw: Found the probable reason: `235fa3a9-feb6-434c-8801-34370c63af50` is an old recording ID that got merged into `3e832e23-d08a-495f-82e2-8a3632947673`. 

But that explains the issue you describe only partially.

That means when Picard would load the release it would get definitely `3e832e23-d08a-495f-82e2-8a3632947673`, even if the file had  `235fa3a9-feb6-434c-8801-34370c63af50` in the tags. But then you would see a change from old value to new value. After saving the file should have only the new value in the tags.

But maybe the file has this stored in some non-standard way? What file format is it? For ID3 tags the MusicBrainz default is to store the recording ID in a frame `UFID` with and owner identifier of "http://musicbrainz.org". But e.g. the MusicBrainz foobar2000 plugin uses `TXXX:musicbrainz_trackid ` instead.

Gabrielcarvfer: Uh, found another problem. The extracted features have the values inside lists... So a few of the failures were due to that...

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Both are fine for me. Changes are minimal.
https://github.com/metabrainz/picard/pull/1854/commits/052c04b62dfa14e7e2a7e5fd6d7e2384cb055691

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Done.

picard/tagger.py
Gabrielcarvfer: Should be better now. :)

phw:
Regarding the logos: I think what you created is fine. But I would probably use the simple version of the logo instead:

https://github.com/metabrainz/metabrainz-logos/blob/master/logos/AcousticBrainz/SVG/AcousticBrainz_logo_icon.svg

For the @2x variants the detailed one actually might be fine, though. At least for the 22x22, which would actually be 44x44 then.

There used to be also medium detail variants of each logo. For some reason they god removed from metabrainz-logos, but e.g. Picard still uses it for some icon size. The medium detail logo for AB would be:

https://github.com/metabrainz/metabrainz-logos/blob/351925036dbaf78095a1f6de1cfed18247b928c2/final-logos/AcousticBrainz/PIXEL/PNG/AcousticBrainz_logo_small_notext.png

I think it would work better then the full detail logo for our use case.

So my suggestion: Try with either the no details or mid details logo for the normal resolution. For the high res @2x variants we could go higher, but probably should test on a high res display (which I can do with some effort).

zas (changes requested):

picard/acousticbrainz/extractor.py
```diff
@@ -0,0 +1,56 @@
...
+    ret = subprocess.run([essentia_path], stdout=subprocess.PIPE)
+
+    # filters the useful line
+    # 'built with Essentia version v2.1_beta2-1-ge3940c0\r'
+    out = ret.stdout.decode(sys.stdout.encoding).split('\n')[-2]
+
+    # removes carrier return (if it is there)
+    # 'built with Essentia version v2.1_beta2-1-ge3940c0\r'
+    cr = out[-1] == '\r'
+    out = out[:-cr]
```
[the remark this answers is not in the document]
Gabrielcarvfer: Switching to 
```
version_regex = re.compile(r"Essentia version (.*[^ \r\n])")
version = version_regex.findall(out)[0]
```

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,139 @@
...
+    process.waitForFinished(msecs=-1)  # Do not timeout
+    results = (output_file.name, process.exitCode(), process.readAllStandardOutput().data().decode('utf-8'))
+
+    # Add feature extractor sha to the output file
+    try:
+        with open(output_file.name, "r", encoding="utf-8") as f:
+            features = json.load(f)
+        features["metadata"]["version"]["essentia_build_sha"] = acousticbrainz_extractor_sha
+        with open(output_file.name, "w", encoding="utf-8") as f:
+            json.dump(features, f, indent=4)
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

phw:
@Gabrielcarvfer The new icons work quite well. The simple shape looks really cleaner on my normal display. And here is a screenshot from a high DPI display, again I think the mid detail icon is a very good choice here.

grafik

Just not sure about the button label. "Submit AcousticBrainz features" is really very long, but every shorter version I can think of looses clarity. But maybe we can get away with something like "AcousticBrainz analysis" or "AscousticBrainz submission"?

Gabrielcarvfer:
Yes, I wasn't sure how to shorten the label. Thought of adding a second line, but it looks completely out of place. 

"Submit AB features" isn't very clear for people who don't know what AB means.
"Submit Acoustic features" is the best I can think of, but people may get confused with AcoustID.
"AcousticBrainz analysis" makes me think it's going to give me some interesting high-level features, but it's basically the opposite of what is happening: extracting low level features for others to work with.

phw:
I fully agree with you here and have no good idea on how to solve it :(

But probably same as with the "Generate AcoustId fingerprints" action. I think you should add the AB feature extraction to the Tools menu, and that's were it is available by default. The button is disabled by default as it is now, and if someone wants to have it in the toolbar they can add it. And after all there is also the option to hide the labels.

Gabrielcarvfer:
I've removed the Brainz part and it doesn't look that bad (side-by-side logos help, at least for me)
image

> I think you should add the AB feature extraction to the Tools menu, and that's were it is available by default. 

Done.

picard/tagger.py
zas: I see no problem in changing it, but please add a comment explaining how the new value was chosen

picard/acousticbrainz/extractor.py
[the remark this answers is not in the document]
zas: Perhaps just catch exception in case there's no match at all.
For example, if text is 'Essentia Version xxx', there will be no index 0 (empty list)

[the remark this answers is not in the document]
Gabrielcarvfer: Yup, now it should be properly dealt with and logged. :)

Gabrielcarvfer:
Question, can I add an ETA field on the statusbar? :D 

Just need to plug seconds to days, hours, minutes, seconds and print it.

```
            seconds = max(eta_seconds, 0)
            days = int(seconds/86400)
            seconds = seconds - days*86400
            hours = int(seconds/3600)
            seconds = seconds - hours*3600
            minutes = int(seconds/60)
            seconds = seconds - minutes*60
            estimated_remaining_time = ("%.dd:%dh:%dm:%ds" % (days, hours, minutes, seconds))
            print(estimated_remaining_time)
```

Gabrielcarvfer:
\*-\*
image

zas:
> Question, can I add an ETA field on the statusbar? :D
> Or should I stop with the feature creep?

It would be better to do it in another PR (and create a matching ticket for changelog).
+1 for the feature

Gabrielcarvfer:
Hmm, having a hard time thinking of icons to represent features extraction state. I thought of playing with AB logo colors for grey/yellow/blue/green/red, but for now I'm reusing the FileItem.icon_file_pending, FileItem.icon_saved and TrackItem.icon_error.

zas (changes requested):
Please add tests.

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,162 @@
...
+
+
+def extractor(input_path):
+    # Create a temporary file with AcousticBrainz output
+    output_file = NamedTemporaryFile("w", suffix=".json")
+    output_file.close()  # close file to ensure other processes can write to it
+
+    # Call the features extractor and wait for it to finish
+    process = QtCore.QProcess()
+    process.setProperty('picard_finished', False)
```
phw, line 106: We need this only if we plan to read it later. I think the reason why this is used when calling fpcalc is that it registers both finished and error signals, and looks like there are cases where both get called.

I wonder whether it would be a better alternative here instead of `run_task` only use the `QProcess` with finished and error signals connected? After all we spawn a separate process anyway, so not really a need to keep a thread waiting. Or do I miss something?

phw (changes requested):

phw:
> BTW, the portable built seems to be borked with the built-in extractor. Didn't really look into it because I'm not sure how we are proceeding with the extractor anyways.

I think we should handle it directly as we do with fpcalc: Bundle it with the installer versions (the size impact is not as bad as I had originally feared), but make it configurable in options to set a path. Empty path means auto detecting its location (which is important for some package setups). That way this also works well also in installs like on Linux where the extractor comes separate. Options should of course provide a link to download the binary.

grafik

Download can open https://acousticbrainz.org/download

Important is that Picard must start and run without a valid extractor binary. The portable installer currently seems to crash Picard on start when running the exe in `check_extractor_version`. Not sure why exactly, maybe it is trying to run it too early during startup. The streaming extractor should get extracted with the rest of the app to `~/AppData/Local/Temp/` into a unique directory (`_MEI*`), same as fpcalc. I tested this, and in general the streaming_extractor_music.exe is there and callable.

> I thought of playing with AB logo colors for grey/yellow/blue/green/red, but for now I'm reusing the FileItem.icon_file_pending, FileItem.icon_saved and TrackItem.icon_error.

I'm not exactly great with icons myself. I tested this, and for me this would work until someone comes around with something better.

Gabrielcarvfer:
Working on the tests. Taking quite some time due to the callbacks going back and forth. =x

Gabrielcarvfer:
Question: can I add a song to the test/data folder? If I can, what kind of song should I add (public domain?) I need something to test the feature extraction.

phw:
> can I add a song to the test/data folder? If I can, what kind of song should I add (public domain?)

I think we could add a test file. But indeed it's content should be in the public domain or at least with a permissive enough license so we can include and distribute it with the source. Also it should be small.

But is it really needed? After all we don't need to test `streaming_extractor_music` really. It also shouldn't be required to have this binary in order to run the test suite. Maybe instead add a small script as a mock replacement for streaming_extractor_music that returns fake results?

Gabrielcarvfer:
> But is it really needed? After all we don't need to test `streaming_extractor_music` really. It also shouldn't be required to have this binary in order to run the test suite. Maybe instead add a small script as a mock replacement for streaming_extractor_music that returns fake results?

Derp, I didn't think of that.  Yup, a mock script should be enough.

Gabrielcarvfer:
I have a few issues with QProcess and relative paths to the mock extractor for the test, and the thread_pool thing from tagger interacting very weirdly with the tests (I guess that is why I didn't found any test using it). 

Ended up running things on the main thread and blocking it while waiting for the extraction. Still going to fix it, but wanted to share the tests earlier.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Ended up going the `run_executable` route.

Gabrielcarvfer:
Hopefully fixed the weird error causing random test failures.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Ended up using subprocess.run instead of QProcesses, but we could do that if I manage to discover why it is failing to run the mock extractor.

picard/tagger.py
Gabrielcarvfer: Ended up removing that. Not really sure if we will have any problem in the real world, as extraction is currently waiting the API checking for duplicates before extracting anything. Kinda slows down the entire process, but also doesn't hog the computer extracting unnecessary stuff nor updates unnecessary data.

Gabrielcarvfer:
About the codacy warnings: 
- SHA1 is used by the OG AB-client (https://github.com/MTG/acousticbrainz-client/blob/master/abz/config.py#L115)
- securing the subprocess.run is just as hard as QProcess. The only thing that comes to my mind is checking whatever is going to be executed against a list of known hashes (just like the MSFT SmartScreen thing)

Gabrielcarvfer:
My locally built portable version works, but the CI one fails to run the portable hook. (ノಠ益ಠ)ノ彡┻━┻

picard/ui/itemviews.py
```diff
@@ -409,18 +412,23 @@ def paintSection(self, painter, rect, index):
             super().paintSection(painter, rect, index)
             painter.restore()
             paint_fingerprint_icon(painter, rect, FileItem.icon_fingerprint_gray)
+        elif index == MainPanel.ACOUSTICBRAINZ_COLUMN:
+            painter.save()
+            super().paintSection(painter, rect, index)
+            painter.restore()
+            paint_fingerprint_icon(painter, rect, FileItem.icon_acousticbrainz)
```
phw, line 419: I think we should rename this helper function to "paint_column_icon"

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,147 @@
...
+    TITLE = N_("AcousticBrainz")
+    PARENT = None
+    SORT_ORDER = 45
+    ACTIVE = True
+    HELP_URL = '/config/options_acousticbrainz.html'
+
+    options = [
+        BoolOption("setting", "use_acousticbrainz", False),
+        TextOption("setting", "acousticbrainz_extractor", ""),
+        TextOption("setting", "acousticbrainz_extractor_version", ""),
```
phw, line 65: Do we need this as an option? It only seems to get set, not read. The only place the version is display is in options, and there the extractor is run before to get the version anyway
.

picard/util/__init__.py
```diff
@@ -290,19 +291,60 @@ def find_existing_path(path):
     return decode_filename(path)
 
 
+def _add_windows_executable_extension(*executables):
+    processed_executables = []
+    for e in executables:
+        if ".py" in e:
+            processed_executables.append(e)
+            continue
+        if ".exe" in e:
```
phw, line 300: Rather use `e.endswith('.py')` and `e.endswith('.exe')`?

picard/acousticbrainz/extractor.py
```diff
@@ -0,0 +1,54 @@
...
+
+import hashlib
+import re
+
+from picard.util import run_executable
+
+
+def precompute_extractor_sha(essentia_path):
+    # Precompute extractor sha1
+    h = hashlib.sha1()
```
phw, line 29: As this has no security implication mark it with `# nosec` to get rid of the warning:

    h = hashlib.sha1()  # nosec

picard/ui/options/acousticbrainz.py
Gabrielcarvfer: The version? Not really needed, unless we want to update AB features that were already uploaded by an older extractor. Not sure if it is the case, so included it anyways.

picard/util/__init__.py
Gabrielcarvfer: Way better. :D

picard/acousticbrainz/extractor.py
Gabrielcarvfer: Ok

zas:
@Gabrielcarvfer : can you fix or silent Codacy issues?

Gabrielcarvfer:
> @Gabrielcarvfer : can you fix or silent Codacy issues?

Done, but I think the portable version from the CI still borked.

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,147 @@
...
+            self.ui.disable_acoustic_features.setChecked(True)
+        self.ui.acousticbrainz_extractor.setPlaceholderText("streaming_extractor_music")
+        self.ui.acousticbrainz_extractor.setText(self._config.setting["acousticbrainz_extractor"])
+        self.update_groupboxes()
+
+    def save(self):
+        if self.ui.use_acoustic_features.isChecked():
+            self._config.setting["use_acousticbrainz"] = True
+        else:
+            self._config.setting["use_acousticbrainz"] = False
```
zas, line 96: It can be simplified to:

```python
self._config.setting["use_acousticbrainz"] = self.ui.use_acoustic_features.isChecked()
```

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,147 @@
...
+        self._config.setting["acousticbrainz_extractor"] = self.ui.acousticbrainz_extractor.text()
+
+    def update_groupboxes(self):
+        if self.ui.use_acoustic_features.isChecked():
+            self.ui.acousticbrainz_settings.setEnabled(True)
+            # If current path to extractor isn't valid, look for an extractor
+            if not os.path.exists(self.ui.acousticbrainz_extractor.text()):
+                self.ui.acousticbrainz_extractor.setText(get_extractor())
+        else:
+            self.ui.acousticbrainz_settings.setEnabled(False)
```
zas, line 106: It could be shorten a bit:

```python
        enabled = self.ui.use_acoustic_features.isChecked()
        self.ui.acousticbrainz_settings.setEnabled(enabled)
        if enabled and not os.path.exists(self.ui.acousticbrainz_extractor.text()):
            # If current path to extractor isn't valid, look for an extractor
            self.ui.acousticbrainz_extractor.setText(get_extractor())
        self._acousticbrainz_extractor_check()
```

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,147 @@
...
+        self._acousticbrainz_extractor_check()
+
+    def acousticbrainz_extractor_browse(self):
+        path, _filter = QtWidgets.QFileDialog.getOpenFileName(self, "", self.ui.acousticbrainz_extractor.text())
+        if path:
+            path = os.path.normpath(path)
+            self.ui.acousticbrainz_extractor.setText(path)
+
+    def acousticbrainz_extractor_download(self):
+        webbrowser2.open('https://acousticbrainz.org/download')
```
zas, line 116: Please use a constant stored in `picard/const/__init__.py` for url.

picard/util/__init__.py
```diff
@@ -301,19 +302,57 @@ def find_existing_path(path):
     return decode_filename(path)
 
 
+def _add_windows_executable_extension(*executables):
+    processed_executables = []
+    for e in executables:
+        if e.endswith(".py") or e.endswith(".exe"):
+            processed_executables.append(e)
+            continue
+        processed_executables.append(e + ".exe")
```
zas, line 311: We could use a list comprehension and make it a one-line function:

```python
def _add_windows_executable_extension(*executables):
    return [e if e.endswith(('.py', '.exe')) else e + '.exe' for e in executables]
```

test/test_acousticbrainz.py
```diff
@@ -0,0 +1,206 @@
...
+        handler("""APIBadRequest""", "", 400)
+
+    def test_submit_failed(self):
+        self.file.set_pending()
+        self.tagger.webservice.post = Mock(wraps=self.mock_post_fail)
+        ab_feature_extraction(self.tagger,
+                              self.file.metadata["musicbrainz_recordingid"],
+                              "test.mp3",
+                              partial(ab_extractor_callback, self.tagger, self.file)
+                              )
```
zas, line 197: We usually split a bit differently, first argument is moved to next line and indented in order to reclaim room on right side:

```python
ab_feature_extraction(
    self.tagger,
    self.file.metadata["musicbrainz_recordingid"],
    "test.mp3",
    partial(ab_extractor_callback, self.tagger, self.file)
)
```

test/test_acousticbrainz.py
```diff
@@ -0,0 +1,206 @@
...
+
+    def test_submit_succeeded(self):
+        self.file.set_pending()
+        self.tagger.webservice.post = Mock(wraps=self.mock_post_succeed)
+        ab_feature_extraction(self.tagger,
```
zas, line 202: See indentation comment above

picard/ui/options/acousticbrainz.py
zas: Instead we could add the version to the feature files.
What's the purpose of `acousticbrainz_extractor_sha` (which is added to feature files)?

zas (changes requested):

picard/ui/options/acousticbrainz.py
Gabrielcarvfer: We already add them to the feature files. Not completely sure, but the original client does that. 
I believe the server can use it to treat extracted features with the same mbid differently (either keeping both or discarding the oldest). I'm using it not to send an already existing version of the features to the server.

picard/acousticbrainz/__init__.py
```diff
@@ -71,14 +74,15 @@ def ab_setup_extractor(tagger, config=None):
         version = check_extractor_version(acousticbrainz_extractor, tagger)
         if version:
             config.setting["use_acousticbrainz"] = True
-            config.setting["acousticbrainz_extractor_version"] = version
-            config.setting["acousticbrainz_extractor_sha"] = sha
+            ACOUSTICBRAINZ_EXTRACTOR_VERSION = version
+            ACOUSTICBRAINZ_EXTRACTOR_SHA = sha
             tagger.log.debug("AcousticBrainz is available")
-            return
+            return version
```
zas, line 80: Since version is returned, it seems to me `ACOUSTICBRAINZ_EXTRACTOR_VERSION` isn't needed.

picard/acousticbrainz/__init__.py
```diff
@@ -71,14 +74,15 @@ def ab_setup_extractor(tagger, config=None):
         version = check_extractor_version(acousticbrainz_extractor, tagger)
         if version:
             config.setting["use_acousticbrainz"] = True
-            config.setting["acousticbrainz_extractor_version"] = version
-            config.setting["acousticbrainz_extractor_sha"] = sha
+            ACOUSTICBRAINZ_EXTRACTOR_VERSION = version
+            ACOUSTICBRAINZ_EXTRACTOR_SHA = sha
```
zas, line 78: Since `ACOUSTICBRAINZ_EXTRACTOR_SHA` is dynamically set, please use lowercased name and since it's a global for internal use (I don't we meant to export it outside this module), underscorde prefix makes sense: `_extractor_sha`

zas (changes requested):

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Yup, removing it.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Done. o>

rdswift (commented):
I tested the portable version on my Windows 7 system, and it seemed to work as expected, once I got past the Option setting problem below.  Unfortunately, I don't have any files that I haven't already submitted, but the logs did correctly show that the information had already been submitted when I tried.

One thing that I did notice that seems to be a bug, is that the first time I open the Options page to the AcousticBrainz page after starting Picard, it doesn't show the option as enabled (as had been saved in the options on the previous Picard session).  When I enable the option, the path to the extractor exe file is cleared.  If I cancel and close the Options dialog and then re-open it, my previously saved settings are back including the full path to the extractor.  If I simply cancel without making a change to the enabled status first, the next time I open the Options it still shows as disabled.

Also, until I go through that process of opening Options, enabling the AcousticBrainz settings and then canceling the change, the extractor remains disabled in the Tools menu.

rdswift:
Also, is this an option that we want to be managed using Options Profiles?  If so, we need to add a new `SETTINGS_GROUP` dictionary item for the option page in `picard/profile.py`.  I can post the necessary code changes if you like.

Gabrielcarvfer:
> The first time I open the Options page to the AcousticBrainz page after starting Picard, it doesn't show the option as enabled (as had been saved in the options on the previous Picard session). 

Starting disabled is the expected behavior. Not restoring the previous setting is not. 

> Also, until I go through that process of opening Options, enabling the AcousticBrainz settings and then canceling the change, the extractor remains disabled in the Tools menu.

Saving the setting with the cancel button certainly wasn't expected. Working on a fix xD

rdswift:
> > The first time I open the Options page to the AcousticBrainz page after starting Picard, it doesn't show the option as enabled (as had been saved in the options on the previous Picard session).
> 
> Starting disabled is the expected behavior. Not restoring the previous setting is not.
> 
> > Also, until I go through that process of opening Options, enabling the AcousticBrainz settings and then canceling the change, the extractor remains disabled in the Tools menu.
> 
> Saving the setting with the cancel button certainly wasn't expected. Working on a fix xD

Sorry, I guess I wasn't as clear as I should have been.  The first time Picard is started with this new option, the settings were indeed disabled as they should be.  I enabled the setting and saved the changes with "Make it so!".  Then closed Picard.  On the next run of Picard, that's when the the problem that I described shows up.

Gabrielcarvfer:
Oh lord... Stupid IDE converted everything to LF instead of CRLF >.>

rdswift:
This is better because it at least remembers that I have the option enabled, but it clears the saved path to the executable when the options page is loaded and displays the "Please select a valid extractor executable." error message.  The path data still remains in the Config.ini file, but it seems to be getting cleared on the Options page.

I think that the problem is that it stores the temporary path to the Config.ini file when you "Make it so!", but that path is deleted when the Picard session ends, and a different temporary path is created the next time you start Picard (portable) thus rendering the path stored in the config file invalid.  When I save the path to the version that I have installed in my `C:\Program Files\AcousticBrainz\` directory, everything works as expected and is properly retained between sessions.

I wonder if a workaround might be to check the stored path and if it begins with `C:\Users\{userid}\AppData\Local\Temp\` replace it with the new temporary path created for the session?  It's a bit of a hack, but the only other thing I could suggest is to copy the exe file to the same directory as the Config.ini file (overwriting if it already exists) at the start of the Picard (portable) session, and then default to that location.  At least that would be stable between sessions.  I believe that would only be required for the portable version, since the installed version should have a stable location for the files.

Gabrielcarvfer:
> I think that the problem is that it stores the temporary path to the extractor executable in the Config.ini file when you "Make it so!", but that path is deleted when the Picard session ends, and a different temporary path is created the next time you start Picard (portable) thus rendering the path stored in the config file invalid. When I save the path to the version that I have installed in my `C:\Program Files\AcousticBrainz\` directory, everything works as expected and is properly retained between sessions.

Hmm, then it is failing to find the extractor.

> I wonder if a workaround might be to check the stored path and if it begins with `C:\Users\{userid}\AppData\Local\Temp\` replace it with the new temporary path created for the session? It's a bit of a hack

I think I can just look for the configured path, and if it doesn't exist, just search again. Kind of defeats the purpose of setting the extractor, but whatever.

Gabrielcarvfer:
Hmm, going to work a bit more.

rdswift:
> Hmm, is the _MEIPASS included in the PATH? If not, it won't work at all.

The path on my system is `C:\Users\bob\AppData\Local\Temp\_MEI197482\streaming_extractor_music.exe` and with each new session of Picard, the `_MEI197482` portion changes.

Gabrielcarvfer:
> > Hmm, is the _MEIPASS included in the PATH? If not, it won't work at all.
> 
> The path on my system is `C:\Users\bob\AppData\Local\Temp\_MEI197482\streaming_extractor_music.exe` and with each new session of Picard, the `_MEI197482` portion changes.

I think it is included, or fpcalc would not be found. I thought on falling back to get_extractor but it broke the tests. :x

rdswift:
> I think it is included, or fpcalc would not be found. I thought on falling back to get_extractor but it broke the tests. :x

It appears that Picard allows the path to fpcalc to be left blank when enabling that option and then it uses the one in the temporary directory if the option setting is empty.  Would that work for the streaming_extractor_music?  I mean if the path was allowed to be blank, and then just use the executable in the temporary directory?

Gabrielcarvfer:
Would that work for the streaming_extractor_music? I mean if the path was allowed to be blank, and then just use the executable in the temporary directory?

I've tried doing that, but it is complicated. Would you mind testing it again? Made a few changes and I hope it works now. It will automatically detect the extractor, but won't change the default config until the user clicks to confirm it.

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+def ab_available():
+    return _acousticbrainz_extractor_sha is not None
+
+
+def ab_check_version(extractor, tagger):
+    extractor_path = find_executable(extractor)
+    return check_extractor_version(extractor_path, tagger)
+
+
+def ab_setup_extractor(tagger, config=None):
```
phw, line 74: `tagger` is not actually needed by this function. It is only used for `tagger.log` calls, but the usual way to use the logging is by importing `picard.log` directly:

```
from picard import log

log.debug('foo')
```

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+            # Run extractor on a different thread and call the callback when done
+            run_task(partial(extractor, input_path), extractor_callback)
+
+
+def extractor(input_path):
+    config = get_config()
+
+    # Create a temporary file with AcousticBrainz output
+    output_file = NamedTemporaryFile("w", suffix=".json")
+    output_file.close()  # close file to ensure other processes can write to it
```
phw, line 165: Does the temp file get removed somewhere? I think we should clean it up, or does anything speak against that?

picard/cluster.py
```diff
@@ -227,6 +227,9 @@ def can_view_info(self):
     def can_submit(self):
         return not self.special and self.files
 
+    def can_extract(self):
```
phw, line 230: Like other of those feature properties I think `can_extract should be defined for the `Item` class (in `ui/item.py`). This avoids that you need to declare it in all those item subclasses.

picard/tagger.py
```diff
@@ -842,6 +851,34 @@ def finished(file, result):
...
 
+    def extract_and_submit_acousticbrainz_features(self, objs):
+        """Extract AcousticBrainz features and submit them."""
+        config = get_config()
+        if not config.setting["use_acousticbrainz"]:
+            return
+
+        for file in iter_files_from_objects(objs):
+            # Skip unmatched files
+            if not file.metadata["musicbrainz_recordingid"]:
```
phw, line 862: `if not file.can_extract()`?

picard/track.py
```diff
@@ -235,6 +235,13 @@ def can_edit_tags(self):
     def can_view_info(self):
         return self.num_linked_files == 1 or self.metadata.images
 
+    def can_extract(self):
+        """Return if this object has a linked file with a recordingId, required by the AcosuticBrainz feature extractor."""
+        for file in self.files:
+            if file.metadata["musicbrainz_recordingid"]:
```
phw, line 241: Maybe `if file.can_extract():` ?

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,146 @@
...
+        webbrowser2.open(ACOUSTICBRAINZ_DOWNLOAD_URL)
+
+    def _acousticbrainz_extractor_check(self):
+        if not self.ui.use_acoustic_features.isChecked():
+            self._acousticbrainz_extractor_set_success("")
+            return
+
+        if not os.path.exists(self.ui.acousticbrainz_extractor.text()):
+            # If current path to extractor isn't valid, look for an extractor
+            self.ui.acousticbrainz_extractor.setText(find_extractor())
```
phw, line 120: This currently makes it impossible to manually change the path, as it immediately tries to correct it and guess it.

> I think I can just look for the configured path, and if it doesn't exist, just search again. Kind of defeats the purpose of setting the extractor, but whatever.

The issue is that for the portable version the path changes on each start. Similar issue happens with the Snap version, there the path changes with each app update.

For AcoustID we hence went with the solution that it defaults to an empty configured path, and when the path is empty it always gets guessed on invocation. See my changes in https://github.com/metabrainz/picard/pull/1724 . I would suggest to make the AB extractor detection work the same way. So instead of saving the auto detection to the config, allow an empty path and if it is empty auto detect when actually using the feature.

picard/util/time.py
```diff
@@ -1,49 +1,49 @@
-# -*- coding: utf-8 -*-
```
phw, line 1: I committed the line ending fixes separately in https://github.com/metabrainz/picard/commit/28c6118bedfefff11f1b3defbb56d753c8aed5a9

phw (changes requested):
Overall I'm happy to merge this, just some minor comments. But functionality and behavior is great, thanks a lot.

The biggest thing is handling of the path detection, what Gabriel and Rob have already started discussing above. Because for portable and snap version things are a but tricky, see my comment. But I'd be fine with merging the rest first and then do fixes to the path handling, if that makes it easier. @Gabrielcarvfer However you prefer.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Technically this deletes the file right away, but I wanted a safe temporary filename to use. They won't be automatically deleted, just like the coverart files. =x

phw: Got it, so it actually gets really created later.

Cover art actually get deleted when Picard exits, see https://github.com/metabrainz/picard/blob/master/picard/coverart/image.py#L72 .

But probably we could remove these files directly after they have been used? The data is not so much, but potentially a lot of files will be created?

picard/acousticbrainz/__init__.py
```diff
@@ -0,0 +1,221 @@
...
+        config = get_config()
+    if config.setting["use_acousticbrainz"]:
+        acousticbrainz_extractor = get_extractor(config)
+        tagger.log.debug("Checking up AcousticBrainz availability")
+        if acousticbrainz_extractor:
+            sha = precompute_extractor_sha(acousticbrainz_extractor)
+            version = check_extractor_version(acousticbrainz_extractor, tagger)
+            if version:
+                _acousticbrainz_extractor_sha = sha
+                tagger.log.debug("AcousticBrainz is available")
```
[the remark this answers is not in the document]
Gabrielcarvfer: Done.

picard/acousticbrainz/__init__.py
Gabrielcarvfer: Done.

picard/cluster.py
Gabrielcarvfer: Done.

picard/tagger.py
Gabrielcarvfer: That is a pretty good question.

picard/track.py
Gabrielcarvfer: Yes. Leftover from when I assumed only files associated with tracks would get extracted, which is true, so I didn't include the file.can_extract. =x

picard/ui/options/acousticbrainz.py
Gabrielcarvfer: > So instead of saving the auto detection to the config, allow an empty path and if it is empty auto detect when actually using the feature.

Yup, it works much better doing that. :D
I really didn't want to keep the field empty (which in is pretty of dumb, since there is a placeholder anyways...).

Gabrielcarvfer:
> The biggest thing is handling of the path detection

I think it is solved now. 

> I'd be fine with merging the rest first and then do fixes to the path handling, if that makes it easier. 

IMHO, I'd prefer to fix everything that can be fixed right away, then squash and merge.
BTW, I didn't do the signing part on the extractors. Have no idea how this is handled.

picard/ui/options/acousticbrainz.py
phw: We struggled with this for the fpcalc case as well. The original behavior that was implemented was that Picard would detect when it was running packaged by PyInstaller and if that was the case it auto detected on startup every time. But this had the downside that you could not intentionally configure an alternative path without loosing it. It also was unnecessary most of the time for macOS and Windows installer case and did not capture the snap case.

With "empty = autodetection" this actually now works well, and gives a clear option to have it auto detect (which probably covers most users needs anyway). I'd just make sure to not run  the auto detection too often. I think the fpcalc case only does auto detection once when you click scan.

phw:
Yes, path handling seems to work well now.

> BTW, I didn't do the signing part on the extractors. Have no idea how this is handled.

Currently the macOS build process resigns everything with the MetaBrainz signature.

I considered adding checksums to the build process for fpcalc (and then also for the fingerprinter), so we only sign binaries we trust. That's probably something we should add at some point.

Side note: Originally we kept existing signatures and only signed unsigned components, but recent Python releases caused signing issues, so we actually remove their signature and do our own. Also something to maybe check again.

rdswift:
> > The biggest thing is handling of the path detection
> 
> I think it is solved now.

I just tried the portable version from the latest build and it seems to work well when I leave the extractor program path blank.  :+1: 

The only thing (very minor) is that it no longer displays the default path to the extractor that will be used, unlike the display for the fpcalc path which shows the actual file path.  It would be nice if the information displayed could be consistent.  See images below.

image

image

picard/acousticbrainz/__init__.py
Gabrielcarvfer: I've seen this was there, but my temp folder always had tons of leftovers, so I assumed it didn't exactly work as intended. Testing it right now and it does work just fine, I was the one killing it with the IDE. Maybe an additional signal handler? =x

Gabrielcarvfer:
Do not merge yet, found another issue introduced with the last changes.

Gabrielcarvfer:
> The only thing (very minor) is that it no longer displays the default path to the extractor that will be used, unlike the display for the fpcalc path which shows the actual file path. It would be nice if the information displayed could be consistent. See images below.
> 
> image
> 
> image

Hmm, no idea what should be done here. I mean, I can check if there is a configured extractor, search for it, if it fails fallback to automatic search, and if it fails put the name.

rdswift:
> Hmm, no idea what should be done here. I mean, I can check if there is a configured extractor, search for it, if it fails fallback to automatic search, and if it fails put the name. Feels kind of unnecessary though.

In one of your earlier revisions, you were actually showing the full path.  Perhaps this could be done only if the textbox is blank?  I suspect that you're already doing the lookup anyway to provide the version information that is displayed below the textbox.

Gabrielcarvfer:
> Perhaps this could be done only if the textbox is blank? I suspect that you're already doing the lookup anyway to provide the version information that is displayed below the textbox.

Done.

rdswift:
> Done.

Excellent!  Thank-you, sir.

phw (dismissed):
@Gabrielcarvfer I replaced some more usages of tagger.log in https://github.com/metabrainz/picard/pull/1854/commits/2614aae119ea24df648b9a8acdcd0b4df10b5355 , could you double check I did not break something?

Commits should be squashed a bit, but otherwise I think we are good to merge. Good work, thanks a lot!

picard/tagger.py
```diff
@@ -842,6 +852,33 @@ def finished(file, result):
...
+                # Check if file was either already processed or sent to the AcousticBrainz server
+                if file.acousticbrainz_features_file:
+                    ab_extractor_callback(self, file, (file.acousticbrainz_features_file, 0, "Writing results"), False)
+                elif file.acousticbrainz_is_duplicate:
+                    ab_extractor_callback(self, file, (None, 0, "Duplicate"), False)
+                else:
+                    file.acousticbrainz_error = False
+                    # Launch the acousticbrainz on a separate process
+                    log.debug("Extracting AcousticBrainz features from %s" % file.filename)
+                    ab_feature_extraction(self,
```
zas, line 877:
```python
ab_feature_extraction(
    self,
    ...
)
```

picard/ui/itemviews.py
```diff
@@ -359,7 +362,7 @@ def show_column(self, column, show):
             if self.sectionSize(column) == 0:
                 self.resizeSection(column, self.defaultSectionSize())
             self._visible_columns.add(column)
-            if column == MainPanel.FINGERPRINT_COLUMN:
+            if column in [MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN]:
```
zas, line 365: Please use set instead of list, it doesn't make much difference in practice, but `in` set is faster than `in` list:

```suggestion
            if column in {MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN}:
```

picard/ui/itemviews.py
```diff
@@ -408,19 +411,24 @@ def paintSection(self, painter, rect, index):
...
+            painter.save()
+            super().paintSection(painter, rect, index)
+            painter.restore()
+            paint_column_icon(painter, rect, FileItem.icon_acousticbrainz)
         else:
             super().paintSection(painter, rect, index)
 
     def on_sort_indicator_changed(self, index, order):
-        if index == MainPanel.FINGERPRINT_COLUMN:
+        if index in [MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN]:
```
zas, line 424:
```suggestion
        if index in {MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN}:
```

picard/ui/itemviews.py
```diff
@@ -408,19 +411,24 @@ def paintSection(self, painter, rect, index):
...
             self.setSortIndicator(-1, QtCore.Qt.AscendingOrder)
 
     def lock(self, is_locked):
         super().lock(is_locked)
-        fingerprint_column_index = MainPanel.FINGERPRINT_COLUMN
-        if not self.is_locked and self.count() > fingerprint_column_index:
-            self.setSectionResizeMode(fingerprint_column_index, QtWidgets.QHeaderView.Fixed)
+        for column_index in [MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN]:
```
zas, line 429:
```suggestion
        for column_index in {MainPanel.FINGERPRINT_COLUMN, MainPanel.ACOUSTICBRAINZ_COLUMN}:
```

picard/ui/itemviews.py
```diff
@@ -432,7 +440,7 @@ def __init__(self, window, parent=None):
         self.panel = parent
         # Should multiple files dropped be assigned to tracks sequentially?
         self._move_to_multi_tracks = True
-        self.setHeaderLabels([_(h) if n != '~fingerprint' else ''
+        self.setHeaderLabels([_(h) if n not in ['~fingerprint', '~acousticbrainz'] else ''
```
zas, line 443:
```suggestion
        self.setHeaderLabels([_(h) if n not in {'~fingerprint', '~acousticbrainz'} else ''
```

picard/ui/itemviews.py
```diff
@@ -1158,3 +1175,24 @@ def decide_fingerprint_icon_info(file):
             icon = QtGui.QIcon()
             tooltip = _('No fingerprint was calculated for this file, use "Scan" or "Generate AcoustID Fingerprints" to calculate the fingerprint.')
         return (icon, tooltip)
+
+    @staticmethod
+    def decide_ab_icon_info(file):
+        if file.acousticbrainz_error:
+            icon = TrackItem.icon_error
+            tooltip = _('Extraction or submission of Acoustic features failed')
+            return (icon, tooltip)
```
zas, line 1184: Since the method ends with `return (icon, tooltip)`, what about:

```python
        if file.acousticbrainz_error:
            icon = TrackItem.icon_error
            tooltip = _('Extraction or submission of Acoustic features failed')
        elif file.acousticbrainz_is_duplicate:
```

picard/ui/options/acousticbrainz.py
```diff
@@ -0,0 +1,159 @@
...
+    def acousticbrainz_extractor_download(self):
+        webbrowser2.open(ACOUSTICBRAINZ_DOWNLOAD_URL)
+
+    def _acousticbrainz_extractor_check(self):
+        if not self.ui.use_acoustic_features.isChecked():
+            self._acousticbrainz_extractor_set_success("")
+            return
+
+        extractor_path = self.ui.acousticbrainz_extractor.text()
+        if extractor_path == "":
```
zas, line 131:
```suggestion
        if not extractor_path:
```

zas (changes requested):
Almost ready to merge, good job!

picard/ui/itemviews.py
Gabrielcarvfer: Done.

picard/tagger.py
Gabrielcarvfer: Should I do the same with calls above?

Gabrielcarvfer:
Squashed everything together, but still have the full history if you prefer to split this into multiple commits.

picard/tagger.py
zas: As you wish, I think it will be better as it will shorten lines a bit

Gabrielcarvfer: The error parameter is required by the run_task thing, but AFAIK it is more for failing to run a task than giving a bad result. At least it sounds more logical to me to return a bad result as a result.

phw: The error Parameter will hold exceptions that were caused by the run task. So it probably should not be fully ignored but handled

Gabrielcarvfer: The only thing that raises an exception os the version extraction (IndexError for parsing, TimeoutException if its not an AB extractor), but I just catch it and return None.

phw: Ok

phw (approved):

zas (approved):
LGTM

Gabrielcarvfer:
Just noticed that the ETA doesn't really work for AB since the infostatus menu isn't updated if no changes happened to the number of pending requests/files. ._.

phw:
> Just noticed that the ETA doesn't really work for AB (for a small number of files) since the infostatus menu isn't updated if no changes happened to the number of pending requests/files. Not a gigantic issue though. ._.

Seems to work for me, at least the pending status of the file changes and hence the estimate updates. But you probably mean in between, so that estimate updates between the different things that need to happen for a single file?

But yes, not a big deal and kind of a separate issue to improve the estimate for this case. Let's merge the AB main feature for now :) Thanks a lot for this again

rdswift:
While trying to get screen shots to add to the Picard User Guide documenting this new functionality, I encountered a bug that disabled the AcousticBrainz options page.  I added a ticket for this at https://tickets.metabrainz.org/browse/PICARD-2265

Gabrielcarvfer:
> I encountered a bug that disabled the AcousticBrainz options page. 

@rdswift probably missing the entry with use_acousticbrainz and acousticbrainz_extractor in the config_upgrade.py. =x

rdswift:
> > I encountered a bug that disabled the AcousticBrainz options page.
> 
> @rdswift probably missing the entry with use_acousticbrainz and acousticbrainz_extractor in the config_upgrade.py. =x

I figure it must be something simple, because it worked well when I was testing on the portable version.

> BTW, I'm trying to translate the docs to pt-br, but this new platform is pretty complicated. It gives no context to where the text is and I have no idea what the keywords are for. ._.

First off...  THANK-YOU for helping with the translation effort!  It is very much appreciated.  I'd like to add you to the list of contributors (see the [Contributors](https://picard-docs.musicbrainz.org/en/about_picard/acknowledgements.html#contributors) section).  Do you want to be listed as "Gabriel Ferreira" or some other way?

Yeah, I'm still trying to figure out how to use Weblate effectively as well.

When it comes to seeing the strings in context, I've found that it sort of does that if you click on the "Browse" or "Zen" button on the "All strings" line for a file.

https://translate.uploadedlobster.com/projects/picard-docs/about_picardacknowledgements/pt_BR/

phw:
> probably missing the entry with use_acousticbrainz and acousticbrainz_extractor in the config_upgrade.py. =x

No, there is no config upgrade involved here. config migrations are only needed if something needs to be converted. The issue here is that in some cases `ab_check_version` ends up with being called with `None`. E.g. at https://github.com/metabrainz/picard/blob/master/picard/ui/options/acousticbrainz.py#L94-L98 , if the first call fails and find_extractor fails to find an extractor then `ab_check_version` gets called again with `None`. Also `self._searched_extractor` can end up as None, and in that case the `ab_check_version` also later gets called with `None`. Probably `ab_check_version` should take care of that. Maybe even `find_executable` should work with a `None` argument.

Did not happen in my tests because I tested on Linux, and the code fails only when adding the Windows executable extensions. Also in my tests I always had the extractor in my path.

rdswift:
> Did not happen in my tests because I tested on Linux, and the code fails only when adding the Windows executable extensions. Also in my tests I always had the extractor in my path.

I guess I just have this ability of just being able to break stuff.  :stuck_out_tongue_winking_eye:   I just wish I had found it sooner.

Gabrielcarvfer:
> The issue here is that in some cases `ab_check_version` ends up with being called with `None`. E.g. at https://github.com/metabrainz/picard/blob/master/picard/ui/options/acousticbrainz.py#L94-L98 , if the first call fails and find_extractor fails to find an extractor then `ab_check_version` gets called again with `None`. Also `self._searched_extractor` can end up as None, and in that case the `ab_check_version` also later gets called with `None`. Probably `ab_check_version` should take care of that. Maybe even `find_executable` should work with a `None` argument.
> 
> Did not happen in my tests because I tested on Linux, and the code fails only when adding the Windows executable extensions. Also in my tests I always had the extractor in my path.

I was thinking it was the case, but didn't test it when I said that. I've opened a new MR with a fix. (I did test this case, but before adding the placeholder thing). =x

Gabrielcarvfer:
> I'd like to add you to the list of contributors 

I haven't even started it yet. Calm down. xD

> When it comes to seeing the strings in context, I've found that it sort of does that if you click on the "Browse" or "Zen" button on the "All strings" line for a file.

It doesn't really show much of the context. Makes me want to translate files locally and just open a MR on github =x

rdswift:
> It doesn't really show much of the context. Makes me want to translate files locally and just open a MR on github =x

I understand completely.

I certainly have no issue if you'd prefer to clone the repo and update the *.po files locally, and then submit the updates though a PR.  That's basically what I've been doing so far, using the free version of [Poedit](https://poedit.net/) for updating the *.po files locally.  That way I can also build the output html and pdf files so that I can see everything in context before I submit my updates.  Ultimately the changes also end up on Weblate (automatically) and will undergo the automated checking and be added to the translation memory.

We're trying to be as flexible as possible, because we value the work of the volunteer translators, and want to allow them to work in the manner that they find most comfortable and productive.  If we make the process rigid and difficult, then we're doomed to failure because we'll never get anyone to volunteer under those conditions.

phw:
FYI: I have for now disabled the bundling of the essentia extractor binary in the macOS builds in https://github.com/metabrainz/picard/commit/d9d748e423c37b18095d8340573c1a2b49253677 . The current binary is compiled with an SDK older than 10.9, and that breaks the notarization process. I have already contacted the AB team on IRC about this, we'll see to get a build process setup on Github Actions that will produce a fresh binary that fulfills their requirements.



> It doesn't really show much of the context. Makes me want to translate files locally and just open a MR on github =x

@Gabrielcarvfer  Keep in mind that this translation project is still in evaluation. We haven't officially announced it, yet, and we have still seeing how well this works with the Github and deployment workflow. Currently I go through some German translations, mostly with the goal to spot some things I can improve on the Weblate side.

Anyway, there are a few things that help: Weblate shows nearby strings below the translation box (the first tab), so you can actually see the context. Also Zen mode can help, but I personally miss the translation memory.

Then there are the auto suggestions, which are using the translation memory across projects. As the full translations of MusicBrainz and Picard are also imported this helps a lot with e.g. reusing Picard's translations for option labels etc.

The documentation files are also organized by chapter on the docs. So if you are translating the component config/options_genres look at the documentation at https://picard-docs.musicbrainz.org/en/config/options_genres.html.

And yes, you totally can also locally translate the .po files locally and open a pull request. Translations are synced two way. When you use Poedit you can also locally import the translation memory, which can be a huge help. I have uploaded a fresh full export of all strings to https://transfer.sh/1Dfgvqv/weblate-musicbrainz.tmx.zip

Just keep in mind that we'd likely expect you to solve any conflicts should there be some when merging the pull request. I don't speak Portuguese and if in doubt can't decide how this should be :) But it doesn't look like we will have too many translators all changing the same languages, we can be lucky if there are a few ;)

PR #1856, opened 2021-06-30
Display profile setting items as tooltips

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Display profile setting override values as tooltips.

# Problem

There was no easy way of seeing what the settings override values were for a profile.

# Solution

Provide the current override value for the profile as a tooltip on the setting in the settings tree.

# Action

This currently uses tooltips, but could easily be converted to display the settings in a text box on the dialog when the setting is selected if that is preferred.

### DIFF
--- a/picard/ui/profileeditor.py
+++ b/picard/ui/profileeditor.py
@@ -35,6 +35,7 @@ from picard.config import (
 )
 from picard.const import PICARD_URLS
 from picard.profile import UserProfileGroups
+from picard.script import get_file_naming_script_presets
 
 from picard.ui import (
     PicardDialog,
@@ -235,11 +236,62 @@ class ProfileEditorDialog(SingletonDialog, PicardDialog):
                 child_item.setFlags(QtCore.Qt.ItemIsEnabled | QtCore.Qt.ItemIsUserCheckable)
                 state = QtCore.Qt.Checked if settings and setting.name in settings else QtCore.Qt.Unchecked
                 child_item.setCheckState(self.TREEWIDGETITEM_COLUMN, state)
+                if settings and setting.name in settings and settings[setting.name] is not None:
+                    value = settings[setting.name]
+                else:
+                    value = None
+                child_item.setToolTip(self.TREEWIDGETITEM_COLUMN, self.make_setting_value_text(setting.name, value))
                 widget_item.addChild(child_item)
             self.ui.settings_tree.addTopLevelItem(widget_item)
             if title in self.expanded_sections:
                 widget_item.setExpanded(True)
         self.building_tree = False
+
+    def make_setting_value_text(self, key, value):
+        ITEMS_TEMPLATE = "\n  - %s"
+        NONE_TEXT = _("None")
+        config = get_config()
+        flag = False
+        if value is None:
+            value_text = "None"
+        elif key == "selected_file_naming_script_id":
+            presets = {x["id"]: x["title"] for x in get_file_naming_script_presets()}
+            if value in config.setting["file_renaming_scripts"]:
+                value_text = config.setting["file_renaming_scripts"][value]["title"]
+            elif value in presets:
+                value_text = presets[value]
+            else:
+                value_text = _("Unknown script")
+        elif key == "list_of_scripts":
+            if config.setting[key]:
+                scripts = config.setting[key]
+                value_text = _("Enabled tagging scripts of %i found:") % len(scripts)
+                for (pos, name, enabled, script) in scripts:
+                    if enabled:
+                        flag = True
+                        value_text += ITEMS_TEMPLATE % name
+                if not flag:
+                    value_text += " %s" % NONE_TEXT
+            else:
+                value_text = _("No scripts in list")
+        elif key == "ca_providers":
+            providers = config.setting[key]
+            value_text = _("Enabled providers of %i listed:") % len(providers)
+            for (name, enabled) in providers:
+                if enabled:
+                    flag = True
+                    value_text += ITEMS_TEMPLATE % name
+            if not flag:
+                value_text += " %s" % NONE_TEXT
+        elif isinstance(value, str):
+            value_text = '"%s"' % value
+        elif type(value) in {bool, int, float}:
+            value_text = str(value)
+        elif type(value) in {set, tuple, list, dict}:
+            value_text = _("List of %i items") % len(value)
+        else:
+            value_text = _("Unknown value format")
+        return value_text
 
     def current_item_changed(self, new_item, old_item):
         """Update the display when a new item is selected in the profile list.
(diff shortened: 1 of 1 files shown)


picard/ui/profileeditor.py
```diff
@@ -235,12 +236,74 @@ def make_setting_tree(self, settings=None):
...
+    def get_file_naming_script_name(self, script_id):
+        config = get_config()
+        if script_id in config.setting["file_renaming_scripts"]:
+            return config.setting["file_renaming_scripts"][script_id]["title"]
+        presets = {x["id"]: x["title"] for x in get_file_naming_script_presets()}
+        if script_id in presets:
+            return presets[script_id]
+        return _("Unknown script")
+
+    def get_list_of_enabled_scripts(self, key, ITEMS_TEMPLATE, NONE_TEXT):
```
zas, line 278: Capitalized names are for constants, and constants shouldn't be parameters.

Instead define them at class level.
Also since it's specific to _list_of_scripts_ we don't even need to pass `key`.

```python
def get_list_of_enabled_scripts(self):
    scripts = config.setting['list_of_scripts']
    ....
           value_text += self.ITEMS_TEMPLATE % name
 ```

File: picard/ui/profileeditor.py [hunk 2]
```diff
@@ -235,12 +236,74 @@ def make_setting_tree(self, settings=None):
...
+        config = get_config()
+        flag = False
+        if config.setting[key]:
+            scripts = config.setting[key]
```
zas, line 282:
```python  
scripts = config.setting[key]
if scripts:
```

picard/ui/profileeditor.py
```diff
@@ -235,12 +236,74 @@ def make_setting_tree(self, settings=None):
...
+                if enabled:
+                    flag = True
+                    value_text += ITEMS_TEMPLATE % name
+            if not flag:
+                value_text += " %s" % NONE_TEXT
+        else:
+            value_text = _("No scripts in list")
+        return value_text
+
+    def get_list_of_enabled_ca_providers(self, key, ITEMS_TEMPLATE, NONE_TEXT):
```
zas, line 294: Same here, no need of parameters since that's specific to _ca_providers_

zas (changes requested):
Not a perfect solution, but better than nothing ;)

Perhaps, instead of just **Value not set** we could also display from where the value will come (first profile below the current one).

Something like:
**Value not set (set in profile2 profile)**
**Value not set (set in default profile)**

rdswift:
> Not a perfect solution, but better than nothing ;)

I'm not completely happy with it either, but I figured it would give us a start of something to pick away at while we come up with a better idea for displaying the information.

> Perhaps, instead of just **Value not set** we could also display from where the value will come (first profile below the current one).
> 
> Something like:
> **Value not set (set in profile2 profile)**
> **Value not set (set in default profile)**

I like that, but should it be first profile or first "enabled" profile?  I've initially set it up as first enabled profile, but it's easy to switch.

zas:
> should it be first profile or first "enabled" profile? I've initially set it up as first enabled profile

Yes, first enabled profile is likely more useful

rdswift:
> Yes, first enabled profile is likely more useful

And it updates the tooltip automatically as profiles are enabled/disabled or if the profile order changes.

picard/ui/profileeditor.py
```diff
@@ -235,12 +241,97 @@ def make_setting_tree(self, settings=None):
...
+        if key == "list_of_scripts":
+            return self.get_list_of_scripts()
+        if key == "ca_providers":
+            return self.get_list_of_ca_providers()
+        if isinstance(value, str):
+            return '"%s"' % value
+        if type(value) in {bool, int, float}:
+            return str(value)
+        if type(value) in {set, tuple, list, dict}:
+            return _("List of %i items") % len(value)
```
zas, line 270: It would probably be safer to use `isinstance()` here too:

```python
if isinstance(value, set) or isinstance(value, tuple) ....
```

picard/ui/profileeditor.py
```diff
@@ -235,12 +241,97 @@ def make_setting_tree(self, settings=None):
...
+            return _("Value not set (set in \"%s\" profile)") % profile
+        if key == "selected_file_naming_script_id":
+            return self.get_file_naming_script_name(value)
+        if key == "list_of_scripts":
+            return self.get_list_of_scripts()
+        if key == "ca_providers":
+            return self.get_list_of_ca_providers()
+        if isinstance(value, str):
+            return '"%s"' % value
+        if type(value) in {bool, int, float}:
```
zas, line 267: I'd rather use `isinstance()` than strict `type()`

picard/ui/profileeditor.py
```diff
@@ -235,12 +241,97 @@ def make_setting_tree(self, settings=None):
...
+            return str(value)
+        if type(value) in {set, tuple, list, dict}:
+            return _("List of %i items") % len(value)
+        return _("Unknown value format")
+
+    def get_file_naming_script_name(self, script_id):
+        config = get_config()
+        if script_id in config.setting["file_renaming_scripts"]:
+            return config.setting["file_renaming_scripts"][script_id]["title"]
```
zas, line 276: Perhaps shorten lines a bit here:

```python
scripts = config.setting["file_renaming_scripts"]
if script_id in scripts:
    return scripts[script_id]["title"]
```

picard/ui/profileeditor.py
```diff
@@ -235,12 +241,97 @@ def make_setting_tree(self, settings=None):
...
+    def get_list_of_scripts(self):
+        config = get_config()
+        scripts = config.setting["list_of_scripts"]
+        if scripts:
+            value_text = _("Tagging scripts (%i found):") % len(scripts)
+            for (pos, name, enabled, script) in scripts:
+                if enabled:
+                    value_text += self.ITEMS_TEMPLATE_CHECKED % name
+                else:
+                    value_text += self.ITEMS_TEMPLATE_UNCHECKED % name
```
zas, line 291: Using a dict for templates with boolean as key would simplify this:

```python

self.ITEMS_TEMPLATES = {
    True: "\n  ■ %s",
    False: "\n  □ %s",
}

....
value_text += self.ITEMS_TEMPLATES[enabled] % name
```

picard/ui/profileeditor.py
```diff
@@ -235,12 +241,97 @@ def make_setting_tree(self, settings=None):
...
+
+    def get_list_of_ca_providers(self):
+        config = get_config()
+        providers = config.setting["ca_providers"]
+        value_text = _("CA providers (%i found):") % len(providers)
+        for (name, enabled) in providers:
+            if enabled:
+                value_text += self.ITEMS_TEMPLATE_CHECKED % name
+            else:
+                value_text += self.ITEMS_TEMPLATE_UNCHECKED % name
```
zas, line 304: See comment above

zas (changes requested):

zas (approved):

PR #1857, opened 2021-07-01
Update menu bar settings to reflect any changes to enabled profiles

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Update menu bar after profile editor closes.

# Problem

Changes made in the profile editor, such as enabling/disabling profiles which changed the active file naming script, were not being reflected in the main screen menu bar.

# Solution

Update menu bar settings to reflect any changes to enabled profiles when the profile editor is closed.

# Action

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -593,6 +593,13 @@ class ScriptEditorDialog(PicardDialog):
         if not self.loading:
             self.select_script(skip_check=True)
 
+    def reload_after_profile(self):
+        """Update after changes to profile settings.
+        """
+        config = get_config()
+        id = config.setting["selected_file_naming_script_id"]
+        self.set_selected_script_id(id, skip_check=False)
+
     def docs_browser(self):
         """Open the scriping documentation in a browser.
         """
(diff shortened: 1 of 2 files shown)


picard/ui/scripteditor.py
```diff
@@ -593,6 +593,13 @@ def load(self):
         if not self.loading:
             self.select_script(skip_check=True)
 
+    def reload_after_profile(self):
+        """Update after changes to profile settings.
+        """
+        config = get_config()
+        id = config.setting["selected_file_naming_script_id"]
```
zas, line 600: `id` is a builtin function, so it is best to avoid using it as variable name (I know, it sucks).
But `script_id` would work well here instead.

zas (changes requested):

picard/ui/scripteditor.py
rdswift: I've done that so often you'd think I'd remember it already.  Apparently not.  :frowning:

zas (approved):

PR #1858, opened 2021-07-01
Disallow entering a blank profile title

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Disallow entering a blank profile title.

# Problem

When changing the title of a user profile, the title could be entered as a blank (empty string).

# Solution

Implement a check when the profile title is changed to disallow entering a blank profile title.  Also remove any leading or trailing whitespace.

# Action

### DIFF
--- a/picard/ui/profileeditor.py
+++ b/picard/ui/profileeditor.py
@@ -94,6 +94,7 @@ class ProfileEditorDialog(SingletonDialog, PicardDialog):
         self.move_view = MoveableListView(self.ui.profile_list, self.ui.move_up_button,
                                           self.ui.move_down_button)
 
+        self.ui.profile_list.itemChanged.connect(self.profile_item_changed)
         self.ui.profile_list.currentItemChanged.connect(self.current_item_changed)
         self.ui.profile_list.itemSelectionChanged.connect(self.item_selection_changed)
         self.ui.profile_list.itemChanged.connect(self.profile_data_changed)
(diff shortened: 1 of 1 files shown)


zas (commented):
Is the case properly handled at config load time? Ie. if it was set to blank in config file, Picard should just ignore it.

rdswift:
> Is the case properly handled at config load time? Ie. if it was set to blank in config file, Picard should just ignore it.

It currently ignores the fact that it is blank and adds it to the list widget with a blank title.  Whenever you make a change to the item such as enable, disable or change name (but not when updating the associated settings) the system will complain and changes the blank to a default name.  Funny thing is that if you don't touch that entry, you can actually save it back to the list with a blank title.

Would you prefer to not add it to the list widget at all, and thus it would be removed when the profiles are saved using "Make It So!"?

zas:
> Would you prefer to not add it to the list widget at all, and thus it would be removed when the profiles are saved using "Make It So!"?

If the blank title doesn't break anything, no need to change. We just need to be sure nothing breaks (for example, preventing Picard to load) even if values from config are corrupted (intentionnally or not).

rdswift:
> If the blank title doesn't break anything, no need to change. We just need to be sure nothing breaks (for example, preventing Picard to load) even if values from config are corrupted (intentionnally or not).

Understood.  The only thing that I can think that would cause it to crash is if the setting data changed type, for example if not all of the tuples in the `config.profiles["user_profiles"]` list contained four elements (`int`, `str`, `bool`, `str`).  I didn't include a lot of bullet-proof checking because I hadn't considered something external modifying the settings file.

zas (approved):

zas (approved):

PR #1859, opened 2021-07-01
Check is naming script is used in a profile before deleting

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Check is naming script is used in a profile before deleting.

# Problem

There was no check, so a user could delete a file naming script that was set for use in a profile.

# Solution

Check is naming script is used in a profile before deleting.

# Action

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -35,6 +35,7 @@ from picard import log
 from picard.config import (
     BoolOption,
     Option,
+    SettingConfigSection,
     TextOption,
     get_config,
 )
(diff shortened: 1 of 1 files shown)


zas:
LGTM.

By the way, what happens if a script was removed from config file while still used in a profile? I mean we are supposed to handle valid changes made directly to config files

rdswift:
> By the way, what happens if a script was removed from config file while still used in a profile? I mean we are supposed to handle valid changes made directly to config files

Currently it logs an error that the script was not found, but does not update the settings to replace it with a valid script id (or None).  Should we change the profile setting for that key back to None?  That might be the cleanest way of handling it.

zas:
> Should we change the profile setting for that key back to None? That might be the cleanest way of handling it.

Yes, I agree..

rdswift:
> > Should we change the profile setting for that key back to None? That might be the cleanest way of handling it.
> 
> Yes, I agree..

I actually decided to put the profile and settings cleanup in MainMenu so that any problems are corrected before being accessed anywhere else in the program.

zas (approved):

PR #1861, opened 2021-07-02
PICARD-2237: Realistic progress and ETA

Gabrielcarvfer:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
More realistic progress and addition of an estimated time of completion/action (ETA) to the statusbar.

# Problem

* JIRA ticket (_optional_): PICARD-2237

Different jobs can be wildly different in duration, due to CPU (AcousticBrainz/AcoustID), IO (loading files while writing covers and other temporary files), and API limitations (rate limits).

# Solution

Estimate progress and ETA based not on the number of pending jobs and fixed weights, but actually batch duration, number of completed jobs and number of pending jobs, which reflect more directly the different tasks. 

We also pick the longest ETA between the ETA for file and ETA for network requests. This works because it is the biggest bottleneck for completion of the batch being processed and prevents a big number of fast requests from biasing a general ETA towards a lower estimate.

We could also estimate the ratio of requests per files operations * number of pending file operations and put that into the network estimate to make the ETA closer to the top value faster.

Examples:
Loading files (early on the run)
image
Loading files (when metadata is being pulled and requests are slower)
image
Scanning 18k files (few seconds after clicking)
image
Scanning 18k files (a few minutes after clicking)
image
Scanning 18k files (another few minutes)
image

# Action


Gabrielcarvfer:
Still thinking how we can make that ETA grow faster based on the number of pending files and still be realistic without overcomplicating things and reduce the number of estimates to speed things up.

Gabrielcarvfer:
It is kind of frightening when the web requests start to be processed, but the results are way nicer if you wait a few minutes.

Up to 2:20 its just loading the library. After that there is a huge jump when I click to scan the 18k files. It keeps increasing as the time of fpcalc will slowly increase the processing time per file, but shouldn't be much more than that.
https://user-images.githubusercontent.com/1678454/124340469-e989ee80-db8b-11eb-840a-a64d8d80112f.mp4

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
+            if self._eta_label:
+                seconds = max(eta_seconds, 0)
+
+                # calculate eta in d:h:m:s format
+                days = int(seconds*INV_SECS_IN_DAY)
+                seconds = seconds - days*SECS_IN_DAY
+                hours = int(seconds*INV_SECS_IN_HOUR)
+                seconds = seconds - hours*SECS_IN_HOUR
+                minutes = int(seconds*INV_SECS_IN_MINUTE)
+                seconds = seconds - minutes*SECS_IN_MINUTE
```
zas, line 116: This could be made simpler and more generic:

```python
SECS_IN_DAY = 86400
SECS_IN_HOUR = 3600
SECS_IN_MINUTE = 60


def euclidian_div(a, b):
    return (a // b, a % b)


def seconds_to_dhms(seconds):
    days, seconds = euclidian_div(seconds, SECS_IN_DAY)
    hours, seconds = euclidian_div(seconds, SECS_IN_HOUR)
    minutes, seconds = euclidian_div(seconds, SECS_IN_MINUTE)

    return (days, hours, minutes, seconds)
```

Such functions could be moved to utils and will be easier to test.

picard/ui/mainwindow.py
```diff
@@ -386,6 +386,10 @@ def create_statusbar(self):
             "you \"Search\" or \"Open in Browser\" from Picard, clicking the "
             "\"Tagger\" button on the web page loads the release into Picard."
         ))
+        self.eta_label = QtWidgets.QLabel()
+        self.eta_label.setMaximumSize(QtWidgets.QLabel("ETA: 00d:00h:00m:00s").size())
```
zas, line 390: This should be initialized with actual format from a function so format is in one place (something like `eta_dhms(0)` returning translated string)

picard/ui/statusindicator.py
```diff
@@ -74,7 +137,7 @@ def set_progress(self, progress):
 
     class WindowsTaskbarStatusIndicator(AbstractProgressStatusIndicator):
         def __init__(self, window):
-            super().__init__()
+            super().__init__(window)
```
zas, line 140: Is this change related to the PR?

zas (changes requested):

zas:
> It is kind of frightening when the web requests start to be processed,

Perhaps we could smooth it out so the increase isn't too frightening

Sophist-UK:
It seems to me that to make an accurate estimate we need to maintain some statistics (total elapsed time for operations and number of operations) for file operations, and by network domain for network requests.

I would suggest that we note the start time for each operation, and when the operation ends we add the elapsed time to the total and increment the number of operations. This allows us to calculate the average time for operations and thus the expected time for remaining operations.

Then we can calculate the expected time for files / each network domain as `rate limit * length of queue * average response time`   and the expected finish time is the maximum of these values.

If we wanted to increase the accuracy we could maintain separate statistics for each disk drive / UNC name and for load and save operations and for different types of network calls - but that would also increase the complexity / analysis time as you could not simply use the length of the queue and would either have to scan the calls to analyse them each time you wanted to estimate the ETA or maintain a counter for the length of each type of queue.

I guess you would also either have to have a default figure to calculate the average response time when no calls had yet completed or only estimate an ETA when you had the stats to allow it.

picard/ui/statusindicator.py
Gabrielcarvfer: Part of the janky hack commit. Thinking on how to make it better, but getting a good ETA is my priority.

Gabrielcarvfer:
> It seems to me that to make an accurate estimate we need to maintain some statistics (total elapsed time for operations and number of operations) for file operations, and by network domain for network requests.

Yup.

> I would suggest that we note the start time for each operation, and when the operation ends we add the elapsed time to the total and increment the number of operations. This allows us to calculate the average time for operations and thus the expected time for remaining operations.

That's basically what I'm doing, but not for each operation as we do not have that granularity and probably don't want it either.

> Then we can calculate the expected time for files / each network domain as `rate limit * length of queue * average response time` and the expected finish time is the maximum of these values.

Sort of. Problem is this is only true if you know beforehand that the operation does strictly what it says it does (e.g. loading file only touches the file and doesn't create an/multiple additional request/s).

> If we wanted to increase the accuracy we could maintain separate statistics for each disk drive / UNC name and for load and save operations and for different types of network calls - but that would also increase the complexity / analysis time as you could not simply use the length of the queue and would either have to scan the calls to analyze them each time you wanted to estimate the ETA or maintain a counter for the length of each type of queue.

I don't really want to go that route.

> I guess you would also either have to have a default figure to calculate the average response time when no calls had yet completed or only estimate an ETA when you had the stats to allow it.

Yes, right now I'm trying to tune the ETA so that you don't have to wait until the entire batch ends to get an accurate result when you try to do something else (where the current code drops all statistics and restart from scratch and usually is pretty accurate).
When I don't overshoot by a lot, I end up missing the target by ~25% (generating fingerprint with biased load file statistics).

Sophist-UK:
> > I would suggest that we note the start time for each operation, and when the operation ends we add the elapsed time to the total and increment the number of operations. This allows us to calculate the average time for operations and thus the expected time for remaining operations.
> 
> That's basically what I'm doing, but not for each operation as we do not have that granularity and probably don't want it either.

So how do you do it if not measuring each transaction and averaging?
 
> > Then we can calculate the expected time for files / each network domain as `rate limit * length of queue * average response time` and the expected finish time is the maximum of these values.
> 
> Sort of. Problem is this is only true if you know beforehand that the operation does strictly what it says it does (e.g. loading file only touches the file and doesn't create an/multiple additional request/s).

Some plugins which run only after the main MB WS has completed generate one web call per track. In other words it is entirely normal for the number of pending network calls to go up.

> > If we wanted to increase the accuracy we could maintain separate statistics for each disk drive / UNC name and for load and save operations and for different types of network calls - but that would also increase the complexity / analysis time as you could not simply use the length of the queue and would either have to scan the calls to analyze them each time you wanted to estimate the ETA or maintain a counter for the length of each type of queue.
> 
> I don't really want to go that route.

No - I think it is overkills and if you do that then you have to be careful about e.g. how to deal with loads and saves on the same drive.

> > I guess you would also either have to have a default figure to calculate the average response time when no calls had yet completed or only estimate an ETA when you had the stats to allow it.
> 
> Yes, right now I'm trying to tune the ETA so that you don't have to wait until the entire batch ends to get an accurate result when you try to do something else (where the current code drops all statistics and restart from scratch and usually is pretty accurate).
> When I don't overshoot by a lot, I end up missing the target by ~25% (generating fingerprint with biased load file statistics).

Yes, whatever algorithm is used needs to still work when the user adds more actions to the queue whilst the first set is still in progress and it needs to cope with e.g. them being entirely an different type of action e.g. saves during file loads and release loads. My algorithm does that but would not handle the increase in network calls when Track plugins add one per track - to take that into account you would need to e.g. calculate the number of network calls made for completed albums cf. no of tracks and use that to estimate future calls. But I imagine that is going to be similar for any other algorithm - without understanding how you are currently trying to estimate, I cannot comment on how this would apply to yours.

Gabrielcarvfer:
> So how do you do it if not measuring each transaction and averaging?

I just expanded the statusindicator implementation and measured the time since the batch started.
You can calculate the time per file dividing the number of processed files and current timestamp - timestamp of when the batch started. It is an average, but not per transaction, but per sampling interval of the statusindicator (every 100ms).

Which is great, as the uniform sampling = estimate for the average value is close to the real average, except when different tasks are introduced.

> Some plugins which run only after the main MB WS has completed generate one web call per track. In other words it is entirely normal for the number of pending network calls to go up.

Yes, nothing I can do for plugins. They will slowly creep up.

> Yes, whatever algorithm is used needs to still work when the user adds more actions to the queue whilst the first set is still in progress and it needs to cope with 

That is what I'm thinking. Right now the time per file and time per request are based on the batch duration divided by the number of completed file/request operations. If I want this value to naturally increase over time (kind of moving an averaging window forward), I need to reduce the number of completed operations. These are currently calculated as             `previous_done_files[t] = self._max_pending_files[t-1] - self._last_pending_files[t-1]`

where

```
self._max_pending_files[t] = max(self._max_pending_files[t-1], previous_done_files[t] + pending_files[t])
self._last_pending_files[t] = pending_files[t]
```

Sophist-UK:
On reflection, using elapsed time for completely loaded times vs. total number of files to load, that seems to me might be pretty accurate - with the proviso that albums are of a similar number of tracks. If you had e.g. AcoustBrainz plugin which does a call per track, and were to load a bunch of NAT tracks from your singles collection first and then a bunch of albums, then the estimate is likely to be completely wrong.

Adding more loads of a similar type should not impact the estimate - but adding saves will also be likely to be completely wrong.

How difficult would it be to calculates ETAs separately for Loads and Saves and show the maximum of the two?

Gabrielcarvfer:
> How difficult would it be to calculates ETAs separately for Loads and Saves and show the maximum of the two?

Considering there's no distinction between different tasks of the same type, as far as I know, it seems rather complicated.

I believe the current estimate is pretty good as is, except for the bias. Looking for statistical tricks to fix the issue.

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
-        progress = 1 - (total_pending / self._max_pending)
-        self.set_progress(progress)
+        if total_pending != last_pending:
+            current_time = time.time()
+
+            # time since we started processing this batch
+            diff_time = current_time-self._prev_time
+
+            # file eta
+            f_eta_seconds = 0
```
zas, line 88: Use `file_eta_seconds` instead of `'f_eta_seconds`

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
+            if previous_done_files:
+                # we assume an additional network request per slower pending file
+                # operation to estimate an upper bound (e.g. AB, AcoustID)
+                f_eta_seconds = (diff_time / previous_done_files) * (pending_files + (pending_requests*previous_done_files/previous_done_requests if previous_done_requests > 0 else pending_requests))
+
+            # network eta
+            n_eta_seconds = 0
```
zas, line 95: Use `network_eta_seconds` instead of `'n_eta_seconds`

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
-        self._max_pending = max(self._max_pending, previous_done + total_pending)
-        self._last_pending = total_pending
-
-        if total_pending == 0 or self._max_pending <= 1:  # No need to show progress for single item
-            self._max_pending = 0
+        total_pending = pending_files + pending_requests
+        last_pending = self._last_pending_files + self._last_pending_requests
+        if total_pending != last_pending:
+            previous_done_files = self._max_pending_files - self._last_pending_files
+            previous_done_files = previous_done_files if previous_done_files >= 0 else 0
```
zas, line 61: `previous_done_files = max(0, previous_done_files)`

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
+            previous_done_requests = self._max_pending_requests - self._last_pending_requests
+            previous_done_requests = previous_done_requests if previous_done_requests >= 0 else 0
```
zas, line 63: `previous_done_requests = max(0, previous_done_requests)`

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
+            self._max_pending_requests = max(self._max_pending_requests, previous_done_requests + pending_requests)
+            self._last_pending_files = pending_files
+            self._last_pending_requests = pending_requests
+
+        if total_pending == 0 or (self._max_pending_files == 0 and self._max_pending_requests == 0):
+            self._max_pending_files = 0
+            self._max_pending_requests = 0
+            self._last_pending_files = 0
+            self._last_pending_requests = 0
+            self._prev_time = time.time()
```
zas, line 74: This block is duplicated in `__init__()`, move it to a method.

zas (changes requested):

picard/ui/mainwindow.py
phw: In addition I even think this should be initialited in the actual implementation. See my comment on refactoring this below.

phw (changes requested):
In general I think this is a very good move and something we should continue on. I have some suggestions to refactor the code below.

About the display of the remaining time. "ETA: 00d:00h:04m:32s" is not really easy to parse. A lot of detailed information, with the relevant parts hard to spot. Maybe we could limit it to the actual available data, e.g. if the highest unit is minutes just display "4m 32s". I think that reads way better.

Personally I probably would go a step further and completely omit seconds unless we are under one minute. The display of seconds wrongly indicates a precision that is not available. Why not even just go with the highest unit for rough estimate? "5 minutes", "32 seconds" or "2 days" (that's scary). Easy to read, as rough as the estimate is and easy to translate.

Or show a progress bar with the time display as a tooltip or label?

Also when loading a library I noticed that the ETA first is rather low, sometimes just showing one second. Then it jumps up to over an hour when file loading has finished and the network requests are queued up, which then rapidly drops. Maybe we can do better for an initial estimate of network requests, because as a rule of thumb we can assume a lot of network requests are taking ~1 second due to MB rate limit. Actually when I in the past did performance measurements as soon as we came to network requests dividing the number of pending requests by 60 often gave a quite accurate results. Of course sometimes additional requests come up, but those are usually faster and they would go into the calculation.

picard/ui/statusindicator.py
```diff
@@ -30,33 +33,93 @@
...
+                seconds = max(eta_seconds, 0)
+
+                # calculate eta in d:h:m:s format
+                days = int(seconds*INV_SECS_IN_DAY)
+                seconds = seconds - days*SECS_IN_DAY
+                hours = int(seconds*INV_SECS_IN_HOUR)
+                seconds = seconds - hours*SECS_IN_HOUR
+                minutes = int(seconds*INV_SECS_IN_MINUTE)
+                seconds = seconds - minutes*SECS_IN_MINUTE
+                estimated_remaining_time = ("ETA: %.2dd:%.2dh:%.2dm:%.2ds" % (days, hours, minutes, seconds))
```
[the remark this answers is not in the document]
rdswift: > Also not all languages have short equivalent to 'ETA' acronym.

Perhaps show the AcousticBrainz icon rather than "ETA"?

rdswift:
> About the display of the remaining time. "ETA: 00d:00h:04m:32s" is not really easy to parse. A lot of detailed information, with the relevant parts hard to spot. Maybe we could limit it to the actual available data, e.g. if the highest unit is minutes just display "4m 32s". I think that reads way better.

Or possibly one level below the highest unit?  For example: "1d 6h"; "2h 13m"; "4m 32s".

picard/ui/statusindicator.py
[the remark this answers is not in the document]
Gabrielcarvfer: It's not only for AB, but for other file/network requests. Maybe a little clock icon? 
I don't want to take much more space with a large string. "Remaining Time" is probably the shortest I can think of in terms of text.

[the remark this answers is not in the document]
rdswift: I like your idea of a clock icon.  It takes up minimal real estate, and doesn't require translation because it is pretty much universally understood.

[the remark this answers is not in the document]
Sophist-UK: > Perhaps show the AcousticBrainz icon rather than "ETA"?

Eh?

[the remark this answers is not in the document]
kellnerd: Maybe an hourglass icon would be even more intuitive?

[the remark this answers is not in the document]
Sophist-UK: Animated hourglass, turning over every 30s?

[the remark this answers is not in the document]
Gabrielcarvfer: I like the hourglass idea. 

Not sure how Qt handles animated stuff. I definitely don't want to animate that manually 🤯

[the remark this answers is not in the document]
rdswift: Hourglass sounds like an even better idea.  Great suggestion!

Gabrielcarvfer:
Got some free time and started working on this again. No need to review just yet, still working on previous feedback.

So far:
- Less hacky EtaStatusIndicator
- Moving seconds to day/hour/minute/second (dhms) to a separate function
- Only picking up to the 2 most significant values (dhms)
- Not showing estimate for a single pending file/network request
- Pretty good estimate as far as my testing goes
- If you wait for file requests to end, the time per file and ratio between files and requests is reset. If you add more file operations, it will converge to a better estimate than keeping slow/fast file ops together.
- Remove duplicate code blocks 

Missing: 
- Hourglass icon
- Smooth out peaks when you have new file ops while network ops are still running

Gabrielcarvfer:
I really suck with UI stuff. No matter what I do, it still looks terrible.
22x22 with black border
image
22x22 without black border
image
256x256 scaled to 22x22 with black border
image

phw:
Not doing a full review, just a quick test: The ETA display is now better, both visually and from the displayed time. Great work!

Currently this does not work on Linux, because the Linux implementation in UnityLauncherEntryStatusIndicator is not correct. But this should be refactored anyway, with the ETA status indicator beings it's own class that gets instantiated independent of the desktop status indicators (see my comment above on how I'd refactor this).

> I really suck with UI stuff. No matter what I do, it still looks terrible.

You are in good company then :D Not really one of the strengths here. Anyway, I think this is definitely a good improvement, much more readable. We can change the icon later on.

Gabrielcarvfer:
> Not doing a full review, just a quick test: The ETA display is now better, both visually and from the displayed time. Great work!

Thanks :)

> Currently this does not work on Linux, because the Linux implementation in UnityLauncherEntryStatusIndicator is not correct. But this should be refactored anyway, with the ETA status indicator beings it's own class that gets instantiated independent of the desktop status indicators (see my comment above on how I'd refactor this).

Yes. I moved it to the infostatus part, where it seems to make much more sense. Still hacky passing the progress from the infostatus to the statusindicator in `mainwindow.update_statusbar_stats`, but no `self.eta_label = window.tagger.window.eta_label` nonsense.

> You are in good company then :D Not really one of the strengths here. Anyway, I think this is definitely a good improvement, much more readable. We can change the icon later on.

I tried my best pixel-art. Still sucks.

picard/ui/statusindicator.py
Gabrielcarvfer: I thought using the integer div and mod would be slower, but ended up being faster. *the more you know*

zas (changes requested):

picard/util/time.py
```diff
@@ -0,0 +1,51 @@
...
+    return days, hours, minutes, seconds
+
+
+def get_timestamp(seconds):
+    timestamp = ""
+    msv = 0
+    for (value, unit) in zip(seconds_to_dhms(seconds), ["d", "h", "m", "s"]):
+        if value == 0 and msv == 0:
+            continue
+        timestamp += "%.2d%s " % (value, unit)
```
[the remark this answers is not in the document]
Gabrielcarvfer: I see. The code is equivalent.

zas (approved):
LGTM

phw (approved):
LGTM

The only issue is a small architectural thing: I find it a bit odd that the progress calculation now is inside one specific widget and gets propagated from there to other status indicator implementations. Clearer would be to have the progress calculation in a separate object independent of a specific UI. But let's merge this, this can be refactored separately.

PR #1865, opened 2021-07-10
PICARD-2239: Allow saving options settings to specified profile

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Allows user to change the settings only in the specified profile.

# Problem

* JIRA ticket (_optional_): PICARD-2239

Users requested an easier method of updating the option settings for a specific profile.  See discussion on the [Community Forum](https://community.metabrainz.org/t/introducing-option-profiles-in-picard/540379/7)

# Solution

Add a combo box to the Options... page (above the "Make it so!" button) to select which profile to save the updates. The default is "Automatically select" in which case the changes are written to the first enabled profile in the stack that manages the option (which is the current action). If a profile is selected, then settings are saved to **only** that profile, and only for options managed by that profile. All other changes are ignored.

Currently the settings displayed are those for the first enabled profile in the stack that manages that option. This can be a bit misleading when making changes, especially if the selected profile isn't currently enabled, or is being overridden by another enabled profile higher up in the stack. With this change, when a specific profile is selected, the settings displayed are updated to show as if the stack had only the selected profile enabled.  The values displayed are reset to normal (the settings displayed are those for the first enabled profile in the stack that manages that option) if the "Automatically select" option is selected.

When saving to a selected profile using "Make it so!", the Options... dialog remains open with a pop-up message displayed to confirm that the specified profile has been updated.  This allows for quickly selecting a different profile, making a change, and then updating the profile.  If there is no specific profile selected, the Options... dialog closes as it does currently.

**EDIT:**

Also add a new dialog that can be opened from each Options... page to display which profiles (if any) are associated with each of the options on the page.

# Action

If and when this is accepted and merged, the documentation will need to be updated on https://github.com/rdswift/picard-docs/pull/104, including:

- Configuration options settings overview
- All option settings page images
- Option profiles documentation


picard/ui/options/dialog.py
```diff
@@ -166,6 +183,48 @@ def __init__(self, default_page=None, parent=None):
                 self.disable_page(page.NAME)
         self.ui.pages_tree.setCurrentItem(self.default_item)
 
+        self.ui.save_to_profile.clear()
+        self.ui.save_to_profile.addItem(_("Automatically select profile"), None)
```
zas, line 187: When no profile exists, perhaps hide this?

zas:
Overall it looks good to me.

picard/ui/options/dialog.py
```diff
@@ -263,3 +371,62 @@ def _show_dialog(self, msg, function):
...
+            item.setEditable(False)
+            row = [item]
+            attached = []
+            for profile in profiles:
+                if name in settings[profile["id"]]:
+                    attached.append("{0}{1}".format(profile["title"], _(" [Enabled]") if profile["enabled"] else "",))
+            attached_profiles = "\n".join(attached) if attached else _("None")
+            item = QtGui.QStandardItem(attached_profiles)
+            item.setEditable(False)
+            row.append(item)
```
zas, line 421: It took me a while to notice `item` here is different from `item` above, perhaps just use 2 different variable names.

rdswift: Changed to use different names as suggested.  Makes it a lot less confusing.  Good catch!

rdswift:
> When saving to a selected profile using "Make it so!", the Options... dialog remains open with a pop-up message displayed to confirm that the specified profile has been updated. This allows for quickly selecting a different profile, making a change, and then updating the profile. If there is no specific profile selected, the Options... dialog closes as it does currently.

Is the pop-up too much?  Would it be better to just briefly display a confirmation message on the same line as the rest of the profile-specific stuff?

zas:
> Is the pop-up too much? Would it be better to just briefly display a confirmation message on the same line as the rest of the profile-specific stuff?

I don't think it is necessary. Also I noticed there's no popup when I have one profile enabled, and let the selector on automatic, while I have the popup when selecting the profile.

I wonder if it would be useful to be able to save the option to normal options (via the selector).

rdswift:
> Also I noticed there's no popup when I have one profile enabled, and let the selector on automatic, while I have the popup when selecting the profile.

I am not currently showing the popup when saving the settings normally (saving to the first enabled profile) because this is the normal behavior.  In the case of there being a selected profile, the settings are **only** saved to that profile, and all options not managed by the profile are ignored (and not saved to the user settings).  Thinking about it, that may not be intuitively clear to the user.

> I wonder if it would be useful to be able to save the option to normal options (via the selector).

That's a good idea.

Perhaps a radical shift in thinking instead...

Only use the profile stack when retrieving an option setting value.  When saving settings, only save them to the user settings unless a profile is selected.  If a profile is selected, only save the settings to that profile and don't save any settings not managed by that profile.  The combobox label could be changed to "Save settings to:" and the items would be "User settings" (the default selection) followed by each of the defined profiles.  If there are no profiles defined, then that whole profiles combobox line would be hidden (as you suggested earlier).

rdswift:
> > Is the pop-up too much? Would it be better to just briefly display a confirmation message on the same line as the rest of the profile-specific stuff?
> 
> I don't think it is necessary.

Sorry, but just to confirm that I'm understanding you correctly...  You don't think that the pop-up is necessary and can be removed?  In that case, should there be any sort of indication of changes made, or should we just close the Options dialog every time after saving the settings?  Closing the dialog every time sort of makes sense to me in that the behavior is then consistent.

zas (approved):
To be frank it works well enough for me, we can improve it with further PRs.
So, ready to merge for me.

phw (approved):
Works well and codewise not comments. I'm ok with merging this.

My concerns about this feature in general:

- The options UI now looks very busy. I think we can fix this by re-arranging things a bit, but not yet quite sure how
- The attached profiles dialog is not easy to understand. Great would be if we would have a UI that makes it clear inside the dialogs which options are being changed.
- The complexity of this feature in general :) It is very powerful, but the discussion on the forum showed it is hard to explain.

But all things we can improve upon

rdswift:
> * The options UI now looks very busy. I think we can fix this by re-arranging things a bit, but not yet quite sure how
> * The attached profiles dialog is not easy to understand. Great would be if we would have a UI that makes it clear inside the dialogs which options are being changed.
> * The complexity of this feature in general :) It is very powerful, but the discussion on the forum showed it is hard to explain.

Ticket PICARD-2246 created with an initial suggestion.

PR #1866, opened 2021-07-12
Unify labels in profiles and options dialog

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [x] Other

# Problem

The profile editor uses option labels that are different than what is being used in the options dialog. In some cases this makes sense, but in others the wording is less clear and/or the duplication causes unnecessary additional translation work. 

# Solution
If it makes sense use the same label for the profile dialog as is being used in the options to avoid unnecessary differences and less clear wording.


rdswift:
I really didn't like creating duplicates of all the same descriptions in the `UserProfileGroups()` class (which I guess I should rename to `OptionProfileGroups()`) for exactly this reason; however, I didn't see any easy way to retrieve the existing strings for the options.  This causes a lot of duplicate text to be stored in the application.

Thinking out loud, I wonder if it's worth creating a new class (or dictionary) to hold all the option setting text strings keyed by the option setting key.  That would ensure consistency because everything would be stored only once, but it would require significant changes to all of the option settings pages to set all the labels on the page, and would complicate the addition of new options.

Another method might be to have the `UserProfileGroups()` class try to retrieve the text for each of the labels from the appropriate `UI_*OptionsPage()` class definitions.  If it's possible, this might actually make more sense, and cause less additional work (and complication) in the future because it would only impact the addition of new options added to the profile groups.

Is this something that you'd like me to investigate further, or is it not worth it at this point?

zas (approved):

phw:
> Thinking out loud, I wonder if it's worth creating a new class (or dictionary) to hold all the option setting text strings keyed by the option setting key. That would ensure consistency because everything would be stored only once, but it would require significant changes to all of the option settings pages to set all the labels on the page, and would complicate the addition of new options.

I don't think this would really solve anything. As you said it would complicate adding new options, it would also complicate editing the options UI. In the end we likely always need to update three different places and replace one kind of redundancy with a separate one. It also doesn't completely eliminate the problem with having two labels in some instances, because there are still some differences.

There might be some value in having kind of a default label defined with each option, maybe as part of the option definition itself. But then is the question how to get this into the UI. We could define specific option widgets, such as a BoolOptionWidget or something. But then we either have every option page very generic or we end up having a lot of custom widgets per option anyway. Overall this route also sounds a bit like overcomplicating things.

I think I'd rather have some duplication of labels where it makes sense :)

PR #1867, opened 2021-07-21
PICARD-2151: Add $year(), $month() and $day() helper functions to extract date info

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Add new `$year()`, `$month()` and `$day()` functions.

# Problem

* JIRA ticket (_optional_): PICARD-2151

Request for helper functions to extract the different parts of a date string.

# Solution

Add new `$year()`, `$month()` and `$day()` functions.

# Action

Reminder to update the documentation on https://github.com/rdswift/picard-docs/pull/104

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1408,7 +1408,7 @@ def _split_date(date_to_parse, date_order="ymd", separator="-"):
 
 
 @script_function(documentation=N_(
-    """`$year(date, date_order="ymd", separator="-")`
+    """`$year(date,date_order="ymd",separator="-")`
 
 Returns the year portion of the specified date.  The default order is "ymd" and the default separator is a hyphen.
 The default date order can be changed by specifying either "dmy" or "mdy".  The default separator can be changed by
@@ -1420,7 +1420,7 @@ def func_year(parser, date_to_parse, date_order='ymd', separator="-"):
 
 
 @script_function(documentation=N_(
-    """`$month(date, date_order="ymd", separator="-")`
+    """`$month(date,date_order="ymd",separator="-")`
 
 Returns the month portion of the specified date.  The default order is "ymd" and the default separator is a hyphen.
 The default date order can be changed by specifying either "dmy" or "mdy".  The default separator can be changed by
@@ -1432,7 +1432,7 @@ def func_month(parser, date_to_parse, date_order='ymd', separator="-"):
 
 
 @script_function(documentation=N_(
-    """`$day(date, date_order="ymd", separator="-")`
+    """`$day(date,date_order="ymd",separator="-")`
 
 Returns the day portion of the specified date.  The default order is "ymd" and the default separator is a hyphen.
 The default date order can be changed by specifying either "dmy" or "mdy".  The default separator can be changed by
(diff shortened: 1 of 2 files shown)


picard/script/functions.py
```diff
@@ -1381,3 +1381,97 @@ def func_countryname(parser, country_code, translate=""):
...
+        separator = None
+    parts = date_to_parse.strip().split(separator)
+    parts.extend(['', '', ''])
+    date_order = date_order.lower()
+    if date_order == 'dmy':
+        return parts[2], parts[1], parts[0]
+    elif date_order == 'mdy':
+        return parts[2], parts[0], parts[1]
+    else:
+        return parts[0], parts[1], parts[2]
```
zas, line 1407: We could use a `namedtuple` for the result.

```python
DateTuple = namedtuple('DateTuple', ('year', 'month', 'day'))

....

return DateTuple(parts[0], parts[1], parts[2])

```

So in caller functions, we can write:

```python
return _split_date(date_to_parse, date_order, separator).year
```

zas (approved):
LGTM

phw (commented):
Great to have these functions :)

My only concern would be the date parsing, we now have different ways of interpreting the dates across the system. We have:

- [`picard.util.extract_year_from_date`](https://github.com/metabrainz/picard/blob/master/picard/util/__init__.py#L725), which uses [dateutil.parser.parse](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse)
- [`picard.util.sanitize_date`](https://github.com/metabrainz/picard/blob/master/picard/util/__init__.py#L209) and some [other locations](https://github.com/metabrainz/picard/blob/master/picard/track.py#L425) where it just assumes `YYYY-MM-DD`

Maybe in this case it would be better to use [dateutil.parser.parse](https://dateutil.readthedocs.io/en/stable/parser.html#dateutil.parser.parse). We'd still want indicators to choose between the ymd, dmy and mdy orders, but dateutil.parser.parse has two parameter's `dayfirst` and `yearfirst` for this. We could keep your `ymd` etc. implementation, as I think this is easier to handle, but translate them to the boolean values:

|date_order|dayfirst|yearfirst|
|-|-|-|
|ymd|False|True|
|dmy|True|False|
|mdy|False|False|

rdswift:
I changed to using the `dateutil.parser.parse()` function as suggested, but there seems to be a problem.  When testing using an empty string or string of random characters as the date input, the exceptions are not being properly captured.  This only seems to happen in the test suite, and it seems to work properly in the script editor.  For now, I just deleted the tests for invalid dates, and added a check for an empty date string in the helper function.

Also, using the `dateutil` parser is a lot more forgiving with respect to missing elements, so I am using the default date of 1900-01-01 to provide the missing bits rather than returning an empty string in the `$year()`, `$month()` and `$day()` functions as before.  This bothers me a bit because the user can no longer check for example if the `release_date` contains just a year, a year and month, or a full date.  Is this okay, or do you have any suggestions?

rdswift:
Just ran into another issue...  There's currently no way to enter this into a script and save it using anything other than a hard-coded date because it shows the "Invalid date" error.  I'll have a look at fixing that.

picard/script/functions.py
```diff
@@ -1370,7 +1372,7 @@ def func_unique(parser, multi, case_sensitive="", separator=MULTI_VALUED_JOINER)
 
 
 @script_function(documentation=N_(
-    """`$countryname(country_code, translate="")`
+    """`$countryname(country_code,translate="")`
```
zas, line 1375: This change is unrelated to this PR, I committed it as https://github.com/metabrainz/picard/commit/40a03a14f5df143e2831d3032f94d46f93306b6c

zas (changes requested):

phw:
> This bothers me a bit because the user can no longer check for example if the `release_date` contains just a year, a year and month, or a full date. Is this okay, or do you have any suggestions?

Oh yes, the partial dates :( Those don't work well with dateutil, but are essential to our use case.

So probably better to go back to a custom implementation. I would be fine with what you previously had, that looked good function wise. In a follow up step we could then start extracting this logic to our own helper function and using this elsewhere where appropriate, completely replacing dateutil.

picard/script/functions.py
```diff
@@ -1381,3 +1383,105 @@ def func_countryname(parser, country_code, translate=""):
...
+        date_order (str, optional): Order of date elements. Can be "ymd", "mdy" or "dmy". Defaults to "ymd".
+
+    Returns:
+        datetime: Object derived from the parameters provided
+    """
+    date_order = date_order.lower()
+    if date_order not in {'ymd', 'mdy', 'dmy'}:
+        date_order = 'ymd'
+    dayfirst = date_order[0] == 'd'
+    yearfirst = date_order[0] == 'y'
```
zas, line 1402: The code above use 3 tests, and sets `date_order` to `ymd` while it isn't used afterwards.

I think following code is equivalent with less tests and sets:

```python
if date_order == 'dmy':
    dayfirst, yearfirst = True, False
elif date_order == 'mdy':
    dayfirst, yearfirst = False, False
else:  # defaults to 'ymd'
    dayfirst, yearfirst = False, True
```

What do you think?

zas (changes requested):

rdswift:
Based on the discussion here and in IRC, I reverted the `dateutil` stuff, used a regular expression to split the date parts, and rebased against the `master` branch.

picard/script/functions.py
```diff
@@ -1381,3 +1381,92 @@ def func_countryname(parser, country_code, translate=""):
...
+
+    Args:
+        date_to_parse (str): Date string to parse
+        date_order (str, optional): Order of date elements. Can be "ymd", "mdy" or "dmy". Defaults to "ymd".
+
+    Returns:
+        tuple: Tuple of the date parts as (year, month, day)
+    """
+
+    parts = re.split('[^0-9]+', date_to_parse.strip())
```
zas, line 1400: Just a minor thing, we can use `\D` (=any non-digit) instead of `[^0-9]`:

```suggestion
    parts = re.split(r'\D+', date_to_parse.strip())
```

zas (changes requested):
It looks good to me.

zas (approved):
Lgtm

picard/script/functions.py
```diff
@@ -1381,3 +1381,92 @@ def func_countryname(parser, country_code, translate=""):
...
+def func_year(parser, date_to_parse, date_order='ymd'):
+    return _split_date(date_to_parse, date_order).year
+
+
+@script_function(documentation=N_(
+    """`$month(date,date_order="ymd")`
+
+Returns the month portion of the specified date.  The default order is "ymd".  This can be changed by specifying
+either "dmy" or "mdy".  If the date is invalid an empty string will be returned.
+"""
```
zas, line 1427: See https://github.com/metabrainz/picard/pull/1868#discussion_r675906402

phw (approved):

zas (approved):
LGTM

PR #1869, opened 2021-07-23
PICARD-2246: Highlight items in Options dialog when profile is selected for output

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**:  Add highlighting of options manged by the selected profile in the Options dialog.

# Problem

* JIRA ticket (_optional_): PICARD-2246

From the discussion:

> * The attached profiles dialog is not easy to understand. Great would be if we would have a UI that makes it clear inside the dialogs which options are being changed.

# Solution

Add highlighting of options manged by the selected profile in the Options dialog.

# Action

Reminder to update the documentation.

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -30,6 +30,8 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
+from collections import namedtuple
+
 from PyQt5 import (
     QtCore,
     QtGui,
@@ -43,7 +45,10 @@ from picard.config import (
     TextOption,
     get_config,
 )
-from picard.profile import UserProfileGroups
+from picard.profile import (
+    USER_SETTINGS_PROFILE_ID,
+    UserProfileGroups,
+)
 from picard.util import (
     restore_method,
     webbrowser2,
@@ -185,13 +190,12 @@ class OptionsDialog(PicardDialog, SingletonDialog):
                 self.disable_page(page.NAME)
         self.ui.pages_tree.setCurrentItem(self.default_item)
 
-        # Set initially displayed values to the user's base settings (with no profiles enabled)
-        # config.setting.set_profile("user_settings")
+        self.USER_SETTINGS_TITLE = _("User base settings")
 
         if config.profiles[SettingConfigSection.PROFILES_KEY]:
             self.ui.profile_frame.show()
             self.ui.save_to_profile.clear()
-            self.ui.save_to_profile.addItem(_("User base settings"), "user_settings")
+            self.ui.save_to_profile.addItem(self.USER_SETTINGS_TITLE, USER_SETTINGS_PROFILE_ID)
             index = 0
             for idx, item in enumerate(config.profiles[SettingConfigSection.PROFILES_KEY], start=1):
                 self.ui.save_to_profile.addItem(item["title"], item["id"],)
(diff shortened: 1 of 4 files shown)


phw:
This is looking good. I like the generic approach, I was concerned that we would need to adapt each option dialog separately for something like this to work.

rdswift:
> This is looking good. I like the generic approach.

The one thing that's not so generic is that it requires each of the option dialogs to have their ui at `self.ui` so that there is a common starting point for finding the widgets to highlight.  Fortunately, this is the case for all of the existing option pages, but is something to bear in mind for any new pages added in the future.

rdswift:
Now that the basics are there, any suggestions for color combinations for the highlighting?  I'm thinking that we want something noticeable but not overwhelming, thus the reason for allowing different combinations for light and dark themes.

rdswift:
During my testing, I found that I often forgot to go back to the "User settings" target before saving changes to the theme, and only realized my mistake when I restarted Picard and the theme hadn't changed.  What do you think of having some sort of visual indicator on the dialog when a profile is selected as the output target?  I'm thinking perhaps a red border around the output target selector box when a profile is selected.  Another option would be to pop up a message reminding the user that the settings would only be saved to the profile and only for options managed by the profile.

picard/ui/options/dialog.py
```diff
@@ -225,11 +226,34 @@ def show_attached_profiles_dialog(self):
...
+            page_name = page.PARENT if page.PARENT in UserProfileGroups.SETTINGS_GROUPS else page.NAME
+            if page_name in UserProfileGroups.SETTINGS_GROUPS:
+                for opt in UserProfileGroups.SETTINGS_GROUPS[page_name]['settings']:
+                    for opt_field in opt.fields:
+                        try:
+                            obj = getattr(page.ui, opt_field)
+                        except AttributeError:
+                            continue
+                        if opt.name in profile_settings:
+                            style = highlight % opt_field
```
zas, line 253: The highlight is great, though it isn't obvious it is related to Options profiles.
Would it be possible to add a tooltip?

picard/ui/options/dialog.py
```diff
@@ -197,6 +190,22 @@ def __init__(self, default_page=None, parent=None):
...
 
+        self.USER_SETTINGS_TITLE = _("User base settings")
+
+        if config.profiles[SettingConfigSection.PROFILES_KEY]:
+            self.ui.profile_frame.show()
+            self.ui.save_to_profile.clear()
+            self.ui.save_to_profile.addItem(self.USER_SETTINGS_TITLE, USER_SETTINGS_PROFILE_ID)
+            index = 0
+            for idx, item in enumerate(config.profiles[SettingConfigSection.PROFILES_KEY], start=1):
+                self.ui.save_to_profile.addItem(item["title"], item["id"],)
```
zas, line 201:
```suggestion
                self.ui.save_to_profile.addItem(item["title"], item["id"])
```

picard/ui/options/dialog.py
```diff
@@ -224,12 +233,62 @@ def show_attached_profiles_dialog(self):
...
+            return self.USER_SETTINGS_TITLE
+        config = get_config()
+        for item in config.profiles[SettingConfigSection.PROFILES_KEY]:
+            if item["id"] == profile_id:
+                return item["title"]
+        return _('Unknown profile')
+
     def switch_profile(self, index):
+        HighlightColors = namedtuple('HighlightColors', ('fg', 'bg'))
+        highlight = "#%s { color: %s; background-color: %s; }"
```
zas, line 247: `highlight` -> `highlight_fmt` ?
It would denote it's a format string.

zas (changes requested):

zas (approved):
LGTM

phw (approved):
Really cool funtionality, thanks!

rdswift:
@phw,

> Really cool functionality, thanks!

Thanks, but this was really a team effort.  The ideas for highlighting the items and auto-selecting the first enabled profile came from discussions with you and @zas.

PR #1871, opened 2021-07-27
Modify date-related functions to optionally split on multiple chars

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Add option to split input date string on single character (default) or multiple characters.

# Problem

The current version (not yet released) of the `$year()`, `$month()`, `$day()` and `$dateformat()` scripting functions splits the input date string on one or more non-digit characters.  This leads to unexpected results such as `$month(2021--16)` returning "16" rather than the expected "".

# Solution

Split the input date string at each non-digit character and add an option to trigger the split at one or more non-digit character.

# Action

Remember to update the documentation.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1386,18 +1386,22 @@ def func_countryname(parser, country_code, translate=""):
 DateTuple = namedtuple('DateTuple', ('year', 'month', 'day'))
 
 
-def _split_date(date_to_parse, date_order="ymd"):
+def _split_date(date_to_parse, date_order="ymd", multi_char=False):
     """Split the specified date into parts.
 
     Args:
         date_to_parse (str): Date string to parse
         date_order (str, optional): Order of date elements. Can be "ymd", "mdy" or "dmy". Defaults to "ymd".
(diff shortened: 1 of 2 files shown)


zas:
> This leads to unexpected results such as `$month(2021--16)` returning "16" rather than the expected "".

Actually I'm not sure what to expect in such case. Dates are usually separated by one character, or none in the case of "YYYYMMDD" format (I would rather support that), but very rarely by more than one character, or those are whitespaces ("2021 - 12 - 17"). I really don't expect "2021--16" to be a valid format.

I'm not fond of this move as it makes code more complex, or we should let the user specifies a full format (which would lead to extra complexity too).

Did we get reports from users asking for this?

rdswift:
> I'm not fond of this move as it makes code more complex, or we should let the user specifies a full format (which would lead to extra complexity too).

Good point.

> Did we get reports from users asking for this?

Just from me, and I can live without it.  Would you like me to cancel this PR?

phw:
I agree with zas on this.

PR #1872, opened 2021-07-27
PICARD-2250: Add `$parsedate()` function

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Add new `$parsedate()` function that uses the `dateutil.parser.parse` method.

# Problem

* JIRA ticket (_optional_): PICARD-2250

There is currently no scripting function capable of parsing dates in a variety of formats, including months as text.

# Solution

Add a new `$parsedate()` function that uses the `dateutil.parser.parse` method.  This allows parsing input dates such as "September 23, 2017" or "23 Sept 2017" into "YYYY-MM-DD" format for use with other date-related scripting functions.

# Action

Update documentation.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -44,6 +44,8 @@ import operator
 import re
 import unicodedata
 
+from dateutil.parser import parse as du_parse
+
 from picard.const.countries import RELEASE_COUNTRIES
 from picard.metadata import MULTI_VALUED_JOINER
 from picard.script.parser import (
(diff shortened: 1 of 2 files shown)


picard/script/functions.py
```diff
@@ -1493,3 +1495,33 @@ def func_dateformat(parser, date_to_parse, date_format=None, date_order='ymd'):
...
+
+
+@script_function(documentation=N_(
+    """`$parsedate(date,first_element=None)`
+
+Parses the input date string and returns the resulting date in the form "YYYY-MM-DD". The
+function will try to automatically determine the date based on the content of each element
+(e.g. text month, day greater than 12, four-digit year, etc.). When date information is
+ambiguous (e.g. 06-07-08) the system will parse it in the order "mdy", but you can force
+it to use "dmy" or "ymd" by setting `first_element` to "d" or "y" respectively.
```
zas, line 1507: It would be more consistent with other date-related functions to use `date_order="ymd"` as default, and accept "dmy" and "mdy".

phw:
My only concern is that I had thought after our recent discussion that we might get completely rid  of the dateutil dependency.

Is there any concrete use case for this function?

@zas what do you think?

picard/script/functions.py
rdswift: I agree that would be more consistent with the other date-related functions.

rdswift:
> My only concern is that I had thought after our recent discussion that we might get completely rid of the dateutil dependency.

Oh right.  I forgot about that.

> Is there any concrete use case for this function?

I have no examples.

I was only adding this to complete our date handling functions.  Since nobody else has requested such functionality, perhaps it might be best to drop it for now and cancel this PR.

zas:
> @zas what do you think?

Well, we can introduce such feature at any time, and we wanted to get rid of `dateutil` dependency, so I would keep this PR for later.

picard/script/functions.py
rdswift: Even though we're putting this on hold for now, I pushed the changes to use "ymd" and "dmy" as the argument options so I didn't forget about it.

PR #1873, opened 2021-07-29
Catch exception when re-parenting SingletonDialog instance

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Catch exception

# Problem

When the automatic parent change code was added in PR #1852 it introduced a bug that would crash Picard when opening the scripting documentation from the tagging script options dialog.  The exception was:

```
  File ".\picard\ui\options\scripting.py", line 111, in show_scripting_documentation
    ScriptingDocumentationDialog.show_instance(parent=self)
  File ".\picard\ui\__init__.py", line 155, in show_instance
    if 'parent' in kwargs and instance.parent() != kwargs['parent']:
TypeError: 'ScriptingOptionsPage' object is not callable
```

# Solution

Catch the exception generated.

# Action

### DIFF
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -151,12 +151,17 @@ class SingletonDialog:
     @classmethod
     def show_instance(cls, *args, **kwargs):
         instance = cls.get_instance(*args, **kwargs)
+        # Get the current parent
+        if hasattr(instance, 'parent'):
+            if callable(instance.parent):
+                parent = instance.parent()
+            else:
+                parent = instance.parent
+        else:
+            parent = None
         # Update parent if changed
-        try:
-            if 'parent' in kwargs and instance.parent() != kwargs['parent']:
-                instance.setParent(kwargs['parent'])
-        except TypeError:
-            pass
+        if 'parent' in kwargs and parent != kwargs['parent']:
+            instance.setParent(kwargs['parent'])
         instance.show()
         instance.raise_()
         instance.activateWindow()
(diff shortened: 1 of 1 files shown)


zas:
> if 'parent' in kwargs and instance.parent() != kwargs['parent']:

Shouldn't it be `instance.parent != kwargs['parent']` ??

phw:
> Shouldn't it be `instance.parent != kwargs['parent']` ??

No, QWidget.parent() is a method.

@rdswift What kind of errors can happen here?

But patch looks fine to me

rdswift:
> @rdswift What kind of errors can happen here?

The only one I've encountered is the `TypeError` message that I posted, and I'm not 100% sure that I understand why that is happening.  That actually has me wondering if we're setting the correct parent for the script editor dialog, which might be contributing to the window issue on macOS.  I'm looking into this a bit further.

rdswift:
Actually, the problem turned out to be that the `ScriptingDocumentationDialog` object is overwriting the `.parent()` method with a `.parent` attribute in the class definition. 

https://github.com/metabrainz/picard/blob/51db740a593a7c4b3a4d8236765860cc1a4a0eb9/picard/ui/options/scripting.py#L74

To be on the safe side, the re-parenting routine now checks for either `.parent` or `.parent()`, and defaults to `None` if neither is found.

zas (approved):
LGTM

PR #1874, opened 2021-07-29
PICARD-2251: Add basic tagging script import/export functions

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Add basic tagging script import/export functions.

# Problem

* JIRA ticket (_optional_): PICARD-2251

Allow importing and exporting tagging scripts similarly to the functionality provided for file naming scripts.

# Solution

Add basic tagging script import/export functions:

- Add buttons to UI and rearrange buttons
- Make all button sizes consistent within the UI (using QToolButton)
- Add import and export functions
- Add missing button tooltips

# Action

If this concept is accepted, we should consider refactoring to reduce code duplication between the file naming and tagging script import/export functions.


phw:
Definitely makes sense to have this functionality here, too. I'll give this a test run to see the UI

picard/ui/options/scripting.py
```diff
@@ -106,10 +123,141 @@ def __init__(self, parent=None):
...
+        if items:
+            item = items[0]
+            script_text = item.script
+            script_title = item.name
+        else:
+            script_text = ""
+            script_title = ""
+
+        if script_text:
+            script_item = PicardScript(title=script_title, script=script_text)
```
zas, line 219: Since `item` is only used there, I would reduce indentation and return immediatly if no items are selected.

```python
items = self.ui.script_list.selectedItems()
if not items:
   return

script_text = items[0].script
script_title = items[0].name

if script_text:
    script_item = PicardScript(title=script_title, script=script_text)
```

zas (changes requested):

picard/ui/options/scripting.py
rdswift: Much cleaner the way you suggested.  Thanks.

I also added a default script title if `item.name` is blank, which Picard currently allows.  Should I fix this so that blank names are not accepted when changing the name of a script?

picard/script/serializer.py
```diff
@@ -245,6 +250,24 @@ def create_from_yaml(cls, yaml_string, create_new_id=True):
             new_object._set_new_id()
         return new_object
 
+    @property
+    def filename(self):
+        return self.make_filename(self.title)
+
+    @classmethod
+    def make_filename(cls, title=None):
```
zas, line 258: Instead of a class method we could make this one a bit more generic and move it to utils:

```python
def make_filename_from_title(title=None, default=None):
    if default is None:
        default = _("No Title")
    if not title:
        title = default
    filename = sanitize_filename(title, win_compat=IS_WIN)
    if IS_WIN:
        filename = replace_win32_incompat(filename)
    return filename
```

And change caller accordingly:
```python

    @property
    def filename(self):
        return make_filename_from_title(self.title, _("Unnamed Script"))
```

It could be tested independently.

What do you think?

picard/ui/options/scripting.py
```diff
@@ -106,10 +123,141 @@ def __init__(self, parent=None):
...
+        script_title = item.name if item.name.strip() else _("Unnamed Script")
+
+        if script_text:
+            script_item = PicardScript(title=script_title, script=script_text)
+            script_filename = ".".join((script_item.filename, self.default_script_extension))
+            default_path = os.path.normpath(os.path.join(self.default_script_directory, script_filename))
+            dialog_title = _("Export Script File")
+            dialog_file_types = self._get_dialog_filetypes()
+            options = QtWidgets.QFileDialog.Options()
+            options |= QtWidgets.QFileDialog.DontUseNativeDialog
```
phw, line 224: Is there any specific reason this does not allow the native dialog? Usually we should prefer the native one, unless we need some functionality that only the Qt dialog provides.

phw (commented):
Looks good to me to merge, just a question about setting DontUseNativeDialog. If that's not really needed we should not set it.

picard/ui/options/scripting.py
rdswift: No reason.  Just got caught up with a copy'n'paste from elsewhere in the code.  I'll fix that.

zas (approved):
LGTM

phw (approved):

PR #1877, opened 2021-08-01
Add options for editing the script title

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**:  Add context menu item on script title to open the metadata editor.

# Problem

The file naming script editor does not currently allow changing the title of a script from the main screen, but there was no indication to the user how the title could be changed.

# Solution

Add options for editing the script title.  Also fix the title display of the metadata editor window.

- Add double click action to open the script metadata editor
- Add right click context menu to open the script metadata editor
- Fix title display of script metadata editor window

# Action

Remember to update the documentation.

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -475,6 +475,11 @@ class ScriptEditorDialog(PicardDialog, SingletonDialog):
         self.ui.buttonbox.addButton(QtWidgets.QDialogButtonBox.Help)
         self.ui.buttonbox.helpRequested.connect(self.show_help)
 
+        # Add links to edit script metadata
+        self.ui.script_title.installEventFilter(self)
+        self.ui.script_title.setContextMenuPolicy(QtCore.Qt.ActionsContextMenu)
+        self.ui.script_title.addAction(self.details_action)
+
         self.ui.file_naming_format.setEnabled(True)
 
         # Add scripting documentation to parent frame.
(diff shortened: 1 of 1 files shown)


picard/ui/scripteditor.py
```diff
@@ -649,6 +654,8 @@ def eventFilter(self, object, event):
         evtype = event.type()
         if evtype in {QtCore.QEvent.WindowActivate, QtCore.QEvent.FocusIn}:
             self.update_examples()
+        elif object == self.ui.script_title and evtype == QtCore.QEvent.MouseButtonDblClick:
+            self.details_action.trigger()
```
zas, line 658: Should'nt we return `True` to prevent further event processing in this case?

https://doc.qt.io/qtforpython/overviews/eventsandfilters.html#event-filters

phw: I think yes. It probably has no real effect here, because there is not other default action on double click, but still would be the proper way to do this to avoid potential other effects.

phw:
@rdswift Works well, that's definitely great. Allows for an easy way to share scripts.

But as you suggested on IRC I would actually go for using the script name as the filename. We need to create a safe file name, but that's something we have already solved in Picard. We have the `sanitize_filename`  and for Windows also the `replace_win32_incompat` functions.

rdswift:
> But as you suggested on IRC I would actually go for using the script name as the filename.

I'll do that as a separate PR because it's actually a bit different in that I propose to do that for both tagging script and file naming scripts.

zas (approved):
LGTM

PR #1880, opened 2021-08-02
Option settings save issues regarding profiles

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other

# Problem

From IRC comments by @phw (outsidecontext):

> **\<outsidecontext>** I got two strange things with the new profile setup that we should improve.
> **\<outsidecontext>** 1. If you start with an empty Picard config and have not profiles at all, the profile selection in option dialog is not shown (good). But saving always shows the message "Changes will only be saved to the profile Unknown profile". Instead it should save directly to base settings
> **\<outsidecontext>** 2. If you add a profile this is active by default in options. If you change settings not part of the profile they are not saved at all. That is unexpected. I am not sure what the best solution is. Two ideas:
> **\<outsidecontext>** a) have "User base settings" active by default
> **\<outsidecontext>** b) Save all changed options that are not part of the profile to the base settings instead.
> **\<outsidecontext>** b) is probably difficult to implement?

# Solution

For item 1, check for the existence of defined profiles before displaying popup warning message.

Item 2 needs a decision (see below).

# Action

It feels like we may not all have the same understanding of how Item 2 should be addressed.  It was originally coded to save the individual option settings to the first enabled profile managing that setting (with the "User base settings" always enabled).  This was changed in PR #1865 as a result of [user feedback / discussion in the forum](https://community.metabrainz.org/t/introducing-option-profiles-in-picard/540379/7), requesting a more straightforward and intuitive approach.  Should we revert the changes made and go back to the original design, or keep the current design, or something different?


zas:
> **<outsidecontext>** 2. If you add a profile this is active by default in options. If you change settings not part of the profile they are not saved at all. That is unexpected. I am not sure what the best solution is. Two ideas:
> **<outsidecontext>** a) have "User base settings" active by default
> **<outsidecontext>** b) Save all changed options that are not part of the profile to the base settings instead.
> **<outsidecontext>** b) is probably difficult to implement?

To me, option that isn't part of top active profile should be saved to first active profile having this option in its set, defaulting to bottom profile (aka User base settings).

The underlying logic is the following:

- save option A
- explore active profiles stack from top to bottom (bottom = User base settings)
- if option A is in set, save to this profile and stop

The profile selector was added in Options, but imho is misleading when user have more than one active profiles. Because profile sets are related to each individual options, but profile selector is displayed for each page of options.
At best, we can display the highest profile in the stack having an option displayed in the page in its set.

So if we have 2 options on the same page, belonging to different active proflles:
- PA containing option A is above PB containing option B
- both options A & B are on same options page
- the selector displays PA as selected (because it contains option A and is above PB)
- user can select either PA or PB

This creates some confusion.

Now, I'm not sure what is the proper solution:
- remove the selector totally ?
- keep the selector but improve it somehow ?

rdswift:
> To me, option that isn't part of top active profile should be saved to first active profile having this option in its set, defaulting to bottom profile (aka User base settings).

So as I understand it, you're saying to revert the changes and go back to the original design that the users found so confusing?  In that case, there's no need for the profile selector combo box other than to highlight which options are managed by the selected profile.  It won't have any impact on how the option settings are saved, so I suggest changing the label to something like "Highlight settings for profile:" and change the default from "User base settings" to "No profile selected".

Once you've confirmed this, I'll make the changes.  Thanks.

zas:
> **EDIT 2:** Building on that, perhaps remove the selector entirely, highlight all options associated with active profiles (with a tooltip showing which profile will receive the update), and just include a note (label) indicating that "Settings highlighted will be saved to a profile instead of the base settings".

Yes, it seems a sane option.

@phw: what's your take on this?

rdswift:
I pushed a commit for the "EDIT 2" option.  That way you will be able to see it in action, and may help clarify my thinking.

zas:
> I pushed a commit for the "EDIT 2" option. That way you will be able to see it in action, and may help clarify my thinking.

I just tested, it doesn't work well for me, mainly because accessing Profile editor or switching target profiles before saving are missing.

I wonder this could be solved by integrating Profile editor directly in Options dialog:
- one could switch to Profiles sub-pages to modify profiles and/or check which are set/enabled
- since profiles are directly connected to Options it makes sense
- it would reduce the number of dialogs (getting rid of separate Profile editor and Attached Profiles dialogs, using one option page instead)
- we can keep highlight and tooltip to indicate per-option profile
- user can edit profiles and those will be saved on Make It So button

But, to be frank, I'm not sure what's the proper UI yet, for sure, EDIT 2 approach doesn't solve much for me.

rdswift:
I agree that this is far from ideal, but I thought this was what you wanted to see -- always display and update the settings based on the currently enabled profiles in their specified order.  Obviously I was wrong.

I don't think it would be that difficult to move the profile editor to a page in the Options... dialog (famous last words :wink:), but I'll have to think about how the processing would be handled.  For example, any profile changes would have to be immediately shown in the other option pages, but not be saved until `Make it so!`.  This means that all highlighting would have to be done based on the settings on the Profiles page rather than the settings in `config.settings`, and would have to be updated every time a change is made to a profile.  Also, we would need to make sure that the Profile page settings are saved to `config.profile` before saving any of the other pages.

I believe that the main issue is going to be updating the values in all of the pages whenever a profile change is made, because the settings selected are based on the profile information in `config.profiles`.  I have an idea of how this might be done, but I'm afraid that updating all of the settings displayed (even if we limit the page updates to those containing "profile-able" options) for every change made to a profile might still be noticeably slow.

In any event, clearly the current commit is unacceptable.

rdswift:
This latest commit moves the profiles editor to the second page of the Options... dialog.  Highlighting and settings are updated after each change to a profile, always showing the values based on the first enabled profile.  Settings will be saved to the first enabled profile.  The update speed isn't as bad as I expected, except when you change all profile settings for a section at once by clicking the checkbox at the settings group (page) level.  This is because the update settings code is called once for each of the child box changes.  Definitely a possible area for improvement.

rdswift:
I'm not sure why CodeQL is suddenly complaining about two files that are not part of this PR.  It appears that it is triggering on the variable name "id".  Do you want me to make those changes to shut CodeQL up, even though they have nothing to do with this PR?

zas:
This last iteration is much better to me:

- integration of Profile editor in Options dialog makes much more sense
- the code is simplified
- it limits Profiles <-> Options dialogs switch

There are two things still bothering me:

- Profiles Help and Attached Profiles buttons are always displayed, though not always useful
- Profiles Help should only be displayed in Profiles Pane imho
- Attached Profiles should only be displayed if there are "profiled" options in the currently displayed pane
- wasting one full "line" for only one or two buttons at the bottom of the dialog isn't great,

We could perhaps move Attached Profiles button on bottom line, and just enable/disable it whenever there are options with profiles in the page, and get rid of the info message about highlight (we have tooltips now).

Something like:


image

What do you think?

phw:
Yes, I think this works better. It avoids the most confusing part that some options would not get saved, and the highlighting still makes clear there is a profile and what options are affected.

Is it possible to show the profile name in the "Highlighted options will be saved to a profile" message, so it shows "Highlighted options will be saved to "My profile 1"" or such?

rdswift:
> There are two things still bothering me:
> 
> * Profiles Help and Attached Profiles buttons are always displayed, though not always useful
> * Profiles Help should only be displayed in Profiles Pane imho
> * Attached Profiles should only be displayed if there are "profiled" options in the currently displayed pane
> * wasting one full "line" for only one or two buttons at the bottom of the dialog isn't great,
> 
> We could perhaps move Attached Profiles button on bottom line, and just enable/disable it whenever there are options with profiles in the page, and get rid of the info message about highlight (we have tooltips now).

I like this.  Also, there's no need for the separate "Help" button.  That was a hold over from the stand-alone dialog that I forgot to remove.

> Maybe change the wording to "Highlighted options will be saved to an active profile"?

Or as @zas says, do we need this line at all?  Also, the name of the profile that will be updated is shown as a tooltip on all highlighted options.

> I also think we need some quick switch in the UI to toggle profiles. What about a sub menu entry in "Options", in which all profiles are listed and you can enable / disable them (similar to "Select file naming script...", but allowing multiple active ones).

I was thinking the same thing.  I'll see if I can add something like that.

phw:
> Or as @zas says, do we need this line at all? Also, the name of the profile that will be updated is shown as a tooltip on all highlighted options.

I found it useful to figure out what is going on, but maybe we can get away without this being displayed.

zas:
> > Or as @zas says, do we need this line at all? Also, the name of the profile that will be updated is shown as a tooltip on all highlighted options.
> 
> I found it useful to figure out what is going on, but maybe we can get away without this being displayed.

Yes, but as it uses a full line height, and since we have tooltips explaining what's going on, I think we can rid of it. Most users will not use profiles I guess, and we can add a note about this highlight in the help doc.

phw:
Agreed, let's do it that way.

rdswift:
> Agreed, let's do it that way.

Crap!  I didn't see this before I pushed my latest commit that revised the text of the highlight note and only showed it on pages where there are actually options highlighted.  I'll push another commit to remove it completely, but please have a look at this version to see if you might change your minds and want to keep it.

phw:
I like the wording better. I'm personally a bit unsure about this. It uses some extra space, and zas is right that it can be explained elsewhere. But I also find it useful information: Here is a screenshot

grafik

zas:
> It uses some extra space, and zas is right that it can be explained elsewhere. But I also find it useful information:

I agree it's useful information, but once user understood once what it means, it doesn't help much anymore, and just take space.
Also such color-coded information isn't really [accessible](https://en.wikipedia.org/wiki/Computer_accessibility) (as already said, we have too much of those in Picard already), so it shouldn't be too pro-eminent IMHO.

The usual approach for such information is rather to display it as a message dialog with checkbox to not show again instead, when user displays an option pane with such highlight.

zas:
@rdswift it works much  better like this, at least for me. Profiles can still be easily switching from main menu, but configuration is easier from Options. Good work!

phw (dismissed):
Yes, this definitely is an improvement. For users not using profiles everything basically just works as before, but also if you are using profiles it has become easier to handle. LGTM

zas (approved):
Better UI on almost all points, and code is simpler. Good work!

PR #1883, opened 2021-08-08
Add profile option setting tooltips and fix saving persistent settings

rdswift:
# Summary

* This is a…
    * [x] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

In the course of changing the profile editor from a stand-alone dialog to an options page, the tooltips showing the current value for a profile option setting were lost.

Also, the persistent settings were no longer being updated because the dialog close event was not being triggered.

# Solution

Re-implement the profile option settings tooltips, and move saving the persistent settings to the `save` method on the page.

# Action


zas (approved):
LGTM

PR #1885, opened 2021-08-08
PICARD-157: Flexible language alias

vladkar:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

    Picard has great feature: "translate artist name". It could be used when you have e.g. 
    Japanese music and you don't know Japanese and even cant navigate, so it would be a 
    good point to represent it in English. However, this feature translates everything (if you
    process **all** your collection at once) which is not very convenient if you want to keep
    some languages. For example you want to have all your collection in English except your
    native language music (in my case I want to translate Japanese, but I don't want to translate Russian).
    
    This feature gives possibility to ignore selected most popular character set or script
    (Latin, Cyrillic, Greek, Arabic, Hebrew, Chinese, Hangul, Hiragana, Katakana, Thai) 
    when you use "translate artist name"

# Solution

    Implementation is relatively simple. It includes one checkbox (to use ignore feature) 
    and one droplist (to select required character set) in "Options/Metadata" and logic to 
    process it.

# Action

    Feel free to review and comment any part of the PR.


picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+    def only_alphabet_chars(self, unistr, alphabet):
+        return all(self.is_in_alphabet(uchr, alphabet)
+                   for uchr in unistr if uchr.isalpha())
+
+    def detect_alphabet(self, unistr):
+        return set(ud.name(char).split(' ')[0]
+                   for char in unistr if char.isalpha())
+
+    def is_greek(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'GREEK') else False
```
zas, line 51: You could simplify:

```suggestion
        return self.only_alphabet_chars(unistr, 'GREEK')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_cyrillic(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'CYRILLIC') else False
```
zas, line 54:
```suggestion
        return self.only_alphabet_chars(unistr, 'CYRILLIC')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_latin(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'LATIN') else False
```
zas, line 57:
```suggestion
        return self.only_alphabet_chars(unistr, 'LATIN')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_arabic(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'ARABIC') else False
```
zas, line 60:
```suggestion
        return self.only_alphabet_chars(unistr, 'ARABIC')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_hebrew(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'HEBREW') else False
```
zas, line 63:
```suggestion
        return self.only_alphabet_chars(unistr, 'HEBREW')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    # NOTE: this only detects Chinese script characters (Hanzi/Kanji/Hanja).
+    # it does not detect other CJK script characters like Hangul or Katakana
+    def is_cjk(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'CJK') else False
```
zas, line 68:
```suggestion
        return self.only_alphabet_chars(unistr, 'CJK')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_hangul(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'HANGUL') else False
```
zas, line 71:
```suggestion
        return self.only_alphabet_chars(unistr, 'HANGUL')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_hiragana(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'HIRAGANA') else False
```
zas, line 74:
```suggestion
        return self.only_alphabet_chars(unistr, 'HIRAGANA')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_katakana(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'KATAKANA') else False
```
zas, line 77:
```suggestion
        return self.only_alphabet_chars(unistr, 'KATAKANA')
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+
+    def is_thai(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'THAI') else False
```
zas, line 80:
```suggestion
        return self.only_alphabet_chars(unistr, 'THAI')
```

zas (changes requested):

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,80 @@
...
+        return self.only_alphabet_chars(unistr, 'HANGUL')
+
+    def is_hiragana(self, unistr):
+        return self.only_alphabet_chars(unistr, 'HIRAGANA')
+
+    def is_katakana(self, unistr):
+        return self.only_alphabet_chars(unistr, 'KATAKANA')
+
+    def is_thai(self, unistr):
+        return True if self.only_alphabet_chars(unistr, 'THAI') else False
```
zas, line 80:
```suggestion
        return self.only_alphabet_chars(unistr, 'THAI')
```

zas (changes requested):
You'll need to fix tests (I enabled workflows)

vladkar:
@zas thank you for your feedback, I updated the tests, but it seems I can't run workflows by myself.

picard/ui/options/metadata.py
```diff
@@ -30,7 +31,7 @@
     TextOption,
     get_config,
 )
-from picard.const import ALIAS_LOCALES
+from picard.const import ALIAS_LOCALES, CHAR_SETS
```
zas, line 34:
```suggestion
from picard.const import (
    ALIAS_LOCALES,
    CHAR_SETS,
)
```

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,79 @@
...
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY,
+# FITNESS FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE
+# AUTHORS OR COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import unicodedata as ud
+from collections import defaultdict
```
zas, line 26: You need to re-order imports, `isort` can help (see https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#coding-style)

```patch
-import unicodedata as ud
 from collections import defaultdict
+import unicodedata as ud
```

zas (changes requested):

vladkar:
- I reordered imports that was added or modified within this pr
- Tests were passed on my local pc and in ci pipeline too

rdswift (commented):
I tested the changes on Windows 7, and everything seems to work as expected.  I couldn't find any issues.

picard/mbjson.py
```diff
@@ -192,6 +194,16 @@ def _translate_artist_node(node):
...
         locale = config.setting["artist_locale"]
         lang = locale.split("_")[0]
+
+        if config.setting['translate_artist_names_exception']:
+            char_set_exception = config.setting["artist_locale_exception"]
+
+            ad = alphabet_detector.AlphabetDetector()
+            char_sets = ad.detect_alphabet(node['name'])
+
+            if char_set_exception in char_sets:
```
zas, line 204: Since `char_set_exception` is only used here, I think we can get rid of it:

```python
            if config.setting["artist_locale_exception"] in char_sets:
```

And if we go for a simple method:

```python
    if (config.setting['translate_artist_names_exception'] and 
        config.setting["artist_locale_exception"] in detect_alphabet(node['name'])):
        return node['name'], node['sort-name']
```

zas (changes requested):

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,79 @@
...
+            return self.alphabet_letters[alphabet].setdefault(
+                uchr, alphabet in ud.name(uchr))
+
+    def only_alphabet_chars(self, unistr, alphabet):
+        return all(self.is_in_alphabet(uchr, alphabet)
+                   for uchr in unistr if uchr.isalpha())
+
+    def detect_alphabet(self, unistr):
+        return set(ud.name(char).split(' ')[0]
+                   for char in unistr if char.isalpha())
```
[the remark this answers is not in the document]
vladkar: To be honest I needed working solution for this problem and it exists (https://github.com/EliFinkelshteyn/alphabet-detector). But instead of having additional dependency I decided to assimilate it with slight changes. It is simple, it works, it was tested by main author, and license allows to do it. If you say we have to rewrite it - we could but should we?

picard/ui/options/metadata.py
```diff
@@ -97,6 +103,16 @@ def load(self):
             if locale == current_locale:
                 combo_box.setCurrentIndex(i)
 
+        self.ui.translate_artist_names_exception.setChecked(config.setting["translate_artist_names_exception"])
+
+        combo_box = self.ui.artist_locale_exception
+        current_locale = config.setting["artist_locale_exception"]
+        for i, (locale, name, level) in enumerate(iter_sorted_locales(CHAR_SETS)):
+            label = "    " * level + name
```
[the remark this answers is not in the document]
vladkar: Expression was simplified (push is coming). Initially I used the same approach as 10 lines above (and it works).

picard/util/alphabet_detector.py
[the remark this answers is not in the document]
vladkar: Anyway done with simplification, will add test later

picard/ui/options/metadata.py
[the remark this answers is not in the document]
vladkar: Done

picard/mbjson.py
```diff
@@ -192,6 +194,16 @@ def _translate_artist_node(node):
     if config.setting['translate_artist_names']:
         locale = config.setting["artist_locale"]
         lang = locale.split("_")[0]
+
+        if config.setting['translate_artist_names_exception']:
+            char_set_exception = config.setting["artist_locale_exception"]
+
+            ad = alphabet_detector.AlphabetDetector()
+            char_sets = ad.detect_alphabet(node['name'])
```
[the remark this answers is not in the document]
vladkar: Done

picard/mbjson.py
vladkar: This simplification is subjective, for example for me the form I used is simpler. Anyway done.

vladkar:
@zas thank you for your feedback

picard/util/alphabet_detector.py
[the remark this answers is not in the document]
vladkar: Test for alphabet detection was added

rdswift:
This may sound like a dumb question, but would there ever be a case where the user would want to select more than one exception?

picard/ui/options/metadata.py
```diff
@@ -97,6 +103,16 @@ def load(self):
             if locale == current_locale:
                 combo_box.setCurrentIndex(i)
 
+        self.ui.translate_artist_names_exception.setChecked(config.setting["translate_artist_names_exception"])
+
+        combo_box = self.ui.artist_locale_exception
+        current_locale = config.setting["artist_locale_exception"]
+        for i, (locale, name, level) in enumerate(iter_sorted_locales(CHAR_SETS)):
+            label = name
+            combo_box.addItem(label, locale)
```
rdswift, line 112: Could this be shortened to:

```
combo_box.addItem(name, locale)
```

Unless I'm missing something, `label` isn't used anywhere else.

rdswift:
Am I correct in assuming that this functionality only comes into play if the "Translate artist names to this locale where possible:" option is checked?  If so, then I suggest that when it is not checked, both the "Ignore artist name translation for character set:" checkbox and the combo box selector should be disabled.

Similarly, the combo box selector should be disabled when the "Ignore artist name translation for character set:" option is not checked.

rdswift:
I believe that the MusicBrainz website refers to these as "script" rather than "character set" (with a more extensive list).  I wonder if we should be using the same terminology here for the sake of consistency, even though I like the term "character set"?

phw (commented):
This is great. I think essentially this fixes PICARD-157

We have some related tickets like PICARD-411 and PICARD-412 regarding this functionality. I think especially PICARD-411 could be tackled once we have this merged.

But this looks good. Once @rdswift's minor comment is addressed I think this could be merged. I think the commits also should be squashed, I think the change is small enough for a single commit. Would be great if you could squash the commits, but if not we can do it while merging.

picard/mbjson.py
```diff
@@ -192,6 +194,11 @@ def _translate_artist_node(node):
     if config.setting['translate_artist_names']:
         locale = config.setting["artist_locale"]
         lang = locale.split("_")[0]
+
+        if (config.setting['translate_artist_names_exception'] and
+                config.setting["artist_locale_exception"] in detect_alphabet(node['name'])):
```
phw, line 199: Our coding style wants the `and` to be at the beginning of the line, so this should do.

```
        if (config.setting['translate_artist_names_exception']
                and config.setting["artist_locale_exception"] in detect_alphabet(node['name'])):
```

phw (changes requested):

picard/util/alphabet_detector.py
```diff
@@ -0,0 +1,30 @@
...
+# LIABILITY, WHETHER IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM,
+# OUT OF OR IN CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN
+# THE SOFTWARE.
+
+import unicodedata as ud
+
+
+def detect_alphabet(unistr):
+    return set(ud.name(char).split(' ')[0]
+               for char in unistr if char.isalpha())
```
phw, line 30: Just a thought: Maybe an improvement would be to not only return a set, but a weighted list depending on the number of occurrences for a script. E.g. Japanese names sometimes also contain Latin characters. If you had configured artist_locale_exception for latin that name would still be considered latin. But if we would weight the scripts by occurence we could say e.g. primary script is "HIRAGANA", even though it also contains "LATIN".

rdswift: This could also provide some intelligent basis for selecting which alias to display if we eventually allow multiple script/char sets to be enabled.

vladkar:
Update summary:
1. Character set was renamed to script
2. UI toggle was added for relevant UI elements (when translation on/off)
3. List of script exceptions might be new feature, but out of scope for this PR

rdswift:
@vladkar 

I just noticed that you have applied your changes directly to the `picard/ui/ui_options_metadata.py` file rather than the `ui/options_metadata.ui` file.  This means that your changes will be lost whenever a change is made to the `ui/options_metadata.ui` file because `picard/ui/ui_options_metadata.py` is auto-generated (and should not be edited).  I'll submit a PR against your repo for the necessary changes.

rdswift:
@vladkar 

For some reason, I am unable to create a pull request against your repo.  The changes to be merged can be found in https://github.com/rdswift/picard/commit/b9cdd305032cb808fdf484d0315c9b012600abd1

vladkar:
@rdswift 

Thank you for the fix, I managed to merge it.
I also slightly updated .gitignore by adding there standard python virtual env (normally it is a part of typical python .gitignore). Not sure it deserves separate pr.

zas:
> I also slightly updated .gitignore by adding there standard python virtual env (normally it is a part of typical python .gitignore). Not sure it deserves separate pr.

It is unrelated to this PR, so I cherry-picked this change and pushed it as https://github.com/metabrainz/picard/commit/99ed558176cf1c7283551db842ca07bc08913fe5

This PR will need a rebase anyway.

zas:
@vladkar can you please fix coding style issues (isort) and clean up this PR a bit (squash commits, make history cleaner) ?

vladkar:
@zas 
isort: I honestly tried to reproduce the behavior of isort multiple times, but it is quite peculiar and not so obvious (compared, for example, to black formatter I normally use), so I kindly ask your help with it

squash: github can do it automatically when pr is merged (if squash option is selected)

zas:
> isort: I honestly tried to reproduce the behavior of isort you use multiple times, but it is quite peculiar and not so obvious (compared, for example, to black formatter I normally use), so I kindly ask your help with it

Did you read https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#coding-style ?
Did you run `isort -rc .` ?

Note: check dev requirements, recent versions of isort have a different behavior.

> 
> squash: github can do it automatically when pr is merged (if squash option is selected)

Yes, of course, but this isn't what I'm asking.
I'd like you to do a manual interactive rebase, to squash few commits (not all) and make a clean PR (this one has far too many commits for a simple set of changes).
You may want to rewrite few commit messages (fixing capitalization while at it) too.

vladkar:
I did read it and as I mentioned I tried it multiple times:
 - "isort -rc ." did not work out of the box for me, it throws exception on some files (mb they should be added to skip list in isort.cfg?), but this is not the main problem
 - isort makes changes for some other files in the project I never touch
 - isort does not make changes on files I add/edit
 - and I am sure it uses provided isort.cfg
 - of cource it is not my responsibility, but threre are more convinient and popular (means with better support) tools to solve same problem (https://python.libhunt.com/compare-isort-vs-black)
 
I don't know what is the exact reason of such behavior (version of isort or whatever), but as I mentioned, I can not reproduce the expected behavior of isort.

vladkar:
I'll squash all the stuff when we agree on the rest.

phw (approved):
Looks great, thanks a lot!

> UPD: the reason was in conflict between installed versions of isort
Yes, isort is very tricky in this regard. The current version 5.x breaks things again, but we should update at some point and try to configure it properly. Good thing is that isort configuration has become more stable with newer releases.

zas (approved):

phw:
I squashed and merged this locally, see https://github.com/metabrainz/picard/commit/7224961bc47870e69f42702c05a33fd1a8a4ea96

PR #1887, opened 2021-08-20
PICARD-2263: Save performers without role to TMCL / TIPL

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2263

Tags "performer" (without role) do not get saved to TMCL (or TIPL for ID3v2.3), instead a separate TXXX:performer is created. "performer:" (colon, but empty role) did not get saved at all.

# Solution

Save to TMCL with a role of "performer".

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -292,8 +292,12 @@ class ID3File(File):
                         metadata.add(name, text)
             elif frameid == "TMCL":
                 for role, name in frame.people:
-                    if role or name:
+                    if role == 'performer':
+                        role = ''
+                    if role:
                         metadata.add('performer:%s' % role, name)
+                    else:
+                        metadata.add('performer', name)
             elif frameid == "TIPL":
                 # If file is ID3v2.3, TIPL tag could contain TMCL
                 # so we will test for TMCL values and add to TIPL if not TMCL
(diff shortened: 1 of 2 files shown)


zas (dismissed):
LGTM

picard/formats/id3.py
```diff
@@ -292,16 +292,25 @@ def _load(self, filename):
                         metadata.add(name, text)
             elif frameid == "TMCL":
                 for role, name in frame.people:
+                    if role == 'performer':
+                        role = ''
                     if role or name:
                         metadata.add('performer:%s' % role, name)
+                    elif name:
+                        metadata.add('performer', name)
```
kellnerd, line 300: Something looks wrong here, won't the `elif` never run since the above `if` already covers the case that there is a name? I think this should be:
```suggestion
                    if role and name:
                        metadata.add('performer:%s' % role, name)
                    elif name:
                        metadata.add('performer', name)
```

phw: Indeed, thanks for spotting. Actually this can be simplified, no need to check if name is non empty. `metadata.add` takes care of that properly, so the code can be the same as in the "TIPL" case below. Fixed.

zas: @kellnerd good catch, thanks

zas (approved):

PR #1889, opened 2021-08-23
PICARD-2266: Add gapless, podcast, podcasturl, show and showsort to list of known tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2266

The tags gapless, podcast, podcasturl, show and showsort are all supported and mapped by Picard, but not included in the list of known tags.

# Solution

Extend `picard.util.tags`

# Action


zas (approved):

rdswift:
I see these are already included in the tag mapping documentation.  Should they be added to the available tags documentation at https://picard-docs.musicbrainz.org/en/variables/tags_basic.html ?

phw:
>  Should they be added to the available tags documentation at https://picard-docs.musicbrainz.org/en/variables/tags_basic.html ?

In general we should document them, but I'm not sure how. Some of these tags are really specific, e.g. podcast and gapless are basically an iTunes thing, boolean tags that only really are supported for MP4 and maybe MP3. And there are more similar tags that are mapped but by default not filled by Picard, such as ` bpm`, `mood` or `key`.

Currently the tags_basic and tags_advanced sections focus on tags that are being actively filled by Picard (with some exceptions we probably should clarify, like website and acoustid_fingerprint).

I would not document these tags as basic, they definitely are not. My idea was to either add another section to tags_advanced, where we list supported tags not filled by Picard. But into this category also e.g. movement etc. tags fall, and those have their own classical tags section. So this would partly overlap.

Another idea was that we add an entirely new "Other tags" section below "Tags & Variables", were we list only those additional tags Picard knows about, but which it does not fill and which are not listed in other sections (with the exception of "Tags from plugins", I think this can duplicate tags as it is more about how a specific plugin fills these tags).

rdswift:
> In general we should document them, but I'm not sure how.

That's pretty much where I'm at as well.  We sort of have them documented already in the tag mapping so there's no huge rush.

I kinda like your idea of a new section, perhaps called "Other Supported Tags" where the tags could be listed with a bit of a general explanation as to why they are different (similar to the general explanation of how "Advanced" tags differ from "Basic" tags).

PR #1890, opened 2021-08-26
PICARD-2269: Fix "TypeError: arguments did not match any overloaded call" with Python 3.10

sbraz:
# Summary

* This is a…
    * [x] Bug fix

Fix `TypeError: arguments did not match any overloaded call` that occurs when running Picard with Python 3.10.

# Problem

Hi,
I tried running Picard with Python 3.10.0rc1, all tests pass but the GUI doesn't open because of errors such as:
```
  File "./picard/ui/coverartbox.py", line 74, in __init__
    self.shadow = self.shadow.scaled(w, h, QtCore.Qt.KeepAspectRatio, QtCore.Qt.SmoothTransformation)
TypeError: arguments did not match any overloaded call:
  scaled(self, int, int, aspectRatioMode: Qt.AspectRatioMode = Qt.IgnoreAspectRatio, transformMode: Qt.TransformationMode = Qt.FastTransformation): argument 1 has unexpected type 'float'
  scaled(self, QSize, aspectRatioMode: Qt.AspectRatioMode = Qt.IgnoreAspectRatio, transformMode: Qt.TransformationMode = Qt.FastTransformation): argument 1 has unexpected type 'float'
```

This is probably caused by the following change:
> Builtin and extension functions that take integer arguments no longer accept Decimals, Fractions and other objects that can be converted to integers only with a loss (e.g. that have the `__int__()` method but do not have the `__index__()` method). (Contributed by Serhiy Storchaka in [bpo-37999](https://bugs.python.org/issue37999).)

# Solution

Casting the result of `self.scaled` to `int` seems to be enough.

# Action

 I don't understand why that never failed with Python 3.9, if someone does, can they please explain this to me?

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -135,7 +135,7 @@ class CoverArtThumbnail(ActiveLabel):
             event.acceptProposedAction()
 
     def scaled(self, *dimensions):
-        return (self.pixel_ratio * dimension for dimension in dimensions)
+        return (round(self.pixel_ratio * dimension) for dimension in dimensions)
 
     def show(self):
         self.set_data(self.data, True)
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -141,9 +141,9 @@ def get_match_color(similarity, basecolor):
     c1 = (basecolor.red(), basecolor.green(), basecolor.blue())
     c2 = (223, 125, 125)
     return QtGui.QColor(
-        c2[0] + (c1[0] - c2[0]) * similarity,
-        c2[1] + (c1[1] - c2[1]) * similarity,
-        c2[2] + (c1[2] - c2[2]) * similarity)
+        int(c2[0] + (c1[0] - c2[0]) * similarity),
+        int(c2[1] + (c1[1] - c2[1]) * similarity),
+        int(c2[2] + (c1[2] - c2[2]) * similarity))
 
 
 class MainPanel(QtWidgets.QSplitter):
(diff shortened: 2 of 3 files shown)


sbraz:
I have applied the proposed change (but I am just calling `round`, no need for `int` since _The return value is an integer if ndigits is omitted or None_).

picard/ui/coverartbox.py
```diff
@@ -135,7 +135,7 @@ def dropEvent(self, event):
             event.acceptProposedAction()
 
     def scaled(self, *dimensions):
-        return (self.pixel_ratio * dimension for dimension in dimensions)
+        return (int(self.pixel_ratio * dimension) for dimension in dimensions)
```
zas, line 138: > I have applied the proposed change (but I am just calling `round`, no need for `int` since _The return value is an integer if ndigits is omitted or None_).

Actually you're just calling `int()`, I guess `round()` is preferable in this case.

Also, yes, no need of `int(round())` with Python >= 3.6.

It wasn't always the case, hence my error:

```
zas:~$ python2
Python 2.7.18 (default, Mar  8 2021, 13:02:45) 
[GCC 9.3.0] on linux2
Type "help", "copyright", "credits" or "license" for more information.
>>> type(round(0.5))
<type 'float'>
>>> 
zas:~$ python3
Python 3.8.10 (default, Jun  2 2021, 10:49:15) 
[GCC 9.4.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> type(round(0.5))
<class 'int'>
```

zas (changes requested):

phw (approved):

zas (approved):
LGTM

picard/ui/coverartbox.py
sbraz: Yeah sorry about that, I'd forgotten to push my local changes, thanks for the merge :)

PR #1893, opened 2021-08-31
PICARD-2271: Add Options page to allow removing unused settings from INI file.

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2271

Over the course of trying out plugins and such, the configuration INI file can become bloated with orphaned settings.

# Solution

Provide a new Options page (called "Maintenance") that lists all orphaned settings found in the INI file, and allows the user to mark the settings for deletion.

This allows you to remove unused option settings from the configuration file.

Settings that are found in the configuration file that do not appear on any option settings page will be listed. If the configuration file does not contain any unused option settings, then the list will be empty.

Note that unused option settings could come from plugins that have been uninstalled, so care must be taken to not remove settings that you may want to use later when the plugin is reinstalled.  Options belonging to plugins that are installed, including those that are currently disabled, will not be listed for possible removal.

To remove one or more settings, select the settings to remove by checking the box next to the setting, and then enable the removal by checking the `Remove selected options` box. When you choose `Make It So!` to save your option settings, the selected items will be removed.

# Action

If this change is accepted, the documentation will need to be updated.


picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,168 @@
...
+        ]).union(key_options)
+
+        def _add_options_from_page(page, current_options):
+            if not hasattr(page, 'options'):
+                return
+            for option in page.options:
+                if option.section == "setting":
+                    current_options.add(option.name)
+
+        for ep in _extension_points:
```
zas, line 102: Disabled plugins will not be present in this list (see https://github.com/metabrainz/picard/blob/f950188f1f460b4e0b7a5cf599868e9b1c60c43b/picard/plugin.py#L102-L107).
Is this intended?

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,168 @@
...
+            'file_renaming_scripts',
+            'selected_file_naming_script_id',
+            'log_verbosity',
+            # Items missed if TagsCompatibilityWaveOptionsPage does not register.
+            'remove_wave_riff_info',
+            'wave_riff_info_encoding',
+            'write_wave_riff_info',
+        ]).union(key_options)
+
+        def _add_options_from_page(page, current_options):
```
zas, line 95: I'm a bit bothered by this function.
Passing `current_options` as parameter is misleading, as it is also declared outside the function's scope, and it modifies its argument's value.
Also it isn't strictly necesserary since it is only used in following loop.

No suggestion though, as removing it would likely increase indentation level too much. Perhaps we can use list comprehensions below.

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,168 @@
...
+        self.ui.enable_cleanup.setChecked(False)
+        self.ui.tableWidget.clearContents()
+        self.ui.tableWidget.setRowCount(len(orphan_options))
+        for row, item in enumerate(sorted(orphan_options)):
+            tableitem = QtWidgets.QTableWidgetItem(item)
+            tableitem.setFlags(tableitem.flags() | QtCore.Qt.ItemIsUserCheckable)
+            tableitem.setCheckState(False)
+            self.ui.tableWidget.setItem(row, 0, tableitem)
+            text = self.make_setting_value_text(item)
+            line_height = max(25, 18 * min(5, text.count("\n") + 1))
```
zas, line 125: Why this calculation? Please explain.

zas (changes requested):
I like the feature.

picard/ui/options/maintenance.py
rdswift: > Disabled plugins will not be present in this list

That's what I though too, but in my testing (Windows 7 running from source code) the disabled plugin settings still appeared in the list.  Only the plugins that hadn't been downloaded (installed) didn't appear.

picard/ui/options/maintenance.py
rdswift: This is to expand the row height to a reasonable size if there are multiple lines in a "Value".  The default size is only one line with tiny scroll bar.  Should I remove this and just leave the default line height, or leave it in and add a comment with an explanation?

picard/ui/options/maintenance.py
rdswift: I've refactored to remove the function, and still managed to stay within the flake8 indentation limits.

picard/ui/options/maintenance.py
zas: I just tested, with ReplayGain plugin, it registers following options:

```
replaygain_vorbisgain_command
replaygain_vorbisgain_options
replaygain_mp3gain_command
replaygain_mp3gain_options
replaygain_metaflac_command
replaygain_metaflac_options
replaygain_wvgain_command
replaygain_wvgain_options
```

If I print `option.name` (line 100), they do show when plugin is enabled, but they don't if not. That's what it is expected from the code above. So `current_options` does actually differs whether the plugin is enabled or disabled.

That said, it doesn't change the result as none of those options are in final list:

```python
orphan_options = file_options.difference(current_options)
```

When plugin is enabled, `current_options` set is bigger, but `file_options` is unchanged:

```python
>>> file_options = {1, 2, 3}
>>> current_options = {2, 3, 4}
>>> file_options.difference(current_options)
{1}
>>> current_options = {2, 3, 4, 5}
>>> file_options.difference(current_options)
{1}
```

picard/ui/options/maintenance.py
zas: It looks good.

picard/ui/options/maintenance.py
zas: Comment is fine, thanks.

picard/ui/options/maintenance.py
rdswift: Okay, now I think I understand.  I was only looking at the `orphan_options` difference set and it wasn't changing.  I didn't catch the change in the numbers of the current options.

Is this okay as-is, or would you like me to change it so that the `current_options` number also includes the disabled plugins?

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,182 @@
...
+            _(self.INSTRUCTIONS)
+            + _("\n\nThe configuration file currently contains {0} option settings, {1} which are unused.").format(
+                len(file_options),
+                len(orphan_options)
+            )
+        )
+        self.ui.enable_cleanup.setChecked(False)
+        self.ui.tableWidget.clearContents()
+        self.ui.tableWidget.setRowCount(len(orphan_options))
+        for row, item in enumerate(sorted(orphan_options)):
```
zas, line 110: `item` -> `option_name` ?

We already have too much items around ;)

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,182 @@
...
+            tableitem = QtWidgets.QTableWidgetItem(item)
```
zas, line 111: As explained below, it's prolly safer to differentiable table cell label from its data:

```python
            tableitem = QtWidgets.QTableWidgetItem("label:" + option_name)  # as example
            tableitem.setData(QtCore.Qt.UserRole, option_name)
```

And change code below to use data instead of text:

```python
 to_remove.add(item.data(QtCore.Qt.UserRole))
```

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,182 @@
...
+        self.enable_cleanup_changed()
+
+    def save(self):
+        if not self.ui.enable_cleanup.checkState() == QtCore.Qt.Checked:
+            return
+        to_remove = set()
+        for idx in range(self.ui.tableWidget.rowCount()):
+            item = self.ui.tableWidget.item(idx, 0)
+            if item.checkState() == QtCore.Qt.Checked:
+                to_remove.add(item.text())
```
zas, line 135: It isn't very safe to use `item.text()` to retrieve option's name (I guess it works well enough in this case, but ...).

Instead we should use `setData()` and `data()` mechanism.

zas (changes requested):

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,182 @@
...
+        NONE_TEXT = _("None")
+        config = get_config()
+        value = config.setting.raw_value(key)
+        if value is None:
+            return NONE_TEXT
+        if isinstance(value, str):
+            return '"%s"' % value
+        if type(value) in {bool, int, float}:
+            return str(value)
+        if type(value) in {set, tuple, list, dict}:
```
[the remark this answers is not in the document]
rdswift: > Why not just use `repr()` (eventually truncated in length for display) and totally get rid of type checking?

I'm ashamed to say that I never thought of it.  That is MUCH nicer than what I was trying to do.  Thank-you!

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,166 @@
...
+            return
+        to_remove = set()
+        for idx in range(self.ui.tableWidget.rowCount()):
+            item = self.ui.tableWidget.item(idx, 0)
+            if item.checkState() == QtCore.Qt.Checked:
+                to_remove.add(item.data(QtCore.Qt.UserRole))
+        if to_remove and QtWidgets.QMessageBox.question(
+            self,
+            _('Confirm Remove'),
+            _("Are you sure you want to remove the {0} selected option settings?").format(len(to_remove)),
```
zas, line 141: Perhaps we should keep it simple and just remove number of settings to remove in this message, else we'll have to handle plural form (feasible of course, but is it worth?):

image

What about simplifying to:

```python
_("Are you sure you want to remove selected option settings?"),
```

After all, there's one checkbox to enable the deletion, then one for each setting, plus a confirmation message; the number of selected settings isn't really needed imho.

If not, then we need to use `ngettext()` instead of `_()`.

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,166 @@
...
+        self.enable_cleanup_changed()
+
+    def save(self):
+        if not self.ui.enable_cleanup.checkState() == QtCore.Qt.Checked:
+            return
+        to_remove = set()
+        for idx in range(self.ui.tableWidget.rowCount()):
+            item = self.ui.tableWidget.item(idx, 0)
+            if item.checkState() == QtCore.Qt.Checked:
+                to_remove.add(item.data(QtCore.Qt.UserRole))
```
zas, line 137: Perhaps we could break down this method a bit and use a generator, something like:

```python
    def column_items(self, column):
        for idx in range(self.ui.tableWidget.rowCount()):
            yield self.ui.tableWidget.item(idx, column)

    def selected_options(self):
        for item in self.column_items(0):
            if item.checkState() == QtCore.Qt.Checked:
                yield item.data(QtCore.Qt.UserRole)

    def select_all_changed(self):
        state = self.ui.select_all.checkState()
        for item in self.column_items(0):
            item.setCheckState(state)

    def save(self):
        if not self.ui.enable_cleanup.checkState() == QtCore.Qt.Checked:
            return
        to_remove = set(self.selected_options())
```

It doesn't change stuff much, but results in smaller methods, which is usually better.

zas (changes requested):

picard/ui/options/maintenance.py
rdswift: Agreed that the number can be removed.  I actually had this in for testing during development and forgot to remove it.

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,178 @@
...
+    def make_setting_value_text(self, key):
+        config = get_config()
+        value = config.setting.raw_value(key)
+        return repr(value)
+
+    # def select_all_changed(self):
+    #     state = self.ui.select_all.checkState()
+    #     for idx in range(self.ui.tableWidget.rowCount()):
+    #         item = self.ui.tableWidget.item(idx, 0)
+    #         item.setCheckState(state)
```
zas, line 170: @rdswift you can remove those lines

zas (changes requested):

picard/ui/options/maintenance.py
rdswift: Crap!  I thought I had removed the old method, but I guess I had just commented it out.  Sorry about that.

zas (dismissed):
LGTM

phw (commented):
Looks good, good work. Amazing what stuff I found in my config file :)

I'd just move it under the Advanced section, what do you think?

rdswift:
> Amazing what stuff I found in my config file :)

Me too.  It's amazing all the stuff that accumulates in your test configuration file during development.  :stuck_out_tongue_winking_eye: 

> I'd just move it under the Advanced section, what do you think?

Perfect!  I'll make the change right away.

phw (approved):

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,176 @@
...
+        current_options = set([
+            # Include options that are required but are not entered directly from the options pages.
+            'file_renaming_scripts',
+            'selected_file_naming_script_id',
+            'log_verbosity',
+            # Items missed if TagsCompatibilityWaveOptionsPage does not register.
+            'remove_wave_riff_info',
+            'wave_riff_info_encoding',
+            'write_wave_riff_info',
+        ]).union(key_options)
```
zas, line 92: What about moving this set to a module level constant? It would prolly ease maintenance (than have it buried in the code), something like:

```python
OPTIONS_NOT_IN_PAGES = {
    # Include options that are required but are not entered directly from the options pages.
    'file_renaming_scripts',
    'selected_file_naming_script_id',
    'log_verbosity',
    # Items missed if TagsCompatibilityWaveOptionsPage does not register.
    'remove_wave_riff_info',
    'wave_riff_info_encoding',
    'write_wave_riff_info',
}

....
         current_options = OPTIONS_NOT_IN_PAGES.union(key_options)

```

picard/ui/options/maintenance.py
```diff
@@ -0,0 +1,176 @@
...
+        # All setting options included in the INI file.
+        config.beginGroup("setting")
+        file_options = set(config.childKeys())
+        config.endGroup()
+
+        orphan_options = file_options.difference(current_options)
+
+        self.ui.description.setText(
+            _(self.INSTRUCTIONS)
+            + _("\n\nThe configuration file currently contains %d option settings, %d which are unused.") % (
```
zas, line 103: I just noticed, but this is incorrect from i18n point of view (any simple string concatenation is).

We should rather have an extra widget for this:

```python
self.ui.config_file_contains.setText(
    _("The configuration file currently contains %d option settings, %d which are unused.") % (
        len(file_options),
        len(orphan_options)
       )
    )
) 
```

Since `self.ui.description` will remain untouched, no need to set text again for it.

zas (changes requested):

picard/ui/options/maintenance.py
rdswift: Can do, but I wonder if it would it be better to just include that line in the `INSTRUCTIONS` string.  Or do we want to keep the two separate?

zas: > Can do, but I wonder if it would it be better to just include that line in the `INSTRUCTIONS` string

But it wouldn't work because of https://github.com/metabrainz/picard/pull/1893/files#diff-951dd86183cd3da6256f3118e7ea870fcd39ec1e1cdeb159ff238ff8f664ea14R71

rdswift: Right.  I missed that.  I'll push the change right away.

zas (dismissed):
LGTM

zas (approved):
LGTM

zas (approved):

PR #1894, opened 2021-09-03
Simplify AcousticBrainz extractor settings

zas:
- do not output error when this feature was never set
- display a tooltip on download button
- (always) set a translatable place holder for path input field
- let user edit the path directly
- ensure extractor validity is checked after browsing for it

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/acousticbrainz/__init__.py
+++ b/picard/acousticbrainz/__init__.py
@@ -118,7 +118,7 @@ def ab_extractor_callback(tagger, file, result, error):
                   (file.metadata["musicbrainz_recordingid"], file.filename,))
 
         # Submit results
-        ab_submit_features(tagger, file, ab_metadata_file)
+        ab_submit_features(tagger, file)
     elif 'Duplicate' in error:
         log.debug("AcousticBrainz already has an entry for recording %s: %s" %
                   (file.metadata["musicbrainz_recordingid"], file.filename,))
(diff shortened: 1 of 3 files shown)


phw (changes requested):
We intentionally did not want to persist the found path if it is empty automatically. Instead empty path will trigger auto detection. I think this PR breaks this, because it will persist the detected path to settings.

We have setups were the path is rather dynamic and changes. E.g. for portable Windows version in changes on each start of Picard, for Snaps it changes on each app update. And even on macOS it changes when the user moves the app around.

Hence we should keep the use case of having the path empty. That is also consistent with how fpcalc is handled.

That also means `check_extractor_version` must still check for a version even if path is empty. But I agree we should not log errors in this case. Maybe if path is empty and no binary is found just log a info or warning.

There is also the "Use acoustic feature extraction" option we could check. If that's disabled we should definitely not try to detect the binary at all.

rdswift (commented):
@phw has already mentioned all of the concerns that I had regarding use of the Windows Portable version.  The rest looks good to me.

zas:
> We intentionally did not want to persist the found path if it is empty automatically.

Oh I missed this, I'll rework the PR.

So, when in "auto-detect" mode, we shouldn't save any path (and remove any existing path from config), right?

Can you confirm "auto-detect" should be triggered when:
- no path is set (=empty field), either one first run or because user cleared the existing path
- when invalid path is set (=something in the field but not a valid executable or invalid version)

On `save`, it reads the text from field, so I guess this is the reason a place holder was used instead.

So basically, place holder should be set to the path of the found path (if any), but no value should be set.

zas:
I made changes so path is never saved if auto-detected.
Please test & review.

rdswift:
I just checked the portable version, and it seems to work well.  :+1:

rdswift (commented):
I just checked the portable version, and it seems to work well.  :+1:

zas:
@phw please re-review, I addressed your comments (I think) and went a bit further in cleaning up the code. I detected few potential issues I tried to solve here.

phw (approved):
Works well now, good improvements

PR #1895, opened 2021-09-05
AcousticBrainz fixes

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

This fixes a some issues encountered during testing:

1. Update the action status in main window if AcousticBrainz gets enabled / disabled
2. Fix submission failing if no recording ID is stored in the tags

The second issues requires some discussion. The issue I encountered was a KeyError exception in https://github.com/metabrainz/picard/blob/master/picard/acousticbrainz/__init__.py#L198 if the features file did not extract a recording ID from existing tags.

But the fix changes more. Previously the submission would not be done if the recording ID stored in tags differs from the currently matched recording ID. I think this is not very clear, because submission would fail, but the user would just assume they have the files matched and can submit them, and since feature extraction was performed the long time this takes gives the impression something was actually done. Basically we are using a mix of new and old data to do the submission.

We have two ways to improve this:

1. Always use the matched recording ID (i.e. the new metadata) for the submission. That means when I have matched a file to a recording the features will be submitted for this recording ID, regardless of existing metadata. This is what I have currently done. It is consistent with what we do for AcoustID fingerprint submission. But it might make it more likely that data gets submitted for wrong recording.

2. Only allow submission for saved files. That would mean we would not even enable submission for files with pending changes and would not start the analysis process for these files. Only the saved metadata gets used. It means we have a separate safety step that requires the user to actually save the files. Basically it is the same as if the user tagged the files with Picard and used the separate AB submission tool afterwards on the files. We might need to add some kind of info that unsaved files cannot be submitted to AB, not sure how this would be done best.

### DIFF
--- a/picard/acousticbrainz/__init__.py
+++ b/picard/acousticbrainz/__init__.py
@@ -194,23 +194,15 @@ def ab_submit_features(tagger, file):
     with open(file.acousticbrainz_features_file, "r", encoding="utf-8") as f:
         features = json.load(f)
 
-    try:
-        musicbrainz_recordingid = features['metadata']['tags']['musicbrainz_trackid'][0]
-    except KeyError:
-        musicbrainz_recordingid = file.metadata['musicbrainz_recordingid']
-        features['metadata']['tags']['musicbrainz_trackid'] = [musicbrainz_recordingid]
-    # Check if extracted recording id matches the current file (recording ID may have been merged with others)
-    if musicbrainz_recordingid != file.metadata['musicbrainz_recordingid']:
-        # If it doesn't, skip the submission
-        log.debug("AcousticBrainz features recordingId do not match the file metadata: %s" % file.filename)
-        submit_features_callback(file, None, None, True)
-        return
+    # Always submit for currently matched recording ID
+    musicbrainz_recordingid = file.metadata['musicbrainz_recordingid']
+    features['metadata']['tags']['musicbrainz_trackid'] = [musicbrainz_recordingid]
 
     # Submit features to the server
     tagger.webservice.post(
         host=ACOUSTICBRAINZ_HOST,
         port=ACOUSTICBRAINZ_PORT,
-        path="/%s/low-level" % file.metadata["musicbrainz_recordingid"],
+        path="/%s/low-level" % musicbrainz_recordingid,
         data=json.dumps(features),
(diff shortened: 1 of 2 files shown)


zas (approved):

zas:
> We have two ways to improve this:
> 
>     1. Always use the matched recording ID (i.e. the new metadata) for the submission. That means when I have matched a file to a recording the features will be submitted for this recording ID, regardless of existing metadata. This is what I have currently done. It is consistent with what we do for AcoustID fingerprint submission. But it might make it more likely that data gets submitted for wrong recording.
> 
>     2. Only allow submission for saved files. That would mean we would not even enable submission for files with pending changes and would not start the analysis process for these files. Only the saved metadata gets used. It means we have a separate safety step that requires the user to actually save the files. Basically it is the same as if the user tagged the files with Picard and used the separate AB submission tool afterwards on the files. We might need to add some kind of info that unsaved files cannot be submitted to AB, not sure how this would be done best.

What about allowing both, defaulting to the safest?

rdswift:
> I think this is not very clear, because submission would fail, but the user would just assume they have the files matched and can submit them, and since feature extraction was performed the long time this takes gives the impression something was actually done.

Perhaps set a "processing_success" flag to True when starting the processing for the batch, updating it to False if an error is encountered for one of the files (and log the error).  Then when the processing items counter reaches zero (at the end of the batch) open a dialog if the "processing_success" flag is False with a message like, "There were errors encountered while submitting to AcousticBrainz.  Please see the log for details."  If the flag is True, no dialog is necessary, although you could open one saying that the submissions were successful if you like (because it does take a significant amount of time to do the processing).  You could even have an option setting for whether or not to display the "Success" message.

rdswift:
> What about allowing both, defaulting to the safest?

And allow the user to decide through an option setting?

zas:
> And allow the user to decide through an option setting?

Yes.
Defaulting to safest mode (option 2)

phw:
Is this really worth an option? I think this is something that should work properly without configuration. A user option that is hard to explain sounds like a not very good work around to me.

Let's merge this and I'll try in another PR how this could work if we submit only saved files.

PR #1898, opened 2021-09-08
PICARD-412: Artist translation option enhancements

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**: Add weighting options to the script character set detector, and allow the user to set multiple translation ignore character set scripts based on a user-specified threshold of weighted content for the script in the artist name.

# Problem

* JIRA ticket (_optional_): PICARD-412

The translate artist name to this locale option only allowed a single locale, and did not allow for a fallback.

The original implementation of ignoring artist name translation allowed only a single character set script to be selected.  There was no allowance for weighting of how much of the artist name string matched the selected character set script.  If a single character matched the selected script, then translation would be ignored.

# Solution

- Add ability to select multiple translation locales and specify the order in which they are tried.  Processing stops when a translation is found for a locale.
- Add weighting options to the script character set detector.
- Allow the user to set multiple translation ignore character set scripts based on a user-specified threshold of weighted content for the script in the artist name.
- Update tests.

# Action


rdswift:
With the latest commits, I think this might address [PICARD-412](https://tickets.metabrainz.org/browse/PICARD-412).  I'm not entirely certain if the PR addresses [PICARD-411](https://tickets.metabrainz.org/browse/PICARD-411) though.

picard/ui/options/metadata.py
```diff
@@ -156,9 +173,78 @@ def set_nat_name_default(self):
...
+        self.move_view = MoveableListView(self.ui.selected_locales, self.ui.move_up, self.ui.move_down)
+        self.ui.add_locale.clicked.connect(self.add_locale)
+        self.ui.remove_locale.clicked.connect(self.remove_locale)
+        self.ui.selected_locales.currentRowChanged.connect(self.set_button_state)
+        self.load()
+
+    def load(self):
+        self.ui.selected_locales.clear()
+        for locale in self.parent().current_locales:
+            label = ("    " if '_' in locale else "") + ALIAS_LOCALES[locale]
```
zas, line 202: Locale should be translated.

I also see another issue with those indentations: this doesn't work for Right to Left languages (spacing should be on the other side).

So we should prolly have a translatable format string + translated locale name:

```python
        def indented_translated_locale(locale, level):
            return _("{indent}{locale}").format(
                indent="    " * level,
                locale=_(ALIAS_LOCALES[locale])
            )

        for locale in self.parent().current_locales:
            level = 1 if '_' in locale else 0
            label = indented_translated_locale(locale, level)
            item = QtWidgets.QListWidgetItem(label)
            item.setData(QtCore.Qt.UserRole, locale)
            self.ui.selected_locales.addItem(item)
        self.ui.selected_locales.setCurrentRow(0)

        self.ui.available_locales.clear()
        for (locale, name, level) in iter_sorted_locales(ALIAS_LOCALES):
            label = indented_translated_locale(locale, level)
            item = QtWidgets.QListWidgetItem(label)
            item.setData(QtCore.Qt.UserRole, locale)
            self.ui.available_locales.addItem(item)
        self.ui.available_locales.setCurrentRow(0)
```

Note, in the loop above, `name` has no use, perhaps we should also modify `iter_sorted_locales()`

picard/ui/options/metadata.py
```diff
@@ -126,12 +133,22 @@ def load(self):
 
         self.set_enabled_states()
 
+    def make_locales_text(self):
+        locales = list(ALIAS_LOCALES[key] for key in self.current_locales)
```
zas, line 137: Locale names aren't translated here.

Something like:

```python
def translated_locales():
    for locale in self.current_locales:
        yield _(ALIAS_LOCALES[locale])

self.ui.selected_locales.setText('; '.join(translated_locales())
```

picard/util/script_detector_weighted.py
```diff
@@ -0,0 +1,94 @@
...
+    total_weighting = 0
+    for character in string_to_check:
+        if character.isalpha():
+            script_id = ud.name(character).split(' ')[0]
+            weighting_factor = SCRIPT_WEIGHTING_FACTORS[script_id] if script_id in SCRIPT_WEIGHTING_FACTORS else 1
+            scripts[script_id] = (scripts[script_id] if script_id in scripts else 0) + weighting_factor
+            total_weighting += weighting_factor
+    # Normalize weightings to a float between 0 and 1 inclusive.
+    for key in scripts:
+        scripts[key] /= total_weighting
```
zas, line 76: There's a risk of division by zero if `string_to_check` is empty or if there's no alpha character in the string.
We should prolly ensure it doesn't happen.

zas (changes requested):
Nice improvement.
Few comments about i18n to address.

picard/mbjson.py
```diff
@@ -192,46 +193,52 @@ def _translate_artist_node(node):
...
+                        score = 0.0
+                    parts.append((score, 5))
+                    comb = linear_combination_of_weights(parts)
+                    if comb > result[0]:
+                        result = (comb, (alias['name'], alias["sort-name"]))
+                transl, translsort = result[1]
+                if transl:
+                    # Matching locale found.  Skip further processing.
+                    return (transl, translsort)
+            if not transl:
```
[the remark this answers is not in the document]
rdswift: > I think this part of the code needs a rewrite.

I agree.  In fact, I thought about this overnight (couldn't sleep) and I want to change the logic a bit so that it does a first pass through the selected locales list looking for an exact match on the full locale, and if there is no match then do another pass looking for a match on the root locales from the items in the selected locales list.  That should provide the user more control if they specify multiple sub-locales for the same root locale.  To make this a bit more efficient, I plan to do a single pass through the aliases and create a dictionary of available aliases based on the best score for each locale found.  The selected alias checks would then be done against this dictionary rather than parsing the aliases list multiple times and recalculating the scores for the items each time.

picard/ui/options/metadata.py
rdswift: Yeah, I completely forgot about the translations for the UI lists and such.  I'll have a look at that.

picard/util/script_detector_weighted.py
rdswift: Are you sure?  I thought about this when I wrote it, and if there are no alpha characters in the string or the string is empty, then the `scripts` dictionary should be empty thus that normalization routine won't be processed.  I'll look again (and perhaps add a test) just to be sure.  It will be easy enough to include a safety net, just to be absolutely sure.

zas: I guess I misread the code (I shouln't review before my third coffee)... You're right, it shouldn't happen.

That said, adding tests is the the way to go ;)

zas (approved):
LGTM

phw (approved):
This is awesome! Really great addition.

What we maybe could do is move the configuration for script exceptions also to a dialog, having it behave similar to the script selection (which I like very much). My reasoning would be that script selection as it is presented is kind of a sophisticated feature and it usually would not get changed that frequently. But currently it kind of dominates the Options > Metadata view, making other more important (for many users) settings less prominent.

I think it would work well with this advanced UI in a dialog, and then  just showing the selected scripts with weighting in the main UI, e.g. "Latin (0), Hebrew (50)".

But let's merge this as is first

zas:
> What we maybe could do is move the configuration for script exceptions also to a dialog, having it behave similar to the script selection (which I like very much). My reasoning would be that script selection as it is presented is kind of a sophisticated feature and it usually would not get changed that frequently. But currently it kind of dominates the Options > Metadata view, making other more important (for many users) settings less prominent.
> 
> I think it would work well with this advanced UI in a dialog, and then just showing the selected scripts with weighting in the main UI, e.g. "Latin (0), Hebrew (50)".

Yes, I thought about that too, it would unify interface and totally makes sense.

rdswift:
Just to make sure I understand, you would like to maintain a separate weighting for each selected script?

zas:
> Just to make sure I understand, you would like to maintain a separate weighting for each selected script?

It was about "move the configuration for script exceptions also to a dialog,"

phw:
I think it was the "Latin (0), Hebrew (50)" that caused the confusion. I don't know why, but when I wrote this down I somehow thought it was pet script. 

But I don't think that's necessary, my comment really was only concerned with moving the UI to a dialog. Sorry for the confusion

rdswift:
No problem.  I'm actually setting up the threshhold weightings on a per script basis.  Not much of a change and provides greater flexibility.

PR #1899, opened 2021-09-10
Translation exception script settings update

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [x] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Update the Metadata options page and add script-specific threshold settings for the translation exception scripts.

# Problem

Translation exception script selection was using up too much space on the Metadata options page, and was inconsistent with the locale selection dialog introduces in PR #1898.  All translation exception scripts used the same threshold value.

# Solution

Update the Metadata options page and add script-specific threshold settings.

- Move exception script selection to new dialog, consistent with the locale selection dialog.
- Add script-specific weighting thresholds.

Exception script processing now checks each selected character set script in order, and if the percentage weighting of the artist name string for that script is greater than or equal to the threshold value set for the script, the translation process is aborted.

# Action

Update the documentation.  See: https://github.com/rdswift/picard-docs/pull/125


picard/ui/options/metadata.py
```diff
@@ -140,18 +133,22 @@ def translated_locales():
 
         self.ui.selected_locales.setText('; '.join(translated_locales()))
 
+    def make_scripts_text(self):
+        def translated_scripts():
+            for script in self.current_scripts:
+                yield "{script} ({weighting}%)".format(
+                    script=_(SCRIPTS[script[0]]),
+                    weighting=script[1],
+                )
```
zas, line 142: We could use `ScriptExceptionSelector.make_label()`

picard/ui/ui_options_metadata.py
```diff
@@ -173,8 +126,7 @@ def retranslateUi(self, MetadataOptionsPage):
         self.translate_artist_names.setText(_("Translate artist names to these locales where possible:"))
         self.select_locales.setText(_("Select"))
         self.translate_artist_names_script_exception.setText(_("Ignore artist name translation for these scripts:"))
-        self.label_2.setText(_("<html><head/><body><p>Only scripts where the percentage of the artist name that is written using that script are above the specified minimum weighting will be considered.</p></body></html>"))
-        self.label.setText(_("Minimum Weighting:"))
+        self.select_scripts.setText(_("Select"))
```
zas, line 129: Since those "Select" buttons open a dialog, they should be labelled "Select…"

zas (changes requested):
Nice improvement.

On first open, the dialog was a bit small here.
Also I think we should add a tooltip to document what the threshold.

picard/ui/options/metadata.py
```diff
@@ -261,4 +263,106 @@ def save_changes(self):
...
+            label = self.make_label(script_id=script[0], script_weighting=script[1])
+            item = QtWidgets.QListWidgetItem(label)
+            item.setData(QtCore.Qt.UserRole, script)
+            self.ui.selected_scripts.addItem(item)
+        if self.ui.selected_scripts.count() > 0:
+            self.ui.selected_scripts.setCurrentRow(0)
+        self.set_weighting_selector()
+
+        self.ui.available_scripts.clear()
+        for script_id in SCRIPTS:
```
phw, line 299: I think it would be great to have the scripts sorted by localized label, something like produced by this:

```
for script_id, label in sorted([(k, _(v)) for k, v in SCRIPTS.items()], key=lambda i: i[1]):
    ...
```

phw (changes requested):
Yes, that looks very good.

> Also I think we should add a tooltip to document what the threshold.

Agreed. Because this is not intuitively understood I liked the detailed description in the current version on master. I think a tooltip for the threshold would work well.

picard/config_upgrade.py
```diff
@@ -435,6 +435,20 @@ def upgrade_to_v2_7_0_dev_4(config):
...
+def upgrade_to_v2_7_0_dev_5(config):
+    """Replace artist_script_exceptions with script_exceptions and remove artist_script_exception_weighting"""
+    _s = config.setting
+    ListOption("setting", "script_exceptions", [])
+    weighting = _s["artist_script_exception_weighting"] if _s["artist_script_exception_weighting"] else 0
+    script_exceptions = []
+    if _s["artist_script_exceptions"]:
+        for script_exception in _s["artist_script_exceptions"]:
+            script_exceptions.append((script_exception, weighting))
+    _s["script_exceptions"] = script_exceptions
```
zas, line 447: Those 5 lines could be replaced by:

```python
artist_script_exceptions = _s["artist_script_exceptions"] or []
_s["script_exceptions"] = [(script_exception, weighting) for script_exception in artist_script_exceptions]
```

picard/mbjson.py
```diff
@@ -194,10 +194,10 @@ def _translate_artist_node(node):
     transl, translsort = None, None
     if config.setting['translate_artist_names']:
         if config.setting['translate_artist_names_script_exception']:
-            threshhold = config.setting["artist_script_exception_weighting"] / 100
-            detected_scripts = list_script_weighted(node["name"], threshhold)
-            for script_id in config.setting["artist_script_exceptions"]:
-                if script_id in detected_scripts:
+            detected_scripts = detect_script_weighted(node["name"])
+            for script_item in config.setting["script_exceptions"]:
+                script_id, script_weighting = script_item
```
zas, line 199: `script_item` isn't stricly needed:

```python
for script_id, script_weighting in config.setting["script_exceptions"]:
    if script_id in detected_scripts and detected_scripts[script_id] >= script_weighting / 100:
        return node['name'], node['sort-name']
```

picard/ui/options/metadata.py
```diff
@@ -179,12 +176,17 @@ def set_enabled_states(self):
         self.ui.select_locales.setEnabled(translate_checked)
         self.ui.selected_locales.setEnabled(translate_checked)
         self.ui.translate_artist_names_script_exception.setEnabled(translate_checked)
-        self.ui.ignore_script_frame.setEnabled(translate_checked and translate_exception_checked)
+        self.ui.selected_scripts.setEnabled(translate_checked and translate_exception_checked)
+        self.ui.select_scripts.setEnabled(translate_checked and translate_exception_checked)
```
zas, line 180: We could reduce code redundancy a bit using a temporary variable:

```python
select_scripts_enabled =  translate_checked and translate_exception_checked       
self.ui.selected_scripts.setEnabled(select_scripts_enabled)
self.ui.select_scripts.setEnabled(select_scripts_enabled)
```

zas (changes requested):

rdswift:
> > Also I think we should add a tooltip to document what the threshold.
> 
> Agreed. Because this is not intuitively understood I liked the detailed description in the current version on master. I think a tooltip for the threshold would work well.

For now, I've added a description right on the dialog as well as tooltips.  It probably doesn't need both (although it doesn't hurt), so we can keep the one you prefer and rove the other if you like.

picard/ui/options/metadata.py
```diff
@@ -261,4 +262,118 @@ def save_changes(self):
...
+        self.ui.setupUi(self)
+        self.ui.button_box.accepted.connect(self.save_changes)
+        self.ui.button_box.rejected.connect(self.reject)
+        self.move_view = MoveableListView(self.ui.selected_scripts, self.ui.move_up, self.ui.move_down)
+        self.ui.add_script.clicked.connect(self.add_script)
+        self.ui.remove_script.clicked.connect(self.remove_script)
+        self.ui.selected_scripts.currentRowChanged.connect(self.selected_script_changed)
+        self.ui.weighting_selector.valueChanged.connect(self.weighting_changed)
+
+        weighting_tooltip_text = (
```
zas, line 278: This text should be translated.

```suggestion
        weighting_tooltip_text = _(
```

zas (changes requested):

zas:
> For now, I've added a description right on the dialog as well as tooltips

I tested, I would only keep the tooltip: the main reason is that once the user understood what it does the description isn't useful anymore, so I think the tooltip is enough.

Apart that, LGTM

rdswift:
> I tested, I would only keep the tooltip: the main reason is that once the user understood what it does the description isn't useful anymore, so I think the tooltip is enough.

Excellent point, and one that I hadn't considered.

zas (approved):
LGTM

phw (approved):

PR #1900, opened 2021-09-12
PICARD-2233: Composer "credited as" must not overwrite composersort

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2233

If "credited as" is set on a composer relationship, the credited name gets used both for composer and composersort. This is inconsistent with how target credits get used for artist / artistsort and causes a composersort tag without proper sort name.

# Solution

Do not use "credited as" for composersort. Added tests specific to reading composer / composersort.


rdswift (commented):
This makes sense to me, and seems to work well.

zas (approved):
LGTM

PR #1901, opened 2021-09-14
PICARD-2278: Show path to config file in About dialog.

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other
* **Describe this change in 1-2 sentences**: Add a line to the "About" dialog showing the path and filename of the current configuration file.

# Problem

* JIRA ticket (_optional_): PICARD-2278

There is no way to see which configuration file is currently being used, or any (easy) way for the user to find out the location of the configuration file.

# Solution

Add a line to the "About" dialog showing the path and filename of the current configuration file.

# Action

Possibly update the Troubleshooting section of the documentation.

### DIFF
--- a/picard/ui/aboutdialog.py
+++ b/picard/ui/aboutdialog.py
@@ -28,6 +28,7 @@
 
 from PyQt5 import QtCore
 
+from picard.config import get_config
 from picard.const import PICARD_URLS
 from picard.formats import supported_extensions
 from picard.util import versions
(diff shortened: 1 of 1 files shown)


phw (commented):
Looks good to me. Just not sure if About is the best place for it. Another place would be the new "Maintenance" section, as this is already concerned with config file things. What do you think?

The filename could also be linked to open the folder in the file browser

zas:
> Just not sure if About is the best place for it. Another place would be the new "Maintenance" section, as this is already concerned with config file things.

Yes, also unsure about the right place for this info.

The question is: if a user is wondering about config file path, where does he expect to find this info?
It is likely more About dialog than Maintenance. But both seem inadequate to me.

What about letting it as is (in About) and eventually move it if we find a better place?

phw:
> What about letting it as is (in About) and eventually move it if we find a better place?

Yes, I'm fine with that.

phw (approved):

zas (approved):

PR #1902, opened 2021-09-15
Only allow AB submission for saved files

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

This is an extension to the discussion in https://github.com/metabrainz/picard/pull/1895 . There I fixed the issue of submission crashing when a file did not have a recording ID saved in the tags. In the fix I changed the behavior to always submit with currently matched files, that means overwriting the metadata the streaming extractor had read from the file with our currently matched recording ID.

But the original intention of the code was to ensure a file has been properly matched to only allow clean submissions. Similar to how the current AB submission tools are supposed to be run on files already properly tagged with MB tags.

# Solution
- Only allow submission of saved files
- Restore the original check that compared loaded metadata with existing metadata (removed previously in https://github.com/metabrainz/picard/pull/1895/commits/2534b9c59c2db018c66aa73675d3306f42c0375c)
- Handle the case that the loaded track ID is empty

# Action
Not fully sure about this, let's discuss. @Gabrielcarvfer , what's you take on this?

### DIFF
--- a/picard/acousticbrainz/__init__.py
+++ b/picard/acousticbrainz/__init__.py
@@ -4,6 +4,7 @@
 #
 # Copyright 2014 Music Technology Group - Universitat Pompeu Fabra
 # Copyright 2020-2021 Gabriel Ferreira
+# Copyright 2021 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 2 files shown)


zas (approved):
To me, it makes sense as we clearly want to avoid submitting incorrect data.

Gabrielcarvfer:
I agree with both of you. I thought I checked for that, but apparently I didn't. The specific check before the submission only checks for the recording id, and only included it because I've seen mutagen/AB get different numbers, which ended up being due to a merged recording.

PR #1903, opened 2021-09-15
Add config file name and link to Maintenance options page

rdswift:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [x] Other
* **Describe this change in 1-2 sentences**:  Rearrange the Maintenance options page and add config filename info and link to directory.

# Problem

Further to the discussion in https://github.com/metabrainz/picard/pull/1901 this enables some opportunity to explore options for location of the current config file information.

# Solution

Rearrange the Maintenance options page and add config filename info and link to directory.

# Action

If accepted, the documentation should be updated to reflect the changes.


zas:
I'm not fond of read-only QLineEdit because it looks like one can edit it, but I guess that's the simplest way to display paths (a QLabel would not horizontally scroll by default on long paths).

zas (approved):
Simple and useful addition.
Now I wonder if we still need it in About. It looks like to me the path to config file isn't important for most users.

phw (approved):

phw:
> Now I wonder if we still need it in About. It looks like to me the path to config file isn't important for most users.

Yes, that's what I thought as well. Great to have a place for that, but IMHO it can be a bit more hidden than about. I personally would just keep it here in Maintenance and remove it from about. But really no strong opinion either way.

PR #1904, opened 2021-09-15
PICARD-2279: Allow saving AcoustID fingerprint in acoustid_fingerprint tag

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2279

Picard supports reading existing fingerprints from the `acoustid_fingerprint` tag, but it has no option to save calculated fingerprints there.

While usually this was considered to be not needed, as the fingerprint can always be calculated again from the audio and can add rather long data to the file, there are some use cases and some users using this. E.g. one user on the forums described a workflow where they add this tag directly after ripping to avoid having to redo the calculation later on.

See e.g. also discussions on the forums like https://community.metabrainz.org/t/picard-2-3-option-to-generate-fingerprints-automatically-to-submit-it/464069/5

# Solution

Add an option for saving the fingerprint to `acoustid_fingerprint`, but keep it disabled by default.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -632,6 +632,9 @@ class File(QtCore.QObject, Item):
             self.acoustid_length = length or self.metadata.length // 1000
             self.tagger.acoustidmanager.add(self, None)
             self.acoustid_update()
+        config = get_config()
+        if config.setting['save_acoustid_fingerprints']:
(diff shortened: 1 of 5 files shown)


zas (approved):
LGTM

PR #1907, opened 2021-09-26
PICARD-2189: Add appdirs module to fix getting proper cache location

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2189

The cache location previously was wrong because it was created before app and org name had been set. The new module fixes this and allows more flexible replacement of the implementation.

# Solution

- Fix cache location by ensuring `QCoreApplication.setApplicationName` and `QCoreApplication.setOrganizationName` get called before QStandardLocations are being used
- Introduce an new `picard.const.appdirs` module, to centralize the place where we deal with these directories. This makes it easier to test and to exchange the implementation in the future.
- Added platform specific tests, this will allow us to change implementation 

# Action

At it's core this PR only deals with fixing the cache location. But there are some inconsistencies in the way we use the paths, depending on platform. I think this PR makes it easier to address these in the future.

- On macOS the config file is stored at `~/.config/MusicBrainz`. Instead it should be placed in ` ~/Library/Preferences/MusicBrainz/Picard`. This needs a migration. I think this best is addressed when implementing PICARD-2227
- Likewise on Linux: We have both `~/.config/MusicBrainz/Picard.ini` and a `~/.config/MusicBrainz/Picard` folder. I think we should be consistent and place the configuration inside that Picard subfolder when doing PICARD-2227
- On Linux plugins are placed in the config folder at `~/.config/MusicBrainz/Picard/plugins`. Better would be to place the plugins folder in `QStandardPaths.AppDataLocation`, which would resolve to `.local/share/MusicBrainz/Picard/plugins`. We could address this by declaring the new folder the plugin folder, but still use the old path as legacy (as our plugin system supports using multiple paths). Should be investigated for other OS as well. PICARD-522

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -34,28 +34,20 @@
 
 import builtins
 from collections import OrderedDict
-import os
 
-from PyQt5.QtCore import QStandardPaths
-
-from picard import (
-    PICARD_APP_NAME,
-    PICARD_ORG_NAME,
-    PICARD_VERSION,
-)
+from picard import PICARD_VERSION
 from picard.const.attributes import MB_ATTRIBUTES
+from picard.const import appdirs
 
 # Install gettext "noop" function in case const.py gets imported directly.
 builtins.__dict__['N_'] = lambda a: a
 
 
 # Config directory
-_appconfiglocation = QStandardPaths.writableLocation(QStandardPaths.AppConfigLocation)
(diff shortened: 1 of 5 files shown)


zas (approved):
Much better, LGTM

CyberSkull:
Great work!

PR #1910, opened 2021-09-28
PICARD-2261: Do not strip whitespace from file tags

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2261

Fixes e.g. breaking tags with ending newlines, like e.g. some iTunes specific tags like `itunnorm`.

# Solution

Do not run `Metadata.strip_whitespace` after applying scripts to tags.

# Action

I am actually not sure if there is a need to run this at all. I kept the stripping when applying the scripts to the track metadata, because this is also done in https://github.com/metabrainz/picard/blob/master/picard/album.py#L383-L398 . But I'm not sure if this really is something that should be done.

It also means my change might cause slightly different results when scripts get applied to files, and this might cause differences in scripts having unintentional additional whitespace. It might even mean that in case of multiple scripts a follow up script does not behave as expected, as it might expect the tag without that additional whitespace.

The IMHO only clean solution to this would be to either do no whitespace stripping at all, or only for those tags actually modified by the script.

Originally there was whitespace trimming added in https://github.com/metabrainz/picard/commit/c165b10b4b913be64de0fb0cb295b9bf68f21974
and removed again for files only in https://github.com/metabrainz/picard/commit/c31c37fff3da2397e068e2b8b8b572dfe1e4a8ab , because it was causing issues like this.

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -198,14 +198,15 @@ class Track(DataObject, FileListItem):
             self.tagger.window.refresh_metadatabox()
 
     @staticmethod
-    def run_scripts(metadata):
+    def run_scripts(metadata, strip_whitespace=False):
         for s_name, s_text in enabled_tagger_scripts_texts():
             parser = ScriptParser()
             try:
                 parser.eval(s_text, metadata)
             except ScriptError:
                 log.exception("Failed to run tagger script %s on track", s_name)
-            metadata.strip_whitespace()
+            if strip_whitespace:
+                metadata.strip_whitespace()
 
     def update(self):
         if self.item:
@@ -416,7 +417,7 @@ class NonAlbumTrack(Track):
         self._customize_metadata()
         run_track_metadata_processors(self.album, m, recording)
         self.orig_metadata.copy(m)
-        self.run_scripts(m)
+        self.run_scripts(m, strip_whitespace=True)
         self.loaded = True
         self.status = None
         if self.callback:


phw:
@zas Really interested in your opinion on the whitespace trimming in general. I just noticed that the current behavior also causes some oddities. E.g. consider the following script:

```
$set(foo,bar    )
$if($eq(%foo%,bar),$set(foo,baz))
```

`%foo%` will be set to `bar    `, that means the $if fails. After the script is run whitespace is trimmed and the result is the tag will have `bar`.

But if I split this into two scripts:

```
$set(foo,bar    )
```

```
$if($eq(%foo%,bar),$set(foo,baz))
```

Now the first script is run, result is trimmed, `%foo%` is `bar`. The second script now succeeds, the end result is that the tag gets set to `baz`. I think this is counterintuitive, I always though about it being equivalent of having a single large script or this script being split up into multiple smaller ones in the same order.

But I also know what why this was added: Due to the difficult whitespace handling in tagger script this should fix things like:

```
$set(foo, bar)
$set(foo,$if2(%artist%,%albumartist%
))
```

or similar. Not really sure about this. But breaking existing tags that rely on whitespace is not good.

Some alternative approaches to this patch and in general:

- Have the scripting functions like $set already strip the parameters
- Keep the whitespace trimming, but have a list of exceptions (but that can never be complete)

zas:
> $set(foo, bar)
> $set(foo,$if2(%artist%,%albumartist%
> ))

IMHO, it is a language definition issue, whitespaces aren't supposed to be separators.

So I agree with this fix, though it might break some scripts.

zas (approved):
LGTM

phw:
> Couldn't the scripts just use the existing `$strip()` function if they want to ensure that the whitespace is removed?

Yes, definitely. The concerns are more about compatibility and eventually breaking existing scripts in subtle ways. I think we should probably mention this in the release announcement.

PR #1911, opened 2021-10-03
macOS: Build with newer essentia extractor binary

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1734

The official binaries for the Essentia extractor offered by AcousticBrainz cannot be used because they are build with an SDK older than SDK 10.9 and hence cannot be notarized. See also https://github.com/metabrainz/picard/commit/d9d748e423c37b18095d8340573c1a2b49253677

# Solution

Use the for now unofficial builds from https://github.com/phw/essentia-extractor-builds/releases/tag/v2.1_beta2_1 which are compiled with newer SDK. 

# Action

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -30,7 +30,7 @@ jobs:
     env:
       DISCID_VERSION: 0.6.2
       FPCALC_VERSION: 1.5.0
-      # ABEXTRACTOR_VERSION: v2.1_beta2-2-gbb40004
+      ABEXTRACTOR_VERSION: v2.1_beta2-4
       PYTHON_VERSION: 3.9.7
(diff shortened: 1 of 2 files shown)


phw:
I think we should use those binaries for now. So I'd merge this and do our beta release with this. But we still need to check whether they deliver the results as expected by AB.

@alastair Your input would be welcomed here. I'm not sure how much the values calculated can differ from other clients. Also I cannot run the older macOS binary and hence cannot compare. Should I compare with the current official Linux binary? Can you assist in verifying the binaries?

zas (approved):
LGTM, but I can't test.

phw:
Merged, we can do wider testing with the upcoming 2.7 beta.

PR #1912, opened 2021-10-04
PICARD-2199: Fix saving ratings to ID3 with non-latin1 characters in email

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2199

If you configure Options > Metadata > Ratings > E-mail to include characters that cannot be encoded as latin-1 saving a file triggers an exception.

# Solution

Use `id3text` helper function to ensure data stored in the file is encodable in latin1.

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -473,10 +473,10 @@ class ID3File(File):
             elif name == 'musicbrainz_recordingid':
                 tags.add(id3.UFID(owner='http://musicbrainz.org', data=bytes(values[0], 'ascii')))
             elif name == '~rating':
-                rating_email = id3text(config.setting['rating_user_email'], 0)
+                rating_user_email = id3text(config.setting['rating_user_email'], 0)
                 # Search for an existing POPM frame to get the current playcount
                 for frame in tags.values():
-                    if frame.FrameID == 'POPM' and frame.email == rating_email:
+                    if frame.FrameID == 'POPM' and frame.email == rating_user_email:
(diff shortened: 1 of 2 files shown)


zas (changes requested):
I think we should unify variable names while at it. We have `rating_user_email`, `user_email` and `rating_email`. What about using `rating_user_email` everywhere?

phw:
Good point, did that in https://github.com/metabrainz/picard/pull/1912/commits/d0cf836d9f74af151dcda8874ddcbf1e45dfbc57

zas:
Also [`id3text()`](https://github.com/metabrainz/picard/blob/ee06ed20f3b6ec17d16292045724921773dde597/picard/formats/id3.py#L85) is always used with `0` as second parameter.

Is there a reason to keep this parameter? (or at least, we could make it defaulting to 0).

zas (approved):
LGTM

phw:
> Is there a reason to keep this parameter? (or at least, we could make it defaulting to 0).
> EDIT: not always 

Yes, the main use of this function is actually at https://github.com/metabrainz/picard/blob/master/picard/formats/id3.py#L439 . All other cases are actually exceptions where the ID3 spec explicitly requires latin1 encoding.

Hence I wouldn't add a default really to make this clear. But we probably should add constants or an enum for the possible values. I'll do a separate PR with that.

PR #1918, opened 2021-10-13
PICARD-2655: Improve matching for VA releases

zas:
- if an album was incorrectly tagged with a label name or whatever, instead of Various Artists
Picard tries to match Various Artists (from database) to the name, usually leading to a very low similarity, reducing a lot the chance to find the correct release

- a contrario, if an album was tagged with Various Artists, it is very likely it is a VA compilation, so increase the weight

# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-2655](https://tickets.metabrainz.org/browse/PICARD-2655)

# Solution

# Action

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -46,6 +46,7 @@ from functools import partial
 from PyQt5.QtCore import QObject
 
 from picard.config import get_config
+from picard.const import VARIOUS_ARTISTS_ID
 from picard.mbjson import (
     artist_credit_from_node,
     get_score,
@@ -253,8 +254,22 @@ class Metadata(MutableMapping):
 
         if "albumartist" in self and "albumartist" in weights:
             a = self["albumartist"]
-            b = artist_credit_from_node(release['artist-credit'])[0]
-            parts.append((similarity2(a, b), weights["albumartist"]))
+            release_artists = release['artist-credit']
+            b = artist_credit_from_node(release_artists)[0]
+            artist_weight = weights["albumartist"]
+            if a.lower() in {'various', 'various artists'}:
+                # if artist in tag is 'various' or 'various artists',
+                # it is very likely we look for a VA compilation
+                # so increase the artist's weight
+                artist_weight *= 2
+            if release_artists[0]['artist']['id'] == VARIOUS_ARTISTS_ID:
+                # it is fairly common VA release are tagged with label's name
+                # so if a release is credited to VA, assume it is similar
+                # to artist in tag
+                sim = 1.0
+            else:
+                sim = similarity2(a, b)
+            parts.append((sim, artist_weight))
 
         if "totaltracks" in weights:
             try:


picard/metadata.py
```diff
@@ -253,8 +254,22 @@ def compare_to_release_parts(self, release, weights):
...
+            if a.lower() in {'various', 'various artists'}:
+                # if artist in tag is 'various' or 'various artists',
+                # it is very likely we look for a VA compilation
+                # so increase the artist's weight
+                artist_weight *= 2
+            if release_artists[0]['artist']['id'] == VARIOUS_ARTISTS_ID:
+                # it is fairly common VA release are tagged with label's name
+                # so if a release is credited to VA, assume it is similar
+                # to artist in tag
+                sim = 1.0
```
Sophist-UK, line 269: I am entirely unsure that this logic is at all correct for various reasons (though it may equally be that I don't understand how this is supposed to work) but:
1. I think that we should be considering a situation where we give the artist similarity a value of 1 ONLY if the album artist is "various" or "various artists" **AND** the release we are matching has a VA_ID. As the code stands if the album artist is "The Beatles" and the release we are matching is VA, artist similarity is 1. Or are we going to assume that any release which is VA is effectively going to match any album artist regardless of whether it is "various"?

2. What happens if the album is not in english e.g. the album is a portuguese language album and the album artist is "vários artistas"? I think we need a much longer list of VA textual names which covers all languages that we support.

3. I am unclear why we are doubling the weighting for artist if the album is "various" rather than halving it. In the specific use case described in the PR description, if the album artist is "various" and it is failing to match a release because the release artist is the label rather than "various" then surely we should give **less** weight to the artists not matching rather than more weight? 

Do we need some unit tests with various albums and various potential releases to test that we get the best release selected?

zas: I'd love have tests for that... thanks for volunteering ;)

Sophist-UK: I am happy to volunteer to write tests for this providing that you are prepared to wait until I have the time to spend on open source and for this to get to the top of my queue of open source stuff I have planned. 😉

phw: I agree with Sophist-UK that we likely should combine the tests of tags indicating VA and the release being VA.

The goal of the patch is to improve the matching of files from a VA release to VA. And one of the problems here is that pure matching of album artist tag with the MB artist name ("Various Artists") often does not yield good results.

The code will indeed improve the situation. But at the same time it will make matching a VA release from MB to **any** file more likely, as all VA releases will always have full albumartist similarity.That might increase the problem that files get tagged too much against compilation, something many users already struggle with.

Combining the check of existing album artist tags indicating VA **and** the release being VA would avoid this. For the localization I would suggest to check for `a.lower() in {'various', 'various artists', config.setting['va_name'].lower()}`.

This will not solve the case where the album artist is set to the label name or something completely different, though.

Sophist-UK: @phw 

1. I am really not sure that we should use `config.setting['va_name']` as this is stated in the documentation to be the **output** name when you use a VA_ID release - not the least for file naming so that the user can group together albums) - so this is (usually) the users local idiom for VA. But (clearly) a user can be trying to match a VA album from a different country, so Picard needs to recognise these. This is why I suggested a list (constant -> one off effort) of V/VA for every locale supported. And this same list could then be used by MB for helping improve the data, so the same effort could then also be reused. (P.S. It won't be perfect as a first iteration, but Google / Deepl translate can give you a first cut for translations of VA without too much effort. If it will help I can find the time do this and post the results here for reuse.)

2. There are probably three use cases to consider, two where the album artist is set to the label name:
    1. The album artist MBID **is** the special VA_ID but the album artist text is set to something other than (one of the translations of) "VA". Here Picard can know for certain that the release is a VA.
    2. The album artist MBID is **not** the special VA_ID even though the album is in fact a VA. I am not sure that we can do  anything in Picard about this because in most cases it is indistinguishable from a non-VA album.
    3a. Special case of 2., where the MBID is **not** VA_ID **but** the text album artist is "Various Artists" (or a translation). Here Picard can also reasonably assume that this is a VA.
    3b. Special case of 2., where the MBID is **not** VA_ID **but** the album artist text is the same as (or very similar to) the label.

    Of course the principle cause of all three use cases is bad data - so the primary fix should be to have an MB report to identify use cases 1 and 3 and have editors fix the data. (Or indeed, have some code for case 3a to auto-create (voteable) edits to fix the album artist MBID.) However, as per this PR Picard should still try to handle bad data because it will never be perfect - and the code should probably deal with cases 1 and 3a and perhaps also 3b (since it is only tweaking the fuzzy match and not making hard changes to the data).

3. I think that fudges like this to attempt to correct bad MBID data should have an `options` switch to enable / disable it.

PR #1926, opened 2021-10-21
PICARD-2304: let the user toggle menu icons

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

Some users prefer to disable menu icons.
There were disabled on MacOsX, with no mean to enable it

* JIRA ticket (_optional_): PICARD-2304

# Solution

- those were disabled for MACOSX, it is now disabled by default on those platforms, but user can enable it
- a new boolean option is used to store the status: show_menus_icons

# Action


phw:
> BTW, should I name the option "show_menus_icons" or "show_menu_icons"?

I'd prefer `show_menu_icons`, menus icons sounds weird :) But otherwise LGTM

PR #1935, opened 2021-10-31
Check in sets

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

In conditionals instead of comparing a value against multiple values in a list or tuple use sets. It doesn't make much difference in practice overall, but it technically makes sense. Also in my tests using  `x in {'foo', 'bar', 'baz'}` is 8% faster then `x in ['foo', 'bar', 'baz']` or `x in ('foo', 'bar', 'baz')`. There is no difference in using tuples or lists.

See also discussion at https://github.com/metabrainz/picard/pull/1933

### DIFF
--- a/test/formats/test_apev2.py
+++ b/test/formats/test_apev2.py
@@ -52,14 +52,14 @@ from .common import (
 from .coverart import CommonCoverArtTests
 
 
-VALID_KEYS = [
+VALID_KEYS = {
     ' valid Key}',
     '{ $ome tag~}',
     'xx',
     'x' * 255,
-]
+}
 
-INVALID_KEYS = [
+INVALID_KEYS = {
     'invalid\x7fkey',
     'invalid\x19key',
     '',
@@ -69,10 +69,10 @@ INVALID_KEYS = [
     'TAG',
     'OggS',
     'MP+',
-]
+}
 
 
-SUPPORTED_TAGS = list(set(TAGS.keys()) - apev2.UNSUPPORTED_TAGS)
+SUPPORTED_TAGS = set(TAGS) - apev2.UNSUPPORTED_TAGS
 
 
 class CommonApeTests:
@@ -85,9 +85,9 @@ class CommonApeTests:
 
         def test_supports_tags(self):
             supports_tag = self.format.supports_tag
-            for key in VALID_KEYS + SUPPORTED_TAGS:
+            for key in VALID_KEYS | SUPPORTED_TAGS:
                 self.assertTrue(supports_tag(key), '%r should be supported' % key)
-            for key in INVALID_KEYS + list(apev2.UNSUPPORTED_TAGS):
+            for key in INVALID_KEYS | apev2.UNSUPPORTED_TAGS:
                 self.assertFalse(supports_tag(key), '%r should be unsupported' % key)
 
         @skipUnlessTestfile
--- a/test/formats/test_apev2.py
+++ b/test/formats/test_apev2.py
@@ -72,7 +72,7 @@ INVALID_KEYS = [
 ]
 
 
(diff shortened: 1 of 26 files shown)


test/formats/test_apev2.py
```diff
@@ -87,7 +87,7 @@ def test_supports_tags(self):
             supports_tag = self.format.supports_tag
             for key in VALID_KEYS + SUPPORTED_TAGS:
                 self.assertTrue(supports_tag(key), '%r should be supported' % key)
-            for key in INVALID_KEYS + apev2.UNSUPPORTED_TAGS:
+            for key in INVALID_KEYS + list(apev2.UNSUPPORTED_TAGS):
```
zas, line 90: Here we can go a bit further by also converting `VALID_KEYS` and `INVALID_KEYS` to sets:

```python
VALID_KEYS = {
    ' valid Key}',
    '{ $ome tag~}',
    'xx',
    'x' * 255,
}

INVALID_KEYS = {
    'invalid\x7fkey',
    'invalid\x19key',
    '',
    'x',
    'x' * 256,
    'ID3',
    'TAG',
    'OggS',
    'MP+',
}


SUPPORTED_TAGS = set(TAGS) - set(apev2.UNSUPPORTED_TAGS)


class CommonApeTests:

    class ApeTestCase(CommonTests.TagFormatsTestCase):
        def setup_tags(self):
            super().setup_tags()
            self.unsupported_tags['r128_album_gain'] = '-2857'
            self.unsupported_tags['r128_track_gain'] = '-2857'

        def test_supports_tags(self):
            supports_tag = self.format.supports_tag
            for key in VALID_KEYS | SUPPORTED_TAGS:
                self.assertTrue(supports_tag(key), '%r should be supported' % key)
            for key in INVALID_KEYS | apev2.UNSUPPORTED_TAGS:
                self.assertFalse(supports_tag(key), '%r should be unsupported' % key)
```

Note the use of `|` operator instead of `+`

Also, since `set(TAGS)` is equivalent to `set(TAGS.keys()`

zas (changes requested):
I was working on similar patch, so +1 from me of course :)

zas:
To be exact, there are some cases where `in` a tuple is faster than `in` a set, mainly when searched element at the start of elements we search in.
Also sets will use more memory:

```python
>>> from sys import getsizeof
>>> getsizeof(('a', 'b', 'c'))
64
>>> getsizeof(['a', 'b', 'c'])
80
>>> getsizeof({'a', 'b', 'c'})
216
```

There are no difference (in most cases) with lists vs tuples (of constant elements) because Python convert lists to tuples in such cases:

```python
import dis

def in_tuple(x):
    return x in ('a', 'b', 'c')

def in_list(x):
    return x in ['a', 'b', 'c']

dis.dis(in_tuple)
dis.dis(in_list)
```

```
  5           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 (('a', 'b', 'c'))
              4 COMPARE_OP               6 (in)
              6 RETURN_VALUE
  9           0 LOAD_FAST                0 (x)
              2 LOAD_CONST               1 (('a', 'b', 'c'))
              4 COMPARE_OP               6 (in)
              6 RETURN_VALUE
```
But sets will be consistently faster when the element isn't in the set at all, or when the set is big enough.

But the whole point here is rather stylistic: using the right type for the right case.
If we don't bother order, and intend to use `in` / `not in` to find an element in a list of constant elements, sets make more sense.

test/formats/test_apev2.py
phw: Indeed, that simplifies code even a bit and makes it more readable. Did the changes in https://github.com/metabrainz/picard/pull/1935/commits/aed015c6570cebba5f0a69b12962bd250e1fc630

zas (approved):

PR #1936, opened 2021-11-06
PICARD-2318: Allow AB submission only for files matched to tracks

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2318

This allows submission of AB features only for files matched to tracks, see discussion on https://community.metabrainz.org/t/submitting-acousticbrainz-features-only-possible-from-the-cluster-pane/556418

I also wonder if we should lift the "only saved" files limit. Maybe this should indeed behave more like AcoustID submission and allow submission always if files are matched to tracks. And also then use the recording ID that is currently set for that track from Picard's metadata (basically reverting https://github.com/metabrainz/picard/commit/35976665c9690269b378f64cf08d5b4909119688).

Currently the feature is a bit hard to detect. Also this would allow us to maybe also handle https://community.metabrainz.org/t/please-help-us-test-picard-2-7-beta-1/554211/9 where users use different tag names for the recording ID (even though that needs some additional thought).


picard/file.py
```diff
@@ -710,7 +710,9 @@ def can_view_info(self):
         return True
 
     def can_extract(self):
-        return self.is_saved() and self.metadata["musicbrainz_recordingid"]
+        from picard.track import Track
+        return (self.is_saved() and self.metadata["musicbrainz_recordingid"]
+            and isinstance(self.parent, Track))
```
zas, line 715: I would first test for Track as it is likely faster:

```python
 return (
    isinstance(self.parent, Track)
    and self.is_saved()
    and self.metadata["musicbrainz_recordingid"]
)
```

zas (dismissed):
LGTM

rdswift:
> I also wonder if we should lift the "only saved" files limit.

If we do end up going this way, please prompt me to update the documentation accordingly.

PR #1937, opened 2021-11-07
PICARD-2321: Fix keep coverart for tracks and display of cover art changes for albums

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2321, PICARD-2322

- On tracks with attached files cover art changes are not reset in cover art box when using "Keep original cover art"
- An album loaded with already matched files does not indicate cover art changes correctly

# Solution
- Fix updating tracks in d6435a894135db779e2b56a026a6623b0b690ef2 by remembering the original cover art for a file. This was important to not regress on the fix from 1cb8c77d6ed4737e34bdbeec48581d687531afe8 (after removing all files a track had no cover art attached anymore).
- Albums update their cover art after matching all files (5d2bde5b7a16638af454a23a66fd1c230d81d4fa)
- A small optimization of `imagelist.add_metadata_images` in d4ad11addf8806c18ed6041f27b80aaaf2cc3a8a to only construct a new `ImageList`  if the newly added images are not already present.

### DIFF
--- a/picard/util/imagelist.py
+++ b/picard/util/imagelist.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2017 Antonio Larrosa
 # Copyright (C) 2017 Sambhav Kothari
-# Copyright (C) 2018, 2020 Philipp Wolfer
+# Copyright (C) 2018, 2020-2021 Philipp Wolfer
 # Copyright (C) 2019-2020 Laurent Monin
 # Copyright (C) 2021 Gabriel Ferreira
 #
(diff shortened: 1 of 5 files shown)


phw:
@zas These are strictly multiple separate issues, but I found it easier to change and test this in one go. Best to review by looking at the separate commits. If there is an issue with one of the commits we can put it into a separate PR and discuss things there.

picard/util/imagelist.py
```diff
@@ -219,9 +219,12 @@ def _add_images(metadata, added_images):
         return
```
zas, line 219: We should return a boolean here too

zas (changes requested):

zas (approved):
LGTM

PR #1938, opened 2021-11-09
PICARD-2324: Fix renaming of WavPack correction files

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2324

Renaming / moving WavPack files is supposed to also rename / move a WavPack correction file (with the same filename, but .wvc extension). This broke with the changes for PICARD-1715 due to https://github.com/phw/picard/commit/bcc81d86eb1cfcbe1f72043feec8d50a9accc1f5#diff-8e9e6145ecc7e912b2d643f5064e6f196d3668bd010ffa6037775232c2aba236R357

# Solution

Re-implement renaming the .wvc file by using the actual new filename of the .wv file. This avoids the entire scripting logic to construct the filename, and hence avoids issues with using the `~extension` from metadata. Also avoids issues where a script might cause different naming due to e.g. extension checks.

Also extended the test so it actually fails with the original code.

# Action

### DIFF
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -316,14 +316,19 @@ class WavPackFile(APEv2File):
     NAME = "WavPack"
     _File = mutagen.wavpack.WavPack
 
+    def _move_or_rename_wvc(self, old_filename, new_filename):
+        wvc_filename = replace_extension(old_filename, ".wvc")
+        if not isfile(wvc_filename):
+            return
+        wvc_new_filename = replace_extension(new_filename, ".wvc")
+        wvc_new_filename = get_available_filename(wvc_new_filename, wvc_filename)
+        log.debug('Moving Wavepack correction file %r => %r', wvc_filename, wvc_new_filename)
+        move_ensure_casing(wvc_filename, wvc_new_filename)
+
     def _move_additional_files(self, old_filename, new_filename, config):
         """Includes an additional check for WavPack correction files"""
-        wvc_filename = replace_extension(old_filename, ".wvc")
-        if (config.setting["rename_files"] or config.setting["move_files"]) and isfile(wvc_filename):
-            wvc_new_filename = replace_extension(new_filename, ".wvc")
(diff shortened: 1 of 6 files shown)


zas (approved):
LGTM

picard/formats/apev2.py
```diff
@@ -311,14 +316,15 @@ class WavPackFile(APEv2File):
...
+    def _move_additional_files(self, old_filename, new_filename, config):
         """Includes an additional check for WavPack correction files"""
-        wvc_filename = old_filename.replace(".wv", ".wvc")
-        if isfile(wvc_filename):
-            config = get_config()
-            if config.setting["rename_files"] or config.setting["move_files"]:
-                self._rename(wvc_filename, metadata, config.setting)
-        return File._save_and_rename(self, old_filename, metadata)
+        wvc_filename = replace_extension(old_filename, ".wvc")
+        if (config.setting["rename_files"] or config.setting["move_files"]) and isfile(wvc_filename):
```
zas, line 322: I think we can move wvc code to its own method in order to keep the code cleaner:

```python
def _move_or_rename_wvc(self, old_filename, new_filename):
    wvc_filename = replace_extension(old_filename, ".wvc")
    if not isfile(wvc_filename):
        return
    wvc_new_filename = replace_extension(new_filename, ".wvc")
    wvc_new_filename = get_available_filename(wvc_new_filename, wvc_filename)
    log.debug('Moving Wavepack correction file %r => %r', wvc_filename, wvc_new_filename)
    move_ensure_casing(wvc_filename, wvc_new_filename)

def _move_additional_files(self, old_filename, new_filename, config):
    """Includes an additional check for WavPack correction files"""
    if config.setting["rename_files"] or config.setting["move_files"]:
        self._move_or_rename_wvc(old_filename, new_filename)
    return super()._move_additional_files(old_filename, new_filename, config)
```

zas (changes requested):

zas (approved):
LGTM

PR #1940, opened 2021-11-09
Rework additional files methods

zas:
# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -564,7 +564,6 @@ class File(QtCore.QObject, Item):
 
     def _get_additional_files_moves(self, old_path, new_path, config):
         patterns = self._compile_move_additional_files_pattern(config)
-        moves = set()
         if patterns:
             try:
                 with os.scandir(old_path) as scan:
@@ -575,11 +574,10 @@ class File(QtCore.QObject, Item):
                                 continue
                             if pattern_regex.match(entry.name):
                                 new_file_path = os.path.join(new_path, entry.name)
-                                moves.add((entry.path, new_file_path))
+                                yield (entry.path, new_file_path)
                                 break  # we are done with this file
             except OSError as why:
                 log.error("Failed to scan %r: %s", old_path, why)
-        return moves
 
     def _apply_additional_files_moves(self, old_path, new_path, config):
         for old_file_path, new_file_path in self._get_additional_files_moves(old_path, new_path, config):
--- a/picard/file.py
+++ b/picard/file.py
@@ -553,8 +553,11 @@ class File(QtCore.QObject, Item):
             if new_path != old_path:
                 patterns_string = config.setting["move_additional_files_pattern"]
                 patterns = self._compile_move_additional_files_pattern(patterns_string)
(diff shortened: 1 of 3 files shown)


picard/file.py
```diff
@@ -547,50 +547,37 @@ def _save_images(self, dirname, metadata):
...
-                            break  # we are done with this file
-        except OSError as why:
-            log.error("Failed to scan %r: %s", old_path, why)
-        return moves
+        return {
+            (re.compile(fnmatch.translate(pattern), re.IGNORECASE), pattern.startswith('.'))
+            for pattern in set(config.setting["move_additional_files_pattern"].lower().split())
+        }
+
+    def _get_additional_files_moves(self, old_path, new_path, config):
```
phw, line 562: I'd keep the patterns passed as a parameter here. Then you also don't need to pass the config object and could have tests just for this method.

picard/file.py
```diff
@@ -547,50 +547,37 @@ def _save_images(self, dirname, metadata):
...
+                                new_file_path = os.path.join(new_path, entry.name)
+                                yield (entry.path, new_file_path)
+                                break  # we are done with this file
+            except OSError as why:
+                log.error("Failed to scan %r: %s", old_path, why)
 
-    def _apply_additional_files_moves(self, moves):
-        for old_file_path, new_file_path in moves:
+    def _apply_additional_files_moves(self, old_path, new_path, config):
+        for old_file_path, new_file_path in self._get_additional_files_moves(old_path, new_path, config):
```
phw, line 580: Same here, I'd keep the `moves` as a parameter. This method can be nicely tested, especially when you can easily pass a list of move entries. The method also doesn't care about whether `moves` is an iterator or a list, so this eases testing as well.

phw (changes requested):
The changes make sense. I'd just take the opportunity to add unit tests for the separate methods as well and hence keep the parameter chain.

zas:
> The changes make sense. I'd just take the opportunity to add unit tests for the separate methods as well and hence keep the parameter chain.

I didn't add tests because main method is tested in https://github.com/metabrainz/picard/blob/master/test/test_filesystem.py
But I agree we could test each method individually

phw (approved):
Good job, LGTM

> I didn't add tests because main method is tested

Agreed, not really necessary. But the way the methods are now we could have test or also reuse them more easily if needed.

zas:
> > 
> 
> Agreed, not really necessary. But the way the methods are now we could have test or also reuse them more easily if needed.

I'll go a bit further, and add tests.

BTW, it is even easier to not pass `config` at all, see 93a8d83

zas:
I added more tests, some are a bit redundant, but difficult to avoid.
Overall code is simplified, and has better coverage.

phw (approved):
Great change. I like how it simplifies the tests and makes those more readable.

PR #1941, opened 2021-11-09
PICARD-2328: Fix "arguments did not match any overloaded call" error with Python 3.10

phw:
# Summary

* This is a…
    * [x] Bug fix
    * [ ] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2328

This adds another instance of the "arguments did not match any overloaded call" error with Python 3.10, for which we already had fixes in PICARD-2269.

# Solution

Ensure integer values for 

# Action


zas (approved):
This Python change is likely a never-ending story ;)

PR #1942, opened 2021-11-09
PICARD-2327: Detect hidden files on macOS

phw:
# Summary

* This is a…
    * [ ] Bug fix
    * [x] Feature addition
    * [ ] Refactoring
    * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2327

Picard uses picard.util.is_hidden to check whether a file is hidden. This is used in some places, e.g. Picard by default skips loading hidden files, unless explicitly added.

This works on Windows (with Windows hidden file attribute). On other platforms dot files are considered hidden. But on macOS it does not honor the hidden file attribute.

# Solution

Use macOS API to query hidden state of files. As there are multiple ways why a file can be considered hidden this provides an up-to-date solution even in case of changes. See also https://stackoverflow.com/questions/284115/cross-platform-hidden-file-detection

Added a macOS specific test.

# Action

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -61,12 +61,9 @@ from picard.const import MUSICBRAINZ_SERVERS
 from picard.const.sys import (
     FROZEN_TEMP_PATH,
     IS_FROZEN,
+    IS_MACOS,
     IS_WIN,
 )
-
-
-if IS_WIN:
-    from ctypes import windll
 
 
 class LockableObject(QtCore.QObject):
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

Codacy complains about an unrelated thing (subprocess), I guess we can ignore it (or make it silent?)

phw:
> Codacy complains about an unrelated thing (subprocess), I guess we can ignore it (or make it silent?)

That's Bandit giving us a heads-up that subprocess might be dangerous. It's fine here, so I made it being ignored with `# nosec` on that line.

PR #1948, opened 2021-11-11
_parse_linux_cdrom_info(): optimize a bit

zas:
- iterate through f directly, no need of readline()
- look for 'drive name:' prefix first, and if found, look for 'Can play audio:'
- break ASAP
- do not split lines, use startswith() and splice on length

# Summary

* This is a…
    * [ ] Bug fix
    * [ ] Feature addition
    * [x] Refactoring
    * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/util/cdrom.py
+++ b/picard/util/cdrom.py
@@ -68,17 +68,17 @@ def _generic_iter_drives():
 def _parse_linux_cdrom_info(f):
     drive_names = []
     drive_audio_caps = []
-    while True:
-        line = f.readline()
-        if not line:
+    DRIVE_NAME = 'drive name:'
+    CAN_PLAY_AUDIO = 'Can play audio:'
+    for line in f:
+        if line.startswith(DRIVE_NAME):
+            drive_names = line[len(DRIVE_NAME):].split()
             break
-        if ":" in line:
-            key, values = line.split(':', 1)
-            if key == 'drive name':
-                drive_names = values.split()
-            elif key == 'Can play audio':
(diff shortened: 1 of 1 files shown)


picard/util/cdrom.py
```diff
@@ -68,17 +68,17 @@ def _generic_iter_drives():
...
-            if key == 'drive name':
-                drive_names = values.split()
-            elif key == 'Can play audio':
-                drive_audio_caps = [v == '1' for v in values.split()]
-                break  # no need to continue past this line
+    if drive_names:
+        for line in f:
+            if line.startswith(CAN_PLAY_AUDIO):
+                drive_audio_caps = [v == '1' for v in line[len(CAN_PLAY_AUDIO):].split()]
+                break
```
phw, line 81: Instead of two loops maybe a single loop:

```python
def _parse_linux_cdrom_info(f):
    drive_names = []
    drive_audio_caps = []
    DRIVE_NAME = 'drive name:'
    CAN_PLAY_AUDIO = 'Can play audio:'
    for line in f:
        if line.startswith(DRIVE_NAME):
            drive_names = line[len(DRIVE_NAME):].split()
            if not drive_names:
                break
        elif line.startswith(CAN_PLAY_AUDIO):
            drive_audio_caps = [v == '1' for v in line[len(CAN_PLAY_AUDIO):].split()]
            break
    yield from zip(drive_names, drive_audio_caps)
```
```

phw (approved):

PR #1950, opened 2021-11-12
PICARD-2334: Have script editor consistently save to configuration

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

* JIRA ticket (_optional_): PICARD-2334

Further to the discussion on PICARD-2330 and [IRC](https://chatlogs.metabrainz.org/libera/metabrainz/msg/4908305/), there is some confusion when saving scripts in the script editor because the changes are saved as pending edits requiring clicking "Make It So!" if the Options... dialog is open, but are saved directly to the configuration settings if the Options... dialog is not open.  Also having all new scripts created with the same name is inconvenient.  There is also a bug that could display the "unsaved edits" warning incorrectly after changes have been made to a script's metadata.

# Solution

For the sake of consistency, the proposal is to always have the script editor "Save" action write the changes directly to the configuration settings, without requiring the "Make It So!" action if the Options... dialog is open.  This would only apply to the file naming scripts and selected file naming script, and would not change the way that all other settings are saved.

New scripts will be created with a number in the title to help avoid confusion due to duplicate names.

Have the "Save" action when updating script metadata cascade the save through to the configuration settings, thus avoiding the incorrect "unsaved edits" warning.

# Action

Review the documentation to determine if an update is required to clarify the changes.


picard/ui/scripteditor.py
```diff
@@ -738,12 +739,15 @@ def _insert_item(self, script_item):
...
         self.restore_selected_script_index = idx
+        self.save_to_config()
 
     def new_script(self):
         """Add a new (empty) script to the script selection combo box and script list.
         """
         if self.unsaved_changes_confirmation():
-            script_item = FileNamingScript(script=DEFAULT_FILE_NAMING_FORMAT).to_dict()
+            count = len(self.naming_scripts) + 1
+            title = "{0} ({1})".format(DEFAULT_SCRIPT_NAME, count)
```
zas, line 749: Script title should be translated and I think count based on number of scripts doesn't work well.
I tried something like:

```python
    def new_script_name(self):
        import re
        translated_default_script_name = _(DEFAULT_SCRIPT_NAME)
        escaped = re.escape(translated_default_script_name)
        regex = re.compile('^' + escaped + '(?: \\((\\d+)\\))?$')
        count = 0
        for script in self.naming_scripts.values():
            m = regex.match(script['title'])
            if m:
                num = m.group(1)
                if num is not None:
                    count = max(count, int(num))
                else:
                    count += 1
        return "{0} ({1})".format(translated_default_script_name, count + 1)

    def new_script(self):
        """Add a new (empty) script to the script selection combo box and script list.
        """
        if self.unsaved_changes_confirmation():
            title = self.new_script_name()
            script_item = FileNamingScript(title=title, script=DEFAULT_FILE_NAMING_FORMAT).to_dict()
            self._insert_item(script_item)
```

Much more complex, but it handles translated default script name, and number scripts in a fancier manner.

For example, let's say we start with no script:

- add new script -> Mon script (1)
- rename it -> Mon script
- add new script -> Mon script (2)
- add new script -> Mon script (3)
- delete Mon script (2)
- add new script -> Mon script (4)

I didn't test it fully (we should write tests for it anyway)

zas (changes requested):
I noticed if I create a new script (Add new script) and do not change anything to it, then press Cancel... the script is still saved.

picard/ui/scripteditor.py
rdswift: I like this much better than the simple version I had incorporated, which is what's used for script name generation for tagging scripts.  When I make the change for the naming script, should I also use the same code for generating the new tagging script titles?

rdswift:
> I noticed if I create a new script (Add new script) and do not change anything to it, then press Cancel... the script is still saved.

Yes because new scripts are saved automatically when they are added in the script editor.  All script editor saves go directly to the configuration settings and not through the pending changes managed by the Options... dialog's "Make It So!" and "Cancel" buttons so that it is consistent regardless of whether or not the Options... dialog is open.  If that's not what we want, then I misunderstood the discussion.

I can see that auto-saving new scripts might not be the expected behavior, so perhaps we need to maintain some sort of "pending edits" dictionary in the script editor and add a "Make It So!" button to write them to the configuration.  I'm not really keen on that though because I think it would end up being even more confusing.  Perhaps just not automatically save the new scripts, and only save them when the user clicks the "Save" button?

zas:
> . Perhaps just not automatically save the new scripts, and only save them when the user clicks the "Save" button?

Yes, this is what I was thinking about, it is a bit weird nothing is saved in Options til user confirms (via Make It So) and in this dialog everything is saved on the fly without any user intervention, it makes me wonder about the action of Cancel / Save buttons.

So, to me, we should save scripts on Save in the dialog, and cancel any changes on Cancel, independently from Options dialog.

@phw what do you think?

picard/ui/scripteditor.py
zas: I'll prepare a PR for you to merge in this one.

zas: @rdswift PR ready -> https://github.com/rdswift/picard/pull/10

rdswift:
> So, to me, we should save scripts on Save in the dialog, and cancel any changes on Cancel, independently from Options dialog.

Sort of like a parallel Options dialog, with the same functionality, but only applying to file naming scripts?  I like that.  Should we rename the buttons in the script editor to "Make It So!", "Restore", "Cancel" and "Help" to be somewhat consistent?  In his case:

- "Make It So!" would save the current state of the script editor (i.e. scripts and selected script) and close the script editor.
- "Restore" would set the current state of the script editor to the values currently saved in the configuration settings, but leave the script editor open.  This should first raise a confirmation dialog to allow the user to cancel the operation.
- "Cancel" would close the script editor without saving any changes to the configuration settings, and without raising any "unsaved changes" warning.

The only time the "unsaved changes" warning would be displayed would be if the user tried closing the script editor using the "X" in the window header.

There may be some challenges keeping the selected script synchronized with the combo box in the Options dialog and main screen quick select menu, but that's no different than the current situation.

Now, hopefully I understood you correctly and haven't taken off in a different direction (again).  :stuck_out_tongue_winking_eye:

zas:
> Now, hopefully I understood you correctly and haven't taken off in a different direction (again).

Yes, that's what I think. But I'm interested about @phw take on this.

picard/const/__init__.py
```diff
@@ -186,7 +186,6 @@
     "%title%"
 
 
-DEFAULT_NUMBERED_SCRIPT_NAME = N_("My script %d")
 DEFAULT_SCRIPT_NAME = N_("My script")
 DEFAULT_COVER_IMAGE_FILENAME = "cover"
 DEFAULT_NUMBERED_PROFILE_NAME = N_("My profile %d")
```
zas, line 191: Perhaps we can also use `unique_numbered_title()` for profiles?

rdswift: Excellent idea.  I wish I had thought of it.  I'll have a look.

test/test_util_get_base_title.py
```diff
@@ -0,0 +1,67 @@
...
+
+
+class GetBaseTitle(PicardTestCase):
+    def test_base_title_0(self):
+        test_title = 'title'
+        title = get_base_title(test_title)
+        self.assertEqual(title, 'title')
+
+    def test_base_title_1(self):
+        test_title = 'title {0}'.format(_(DEFAULT_COPY_TEXT),)
```
zas, line 35: That's a bad idea to rely on translated texts in tests, instead I would pass the string as an argument to `get_base_title()`

picard/util/__init__.py
```diff
@@ -891,3 +894,30 @@ def wildcards_to_regex_pattern(pattern):
...
+            else:
+                count += 1
+    return "{0} ({1})".format(default_title, count + 1)
+
+
+def get_base_title(title):
+    """Extract the base portion of a title,
+       removing the "(copy)" and number portion from the end.
+    """
+    re_text = "^(.*)\\s+{0}(\\s+\\(\\d*\\))?$".format(re.escape(_(DEFAULT_COPY_TEXT)))
```
zas, line 921: I would pass `suffix`:

```python
def get_base_title(title, suffix):
    escaped_suffix = re.escape(suffix)
    re_text = "^(.*)\\s+" + escaped_suffix + "(\\s+\\(\\d*\\))?$"
```

Then you can use it with:

```python
get_base_title(title, _(DEFAULT_COPY_TEXT))
```

And, in tests, with a constant string:

```python
get_base_title(title, 'copy')
```

Or even, `c?py` to ensure it is escaped.

test/test_util_get_base_title.py
zas: See my comment above.

It would be more readable without `format()`:

```python
test_title = 'title copy (1)'
title = get_base_title(test_title, 'copy')
```
	
@zas

zas (changes requested):

picard/util/__init__.py
rdswift: Great suggestion.  Thanks.  I did make one change though by making the `suffix` parameter optional and using the `DEFAULT_COPY_TEXT` value if it's missing.  Is that okay?  I know that it will break one of the tests if we ever change the value of `DEFAULT_COPY_TEXT`.  Should I remove that test, because (as you noted earlier) it's a bad idea to rely on translated texts in tests.

zas: >  Thanks. I did make one change though by making the `suffix` parameter optional and using the `DEFAULT_COPY_TEXT` value if it's missing. Is that okay?

Actually I think you should split this in two parts, a simple function for tests, and a convenient for code:

Near its uses:

```python
from picard.util import get_base_title_suffix

DEFAULT_COPY_TEXT = N_("(copy)")

def get_base_title(title):
   return get_base_title_suffix(title, _(DEFAULT_COPY_TEXT))
```

In `picard/util`:

```python
def get_base_title_suffix(title, suffix):
    escaped_suffix = re.escape(suffix)
...

In tests, use `get_base_title_suffix()`

Feel free to use better names ;)

Comment (rdswift) [reply picard/util/__init__.py]: I see where that makes a lot of sense.  I'll make the changes.  Thanks for your patience.

**Comment by zas** [comment]
@rdswift time to clean up:

- commit 8a4570df447872 has a title that doesn't match what it does
- the PR (as is) doesn't address the issue
- commits should be squashed

So, I would create another PR including only script/file titles changes.

What do you think?

**Comment by rdswift** [comment]
> What do you think?

Agreed.  Will do.

**Comment by zas** [comment]
> I noticed if I create a new script (Add new script) and do not change anything to it, then press Cancel... the script is still saved.

Was this addressed?

**Comment by rdswift** [comment]
> > I noticed if I create a new script (Add new script) and do not change anything to it, then press Cancel... the script is still saved.
> 
> Was this addressed?

Not yet.  I planned to look at that as part of the whole change plan regarding your proposed methodology because I suspect that it's tightly integrated.

**Comment by phw** [comment]
I fully agree on making this dialog consistent with options and save separately. @rdswift I'd fully go with your suggestion in harmonizing the behavior. Including renaming the button and change the behavio to save + close. I think that should avoid all the current misunderstandings :)

**Comment by rdswift** [comment]
> I fully agree on making this dialog consistent with options and save separately.

Okay, I'll get started.  Just a couple of questions about synchronizing that I'd like some input:

-----

**Case 1**

Steps:

1. From main Picard screen, open the script editor.
2. Open the Options dialog.
3. In the script editor, add a new script and select it.  This should update the selected script in the Options dialog.
4. Close the script editor without saving.

I assume the Options dialog should be reset to the currently selected script from the saved settings because the new script was not saved so the selector in Options points to a non-existent script.  Is this a reasonable action?

-----

**Case 2**

Steps:

1. From main Picard screen, open the script editor.
2. Open the Options dialog.
3. In the script editor, select a different existing script.  This should update the selected script in the Options dialog.
4. Close the script editor without saving.

In this case the selected script in Options still points to an existing script, so is it reasonable to keep this selection in Options rather than resetting the selection as proposed in Case 1?

-----

**Case 3**

Steps:

1. From main Picard screen, open the script editor.
2. In the script editor, add a new script and select it.  This should update the selected script in the quick selector in the Picard main menu.
3. Close the script editor without saving.

I assume the quick selector should be reset to the currently selected script saved in the settings because the new script was not saved so the selector in the quick selector points to a non-existent script.  Is this a reasonable action?

-----

**Case 4**

Steps:

1. From main Picard screen, open the script editor.
2. In the script editor, select a different existing script.  This should update the selected script in the quick selector in the Picard main menu.
4. Close the script editor without saving.

In this case the selected script in the quick selector still points to an existing script, so is it reasonable to keep this selection in the quick selector rather than resetting the selection as proposed in Case 3?

-----

**Case 5**

Steps:

1. From main Picard screen, open the script editor.
2. Open the Options dialog.
3. In the script editor, add a new script and select it.  This should update the selected script in the Options dialog.
4. Create a new profile that manages the file naming script and activate the profile.
5. Save the Option settings using "Make It So!" in the Options dialog.  This associates the new (unsaved) script with the new profile.
4. Close the script editor without saving.

Unless we do some special processing to save the new script when it gets attached to the new profile (even without the user initiating a save in the script editor), we're going to end up with the profile pointing to a non-existent script.  Should we remove the setting from the profile (i.e. set the selected script to the initial Null value), or do you have any other suggestions?  I suppose we could raise a warning dialog box  when closing the script editor about the script being assigned to a profile but not saved, and that it would be removed from the profile.

This is the case that bothers me the most because the user is going to end up with something inconsistent with the rest of the processing, and this unexpected result could lead to confusion.

**Comment by phw** [comment]
@rdswift Tricky, not 100% sure on some cases.

About Case 1 and 3:
Would it maybe be better not to update the selected script in options / quick selector unless saved? Maybe we should really only synchronize settings when saving. Which probably means we need an action to save without closing the editor dialog (because we want to allow the use case where the editor is kept open for manipulating the script while tagging)?

Or we say changes to the script list get always saved immediately? That means adding or removing a script is permanent and applies immediately. I think that would also solve case 5. Or is it too confusing?

**Comment by rdswift** [comment]
> About Case 1 and 3: Would it maybe be better not to update the selected script in options / quick selector unless saved? Maybe we should really only synchronize settings when saving. Which probably means we need an action to save without closing the editor dialog (because we want to allow the use case where the editor is kept open for manipulating the script while tagging)?

We can do that, but would we then have separate "Save" and "Close" buttons as we do now (different behavior than the Options dialog), and still have the problem about closing the script editor and re-opening it and having changes lost?  Also, if we don't synchronize with the Options dialog when making changes there could be some confusion because the preview lists in the two screens will be different, and the script selector combo boxes may have different content.

> Or we say changes to the script list get always saved immediately? That means adding or removing a script is permanent and applies immediately. I think that would also solve case 5. Or is it too confusing?

That's actually one issue that @zas had reported as being confusing because changes were being saved without clicking the "Save" button.  So far, this seems to be a "no win" situation.  No matter what we decide, I suspect we may end up needing to have to have some sort of pop-up dialog explaining what's happening because it won't be consistently intuitive.  :frowning_face:

**Comment by rdswift** [comment]
Just had a thought about Case 5.  What if when the user clicks "Make It So!" in the the Options dialog, it automatically processes a "Make It So!" in the script editor (without closing the script editor)?  That should be consistent behavior, and avoid any orphaned selected script settings.

**Comment by rdswift** [comment]
Sorry for the massive changes in a forced push, but after rebasing I sort of got carried away.  I think this now addresses most of the issues with unexpected behavior and saving things automatically with no user intervention.  It now doesn't save changes until the user clicks the "Make It So!" button.  When the user resets or closes the script editor, it now checks for profiles linked to unsaved scripts, and provides a warning dialog.  It should be easy enough to change this behavior if we want to do an automatic "Make It So!" in the script editor whenever the user clicks "Make It So!" in the Options dialog.  I've also made some improvements to the real-time synchronization between the script editor and the Options dialog and Main Window.

**Review by zas — CHANGES_REQUESTED** [review]
Overall it is much better, behavior & code-wise.

I got an exception though, while testing, I deleted few unused scripts, pressed Make It So!:

```Traceback (most recent call last):
  File "/home/zas/src/picard/./picard/ui/options/renaming.py", line 183, in show_script_editing_page
    self.script_editor_dialog = ScriptEditorDialog.show_instance(parent=self, examples=self.examples)
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 443, in show_instance
    instance = super().show_instance(*args, **kwargs)
  File "/home/zas/src/picard/./picard/ui/__init__.py", line 153, in show_instance
  File "/home/zas/src/picard/./picard/ui/__init__.py", line 147, in get_instance
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 524, in __init__
    self.load()
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 650, in load
    self.original_script_title = self.naming_scripts[self.original_script_id]['title']
KeyError: 'Preset 1'
```

rdswift:
> I got an exception though, while testing, I deleted few unused scripts, pressed Make It So!:

Good catch!  I don't know how I missed that.  I was only trying to retrieve the title from the list of user scripts and not from the presets.  It now uses the full list, which should fix it.

zas (changes requested):
I encountered another issue:

```
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 697, in closeEvent
    self.reset_script_in_settings()
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 749, in reset_script_in_settings
    script_text = self.naming_scripts[self.selected_script_id]['script']
KeyError: 'Preset 1'
```
It happened when I selected Exit without saving in the menu.

Also Preset entries are suddenly prefixed with "User: "

image

zas (changes requested):
I managed to trigger another one.
From Options > Filenaming > Edit scripts dialog, I deleted all user scripts, kept only presets, make it so, boom.

```python
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 697, in closeEvent
    self.reset_script_in_settings()
  File "/home/zas/src/picard/./picard/ui/scripteditor.py", line 749, in reset_script_in_settings
    script_text = self.all_scripts()[self.selected_script_id]['script']
KeyError: '8ac08dea-811b-4762-a678-c5844eb0dbd5'
```


Also, I noticed something:

one can edit metadata for presets, but then a message saying changes weren't saved appear when exiting dialog, it's a bit weird. I would expect the presets metadata to be clearly read-only (and the message removed).

rdswift:
> I managed to trigger another one.

Crap!  That one change has sure generated a lot of errors.  I've made the correction, and I **THINK** that I have them all now.

> Also, I noticed something:
> 
> one can edit metadata for presets, but then a message saying changes weren't saved appear when exiting dialog, it's a bit weird. I would expect the presets metadata to be clearly read-only (and the message removed).

Yeah, that's how it should be.  I had missed two possibilities for updates which were being caught in the closeing checks.  Should be fixed now.  Presets should now be completely read-only.

zas (commented):
I noticed a small thing that could prolly be improved:

I select the active user script, I delete it, the active script is moved to another one, but I cancel in the dialog, the active script is not restored to the one it was, even though nothing has changed.

I think we should keep track of active user script, and restore it if no change actually occurs.

rdswift:
> I think we should keep track of active user script, and restore it if no change actually occurs.

I think I had that in at one point but it got lost along the way.

Please have a look at the latest and let me know if this is what you had in mind.  It is a quick hack that needs to be optimized, but I want to make sure I'm on th e right track first.

zas:
> > I think we should keep track of active user script, and restore it if no change actually occurs.
> 
> I think I had that in at one point but it got lost along the way.
> 
> Please have a look at the latest and let me know if this is what you had in mind. It is a quick hack that needs to be optimized, but I want to make sure I'm on the right track first.

Yes, it now works as expected. Good job.

rdswift:
In addition to the cleanup, I made one change in the way the script editor operates.  It now blocks selecting a new script (from either the script editor or the parent) or saving using "Make It So!" (in the script editor) if there is an error in the current script.

zas:
Issues I reported were all fixed.

I found another small issue:

If I export a script, then re-import it, I end with 2 scripts with same title:

image

In this case, I think we should ask the user what to do:
- overwrite existing script of the same name
- create a new script (appending a number to name)

What do you think?

rdswift:
> In this case, I think we should ask the user what to do:
> 
> * overwrite existing script of the same name
> * create a new script (appending a number to name)
> 
> What do you think?

That makes sense to me.  I'll see what I can come up with.

zas (approved):
LGTM

picard/ui/scripteditor.py
```diff
@@ -917,36 +1085,45 @@ def delete_script(self):
...
+                    "\n\n"
+                    "Profile: %s"
+                ) % profile.title,
                 QtWidgets.QMessageBox.Ok,
                 self
             ).exec_()
             return
         if confirmation_dialog(self, _('Are you sure that you want to delete the script?')):
+            widget = self.ui.preset_naming_scripts
             idx = self.ui.preset_naming_scripts.currentIndex()
```
zas, line 1099:
```suggestion
            idx = widget.currentIndex()
```

zas (changes requested):

picard/ui/scripteditor.py
```diff
@@ -662,21 +711,134 @@ def eventFilter(self, object, event):
...
+            list: List of Profile named tuples
+        """
+        profiles_list = []
+        config = get_config()
+        profiles = config.profiles[self.PROFILES_KEY]
+        profile_settings = config.profiles[self.SETTINGS_KEY]
+        for profile in profiles:
+            settings = profile_settings[profile["id"]]
+            if self.SELECTED_SCRIPT_KEY in settings:
+                profiles_list.append(self.Profile(profile["id"], profile["title"], settings[self.SELECTED_SCRIPT_KEY]))
```
zas, line 810:
```suggestion
                profiles_list.append(
                    self.Profile(
                        profile["id"],
                        profile["title"],
                        settings[self.SELECTED_SCRIPT_KEY]
                    )
                )
```

zas (approved):

zas (dismissed):

phw (changes requested):
This is indeed much better, great work.

I found two small issues, though:

1. If I start Picard with empty picard.ini and try to open the file editor dialog I get:

```
E: 18:35:46,829 script.get_file_naming_script:127: Unable to retrieve the file naming script ''
  File ".\picard\ui\mainwindow.py", line 1645, in open_file_naming_script_editor
    self.script_editor_dialog = ScriptEditorDialog.show_instance(parent=self, examples=self.examples)
  File ".\picard\ui\scripteditor.py", line 444, in show_instance
    instance = super().show_instance(*args, **kwargs)
  File ".\picard\ui\__init__.py", line 153, in show_instance
  File ".\picard\ui\__init__.py", line 147, in get_instance
  File ".\picard\ui\scripteditor.py", line 527, in __init__
    self.load()
  File ".\picard\ui\scripteditor.py", line 663, in load
    self.original_script_title = self.all_scripts()[self.original_script_id]['title']
KeyError: ''
```

2. If I open the script editor from Options. If I change the script in the script editor to a user script other than my first user script, then the selection will at first also change in options dialog. But after hotting "Make it so" in script dialog it jumps back to previous selection in options dialog.

    This strangely does not happen when I change to a preset or to the first user script. In those cases the option dialog shows the change and keeps it after saving and closing the script editor with "Make it so".

rdswift:
> 1. If I start Picard with empty picard.ini and try to open the file editor dialog I get:

Good catch!  I added a check for missing or invalid selected script id in the loading() method, so this should now default to "Preset 1".

> 2. If I open the script editor from Options. If I change the script in the script editor to a user script other than my first user script, then the selection will at first also change in options dialog. But after hotting "Make it so" in script dialog it jumps back to previous selection in options dialog.
>    This strangely does not happen when I change to a preset or to the first user script. In those cases the option dialog shows the change and keeps it after saving and closing the script editor with "Make it so".

This I haven't been able to duplicate here.  Can you provide more details such as whether you are starting with a blank ini file and the exact sequence of steps?  Thanks.

picard/ui/scripteditor.py
```diff
@@ -656,6 +656,8 @@ def load(self, reload=False):
         config = get_config()
         self.naming_scripts = config.setting[self.SCRIPTS_LIST_KEY]
         self.selected_script_id = config.setting[self.SELECTED_SCRIPT_KEY]
+        if not self.selected_script_id or self.selected_script_id not in self.naming_scripts:
+            self.selected_script_id = "Preset 1"
```
zas, line 660: Instead of "Preset 1", I think we should store this in a constant (`DEFAULT_PRESET_ID`?)

zas (changes requested):

rdswift:
> 2\. If I open the script editor from Options. If I change the script in the script editor to a user script other than my first user script, then the selection will at first also change in options dialog. But after hotting "Make it so" in script dialog it jumps back to previous selection in options dialog.
>     This strangely does not happen when I change to a preset or to the first user script. In those cases the option dialog shows the change and keeps it after saving and closing the script editor with "Make it so".

@phw, following our discussion it did appear to be a result of a race condition, with the Options dialog not getting the updated information directly from the script editor on closing.  I tried writing the updated information to the config object in the editor and then reading it from config in the Options dialog and that seems to have corrected the issue.  I still haven't been able to reproduce the error you described, so you'll have to test this revision and let me know if it solved the issue.  Thanks.

phw (dismissed):
With the latest changes this works without issues for me. Great work!

rdswift:
> With the latest changes this works without issues for me. Great work!

Except it causes a regression that breaks the behavior with an issue that @zas had brought up earlier (and had been fixed).  I'll need to have a look again.

rdswift:
@phw, this latest change fixes the regression and still seems to work here regarding your issue.  Can you please check again to make sure that it works for you.  Thanks.

zas:
> @phw, this latest change fixes the regression and still seems to work here regarding your issue. Can you please check again to make sure that it works for you. Thanks.

@rdswift can you squash commits e7c5d71 and e4bf13a ?

`git rebase -i e7c5d714d^^^` and fix, then force push.

image

zas (approved):

phw:
The last commit e253ef4956fc6150be278613a515bce4619b6290 fixed the issue for me.

PR #1951, opened 2021-11-13
Optimize file AcoustID dequeueing

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

- when a file is moved, or removed, the AcoustID tasks queue is totally copied and rewritten, even if file isn't in it
- 

# Solution

- introduce AcoustIDTask named tuple for convenience
- implement a class that acts as both a deque and a defauldict of sets
- the deque is used as previously to get next task
- the defaultdict of sets ensures we can find tasks to remove in O(1)
- popleft is modified to check if the file was previously removed

# Action

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -23,7 +23,11 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-from collections import deque
+from collections import (
+    defaultdict,
+    deque,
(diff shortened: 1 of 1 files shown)


zas:
Logic is flawed.
It doesn't work if file is removed from queue then re-added

zas:
I tried another approach

zas:
This doesn't work, too complicated and not thread-safe.

PR #1952, opened 2021-11-13
Fix AcoustID succesful lookup debug logging

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

- it was logged more than once

# Solution

- move outside the loop
- add the number of found recordings

# Action

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -113,12 +113,23 @@ class AcoustIDClient(QtCore.QObject):
                                 parsed_recording['score'] = score * result_score
                                 parsed_recording['acoustid'] = result['id']
                                 recording_list.append(parsed_recording)
-                        log.debug("AcoustID: Lookup successful for '%s'", task.file.filename)
 
-                    # Set AcoustID in tags if there was no matching recording
-                    if results and not recording_list:
-                        task.file.metadata['acoustid_id'] = results[0]['id']
-                        task.file.update()
+                    if results:
+                        if not recording_list:
+                            # Set AcoustID in tags if there was no matching recording
+                            task.file.metadata['acoustid_id'] = results[0]['id']
+                            task.file.update()
+                            log.debug(
+                                "AcoustID: Found no matching recordings for '%s',"
+                                " setting acoustid_id tag to %r",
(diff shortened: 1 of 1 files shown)


picard/acoustid/__init__.py
```diff
@@ -106,12 +106,18 @@ def _on_lookup_finished(self, next_func, file, document, http, error):
...
-                    # Set AcoustID in tags if there was no matching recording
-                    if results and not recording_list:
-                        file.metadata['acoustid_id'] = results[0]['id']
-                        file.update()
+                    if results:
+                        if not recording_list:
+                            # Set AcoustID in tags if there was no matching recording
+                            file.metadata['acoustid_id'] = results[0]['id']
+                            file.update()
+                        else:
```
phw, line 115: I guess we should have some debug output here as well. Something like "AcoustID: Found for '%s', but recording matches below threshold".

phw (changes requested):

picard/acoustid/__init__.py
zas: Done, and rebased

picard/acoustid/__init__.py
```diff
@@ -113,12 +113,22 @@ def _on_lookup_finished(self, task, document, http, error):
...
-                    # Set AcoustID in tags if there was no matching recording
-                    if results and not recording_list:
-                        task.file.metadata['acoustid_id'] = results[0]['id']
-                        task.file.update()
+                    if results:
+                        if not recording_list:
+                            # Set AcoustID in tags if there was no matching recording
+                            task.file.metadata['acoustid_id'] = results[0]['id']
+                            task.file.update()
+                            log.debug(
```
phw, line 122: Actually the message was not correct, sorry. It's not about the threshold here. This just catches the case where there is an AcoustID that is not matched to any MB recording (as it can happen if AcoustID submission is done with a tool other than Picard, or when all links to recordings get removed)

zas: Ah yes, fixed in c6b7067

phw (approved):

PR #1954, opened 2021-11-16
Add functions for creating unique numbered script/profile titles

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**: Adds standard functions for creating unique numbered titles for scripts and profiles.

# Problem

New tagger scripts were uniquely numbered based on the number of existing scripts, causing unusual numbering and possible duplicate names if scripts were deleted and new scripts added.  Profiles and naming scripts were not uniquely named.

# Solution

Add standard functions for creating unique numbered titles for scripts and profiles.  This includes `unique_numbered_title()` which produces a properly numbered title based on a base name and list of the current titles, and `get_base_title()` / `get_base_title_with_suffix()` which extracts the base title from a title with a specified suffix to properly number multiple copies of a title.

- Add utility function unique_numbered_title() -- code provided by @zas 
- Add utility function to extract base title to allow creating unique numbered title for titles that have been copied.

# Action

None.


zas (approved):
LGTM

picard/util/__init__.py
```diff
@@ -891,3 +894,38 @@ def wildcards_to_regex_pattern(pattern):
...
+    count = 0
+    for title in existing_titles:
+        m = regex.match(title)
+        if m:
+            num = m.group(1)
+            if num is not None:
+                count = max(count, int(num))
+            else:
+                count += 1
+    return "{0} ({1})".format(default_title, count + 1)
```
phw, line 914: I think to be more flexible with localization we should have the format string translated, e.g.

    "{title} ({count})".format(title=default_title, count=count + 1)

That allows localization to have this formatted as it best fits for the language, including replacing the `(` and `)` with different characters.

phw (changes requested):
I see some potential of reusing `unique_numbered_title` elsewhere :)

Overall this looks great, but it regresses a bit in making the resulting string localizable. See my comment.

picard/util/__init__.py
rdswift: I agree, but that will also require the regex on Line 904 to somehow be translated.  Any suggestions how that might be handled?

zas: Not trivial, we could perhaps parse the translated format string but how to handle `(?: )` part?
If the fmt string is reverse '({count}) {title}' for example, the optional part is anything but `{title}', so split in two parts:
`({count}) ` and `{title}`, replace `{count}` with  `(\\d+)` and wrap with `(?:` + `)`.
It would also mean translate the format string inside the method.

I'll give it a try.

rdswift: > It would also mean translate the format string inside the method.

And the regex used on Line 922 in the `get_base_title_suffix()` method.

zas: https://github.com/rdswift/picard/pull/11

zas: I didn't do `get_base_title_suffix()` but it probably be done the same way

picard/util/__init__.py
```diff
@@ -926,13 +926,25 @@ def wrap_count(p):
...
+            return '(?:' + re.escape(p).replace('\\{count\\}', '\\d*') + ')?'
+        else:
+            return p
+
     escaped_suffix = re.escape(suffix)
-    re_text = "^(.*)\\s+" + escaped_suffix + "(\\s+\\(\\d*\\))?$"
-    match_obj = re.fullmatch(re_text, title)
+    joiner = '(.*?)(?:\\s*' + escaped_suffix + ')?'
+    regstr = '^' + joiner.join([wrap_count(p) for p in parts]) + '$'
+    match_obj = re.fullmatch(regstr, title)
```
phw, line 947: Alternative suggestion, which I think is a bit more clear to understand:

```python
reg_title = r'(?P<title>.*?)(?:\s*' + escaped_suffix + ')?'
reg_count = r'(?:\d+)?'
regstr = '^' + re.escape(fmt)
    .replace(r'\{title\}', reg_title)
    .replace(r'\{count\}', reg_count)
    .replace(' ', r'\s+') + '$'
match_obj = re.fullmatch(regstr, title)
return match_obj['title'] if match_obj else title
```

phw (commented):
Yes, did not think about the regex. Your solutions work, though. Makes things a bit more complex and not so nice, but maybe we could put this into a helper function. Or we could get away with not localizing this, it should be ok mostly.

picard/util/__init__.py
rdswift: I tried out your suggested code, but it fails most of the tests because it doesn't take into account the brackets around `{count}` in the format string.  I'll play with it a bit and see if I can work around that.

rdswift: I think this latest commit incorporates your suggestion to improve readability and clarifies what's happening.

zas: We can reduce code redundancy: see https://github.com/rdswift/picard/pull/12
Also we don't need `^` / `$` if we use `re.fullmatch()`

phw (approved):

zas (approved):

PR #1956, opened 2021-11-18
Introduce new class ABExtractor in order to get extractor dynamically

zas:
- use a cache to prevent useless version & sha calculations
- based cache key on stat.st_mtime_ns + path
- declare ab_extractor as a property of tagger class
- pass tagger to all methods that needs it

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/acousticbrainz/__init__.py
+++ b/picard/acousticbrainz/__init__.py
@@ -59,6 +59,9 @@ ABExtractorProperties = namedtuple('ABExtractorProperties', ('path', 'version', 
 class ABExtractor:
 
     def __init__(self):
+        self._init_cache()
+
+    def _init_cache(self):
         self.cache = defaultdict(lambda: None)
 
     def get(self, config=None):
@@ -101,6 +104,8 @@ class ABExtractor:
                     sha=sha,
                     mtime_ns=mtime_ns
                 )
+                # clear the cache, we keep only one entry
+                self._init_cache()
                 self.cache[(result.path, result.mtime_ns)] = result
                 log.debug('ABExtractor: caching: %r', result)
                 return result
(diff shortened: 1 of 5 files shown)


picard/acousticbrainz/__init__.py
```diff
@@ -48,63 +51,97 @@
...
+
+        try:
+            statinfo = os.stat(extractor_path)
+            mtime_ns = statinfo.st_mtime_ns
+        except OSError as exc:
+            log.warning('ABExtractor: cannot stat extractor: %s', exc)
+            return None
+
+        # check if we have this in cache already
+        cached = self.cache[(extractor_path, mtime_ns)]
```
phw, line 88: Do we really need a list of cache entries? I'd actually go with a single cache item. If path changes it can be re-evaluated. That avoids the question what to do with stale items. Not really an issue, because this usually wouldn't change that often during one run of Picard, but still it seems odd to keep the old entries around.

phw (commented):
This is a good improvement. I'm confident the AB tests are now stable again.

Just a small question for discussion, but otherwise LGTM

picard/acousticbrainz/__init__.py
zas: That's not a list but a defaultdict.
Actually it was more because that's simpler to test if there's something in the cache (key is a (path, time) tuple / O(1) check). I thought about potential old entries, but it will happen only in case of changes, so at worse we'll have few entries in the dictionary.
I thought about that of course, but I concluded it wasn't worth bothering.
Without any change to extractor path or binary during runtime, we'll have just one entry (and none at startup).

zas: Done in e72c552df0

phw (approved):

PR #1957, opened 2021-11-20
PICARD-2325: Show error message on login errors

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2325

If the OAuth login fails Picard does not show any error message. The access code input dialog just closes.

# Solution
In options dialog, show an error message. If the authorization dialog is shown use a dialog to display the error.

# Action
The use case where we display a dialog is not very nice. Picard opens multiple message boxes for this (first dialog asking if the user wants to login, second to enter the access token, and potentially a third with error message), and this could be repeated for every outstanding request.

It would be nicer to have a single dialog, that handles it all and which then also sets the credentials for all future requests.

Unfortunately this does not work with the current network implementation, see my comment on PICARD-1638. This should be solved, but this is out of scope for this PR. Hence the message box solution.


phw:
I added fa73b7d3e00909f67a40f520fb2efded113bb637, which changes the layout definition for the login view a bit. Visually it is identical, but instead of this awkward grid layout it uses a vertical layout + horizontal layout. I am quite sure this makes the workaround for PICARD-1654 obsolete, which even might be the cause for PICARD-2320

picard/oauth.py
```diff
@@ -244,14 +247,24 @@ def fetch_username(self, callback):
...
                 successful = True
         except Exception as e:
             log.error('OAuth: Unexpected error handling username fetch response: %r', e)
+            error_msg = _('Unexpected authentication error')
         finally:
-            callback(successful)
+            callback(successful=successful, error_msg=error_msg)
+
+    def _extract_error_description(self, http, data):
+        if self.webservice.http_response_code(http) == 400:
```
zas, line 266: Shouldn't we handle any error code here?

phw: In that case we have a defined behavior and know how to extract the error message.

What would you do for other cases? Return a generic message, maybe with the HTTP code?

zas: I was just thinking about something like:

 ```python
http_error_code = self.webservice.http_response_code(http)
if http_error_code >= 400:
    response = load_json(data)
    if 'error_description' in response:
        return response['error_description'] + " (%d)" % http_error_code
```

phw: In practice this would just fail completely in most cases except the expected 400, the else would most likely not be hit. That's because as long as the endpoint will be reached and reacts the only error it will return is a 400. The other cases are more serious, such as a 5xx code by the webserver, or maybe a generic 404 error. Those likely won't parse as JSON.

I did it a bit different in c03716a25b752365f5f0c206b5982246130c326f so it shows a fallback message whenever parsing the JSON or reading `error_description`  fails. I also wouldn't expose the HTTP code in the 400 case to the UI. That's already part of the webserver module's logging.

zas (approved):
LGTM

PR #1958, opened 2021-11-21
Album cleanup

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -517,7 +517,7 @@ class Album(DataObject, Item):
             self.update(update_tracks=False)
             add_metadata_images(self, [file])
 
-    def _remove_file(self, track, file, new_album=True):
+    def remove_file(self, track, file, new_album=True):
         self._files -= 1
         if new_album:
             self.update(update_tracks=False)
--- a/picard/album.py
+++ b/picard/album.py
@@ -660,6 +660,11 @@ class Album(DataObject, Item):
                     count += 1
         return count
 
+    def _iter_unsaved_files(self):
+        for file in self.iterfiles(save=True):
+            if not file.is_saved():
+                yield file
+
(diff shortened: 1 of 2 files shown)


picard/album.py
```diff
@@ -645,12 +645,7 @@ def is_complete(self):
             return True
 
     def is_modified(self):
-        if self.tracks:
-            for track in self.tracks:
-                for file in track.files:
-                    if not file.is_saved():
-                        return True
-        return False
+        return self.get_num_unsaved_files() > 0
```
phw, line 648: Not sure about this. It saves us a few lines of code, but for long list of tracks returning on first modified is a serious performance improvement. What about something like this (untested)?

```python
    def is_modified(self):
        for file in self._iter_unsaved_files():
            return True
        return False

    def get_num_unsaved_files(self):
        count = 0
        for file in self._iter_unsaved_files():
            count += 1
        return count

    def _iter_unsaved_files(self):
        for track in self.tracks:
            for file in track.files:
                if not file.is_saved():
                    yield file
```

phw: Or even for _iter_unsaved_files:

```python
    def _iter_unsaved_files(self):
        for file in self.iterfiles(save=True):
            if not file.is_saved():
                yield file

phw (changes requested):

picard/album.py
zas: Yes, you're right, that's much better. Done in f3cb6735458179558ae82910efb33d1424b0c562

zas:
I re-organized commits to include changes suggested by @phw

phw (approved):

PR #1963, opened 2021-11-23
PICARD-2339: Faster clustering implementation

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2339

The existing clustering code is using the Levenshtein distance to calculate similarity, which caused a O(n^2) performance. But since only exactly similar matches are being used (similarity threshold 1.0) this was is not strictly necessary.

If we don't use a threshold (and Picard never did) then we can simplify the code.

# Solution

This new implementation uses simple comparison for string equality and performs in O(n).

Added tests to verify consistent behavior.

There is a small change in behavior: Previously if there were multiple different spelling for what is considered the same album title the last encountered spelling would be used. The new implementation uses the first one.

See https://github.com/metabrainz/picard/commit/760af4dc7a45801df2a86eee0afc794d172d26a9#diff-2d1bf33b26a905bd43b846bc0f989f2b3ff96f82da68719d465a330ff0e20ebeL94

This patch is based on my older work on https://github.com/phw/picard/commit/e31534dbe712b51d0b631e031351be9481a093fa

# Action

We should do a performance comparison.

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -36,13 +36,11 @@
 
 
 from collections import defaultdict
-from enum import IntEnum
-from heapq import (
-    heappop,
-    heappush,
-)
 import ntpath
-from operator import attrgetter
+from operator import (
+    attrgetter,
+    itemgetter,
+)
 import re
 
 from PyQt5 import QtCore
@@ -54,7 +52,6 @@ from picard.metadata import (
     Metadata,
     SimMatchRelease,
 )
-from picard.similarity import similarity
 from picard.util import (
     album_artist_from_path,
     find_best_match,
@@ -306,12 +303,19 @@ class Cluster(FileList):
             self.lookup_task = None
 
     @staticmethod
-    def cluster(files, threshold, tagger=None):
+    def cluster(files, tagger=None):
+        """Group the provided files into clusters, based on album tag in metadata.
+
+        Args:
+            files: List of File objects.
+            tagger: Tagger instance (optional). Only needed for statusbar updates.
+
+        Yields:
+            FileCluster objects
+        """
+        cluster_list = defaultdict(FileCluster)
         config = get_config()
         win_compat = config.setting["windows_compatibility"] or IS_WIN
-        artist_dict = ClusterDict()
-        album_dict = ClusterDict()
-        tracks = []
         num_files = len(files)
 
         # 10 evenly spaced indexes of files being clustered, used as checkpoints for every 10% progress
@@ -320,6 +324,7 @@ class Cluster(FileList):
         for i, file in process_events_iter(enumerate(files)):
             artist = file.metadata["albumartist"] or file.metadata["artist"]
             album = file.metadata["album"]
+
             # Improve clustering from directory structure if no existing tags
             # Only used for grouping and to provide cluster title / artist - not added to file tags.
             if win_compat:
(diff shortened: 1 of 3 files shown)


picard/cluster.py
```diff
@@ -443,183 +415,35 @@ def lookup_metadata(self):
...
+        self._artists = defaultdict(lambda: defaultdict(lambda: 0))
+        self._titles = defaultdict(lambda: 0)
 
-    def get_token(self, index):
-        word, token = self.ids[index]
-        return token
+    def add(self, album, artist, file):
+        self.files.append(file)
+        self._artist_counts[tokenize(artist)] += 1
+        self._artists[tokenize(artist)][artist] += 1
```
zas, line 428: I don't think we need `self._artist_counts` since we can use `len(self._artists[artist_token])` instead.

zas (changes requested):

picard/cluster.py
phw: It would actually be `sum(self._artists[artist_token])`. Not sure. We would first need to take the artist with highest sum, and then from that the individual highest "spelling variation". The separate artist_counts looks more efficient. But I'm not really happy about the implementation of the `artist` property. There must be a better way :D

zas: As far I understand this code, `self._artist_counts[tokenize(artist)] == len(self._artists[tokenize(artist)])` but that's not what you want it seems.
If you want artist full name having most occurrences in `self._artists[tokenize(artist)]`, that's `max(self._artists[tokenize(artist)])`.

phw: Still no :) It's a two step process of first choosing the most common artist, and then the most common spelling. We need an example. We have a cluster with different files and the following artist entries:

- artist 1
- ARTIST 1
- ARTIST 1
- artist1
- artist 2
- artist 2
- artist 2

Because of `tokenize` we have two entries in `_artists`: `artist1`, `artist2`.

`artist1` has 4 occurrences, `artist2` has 3. So we want to use `artist1`. That`s currently memorized in `_artist_count`.

`artist1` has 3 different spellings, "artist 1", "ARTIST 1" and "artist1". "ARTIST 1" is the most common, with 2 occurrences. That's what we want to use. These counts of different spellings for `artist1` are currently memorized in `self._artists["artist1"]`.

So to clarify: In the example `self._artist_counts[tokenize("artist1")] == 4`, but `len(self._artists[tokenize("artist1")]) == 3`. 

> If you want artist full name having most occurrences in self._artists[tokenize(artist)], that's max(self._artists[tokenize(artist)], key=self._artists[tokenize(artist)].get).

That's actually nicer then the itemgetter approach I used, so that means it could be:

```
@property
def artist(self):
    tokenized_artist = max(self._artist_counts, key=self._artist_counts.get)
    return max(self._artists[tokenized_artist], key=self._artists[tokenized_artist].get)

@property
def title(self):
    # Find the most common title
    return max(self._titles, key=self._titles.get)
```

zas: Yes, this is what I figured, see irc PM

phw:
@zas: I think this is ready know for full review. I extended the tests a bit and added missing behavior for empty artists. The old implementation then hard coded "Various Artists". I made this use the va_name from settings, since this is a slight change in behavior I added PICARD-2340 for this.

Next step would be to run the clustering itself in a thread to offload it from main thread. Although it is now reasonable fast I think we still should do that. It is pure calculation, and the more files you cluster the longer it takes. But I'll do this in a separate PR, not sure what side effects this might have.

zas (approved):
LGTM, good work!

picard/cluster.py
```diff
@@ -320,61 +324,31 @@ def cluster(files, threshold, tagger=None):
         for i, file in process_events_iter(enumerate(files)):
             artist = file.metadata["albumartist"] or file.metadata["artist"]
             album = file.metadata["album"]
+
             # Improve clustering from directory structure if no existing tags
             # Only used for grouping and to provide cluster title / artist - not added to file tags.
             if win_compat:
                 filename = ntpath.splitdrive(file.filename)[1]
             else:
                 filename = file.filename
```
phw, line 333: @zas This is a part I did not touch here, but I wonder about the need for this. If we are on Windows I think `ntpath.splitdrive` should always be done by `album_artist_from_path`, so you don't end up with an artist named "C:".

But do you see any need to have this for other platforms if `config.setting["windows_compatibility"]`  is True? It's really only needed to split this Windows drive letter like "C:\\" or that UNC path to a share, such as "\\\\MyNAS\\theshare\\". But I would not have such paths on e.g. Linux or macOS, and ntpath.splitdrive has no effect on paths like e.g. "/mount/mynas/theshare/xyz".

So I would propose we move the call to `ntpath.splitdrive` to  `album_artist_from_path`, but only check for IS_WIN.

zas: Yes, I agree.
But I would do that in another PR.

PR #1965, opened 2021-11-23
Have album_artist_from_path strip the Windows drive letter

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

`album_artist_from_path` is called by the clustering, which removes the Windows drive letter from the path to avoid that the drive gets detected as artist or album name.

This functionality should be part of `album_artist_from_path` itself. It also only applies to Windows, not for other systems, even in Windows compatibility mode (see discussion at https://github.com/metabrainz/picard/pull/1963#discussion_r755199555)

# Solution

Extend `album_artist_from_path` to use `ntpath.splitdrive` on Windows. Add tests.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -48,6 +48,7 @@ import json
 import ntpath
 from operator import attrgetter
 import os
+from pathlib import PurePath
 import re
 import subprocess  # nosec: B404
 import sys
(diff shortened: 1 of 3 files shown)


picard/util/__init__.py
```diff
@@ -541,9 +541,19 @@ def linear_combination_of_weights(parts):
...
+        filename: The full file path
+        album: Default album name
+        artist: Default artist name
+
+    Returns:
+        A tuple (album, artist)
     """
     if not album:
+        if IS_WIN:
+            filename = ntpath.splitdrive(filename)[1]
```
zas, line 556: We can even get rid of this using `pathlib.PurePath`

https://github.com/phw/picard/pull/13

zas (changes requested):

zas (approved):

PR #1971, opened 2021-11-25
Convert DataObject.genres to Counter, simplfify _convert_folksonomy_tags_to_genre()

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -43,7 +43,6 @@ from collections import (
     Counter,
     defaultdict,
 )
-from itertools import filterfalse
 import re
 import traceback
 
@@ -120,8 +119,10 @@ class TagGenreFilter:
                 return True
         return False
 
-    def filter(self, list_of_tags):
-        return list(filterfalse(self.skip, list_of_tags))
+    def filter(self, counter):
+        for name, count in counter:
+            if not self.skip(name):
+                yield (name, count)
 
 
 class TrackArtist(DataObject):
@@ -328,9 +329,7 @@ class Track(DataObject, FileListItem):
         # Filter by name and usage
         genres_filter = TagGenreFilter(filters)
         genres_list = []
-        for name, count in most_common_genres:
-            if genres_filter.skip(name):
-                continue
+        for name, count in genres_filter.filter(most_common_genres):
             percent = 100 * count // topcount
             if percent < minusage:
                 break
--- a/picard/track.py
+++ b/picard/track.py
@@ -345,22 +345,22 @@ class Track(DataObject, FileListItem):
 
     def _convert_folksonomy_tags_to_genre(self):
         config = get_config()
-        # Combine release and track tags
-        tags = Counter(self.genres)
-        tags += self.album.genres
+        # Combine release and track genres
+        genres = Counter(self.genres)
+        genres += self.album.genres
         if self.album.release_group:
-            tags += self.album.release_group.genres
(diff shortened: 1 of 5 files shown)


picard/track.py
```diff
@@ -311,41 +314,45 @@ def _customize_metadata(self):
     def _convert_folksonomy_tags_to_genre(self):
         config = get_config()
         # Combine release and track tags
-        tags = dict(self.genres)
-        self.merge_genres(tags, self.album.genres)
+        tags = Counter()
+        tags += self.genres
```
phw, line 318: This could be 

```
tags = Counter(self.genres)
```

phw (dismissed):

zas:
@phw sorry, I rebased a bunch of times, you can review now ;)

picard/track.py
```diff
@@ -308,48 +311,62 @@ def _customize_metadata(self):
...
+        most_common_genres = genres.most_common(limit)
+        topcount = most_common_genres[0][1]
+
+        # Filter by name and usage
+        genres_filter = TagGenreFilter(filters)
+        genres_list = []
+        for name, count in most_common_genres:
+            if genres_filter.skip(name):
+                continue
+            percent = 100 * count // topcount
```
phw, line 334: `TagGenreFilter` already has a `filter` method, which we don't actually use. I think we should have `TagGenreFilter.filter(genres)` which accepts tuples of genre + count and yields only the unfiltered ones so we can do:

```python
for name, count in genres_filter.filter(most_common_genres):
    percent = 100 * count // topcount
    ...
```

What do you think?

phw (changes requested):

picard/track.py
zas: Good idea, done in 8d87135

picard/track.py
```diff
@@ -308,48 +312,60 @@ def _customize_metadata(self):
...
+
+        # Filter by name and usage
+        genres_filter = TagGenreFilter(filters)
+        genres_list = []
+        for name, count in genres_filter.filter(most_common_genres):
+            percent = 100 * count // topcount
+            if percent < minusage:
+                break
+            name = _TRANSLATE_TAGS.get(name, name.title())
+            genres_list.append(name)
```
phw, line 337: As the following code block is very small I would do this as 

```
if percent >= minusage:
    name = _TRANSLATE_TAGS.get(name, name.title())
    genres_list.append(name)
``` 

I think that reads clearer. In general I think we should avoid the early `break`  / `continue`  calls except for cases where it clearly makes sense or avoids overly deep indentation.

phw (commented):
This code is a great simplification. Just a minor nitpick from my side

picard/track.py
zas: Nope, that would be different.
The list we iterate in is sorted, the break actually reduces the number of iterations, inverting the test and removing it would lead to the same result, but with (possibly) more iterations.

phw: Righ, I misread that.

phw (approved):

phw:
Looks like this conflicts with the changes in https://github.com/metabrainz/picard/commit/543709eab244a926a8f5972d50b6a9fed75ab3c9 . But once that's solved LGTM

PR #1979, opened 2021-11-28
Refactor webservice queue

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

Move the request task queue handling of the `WebService`  class to a separate `RequestPriorityQueue` class.

This simplifies testing of the functionality and reduces complexity of the `WebService` itself. Also applied some minor additional refactoring.

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -368,7 +368,6 @@ class WebService(QtCore.QObject):
         self._active_requests = {}
         self._queue = RequestPriorityQueue(ratecontrol)
         self.num_pending_web_requests = 0
-        self._last_num_pending_web_requests = -1
 
     def _init_timers(self):
         self._timer_run_next_task = QtCore.QTimer(self)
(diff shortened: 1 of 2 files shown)


zas (approved):
Much cleaner code, good job!

picard/webservice/__init__.py
```diff
@@ -260,6 +260,64 @@ def _init_headers(self):
...
+            self._count -= 1
+        except Exception as e:
+            log.debug(e)
+
+    def run_ready_tasks(self):
+        delay = sys.maxsize
+        for prio in sorted(self._queues.keys(), reverse=True):
+            prio_queue = self._queues[prio]
+            if not prio_queue:
+                del(self._queues[prio])
```
zas, line 304: I just noticed this syntax but according to https://docs.python.org/3/reference/simple_stmts.html#the-del-statement `del x` is the correct syntax, `del(x)` is interpreted as `del (x)`, and parentheses are ignored (see https://docs.python.org/3/reference/simple_stmts.html#grammar-token-python-grammar-target_list).
Elsewhere in the code we use `del x`, is there a specific reason to not do the same here?

PR #1981, opened 2021-12-03
PICARD-2351: Fix system language detection on Windows

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2351

System language detection for Windows broke in Picard 2.7.0b3 with https://github.com/metabrainz/picard/commit/0816eb61bfac1928611ce3b8c29b23479929560f

# Solution

It turns out that the call to `locale.setlocale(locale.LC_ALL, current_locale)` on Windows did not work most of the time because of the encoding. E.g. on a German Windows 11 the locale we used was `de_DE.cp1252`, but passing this to `locale.setlocale`  raises an `locale.Error`  with "unsupported locale setting".

So let's just set the locale based on the language alone, only `de_DE`.

The reason why language selection worked anyway is because prior to 0816eb61bfac1928611ce3b8c29b23479929560f we set the `de_DE.cp1252` in `os.environ['LANGUAGE']` and `os.environ['LANG']` anyway, and that at least made gettext load the proper language.

# Action

### DIFF
--- a/picard/i18n.py
+++ b/picard/i18n.py
@@ -44,9 +44,7 @@ if IS_WIN:
     def _init_default_locale():
         try:
             current_locale = locale.windows_locale[windll.kernel32.GetUserDefaultUILanguage()]
-            current_locale += '.' + locale.getpreferredencoding()
-            locale.setlocale(locale.LC_ALL, current_locale)
-            return current_locale
+            return locale.setlocale(locale.LC_ALL, current_locale)
         except KeyError:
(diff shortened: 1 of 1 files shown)


zas (approved):

phw:
I also explicitly tested this on Linux and macOS. The code paths there were not affected and system language detection works on those platforms.

PR #1983, opened 2021-12-04
Add more tests and optimize similarity2()

zas:
- we mainly use this method to match tracks & files, and the common case is perfect match
- astrcmp() call is expensive (even for the C version)
- if lengths are identical, and lists are perfectly identical, just return 1.0
- we don't need total, as it is equal to len(alist) when exiting the loop
- with test `if not alen or not blen` we are sure a division by zero cannot occur

The gain is marginal on random strings, but in practice we compare much more identical strings than different strings.

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/similarity.py
+++ b/picard/similarity.py
@@ -53,12 +53,22 @@ _split_words_re = re.compile(r'\W+', re.UNICODE)
 
 def similarity2(a, b):
     """Calculates similarity of a multi-word strings."""
+    if not a or not b:
+        return 0.0
+    if a == b:
+        return 1.0
+
     alist = list(filter(bool, _split_words_re.split(a.lower())))
     blist = list(filter(bool, _split_words_re.split(b.lower())))
-    total = 0
-    score = 0.0
-    if len(alist) > len(blist):
+
+    alen, blen = len(alist), len(blist)
+    if not alen or not blen:
+        return 0.0
+    if alen > blen:
         alist, blist = blist, alist
+        alen, blen = blen, alen
+
+    score = 0.0
     for av in alist:
         ms = 0.0
         mp = None
@@ -71,9 +81,6 @@ def similarity2(a, b):
             score += ms
             if ms > 0.6:
                 del blist[mp]
-        total += 1
-    total += len(blist) * 0.4
-    if total:
-        return score / total
-    else:
-        return 0
+
+    # division by zero cannot happen, alen > 0 at this point
+    return score / (alen + len(blist) * 0.4)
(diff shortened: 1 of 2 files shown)


picard/similarity.py
```diff
@@ -53,12 +53,22 @@ def similarity(a1, b1):
 
 def similarity2(a, b):
     """Calculates similarity of a multi-word strings."""
+    if not a or not b:
+        return 0.0
+    if a == b:
+        return 1.0
```
phw, line 59: I think the equality check should go first. I see tests currently go the opposite, but I'd expect `similarity2("", "") == 1.0`. What did the old code do?

picard/similarity.py
```diff
@@ -53,12 +53,22 @@ def similarity(a1, b1):
...
+
     alist = list(filter(bool, _split_words_re.split(a.lower())))
     blist = list(filter(bool, _split_words_re.split(b.lower())))
-    total = 0
-    score = 0.0
-    if len(alist) > len(blist):
+
+    alen, blen = len(alist), len(blist)
+    if not alen or not blen:
+        return 0.0
```
phw, line 66: Here again, what is expected if both are 0?

phw (changes requested):
Good improvements. Just not sure about the results for empty inputs. Does the previous code pass your new tests?

picard/similarity.py
zas: I was also expecting `similarity2("", "") == 1.0` but that's not what the old code does, so I preserved this behavior in order to ensure nothing breaks.

picard/similarity.py
zas: Check my first commit, it had tests before I made any change, I preserved the previous behavior, so if a or b (or both) are empty, it returns 0.0.

zas:
> Does the previous code pass your new tests?

Yes, this is why changes for tests are in the first commit of this PR.
That said, the case of empty strings is unlikely to happen at runtime (it would mean no title for a track or a file).
So I don't know if this behavior is a real choice, but perhaps changing it would have zero impact on overall behavior. I preserved the previous behavior because I didn't want to take this risk at this point.

phw (approved):
Ok, thanks for the clarification. I'm in favor of preserving the behavior then, just to be sure we do not introduce some subtle issue with existing behavior.

But as you said it might just be accidental behavior and not an intentional choice. We could change this in a future release.

zas:
> We could change this in a future release.

Yes, it will let us simplify the code a bit while being much more logical.

PR #1984, opened 2021-12-04
PICARD-2349: Use consistent terminology for "standalone recordings"

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Standardize on "Standalone recordings" rather than "Non-album tracks"

# Problem

* JIRA ticket (_optional_): PICARD-2349

For historical reasons Picard uses both the terms "standalone recording" and the older "non-album tracks". We should unify this.

This requires updates to existing translations, even though some translations might end up being unchanged because they maybe already use consistent terms.

# Solution

Standardize on "Standalone recordings" rather than "Non-album tracks".

# Action

### DIFF
--- a/picard/profile.py
+++ b/picard/profile.py
@@ -56,7 +56,7 @@ class UserProfileGroups():
         "title": N_("Metadata"),
         "settings": [
             SettingDesc("va_name", N_("Various Artists name"), ["va_name"]),
-            SettingDesc("nat_name", N_("Non-album tracks name"), ["nat_name"]),
+            SettingDesc("nat_name", N_("Standalone recordings name"), ["nat_name"]),
             SettingDesc("translate_artist_names", N_("Translate artist names"), ["translate_artist_names"]),
             SettingDesc("artist_locales", N_("Translation locales"), ["selected_locales"]),
             SettingDesc("translate_artist_names_script_exception", N_("Translate artist names exception"), ["translate_artist_names_script_exception"]),
--- a/picard/ui/options/metadata.py
+++ b/picard/ui/options/metadata.py
@@ -83,7 +83,7 @@ class MetadataOptionsPage(OptionsPage):
 
     options = [
         TextOption("setting", "va_name", "Various Artists"),
-        TextOption("setting", "nat_name", "[non-album tracks]"),
+        TextOption("setting", "nat_name", "[standalone recordings]"),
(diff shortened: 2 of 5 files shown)


zas (approved):

phw:
I would prefer not to merge this for the 2.7 release to not introduce additional translation changes.

Also in regards to the ticket I created, I intended to actually also change internal naming in code. But that's something I'd delay for Picard 3 really

rdswift:
> I would prefer not to merge this for the 2.7 release to not introduce additional translation changes.
> 
> Also in regards to the ticket I created, I intended to actually also change internal naming in code. But that's something I'd delay for Picard 3 really

Understood.  In that case, should I cancel this PR and ressurect it after the 2.7 release, and include the change to rename the settings keys?

phw:
> Understood. In that case, should I cancel this PR and resurect it after the 2.7 release, and include the change to rename the settings keys?

I don't mind the string changes separately. But even those I would probably do for a Picard 2.8 to avoid translation changes for 2.7. So let's just keep this PR open as it is and merge it once we are out. Usually we keep the master branch open for a short while for a patch release after a 2.x release. The internal code changes we should then do for Picard 3.

Maybe for some background info because mentioning Picard 3 will now come up more often: zas and I had a rough planing on how to proceed. For various reasons it gets time to think about a Picard 3. The main driver behind this is that we need to start updating to Qt6 if we want to support Apple Silicon, the new ARM processor of the newer macs. That means also we will break compatibility with some older systems, and it means breaking some plugins. But it also means we can do some changes to Picard that we did not do because of preserving compatibility. I don't currently think we can do everything we would like to do, but there are a few core things we should tackle (e.g. improvements to plugins, replacement of QSettings and new networking / webservice module).

So the rough plan is that we do the 2.7 + necessary patch releases, than go for a 2.8 which we consider kind of a preparation for the Qt6 migration. That 2.8 I think we should maintain for a while and do patch releases. But after the 2.8 release we in general move to Python 3 development. We should at least aim to put out an alpha version of Picard 3 out in the second half of next year. But otherwise we'll release it when it is ready. That also means we should see to not make the focus of Picard 3 too wide, we can do releases afterwards.

PR #1986, opened 2021-12-05
Mainwindow cleanup: get rid of get_files_from_objects()

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1351,22 +1351,25 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             for obj in self.selected_objects:
                 if obj is None:
                     continue
-                if obj.can_analyze():
-                    can_analyze = True
-                if obj.can_save():
-                    can_save = True
-                if obj.can_remove():
-                    can_remove = True
-                if obj.can_refresh():
-                    can_refresh = True
-                if obj.can_autotag():
-                    can_autotag = True
-                if obj.can_submit():
-                    can_submit = True
-                if obj.can_extract():
-                    can_extract = True
+                # using x = x or obj.x() form prevents calling function
+                # if x is already True
+                can_analyze = can_analyze or obj.can_analyze()
+                can_autotag = can_autotag or obj.can_autotag()
+                can_extract = can_extract or obj.can_extract()
+                can_refresh = can_refresh or obj.can_refresh()
+                can_remove = can_remove or obj.can_remove()
+                can_save = can_save or obj.can_save()
+                can_submit = can_submit or obj.can_submit()
                 # Skip further loops if all values now True.
-                if can_analyze and can_save and can_remove and can_refresh and can_autotag and can_submit and can_extract:
+                if (
+                    can_analyze
+                    and can_autotag
+                    and can_extract
+                    and can_refresh
+                    and can_remove
+                    and can_save
+                    and can_submit
+                ):
                     break
         can_extract = can_extract and self.tagger.ab_extractor.available()
         self.remove_action.setEnabled(can_remove)
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1266,7 +1266,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         elif isinstance(selected, Track):
             dialog_class = TrackInfoDialog
         else:
-            selected = self.tagger.get_files_from_objects(self.selected_objects)[0]
+            selected = next(iter_files_from_objects(self.selected_objects))
             dialog_class = FileInfoDialog
         dialog = dialog_class(selected, self)
         dialog.ui.tabWidget.setCurrentIndex(default_tab)
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -802,10 +802,11 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         config.setting["dont_write_tags"] = not checked
 
     def get_selected_or_unmatched_files(self):
-        files = self.tagger.get_files_from_objects(self.selected_objects)
-        if not files:
-            files = self.tagger.unclustered_files.files
-        return files
+        if self.selected_objects:
+            files = list(iter_files_from_objects(self.selected_objects))
+            if files:
+                return files
+        return self.tagger.unclustered_files.files
 
     def open_tags_from_filenames(self):
         files = self.get_selected_or_unmatched_files()
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1338,28 +1338,36 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         can_view_info = bool(single and single.can_view_info())
         can_browser_lookup = bool(single and single.can_browser_lookup())
         is_file = bool(single and isinstance(single, (File, Track)))
-        have_files = bool(self.tagger.get_files_from_objects(self.selected_objects))
-        have_objects = bool(self.selected_objects)
-        for obj in self.selected_objects:
-            if obj is None:
-                continue
-            if obj.can_analyze():
-                can_analyze = True
-            if obj.can_save():
-                can_save = True
-            if obj.can_remove():
-                can_remove = True
-            if obj.can_refresh():
-                can_refresh = True
-            if obj.can_autotag():
-                can_autotag = True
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -1210,10 +1207,11 @@ def open_folder(self):
     def _ensure_fingerprinting_configured(self, callback):
         config = get_config()
 
-        def on_finished(result):
-            callback(config.setting['fingerprinting_system'])
         if not config.setting['fingerprinting_system']:
             if self._show_analyze_settings_info():
+                def on_finished(result):
+                    callback(config.setting['fingerprinting_system'])
```
phw, line 1213: Maybe use functools.partial:

```
on_finished = partial(callback, config.setting['fingerprinting_system'])
```

picard/ui/mainwindow.py
```diff
@@ -1332,32 +1329,44 @@ def update_actions(self):
         can_autotag = False
         can_submit = False
         can_extract = False
-        single = self.selected_objects[0] if len(self.selected_objects) == 1 else None
+
+        try:
+            single = self.selected_objects[0]
+        except IndexError:
+            single = None
```
phw, line 1336: This is different. `single` was only set if exactly one object had been selected. It is being used to enable actions that should only be enabled for a single selection.

phw (changes requested):
Good improvements. I especially like how it gets completely rid of `Tagger.get_files_from_objects` and at least in case of `MainWindow.update_actions` makes it much more efficient.

I think all uses of `Tagger.get_files_from_objects` have now been removed and we only need to keep it around for plugins. Let's explicitly add a note about deprecation to the docstring so we don't forget and can remove it in Picard 3.

Also would you mind if we don't merge this yet for Picard 2.7? After review I'm pretty sure it does what it should except for the single selection issue. But let's keep things safe :)

picard/ui/mainwindow.py
zas: Obviously, I missed that, I'll rework that.

picard/ui/mainwindow.py
zas: Hmmm, actually `partial()` cannot be used like this as `on_finished()` takes one (unused) argument (`result`).

zas:
> Also would you mind if we don't merge this yet for Picard 2.7? After review I'm pretty sure it does what it should except for the single selection issue. But let's keep things safe :)

I splitted this PR, and we'll merge `get_files_from_objects` changes after 2.7.

Cosmetic changes are now in https://github.com/metabrainz/picard/pull/1989 and I rebased this branch, dropping the change to single.

picard/ui/mainwindow.py
phw: Indeed, I missed that.

picard/ui/mainwindow.py
zas: This change was removed; code rebased.

picard/ui/mainwindow.py
```diff
@@ -1338,28 +1338,36 @@ def update_actions(self):
...
+                if obj.can_remove():
+                    can_remove = True
+                if obj.can_refresh():
+                    can_refresh = True
+                if obj.can_autotag():
+                    can_autotag = True
+                if obj.can_submit():
+                    can_submit = True
+                if obj.can_extract():
+                    can_extract = True
```
phw, line 1367: We could also do:

```
can_analyze |= obj.can_analyze()
can_save |= obj.can_save()
can_remove |= obj.can_remove()
can_refresh |= obj.can_refresh()
can_autotag |= obj.can_autotag()
can_submit |= obj.can_submit()
can_extract |= obj.can_extract()
```

phw: I think that even prevents calling the method if the variable is already True.

picard/ui/mainwindow.py
```diff
@@ -1440,7 +1448,9 @@ def update_selection(self, objects=None, new_selection=True, drop_album_caches=F
                                                history=None)
         elif coverart_visible and new_selection:
             # Create a temporary file list which allows changing cover art for all selected files
-            files = self.tagger.get_files_from_objects(objects)
+            files = list(iter_files_from_objects(objects))
+            # TODO: it should be possible to pass the generator directly to FileList(),
+            # but underlying code has to be checked
```
phw, line 1453: Actually no, at least not in general. The passed file list is passed on to `FileListItem`, and that just sets the passed value. And some entity objects assume this is a list and use `len()` or index access on files.

It might for some reason work in this specific case where we have this temporary `FileList` currently only for cover art updating. But if you pass on an iterator you also pass on state, and I wouldn't keep this iterator around for storing the actual list of files attached to an object for longer. E.g. `iter_files_from_objects(objects)` depends on iterating over the list of files of each object, and that might change. So the files we currently think we pass on to FileList might not be the ones that get later used when this iterator is finally evaluated.

picard/ui/mainwindow.py
zas: Done in aefd5e7f3bbcb1477b62e91f3296507a38a6f4c1

picard/ui/mainwindow.py
zas: I removed the TODO comment and rebased

picard/ui/mainwindow.py
```diff
@@ -1258,18 +1259,17 @@ def show_more_albums(self):
...
+        if isinstance(selected, Album):
+            dialog_class = AlbumInfoDialog
+        elif isinstance(selected, Cluster):
+            dialog_class = ClusterInfoDialog
+        elif isinstance(selected, Track):
+            dialog_class = TrackInfoDialog
         else:
-            file = self.tagger.get_files_from_objects(self.selected_objects)[0]
-            dialog = FileInfoDialog(file, self)
+            selected = next(iter_files_from_objects(self.selected_objects))
```
phw, line 1270: We had this issue with the previous code also, but it `self.selected_objects` guaranteed to yield a file here?

zas: Actually, no. Plus there's a potential race condition if `selected_objects` is modified by another thread, so the test at start of this method doesn't guarantee we have objects. Nor we have guarantee first item is a `File` item.
So it could be safer to use a `try/except` block, catching `StopIteration`.

Also, `FileInfoDialog` could raise an exception if we pass objects which aren't handled by it, but that's another issue.

```python
try:
    selected = next(iter_files_from_objects(self.selected_objects))
except StopIteration:
   return
dialog_class = FileInfoDialog
```

And at start of method, we could replace:

```python
if not self.selected_objects:
    return
selected = self.selected_objects[0]
```

with

```python
try:
    selected = self.selected_objects[0]
except IndexError:
    return
```

zas: See bb80e23

rdswift:
> I think all uses of `Tagger.get_files_from_objects` have now been removed and we only need to keep it around for plugins. Let's explicitly add a note about deprecation to the docstring so we don't forget and can remove it in Picard 3.

I don't know how you normally handle depecation notices, but would it also make sense to include an entry in the log when the deprecated method is called?  Then again, who actually reads the logs...  :wink:

zas:
> > I think all uses of `Tagger.get_files_from_objects` have now been removed and we only need to keep it around for plugins. Let's explicitly add a note about deprecation to the docstring so we don't forget and can remove it in Picard 3.
> 
> I don't know how you normally handle depecation notices, but would it also make sense to include an entry in the log (warning level) when the deprecated method is called? Then again, who actually reads the logs... wink

https://pypi.org/project/Deprecated/ could be handy

phw:
@zas shall we see to get this merged for 2.7.1? I think this looks great, and actually should be an improvement. I combination with the already merged #1967 we should have reduced the calls to some of the item capabilitiy functions significantly in some cases.

Also +1 for https://pypi.org/project/Deprecated/, we should do this for Picard 2.8 (actually go through the code and mark methods accordingly)

zas:
> @zas shall we see to get this merged for 2.7.1? I think this looks great, and actually should be an improvement. I combination with the already merged #1967 we should have reduced the calls to some of the item capabilitiy functions significantly in some cases.
> 

Ok, for 2.7.1

> Also +1 for https://pypi.org/project/Deprecated/, we should do this for Picard 2.8 (actually go through the code and mark methods accordingly)

Ok, for 2.8 ;)

phw (approved):

PR #1991, opened 2021-12-07
dropMimeData(): use iter_files_from_objects()

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -77,6 +77,7 @@ from picard.track import (
 )
 from picard.util import (
     icontheme,
+    iter_files_from_objects,
     natsort,
     normpath,
     restore_method,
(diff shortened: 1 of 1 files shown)


picard/ui/itemviews.py
```diff
@@ -784,10 +785,12 @@ def dropMimeData(self, parent, index, data, action):
         # application/picard.album-list
         albums = data.data("application/picard.album-list")
         if albums:
-            albums = [self.tagger.load_album(id) for id in bytes(albums).decode().split("\n")]
+            album_ids = bytes(albums).decode().split("\n")
+            log.debug("Dropped albums = %r", album_ids)
+            files = list(iter_files_from_objects(self.tagger.load_album(id) for id in album_ids))
```
phw, line 790: I think we don't need the list conversion here, passing on the iterator would be ok. Especially because this still runs on the main thread later on in tagger.move_files.

The advantage would be that we move even more work out of the drop handler

zas: Done & rebased in 8f90884

phw (approved):

PR #1996, opened 2021-12-14
PICARD-2360: When comparing metadata to a release consider the actual medium track count

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2360

tracktotal in a file is supposed to have the total number of tracks on the specific medium. Comparing to the total tracks of a release across all medium hence gives wrong results.

# Solution

# Action

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -252,8 +252,15 @@ class Metadata(MutableMapping):
 
         try:
             a = int(self["totaltracks"])
-            b = release['track-count']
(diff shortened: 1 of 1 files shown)


phw:
Please don't merge this, I'm still investigating the details. Also this probably is not a full fix for PICARD-2360 but rather a part of it (and maybe should be put into a separate issue). The other part might be giving the preferred format (and also preferred release country) more weight.

PR #1997, opened 2021-12-15
PICARD-2362: Do not include __pycache__ dirs in sdist.

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2362

The published sdist contains `__pycache__` directories in test/data. This breaks reproducible builds.

# Solution

Exclude those __pycache__ directories

# Action

I'm not totally happy with that solution, as it needs to explicitly define the exact paths to exclude. I found no more generic way with wildcards or similar that would exclude them. The problem seems to be that the `recursive-include test/data *` includes everything regardless, and only explicitly excluding it helped in my testing. Suggestions for more generic fix welcome.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -10,6 +10,8 @@ include *.txt
 
 recursive-include test *.py
 recursive-include test/data *
+recursive-exclude test/data/testplugins/module/dummyplugin/__pycache__ *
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #1998, opened 2021-12-17
PICARD-2363: Fix ID3v2.3 showing as changed after saving

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2363

Copying the new metadata over to the original metadata after saving must consider format specific conversions.

This was caused by a5fcde1 as part of PICARD-2174. This PR here actually complements the fix done in #1982.

# Solution

When copying new metadata to original metadata after saving apply the format conversions.

I chose this approach as it might avoid some other similar issues, and it matches the behavior before a5fcde1 . The alternative would be to have https://github.com/phw/picard/blob/40b2b76ce7f150321d40bdf8e2051023d97a8a6a/picard/file.py#L689 also use `self.format_specific_metadata()`, but that would add additional calls every time update is being called, something we should avoid. Doing this once on saving, when the values get copied anyway, is more efficient.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -182,6 +182,17 @@ class File(QtCore.QObject, Item):
         """
         return metadata.getall(tag)
 
+    def _format_specific_copy(self, metadata, settings=None):
+        """Creates a copy of metadata, but applies format_specific_metadata() to the values.
+        """
+        copy = Metadata(
+            deleted_tags=metadata.deleted_tags,
+            images=metadata.images,
+            length=metadata.length)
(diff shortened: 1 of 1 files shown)


zas (dismissed):
LGTM

phw:
@zas this makes it nicer: 823756ca04f552d831533198e1cfe0cf68d0187d

zas (approved):

PR #2001, opened 2021-12-18
PICARD-2366: fix auto analyze

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2366

Auto analyze / auto cluster only works when files are added via drag and drop, but not if added via Add files / Add folder dialogs. That's because in this case there is not drop target, but Tagger.move_files still did consider them having been moved to a proper target.

# Solution

Handle `None`  target.

Also ensure the real target the file has been moved to is considered, as files might get moved to different target for various reasons. E.g. dropping a file onto a file will move the file to the target file's parent. If that is unclustered files the real target is unclustered files and the auto features should trigger.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -468,7 +468,7 @@ class Tagger(QtWidgets.QApplication):
                 file_moved = True
 
         if not file_moved:
-            self.move_file(file, target)
+            target = self.move_file(file, target)
             if target and target != self.unclustered_files:
                 file_moved = True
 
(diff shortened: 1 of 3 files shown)


picard/tagger.py
```diff
@@ -485,21 +485,26 @@ def _file_loaded(self, file, target=None, remove_file=False, unmatched_files=Non
...
+                target = target.parent
+            # To be able to move a file the target must implement add_file(file)
+            if not hasattr(target, 'add_file'):
+                # Ensure a file always has a parent so it shows up in UI
+                if not file.parent:
+                    target = self.unclustered_files
+                # Unsupported target, do not move the file
+                else:
+                    target = None
+            if target is not None:
```
zas, line 505: I think `File.move()` should handled the `target is None` case itself.

https://github.com/metabrainz/picard/blob/1415787a873e71667b605a68afd00305d2d60417/picard/file.py#L611-L623

If we call `File.move(None)`, it will likely raise an exception when hitting https://github.com/metabrainz/picard/blob/1415787a873e71667b605a68afd00305d2d60417/picard/file.py#L618 or https://github.com/metabrainz/picard/blob/1415787a873e71667b605a68afd00305d2d60417/picard/file.py#L622

So I think we should check for it:

```python
    def move(self, parent):
        if parent is not None and parent != self.parent:
```

zas (approved):
It makes sense.

picard/tagger.py
phw: Good point. Or actually even do check for `hasattr(parent, 'add_file')`

phw: @zas Please check 598417a22a940b8a7104eb92339575c111f60eef

zas (approved):

PR #2002, opened 2021-12-18
PICARD-2367: Fix crash when moving album to cluster

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2367

Dragging an album to the "Clusters" item triggers a crash with 
```
  File "/home/phw/devel/musicbrainz/picard/./picard/tagger.py", line 528, in move_files
    self.cluster(files)
  File "/home/phw/devel/musicbrainz/picard/./picard/tagger.py", line 910, in cluster
    if len(objs) <= 1 or self.unclustered_files in objs:
TypeError: object of type 'generator' has no len()
```

That's caused by the change in #1991, where the dropped files are now passed as an iterator instead of a list.

# Solution

Make tagger.cluster be able to handle an iterator instead of a list

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -46,6 +46,7 @@
 import argparse
 from concurrent.futures import ThreadPoolExecutor
 from functools import partial
+import itertools
 import logging
 import os.path
 import platform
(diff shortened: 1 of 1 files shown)


zas (dismissed):
LGTM

phw:
@zas Did a small change, from

```
if len(files) > 1:
    cluster = self.load_cluster(file_cluster.title, file_cluster.artist)
    cluster.add_files(files)
else:
    self.unclustered_files.add_files(files)
```

to 

```
if len(files) > 1:
    cluster = self.load_cluster(file_cluster.title, file_cluster.artist)
else:
    cluster = self.unclustered_files
cluster.add_files(files)
```

zas (approved):

PR #2005, opened 2021-12-21
PICARD-2372: Always save profiles first before other settings

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2372

Saving options does not always save settings to profiles. That's because the profiles can get updated after other settings

# Solution

Always save profiles first before other settings.
Ensures the profiles are up to date and all saved settings end up in the profiles without getting overwritten later.

# Action


zas (approved):
Nice catch

PR #2006, opened 2021-12-21
PICARD-2373: Make `$slice()` 'end' parameter optional

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Make the `$slice()` 'end' parameter optional.

# Problem

* JIRA ticket (_optional_): PICARD-2373

The documentation for the `$slice()` scripting function is incorrect because it shows the `end` parameter as optional.  The parameter can be left empty, but it still requires that it be included (thus requiring an additional comma in the call).  In discussing corrections to the documentation on https://github.com/metabrainz/picard-docs/pull/141 it was decided to correct the operation of the function to make the `end` parameter truly optional.

# Solution

Make the `$slice()` 'end' parameter optional.  Update tests.

# Action

Update documentation.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1258,12 +1258,13 @@ Returns a multi-value variable containing the elements between the `start` and
 The following example will create a multi-value variable with all artists
     in `%artists%` except the first, which can be used to create a "feat." list.
 
-Example:
+Examples:
 
-    $setmulti(supporting_artists,$slice(%artists%,1,))
+    $setmulti(supporting_artists,$slice(%artists%,1))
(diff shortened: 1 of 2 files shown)


phw (approved):
Makes sense to me, let's merge that.

I'd like to keep translation changes to a minimum for a 2.7.2 release, which I think we should do in mid January. But I think this specific minor change would be ok for this if we do it now.

zas (approved):

PR #2007, opened 2021-12-22
PICARD-2375: Fix progress indicator staying at 0% for pending network requests

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2375

If there are only pending network requests but no file requests the loading progress stays near 0%.

The issue is caused by https://github.com/metabrainz/picard/blob/master/picard/ui/infostatus.py#L96-L101

There the starting time gets reset on each call if there are no more pending files to load, ignoring any requests. So once there are only pending requests it always has basically 100% progress still to do.

# Solution

Consider both file and network requests and only reset if they both reached zero.

# Action


zas (approved):
LGTM

PR #2008, opened 2021-12-22
PICARD-2374: Remove rate limit for archive.org

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2374

Fetching cover art from cover art archive involves both requests to coverartarchive.org for the webservice and archive.org for the actual image downloads.

Picard currently does remove the default rate limit for coverartarchive.org, but keeps the 1 second ddefault delay for archive.org.

As archive.org does not apply a rate limit, we should remove this delay as well

# Solution
Set rate limit for archive.org to 0. This also applies to subdomains. While the ratecontrol module keeps the adjusted delays separate for each subdomain, it copies the minimum delay from the main domain.

Also moved setting the limit to the CAA implementation, as I think knowledge of which hosts get queried should be kept close to the modules using them.

# Action

### DIFF
--- a/picard/coverart/providers/caa.py
+++ b/picard/coverart/providers/caa.py
@@ -68,6 +68,7 @@ from picard.coverart.utils import (
     CAA_TYPES,
     translate_caa_type,
 )
+from picard.webservice import ratecontrol
 
 from picard.ui import PicardDialog
 from picard.ui.ui_provider_options_caa import Ui_CaaOptions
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

PR #2009, opened 2021-12-22
PICARD-2377: Fix potential crash on emptydirs initialization

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2377

If one of the QStandardPaths pointed to a no longer available location Picard would crash on startup

# Solution
Handle the exception. The actual exception thrown was a `FileNotFoundError`, but https://docs.python.org/3/library/os.path.html#os.path.realpath documents it as throwing `OSError` (which is the superclass of `FileNotFoundError`), so handle that instead.

# Action


zas (approved):
LGTM

PR #2010, opened 2021-12-22
PICARD-2379: Add clarification text to scripting options ui

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add clarification text to scripting options ui.

# Problem

* JIRA ticket (_optional_): PICARD-2379

It is not intuitive what the check boxes signify on the scripting options page.

# Solution

Add clarification text to scripting options ui.  Example showing proposed changes:

image

As I understand the plans, changes requiring additional translation will be held until the Picard 2.8 release.

# Action

Update screen shot in the documentation.

### DIFF
--- a/picard/ui/ui_options_script.py
+++ b/picard/ui/ui_options_script.py
@@ -20,6 +20,14 @@ class Ui_ScriptingOptionsPage(object):
         self.enable_tagger_scripts.setObjectName("enable_tagger_scripts")
         self.verticalLayout = QtWidgets.QVBoxLayout(self.enable_tagger_scripts)
         self.verticalLayout.setObjectName("verticalLayout")
+        self.horizontalLayout_2 = QtWidgets.QHBoxLayout()
+        self.horizontalLayout_2.setContentsMargins(-1, -1, -1, 0)
+        self.horizontalLayout_2.setObjectName("horizontalLayout_2")
+        self.label = QtWidgets.QLabel(self.enable_tagger_scripts)
+        self.label.setWordWrap(True)
+        self.label.setObjectName("label")
+        self.horizontalLayout_2.addWidget(self.label)
+        self.verticalLayout.addLayout(self.horizontalLayout_2)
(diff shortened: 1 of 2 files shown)


picard/ui/ui_options_script.py
```diff
@@ -108,7 +116,8 @@ def setupUi(self, ScriptingOptionsPage):
 
     def retranslateUi(self, ScriptingOptionsPage):
         _translate = QtCore.QCoreApplication.translate
-        self.enable_tagger_scripts.setTitle(_("Tagger Script(s)"))
+        self.enable_tagger_scripts.setTitle(_("Enable Tagger Script(s)"))
+        self.label.setText(_("<html><head/><body><p>Checked scripts will be executed automatically, once for each track in the release, when Picard retrieves release information from MusicBrainz.</p></body></html>"))
```
phw, line 120: Maybe "Activated" instead of "Checked"? But not sure. For me "Checked" just makes it look more like "Select", and that usually is done to do something with the selected items. But here it is clearly to activate / disable the script.

We could also call it "Enable", but that conflicts with the "Enable" of the entire feature, so it's confusing.

phw (dismissed):
In general yes. But Let's do it for 2.8, yes

picard/ui/ui_options_script.py
rdswift: How about something like, "Scripts that have been activated below will be executed automatically, once for each track in the release, when Picard retrieves release information from MusicBrainz."?

phw: Yes, good approach. It's technically not fully correct, as the script need also to be re-run also when matching files. But the goal is to have the scripts process the metadata of each loaded track, the rest is an implementation details. So maybe even simpler:

> Scripts that have been activated below will be executed automatically for each track of a release loaded from MusicBrainz.

phw (approved):

PR #2011, opened 2021-12-23
PICARD-2378: Ensure AB submission submits only base filename

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2378

AB submission on Windows submits the entire file path. It's essentially a bug in the extractor binary (https://github.com/MTG/essentia/issues/1209). But as this can reveal more information about the files than the user is comfortable with let's handle that on Picard side.

# Solution

Set the `file_name` based on File.base_filename

# Action
If we choose to allow submission without saving (just matched files) then we would need to override the entire `.metadata.tags` dict anyway.

### DIFF
--- a/picard/acousticbrainz/__init__.py
+++ b/picard/acousticbrainz/__init__.py
@@ -256,6 +256,13 @@ def ab_submit_features(tagger, file):
         submit_features_callback(file, None, None, True)
         return
 
+    # Set the filename. Especially on Windows the feature extractor fails to set the file basename only.
+    # As Picard already handles this just add the correct data ourselves.
+    try:
+        features['metadata']['tags']['file_name'] = file.base_filename
+    except KeyError:
+        log.warning('AcousticBrainz: Failed to set file_name in features data')
+
     # Submit features to the server
     tagger.webservice.post(
         host=ACOUSTICBRAINZ_HOST,


picard/acousticbrainz/__init__.py
```diff
@@ -256,6 +256,13 @@ def ab_submit_features(tagger, file):
         submit_features_callback(file, None, None, True)
         return
 
+    # Set the filename. Especially on Windows the feature extracture fails to set the file basename only.
```
zas, line 259: "extracture" ?

picard/acousticbrainz/__init__.py
```diff
@@ -256,6 +256,13 @@ def ab_submit_features(tagger, file):
...
+    # As Picard already handles this just add the correct data ourselves.
+    try:
+        features['metadata']['tags']['file_name'] = file.base_filename
+    except KeyError:
+        log.warning('AcousticBrainz: Failed to set file_name in features data')
```
zas, line 264: Can you clarify in which case(s) this exception is triggered?

When it happens, does it still make sense to post data?

phw: It would if the structure returned by the extractor did not include metadata or tags. This really is just to handle this exception in case of invalid data.

I think if the data is still acceptable without this is something we can have the AB server decide. The thinking is that both streaming_extractor_music and AB server work together. So user might exchange streaming extractor music to a newer version, which maybe has different data. No need for the feature in Picard to break and crash just because of that.

picard/acousticbrainz/__init__.py
phw: Fixed it

zas (approved):

PR #2018, opened 2021-12-28
Rework AcoustIdAPIHelper and add few tests

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/webservice/api_helpers.py
+++ b/picard/webservice/api_helpers.py
@@ -261,29 +261,42 @@ class MBAPIHelper(APIHelper):
 
 class AcoustIdAPIHelper(APIHelper):
 
+    acoustid_host = ACOUSTID_HOST
+    acoustid_port = ACOUSTID_PORT
+    client_key = ACOUSTID_KEY
+    client_version = PICARD_VERSION_STR
+
     def __init__(self, webservice):
-        super().__init__(ACOUSTID_HOST, ACOUSTID_PORT,
-                         '/v2/', webservice)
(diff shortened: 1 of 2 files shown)


picard/webservice/api_helpers.py
```diff
@@ -261,29 +261,42 @@ def delete_from_collection(self, collection_id, releases, handler):
...
     def __init__(self, webservice):
-        super().__init__(ACOUSTID_HOST, ACOUSTID_PORT,
-                         '/v2/', webservice)
+        super().__init__(
+            self.acoustid_host, self.acoustid_port, '/v2/', webservice
+        )
 
     @staticmethod
-    def _encode_acoustid_args(args, format_='json'):
+    def _static_encode_acoustid_args(args, client_key, client_version, format_):
```
phw, line 275: Should we really have this additional static method? The testes don't run `_encode_acoustid_args` currently, but I think they should and they easily could by creating a `AcoustIdAPIHelper`  with custom values for the host, port etc. And if we run `_encode_acoustid_args` in tests there is not need for `_static_encode_acoustid_args`  anymore.

zas: > there is not need for _static_encode_acoustid_args anymore

That was my plan, it was there to ease the transition. See 728b110bac42210fed52a0570782192fb7cccb4f

phw (approved):

PR #2019, opened 2021-12-28
Ensure rating submission XML is valid

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

MBAPIHelper._xml_ratings does not handle encoding the input parameters.

# Solution
Properly XML encode provided data for rating submission. Also ensure rating is an integer as expected.

In the current implementation both should already be guaranteed. `track.id`  is always a MBID as given by MB service, and the rating widget's `self._rating`  is always an integer. But for a complete implementation in the context of `MBAPIHelper`  and to avoid potential issues both should be handled.

# Action

### DIFF
--- a/picard/ui/ratingwidget.py
+++ b/picard/ui/ratingwidget.py
@@ -29,6 +29,7 @@ from PyQt5 import (
     QtWidgets,
 )
 
+from picard import log
 from picard.config import get_config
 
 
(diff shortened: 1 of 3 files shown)


zas (approved):
Good move.

phw:
@zas Do you think we should do anything about https://github.com/metabrainz/picard/pull/2019/checks?check_run_id=4652255922 ? I actually think we can just ignore it, using quoteattr to replace the common entities should really be enough for our use case of building this string I guess. If someone wants to attack the MB server with a bad XML they don't need to use Picard for that really.

zas:
> @zas Do you think we should do anything about https://github.com/metabrainz/picard/pull/2019/checks?check_run_id=4652255922 ? I actually think we can just ignore it, using quoteattr to replace the common entities should really be enough for our use case of building this string I guess. If someone wants to attack the MB server with a bad XML they don't need to use Picard for that really.

I agree we can just ignore it, but we need to silent the warning explicitly to make Codacy happy.

phw:
> but we need to silent the warning explicitly to make Codacy happy.

Did that in 1370fc5064b28dd3818e2f27778f166895a06a47

zas (approved):

PR #2027, opened 2022-01-06
PICARD-2390: Fixed display of "default listening port" label

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2390

With some localization and with UI scaling the label was only partially visible

# Solution

Make this use the same layout as we use for all other number input fields, with the input being on the very right.

While originally the idea was to not put so much whitespace between the label and the input, I could not make it work reliable. Hence this change, which is at least consistent with other similar options. If someone finds a solution for a different layout the makes the label and input look more connected we can do this.

Before with German localization on Linux:

grafik

With the fix:

grafik

Before on Windows 10 with font scaling 125%:

grafik

With the fix:

grafik


zas (approved):
Good fix.

rdswift:
I assume that this also closes PICARD-2272 ?

phw:
@rdswift Yes, I think so. I hadn't seen this issue, but looks like the same formatting issue with window scaling

PR #2028, opened 2022-01-06
Oauth code tidy up

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/oauth.py
+++ b/picard/oauth.py
@@ -155,11 +155,16 @@ class OAuthManager(object):
                 self.refresh_access_token(callback)
 
     def get_authorization_url(self, scopes):
-        params = {"response_type": "code", "client_id":
-                  MUSICBRAINZ_OAUTH_CLIENT_ID, "redirect_uri":
-                  "urn:ietf:wg:oauth:2.0:oob", "scope": scopes}
-        url = build_qurl(self.host, self.port, path="/oauth2/authorize",
-                         queryargs=params)
+        params = {
+            "response_type": "code",
+            "client_id": MUSICBRAINZ_OAUTH_CLIENT_ID,
+            "redirect_uri": "urn:ietf:wg:oauth:2.0:oob",
+            "scope": scopes,
+        }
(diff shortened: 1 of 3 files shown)


picard/oauth.py
```diff
@@ -172,28 +177,37 @@ def set_access_token(self, access_token, expires_in):
...
+        log.debug("OAuth: refreshing access_token with a refresh_token %s", self.refresh_token)
+        path = "/oauth2/token"
+        params = {
+            "grant_type": "refresh_token",
+            "refresh_token": self.refresh_token,
+            "client_id": MUSICBRAINZ_OAUTH_CLIENT_ID,
+            "client_secret": MUSICBRAINZ_OAUTH_CLIENT_SECRET,
+        }
+        self.webservice.post(
+            self.host, self.port, path, self._query_data(params),
```
phw, line 199: Instead of `self._query_data` we could also use `urllib.parse.urlencode` and get rid of the custom Qt based implementation.

phw (changes requested):

picard/oauth.py
zas: Yup, agreed.

See d1de2a7208 and 8b787482f

test/test_oauth.py
```diff
@@ -0,0 +1,36 @@
...
+from picard.oauth import OAuthManager
+
+
+class OAuthManagerTest(PicardTestCase):
+
+    def test_query_data(self):
+        params = {
+            'a&b': 'a b',
+            'c d': 'c&d',
+            '': '',
```
phw, line 33: I'd add some test entries which include `=` here, to make sure that gets encoded as well

phw (approved):

test/test_oauth.py
zas: Done in 53ec0e0

phw (approved):

PR #2029, opened 2022-01-06
PICARD-2392: Allow empty strings in multi-value variables

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**:  Allow empty strings in multi-value variables.

# Problem

* JIRA ticket (_optional_): PICARD-2392

Picard currently does not allow keeping empty strings in a multi-value variable, and automatically strips them out.  This can lead to problems for example if the user is trying to manage a list of artist join phrases, where the resulting multi-value may have elements removed so that it no longer matches the metadata.  This is the case with the changes to the ***AdditionalArtistsVariables*** plugin in https://github.com/metabrainz/picard-plugins/pull/320 requiring a non-blank placeholder to be substituted which would have to be removed by the user when processing the variable in a script.

# Solution

The change includes the following:

- Keep empty strings in multi-value lists
- Add scripting function to remove empty strings from multi-value
- Update tests

In addition, the criteria for removing a variable (marking it as deleted in the metadata) is changed to only if the variable has zero or one element, and the element is an empty string.

# Action

Please review this to ensure that this new behavior is something that should actually be revised, and that it doesn't accidentally cause other problems not checked in the tests.

If approved, this could be held until the Picard 2.8 release to avoid additional translation requirements.

If approved, the documentation will need to be updated to include the new script function.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -433,8 +433,9 @@ class Metadata(MutableMapping):
         name = self.normalize_tag(name)
         if isinstance(values, str) or not isinstance(values, Iterable):
             values = [values]
-        values = [str(value) for value in values if value or value == 0]
-        if values:
+        values = [str(value) for value in values if value or value == 0 or value == '']
+        # Remove if there is only a single empty or blank element.
+        if values and (len(values) > 1 or values[0]):
             self._store[name] = values
             self.deleted_tags.discard(name)
         elif name in self._store:
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1503,8 +1503,8 @@ Removes all empty string elements from the multi-value variable.
 
 Example:
 
-    $$setmulti(test,one; ; two; three)
-    $cleanmulti(%test%)
+    $setmulti(test,one; ; two; three)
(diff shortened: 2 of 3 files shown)


phw (dismissed):
Makes sense to me and I guess it should work, but not 100% sure we don't have any unexpected side effect. So definitely I'd put this into 2.8.

zas (dismissed):

picard/metadata.py
```diff
@@ -433,8 +433,9 @@ def set(self, name, values):
         name = self.normalize_tag(name)
         if isinstance(values, str) or not isinstance(values, Iterable):
             values = [values]
-        values = [str(value) for value in values if value or value == 0]
-        if values:
+        values = [str(value) for value in values if value or value == 0 or value == '']
+        # Remove if there is only a single empty or blank element.
+        if values and (len(values) > 1 or values[0]):
```
phw, line 438: I think this should be fine, I can't think of no side effect right now.

phw (approved):

PR #2032, opened 2022-01-07
Add `$appendmulti()` function

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds a new `$appendmulti()` function.

# Problem

Picard does not currently provide an easy way to append an element to a multi-value variable.  A potential use case is adding "My collection" to `%genres%`.

# Solution

Add a new `$appendmulti()` function.

# Action

If approved, do not merge until Picard v2.8 so as not to create additional translation work.

If approved, the documentation will need to be updated to add the new function.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1479,6 +1479,25 @@ def func_dateformat(parser, date_to_parse, date_format=None, date_order='ymd'):
 
 
 @script_function(eval_args=False, documentation=N_(
+    """`$appendmulti(name,value[,...])`
+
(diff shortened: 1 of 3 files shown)


rdswift:
After giving this a bit more thought, this function isn't necessary.  Items can be easily appended using:

```
$setmulti(foo,one; two)
$setmulti(foo,%foo%; three)
```

Unless anyone can provide a good reason for keeping this, I'll cancel the PR.

PR #2036, opened 2022-01-16
PICARD-2398: Load recording relationships separately for huge releases

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2398

The MB API for performance reasons does not include recording level relationships if an release contains too many tracks. This leads to missing data if the user has enabled track relationships.

# Solution

Detect the case of missing recording relationships and issue separate requests to load the recordings for a release. Since these requests are limited to 100 results multiple requests with offsets need to be done.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Hendrik van Antwerpen
 # Copyright (C) 2008 ojnkpjg
-# Copyright (C) 2008-2011, 2014, 2018-2021 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014, 2018-2022 Philipp Wolfer
 # Copyright (C) 2009 Nikolai Prokoschenko
 # Copyright (C) 2011-2012 Chad Wilson
 # Copyright (C) 2011-2013, 2019 Michael Wiencek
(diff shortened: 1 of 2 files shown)


picard/album.py
```diff
@@ -285,6 +303,42 @@ def _release_request_finished(self, document, http, error):
...
+        )
+
+    def _recordings_request_finished(self, document, http, error):
+        self._requests -= 1
+        if error:
+            self.error_append(http.errorString())
+            self._finalize_loading(error)
+        else:
+            for recording in document.get('recordings', []):
+                self._recordings_map[recording.get('id')] = recording
```
zas, line 333: If `recording.get('id')` is ensured to always work, we can write `recording['id']` and if it doesn't, I'm not sure we want to store the value for the `None` key. Since those come the web service, I'd rather ensure we have an `id`:

```python
for recording in document.get('recordings', []):
    recording_id = recording.get('id')
    if recording_id:
        self._recordings_map[recording_id] = recording
```

What do you think?

zas (changes requested):

picard/album.py
```diff
@@ -285,6 +303,42 @@ def _release_request_finished(self, document, http, error):
...
+            'recording-rels',
+            'release-rels',
+            'url-rels',
+            'work-rels'
+        }
+        log.debug('Loading recording relationships for %r (offset=%i, limit=%i)' % (self, offset, limit))
+        self._requests += 1
+        self.load_task = self.tagger.mb_api.browse_recordings(
+            self._recordings_request_finished,
+            inc=tuple(inc),
```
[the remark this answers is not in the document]
phw: Changed

picard/album.py
phw: Yes, I changed this and rebased

zas (approved):
LGTM

PR #2038, opened 2022-01-16
AB tests: Log debug message on extractor setup

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem

Attempt to capture debug output for spurious test failures in AB extractor setup. See e.g. test run at https://github.com/metabrainz/picard/runs/4831702643?check_suite_focus=true

PR #2041, opened 2022-01-19
Set log level in tests to DEBUG

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Log output was not shown in failing tests, making some issues hard to figure out. See https://github.com/metabrainz/picard/pull/2038 for background.

# Solution

Lower log level for tests to `DEBUG`. This allows pytest to capture the log output and display it in context of the test.

### DIFF
--- a/test/formats/test_vorbis.py
+++ b/test/formats/test_vorbis.py
@@ -21,7 +21,6 @@
 
 
 import base64
-import logging
 import os
 
 from mutagen.flac import (
@@ -35,10 +34,7 @@ from test.picardtestcase import (
     create_fake_png,
 )
 
-from picard import (
-    config,
-    log,
-)
+from picard import config
 from picard.coverart.image import CoverArtImage
 from picard.formats import vorbis
 from picard.formats.util import open_ as open_format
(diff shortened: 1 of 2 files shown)


phw:
@zas I was not sure about the `loglevel` context manager I added in  #2038. I though we'd need it in some cases, but after looking through it I don't see any case where we would. IMHO it is better to log in debug mode to have in case of failing tests the full details. And as long as tests are green the output is hidden anyway.

The context manager was a nice solution, but if we don't use it we should not add it to the code :)

phw:
It's actually nice that the tests failed with the `check_extractor_version`  timeout issue here, because this demonstrates how this change is useful. See the pytest output in

https://github.com/metabrainz/picard/runs/4864203315?check_suite_focus=true

For a successful run pytest is clean, no log output that pollutes the actual output of pytest. But for the failing test the log output is included cleanly as part of the failure report. If we had had this earlier it would have been easy to identify this `check_extractor_version` timeout issue. Without this logging we were poking in the dark.

zas (approved):

PR #2042, opened 2022-01-19
Update dependencies for macOS and Windows builds

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

Update the dependencies for the macOS and Windows binary builds to stay up-to-date.

For the pip dependencies I am running with latest versions locally for some while now without issue.

For macOS let's stay with Python 3.9 for now, we can consider upgrading to 3.10 with Picard 3. Python 3.10 should work in general, but it causes some deprecation warnings which we can not fully address when supporting Python < 3.10.

Also macOS builds need some testing with latest pyobjc on macOS 10.12.

For Windows we stick with Python 3.8 for now to maintain Windows 7 compatibility. Again this is something we can change in Picard 3.


phw:
Tested on macOS 10.12, 12.1 and Windows 10 / 11

zas (approved):

PR #2043, opened 2022-01-20
Delete CODE_OF_CONDUCT.md and use org default

reosarevok:
This seems like an outdated copy of the org's CoC that adds nothing.
Removing so this uses the default .github repo one that is kept up to date.
Feel free to reject this PR if you want to keep a separate one for some reason, rather than this being a historical artifact :)


phw (approved):
Thanks, makes sense. I don't think there is an actual reason, just as you said a historical artifact :)

PR #2044, opened 2022-01-20
PICARD-2398: Ensure recording relationships are available to album metadata processors

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2398

This is an extension to #2036, which handled the loading of recording relationships for large releases, where the MB API omits the per recording relationships for performance reasons.

In the previous commit these separately loaded relationships where not available to plugins registering an album metadata processor, because those were run before the relationship loading. This PR fixes this.

# Solution
Call `run_album_metadata_processors` after the web service calls to load the album's recordings.

This requires that the separately loaded recording relationships need to get merged into the loaded `self._release_node`  earlier, handling this merging in `_finalize_loading_track` is too late.

It is also important to keep the order of running the album metadata processors first, and only running the track metadata processors after the album processors have finished.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -191,6 +191,12 @@ class Album(DataObject, Item):
         """Returns the list of album artists (as AlbumArtist objects)"""
         return self._album_artists
 
+    def _run_album_metadata_processors(self):
+        try:
+            run_album_metadata_processors(self, self._new_metadata, self._release_node)
+        except BaseException:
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2045, opened 2022-01-22
PICARD-2396: Do not link AcoustIDs to recordings if there is a large difference in length

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2396

Picard currently submits the AcoustID fingerprint in combination with a recording MBID, to link the AcoustID and recording. This submission currently makes no further validation of the data.

The discussion at https://community.metabrainz.org/t/report-showing-acoustids-likely-to-be-bad-link-to-musicbrainz-recordings/551047 has identified that recordings linked to AcoustIDs with clearly wrong durations are a clear indication of wrong matching. This is something Picard can easily detect.

# Solution

If the duration difference between the MB recording and the fingerprint is larger than 30 seconds, than do not submit the MBID. Still submit the AcoustID fingerprint, but combine it with textual metadata only.

30 seconds where chose because this is also what the AcoustID server uses when considering if two fingerprints could be for the same recording (see https://github.com/acoustid/acoustid-server/blob/master/acoustid/const.py#L21)

# Action

### DIFF
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2017 Sambhav Kothari
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2018, 2020-2021 Laurent Monin
-# Copyright (C) 2020 Philipp Wolfer
+# Copyright (C) 2020, 2022 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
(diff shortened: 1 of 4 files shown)


phw:
This is a bit difficult to test. Testing the exact behavior would cause most of the time an intentional wrong submission. Hence I tested it with code patched so `Submission.valid_duration`  was always false, thus triggering the fallback code of metadata submission. Then I submitted correctly matched recordings for https://acoustid.org/track/b5123596-4a8d-43f3-98b7-3cdf8ba2ec39 and checked the results of the metadata submission.

picard/acoustid/manager.py
```diff
@@ -31,27 +31,64 @@
...
 
+    @property
+    def valid_duration(self):
+        return self.metadata is None or abs(self.duration * 1000 - self.metadata.length) <= FINGERPRINT_MAX_ALLOWED_LENGTH_DIFF_MS
+
+    @property
     def is_submitted(self):
         return not self.recordingid or self.orig_recordingid == self.recordingid
 
+    def get_args(self):
```
zas, line 67: Maybe, make this a property:

```python
def args(self):
```

zas (dismissed):
LGTM

picard/acoustid/manager.py
phw: Yes, agreed. I made it a property now

zas (approved):

PR #2046, opened 2022-01-23
PICARD-2292:  Fix setting "no lyrics" for recording with mixed instrumental works

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2292

Since PICARD-1117 the language is set to "zxx" (no lyrics) for all recordings being linked to a work with the "instrumental" flag. Also in this case the lyricist tag is being cleared.

But if a recording is linked to multiple works, e.g. a medley like https://musicbrainz.org/recording/91b7e04e-529b-4119-bc30-867db56fd400, then this should only happen for the specific work relations. Currently a single instrumental link to a recording unsets the languages and lyricists for all other work relationships.

# Solution

If a work relationship is instrumental then add the language "zxx" and only omit the lyricist for this work. All other languages and lyricist for non instrumental performances are kept.

Also added specific test case for this situation.

# Action

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -138,7 +138,7 @@ def _parse_attributes(attrs, reltype, attr_credits):
     return ' '.join([prefix, result]).strip()
 
 
-def _relations_to_metadata(relations, m):
+def _relations_to_metadata(relations, m, instrumental=False):
     config = get_config()
     use_credited_as = not config.setting['standardize_artists']
     use_instrument_credits = not config.setting['standardize_instruments']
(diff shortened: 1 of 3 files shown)


picard/mbjson.py
```diff
@@ -441,20 +443,19 @@ def recording_to_metadata(node, m, track=None):
...
-        m.unset('lyricist')
-        m['language'] = 'zxx'
 
 
-def performance_to_metadata(relation, m):
+def get_relation_attributes(relation):
     if 'attributes' in relation:
-        for attribute in relation['attributes']:
-            m.add_unique("~performance_attributes", attribute)
+        return [a for a in relation['attributes']]
```
zas, line 450: Could we return a set instead? Or does order matter?

zas (approved):

picard/mbjson.py
phw: Probably yes, but I did not want to change that here. It'll obviously mess with order of these attributes in the tags, but the order is kind of undefined anyway. Maybe we should make it a set, but then sort them alphabetically in the tags? That needs an additional processing after all has been parsed, so better do this in a separate PR.

PR #2048, opened 2022-01-24
PICARD-1455: Allow disc lookup from EAC / XLD log file

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1455

The popular CD ripping tools EAC (Windows) and XLD (macOS) can generate a log file of the ripping process. This log file contains the complete information from the CD TOC.

Adding the ability to Picard to perform a disc ID lookup based on these log files will allow users of these tools to do a disc lookup without the need for putting the physical CD again into their drive.

# Solution

This adds a submenu "From EAC / XLD log file..." to the CD drive selection sub menu:

grafik

Upon selecting it the user can choose a log file:

grafik

The log file gets parsed and the calculated disc ID gets processed normally, showing the CD lookup result window.

Examples of both a XLD and an EAC log file are part of the test data.

# Action

There could be other formats for which a disc ID calculation could be interesting, e.g. CD image files or CUE sheets. One way to extend this would be to add an extension point plugins could use to add additional disc ID extractors.

I'd like to do this on basis of this PR. But first I'd like to have the basic functionality of being able to parse the XLD / EAC logs, as both are very common and provide good quality output.


rdswift:
Nice feature.  When this is merged, we'll need to remember to update the documentation accordingly.

picard/discid/eaclog.py
```diff
@@ -0,0 +1,96 @@
...
+    for line in lines:
+        # to allow internationalized EAC output where column headings
+        # may differ
+        if re.match(r""" \s*
+                   .+\s+ \| (?#track)
+                \s+.+\s+ \| (?#start)
+                \s+.+\s+ \| (?#length)
+                \s+.+\s+ \| (?#start sec)
+                \s+.+\s*$   (?#end sec)
+                """, line, re.X):
```
zas, line 48: Since that's in a loop, it would be better to compile the regex before entering the loop.

picard/discid/eaclog.py
```diff
@@ -0,0 +1,96 @@
...
+            \s*\|\s*
+            (?P<start_time>[0-9:.]+)
+            \s*\|\s*
+            (?P<length_time>[0-9:.]+)
+            \s*\|\s*
+            (?P<start_sector>\d+)
+            \s*\|\s*
+            (?P<end_sector>\d+)
+            \s*$
+            """, line, re.X)
```
zas, line 66: Since that's in a loop, it would be better to compile the regex before entering the loop.

zas (changes requested):

picard/discid/eaclog.py
phw: Done in 997d16caf316f90c2432544eaa05e58b8d0add55

zas (approved):

thwaller:
That is a great way to incorporate it!

PR #2049, opened 2022-01-24
PICARD-2407: Send tport with Submit cluster as release

jesus2099:
PICARD-2407, MBS-12179

When you Add cluster as release, the tport is lost on MBS and Green Tagger buttons disappear.

My code here does certainly not compile.
I need help finalising this PR.
I know nothing about Python.
I saw `self.local_port` from here https://github.com/metabrainz/picard/blob/14ce07756eb101d62100fb6ab8800ae644c372a2/picard/browser/filelookup.py#L77
But it certainly does not work in this context, here, unfortunately.

However, I did some tests on how to send tport=8000:

It does not work when it is sent with the other `POST` parameters.

> <ins>[action="/release/add" + POST parameters include tport=8000](http://jesus2099.gitlab.io/forward-request.html?_method=post&_action=https%3A%2F%2Fmusicbrainz.org%2Frelease%2Fadd&tport=8000&name=Tutu&artist_credit.names.0.artist.name=Toto) (**URL fixed**)</ins>
> NG: Green tagger buttons disappear, even if they were there before adding the cluster as release
> _Unknown field: tport_

It works when added in the `action` URL, even if it is ugly to mix `GET` and `POST` like this. :)

> [action="/release/add?tport=8000" + POST parameters](http://jesus2099.gitlab.io/forward-request.html?_method=post&_action=https%3A%2F%2Fmusicbrainz.org%2Frelease%2Fadd%3Ftport%3D8000&name=Tutu&artist_credit.names.0.artist.name=Toto)
> OK: Green tagger buttons appear

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/browser/addrelease.py
+++ b/picard/browser/addrelease.py
@@ -89,19 +89,21 @@ def serve_form(token):
     try:
         payload = jwt.decode(token, __key, algorithms=__algorithm)
         log.debug('received JWT token %r', payload)
+        tagger = QCoreApplication.instance()
+        tport = tagger.browser_integration.port
         if 'cluster' in payload:
-            cluster = _find_cluster(payload['cluster'])
+            cluster = _find_cluster(tagger, payload['cluster'])
             if not cluster:
                 raise NotFoundError('Cluster not found')
-            return _get_cluster_form(cluster)
+            return _get_cluster_form(cluster, tport)
         elif 'file' in payload:
-            file = _find_file(payload['file'])
+            file = _find_file(tagger, payload['file'])
             if not file:
                 raise NotFoundError('File not found')
             if payload.get('as_release', False):
-                return _get_file_as_release_form(file)
+                return _get_file_as_release_form(file, tport)
             else:
-                return _get_file_as_recording_form(file)
+                return _get_file_as_recording_form(file, tport)
         else:
             raise InvalidTokenError
     except jwt.exceptions.InvalidTokenError:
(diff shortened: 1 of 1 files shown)


phw:
Yes, it is a bit tricky at this point. You need to get the port from the browser integration, which you can get from the main application instance.

The port can be retrieved via:

```
QCoreApplication.instance().browser_integration.port
```

jesus2099:
I have this behaviour with Vivaldi (Chrome):

image

So it's _normal_, or shared by everyone, at least.

jesus2099:
I changed the file accordingly.
But @mwiencek or @reosarevok, could you tell if this should not be rather fixed on MBS side?
MBS is forgetting `tport` on Picard _Submit cluster as release…_.

jesus2099:
I opened a ticket: https://tickets.metabrainz.org/browse/MBS-12179

mwiencek:
On the website tport is remembered by storing it in the user session, but the session cookie uses `SameSite=Lax`, so browsers don't allow access to it in this case.  This is also why you appear to be logged out initially.  Since it doesn't find any existing session, it creates a new one (with no tport set).  I'm not sure if this can be easily worked around in musicbrainz-server...

Interestingly, when I used the the add-cluster plugin (FF 96 on Linux) I get a login screen instead of "Confirm Form Submission."  It must be sending an "Origin" header in your case for you to be getting that screen?

phw:
@jesus2099 Thanks for the update. I tested it, but unfortunately it does not work this way :) The "action" parameter is only part of what is being used to build the final URL, and it is treated as the path. Hence the question mark gets url encoded. The query parameters instead need to be passed as separate argument. I did a fix in https://github.com/jesus2099/metabrainz_picard/pull/1 , please check and merge that into your branch, then this PR should show the changes.

I also added PICARD-2407 to track the Picard side of this (we need this to generate the changelog)

@mwiencek 
> I'm not sure if this can be easily worked around in musicbrainz-server...

Not sure either. But interesting is:

1. it does not loose the login session once you proceed, only the tport is lost
2. it does not loose the tport either when using "add file as standalone recording" via `https://musicbrainz.org/recording/create`

Hence I suspect an issue either specific to `/release/add`.

> Interestingly, when I used the the add-cluster plugin (FF 96 on Linux) I get a login screen instead of "Confirm Form Submission." It must be sending an "Origin" header in your case for you to be getting that screen?

That's the main difference between the addrelease plugin and the built-in functionality in Picard 2.7, and one of the reasons why we did this move. Instead of relying on the HTML form being loaded from the file system, the built-in add release functionality utilizes Picard's built-in web server.

zas (approved):

phw (approved):

mwiencek:
> it does not loose the login session once you proceed, only the tport is lost

That's due to the `remember_me` cookie, which will log you in even if your session expires (or in this case, is lost).  But it only works if you tick "remember me" on the login page.

> it does not loose the tport either when using "add file as standalone recording" via https://musicbrainz.org/recording/create

Odd, it does get lost for me with /recording/create too (tried in production and my dev server).  Does your `musicbrainz_server_session` cookie not change?

> That's the main difference between the addrelease plugin and the built-in functionality in Picard 2.7

I didn't notice it was built-in now.  Thanks!  Now I get the same behavior there.

jesus2099:
Oh I did not notice the same issue with `/recording/create` but @phw has fixed it as well: https://github.com/metabrainz/picard/commit/27c454fb34f234b59d60bde9b673b3afc630e552#diff-0ce8922143bd631ec8c94ec721b4b1692be104f506e6df0057058dea5cb6598dR162-R169 !

phw:
I see, so without remember me this is broken.

> Odd, it does get lost for me with /recording/create too (tried in production and my dev server). Does your `musicbrainz_server_session` cookie not change?

Strange, I tested this several times (without the fix), and after the "add as recording" my tagger buttons where still intact, while they always got lost for "add as release". I'll test again. But looking at the details of the issue with the cookie behavior the expectation would be indeed that this case is exactly the same.

jesus2099:
I was not very clear in [my previous comment](#issuecomment-1022293117) (now edited), but for me, the tagger links were lost with `/recording/create` too.

PR #2051, opened 2022-01-26
Various documentation's hyperlink updates

jesus2099:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

- Update doc broken hyperlinks to cached versions
- Update HTTP hyperlinks to HTTPS when possible

# Action

Do you have a better link than https://web.archive.org/web/20210412222406/download.microsoft.com/download/A/5/4/A54BADB6-9C3F-478D-8657-93B3FC9FE62D/vcsetup.exe for **Visual C++ 2008 Express**?


INSTALL.md
```diff
@@ -8,23 +8,23 @@ Before installing Picard from source, you need to check you have the following d
...
+* [Python 3.6 or newer](https://python.org/download)
+* [PyQt 5.11 or newer](https://riverbankcomputing.com/software/pyqt/download)
 * [Mutagen 1.37 or newer](https://mutagen.readthedocs.io/)
 * [PyYAML 5.1 or newer](https://pyyaml.org/)
 * [python-dateutil](https://dateutil.readthedocs.io/en/stable/)
 * gettext:
   * [Windows](https://mlocati.github.io/articles/gettext-iconv-windows.html)
 * a compiler
-  * Windows should work with [Visual C++ 2008 Express](http://go.microsoft.com/?linkid=7729279)
+  * Windows should work with [Visual C++ 2008 Express](https://web.archive.org/web/20210412222406/download.microsoft.com/download/A/5/4/A54BADB6-9C3F-478D-8657-93B3FC9FE62D/vcsetup.exe)
```
phw, line 19: We should probably rather link to a newer version. I'd not really recommend anyone to install VS 2008 today, and I can't tell if this works. Visual Studio 2019 Community would do, that's what I have on my Windows 10 test VM. I guess the new VS Community 2022 will do as well. Maybe we should just link to https://visualstudio.microsoft.com/de/downloads/

A more minimal install is actually just installing the Visual Studio Build Tools (also available from same . Those install just the compilers, build tools and other essential stuff to be able to compile software. But not the Visual Studio IDE.

jesus2099: So I put this instead? :)

```diff
- [Visual C++ 2008 Express](http://go.microsoft.com/?linkid=7729279)
+ [Visual Studio](https://visualstudio.microsoft.com/downloads/)
```

phw: @jesus2099 Actually let me check. This depends on the Python version being used, and I think for the recent Python releases we support it should be Visual Studio 2019. But I don't see a nice download page for that

phw: There is really no nice download page for this. This gives download directions for older versions:

https://visualstudio.microsoft.com/vs/older-downloads/

But all kind of stuff there. Clicking the downloads link for Visual Studio 2019 gives again a list of downloads for various editions in different versions.

Maybe do it as before and provide a direct download link to the exe?

```diff
- [Visual C++ 2008 Express](http://go.microsoft.com/?linkid=7729279)
+ [Visual Studio Community 2019](https://aka.ms/vs/16/release/vs_community.exe)
```

At least we know this version works and people get the correct thing when following the link.

jesus2099: Thanks again @phw, I did this! :)

setup.py
```diff
@@ -517,7 +517,7 @@ def versiontuple(v):
         return tuple(map(int, (v.split("."))))
 
     # input_dirs are incorrectly handled in babel versions < 1.0
-    # http://babel.edgewall.org/ticket/232
+    # https://web.archive.org/web/20150910064954/babel.edgewall.org/ticket/232
```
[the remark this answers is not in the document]
jesus2099: Yes, I don't know Python so I prefer keeping this comment as long as the `input_dirs_workaround` is kept on the following line. 😅

phw (approved):

PR #2052, opened 2022-01-26
Remove legacy genre tag translation

atj:
The translation mapping was originally added as part of the last.fm
plugin in commit 99dc3e54, which was later used as the basis of the
initial folksonomy genre implementation.

This change fixes a bug where a track tagged with the "electronic" and
"electronica" genres would end up with "electronic" twice.

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Screenshot_2022-01-26_17-10-06

# Solution

Removed the genre translation dict as per the commit message. I've been unable to find any reasons to explain why this behaviour was added in the first place - my assumption is that it was to match the Last.fm tag schema?

Note that this change will cause tag changes for all tracks with the following genres:

* hip hop
* synth-pop
* electronica

This could cause some user confusion but I don't see any other reasonable option...

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -82,13 +82,6 @@ from picard.util.textencoding import asciipunct
 from picard.ui.item import FileListItem
 
 
-_TRANSLATE_TAGS = {
-    "hip hop": "Hip-Hop",
-    "synth-pop": "Synthpop",
-    "electronica": "Electronic",
-}
-
-
 class TagGenreFilter:
 
     def __init__(self, filters):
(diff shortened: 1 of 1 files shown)


zas (approved):

phw (approved):
> This could cause some user confusion but I don't see any other reasonable option...

I think this is less confusion than this hard coded and very opinionated  conversion can cause :)

atj:
Thanks!

PR #2054, opened 2022-01-27
PICARD-2409: Search for cdtoc URL triggers CD lookup

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2409

Searching for or drag and dropping a disc ID URL (such as https://musicbrainz.org/cdtoc/vtlGcbJUaP_IFdBUC10NGIhu2E0-) should trigger the CD lookup dialog

# Solution

Handle this URL and show the CD lookup dialog.

One difference to a regular disc ID lookup is, that submission does not work. In this case we only have the disc ID, e.g. 
"vtlGcbJUaP_IFdBUC10NGIhu2E0-". For submission also the full TOC details would be needed. Hence the CD lookup dialog hides the submission buttons.

Also added tests for this and other untested methods in `FileLookup`.

### DIFF
--- a/picard/browser/filelookup.py
+++ b/picard/browser/filelookup.py
@@ -51,7 +51,7 @@ from picard.ui.searchdialog.album import AlbumSearchDialog
 class FileLookup(object):
 
     RE_MB_ENTITY = re.compile(r"""
-        \b(?P<entity>release-group|release|recording|work|artist|label|url|area|track)?
+        \b(?P<entity>area|artist|instrument|label|place|recording|release|release-group|series|track|url|work)?
         \W*(?P<mbid>[a-f0-9]{8}(?:-[a-f0-9]{4}){3}-[a-f0-9]{12})
     """, re.VERBOSE | re.IGNORECASE)
 
--- a/picard/browser/filelookup.py
+++ b/picard/browser/filelookup.py
@@ -49,6 +49,11 @@ from picard.ui.searchdialog.album import AlbumSearchDialog
 
 class FileLookup(object):
 
+    RE_MB_ENTITY = re.compile(r"""
+        \b(?P<entity>release-group|release|recording|work|artist|label|url|area|track)?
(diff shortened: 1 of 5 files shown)


picard/browser/filelookup.py
```diff
@@ -49,6 +50,16 @@
...
 class FileLookup(object):
 
+    RE_MB_ENTITY = re.compile(r"""
+        \b(?P<entity>area|artist|instrument|label|place|recording|release|release-group|series|track|url|work)?
+        \W*(?P<mbid>[a-f0-9]{8}(?:-[a-f0-9]{4}){3}-[a-f0-9]{12})
+    """, re.VERBOSE | re.IGNORECASE)
+
+    RE_MB_CDTOC = re.compile(r"""
+        \b(?P<entity>cdtoc)
+        \W*(?P<mbid>[a-z0-9-_.]{28})
```
zas, line 60: This is misleading, since this part of the expression doesn't actually match a mbid (nor cdtoc is an actual entity but ok).

I think we should just use `discid` and change the code below.

```python
        if m is None:
            m = self.RE_MB_CDTOC.search(string)
            if m is None:
                return False
...
        if entity != 'cdtoc':
            mbid = m.group('mbid').lower()
          
...
        elif entity == 'cdtoc':
            discid = m.group('discid')
            disc = Disc(id=discid)
            disc.lookup()
```

zas (changes requested):

picard/browser/filelookup.py
phw: We could also rename the variable to `id`

phw: I renamed the variable to ID. The URLs are in both cases exceptionally the same, that way we don't have to deal with different code paths for reading the IDs later on.

zas (approved):

PR #2055, opened 2022-01-27
Clarify script types in strings in the UI files

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**: Clarify the script types in the UI files

# Problem

The word "script" is used in a few places in the Picard UI, but refers to different things which could lead to confusion.  For example, it could refer to a tagging script, file naming script, or the script used for a selected locale.

# Solution

Clearly identify the type of script (i.e.: "tagging script", "file naming script" and "language script") in the UI.  This should help clarify for the translations.

# Action

### DIFF
--- a/picard/ui/ui_exception_script_selector.py
+++ b/picard/ui/ui_exception_script_selector.py
@@ -102,11 +102,11 @@ class Ui_ExceptionScriptSelector(object):
 
     def retranslateUi(self, ExceptionScriptSelector):
         _translate = QtCore.QCoreApplication.translate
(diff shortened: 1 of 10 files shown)


zas (approved):

phw (approved):
Thanks for doing this. That should make it also much clearer for translators how to translate something.

@zas For your info: @rdswift and myself discussed the idea that if a string here would look too strange in the English UI there would still be the possibility to update the English .po files with a better string. Not sure if that is needed, but we have the option.

phw (approved):

PR #2057, opened 2022-01-30
Omit standard ports in URLs constructed by build_qurl

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

URLs constructed by `picard.util.build_qurl`  always contain the port, e.g. `https://musicbrainz.org:443/`. This is not necessary if the default ports 80 for http or 443 for https are being used.

# Solution

Skip setting the port if the default is being used. Only set the scheme.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -580,13 +580,14 @@ def build_qurl(host, port=80, path=None, queryargs=None):
     """
     url = QtCore.QUrl()
     url.setHost(host)
-    url.setPort(port)
 
-    if host in MUSICBRAINZ_SERVERS or port == 443:
+    if port == 443 or host in MUSICBRAINZ_SERVERS:
         url.setScheme("https")
-        url.setPort(443)
+    elif port == 80:
+        url.setScheme("http")
     else:
         url.setScheme("http")
+        url.setPort(port)
 
     if path is not None:
         url.setPath(path)
--- a/test/test_browser.py
+++ b/test/test_browser.py
@@ -50,8 +50,9 @@ class BrowserLookupTest(PicardTestCase):
 
     def assert_mb_url_matches(self, url, path, query_args=None):
         parsed_url = urlparse(url)
-        expected_host = "%s:%s" % (SERVER, PORT)
+        expected_host = SERVER
         self.assertEqual(expected_host, parsed_url.netloc, '"%s" hostname does not match "%s"' % (url, expected_host))
+        self.assertEqual('https' if PORT == 443 else 'http', parsed_url.scheme)
         self.assertEqual(path, parsed_url.path, '"%s" path does not match "%s"' % (url, path))
         if query_args is not None:
             actual_query_args = {k: v[0] for k, v in parse_qs(parsed_url.query).items()}
--- a/test/test_disc.py
+++ b/test/test_disc.py
@@ -39,7 +39,7 @@ class MockDisc:
     mcn = '5029343070452'
     tracks = list(range(0, 11))
     toc_string = ' '.join(str(i) for i in test_toc)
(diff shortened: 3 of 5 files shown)


Sophist-UK:
@phw Philipp - Just a reminder that there are other places in Picard where port is calculated e.g.
* https://github.com/metabrainz/picard/blob/486d0fd2dc55a5bc6b6b72a0a55121fa4f1c0865/picard/webservice/__init__.py#L457-L500
* https://github.com/metabrainz/picard/blob/486d0fd2dc55a5bc6b6b72a0a55121fa4f1c0865/picard/coverart/image.py#L169-L178
* https://github.com/metabrainz/picard/blob/486d0fd2dc55a5bc6b6b72a0a55121fa4f1c0865/picard/ui/coverartbox.py#L398-L412

There may be others...

phw:
@Sophist-UK these cases are the opposite: they take an URL and figure out the host and port in order to call the webservice API.

What is being changed here is the other way around: constructing the URL from host and port for performing than the actual request.

The way the webservice API works with host and port is really a bit awkward and we should refactor it. Rather it should be URL based, with the option to construct Webservice instances that work against a specific base URL so you only have to specify the path. Maybe we should take the chance in Picard 3 and have an updated webservice Implementation

Sophist-UK:
@phw Philipp The conceptual issue is the same - the port is being explicitly set when it could be implied from the http/https.

It seems to me that if an explicit port causes problems with one site, it might cause similar problems elsewhere and perhaps better therefore to make similar changes elsewhere in Picard.

phw:
> The conceptual issue is the same - the port is being explicitly set when it could be implied from the http/https.

No, it is the exact opposite operation. The core of the issue is the API of `picard.webservice.WebService`. Let's take as an example the definition of get method:

```
def get(self, host, port, path, handler, parse_response_type=DEFAULT_RESPONSE_PARSER_TYPE,
            priority=False, important=False, mblogin=False, cacheloadcontrol=None, refresh=False,
            queryargs=None):
    ...
```

There is no URL there and no scheme. What it gets is `host`, `port` and `path`. The WebService class then needs to construct a URL from it. Since there is no explicit mention of the scheme, it needs to derive this from what it got, which is the port. So based on the port it constructs a URL like `https://example.com:443`, `http://example.com:80` or `http://example.com:8000`.

Now 443 and 80 are the default ports for HTTPS and HTTP respectively. While setting them explicitly is not wrong, it is also not needed. And since at least the tango.info server chokes on that, and probably also others, let's omit it. That's what this PR is doing, and this applies to all requests being done through the WebService class.

Now the other examples you showed are on the other side, on the caller side of the WebService API. In order to make a query with e.g. `WebService.get` you need a host and port. But if you only have a URL you need to get the port from that. The examples all also use `QUrl` for deconstructing the URL. So you could do something like this:

```python
url = QUrl('https://example.com')
host = url.host()
port = url.port()
```

One could suspect that port is now set to 443. But it isn't, because the port is not explicitly set in the URL `QUrl.port()` actually returns `None`. So to get the correct port the code must be instead something like this:

```python
url = QUrl('https://example.com')
host = url.host()
if url.scheme == 'https':
    port = url.port(443)
else:
    port = url.port(80)
```

The values passed to `QUrl.port` are default values, which will be used if there is no explicit URL set on the host.

So no, these examples do not set the port explicitly but actually do derive it from http/https.

So at the core of the issue is the somewhat limitating API of the webservice module. But for fixing the issue at hand with having the default ports not being part of the URL this fix here does it, but the other locations need to keep their functionality.

zas (approved):

Sophist-UK:
@phw Ah - apologies.

PR #2059, opened 2022-02-05
PICARD-2411: Add option to fix seektable for flac files

zas:
This is to fix problem with Kodi which is unable to load files
that have empty SeekTable, so by default a single SeekPoint is added.
An error message is logged if the file has no samples.

# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2411

# Solution

# Action


phw (approved):
Clean patch, and as this is an existing issue with some files I see this in line with some other options, like removing ID3 tags from FLAC / APE

PR #2063, opened 2022-02-13
Fix crash on loading files with ACOUSTID_FINGERPRINT tag.

cybersphinx:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Fix crash introduced by https://github.com/metabrainz/picard/pull/2045 (PICARD-2396), specifically 62da04f831ca64e755ffdf76ac46956b4059c221.

# Problem

On loading some files, Picard crashes with the following message: 

```
  File "/home/user/picard/picard/acoustid/manager.py", line 53, in <genexpr>
    return int(sum((len(key) + len(value) + 2 for key, value in self.args.items())) * 1.03)
TypeError: object of type 'NoneType' has no len()
```

After some investigation, it seems that files with an ACOUSTID_FINGERPRINT tag cause this. To reproduce use e.g. this [testfile.zip](https://github.com/metabrainz/picard/files/8056446/testfile.zip) (public domain, see https://kimikoishizaka.bandcamp.com/album/bach-well-tempered-clavier-book-1) or take a random file with just basic metadata and add an ACOUSTID_FINGERPRINT tag (doesn't have to be a valid fingerprint.)

# Solution

A debug `print(self.args.items())` (https://github.com/cybersphinx/picard/commit/2b25df9b2432b0f49d3860a6fa4d47012a66b975) before the offending line shows that `'mbid'` is `None`. So this fix checks if `self.recordingid` is `None` to make sure `'mbid'` isn't empty.

Incidentally, testing this fix with the abovementioned debug output, for files without a fingerprint tag that are not in Musicbrainz (after fingerprint scanning) the items to submit now contain all the regular tags (artist, album, …) while before `'mbid'` would be an empty string and so no other metadata included.

(Also, this seems to be called quite often, for the testfile the debug output is printed four times, for the source file directly from Bandcamp after fingerprint scanning it shows six times.)

# Action

### DIFF
--- a/picard/acoustid/manager.py
+++ b/picard/acoustid/manager.py
@@ -76,7 +76,7 @@ class Submission(object):
             puid = metadata['musicip_puid']
             if puid:
                 args['puid'] = puid
-        if self.valid_duration:
+        if self.valid_duration and self.recordingid:
             args['mbid'] = self.recordingid
         elif metadata:
             args['track'] = metadata['title']


cybersphinx:
> Actualy I'm a bit puzzled by this. Submission without having the file matched to a recording ID should not even be possible.

I assumed it was for tracks without Musicbrainz entry and ends up as the "Additional user-submitted metadata" like e.g. here https://acoustid.org/track/7da4bc46-ba8c-49cd-aa80-8f175b177c42

> I could not reproduce this.

For example https://winterquilt.bandcamp.com/album/qilphot-original-d-mo-soundtrack is not on Musicbrainz, if I scan the first track to fingerprint it I get 
```
dict_items([('fingerprint', 'AQAD...'), ('duration', '210'), ('mbid', '')])
```
without my patch and 
```
dict_items([('fingerprint', 'AQAD...'), ('duration', '210'), ('track', 'As the Village Lays Aslumber …'), ('artist', 'winterquilt'), ('album', 'QILPHOT - ORIGINAL DÆMO SOUNDTRACK'), ('albumartist', 'winterquilt'), ('trackno', '1'), ('discno', ''), ('year', '2021')])
```
with it. (But I haven't tried submitting it, maybe that's what you meant.)

phw (approved):
Yes, submission is what I thought would have been possible. That currently should only work for matched files, not for unmatched ones.

But the behavior of `Submission.args`  on  that level is fine. If we would allow submission of unmatched files this is exactly how it would need to behave.

Anyway, thanks a lot for the fix here

PR #2065, opened 2022-02-16
PICARD-2421: Consider MB track ID when matching tagged files to albums

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2421

Improve the matching of files tagged with MBIDs to tracks of a recording.

Instead of only matching by recording ID also consider the track ID. In case of ambiguity use the track duration to find the better match. This gives better results if recordings for tracks got changed.

See discussion at https://community.metabrainz.org/t/picard-confused-when-recordingids-swap-around-in-a-release/572514/ for more details.

# Solution

When matching also include the `musicbrainz_trackid`  in the track cache and use duration for similarity score.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -650,12 +650,16 @@ class Album(DataObject, Item):
         def build_tracks_cache():
             for track in tracks:
                 tm_recordingid = track.orig_metadata['musicbrainz_recordingid']
+                tm_trackid = track.orig_metadata['musicbrainz_trackid']
                 tm_tracknumber = track.orig_metadata['tracknumber']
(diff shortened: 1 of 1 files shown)


phw:
I tested this also with the files provided by Ivan Dobsky. It leads to the expected results, matching the files to the proper track by track ID. Also if the files would have been originally tagged the opposite way this will still give the correct result by moving the files around to the matching recording ID.

zas (approved):
LGTM

PR #2066, opened 2022-02-16
PICARD-2422: Remove AcousticBrainz analysis and submission

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2422

See https://blog.metabrainz.org/2022/02/16/acousticbrainz-making-a-hard-decision-to-end-the-project/ for the background.

# Solution

Remove all AB related code changes


phw:
@Gabrielcarvfer Pinging you just for raising your attention to this, since you added the feature in the first place. Thanks a lot for this again. I'm sad to see this go. Hopefully Picard can be part of future plans.

Gabrielcarvfer:
It is a shame, but I kind of expected since I've already read a few posts from Alastair talking about the quality of the data. 
Facebook research has some interesting models to split instrument and voice tracks from audio, probably could be used to provide more meaningful information (https://github.com/facebookresearch/demucs).

rdswift (commented):
Huge change, but looks good to me.

For the documentation changes, can (should) we link the PR to the same ticket or create a new ticket?

phw:
I think we can use the same ticket. Regarding documentation I think we need a new ticket type Documentation instead of the component. That would make it easier to handle. Currently we can't properly add documentation tickets with a target release version without messing with the changelog :( I'll ask yvanzo if he can configure this

zas (approved):
That's a lot of code to remove, sad to see this feature disappear, but since AB is gone, nothing else to do.
The patch looks sane.

rdswift:
> That's a lot of code to remove, sad to see this feature disappear, but since AB is gone, nothing else to do.

True, but I suspect that there will need to be a similar amount of different code to add when there is a decision made regarding what type of functionality will be required to support the data collection system replacing AcousticBrainz.  :stuck_out_tongue_winking_eye:

PR #2067, opened 2022-02-17
Refactor similarity utils

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

The utility functions `sort_by_similarity` and `find_best_match` currently expect as parameter a callable that returns an iterable. I think it is a cleaner API if they accept an iterable directly.

# Solution

Make the functions work on the iterable, move calling of the functions to the outside.


zas (approved):
Much cleaner & pythonic, good move.

PR #2069, opened 2022-02-27
PICARD-2425: Handle possible exceptions with os.path.realpath

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2425

`os.path.realpath` can throw `OSError`  for different reasons. E.g. PICARD-2425 reports an issue with drives mounted via SSHFS-Win. In general on Windows this can cause exceptions if drives are mounted without mount manager. https://github.com/billziss-gh/sshfs-win/issues/243 has some more info about this.

Due to missing exception handling this can cause crashes when saving files or when generating filename previews in options.

# Solution

Handle `OSError`  on `os.path.realpath`  calls. If it makes sense use `picard.util.normpath`, which already handles this.

Note that we still log warnings in these cases, which I think makes sense.

# Action

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -78,6 +78,7 @@ from picard.util import (
     find_best_match,
     format_time,
     is_absolute_path,
+    normpath,
     thread,
     tracknum_and_title_from_filename,
 )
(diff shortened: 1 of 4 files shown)


zas (approved):
LGTM

PR #2073, opened 2022-03-06
PICARD-1570: Allow accessing files with long paths on Windows

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1570

On traditional Windows API the file path length is limited to 259 characters (+1 null character). Picard cannot read or write longer paths.

The Windows API however supports accessing longer paths, if the path gets prefixed with `\\?\`. This internally enables long path handling.

# Solution

Extend `picard.util.normpath`  to extend paths with the long path prefix `\\?\` on Windows if the path exceeds 259 characters in length. This enables both reading and writing to such files

# Action
Please note a few things about this:

1. This solution works independent of the new long path support that can be enabled in Windows 10. If that is enabled Picard already can handle reading and writing long paths. But this change allows this also if on Windows 10 / 11 long path handling is disabled (the default) or even on on Windows 7 / 8.
2. This change does not affect that Picard enforces the path limit when creating path and file names. This will be handled in a separate PR (PICARD-2076) and requires some changes to the options to make this work nicely for the user. But this change is a prerequisite for this, and it allows the user to both load files with long path and save tags to them.
3. There is a related issue that long paths are not seen by Picard with drag and drop, see PICARD-2429 . I have a patch for that (work in progress), and also enabling long path support on Windows 10 resolves this issue as well.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -588,7 +588,7 @@ class Tagger(QtWidgets.QApplication):
     def _scan_paths_recursive(paths, recursive, ignore_hidden):
         local_paths = list(paths)
         while local_paths:
-            current_path = local_paths.pop(0)
+            current_path = normpath(local_paths.pop(0))
             try:
                 if os.path.isdir(current_path):
                     for entry in os.scandir(current_path):
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -72,6 +72,17 @@ from picard.const.sys import (
 )
 
 
+# Windows path length constraints:
+# the entire path's length
+WIN_MAX_FILEPATH_LEN = 259
+# the entire parent directory path's length, *excluding* the final separator
+WIN_MAX_DIRPATH_LEN = 247
+# a single node's length (this seems to be the case for older NTFS)
+WIN_MAX_NODE_LEN = 226
+# Prefix for long paths in Windows API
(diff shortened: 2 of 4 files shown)


phw:
@rdswift Could you try to test this on Windows 7 with a file with a long path (> 259 characters)? I think it should work, but could not test it myself.

zas (approved):
LGTM

rdswift:
> I think it should work, but could not test it myself.

Nor could I because Windows 7 won't allow me to create a path that exceeds the normal maximum length.

rdswift (commented):
The code looks good to me (except your note about appending a prefix, which I think should actually be "prepending"  :stuck_out_tongue_winking_eye: ).  Unfortunately I couldn't test it here because my Windows 7 system won't allow me to create a long path for testing.

phw:
@rdswift thanks for checking, I'll fix the note.

I'd say we move forward with this then. I have a second patch set in the works to allow users to lift the 259 character limit on Windows. It's nearly ready, but this will allow fully testing the long path support as then Picard itself can generate such files.

phw:
I changed "append" to "prepend in the comment

zas (approved):

PR #2074, opened 2022-03-08
Remove enabledSlot() and refactor renaming options UI

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

Some refactoring and code cleanup

# Solution
- Remove `picard.util.enabledSlot`, this was a workaround for older PyQt4 versions which is no longer needed. It even would not apply to the cases here if connecting toggled to setEnabled
- Remove some code dealing with disabling / enabling UI elements in renaming options if "move files" gets toggled. This is not needed since the move files option is a checkable QGroupBox, which takes care of disabling elements if unchecked.

# Action


zas (approved):
Good move

PR #2075, opened 2022-03-08
PICARD-2076: Remove Windows 259 characters path length limit

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2076, PICARD-2429

This change allows the user to disable the 259 character path limit Picard would usually enforce in Windows compatibility mode when renaming and / or moving files. This is possible both on Windows and on other platforms with win compat enabled.

The reason for the 259 character limit are legacy Windows API limitations (MAX_PATH = 260, which includes a trailing NULL character). But Windows for some time already supports longer paths with the `\\?\` prefix (see PICARD-1570), and since Windows 10 version 1607 this path limit even can be completely disabled in the Windows registry.

# Solution

- Add an option `windows_long_paths`, which when enabled no longer restricts path length to 259 characters in win_compat mode.
- Added a method `picard.util.system_supports_long_paths` which detects whether long path support is enabled on Windows. All other systems are assumed to have long path support (as before).
- When enabling this option on Windows without long path support a warning dialogue is shown to inform the user about possible consequences
- The `\\?\` prefix added in PICARD-1570 is only used if long path support is not enabled. Hence on Windows 10 1607 or later with long path support enabled the path appear normally
- Added debug output on startup to indicate the status of `picard.util.system_supports_long_paths` (to help debug)

# Action

I tested this on two systems, Windows 10 with long path support enabled and Windows 11 without long path support enabled. On both system I had no issue saving files with long paths using Picard. As expected the system with long path support disabled used the `\\?\` prefix, the other not.

Things to do now:

- This needs testing on Windows 7, @rdswift can you look into this?
- The new option must be documented in picard-docs
- picard-docs should also explain the downsides of this Picard option and that enabling long path support in Windows 10 1607 or later is recommended

The downsides of enabling long paths in Picard are:

- In general some software might have problems accessing the files
- In particular Windows Explorer can have problems. I think in older Windows version it can plainly fail to access long folders completely. But even in current Windows 10 / 11 (even with long path support enabled), Explorer will not allow users to rename the files or create files / folders inside folders that exceed the limit. You can however browse these folders and open files
- Drag and drop to Picard is not working, unless long path support is enabled in Windows registry (see PICARD-2429)

Advantages of enabling long path support in Windows 10 / 11:

- Paths can be represented without `\\?\` prefix
- Drag and drop is working (see PICARD-2429)


rdswift:
Tested with the portable version.  It creates the target directory with the long path but fails when trying to copy the file.  Log shows:

```
D: 11:32:14,445 formats\vorbis._save:239: Saving file 'C:\\CD Rip\\Ripped\\JEEN. - Test File.flac'
E: 11:32:14,466 util\thread.future_callback:59: Traceback (most recent call last):
  File "util\thread.py", line 55, in future_callback
  File "concurrent\futures\_base.py", line 437, in result
  File "concurrent\futures\_base.py", line 389, in __get_result
  File "concurrent\futures\thread.py", line 57, in run
  File "file.py", line 362, in _save_and_rename
  File "file.py", line 529, in _rename
  File "os.py", line 213, in makedirs
  File "os.py", line 213, in makedirs
  File "os.py", line 213, in makedirs
  [Previous line repeated 1 more time]
  File "os.py", line 223, in makedirs
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\CD Rip\\Tagged\\Picard Portable\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard'

E: 11:32:14,469 ui\item.error_append:103: <FLACFile 'JEEN. - Test File.flac'>: [WinError 3] The system cannot find the path specified: 'C:\\CD Rip\\Tagged\\Picard Portable\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard\\This is a long path name to test Picard'
D: 11:32:14,469 file.update:700: Updating file <FLACFile 'JEEN. - Test File.flac'>
```

phw:
Thanks for testing. Yes, I think there is a missing normpath file for creating the directories. In my tests the directories created probably fittest into the path limit and only the file then exceeded it. I'll check

rdswift:
I just did another test keeping the directory path short but making the file name over 259 characters, and it still failed.  Script was:

```
$set(_longpath,This is a long path name to test Picard)
Long File Name Test/
01 %_longpath% -
02 %_longpath% -
03 %_longpath% -
04 %_longpath% -
05 %_longpath% -
06 %_longpath% -
07 %_longpath% -
08 %_longpath% -
09 %_longpath% -
10 %_longpath% -
%title%
```

and the log showed:

```
D: 12:34:10,533 file._rename:531: Moving file 'C:\\CD Rip\\Ripped\\JEEN. - Test File.flac' => 'C:\\CD Rip\\Tagged\\Picard Portable\\Long File Name Test\\01 This is a long path name to test Picard -02 This is a long path name to test Picard -03 This is a long path name to test Picard -04 This is a long path name to test Picard -05 This is a long path name to test Picard -06 This is a long path name to.flac'
E: 12:34:10,544 util\thread.future_callback:59: Traceback (most recent call last):
  File "shutil.py", line 791, in move
FileNotFoundError: [WinError 3] The system cannot find the path specified: 'C:\\CD Rip\\Ripped\\JEEN. - Test File.flac' -> 'C:\\CD Rip\\Tagged\\Picard Portable\\Long File Name Test\\01 This is a long path name to test Picard -02 This is a long path name to test Picard -03 This is a long path name to test Picard -04 This is a long path name to test Picard -05 This is a long path name to test Picard -06 This is a long path name to.flac'

During handling of the above exception, another exception occurred:

  File "util\thread.py", line 55, in future_callback
  File "concurrent\futures\_base.py", line 437, in result
  File "concurrent\futures\_base.py", line 389, in __get_result
  File "concurrent\futures\thread.py", line 57, in run
  File "file.py", line 362, in _save_and_rename
  File "file.py", line 532, in _rename
  File "util\filenaming.py", line 483, in move_ensure_casing
  File "shutil.py", line 811, in move
  File "shutil.py", line 435, in copy2
  File "shutil.py", line 264, in copyfile
FileNotFoundError: [Errno 2] No such file or directory: 'C:\\CD Rip\\Tagged\\Picard Portable\\Long File Name Test\\01 This is a long path name to test Picard -02 This is a long path name to test Picard -03 This is a long path name to test Picard -04 This is a long path name to test Picard -05 This is a long path name to test Picard -06 This is a long path name to.flac'

E: 12:34:10,545 ui\item.error_append:103: <FLACFile 'JEEN. - Test File.flac'>: [Errno 2] No such file or directory: 'C:\\CD Rip\\Tagged\\Picard Portable\\Long File Name Test\\01 This is a long path name to test Picard -02 This is a long path name to test Picard -03 This is a long path name to test Picard -04 This is a long path name to test Picard -05 This is a long path name to test Picard -06 This is a long path name to.flac'
D: 12:34:10,545 file.update:700: Updating file <FLACFile 'JEEN. - Test File.flac'>
```

Note that it appears that Picard stopped building the file name part way through, when continuing would have exceeded the 259 character limit.

phw:
The odd part is that the paths don't have the `\\?\`  prefix, but they should have. I can't reproduce this here. Can you check what ` picard.util.system_supports_long_paths()`  reports on your system? It should be `False`, but to me it seems it returns `True`  in your case.

phw:
@rdswift I think I found the issue. The problem was with `picard.util.normpath`. If you had a long path selected Windows might instead return a DOS compatible path name, something like:

> C:\Users\Developer\Music\GEORGE~1\GEORGE~1\GEORGE~1\GEORGE~1\GEORGE~1\GEORGE~1\GEORGE~1\GEORGE~1\BADTOT~1

This is shorter then the max length, hence the ` \\?\` does not get added. But running `os.path.realpath`  on this expands it to the full path names, in this case it was:

> C:\Users\Developer\Music\George Thorogood & the Destroyers\George Thorogood & the Destroyers\George Thorogood & the Destroyers\George Thorogood & the Destroyers\George Thorogood & the Destroyers\George Thorogood & the Destroyers\George Thorogood\George Thorogood & the Destroyers\Bad to the Bone\

This is longer than max path, but still no prefix added. I know turned this around so the prefix gets added before realpath gets called.

rdswift:
> but to me it seems it returns `True` in your case.

Yup, it is returning True.

phw:
> Yup, it is returning True.

@rdswift That's odd. You are on Windows 7? Could you check with regedit if you have a key `LongPathsEnabled` in the registry under `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem` and if it is set to 1?

Usually this should not exist on older Windows versions, but could of course exist, and that would confuse Picard currently. Maybe this needs to be combined with a Windows version check.

picard/util/__init__.py
```diff
@@ -175,20 +179,45 @@ def decode_filename(filename):
...
 
+def system_supports_long_paths():
+    """Detects long path support.
+
+    On Windows returns True, only if long path support is enabled in the registry (Windows 10 1607 or later).
+    All other systems return always True.
+    """
+    if not IS_WIN:
+        return True
+    else:
```
zas, line 190: We could drop this `else` since the other branch returns, in order to reduce indentation level for the rest of the code.

zas (changes requested):

rdswift:
I added the following lines to the `text_paragraphs` list in `aboutdialog.py`

```
            strong('Supports long paths: {0}'.format(system_supports_long_paths(),)),
            strong('platform.version(): {0}'.format(platform.version(),)),
            strong('sys.getwindowsversion(): {0}'.format(sys.getwindowsversion(),)),
```

which produced the following output:

image

rdswift:
After changing the value of registry key `HKEY_LOCAL_MACHINE\SYSTEM\CurrentControlSet\Control\FileSystem\LongPaths\Enabled` to '0', both the long path and long filename tests worked perfectly.

picard/util/__init__.py
```diff
@@ -175,20 +179,57 @@ def decode_filename(filename):
...
+    try:
+        # Long path support can be enabled in Windows 10 version 1607 or later
+        if _check_windows_min_version(10, 14393):
+            with winreg.OpenKey(winreg.HKEY_LOCAL_MACHINE,
+                                r"SYSTEM\CurrentControlSet\Control\FileSystem") as key:
+                supported = winreg.QueryValueEx(key, "LongPathsEnabled")[0] == 1
+        else:
+            supported = False
+        system_supports_long_paths._supported = supported
+        return supported
```
rdswift, line 212: I believe these should be `_supported` rather than `supported`,

rdswift (changes requested):

picard/util/__init__.py
rdswift: Wrong.  I didn't look close enough.  Sorry about that.

rdswift (approved):

picard/util/__init__.py
phw: No, should be fine that way. `supported`  is the local variable, `system_supports_long_paths._supported`  is the cache value (to avoid the rather expensive check with registry lookup)

zas (approved):
LGTM

PR #2076, opened 2022-03-08
PICARD-993: updated error symbol

Aerozol:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Replacing stop error symbol with exclamation symbol as in ticket.

* JIRA ticket: PICARD-993


phw:
As discussed on the ticket I'll check what is needed to distingiush between different error conditions here.

phw:
@Aerozol I think we can handle errors caused by write permissions and errors caused by file not being found at the path separately. I started working on the necessary changes.

Could you provide me the question mark and lock icon in the same formats as this here (SVG and the 16x16 PNGs)? Then I will finalize the patch.

Aerozol:
Great @phw, here are all the files: https://drive.google.com/drive/folders/173oD54B28LN4Vg0vK7yMy2jTcI1eIcnJ?usp=sharing
Should be able to plug them straight into the resources/images folder, let me know if you want any changes

phw:
This is now part of #2078

PR #2081, opened 2022-03-12
PICARD-2420: Fix AcoustID submission getting enabled for files with fingerprint in tags

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2420

If one adds a file to Picard that has an existing `acoustid_fingerprint` and an existing `musicbrainz_recordingid` tag, that file will be loaded and assigned to the corresponding recording. This immediately will enable the fingerprint submission, because the file originally gets added to the AcoustId manager without recording ID, then gets assigned to the recording. This makes this a new combination.

# Solution

The patch changes the behavior so on loading the file gets added to the AcoustId manager with the recording ID already assigned. That way it counts as already submitted.

I'd like to put this up for discussion, because I am unsure what expected behavior is. The current behavior can lead to a AcoustID / MBID combination submitted every time the user loads the file into Picard.

The new behavior will treat this already tagged combination as already submitted and won't submit.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -255,7 +255,7 @@ class File(QtCore.QObject, Item):
         if not config.setting["ignore_existing_acoustid_fingerprints"]:
             fingerprints = self.metadata.getall('acoustid_fingerprint')
             if fingerprints:
-                self.set_acoustid_fingerprint(fingerprints[0])
+                self.set_acoustid_fingerprint(fingerprints[0], recordingid=self.metadata['musicbrainz_recordingid'])
         run_file_post_load_processors(self)
         self.update()
         callback(self)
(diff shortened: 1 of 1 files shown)


zas:
> I'd like to put this up for discussion, because I am unsure what expected behavior is. The current behavior can lead to a AcoustID / MBID combination submitted every time the user loads the file into Picard.
> 
> The new behavior will treat this already tagged combination as already submitted and won't submit.

The new behavior looks much safer,  but I think we need to ensure two things:
- that the user is aware of what's happening
- that the user is able to force a submission

phw:
> * that the user is able to force a submission

I think that's always possible with "generate AcoustID fingerprints" I think, but need to actually test it.



> * that the user is aware of what's happening

Not sure what to do here. Suggestions?

zas:
> Not sure what to do here. Suggestions?

At least ensure we update status and log when it happens.

rdswift:
> Not sure what to do here. Suggestions?

Perhaps something with the fingerprint icon for the track?  Show as already submitted?

zas (approved):
LGTM

phw:
Yes, let's merge this. I think this behavior is indeed an improvement over the previous.

PR #2082, opened 2022-03-12
An editable QComboBox automatically sets up a QCompleter

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

A `QComboBox`  which is set to being editable automatically sets up a `QCompleter` for the values in the combo box. No need to initialize a new completer, we just need to configure it for using popup auto completion (instead of default inline completion).


zas (approved):
Less code for the same result, good work. ;)

PR #2086, opened 2022-04-07
PICARD-2451: Remove unicode zero-width space characters from file path

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Removes unicode zero-width spaces from renaming file output path.

# Problem

As discussed in https://github.com/metabrainz/picard-plugins/pull/326 it is currently possible for Picard to create an output path which includes a unicode zero-length space character (\\u200B).  This can lead to unexpected consequences for the user.

* JIRA ticket (_optional_): PICARD-2451

# Solution

Automatically strip out the zero-width space character.  If there is a valid use case to keep the character in the output path, perhaps a setting can be added to the File Renaming Options to allow keeping the character.

# Action

None.


phw (approved):
I don't see a downside in this. @zas anything?

zas (approved):
LGTM

PR #2088, opened 2022-04-08
Optimize access to profile settings

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

On every settings read / write access the list of profile setting names got compiled again. This list does not change on runtime, so instead compute it only once.


zas (approved):
LGTM

rdswift (approved):
Makes perfect sense to me.  Well done.

PR #2094, opened 2022-04-23
PICARD-2462: Add functions to store and retrieve persistent script variables.

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds functions to store, retrieve and clear variable values that persist between tracks and albums.

# Problem

Picard does not provide the ability to persist script variable values between different tracks or albums.

* JIRA ticket (_optional_): PICARD-2462

# Solution

Add four new scripting functions to allow storing, retrieving and clearing persistent script variables:

- `$setp(name,value)` : Sets the persistent variable `name` to `value`.
- `$unsetp(name)` : Unsets the persistent variable `name`.
- `$getp(name)` : Gets the persistent variable `name`.
- `$clearp()` : Clears all persistent variables.

Add tests for the new functions.

# Action

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -67,6 +67,29 @@ except ImportError:
 Bound = namedtuple("Bound", ["lower", "upper"])
 
 
+class PersistentVariables:
+    variables = {}
+
+    @classmethod
+    def clear_variables(cls):
+        cls.variables = {}
+
+    @classmethod
+    def set_variable(cls, key, value):
+        if key:
+            cls.variables[key] = value
+
+    @classmethod
+    def unset_variable(cls, key):
+        cls.variables.pop(key, None)
+
+    @classmethod
+    def get_variable(cls, key):
+        if key not in cls.variables:
+            return ""
+        return cls.variables[key]
+
+
 class FunctionRegistryItem:
     def __init__(self, function, eval_args, argcount, documentation=None,
                  name=None, module=None):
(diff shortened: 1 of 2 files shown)


rdswift:
@phw I just read your note on the [Community Forum](https://community.metabrainz.org/t/new-ideas-for-enhancement/581619/13) about typically releasing new scripting functions as plugins first, and possibly including them in Picard's base code later.  Would you prefer that I do that with these changes, and cancel this PR?

phw:
Interesting idea, but the results look a bit random to me. There is no guarantee in which order scripts are getting run. Definitely not for which albums, and even for tracks the order is more an implementation details. E.g. script gets rerun when matching files to tracks, and this can be in any order.

Do you have a concrete example how this can be used?

> typically releasing new scripting functions as plugins first, and possibly including them in Picard's base code later. Would you prefer that I do that with these changes, and cancel this PR?

I would not say typically, but in this case I thing yes, I'd prefer to have this as a plugin. In this case I think it can be considered experimental, and when people use it we could explore the actual use cases and the limitations users face with this functionality. That would allow us to maybe come up with a improved implementation inside Picard.

rdswift:
> Do you have a concrete example how this can be used?

A possible application would be to find and store the earliest track recording date for an album to include in the oputput path.  For this you might have two tagging scripts:

```
$noop( Clear the persistent variable.  This would be run manually before any track processing. )
$unsetp(earliest_recording_date)
```

and

```
$noop( Set the persistent variable.  This would be run automatically during any track processing. )
$if(%_recording_firstreleasedate%,
    $if($gt(%_recording_firstreleasedate%,$getp(earliest_recording_date)),
        $setp(earliest_recording_date,%_recording_firstreleasedate%)
    )
)
```

The file naming script could then retrieve the resulting value as `$getp(earliest_recording_date` to include in the file path.

It could also be used to mark an album as having explicit lyrics if "explicit" appears in any track's `%_recording_comment%` tag, or even something as frivolous as count how many tracks had been saved in the current Picard session.

rdswift:
I'll convert this to a plugin as suggested.

PR #2095, opened 2022-04-24
PICARD-2459: Handle missing coverart temp files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

The cover art images store the image data in temporary files. If the temporary file was removed Picard would crash. This happens for example on macOS if Picard is running for more than 3 days, as macOS removes temporary files not getting accessed for this time.

# Solution

This fix consists of two parts:

- The first commit handles the crash and introduces a placeholder pixmap being shown instead of the real data. This makes Picard more error tolerant and avoids crashes. Note that the situation is still considered an error, e.g. if you save a file with a cover art file attached with missing data Picard will show a saving error because the cover art cannot be accessed. The user needs to assign a new cover art file or e.g. reload the release.
- The seconds commit attempts to prevent the specific situation with the temp file cleanup on macOS by periodically (every 1 hour) touching the temporary files. As the timestamp will periodically update macOS will not delete the files.

To test the new handling of missing temp files one can manually remove cover art temp files while Picard is running and see the replacement pixmap being used in cover art box and info dialog.


picard/coverart/image.py
```diff
@@ -115,6 +134,13 @@ def data(self):
     def filename(self):
         return self._filename
 
+    def touch(self):
+        if self._filename:
+            try:
+                Path(self._filename).touch()
```
zas, line 140: I think using `touch()` alone isn't totally correct, because it may (re-)create the file if it was removed. Also, according to https://superuser.com/questions/187071/in-macos-how-often-is-tmp-deleted/187105#187105 the cleaner script checks for atimes too, but this is not reliable (dunno on macos, but on linux one can disable atime updates), so `touch()` is appropriate as it will change mtime.
So I think we just need to ensure the file still exists before touching it, there will be a race condition anyway since we use filename.

Also, I think we need something more generic that can be used for other temporary files or dirs, so instead of having a timer in cover art, and a code to touch the file I would create a "main" timer (running every X minutes), that touch a list of files build by a register function.

Something like:

```python
files_to_touch_on_regular_basis = set()

def  touch_on_regular_basis(filepath):
  files_to_touch_on_regular_basis.add(filepath)

def run_timer_touch_on_regular_basis():
   for filepath in files_to_touch_on_regular_basis:
       # check if filepath still exists
       # if it does, touch
       # if not, remove from files_to_touch_on_regular_basis and do nothing
```

zas (changes requested):

picard/coverart/image.py
phw: Ok, yes. That makes sense. I'll rework the patch and put that functionality into a separate module. This all feels a bit awkward still, but I think it is still the better solution than using a custom temp folder.

I'll also split the PR then and submit the first commit separately. I think both are really independent of each other, and I need a few days to find the time for reworking the temp file handling.

phw:
Closing this for now in favor of #2096, will add a separate PR for the temp file touching.

PR #2096, opened 2022-04-25
PICARD-2459: Handle cover art temp files missing

phw:
Do not crash if the temporary file for cover art is unexpectedly missing. Instead show a place holder pixmap.

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

The cover art images store the image data in temporary files. If the temporary file was removed Picard would crash. This happens for example on macOS if Picard is running for more than 3 days, as macOS removes temporary files not getting accessed for this time.

This PR provides the first part of #2095, where Picard now handles the exception without crashing. I'll add another separate PR to address the keep-alive of the temporary files.

# Solution

Handle the crash and introduce a placeholder pixmap being shown instead of the real data. This makes Picard more error tolerant and avoids crashes. Note that the situation is still considered an error, e.g. if you save a file with a cover art file attached with missing data Picard will show a saving error because the cover art cannot be accessed. The user needs to assign a new cover art file or e.g. reload the release.


phw:
Just to demonstrate how this can look like if files are missing:

grafik

zas (approved):
LGTM

PR #2097, opened 2022-04-26
PICARD-2463: Fix coverart per screen pixel ratio

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2463

The cover art box considers the screen scaling to render the cover art in the proper resolution. E.g. on a High DPI / retina display with a scaling factor of 2 it will render the images in twice the displayed size.

But this code does not consider multiple screens with different scaling factors. If there are multiple screens the scaling factor of the primary will be used for all screens. This can result in pixelated cover art display if the Picard window is on a a screen with different scaling factor. E.g. the attached screenshot shows the Picard window on a screen with scaling factor set to 1.25, but rendered with a scaling factor of 1.0 (from primary screen). If the primary screen is changed to the one with 1.25 scaling factor the cover art will render nicely on this display, but when moving the window to the one with 1.0 it will look pixelated again.

# Solution

- Handle `screenChanged` events in cover art to detect Picard window changing the screen. This requires setting `WA_NativeWindow` on the main window to have a native window handle available. In case pixel ratio changed re-render the currently displayed cover art.
- Make the pixel ratio part of the key for the LRU cache, so changes in pixel ration invalidate current cache
- The default shadow and missing data pixmaps also get pushed in the cache now
- Apply some refactoring by moving the code to render the cover art stack pixmap into own method to improve readability

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -67,6 +67,7 @@ from picard.ui.widgets import ActiveLabel
 
 
 THUMBNAIL_WIDTH = 128
+COVERART_WIDTH = THUMBNAIL_WIDTH - 7
 
 
 class CoverArtThumbnail(ActiveLabel):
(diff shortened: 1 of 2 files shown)


phw:
For example I have two screens, the primary using 1.0 scaling, the second using 1.25 scaling. Without this patch this is how the cover art box looks like on the scaled screen:

grafik

If I make the scaled screen the primary it looks good on that screen, but similar scaling issues display when I then move the Picard window to the unscaled screen.

With the fix this is how the cover art looks like on the scaled screen:

grafik

Moving the window between the two screens now updates the box properly and it displays crisp on both displays.

zas (dismissed):
LGTM

phw:
@zas Needed a small fix, the call to `set_data` was wrong, see 0641874c5617549591db242f57e9e91bf5055dca

zas:
> @zas Needed a small fix, the call to `set_data` was wrong, see [bc5c55a](https://github.com/metabrainz/picard/commit/bc5c55a063a4132799d007bb3ae6a18c350870af)

Oh, I missed that :)

zas (approved):
LGTM

phw:
Found another issue PICARD-2464. Tried to quickly fix it here, but fix was borked. Also technically a separate issue, so let's keep this separate

PR #2098, opened 2022-04-26
PICARD-2464: Fix rendering of cover art stack on HiDPI displays

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2464

Sizes where applied wrong when the pixel ratio > 1.0 was used, resulting in too small thumbnails inside the stack:

grafik

# Solution

Solved by using logical pixel coordinates in the QPainter and setting the pixel ratio of the underlying QPixmap correctly. That way all painting can use logical pixels and the painter takes care to upscale everything accordingly. The only places where we need to consider the scaling ourself is when setting the size of a QPixmap (on creation or when scaling it) and when accessing `width()` and `height()` of a QPixmap.

Result now renders properly:

grafik

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -211,26 +211,33 @@ class CoverArtThumbnail(ActiveLabel):
                 pixmap = self.render_cover_stack(self.data, has_common_images)
             self._pixmap_cache[key] = pixmap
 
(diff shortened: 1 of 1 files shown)


zas (approved):
Good catch, code is simpler and works better, LGTM.

PR #2099, opened 2022-04-26
PICARD-2459: Prevent macOS from removing temporary cover art files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

macOS cleans up temporary files if they haven't been accessed for three days. If Picard is left running for a longer period of time the files became missing.

# Solution

Add a module `picard.util.periodictouch` which sets up a timer to periodically touch a list of files. Register the cover art temp files there.

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -54,6 +54,7 @@ from picard.util import (
     encode_filename,
     imageinfo,
     is_absolute_path,
+    periodictouch,
 )
 from picard.util.scripttofilename import script_to_filename
 
@@ -77,6 +78,7 @@ class DataHash:
                 with os.fdopen(fd, "wb") as imagefile:
                     imagefile.write(data)
                 _datafiles[self._hash] = self._filename
+                periodictouch.register_file(self._filename)
                 log.debug("Saving image data %s to %r" % (self._hash, self._filename))
             else:
                 self._filename = _datafiles[self._hash]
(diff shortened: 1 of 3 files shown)


picard/coverart/image.py
```diff
@@ -77,6 +78,7 @@ def __init__(self, data, prefix='picard', suffix=''):
                 with os.fdopen(fd, "wb") as imagefile:
                     imagefile.write(data)
                 _datafiles[self._hash] = self._filename
+                periodictouch.register_file(self._filename)
```
zas, line 81: I think those functions should be no-op on anything but macos.

zas (changes requested):

picard/coverart/image.py
phw: I didn't want to push that logic all over the place and keep the decision of whether to start this at a single place to tagger.py (which basically controls bootstrapping the application, and hence I think there this decision belongs to).

But I changed the module to not fill the temp file set if the timer is not active, see https://github.com/metabrainz/picard/pull/2099/commits/7e3e8954cb4ea3dc201500231699c9e7e7e63569

zas (approved):
LGTM

picard/coverart/image.py
zas: Yes, it is even better. Nice work.

PR #2100, opened 2022-04-29
Simplify album and track similarity search

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

Refactor `AlbumSearchDialog.show_similar_albums`  and  `TrackSearchDialog.show_similar_tracks()` 

# Solution

- De-duplicate code
- Reuse existing `search` method
- Set the dialog to have forced advanced search. Previously the search would initially be performed as an advanced search, but if the user had simple search active the search string was filled as only a simple title search. That meant if the user hit the search button there would be different results. In forced mode the dialog uses advanced search independent of user settings (and does also not alter user settings)
- Removed duplicated code by introducing `build_lucene_query`
- Added tests for API helpers `build_lucene_query` and `escape_lucene_query`

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -1376,8 +1376,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         if not isinstance(obj, File):
             log.debug('show_more_tracks expected a File, got %r' % obj)
             return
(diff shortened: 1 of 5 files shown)


zas (approved):
LGTM

PR #2104, opened 2022-05-08
PICARD-2470: Allow UTF-16 encoded EAC log files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2470

EAC seems to save UTF-16 encoded files. There circulate TOC files in UTF-8 encoding, but mostly these are probably the result of copy and pasting the log.

# Solution
Detect the encoding based on the BOM (which for UTF-16 should always be present).


zas (approved):
LGTM.

It lacks tests for others utf16/32 BOMs, but not sure that's worth since it uses the same code.

phw:
> It lacks tests for others utf16/32 BOMs, but not sure that's worth since it uses the same code.

Yes. I think the only encoding EAC is using is UTF-16 LE anyway. I just added the other BOMs for completeness, doesn't make things more complicated. UTF-32 is extremely rare anyway (many editors can not even save it), and UTF-16 BE is also rare because BE systems are :)

PR #2105, opened 2022-05-08
PICARD-2473: EAC log disc lookup must be available on systems without CD drive

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2473

On Windows systems if there is not CD drive detected the CD lookup gets completely disabled. This lead to the "Lookup from ripper log" action to not be available at all.

This behaved a bit different on Linux where CD lookup was enabled regardless, and would always attempt to lookup the default "/dev/cdrom".

# Solution
Make the lookup disc ID action always enabled. If there is not drive make the ripper log file lookup the default.

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -91,6 +91,7 @@ from picard.util import (
     icontheme,
     iter_files_from_objects,
     iter_unique,
+    reconnect,
     restore_method,
     thread,
     throttle,
@@ -731,13 +732,13 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         # TR: Keyboard shortcut for "Lookup CD"
         action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
         action.triggered.connect(self.tagger.lookup_cd)
-        action.setEnabled(False)
         self.cd_lookup_action = action
 
         self.cd_lookup_menu = QtWidgets.QMenu(_("Lookup &CD..."))
         self.cd_lookup_menu.triggered.connect(self.tagger.lookup_cd)
         if discid is None:
             log.warning("CDROM: discid library not found - Lookup CD functionality disabled")
+            self.cd_lookup_action.setEnabled(False)
         else:
             thread.run_task(get_cdrom_drives, self._update_cd_lookup_actions)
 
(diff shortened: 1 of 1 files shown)


picard/ui/mainwindow.py
```diff
@@ -903,31 +904,39 @@ def _update_cd_lookup_actions(self, result=None, error=None):
...
 
+    def _update_cd_lookup_default_action(self, action):
+        if action:
+            reconnect(self.cd_lookup_action.triggered, action.trigger)
+        else:
+            reconnect(self.cd_lookup_action.triggered, self.tagger.lookup_cd)
+
     def _update_cd_lookup_button(self):
-        if len(self.cd_lookup_menu.actions()) > 1:
+        if len(self.cd_lookup_menu.actions()) > 0:
```
zas, line 939:
```suggestion
        if self.cd_lookup_menu.actions():
```

picard/ui/mainwindow.py
```diff
@@ -903,31 +904,39 @@ def _update_cd_lookup_actions(self, result=None, error=None):
...
-                    self.cd_lookup_action.setShortcut(QtGui.QKeySequence())
-                    action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
-        self._set_cd_lookup_from_file_actions()
+                    self._update_cd_lookup_default_action(action)
+        self._set_cd_lookup_from_file_actions(has_drives)
         self._update_cd_lookup_button()
 
-    def _set_cd_lookup_from_file_actions(self):
+    def _set_cd_lookup_from_file_actions(self, has_drives):
         if len(self.cd_lookup_menu.actions()) > 0:
```
zas, line 925:
```suggestion
        if self.cd_lookup_menu.actions():
```

picard/ui/mainwindow.py
```diff
@@ -903,31 +904,39 @@ def _update_cd_lookup_actions(self, result=None, error=None):
...
             shortcut_drive = config.setting["cd_lookup_device"].split(",")[0] if len(drives) > 1 else ""
-            self.cd_lookup_action.setEnabled(discid is not None)
-            self.cd_lookup_menu.clear()
             for drive in drives:
                 action = self.cd_lookup_menu.addAction(drive)
                 action.setData(drive)
                 if drive == shortcut_drive:
-                    # Clear existing shortcode on main action and assign it to sub-action
-                    self.cd_lookup_action.setShortcut(QtGui.QKeySequence())
-                    action.setShortcut(QtGui.QKeySequence(_("Ctrl+K")))
```
zas, line 919: What happens to this shortcut?

picard/ui/mainwindow.py
```diff
@@ -903,31 +904,39 @@ def _update_cd_lookup_actions(self, result=None, error=None):
             self.update_cd_lookup_drives(result)
 
     def update_cd_lookup_drives(self, drives):
+        self.cd_lookup_menu.clear()
+        self.cd_lookup_action.setEnabled(discid is not None)
         if not drives:
             log.warning("CDROM: No CD-ROM drives found - Lookup CD functionality disabled")
+            has_drives = False
         else:
+            has_drives = True
```
zas, line 913: I don't think we need `has_drives`, since we can use `drives` or `bool(drives)`.

zas (changes requested):

picard/ui/mainwindow.py
phw: It stays on `self.cd_lookup_action` , which then just triggers the default action from the menu. That's way easier then to change the shortcut only. But I also needed this so it triggers the eaclog action properly.

picard/ui/mainwindow.py
phw: Done

picard/ui/mainwindow.py
phw: Done

picard/ui/mainwindow.py
phw: True, I changed it to just pass `drives`  around

zas (approved):

PR #2107, opened 2022-05-10
PICARD-2472: Cluster only unclustered files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2472

Clustering will apply to files selected in the right pane, which is rather unexpected. It should only apply to unclustered files. Furthermore the cluster action only is activated if there is at least one unclustered file, so only under this condition the user can also apply it to matched files.

# Solution
Limit clustering to unclustered files

@zas can you test this and see if this meets your expected behavior?

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -331,11 +331,11 @@ class UnclusteredFiles(Cluster):
 
     def add_files(self, files, new_album=True):
         super().add_files(files, new_album=new_album)
-        self.tagger.window.enable_cluster(self.get_num_files() > 0)
+        self.tagger.window.enable_cluster(bool(self.files))
 
     def remove_file(self, file, new_album=True):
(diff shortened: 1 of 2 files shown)


zas (approved):
I tested, and the new behavior makes much more sense.
One can cluster independently of what is selected in the right pane (so it doesn't re-cluster files already matched), and it is still possible to cluster unclustered selected files.
To me, this behavior is much more sane and predictable.

phw:
> To me, this behavior is much more sane and predictable.

I think it is also what was behind the intention of the code. Especially the code deactivating the clustering action if there are no unclustered files strongly indicates this.

PR #2110, opened 2022-05-24
PICARD-2486: Add text-based comparison script functions

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds new text-based comparison scriping functions.

# Problem

The current scripting comparison functions `$gt()`, `$gte()`, `$lt()` and `$lte()` operate on interger values only. This does not support things like date comparisons. See the discussion at https://community.metabrainz.org/t/how-set-date-to-minimum-of-recording-firstreleasedate-and-originaldate/585648 and https://community.metabrainz.org/t/feature-request-numeric-functions-min-and-max-in-picards-scripting-language/585669 for more information.

* JIRA ticket (_optional_): PICARD-2486

# Solution

~Adds four new text-based comparison scripting functions `$textlt()`, `$textlte()`, `$textgt()` and `$textgte()`, plus text-based `$textmin()` and `$textmax()` functions.~

Adds an optional argument to the existing `$lt()`, `$lte()`, `$gt()` and `$gte()` functions to specify whether the arguments should be processed as **int**, **float**, **text** (case-sensitive text) or **nocase** (case-insensitive text).  If no argument is provided the comparison will be processed as **int** to maintain backward compatibility.

Also adds two new functions `$min()` and `$max()`, with the type of processing specified as one of **int**, **float** or **text** (case-sensitive text) as the first argument.

# Action

I assume that, if accepted, these changes would be applied to Picard v3.0 (or v2.9 if there is such a release).  If accepted for other than v3.0 then the function description docstrings will need to be updated accordingly.

I wasn't sure which branch to merge into, so I used the `master` branch.  We can change the target if necessary.

If accepted, the Picard documentation will need to be updated.

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -688,6 +688,7 @@ def _cmp(op, x, y, _type):
         _type = 'auto'
     _typer = None
     if _type == 'auto':
+        _type = 'text'
         for _test_type in [int, float]:
             try:
                 _type = (_test_type(x), _test_type(y))
@@ -695,8 +696,6 @@ def _cmp(op, x, y, _type):
                 break
             except ValueError:
                 pass
-        if _type == 'auto':
-            _type = 'text'
     if _type == 'text':
         return "1" if op(x, y) else ""
     elif _type == 'nocase':
@@ -1566,6 +1565,7 @@ def _type_args(_type, *args):
         _type = 'auto'
     _typer = None
     if _type == 'auto':
+        _type = 'text'
         for _test_type in [int, float]:
             try:
                 _type = set(_test_type(item) for item in haystack)
@@ -1573,8 +1573,6 @@ def _type_args(_type, *args):
                 break
             except ValueError:
                 pass
-        if _type == 'auto':
-            _type = 'text'
     _typer = None
     if _type == 'int':
         _typer = int
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1517,3 +1517,71 @@ def func_cleanmulti(parser, multi):
     values = [str(value) for value in parser.context.getall(name) if value or value == 0]
     parser.context[multi] = values
     return ""
+
+
+@script_function(documentation=N_(
+    """$textlt(x, y)
+
+Returns true if `x` is less than `y` using a text comparison.
+
+_Since Picard 3.0_"""
+))
+def func_textlt(parser, x, y):
+    return "1" if x < y else ""
+
+
+@script_function(documentation=N_(
+    """$textlte(x, y)
+
+Returns true if `x` is less than or equal to `y` using a text comparison.
+
+_Since Picard 3.0_"""
+))
+def func_textlte(parser, x, y):
+    return "" if x > y else "1"
+
+
+@script_function(documentation=N_(
+    """$textgt(x, y)
+
+Returns true if `x` is greater than `y` using a text comparison.
+
+_Since Picard 3.0_"""
+))
+def func_textgt(parser, x, y):
+    return "1" if x > y else ""
+
+
+@script_function(documentation=N_(
+    """$textgte(x, y)
+
+Returns true if `x` is greater than or equal to `y` using a text comparison.
+
+_Since Picard 3.0_"""
+))
+def func_textgte(parser, x, y):
+    return "" if x < y else "1"
+
+
+@script_function(documentation=N_(
+    """$textmin(x,...)
+
+Returns the minimum value using a text comparison.
+Can be used with an arbitrary number of arguments.
+
+_Since Picard 3.0_"""
+))
+def func_textmin(parser, x, *args):
+    return min((*args, x))
+
+
+@script_function(documentation=N_(
+    """$textmax(x,...)
+
+Returns the maximum value using a text comparison.
+Can be used with an arbitrary number of arguments.
+
+_Since Picard 3.0_"""
+))
+def func_textmax(parser, x, *args):
+    return max((*args, x))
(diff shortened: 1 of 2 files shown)


rdswift:
For the sake of completeness, should there also be integer-based `$min()` and `$max()` functions, even though I can't think of a use case?

zas:
I'm a bit bothered by introduction of another full set of new comparison functions for text and I wonder if we should handle text type in existing comparison functions instead.
Type of comparison could be specified using a third parameter (this family of functions only accepts 2 parameters).

What I'm thinking about is something like:

```
$lt(1, 2) -> $lt(1, 2) or $lt(1, 2, int)
$textlt(a, b) -> $lt(a, b, text)
```

@phw, @rdswift : what do you think?

picard/script/functions.py
```diff
@@ -681,11 +681,14 @@ def func_ne(parser, x, y):
...
 @script_function(documentation=N_(
-    """`$lt(x,y)`
+    """`$lt(x,y[,text])`
 
-Returns true if `x` is less than `y`."""
+Returns true if `x` is less than `y`.
+Values `x` and `y` are treated as integers unless `text` is set."""
 ))
-def func_lt(parser, x, y):
+def func_lt(parser, x, y, text=None):
```
zas, line 689: If we want to add more types of comparison, it would perhaps be better to have something like:

```python
def func_lt(parser, x, y, type=None):
    if type == 'text':
```

The idea behind this is that it will be possible to add new types, for example, 'date' or 'float'.
It will default to 'int', but this type should also be handled as a valid type.

zas (changes requested):

picard/script/functions.py
rdswift: Good point.  I've modified all functions to allow specifying the processing type.  In addition, I added floating point processing to the `$gt`, `$gte`, `$lt` and `$lte` functions, and integer and floating point processing options to the `$max` and `$min` functions.

rdswift:
In a [related PR](https://github.com/rdswift/picard/pull/14) @zas commented:

I wonder if we could extend this to eq(), ne() and not().

(Copied here to keep the discussion in one place.)

rdswift:
> I wonder if we could extend this to eq(), ne() and not().

We could, but it might be a bit odd because (as you pointed out in IRC) the default processing type is different for these functions.

zas (dismissed):

picard/script/functions.py
```diff
@@ -1517,3 +1533,72 @@ def func_cleanmulti(parser, multi):
...
+    return haystack
+
+
+def _extract(_func, _type, *args):
+    try:
+        haystack = _type_args(_type, *args)
+    except ValueError:
+        return ""
+
+    if _type == 'ncase':
```
zas, line 1564: What does `ncase` mean?

phw:
> Type of comparison could be specified using a third parameter (this family of functions only accepts 2 parameters).

I agree that another set of functions becomes difficult. And as a user I'd really expect this to do the right thing. Maybe we could even do an `auto` mode (in addition to the explicit `int`  and `text` modes) that does int comparison if both values are numeric strings, otherwise text comparison?

picard/script/functions.py
rdswift: It's a non case-sensitive text comparison.  I'm open to suggestions for a less cryptic / confusing name.  :wink:

rdswift:
> Maybe we could even do an `auto` mode (in addition to the explicit `int` and `text` modes) that does int comparison if both values are numeric strings, otherwise text comparison?

That sounds like a good idea.  I can take a look at adding that.  Should "auto" become the default?  Even though this new default would be different, I don't think it would break any scripts.

rdswift:
I've implemented the `auto` processing type option, which first uses `int` or `float` in that order if all arguments pass the type, falling back to `text` if both `int` and `float` raise exceptions.  For now, I've made this the default if no processing option is specified.  Is this okay?

Also, there are two different text comparison types available: `text` is a case-sensitive compare, and `nocase` is a case-insensitive compare.  The fallback for `auto` is currently `text`, but I'm wondering if it should be changed to `nocase` instead because that is the most flexible comparison method.  Any preference on this?

picard/script/functions.py
```diff
@@ -680,60 +680,97 @@ def func_ne(parser, x, y):
...
+def _cmp(op, x, y, _type):
+    """Compare x vs y using op method and specified _type
+       op is expected to be a method from operator module
+    """
+    if not _type:
+        _type = 'auto'
+    _typer = None
+    if _type == 'auto':
+        _type = 'text'
+        for _test_type in [int, float]:
```
zas, line 692: Just a minor thing, but a tuple (to preserver order) would be better than a list
```suggestion
        for _test_type in (int, float):
```

picard/script/functions.py
```diff
@@ -1517,3 +1554,94 @@ def func_cleanmulti(parser, multi):
...
+    haystack = set()
+    # Automatically expand multi-value arguments
+    for item in args:
+        haystack = haystack.union(set(x for x in item.split(MULTI_VALUED_JOINER)))
+    if not _type:
+        _type = 'auto'
+    _typer = None
+    if _type == 'auto':
+        _type = 'text'
+        for _test_type in [int, float]:
```
zas, line 1569:
```suggestion
        for _test_type in (int, float):
```

zas (changes requested):

zas (approved):

PR #2112, opened 2022-05-30
PICARD-2489: Handle strxfrm raising OSError

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2489

locale.strxfrm can raise OSError on some systems. Handle this gracefully to avoid Picard crashing.

# Solution

Add a function `picard.util.strxfrm` that wraps `locale.strxfrm`, but never raises `OSError` but only returns the original string (converted to lowercase for some normalization). Worst case strxfrm will fail on a string and sorting will not be fully correct, but Picard continues to function. OSError will be logged.

### DIFF
--- a/picard/ui/collectionmenu.py
+++ b/picard/ui/collectionmenu.py
@@ -23,8 +23,6 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-import locale
-
 from PyQt5 import (
     QtCore,
     QtGui,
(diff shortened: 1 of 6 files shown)


zas (approved):
LGTM

phw:
Just to have this info here:

- I added some details on what is happening on FreeBSD on https://tickets.metabrainz.org/browse/PICARD-2489
- In general Python's `locale.strxfrm` can throw `OSError` (but it is not documented that it does), see https://github.com/python/cpython/blob/677320348728ce058fa3579017e985af74a236d4/Modules/_localemodule.c#L418 . If it does depends on the libc implementation, but e.g. despite FreeBSD I also found a similar condition in Haiku's source code. So it's safer to handle that.

PR #2113, opened 2022-05-30
PICARD-2491: Handle crash when loading albums with MBID redirect

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2491

After loading an album that leads to an MBID redirect _parse_result would be False, but since 8fa8bd8 (PICARD-2398) this would trigger advanced relationship reload for the tracks. This would trigger finalizing the loading without data being parsed.

# Solution

Introduce ParseResult enum to separate the different results of parsing (parsed, not parsed because of redirect, parsed but missing track relationships that need loading).

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -120,6 +120,12 @@ class AlbumStatus(IntEnum):
     LOADED = 3
 
 
+class ParseResult(IntEnum):
+    PARSED = 0
+    REDIRECT = 1
+    MISSING_TRACK_RELS = 2
+
+
 class Album(DataObject, Item):
 
     metadata_images_changed = QtCore.pyqtSignal()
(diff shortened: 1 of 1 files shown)


picard/album.py
```diff
@@ -288,19 +297,21 @@ def _release_request_finished(self, document, http, error):
                         error = False
             else:
                 try:
-                    parsed = self._parse_release(document)
+                    parse_result = self._parse_release(document)
                     config = get_config()
-                    if not parsed and config.setting['track_ars']:
+                    if parse_result == ParseResult.MISSING_TRACK_RELS:
```
zas, line 302: Isn't `config.setting['track_ars']` test still needed?

phw: It's already handled in `_parse_release`. The state `ParseResult.MISSING_TRACK_RELS` will only be set if this config setting is used.

That is another nice effect of introducing this enum, it simplifies the code here. The double check for the config setting was only done because of the ambiguity of the boolean return value.

zas: Thanks for the explanation.

zas (approved):

PR #2115, opened 2022-06-03
Album load_tracks(): reduce code redundancy

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -462,9 +462,10 @@ class Album(DataObject, Item):
         # Generate a list of unique media, but keep order of first appearance
         self._new_metadata['media'] = " / ".join(list(OrderedDict.fromkeys(all_media)))
 
+        multiartists = len(artists) > 1
         for track in self._new_tracks:
             track.metadata["~totalalbumtracks"] = totalalbumtracks
-            if len(artists) > 1:
+            if multiartists:
                 track.metadata["~multiartist"] = "1"
         del self._release_node
         del self._release_artist_nodes
--- a/picard/album.py
+++ b/picard/album.py
@@ -413,6 +413,12 @@ class Album(DataObject, Item):
         all_media = []
         absolutetracknumber = 0
 
+        def _load_track(node, mm, artists, extra_metadata):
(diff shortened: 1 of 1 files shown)


picard/album.py
```diff
@@ -411,7 +411,13 @@ def _finalize_loading_track(self, track_node, metadata, artists, extra_metadata=
...
         artists = set()
         all_media = []
-        absolutetracknumber = 0
+
+        def _load_track(node, mm, artists, extra_metadata):
+            _load_track.absolutetracknumber += 1
+            extra_metadata['~absolutetracknumber'] = _load_track.absolutetracknumber
+            self._finalize_loading_track(node, mm, artists, extra_metadata)
+
+        _load_track.absolutetracknumber = 0
```
phw, line 420: why not keep this as `absolutetracknumber` variable before defining `_load_track`?

zas: I think it would require to use nonlocal keyword, I'll give it a try

zas: I did the change in [ea1fd44](https://github.com/metabrainz/picard/pull/2115/commits/ea1fd44fbee782794ae4ccccc78277c536aa66a6). Not sure which one is better, I tend to prefer the function attribute approach because it clearly links the counter to this function but the `nonlocal` approach is perhaps easier to understand.

What do you think?

phw: The inline definition of the function forms a closure which includes the variables defined in the scope of the `_load_tracks`  method, so this is fine.

I personally prefer how it is now. I don't really see this variable as belonging to that helper function. Rather the opposite, that variable exists in the outer context to be used there. I definitely find it easier to understand this way. But no strong opinion, it's maybe just because I'm so used to the usage of anonymous functions from other languages as well.

phw (approved):

picard/album.py
zas: Agreed, I rebased and only kept the nonlocal approach

phw (approved):

PR #2117, opened 2022-06-07
PICARD-2498: HTML escape plugin metadata before display in options

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2498

Currently all plugin metadata can potentially contain HTML which gets rendered when displaying the plugin details in options. This is not intentional, the only field that can currently contain HTML is the description.

This was revealed because the author list in the acousticbrainz plugin uses the format "Author Name <email>".

# Solution
HTML escape all plugin metadata fields except for the description.

For authors also try to detect the "Author Name <email>" and convert the author names into mailto links.

### DIFF
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -614,9 +614,10 @@ class PluginsOptionsPage(OptionsPage):
     @staticmethod
     def link_authors(authors):
         formatted_authors = []
-        re_author = re.compile(r"(?P<author>.*?)\s*<(?P<email>.*?)>")
+        re_author = re.compile(r"(?P<author>.*?)\s*<(?P<email>.*?@.*?)>")
         for author in authors.split(','):
-            match = re_author.match(author.strip())
+            author = author.strip()
+            match = re_author.fullmatch(author)
             if match:
                 author_str = '<a href="mailto:{email}">{author}</a>'.format(
                     email=escape(match['email']),
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -34,6 +34,7 @@ from functools import partial
 from html import escape
 from operator import attrgetter
 import os.path
+import re
 
 from PyQt5 import (
     QtCore,
@@ -601,7 +602,7 @@ class PluginsOptionsPage(OptionsPage):
             text.append(plugin.description + "<hr width='90%'/>")
         infos = [
             (_("Name"), escape(plugin.name)),
-            (_("Authors"), escape(plugin.author)),
+            (_("Authors"), self.link_authors(plugin.author)),
             (_("License"), plugin.license),
             (_("Files"), escape(plugin.files_list)),
         ]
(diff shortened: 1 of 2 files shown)


picard/ui/options/plugins.py
```diff
@@ -599,16 +601,32 @@ def refresh_details(self, item):
...
         ]
         for label, value in infos:
             if value:
                 text.append("<b>{0}:</b> {1}".format(label, value))
         self.ui.details.setText("<p>{0}</p>".format("<br/>\n".join(text)))
 
+    @staticmethod
+    def link_authors(authors):
+        formatted_authors = []
+        re_author = re.compile(r"(?P<author>.*?)\s*<(?P<email>.*?)>")
```
zas, line 617: In the email part, I think we should match `@` too.
It would avoid creating invalid mailto links for `some<thing>`.

picard/ui/options/plugins.py
```diff
@@ -599,16 +601,32 @@ def refresh_details(self, item):
...
-            (_("Files"), plugin.files_list),
+            (_("Files"), escape(plugin.files_list)),
         ]
         for label, value in infos:
             if value:
                 text.append("<b>{0}:</b> {1}".format(label, value))
         self.ui.details.setText("<p>{0}</p>".format("<br/>\n".join(text)))
 
+    @staticmethod
+    def link_authors(authors):
```
zas, line 615: Write a test set for this method?

picard/ui/options/plugins.py
```diff
@@ -599,16 +601,32 @@ def refresh_details(self, item):
...
+        formatted_authors = []
+        re_author = re.compile(r"(?P<author>.*?)\s*<(?P<email>.*?)>")
+        for author in authors.split(','):
+            match = re_author.match(author.strip())
```
zas, line 619:
```suggestion
            match = re_author.fullmatch(author.strip())
```

Else it will match stuff like `some<thing>else` (I think).

zas (changes requested):

phw:
@zas I added the test and addressed your comments. Tests also revealed some issue with extra whitespace being added in case without e-mail.

test/test_ui_options_plugins.py
```diff
@@ -0,0 +1,32 @@
...
+
+from picard.ui.options.plugins import PluginsOptionsPage
+
+
+class PluginsOptionsPageTest(PicardTestCase):
+
+    def test_link_authors(self):
+        self.assertEqual(
+            '<a href="mailto:coyote@acme.com">Wile E. Coyote</a>, Road &lt;Runner&gt;',
+            PluginsOptionsPage.link_authors('Wile E. Coyote <coyote@acme.com>, Road <Runner>'),
```
zas, line 31: I see an extra complication, those "mailbox" addresses can use double quotes (I think): "a,b" <x@y> can happen in a list of addresses.

Not sure it is worth bothering.

phw: That should be handled already because quotes get escaped

zas: Are you sure a comma inside double quotes is handled properly? I didn't test, but the `split(",")` may lead to an unexpected result.

zas, line 31: What I suggest is testing this case:

```suggestion
            PluginsOptionsPage.link_authors('Wile E. Coyote <coyote@acme.com>, Road <Runner>, "a,b"<x@y>'),
```

Is it a valid case?

zas: Yes, it would imply much complicated code, so let's not bother with that.

zas (approved):

PR #2118, opened 2022-06-09
PICARD-2480: Detect and ignore data tracks in EAC logs

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2480

Picard calculates wrong disc IDs for EAC logs from Enhanced CDs (with trailing data track).

Also refactor the code to use a `TocEntry`  named tuple.

# Solution

Check the gap between the last track and the track before. If the gap matches 11400 sectors as specified for Enhanced CDs (compare https://en.wikipedia.org/wiki/Enhanced_CD) the last track is considered a data track and removed from the calculation.


zas (approved):
LGTM

PR #2120, opened 2022-06-20
PICARD-2504: Increase default query limit and make it configurable

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2504

Currently Picard uses the MB default query limit of 25 for all web service requests. That means it considers at most 25 results. As Picard applies it's own filtering, which can be adjusted e.g. by "preferred releases" settings) this means the selection is always limited to 25 items.

It's quite possible the best match after applying the weights (e.g. for release types) is after the first 25 results. Consider increasing the query limit by default, and maybe make it configurable.

# Solution

Make the query limit configurable in Options > Advanced and increase the default from 25 to 50


phw:
I intentionally left the `QUERY_LIMIT` constant in case it is being imported and used by some plugin.

phw:
Yes, thanks. I added it to the option profiles

ui/options_advanced.ui
```diff
@@ -20,13 +20,65 @@
...
+        <item>
+         <widget class="QLabel" name="label_query_limit">
+          <property name="sizePolicy">
+           <sizepolicy hsizetype="Expanding" vsizetype="Preferred">
+            <horstretch>0</horstretch>
+            <verstretch>0</verstretch>
+           </sizepolicy>
+          </property>
+          <property name="text">
+           <string>Max. number of entities to return per MusicBrainz query</string>
```
rdswift, line 50: Minor point, but should this be "Maximum number of entries..." rather than using the abbreviation?  This should not impact the width of the dialog because it will still be shorter than the previous label (I think)?

ui/options_advanced.ui
```diff
@@ -20,13 +20,65 @@
...
+        <item>
+         <widget class="QSpinBox" name="query_limit">
+          <property name="accelerated">
+           <bool>true</bool>
+          </property>
+          <property name="minimum">
+           <number>1</number>
+          </property>
+          <property name="maximum">
+           <number>100</number>
```
zas, line 63: I'm not sure about those min/max values.
A value of 1 doesn't look very useful. What are the drawbacks when setting a value above 100?

I'd suggest we use a low value of 10 (it used to be 25), and add a tool tip to explain the drawbacks of setting the value too high or too low.

Also do we really need a granularity of 1 for this ? Not sure 30 or 31 do much difference. A step of 5 may be enough, dunno.

zas (commented):
Overall it looks good.

ui/options_advanced.ui
rdswift: Does MB have a maximum number that it will honor?  If so then that might be the maximum value allowed in this option.  I agree that a granularity of 5 might be more reasonable.

phw: 1 and 100 are the minimum/maximum for the MB API, if you set it outside this range it'll fallback to the default (which is 25). So we should not allow the user to set such nonsense.

I agree that setting 50 or 51 doesn't make any real difference, but I also don't see any reason why we should make this more complicated than it is and artificially limit the values.

phw: The step approach really would be weird, especially with a limit of 1. The step just changes what the arrow buttons do, but not the selectable choices, and since it is an input you can enter every number you want (inside the range). We could handle that, but it starts getting really confusing if the values e.g. get rounded after input. It just looks random and arbitrary for the user.

But another more radical suggestion: We make it a dropdown and even more strictly limit it to a choice of just 25, 50, 75 and 100. That fulfills the main concern to give the user the ability to increase the default limit and avoids changing it to values that don't make sense (I think below 25 it starts getting pointless already). What do you think? ec8c71ab63e7ca6e3579cd045e7a842abcc07a94

grafik

picard/const/__init__.py
```diff
@@ -142,7 +142,7 @@
 }
 
 # Default query limit
-QUERY_LIMIT = 25
+QUERY_LIMIT = 50
```
zas, line 145: May be rename it to `DEFAULT_QUERY_LIMIT` and drop comment above (or replace it with an explanation).

zas (changes requested):

picard/const/__init__.py
phw: I think than we can also drop the constant alltogether, I just kept it in case someone uses it for own requests in custom plugins.

zas (approved):

PR #2122, opened 2022-06-25
timeout added to pytest

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
Use timeout in workflows for unit tests to prevent tests from going into an infinite loop. Admins won't need to cancel tests themselves, contributors will receive a quicker feedback.

# Problem
My GSoC PR made me think that timeout in the tests is a useful feature so here it is :)

# Solution
`timeout-minutes` set to 30, the exact value is a subject of discussion.

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -82,7 +82,7 @@ jobs:
       env:
         DEPENDENCIES: ${{ matrix.dependencies }}
     - name: Test with pytest
-    	timeout-minutes: 30
+      timeout-minutes: 30
       run: |
         pip install pytest pytest-randomly pytest-cov
         pytest --verbose test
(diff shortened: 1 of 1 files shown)


.github/workflows/run-tests.yml
```diff
@@ -81,6 +82,7 @@ jobs:
       env:
         DEPENDENCIES: ${{ matrix.dependencies }}
     - name: Test with pytest
+    	timeout-minutes: 30
```
phw, line 85: Can you fix the indentation here?

phw (changes requested):
Yes, that makes sense. The timeout AFAIK also only applies on inactivity (no console output), so even if we'd had a test suit that overall ran longer it should still work.

Let's merge that after fix of indentation

.github/workflows/run-tests.yml
skelly37: Yeah, thanks for the noticing, should be ok now

phw (approved):
Thanks a lot!

PR #2124, opened 2022-06-26
PICARD-2506: Fixed calling fpcalc with long path filename

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2506

Calls to fpcalc on Windows 10 / 11 fail, if the path to the audio file exceeds the 259 character path name limit and Windows is configured for long path support. The reason is that on Windows with long path support Picard keeps the path as is. But fpcalc (due to ffmpeg, see https://trac.ffmpeg.org/ticket/8885) does not support long paths yet and hence fails.

The solution is to enforce the Windows long path prefix `\\?\` whenever passing paths to fpcalc.

# Solution

On Windows always add the long path prefix \\?\ if the path exceeds this limit, independent of the OS configuration for long path supports. This fixes ffmpeg currently not being able to handle long paths.

Added util method win_prefix_longpath() to help with that.

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2017-2018 Sambhav Kothari
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2018-2021 Laurent Monin
-# Copyright (C) 2018-2021 Philipp Wolfer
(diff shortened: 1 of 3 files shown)


zas (approved):
LGTM

PR #2125, opened 2022-06-26
PICARD-2508: Unexpected script execution results when using file naming script editor

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Use a copy of the original file metadata to avoid processing multiple script changes when using the file naming script editor.

# Problem

Tagging scripts are executed whenever updating the file naming examples in the file naming script editor.  This could result in changes being applied multiple times unexpectedly.

* JIRA ticket (_optional_): PICARD-2508

# Solution

Use a copy of the original file metadata to avoid processing multiple script changes when using the file naming script editor.

# Action

None.

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 #
-# Copyright (C) 2021 Bob Swift
+# Copyright (C) 2021-2022 Bob Swift
 # Copyright (C) 2021 Laurent Monin
 # Copyright (C) 2021-2022 Philipp Wolfer
 #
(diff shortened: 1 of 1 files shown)


rdswift:
Note that the preview in the script editor still shows the changes applied twice.  The ideal way to resolve this would be to use a copy of the metadata after retrieved from MusicBrainz but before any tagging scripts have been applied, but I'm not sure if this is available anywhere.

rdswift:
Actually, looking at this further, I'm wondering if we need to execute the tagging scripts at all when updating the examples?

phw (approved):
Thanks a lot for finding and fixing this.

PR #2126, opened 2022-06-28
Timeouts added to all the unit tests

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Seems like I've missed quite a lot of tests in my previous commit so here's the full refactored codebase (I really hope I haven't missed anything this time).

# Problem

Even if tests fail, they might get stuck and won't end themselves (despite timeouts in code) after Errors, so we have to create shorter timeouts in the workflow files.

# Solution

Timeouts added

# Action

Nope

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -120,7 +120,7 @@ jobs:
       run: picard --long-version --no-crash-dialog
     - name: Verify sdist package
       if: runner.os != 'Windows'
-      timeout-minutes: 40
+      timeout-minutes: 30
       run: |
         pip install pytest
         scripts/package/run-sdist-test.sh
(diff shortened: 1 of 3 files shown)


.github/workflows/run-tests.yml
```diff
@@ -120,6 +120,7 @@ jobs:
       run: picard --long-version --no-crash-dialog
     - name: Verify sdist package
       if: runner.os != 'Windows'
+      timeout-minutes: 40
```
phw, line 123: Any specific reason for the 40 here?

phw (approved):
I guess this is fine. That should be plenty of time and should help us finish the run earlier in case of endless loops and blocking issues. And if a run exceeds this time for some reason outside our control it would be very unusual and we can restart the process manual.

.github/workflows/run-tests.yml
phw: Let's make it 30 anyway, keep it consistent. It should still be under 10 minutes (and that's long already), so 30 is fine

zas (approved):

PR #2128, opened 2022-07-05
PICARD-2515: Ensure color chooser buttons can be styled on macOS

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2515

On macOS the style override prevents styling these buttons. Hence explicitly reset the style for these widgets only.

# Solution

# Action

### DIFF
--- a/picard/ui/options/interface_colors.py
+++ b/picard/ui/options/interface_colors.py
@@ -29,6 +29,7 @@ from PyQt5 import (
 )
 
 from picard.config import Option
+from picard.const.sys import IS_MACOS
 
 from picard.ui.colors import (
     InterfaceColors,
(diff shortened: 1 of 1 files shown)


phw:
@zas This is ready know. I think this is the best approach to fix this. In theory the style override should not have any effect on the rendering, but for some reason it breaks these buttons. Resetting the style to the default "macintosh" (in Qt6 it will be called "macos") for the buttons only fixes the issue.

This change also changes the way the style override is applied. As it is only needed on macOS (to fix PICARD-2417) I limited it to this platform, to prevent similar unintended side effects to happen elsewhere.

zas (approved):
LGTM

PR #2130, opened 2022-07-20
PICARD-2519: Allow passing supported URLs on command line

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Currently Picard supports passing file paths on command line to open the respective files. Picard also supports handling some URL patterns by drag and drop, e.g. MusicBrainz recordings, releases, release group and cdtoc URLs.

Extend Picard to also handle these supported URLs when passed via command line.

# Problem

* JIRA ticket (_optional_): PICARD-2519

# Solution
call urllib.parse.urlparse on each argument marked as FILE to determine whether it's a file or URL, then pass it to the actual Tagger.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -197,7 +197,7 @@ class Tagger(QtWidgets.QApplication):
         config = get_config()
         theme.setup(self)
 
-        self._cmdline_files = picard_args.FILE
+        self._cmdline_files = picard_args.FILE_OR_URL
         self.autoupdate_enabled = autoupdate
         self._no_restore = picard_args.no_restore
         self._no_plugins = picard_args.no_plugins
@@ -1097,7 +1097,7 @@ If a new instance will not be spawned:
                         help="display version information and exit")
     parser.add_argument("-V", "--long-version", action='store_true',
                         help="display long version information and exit")
-    parser.add_argument('FILE', nargs='*')
+    parser.add_argument('FILE_OR_URL', nargs='*')
 
     return parser.parse_known_args()[0]
 
@@ -1139,7 +1139,7 @@ def main(localedir=None, autoupdate=True):
     }
 
     to_be_added = []
-    for x in picard_args.FILE:
+    for x in picard_args.FILE_OR_URL:
         if not urlparse(x).netloc:
             x = os.path.abspath(x)
         to_be_added.append(x)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -325,33 +325,41 @@ class Tagger(QtWidgets.QApplication):
             if messages:
                 self.load_to_picard(messages)
 
-    def load_to_picard(self, items):
-        files = set()
-        mbids = set()
-        urls = set()
+    @staticmethod
+    def _parse_file_or_url_items(items):
+        out = {
+            "files": set(),
+            "mbids": set(),
+            "urls": set(),
+        }
 
         for item in items:
             parsed = urlparse(item)
             if not parsed.scheme:
-                files.add(item)
+                out["files"].add(item)
             elif parsed.scheme == "file":
                 # remove file:// prefix safely
-                files.add(item[7:])
+                out["files"].add(item[7:])
+            elif parsed.scheme == "mbid":
+                out["mbids"].add(parsed.netloc + parsed.path)
             elif parsed.scheme in {"http", "https"}:
                 # .path returns / before actual link
-                urls.add(parsed.path[1:])
-            elif parsed.scheme == "mbid":
-                mbids.add(parsed.netloc + parsed.path)
+                out["urls"].add(parsed.path[1:])
 
-        if files:
-            self.add_paths(files)
+        return out
+
+    def load_to_picard(self, items):
+        parsed_items = self._parse_file_or_url_items(items)
+
+        if parsed_items["files"]:
+            self.add_paths(parsed_items["files"])
             self.bring_tagger_front()
-        if urls or mbids:
+        if parsed_items["urls"] or parsed_items["mbids"]:
             file_lookup = self.get_file_lookup()
-            for url in urls:
-                thread.to_main(file_lookup.mbid_lookup, url, None, None, False)
-            for mbid in mbids:
+            for mbid in parsed_items["mbids"]:
                 thread.to_main(file_lookup.mbid_lookup, mbid, None, None, False)
+            for url in parsed_items["urls"]:
+                thread.to_main(file_lookup.mbid_lookup, url, None, None, False)
             self.bring_tagger_front()
 
     def enable_menu_icons(self, enabled):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -334,7 +334,7 @@ class Tagger(QtWidgets.QApplication):
             if not parsed.scheme:
                 files.append(item)
             elif parsed.scheme == "file":
-                files.append(item.replace("file://", ''))
+                files.append(parsed.path)
             elif parsed.scheme in {"http", "https"}:
                 # .path returns / before actual link
                 urls.append(parsed.path[1:])
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -332,7 +332,7 @@ class Tagger(QtWidgets.QApplication):
             if not parsed.netloc:
                 files.append(item)
             else:
-                # .path returns / before actual cdtoc
+                # .path returns / before actual link
                 urls.append(parsed.path[1:])
 
         if files:
@@ -341,7 +341,7 @@ class Tagger(QtWidgets.QApplication):
             print(urls)
             file_lookup = self.get_file_lookup()
             for url in urls:
-                file_lookup.mbid_lookup(url, browser_fallback=False)
+                thread.to_main(file_lookup.mbid_lookup, url, None, None, False)
 
     def enable_menu_icons(self, enabled):
         self.setAttribute(QtCore.Qt.ApplicationAttribute.AA_DontShowIconsInMenus, not enabled)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -142,7 +142,6 @@ from picard.webservice.api_helpers import (
 import picard.resources  # noqa: F401 # pylint: disable=unused-import
 
 from picard.ui import theme
-from picard.ui.itemviews import BaseTreeView
 from picard.ui.mainwindow import MainWindow
 from picard.ui.searchdialog.album import AlbumSearchDialog
 from picard.ui.searchdialog.artist import ArtistSearchDialog
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -339,7 +339,7 @@ class Tagger(QtWidgets.QApplication):
                 # .path returns / before actual link
                 urls.append(parsed.path[1:])
             elif parsed.scheme == "mbid":
-                mbids.append(item.replace("mbid://", ''))
+                mbids.append(parsed.netloc + parsed.path)
 
         if files:
             self.add_paths(files)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1140,13 +1140,13 @@ def main(localedir=None, autoupdate=True):
         picard_args.stand_alone_instance,
     }
 
-    to_be_added = []
-    for x in picard_args.FILE_OR_URL:
-        if not urlparse(x).netloc:
-            x = os.path.abspath(x)
-        to_be_added.append(x)
-
     if not should_start:
+        to_be_added = []
+        for x in picard_args.FILE_OR_URL:
+            if not urlparse(x).netloc:
+                x = os.path.abspath(x)
+            to_be_added.append(x)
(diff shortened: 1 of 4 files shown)


skelly37:
I think that handling the `QObject::startTimer: Timers cannot be started from another thread` will make it actually done.

skelly37:
Okay, I've found out that `picard.webservice.WebService.add_task` is responsible for raising `QObject::startTimer` errors twice. 

Seems like the WebService was designed to be called by the main thread, while the pipe listener runs on another one, so it cannot start the actual task (and even breaks the currently working drag'n'drop feature). I mean, drag'n'drop works until we try to pass some URL via command-line to an existing instance.

skelly37:
@zas @phw So basically I need help from the more proficient in Qt one of you

rdswift:
> Seems like the WebService was designed to be called by the main thread, while the pipe listener runs on another one, so it cannot start the actual task (and even breaks the currently working drag'n'drop feature).

I wonder if you could accomplish this using a signal?  According to [this article on setting handlers for asynchronous events](https://docs.python.org/3/library/signal.html), "***Python signal handlers are always executed in the main Python thread of the main interpreter, even if the signal was received in another thread.***"

There are also a couple of threads on Stack Overflow ([Python signals between threads](https://stackoverflow.com/questions/6855984/python-signals-between-threads) and [Emit signal in standard python thread](https://stackoverflow.com/questions/19078249/emit-signal-in-standard-python-thread)) that might provide some insights.

phw:
> I wonder if you could accomplish this using a signal?

Indeed that is probably the way, and Picard has a helper for that. You can use the `to_main` function from `picard.util.thread` to run a task on the main thread.



> Album, Artist (?), Track. The only thing with just entering the MBID is that Picard won't know which type of record so it may have to try as many as three calls to the API to get a response. Perhaps the command line entry could (should) include a type identifier along with the MBID, like "Album:1fa3b874-dbef-4c86-9681-3f6d6877ddba"?

I would limit it to URLs for now, otherwise it can get confused with normal valid local paths. If we would introduce MBIDs, they should be with identify identifier in the form `release:{mbid}`, `recording:{mbid}` or `artist:{mbid}`. This is the short form of the full MB URLs, which is in this form already supported.

skelly37:
I tend to agree with @phw on extending the current functionality to just URLs. The GUI automations we've discussed on IRC seem to be more useful than @rdswift's suggestion imo, especially because MBID isn't something that a user would enter like other CLI arguments.

skelly37:
Also thanks for the decent feedback, gonna read the articles and mentioned methods :)

rdswift:
> I would limit it to URLs for now, otherwise it can get confused with normal valid local paths.

Okay.  Perhaps somthing to consider in the future to better facilitate batch processing of an automated rip -> tag workflow.

> If we would introduce MBIDs, they should be with identify identifier in the form `release:{mbid}`, `recording:{mbid}` or `artist:{mbid}`.

Agreed.

zas:
> > 
> 
> I would limit it to URLs for now, otherwise it can get confused with normal valid local paths. If we would introduce MBIDs, they should be with identify identifier in the form `release:{mbid}`, `recording:{mbid}` or `artist:{mbid}`. This is the short form of the full MB URLs, which is in this form already supported.

What about allowing URLs in the form of:

```
mbid://<entity-type>/<mbid>
```

Examples:

```
mbid://release/dbd0ce67-cae6-33eb-8f5a-1143a30c2353
mbid://artist/83d91898-7763-47d7-b03b-b92132375c47
```

skelly37:
> > > 
> > 
> > 
> > I would limit it to URLs for now, otherwise it can get confused with normal valid local paths. If we would introduce MBIDs, they should be with identify identifier in the form `release:{mbid}`, `recording:{mbid}` or `artist:{mbid}`. This is the short form of the full MB URLs, which is in this form already supported.
> 
> What about allowing URLs in the form of:
> 
> ```
> mbid://<entity-type>/<mbid>
> ```

Hm, it could be easily parsed by `urllib.parse.urlparse` to separate MBID from HTTP(S). Worth considering.

skelly37:
Okay, I've added the "mbid://" option, you might want to review the code now :)

picard/tagger.py
```diff
@@ -321,8 +322,35 @@ def pipe_server(self):
...
+        files = []
+        mbids = []
+        urls = []
+
+        for item in items:
+            parsed = urlparse(item)
+            if not parsed.scheme:
+                files.append(item)
+            elif parsed.scheme == "file":
+                files.append(item.replace("file://", ''))
```
zas, line 337:
```suggestion
                files.append(parsed.path)
```

picard/tagger.py
```diff
@@ -321,8 +322,35 @@ def pipe_server(self):
...
+            elif parsed.scheme in {"http", "https"}:
+                # .path returns / before actual link
+                urls.append(parsed.path[1:])
+            elif parsed.scheme == "mbid":
+                mbids.append(item.replace("mbid://", ''))
```
zas, line 342:
```suggestion
                mbids.append(parsed.netloc + parsed.path)
```

picard/tagger.py
```diff
@@ -321,8 +322,35 @@ def pipe_server(self):
...
             messages = [x for x in self.pipe_handler.read_from_pipe() if x not in IGNORED]
             if messages:
-                self.add_paths(messages)
-                self.bring_tagger_front()
+                self.load_to_picard(messages)
+
+    def load_to_picard(self, items):
+        files = []
+        mbids = []
+        urls = []
```
zas, line 330: I think we could uniquify arguments here, using sets instead of lists.

zas (changes requested):

picard/tagger.py
zas: Hmmm, actually a bad idea, as it may strip "#" or "?"

picard/tagger.py
```diff
@@ -334,7 +334,7 @@ def load_to_picard(self, items):
             if not parsed.scheme:
                 files.append(item)
             elif parsed.scheme == "file":
-                files.append(item.replace("file://", ''))
+                files.append(parsed.path)
```
zas: Actually a bad idea, sorry.
It will fail with paths like "file:///some#thing"

zas (changes requested):

picard/tagger.py
skelly37: Okay, better idea: 

`item[7:]`

picard/tagger.py
```diff
@@ -1110,9 +1138,15 @@ def main(localedir=None, autoupdate=True):
...
+    to_be_added = []
+    for x in picard_args.FILE_OR_URL:
+        if not urlparse(x).netloc:
+            x = os.path.abspath(x)
+        to_be_added.append(x)
+
     if not should_start:
         try:
-            pipe_handler = pipe.Pipe(app_name=PICARD_APP_NAME, app_version=PICARD_FANCY_VERSION_STR, args=[os.path.abspath(x) for x in picard_args.FILE])
+            pipe_handler = pipe.Pipe(app_name=PICARD_APP_NAME, app_version=PICARD_FANCY_VERSION_STR, args=to_be_added)
```
zas, line 1149: `to_be_added` is only used here. We should move the preprocessing code above here too.

picard/tagger.py
```diff
@@ -321,8 +322,36 @@ def pipe_server(self):
...
+            if not parsed.scheme:
+                files.add(item)
+            elif parsed.scheme == "file":
+                # remove file:// prefix safely
+                files.add(item[7:])
+            elif parsed.scheme in {"http", "https"}:
+                # .path returns / before actual link
+                urls.add(parsed.path[1:])
+            elif parsed.scheme == "mbid":
+                mbids.add(parsed.netloc + parsed.path)
```
zas, line 343: To ease testing and to keep functions short, it would be preferable if this part of the code was in its own method.

```python
@staticmethod
def _parse_file_or_url_items(items):
    files = set()
    mbids = set()
    urls = set()

    ....

    return files, mbids, urls

def load_to_picard(self, items):
    files, mbids, urls = self._parse_file_or_url_items(items)
```

Then we can add tests for this method alone.

zas (changes requested):

picard/tagger.py
skelly37: Seems nice, thanks. Though I'd rather return a dict instead of `return files, mbids, urls`. We're planning to parse commands as well in the future, four returned values seem to be more error-prone and overcomplicated.

picard/tagger.py
```diff
@@ -325,33 +325,41 @@ def pipe_server(self):
...
-        files = set()
-        mbids = set()
-        urls = set()
+    @staticmethod
+    def _parse_file_or_url_items(items):
+        out = {
+            "files": set(),
+            "mbids": set(),
+            "urls": set(),
+        }
```
zas, line 334: Alternatively, we could use a class, see https://github.com/skelly37/picard/pull/3

zas (approved):
LGTM

picard/tagger.py
```diff
@@ -322,8 +344,21 @@ def pipe_server(self):
...
+
+        if parsed_items.files:
+            self.add_paths(parsed_items.files)
+            self.bring_tagger_front()
+        if parsed_items.urls or parsed_items.mbids:
+            file_lookup = self.get_file_lookup()
+            for mbid in parsed_items.mbids:
+                thread.to_main(file_lookup.mbid_lookup, mbid, None, None, False)
+            for url in parsed_items.urls:
+                thread.to_main(file_lookup.mbid_lookup, url, None, None, False)
```
rdswift, line 360: Minor point, but since we're dealing with sets here I think this could be shortened a bit by using the combined sets as the iterator (unless you want to keep the duplicated code for clarity).  I haven't tested this, but I think the following should work:

```suggestion
            for item in parsed_items.mbids.union(parsed_items.urls):
                thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
```

zas: I tend to prefer the set union using the `|` operator which is easier to read to me, dunno which one you prefer:

```suggestion
            for item in parsed_items.mbids | parsed_items.urls:
                thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
```

rdswift: You're right.  That is easier to read.  I actually prefer it the way you have it written.

skelly37: I'm okay with that

picard/tagger.py
```diff
@@ -1071,7 +1103,7 @@ def process_picard_args():
                         help="display version information and exit")
     parser.add_argument("-V", "--long-version", action='store_true',
                         help="display long version information and exit")
-    parser.add_argument('FILE', nargs='*')
+    parser.add_argument('FILE_OR_URL', nargs='*')
```
zas, line 1106: Please add an help

skelly37: I've copied it from my docs PR

picard/tagger.py
```diff
@@ -1071,7 +1103,8 @@ def process_picard_args():
                         help="display version information and exit")
     parser.add_argument("-V", "--long-version", action='store_true',
                         help="display long version information and exit")
-    parser.add_argument('FILE', nargs='*')
+    parser.add_argument('FILE_OR_URL', nargs='*',
+                        help="the file(s), URL(s) and MBID(s) to load")
```
rdswift, line 1107: Should we include a brief description of the formats of the inputs? Most people will probably understand the files and URLs, but not the required format for MBIDs.

skelly37: I don’t think so. The help messages should be short and substantive. It should be documented in some place (my suggestion is picard docs and maybe the official wiki article about MBID) but not in --help imo.

zas (approved):
LGTM

picard/tagger.py
```diff
@@ -322,8 +344,19 @@ def pipe_server(self):
...
+        parsed_items = ParseItemsToLoad(items)
+
+        if parsed_items.files:
+            self.add_paths(parsed_items.files)
+            self.bring_tagger_front()
+        if parsed_items.urls or parsed_items.mbids:
+            file_lookup = self.get_file_lookup()
+            for item in parsed_items.mbids | parsed_items.urls:
+                thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
+            self.bring_tagger_front()
```
zas, line 359: I was bothered by calling `bring_tagger_front()` twice, so I added `ParseItemsToLoad.__bool__()` in  https://github.com/skelly37/picard/pull/4
What do you think?

zas (changes requested):

picard/tagger.py
skelly37: Good idea, thanks, merged.

zas (approved):

phw (approved):

phw:
Works great, thanks a lot

PR #2131, opened 2022-07-21
PICARD-2524: prevent crashes when doing multiple searches while one i…

zas:
…s running

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2524

# Solution

# Action

### DIFF
--- a/picard/ui/searchdialog/album.py
+++ b/picard/ui/searchdialog/album.py
@@ -62,6 +62,7 @@ class CoverWidget(QtWidgets.QWidget):
     def __init__(self, parent, width=100, height=100):
         super().__init__(parent)
         self.layout = QtWidgets.QVBoxLayout(self)
+        self.destroyed.connect(self.invalidate)
         self.layout.setContentsMargins(0, 0, 0, 0)
         self.layout.setAlignment(QtCore.Qt.AlignmentFlag.AlignCenter)
         self.loading_gif_label = QtWidgets.QLabel(self)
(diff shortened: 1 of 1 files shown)


picard/ui/searchdialog/album.py
```diff
@@ -96,6 +99,12 @@ def showEvent(self, event):
         super().showEvent(event)
         self.shown.emit()
 
+    def invalidate(self):
+        self.layout = None
+
+    def __del__(self):
+        self.shown.disconnect()
```
phw, line 106: Why is this needed? This should happen automatically. Or was there some issue still producible without this change?.

picard/ui/searchdialog/album.py
```diff
@@ -62,6 +62,7 @@ class CoverWidget(QtWidgets.QWidget):
     def __init__(self, parent, width=100, height=100):
         super().__init__(parent)
         self.layout = QtWidgets.QVBoxLayout(self)
+        self.layout.destroyed.connect(self.invalidate)
```
phw, line 65: I think this could also connect to `self.destroyed`, as it is the widget itself that primarily is the one getting removed. Would have the same effect, though.

phw (commented):
Yes, I think this is generally the right approach. It's also what we do in Option pages if there are async tasks modifying UI that might no longer exists at the time the callback gets executed.

picard/ui/searchdialog/album.py
zas: It was a leftover from experiments, unneeded.

picard/ui/searchdialog/album.py
zas: Yup, done.

phw (approved):

PR #2132, opened 2022-07-22
Use macos-11 image for Github actions, upgrade to latest PyInstaller

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

- Github has deprecated the macOS-10.15 package and will completely remove support for it soon, see https://github.com/actions/virtual-environments/issues/5583
- macOS packaging is currently failing with some issue building pyinstaller, see e.g. https://github.com/metabrainz/picard/runs/7447653204 . It's unclear whether this is related to the macOS-10.15 deprecation

# Todo

Check whether the build packages are still usable on older macOS.

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -66,7 +66,8 @@ jobs:
         cd pyinstaller/bootloader
         python3 ./waf --verbose all
         cd ..
-        pip3 install .
+        pip3 install -r requirements.txt
+        python3 setup.py install
       env:
         PYINSTALLER_VERSION: v5.2
         CFLAGS: -mmacosx-version-min=${{ matrix.macos-deployment-version }}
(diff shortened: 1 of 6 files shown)


phw:
@zas This will pass the tests again, and from what I can tell it produces working packages.

It contains a bit more than I originally wanted:

- It updates the macOS image for GHA, but this did not solve the build errors. But as we need to do this anyway before end of August we can do this here.
- It fixes the build error. I think some setuptools related update caused this, but I could not exactly pin it down. Anyway, https://github.com/metabrainz/picard/pull/2132/commits/35bdb23caeb8133063f10dc77b5253bc44409e29 fixes the build error, by following the instructions on https://pyinstaller.org/en/stable/installation.html#installing-from-the-archive instead of using `pip install`.
- I generally updated PyInstaller to latest version, as we should do this anyway.

As far as I could tell we might still have builds that run as before. But I think we should release a Picard 2.8.3 (with at least the fix in #2131 and maybe others) before end of August, using the old GHA image for macOS, just to be safe.

zas (approved):
LGTM

PR #2133, opened 2022-07-22
exit patch

skelly37:
# Summary

Exit fails when there's no pipe handler

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem

* JIRA ticket (_optional_): None

# Solution

`tagger.pipe_handler` now defaults to `None` and is None-checked in the `tagger.exit()`

# Action

@phw  here you are :)

PR #2135, opened 2022-07-22
PICARD-2521: Cannot exit properly after removing the pipe file

skelly37:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

The single instance mode via named pipes got merged and works almost perfectly. Almost... because when the pipe file gets deleted (outside of the Picard, e.g. `rm`), Picard cannot exit properly and gets stuck during the exit function's call.

# Problem

* JIRA ticket (_optional_): PICARD-2521

# Solution

1) save the tagger's exit code
2) check whether the pipe file was removed unexpectedly (new `Pipe` attribute)
- if yes: `os._exit(exit_code)`
- if no: `sys.exit(exit_code)`

# Action

(https://github.com/metabrainz/picard/pull/2133) included since they're both about fixing the exit process, so they could be in a single PR

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -46,6 +46,7 @@
 
 import argparse
 from concurrent.futures import ThreadPoolExecutor
+from concurrent.futures._base import TimeoutError
 from functools import partial
 import itertools
 import logging
@@ -209,12 +210,14 @@ class Tagger(QtWidgets.QApplication):
         # Default thread pool
         self.thread_pool = ThreadPoolExecutor()
 
+        self.pipe_pool = ThreadPoolExecutor()
+        self.pipe_handler = pipe_handler
+
         # if the instance is forced, we get None instead of an actual handler
         # even though there's always something provided as pipe_handler, I created a default argument to make it more obvious
-        if pipe_handler:
-            self.pipe_handler = pipe_handler
+        if self.pipe_handler:
             self.pipe_handler.pipe_running = True
-            self.thread_pool.submit(self.pipe_server)
+            self.pipe_pool.submit(self.pipe_server)
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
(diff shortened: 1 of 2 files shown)


picard/tagger.py
```diff
@@ -426,6 +426,7 @@ def exit(self):
         self.webservice.stop()
         self.run_cleanup()
         QtCore.QCoreApplication.processEvents()
+        os._exit(0)
```
zas, line 429: I'm a bit bothered by this "solution" because it causes the program to exit immediately, without flushing buffers for example.
`exit()` is called from https://github.com/skelly37/picard/blob/5b453543c6ce3a93a84b995ccb9a0148305f67a0/picard/tagger.py#L1025 but we can see we expect some extra code to run after its call.

Also we expect to return an exit code on exit. See https://github.com/skelly37/picard/blob/5b453543c6ce3a93a84b995ccb9a0148305f67a0/picard/tagger.py#L1150 and https://github.com/skelly37/picard/blob/5b453543c6ce3a93a84b995ccb9a0148305f67a0/picard/tagger.py#L443

zas (changes requested):

picard/tagger.py
skelly37: That's true, the only solution I see for now is to replace 

`sys.exit(tagger.run())` with `os._exit(tagger.run())`

So the question is whether it is more important to prevent Picard from freezing or to exit nicely.

skelly37:
This way we keep the current logic as it should be: with `sys.exit`.

But if there's something left in the separate pipe pool we can run force quit.

picard/tagger.py
skelly37: True, I misread the code. The "dirty" exit should happen as rarely as possible. It's okay now imo.

skelly37:
Actually, it won't work as intended. Threads are re-used.

zas (approved):
LGTM

phw (approved):

PR #2137, opened 2022-07-31
PICARD-2526: Allow starting processing actions from the command line

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Users will be able to pass some commands to the existing instance with `-e`/`--exec` flag.

# Problem

It would be useful to tell the existing Picard instance what to do, e.g. QUIT, useful in combination with the other single-instance mode improvements

* JIRA ticket (_optional_): PICARD-2526

# Solution

- add command:// prefix to the commands
- send them as any other arg
- parse with urlparse, split by ";"
- execute

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -374,6 +374,11 @@ class Tagger(QtWidgets.QApplication):
         if parsed_items.non_executable_items():
             self.bring_tagger_front()
 
+    def get_album_pane_tracks(self):
+        for album in self.albums.values():
+            for track in album.iterfiles():
+                yield track
+
     def enable_menu_icons(self, enabled):
         self.setAttribute(QtCore.Qt.ApplicationAttribute.AA_DontShowIconsInMenus, not enabled)
 
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -256,9 +256,9 @@ REMOTE_COMMANDS = {
         "handle_command_remove_saved",
         help_text="Remove all saved releases from the album pane.",
     ),
-    "SAVE_COMPLETE": RemoteCommand(
-        "handle_command_save_complete",
-        help_text="Save all completely matched releases in the album pane.",
+    "SAVE_MODIFIED": RemoteCommand(
+        "handle_command_save_modified",
+        help_text="Save all modified files from the album pane.",
     ),
     "SCAN": RemoteCommand(
         "handle_command_scan",
@@ -519,7 +519,7 @@ class Tagger(QtWidgets.QApplication):
             if track.state == File.NORMAL:
                 track.remove()
 
-    def handle_command_save_complete(self, argstring):
+    def handle_command_save_modified(self, argstring):
         for track in self.get_album_pane_tracks():
             if track.state == File.CHANGED:
                 track.save()
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1217,6 +1217,9 @@ List of the commands available to execute in Picard from the command-line:
                                       files in the album pane
 
 - Commands are case insensitive (not args!), i.e. SHOW == show == ShOw
+- If there is no running instance, picard -e blocks Picards startup and does nothing
+- If there is an existing instance, picard -e will send all the positional arguments in the given order.
+  Then the commands. So the files passed to the new instance with a command will be processed as well.
 - Arguments are optional and not providing them will not cause Picard to crash
   - LOOKUP_CD without the argument will default to the first (alphabetically) available disc drive
   - REMOVE without the argument will not do anything
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -178,6 +178,7 @@ def plugin_dirs():
 class ParseItemsToLoad:
 
     def __init__(self, items):
+        self.commands = []
         self.files = set()
         self.mbids = set()
         self.urls = set()
@@ -186,6 +187,9 @@ class ParseItemsToLoad:
             parsed = urlparse(item)
             if not parsed.scheme:
                 self.files.add(item)
+            elif parsed.scheme == "command":
+                for x in item[10:].split(';'):
+                    self.commands.append(x.strip())
             elif parsed.scheme == "file":
                 # remove file:// prefix safely
                 self.files.add(item[7:])
@@ -195,8 +199,15 @@ class ParseItemsToLoad:
                 # .path returns / before actual link
                 self.urls.add(parsed.path[1:])
 
-    def __bool__(self):
+    # needed to indicate whether Picard should be brought to the front
+    def non_executable_items(self):
         return bool(self.files or self.mbids or self.urls)
+
+    def __bool__(self):
+        return bool(self.commands or self.files or self.mbids or self.urls)
+
+    def __str__(self):
+        return "files: %r mbids: %r urls: %r commands: %r" % (self.files, self.mbids, self.urls, self.commands)
 
 
 class Tagger(QtWidgets.QApplication):
@@ -360,7 +371,7 @@ class Tagger(QtWidgets.QApplication):
             for item in parsed_items.mbids | parsed_items.urls:
                 thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
 
-        if parsed_items:
+        if parsed_items.non_executable_items():
             self.bring_tagger_front()
 
     def enable_menu_icons(self, enabled):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1215,7 +1215,10 @@ List of the commands available to execute in Picard from the command-line:
                                       files in the album pane
 
 - Commands are case insensitive (not args!), i.e. SHOW == show == ShOw
-- Arguments do not have to be entered, as they are required only for the two of commands
+- Arguments are optional and not providing them will not cause Picard to crash
+  - LOOKUP_CD without the argument will default to the first (alphabetically) available disc drive
+  - REMOVE without the argument will not do anything
+  - other commands ignore the arguments and act the same regardless of their amount
 - Only one path can be send per one REMOVE command""")
 
 
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -456,7 +456,7 @@ class Tagger(QtWidgets.QApplication):
     def handle_command_remove(self, argstring):
         for file in self.get_all_file_objects():
             if argstring == file.filename:
-                argstring.remove()
+                file.remove()
                 return
 
     def handle_command_remove_all(self, argstring):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -253,6 +253,8 @@ class Tagger(QtWidgets.QApplication):
             self.pipe_handler.pipe_running = True
             self.thread_pool.submit(self.pipe_server)
 
+        self._init_remote_commands()
+
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
         # expects instant feedback instead of waiting for a long list of
@@ -358,10 +360,15 @@ class Tagger(QtWidgets.QApplication):
         while self.pipe_handler.pipe_running:
             messages = [x for x in self.pipe_handler.read_from_pipe() if x not in IGNORED]
             if messages:
-                self.load_to_picard(messages)
+                log.debug("pipe messages: %r", messages)
+                thread.to_main(self.load_to_picard, messages)
 
     def load_to_picard(self, items):
         parsed_items = ParseItemsToLoad(items)
+        log.debug(str(parsed_items))
+
+        for command in parsed_items.commands:
+            self.handle_command(command)
 
         if parsed_items.files:
             self.add_paths(parsed_items.files)
@@ -378,6 +385,60 @@ class Tagger(QtWidgets.QApplication):
         for album in self.albums.values():
             for track in album.iterfiles():
                 yield track
+
+    def _init_remote_commands(self):
+        self.commands = {
+            "CLUSTER": self.handle_command_cluster,
+            "FINGERPRINT": self.handle_command_fingerprint,
+            "LOOKUP": self.handle_command_lookup,
+            "QUIT": self.handle_command_quit,
+            "REMOVE_SAVED": self.handle_command_remove_saved,
+            "SAVE_COMPLETE": self.handle_command_save_complete,
+            "SCAN": self.handle_command_scan,
+            "SHOW": self.handle_command_show,
+            "SUBMIT_FINGERPRINTS": self.handle_command_submit_fingerprints,
+        }
+
+    def handle_command(self, command):
+        try:
+            cmd, *args = command.split(' ', 1)
+            argstring = next(iter(args), "")
+            thread.to_main(self.commands[cmd.upper()], argstring.strip())
+        except KeyError:
+            log.error("Unknown command: %r", command)
+
+    def handle_command_cluster(self, argstring):
+        self.cluster(self.unclustered_files.files)
+
+    def handle_command_fingerprint(self, argstring):
+        for album_name in self.albums:
+            self.analyze(self.albums[album_name].iterfiles())
+
+    def handle_command_lookup(self, argstring):
+        self.autotag(self.unclustered_files.files)
+
+    def handle_command_quit(self, argstring):
+        self.exit()
+        self.quit()
+
+    def handle_command_remove_saved(self, argstring):
+        for track in self.get_album_pane_tracks():
+            if track.state == File.NORMAL:
+                track.remove()
+
+    def handle_command_save_complete(self, argstring):
+        for track in self.get_album_pane_tracks():
+            if track.state == File.CHANGED:
+                track.save()
+
+    def handle_command_scan(self, argstring):
+        self.analyze(self.unclustered_files.files)
+
+    def handle_command_show(self, argstring):
+        self.bring_tagger_front()
+
+    def handle_command_submit_fingerprints(self, argstring):
+        self.acoustidmanager.submit()
 
     def enable_menu_icons(self, enabled):
         self.setAttribute(QtCore.Qt.ApplicationAttribute.AA_DontShowIconsInMenus, not enabled)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -386,12 +386,18 @@ class Tagger(QtWidgets.QApplication):
             for track in album.iterfiles():
                 yield track
 
+    def get_all_file_objects(self):
+        return self.unclustered_files.files + [x for x in self.get_album_pane_tracks()] + [x for x in self.clusters.iterfiles()]
+
     def _init_remote_commands(self):
         self.commands = {
             "CLUSTER": self.handle_command_cluster,
             "FINGERPRINT": self.handle_command_fingerprint,
             "LOOKUP": self.handle_command_lookup,
             "QUIT": self.handle_command_quit,
+            # due to the pipe protocol limitations
+            # we currently can handle only one file per `remove` command
+            "REMOVE": self.handle_command_remove,
             "REMOVE_SAVED": self.handle_command_remove_saved,
             "SAVE_COMPLETE": self.handle_command_save_complete,
             "SCAN": self.handle_command_scan,
@@ -420,6 +426,12 @@ class Tagger(QtWidgets.QApplication):
     def handle_command_quit(self, argstring):
         self.exit()
         self.quit()
+
+    def handle_command_remove(self, argstring):
+        for file in self.get_all_file_objects():
+            if argstring == file.filename:
+                argstring.remove()
+                return
 
     def handle_command_remove_saved(self, argstring):
         for track in self.get_album_pane_tracks():
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -428,14 +428,14 @@ class Tagger(QtWidgets.QApplication):
 
     def handle_command_lookup_cd(self, argstring):
         disc = Disc()
+        devices = get_cdrom_drives()
 
         if not argstring:
-            devices = get_cdrom_drives()
             if devices:
                 device = devices[0]
             else:
                 device = None
-        elif argstring.startswith("/dev/"):
+        elif argstring in devices:
             device = argstring
         else:
             thread.run_task(
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1192,6 +1192,33 @@ def longversion():
     print(versions.as_string())
 
 
+def print_help_for_commands():
+    print("""usage: picard -e [command] [arguments ...]
+or
+picard -e [command 1] [arguments ...] -e [commands 2] [arguments ...]
+
+List of the commands available to execute in Picard from the command-line:
+
+  - CLUSTER                           cluster all files in the cluster pane
+  - FINGERPRINT                       calculate acoustic fingerprints for all (matched) files in
+                                      the album pane
+  - LOOKUP                            lookup all clusters in the cluster pane.
+  - LOOKUP_CD [device/log file]       read CD from the selected drive and lookup on MusicBrainz.
+  - QUIT                              exit the running instance of Picard
+  - REMOVE [absolute path (1 file)]   remove the file from Picard
+  - REMOVE_ALL                        remove all files from Picard
+  - REMOVE_SAVED                      remove all saved releases from the album pane
+  - SAVE_COMPLETE                     save all completely matched releases in the album pane
+  - SCAN                              scan all files in the cluster pane
+  - SHOW                              make the running instance the currently active window
+  - SUBMIT_FINGERPRINTS               submit outstanding acoustic fingerprints for all (matched)
+                                      files in the album pane
+
+- Commands are case insensitive (not args!), i.e. SHOW == show == ShOw
+- Arguments do not have to be entered, as they are required only for the two of commands
+- Only one path can be send per one REMOVE command""")
+
+
 def process_picard_args():
     parser = argparse.ArgumentParser(
         formatter_class=argparse.RawDescriptionHelpFormatter,
@@ -1282,6 +1309,10 @@ def main(localedir=None, autoupdate=True):
 
         if picard_args.exec:
             for e in picard_args.exec:
+                if "help" in [x.lower().strip() for x in e]:
+                    print_help_for_commands()
+                    to_be_added.clear()
+                    break
                 to_be_added.append("command://" + " ".join(e))
 
         if to_be_added:
@@ -1299,8 +1330,8 @@ def main(localedir=None, autoupdate=True):
 
         # pipe has sent its args to existing one, doesn't need to start
         if not should_start:
+            log.debug("No need for spawning a new instance, exiting...")
             # just a custom exit code to show that picard instance wasn't created
-            log.info("No need for spawning a new instance, exiting...")
             sys.exit(EXIT_NO_NEW_INSTANCE)
     else:
         pipe_handler = None
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -649,7 +649,6 @@ class Tagger(QtWidgets.QApplication):
         QtCore.QCoreApplication.processEvents()
 
     def _run_init(self):
-        print(self._to_load)
         if self._to_load:
             self.load_to_picard(self._to_load)
         del self._to_load
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -118,7 +118,6 @@ from picard.track import (
 )
 from picard.util import (
     check_io_encoding,
-    decode_filename,
     encode_filename,
     is_hidden,
     iter_files_from_objects,
@@ -225,7 +224,7 @@ class Tagger(QtWidgets.QApplication):
     _debug = False
     _no_restore = False
 
-    def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
+    def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None, to_load=None):
 
         super().__init__(sys.argv)
         self.__class__.__instance = self
@@ -233,7 +232,8 @@ class Tagger(QtWidgets.QApplication):
         config = get_config()
         theme.setup(self)
 
-        self._cmdline_files = picard_args.FILE_OR_URL
+        self._to_load = to_load
+
         self.autoupdate_enabled = autoupdate
         self._no_restore = picard_args.no_restore
         self._no_plugins = picard_args.no_plugins
@@ -368,9 +368,6 @@ class Tagger(QtWidgets.QApplication):
         parsed_items = ParseItemsToLoad(items)
         log.debug(str(parsed_items))
 
-        for command in parsed_items.commands:
-            self.handle_command(command)
-
         if parsed_items.files:
             self.add_paths(parsed_items.files)
 
@@ -378,6 +375,9 @@ class Tagger(QtWidgets.QApplication):
             file_lookup = self.get_file_lookup()
             for item in parsed_items.mbids | parsed_items.urls:
                 thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
+
+        for command in parsed_items.commands:
+            self.handle_command(command)
 
         if parsed_items.non_executable_items():
             self.bring_tagger_front()
@@ -412,6 +412,7 @@ class Tagger(QtWidgets.QApplication):
         try:
             cmd, *args = command.split(' ', 1)
             argstring = next(iter(args), "")
+            log.debug("Executing command: %r", cmd.upper())
             thread.to_main(self.commands[cmd.upper()], argstring.strip())
         except KeyError:
             log.error("Unknown command: %r", command)
@@ -587,9 +588,10 @@ class Tagger(QtWidgets.QApplication):
         QtCore.QCoreApplication.processEvents()
 
     def _run_init(self):
-        if self._cmdline_files:
-            self.load_to_picard([decode_filename(f) for f in self._cmdline_files])
-            del self._cmdline_files
+        print(self._to_load)
+        if self._to_load:
+            self.load_to_picard(self._to_load)
+        del self._to_load
 
     def run(self):
         self.update_browser_integration()
@@ -1302,9 +1304,8 @@ def main(localedir=None, autoupdate=True):
         picard_args.no_plugins,
         picard_args.stand_alone_instance,
     }
-
+    to_be_added = []
     if not should_start:
-        to_be_added = []
         for x in picard_args.FILE_OR_URL:
             if not urlparse(x).netloc:
                 x = os.path.abspath(x)
@@ -1322,14 +1323,14 @@ def main(localedir=None, autoupdate=True):
             # note: log level isn't defined yet, it defaults to info, log.debug() would not work here
             log.info("Sending messages to main instance: %r" % to_be_added)
 
-        if picard_args.exec is None:
-            try:
-                pipe_handler = pipe.Pipe(app_name=PICARD_APP_NAME, app_version=PICARD_FANCY_VERSION_STR, args=to_be_added)
-                should_start = pipe_handler.is_pipe_owner
-            except pipe.PipeErrorNoPermission as err:
-                log.error(err)
-                pipe_handler = None
-                should_start = True
+        try:
+            pipe_handler = pipe.Pipe(app_name=PICARD_APP_NAME, app_version=PICARD_FANCY_VERSION_STR,
+                                     args=to_be_added)
+            should_start = pipe_handler.is_pipe_owner
+        except pipe.PipeErrorNoPermission as err:
+            log.error(err)
+            pipe_handler = None
+            should_start = True
 
         # pipe has sent its args to existing one, doesn't need to start
         if not should_start:
@@ -1346,7 +1347,7 @@ def main(localedir=None, autoupdate=True):
     except ImportError:
         pass
 
-    tagger = Tagger(picard_args, localedir, autoupdate, pipe_handler=pipe_handler)
+    tagger = Tagger(picard_args, localedir, autoupdate, pipe_handler=pipe_handler, to_load=to_be_added)
 
     # Initialize Qt default translations
     translator = QtCore.QTranslator()
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -459,13 +459,14 @@ class Tagger(QtWidgets.QApplication):
         self.commands = {name: getattr(self, remcmd.method_name) for name, remcmd in REMOTE_COMMANDS.items()}
 
     def handle_command(self, command):
+        cmd, *args = command.split(' ', 1)
+        argstring = next(iter(args), "")
+        cmd = cmd.upper()
+        log.debug("Executing command: %r", cmd)
         try:
-            cmd, *args = command.split(' ', 1)
-            argstring = next(iter(args), "")
-            log.debug("Executing command: %r", cmd.upper())
-            thread.to_main(self.commands[cmd.upper()], argstring.strip())
+            thread.to_main(self.commands[cmd], argstring.strip())
         except KeyError:
-            log.error("Unknown command: %r", command)
+            log.error("Unknown command: %r", cmd)
 
     def handle_command_cluster(self, argstring):
         self.cluster(self.unclustered_files.files)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -398,6 +398,7 @@ class Tagger(QtWidgets.QApplication):
             # due to the pipe protocol limitations
             # we currently can handle only one file per `remove` command
             "REMOVE": self.handle_command_remove,
+            "REMOVE_ALL": self.handle_command_remove_all,
             "REMOVE_SAVED": self.handle_command_remove_saved,
             "SAVE_COMPLETE": self.handle_command_save_complete,
             "SCAN": self.handle_command_scan,
@@ -432,6 +433,10 @@ class Tagger(QtWidgets.QApplication):
             if argstring == file.filename:
                 argstring.remove()
                 return
+
+    def handle_command_remove_all(self, argstring):
+        for file in self.get_all_file_objects():
+            file.remove()
 
     def handle_command_remove_saved(self, argstring):
         for track in self.get_album_pane_tracks():
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -256,6 +256,10 @@ REMOTE_COMMANDS = {
         "handle_command_remove_saved",
         help_text="Remove all saved releases from the album pane.",
     ),
+    "SAVE_MATCHED": RemoteCommand(
+        "handle_command_save_matched",
+        help_text="Remove all matched releases from the album pane."
+    ),
     "SAVE_MODIFIED": RemoteCommand(
         "handle_command_save_modified",
         help_text="Save all modified files from the album pane.",
@@ -519,6 +523,11 @@ class Tagger(QtWidgets.QApplication):
         for track in self.get_album_pane_tracks():
             if track.state == File.NORMAL:
                 track.remove()
+
+    def handle_command_save_matched(self, argstring):
+        for album in self.albums.values():
+            for track in album.iter_unsaved_files():
+                track.save()
 
     def handle_command_save_modified(self, argstring):
         for track in self.get_album_pane_tracks():
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1214,7 +1214,9 @@ If a new instance will not be spawned:
     parser.add_argument("-d", "--debug", action='store_true',
                         help="enable debug-level logging")
     parser.add_argument("-e", "--exec", nargs="+", action='append',
-                        help="send command (arguments can be entered after space) to a running instance")
+                        help="send command (arguments can be entered after space) to a running instance "
+                        "(use `-e help` for a list of the available commands",
+                        metavar="COMMAND")
     parser.add_argument("-M", "--no-player", action='store_true',
                         help="disable built-in media player")
     parser.add_argument("-N", "--no-restore", action='store_true',
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -55,6 +55,7 @@ import re
 import shutil
 import signal
 import sys
+from textwrap import fill
 from urllib.parse import urlparse
 
 from PyQt5 import (
@@ -208,6 +209,70 @@ class ParseItemsToLoad:
 
     def __str__(self):
         return "files: %r mbids: %r urls: %r commands: %r" % (self.files, self.mbids, self.urls, self.commands)
+
+
+class RemoteCommand:
+    def __init__(self, method_name, help_text=None, help_args=None):
+        self.method_name = method_name
+        self.help_text = help_text or ""
+        self.help_args = help_args or ""
+
+
+REMOTE_COMMANDS = {
+    "CLUSTER": RemoteCommand(
+        "handle_command_cluster",
+        help_text="Cluster all files in the cluster pane.",
+    ),
+    "FINGERPRINT": RemoteCommand(
+        "handle_command_fingerprint",
+        help_text="Calculate acoustic fingerprints for all (matched) files in the album pane.",
+    ),
+    "LOOKUP": RemoteCommand(
+        "handle_command_lookup",
+        help_text="Lookup all clusters in the cluster pane.",
+    ),
+    "LOOKUP_CD": RemoteCommand(
+        "handle_command_lookup_cd",
+        help_text="Read CD from the selected drive and lookup on MusicBrainz. "
+        "Without argument, it defaults to the first (alphabetically) available disc drive",
+        help_args="[device/log file]",
+    ),
+    "QUIT": RemoteCommand(
+        "handle_command_quit",
+        help_text="Exit the running instance of Picard.",
+    ),
+    # due to the pipe protocol limitations
+    # we currently can handle only one file per `remove` command
+    "REMOVE": RemoteCommand(
+        "handle_command_remove",
+        help_text="Remove the file from Picard. Do nothing if no argument.",
+        help_args="[absolute path (1 file)]",
+    ),
+    "REMOVE_ALL": RemoteCommand(
+        "handle_command_remove_all",
+        help_text="Remove all files from Picard.",
+    ),
+    "REMOVE_SAVED": RemoteCommand(
+        "handle_command_remove_saved",
+        help_text="Remove all saved releases from the album pane.",
+    ),
+    "SAVE_COMPLETE": RemoteCommand(
+        "handle_command_save_complete",
+        help_text="Save all completely matched releases in the album pane.",
+    ),
+    "SCAN": RemoteCommand(
+        "handle_command_scan",
+        help_text="Scan all files in the cluster pane.",
+    ),
+    "SHOW": RemoteCommand(
+        "handle_command_show",
+        help_text="Make the running instance the currently active window.",
+    ),
+    "SUBMIT_FINGERPRINTS": RemoteCommand(
+        "handle_command_submit_fingerprints",
+        help_text="Submit outstanding acoustic fingerprints for all (matched) files in the album pane.",
+    ),
+}
 
 
 class Tagger(QtWidgets.QApplication):
@@ -391,22 +456,7 @@ class Tagger(QtWidgets.QApplication):
         return self.unclustered_files.files + list(self.get_album_pane_tracks()) + list(self.clusters.iterfiles())
 
     def _init_remote_commands(self):
-        self.commands = {
-            "CLUSTER": self.handle_command_cluster,
-            "FINGERPRINT": self.handle_command_fingerprint,
-            "LOOKUP": self.handle_command_lookup,
-            "LOOKUP_CD": self.handle_command_lookup_cd,
-            "QUIT": self.handle_command_quit,
-            # due to the pipe protocol limitations
-            # we currently can handle only one file per `remove` command
-            "REMOVE": self.handle_command_remove,
-            "REMOVE_ALL": self.handle_command_remove_all,
-            "REMOVE_SAVED": self.handle_command_remove_saved,
-            "SAVE_COMPLETE": self.handle_command_save_complete,
-            "SCAN": self.handle_command_scan,
-            "SHOW": self.handle_command_show,
-            "SUBMIT_FINGERPRINTS": self.handle_command_submit_fingerprints,
-        }
+        self.commands = {name: getattr(self, remcmd.method_name) for name, remcmd in REMOTE_COMMANDS.items()}
 
     def handle_command(self, command):
         try:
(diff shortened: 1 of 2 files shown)


picard/tagger.py
```diff
@@ -347,11 +351,16 @@ def pipe_server(self):
         while self.pipe_handler.pipe_running:
             messages = [x for x in self.pipe_handler.read_from_pipe() if x not in IGNORED]
             if messages:
+                print(messages)
```
rdswift, line 354: I'm guessing these print statements were for your local debugging during development and shoud be removed?

picard/tagger.py
```diff
@@ -347,11 +351,16 @@ def pipe_server(self):
...
                 self.load_to_picard(messages)
 
     def load_to_picard(self, items):
         parsed_items = ParseItemsToLoad(items)
 
+        for command in parsed_items.commands:
+            print(command)
```
rdswift, line 361: Remove?

picard/tagger.py
```diff
@@ -1144,6 +1161,7 @@ def main(localedir=None, autoupdate=True):
         return longversion()
 
     # any of the flags that change Picard's workflow significantly should trigger creation of a new instance
+    print(picard_args.exec)
```
rdswift, line 1164: Remove?

picard/tagger.py
skelly37: Yeah, it’s all for debugging purposes. As you can see, QUIT doesn’t work and I haven’t implemented most of the commands yet. So it’ll remain here for a while.

picard/tagger.py
```diff
@@ -186,6 +187,9 @@ def __init__(self, items):
             parsed = urlparse(item)
             if not parsed.scheme:
                 self.files.add(item)
+            elif parsed.scheme == "command":
+                for x in item.replace("command://", '').upper().split(';'):
```
zas, line 191: I wonder about how this would play if we want to pass arguments along commands: "CMD1 ARG1 ARG2; CMD2 ARG3"
It might be useful, for example to target a specific cluster or album.
Also I'm not sure about the handling of multiple commands using ";" separator, instead of repeating "-e" option.
@phw : what do you think about allowing arguments?


I don't think we should convert to uppercase here, but instead do case-insensitive comparison in `handle_command()`.

`item.replace("command://", '')` -> `item[10:]`

rdswift: I agree that repeating the -e option with command and argument to use makes more sense from a user perspective.

skelly37: I agree with the repeating -e, good point.

I’m not sure what do you mean by a *case-insensitive comaprison* tho @zas

zas: I meant just do the conversion when needed.

```python
    def handle_command(self, command):
        command = command.upper()
        if command == "SHOW":
            self.bring_tagger_front()
        elif command == "QUIT":
            self.exit()
```

BTW, I was thinking about something like:

```python
    self.commands = {
        'SHOW': self.handle_command_show,
        'QUIT': self.handle_command_quit,
        'CMD_WITH_ARGS': self.handle_command_with_args,
    }

...

    def handle_command(self, command):
        try:
            cmd, argstring = command.strip().split(" ", 1)
            commands[cmd.upper()](argstring.strip())
        except KeyError:
            log.error("Unknown command: %r", command)

    def handle_command_show(self, argstring):
        self.bring_tagger_front()

    def handle_command_quit(self, argstring):
        self.exit()

    def handle_command_with_args(self, argstring):
        if not argstring:
            log.error("Missing argument")
        do_something_with(argstring)
```

(indentation may be incorrect, just an example)

It would be easier to extend, and commands can be added dynamically.

rdswift: When splitting, you may also need to trap `ValueError: not enough values to unpack (expected 2, got 1)` in case there is no space (no args for the command).

skelly37: Okay, there's a problem with your idea:

```
〉picard -e a -e b -e c
c
```

Basically it's a flag that can have only one value. Filepaths/URLs are positional arguments, so I don't see the other way than `picard -e "SHOW; QUIT"`

skelly37: The problem is that we cannot use a flag (which is `-e/--exec`) more than once because it gets overriden. So even if we add `nargs='*'`, we end up with something like this:

```
〉picard -e a -e b -e c
['c']

〉picard -e a b c
['a', 'b', 'c']
```

So we've returned to the start point, in that case I stick to my `picard -e "SHOW; CLUSTER; QUIT"` idea

zas: > The problem is that we cannot use a flag (which is -e/--exec) more than once because it gets overriden.

Try this:

```python
>>> p = argparse.ArgumentParser()
>>> p.add_argument("-e", nargs='+', action='append')
_AppendAction(option_strings=['-e'], dest='e', nargs='+', const=None, default=None, type=None, choices=None, help=None, metavar=None)
>>> p.parse_args("-e a b c -e d e -e f".split())
Namespace(e=[['a', 'b', 'c'], ['d', 'e'], ['f']])
>>> p.parse_args("-e".split())
usage: [-h] [-e E [E ...]]
: error: argument -e: expected at least one argument
```

picard/tagger.py
```diff
@@ -369,14 +378,25 @@ def load_to_picard(self, items):
...
             self.bring_tagger_front()
 
     def handle_command(self, command):
-        if command == "SHOW":
-            self.bring_tagger_front()
-        elif command == "QUIT":
-            self.exit()
+        try:
+            print(command)
+            to_exec = command.split(" ", 1)
```
zas, line 387: A more pythonic approach:

```python
cmd, *args = command.split(' ', 1)
argstring = next(iter(args), "")
thread.to_main(self.commands[cmd.upper()], argstring.strip())
```

skelly37: I wouldn't come up with such thing myself, thanks a lot!

picard/tagger.py
skelly37: Aight, thanks

skelly37:
Hm

- The *core* works
- `SHOW` works
- `QUIT` gets stuck as when the pipe was introduced, then removed (but force quit will not be a reasonable solution here)
- print leftovers removed
- the cli flag got refactored

Time to implement other commands and think about the `QUIT` in the meantime, I guess...

zas:
@skelly37 https://github.com/skelly37/picard/pull/5 adds a bit more logging

picard/tagger.py
```diff
@@ -391,7 +391,8 @@ def handle_command_show(self, argstring):
         self.bring_tagger_front()
 
     def handle_command_quit(self, argstring):
-        self.exit()
+        # hotfix, should be replaced by something better imo
+        thread.to_main(self.sighandler)
```
skelly37, line 395: I'm not sure about it but at least it works and I can move further with the implementation

picard/tagger.py
```diff
@@ -396,6 +396,9 @@ def handle_command(self, command):
     def handle_command_cluster(self, argstring):
         self.cluster(self.unclustered_files.files)
 
+    def handle_command_lookup(self, argstring):
+        self.autotag(self.unclustered_files.files)
```
skelly37, line 400: @rdswift that's the scan you meant, right?

skelly37:
Update:

- `QUIT` got fixed and has a proper flow
- 5/9 commands are already implemented (`QUIT` + `SHOW` + all the cluster pane ones)
- Only album pane ones left to do
- After that we can discuss ideas/design of the functions that actually take the arguments suggested by @zas 

@phw, @zas What's wrong [there](https://github.com/metabrainz/picard/runs/7680182896?check_suite_focus=true)? I don't understand...

zas:
> 
> @phw, @zas What's wrong [there](https://github.com/metabrainz/picard/runs/7680182896?check_suite_focus=true)? I don't understand...

This was a temporary server issue (it was returning http 500) when trying to upload [SARIF](https://docs.github.com/en/code-security/code-scanning/integrating-with-code-scanning/sarif-support-for-code-scanning) results from [CodeQL action](https://github.com/github/codeql-action).
This is configured by this [workflow](https://github.com/metabrainz/picard/blob/695d2aa7597fb6723da06ca6081413ffe5fcd58a/.github/workflows/codeql-analysis.yml).

I re-ran the failed job, and it worked.

picard/tagger.py
```diff
@@ -234,6 +245,18 @@ def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
...
+        self.commands = {
+            "CLUSTER": self.handle_command_cluster,
+            "FINGERPRINT": self.handle_command_fingerprint,
+            "LOOKUP": self.handle_command_lookup,
+            "QUIT": self.handle_command_quit,
+            "REMOVE_SAVED": self.handle_command_remove_saved,
+            "SAVE_COMPLETE": self.handle_command_save_complete,
+            "SCAN": self.handle_command_scan,
+            "SHOW": self.handle_command_show,
+            # SUBMIT_FINGERPRINTS ???
```
skelly37, line 257: I cannot seem to find a function submitting fingerprints, any ideas?

zas: > I cannot seem to find a function submitting fingerprints, any ideas?

https://github.com/metabrainz/picard/blob/695d2aa7597fb6723da06ca6081413ffe5fcd58a/picard/acoustid/manager.py#L153

zas:
@skelly37 your last commit is titled "consider it ready", it wouldn't help one looking at commit history.

In general, I think you should be more careful about your commit messages.Things like "almost done" or "typo fix v2" don't help anyone to understand what's the commit is about.
Also, please use correct English (capitalization, punctuation, verbs).

In this case, since that was a draft pull request for a while, I think an interactive rebase / history cleanup is needed before merging.

This was also the case with your others PRs. Nothing catastrophic, but we usually prefer clean explicit commit messages (and sometimes we get lazy too, so our commit messages aren't always fantastic...).

Please just take care about that.

picard/tagger.py
```diff
@@ -234,6 +245,18 @@ def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
...
+            "FINGERPRINT": self.handle_command_fingerprint,
+            "LOOKUP": self.handle_command_lookup,
+            "QUIT": self.handle_command_quit,
+            "REMOVE_SAVED": self.handle_command_remove_saved,
+            "SAVE_COMPLETE": self.handle_command_save_complete,
+            "SCAN": self.handle_command_scan,
+            "SHOW": self.handle_command_show,
+            "SUBMIT_FINGERPRINTS": self.handle_command_submit_fingerprints,
+        }
+
```
zas, line 259: Perhaps it would be easier to move this near `handle_command_*()` methods:

Here call:
```python
    self._init_remote_commands()
```

Near  `handle_command_*()` methods:

```python
    def _init_remote_commands(self):
        self.commands = {
            "CLUSTER": self.handle_command_cluster,
            "FINGERPRINT": self.handle_command_fingerprint,
            "LOOKUP": self.handle_command_lookup,
            "QUIT": self.handle_command_quit,
            "REMOVE_SAVED": self.handle_command_remove_saved,
            "SAVE_COMPLETE": self.handle_command_save_complete,
            "SCAN": self.handle_command_scan,
            "SHOW": self.handle_command_show,
            "SUBMIT_FINGERPRINTS": self.handle_command_submit_fingerprints,
        }
```

picard/tagger.py
```diff
@@ -360,9 +388,57 @@ def load_to_picard(self, items):
...
-        if parsed_items:
+        if parsed_items.non_executable_items():
             self.bring_tagger_front()
 
+    def get_album_pane_tracks(self):
+        tracks = set()
+        for album_name in self.albums:
+            for track in self.albums[album_name].iterfiles():
+                tracks.add(track)
+        return tracks
```
zas, line 399: Not sure the conversion to `set` is needed here, perhaps we can just use:

```python
def get_album_pane_tracks(self):
    for album in self.albums.values():
        for track in album.iterfiles():
            yield track
```

At worse, if required, conversion to set can be done in caller:

```python
for track in set(self.get_album_pane_tracks()):
```

What do you think?

zas (changes requested):

skelly37:
> @skelly37 your last commit is titled "consider it ready", it wouldn't help one looking at commit history.
> 
> In general, I think you should be more careful about your commit messages.Things like "almost done" or "typo fix v2" don't help anyone to understand what's the commit is about. Also, please use correct English (capitalization, punctuation, verbs).
> 
> In this case, since that was a draft pull request for a while, I think an interactive rebase / history cleanup is needed before merging.
> 
> This was also the case with your others PRs. Nothing catastrophic, but we usually prefer clean explicit commit messages (and sometimes we get lazy too, so our commit messages aren't always fantastic...).
> 
> Please just take care about that.

Sorry about that, I get your point.

Though, I'm personally used to the *squash on merge* flow, where commits serve mostly only the in-PR purposes. And actually I was going to mention this issue, too. Why doesn't Picard use squash on merge? The CI/CD works just fine so bisect is rather not a case and I can't think of any other reason not to squash.

picard/tagger.py
skelly37: The reason I was adding them to a set was because I forgot/didn't know about the `values` method. So it is not required and the `yield` approach is much better, thanks :)

zas:
> Why doesn't Picard use squash on merge?

Usually we want to preserve dev history, it is better to have smaller commits, especially when it comes to use bisect to find when an issue was discovered.
Of course, it all depends on the dev process, but small logical units often make sense.
I use to squash commits, doing interactive rebase before the PR is merged, so, at the end, a simple PR can be one commit, but for bigger PRs squashing all may be too much, and having more steps is usually preferred.

Commits that are squashed are typically:
- commits cancelling each other (like added/removed code)
- commits fixing a previous commit (like typos)
- commits fixing a previous commit that broke tests

Each commit is supposed to test successfully, that's not always easy for big feature introduction, but often that's more about code organization: introducing a bunch of methods without their calls (so those are no-op) until everything is in place, then add a commit enabling the added code.

There's nothing written in stone in this matter, but the general rule is:
- small "logic" commits
- clear commit messages
- commit doesn't break tests

zas:
For example, https://github.com/metabrainz/picard/pull/2137/commits/e012e84d22d23dd6a093426d25500ed80f910959 should be 2 small commits:

- one introducing `_init_remote_commands()`
- one changing the loop in `get_album_pane_tracks()`

Why? Because those are independent changes.

skelly37:
@zas Okay, understood, thanks for explaining your convention to me.

My suggestion is then to split it into such commits commits:
1) `ParseItemsToLoad` got prepared to parse commands // all the changes that happened there
2) Method to get all the tracks from album pane for Tagger // `get_album_pane_tracks`
3) Command handling introduced to Tagger // L389-L441
4) Commands are actually passed via command-line // `parser.add_argument` and needed changes in the main function

Are you okay with that, assuming that none of you will request changes not related to the git?

skelly37:
ping @phw

zas (approved):
LGTM

picard/tagger.py
```diff
@@ -386,12 +386,19 @@ def get_album_pane_tracks(self):
             for track in album.iterfiles():
                 yield track
 
+    def get_all_file_objects(self):
+        return self.unclustered_files.files + [x for x in self.get_album_pane_tracks()] + [x for x in self.clusters.iterfiles()]
```
zas, line 390: I think that's equivalent to:

```suggestion
        return self.unclustered_files.files + list(self.get_album_pane_tracks()) + list(self.clusters.iterfiles())
```

picard/tagger.py
```diff
@@ -421,6 +428,16 @@ def handle_command_quit(self, argstring):
         self.exit()
         self.quit()
 
+    def handle_command_remove(self, argstring):
+        for file in self.get_all_file_objects():
+            if argstring == file.filename:
+                argstring.remove()
```
zas, line 434:
```suggestion
                file.remove()
```

?

zas (changes requested):

picard/tagger.py
skelly37: Yeah, I was experimenting a bit and finally left with some typo, thanks for noticing

skelly37:
Okay, all the ideas implemented, the PR is ready imo.

I'd just like to ask anyone to manually test the `lookup_cd` command. I have no music CDs here so I cannot actually test if I've rewritten it correctly.

phw (dismissed):

requirements-build.txt
```diff
@@ -1,3 +1,2 @@
 Babel==2.9.1
 PyInstaller==5.3
-setuptools<45.0.0
```
phw, line 3: This got already removed on master, I think rebasing this branch against master would solve that.

phw:
Sorry for the long silence, I had been totally occupied by life here. Things look good to me. I can theoretically test the CD lookup feature, but right now don't have my CD drive handy. I'll see to do this later, but would be happy with merging this before.

Wanted to comment on the squash merge question a bit. zas has said it basically all already, and I fully agree here. When merging a single commit I usually go with the rebase and merge approach, so we get a simpler history. Also if it's really a simple change with some of those fixup commits I might choose a squash and merge.

But I'm not really a fan of squash and merge as a rule for every PR. IMHO many more complex changes warrant to be separated into multiple logical commits, that just makes the change easier to understand. Also having an explicit merge commit then makes it clear these changes formed a PR.

But what's right is that we don't always are very strict and consistent with this, and haven't really formalized the rules here. We had some rather wild commit histories merged, we sometimes have single commits with separate merge commits.

requirements-build.txt
skelly37: You've approved and commented so I'm confused

Should I do it now or is it just a tip for the future PRs?

skelly37: Done :)

zas (dismissed):
LGTM

phw (dismissed):

rdswift (dismissed):
Looks okay to me.  I notice that the available commands are not displayed in any help, so we'll need to make sure they are documented well in the on-line documentation.

Something we might consider is to add a `-e --help` or `--help -e` or `--e-help` or some other option to display the available commands.

**EDIT:** Because `-e` always requires a command as an argument, perhaps the easiest way would be to have `help` as one of the commands so if the user enters `-e help` the available commands for the `-e` command line argument would be displayed.

skelly37:
> Looks okay to me.  I notice that the available commands are not displayed in any help, so we'll need to make sure they are documented well in the on-line documentation.
> 
> Something we might consider is to add a `-e --help` or `--help -e` or `--e-help` or some other option to display the available commands.
> 
> **EDIT:** Because `-e` always requires a command as an argument, perhaps the easiest way would be to have `help` as one of the commands so if the user enters `-e help` the available commands for the `-e` command line argument would be displayed.

-e help seems okay to me. Should be done by the end of the day.

picard/tagger.py
```diff
@@ -1094,6 +1213,8 @@ def process_picard_args():
                         help="location of the configuration file (starts a stand-alone instance)")
     parser.add_argument("-d", "--debug", action='store_true',
                         help="enable debug-level logging")
+    parser.add_argument("-e", "--exec", nargs="+", action='append',
+                        help="send command (arguments can be entered after space) to a running instance")
```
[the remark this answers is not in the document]
skelly37: @rdswift and how do you see `-e help`?

My idea is to
1) add your suggestion
2) scan `exec` args for help. If help is present, print help instead of passing the args to the pipe
3) just use `print`(?) for some help message, as consistent with `picard --help` as possible

Any better ideas or should I go this way?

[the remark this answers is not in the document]
rdswift: What you suggest sounds good to me.  I also think if you could include a brief note about being able to have multiple commands and them processed in the order they appear.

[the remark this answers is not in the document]
rdswift: Just to clarify, if a command with arguments is entered does the wole thing need to be in quotes?  For example `-e lookup path/to/my/eac/log/file` would need to be entered as `-e "lookup path/to/my/eac/log/file"`?  If so, the brief help should probably mention that as well.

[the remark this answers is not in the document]
skelly37: Nah, you don't need the quotes, unless the path contains spaces (which is an expected behavior imo)

rdswift:
> Nah, you don't need the quotes, unless the path contains spaces (which is an expected behavior imo)

I agree that is the expected behavior.

Having said that, do we have a problem if we try to pass both a command AND a file to load?  For example, `picard file/to/load` is a valid command line as is `picard -e lookup file/to/lookup`.  What happens if the two are combined as `picard -e lookup file/to/lookup file/to/load`?  I suspect that the `file/to/load` argument will also get passed on to the `-e lookup` command along with `file/to/lookup`, and I believe that will be a problem.

rdswift (approved):
Looks okay to me, but please have a look at the issue that I raised about including both a command and a file to load on the same command line.

skelly37:
> What happens if the two are combined as `picard -e lookup file/to/lookup file/to/load`?  I suspect that the `file/to/load` argument will also get passed on to the `-e lookup` command along with `file/to/lookup`, and I believe that will be a problem.


Depends.

- If there is an existing instance, picard -e will send all the positional arguments in the given order.
- If there is no running instance, nothing will happen, as picard -e blocks Picards startup

zas:
@skelly37 nice to have help for remote commands, have a look at https://github.com/skelly37/picard/pull/7

rdswift:
> * If there is no running instance, nothing will happen, as picard -e blocks Picards startup

In this case, would it make sense to start an instance and process the commands rather than (silently) do nothing?

skelly37:
> > * If there is no running instance, nothing will happen, as picard -e blocks Picards startup
> 
> In this case, would it make sense to start an instance and process the commands rather than (silently) do nothing?

Imo it makes sense to execute stuff only on an existing instance; *execute a command in a running Picard instance.* If others insist I might change it but I'd prefer not to. Take for example: `picard -e QUIT`. Would it make sense to start Picard only to quit it?

The current logic is as intuitive as possible for me.

rdswift:
> * If there is an existing instance, picard -e will send all the positional arguments in the given order. Then the commands. So yeah, you are right, all files will be looked up alongside with the file/to/load in your scenario

This still bothers me a little bit, because the command line behavior of adding a file via the command line differs depending on whether or not there is a `-e command` specified before the path of the file to add.  I wonder if it will work properly if the file to add is specified before ANY `-e command` options.  I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.

zas:
> I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.

Yes, that's a bit confusing. I wonder if we should add a `LOAD url/file` command for convenience so one can only use `-e` commands (optionally).

`picard -e LOAD filepath1 filepath2 filepath3 -e CLUSTER -e LOOKUP `

skelly37:
> > * If there is an existing instance, picard -e will send all the positional arguments in the given order. Then the commands. So yeah, you are right, all files will be looked up alongside with the file/to/load in your scenario
> 
> This still bothers me a little bit, because the command line behavior of adding a file via the command line differs depending on whether or not there is a `-e command` specified before the path of the file to add. I wonder if it will work properly if the file to add is specified before ANY `-e command` options. I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.

Any `-e` argument prevents Picard from starting, regardless of its type and the order of other arguments, see: https://github.com/metabrainz/picard/pull/2137/files#diff-754164b29034aa27e27d501630b16fe491c78108500b10caeab6b272bc240fbdR1328

This explains it well, imo:

  If there is no running instance, picard -e blocks Picards startup and does
nothing.

skelly37:
> > I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.
> 
> Yes, that's a bit confusing. I wonder if we should add a `LOAD url/file` command for convenience so one can only use `-e` commands (optionally).
> 
> `picard -e LOAD filepath1 filepath2 filepath3 -e CLUSTER -e LOOKUP `

So you're saying that we should ditch the positional argument FILE_OR_URL in favor of adding a LOAD command which would be the only way to load something to picard?

zas:
> > > I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.
> > 
> > 
> > Yes, that's a bit confusing. I wonder if we should add a `LOAD url/file` command for convenience so one can only use `-e` commands (optionally).
> > `picard -e LOAD filepath1 filepath2 filepath3 -e CLUSTER -e LOOKUP `
> 
> So you're saying that we should ditch the positional argument FILE_OR_URL in favor of adding a LOAD command which would be the only way to load something to picard?

Nope, in addition. Basically `picard <FILE>` would be equivalent to `picard -e LOAD <FILE>`. The idea is to let user use only `-e` commands to achieve the same thing.

skelly37:
> > > > I think it might, in which case the documentation should probably note that any `-e command` options should only appear at the end of the command line.
> > > 
> > > 
> > > Yes, that's a bit confusing. I wonder if we should add a `LOAD url/file` command for convenience so one can only use `-e` commands (optionally).
> > > `picard -e LOAD filepath1 filepath2 filepath3 -e CLUSTER -e LOOKUP `
> > 
> > 
> > So you're saying that we should ditch the positional argument FILE_OR_URL in favor of adding a LOAD command which would be the only way to load something to picard?
> 
> Nope, in addition. Basically `picard <FILE>` would be equivalent to `picard -e LOAD <FILE>`. The idea is to let user use only `-e` commands to achieve the same thing.

In that case we have to either redesign the argstring handling or allow only one file per LOAD (like in REMOVE), because paths can contain spaces

rdswift:
> In that case we have to either redesign the argstring handling or allow only one file per LOAD (like in REMOVE), because paths can contain spaces

At the risk of introducing some scope creep, I think it might be cleanest overall to update the arg handling as it is passed to the pipe in the line:

```
                pipe_handler = pipe.Pipe(app_name=PICARD_APP_NAME, app_version=PICARD_FANCY_VERSION_STR, args=to_be_added)
```

Currently `to_be_added` is an iterable of strings.  If it was changed to an iterable of tuples of (command, command_args) where `command` is a string and `command_args` is a list of strings, this would simplify argument processing because argparse would take care of quoted arguments for you automatically.  This could also make use of the `LOAD` command proposed by @zas because the contents of the argparse `FILE_OR_URL` could automatically be converted to a `LOAD` command (ultimately handled with a single processing path).  It would also remove the single file argument restriction for the `REMOVE` command.

To reuse most of your existing pipe and command processing code, commands with multiple arguments could be sent to the pipe as separate command/argument message strings.

phw (approved):

rdswift:
> Take for example: `picard -e QUIT`. Would it make sense to start Picard only to quit it?

Actually, I believe that it would make sense because it would allow the user to process a command stack such as:

```
picard -e load my/ripped/album/directory -e cluster -e lookup -e save_complete -e fingerprint -e submit_fingerprints -e remove_saved -e quit
```

This would all be handled in a single running instance rather than first having to create an instance for processing and then create another instance to pass the command stack to the first instance.  We might even consider adding a `DUMP_LOG` command to dump the log entries to a file specified as an argument to the command, like `-e dump_log my/log/for/this/album.txt` which could be included in the command stack just prior to `-e quit`.

Don't get me wrong, I think what you've done with this so far is great.  I just think that with a little tweaking it could be so much more versatile.  Having said that, perhaps these additional tweaks should be on a separate PR after this one has been merged.

rdswift:
One final (?) item...  I see that we have a `SAVE_COMPLETE` command to save fully matched releases, but nothing for saving partially matched releases which are quite common when the user chooses to `SCAN` rather than `LOOKUP`.  Perhaps we should also have a `SAVE_ALL` command to accommodate that type of workflow.

rdswift:
Sorry, but I just thought of another potential command (enhancement), that being the ability to read the commands from a text file.  Perhaps the command could be called `BATCH` or `INCLUDE`.  For example, if the text file contained:

```
load "/my/ripped/files/directory/selected album"
cluster
lookup
save_complete
fingerprint
submit_fingerprints
remove_saved
dump_log /var/log/picard_selected_album.log
quit
```

then the command stack in the text file could be executed by calling `picard -e batch "/my/command/batch/file"`.  You could use `shlex.split()` to parse each non-empty line from the text file and then add them to the queue in the same way as if they had been entered on the command line.

This would allow the users to build potentially lengthy command stacks (either by hand or dynamically through their ripping workflow) and defer Picard processing until later using a batching process.

skelly37:
> Sorry, but I just thought of another potential command (enhancement), that being the ability to read the commands from a text file. Perhaps the command could be called `BATCH` or `INCLUDE`. For example, if the text file contained:
> 
> ```
> load "/my/ripped/files/directory/selected album"
> cluster
> lookup
> save_complete
> fingerprint
> submit_fingerprints
> remove_saved
> dump_log /var/log/picard_selected_album.log
> quit
> ```
> 
> then the command stack in the text file could be executed by calling `picard -e batch "/my/command/batch/file"`. You could use `shlex.split()` to parse each non-empty line from the text file and then add them to the queue in the same way as if they had been entered on the command line.
> 
> This would allow the users to build potentially lengthy command stacks (either by hand or dynamically through their ripping workflow) and defer Picard processing until later using a batching process.

That would be really cool, thanks.



> One final (?) item... I see that we have a `SAVE_COMPLETE` command to save fully matched releases, but nothing for saving partially matched releases which are quite common when the user chooses to `SCAN` rather than `LOOKUP`. Perhaps we should also have a `SAVE_ALL` command to accommodate that type of workflow.

It saves all changed files, if I understood the enums correctly.


> Currently `to_be_added` is an iterable of strings. If it was changed to an iterable of tuples of (command, command_args) where `command` is a string and `command_args` is a list of strings, this would simplify argument processing because argparse would take care of quoted arguments for you automatically. This could also make use of the `LOAD` command proposed by @zas because the contents of the argparse `FILE_OR_URL` could automatically be converted to a `LOAD` command (ultimately handled with a single processing path). It would also remove the single file argument restriction for the `REMOVE` command.
> 
> To reuse most of your existing pipe and command processing code, commands with multiple arguments could be sent to the pipe as separate command/argument message strings.

So you're saying that `picard -e REMOVE file1 file2` should be split into:
- `picard -e REMOVE file1`
- `picard -e REMOVE file2`
? This might be indeed a good solution.


> > Take for example: `picard -e QUIT`. Would it make sense to start Picard only to quit it?
> 
> Actually, I believe that it would make sense because it would allow the user to process a command stack such as:
> 
> ```
> picard -e load my/ripped/album/directory -e cluster -e lookup -e save_complete -e fingerprint -e submit_fingerprints -e remove_saved -e quit
> ```
> 
> This would all be handled in a single running instance rather than first having to create an instance for processing and then create another instance to pass the command stack to the first instance. We might even consider adding a `DUMP_LOG` command to dump the log entries to a file specified as an argument to the command, like `-e dump_log my/log/for/this/album.txt` which could be included in the command stack just prior to `-e quit`.

Yeah, now I tend to agree with you, `picard -e` shouldn't prevent Picard from starting.

> Don't get me wrong, I think what you've done with this so far is great. I just think that with a little tweaking it could be so much more versatile. Having said that, perhaps these additional tweaks should be on a separate PR after this one has been merged.

Thanks :) Though I'm not sure how much should we add in this PR. Imo we should add the reading from file and `load` in this PR, but patch the multi-argument commands and `picard -e` blocking startup in the second one. How about that?

rdswift:
> > One final (?) item... I see that we have a `SAVE_COMPLETE` command to save fully matched releases, but nothing for saving partially matched releases which are quite common when the user chooses to `SCAN` rather than `LOOKUP`. Perhaps we should also have a `SAVE_ALL` command to accommodate that type of workflow.
> 
> It saves all changed files, if I understood the enums correctly.

Okay, I just took a look at the code and I see that it actually saves all matched files as you said, rather than just the releases that are marked as "complete" (indicating that all tracks on the album/release have one [and only one?] file attached).  That being the case, I suggest that the current `SAVE_COMPLETE` command be renamed to `SAVE_MATCHED` or `SAVE_ALL` to avoid confusion, and perhaps consider adding a `SAVE_COMPLETE` command that only saves files where the release is marked as "complete".

> So you're saying that `picard -e REMOVE file1 file2` should be split into:
> 
> * `picard -e REMOVE file1`
> * `picard -e REMOVE file2`
>   ? This might be indeed a good solution.

Not exactly (unless I misunderstood your comment).  What I meant was that the user could still enter it as `picard -e REMOVE file1 file2` and when you process it you send it to the pipe as two separate commands as if it had been entered as `picard -e REMOVE file1 -e REMOVE file2`.  To accomplish this I suggest replacing line 1364 which is currently:

```
                to_be_added.append("command://" + " ".join(e))
```

with something like (**WARNING: Untested Code**):

```
                if len(e) > 1:
                    for _arg in e[1:]:
                        to_be_added.append("command://{0} {1}".format(e[0], _arg))
                else:
                    to_be_added.append("command://" + e[0])
```

> > Don't get me wrong, I think what you've done with this so far is great. I just think that with a little tweaking it could be so much more versatile. Having said that, perhaps these additional tweaks should be on a separate PR after this one has been merged.
> 
> Thanks :) Though I'm not sure how much should we add in this PR. Imo we should add the reading from file and `load` in this PR, but patch the multi-argument commands and `picard -e` blocking startup in the second one. How about that?

That works for me, but I'll defer to @zas and @phw for guidance.  I'd still l ike to see the `LOG_DUMP` (and possibly `LOG_CLEAR`) command added at some point as well.

skelly37:
> Okay, I just took a look at the code and I see that it actually saves all matched files as you said, rather than just the releases that are marked as "complete" (indicating that all tracks on the album/release have one [and only one?] file attached). That being the case, I suggest that the current `SAVE_COMPLETE` command be renamed to `SAVE_MATCHED` or `SAVE_ALL` to avoid confusion, and perhaps consider adding a `SAVE_COMPLETE` command that only saves files where the release is marked as "complete".

Okay, I can rename it to `SAVE_MODIFIED` (better than `SAVE_ALL`, as some files might be untouched or broken). The actual `SAVE_COMPLETE` would require more research from me, so I'd leave it for the next PR.

> 
> > So you're saying that `picard -e REMOVE file1 file2` should be split into:
> > 
> > * `picard -e REMOVE file1`
> > * `picard -e REMOVE file2`
> >   ? This might be indeed a good solution.
> 
> Not exactly (unless I misunderstood your comment). What I meant was that the user could still enter it as `picard -e REMOVE file1 file2` and when you process it you send it to the pipe as two separate commands as if it had been entered as `picard -e REMOVE file1 -e REMOVE file2`. To accomplish this I suggest replacing line 1364 which is currently:

Actually, you misunderstood it and that's what I meant :) Thanks

> > > Don't get me wrong, I think what you've done with this so far is great. I just think that with a little tweaking it could be so much more versatile. Having said that, perhaps these additional tweaks should be on a separate PR after this one has been merged.
> > 
> > 
> > Thanks :) Though I'm not sure how much should we add in this PR. Imo we should add the reading from file and `load` in this PR, but patch the multi-argument commands and `picard -e` blocking startup in the second one. How about that?
> 
> That works for me, but I'll defer to @zas and @phw for guidance. I'd still l ike to see the `LOG_DUMP` (and possibly `LOG_CLEAR`) command added at some point as well.

- `LOG_DUMP <filename>`: dump picard log into some file
- `LOG_CLEAR`: clear the current logger state

Did I get it right?

skelly37:
Okay, so after reading your (very helpful btw :>) feedback I suggest merging it now, while leaving:

- `LOG_DUMP`
- `LOG_CLEAR`
- `LOAD`
- `SAVE_COMPLETE`
- `picard -e` does not necessarily prevent Picard's startup
- *positional arguments removed in favor of `picard -e LOAD <filename>` ?*

for the next PR

rdswift:
> * `LOG_DUMP <filename>`: dump picard log into some file
> * `LOG_CLEAR`: clear the current logger state
> 
> Did I get it right?

I think so.  The relevant code for saving and clearing the log can be found in the `_save_log_as_do()` and `_clear_log_do()` modules in `picard/ui/logview.py`.  There may need to be some code duplication or refactoring required to avoid the confirmation, file selection, and error dialogs.

> Okay, so after reading your (very helpful btw :>) feedback I suggest merging it now, while leaving:
> 
> * `LOG_DUMP`
> * `LOG_CLEAR`
> * `LOAD`
> * `SAVE_COMPLETE`
> * `picard -e` does not necessarily prevent Picard's startup
> * _positional arguments removed in favor of `picard -e LOAD <filename>` ?_
> 
> for the next PR

Fine by me, but I think you might be misunderstanding the positional arguments portion.  If I understood @zas correctly, I believe that the positional argument for loading a file on start-up should remain (if for no other reason than to keep it backwards compatible) but also add the `LOAD` command to allow both options.  As I understand it, for the positional argument loading to work the files/directories must be specified on the command line before any `-e` commands.

skelly37:
> > * `LOG_DUMP <filename>`: dump picard log into some file
> > * `LOG_CLEAR`: clear the current logger state
> > 
> > Did I get it right?
> 
> I think so. The relevant code for saving and clearing the log can be found in the `_save_log_as_do()` and `_clear_log_do()` modules in `picard/ui/logview.py`. There may need to be some code duplication or refactoring required to avoid the confirmation, file selection, and error dialogs.

Understood, thanks
 

> Fine by me, but I think you might be misunderstanding the positional arguments portion. If I understood @zas correctly, I believe that the positional argument for loading a file on start-up should remain (if for no other reason than to keep it backwards compatible) but also add the `LOAD` command to allow both options. As I understand it, for the positional argument loading to work the files/directories must be specified on the command line before any `-e` commands.

Actually, we're working on Picard 3.0.0. In semantic versioning, incrementing major is connected with breaking changes, so I don't think we should worry about the backwards compatibility. Actually, we could enforce the proper arguments flow, i.e. when something will get loaded, processed etc. Not to mention how much code we could remove thanks to it. So as for me I'm strongly for deprecating positional arguments in favor of the `picard -e LOAD`. 

@zas @phw what's your take on this?

zas:
> > > * `LOG_DUMP <filename>`: dump picard log into some file
> > > * `LOG_CLEAR`: clear the current logger state
> > > 
> > > Did I get it right?
> > 
> > 
> > I think so. The relevant code for saving and clearing the log can be found in the `_save_log_as_do()` and `_clear_log_do()` modules in `picard/ui/logview.py`. There may need to be some code duplication or refactoring required to avoid the confirmation, file selection, and error dialogs.
> 
> Understood, thanks
> 
> > Fine by me, but I think you might be misunderstanding the positional arguments portion. If I understood @zas correctly, I believe that the positional argument for loading a file on start-up should remain (if for no other reason than to keep it backwards compatible) but also add the `LOAD` command to allow both options. As I understand it, for the positional argument loading to work the files/directories must be specified on the command line before any `-e` commands.
> 
> Actually, we're working on Picard 3.0.0. In semantic versioning, incrementing major is connected with breaking changes, so I don't think we should worry about the backwards compatibility. Actually, we could enforce the proper arguments flow, i.e. when something will get loaded, processed etc. Not to mention how much code we could remove thanks to it. So as for me I'm strongly for deprecating positional arguments in favor of the `picard -e LOAD`.
> 
> @zas @phw what's your take on this?

I'm not sure about this, positional arguments shouldn't be much an issue, the order issue come from the fact we allow multiple arguments after -e commands, so it is impossible to distinguish `-e cmd arg1 arg2` from `-e cmd arg` `posarg1` hence the need to swap them and write `posarg1` `-e cmd arg`. But this only occur if one is mixing `-e` commands and positional arguments.
To me, that's not much an issue, and I don't think we should remove positional arguments for this corner case.
I agree that's a bit weird, and it might confuse new users though. We could also limit number of args passed to `-e cmd` but it will likely make the code overcomplicated,
I tend to think we should keep it as is, and just let the user use either `picard -e LOAD file1 file2` or just `picard file1 file2`, while documenting the case of mixing those, and impact of order of arguments.

phw?

rdswift:
> Okay, I can rename it to `SAVE_MODIFIED` (better than `SAVE_ALL`, as some files might be untouched or broken). The actual `SAVE_COMPLETE` would require more research from me, so I'd leave it for the next PR.

@zas @phw Is `SAVE_MODIFIED` okay?  It seems that the code will only save the file if it has been modified, which matches the command title so that looks to be okay.

Does this raise another issue in that matched files will not be saved unless they have been modified, and there is no `-e` command to save matched files.  Does this mean that there is no way to process files to rename and/or move unless the file metadata has changed?  Is there a need to also have a `SAVE_MATCHED` type of command?

zas:
I'd say yes to both: `SAVE_ALL` -> `SAVE_MODIFIED` and add `SAVE_MATCHED`

skelly37:
> I'd say yes to both: `SAVE_ALL` -> `SAVE_MODIFIED` and add `SAVE_MATCHED`

So you're saying we should add `SAVE_MATCHED` in this PR as well?

zas:
> > ``
> 
> So you're saying we should add `SAVE_MATCHED` in this PR as well?

Yes, if possible.

skelly37:
Fine, I’ll do some research then

skelly37:
@zas @rdswift I'm wondering if I understood the task correctly. Can I achieve this by:

```python
for file in self.get_all_file_objects():
    if file.acoustid_length:
        file.remove()
```?

Or anything else, basically a matched track should have any metadata.

**Comment by zas** [comment]
> @zas @rdswift I'm wondering if I understood the task correctly. Can I achieve this by:

You mean for `SAVE_MATCHED` ?

Have a look at https://github.com/metabrainz/picard/blob/224096df0be6452c436a658f488168c04b487e9e/picard/album.py#L749-L764

Basically, I think it would be iterating over loading albums and call something like `_iter_unsaved_files()`.

@rdswift does it match your idea?

**Comment by rdswift** [comment]
> @rdswift does it match your idea?

Yup.  That's what I had in mind.  If I read the code correctly, this would provide an iterable of files that were matched to a track (and no other files matched to that track) that has not already been saved.

**Comment by skelly37** [comment]
Okay, I'm done (unless you request some changes during review).

- I've added the `SAVE_MATCHED` command and based on my manual testing it works fine
- `picard -e` does not prevent startup anymore (but it does not properly await the positional arguments, as I could not find any way to inhibit command execution until `len(parsed_items.files) == len(tagger.get_all_file_objects())`). Any way I've tried basically bricked Picard. So I suggest to handle it in a separate patch (like we did with Picard exit after unexpected pipe removal)
- `picard -e help` prevents startup which should be obvious
- Help messages are up to date

@phw @zas I think this PR is ready for review rn. I will handle `LOG_*` and `LOAD` commands in a separate, smaller patch.

**Comment by rdswift** [comment]
I believe that your latest commit will save all (unsaved) files matched to tracks, but this includes the case when there is more than one file matched to a track.  I sort of had in mind that it should only save properly matched files (only one file matched to a particular track).  If @zas and @phw feel it's okay as you have it currently, then I guess I'm okay with it.

**Comment by skelly37** [comment]
Some strange changes, looks like I can request only one reviewer...

> I believe that your latest commit will save all (unsaved) files matched to tracks, but this includes the case when there is more than one file matched to a track. I sort of had in mind that it should only save properly matched files (only one file matched to a particular track). If @zas and @phw feel it's okay as you have it currently, then I guess I'm okay with it.

Anyway, I'm afraid I don't understand how should I determine which file is the one that should be saved.

**Comment by rdswift** [comment]
> Anyway, I'm afraid I don't understand how should I determine which file is the one that should be saved.

From my perspective, if there's more than one file attached to a track, that is an error and none of them should be saved because, as you say, you don't know which one is correct.

**Comment by skelly37** [comment]
> > Anyway, I'm afraid I don't understand how should I determine which file is the one that should be saved.
> 
> From my perspective, if there's more than one file attached to a track, that is an error and none of them should be saved because, as you say, you don't know which one is correct.

That is a great point, thanks. I think that the recent commit solves the issue and PR might get final review and merge

**Comment by skelly37** [comment]
I'm not sure if we should apply the Codacy warning. I don't see a need to `await`/`asyncio.run` all the `Item.load` ocurrences. We await only in `tagger.load_to_picard()` to ensure the proper flow, i.e. commands won't execute before files are loaded. 

On the other hand, refactoring would fulfill the Liskov Substitution Principle, as the overriden methods should behave in the similar way. 

I'm not really sure what to do, if you find it useful to change the base class as well, I will do it.

File: picard/tagger.py [hunk 15]
```diff
@@ -851,7 +852,7 @@ def _scan_paths_recursive(paths, recursive, ignore_hidden):
             except OSError as err:
                 log.warning(err)
 
-    def add_paths(self, paths, target=None):
+    async def add_paths(self, paths, target=None):
```
zas, line 855: I don't get why this is introduced at this point. Can you explain why it is needed?

Additionally `add_paths()` is very likely to be used by external plugins, making it a coroutine will break them.

It should be noted that asyncio wasn't used at all in the code until its introduction here. I don't say we shouldn't use asyncio but I fear it breaks things.

@phw what do you think?

skelly37: I really did not want to introduce asyncio to the codebase but I could not come up with a better idea to actually **await** for the files to be loaded. If you have any better idea how to run commands after all the files are already loaded, I’ll gladly remove the asyncio.

skelly37: If i just tried adding some thread (pool) and block the function until the amount of all files in picard matches the number of arguments, I ended up with the whole app getting frozen

zas: > I really did not want to introduce asyncio to the codebase but I could not come up with a better idea to actually **await** for the files to be loaded

That's fine, I have nothing against using asyncio if it helps, I'm just worried those changes could break something.

skelly37: Actually, seems like the issue persists even with asyncio (but there's some improvement). I suppose that we might need to write some blocking file-loading flow for the sake of `picard -e LOAD` command in the next PR

skelly37: So @phw, the question for you is whether we merge at the current state or reset to https://github.com/metabrainz/picard/pull/2137/commits/4d0867faaef54d21bbde3a8d784a44029c6946df and re-apply the `picard -e SAVE_MATCHED` patch, while leaving the `asyncio` stuff for a next ticket

skelly37: After further research: I tend to reset to the mentioned commit. We need to introduce second type of all executed commands: blocking ones. This is too much for this PR imo. 

Even though we await load, there is still save, lookup, scan etc. Let's leave it for now.

skelly37: We do not want to run GUI events in one thread, but the command-line commands should be executed in a blocking way to ensure proper execution order. 

So my diagnosis is:
- each new command requires research if it uses any threaded code. If yes, it has to be cloned without this threaded code (or the futures have to be joined at some place).

And my suggestion is:
- reset to https://github.com/metabrainz/picard/commit/4d0867faaef54d21bbde3a8d784a44029c6946df, re-apply `picard -e SAVE_MATCHED` patch and merge
- in the second PR let's take care of the execution order: i.e. join threads & await functions in their clone versions for our CLI purposes.
- in the third one let's add:
    - `LOAD`
    - `LOG_DUMP`
    - `LOG_CLEAR`
    - `FROM_FILE`: (load commands from a given text file)
    
Sorry for the mess but it's really complex to me, I'm still learning the codebase.

What's your take on this @phw @zas?

zas: I agree with your suggestion.

picard/tagger.py
```diff
@@ -213,15 +293,16 @@ class Tagger(QtWidgets.QApplication):
     _debug = False
     _no_restore = False
 
-    def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
+    def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None, to_load=None):
```
skelly37, line 296: Since you've told me to keep such changes away from the feature PR, when this one gets merged, I'll refactor the class to take one arg less by embedding the remaining picard args into `to_load`

picard/tagger.py
```diff
@@ -360,9 +445,104 @@ def load_to_picard(self, items):
...
+        if parsed_items.non_executable_items():
             self.bring_tagger_front()
 
+    def get_album_pane_tracks(self):
+        for album in self.albums.values():
+            for track in album.iterfiles():
+                yield track
+
+    def get_all_file_objects(self):
+        return self.unclustered_files.files + list(self.get_album_pane_tracks()) + list(self.clusters.iterfiles())
```
zas, line 460:
```suggestion
        yield from self.unclustered_files.files
        yield from self.get_album_pane_tracks()
        yield from self.clusters.iterfiles()
```

zas (changes requested):
It looks good to be merged, improvements can come later.

Is there a remote command to remove empty clusters/albums? If not, I think we could add one.

skelly37:
> Is there a remote command to remove empty clusters/albums? If not, I think we could add one.

There's no such command and I like the idea. I'll add it to the TODO list alongside with logs, loading files and loading commands from file.

picard/tagger.py
```diff
@@ -360,9 +445,106 @@ def load_to_picard(self, items):
...
+
+    def handle_command_remove_saved(self, argstring):
+        for track in self.get_album_pane_tracks():
+            if track.state == File.NORMAL:
+                track.remove()
+
+    def handle_command_save_matched(self, argstring):
+        for album in self.albums.values():
+            for track in album.iter_correctly_matched_tracks():
+                track.files[0].save()
```
rdswift, line 532: Just to clarify, `album.iter_correctly_matched_tracks()` returns tracks with only one file attached, which is why saving only `track.files[0]` is appropriate, right?

zas (dismissed):
LGTM

rdswift (approved):
The one main change that I would like to see is replacing Line 1377 in `tagger.py` with something like (**untested**):

```
                cmd_args = e[1:] if len(e) > 1 else ['']
                for cmd_arg in cmd_args:
                    to_be_added.append("command://{0} {1}".format(e[0], cmd_arg))
```

to allow multiple files / arguments for the commands, but this can wait until the next round of changes.

phw (approved):

picard/tagger.py
```diff
@@ -360,9 +445,106 @@ def load_to_picard(self, items):
...
+        for command in parsed_items.commands:
+            self.handle_command(command)
+
+        if parsed_items.non_executable_items():
             self.bring_tagger_front()
 
+    def get_album_pane_tracks(self):
+        for album in self.albums.values():
+            for track in album.iterfiles():
+                yield track
```
phw, line 457: This could be

```python
yield from album.iterfiles()
```

Or even the entire method could also be done with:

```python
def get_album_pane_tracks(self):
   return iter_files_from_objects(self.albums)
```

Also I find the name a bit misleading, it is not yielding Track objects, but files, so maybe also name this ` iter_album_files`?

picard/tagger.py
```diff
@@ -360,9 +445,106 @@ def load_to_picard(self, items):
...
+
+    def get_all_file_objects(self):
```
phw, line 459: Can we maybe name it `iter_all_files`? Currently this is easily confused with the similar named `get_files_from_objects` (which we probably should remove anyway, but still)

phw (changes requested):

zas (dismissed):

phw (dismissed):

rdswift (dismissed):

picard/tagger.py
```diff
@@ -468,9 +649,10 @@ def exit(self):
         QtCore.QCoreApplication.processEvents()
 
     def _run_init(self):
-        if self._cmdline_files:
-            self.load_to_picard([decode_filename(f) for f in self._cmdline_files])
-            del self._cmdline_files
+        print(self._to_load)
```
phw, line 652: Just noticed this, is this a debugging left over?

skelly37:
> The one main change that I would like to see is replacing Line 1377 in `tagger.py` with something like (**untested**):
> 
> ```
>                 cmd_args = e[1:] if len(e) > 1 else ['']
>                 for cmd_arg in cmd_args:
>                     to_be_added.append("command://{0} {1}".format(e[0], cmd_arg))
> ```
> 
> to make more effective use of `argparse` and allow multiple files / arguments for the commands, but this can wait until the next round of changes.

I haven't implemented multiple files support on purpose. File paths can contain whitespaces, so it could lead to an unexpected behavior, as you have used ' ' as a separator. But anyway, `\0` is the ONLY correct separator but it's already used as a separator for the pipe messages. I prioritized working correctly in the edge cases over some kind of convenience. Unless you have some better idea or I do not understand something, I'd drop this idea.

rdswift:
> > The one main change that I would like to see is replacing Line 1377 in `tagger.py` with something like (**untested**):
> > ```
> >                 cmd_args = e[1:] if len(e) > 1 else ['']
> >                 for cmd_arg in cmd_args:
> >                     to_be_added.append("command://{0} {1}".format(e[0], cmd_arg))
> > ```
> >   
> > to make more effective use of `argparse` and allow multiple files / arguments for the commands, but this can wait until the next round of changes.
> 
> I haven't implemented multiple files support on purpose. File paths can contain whitespaces, so it could lead to an unexpected behavior, as you have used ' ' as a separator. But anyway, `\0` is the ONLY correct separator but it's already used as a separator for the pipe messages. I prioritized working correctly in the edge cases over some kind of convenience. Unless you have some better idea or I do not understand something, I'd drop this idea.

I must be missing something because I don't understand your reluctance to accept multiple arguments for a command.  If you look closely at the code I proposed, it parses the multiple arguments and creats a separate `to_be_added` item for each argument.  This should avoid any issues with trying to separate multiple arguments for passing to the pipe.

skelly37:
> > > The one main change that I would like to see is replacing Line 1377 in `tagger.py` with something like (**untested**):
> > > ```
> > >                 cmd_args = e[1:] if len(e) > 1 else ['']
> > >                 for cmd_arg in cmd_args:
> > >                     to_be_added.append("command://{0} {1}".format(e[0], cmd_arg))
> > > ```
> > > 
> > > 
> > >     
> > >       
> > >     
> > > 
> > >       
> > >     
> > > 
> > >     
> > >   
> > > to make more effective use of `argparse` and allow multiple files / arguments for the commands, but this can wait until the next round of changes.
> > 
> > 
> > I haven't implemented multiple files support on purpose. File paths can contain whitespaces, so it could lead to an unexpected behavior, as you have used ' ' as a separator. But anyway, `\0` is the ONLY correct separator but it's already used as a separator for the pipe messages. I prioritized working correctly in the edge cases over some kind of convenience. Unless you have some better idea or I do not understand something, I'd drop this idea.
> 
> I must be missing something because I don't understand your reluctance to accept multiple arguments for a command. If you look closely at the code I proposed, it parses the multiple arguments and creats a separate `to_be_added` item for each argument. This should avoid any issues with trying to separate multiple arguments for passing to the pipe.

`picard -e my/fi le.mp3 e xa/mple.mp3` -> How to determine the correct file names via argparse?

rdswift:
> `picard -e my/fi le.mp3 e xa/mple.mp3` -> How to determine the correct file names via argparse?

Arguments which include spaces should be quoted.  That's standard behavior.  For example, they should be entered as:

`picard -e LOAD "my/fi le.mp3" "e xa/mple.mp3"`

In addition, have you tested your current processing if a file path contains more than one consecutive space?  I think you'll find that multiple spaces will be collapsed to a single space.  This will not be the case if the arguments containing spaces (single or multiple) are enclosed within quotes.

skelly37:
Okay, this way. Sorry, my brain got a lag :) Thanks for the suggestion, now I agree it has to be done.

PR #2138, opened 2022-08-18
PICARD-2520: Null bytes in tag values can cause crashes when sorting

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2520

When a audio file contains unexpected null bytes in the tag values this can cause Picard to crash when sorting by this tag value. This is due `strxfrm`  not allowing null bytes as part of the input.

# Solution
strxfrm fails if a string contains a null char. Sanizize the input strings for strxfrm by removing any null chars.

# Action

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -1086,8 +1086,9 @@ def any_exception_isinstance(error, type_):
 def strxfrm(string):
     """Transforms a string to one that can be used in locale-aware comparisons.
 
-    Wrapper around locale.strxfrm, that never throws OSError. If an OSError
-    occurs this function will return the string converted to lower case.
+    Wrapper around locale.strxfrm, that never throws OSError or ValueError. If an
+    error occurs this function will return the string converted to lower case.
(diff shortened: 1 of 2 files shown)


phw:
@zas What do you think about PICARD-2534? Should we maybe implement removal of trailing null chars centrally in `Metadata`?

zas:
> @zas What do you think about [PICARD-2534](https://tickets.metabrainz.org/browse/PICARD-2534)? Should we maybe implement removal of trailing null chars centrally in `Metadata`?

Is there any drawback doing so?
I don't think null chars are of any use in the context of metadata, so implementing centrally is perhaps even better as it would prevent (yet undetected) crashes (as the one fixed here).

zas (approved):

phw:
> Is there any drawback doing so?

I don't think so, for me it also just makes sense. Can't rule out there might be some obscure case, though. But we should just go for it.

PR #2139, opened 2022-08-29
Refactor handling of Picard command-line args

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
I have removed the need to duplicate the arguments passed during Tagger's initialization

# Problem

After merging the PR introducing the command-line arguments processing, Tagger now gets both unparsed arguuments & parsed items to load by the app (files, MBIDs, URLs & commands)

# Solution

- Merge command-line arguments and remote commands into `picard_args`.
- Name pipe with an unique identifier

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1405,7 +1405,7 @@ def main(localedir=None, autoupdate=True):
     if picard_args.stand_alone_instance:
         identifier = uuid4().hex
     else:
-        identifier = md5(picard_args.config_file.encode('utf8')).hexdigest() if picard_args.config_file else 'main'
+        identifier = md5(picard_args.config_file.encode('utf8')).hexdigest() if picard_args.config_file else 'main'  # nosec: B303
         identifier += '_NP' if picard_args.no_plugins else ''
 
     if picard_args.processable:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1280,7 +1280,7 @@ class PicardArgs:
                 if "HELP" in [x.upper().strip() for x in e]:
                     print_help_for_commands()
                     sys.exit(0)
-                args = e[1:] if len(e) > 1 else ['']
+                args = e[1:] or ['']
                 for arg in args:
                     self.processable.append(f"command://{e[0]} {arg}")
 
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1287,7 +1287,7 @@ class PicardArgs:
             self.load_from_argparse(argparse_args)
 
     def load_from_argparse(self, argparse_args):
-        self.CONFIG_FILE = argparse_args.config
+        self.CONFIG_FILE = argparse_args.config_file
         self.DEBUG = argparse_args.debug
         self._EXEC = argparse_args.exec
         self.NO_PLAYER = argparse_args.no_player
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -364,7 +364,7 @@ class Tagger(QtWidgets.QApplication):
         log.debug("Starting Picard from %r", os.path.abspath(__file__))
         log.debug("Platform: %s %s %s", platform.platform(),
                   platform.python_implementation(), platform.python_version())
-        log.debug("Versions: %s", picard_args.long_version)
+        log.debug("Versions: %s", versions.as_string())
         log.debug("Configuration file path: %r", config.fileName())
 
         log.debug("User directory: %r", os.path.abspath(USER_DIR))
@@ -1248,60 +1248,6 @@ class Tagger(QtWidgets.QApplication):
         self.signalnotifier.setEnabled(True)
 
 
-class PicardArgs:
-
-    def __init__(self, argparse_args):
-        self.__load_from_argparse(argparse_args)
-
-    def __load_from_argparse(self, argparse_args):
-        self.config_file = argparse_args.config_file
-        self.debug = argparse_args.debug
-        self._exec = argparse_args.exec
-        self.no_player = argparse_args.no_player
-        self.no_restore = argparse_args.no_restore
-        self.no_plugins = argparse_args.no_plugins
-        self.no_crash_dialog = argparse_args.no_crash_dialog
-        self.stand_alone_instance = argparse_args.stand_alone_instance
-        self._file_or_url = argparse_args.FILE_OR_URL
-        self.long_version = versions.as_string()
-
-        self.version = ""
-        self.processable = []
-
-        self.version = self.__get_version(short=argparse_args.version, long=argparse_args.long_version)
-        self.__parse_loadable_items()
-
-    # Add string output for debugging purposes
-    def __str__(self):
-        attributes = ('config_file', 'debug', 'no_player', 'no_restore', 'no_plugins',
-            'no_crash_dialog', 'stand_alone_instance', 'version', 'processable')
-        return str({a: getattr(self, a, None) for a in attributes})
-
-    def __parse_loadable_items(self):
-        for x in self._file_or_url:
-            if not urlparse(x).netloc:
-                x = os.path.abspath(x)
-            self.processable.append(x)
-
-        if self._exec:
-            for e in self._exec:
-                if "HELP" in [x.upper().strip() for x in e]:
-                    print_help_for_commands()
-                    sys.exit(0)
-                args = e[1:] or ['']
-                for arg in args:
-                    self.processable.append(f"command://{e[0]} {arg}")
-
-        del self._exec
-        del self._file_or_url
-
-    def __get_version(self, short, long):
-        if short:
-            return f"{PICARD_ORG_NAME} {PICARD_APP_NAME} {PICARD_FANCY_VERSION_STR}"
-        elif long:
-            return self.long_version
-
-
 def print_help_for_commands():
     maxwidth = 80
     helpcmd = []
@@ -1370,7 +1316,23 @@ If a new instance will not be spawned:
     parser.add_argument('FILE_OR_URL', nargs='*',
                         help="the file(s), URL(s) and MBID(s) to load")
 
-    return parser.parse_known_args()[0]
+    args = parser.parse_args()
+    args.remote_commands_help = False
+
+    args.processable = []
+    for x in args.FILE_OR_URL:
+        if not urlparse(x).netloc:
+            x = os.path.abspath(x)
+        args.processable.append(x)
+
+    if args.exec:
+        for e in args.exec:
+            args.remote_commands_help = args.remote_commands_help or "HELP" in {x.upper().strip() for x in e}
+            remote_command_args = e[1:] or ['']
+            for arg in remote_command_args:
+                args.processable.append(f"command://{e[0]} {arg}")
+
+    return args
 
 
 def main(localedir=None, autoupdate=True):
@@ -1396,10 +1358,14 @@ def main(localedir=None, autoupdate=True):
 
     signal.signal(signal.SIGINT, signal.SIG_DFL)
 
-    picard_args = PicardArgs(process_picard_args())
+    picard_args = process_picard_args()
 
+    if picard_args.long_version:
+        return print(versions.as_string())
     if picard_args.version:
-        return print(picard_args.version)
+        return print(f"{PICARD_ORG_NAME} {PICARD_APP_NAME} {PICARD_FANCY_VERSION_STR}")
+    if picard_args.remote_commands_help:
+        return print_help_for_commands()
 
     # any of the flags that change Picard's workflow significantly should trigger creation of a new instance
     if picard_args.stand_alone_instance:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1278,7 +1278,7 @@ class PicardArgs:
                 if "HELP" in [x.upper().strip() for x in e]:
                     print_help_for_commands()
                     sys.exit(0)
-                args = e[1:] if len(e) > 1 else []
+                args = e[1:] if len(e) > 1 else ['']
                 for arg in args:
                     self.processable.append(f"command://{e[0]} {arg}")
 
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -297,19 +297,19 @@ class Tagger(QtWidgets.QApplication):
 
         super().__init__(sys.argv)
         self.__class__.__instance = self
-        setup_config(self, picard_args.CONFIG_FILE)
+        setup_config(self, picard_args.config_file)
         config = get_config()
         theme.setup(self)
 
-        self._to_load = picard_args.PROCESSABLE
+        self._to_load = picard_args.processable
 
         self.autoupdate_enabled = autoupdate
-        self._no_restore = picard_args.NO_RESTORE
-        self._no_plugins = picard_args.NO_PLUGINS
+        self._no_restore = picard_args.no_restore
+        self._no_plugins = picard_args.no_plugins
 
         self.set_log_level(config.setting['log_verbosity'])
 
-        if picard_args.DEBUG or "PICARD_DEBUG" in os.environ:
+        if picard_args.debug or "PICARD_DEBUG" in os.environ:
             self.set_log_level(logging.DEBUG)
 
         # Default thread pool
@@ -411,7 +411,7 @@ class Tagger(QtWidgets.QApplication):
         self.mbid_redirects = {}
         self.unclustered_files = UnclusteredFiles()
         self.nats = None
-        self.window = MainWindow(disable_player=picard_args.NO_PLAYER)
+        self.window = MainWindow(disable_player=picard_args.no_player)
         self.exit_cleanup = []
         self.stopping = False
 
@@ -1253,47 +1253,47 @@ class PicardArgs:
             self.load_from_argparse(argparse_args)
 
     def load_from_argparse(self, argparse_args):
-        self.CONFIG_FILE = argparse_args.config_file
-        self.DEBUG = argparse_args.debug
-        self._EXEC = argparse_args.exec
-        self.NO_PLAYER = argparse_args.no_player
-        self.NO_RESTORE = argparse_args.no_restore
-        self.NO_PLUGINS = argparse_args.no_plugins
-        self.NO_CRASH_DIALOG = argparse_args.no_crash_dialog
-        self.STAND_ALONE_INSTANCE = argparse_args.stand_alone_instance
-        self.VERSION = argparse_args.version
-        self.LONG_VERSION = argparse_args.long_version
-        self._FILE_OR_URL = argparse_args.FILE_OR_URL
-        self.PROCESSABLE = []
+        self.config_file = argparse_args.config_file
+        self.debug = argparse_args.debug
+        self._exec = argparse_args.exec
+        self.no_player = argparse_args.no_player
+        self.no_restore = argparse_args.no_restore
+        self.no_plugins = argparse_args.no_plugins
+        self.no_crash_dialog = argparse_args.no_crash_dialog
+        self.stand_alone_instance = argparse_args.stand_alone_instance
+        self.version = argparse_args.version
+        self.long_version = argparse_args.long_version
+        self._file_or_url = argparse_args.FILE_OR_URL
+        self.processable = []
 
         self.__parse_loadable_items()
 
     def __parse_loadable_items(self):
-        for x in self._FILE_OR_URL:
+        for x in self._file_or_url:
             if not urlparse(x).netloc:
                 x = os.path.abspath(x)
-            self.PROCESSABLE.append(x)
+            self.processable.append(x)
 
-        if self._EXEC:
-            for e in self._EXEC:
+        if self._exec:
+            for e in self._exec:
                 if "HELP" in [x.upper().strip() for x in e]:
                     print_help_for_commands()
                     sys.exit(0)
                 args = e[1:] if len(e) > 1 else []
                 for arg in args:
-                    self.PROCESSABLE.append(f"command://{e[0]} {arg}")
+                    self.processable.append(f"command://{e[0]} {arg}")
 
-        del self._EXEC
-        del self._FILE_OR_URL
+        del self._exec
+        del self._file_or_url
 
     def __delete_version_args(self):
-        del self.VERSION
-        del self.LONG_VERSION
+        del self.version
+        del self.long_version
 
     def get_version(self):
-        if self.VERSION:
+        if self.version:
             return f"{PICARD_ORG_NAME} {PICARD_APP_NAME} {PICARD_FANCY_VERSION_STR}"
-        elif self.LONG_VERSION:
+        elif self.long_version:
             return versions.as_string()
         else:
             self.__delete_version_args()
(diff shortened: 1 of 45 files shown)


skelly37:
TODO: Support multiple arguments in commands like `picard -e LOAD`
REF: https://github.com/metabrainz/picard/pull/2137#issuecomment-1230697994

skelly37:
@zas @phw Enjoy!

@rdswift I have also included your suggestion here so you might want to take a look

skelly37:
Everything after [d329ff0](https://github.com/metabrainz/picard/pull/2139/commits/d329ff0df5b3a5efb587775df0d5232b5ef54aac) (except [6a70deb](https://github.com/metabrainz/picard/pull/2139/commits/6a70debde6ffd9eb497d540226fde15479ad448c) of course) is a questionable suggestion, I can remove some commit, no hard feelings

picard/tagger.py
```diff
@@ -1062,7 +1062,7 @@ def _lookup_disc(self, disc, result=None, error=None):
         self.restore_cursor()
         if error is not None:
             QtWidgets.QMessageBox.critical(self.window, _("CD Lookup Error"),
-                                           _("Error while reading CD:\n\n%s") % error)
+                                           _("Error while reading CD:\n\n%s"), error)
```
zas, line 1065: I think this change is incorrect, our log.* functions support passing format arguments but here that's a Qt message box.

picard/tagger.py
```diff
@@ -1246,12 +1246,58 @@ def sighandler(self):
...
+        self._EXEC = argparse_args.exec
+        self.NO_PLAYER = argparse_args.no_player
+        self.NO_RESTORE = argparse_args.no_restore
+        self.NO_PLUGINS = argparse_args.no_plugins
+        self.NO_CRASH_DIALOG = argparse_args.no_crash_dialog
+        self.STAND_ALONE_INSTANCE = argparse_args.stand_alone_instance
+        self.VERSION = argparse_args.version
+        self.LONG_VERSION = argparse_args.long_version
+        self._FILE_OR_URL = argparse_args.FILE_OR_URL
+        self.PROCESSABLE = []
```
zas, line 1267: Why capitalized attribute's names?

I have the feeling we could subclass `argparse_args` object and extends it, rather than copying all of its attributes.

picard/tagger.py
```diff
@@ -1246,12 +1246,58 @@ def sighandler(self):
...
+        del self.VERSION
+        del self.LONG_VERSION
+
+    def get_version(self):
+        if self.VERSION:
+            return f"{PICARD_ORG_NAME} {PICARD_APP_NAME} {PICARD_FANCY_VERSION_STR}"
+        elif self.LONG_VERSION:
+            return versions.as_string()
+        else:
+            self.__delete_version_args()
```
zas, line 1299: Why `del` those attributes?

picard/tagger.py
```diff
@@ -1348,39 +1394,26 @@ def main(localedir=None, autoupdate=True):
...
-    if picard_args.version:
-        return version()
-    if picard_args.long_version:
-        return longversion()
+    picard_args = PicardArgs(process_picard_args())
+
+    version = picard_args.get_version()
+    if version:
+        return version
+    del version
```
zas, line 1402: Why `del`?

picard/tagger.py
```diff
@@ -1392,6 +1425,9 @@ def main(localedir=None, autoupdate=True):
             log.debug("No need for spawning a new instance, exiting...")
             # just a custom exit code to show that picard instance wasn't created
             sys.exit(EXIT_NO_NEW_INSTANCE)
+
+        del should_start
```
zas, line 1429: Why `del`?

picard/tagger.py
```diff
@@ -1402,7 +1438,10 @@ def main(localedir=None, autoupdate=True):
     except ImportError:
         pass
 
-    tagger = Tagger(picard_args, localedir, autoupdate, pipe_handler=pipe_handler, to_load=to_be_added)
+    tagger = Tagger(picard_args, localedir, autoupdate, pipe_handler=pipe_handler)
+
+    del pipe_handler
+    del picard_args
```
zas, line 1444: Why `del`?

zas (changes requested):

picard/tagger.py
skelly37: Right, my overlook

picard/tagger.py
skelly37: >Why capitalized attribute's names?
To show they're constant


>I have the feeling we could subclass argparse_args object and extends it, rather than copying all of its attributes.
I think it would compound the readability and simplicity of this class

picard/tagger.py
skelly37: Not needed further, minor memory optimization, as all other `del`s. What you're trying to say that explicit freeing the memory ASAP is not the pythonic way, right? But they are really not needed during runtime, yet they'd be deleted after the GUI is closed (unless I misunderstand something in Python's memory management). The same goes for all other `del`s

zas: Well, `del` doesn't free memory by itself. That's not like a C `free()`. It marks objects as unused, and memory will eventually be freed when Python garbage collector runs.

It makes sense to call `del` to remove an unused variable in some cases, but IMHO you shouldn't bother too much (but perhaps if the variable contains a LOT of data). Or move the variable to a function, call the function, and variable will be freed on return.

picard/tagger.py
zas: Even though there is no strict definition of a constant in Python (well, they are _usually_ defined at module level), those are normal class attributes, so I don't think they should be capitalized.
A class constant can be capitalized, but it is usually never modified, that's not the case here for `PROCESSABLE` which is initialized then modified.
I don't think upper-cased names add anything here.

picard/tagger.py
```diff
@@ -1246,12 +1246,58 @@ def sighandler(self):
...
+            self.PROCESSABLE.append(x)
+
+        if self._EXEC:
+            for e in self._EXEC:
+                if "HELP" in [x.upper().strip() for x in e]:
+                    print_help_for_commands()
+                    sys.exit(0)
+                args = e[1:] if len(e) > 1 else []
+                for arg in args:
+                    self.PROCESSABLE.append(f"command://{e[0]} {arg}")
```
rdswift, line 1284: By defaulting to an empty list rather than `['']` I think that any commands without arguments such as `-e CLUSTER_ALL` will not be added to `self.PROCESSABLE`.  That's why my suggestion was to default to `['']`.

picard/tagger.py
```diff
@@ -1246,12 +1246,58 @@ def sighandler(self):
         self.signalnotifier.setEnabled(True)
 
 
-def version():
-    print("%s %s %s" % (PICARD_ORG_NAME, PICARD_APP_NAME, PICARD_FANCY_VERSION_STR))
+class PicardArgs:
 
+    def __init__(self, argparse_args=None):
```
zas, line 1251: Is there a reason to make `argparse_args` optional?

picard/tagger.py
```diff
@@ -1246,12 +1246,58 @@ def sighandler(self):
...
+        if argparse_args is not None:
+            self.load_from_argparse(argparse_args)
 
-def longversion():
-    print(versions.as_string())
+    def load_from_argparse(self, argparse_args):
```
zas, line 1255: This method is only called from `__init__()` so the code can be moved there and this method removed.

zas (changes requested):

picard/tagger.py
skelly37: Well, actually no.

po/appstream/en_CA.po
```diff
@@ -9,7 +9,7 @@
 #, fuzzy
 msgid ""
 msgstr ""
-"Project-Id-Version: PACKAGE VERSION\n"
+"Project-Id-Version: PACKAGE version\n"
```
zas, line 12: This shouldn't have been changed ;)

zas (changes requested):

picard/tagger.py
```diff
@@ -1246,12 +1246,52 @@ def sighandler(self):
...
 
-def version():
-    print("%s %s %s" % (PICARD_ORG_NAME, PICARD_APP_NAME, PICARD_FANCY_VERSION_STR))
+class PicardArgs:
 
+    def __init__(self, argparse_args):
+        self.__load_from_argparse(argparse_args)
 
-def longversion():
-    print(versions.as_string())
```
zas, line 1254: Why is this disappeared?

picard/tagger.py
```diff
@@ -1246,12 +1246,52 @@ def sighandler(self):
         self.signalnotifier.setEnabled(True)
 
 
-def version():
-    print("%s %s %s" % (PICARD_ORG_NAME, PICARD_APP_NAME, PICARD_FANCY_VERSION_STR))
```
zas, line 1250: Why is this disappeared?
Those methods are required.

zas (changes requested):

po/appstream/en_CA.po
skelly37: A little fckup provided by PyCharm (: Gimme 1min

picard/tagger.py
skelly37: Where is it used? I cannot find any usage

picard/tagger.py
skelly37: Where is it used? I cannot find any usage

skelly37: So basically that's why I moved the logic to `PicardArgs` :)

rdswift: I believe that move is also the reason for the failing tests.

zas: > Where is it used? I cannot find any usage

On master:

```bash
❯ python tagger.py -V             
Picard 2.9.0.dev1, Python 3.10.4, PyQt 5.15.7, Qt 5.15.2, Mutagen 1.45.1, Discid discid 1.2.0, libdiscid 0.6.2, astrcmp Python, SSL OpenSSL 1.1.1m  14 Dec 2021
❯ python tagger.py -v
MusicBrainz Picard 2.9.0.dev1

```

On this branch, behavior changed and Picard starts instead.

rdswift: The problem is that you changed lines 1353-1355.

skelly37: Okay, I see that I forgot to assign the value to the `PicardArgs.version` and it always remained the default `""` empty string

rdswift: What output do you get when you start it with the `--long-version ` command line option?

skelly37: @rdswift It works exactly the same for the end user, but the functions are no longer stand-alone code but wrapped in the class.

skelly37: See PicardArgs. __get_version(), PicardArgs.version value depends on the flag provided by the user.

rdswift: Okay, thanks.

rdswift: Thinking about this a bit more...  Not that it makes a big difference, but I sort of question the need to store the version string in a variable when it is only used once.  Then again, maybe it is stored as an empty string if neither `-v` nor `-V` are specified on the command line, and Picard exits after displaying the string when `-v` or `-V` are specified.

skelly37: > I sort of question the need to store the version string in a variable when it is only used once. Then again, maybe it is stored as an empty string if neither -v nor -V are specified on the command line, and Picard exits after displaying the string when -v or -V are specified.

If we wanted to introduce just `PicardArgs.get_version()`, we'd end up having to store the booleans from `argparse` to determine which wersion actually return, so we still end up with 2 variables. And since @zas advised me not to overuse `del`, I left it there.

picard/tagger.py
```diff
@@ -1246,12 +1246,52 @@ def sighandler(self):
...
+            if not urlparse(x).netloc:
+                x = os.path.abspath(x)
+            self.processable.append(x)
+
+        if self._exec:
+            for e in self._exec:
+                if "HELP" in [x.upper().strip() for x in e]:
+                    print_help_for_commands()
+                    sys.exit(0)
+                args = e[1:] if len(e) > 1 else ['']
```
zas, line 1283:
```suggestion
                args = e[1:] or ['']
```

```python
>>> e = []
>>> e[1:] if len(e) > 1 else ['']
['']
>>> e[1:] or ['']
['']
>>> e = ['a']
>>> e[1:] if len(e) > 1 else ['']
['']
>>> e[1:] or ['']
['']
>>> e = ['a', 'b']
>>> e[1:] if len(e) > 1 else ['']
['b']
>>> e[1:] or ['']
['b']
```

picard/tagger.py
```diff
@@ -1246,12 +1246,52 @@ def sighandler(self):
...
+                for arg in args:
+                    self.processable.append(f"command://{e[0]} {arg}")
+
+        del self._exec
+        del self._file_or_url
```
zas, line 1288: Those `del` are likely useless, since they point to data from `argparse`.

Try this example which tries to simulate what we have here:

`pip install memory_profiler`

```python
from memory_profiler import profile


class A:
    def __init__(self):
        self.v = "a" * 1000000


class B:
    def __init__(self, obj):
        self._v = obj.v

    def clear_v(self):
        del self._v


@profile
def main():
    obj_a = A()
    obj_b = B(obj_a)
    obj_b.clear_v()
    del obj_b
    del obj_a


if __name__ == '__main__':
    main()
```

```
Line #    Mem usage    Increment  Occurrences   Line Contents
=============================================================
    17     35.2 MiB     35.2 MiB           1   @profile
    18                                         def main():
    19     36.3 MiB      1.0 MiB           1       obj_a = A()
    20     36.3 MiB      0.0 MiB           1       obj_b = B(obj_a)
    21     36.3 MiB      0.0 MiB           1       obj_b.clear_v()
    22     36.3 MiB      0.0 MiB           1       del obj_b
    23     35.4 MiB     -0.9 MiB           1       del obj_a
```

It is important to understand that Python only copies data if needed, and it uses references to this data.
In the case here, when you write `self._exec = argparse_args.exec` you just create a reference to the data initially stored by `argparse_args.exec`, so calling `del self._exec` don't actually free the memory used by the data, at least until you also `del argparse_args.exec`.

If we'd go for memory usage optimization it is rather elsewhere I'd look first, mainly in all the code processing data from MB or cover art for (possibly) millions of tracks... The code here is executed only once.

Note that I'm not against memory usage optimization, but to be clear, Python isn't the language for that... and if you really want to do that, use a memory profiler.

zas (changes requested):

zas (changes requested):

zas:
See https://github.com/skelly37/picard/pull/8

rdswift:
See https://github.com/skelly37/picard/pull/9

picard/tagger.py
```diff
@@ -208,7 +208,7 @@ def __bool__(self):
         return bool(self.commands or self.files or self.mbids or self.urls)
 
     def __str__(self):
-        return "files: %r mbids: %r urls: %r commands: %r" % (self.files, self.mbids, self.urls, self.commands)
+        return f"files: {self.files}  mbids: f{self.mbids}  urls: {self.urls}  commands: {self.commands}"
```
zas, line 211: Minor thing, but replacement  for `"%r" % obj` is `f"{repr(obj)}` (or `f"{obj!r}'`).
Not sure this change was intentional, in many cases the result is the same, but not always.

But anyway using Python3 formatted strings is better, so +1 from me.

skelly37: Right, thanks

picard/tagger.py
skelly37: Really well explained and helpful, thanks

picard/tagger.py
```diff
@@ -1348,52 +1396,35 @@
...
-        if picard_args.exec:
-            for e in picard_args.exec:
-                if "HELP" in [x.upper().strip() for x in e]:
-                    print_help_for_commands()
-                    sys.exit(0)
-                to_be_added.append("command://" + " ".join(e))
+    if picard_args.stand_alone_instance:
+        identifier = uuid4().hex
+    else:
+        identifier = md5(picard_args.config_file.encode('utf8')).hexdigest() if picard_args.config_file else 'main'
```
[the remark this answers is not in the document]
zas: @skelly37 we may add `# nosec: B303` to this line, here `md5()` use is intended.

picard/tagger.py
```diff
@@ -801,14 +801,14 @@ def add_files(self, filenames, target=None):
         for filename in filenames:
             filename = normpath(filename)
             if ignore_hidden and is_hidden(filename):
-                log.debug("File ignored (hidden): %r" % (filename))
+                log.debug("File ignored (hidden): %r", filename)
```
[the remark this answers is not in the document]
skelly37: So should I revert all the f-string refactors and then we can merge?

picard/util/pipe.py
```diff
@@ -137,6 +139,8 @@ def __init__(self, app_name: str, app_version: str, args: Optional[Iterable[str]
         if not isinstance(app_name, str) or not isinstance(app_version, str):
             raise PipeErrorInvalidAppData
 
+        self._identifier = identifier.replace(' ', '_') if identifier else 'main'
```
zas, line 142: @rdswift can `identifier` contain spaces the way it is build atm?

picard/util/pipe.py
```diff
@@ -190,7 +194,7 @@ def __generate_filenames(self, app_name: str, app_version: str) -> List[str]:
         for directory in self.PIPE_DIRS:
             if directory:
                 _pipe_names.append(os.path.join(os.path.expanduser(directory),
-                                                f"{app_name}_v{app_version}_pipe_file"))
+                                                f"{app_name}_v{app_version}_{self._identifier}_pipe_file"))
```
zas, line 197: For extra safety we could wrap filename with `sanitize_filename()` (which replaces path separators).

zas (approved):

picard/tagger.py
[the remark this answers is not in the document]
skelly37: btw, a suggestion+question

Is there any reason why resolving all conversations is not required before merging? See: 
image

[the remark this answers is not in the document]
zas: Keep them, those are minor changes, but next time try to keep unrelated changes outside ;)

picard/util/pipe.py
skelly37: You mean this? https://pypi.org/project/sanitize-filename/

picard/util/pipe.py
rdswift: No.  Good catch.  I had originally included the `.replace()` because I was first passing the whole config file path (which could contain spaces) but then decided to use the hash and forgot to remove the `.replace()`.

skelly37: I don't see any way to add spaces to the identifier but I think it is a good security feature that may prevent future errors.

skelly37: Or we can simply add a test case for pipe where we would assert that this variable contains no spaces, which gives better feedback for the developer

picard/tagger.py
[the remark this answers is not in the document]
skelly37: > btw, a suggestion+question
> 
> Is there any reason why resolving all conversations is not required before merging? See: image

ping @zas to ensure that you have seen the question (as the time difference is small and you could've not refresh).

picard/util/pipe.py
skelly37: I mean, the question is up to you @rdswift 

Could whitespaces in this string potentially break Picard?

If yes, I will add a proper test case. If not, I will remove the `.replace()` call :)

picard/util/pipe.py
rdswift: Nope.  See Line 314 in `util/__init__.py`.

picard/util/pipe.py
rdswift: I actually added it because I wasn't sure if a space would break the pipe name.  :stuck_out_tongue_winking_eye:   If that's not a concern, then there's likely no reason to keep it.

picard/util/pipe.py
skelly37: > Nope. See Line 314 in `util/__init__.py`.

Thanks

---

So you suggest to call the sanitizer on the `os.path.join()` result, right?

zas:
```suggestion
                                                sanitize_filename(f"{app_name}_v{app_version}_{self._identifier}_pipe_file")))
```

rdswift: I *think* he's suggesting to wrap `f"{app_name}_v{app_version}_{self._identifier}_pipe_file"`.  If you whap the whole join it would remove the path separators from the name that you're appending.

picard/tagger.py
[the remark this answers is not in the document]
zas: Well, no particular reason, I can enable it, and if it becomes annoying we can always disable it.
@phw^^

picard/util/pipe.py
zas: > So you suggest to call the sanitizer on the `os.path.join()` result, right?

Nope, only on the filename, see suggestion above.

picard/util/pipe.py
```diff
@@ -194,7 +194,7 @@ def __generate_filenames(self, app_name: str, app_version: str) -> List[str]:
         for directory in self.PIPE_DIRS:
             if directory:
                 _pipe_names.append(os.path.join(os.path.expanduser(directory),
-                                                f"{app_name}_v{app_version}_{self._identifier}_pipe_file"))
+                                                sanitize_filename(f"{app_name}_v{app_version}_{self._identifier}_pipe_file")))
```
zas, line 198: You need to import it ;)

zas (changes requested):

picard/util/pipe.py
skelly37: Already done ;) https://github.com/metabrainz/picard/pull/2139/commits/6c510ed814b77fa9d6e7b4f6387bfef60a0d7ad5

skelly37:
Sorry, I'm not on my main machine and always forget to set up the pre-commit hook on machines for journey...

zas (approved):
LGTM

rdswift (commented):
This is perhaps a really minor point, but does the comment at Line 322 of `tagger.py` still apply?

skelly37:
> This is perhaps a really minor point, but does the comment at Line 322 of `tagger.py` still apply? Otherwise, looks good to me.

You are right, it should be removed, thanks for noticing. And imho, it's not *a really minor point*. Outdated comments can cause a big mess and I strongly advise everyone not to ignore them.

rdswift (approved):

skelly37:
These MacOS 3.7 tests (and only them) tend to break rarely. I think this should be researched some time.

skelly37:
picard --help also needs to be cleaned from (starts a stand-alone instance) artifacts. I have just noticed it so please do not merge before such commit :)

phw (dismissed):
This is a good improvement. The code looks cleaner that way again.

Given the changes I think this could be put into a single commit, we should use squash on merge here on Github.

skelly37:
@rdswift please take a quick look if I have documented your changes fully correctly in the last commit :)

rdswift (approved):
The help changes look good to me.  Well done.

zas (approved):

zas:
Grrr, I rebased instead of squashing commits. Well, that's done.

PR #2140, opened 2022-08-30
PICARD-2539: Avoid crashes on repeated calls to Album._finalize_loading

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2539

`Album._finalize_loading` is supposed to be called only once after all requests have finished. But wrong usage, e.g. by plugins, can cause additional calls.

# Solution

Make sure the method exits early if called after loading the album has finished and log this event for further debugging.

I was unable to fully reproduce the situation as it happened on PICARD-2539. But I could easily trigger it by adding an additional call to `Album._finalize_loading`, which is something that can easily happen in a plugin. It makes sense to have this function be more fault tolerant.

The additional logging of the caller should make it easier to find and fix causes for these additional calls. We can't rule out currently that there is maybe a bug in Picard itself, that can lead to a situation like reported in the ticket (even though I don't see how this would happen looking at the source).

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -528,13 +528,27 @@ class Album(DataObject, Item):
             self.tagger.window.refresh_metadatabox()
 
     def _finalize_loading(self, error):
+        if self.loaded:
+            # This is not supposed to happen, _finalize_loading should only
+            # be called once after all requests finished.
+            import inspect
+            stack = inspect.stack()
+            args = [self]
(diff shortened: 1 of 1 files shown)


zas (approved):
I couldn't reproduce it either, but your explanation and fix make sense.
Extra logging is welcome.

PR #2142, opened 2022-09-04
Added command: LOAD

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
A command that allows to load files as a part of the command pipeline

# Problem

It would be useful to load a file after some commands have already been executed as a part of the pipeline. The `load` command extends positional arguments' functionality.

# Solution

Added the needed command

# Action
Should be squashed on merge

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -490,9 +490,6 @@ class Tagger(QtWidgets.QApplication):
             log.error("Cannot LOAD a command: %s", argstring)
             return
 
-        if not argstring.startswith("file://") and not argstring.startswith("mbid://") and not argstring.startswith("url://"):
-            argstring = "file://" + argstring
-
         self.load_to_picard((argstring,))
 
     def handle_command_lookup(self, argstring):
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -229,6 +229,10 @@ REMOTE_COMMANDS = {
         "handle_command_fingerprint",
         help_text="Calculate acoustic fingerprints for all (matched) files in the album pane.",
     ),
+    "LOAD": RemoteCommand(
+        "handle_command_load",
+        help_text="Load 1 or more files to Picard.",
+    ),
     "LOOKUP": RemoteCommand(
         "handle_command_lookup",
         help_text="Lookup all clusters in the cluster pane.",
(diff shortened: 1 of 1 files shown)


rdswift:
Could this be used to load URLs and MBIDs as well?  If so you might be better off passing the argument list to `load_t-Picard()` for processing rather than adding the arguments directly using `add_paths()`.

skelly37:
> Could this be used to load URLs and MBIDs as well? If so you might be better off passing the argument list to `load_to_Picard()` for processing rather than adding the arguments directly using `add_paths()`.

Could be, I wasn't sure if it should be added. What's your take on this @zas @phw ?

rdswift:
> Could be, I wasn't sure if it should be added. What's your take on this @zas @phw ?

My thinking was that it should operate the same as files, etc. entered on the command line, as described by @zas in his comment on https://github.com/metabrainz/picard/pull/2137.

> > So you're saying that we should ditch the positional argument FILE_OR_URL in favor of adding a LOAD command which would be the only way to load something to picard?

> Nope, in addition. Basically picard <FILE> would be equivalent to picard -e LOAD <FILE>. The idea is to let user use only -e commands to achieve the same thing.

skelly37:
Ping @phw @zas , fully done imo

picard/tagger.py
```diff
@@ -480,6 +485,16 @@ def handle_command_fingerprint(self, argstring):
         for album_name in self.albums:
             self.analyze(self.albums[album_name].iterfiles())
 
+    def handle_command_load(self, argstring):
+        if argstring.startswith("command://"):
+            log.error("Cannot LOAD a command: %s", argstring)
+            return
+
+        if not argstring.startswith("file://") and not argstring.startswith("mbid://") and not argstring.startswith("url://"):
+            argstring = "file://" + argstring
```
zas, line 494: This prevents something like `-e load https://musicbrainz.org/release/f1e04a39-65fc-48c9-b9f9-fa1dbe7e9326`.
Instead a check like the one at https://github.com/metabrainz/picard/blob/974547ce32718b35d8da85445887856fa4062fb5/picard/tagger.py#L1335 could be used.

zas (changes requested):

picard/tagger.py
skelly37: My bad, seems like I've gone too automatic... Tho I'd rather keep it consistent with our `ParseItemsToLoad` class, see yourself

skelly37: Actually, none of it is needed. See the recent commit.

zas (approved):
LGTM

phw (approved):

PR #2143, opened 2022-09-04
Added command: FROM_FILE

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
It would be useful to keep command pipeline in a reusable file

# Problem

Some operations are repeatable so using such file can greatly automate the workflow

# Solution

1. Added the command that takes a file path and iterates all over its lines to process the commands

# Action
Should be squashed on merge

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -486,24 +486,26 @@ class Tagger(QtWidgets.QApplication):
         for album_name in self.albums:
             self.analyze(self.albums[album_name].iterfiles())
 
-    def handle_command_from_file(self, argstring):
+    @staticmethod
+    def _read_lines_from_file(filepath):
         try:
-            lines = [line.strip() for line in open(argstring).readlines()]
-        except OSError:
-            log.error("Error opening command file: %s", argstring)
-            return
-        except Exception:
-            log.error("Error reading command file: %s", argstring)
-            return
+            yield from (line.strip() for line in open(filepath).readlines())
+        except Exception as e:
+            log.error("Error reading command file '%s': %s" % (filepath, e))
 
+    @staticmethod
+    def _parse_commands_from_lines(lines):
         for line in lines:
-            elements = shlex.split(line.strip())
-            # Skip blank lines or comment lines beginning with an octothorpe (#)
-            if elements[0].startswith('#'):
+            if not line or line.startswith('#'):
                 continue
+            elements = shlex.split(line)
             command_args = elements[1:] or ['']
             for element in command_args:
-                self.handle_command(f"command://{elements[0]} {element}")
+                yield f"command://{elements[0]} {element}"
+
+    def handle_command_from_file(self, argstring):
+        for command in self._parse_commands_from_lines(self._read_lines_from_file(argstring)):
+            self.handle_command(command)
 
     def handle_command_lookup(self, argstring):
         self.autotag(self.unclustered_files.files)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -497,7 +497,8 @@ class Tagger(QtWidgets.QApplication):
 
         for line in lines:
             elements = shlex.split(line.strip())
-            if not elements:
+            # Skip blank lines or comment lines beginning with an octothorpe (#)
+            if not elements or elements[0].startswith('#'):
                 continue
             command_args = elements[1:] or ['']
             for element in command_args:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -485,14 +485,22 @@ class Tagger(QtWidgets.QApplication):
             self.analyze(self.albums[album_name].iterfiles())
 
     def handle_command_from_file(self, argstring):
-        for command in [line.strip() for line in open(argstring).readlines()]:
-            elements = command.split(' ')
+        try:
+            lines = [line.strip() for line in open(argstring).readlines()]
+        except OSError:
+            log.error("Error opening command file: %s", argstring)
+            return
+        except Exception:
+            log.error("Error reading command file: %s", argstring)
+            return
 
-            if len(elements) > 1:
-                for element in elements[1:]:
-                    self.handle_command(f"command://{elements[0]} {element}")
-            elif elements:
-                self.handle_command(f"command://{elements[0]} ")
+        for line in lines:
+            elements = shlex.split(line.strip())
+            if not elements:
+                continue
+            command_args = elements[1:] or ['']
+            for element in command_args:
+                self.handle_command(f"command://{elements[0]} {element}")
 
     def handle_command_lookup(self, argstring):
         self.autotag(self.unclustered_files.files)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -499,7 +499,7 @@ class Tagger(QtWidgets.QApplication):
         for line in lines:
             elements = shlex.split(line.strip())
             # Skip blank lines or comment lines beginning with an octothorpe (#)
-            if not elements or elements[0].startswith('#'):
+            if elements[0].startswith('#'):
                 continue
             command_args = elements[1:] or ['']
             for element in command_args:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -229,6 +229,10 @@ REMOTE_COMMANDS = {
         "handle_command_fingerprint",
         help_text="Calculate acoustic fingerprints for all (matched) files in the album pane.",
     ),
+    "FROM_FILE": RemoteCommand(
+        "handle_command_from_file",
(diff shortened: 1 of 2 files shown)


skelly37:
Oh, 3.9 this time. Seems like the whole testing process on Mac is problematic.

picard/tagger.py
```diff
@@ -480,6 +484,16 @@ def handle_command_fingerprint(self, argstring):
...
 
+    def handle_command_from_file(self, argstring):
+        for command in [line.strip() for line in open(argstring).readlines()]:
+            elements = command.split(' ')
+
+            if len(elements) > 1:
+                for element in elements[1:]:
+                    self.handle_command(f"command://{elements[0]} {element}")
+            elif elements:
+                self.handle_command(f"command://{elements[0]} ")
```
rdswift, line 495:
```suggestion
    def handle_command_from_file(self, argstring):
        try:
            lines = [line.strip() for line in open(argstring).readlines()]
        except OSError:
            log.error("Error opening command file: %s", argstring)
            return
        except Exception:
            log.error("Error reading command file: %s", argstring)
            return

        for line in lines:
            elements = shlex.split(line.strip())
            if not elements:
                continue
            command_args = elements[1:] or ['']
            for element in command_args:
                self.handle_command(f"command://{elements[0]} {element}")
```

Note that you would also have to include the `shlex` module.

This incorporates some error checking to avoid an unhandled exception if there is a problem reading the command file, and uses the same optimizations that evolved in your previous PR.

picard/tagger.py
```diff
@@ -480,6 +485,24 @@ def handle_command_fingerprint(self, argstring):
...
+            log.error("Error opening command file: %s", argstring)
+            return
+        except Exception:
+            log.error("Error reading command file: %s", argstring)
+            return
+
+        for line in lines:
+            elements = shlex.split(line.strip())
+            if not elements:
+                continue
```
rdswift, line 501: In case you want to be able to have comment lines in the command file.  Since the first element should be a command (and none of them starts with an octothorpe), it should be safe to skip any line starting with an octothorpe (and/or other character if preferred) as a comment.

```suggestion
            # Skip blank lines or comment lines beginning with an octothorpe (#)
            if not elements or elements[0].startswith('#'):
                continue
```

skelly37: Thanks!

zas (changes requested):
I think we should split that in smaller methods to ease testing.
Also no need to call `strip()` twice on lines.
Since we have a catchall exception block I don't think we need to handle OSError separately.

The code I'm thinking about is something like:


```python
    @staticmethod
    def _read_lines_from_file(filepath):
        try:
            yield from (line.strip() for line in open(filepath).readlines())
        except Exception as e:
            log.error("Error reading command file '%s': %s" % (filepath, e))

    @staticmethod
    def _parse_commands_from_lines(lines):
        for line in lines:
            if not line or line.startswith('#'):
                continue
            elements = shlex.split(line)
            if not elements:
                continue
            command_args = elements[1:] or ['']
            for element in command_args:
                yield f"command://{elements[0]} {element}"

    def handle_command_from_file(self, argstring):
        for command in self._parse_commands_from_lines(self._read_lines_from_file(argstring)):
            self.handle_command(command)
```

This way it will be easier to write tests for `_parse_commands_from_lines()`.

rdswift:
One thing that just occurred to me is that it may be possible to create an infinite loop by including a circular reference in the command file.  The simplest way that I can think of to avoid this is by not processing any `FROM_FILE` commands found in the command file.

This could be done by adding something like the following test just before the `command_args = elements[1:] or ['']` line:

```
            if elements[0].upper() == 'FROM_FILE':
                log.warning('Nested FROM_FILE commands not allowed.')
                continue
```

The down side to this is that you will not be able to have any nested `FROM_FILE` commands.  Alternately, you could maintain a set `from_files` in `tagger`, adding the file paths as they are being processed and removing them once they have been processed, and then checking if the file is in the set before processing.  Something like:

```
        self.from_files = set()
```

in `tagger.__init__()` and then modify `handle_command_from_file()` something like:

```
    def handle_command_from_file(self, argstring):
        if argstring in self.from_files:
            log.error(f'Circular reference found in FROM_FILE processing: {argstring}')
            return
        try:
            lines = [line.strip() for line in open(argstring).readlines()]
        except OSError:
            log.error("Error opening command file: %s", argstring)
            return
        except Exception:
            log.error("Error reading command file: %s", argstring)
            return

        self.from_files.add(argstring)

        for line in lines:
            elements = shlex.split(line.strip())
            # Skip blank lines or comment lines beginning with an octothorpe (#)
            if not elements or elements[0].startswith('#'):
                continue
            if elements[0].upper() == 'FROM_FILE':
                log.warning('Nested FROM_FILE commands not allowed.')
                continue
            command_args = elements[1:] or ['']
            for element in command_args:
                self.handle_command(f"command://{elements[0]} {element}")

        self.from_files.discard(argstring)
```

This might be overkill, but it would allow nested `FROM_FILE` commands.  I don't know whether or not that is important.

rdswift:
> ```python
>             if not elements:
>                 continue
> ```

Is this needed?  Shouldn't blank lines already be handled just prior?

skelly37:
> One thing that just occurred to me is that it may be possible to create an infinite loop by including a circular reference in the command file. The simplest way that I can think of to avoid this is by not processing any `FROM_FILE` commands found in the command file.

Really bad idea, curbs modularity, I really wouldn't like to implement such solution

> Alternately, you could maintain a set `from_files` in `tagger`, adding the file paths as they are being processed and removing them once they have been processed, and then checking if the file is in the set before processing. Something like:
> 
> ```
>         self.from_files = set()
> ```
> 
> in `tagger.__init__()` and then modify `handle_command_from_file()` something like:
> 
> ```
>     def handle_command_from_file(self, argstring):
>         if argstring in self.from_files:
>             log.error(f'Circular reference found in FROM_FILE processing: {argstring}')
>             return
>         try:
>             lines = [line.strip() for line in open(argstring).readlines()]
>         except OSError:
>             log.error("Error opening command file: %s", argstring)
>             return
>         except Exception:
>             log.error("Error reading command file: %s", argstring)
>             return
> 
>         self.from_files.add(argstring)
> 
>         for line in lines:
>             elements = shlex.split(line.strip())
>             # Skip blank lines or comment lines beginning with an octothorpe (#)
>             if not elements or elements[0].startswith('#'):
>                 continue
>             if elements[0].upper() == 'FROM_FILE':
>                 log.warning('Nested FROM_FILE commands not allowed.')
>                 continue
>             command_args = elements[1:] or ['']
>             for element in command_args:
>                 self.handle_command(f"command://{elements[0]} {element}")
> 
>         self.from_files.discard(argstring)
> ```
> 
> This might be overkill, but it would allow nested `FROM_FILE` commands. I don't know whether or not that is important.

I don't think we should handle infinite loops anyhow. It's like any other scripting language, the programmer is responsible for handling infinite loops and broken recursive call. I'm against such solution but if more of you'd insist on it, I'll choose the second way.

skelly37:
Alternatively, we could track the processed files as you've suggested but instead of skipping them, `log.warning({source_file}: Sourced already used file: {file_to_load})`. I still prefer to do nothing but eventually this way it can also be okay.

skelly37:
*writing tests for the parsing method rn*

test/test_parsing_files_with_commands.py
```diff
@@ -0,0 +1,29 @@
...
+
+from picard.tagger import Tagger
+
+
+class TestParsingFilesWithCommands(PicardTestCase):
+    MOCK_FILE_CONTENTS = (
+        "FROM_FILE file0.mp3",
+        "",
+        "#commented command",
+        "LOAD file1.mp3 file2.mp3",
```
zas, line 11: Maybe also test a line with whitespaces at the start/end (for `strip()`).

zas (changes requested):

test/test_parsing_files_with_commands.py
skelly37: done :)

zas (approved):
LGTM

phw (approved):
Nice functionality :)

zas:
@skelly37 please fix conflicts so we can merge

skelly37:
@phw @zas  mergeable :)

zas (approved):

rdswift:
I just tested this and it seems to be failing for all commands in the command file.  It appears to be trying to execute `COMMAND://LOAD` rather than `LOAD` and fails with a KeyError.

PR #2144, opened 2022-09-06
`CLEAR_LOGS` command

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

A command allowing to clear Picard logs.

# Solution

TBD

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -485,9 +485,8 @@ class Tagger(QtWidgets.QApplication):
             self.analyze(self.albums[album_name].iterfiles())
 
     def handle_command_log_clear(self, argstring):
-        log.main_tail.clear()
-        log.history_tail.clear()
-        self.window.history_dialog.update_log_tail()
+        self.window.log_dialog.clear()
+        self.window.history_dialog.clear()
 
     def handle_command_lookup(self, argstring):
         self.autotag(self.unclustered_files.files)
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -221,6 +221,10 @@ class RemoteCommand:
 
 
 REMOTE_COMMANDS = {
+    "CLEAR_LOGS": RemoteCommand(
+        "handle_command_clear_logs",
+        help_text="Clear the Picard logs",
+    ),
     "CLUSTER": RemoteCommand(
         "handle_command_cluster",
         help_text="Cluster all files in the cluster pane.",
@@ -228,10 +232,6 @@ REMOTE_COMMANDS = {
     "FINGERPRINT": RemoteCommand(
         "handle_command_fingerprint",
         help_text="Calculate acoustic fingerprints for all (matched) files in the album pane.",
-    ),
-    "LOG_CLEAR": RemoteCommand(
-        "handle_command_log_clear",
-        help_text="Clear the Picard logs",
     ),
     "LOOKUP": RemoteCommand(
         "handle_command_lookup",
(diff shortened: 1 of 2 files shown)


skelly37:
It works fine except for one edge case: 

the log window is opened while receiving `LOG_CLEAR` — if you do something like this, no further `LOG_CLEAR` will work properly

picard/ui/logview.py
```diff
@@ -324,3 +324,7 @@ class HistoryView(LogViewCommon):
 
     def __init__(self, parent=None):
         super().__init__(log.history_tail, _("Activity History"), parent=parent)
+
+    def update_log_tail(self):
+        self.log_tail.clear()
+        self.display(clear=True)
```
zas, line 330: This only concerns the History View.

See https://github.com/skelly37/picard/pull/10

picard/tagger.py
```diff
@@ -229,6 +229,10 @@ def __init__(self, method_name, help_text=None, help_args=None):
         "handle_command_fingerprint",
         help_text="Calculate acoustic fingerprints for all (matched) files in the album pane.",
     ),
+    "LOG_CLEAR": RemoteCommand(
```
zas, line 232: Others remote commands start with a verb, so I think `CLEAR_LOGS` would be more consistent (and plural because it clears all logs).

Now I'm not sure we need to clear all logs in one command, we could also provide a generic command `CLEAR_LOG <logname>`, with `logname` being either  `main` or `history`.

@rdswift what's your take on this?

zas (changes requested):

picard/tagger.py
rdswift: That is likely better, but we'll need to provide some explanation somewhere about the difference between the two (because even I'm not sure :wink: ).

zas: Ok, let's keep it simple for now, we renamed the command `CLEAR_LOGS` and if we need in the future we can still introduce `CLEAR_LOG <logname>`.

@skelly37 can you make this change?

rdswift:
Will we also be introducing a `WRITE_LOG` command to dump the log to a file, or is it already in the works and I missed it?

picard/tagger.py
rdswift: > Others remote commands start with a verb, so I think `CLEAR_LOGS` would be more consistent (and plural because it clears all logs).

In that case, should we also be renaming the `FROM_FILE` command to something like `READ_COMMANDS_FROM_FILE` (or something a bit shorter, but still starting with a verb)?

skelly37:
> Will we also be introducing a `WRITE_LOG` command to dump the log to a file, or is it already in the works and I missed it?

It's on a TODO list, see my PR on docs

picard/tagger.py
skelly37: > Ok, let's keep it simple for now, we rename the command `CLEAR_LOGS` and if we need in the future we can still introduce `CLEAR_LOG <logname>`.
> 
> @skelly37 can you make this change?

Just rename into `CLEAR_LOGS`?

skelly37: > > Others remote commands start with a verb, so I think `CLEAR_LOGS` would be more consistent (and plural because it clears all logs).
> 
> In that case, should we also be renaming the `FROM_FILE` command to something like `READ_COMMANDS_FROM_FILE` (or something a bit shorter, but still starting with a verb)?
> 
> **EDIT:** Perhaps `IMPORT_COMMAND_FILE` or maybe `IMPORT_COMMANDS`?

I don't think so, `FROM_FILE` sounds pretty fine and intuitive.

zas: > > > Others remote commands start with a verb, so I think `CLEAR_LOGS` would be more consistent (and plural because it clears all logs).
> > 
> > 
> > In that case, should we also be renaming the `FROM_FILE` command to something like `READ_COMMANDS_FROM_FILE` (or something a bit shorter, but still starting with a verb)?
> > **EDIT:** Perhaps `IMPORT_COMMAND_FILE` or maybe `IMPORT_COMMANDS`?
> 
> I don't think so, `FROM_FILE` sounds pretty fine and intuitive.

I tend to agree, `FROM_FILE` works for me.

skelly37:
> Will we also be introducing a `WRITE_LOG` command to dump the log to a file, or is it already in the works and I missed it?

@rdswift one more question:

Should it be rather pop or read? Both approaches make sense to me, so I'd like to see how do you see it as a requester.

rdswift:
> Should it be rather pop or read? Both approaches make sense to me, so I'd like to see how do you see it as a requester.

I think `read` makes more sense to me.

zas (approved):

PR #2146, opened 2022-09-07
Added configurable way to dump Picard logs into a file

skelly37:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

A commands to write logs into a file

# Solution

Added `WRITE_LOGS`

# Action

@zas @phw please take a look if you like the approach :)

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -441,8 +441,6 @@ class Tagger(QtWidgets.QApplication):
         self.exit_cleanup = []
         self.stopping = False
 
-        self.saved_logs_verbosity = logging.DEBUG
-
         # On macOS temporary files get deleted after 3 days not being accessed.
         # Touch these files regularly to keep them alive if Picard
         # is left running for a long time.
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -291,6 +291,11 @@ REMOTE_COMMANDS = {
         "handle_command_scan",
         help_text="Scan all files in the cluster pane.",
     ),
+    "SET_SAVED_LOGS_VERBOSITY": RemoteCommand(
+        "handle_command_set_saved_logs_verbosity",
+        help_text="Set verbosity of logs that will be saved with WRITE_LOGS",
+        help_args="[verbosity level: DEBUG/D, INFO/I, WARNING/W, ERROR/E]",
+    ),
     "SHOW": RemoteCommand(
         "handle_command_show",
         help_text="Make the running instance the currently active window.",
@@ -298,6 +303,11 @@ REMOTE_COMMANDS = {
     "SUBMIT_FINGERPRINTS": RemoteCommand(
         "handle_command_submit_fingerprints",
         help_text="Submit outstanding acoustic fingerprints for all (matched) files in the album pane.",
+    ),
+    "WRITE_LOGS": RemoteCommand(
+        "handle_command_write_logs",
+        help_text="Write Picard logs to a given path.",
+        help_args="[absolute path to 1 file]",
     ),
 }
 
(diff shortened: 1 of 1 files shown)


skelly37:
1 minor issue, `LOG_CLEAR` does not seem to affect the command and basically the whole log history gets dumped :(

picard/tagger.py
```diff
@@ -605,12 +617,28 @@ def handle_command_save_modified(self, argstring):
...
 
+    def handle_command_set_saved_logs_verbosity(self, argstring):
+        if argstring.upper() in {"DEBUG", "D"}:
+            self.saved_logs_verbosity = logging.DEBUG
+        elif argstring.upper() in {"INFO", "I"}:
+            self.saved_logs_verbosity = logging.INFO
+        elif argstring.upper() in {"WARNING", "W"}:
+            self.saved_logs_verbosity = logging.WARNING
+        elif argstring.upper() in {"ERROR", "E"}:
+            self.saved_logs_verbosity = logging.ERROR
```
zas, line 628: I think we should left this verbosity thing out for now, and just dump the whole log.

picard/tagger.py
```diff
@@ -605,12 +617,28 @@ def handle_command_save_modified(self, argstring):
...
+
     def handle_command_show(self, argstring):
         self.bring_tagger_front()
 
     def handle_command_submit_fingerprints(self, argstring):
         self.acoustidmanager.submit()
 
+    def handle_command_write_logs(self, argstring):
+        with open(argstring, 'w') as f:
```
zas, line 637: We definitively need to catch exceptions here, for example, if the target file cannot be written, a PermissionError might be raised, leading to a crash of the application.
We should catch all exceptions, and log them to avoid any issue.

zas (changes requested):

picard/tagger.py
skelly37: Disagreed, the command does not have a complex logic, yet it helps a lot with writing a clean log files.

picard/tagger.py
skelly37: Just `except Exception as e` and then log it?

zas: Yes

picard/tagger.py
zas: The thing is that what is logged depends on actual log level, so if Picard is set to INFO level, changing this to DEBUG will not change anything.
I don't say this isn't useful, but we need to think about how to implement it a bit more. What I suggest is to let this out this PR, and do that in an other.

picard/tagger.py
```diff
@@ -611,6 +618,14 @@ def handle_command_show(self, argstring):
     def handle_command_submit_fingerprints(self, argstring):
         self.acoustidmanager.submit()
 
+    def handle_command_write_logs(self, argstring):
+        try:
+            with open(argstring, 'w') as f:
+                for x in self.window.log_dialog.log_tail.contents():
+                    f.write(f"{x.message}\n")
+        except Exception as e:
+            log.error("Error writing logs to a file: %r", e)
```
zas, line 627:
```suggestion
            log.error("Error writing logs to a file: %s", e)
```

For exceptions, since our goal is to display a human-readable message, better use `str(e)` rather than `repr(e)`.

picard/tagger.py
```diff
@@ -436,6 +441,8 @@ def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
         self.exit_cleanup = []
         self.stopping = False
 
+        self.saved_logs_verbosity = logging.DEBUG
```
zas, line 444: Unused, it can be removed (along the blank line below)

zas (changes requested):

zas (approved):
LGTM

PR #2147, opened 2022-09-11
PICARD-2539: Always call _run_album_metadata_processors before lowering album._requests

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2539

Fixes extra calls to `album._finalize_loading()` happening when cover art saving is enabled but all providers are disabled and huge releases get loaded with recording ARs enabled.

This fixes the root cause for the crash already fixed in #2140

# Solution
Ensure that `self._requests -= 1` always happens after `self._run_album_metadata_processors()`. This is also what happens for the normal (non-huge releases) loading. See https://tickets.metabrainz.org/browse/PICARD-2539?focusedCommentId=63951&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-63951 for a more in-depth explanation.

# Action

The self._requests handling is still an awkward API, and this should be refactored. But we have PICARD-1403 for this.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -340,9 +340,9 @@ class Album(DataObject, Item):
         )
 
     def _recordings_request_finished(self, document, http, error):
-        self._requests -= 1
         if error:
             self.error_append(http.errorString())
+            self._requests -= 1
             self._finalize_loading(error)
         else:
             for recording in document.get('recordings', []):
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2149, opened 2022-09-23
PICARD-2549: Fixed similarity not getting updated if only difference is cover art

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2549

If a file is moved to a track where the metadata matches perfectly but only cover art is different, then the similarity score of this file was not recalculated. That means the file retains the similarity of the previous (probably bad) match.

# Solution

Reset the similarity to 1.0 in all cases when metadata match was successful. This matches the existing behavior when such files would be matched directly to the perfect match: The file shows as changed, but with metadata similariry set to 100%.

We could consider showing a lower score in all cases where only cover art does not match. But given that we have many cases where the cover art known for the file does not match the loaded cover art (see e.g. PICARD-1001) I think this would not be helpful.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -709,12 +709,12 @@ class File(QtCore.QObject, Item):
                         self.state = File.CHANGED
                     break
             else:
-                if (self.metadata.images
+                self.similarity = 1.0
+                if self.state in (File.CHANGED, File.NORMAL):
+                    if (self.metadata.images
                         and self.orig_metadata.images != self.metadata.images):
-                    self.state = File.CHANGED
-                else:
-                    self.similarity = 1.0
-                    if self.state == File.CHANGED:
+                        self.state = File.CHANGED
+                    else:
                         self.state = File.NORMAL
         if signal:
             log.debug("Updating file %r", self)
--- a/test/test_file.py
+++ b/test/test_file.py
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 #
-# Copyright (C) 2018-2021 Philipp Wolfer
+# Copyright (C) 2018-2022 Philipp Wolfer
 # Copyright (C) 2019-2022 Laurent Monin
 # Copyright (C) 2021 Bob Swift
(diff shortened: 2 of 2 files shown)


picard/file.py
```diff
@@ -709,12 +709,12 @@ def update(self, signal=True):
                         self.state = File.CHANGED
                     break
             else:
-                if (self.metadata.images
+                self.similarity = 1.0
+                if self.state in (File.CHANGED, File.NORMAL):
```
zas, line 713: Just a minor thing, but using a set instead of a tuple  seems more correct to me:

```suggestion
                if self.state in {File.CHANGED, File.NORMAL}:
```

picard/file.py
```diff
@@ -709,12 +709,12 @@ def update(self, signal=True):
                         self.state = File.CHANGED
                     break
             else:
-                if (self.metadata.images
+                self.similarity = 1.0
```
zas, line 712: I wonder if we should have a constant (`MAXSIMILARITY`?) instead of `1.0`

zas (changes requested):

picard/file.py
phw: Definitely, yes, I changed it

picard/file.py
phw: Not really sure about that specific case. We have multiple places where we deal with scoring between 0.0 and 1.0 (basically percentages). They are similar because of the range, would need to be checked if they are actually always the same and would reuse the same constant.

But actually because this is such a clear concept I think a constant doesn't help readability here. Quite the opposite, reading ` 1.0`  makes it more visible what's happening than a constant. It's the same as if we had e.g. percentage values between 0 and 100. I'd find reading a `100`  in code much clearer than a constant `ONE_HUNDRED_PERCENT`.

I'd see it differently if we had some other strange max, such as `0.97` or so. Then I'd definitely argue for a constant.

zas (approved):
LGTM

PR #2152, opened 2022-09-26
PICARD-2556: fix player popover dialogs on Wayland

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2556

Under Wayland the player toolbar popover dialogs, used for audio volume and playback speed settings, are

1. causing a crash due to wrong data type of coordinates
2. not properly positioned

# Solution

1. Ensure position and slider value are set as integer
2. Add custom handling for calculating the popover position on Wayland

Note that there is a small difference in behavior: On systems other then Wayland the code makes sure the popover is shown centered above / below the parent widget if possible, but repositions it to ensure it is on screen. This is relevant if the parent is near the screen edges.

On Wayland that's no possible, because the position relative to the screen cannot be determined. Hence instead of the screen the logic now repositions the popovers so they are inside the main window. That ensures the popovers are fully visible in e.g. maximized state.

### DIFF
--- a/picard/ui/widgets/__init__.py
+++ b/picard/ui/widgets/__init__.py
@@ -116,7 +116,7 @@ class Popover(QtWidgets.QFrame):
             y = -self.height()
         else:  # bottom
             y = parent.height()
-        pos = parent.mapToGlobal(QtCore.QPoint(x, y))
+        pos = parent.mapToGlobal(QtCore.QPoint(int(x), int(y)))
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2153, opened 2022-09-26
PICARD-2557: Revert "An editable QComboBox automatically sets up a QCompleter"

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2557

The edit tag dialog's auto completion for the tag name will show the currently typed value as the first suggestion in autocomplete, which is not helpful for the user and always requires an extra button press to skip it.

This was caused by removing the custom QCompleter implementation and using the default QComboBox  completer implementation.

# Solution
This reverts commit e0d58d50987a3f9c7cf52e23d728f1f2cf0a153f and restores the custom implementation.

# Action


zas (approved):
LGTM

PR #2154, opened 2022-09-26
PICARD-2331: Port to PyQt6

phw:
# Summary
Porting Picard to PyQt6.

# Problem
* JIRA ticket (_optional_): PICARD-2331

# Solution

- [x] Upgrade all references to PyQt6 and apply general API changes
- [x] Replacement for `QAction.menu` (for CD drive selection button) (done in #2169)
- [x] Remove custom network timeout implementation (now supported by Qt6 itself)
- [x] Bump API version to 3.0
- [x] Remove macOS 10.12 / 10.13 builds (no longer supported by Qt6)
- [x] Update resource loading for PyQt6 (requires PyQt >= 6.3.1)
- [x] Handle no longer available `QtWinExtras.QWinTaskbarButton` (see https://www.qt.io/blog/qt-extras-modules-in-qt-6, [QTBUG-94008](https://bugreports.qt.io/browse/QTBUG-89564), [QTBUG-89564](https://bugreports.qt.io/browse/QTBUG-94008)) -> disabled for now
- [x] Migrate media player (only works stable with Qt >= 6.5)
- [x] Investigate failings tests on Linux CI
- [x] Test on Linux
- [x] Fix macOS code signing
- [x] Test on macOS
- [x] Test on Windows
- [x] Fix DBus implementation for task bar progress on Linux
- [x] Fix crash of edit tag dialog
- [x] Options > Plugins sometimes throws exception if running with config upgrade (needs investigation)
- [ ] ~Investigate segmentation faults during drag and drop~ *issue happens only with pre-compiled Qt6 from pypi*
- [ ] ~Dark theme glitches on Linux / GNOME <- Those are likely a Qt or GNOME issue and out of scope for this PR~ *issue happens only with pre-compiled Qt6 from pypi*


phw:
This is mostly complete now. Most things should work, but definitely needs more testing. There might be some incompatibility hidden with some code paths.

Media player has been ported, but current state does not behave well. Playback often stops when doing changes to volume, playback rate or position, without any obvious reason. If QtMultimedia behaves so unstable it might indicate that this is not yet ready. But I need to investigate more and test on other systems. We wanted to move this over to a plugin anyway, but in a first step I wanted to transfer the functionality.

Sophist-UK:
@phw Philipp 

Hmm ... I cannot believe just how much work is being needed to go from Qt5 to Qt6!!!!

phw:
@Sophist-UK For the most parts is the renaming if Qt5 to Qt6. And then there were the anticipated changes, like for the media player and networking. But yes, also quite some smaller changes I did not expect. Part of this is that also PyQt took the chance to cleanup their code.

The changes in #1922, #2035, and #2169 also belong to the list of things needed for the switch. Just that those already could be done while staying with PyQt5.

regorxxx:
Note QT6 breaks compatibility for anything below Win 10.
https://forum.qt.io/topic/140132/does-qt6-built-applications-run-on-windows-7

phw:
Yes it does, that's already known. It's also the reason why Python gets upgraded here. Picard 2.9 stays with Python 3.8 for compatibility with Windows 7.

phw:
@zas @rdswift I think we are now good to move forward with this. The port seems complete and working. If necessary we can tweak later, but I don't want to keep rebasing forever. For Picard 2 we can do changes to the 2.10.x branch for bugfixes and eventual backports.

@rdswift I'm concerned about your setup and contributions. As I understand you are running Windows 7, and Qt6 officially dropped support for this OS. The official binary builds, which are also being installed by PyQt6 by default, are, AFAIK, not compatible with Windows 7 anymore.

How is your situation here? Do you have the ability to still run Picard?

Unfortunately we cannot avoid this forever. We also want to use newer Python versions. The last Python version compatible with Windows 7 is Python 3.8 (which we currently use for the Windows build). But support for this runs out in a year as well.

rdswift:
Don't worry about me.  I actually do most of my development work and music tagging and management on a Linux system.  I also have a Windows 10 machine for development and testing of one commercial package I maintain, so I can use that for Picard testing as well.  I'm trying to retire the Windows 7 box, but haven't yet migrated all my accounting stuff (and some other specialty software) over to the Windows 10 machine yet.

zas:
I'm all to merge this asap, now we released 2.10, at worse we'll have some few extra dot releases to do, but our goal is Picard 3.

zas (approved):

zas:
I get weird colors in dark mode:

image

phw:
@zas Yes, happens for me as well when using the pre-compiled Qt binaries from pypi. But works without issue when using the distribution's packages. Please try this.

phw:
@zas @rdswift Ok, let's finally move forward. I'm going to merge this so we can continue on this basis. The binary builds for macOS still need some work and currently don't run, but I'll investigate deeper after merging. In general macOS works when running from source.

PR #2156, opened 2022-09-29
PICARD-2554: Only do AcoustID submission for matched files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2554

Submission was also possible for files having existing recording MBID in the tags, even if they were on the left side as unmatched files. The intention was to only allow submission for files matched to tracks, but this broke with https://github.com/metabrainz/picard/commit/a1c09fc1b68c5202624ca5966170805ebf313b94 because clusters then also got a `orig_metadata`  property.

# Solution

Ensure submission of fingerprint / recording ID pairs is only done for files matched to tracks. Kept the implementation flexible by adding a `Item.can_link_fingerprint`  property so we can also have other entities then tracks be able to have files attached with AcoustID submission enabled. Also having an explicit property avoids side effects as in the previous implementation.


zas (approved):
LGTM

PR #2157, opened 2022-09-29
PICARD-2564: Keep option UI for ascii_filenames and windows_compatibility activated

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2564

The options screen for file renaming currently attempts to set the two options "ascii_filenames" and "windows_compatibility" to inactive, when neither rename nor move files is active. There are two problems with this:

The code does not set the initial state correctly, so the options are changeable by the user initially, but when both rename and move files get deactivated then the other options suddenly get grayed out
The options are also used for cover art file naming, independent of the renaming or moving of music files.

I hence propose to completely remove this graying out of these two options and instead allow the user to always change them.

# Solution

Allow the user to always change these settings, independent of file renaming / moving, as the same settings are also used for cover art files and in theory all file saving done by Picard.

Also run `toggle_file_naming_format`  initially to ensure the UI state is properly initialized.

### DIFF
--- a/picard/ui/options/renaming.py
+++ b/picard/ui/options/renaming.py
@@ -105,6 +105,7 @@ class RenamingOptionsPage(OptionsPage):
 
         self.ui.move_files.toggled.connect(self.toggle_file_naming_format)
         self.ui.rename_files.toggled.connect(self.toggle_file_naming_format)
+        self.toggle_file_naming_format(None)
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2158, opened 2022-09-30
PICARD-2567: Generate disc ID from fre:ac ripping log

jbosboom:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Slightly relax the EAC ripping log parsing code so it can parse logs from fre:ac.

# Problem

fre:ac's ripping logs contain a TOC table with only minor differences from EAC's format (times always use two digits for each field, and small spacing differences).  However fre:ac prefixes most log lines with a timestamp, so the existing EAC log parsing code cannot parse fre:ac logs.

* JIRA ticket (_optional_): PICARD-2567

# Solution

Removing the `^` anchor from the TOC table line regular expression and using `search` instead of `match` is sufficient to allow the EAC log parser to parse the TOC table from fre:ac logs.  The line regex already handles the differences in spacing and leading zeroes in start times and lengths.  The line regex does not match data track lines because fre:ac writes "DATA" in place of the track number, but data tracks are not used for disc ID calculation anyway. 

Other fre:ac log information:
- fre:ac logs use the `.log` extension.
- Some log lines have trailing whitespace.
- fre:ac logs are written in UTF-8 without BOM even on Windows (tested on 10, not sure if different on 7, 8 or 11).
- fre:ac supports UI translations and will use the selected language for CDDB lookups, but log text not derived from tags, in particular the TOC table header, is not translated.
- The second table, under the "Disc info:" heading, is only present if a CDDB lookup was performed and successful.  Otherwise "CDDB info:    No CDDB data available for this disc!" is printed following the "CDDB disc ID:" line, and "unknown album"/"unknown title"/etc. are used when building filenames by substitution.

# Action

I did not modify the "EAC / XLD / Whipper log files" string because I do not understand how translations work and do not want to break anything.

The test logs were generated from different discs than the existing logs.  If you wanted to avoid adding another TOC tuple literal in the tests, you could install fre:ac and rip your discs again, or forge a fre:ac log file as if you had ripped your discs again.

### DIFF
--- a/picard/disc/eaclog.py
+++ b/picard/disc/eaclog.py
@@ -4,6 +4,7 @@
 #
 # Copyright(c) 2018 Konstantin Mochalov
 # Copyright(c) 2022 Philipp Wolfer
+# Copyright(c) 2022 Jeffrey Bosboom
 #
 # Original code from https://gist.github.com/kolen/765526
 #
@@ -41,7 +42,7 @@ RE_TOC_TABLE_HEADER = re.compile(r""" \s*
     """, re.VERBOSE)
 
 RE_TOC_TABLE_LINE = re.compile(r"""
-    ^\s*
+    \s*
     (?P<num>\d+)
     \s*\|\s*
     (?P<start_time>[0-9:.]+)
@@ -69,14 +70,14 @@ def filter_toc_entries(lines):
             break
 
     for line in lines:
-        m = RE_TOC_TABLE_LINE.match(line)
+        m = RE_TOC_TABLE_LINE.search(line)
         if not m:
             break
         yield TocEntry(int(m['num']), int(m['start_sector']), int(m['end_sector']))
 
 
 def toc_from_file(path):
-    """Reads EAC / XLD log files, generates MusicBrainz disc TOC listing for use as discid.
+    """Reads EAC / XLD / fre:ac log files, generates MusicBrainz disc TOC listing for use as discid.
 
     Warning: may work wrong for discs having data tracks. May generate wrong
     results on other non-standard cases."""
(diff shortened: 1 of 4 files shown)


phw:
Thanks a lot for your contribution, this is great.

We need a ticket for the changelog, so I added PICARD-2567.


> I did not modify the "EAC / XLD / Whipper log files" string because I do not understand how translations work and do not want to break anything.

You can just update the string, we'll take care of updating the translation files once it has been merged. I prefer this approach over having the full translation files updated in the pull request, as parallel translation changes quickly cause merge conflicts. Also it makes the diffs much less readable with changes that are essentially just automatic updates anyway :)

> The test logs were generated from different discs than the existing logs. If you wanted to avoid adding another TOC tuple literal in the tests, you could install fre:ac and rip your discs again, or forge a fre:ac log file as if you had ripped your discs again.

I think it's fine as it is. Does no harm in having some variations of data sources in the tests.

phw (approved):

phw:
I merged it and updated the UI text in https://github.com/metabrainz/picard/commit/a84fae2d7f90b8f6bb8e4abddc0337431810c1c3 . Thanks again for your contribution.

jbosboom:
https://github.com/metabrainz/picard/commit/a84fae2d7f90b8f6bb8e4abddc0337431810c1c3 's commit message references PICARD-2564.  I think you meant [PICARD-2567](https://tickets.metabrainz.org/browse/PICARD-2567).  (Perhaps too late to change now that the commit is published.)

PR #2163, opened 2022-10-27
PICARD-2582: Get rid of legacy distutils imports in setup.py

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2582

Picard's `setup.py` was still importing some legacy distutils components to achieve some functionality.

Also these distutils imports are imported before setuptools gets imported, which is actively discouraged as setuptools might provide modified distutils.

# Solution

Replace all distutils imports with proper replacements (see also [PEP 632](https://peps.python.org/pep-0632/))

# Action

This needs a newer setuptools which provides the `build` command implementation (was added sometime in August this year).

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2006-2008, 2011-2014, 2017 Lukáš Lalinský
 # Copyright (C) 2007 Santiago M. Mola
 # Copyright (C) 2008 Robert Kaye
-# Copyright (C) 2008-2009, 2018-2021 Philipp Wolfer
+# Copyright (C) 2008-2009, 2018-2022 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2011-2012, 2014, 2016-2018 Wieland Hoffmann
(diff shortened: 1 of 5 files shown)


phw:
The `command.build` was added to setupttools 62.4.0, which was released in June 13th '22, so that's the minimum supported version with this patch.

setup.py
```diff
@@ -39,14 +39,12 @@
 
 
 import datetime
-from distutils import log
-from distutils.command.build import build
-from distutils.dep_util import newer
-from distutils.spawn import find_executable
 import glob
 from io import StringIO
+import logging as log
```
zas, line 44: Is this import needed?

phw: Yes, log is used at various places in setup.py in the different commands. `distutils` provided its own `log` module, the recommended replacement is using the `logging` module instead, see https://peps.python.org/pep-0632/#migration-advice .

zas (approved):
LGTM

PR #2164, opened 2022-10-27
Upgrade to isort 5

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

# Problem

We had disabled using newer isort for now. Upgrade the configuration for using the new version and review the changes it applies.

isort 5 now can also handle imports not at top, hence some new locations now get sorted automatically.

Note: I also tried moving the configuration from .isort.cfg to pyproject.toml as described in https://pycqa.github.io/isort/docs/configuration/config_files.html, but my configuration there always got ignored.


phw:
Update: Moved the config to `pyproject.toml` in 2ce4bca05b3d540d7b60b73957122c1414934af6.

Turns out in my previous attempt I had invalid syntax there. Also because of stronger typing in TOML the config needed some minor tweaks.

picard/formats/mutagenext/tak.py
```diff
@@ -34,11 +34,11 @@
 
 try:
     from mutagen.tak import (
-        Open,
         TAK,
+        Open,
```
zas, line 38: This looks incorrect to me, why is it placed after `TAK`?

picard/util/imagelist.py
```diff
@@ -165,6 +165,7 @@ def _update_state(obj, state):
 # TODO: use functools.singledispatch when py3 is supported
 def _get_state(obj):
     from picard.album import Album
+
```
zas, line 168: Why does it add this blank line?

zas (changes requested):

picard/formats/mutagenext/tak.py
phw: Odd indeed, I'll look into it. But I assume it is because of `TAK` being all uppercase, and isort considers these constants and sorts them to the top, just like in https://github.com/metabrainz/picard/blob/master/picard/config.py#L43-L48

This is one of the places isort did not touch before, and hence this was manually sorted. I'll see if this can be prevented, but it might be easier to just live with this case.

picard/util/imagelist.py
phw: Because it separates picard and picard.ui imports according to the rules (separate import groups). The change now is that isort also handles imports placed elsewhere in the file, not only at the top. Which is actually great for us, because we have quite a few conditional imports (e.g. list of imports only in a `if IS_WIN`  block).

picard/formats/mutagenext/tak.py
phw: Yes, it is because of https://pycqa.github.io/isort/docs/configuration/options.html#order-by-type . `TAK` is all uppercase, so goes into the first group., followed by Open and TAKHeaderError which both qualify for the second camel-case group.

We can either add an `# isort:skip` here or just accept this case (I prefer the latter).

phw:
The ignores don't work properly on Windows. This looks like a bug to me (probably because of different character separators. I'll look into it and report it upstream, but as a quick workaround we can probably use `skip_glob` for ignoring `resources.py` as well.

zas (approved):
LGTM

PR #2167, opened 2022-11-01
Fix executable `LOOKUP` command text.  Add `LOOKUP_CLUSTERS` command.

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

The recently added `LOOKUP` executable command indicated in the help text that it would "Lookup all clusters in the cluster pane", but in fact it actually would "Lookup all unclustered files in the cluster pane".  There was no executable command that would "Lookup all clusters in the cluster pane."

# Solution

Change the help text on the existing `LOOKUP` command, and add a new `LOOKUP_CLUSTERS` command.

# Action

If accepted, the documentation will need to be updated to reflect the changes.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -246,6 +246,10 @@ REMOTE_COMMANDS = {
     ),
     "LOOKUP": RemoteCommand(
         "handle_command_lookup",
+        help_text="Lookup all unclustered files in the cluster pane.",
+    ),
+    "LOOKUP_CLUSTERS": RemoteCommand(
+        "handle_command_lookup_clusters",
         help_text="Lookup all clusters in the cluster pane.",
     ),
     "LOOKUP_CD": RemoteCommand(
@@ -546,6 +550,9 @@ class Tagger(QtWidgets.QApplication):
 
     def handle_command_lookup(self, argstring):
         self.autotag(self.unclustered_files.files)
+
+    def handle_command_lookup_clusters(self, argstring):
+        self.autotag(self.clusters)
 
     def handle_command_lookup_cd(self, argstring):
         disc = Disc()
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -246,8 +246,8 @@ REMOTE_COMMANDS = {
     ),
     "LOOKUP": RemoteCommand(
         "handle_command_lookup",
-        help_text="Lookup files in the clustering pane.",
-        help_args="[clustered|unclustered]"
+        help_text="Lookup files in the clustering pane. Defaults to all files.",
+        help_args="[clustered|unclustered|all]"
     ),
     "LOOKUP_CD": RemoteCommand(
         "handle_command_lookup_cd",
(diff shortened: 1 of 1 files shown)


picard/tagger.py
```diff
@@ -246,6 +246,10 @@ def __init__(self, method_name, help_text=None, help_args=None):
     ),
     "LOOKUP": RemoteCommand(
         "handle_command_lookup",
+        help_text="Lookup all unclustered files in the cluster pane.",
+    ),
+    "LOOKUP_CLUSTERS": RemoteCommand(
+        "handle_command_lookup_clusters",
```
zas, line 252: Does it lookup clusters or files in clusters? I think the `LOOKUP` command is unclear about what it does, and adding `LOOKUP_CLUSTERS` is even more confusing.

We could use a command with arguments instead: `LOOKUP <what>` with a default.

What do you think?

rdswift: We could do a single command with an argument.  I actually like that idea.  Perhaps the argument options could be:

- `clustered` - Lookup files that have been clustered.
- `unclustered` - Lookup files that have not been clustered.

Does that help clarify?

phw: I like this idea. Not sure about the default either, though. What about allowing `all` and make that the default?

rdswift: That sounds okay to me.  I'll make the change.

picard/tagger.py
```diff
@@ -546,13 +546,17 @@ def handle_command_load(self, argstring):
...
+            argstring = argstring.upper()
         else:
+            argstring = 'ALL'
+        if argstring not in {'ALL', 'CLUSTERED', 'UNCLUSTERED'}:
             log.error("Invalid LOOKUP command argument: '%s'", argstring)
+            return
+        if argstring in {'ALL', 'CLUSTERED'}:
+            self.autotag(self.clusters)
+        if argstring in {'ALL', 'UNCLUSTERED'}:
+            self.autotag(self.unclustered_files.files)
```
zas, line 559: What about rewriting it as:

```python
    def handle_command_lookup(self, argstring):
        if argstring:
            argstring = argstring.upper()
        if not argstring or argstring == 'ALL':
            self.autotag(self.clusters)
            self.autotag(self.unclustered_files.files)
        elif argstring == 'CLUSTERED':
            self.autotag(self.clusters)
        elif argstring == 'UNCLUSTERED':
            self.autotag(self.unclustered_files.files)
        else:
            log.error("Invalid LOOKUP command argument: '%s'", argstring)
```

zas (changes requested):
It is much better with optional argument IMHO

picard/tagger.py
rdswift: That looks a bit cleaner.  I've pushed the changes.  Thanks.

zas (approved):
LGTM

phw (approved):

PR #2168, opened 2022-11-04
PICARD-2565: Apply make_short_filename and make_save_path to saved image files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2565

Saving cover art files uses script_to_filename and applies the special character replacement for Windows compatibility and ASCII conversion. But it ignores the changes for max. file path length ("make_short_filename") and the general "make_save_path" that does make some generic sanitizing of the final path.

# Solution

Call both `make_short_filename`  and `make_save_path` when saving images.


zas (approved):
LGTM

PR #2169, opened 2022-11-06
Refactor CD lookup submenu

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

With Qt6 `QAction.setMenu`  will get removed. Instead directly place the CD lookup `QMenu` in the menu and on the toolbar button.

This is a part of the PyQt6 port in PICARD-2331 (see #2154). But I think this is one partial aspect we can merge separately in preparation and just consider `QAction.setMenu` deprecated.


zas (approved):
LGTM

PR #2170, opened 2022-11-07
PICARD-2585: Treat unknown parsed schemes as part of the file path

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Treat unknown parsed schemes as part of the file path.

# Problem

* JIRA ticket (_optional_): PICARD-2585

Since the command line parsing was updated to include the ability to pass executable commands, Picard can no longer load a file or directory with a drive specified in Windows (e.g. "C:\Testing\Ripped Files").

# Solution

Treat unknown parsed schemes as part of the file path.

# Action

None.


picard/tagger.py
```diff
@@ -181,6 +181,8 @@ def plugin_dirs():
 
 class ParseItemsToLoad:
 
+    WINDOWS_DRIVE_TEST = re.compile(r"^[a-z]\:", re.IGNORECASE)
```
zas, line 184: I'm not a Windows expert, but I don't think drive names can be a lowercase letter, nor more than one letter.

If this statement is correct, we can go with:

```suggestion
    WINDOWS_DRIVE_TEST = re.compile(r"^[A-Z]\:")
```

zas: Well, ignore, I forgot Win paths are case-insensitive, so `c\:` might be possible.

rdswift: Yeah, I can use lower case drive letters, such as "c:\Ripped Files\Mountain Goats" to access the directcory, and Picard loads the information properly.

rdswift: Any preference on whether to use `re.compile(r"^[a-zA-Z]\:")` or `re.compile(r"^[a-z]\:", re.IGNORECASE)`?

zas (approved):
LGTM

phw (approved):

PR #2171, opened 2022-11-12
PICARD-2588: Revert "Replace QThread with Python treads using concurrent.futures"

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2588

This declares 96f492c357457f60504ebf75b0ab1a5b8747edd7 to be a stupid idea and reverts back to using QThread for all threads. Mixing threading between Qt and Python's is going to cause all kinds of incompatibility issue, see PICARD-2588 for a bit. As we are running a Qt app we must rely on Qt's event loop.

# Solution

This reverts commit 96f492c357457f60504ebf75b0ab1a5b8747edd7 and fixes PICARD-2588.

Also reintroduces the priority parameter, using the values initially set for the various operations. I just wen with using the Qt enum instead of its integer values (something that I think will be mandatory for Qt6 anyway).

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -214,7 +214,8 @@ class File(QtCore.QObject, Item):
     def load(self, callback):
         thread.run_task(
             partial(self._load_check, self.filename),
-            partial(self._loading_finished, callback))
+            partial(self._loading_finished, callback),
+            priority=QtCore.QThread.Priority.LowestPriority)
 
     def _load_check(self, filename):
         # Check that file has not been removed since thread was queued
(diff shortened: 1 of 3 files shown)


zas (approved):
It looks good to me, reasons behind this revert makes a lot of sense.

phw:
Just for future reference: https://stackoverflow.com/questions/1595649/threading-in-a-pyqt-application-use-qt-threads-or-python-threads

PR #2172, opened 2022-11-12
Limit worker threads based on CPUs

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

The default thread pool by default has a max. thread count based on [`QThread.idealThreadCount()`](https://doc.qt.io/qt-5/qthread.html#idealThreadCount), which basically sets the thread count based on the system's available CPU cores.

Especially during file loading this number of threads gets exhausted. But this means there are 8 loading threads running, but there is always at least the main UI thread. Also there are the save thread pool (which might be used when the user is saving files in parallel) and the priority thread pool (which currently is used in metadatabox updates only and is meant for everything that should not be blocked by normal background tasks).

# Solution

Limit number of threads in the main pol to `QThread.idealThreadCount()` minus 3 (main thread, priority pool + save pool). Keep at least 2, as one thread gets used up by the new pipe config.

In my tests with 8 cores this gave 5. It improved responsiveness during loading of many files significantly. But we might consider going a bit higher. We have one thread reserved for the pipe handling, which usually isn't doing much. Then save thread pool is usually idle. Priority gets used if the user clicks some item with metadata.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -346,29 +346,14 @@ class Tagger(QtWidgets.QApplication):
         if picard_args.debug or "PICARD_DEBUG" in os.environ:
             self.set_log_level(logging.DEBUG)
 
-        # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
-        # It's a valid reference, but its start() method doesn't work.
-        self.thread_pool = QtCore.QThreadPool(self)
+        self._setup_thread_pools()
 
         self.pipe_handler = pipe_handler
-
         if self.pipe_handler:
             self.pipe_handler.pipe_running = True
             thread.run_task(self.pipe_server, self._pipe_server_finished)
 
         self._init_remote_commands()
-
-        # Provide a separate thread pool for operations that should not be
-        # delayed by longer background processing tasks, e.g. because the user
-        # expects instant feedback instead of waiting for a long list of
-        # operations to finish.
-        self.priority_thread_pool = QtCore.QThreadPool(self)
-        self.priority_thread_pool.setMaxThreadCount(1)
-
-        # Use a separate thread pool for file saving, with a thread count of 1,
-        # to avoid race conditions in File._save_and_rename.
-        self.save_thread_pool = QtCore.QThreadPool(self)
-        self.save_thread_pool.setMaxThreadCount(1)
 
         if not IS_WIN:
             # Set up signal handling
@@ -463,6 +448,23 @@ class Tagger(QtWidgets.QApplication):
     @property
     def is_wayland(self):
         return self.platformName() == 'wayland'
+
+    def _setup_thread_pools(self):
+        # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
+        # It's a valid reference, but its start() method doesn't work.
+        self.thread_pool = QtCore.QThreadPool(self)
+
+        # Provide a separate thread pool for operations that should not be
+        # delayed by longer background processing tasks, e.g. because the user
+        # expects instant feedback instead of waiting for a long list of
+        # operations to finish.
+        self.priority_thread_pool = QtCore.QThreadPool(self)
+        self.priority_thread_pool.setMaxThreadCount(1)
+
+        # Use a separate thread pool for file saving, with a thread count of 1,
+        # to avoid race conditions in File._save_and_rename.
+        self.save_thread_pool = QtCore.QThreadPool(self)
+        self.save_thread_pool.setMaxThreadCount(1)
 
     def pipe_server(self):
         IGNORED = {pipe.Pipe.MESSAGE_TO_IGNORE, pipe.Pipe.NO_RESPONSE_MESSAGE}
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -453,6 +453,13 @@ class Tagger(QtWidgets.QApplication):
         # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
         # It's a valid reference, but its start() method doesn't work.
         self.thread_pool = QtCore.QThreadPool(self)
+        # Limit threads to ideal thread count (based on CPUs), but reserve
+        # threads for the main thread and the priority and save thread pools.
+        # Also keep always at least 2 threads, as one will be used for the
+        # pipe server for interprocess communication.
+        max_threads = max(2, QtCore.QThread.idealThreadCount() - 3)
+        log.debug(f'Using {max_threads} worker threads.')
+        self.thread_pool.setMaxThreadCount(max_threads)
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user


Sophist-UK:
A long time ago I tried to improve Picard performance by pushing more stuff to worker threads. Part of the reason I had difficulties is the way that Python deals with multi-threading and how it decides which thread to give the GIL to next. Research done by David Beazley about 10 years ago and ignored by Python's BDfL demonstrated that when a thread ceases to be runable or its time-slice expires, all the runable threads compete simultaneously to get the GIL and this has a very big overhead. The GIL rewrite for v3.4 helped a little, but was still fundamentally still a free-for-all fight for the GIL. 

More recently (earlier this year) I tried to revive the idea of a fix for the GIL, based on a simple GIL scheduler which maintained separate CPU-light (i.e. I/O) and CPU-heavy (i.e. computation) runnable thread lists, each list itself being round robin, and where the GIL was handed off to the most appropriate thread rather than starting a fight for it. I wrote a benchmark to show the impact and overheads - and the results can be found [here](https://github.com/faster-cpython/ideas/discussions/328). Unfortunately I was still unable to gain the interest of the ex-BDfL who now runs a Python performance team for Microsoft. ☹️ 

I mention this because I have some insight into the ideal thread count for worker threads. And as you might expect, deciding how many threads to have in the thread pool is nowhere near as simple as relating it to the number of Cores the CPU can run.

For a start, Python is still highly limited by needing a GIL, which in essence restricts execution to a single thread at any one time. (There are exceptions - but these probably don't apply or make little difference to an app like Picard.) The point is that if you only ever have one thread executing, the number of threads you should have in the pool actually bears little relationship if any to the number of cores in the CPU.

On the other hand, Picard uses these threads largely so that file I/O is non-blocking and the GUI remains responsive. I/O threads generally speaking do not use much CPU, though our worker threads do I/O via Mutagen which is probably a bit more CPU intensive (though I have no measurements on just how much). Threads which are waiting for I/O use little or no CPU whilst they wait, so again, no relationship to CPU count.

So, how should we determine the size of the thread pool? Some factors:

a. File I/O throughput - IMO 99% of users will be doing all Picard I/O to a single directory structure on a single disk (either local or over the LAN) - and I suspect that 2 simultaneous file reads will max out the I/O bandwidth (but this can be tested). Doing more file parallel file reads on a spinning disk may cause head seeks and actually reduce the throughput. Additionally, since we e.g. do web service calls once a file (or album's worth of files) have been read, it is probably better to read 100 files two-by-two, than it is to read all 100 simultaneously so that they all finish being read at the same time.

b, I think Picard actually does three forms of background file I/O - directory scans when you ask it to load a directory containing your music library, reads to get the existing metadata, writes when you save metadata. We should think about whether we want e.g. directory scans to have higher "priority" than file loads, and whether file saves should have "priority" over file reads (or vice versa). If it were me I would push directory reads onto the top of the queue and file loads at the back of the queue.

c. It may be that file writes use a different bandwidth than file reads i.e. if the limitation is the LAN bandwidth rather than a spinning disk, you can send a file over the network at the same time as reading a file without having a material impact on the file reads. Not sure about e.g. SATA bandwidth, or indeed what happens inside a SSD or HDD, but the concept still holds true. So if it were me, I think I would have a separate pool of (say) two threads to use for file saves.

I guess you may want to experiment a little with different configurations and do some simple benchmarking to see what works best. 

If it were me I would start by checking whether directory scans are running on worker threads and fix that if needed - and I might experiment with the directory recursion to see if putting multiple subdirectories on different thread tasks which run in parallel boosts the scan speed cf. a single threaded recursion. I might then benchmark directory scans, reads and writes by themselves with several different thread pool sizes, and then try simultaneous scans / reads or reads / writes to see how mixed workloads change things and whether separate thread pools would help.

I would be happy to discuss any of this further (here on on a skype (or similar call) if you think it would help. 😃

phw:
> I think Picard actually does three forms of background file I/O - directory scans when you ask it to load a directory containing your music library, reads to get the existing metadata, writes when you save metadata.

directory scans currently don't run in a thread. I think Gabriel experimented with it, but I don't remember exactly why this is currently not happening in a thread. We should probably try to do that.

File saves are currently prioritized over loading (by QThread priority). But this was implemented long time ago in the very early days, so hard to tell the exact reasoning. Overall saving while reading is in most cases not good for IO anyway.


> So if it were me, I think I would have a separate pool of (say) two threads to use for file saves.

There is already a separate file saving pool, limited to one thread. Everything just lead to bad write performance in most scenarios.

Loading can be more parallelized, as there is more going on then just file IO. E.g. triggering the network requests (which is mostly waiting time) and also the metadata comparison is rather time intensive (where the most compute intensive part of `astrcmp` runs outside the GIL). But yes, clearly a thread limit makes sense. Maybe we need to split it into a separate more limited thread pool for just file loading and handle the rest in a new thread with a more open thread pool.

Gabrielcarvfer:
> More recently (earlier this year) I tried to revive the idea of a fix for the GIL, based on a simple GIL scheduler which maintained separate CPU-light (i.e. I/O) and CPU-heavy (i.e. computation) runnable thread lists, each list itself being round robin, and where the GIL was handed off to the most appropriate thread rather than starting a fight for it. I wrote a benchmark to show the impact and overheads - and the results can be found [here](https://github.com/faster-cpython/ideas/discussions/328). Unfortunately I was still unable to gain the interest of the ex-BDfL who now runs a Python performance team for Microsoft. frowning_face
> 

There are updates on that. Sam Gross has a working implementation without the GIL and is currently working on a PEP to upstream the changes. He recently made a presentation about his work in europython2022. 

Presentation: https://www.youtube.com/watch?v=9OOJcTp8dqE
Repository: https://github.com/colesbury/nogil

Sophist-UK:
@Gabrielcarvfer Thanks for adding these references. As you will see from the GIL discussion I started, I have previously read the NoGil design document, and as it states, a NoGil version of Python is still years away, even assuming that it gets adopted.. Which is why my proposal for a different way of handing off the GIL is still relevant.

So, until such time as NoGil python is a reality, my comments stand - the number of threads in a thread pool cannot currently have any real relationship to the number of CPU cores both because of the GIL and because you can potentially have many, many threads waiting on simultaneous I/O without using any CPU. 

And the latter remains true even after NoGil is implemented.

AFAIK, Picard still doesn't use worker threads (or a worker thread) for CPU-bound activities, but rather only for I/O bound activities, so I still think that the thread pool size = parallelism of I/O is related to I/O bandwidth and not CPU cores.

Of course, if Picard were to start putting CPU-heavy workload on a worker thread (e.g. the metadata summarisation over multiple selected tracks), the situation would be different. In all situations I would be recommending having separate thread pools for CPU-heavy and I/O worker thread activities. For current GIL python, the CPU-heavy threadpool should have one thread, and the I/O threadpool same number as with no CPU-heavy threads. But in this situation (current GIL, 1 CPU heavy thread) beware that the current GIL implementation will result in poor performance. For future NoGIL python, the CPU-heavy threadpool should have a number related to the number of cores (but should also be configurable so that the user can influence the number of cores that Picard soaks up).

Summary - assuming no CPU-heavy worker activities, and current Python, I stand by the advice I gave above.

Gabrielcarvfer:
>directory scans currently don't run in a thread. I think Gabriel experimented with it, but I don't remember exactly why this is currently not happening in a thread. 

As far as I can remember, it didn't really slow things down for fast drives, but made a ton of difference for HDDs (by preventing concurrent access to the disk, which means it can be scanned sequentially which is 10-30x as fast).

Sophist-UK:
Yes - I can quite believe that directory scans would screw up file loading performance on HDDs by causing significantly increased head seeking.

But then again, if you are prioritising directory scans ahead of file loads then if you were in the middle of loading one set of files, and then told Picard simultaneously to scan directories to add new files to the queue, Picard should switch completely from file loads to doing scans, before switching back to doing file loads. So this seems to be an argument for pushing new file loads to the back of the read threadpool queue and pushing directory scans onto the front of the queue (or having separate priority queues) but keeping scans and loads in the same threadpool so as to run either/or rather than in parallel.

phw:
> However, the key point as it relates to this PR is that the optimum number of threads in the pool is NOT related to the CPU count. Sorry.

You don't need to convince me about this, we don't have a disagreement here. Actually it was kind of the point of this plugin, because by default the thread pools get setup based on CPU threads available. Hence the attempt to limit it.

But your details convinced me that it would be better to separate the IO operations into a separate thread pool used just for the saving and not attempt to mix it up with other threaded action. So have a thread pool for saving with maybe only 2-4 threads. Then we can keep the main thread pool with higher count for other tasks.

> Of course, if Picard were to start putting CPU-heavy workload on a worker thread (e.g. the metadata summarisation over multiple selected tracks), the situation would be different.

This is run in a separate thread already. It's a lot Python code, so GIL is hold a lot during these, but still it improves UI responsiveness. 

The other part is clustering. But this used to be more CPU intensive before we replaced the original algorithm with a simpler and faster one in PICARD-2339 . Funny thing is that the more intense code ran on the main thread, while now it got moved to a separate thread.

Then of course plugins can use threads if they want, and they would usually use the main pool for this.

Sophist-UK:
> This is run in a separate thread already. It's a lot Python code, so GIL is hold a lot during these, but still it improves UI responsiveness.

If it isn't already, this must be in a separate thread pool with a single thread.

phw:
> If it isn't already, this must be in a separate thread pool with a single thread.

It is (there is a priority thread pool), but why is it important? If we introduce a separate thread pool for savings this could probably go back to the main thread pool (as saving is the main reason the main thread pool gets filled up).

> If this [clustering] is still CPU intensive, it should be in the same pool as above.

This would stay on the main thread pool.

> What I am not sure about is what should happen if you try to do clustering and metadata summarisation (assume huge numbers of things to cluster or summarise). I am unsure what I would want to happen as a user, much less how to achieve that in practice.

Those two are really independent of each other. The metadata update should usually happen pretty quickly, that's why this is currently run in the "priority" thread pool, so it does not wait for other tasks in the main thread pool to finish. Clustering can run in a separate thread. My expectation when I select something different would be that the metadataview updates in a short time, independent on whether the clustering is currently working or not.

You should take a look at the thread pool setup on https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L351-L371 and search the code for instances of `thread.run_task`, that should answer most of your questions about current behavior.

Might also be worth noting that the process of calculating the clusters is usually the smaller part of the full time it takes to cluster files. The large part is moving the objects around and updating the UI tree views which happens on the main thread.

Overall one of the sources for many of the remaining UI performance issue is that the updating of the UI still happens too granular and is often entangled with additional updates and calculations. E.g. when you move many files around the all get added to their new parents, doing all the necessary updates to metadata etc. while also then triggering UI update one by one. A better separation of the object updates and batching the UI updates would help here a lot. The current entanglement of both means that many parts need to run on the main thread because they include UI updates. But this is an entirely separate topic and does not belong here in more detail.

Sophist-UK:
> Overall one of the sources for many of the remaining UI performance issue is that the updating of the UI still happens too granular and is often entangled with additional updates and calculations. E.g. when you move many files around the all get added to their new parents, doing all the necessary updates to metadata etc. while also then triggering UI update one by one. A better separation of the object updates and batching the UI updates would help here a lot. The current entanglement of both means that many parts need to run on the main thread because they include UI updates. But this is an entirely separate topic and does not belong here in more detail.

Yup - I do understand the difficulties here. Other tools I have worked on have used an event model to loosely couple the GUI to the real work, and have also used a shadow model to separate GUI objects from the underlying real data objects. But this is a non-trivial change!!

phw:
> > This would stay on the main thread pool.
> 
> I thought you said that had been put into a worker thread to make the GUI more responsive?

The main thread **pool** (or maybe let's call it the primary thread pool), not the main thread. Picard currently has three thread pools setup by default:

- A primary thread pool with a default number of threads (what I referred to as main thread pool)
- A thread pool just for file saving limited to a single thread only. This was introduced very early, both because too many parallel writes hurt performance, but also I think there was some problem with saving finishing order. But I don't know the details or if they would still be relevant.
- A separate thread pool that got called "priority thread pool", also limited to one thread. The only thing scheduled there are the metadata updates. The main reason was to keep the UI responsive and not having these updates wait when the primary thread pool was busy with IO tasks

Please really see the code I linked above.

> I agree that the metadata summary should run in a separate pool, and that would have the effect of it not being queued behind a bunch of file I/O jobs, but it needs to be in a separate pool to limit how many CPU-heavy jobs run in parallel. If I have 5000 albums loaded, and select them all, the metadata updates is not quick (or at least wasn't last time I tried it).

As you can see above this is already limited to one thread. One reason was that it was the quickest solution to ensure we don't get a lost update situation when later updates finish before older updates. But there would be other ways to solve this. But if it is beneficial to have this limited regardless then we can keep the status quo.

The metadata update clearly is the slowest non IO action currently. Getting the metadata for ~25k files uses about 2.9 seconds on my system.

Clustering on the other hand does not have this impact. Clustering the 25k files takes around 150ms. As that's not something that is run frequently you won't notice a slow down in loading / saving thread performance due to this.

Sophist-UK:
I agree, 150ms is not too great a time to be on the normal thread pool. and that allows one metadata summary job and one or more Clustering jobs and several I/O jobs to be executing on the various thread pools at the same time.

Gabrielcarvfer:
I was giving it a try, and I believe a barrier that locked the workers until the screen repaint would make it butter smooth.
Since I have no idea how to do that, I did the following. Seems to work just fine.
```
class Runnable(QRunnable):

    def __init__(self, func, next_func, traceback=True):
        super().__init__()
        self.func = func
        self.next_func = next_func
        self.traceback = traceback

    def run(self):
+      if QCoreApplication.instance().hasPendingEvents():
+          time.sleep(0.05+random.random()/4) # make workers sleep 50ms-300ms
        try:
            result = self.func()
        except BaseException:
            if self.traceback:
                log.error(traceback.format_exc())
            to_main(self.next_func, error=sys.exc_info()[1])
        else:
            to_main(self.next_func, result=result)
```

Sophist-UK:
The proposal I was making in the faster-python discussion was to implement a round-robin approach to giving threads the GIL rather than have them compete with some threads getting more than their fair share, and some threads less - and when the thread getting less is the main thread, the GUI starts lagging. More recent versions of Python have a kludge to stop a single thread keeping the GIL at the end of a time-slice, but you can still get two threads that monopolise the GIL, locking out other threads, even the main GUI thread.

What @Gabrielcarvfer 's code would seem to do is delay **the start** of the next job in order to allow `next_func` and any other main thread activities to process before starting a new worker job, which sounds exactly what you want to happen.

It's probably a bit more complicated than this, but that is the gist of it.

However I am not sure whether the delay values you have chosen are optimal or indeed whether a random amount is needed at all, but certainly 50ms-300ms seems way too long to me (especially 300ms!!!!!) compared to a Python time-slice of 5ms.

So I am wondering whether the following code would be better:
``` python
class Runnable(QRunnable):

    def __init__(self, func, next_func, traceback=True):
        super().__init__()
        self.func = func
        self.next_func = next_func
        self.traceback = traceback

    def run(self):
+      while (QCoreApplication.instance().hasPendingEvents()):
+          time.sleep(0.01)

        try:
            result = self.func()
        except BaseException:
            if self.traceback:
                log.error(traceback.format_exc())
            to_main(self.next_func, error=sys.exc_info()[1])
        else:
            to_main(self.next_func, result=result)
```

With this the main thread would get an increasing probability of getting the GIL as each worker thread ends a job and doesn't start a new one until the GUI has processed everything - and eventually it is guaranteed to get the GIL because all worker threads are sleeping.

@Gabriel - do you fancy giving the above code a try?

Sophist-UK:
I also just found [this article](https://tenthousandmeters.com/blog/python-behind-the-scenes-13-the-gil-and-its-effects-on-python-multithreading/) from last year which suggest that changing the default switch interval from 5ms to something significantly smaller can help the throughput of I/O when there are CPU-heavy threads competing for the GIL.

So in addition to the above code (with the 0.01 sec value changed to 2*the chosen switch-interval) it might be worth experimenting with changing the switch interval from 5ms (0.005) to 1ms (0.001) or 0.1ms (0.0001) or even 0.01ms (0.00001). At these very small values, the optimum value might depend on the single core processing speed of the specific PC, but this is IMO worth a try to see whether changing the switch-interval at Picard start-up makes a significant difference.

Sophist-UK:
@Gabrielcarvfer As I frequently say, an ounce of practical experimentation is worth a ton of theory!!

However, I would still suggest you experiment with the values, as 50ms minimum sounds much bigger than it needs to be and 300ms is, in terms of computer speed, an eon. Would 10ms-100ms be a better spread? (This is a gut reaction, not really even theory based, so experimentation will be needed if you haven't already done this.)

Gabrielcarvfer:
@Sophist-UK tried 10-100ms, it just isn't smooth (at least on my desktop).

Tried setting the sys.switch something something you mentioned to 1ms, and it does make everything super smooth, but it's 10x as slow as the 50-300ms sleep I suggested.

Sophist-UK:
The sys.setSwitchTime is a different setting to the sleep time - and you need to optimise them independently, probably through iteration as they will effect each other. But your measurement do show that an ounce of experimentation is worth a ton of theory.

But I still believe that a delay of 300ms is a crazy amount of time to sleep, and that a SwitchTime of < 5ms (in combination with a different sleep amount) might well improve throughput.

jesus2099:
@Sophist-UK 

> a. File I/O throughput - IMO 99% of users will be doing all Picard I/O to a single directory structure on a single disk (either local or over the LAN) - and I suspect that 2 simultaneous file reads will max out the I/O bandwidth (but this can be tested). Doing more file parallel file reads on a spinning disk may cause head seeks and actually reduce the throughput. Additionally, since we e.g. do web service calls once a file (or album's worth of files) have been read, it is probably better to read 100 files two-by-two, than it is to read all 100 simultaneously so that they all finish being read at the same time. I am guessing that a fixed thread pool of 2-4 would max out the file read bandwidth.

Could this be further limited to **1 by 1** file reading (when loading files to display their duration in left-hand panel) and AcoustID scanning?

It would nice when [reading CD drive directly](https://tickets.metabrainz.org/browse/PICARD-1945).

phw:
@jesus2099 We are considering here limiting it to a low number. Maybe 3 or 4. But generally setting it to 1 is too restrictive. Parallel processing of file reading has for the vast majority of use cases a speed benefit. The CD reading really is an edge case.

Also the current planning focuses mostly around loading of files. fpcalc processing runs in separately spawned processes. It's currently limited to 2 parallel tasks already.

What we could do is exposing options for users to manually change this.

But my general recommendation for you would be to first rip the tracks then run all the loading and fpcalc analysis on it. Doing this from CD drive is slow. Picard and fpcalc assume they can freely access the files and seek in them to read from arbitrary positions. If you have the data as files locally on your disc. But if you read only from the virtual file system the virtual file system might need to jump around on  the CD to read the requested data.

Sophist-UK:
> What we could do is exposing options for users to manually change this.
👍

jesus2099:
@phw thanks for your answer!
For the record, I have already tested, even doing it manually, 1 by 1 drag and drop then 1 by 1 scan, is slightly more faster for me than ripping first (with another app), then delete files after AcoustID scan and submit.
If I could deactivate multithreading, it would then be very much faster.

I'm just looking at quick ways of submitting AcoustID, otherwise I would submit them. ;)

But maybe I will find a Linux command line way of doing it, like for [isrcsubmit](http://jonnyjd.github.io/musicbrainz-isrcsubmit/).

Sophist-UK:
> For the record, I have already tested, even doing it manually, 1 by 1 drag and drop then 1 by 1 scan, is slightly more faster for me than ripping first (with another app), then delete files after AcoustID scan and submit.

Faster for the computer, not faster for you because this is a much much much more manual workflow that telling Picard to get on with it and leaving it overnight whilst you sleep.

> If I could deactivate multithreading, it would then be very much faster.

You can't! Picard would lock up without multi-tasking!

However, if the threadpool size were adjustable e.g. with a command line option, then you could have a threadpool size of 1 which is what I think you are asking for.

jesus2099:
> Faster for the computer, not faster for you because this is a much much much more manual workflow that telling Picard to get on with it and leaving it overnight whilst you sleep.

Sorry to insist but no, I still have to match CD track to MB release tracks *1 by 1*, even if I rip, so it could not be done overnight without my intervention.
And I add things CD by CD, and I don't rip, don't tag.

Sophist-UK:
Well, Picard should not need you to match each track manually. 

And if you don't rip and the files are on a read-only CD, how do you tag them?

Are you sure that there isn't a better way than this for you to manage your music library?

phw:
@Sophist-UK jesus2099 is just using Picard to contribute new AcoustID fingerprints. The files are hence not of interest. As this is mostly about submitting new fingerprints the matching is needed.

Sophist-UK:
Ah - well, I guess that this is not really Picard's mission - adding AcoustIDs is normally a by-product addition of using Picard to tag music files and manage your library.

But, if @jesus2099 was managing his music library as e.g. MP3 files I would imagine that he would find this easier.

phw:
I did some performance measuring. The results are kind of interesting.

I measured the time between adding paths to Picard and the last file being loaded (the counter of files to process reaching zero) and did this with varying number of separate loading threads. I usually did each run three times and took the average . I did this with the same library of 4.4k MP3 files, but loaded from various storage mediums. The files are all tagged, but I did the same test twice, once with "Ignore MBIDs" enabled (so the files stay on the left and apart from the loading nothing else happens in Picard) and with this disabled (so the files move over to the right and Picard starts loading releases in parallel).

On Linux I cleaned the disk buffers before each run with `sysctl vm.drop_caches=3`. On macOS I tried to clear the disk buffers with `purge && sync`. On Windows I could not find a way to clear disk buffers (despite doing a lot of IO in between, which I did not want to do), so I did the full test without any clearing. I did a few runs with rebooting in between, but the end results are actually not so different.

Please note that these tests only are about pure loading performance. They don't tell anything about responsiveness, but I'll have a few words about this at the end.

Here are the results:

https://docs.google.com/spreadsheets/d/1N75wfEJlQ7jmO4IdxOREseUAiCQ-7iIiABIT9hpK_Pg/edit

What I find interesting is that there is no clear picture. In some cases loading with more threads slows down loading significantly. This was the case on Linux on my main SSD and with an external USB HDD with NTFS.

But you can also have huge speed gains. The external USB HDD with LUKS+BTRFS needs only 67% of the time with 8 threads than it did with 1 thread. This one surprised me, but it's probably because of the higher CPU use the encryption causes. Less surprising is that parallel loads also increase speed for loading from a network share. But I only did limited runs here due to the long loading time involved, also this was the least stable test setup, as the laptop doing the serving had to be in slight use and Wifi was also in use in parallel.

Then for some setups there is no real conclusion possible. E.g. I tried it on my old mac Mini from 2014. This one has a rather slow HDD. I expected parallel IO to be bad here, but actually there might be a small gain from more threads. But also the variation in loading times was so high (even though the mac was not doing anything else and just running the benchmark) that you can't really draw a definitive conclusion.

Similar on Windows, no big difference in loading times (but more consistent and less varying than on the mac).

Now some notes about responsiveness of the UI: I can't really measure this, but I did try to use Picard while loading is happening. I need to do this some more, but a few things are already interesting:

- Responsiveness varies between setups. Surprisingly the old mac mini is doing quite well
- Higher loading thread count does not generally imply less responsive UI. If you think about it this is not very surprising: The main time in many setups is IO. It has zero impact on the UI responsiveness if you have 8 additional threads sitting there waiting for IO. The GIL is not blocked, the CPU is free to do UI updates.
- I got some of the worst UI locks on the fastest setup! On Linux loading from internal SSD with only a single thread I had some serious lags in the UI (so bad that the app got reported as unresponsive by the window manager). In general the faster setups had more UI lag, the slower ones less. Loading from network share is super slow, but Picard UI is not lagging. This again fits into the previous comment: If you have threads waiting long for IO, it does not impact UI. If you then load only a single file every second and do this update the main thread does not really have heavy work to do. But if the IO is happening quickly and all the work inside the GIL locked Python code for many files happens in bulks then it can impact the UI massively.

This also in parts answers one of the mysteries I had: Over the many years I worked with Picard my hardware got better and there have been improvements to Picard and Python but yet loading many files felt worse than I remembered it to be.

Now I'm not sure what conclusion to draw from that. One thing might be that because there is no hard rule of more or less threads being better, maybe making the number of loading threads configurable is really the way to go.

Sophist-UK:
As I previously suggested, this does show just how an ounce of practical experimentation is worth a pound of theory.

I have no idea what CPU load the network JSON decoding creates but some years ago when it was XML I did profile it and discovered just how CPU intensive XML processing was. But I would guess that the ignore MBID runs would be more indicative of file I/o alone, and the standard runs more indicative of real life.

For the encrypted LUKS+BTRFS runs, again this doesn't surprise me. I am assuming that the decryption is done by the O/S and python does not require the GIL during this, but (obviously) the I/O subsystem is not in use during encryption so more threads would allow this to be used and improve performance. I would anticipate that the same thing would apply for encrypted filesystems on Windows.

What is surprising is that GUI responsiveness was bad when there was a single I/O thread. I am struggling to understand what would cause this - when this single thread was in I/O wait, why would the main thread not get CPU? Is this a python bug do you think? Or something in the Picard code?

phw:
> What is surprising is that GUI responsiveness was bad when there was a single I/O thread. I am struggling to understand what would cause this - when this single thread was in I/O wait, why would the main thread not get CPU?

It is exactly the opposite. It's not that the IO is locking anything, in this scenario there is the least IO wait. For every file being loaded you have IO followed by the post processing done in Python (still in the thread) + update of the GUI on the main thread.

In all the measurements I did the same files were being used, so in each scenario you need to do exactly the same IO independent work.

But if the IO takes only 20 seconds all the additional work for the 4.4k files takes place in a short timeframe, and that can be noticable in UI lags. On the other extreme in the networking case the same processing is stretched over an hour because IO is super slow. The occasional non-IO processing is hardly noticeable in the UI.

Sophist-UK:
@phw Philipp What I think you are saying is that we should go for the best I/O performance, and have a separate thread pool of 1 that we can move more CPU intensive work to in order to allow the GUI to remain responsive (though CPU intensive threads do have some other smaller issues due to the poor GIL transfer implementation).

phw:
@Sophist-UK No, that's not really what I mean with this.The general thing is of course true (move performance intensive work into threads) and there might be some improvements here, but it's not really the core part of that.

What my observations imply is that if we want to improve UI responsiveness in these scenarios of loading many files pretty fast is that we need to look into exactly what is happening beside IO when a file gets loaded. The IO is the rather uninteresting part (because it is just thread idle time). This needs some detailed profiling to see where really the main time is being spent.

But to make some assumptions beforehand I fear this will always be very restricted, as beside IO everything else happens in Python. When loading a file at least theoretically all that is happening outside IO is:

- mutagen parsing the data and building the metadata info
- Picard preparing the metadata for internal use
- Picard setting up all the objects and structures (that's the catch all here, but for files it is not much)
- Running `file_post_load_processors` (but there are barely any plugins utilizing this)
- The UI getting updated

The first two already run in a separate thread, but there is not much potential to release the GIL here. The last one always needs to run on the main thread. Plugins are started on the main thread, I think the idea here is that plugins that need it should spawn their own thread.

So whatever bottleneck we have here, threading is likely not the answer. If you ask me my main candidate for better UI responsiveness would be the UI update part. This is where we got the most significant results in the past by reducing the number of UI updates needed. Gabriel did a lot here. But I think there is still potential.

E.g. right now after loading a file it gets moved to the parent (e.g. unclustered files in my example). The cluster adds the item to its tree view representation and updates the UI. This happens for each file individually. If we would do this for multiple files at once instead and update the tree view only once it might improve performance. This is just a theory and I don't know exactly how much refactoring is needed to achieve this.

But as I said, the next step is deeper profiling to find the bottleneck.

phw:
Turns out that there is a specific piece of code being run on main thread that is slowing the main thread down significantly. It's the metadata comparison in file.update: https://github.com/metabrainz/picard/blob/master/picard/file.py#L693-L719

So @Sophist-UK , you were right, we have a specific CPU intensive part that is being an issue here. Furthermore, if I run this piece of code in a thread the UI becomes very responsive during load again :)

I'm currently checking if this has some unintended side effects.

phw:
I'd appreciate some testing and feedback on https://github.com/metabrainz/picard/pull/2190

Gabrielcarvfer:
> I'd appreciate some testing and feedback on #2190

Nice and smooth with cold cache. Stuttering with hot cache.

phw:
Thanks for testing. This sounds like 5ge same IO paradox again as described above. Essentially the faster the IO the more the rest of the code influences the loading. In it's extreme IO would be 0 and the application would just do all the GIL locked code at once, causing the UI to hang.

I also noticed the code has a problem: the file load threads run with higher priority in the queue. That means the file.update calls all get queued up behind all the loading calls. This means the code does not run at all during loading, and runs all at once after the last file loaded. This gave me the smooth experience during load, but after the last file loaded the GUI hangs for a second while all the update calls get processed.

Ideally the update should be at least the same priority as loading, probably even higher to not delay these updates too much. If I change the priority accordingly the loading is less smooth again, but still more responsive then before.

phw:
Closing this for now, this clearly was the wrong approach. The issues I was trying to address has been better fixed by #2191.

However, we might revisit limiting the parallel worker threads in some way. But not like it was done here.

PR #2173, opened 2022-11-19
PICARD-2546: Sequential processing of executable commands

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Process commands through a queue object.  Track threads created during command processing.  Add a blocking `to_main()` function.

# Problem

As noted in PICARD-2546, commands are processed in separate threads so they cannot be sequenced.

* JIRA ticket (_optional_): PICARD-2546

# Solution

Process command line executable commands through a queue object:

- Implement a queue object to hold the executable command queue
- Implement a command processor thread to parse the command queue
- Implement command file tracking to avoid circular references
- Do not queue additional commands after a `QUIT` has been queued

Track threads created during executable command processing for consideration before marking the command as having been completed:

- Add a new blocking version of the `to_main()` function
- Track the status of threads created while an executable command is being processed
- Block new command execution until current command is complete
- Test for completion includes all threads created during the command processing

Update and add new tests.

# Action

I realize that this method is not perfect because it blocks on threads started during command execution including those that have nothing to do with the command, and currently only blocks threads started with the `Runnable` class.  I welcome any suggestions for improving this.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -55,6 +55,7 @@ import shutil
 import signal
 import sys
 from textwrap import fill
+import time
 from urllib.parse import urlparse
 from uuid import uuid4
 
@@ -395,35 +396,24 @@ class Tagger(QtWidgets.QApplication):
         RemoteCommands.set_running(False)
 
     def run_commands(self):
-        # Provide a set of commands that should not automatically release the 'self.command_running'
-        # flag when the command execution completes.  For example, loading a release creates a number
-        # of sub-tasks on separate threads and requires additional process checks to ensure that all
-        # sub-tasks are complete before the 'self.command_running' flag should be released.
-
-        # no_auto_complete = {'LOAD', 'CLUSTER'}
-
         while not self.stopping:
             if not RemoteCommands.command_queue.empty() and not RemoteCommands.get_running():
                 (cmd, arg) = RemoteCommands.command_queue.get()
                 cmd = cmd.upper()
-                arg = arg.strip()
                 if cmd in self.commands:
+                    arg = arg.strip()
                     log.debug("Executing command: %s %r", cmd, arg)
                     RemoteCommands.set_running(True)
+                    thread.to_main_with_blocking(self.commands[cmd], arg)
 
-                    # FIXME: Find a way to execute each command using a block to ensure completion
-                    # before executing the next command.  Perhaps track completion of all threads
-                    # created while "command_running" is True and only clear_command_running()
-                    # when all those threads have completed?  That may also remove the need for
-                    # the special "no_auto_complete" processing.
-
-                    thread.to_main(self.commands[cmd], arg)
-                    # if cmd in no_auto_complete:
-                    #     break
-                    self.clear_command_running()
+                    # 30 second timeout to avoid hanging up command processing indefinitely
+                    RemoteCommands.wait_for_completion(30)
+                    RemoteCommands.clear_command_running()
+                    log.debug("Completed command: %s %r", cmd, arg)
                 else:
                     log.error("Unknown command: %r", cmd)
                 RemoteCommands.command_queue.task_done()
+            time.sleep(.01)
 
     def _run_commands_finished(self, result=None, error=None):
         if error:
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -50,9 +50,7 @@ from hashlib import md5
 import logging
 import os
 import platform
-import queue
 import re
-import shlex
 import shutil
 import signal
 import sys
@@ -137,6 +135,10 @@ from picard.util import (
 )
 from picard.util.cdrom import get_cdrom_drives
 from picard.util.checkupdate import UpdateCheckManager
+from picard.util.remotecommands import (
+    REMOTE_COMMANDS,
+    RemoteCommands,
+)
 from picard.webservice import WebService
 from picard.webservice.api_helpers import (
     AcoustIdAPIHelper,
@@ -177,25 +179,6 @@ def plugin_dirs():
     if not os.path.exists(USER_PLUGIN_DIR):
         os.makedirs(USER_PLUGIN_DIR)
     yield USER_PLUGIN_DIR
-
-
-class CommandFiles:
-    _command_files = set()
-
-    # Maintain a flag to indicate whether a 'QUIT' command has been queued
-    quit_flag = False
-
-    @classmethod
-    def contains(cls, filepath):
-        return filepath in cls._command_files
-
-    @classmethod
-    def add(cls, filepath):
-        cls._command_files.add(filepath)
-
-    @classmethod
-    def remove(cls, filepath):
-        cls._command_files.discard(filepath)
 
 
 class ParseItemsToLoad:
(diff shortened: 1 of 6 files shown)


rdswift:
Would it make sense to create a new class to hold the command queue as well as the contents of the CommandFiles class?  It could possibly also hold the parsing methods, thus keeping things all together.

picard/tagger.py
```diff
@@ -470,31 +383,49 @@ def pipe_server(self):
...
-                thread.to_main(file_lookup.mbid_lookup, item, None, None, False)
+    def clear_command_running(self, *args, **kwargs):
+        RemoteCommands.set_running(False)
+
+    def run_commands(self):
+        while not self.stopping:
+            if not RemoteCommands.command_queue.empty() and not RemoteCommands.get_running():
+                (cmd, arg) = RemoteCommands.command_queue.get()
+                cmd = cmd.upper()
+                arg = arg.strip()
```
zas, line 402: You can move this line in the `if` branch below, as `arg` is only used there.

zas (changes requested):
It looks like a good improvement, I didn't test it yet, but I will soon.

zas (changes requested):
I noticed multiple problems with this patch:

- `python tagger.py -e 'load /home/zas/Musique` doesn't work, it works with `master`
- `ctrl-C` quitting from command line doesn't work (I even at to `kill -9` the process, it was using 100% cpu), it works fine with `master` branch

rdswift:
> I noticed multiple problems with this patch:
> 
> * `python tagger.py -e 'load /home/zas/Musique` doesn't work, it works with `master`

That's odd, because it works here under both Windows and Linux.  Do the logs show anything such as "Unknown command: 'LOAD"?  Were you using this to send the command to a running instance, or was this a fresh instance?  I'll try doing some more testing.

> * `ctrl-C` quitting from command line doesn't work (I even at to `kill -9` the process, it was using 100% cpu), it works fine with `master` branch

Yeah, I forgot to include a `sleep` in a couple of the processing loops to allow for system processing.  Added now.

zas:
> That's odd, because it works here under both Windows and Linux. Do the logs show anything such as "Unknown command: 'LOAD"? Were you using this to send the command to a running instance, or was this a fresh instance? I'll try doing some more testing.

I retested right now, and it works... weird. I cannot reproduce it anymore.

zas (dismissed):
LGTM

rdswift:
In the course of further testing I found a bug in the function to write the log to a file.

I added a new `PAUSE` command as a (temporary) way of adding a time delay between commands to allow background processing to complete.  I've been using this to help identify tasks that are starting in the background and aren't currently being trackd for completion.

I created a `command_task_tracking` decorator to handle adding and removing tasks from the tracking set, and some related refactoring to simplify and reduce code duplication.  I also took the liberty of adding a default `_no_operation()` function for the `next_func` parameter to the `thread.run_task()` function.

The sequential command processing seems to work reasonably well, although there is still some work to do to catch and track all the threads that run in the background as part of the command processing.  I think that the new decorator should help make that a bit easier as the tasks are identified.

phw:
Tested this locally, seems to work pretty good. It's nice to be able to sequence tasks this way.

rdswift:
In addition to finding and fixing some more bugs in the existing command handling code, I think that the task tracking decorators are **finally** applied so that the commands are processed sequentially.  I suspect that there are still some that can be removed, but they don't seem to impact the performace during my testing.  The commands that I have tested include:

- LOAD single file
- LOAD directory with multiple files
- LOAD directory with subdirectors containing multiple files
- LOAD mbid://662cea72-d3f5-467d-8c45-9260c2f8f47e
- LOAD https://musicbrainz.org/release/662cea72-d3f5-467d-8c45-9260c2f8f47e
- FROM_FILE with additional commands
- FROM_FILE with missing file (logs an error and continues processing)
- FROM_FILE with circular reference (logs a warning, ignores the file and continues processing)
- CLUSTER
- LOOKUP CLUSTERED
- SCAN
- SAVE_MATCHED
- REMOVE_SAVED
- REMOVE_UNCLUSTERED
- REMOVE_EMPTY
- REMOVE_ALL
- FINGERPRINT
- SUBMIT_FINGERPRINTS
- PAUSE
- WRITE_LOGS
- QUIT

I haven't (yet) tested the LOOKUP_CD commands because of not being able to install the necessary python-libdiscid library on my Windows 7 system.

zas (commented):
I'm a bit worried about the spread of `command_task_tracking` decorator. It is unclear to me which methods should have it or not (and why).
I understand which problem it solves, but I think it will make things hard to maintain over the time.

Now I'm not sure if there's any alternative to that.

@phw what do you think?

rdswift:
I'm a novice when it comes to threading.  Is there a built-in way to track the status of all sub threads created from a parent task so that the parent can be blocked until all the sub threads are completed?  Something similar to processes?  That might provide an alternative.

rdswift:
I just had another thought...  Perhaps we could store the number of active threads in the appropriate thread pools when a command is started, and then block the start of the next command until the number of active threads comes back down to that number.  It's not the greatest because it could incorrectly release the block if a previously running thread completes, but that should really only be an issue during startup.  I'll look at this a bit more to see if it makes sense (and if I can get it to work), but it will be a couple of days before I get a chance to look at it.

phw:
I think the alternative might be to use a separate thread pool with only a single thread just for command processing. That way only a single thread ever runs in parallel and the commands hence get serialized.

We also did this for the file saving. One reason was performance, but one had something to do with serializing the saving (I don't remember the details why it was necessary).

rdswift:
I completely changed the methodology that I was using for the command completion testing to now use the counts of the active threads in the thread pools as well as the count of pending webservice requests and acoustid fingerprinting tasks running.  This greatly simplified the code, allowing the removal of all of the `@command_task_tracking` decorators.  I tested all of the executable commands here and they seem to be properly separated for sequential processing.

picard/tagger.py
```diff
@@ -595,6 +533,20 @@ def handle_command_lookup_cd(self, argstring):
             partial(self._lookup_disc, disc),
             traceback=self._debug)
 
+    def handle_command_pause(self, argstring):
+        arg = argstring.strip()
+        if arg:
+            try:
+                _delay = float(arg)
```
zas, line 540: Is there a specific reason to use `_delay` over `delay` here?

zas (changes requested):

picard/tagger.py
rdswift: No reason at all.  I've made the change.

In addition, I've removed the 30 second command timeout.  After thinking about it I couldn't come up with a single situation when it might be needed, but it could cause a long running command to show as being completed prematurely causing problems with the sequencing.  Some examples could be: scanning or looking up a large number of clusters; loading a release with a lot of coverart images; or fingerprinting a large number of files.

zas (approved):
LGTM

phw (approved):

PR #2175, opened 2022-11-28
PICARD-2591: Accessing CD Lookup crashes Picard if there is no cdtoc module loaded

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Disable executable `LOOKUP_CD` commands if there is no `discid` module loaded.

# Problem

The main interface deactivates the Lookup CD items from the tools menu, but they can still be initiated using the executable command processing.

* JIRA ticket (_optional_): PICARD-2591

# Solution

Check if `picard.util.cdrom.discid` is `None` and if so disable the `LOOKUP_CD` executable command.

# Action

None.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -135,6 +135,7 @@ from picard.util import (
     webbrowser2,
 )
 from picard.util.cdrom import (
+    DISCID_NOT_LOADED_MESSAGE,
     discid as _discid,
     get_cdrom_drives,
 )
@@ -577,7 +578,7 @@ class Tagger(QtWidgets.QApplication):
 
     def handle_command_lookup_cd(self, argstring):
         if not _discid:
-            log.error("CDROM: discid library not found - Lookup CD functionality disabled")
+            log.error(DISCID_NOT_LOADED_MESSAGE)
             return
         disc = Disc()
         devices = get_cdrom_drives()
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -99,6 +99,7 @@ from picard.util import (
     webbrowser2,
 )
 from picard.util.cdrom import (
+    DISCID_NOT_LOADED_MESSAGE,
     discid,
     get_cdrom_drives,
 )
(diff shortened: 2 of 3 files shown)


phw (approved):
Thanks for this fix, looks good.

Side note: I removed the milestone on the ticket, as this does not fix an issue of current release version and hence does not need to show up in the changelogs. In general if we fix an issue that got introduced with recent, yet unreleased, development changes and there is a PR we don't necessarily need a ticket in the issue tracker. The PR itself is enough.

But it's Ok to add an issue if unsure. I'll usually will just clear the milestone then, otherwise compiling the changelogs gets more complicated on release date.

picard/tagger.py
```diff
@@ -573,6 +576,9 @@ def handle_command_lookup(self, argstring):
             log.error("Invalid LOOKUP command argument: '%s'", argstring)
 
     def handle_command_lookup_cd(self, argstring):
+        if not _discid:
+            log.error("The Lookup CD function is not available.")
```
phw, line 580: Let's change the wording here. We have the same check already in https://github.com/metabrainz/picard/blob/master/picard/ui/mainwindow.py#L745 , I think it would be a good idea to use the same wording.

But also different to the check in mainwindow let's keep this here an error (as you currently did it). Trying to execute this command without discid being avaible really counts as an error message.

phw (changes requested):
Let's change the

rdswift:
> Side note: I removed the milestone on the ticket, as this does not fix an issue of current release version and hence does not need to show up in the changelogs. In general if we fix an issue that got introduced with recent, yet unreleased, development changes and there is a PR we don't necessarily need a ticket in the issue tracker. The PR itself is enough.

Okay, I'll try to remember that in the future.  I wasn't sure, so I thought the safest would be to add the ticket.  For some reason I forgot that you used the milestones on the tickets to generate the change log for the release and blog post.  Now that you reminded me, removing it makes perfect sense.  Thanks.

picard/tagger.py
rdswift: Done, but in a case like this, should I be adding "CDROM: discid library not found - Lookup CD functionality disabled" as a string variable/constant in `picard/disc/__init__.py` and including it in `picard/tagger.py` and `picard/mainwindow.py` to avoid maintining the same message in two different locations?

rdswift: I actually decided to add the string constant to `picard/util/cdrom.py` since that's the module that was already providing `discid` that was used for testing.  If you don't want this done, please let me know and I can revert the changes.

zas (approved):
LGTM

phw (approved):

PR #2177, opened 2022-12-04
Extended compatibility options for file names

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-356, PICARD-2553, PICARD-2576

Allow more customization of file compatibility options.

# Solution

This addresses multiple issues by providing extend file name compatibility options.

- Moved the options "Replace non-ASCII characters", "Windows compatibility" and "Allow paths longer than 259 characters" to a separate "Compatibility" page (below "File Naming"). This also allows a smaller minimum height for the Options dialog and hence addresses PICARD-2576 at least partially
- Added a new option "Replace spaces with underscores", which allows using underscores instead of spaces in file names (PICARD-2553)
- Adds a "Customize..." dialog for Windows compatibility, where the replacement characters for each of the Windows special characters can be separately specified by the user (default stays the underscore for all characters).

grafik

# Action

This is setting the base for more possible options, which now have a good place in the Options dialog. E.g. defining the replacement character for directory separators used in tags is something that is often requested.

Also we could add an additional general file name character replacement configuration. It would work similar to the Windows compatibility replacement, but allow replacing characters independently. The Windows compatibility would stay as a special important case.

### DIFF
--- a/picard/ui/options/renaming_compat.py
+++ b/picard/ui/options/renaming_compat.py
@@ -57,6 +57,9 @@ from picard.ui.ui_options_renaming_compat import Ui_RenamingCompatOptionsPage
 from picard.ui.ui_win_compat_dialog import Ui_WinCompatDialog
 
 
+DEFAULT_REPLACEMENT = '_'
+
+
 class RenamingCompatOptionsPage(OptionsPage):
 
     NAME = "filerenaming_compat"
@@ -71,13 +74,13 @@ class RenamingCompatOptionsPage(OptionsPage):
         BoolOption("setting", "ascii_filenames", False),
         BoolOption("setting", "replace_spaces_with_underscores", False),
         Option("setting", "win_compat_replacements", {
-            '*': '_',
-            ':': '_',
-            '<': '_',
-            '>': '_',
-            '?': '_',
-            '|': '_',
-            '"': '_',
+            '*': DEFAULT_REPLACEMENT,
+            ':': DEFAULT_REPLACEMENT,
+            '<': DEFAULT_REPLACEMENT,
+            '>': DEFAULT_REPLACEMENT,
+            '?': DEFAULT_REPLACEMENT,
+            '|': DEFAULT_REPLACEMENT,
+            '"': DEFAULT_REPLACEMENT,
         })
     ]
 
@@ -177,6 +180,9 @@ class WinCompatDialog(PicardDialog):
         self.ui.replace_quotationmark.setValidator(WinCompatReplacementValidator())
         self.ui.buttonbox.accepted.connect(self.accept)
         self.ui.buttonbox.rejected.connect(self.reject)
+        reset_button = QtWidgets.QPushButton(_("Restore &Defaults"))
+        self.ui.buttonbox.addButton(reset_button, QtWidgets.QDialogButtonBox.ButtonRole.ResetRole)
+        reset_button.clicked.connect(self.restore_defaults)
         self.load()
 
     def load(self):
@@ -197,6 +203,15 @@ class WinCompatDialog(PicardDialog):
         self.replacements['?'] = self.ui.replace_questionmark.text()
         self.replacements['"'] = self.ui.replace_quotationmark.text()
         super().accept()
+
+    def restore_defaults(self):
+        self.ui.replace_asterisk.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_colon.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_gt.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_lt.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_pipe.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_questionmark.setText(DEFAULT_REPLACEMENT)
+        self.ui.replace_quotationmark.setText(DEFAULT_REPLACEMENT)
 
 
 register_options_page(RenamingCompatOptionsPage)


rdswift:
Very nice!  A quick question...  Does this allow replacing with a string or is it limited to a single character for each entry?  If a string of characters, is this wanted behavior?  Does it allow an empty string as the replacement character?

This is also a reminder to myself that the documentation will need to be updated once this has been merged.

zas:
Maybe it would be convenient to have a button to restore defaults in Windows compatibility dialog.
Apart that, the patch looks good.

phw:
In theory it could support a string, but I decided to be conservative for now and allow only a single character. I don't wanted to allow too much we then need to revert, I'd rather see what use cases users actually will have.

But I allowed the empty string, so forbidden characters can be removed.

My current plan is to add two additional options to the compatibility page:

- Replacement character for directory separators in tags: As this needs to be handled on the tags before the naming script it makes sense to have this as a separate option
- Additional character replacements: This would be a mapping that would be applied after the script ran but before the Windows compatibility. That way Windows compatible characters will still be enforced (if enabled), but the user gets the freedom to do some additional replacements of other characters. Here it probably will be useful to allow strings both for search and replacement. That way one could e.g. replace three dots with ellipsis.

I think all this should solve the common user requests we get for these topics.

phw:
> Maybe it would be convenient to have a button to restore defaults in Windows compatibility dialog.

Good point, I added that

picard/ui/options/renaming_compat.py
```diff
@@ -0,0 +1,215 @@
...
+    ACTIVE = True
+    HELP_URL = '/config/options_filerenaming_compat.html'
+
+    options = [
+        BoolOption("setting", "windows_compatibility", True),
+        BoolOption("setting", "windows_long_paths", system_supports_long_paths() if IS_WIN else False),
+        BoolOption("setting", "ascii_filenames", False),
+        BoolOption("setting", "replace_spaces_with_underscores", False),
+        Option("setting", "win_compat_replacements", {
+            '*': '_',
```
zas, line 74: https://github.com/metabrainz/picard/pull/2177/files#diff-9fd922f0357a6d93bbfac9746d8a3f164d73a879a6b68923fd0f5af83bd148f8R205 defines a local variable `default`, we could use a class level constant and use it here too.

zas (changes requested):

picard/util/__init__.py
```diff
@@ -293,16 +296,20 @@ def sanitize_date(datestr):
...
-        drive, rest = ntpath.splitdrive(string)
-        return drive + _re_win32_incompat.sub(repl, rest)
+        drive, string = ntpath.splitdrive(string)
     else:
-        return _re_win32_incompat.sub(repl, string)
+        drive = ''
+
+    replacements = defaultdict(lambda: repl, replacements or {})
+    for char in {'"', '*', ':', '<', '>', '?', '|'}:
+        string = string.replace(char, replacements[char])
```
zas, line 310: In Python3, there are `str.translate()` and `str.maketrans()`, but not sure if it would be faster and/or more convenient.
Also I think testing the presence of  `char` in `string` before calling `str.replace()` is faster in some cases .

```suggestion
        if char in string:
            string = string.replace(char, replacements[char])
```

In our case, it is unlikely we replace all chars I guess.

A quick performance test (far from complete):


```python
import timeit


s_occurences = {
    0: 'abcdefghijklmnop',
    2: 'abcd"fghijk*mnop',
    4: 'ab:d"fghijk*mn<p',
    6: 'ab:d"fg?i>k*mn<p',
    8: 'a*:d"fg?i>k*mn<:',
    'max': ':*::"::?:>:*::<:',
}


def replA(string):
    for char in {'"', '*', ':', '<', '>', '?', '|'}:
        string = string.replace(char, '_')
    return string


def replB(string):
    for char in {'"', '*', ':', '<', '>', '?', '|'}:
        if char in string:
            string = string.replace(char, '_')
    return string


trans = str.maketrans({'"': '_', '*': '_', ':': '_', '<': '_', '>': '_', '?': '_', '|': '_'})


def replC(string):
    return string.translate(trans)


for k, v in s_occurences.items():
    print(k)
    print("A: %f" % timeit.timeit(stmt=lambda: replA(s_occurences[k]), number=1000000))
    print("B: %f" % timeit.timeit(stmt=lambda: replB(s_occurences[k]), number=1000000))
    print("C: %f" % timeit.timeit(stmt=lambda: replC(s_occurences[k]), number=1000000))
```

```
0
A: 0.376292
B: 0.255101
C: 0.672888
2
A: 0.441930
B: 0.375825
C: 0.630128
4
A: 0.476545
B: 0.476716
C: 0.584534
6
A: 0.500308
B: 0.598776
C: 0.557867
8
A: 0.510370
B: 0.588787
C: 0.484488
max
A: 0.513036
B: 0.590188
C: 0.236326
```

That's tricky, because performance depends a lot on number of occurrences of the searched characters in the original string. The `translate` approach beats everything if the source string is made only of the searched characters (which is unlikely), but else is twice slower.

The extra test is improving performance when the string doesn't have much characters to replace but when there's a lot it is a tad slower.

phw: Given the numbers I think we could go with B then. That we have no or only a few of those characters seems to be the most common scenario for our use case. It is a bit counterintuitive why this is faster, but if it is like this it makes sense to use it.

That `str.translate` is so slow is disappointing, and your timing does not even include the `maketrans` call (which in the current architecture would need to be done each time as well, but that probably could be refactored).

zas (approved):
LGTM

PR #2178, opened 2022-12-05
PICARD-2596: Apply show_menu_icons only when options get saved

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2596

Toggling "Show icons in menus" in Options > User Interface is applied immediately, without saving. When canceling the options the UI displays differently then what was setup.

Instead this change should only be applied when saving the options.

# Solution

Apply show_menu_icons only when options get saved


zas (approved):
LGTM

PR #2179, opened 2022-12-05
PICARD-2576: Moved toolbar customization to separate options page

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2576

The option dialog is constrained in the minimum height. This grew significantly since Picard 2.7. The main culprit had been the file naming options page and the user interface options, both gained additional options in recent releases. The result is that the option window become cumbersome to use or in some cases completely unusable on screens with lower pixel height.

Refactoring of the file naming options already happened in #2177. This PR now tackles the user interface options.

# Solution

Move the action toolbar configuration to a separate option page.

This removes complexity from the general "User Interface" page and allows the options dialog to be resized to a smaller height, while allowing a more comfortable toolbar customization even on larger screens. This makes the options usable again on screens with only 768px height available.

Here is a screenshot of this dialog on Linux running with XFCE and a screen height of only 720px:

picard-xfce-1280x720


phw:
For comparison here is a screenshot of the same dialog in Picard 2.8.4 on the same system:

grafik

rdswift (approved):
I like it.  Allows shrinking of the Options window significantly.  :+1:

zas (approved):
Good move, LGTM

PR #2181, opened 2022-12-05
PICARD-2595: Customizable replacement for directory separators

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2595

Slashes / backslashes, which are used as directory separators on supported operating systems and also are used as directory separator in tagger script, get currently replaced by underscores in tags when generating the file name from tags. This is hard coded.

A common request by users is to be able to customize this character. There is currently no proper way to do this. A workaround is using a tagger script to manipulate the tags directly or set variables that are then used in the naming script. But this workaround does not consider manual changes to tags or changes done after initial loading.

# Solution

Add an option to File Naming > Compatibility that lets the user choose the character used as a replacement for directory separators.


rdswift (approved):
Looks good and seems to work fine for me.

zas (approved):
LGTM

PR #2185, opened 2023-01-03
PICARD-1945: Add option setting for number of threads to use for fpcalc

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add an option setting for the number of fpcalc processing threads allowed.

# Problem

When fingerprinting files on a CD Picard will fork a fpcalc process for each currently selected track to run in parallel.  On a CD drive this will cause the drive head to seek between the multiple tracks slowing down the fingerprinting process significantly.  It would be nice to have an option to specify the maximum number of fpcalc processes to run simultaneously or an option if fpcalc is being run on a CD drive to only fork one process at a time.

* JIRA ticket (_optional_): PICARD-1945

# Solution

Add an option setting to the **Fingerprinting** section to allow the user to adjust the maximum number of threads available to use for the fpcalc processing.  Minimum is 1 thread, maximum is 9 threads, default is 2 threads.

# Action

Update documentation if the change is accepted.


picard/acoustid/__init__.py
```diff
@@ -82,6 +82,10 @@ def init(self):
     def done(self):
         pass
 
+    def get_max_processes(self):
+        config = get_config()
+        return config.setting['fpcalc_threads'] or 2
```
zas, line 87: We should have this `2` as a constant: `DEFAULT_FPCALC_THREADS = 2` and use it also for default config

zas (changes requested):

rdswift:
Any thoughts on how to fix the Codacy error?  It's saying something about too many runs in the SARIF file (16 > 15), but I have no idea where these limits are set, or how I've managed to exceed them with this small change.  Thanks.

jesus2099:
It's so sweet to do all this but I hope it does not clutter the settings panel.

And when Picard is reading the (virtual) files of CD tracks to display their duration and format in left hand panel, it also reads them 2 by 2, or more.

jesus2099:
And I must disclose that I was also looking at [other ways to submit AcoustID from CD](https://community.metabrainz.org/t/easy-cd-acoustid-submit-without-ripping/616860?u=jesus2099) (but far from started).

Sophist-UK:
@jesus2099 Your workflow for acoustids is non-typical. Here are a few ideas that might help:

a. Run Picard from source and change the threadpool size to 1. This should serialise the access to the CD drive.

b. Use a different workflow:

    1. Copy CD to a temp directory
    2. Tag the files in the temp directory with Picard
    3. Submit the acoustids
    4. Delete the files.

phw (approved):
Looks good, thanks a lot.

phw:
> And when Picard is reading the (virtual) files of CD tracks to display their duration and format in left hand panel, it also reads them 2 by 2, or more.

This is a completely different thing and not part of what is being addressed here in this PR. The general number of loading threads is part of the discussion in #2172. As I wrote there we will very likely reduce the number of parallel loading threads for performance reasons. We might also add a config option, but that needs to be seen.

> I hope it does not clutter the settings panel.

As long as this is the only such setting we keep it in the fingerprinting options, I think that's fine. If we would get more such settings related to configure parallelism I think we would move them somewhere into the Advanced section.

phw:
> Any thoughts on how to fix the Codacy error? It's saying something about too many runs in the SARIF file (16 > 15), but I have no idea where these limits are set, or how I've managed to exceed them with this small change.

Generally the code coverage file submitted exceeds some limits. We already had this recently and it just resolved itself later on. I think it is something Codacy related.

I'll just ignore it here for now.

zas (approved):
LGTM

PR #2188, opened 2023-01-07
PICARD-2606: Add writer and lyricist sort names as variables

kellnerd:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
  Adds two new variables `_writersort` and `_lyricistsort`.

# Problem

* Picard only provides the sort names of composers but not of lyricists and writers, probably because only `composersort` has a specified tag mapping for various tag formats.
* JIRA ticket (_optional_): PICARD-2606

# Solution

Instead of defining non-standard tag mappings for `writersort` and `lyricistsort`, this PR only provides them as hidden variables, which can still be mapped to custom tags by the users or be used in filenaming scripts.

# Action

In my original implementation as a Picard plugin, I had also [defined display names](https://github.com/kellnerd/picard-plugins/blob/0ea52b5a55658f6e0ed9c1f7f72799fde4375439/plugins/additional_sort_names/additional_sort_names.py#L50-L54) for the internal tag names which are most likely to be used.
It certainly looks better in the tag view if one maps the hidden sort name variables to tags, but I'm not sure if the `TAG_NAMES` mapping is used for anything else but display purposes. I could easily add these two tag names for future use, so just let me know what you think.

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2017-2022 Laurent Monin
 # Copyright (C) 2018-2022 Philipp Wolfer
 # Copyright (C) 2019 Michael Wiencek
-# Copyright (C) 2020 David Kellner
+# Copyright (C) 2020, 2023 David Kellner
 # Copyright (C) 2020 dukeyin
 # Copyright (C) 2021 Bob Swift
 # Copyright (C) 2021 Vladislav Karbovskii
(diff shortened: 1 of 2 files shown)


phw (approved):
Thanks a lot. It really makes sense to me to have these variables

rdswift:
@kellnerd can you please have a look at https://github.com/metabrainz/picard-docs/pull/183 to make sure that I've correctly captured the descriptions for the new variables in the documentation?  Thanks.

kellnerd:
I've just reworded the title of the commit and the PR to correctly capitalize the ticket's name. Now the automation in JIRA has recognized it.

zas (approved):

phw (approved):

PR #2189, opened 2023-01-07
Debugging Codacy analysis CI errors

phw:
Trying to handle

```
Error: Code Scanning could not process the submitted SARIF file:
rejecting SARIF, as there are more runs than allowed (16 > 15)
```

See https://github.com/github/codeql-action/issues/1245

### DIFF
--- a/.github/workflows/codacy-analysis.yml
+++ b/.github/workflows/codacy-analysis.yml
@@ -42,8 +42,14 @@ jobs:
           # This will handover control about PR rejection to the GitHub side
           max-allowed-issues: 2147483647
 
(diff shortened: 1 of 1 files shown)


phw:
So all the tests on codacy side create more "runs" then are allowed for submission to codeql. It's a bit odd, because each tool provided by codacy generates a "run" entry in the .sarif files. But these even include tools we have disabled in the codacy project config.

I asked about this in https://github.com/codacy/codacy-analysis-cli-action/issues/95

phw:
Seems to work again

PR #2190, opened 2023-01-08
PICARD-2609: Threaded file update

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

PICARD-2609

As observed in https://github.com/metabrainz/picard/pull/2172 loading a lot of files into Picard can lead to the UI stuttering a lot and becoming unresponsive during processing of the files.

# Solution 
Doing some analysis I found that the metadata comparison in `file.update`  has a major impact here.

Doing this part inside a thread hugely impacts the responsiveness of the UI in my tests. Loading my 4.4k test files the UI kept updating and responding to user input.

This patch also removes an additional call to `file.update` directly after loading. This is not necessary here as an update will happen once the file gets moved to the proper parent. 

As far as I can tell making the update here async should not have any negative impact on the functionality. I found no code that relies on the similarity being synchronously updated. Instead the calls just happen to trigger the UI update.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -275,7 +275,6 @@ class File(QtCore.QObject, Item):
             if fingerprints:
                 self.set_acoustid_fingerprint(fingerprints[0])
         run_file_post_load_processors(self)
-        self.update()
         callback(self)
 
     def _copy_loaded_metadata(self, metadata, postprocessors=None):
@@ -683,9 +682,9 @@ class File(QtCore.QObject, Item):
         for name in set(self.metadata) | set(self.orig_metadata):
             if name.startswith('~'):
                 continue
-            if not self.supports_tag(name):
-                continue
             if name in ignored_tags:
+                continue
+            if not self.supports_tag(name):
                 continue
             yield name
 
--- a/picard/file.py
+++ b/picard/file.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2004 Robert Kaye
 # Copyright (C) 2006-2009, 2011-2013, 2017 Lukáš Lalinský
-# Copyright (C) 2007-2011, 2015, 2018-2022 Philipp Wolfer
+# Copyright (C) 2007-2011, 2015, 2018-2023 Philipp Wolfer
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008-2009 Nikolai Prokoschenko
 # Copyright (C) 2009 Carlin Mangar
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -683,13 +682,19 @@ def _tags_to_update(self, ignored_tags):
...
                 continue
+            if not self.supports_tag(name):
+                continue
             yield name
 
     def update(self, signal=True):
+        thread.run_task(
+            self._compare_metadata,
+            partial(self._compare_metadata_finished, signal)
+        )
```
rdswift, line 695: Is using a separate (unspecified) thread pool here going to cause some commands to show as being completed prematurely in the testing in `picard/tagger.py` Lines 404 - 434?

Sophist-UK: We need to be careful that we do not run more than one CPU intensive thread at a time. We should have a separate thread pool for CPU intensive workload with a single thread in it.

Sophist-UK (commented):
I am unsure whether the jitter reported by @Gabrielcarvfer in #2172 is due to not using a threadpool here, or because the hot cache is triggering different code which also needs to be put on a CPU intensive worker thread.

phw:
> I am unsure whether the jitter reported by @Gabrielcarvfer in https://github.com/metabrainz/picard/pull/2172 is due to not using a threadpool here, 

run_task is always using a thread pool. It's `tagger.thread_pool` in this case

> or because the hot cache is triggering different code which also needs to be put on a CPU intensive worker thread.

Yes, exactly that. It's indicated by the fact that he noticed it especially at the faster IO again. These stutters become more visible then.

picard/file.py
phw: @rdswift As I understand it, no. These updates ran after file loading anyway, when the file gets moved to a cluster or release.

@Sophist-UK I moved the `file.update` to use `tagger.priority_thread_pool` instead (previously only used by the metadata view). That is limited to one thread. And I think it also fits with the purpose. Essentially these are updates that are supposed to update the UI, and they should happen in time.

This also solves the issue that these updates must not be put behind all the loading threads, otherwise you just end with a longer UI stutter once all loading has ended. I first attempted to solve this by setting higher thread pool priority for `file.update`, but that lead to much more stuttering.  Having these update calls be run sequentially indeed improved responsiveness for me again.

rdswift: It definitely shouldn't be a problem now that you're running the task in the `tagger.priority_thread_pool`.  Thanks.

rdswift (approved):
It looks fine to me.  My biggest concern was if it would cause commands processed from the command queue to be shown as complete prematurely.  This has been addressed by running the tasks in `tagger.priority_thread_pool`.

Gabrielcarvfer:
> I must say I'm really happy with these changes. I used Picard with this patch the last few days, both for tagging and for testing. It is a huge improvement on my system. I could not find any downsides so far.
> 
> So let's see to get this merged.

It really is butter smooth with the latest changes. :D

zas (approved):
LGTM

PR #2192, opened 2023-01-20
PICARD-2615: Picard does not check for unsaved files when using the QUIT command

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Have `QUIT` command do unsaved files check, with option to bypass the check.

# Problem

When the 'QUIT' command is executed, Picard exits without performing the unsaved files check.

* JIRA ticket (_optional_): PICARD-2615

# Solution

Update the `QUIT` command processing to perform the unsaved files check, allowing the command to be cancelled by the user.  Modify the command to allow an optional argument to force Picard to bypass the unsaved files check.

# Action

Update documentation.

### DIFF
--- a/picard/util/remotecommands.py
+++ b/picard/util/remotecommands.py
@@ -64,7 +64,7 @@ REMOTE_COMMANDS = {
     "LOOKUP_CD": RemoteCommand(
         "handle_command_lookup_cd",
         help_text="Read CD from the selected drive and lookup on MusicBrainz. "
-        "Without argument, it defaults to the first (alphabetically) available disc drive",
+        "Without argument, it defaults to the first (alphabetically) available disc drive.",
         help_args="[device/log file]",
     ),
     "PAUSE": RemoteCommand(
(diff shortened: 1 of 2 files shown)


picard/util/remotecommands.py
```diff
@@ -64,7 +64,7 @@ def __init__(self, method_name, help_text=None, help_args=None):
     "LOOKUP_CD": RemoteCommand(
         "handle_command_lookup_cd",
         help_text="Read CD from the selected drive and lookup on MusicBrainz. "
-        "Without argument, it defaults to the first (alphabetically) available disc drive",
+        "Without argument, it defaults to the first (alphabetically) available disc drive.",
```
zas, line 67: This doesn't look like related to PICARD-2615.

zas (changes requested):

picard/util/remotecommands.py
rdswift: Strictly speaking it isn't, but I thought I would include the change here rather than create a separate PR just to add a period to one description to make it consistent with all the others.  I'll remove it from this PR and create a new one just for addinding the period.

rdswift: I just added https://github.com/metabrainz/picard/pull/2193 to cover the change to the `LOOKUP_CD` command description.

zas (approved):

phw (approved):

PR #2197, opened 2023-03-03
PICARD-2629: Add keyboard shortcut for "Search for similar albums"

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2629

There is currently no keyboard shortcut for "Search for similar albums", even though there is one for "Search for similar tracks" (`Ctrl+T`)

# Solution

Add a new action "Search for similar items" that performs either track or album search depending on context. This works as both actions are only selected for single selection of the corresponding type.

This action is triggered by `Ctrl+T`, so same shortcut for both. The action is also available in Options > Interface > Toolbars and hence can be added to the toolbar.

I kept the original actions for track and album search for use in the context menus. That way it gives more context what is being searched for.


zas (approved):
LGTM

PR #2198, opened 2023-03-13
PICARD-2620: Added new icon for macOS

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2620

# Solution

Provide a new icon for macOS that better integrates into the macOS icon guidelines.


phw:
Here is a screenshot of the new icon (on the right, the left one is the old):

Screen Shot 2023-03-14 at 07 55 32

I could only take the screenshot on macOS 10.12, hence the other icons have different shapes. But on more recent macOS versions the icons of the Apple apps and more and more third-party apps all follow this style with the round cornered shape.

PR #2200, opened 2023-03-16
PICARD-2121 Coverart processing

Pranay-Pandey:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Add Cover Art Post Processing plugin**

# Problem
Picard should allow to post process cover art images provided by the providers. 
* Ignore images smaller than specified width and height
* If image is larger than max. width and height, resize it to max. width and height

* JIRA ticket : PICARD-2121

# Solution

CoverArt Post Processing Plugin.
Plugin extracts the coverArt Images from the metadata of the tracks. 
* If the dimensions of the image is less than a minimum limit, then the image is removed from the metadata. 
* If the dimensions of the image are larger than the specified size then, the image is resized to the set maximum.

# Action
The plugin requires the Pillow library to be installed as it is necessary to resize the image.

### DIFF
--- a/picard/plugins/coverart_post_processing.py
+++ b/picard/plugins/coverart_post_processing.py
@@ -1,0 +1,47 @@
+PLUGIN_NAME = "Cover Art Post Processing"
+PLUGIN_AUTHOR = "Pranay"
+PLUGIN_DESCRIPTION = """
+Post Processing Features for the CoverArt Images.
+Ignore images smaller than specified width and height.
+"""
+PLUGIN_VERSION = '0.1'
+PLUGIN_API_VERSIONS = ['2.2']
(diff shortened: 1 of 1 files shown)


rdswift (commented):
Shouldn't this be in the https://github.com/metabrainz/picard-plugins repository instead of being included in the Picard base code?

Sophist-UK:
I agree with Bob / @rdswift that this should probably be in the Picard Plugins repository instead.

Pranay-Pandey:
Thank you for your review. I will make the necessary changes and resubmit the pull request to the plugin repository. I intend to integrate this feature into the code base in the future.

PR #2201, opened 2023-03-17
PICARD-2625: Improve file naming script editing and new script creation

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Allow editing of all scripts and allow selecting template for creation of new file naming scripts.

# Problem

We have received multiple reports by users struggling to edit the file naming scripts. Those users open the script editor and find themselves unable to edit the script. The reason is that by default they are on one of the Picard presets, which is not editable. The user is expected to either create a new custom script (either an empty one from scratch or as a copy of the preset) and then edit this new custom script.

This is not obvious, and users fail to understand this mechanic. It should be more discoverable how to make a edited custom script from a preset.

* JIRA ticket (_optional_): PICARD-2625

# Solution

Make all scripts editable as "User Scripts" and do away with the read-only "System Scripts" entirely.

The process automatically converts the system scripts to user scripts for existing installations, and create the (currently three) preset scripts as user scripts for new installations (or when the user chooses to reset all option settings).  The current system preset scripts are kept as "Naming Script Templates" which could optionally be selected by the user as a starting point when creating a new script as an option rather than just starting with a blank script.  This should simplify the user interface, because there would only be one type of script and all scripts would be editable (as one would expect in the "File Naming Script Editor").

The current functionality of adding, deleting, copying, importing and exporting scripts would be retained, with the only difference being a sub-menu for selecting the template to use (i.e. blank or one of the standard script templates) when creating a new script.

# Action

### DIFF
--- a/picard/ui/mainwindow.py
+++ b/picard/ui/mainwindow.py
@@ -25,7 +25,7 @@
 # Copyright (C) 2018 Kartik Ohri
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2018 virusMac
-# Copyright (C) 2018, 2021 Bob Swift
+# Copyright (C) 2018, 2021, 2023 Bob Swift
 # Copyright (C) 2019 Timur Enikeev
 # Copyright (C) 2020-2021 Gabriel Ferreira
 # Copyright (C) 2021 Petit Minion
@@ -132,7 +132,6 @@ from picard.ui.passworddialog import (
 from picard.ui.scripteditor import (
     ScriptEditorDialog,
     ScriptEditorExamples,
-    user_script_title,
 )
 from picard.ui.searchdialog.album import AlbumSearchDialog
 from picard.ui.searchdialog.track import TrackSearchDialog
@@ -208,6 +207,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self.script_editor_dialog = None
         self.examples = None
 
+        self.check_and_repair_naming_scripts()
         self.check_and_repair_profiles()
 
         self.setupUi()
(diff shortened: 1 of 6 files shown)


zas (dismissed):
Good work, nice improvement.

Edit: I noticed a duplication of scripts, but I fail to reproduce it, basically a second script with same name and same content was appearing.

rdswift:
> I noticed a duplication of scripts, but I fail to reproduce it, basically a second script with same name and same content was appearing.

I found a way to reproduce it.  Have now added a check so that script names cannot be duplicated.

zas (dismissed):
LGTM

picard/ui/mainwindow.py
```diff
@@ -1787,6 +1787,23 @@ def check_for_update(self, show_always):
...
+        script_id_key = 'selected_file_naming_script_id'
+        if not config.setting[script_key]:
+            scripts = {}
+            for script in get_file_naming_script_presets():
+                scripts[script['id']] = script.to_dict()
+            config.setting[script_key] = scripts
+        naming_script_ids = list(config.setting[script_key].keys())
+        if config.setting[script_id_key] not in naming_script_ids:
+            config.setting[script_id_key] = naming_script_ids[0]
+
```
zas, line 1806: See https://github.com/rdswift/picard/pull/15

zas (changes requested):

picard/ui/mainwindow.py
rdswift: Merged.  Thanks.

zas (approved):

phw (approved):

PR #2202, opened 2023-04-17
PICARD-2634: Use standardized artist names by default

zas:
# Summary

From ticket:

I think the option "Use standardized artist names" should be turned on by default. If the option is turned off, then organizing audio files into artist folders has unexpected results. As an example, you typically wouldn't want to have both "The Prodigy" and "Prodigy" as two different folders for the same artist, just because some of the releases of [this release group](https://musicbrainz.org/release-group/ac9138ce-6331-3f8d-86d7-69f13e4ab4f4) happen to be credited to "Prodigy" (and not to "The Prodigy").

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

* JIRA ticket (_optional_): PICARD-2634

# Solution

Switch option default from False to True.
It only changes the default value (from False to True), so it has no impact on existing configurations.

# Action

### DIFF
--- a/picard/ui/options/metadata.py
+++ b/picard/ui/options/metadata.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2008-2009, 2018-2022 Philipp Wolfer
 # Copyright (C) 2011 Johannes Weißl
 # Copyright (C) 2011-2013 Michael Wiencek
-# Copyright (C) 2013, 2018, 2020-2021 Laurent Monin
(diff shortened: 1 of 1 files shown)


rdswift (approved):
A pretty straightforward change.  Looks fine to me.

phw (approved):
I agree, this makes sense

PR #2206, opened 2023-04-22
id3v2.4 TDRL -> releasedate mapping

certuna:
# Summary
* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Picard currently cannot read or write the official spec Release Time (`TDRL` frame) in id3v2.4 due to a missing mapping of `TDRL` to the internal variable `releasedate` (analogous to `TDRC` -> `date` and `TDOR` -> `originaldate`).

This PR fixes it and provides a gracious fallback + upgrade path for id3v2.3.

# Problem

1. When Picard is instructed to write the variable `releasedate` to id3v2.4, it currently writes the frame `TXXX:releasedate` and not `TDRL`. While this is compliant with id3v2.3 (which does not have the `TDRL` frame defined), this is not compliant with id3v2.4, where it is part of the official standard.

2. With Vorbis Comments, the `RELEASEDATE` frame is written correctly, albeit without date sanitation.

Most other tag editors (mp3tag, Yate, kid3, Ex Falso, ffmpeg, TagLib, etc) have a mapping `releasedate` -> `TDRL`, Picard is the only modern tag editor I'm aware of that is not able to read or write this frame.

* JIRA ticket (_optional_): PICARD-2616

# Solution

I have deliberately not touched any of the default MusicBrainz JSON -> Picard tag mappings, as I understand it is a deliberate decision that Picard does not write anything to the `TDRL` frame by default. However, this mapping at least makes the `releasedate` internal variable accessible to scripts that do.

- added a `TDRL` -> `releasedate` mapping for id3v2.4
- added a `TXXX:releasedate` custom frame fallback for id3v2.3 + upgrade path for v2.3 -> v2.4
- added date sanitation of the `releasedate` field for Vorbis Comments, like Picard already does with `originaldate` and `date`

Note: id3v2.3 fallback is done in the same way Picard currently handles Mood (`TMOO`) which is another frame that exists in v2.4 but not v2.3)

# Action

Please review & comment

FYI: this test script that writes a MusicBrainz release date to the TDRL can be useful for testing:
`$set(releasedate,%date%)`

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -110,6 +110,7 @@ class ID3File(File):
         'TXXX:ALBUMARTISTSORT': 'TSO2',
         'TXXX:COMPOSERSORT': 'TSOC',
         'TXXX:mood': 'TMOO',
+        'TXXX:releasedate': 'TDRL',
     }
 
     __translate = {
@@ -506,6 +507,8 @@ class ID3File(File):
                             tags.add(self.build_TXXX(encoding, 'mood', values))
                     # No need to care about the TMOO tag being added again as it is
                     # automatically deleted by Mutagen if id2v23 is selected
+                        if frameid == 'TDRL':
+                            tags.add(self.build_TXXX(encoding, 'releasedate', values))
                     tags.add(getattr(id3, frameid)(encoding=encoding, text=values))
                     if frameid == 'TSOA':
                         tags.delall('XSOA')
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -137,6 +137,7 @@ class ID3File(File):
         'TPUB': 'label',
         'TDOR': 'originaldate',
         'TDRC': 'date',
+        'TDRL': 'releasedate',
         'TSSE': 'encodersettings',
         'TSOA': 'albumsort',
         'TSOP': 'artistsort',
--- a/picard/formats/vorbis.py
+++ b/picard/formats/vorbis.py
@@ -139,7 +139,7 @@ class VCommentFile(File):
             for value in values:
                 value = value.rstrip('\0')
                 name = origname
-                if name == "date" or name == "originaldate":
+                if name == "date" or name == "originaldate" or name == "releasedate":
                     # YYYY-00-00 => YYYY
                     value = sanitize_date(value)
                 elif name == 'performer' or name == 'comment':
@@ -277,7 +277,7 @@ class VCommentFile(File):
                 continue
             elif name.startswith('lyrics:'):
                 name = 'lyrics'
-            elif name == "date" or name == "originaldate":
+            elif name == "date" or name == "originaldate" or name == "releasedate":
                 # YYYY-00-00 => YYYY
                 value = sanitize_date(value)
             elif name.startswith('performer:') or name.startswith('comment:'):
--- a/picard/ui/edittagdialog.py
+++ b/picard/ui/edittagdialog.py
@@ -64,7 +64,7 @@ class TagEditorDelegate(QtWidgets.QItemDelegate):
         else:
             editor = super().createEditor(parent, option, index)
         completer = None
-        if tag in {'date', 'originaldate'}:
+        if tag in {'date', 'originaldate', 'releasedate'}:
             editor.setPlaceholderText(_('YYYY-MM-DD'))
         elif tag == 'originalyear':
             editor.setPlaceholderText(_('YYYY'))
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -173,6 +173,7 @@ class MainPanel(QtWidgets.QSplitter):
         (N_('Fingerprint status'), '~fingerprint'),
         (N_('Date'), 'date'),
         (N_('Original Release Date'), 'originaldate'),
+        (N_('Release Date'), 'releasedate'),
         (N_('Cover'), 'covercount'),
     ]
 
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -103,6 +103,7 @@ TAG_NAMES = {
     'r128_track_gain': N_('R128 Track Gain'),
     '~rating': N_('Rating'),
     'releasecountry': N_('Release Country'),
+    'releasedate': N_('Release Date'),
     'releasestatus': N_('Release Status'),
     'releasetype': N_('Release Type'),
     'remixer': N_('Remixer'),


picard/formats/vorbis.py
```diff
@@ -139,7 +139,7 @@ def _load(self, filename):
             for value in values:
                 value = value.rstrip('\0')
                 name = origname
-                if name == "date" or name == "originaldate":
+                if name == "date" or name == "originaldate" or name == "releasedate":
```
rdswift, line 142: It might be cleaner if you used a set here similar to your change in `edittagdialog.py`.  Something like:

```python
if name in {'date', 'originaldate', releasedate'}:
```

Same comment applies to your change in Line 280.

certuna: Done, thanks!

rdswift:
One more suggestion...  You might want to be a bit more descriptive in your commit messages.  I suggest having a look at articles like https://initialcommit.com/blog/git-commit-messages-best-practices and https://www.freecodecamp.org/news/how-to-write-better-git-commit-messages/ for some tips.

picard/formats/id3.py
```diff
@@ -505,6 +507,8 @@ def _save(self, filename, metadata):
                             tags.add(self.build_TXXX(encoding, 'mood', values))
                     # No need to care about the TMOO tag being added again as it is
                     # automatically deleted by Mutagen if id2v23 is selected
+                        if frameid == 'TDRL':
```
zas, line 510: `TDRL` is specified in ID3v2.4, not  ID3v2.3.
So I guess we should only do that `if config.setting['write_id3v24']:`

zas (changes requested):
Tests should be added for this.

picard/formats/id3.py
phw: This is already part of the `if config.setting['write_id3v23']` block, same handling as `TMOO`. I think it is fine as it is.

phw (changes requested):
Thanks a lot for tackling this. Before we merge this we should have a look over how other tools handle `TDRL`  and how it maps to other tagging format, especially MP4 and ASF (which had not yet been considered in this PR). https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#audio-metadata-specifications provides some resources for tag specs and mappings in other tools.

The current implementation will not write this tag to ASF, for MP4 it would be treated as a custom tag `----:com.apple.iTunes:releasedate`.

So a quick look over what we have here:

| Software                        | ID3v2.3             | Vorbis         | MP4                                 | ASF | Ape            |
|---------------------------------|---------------------|----------------|-------------------------------------|-----|----------------|
| MP3Tag                          | `TDRL`              | `RELEASETIME`? | -                                   | -   | -              |
| Kid3                            | -                   | `RELEASEDATE`? | `----:com.apple.iTunes:RELEASEDATE` | -   | -              |
| Yate                            | `TXXX:Release Time` | `RELEASETIME`  | `----:com.apple.iTunes:RELEASETIME` | -   | `Release Time` |
| JaudioTagger (Jaikoz, SongKong) | -                   | -              | -                                   | -   | -              |
| MediaMonkey                     | -                   | -              | -                                   | -   | -              |
| MusicBee                        | -                   | -              | -                                   | -   | -              |

Notes;
- JaudioTagger, MediaMonkey, MusicBee do not support TDRL (according to the docs)
- MP3Tag: Vorbis is not specified, but AFAIK it ends up with internal name `RELEASETIME`
- Kid3: MP4 uses uppercase `----:com.apple.iTunes:releasedate`
- Kodi treats `TDRL` just as another source for its DATE, just like `TDRC`. Unsure which it prefers.
- foobar2000: Supports `TDRL`. Internal name is `RELEASE DATE`. Unsure how it handles it for other formats

So this is really a mixed bag (and as a side note is one of the reasons this has not yet been supported by Picard). Anyway, we should just choose. I think the current approach is overall fine. Not supporting ASF seems to be widely the standard way of handling it (and ASF has lost relevance anyway, I can't remember the last time any user had some question about that).

Maybe explicitly add it to the mapping for MP4 as uppercases (`----:com.apple.iTunes:RELEASEDATE`). What do you think?

Sophist-UK:
> Before we merge this we should have a look over how other tools handle `TDRL` and how it maps to other tagging format, especially MP4 and ASF (which had not yet been considered in this PR).

1. It is not just tagging tools that we need to research, but also players - it is equally important to know what players support what tags.

2. In the absence of de-jure or de-facto standard, I think we should be willing to think about creating new de-facto standards. If there is no prior use at all, then we can go ahead and do whatever we consider is sensible - if there is prior use but not a de-facto standard e.g. in one tagger / player, then we should consider whether it is reasonable to stay compatible, or whether there is a good reason to set a different de-facto standard. (IMO we should not avoid doing something new just because no one else has done it yet and we are afraid of doing something and then finding someone else does something different.)
3.

picard/formats/id3.py
zas: Do I miss something? But writing ID3v2.4 tag when `write_id3v23` doesn't look correct to me. What's the purpose of `write_id3v24` then?

phw: In case of v2.3 the code writes a TXXX tag(with `self.build_TXXX`) instead of the TDRL. The TDRL  itself will, if present, get removed by mutagen when downgrading to v2.3.

certuna:
A lot of players these days just use boilerplate libraries like TagLib and mutagen, which both map `releasedate` to `TDRL` (id3v2.4) and `RELEASEDATE` (Vorbis). Kodi, Navidrome and Jellyfin for example all use TagLib.

phw:
I think we are essentially good with this approach. The TDRL tag is nod widely supported, but I agree with the approach taken here. What I'd wish to see is adding an explicit support for `----:com.apple.iTunes:RELEASEDATE` to MP4 in `__r_freeform_tags_ci` (https://github.com/metabrainz/picard/blob/master/picard/formats/mp4.py#L159). Then we would be close how Kid3 supports it, and also be able to handle both lower and uppercase tag name there.

> A lot of players these days just use boilerplate libraries like TagLib and mutagen,

To be fair just because these libraries are used does not necessarily mean the player has support for this tag. It's still the decision of the developers which tags and how they support. And some tools are very open and allow flexible access to a lot of tags, and others are very limited and only support the essentials. But I get your point.

> If you think the id3v2.3 fallback frame TXXX:releasedate is not 'de facto' enough, happy to remove it - then the default behaviour of mutagen kicks in and it would automatically never write anything to v2.3, only to v2.4.

I think it is ok to write it. I just wonder if we should maybe uppercase this. So Picard's original approach was to write TXXX frames as title case. Yate also does this in this case with `TXXX:Release Time`. The reason was because a lot of software did it that way and it seemed to be the format typical approach to name these tags.

But in the last years newer tags had usually been all uppercase, essentially unifying the naming between Vorbis and ID3 TXXX. This makes handling custom tags much easier, as flexible players like foobar2000 can handle these tags independent of what format is being used and without the player needing actual knowledge of the tags. Especially from the foobar2000 side Picard receives frequent criticism for how it handles tags.

An alternative approach would be to use `TDRL` anyway, as e.g. MP3Tag chose to handle it. There is precedence for this (https://github.com/metabrainz/picard/blob/master/picard/formats/mutagenext/compatid3.py#L76-L77), but usually we did this for tags were there was a wider support for this deviating behavior.

> As these are tiny (one-line) commits I thought the code would more or less speak for itself, and this PR discussion would provide the required context - but happy to be more verbose in future commits. Unfortunately I cannot edit past commits I think?

Actually I think in this case we will just squash them on merge into a single commit, it's only a few lines of changes that belong together. So for me the commit history is currently fine as it is. But otherwise I would indeed prefer more descriptive commit messages.

rdswift:
Just adding a note here so that we don't forget...  Once this is all sorted out and merged, we'll need to update the tag mapping appendix information (and spreadsheet) in the Picard documentation.  This should be done in the https://github.com/metabrainz/picard-docs/blob/master/tag_mapping.py file, which is used to automatically update all of the references.

zas:
> Unfortunately I cannot edit past commits I think?

You can, `git rebase -i`  is your friend, check out https://about.gitlab.com/blog/2020/11/23/keep-git-history-clean-with-interactive-rebase/ for examples.

That said, in this case, I think you can just squash all commits in one, and set a clean message for it.

certuna:
> What I'd wish to see is adding an explicit support for `----:com.apple.iTunes:RELEASEDATE` to MP4 in `__r_freeform_tags_ci` (https://github.com/metabrainz/picard/blob/master/picard/formats/mp4.py#L159). Then we would be close how Kid3 supports it, and also be able to handle both lower and uppercase tag name there.

Done.

> I think it is ok to write it. I just wonder if we should maybe uppercase this.

Happy to do this, but shouldn't this then also be done with `mood` (currently written in lowercase as `TXXX:mood` in id3v2.3?)

> Tests should be added for this.

@zas Happy do do this, I was a bit hesitant to just upload new test.mp3/mp4/flac files without discussion/consultation :)
Should I modify the existing `test.mp3`/etc files in `picard/test/data`, or do I create and upload fresh `test-releasedate23.mp3`, `test-releasedate24.mp3`, `test-releasedate.flac` and `test-releasedate.m4a` files?

phw:
> Happy to do this, but shouldn't this then also be done with `mood` (currently written in lowercase as `TXXX:mood` in id3v2.3?)

We usually are rather careful changing tags once they are implemented, unless there is a good reason to change it. Anyway, as part of this PR we should focus on releasedate.

> Should I modify the existing `test.mp3`/etc files in `picard/test/data`, or do I create and upload fresh `test-releasedate23.mp3`, `test-releasedate24.mp3`, `test-releasedate.flac` and `test-releasedate.m4a` files?

Most of the test cases use the existing files as basis to first create the tags in the state to test and then perform the tests. The first step for having basic test coverage is to add the `releasedate` with a value in https://github.com/metabrainz/picard/blob/master/test/formats/common.py#L104 . This verifies the format can write the tag and read it back.

Specific tests for ID3 can be added to https://github.com/metabrainz/picard/blob/master/test/formats/test_id3.py . E.g. to specifically test how the tag gets written to ID3 v2.3 vs v2.4 something like this could work (just some ad-hoc code written here, might contain mistakes):


```python
    class Id3TestCase(CommonTests.TagFormatsTestCase):
        
         # ....

        @skipUnlessTestfile
        def test_releasedate_v23(self):
            config.setting['write_id3v23'] = True
            metadata = Metadata({
                'releasedate': '2023-04-28',
            })
            save_metadata(self.filename, metadata)
            raw_metadata = load_raw(self.filename)
            self.assertEqual(metadata['releasedate', raw_metadata['TXXX:RELEASEDATE')
   
     @skipUnlessTestfile
        def test_releasedate_v24(self):
            config.setting['write_id3v23'] = False
            metadata = Metadata({
                'releasedate': '2023-04-28',
            })
            save_metadata(self.filename, metadata)
            raw_metadata = load_raw(self.filename)
            self.assertEqual(metadata['releasedate', raw_metadata['TDRL')
```

zas:
@certuna can you update the PR as requested?

certuna:
> @certuna can you update the PR as requested?

Yes I will have time this weekend to work on the tests.

I have added the uppercase mapping for MP4 as requested by @phw : https://github.com/metabrainz/picard/commit/38c739a4f89e3e0b58e9b44c6f38b3a5d55c9748

Are there any more changes needed?

phw (approved):
As this looks good code wise I'll going to merge it so we can include it in the 2.9 release. Thanks a lot for the contribution, much appreciated.

I'll add some tests after the merge. If you have anything to add feel free to open a new PR for it.

PR #2208, opened 2023-04-27
Refactor log string substitution

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

For log output Picard often does variable substitutions in the logging strings. In many cases these used the string formatting options provided by Python (all of % operator, `.format()`  and f-strings are used).

This means that the string substitutions are applied even if due to the log level the string never gets outputted. This is especially true for debug log messages in normal operation. This ads an unnecessary overhead.

# Solution
Rely on the string substitutions of the logging module to avoid unnecessary substitutions if log output never gets written.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -304,7 +304,7 @@ class Album(DataObject, Item):
                     parse_result = self._parse_release(document)
                     config = get_config()
                     if parse_result == ParseResult.MISSING_TRACK_RELS:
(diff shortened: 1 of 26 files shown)


zas (approved):
LGTM

PR #2209, opened 2023-04-28
PICARD-900: Use ID3v2.4 UTF-8 by default

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-900

Folks, it's time. It has been 6 years since this was attempted the last time in #559 . But I think in 2023 it is finally time to change Picard to default to ID3v2.4.

Since the last discussion Windows has gained ID3v2.4 support. All Windows version not supporting it are officially no longer supported (which includes older Windows 10 versions).

Also there are more issues discussed were ID3v2.3 is the problem, not the solution. In regards to user support it seems to me it is easier to tell those few people having issues with ID3v2.4 on older devices to change to v2.3 than it is to explain users why they should better use v2.4 and trying to argue why Picard still considers 2.3 the default.

One of the major issues is the lack of proper multi-value support in v2.3. This causes issues like https://community.metabrainz.org/t/certain-albums-not-scrobbling/630833/12

# Solution

Switch the default for new installs to ID3v2.4 UTF-8.

### DIFF
--- a/picard/ui/options/tags_compatibility_id3.py
+++ b/picard/ui/options/tags_compatibility_id3.py
@@ -49,8 +49,8 @@ class TagsCompatibilityID3OptionsPage(OptionsPage):
 
     options = [
         BoolOption("setting", "write_id3v1", True),
-        BoolOption("setting", "write_id3v23", True),
-        TextOption("setting", "id3v2_encoding", "utf-16"),
+        BoolOption("setting", "write_id3v23", False),
+        TextOption("setting", "id3v2_encoding", "utf-8"),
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM.

Yes, it's time :)

rdswift (approved):
I agree completely.  Looks fine to me.

phw:
So let's do this now. We should do the 2.9 beta release finally, then we can gather wider feedback on how this is going.

PR #2210, opened 2023-05-21
PICARD-2627: Ensure pipe listener threads closes on quit

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2627

It can happen that `Tagger.pipe_server` hangs waiting getting data from the pipe. If this happens when Picard is closing the `pipe_server` thread keeps running in the background until it reveives data, which likely won't happen.

# Solution
To fix this write a MESSAGE_TO_IGNORE to the pipe when closing. This causes the pipe read call to return and the loop inside `Tagger.pipe_server`  will stop properly.


zas (approved):
LGTM

PR #2211, opened 2023-05-21
PICARD-2645: Fix handle the quit confirmation when reveiving QUIT command

phw:
…mand

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2645

Sending the QUIT command to Picard with "picard -e QUIT" wrongly handles the quit confirmation dialog. If there are no files to save, the QUIT is ignored. If there are files to save the QUIT is only acknowledged if the user cancels, not confirms, the quit.

# Solution

Properly handle the return value of `show_quit_confirmation` (`True`  means do quit Picard, `False`  don't). Also refactor `Tagger.handle_command_quit` for readability.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -547,12 +547,13 @@ class Tagger(QtWidgets.QApplication):
             log.error("No command pause time specified.")
 
     def handle_command_quit(self, argstring):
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2213, opened 2023-05-28
PICARD-711: Add support for series variables

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-711

Adds support for series variables for release group, release, recording and work series. 

# Solution
Make the series name, id, disambiguation comment and number available as variables. The following variables are getting set:

- ~recording_series
- ~recording_series_id
- ~recording_seriescomment
- ~recording_seriesnumber
- ~release_series
- ~release_series_id
- ~release_seriescomment
- ~release_seriesnumber
- ~releasegroup_series
- ~releasegroup_series_id
- ~releasegroup_seriescomment
- ~releasegroup_seriesnumber
- ~work_series
- ~work_series_id
- ~work_seriescomment
- ~work_seriesnumber

The variable names are modeled after https://github.com/kellnerd/picard-plugins/blob/release_series/plugins/release_series/release_series.py

I thought about adding actual tags instead of just variables. But series are often rather arbitrary. E.g. https://musicbrainz.org/release/b84ee12a-09ef-421b-82de-0441a926375b is part of 8 series.

Having this as variables allows the users to decide if and how they make use of this information. I expect this to be used mostly by advanced users in any case.

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -609,15 +609,15 @@ class Album(DataObject, Item):
             inc |= {
                 'artist-rels',
                 'recording-rels',
+                'release-group-level-rels',
                 'release-rels',
                 'series-rels',
                 'url-rels',
-                'work-rels'
+                'work-rels',
             }
             if config.setting['track_ars']:
                 inc |= {
                     'recording-level-rels',
-                    'release-group-level-rels',
                     'work-level-rels',
                 }
         require_authentication = self.set_genre_inc_params(inc, config) or require_authentication
--- a/picard/album.py
+++ b/picard/album.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Hendrik van Antwerpen
 # Copyright (C) 2008 ojnkpjg
-# Copyright (C) 2008-2011, 2014, 2018-2022 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014, 2018-2023 Philipp Wolfer
 # Copyright (C) 2009 Nikolai Prokoschenko
 # Copyright (C) 2011-2012 Chad Wilson
 # Copyright (C) 2011-2013, 2019 Michael Wiencek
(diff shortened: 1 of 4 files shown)


phw:
Side note: There is another series related ticket [PICARD-802](https://tickets.metabrainz.org/browse/PICARD-802). This one is not directly related to the work being done here. Instead it is about using series for tagging. E.g. a recording series could generally be used similar to how releases are being loaded directly for tagging.

rdswift (approved):
Looks good to me.  Thanks for this.

picard/album.py
```diff
@@ -610,12 +610,14 @@ def load(self, priority=False, refresh=False):
...
                 'recording-rels',
                 'release-rels',
+                'series-rels',
                 'url-rels',
                 'work-rels'
             }
             if config.setting['track_ars']:
                 inc |= {
                     'recording-level-rels',
+                    'release-group-level-rels',
```
kellnerd, line 620: Hmm, I don't get why release group relationships should only be loaded when "Use track relationships" is enabled. Wouldn't it be more intuitive to add this `inc` parameter to the set which is enabled by "Use release relationships"?
(Would still not be 100% accurate as this is about release _group_ rels, but since "track" means recordings and works, it is also kind of intuitive to use "release" to also cover release groups.)

kellnerd (commented):
Great job, another one of my plugins becoming obolete, yay!

Had I known that it would be this simple to support loading series relationships for all these entity types _without additional requests_, I would not have bothered to write a plugin for release group series 😮
I thought release group series would require an additional request in any case, that's why I started with these. Since these were sufficient for  99% of my use cases, I never picked up the idea again to implement support for the other entity types in Picard.

picard/album.py
phw: Having this done with release ARs are enabled was actually my intention. I'll fix that.

phw:
> Had I known that it would be this simple to support loading series relationships for all these entity types without additional requests, I would not have bothered to write a plugin for release group series

@kellnerd Actually I had the same assumption that we would need at least one additional request, and hence thought it's best placed in a plugin. If I'd realized earlier that it was possible without additional request I would have done this change earlier :dagger: 

Anyway, I'm glad you did the plugin and did some preparations on what variables are needed / useful.

phw:
I addressed the comment by @kellnerd . Also noticed I forgot to include the tests in my push yesterday, now also added.

zas (approved):
LGTM

PR #2217, opened 2023-05-31
PICARD-2624: Use mutagen 1.46 with IFF file patch for macOS / Windows packages

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2624

Fixes writing IFF (WAVE / AIFF) files with truncated trailing null bytes.

# Solution

For the Windows / macOS packages include mutagen with the patch at https://github.com/quodlibet/mutagen/pull/517 applied. The branch being pulled from (https://github.com/phw/mutagen/tree/release-1.46-picard) contains the 1.46 release of mutagen with only that patch applied.

For the final release of Picard I hope there will be a new version of mutagen available we will upgrade to. But since this is an important fix it warrants some special handling.


zas (approved):
LGTM

PR #2220, opened 2023-06-02
Enforce keywords arguments only for WSRequest and drop WS*Request sub…

zas:
…classes

- using only keyword arguments will ease future changes
- order of arguments do not matter anymore
- this class isn't used directly for now, I don't think any plugin uses it
- subclasses aren't strictly needed
- since methods have a lot of parameters, making them obvious is much easier to read
- we can pass a dict (**kwargs)
- note the use of * in __init__() enforcing keyword args only.
- use few asserts as safety for "mandatory" arguments

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -144,14 +144,20 @@ class WSRequest(QNetworkRequest):
 
         # mandatory parameters
         self.method = method
-        assert self.method in {'GET', 'PUT', 'DELETE', 'POST'}
+        if self.method not in {'GET', 'PUT', 'DELETE', 'POST'}:
+            raise AssertionError('invalid method')
         self.host = host
-        assert self.host is not None
+        if self.host is None:
+            raise AssertionError('host undefined')
         self.port = int(port)
-        assert self.port >= 0
-
+        if self.port < 0:
+            raise AssertionError('port invalid')
         self.path = path
+        if self.path is None:
+            raise AssertionError('path undefined')
         self.handler = handler
+        if self.handler is None:
+            raise AssertionError('handler undefined')
 
         # optional parameter, must be set before calling build_qurl()
         self.queryargs = queryargs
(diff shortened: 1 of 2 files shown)


zas:
Actually there's one plugin using WSGetRequest : https://github.com/metabrainz/picard-plugins/blob/c03c7f1fdb2546d4e3467d99859406ce6cf0bc05/plugins/tangoinfo/__init__.py#L49

phw (dismissed):
Good move. LGTM

phw:
> Actually there's one plugin using WSGetRequest : https://github.com/metabrainz/picard-plugins/blob/c03c7f1fdb2546d4e3467d99859406ce6cf0bc05/plugins/tangoinfo/__init__.py#L49

Oh yes, that was a very specific workaround. Issue was for some reason that the URL always contained the port number, and the tango.info server did error out in this case.

But as noted in the comments in https://github.com/metabrainz/picard-plugins/pull/323 this was fixed in https://github.com/metabrainz/picard/pull/2057 by not setting the port number if it is the default port. So we should be able to get rid of that workaround, at least for newer Picard.

But we could also keep a class available for compatibility:

```python
class WSGetRequest(WSRequest):
    pass
```

Just that, with a big fat deprecation comment on top :)

phw:
Nah, not even needed..The plugin only loads this class for Picard < 2.8.0dev1, so it's ok to remove it without issue

phw (dismissed):

phw (approved):

picard/webservice/__init__.py
```diff
@@ -97,60 +97,95 @@ def __init__(self, response_type):
...
             mblogin: Hints that this request should be tied to a MusicBrainz
             account, requiring that we obtain an OAuth token first.
             cacheloadcontrol: See `QNetworkRequest.Attribute.CacheLoadControlAttribute`.
             refresh: Indicates a user-specified resource refresh, such as when
-            the user wishes to reload a release.  Marks the request as high priority
-            and disables caching.
+                the user wishes to reload a release.  Marks the request as high priority
+                and disables caching.
             queryargs: `dict` of query arguments.
-            retries: Current retry attempt number.
```
zas, line 122: I removed this one about `retries` since this is a leftover from old changes, it makes no sense in the current code.

phw (approved):

PR #2221, opened 2023-06-03
Webservice tidy up (pass 2)

zas:
Introduce WSRequest url parameter in order to simplify code.

This is a move I wanted to do since a long time.
WSRequest can now accept a QUrl directly, so we don't need to pass host, port, path, queryargs individually.
It helps to simplify the code because it avoids conversion from QUrl to individual parts, then from parts to QUrl.

Url can be passed as a QUrl or as a string (it will be converted to QUrl in constructor).

The whole class is covered by tests.

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -114,6 +114,7 @@ class WSRequest(QNetworkRequest):
         priority=False,
         important=False,
         request_mimetype=None,
+        qurl=None,
     ):
         """
         Args:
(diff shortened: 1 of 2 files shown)


rdswift (approved):
It all looks good to me.  I was going to suggest testing a non-standard port, but then I went through the tests again and found that it was already there.  Well done!

phw (approved):

PR #2225, opened 2023-06-04
Introduce WebService (get|post|put|delete|download)_url() methods

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -37,7 +37,6 @@ from PyQt5.QtCore import (
     QMutex,
     QObject,
     QUrl,
-    QUrlQuery,
 )
 
 from picard import log
@@ -160,7 +159,10 @@ class CoverArtImage:
         else:
             self.types = types
         if url is not None:
-            self.parse_url(url)
+            if not isinstance(url, QUrl):
+                self.url = QUrl(url)
+            else:
+                self.url = url
         else:
             self.url = None
         self.comment = comment
@@ -177,17 +179,6 @@ class CoverArtImage:
             self.support_multi_types = support_multi_types
         if data is not None:
             self.set_data(data)
-
-    def parse_url(self, url):
-        self.url = QUrl(url)
-        self.host = self.url.host()
-        self.port = self.url.port(443 if self.url.scheme() == 'https' else 80)
-        self.path = self.url.path(QUrl.ComponentFormattingOption.FullyEncoded)
-        if self.url.hasQuery():
-            query = QUrlQuery(self.url.query())
-            self.queryargs = dict(query.queryItems())
-        else:
-            self.queryargs = None
 
     @property
     def source(self):
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -465,20 +465,12 @@ class CoverArtBox(QtWidgets.QGroupBox):
             self.load_remote_image(url, fallback_data)
 
         if url.scheme() in {'http', 'https'}:
-            path = url.path()
-            if url.hasQuery():
-                query = QtCore.QUrlQuery(url.query())
-                queryargs = dict(query.queryItems())
-            else:
-                queryargs = {}
-            if url.scheme() == 'https':
-                port = 443
-            else:
-                port = 80
-            self.tagger.webservice.get(url.host(), url.port(port), path,
-                                       partial(self.on_remote_image_fetched, url, fallback_data=fallback_data),
-                                       parse_response_type=None, queryargs=queryargs,
-                                       priority=True, important=True)
+            self.tagger.webservice.download_url(
+                url=url,
+                handler=partial(self.on_remote_image_fetched, url, fallback_data=fallback_data),
+                priority=True,
+                important=True,
+            )
         elif url.scheme() == 'file':
             path = normpath(url.toLocalFile().rstrip("\0"))
             if path and os.path.exists(path):
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -105,6 +105,28 @@ def hostkey_from_url(url):
     return (url.host(), port_from_qurl(url))
 
 
+def host_port_to_url(host, port, path=None, scheme=None, as_string=False):
+    """Convert host & port (with optional path and scheme) to an URL"""
+    url = QUrl()
+    if scheme is None:
+        if port == 443:
+            scheme = 'https'
+        else:
+            scheme = 'http'
+    url.setScheme(scheme)
+
+    if ((scheme == 'https' and port != 443)
+            or (scheme == 'http' and port != 80)):
+        url.setPort(port)
+
+    url.setHost(host)
+
+    if path is not None:
+        url.setPath(path)
+
+    return url.toString() if as_string else url
+
+
 class WSRequest(QNetworkRequest):
     """Represents a single HTTP request."""
     _access_token = None
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -64,6 +64,7 @@ from picard.util import (
 )
 from picard.util.xml import parse_xml
 from picard.webservice import ratecontrol
+from picard.webservice.utils import port_from_qurl
 
 
 COUNT_REQUESTS_DELAY_MS = 250
@@ -89,42 +90,6 @@ class UnknownResponseParserError(Exception):
     def __init__(self, response_type):
         message = "Unknown parser for response type '%s'. Parser for given response type does not exist." % response_type
         super().__init__(message)
-
-
-def port_from_qurl(qurl):
-    """Returns QUrl port or default ports (443 for https, 80 for http)"""
-    if qurl.scheme() == 'https':
-        return qurl.port(443)
-    return qurl.port(80)
-
-
-def hostkey_from_url(url):
-    """Returns (host, port) from passed url (as string or QUrl)"""
-    if not isinstance(url, QUrl):
-        url = QUrl(url)
-    return (url.host(), port_from_qurl(url))
-
-
-def host_port_to_url(host, port, path=None, scheme=None, as_string=False):
-    """Convert host & port (with optional path and scheme) to an URL"""
-    url = QUrl()
-    if scheme is None:
-        if port == 443:
-            scheme = 'https'
-        else:
-            scheme = 'http'
-    url.setScheme(scheme)
-
-    if ((scheme == 'https' and port != 443)
-            or (scheme == 'http' and port != 80)):
-        url.setPort(port)
-
-    url.setHost(host)
-
-    if path is not None:
-        url.setPath(path)
-
-    return url.toString() if as_string else url
 
 
 class WSRequest(QNetworkRequest):
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -629,6 +629,34 @@ class WebService(QtCore.QObject):
         )
         return self.add_request(request)
 
+    def get_url(self, **kwargs):
+        kwargs['method'] = 'GET'
+        kwargs['parse_response_type'] = kwargs.get('parse_response_type', DEFAULT_RESPONSE_PARSER_TYPE)
+        return self.add_request(WSRequest(**kwargs))
+
+    def post_url(self, **kwargs):
+        kwargs['method'] = 'POST'
+        kwargs['parse_response_type'] = kwargs.get('parse_response_type', DEFAULT_RESPONSE_PARSER_TYPE)
+        kwargs['mblogin'] = kwargs.get('mblogin', True)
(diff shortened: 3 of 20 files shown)


zas:
@phw this is still work in progress, and it needs careful review.

picard/webservice/__init__.py
```diff
@@ -149,6 +151,14 @@ def __init__(
 
         if not isinstance(url, QUrl):
             url = QUrl(url)
+
+        if queryargs is not None:
+            query = QtCore.QUrlQuery(url)
+            for k, v in queryargs.items():
+                # FIXME: check if encoding is correct
+                query.addQueryItem(k, str(v))
```
zas, line 159: I'm not sure encoding is correct here. Code is identical to `build_qurl()`.

See https://doc.qt.io/qt-6/qurlquery.html#QUrlQuery-1

test/test_webservice.py
```diff
@@ -521,3 +521,13 @@ def test_set_retries_not_reached(self):
...
+
+    def test_queryargs(self):
+        request = WSRequest(
+            url='http://example.org/path?a=1',
+            method='GET',
+            handler=dummy_handler,
+            queryargs={'a': 2, 'b': 'x%20x', 'c': '1+2', 'd': '&', 'e': '?'},
+        )
+        # FIXME: check encoding
+        self.assertEqual(request.url().toString(), 'http://example.org/path?a=1&a=2&b=x x&c=1+2&d=%26&e=?')
```
zas, line 533: @phw: what do we expect here?

picard/webservice/api_helpers.py
```diff
@@ -78,35 +78,36 @@ def host(self):
...
-
-    def delete(self, path_list, handler, priority=True, important=False,
-               mblogin=True, queryargs=None):
-        path = self.api_path + "/".join(path_list)
-        return self._webservice.delete(self.host, self.port, path, handler,
-                                       priority=priority, important=important, mblogin=mblogin,
-                                       queryargs=queryargs)
+        return host_port_to_url(self.host, self.port, path=path)
+
+    def get(self, path_list, handler, **kwargs):
```
zas, line 85: This move breaks tests, but I'm not sure why. I guess that's because the prototype changed, as tests using MagickMock

test/test_webservice.py
phw: I probably would have assumed the queryargs are considered totally unencoded and supposed to get submitted as is. Which means I'd expect it to encode the parameter `b` as `x%2520x`. Makes sense?

zas: Yes, I noticed `queryargs` are encoded sometimes: https://github.com/metabrainz/picard/blob/5fb2ad4017b16f9eb4bc609c34dba4627663ebe1/picard/webservice/api_helpers.py#L176

picard/webservice/api_helpers.py
```diff
@@ -59,91 +58,74 @@ def _wrap_xml_metadata(data):
 
 
 class APIHelper(object):
+    base_path = "/"
 
-    def __init__(self, host, port, api_path, webservice):
-        self._host = host
-        self._port = port
-        self.api_path = api_path
+    def __init__(self, webservice):
```
phw, line 63: I'd make the base_path a parameter here. Actually I'd consider making it the base URL. Then you can easily initialize an API handler as e.g.

```python
lb_api = APIHelper(ws, base_url="https://api.listenbrainz.org/1/")
lb_api.get(....)
```

Derived sub classes can still be made easily by overriding `__init__`  and providing hard coded defaults.

Essentially this is also why tests break, as the tests test the APIHelper class in standalone use, which this PR currently breaks.

picard/webservice/api_helpers.py
phw: See my comment above: The tests break because the APIHelper got a completely changed interface.

picard/webservice/api_helpers.py
```diff
@@ -59,91 +58,74 @@ def _wrap_xml_metadata(data):
...
-                                       queryargs=queryargs)
+    def url(self):
+        raise NotImplementedError
+
+    def url_from_path_list(self, path_list):
+        url = self.url
+        url.setPath("/".join([self.base_path] + list(path_list)))
+        return url
+
+    def get(self, path_list, handler, **kwargs):
```
phw, line 79: I am not too sure on the `path_list` . Is this really the most convenient interface? Taking my LB API example from above:

```python
lb_api = APIHelper(ws, base_url="https://api.listenbrainz.org/1/")
lb_api.get(('user', 'outsidecontext', 'playing-now'), some_handler)
```

vs.

```python
lb_api = APIHelper(ws, base_url="https://api.listenbrainz.org/1/")
lb_api.get('user/outsidecontext/playing-now', some_handler)
```

To me pathing the path as a string seems easier to read and write. Or maybe be flexible?

phw (changes requested):
Overall this looks really good, great.

I have some concerns around the changes to the `APIHelper`'s interface. I will provide a PR with some suggestions for dicsussion.

picard/webservice/api_helpers.py
zas: Well, I kept it because it was there already.
I was wondering about that too.

test/test_webservice.py
zas: Introduction of `unencoded_queryargs` solves that

picard/webservice/api_helpers.py
zas: Check https://github.com/metabrainz/picard/pull/2225/commits/fb6401cda297b4a0c8b658dfcbe45de7a0daab05

I think that's simpler and easier to understand.

zas:
@rdswift if you have time, give it a test run, especially looking for broken queries (badly built)

picard/webservice/api_helpers.py
```diff
@@ -60,102 +60,84 @@ def _wrap_xml_metadata(data):
...
 
     def __init__(self, webservice):
-        super().__init__(None, None, "/ws/2/", webservice)
-
-    @property
-    def host(self):
         config = get_config()
-        return config.setting['server_host']
+        host = config.setting['server_host']
+        port = config.setting['server_port']
```
zas, line 113: @phw, this change is incorrect I think, because class is instanced once in https://github.com/metabrainz/picard/blob/8ed7d4343511242d6604ed27c2ee2d2a566af02c/picard/tagger.py#L332
But those values can be modified during runtime.

rdswift:
> @rdswift if you have time, give it a test run, especially looking for broken queries (badly built)

I've only been able to do some limited testing, but so far everything seems to be working as expected.

rdswift (commented):
I've only been able to do some limited testing, but so far everything seems to be working as expected.  Nice!

picard/webservice/api_helpers.py
zas: Check https://github.com/metabrainz/picard/pull/2225/commits/9a0bedff20d266f6e32b9b2db460775e38456bb6

phw: Great. Makes also cleaner code in the base class

test/test_webservice.py
```diff
@@ -497,3 +526,74 @@ def test_set_retries_not_reached(self):
...
         self.assertFalse(request.max_retries_reached())
+
+    def test_queryargs(self):
+        request = WSRequest(
+            url='http://example.org/path?a=1',
+            method='GET',
+            handler=dummy_handler,
+            queryargs={'a': 2, 'b': 'x%20x', 'c': '1+2', 'd': '&', 'e': '?'},
+        )
+        expected = 'http://example.org/path?a=1&a=2&b=x x&c=1+2&d=%26&e=?'
```
zas, line 537: @phw I wonder about expected behavior here when url has already a query (`?a=1`), the current code just append `queryargs` key/value elements.
I don't think that's a case happening in current code, but it would probably make more sense to actually replace existing values with those passed from `queryargs`.

test/test_webservice.py
```diff
@@ -497,3 +526,74 @@ def test_set_retries_not_reached(self):
...
+        self.assertEqual(request.url().toString(), expected)
+
+    def test_unencoded_queryargs(self):
+        request = WSRequest(
+            url='http://example.org/path?a=1',
+            method='GET',
+            handler=dummy_handler,
+            unencoded_queryargs={'a': 2, 'b': 'x%20x', 'c': '1+2', 'd': '&', 'e': '?'},
+        )
+        expected = 'http://example.org/path?a=1&a=2&b=x%2520x&c=1%2B2&d=%26&e=%3F'
```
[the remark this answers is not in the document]
phw: Yes, the encoding is definitely preferable. A `+` would sometimes be interpreted as a space.

test/test_webservice.py
phw: I tend to agree that I would have expected replacement. On the other hand there could be even a use case for the behavior: If you access an API where some kind of access token gets transferred by query parameter you could define the base URL as `https://example.com/api/?token=12345`, and then still do extra requests with additional query parameters.

Currently this is all hypothetical though, as this is not used in practice.

In the end I have no strong opinion either way.

picard/webservice/api_helpers.py
```diff
@@ -243,13 +243,11 @@ def delete_from_collection(self, collection_id, releases, handler):
...
 
-    base_path = '/v2'
     client_key = ACOUSTID_KEY
     client_version = PICARD_VERSION_STR
 
-    @property
-    def url(self):
-        return QUrl(ACOUSTID_URL)
+    def __init__(self, webservice):
+        super().__init__(webservice, base_url=ACOUSTID_URL + '/v2')
```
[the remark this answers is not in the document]
phw: I think yes

test/test_webservice.py
zas: I'll keep it as it is now, we can always change it later if needed

picard/webservice/api_helpers.py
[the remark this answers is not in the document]
zas: Done in https://github.com/metabrainz/picard/pull/2225/commits/db4d9cfa0ad29fe947862669dfd1c4a340aa4f19

picard/webservice/__init__.py
```diff
@@ -134,6 +136,8 @@ def __init__(
             important: Indicates that this is an important request.
             request_mimetype: Set the Content-Type header.
             url: URL passed as a string or as a QUrl to use for this request
+            queryargs: Encoded query arguments, a dictionary mapping field names to values
+            unencoded_queryargs: Unencoded query arguments, a dictionary mapping field names to values
```
phw, line 140: I think we should not have both, only one parameter accepting unencoded query args. Or do we need that queryargs argument to be encoded for some use case?

zas: I think keeping both doesn't hurt for now, but getting rid of `queryargs` is my goal.
We'll do another round of cleanup.
The PR is big enough ;)

There's also `MUSICBRAINZ_SERVERS` to look into (host, port -> URL)

picard/ui/coverartbox.py
```diff
@@ -465,20 +465,12 @@ def fetch_remote_image(self, url, fallback_data=None):
...
-            else:
-                queryargs = {}
-            if url.scheme() == 'https':
-                port = 443
-            else:
-                port = 80
-            self.tagger.webservice.get(url.host(), url.port(port), path,
-                                       partial(self.on_remote_image_fetched, url, fallback_data=fallback_data),
-                                       parse_response_type=None, queryargs=queryargs,
-                                       priority=True, important=True)
```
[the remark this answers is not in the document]
zas: Yes, overall it simplifies code a lot, those long lists of arguments passed from methods to methods were the main motivation at start, it is awful to maintain.

We could also create an object to contain all WSRequest options (and pass it as one parameter), but we had to start somewhere...

picard/coverart/image.py
```diff
@@ -178,17 +180,6 @@ def __init__(self, url=None, types=None, comment='', data=None, support_types=No
         if data is not None:
             self.set_data(data)
 
-    def parse_url(self, url):
```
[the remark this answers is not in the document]
zas: https://github.com/metabrainz/picard-plugins/blob/c03c7f1fdb2546d4e3467d99859406ce6cf0bc05/plugins/theaudiodb/__init__.py#L97 says it was for Picard < 2.1.3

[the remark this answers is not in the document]
phw: Yes exactly. My point was that nothing breaks due to this, we can keep that method override. With old code it will work (and even fix the bug for old Picard versions), for the new code it just never gets called.

phw (approved):

PR #2227, opened 2023-06-07
Update calculation of patch versions for Windows store

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

For Windows store we combine the patch version with the build number to ensure continuously growing version numbers for upgrades. Only this allows upgrades also for the development builds of the .msix package.

Currently the patch version gets multiplied by 10000, which means Picard version 2.9.2 becomes 2.9.20000. As Windows  limits the patch number to max. 65535 this means we can only release Picard 2.9.6 and not 2.9.7.

# Solution
Reduce the version numbers to be multiplied to 1000 to allow for two digit patch version numbers. This still leaves us 999 build numbers. As the build numbers are counted as commits since last release tag this should mostly give us enough room (worst case we get stuck at build version 999, which only affects development builds for local testing anyway).

The highest version by this patter is Picard x.x.65 build 535.

Release versions always are for the release tag and hence the build version is set to 0, so the highest release version for e.g. Picard 2.9 will be 2.9.65000.

Compare also previous changes in 1948f0c1164fed590d16dafe9c6bcfd21fe7122f

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -302,7 +302,10 @@ class picard_build(build):
             generate_file('win-version-info.txt.in', 'win-version-info.txt', {**args, **version_args})
 
             default_publisher = 'CN=Metabrainz Foundation Inc., O=Metabrainz Foundation Inc., L=San Luis Obispo, S=California, C=US'
-            store_version = (PICARD_VERSION.major, PICARD_VERSION.minor, PICARD_VERSION.patch * 10000 + self.build_number, 0)
+            # Combine patch version with build number. As Windows store apps require continuously
+            # growing version numbers we combine the patch version with a build number set by the
+            # build script.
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

"As Windows limits the patch number to max. 65535" ... Windows is still in 16 bits...

phw:
Yes, it is a mess really. There is also a build number, but for the store release we are not allowed to use it (it's used internally).

Overall we can't fully match our versions to theirs. But most important issue I wanted to address is that we can release a Picard 2.9.10 if needed. With the previous numbering we would have gotten into trouble.

The build number thing is for internal testing use only. There are scenarios where it breaks, but that's not so important.

PR #2229, opened 2023-06-08
PICARD-2654: Display a message when rating submission fails due to server error

zas:
- a message is written to status bar and history, translatable
- another message is sent via log.error()

It also ensures we have an handler for this request.

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

* JIRA ticket (_optional_): [PICARD-2654](https://tickets.metabrainz.org/browse/PICARD-2654)

# Solution

# Action

### DIFF
--- a/picard/ui/ratingwidget.py
+++ b/picard/ui/ratingwidget.py
@@ -93,6 +93,15 @@ class RatingWidget(QtWidgets.QWidget):
             rating = self._maximum
         return rating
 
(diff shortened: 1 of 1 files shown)


phw (approved):
Looks good. We should have a ticket for this for the changelog

PR #2238, opened 2023-06-14
Use notarytool for macOS notarization

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Apple has deprecated using `altool`  for the app notarization. Instead the new `notarytool`  should be used. See https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool for background.

# Solution

Perform the notarization with the new `notarytool`. This also has the advantage that there is now a clear way to specify the MetaBrainz team by specifying the team ID. In addition we can get rid of the polling because `notarytool`  supports a [`--wait` parameter](https://developer.apple.com/documentation/technotes/tn3147-migrating-to-the-latest-notarization-tool#Submit-a-file).

The entire process is also faster.

Tested successfully on my repo phw/picard . See https://github.com/phw/picard/actions/runs/5264236895

### DIFF
--- a/scripts/package/macos-notarize-app.sh
+++ b/scripts/package/macos-notarize-app.sh
@@ -30,13 +30,14 @@ ditto -c -k --rsrc --keepParent "$APP_BUNDLE" "${APP_BUNDLE}.zip"
 
 # Submit for notarization
 echo "Submitting $APP_BUNDLE for notarization..."
-RESULT=$(xcrun altool --notarize-app --type osx \
-  --file "${APP_BUNDLE}.zip" \
-  --primary-bundle-id org.musicbrainz.Picard \
-  --username "$APPLE_ID_USER" \
-  --password @env:APPLE_ID_PASSWORD \
-  -itc_provider MetaBrainzFoundationInc \
-  --output-format xml)
+RESULT=$(xcrun notarytool submit \
+  --apple-id "$APPLE_ID_USER" \
+  --team-id "$APPLE_ID_TEAM" \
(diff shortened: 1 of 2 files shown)


scripts/package/macos-notarize-app.sh
```diff
@@ -30,56 +30,32 @@ ditto -c -k --rsrc --keepParent "$APP_BUNDLE" "${APP_BUNDLE}.zip"
...
-REQUEST_UUID=$(echo "$RESULT" | xpath \
-  "//key[normalize-space(text()) = 'RequestUUID']/following-sibling::string[1]/text()" 2> /dev/null)
+STATUS=$(echo "$RESULT" | xpath \
+  "//key[normalize-space(text()) = 'status']/following-sibling::string[1]/text()" 2> /dev/null)
 
-if [ -z "$REQUEST_UUID" ]; then
-  echo "Submitting $APP_BUNDLE failed:"
+if [ "$STATUS" = "Accepted" ]; then
+  echo "Notarization of $APP_BUNDLE succeeded!"
+  break
```
zas, line 53:
```suggestion
  exit 0
```

zas (changes requested):

scripts/package/macos-notarize-app.sh
phw: Actually that line needs to be removed completely. A left over from when this was inside the loop. Will fix

zas (approved):
LGTM

PR #2241, opened 2023-06-16
Fix appstream XML and use appstreamcli for validation

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

The AppStream metadata was invalid (multiple `<recommends>` , should be only one).

We have AppStream validation already. But it was using the older ` appstream-util` which did not catch that issue. Replaced this with `appstreamcli`.

### DIFF
--- a/.github/workflows/validate-appstream.yml
+++ b/.github/workflows/validate-appstream.yml
@@ -19,11 +19,14 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.11'
-    - name: Install appstream
+    - name: Install appstream-util
       run: |
         sudo apt-get update
-        sudo apt-get install appstream gettext
+        sudo apt-get install appstream-util gettext
     - name: Validate AppStream metadata
       run: |
         python setup.py build_appdata
-        appstreamcli validate --pedantic --explain org.musicbrainz.Picard.appdata.xml
+        appstream-util validate-relax org.musicbrainz.Picard.appdata.xml
+        # TODO: Consider using appstreamcli once ubuntu-latest provides
+        #       appstreamcli>=0.16, as this tool catches more issues.
+        #appstreamcli validate --pedantic --explain org.musicbrainz.Picard.appdata.xml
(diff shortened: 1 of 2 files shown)


.github/workflows/validate-appstream.yml
```diff
@@ -19,11 +19,11 @@ jobs:
...
-    - name: Install appstream-util
+    - name: Install appstream
       run: |
         sudo apt-get update
         sudo apt-get install appstream-util gettext
     - name: Validate AppStream metadata
       run: |
         python setup.py build_appdata
-        appstream-util validate-relax org.musicbrainz.Picard.appdata.xml
+        appstreamcli validate --pedantic --explain org.musicbrainz.Picard.appdata.xml
```
zas, line 29: https://manpages.ubuntu.com/manpages/jammy/en/man1/appstreamcli.1.html says `appstreamcli` is provided by `appstream` package, which is not installed above.

zas (changes requested):

.github/workflows/validate-appstream.yml
```diff
@@ -19,11 +19,11 @@ jobs:
       uses: actions/setup-python@v4
       with:
         python-version: '3.11'
-    - name: Install appstream-util
+    - name: Install appstream
       run: |
         sudo apt-get update
         sudo apt-get install appstream-util gettext
```
zas, line 25:
```suggestion
        sudo apt-get install appstream gettext
```

Not sure appstream-util is still needed

zas (changes requested):

.github/workflows/validate-appstream.yml
phw: I wanted to change that, but only changed it in the step's description. Oh well, will fix.

phw:
@zas Validation fails, but that's because of appstream 0.15 on the CI. Locally with 0.16 it validates (with some "pedantic" output I keep for information, but it doesn't cause the validation to fail).

I'll revert this back to appstream-cli for now.

phw:
@zas https://github.com/metabrainz/picard/pull/2241/commits/bbd98b97d4f75a3566f41bf2d34b59663a2aed5d

zas (approved):

PR #2246, opened 2023-06-26
PICARD-2671: Add file save confirmation and first run dialogs

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds messagebox dialogs on first run and when saving files.

# Problem

There seems to be a need to inform the users about the hazards of saving files if Picard is incorrectly configured, and to encourage new users to RTFM.

* JIRA ticket (_optional_): PICARD-2671

# Solution

Add a messagebox that displays when Picard is started (on first run), introducing the program and encouraging the user to RTFM to avoid some of the hazards of saving files when Picard is not properly configured.

Add a warning messagebox that displays each time the user initiates a file save action.

Include options to disable the messagebox displays in the future.

# Action

Please review the suggested message texts, and provide suggested improvements.

If this change is accepted, the documentation will need to be updated to include the new configuration setting (to enable or disable the warning message when initiating a save).


picard/ui/savewarningdialog.py
```diff
@@ -0,0 +1,70 @@
...
+        if not config.setting["dont_write_tags"]:
+            actions.append(_("overwrite existing metadata in the files"))
+        if config.setting["rename_files"] and config.setting["move_files"]:
+            actions.append(_("rename and move the files"))
+        elif config.setting["rename_files"]:
+            actions.append(_("rename the files"))
+        elif config.setting["move_files"]:
+            actions.append(_("move the files"))
+        if actions:
+            self.WARNING_TEXT = _("This will ") + ' and '.join(actions) + '.  This action cannot be undone.  Do you want to continue?'
```
zas, line 48: From i18n point of view this is incorrect (this will not translate well to other languages).
That's because of the string concatenation (see https://doc.qt.io/qt-6/i18n-source-translation.html#use-parameters-instead-of-concatenating-strings)

Something like:

```
This action will:
- action1
- action2
- action3

This action cannot be undone.  Do you want to continue?
```

would work better.

The list indicator has to be in the translated string, or each list element appended to a list widget.

To keep it simple, in this case, I would use Qt rich text (pseudo html):

```python
        if not config.setting['dont_write_tags']:
            actions.append(_("overwrite existing metadata in the files"))
        if config.setting['rename_files'] and config.setting['move_files']:
            actions.append(_("rename and move the files"))
        elif config.setting['rename_files']:
            actions.append(_("rename the files"))
        elif config.setting['move_files']:
            actions.append(_("move the files"))
        if actions:
            header = _("This action will:")
            footer = _("<strong>This action cannot be undone.</strong>  Do you want to continue?")
            list_of_actions = ''
            for action in actions:
                 list_of_actions += _('<li>{action}</li>').fomat(action=action)
            self.WARNING_TEXT = _('<p>{header}</p><ul>{list_of_actions}</ul><p>{footer}</p>').format(header=header, list_of_actions=list_of_actions, footer=footer)
```

(totally untested).

picard/ui/firstrundialog.py
```diff
@@ -0,0 +1,74 @@
...
+                "MusicBrainz Picard is an extremely powerful cross-platform music file tagger.  "
+                "In addition to downloading metadata from the MusicBrainz database and updating the tags in your music files, "
+                "it can also automatically rename your files and move them into directories based on a file renaming script."
+                "</p><p>"
+                "Because of Picard's power and flexibility, it has a great number of option settings that can be configured.  "
+                "We encourage all new users to review the on-line documentation found under the 'Help' menu (or by pressing the F1 key) to see what "
+                "each option does.  There are also explanations of all the screens displayed by the program, as well as suggested work flows and tutorials to "
+                "help you get started."
+                "</p><p>"
+                "Finally, we strongly encourage you to work with a copy of your music files initially, until you are sure that Picard is configured properly and "
```
zas, line 48: I'd say "we strongly encourage to work backup original files and to work on small batches." 
(not only initially, experienced uses will ignore this if they're confident enough).

Also, a note about Picard is provided as is, with no guarantee, and that user uses it at his own risk (basically a reminder of the license).

zas (changes requested):
Overall, it looks good

picard/ui/savewarningdialog.py
rdswift: I always seem to screw up the i18n stuff.  Thanks (again) for helping me get it fixed.  I changed it based on your suggested code, but avoided translating the `'<p>{header}</p><ul>{list_of_actions}</ul><p>{footer}</p>'` part because `header`, `list_of_actions` and `footer` were already translated and the remaining html tags shouldn't be translated.  Please have a look to make sure it looks okay.

picard/ui/firstrundialog.py
rdswift: Updated to include your suggestions.  Thanks.

picard/ui/savewarningdialog.py
```diff
@@ -0,0 +1,70 @@
...
+            actions.append(_("rename the files"))
+        if config.setting['move_files']:
+            actions.append(_("move the files"))
+
+        if actions:
+            header = _("This action will:")
+            footer = _("<strong>This action cannot be undone.</strong>  Do you want to continue?")
+            list_of_actions = ''
+            for action in actions:
+                list_of_actions += f'<li>{action}</li>'
```
zas, line 46: Actually those pseudo HTML parts also need to be translated even though they don't contain textual strings(I think it was the case in my example).
Mainly because of `dir` attribute that can be set by a translator for languages not using default left to right direction.
Same goes for HTML below.

See https://doc.qt.io/qt-6/richtext-html-subset.html#block-attributes

zas (changes requested):

zas (approved):
LGTM

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,73 @@
...
+
+class NewUserDialog():
+
+    def __init__(self):
+
+        self.DIALOG_TEXT = _(
+            (
+                "<p>"
+                "MusicBrainz Picard is an extremely powerful cross-platform music file tagger.  "
+                "In addition to downloading metadata from the MusicBrainz database and updating the tags in your music files, "
```
Sophist-UK, line 35: I think an additional sentence would be useful here to explain that the primary purpose of Picard is to tag music files with metadata held in Musicbrainz (overwriting most if not all existing tags) and optionally renaming files based on this same metadata - and also that the majority of usage is to tag albums rather than individual files - and that if this is not your goal, then Picard may not be the right tagger for your needs.

Sophist-UK: Also, I think that this should be a standard disclaimer that is also included word for word in the introductory web pages of the documentation site.

picard/ui/savewarningdialog.py
```diff
@@ -0,0 +1,70 @@
...
+        if not config.setting['dont_write_tags']:
+            actions.append(_("overwrite existing metadata in the files"))
+        if config.setting['rename_files']:
+            actions.append(_("rename the files"))
+        if config.setting['move_files']:
+            actions.append(_("move the files"))
+
+        if actions:
+            header = _("This action will:")
+            footer = _("<strong>This action cannot be undone.</strong>  Do you want to continue?")
```
Sophist-UK, line 43: Personally I think that the list of actions needs to give more details about what can go wrong (based on the same actions) i.e. especially move the files. Perhaps
```
actions.append(_("move the files - only continue if you are happy the your file naming script works as you intend")) 
```

picard/ui/ui_options_interface.py
```diff
@@ -117,6 +120,7 @@ def retranslateUi(self, InterfaceOptionsPage):
         self.builtin_search.setText(_("Use builtin search rather than looking in browser"))
         self.use_adv_search_syntax.setText(_("Use advanced query syntax"))
         self.quit_confirmation.setText(_("Show a quit confirmation dialog for unsaved changes"))
+        self.file_save_warning.setText(_("Show a confirmation dialog when saving files"))
```
Sophist-UK, line 123: I know it may seem a little odd to provide an option to reset the first use warning, but I am personally reasonably confident that at some point someone is going to regret clearing this flag and want to reset it.

picard/ui/newuserdialog.py
zas: I think the text is too long now, and people are unlikely to read it.

image

I suggest we keep it as short as possible, being very clear about risks, but that's all.

1) Read the documentation
2) Work on copies and in small batches
3) Use at your own risk

What do you think?

Sophist-UK: @zas Yes - I think you are right. But unfortunately the alternative would be to simplify it and include a link to the same or an even longer explanation which would then be even less likely to be read. Here is my attempt at making this more punchy:

>                 NEW USERS - READ THIS BEFORE USING PICARD
>
> Picard is designed to download music meta-data and cover art from MusicBrainz (a massive online database of information about music releases) and then use it to overwrite any existing tags and rename and move the files into a structured "library". It is best used with albums, though with careful configuration it can be used with individual tracks. If this is not your intent, then this may not be the best tool for you. 
>
> If this is what you want to achieve, then Picard can be a very powerful tool for you, but if you configure its options incorrectly, you can create havoc with your music collection.
>
> Thousands of users around the world successfully use Picard to tag and structure their music collections, and so that you also can use Picard successfully to tag your music collection we recommend that you:
> 
> 1. Read the HELP documentation - press F1 or use the Help menu. Use it to understand how to use Picard to match your music files against releases in the MusicBrainz database.
> 2. Configure your file naming script carefully.
> 3. Work on copies of your music files and in small batches until you are fully confident that your music files will be handled the way you want them too. Once Picard has updated a file, there is no way to undo the changes.
>
> ✅ Show this message again the next time you start Picard
>
> Picard is open source software written by volunteers and as such is provided as-is and with no warranty - you use Picard at your own risk.

zas: @Sophist-UK your suggestion is even longer, and as is, I'm pretty sure new users will press Ok and don't read.
If people aren't interested in details and don't follow the link or follow it and don't read, well, that's their problem not ours.

We need to provide concise information about risks for audio files to new users. The message has to be short and clear, providing a link is just for 1% of users that will bother clicking it, and for 1% of those that will actually read the page.

What we need instead is a message 80% of new users will read, and as such it has to be very very short and impactful (like a good ad).

To me what you suggest would perfectly fit the linked document, but not the dialog box.

@rdswift @phw what do you think?

Sophist-UK: @zas - Yes, I know mine was actually about 25% longer, but then again it actually says more and in a more punchy way - and I am sure that it can be cut down a bit. Here is a short, blunt alternative:

>            NEW USERS - READ THIS BEFORE USING PICARD
>
> Picard is a very flexible music tagging tool, using MusicBrainz metadata for tags and file naming, but if you don't understand how it works, what its limitations are and how to configure it you can easily screw up your music library.
>
> **We therefore STRONGLY recommend that you:**
> 1. Read the [Picard documentation at https://picard-docs.musicbrainz.org/](https://picard-docs.musicbrainz.org/) before you use this tool on your music collection.**
> 2. Work on copies of your music files and in small batches until you are fully confident that your music files will be handled the way you want them too. Once Picard has updated a file, there is no way to undo the changes.
>
> ✅ Show this message again the next time you start Picard
>
> Picard is open source software written by volunteers and as such is provided as-is and with no warranty - you use Picard at your own risk.

rdswift: I like the shortened version, and have updated the PR accordingly.  I also added the number of files that will be processed to the file save warning dialog.

picard/ui/ui_options_interface.py
rdswift: I'm not sure that's still the case with the shortened version of the new user dialog?

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,72 @@
...
+                "<strong>You use Picard at your own risk.</strong>"
+                "</p>"
+            )
+        ).format(documentation_url=PICARD_URLS['documentation_server'])
+
+        self.show_again = True
+        show_again_text = _("Show this message again the next time you start Picard.")
+
+        self.msg = QtWidgets.QMessageBox()
+        self.msg.setIcon(QtWidgets.QMessageBox.Information)
```
zas, line 56:
```suggestion
        self.msg.setIcon(QtWidgets.QMessageBox.Warning)
```

This message is definitively a warning, so warning icon makes more sense to me. What do you think?

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,72 @@
...
+            (
+                "<h2 align=center>READ THIS BEFORE USING PICARD</h2>"
+                "<p>"
+                "Picard is a very flexible music tagging tool, using MusicBrainz metadata for tags and file naming, "
+                "but if you don't understand how it works, what its limitations are, and how to configure it, "
+                "you can easily mess up your music library."
+                "</p><p>"
+                "We therefore <strong>STRONGLY</strong> recommend that you:"
+                "</p><ol>"
+                "<li>Read the Picard documentation at <a href='{documentation_url}'>{documentation_url}</a> "
```
zas, line 41:
```suggestion
                "<li>Read the Picard  <a href='{documentation_url}'>documentation</a> "
```

could work, though I noticed in dark theme here, link color is hard to read (that's another issue) and we can't set a tooltip (title attribute doesn't work).

image

So not sure about this.

rdswift: I actually prefer it as you've suggested.  As you say, the issue with links on the dark theme is another matter, and shouldn't impact how we decide what text to display.

picard/ui/savewarningdialog.py
phw: I don't think we need to overdo the warning here. It probably doesn't help if this gets too much a wall of text. IMHO it's enough to warn about that saving will do changes and mention the actions.

picard/ui/ui_options_interface.py
phw: I'd still allow bringing this dialog back if wanted.

phw (dismissed):
To me this looks good. Especially the save dialog is really useful.

I think one reason we have avoided adding this new user dialog is because we all thought it would be better to have a nice first use wizard kind of dialog. But it does not help not moving forward with an improvement just because of wanting something nobody is currently working on. So let's go with this change. We can include it in the next release without much trouble and still build upon this and turn this dialog into some more complex thing.

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,72 @@
...
+            (
+                "<h2 align=center>READ THIS BEFORE USING PICARD</h2>"
+                "<p>"
+                "Picard is a very flexible music tagging tool, using MusicBrainz metadata for tags and file naming, "
+                "but if you don't understand how it works, what its limitations are, and how to configure it, "
+                "you can easily mess up your music library."
+                "</p><p>"
+                "We therefore <strong>STRONGLY</strong> recommend that you:"
+                "</p><ol>"
+                "<li>Read the Picard  <a href='{documentation_url}'>documentation</a> "
```
zas, line 41:
```suggestion
                "<li>Read the Picard <a href='{documentation_url}'>documentation</a> "
```

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,72 @@
...
+                "before you use this tool on your music collection.  The link is also available from the Help menu.<br /></li>"
```
zas, line 42:
```suggestion
                "before you use this tool on your music collection. The link is also available from the Help menu.</li>"
```

picard/ui/newuserdialog.py
```diff
@@ -0,0 +1,72 @@
...
+                "<li>Work on copies of your music files and in small batches until you are fully confident that your music files will be handled the way "
+                "you want them to be. <strong>Once Picard has updated a file, there is no way to undo the changes</strong>.</li>"
+                "</ol><p>"
+                "Picard is open source software written by volunteers.  It is provided as-is and with no warranty. "
```
zas, line 46:
```suggestion
                "Picard is open source software written by volunteers. It is provided as-is and with no warranty. "
```

picard/ui/savewarningdialog.py
```diff
@@ -0,0 +1,75 @@
...
+            for action in actions:
+                list_of_actions += _('<li>{action}</li>').format(action=action)
+            if file_count:
+                count_text = _("The number of files to be processed is {file_count:,d}.").format(file_count=file_count)
+                warning_text = _('<p>{header}</p><ul>{list_of_actions}</ul><p>{count_text}</p><p>{footer}</p>').format(
+                    header=header, list_of_actions=list_of_actions, footer=footer, count_text=count_text)
+            else:
+                warning_text = _('<p>{header}</p><ul>{list_of_actions}</ul><p>{footer}</p>').format(header=header, list_of_actions=list_of_actions, footer=footer)
+        else:
+            warning_text = _("There are no actions selected.  No changes will be saved.")
```
zas, line 54:
```suggestion
            warning_text = _("There are no actions selected. No changes will be saved.")
```

zas (changes requested):
Double spaces isn't typographically correct AFAIK.

Sophist-UK:
One more question. As a long time user of Picard, will I get the New User warning the first time I run a release that contains this PR? It's not a big deal to dismiss it, and personally I won't be annoyed by it, but I can see some other long-time users taking offense - but I am unclear how we set the option to True for genuine new users and False for users who have just upgraded.

hacketiwack:
On my side, as a long time user as well, I wouldn't mind seeing this warning.
I wouldn't add more logic in the code just for that corner case.

rdswift:
@zas:
> Double spaces isn't typographically correct AFAIK.

I usually use two spaces to separate sentences in my writing, but I've seen it done both ways.  For example, the second paragraph of the copyright notice in all of the program files has two spaces between the sentences.  I believe this is copied directly from the GPL license text.

In any event, I've changed them all to single space as requested.

@phw:
> I don't think we need to overdo the warning here. It probably doesn't help if this gets too much a wall of text. IMHO it's enough to warn about that saving will do changes and mention the actions.

Do you mean that we should remove the `<strong>...</strong>` bits, or that the length of the text in the warning is appropriate?

@phw:
> I'd still allow bringing this dialog back if wanted.

The file save warning dialog for sure, but do you also mean the new user / first run dialog?  My thinking was that it doesn't really provide anything but an initial (static) warning, and once disabled would likely never be re-enabled.

@Sophist-UK:
> will I get the New User warning the first time I run a release that contains this PR?

Yes.  I agree with @hacketiwack that it is not really worth the extra code checking for a one-time thing.

picard/ui/options/interface.py
```diff
@@ -76,6 +76,7 @@ class InterfaceOptionsPage(OptionsPage):
         BoolOption("setting", "show_menu_icons", True if not IS_MACOS else False),  # On macOS it is not common that the global menu shows icons
         BoolOption("setting", "builtin_search", True),
         BoolOption("setting", "use_adv_search_syntax", False),
+        BoolOption("setting", "show_new_user", True),
```
zas, line 79: I suggest we use a more explicit name for the option: `show_new_user_dialog`

Patch -> https://github.com/rdswift/picard/pull/16

zas (changes requested):

zas (approved):
LGTM

rdswift:
image

This is the current version of the new user dialog.    A couple of questions:

1. Should bullet 2 and the final paragraph contain the bold text, or should the bold be removed (text stays)?  This could also apply to the bold, upper case text in the first line of the second paragraph, but I think that it may be appropriate in that case.
2. Should there be a blank line between the two bullets?  I added this simply because I thought it looked better and provided a visual separation between the two ideas.
3. Should the icon be changed to "Critical" to show more emphasis, as shown below?

image

Sophist-UK:
@rdswift Bob - Personally I am happy with the dialog as it is with the warning triangle and bold text. I think it gives a suitably strong warning without putting people off.

zas:
I'm also fine with the current version.

PR #2250, opened 2023-07-03
PICARD-2679: Fix open folder for Snap install

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2679

If Picard is installed via Snap all "Open folder" actions do not work. This is for example the case for "Open containing folder" for loaded files, but also for opening the plugin folder or config folder in options.

This is supposed to work via xdg-desktop-portal automatically, and it does work for the Flatpak install. But there seems to be an issue with this on Snap, see https://forum.snapcraft.io/t/xdg-open-or-gvfs-open-qdesktopservices-openurl-file-somelocation-file-txt-wont-open-the-file/16824/14 and https://bugreports.qt.io/browse/QTBUG-83939

# Solution
When running as a Snap use xdg-open instead of QDesktopServices. Introduce picard.util.open_local_path helper to wrap this functionality.

# Action

### DIFF
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -52,6 +52,7 @@ from picard.util import (
     bytes2human,
     encode_filename,
     format_time,
+    open_local_path,
     union_sorted_lists,
 )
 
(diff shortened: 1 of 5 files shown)


zas (approved):
LGTM

PR #2247, opened 2023-07-02
PICARD-2627: Handle closing pipe threads on app crash

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem

Same as #2244, but instead of https://github.com/metabrainz/picard/pull/2244/commits/040d6198dae3cc273fc83c50e2543e1f788bafa5 it does https://github.com/metabrainz/picard/commit/9ea4ff1945f78d9dc6c130f049ab17a38fd86e6c

# Solution

When `QApplication.quit()` is called always also run the `exit()` cleanup, that is supposed to cleanly shutdown uses resources and objects needing cleanup. This should even happen in case of an crash as quit is called in https://github.com/metabrainz/picard/blob/master/picard/__init__.py#L135

The commit includes changes that ensure the variables for the various objects needing shutdown, such as the `pipe_handler`, `webservice` and thread pools, are initiated early. That way `exit()` runs cleanly even on an early crash where not the entire application  has been initialized.

@zas Please give this some testing. But in my tests this gives a very clean shutdown, logging only the actual crash reason without follow up exceptions caused during shutdown.

PR #2251, opened 2023-07-04
Pipe reintroduce thread pool

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

@zas: This ads back the thread pool instance to the pipe implementation that we removed in #2247 . The reasoning is that currently a new thread pool is created on every call to `read_from_pipe` and `send_to_pipe` to then only use a single thread. This is quite some overhead and not how a pool of threads should be handled.

This also increases the timeout for the reading thread. Generally it is ok for the reading thread to wait for a while and block reading as it waits for input data. We need the timeout to be able to interrupt the thread, though, and it should be short enough to be not annoying for the user by having to wait too long for exit.

This change in timeout might also help with PICARD-2660, but I'm not totally sure. I first thought we could get rid of the call to the opposite method in read_from_pipe and send_to_pipe. But we need those to ensure the timed out thread exits. Otherwise we get either blocks or unhandled reader threads that "eat" messages.

And we also need the timeout, otherwise we end up in totally blocking situations.

### DIFF
--- a/picard/util/pipe.py
+++ b/picard/util/pipe.py
@@ -160,6 +160,8 @@ class AbstractPipe(metaclass=ABCMeta):
 
         self.unexpected_removal = False
 
+        self.__thread_pool = concurrent.futures.ThreadPoolExecutor(max_workers=6)
+
         for path in self._paths:
             self.path = path
             for arg in self._args:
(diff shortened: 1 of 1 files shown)


phw:
Closing as the PR did not update with new commits. I'll add a new one

PR #2252, opened 2023-07-05
Pipe refactor thread pool

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

@zas: This ads back the thread pool instance to the pipe implementation that we removed in #2247 . The reasoning is that currently a new thread pool is created on every call to `send_to_pipe` to then only use a single thread. This is quite some overhead and not how a pool of threads should be handled.

Furthermore the thread pool is no longer used at all in `read_from_pipe`. It's not needed as the pipe server already runs in a separate thread. This allows us to keep reading from the pipe without timeout, avoiding the constant looping (which only consumes CPU for no good reason). We also don't need any workarounds to close the timed out thread but just need closing the reader thread on application exit.

This change should also avoid the issue at [PICARD-2660](https://tickets.metabrainz.org/browse/PICARD-2660).

~~This also increases the timeout for the reading thread. Generally it is ok for the reading thread to wait for a while and block reading as it waits for input data. We need the timeout to be able to interrupt the thread, though, and it should be short enough to be not annoying for the user by having to wait too long for exit.~~

~~This change in timeout might also help with PICARD-2660, but I'm not totally sure. I first thought we could get rid of the call to the opposite method in read_from_pipe and send_to_pipe. But we need those to ensure the timed out thread exits. Otherwise we get either blocks or unhandled reader threads that "eat" messages.~~

### DIFF
--- a/picard/util/pipe.py
+++ b/picard/util/pipe.py
@@ -279,7 +279,10 @@ class AbstractPipe(metaclass=ABCMeta):
         log.debug("Stopping pipe")
         self.pipe_running = False
         self.send_to_pipe(self.MESSAGE_TO_IGNORE)
-        self.__thread_pool.shutdown(wait=True, cancel_futures=True)
+        try:
+            self.__thread_pool.shutdown(wait=True, cancel_futures=True)
+        except TypeError:  # cancel_futures is not supported on Python < 3.9
+            self.__thread_pool.shutdown(wait=True)
 
 
 class UnixPipe(AbstractPipe):
--- a/picard/util/pipe.py
+++ b/picard/util/pipe.py
@@ -107,7 +107,6 @@ class PipeErrorNoDestination(PipeError):
 class AbstractPipe(metaclass=ABCMeta):
     NO_RESPONSE_MESSAGE: str = "No response from FIFO"
     MESSAGE_TO_IGNORE: str = '\0'
-    TIMEOUT_SECS_READ: float = 5.0
     TIMEOUT_SECS_WRITE: float = 1.5
 
     @classmethod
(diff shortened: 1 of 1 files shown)


picard/util/pipe.py
```diff
@@ -262,31 +255,32 @@ def send_to_pipe(self, message: str, timeout_secs: Optional[float] = None) -> bo
...
+            # If the call raised an exception, this method will raise the same exception.
+            log.error("pipe sender exception: %s", e)
 
         return False
 
     def stop(self):
         log.debug("Stopping pipe")
         self.pipe_running = False
         self.send_to_pipe(self.MESSAGE_TO_IGNORE)
+        self.__thread_pool.shutdown(wait=True, cancel_futures=True)
```
zas, line 283: `cancel_futures` was added in Python 3.9 (see https://docs.python.org/3/library/concurrent.futures.html#concurrent.futures.Executor.shutdown)

picard/util/pipe.py
```diff
@@ -346,7 +340,6 @@ def _reader(self) -> str:
             try:
                 with open(self.path, 'r') as fifo:
                     response = fifo.read()
-            except FileNotFoundError:
```
zas, line 349: Why is this removed?

phw: By accident, will fix

picard/util/pipe.py
phw: https://github.com/metabrainz/picard/pull/2252/commits/7a1df3a20dcf203357dc15177b494d61b2ef93f0

phw: See https://github.com/metabrainz/picard/pull/2252/commits/7a3ca59ba12048b683e3aa3fb38db6f71d1b41cc

phw: See 5b61c2d9077fa3466df9b71b5e02f37b590fc1e8

zas (approved):
Tested, worked as expected for me
Simpler, and likely more reliable

phw:
> Simpler, and likely more reliable

I hope so :) But test results are the same for me.

PR #2254, opened 2023-07-06
PICARD-2680: Modal first use and save confirmation dialogs

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2680

When starting Picard or saving the new warning dialogs might open dislocated from the main window and on some system even show up behind the main window.

# Solution

Make the dialogs application modal and transient to the main window. This is how most of Picards dialogs are being opened already, specifically those message boxes.


rdswift (approved):
Good catch!  The changes look fine to me.

zas (approved):
LGTM

PR #2257, opened 2023-07-16
PICARD-2011: Avoid "RuntimeError: dictionary changed size during iteration" errors

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2011

The `Metadata`  class is not thread safe. If it is accessed in a thread this can lead to errors. E.g. the metadatabox can throw "RuntimeError: dictionary changed size during iteration" when iterating over the metadata while another thread changes it.

Similar `ConfigSection.as_dict()` can throw if `Object.registry` changes. 

# Solution

- Make `Metadata` thread safe by introducing a read / write lock
- Have `ConfigSection.as_dict()` operate on a copy of `Object.registry`

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -89,7 +89,11 @@ class ConfigSection(QtCore.QObject):
         return self.__qt_config.contains(self.key(name))
 
     def as_dict(self):
-        return {key: self[key] for section, key in Option.registry if section == self.__name}
+        return {
+            key: self[key] for section, key
+            in list(Option.registry)
+            if section == self.__name
+        }
 
     def remove(self, name):
         key = self.key(name)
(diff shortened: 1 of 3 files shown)


phw:
@zas This needs a bit more testing. But could you already look over this and test yourself as well?

This is the second time I approach making Metadata thread safe. In my first attempt I was fighting with dead-locks, but this time I think I got it correct.

phw:
From all that I can tell this is working fine. I would appreciate some additional testing, though.

phw:
> I tested and didn't notice any problem (though we lack a reliable mean to reproduce this issue).

Indeed. But at least I could repoduce the issue a couple of times (not intentionally, though), but never with this patch.



> The patch looks safe as the code is mainly about wrapping existing code with locks.

The important part is that no read lock gets created in code that has the write lock. That's why the internal methods `_get`  and `_del` are needed. I think I got this part right this time :)

PR #2259, opened 2023-07-24
PICARD-2688: Show command line help messages in dialog on Windows

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2688

On Windows if there is no console attached to the GUI application Picard can crash when run with "-h".
# Solution

Show the output in a dialog on Windows in case there is no console attached to `sys.stdout`. This applies to output in case of call with wrong arguments or explicit usage of the `-h`, `-v`, `-e HELP` in related switches.

I tested this on Linux and Windows. No change in behavior on Linux, as expected. On Windows it now behaves differently depending on the build and scenario. I tested with  `-h`, `-v`, `-e HELP` and with an invalid argument such as `--foo`.

A straightforward way to test this behavior from source tree is to start it with `pythonw.exe` instead of `python.exe`. `pythonw.exe` is Python build as a windowed application, so it does not show a terminal on launch.

Launching `python.exe tagger.py -h` from terminal gives the known output:

grafik

The same with  `pythonw.exe tagger.py -h` now instead opens the help in a dialog:

grafik

Interesting is the behavior for the build packages. The portable version, which previously crashed when passing `-h`, will now display the help in a dialog as expected.

For the `picard.exe` used in the installer package we have the workaround from https://github.com/metabrainz/picard/blob/master/scripts/pyinstaller/win-startup-hook.py#L28-L34 to redirect output to the console if it was launched from a console. This means

- If `picard.exe -h` gets started from Windows GUI (no console) output will be shown in a dialog. This scenario crashed before this change.
- If started from a console window output will be text output on the console, the same as before

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -153,7 +153,10 @@ from picard.webservice.api_helpers import (
 
 import picard.resources  # noqa: F401 # pylint: disable=unused-import
 
-from picard.ui import theme
+from picard.ui import (
+    FONT_FAMILY_MONOSPACE,
+    theme,
+)
 from picard.ui.mainwindow import MainWindow
 from picard.ui.searchdialog.album import AlbumSearchDialog
 from picard.ui.searchdialog.artist import ArtistSearchDialog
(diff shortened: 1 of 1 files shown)


rdswift (approved):
Looks fine to me.  Nice improvement.

PR #2260, opened 2023-07-24
PICARD-2689: Render the actually dragged element instead of the first selected

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Fixes issues with an empty rectangle being shown during dragging of items if the first selected item is not visible.

# Problem

* JIRA ticket (_optional_): PICARD-2689

https://community.metabrainz.org/t/possible-bug-found-dragn-dropping-items-from-the-right-pane-results-in-a-white-box-if-they-are-outside-the-visible-part/646711

# Solution

Instead of taking the first selected element take the currently active one. The item becomes active when clicking on it while starting the drag. This is probably even more intuitive for the user, as it shows the actual dragged element.

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -708,8 +708,9 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         if items:
             drag = QtGui.QDrag(self)
             drag.setMimeData(self.mimeData(items))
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2261, opened 2023-07-24
Picard-2685:  Preserve calculated from audio file tags when moving files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2685

1. Have files loaded matched to a track
2. Generate AcoustID fingerprint or ReplayGain tags (with the replaygain2 plugin)
3. Move the file to another track or unmatched files

Result: The AcoustID fingerprint or ReplayGain tags get removed

Expected result: As these tags are generated from audio content it would be expected that they move with the file instead of being replaced by the new track's data.

For the `acoustid_id` tag this is already handled.

# Solution

Preserve calculated AcoustID fingerprint and ReplayGain tags set on a file
object when that file object gets moved between tracks.

This generalized the mechanism already in place for `acoustid_id`  and extends it to `acoustid_fingerprint` and all the ReplayGain tags.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -98,7 +98,10 @@ from picard.util.filenaming import (
 )
 from picard.util.preservedtags import PreservedTags
 from picard.util.scripttofilename import script_to_filename_with_metadata
-from picard.util.tags import PRESERVED_TAGS
+from picard.util.tags import (
+    CALCULATED_TAGS,
+    PRESERVED_TAGS,
+)
 
 from picard.ui.item import Item
 
@@ -305,9 +308,14 @@ class File(QtCore.QObject, Item):
             to_metadata[info] = from_metadata[info]
 
     def copy_metadata(self, metadata, preserve_deleted=True):
-        acoustid = self.metadata["acoustid_id"]
         saved_metadata = {}
 
+        # Keep current value for special tags that got calculated from audio content
+        for tag in CALCULATED_TAGS:
+            if tag not in metadata.deleted_tags and self.metadata[tag]:
+                saved_metadata[tag] = self.metadata[tag]
+
+        # Keep original values of preserved tags
         preserved_tags = PreservedTags()
         for tag, values in self.orig_metadata.rawitems():
             if tag in preserved_tags or tag in PRESERVED_TAGS:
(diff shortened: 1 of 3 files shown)


zas (approved):
Good patch, nice improvement

picard/file.py
```diff
@@ -305,9 +308,14 @@ def _copy_file_info_tags(self, to_metadata, from_metadata):
...
 
     def copy_metadata(self, metadata, preserve_deleted=True):
-        acoustid = self.metadata["acoustid_id"]
         saved_metadata = {}
 
+        # Keep current value for special tags that got calculated from audio content
+        for tag in CALCULATED_TAGS:
+            if tag not in metadata.deleted_tags and self.metadata[tag]:
+                saved_metadata[tag] = self.metadata[tag]
+
```
zas, line 317: I wonder if this should not be after next loop instead

zas (changes requested):

picard/file.py
zas: Actually that looks correct as is

zas (approved):

picard/file.py
phw: > Actually that looks correct as is

@zas Looking at it again there is actually a functional difference in the case you put one of the tags, let's say `acoustid_id`  into preserved tags also.

**Previous behavior:**

1. The current value of `acoustid_id` from `file.metadata` gets saved
2. Due to preserved tags the value of `acoustid_id` from `file.orig_metadata`  gets saved
3. Metadata update gets applied
4. `acoustid_id`  value from `file.orig_metadata` gets restored
5. `acoustid_id`  value from `file.metadata` gets restored

Result: The value from `file.metadata` is kept

**New behavior:**

1. The current value of `acoustid_id` from `file.metadata` gets saved
2. Due to preserved tags the value of `acoustid_id` from `file.orig_metadata`  gets saved, overwriting the previously saved value
3. Metadata update gets applied
4. Saved `acoustid_id`  value `file.orig_metadata` gets restored

Result: The value from `file.orig_metadata` is restored

It is really a special case and only applies if the tag has a original value and is also put into the preserved tags configuration.

I'm actually not sure what is expected here. The old behavior kind of disabled the kept tags configuration. So maybe the new behavior is closer to expectation?

zas: @phw yes, I think new behavior is better, so let's keep it as is.

PR #2262, opened 2023-07-28
PICARD-2694: Clear instance of SingletonDialog if destroyed

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2694

If the ScriptingDocumentationDialog gets closed not explicitly but because its parent (the OptionsDialog) gets closed and destroys child widgets then reopening it again crashes Picard.

# Solution

Clear the instance of `SingletonDialog` on `destroyed` instead of `finished`

That fixes issues where the dialog does not get closed as a dialog but the widget just get cleared because the parent gets removed.

### DIFF
--- a/picard/ui/__init__.py
+++ b/picard/ui/__init__.py
@@ -148,7 +148,7 @@ class SingletonDialog:
     def get_instance(cls, *args, **kwargs):
         if not cls._instance:
             cls._instance = cls(*args, **kwargs)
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2263, opened 2023-07-28
PICARD-2693: Do not continue Win installation if uninstall got cancelled

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2693

On Windows when installing Picard with the installer package we trigger the uninstaller in order to completely remove the previous version. This avoids issues with leftover files, especially DLLs for dependencies, from previous install causing version conflicts.

However, it was possible for the user to cancel the uninstall and continue with the installation anyway.

# Solution

When the user cancels the uninstall also abort the installer. Show a prompt informing the user.

Note: As an alternative I had thought about making the uninstaller run automatically non-interactive. However, NSIS does not support runnning non-interactive with UI. If you launch the uninstaller in automatic mode it just would run without UI, causing the installer Window to hang and wait for it to finish without visual feedback. Also I think interactive use is more user friendly overall.


phw:
@zas I tested this on Windows. I know you cannot test, but please have a look in general if you agree with the concept and for stupid things I might have overlooked.

zas (approved):
LGTM

PR #2264, opened 2023-07-28
PICARD-2695: Fix crash in script editor on some unicode escape tooltips

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2695

For some code points, e.g. \u0000, unicodedata.name does not return a name but raises a ValueError.

# Solution

Catch the ValueError and as an alternative just output the unicode code point in the format "U+0000" (as it is common).


zas (approved):
LGTM

PR #2266, opened 2023-07-29
PICARD-1815: Add option to notify user at startup if plugin updates are available

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds an option to notify the user when Picard is started if there are updates available for installed plugins.

# Problem

The only way that a user will know if there is an update available for an installed plugin is by viewing the plugins page in the option settings.

* JIRA ticket (_optional_): PICARD-1815

# Solution

Add a check for new updates available for installed plugins and display a pop-up messagebox if updates are available.  The check is performed during Picard startup, just after the check for program updates.  Add a new option setting to enable or disable this check and notification.

# Action


picard/pluginmanager.py
```diff
@@ -445,3 +447,22 @@ def _plugins_json_loaded(self, response, reply, error, callback=None):
...
         return True
+
+    def check_update(self, parent=None):
+        if self.available_plugins is None:
+            self.query_available_plugins(partial(self._display_update, parent))
+        else:
+            self._display_update(parent)
+
+    def _display_update(self, parent):
+        test_dictionary = {p.module_name: p.version for p in self.available_plugins}
```
zas, line 458: This line should be moved inside the `if` branch below.

Also the name isn't great, what about `available_versions` or something like that?

zas (changes requested):

picard/pluginmanager.py
rdswift: Good catch.  No need to assign the variable unless it is going to be used.

zas (dismissed):
LGTM

picard/pluginmanager.py
```diff
@@ -445,3 +447,22 @@ def _plugins_json_loaded(self, response, reply, error, callback=None):
...
+            available_versions = {p.module_name: p.version for p in self.available_plugins}
+            for plugin in self.plugins:
+                if (plugin.module_name in available_versions and available_versions[plugin.module_name] > plugin.version):
+                    QMessageBox.information(
+                        parent,
+                        _("Picard Plugins Update"),
+                        _("There are updates available for your currently installed plugins."),
+                        QMessageBox.StandardButton.Ok, QMessageBox.StandardButton.Ok
+                    )
+                    break
```
zas, line 468: Since `_display_update()` is (and will be) only used in `check_update()`, I think we can revamp this a bit, see https://github.com/rdswift/picard/pull/17

rdswift: Merged.  Thanks.

zas (approved):
LGTM

phw:
@rdswift @zas  After using this a bit I must say I find the current behavior a bit annoying over time. I think a few things would make this better:

1. The dialog should provide the option to open the plugins page directly, so the user can actually act on the message
2. We probably should have an "update all plugins" button on the plugin options


Alternatively the dialog could allow updating the plugins right away. But that probably requires it to show exactly which plugins would be updated. What do you think?

rdswift:
> I think a few things would make this better:
> 
> 1. The dialog should provide the option to open the plugins page directly, so the user can actually act on the message
> 2. We probably should have an "update all plugins" button on the plugin options

I like both of these suggestions.

> Alternatively the dialog could allow updating the plugins right away. But that probably requires it to show exactly which plugins would be updated. What do you think?

I can't explain why, but for some reason that behavior makes me uncomfortable, however I can see why it might be desirable.  I think I might have subconsciously avoided doing that because of the added complexity required to unload and reload the plugins, and that there may be a situation where the user doesn't want to update one of the plugins for some reason.  By having the updates processed through the existing plugins page, the user maintains full control of what is happening.

This might be over-complicating things, but perhaps each plugin on the plugins page could include checkboxes to indicate whether they should be automatically updated, and whether or not they should be included in the "Updates Available" notification?

In any event, it should be easy enough to show the list of plugins to be updated rather than just stop iterating after finding the first one, and that is good information to display no matter which way we choose to proceed.

rdswift:
I added PICARD-2715 to capture the suggestion and subsequent discussion, and to use for any changes decided.

phw:
> I can't explain why, but for some reason that behavior makes me uncomfortable, however I can see why it might be desirable. I think I might have subconsciously avoided doing that because of the added complexity required to unload and reload the plugins,

@rdswift That's actually exactly how I felt about this, and why I first proposed linking to the plugin options page. I think this just offers more flexibility. We can make performing updates easier on this page. Another thought was that maybe plugins with updates available should be listed first. Or a filter.

Thanks for adding this ticket.

PR #2269, opened 2023-08-05
PICARD-2696: avoid deleting dropped files

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -136,6 +136,7 @@ class CoverArtThumbnail(ActiveLabel):
         event.acceptProposedAction()
 
     def dropEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
         accepted = False
         # Chromium includes the actual data of the dragged image in the drop event. This
         # is useful for Google Images, where the url links to the page that contains the image
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -695,8 +695,8 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         return ["text/uri-list", "application/picard.album-list"]
 
     def dragEnterEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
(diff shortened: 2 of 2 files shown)


picard/ui/coverartbox.py
```diff
@@ -136,6 +136,7 @@ def dragMoveEvent(event):
         event.acceptProposedAction()
 
     def dropEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
```
phw, line 139: Does this work without breaking normal move actions inside Picard? My idea would have been to only do this if the drop originates from outside.

phw (commented):
Should work. Only open question seey comment. I'll test this when I'm back on the next weekend.

picard/ui/coverartbox.py
Sophist-UK: As I commented previously elsewhere, one `proposedAction` is do nothing or Ignore. IMO you should NOT change this to a CopyAction.

phw: Picard only accepts copy and move actions. There is nothing else at that part of code.

picard/ui/itemviews.py
```diff
@@ -695,8 +695,8 @@ def mimeTypes(self):
         return ["text/uri-list", "application/picard.album-list"]
 
     def dragEnterEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
```
phw, line 698: I wonder why in the reported case `event.mimeData().hasUrls()` would be false. After all it is files being dropped. 

It all looks like the reported issue was supposed to be handled, but for some reason it does not work. Really needs deeper debugging.

phw:
@zas It needed a bit more handling, see #2270.

The main culprit was that in our case the parent implementation of `dropEvent` always did do a "acceptProposedAction", so no matter how we changed the actual action it always accepted the move.

PR #2270, opened 2023-08-12
PICARD-2696: Fix file deletion when shift-dropping files on Windows

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2696

When dropping while pressing shift on Windows the files get sent as a "move" event to the application. Accepting this results in file deletion. Picard must always handle such events as "copy" to avoid this.

This replaces #2269

# Solution

For itemview override the acceptance done in `QAbstractItemView.dropEvent` and only do a "copy" if the drop originates from outside Picard (no source widget) or is providing URLs.

For the cover art box generally handle all drops as copy.

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -127,13 +127,9 @@ class CoverArtThumbnail(ActiveLabel):
         else:
             return True
 
-    @staticmethod
-    def dragEnterEvent(event):
-        event.acceptProposedAction()
-
-    @staticmethod
-    def dragMoveEvent(event):
-        event.acceptProposedAction()
+    def dragEnterEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
+        event.accept()
 
     def dropEvent(self, event):
         accepted = False
@@ -170,7 +166,8 @@ class CoverArtThumbnail(ActiveLabel):
                 self.image_dropped.emit(QtCore.QUrl(''), dropped_data)
 
         if accepted:
-            event.acceptProposedAction()
+            event.setDropAction(QtCore.Qt.DropAction.CopyAction)
+            event.accept()
 
     def scaled(self, *dimensions):
         return (round(self.pixel_ratio * dimension) for dimension in dimensions)
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2007 Robert Kaye
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Hendrik van Antwerpen
-# Copyright (C) 2008-2011, 2014-2015, 2018-2022 Philipp Wolfer
+# Copyright (C) 2008-2011, 2014-2015, 2018-2023 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2009 Nikolai Prokoschenko
 # Copyright (C) 2011 Tim Blechmann
@@ -695,7 +695,7 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         return ["text/uri-list", "application/picard.album-list"]
 
     def dragEnterEvent(self, event):
-        if event.mimeData().hasUrls():
+        if not event.source() or event.mimeData().hasUrls():
             event.setDropAction(QtCore.Qt.DropAction.CopyAction)
             event.accept()
         else:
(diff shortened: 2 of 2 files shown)


zas (approved):
LGTM

picard/ui/coverartbox.py
```diff
@@ -127,13 +127,9 @@ def __eq__(self, other):
...
 
-    @staticmethod
-    def dragEnterEvent(event):
-        event.acceptProposedAction()
-
-    @staticmethod
-    def dragMoveEvent(event):
-        event.acceptProposedAction()
+    def dragEnterEvent(self, event):
+        event.setDropAction(QtCore.Qt.DropAction.CopyAction)
```
Sophist-UK, line 131: According to the QT documentation at https://doc.qt.io/qt-6/qt.html#DropAction-enum existing drop actions can be:

Constant | Value | Description
-- | -- | --
Qt::CopyAction | 0x1 | Copy the data to the target.
Qt::MoveAction | 0x2 | Move the data from the source to the target.
Qt::LinkAction | 0x4 | Create a link from the source to the target.
Qt::ActionMask | 0xff |  
Qt::IgnoreAction | 0x0 | Ignore the action (do nothing with the data).
Qt::TargetMoveAction | 0x8002 | On Windows, this value is used when the ownership of the D&D data should be taken over by the target application, i.e., the source application should not delete the data. On X11 this value is used to do a move. TargetMoveAction is not used on the Mac.

I still believe that we should check the existing DropAction and only change it if it is not `Qt::IgnoreAction` on the basis that changing a MoveAction/LinkAction to a CopyAction seems reasonable, but (just in case we ever get one) changing an IgnoreAction to a CopyAction seems wrong. (I am not sure how an IgnoreAction might happen, but I guess that starting a Move and then aborting it would result in an IgnoreAction in order to inform the receiving app that the Drag&Drop which they were already told to expect is no longer going to happen. If this is the basis for an IgnoreAction, we would not IMO want to change it to a CopyAction.)

picard/ui/coverartbox.py
```diff
@@ -170,7 +166,8 @@ def dropEvent(self, event):
                 self.image_dropped.emit(QtCore.QUrl(''), dropped_data)
 
         if accepted:
-            event.acceptProposedAction()
+            event.setDropAction(QtCore.Qt.DropAction.CopyAction)
```
Sophist-UK, line 169: Ditto

picard/ui/itemviews.py
```diff
@@ -767,7 +767,13 @@ def dropEvent(self, event):
         # assigned to the same track.
         if event.keyboardModifiers() == QtCore.Qt.KeyboardModifier.AltModifier:
             self._move_to_multi_tracks = False
-        return QtWidgets.QTreeView.dropEvent(self, event)
+        QtWidgets.QTreeView.dropEvent(self, event)
+        # The parent dropEvent implementation automatically accepts the proposed
+        # action. Override this, for external drops we never support move (which
+        # can result in file deletion, e.g. on Windows).
+        if event.isAccepted() and (not event.source() or event.mimeData().hasUrls()):
+            event.setDropAction(QtCore.Qt.DropAction.CopyAction)
```
Sophist-UK, line 775: Ditto

picard/ui/coverartbox.py
phw: Maybe we should handle this event, but before we change anything in this regard I would like to know how this event can actually be triggered and how it is supposed to be handled. The Qt5 documentation is pretty sparse here.

Looking through Qt's code samples it seems this action sometimes gets handled by not performing any action but accepting the action, e.g. https://github.com/qt/qtbase/blob/33f7eb44ff33e0f8f516a9e8076dfdb6098cfc6f/tests/manual/examples/widgets/itemviews/puzzle/piecesmodel.cpp#L108-L109

But it's all really vague. If you have more information on how this event is used and under which circumstances it triggers we can investigate how to handle it in the different scenarios where drops are supported.

For the itemviews it likely is not a problem anyway, as anything other then CopyAction and MoveAction is not accepted.

phw: I added https://github.com/metabrainz/picard/pull/2271 with how I think it should be handled based on the Qt examples. But I don't see how to verify this is the correct way to handle the ignore message. I very much would like to have some kind of reproducer for this before merging that code.

PR #2272, opened 2023-08-13
PICARD-2698: Fix high CPU usage in Windows pipe implementation

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2698

On Windows Picard is causing constant high CPU usage if idle. This is caused by the pipe server loop. The previous implementation had the "pipe server" (Picard's main instance) poll for the pipe getting created. The connecting "client" then created the named pipe, sent a message and disconnected. 

# Solution

Instead for the main instance to constantly poll for a pipe getting created have the main instance own the pipe and wait for connecting client. `ConnectNamedPipe` will wait until the client connects and only then the message will be read. This is somewhat similar to the Unix pipe implementation.

One caveat is that once the client disconnects the pipe gets closed (see also https://stackoverflow.com/questions/19636548/named-pipe-232-the-pipe-is-being-closed). That means after receiving the message the server needs to recreate the named pipe.

# Action

My knowledge about Windows named pipes is very limited. But I'm a bit concerned that this delay where the pipe stops working and gets recreated could cause issues if multiple instances get launched trying to send data in parallel to the pipe. It might be possible that in the short timeframe where the main  instance looses the pipe another started instance might take over the role of the main instance instead of sending to Picard.

In my tests I could not reproduce this scenario, though. I'm also not sure how this could be effectively be prevented.


test/test_util_pipe.py
```diff
@@ -68,27 +59,22 @@ def test_pipe_protocol(self):
...
             except concurrent.futures._base.TimeoutError:
-                pipe_listener_handler.read_from_pipe()
+                pass
+            finally:
+                pipe_handler.stop()
+                __pool.shutdown()
 
             self.assertEqual(res, message,
                              "Data is sent and read correctly")
+
```
zas, line 80: This line makes code checker unhappy

zas (changes requested):

test/test_util_pipe.py
phw: @zas Fixed

picard/util/pipe.py
```diff
@@ -392,42 +392,54 @@ def __init__(self, app_name: str, app_version: str, args: Optional[Iterable[str]
...
-
+        self.__create_pipe()
         self._remove_temp_attributes()
 
+    def __create_pipe(self):
+        try:
+            self.__pipe = win32pipe.CreateNamedPipe(
+                self.path,
+                win32pipe.PIPE_ACCESS_DUPLEX,
+                win32pipe.PIPE_TYPE_MESSAGE | win32pipe.PIPE_WAIT,
```
zas, line 403: See the note at the end of https://learn.microsoft.com/en-us/windows/win32/ipc/named-pipe-type-read-and-wait-modes

phw: Yes, that's correct. But the other reason why blocking mode is used here is just because it is the best way to handle it. With the blocking wait the thread waits until a client actually sends a message. That way the thread is just idle and does not consume CPU time until a message comes in.

zas (approved):
LGTM

phw:
@rdswift Could you test if this works for you on Windows as well?

rdswift:
> @rdswift Could you test if this works for you on Windows as well?

I downloaded and installed the build artifact, and the CPU usage dropped from 15% with the previous version to 0% with the new version.  Looks good to me.  Were there any specific things you wanted me to try?

phw:
@rdswift Sounds great. Does the command passing (with the `-e` parameter) and the single instance launch both still work properly?

phw:
Thanks a lot for testing

> The only thing that I noticed was starting a second instance (from the command line in powershell) with no arguments just exited immediately (as expected), but didn't make the currently running instance the active window. I never checked this before, so I don't know if that is the expected result. In any event, certainly not a show-stopper in my opinion.

That was like this before. I also once stumbled over this. Not sure, maybe the default should indeed be if there is an instance already but no command was given that the `SHOW` command gets sent.

rdswift:
> maybe the default should indeed be if there is an instance already but no command was given that the `SHOW` command gets sent.

That's a good idea.  I just added ticket PICARD-2717 to capture this.

PR #2273, opened 2023-08-13
PICARD-2710: Fix no loading files on low CPU systems

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2710

If Picard is started on a system with 2 or less CPU cores then the main thread pool only has 1 or 2 threads available.

In the current implementation of the single instance mode one thread is required to run the pipe server, one is required for processing incoming commands.

# Solution

Ensure the thread pool provides at least 3 threads. This way Picard is able to process tasks like file load / save.

Additionally this PR changes the behavior of the run_commands loop. It avoids the run_commands loop constantly looping in the background. Instead the thread terminates if there are no more commands to handle and gets started again on demand if there are new commands to process.

For the majority of users the command feature probably is unused, and even those using it do so only part of the time when using Picard.

This helps with both thread usage PICARD-2710 and also with CPU consumption
PICARD-2698 (even though this loop was rather light in this regard).

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -269,17 +269,11 @@ class Tagger(QtWidgets.QApplication):
         if picard_args.debug or "PICARD_DEBUG" in os.environ:
             self.set_log_level(logging.DEBUG)
 
-        # FIXME: Figure out what's wrong with QThreadPool.globalInstance().
-        # It's a valid reference, but its start() method doesn't work.
+        # Main thread pool used for most background tasks
         self.thread_pool = QtCore.QThreadPool(self)
-
-        self.pipe_handler = pipe_handler
-
-        if self.pipe_handler:
-            self.pipe_handler.pipe_running = True
-            thread.run_task(self.pipe_server, self._pipe_server_finished)
-
-        self._init_remote_commands()
+        # Two threads are needed for the pipe handler and command processing.
+        # At least one thread is required to run other Picard background tasks.
+        self.thread_pool.setMaxThreadCount(max(3, QtCore.QThread.idealThreadCount()))
 
         # Provide a separate thread pool for operations that should not be
         # delayed by longer background processing tasks, e.g. because the user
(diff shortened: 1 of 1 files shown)


rdswift (approved):
This looks good to me.  Nice fix.

Sophist-UK:
The whole idea that the number of threads should be based on the number of cores / hyperthreads is **wrong**. This idea only makes sense when threads are used A) by a language that supports simultaneous use of multiple cores (which Python cannot do because of the GIL - and I can reasonably claim to be a genuine expert on the Python GIL), and B) when the threads are used for CPU intensive tasks rather than for blocking I/O (and Picard uses them for blocking I/O and not CPU intensive tasks).

The whole purpose of Picard using threads is to allow for several blocking I/O tasks and / or tasks like pipelines or drop target which need to wait for something to happen.

IMO for Picard's use case, the number of threads in the pool should either be fixed based on the maximum number of simultaneous tasks, or should be variable based on the I/O capacity of the system it is running on (rather than the CPU parallelism).

zas (approved):
LGTM

phw:
@Sophist-UK Fully agree. But this is pretty much out of scope for this PR, which only is supposed to fix the issue with all threads of the pool being occupied.

We already had this discussion in #2172 . I very much agree with you that it makes sense to limit the number of worker threads and not scale up with CPU. The tricky part is however to decide for a good thread count. My testing in https://github.com/metabrainz/picard/pull/2172#issuecomment-137462725 gave no clear picture, with depending on the setup having much worse or much better performance with higher no. of threads.

Sophist-UK:
I just read #2172 again, and the final part of the discussion said that we were doing the file I/o and then the subsequent CPU intensive processing on the same thread. I will add another comment there about this.

phw:
@Sophist-UK Please comment with this on a relevant ticket. Not sure if there is one already, if not, please create a new one. Comments on closed PRs won't get found in the future.

Sophist-UK:
OK

PR #2278, opened 2023-08-15
Disable plugin updates by default

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

The new plugin update feature is something that I think we can ship in the 2.9.1 release. But as we want to get this release out soon due to the important bug fixes I would feel better if we could make this disabled for now by default.

That way we can ship the feature to interested users and hopefully can get feedback on it, but don't cause unexpected interruption for existing users.

Once we have extended the feature and improved the UI we can consider making it the default.


zas (approved):
Yes, it makes sense.

PR #2279, opened 2023-08-15
PICARD-2715: Show plugins with updates and add button to open plugin options window

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds list of plugin updates available and a button to open the Options -> Plugins screen directly from the plugin update notification dialog

# Problem

The plugin updates available notification dialog did not provide any details or allow easy updating.

* JIRA ticket (_optional_): PICARD-2715

# Solution

Add a list of plugin updates available and a button to open the Options -> Plugins screen directly from the plugin update notification dialog.  Examples:

image

image

Note that the displayed list is limited to a maximum of 3 plugins, with a note indicating if there are more.

# Action

None.

### DIFF
--- a/picard/pluginmanager.py
+++ b/picard/pluginmanager.py
@@ -35,9 +35,13 @@ import zipfile
 import zipimport
 
 from PyQt5 import QtCore
-from PyQt5.QtWidgets import QMessageBox
+from PyQt5.QtWidgets import (
+    QCheckBox,
+    QMessageBox,
+)
 
 from picard import log
+from picard.config import get_config
 from picard.const import (
     PLUGINS_API,
     USER_PLUGIN_DIR,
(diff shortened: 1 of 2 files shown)


phw (dismissed):
Nice and clean. I think this makes the update check really more useful.

rdswift:
Do we still want to default this plugin check to disabled?

Should I add a checkbox to the dialog to disable the checks, similarly to the new user / first run / save warning dialogs?

phw:
I think we should keep it disabled for now. We already introduced multiple new popups. I think it is better to not show another one unasked on a minor update.

rdswift:
I added to option to disable further checks, and refactored a bit.  The dialog now looks like:

image

phw (dismissed):

picard/pluginmanager.py
```diff
@@ -449,26 +482,55 @@ def enabled(self, name):
...
+
+                msg = PluginUpdatesDialog(parent, f'{header}<ul>{plugin_list}</ul>{extra_plugins}{footer}')
+
+                show_options_page, perform_check = msg.show()
+                if parent:
+                    config = get_config()
+                    config.setting['check_for_plugin_updates'] = perform_check
+
+                    if show_options_page:
+                        parent.show_plugins_options_page()
```
rdswift, line 533: Should this be changed to:
```suggestion
                if parent:
                config = get_config()
                config.setting['check_for_plugin_updates'] = perform_check

                if parent and show_options_page:
                    parent.show_plugins_options_page()
```

rdswift: Changes tested and implemented.

zas: Yes I think so. The setting doesn't depend on `parent`.

zas (approved):

phw (approved):

PR #2281, opened 2023-08-16
PICARD-2720: Add .desktop file action to explicitly open new window

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2720

XDG desktop entries allow specifying additional actions as alternative ways to launch an application.

We can use this to add a new option to explicitly open a new standalone window / instance of Picard.

See https://specifications.freedesktop.org/desktop-entry-spec/desktop-entry-spec-latest.html#extra-actions

# Solution
Add a new action to launch `picard --stand-alone-instance`.

grafik


zas (approved):
LGTM

PR #2283, opened 2023-08-17
PICARD-2712: Fix sanitize date removing 0 values

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2712

Sanitize date cleared zero values in between, such turning "YYYY-00-DD" into "YYYY-DD", which would be a completely different date.

# Solution

Fix this by keeping empty values if they are not the last element.

It changes the behavior for some strings with e.g. double slashes, as the empty string in between those is now considered a valid value and treated as `0`.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -312,18 +312,23 @@ def format_time(ms, display_zero=False):
 def sanitize_date(datestr):
     """Sanitize date format.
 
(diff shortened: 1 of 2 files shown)


zas (approved):
Good improvement

PR #2284, opened 2023-08-18
PICARD-2724: Fix potential crash in artist alias translation

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2724

Fix potential crash in artist alias translation

This crash requires very specific circumstances:

- Option `translate_artist_names` must be enabled
- Option `translate_artist_names_script_exception` must be set
- The track search dialog must be opened on a file
- The search result must contain artist aliases
- An artist alias must:
  - be in a script not catched by `translate_artist_names_script_exception` 
  - be set to "primary"
  - have a "locale" set
  - have no type set

In this case the result does not contain the `type-id` key but Picard still tries to read it.

# Solution

Make the code around reading the alias node more robust:

- handle non existent `type-id`
- handle non-existent `primary` (should not happen)
- handle another case where `primary` is set but `locale` is `null`. This actually should not happen as the MB server does not allow setting primary without locale. But it might well be possible that some data exists with this condition

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -246,11 +246,11 @@ def _locales_from_aliases(aliases):
     full_locales = {}
     root_locales = {}
     for alias in aliases:
-        if not alias['primary']:
+        if not alias.get('primary'):
             continue
-        if 'locale' not in alias:
+        full_locale = alias.get('locale')
+        if not full_locale:
             continue
-        full_locale = alias['locale']
         root_locale = full_locale.split('_')[0]
         full_parts = []
         root_parts = []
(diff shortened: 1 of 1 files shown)


phw:
Because I edited the [MusicBrainz Test Artist](https://musicbrainz.org/artist/7e84f845-ac16-41fe-9ff8-df12eb32af55) this can currently be reproduced by:

- In Options > Metadaa
  - Enable `translate_artist_names` and `translate_artist_names_script_exception`
  - For `translate_artist_names_script_exception` select a single script that is not latin
- Right click a file and select "search for similar titles"
- Depending on the searched data the crash might happen. If it doesn't enter as the search term `track:(track:(😀 grinning face Unicode: U+1F600, UTF-8: F0 9F 98 80) artist:(TEST) qdur:(146)) artist:(TEST) qdur:(146)`

zas (approved):
Nice patch, LGTM

PR #2285, opened 2023-08-19
Add note about contributing to user documentation

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**:  Adds a note about contributing to the user documentation in the CONTRIBUTING section.

# Problem

This helps inform potential contributors to the project that there is also an opportunity to contribute to the user documentation.

# Solution

Add a note about contributing to the user documentation in the CONTRIBUTING section.

# Action

None.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -74,6 +74,11 @@ We follow the "typical" GitHub workflow when contributing changes:
 See po/README.md for information about translations.
 
 
+## User Documentation
+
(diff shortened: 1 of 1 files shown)


phw (changes requested):
@rdswift Good idea, that was indeed missing. But I'd actually move this section up, directly after the "Translations" section.

The "Audio Metadata Specifications" is kind of an Appendix, let's keep this at the end.

rdswift:
> The "Audio Metadata Specifications" is kind of an Appendix, let's keep this at the end.

That makes sense.  The new documentation section has now been moved to immediately follow the translations section, as suggested.

PR #2287, opened 2023-08-21
PICARD-2700: fix accumulating series metadata on refresh

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2700

When refreshing series are added once again (and so on), because of the use of `Metadata.add()` over same tags.

# Solution

Call `Metadata.unset()` for series before adding first tag value.

This fix is limited to `*series` tags, but the issue is perhaps concerning more, the whole process of refreshing will need to be reviewed.

# Action

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -27,7 +27,7 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-from collections import namedtuple
+from types import SimpleNamespace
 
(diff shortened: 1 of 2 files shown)


phw (approved):
This looks clean. I like how this also improves existing code by replacing `TargetTypeFuncContext` with `SimpleNamespace` as well.

LGTM

PR #2291, opened 2023-08-24
PICARD-2690: move const translations

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2690

The `picard.const` module contains longs lists, most specifically for locales and written scripts.

This makes translating much harder, as half the strings to translate are from these lists, but they are essentially only used for very few UI elements.

Also it makes it difficult to assess how complete a translation of Picard is. A translations that translates all the UI texts except for the locale list has a low completion percentage on Weblate, but actually it is a very complete translation. Likewise a translation that solely translates the locales list shows with the same or higher completion, but actually the application can hardly be described as being localized.

# Solution

Move translations of the strings in `picard.const` into a separate gettext message domain "picard-constants".

I originally planned to move only the locales there, but with Babel it is easier to handle it by excluding a folder. And actually I think that fits well, as it also includes the scripts. Likewise I think it is no mistake to include the values in `picard/const/__init__.py` as well.

There is a new setup command `picard_regen_constants_pot_file`. We could also include a single meta command to update both, but for now I wanted to keep it separate (see below).

# Action

Of course we should keep existing translations. My plan was as follows:

- Once this has been merged I set up the Weblate component for it
- We leverage Weblate's capability to automatically translate strings from other components to fill the translations again
- Afterwards we can regenerate the main `picard.pot` (removing the strings from it)
- I'll also update the translation docs and unify generating both `picard.pot`  and `constants/constants.pot` with a single command

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -492,32 +492,26 @@ class picard_regen_appdata_pot_file(Command):
 
 
 _regen_pot_description = "Regenerate po/picard.pot, parsing source tree for new or updated strings"
+_regen_constants_pot_description = "Regenerate po/constants/constants.pot, parsing source tree for new or updated strings"
 try:
-    from babel import __version__ as babel_version
     from babel.messages import frontend as babel
-
-    def versiontuple(v):
-        return tuple(map(int, (v.split("."))))
-
-    # input_dirs are incorrectly handled in babel versions < 1.0
-    # https://web.archive.org/web/20150910064954/babel.edgewall.org/ticket/232
-    input_dirs_workaround = versiontuple(babel_version) < (1, 0, 0)
 
     class picard_regen_pot_file(babel.extract_messages):
         description = _regen_pot_description
 
         def initialize_options(self):
-            # cannot use super() with old-style parent class
-            babel.extract_messages.initialize_options(self)
+            super().initialize_options()
             self.output_file = 'po/picard.pot'
             self.input_dirs = 'picard'
-            if self.input_dirs and input_dirs_workaround:
-                self._input_dirs = self.input_dirs
+            self.ignore_dirs = ('const',)
 
-        def finalize_options(self):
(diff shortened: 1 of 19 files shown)


picard/pluginmanager.py
```diff
@@ -273,8 +275,11 @@ def _load_plugin_from_directory(self, name, plugindir):
         if zip_importer:
             name = module_name
             if not zip_importer.find_module(name):
-                error = _("Failed loading zipped plugin %r from %r")
-                self.plugin_error(name, error, name, zipfilename)
+                error = _('Failed loading zipped plugin "%(plugin)s" from "%(filename)s"')
```
zas, line 278: It looks like this PR wasn't based on `master` but on [phw:weblate-unnamed-variables-check](https://github.com/phw/picard/tree/weblate-unnamed-variables-check)

phw: Indeed. Fixed now

zas (approved):
It will ease translations for sure, good move.

setup.py
```diff
@@ -688,8 +685,9 @@ def _picard_get_locale_files():
     locales = []
     path_domain = {
         'po': 'picard',
-        os.path.join('po', 'countries'): 'picard-countries',
         os.path.join('po', 'attributes'): 'picard-attributes',
+        os.path.join('po', 'constants'): 'picard-constants',
+        os.path.join('po', 'countries'): 'picard-countries',
```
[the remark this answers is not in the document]
phw: Absolutely better readable. Done in 8428a17655aaf8fde947d18d49c86a86a56b44c5

rdswift (approved):
I don't think I have anything to add.  It looks good to me.

PR #2292, opened 2023-08-27
Use named parameters in get_timestamp() time formatting strings

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

The function `get_timestamp()` in `picard.util.time` uses parametered strings like `%.2dh %.2dm`. For translation it is hard to see for what purpose those are being used for and how to replace them. Also order cannot be easily changed (should this be necessary in the target language), even though this is unlikely in this specific case.

# Solution

Use named parameters like `%(hours).2dh %(minutes).2dm`

### DIFF
--- a/picard/util/time.py
+++ b/picard/util/time.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2021 Gabriel Ferreira
 # Copyright (C) 2021 Laurent Monin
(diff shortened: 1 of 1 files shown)


rdswift (approved):
This makes sense to me.

zas (approved):
LGTM

PR #2293, opened 2023-08-27
Fix bytes2human using localized strings.

phw:
This partially reverts 3d50184, but still ensures that the partial string in _(...) does not get extracted. Also use named parameter to make the usage of the string more clear.


zas (approved):
LGTM

PR #2294, opened 2023-08-29
Log SSL errors

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/webservice/__init__.py
+++ b/picard/webservice/__init__.py
@@ -43,7 +43,10 @@ from PyQt5 import (
     QtNetwork,
 )
 from PyQt5.QtCore import QUrl
-from PyQt5.QtNetwork import QNetworkRequest
+from PyQt5.QtNetwork import (
+    QNetworkRequest,
+    QSslError,
+)
 
 from picard import (
     PICARD_APP_NAME,
@@ -165,6 +168,11 @@ class WSRequest(QNetworkRequest):
             url.setQuery(query)
 
         super().__init__(url)
+
+        # To simulate an ssl error, uncomment following lines
+        # ssl = self.sslConfiguration()
+        # ssl.setCaCertificates(list())
+        # self.setSslConfiguration(ssl)
 
         # optional parameters
         self.parse_response_type = parse_response_type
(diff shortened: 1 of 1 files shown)


phw (commented):
That's a bit difficult to test in real. But wouldn't this lead to double logging, as we already log all errors in https://github.com/metabrainz/picard/blob/master/picard/webservice/__init__.py#L511 as well?

zas:
> That's a bit difficult to test in real. But wouldn't this lead to double logging, as we already log all errors in https://github.com/metabrainz/picard/blob/master/picard/webservice/__init__.py#L511 as well?

You can use the commented out code to trigger an error, but basically it provides more details than the existing message;

```
E: 12:25:28,378 webservice.ssl_errors:363: SSL error: The issuer certificate of a locally looked up certificate could not be found
E: 12:25:28,378 webservice._handle_reply:534: Network request error for https://picard.musicbrainz.org/api/v2/plugins/ -> SSL handshake failed (QT code 6, HTTP code 0)
E: 12:25:28,379 ui/mainwindow.set_statusbar_message:469: Error loading plugins list: SSL handshake failed
```

The existing message only says the SSL handshake failed, but the new message actually says why.

picard/webservice/__init__.py
```diff
@@ -341,6 +350,20 @@ def __init__(self, parent=None):
...
+        # This can also be used to ignore others if needed
+        ignored_errors = {
+            QSslError.NoError,
+        }
+        has_errors = False
+        for error in errors:
+            if error not in ignored_errors:
+                has_errors = True
+                log.error("SSL error: %s" % error.errorString())
+        if not has_errors:
```
Sophist-UK, line 364: Would it be better to remove any NoError entries from errors, then you can do `if not errors: return` and then loop and do `ignoreSslErrors()`?

zas: I'm reluctant to modify the passed list, and it wouldn't be much an improvement over current code.

phw (approved):

PR #2295, opened 2023-08-30
PICARD-2722: Avoid pipe thread respawning on exit on Windows

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2722

Closing Picard on Windows can lead to an exception when closing the pipe while also have the pipe server thread still running. The reason is a race condition when sending the close message to the pipe. As the pipe gets recreated after it was read from it could happen that the pipe was running again before the application closed.

# Solution

Only re-create the pipe if it was not marked as stopped. Also avoid exceptions when attempting to close the handle.

The second commit of this PR also solves a smaller problem, where on starting Picard under Windows there would always be a log output

```
E: 18:54:33,314 util/pipe.send_to_pipe:274: pipe sender exception: (2, 'CreateFile', 'The system cannot find the file specified.')
```

This was caused by the attempt to write to the pipe, which is done for probing whether a process already has the pipe or not.

### DIFF
--- a/picard/util/pipe.py
+++ b/picard/util/pipe.py
@@ -421,15 +421,23 @@ class WinPipe(AbstractPipe):
                 log.error('Error closing pipe', exc_info=True)
 
     def _sender(self, message: str) -> bool:
-        pipe = win32file.CreateFile(
-            self.path,
-            win32file.GENERIC_READ | win32file.GENERIC_WRITE,
-            self.__SHARE_MODE,
-            None,
-            win32file.OPEN_EXISTING,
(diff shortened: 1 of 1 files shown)


phw:
The issue is hard to reproduce on my system. I could not intentionally trigger it, but it happened from time to time. With the patch it hasn't happened so far.

zas (approved):
I can't test it, but it looks good.

PR #2300, opened 2023-09-06
Rename id to profile_id where applicable

zas:
`profile_id` is explicit.
`id()` is a built-in function, see https://docs.python.org/3.10/library/functions.html#id

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action


rdswift (approved):
This makes a lot of sense to me.  Looks good.  Thanks.

phw (approved):

PR #2306, opened 2023-09-08
Fixed regresson in script editor

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Revert use of qlistwidget_items for getting the script list, as this is a QComboBox and not QListWidget.


zas (approved):
Thanks for catching that, I totally misread this code before changing it.

phw:
Yep, it looked so similar. I was also looking for the "count()" / "item(i)" combo but missed that. looked so similar. Just found out while testing some parts.

PR #2307, opened 2023-09-09
PICARD-2751: Avoid deprecated importlib APIs, full Python 3.12 compatibility

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2751, PICARD-2354, PICARD-2355

Picard's plugin loading system must be updated to avoid long deprecated and inf Python 3.12 finally removed APIs of the importlib and zipimport modules.

This PR replaces #2134

# Solution

At the core the change is to use `find_spec` in combination with `exec_module` to load the modules. A large benefit is that both the normal module loading via importlib as well as the zipimporter can now use the exactly same API.

Unfortunately for zipimporter we need to maintain the legacy loading code with `find_module` + `load_module` for compatibility with Python < 3.10.

This PR maintains full compatibility with the existing plugin systems. Overall there is likely a lot that we could further refactor and trim down, but that is something to be reserved for Picard 3.0.

To properly test the result this PR also enables CI testing with Python 3.12.0rc1

### DIFF
--- a/picard/pluginmanager.py
+++ b/picard/pluginmanager.py
@@ -598,17 +598,17 @@ class PluginMetaPathFinder(MetaPathFinder):
         if not fullname.startswith(_PLUGIN_MODULE_PREFIX):
             return None
         plugin_name = fullname[len(_PLUGIN_MODULE_PREFIX):]
-        for plugin_dir in _plugin_dirs:
-            spec = None
-            if hasattr(zipimport.zipimporter, 'find_spec'):  # Python >= 3.10
+        spec = None
+        if hasattr(zipimport.zipimporter, 'find_spec'):  # Python >= 3.10
+            for plugin_dir in _plugin_dirs:
                 zipfilename = os.path.join(plugin_dir, plugin_name + '.zip')
                 zip_importer = zip_import(zipfilename)
                 if zip_importer:
                     spec = zip_importer.find_spec(fullname)
-            if not spec:
-                spec = importlib.machinery.PathFinder().find_spec(fullname, [plugin_dir])
-            if spec and spec.loader:
-                return spec
+                    break
+        if not spec:
+            spec = importlib.machinery.PathFinder().find_spec(fullname, _plugin_dirs)
+        return spec if spec and spec.loader else None
 
 
 sys.meta_path.append(PluginMetaPathFinder())
--- a/picard/pluginmanager.py
+++ b/picard/pluginmanager.py
@@ -213,6 +213,13 @@ def register_plugin_dir(path):
         _plugin_dirs.append(path)
 
 
+def plugin_dir_for_path(path):
+    for plugin_dir in plugin_dirs():
+        if path.startswith(plugin_dir):
+            return plugin_dir
+    return path
+
+
 class PluginManager(QtCore.QObject):
 
     plugin_installed = QtCore.pyqtSignal(PluginWrapper, bool)
@@ -289,7 +296,7 @@ class PluginManager(QtCore.QObject):
                   len(names))
         for name in sorted(names):
             try:
-                self._load_plugin_from_directory(name, plugindir)
+                self._load_plugin(name)
             except Exception:
                 self.plugin_error(name, _("Unable to load plugin '%s'"), name, log_func=log.exception)
 
@@ -299,7 +306,15 @@ class PluginManager(QtCore.QObject):
                 return (plugin, index)
         return (None, None)
 
-    def _load_plugin_from_directory(self, name, plugindir):
+    def _load_plugin(self, name):
+        existing_plugin, existing_plugin_index = self._get_plugin_index_by_name(name)
+        if existing_plugin:
+            log.debug("Ignoring already loaded plugin %r (version %r at %r)",
+                existing_plugin.module_name,
+                existing_plugin.version,
+                existing_plugin.file)
+            return
+
         spec = None
         module_pathname = None
         zip_importer = None
@@ -309,26 +324,27 @@ class PluginManager(QtCore.QObject):
         # Legacy loading of ZIP plugins. In Python >= 3.10 this is all handled
         # by PluginMetaPathFinder. Remove once Python 3.9 is no longer supported.
         if not hasattr(zipimport.zipimporter, 'find_spec'):
-            zipfilename = os.path.join(plugindir, name + '.zip')
-            zip_importer = zip_import(zipfilename)
-            if zip_importer:
-                if not zip_importer.find_module(name):
-                    errorfmt = _('Failed loading zipped plugin "%(plugin)s" from "%(filename)s"')
-                    self.plugin_error(name, errorfmt, params={
-                        'plugin': name,
-                        'filename': zipfilename,
-                    })
-                    return None
-                module_pathname = zip_importer.get_filename(name)
-                manifest_data = load_zip_manifest(zip_importer.archive)
+            for plugin_dir in plugin_dirs():
+                zipfilename = os.path.join(plugin_dir, name + '.zip')
+                zip_importer = zip_import(zipfilename)
+                if zip_importer:
+                    if not zip_importer.find_module(name):
+                        errorfmt = _('Failed loading zipped plugin "%(plugin)s" from "%(filename)s"')
+                        self.plugin_error(name, errorfmt, params={
+                            'plugin': name,
+                            'filename': zipfilename,
+                        })
+                        return None
+                    module_pathname = zip_importer.get_filename(name)
+                    manifest_data = load_zip_manifest(zip_importer.archive)
+                    break
 
         if not module_pathname:
-            spec = PluginMetaPathFinder().find_spec(full_module_name, [plugindir])
+            spec = PluginMetaPathFinder().find_spec(full_module_name, [])
             if not spec or not spec.loader:
-                errorfmt = _('Failed loading plugin "%(plugin)s" in "%(dirname)s"')
+                errorfmt = _('Failed loading plugin "%(plugin)s"')
                 self.plugin_error(name, errorfmt, params={
                     'plugin': name,
-                    'dirname': plugindir,
                 })
                 return None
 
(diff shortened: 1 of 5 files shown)


hboetes:
FYI I just tried your PR, and it works, but I get these errors.

```
E: 17:15:51,086 config.__init__:424: Option setting/rsgain_command already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:353: in ReplayGain2OptionsPage
        TextOption("setting", "rsgain_command", "rsgain"),
E: 17:15:51,088 config.__init__:424: Option setting/album_tags already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:354: in ReplayGain2OptionsPage
        BoolOption("setting", "album_tags", True),
E: 17:15:51,090 config.__init__:424: Option setting/true_peak already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:355: in ReplayGain2OptionsPage
        BoolOption("setting", "true_peak", False),
E: 17:15:51,091 config.__init__:424: Option setting/reference_loudness already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:356: in ReplayGain2OptionsPage
        BoolOption("setting", "reference_loudness", False),
E: 17:15:51,093 config.__init__:424: Option setting/target_loudness already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:357: in ReplayGain2OptionsPage
        IntOption("setting", "target_loudness", -18),
E: 17:15:51,094 config.__init__:424: Option setting/clip_mode already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:358: in ReplayGain2OptionsPage
        IntOption("setting", "clip_mode", CLIP_MODE_POSITIVE),
E: 17:15:51,096 config.__init__:424: Option setting/max_peak already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:359: in ReplayGain2OptionsPage
        IntOption("setting", "max_peak", 0),
E: 17:15:51,097 config.__init__:424: Option setting/opus_mode already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:360: in ReplayGain2OptionsPage
        IntOption("setting", "opus_mode", OPUS_MODE_STANDARD),
E: 17:15:51,099 config.__init__:424: Option setting/opus_m23 already declared
at /home/han/.config/MusicBrainz/Picard/plugins/replaygain2.zip/replaygain2/__init__.py:361: in ReplayGain2OptionsPage
        BoolOption("setting", "opus_m23", False)
Time: 0h:00m:06s                                                                                                                                                              
```

hboetes:
And it crashes upon changing the configuration.

phw:
@hboetes Thanks for the feedback. The log output is quite interesting. I debugged this a bit. The plugin code gets run twice at exactly the line https://github.com/metabrainz/picard/pull/2307/files#diff-c2f162faf440349fc5556185e9cd8b4b3cf0f35235eb561eb3fd7261651a2fbfR356 . That means the `spec.loader.exec_module(plugin_module)` call is enough for this to happen. Unsure how to avoid this.



> And it crashes upon changing the configuration.

Can you elaborate? Any error output on the terminal?

hboetes:
Yes, this actually happened after pressing the settings button, the first time it complains, the second time it crashes.
```
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/mainwindow.py", line 1313, in show_options
    options_dialog = OptionsDialog.show_instance(page, self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/__init__.py", line 156, in show_instance
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/__init__.py", line 150, in get_instance
                    ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 168, in __init__
    self.add_pages(None, default_page, self.ui.pages_tree)
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 113, in add_pages
    self.add_pages(page.NAME, default_page, item)
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 104, in add_pages
    for foo, bar, page in sorted(pages):
                          ^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'ReplayGain2OptionsPage' and 'ReplayGain2OptionsPage'
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/mainwindow.py", line 1313, in show_options
    options_dialog = OptionsDialog.show_instance(page, self)
                     ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/__init__.py", line 156, in show_instance
               ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/__init__.py", line 150, in get_instance
                    ^^^^^^^^^^^^^^^^^^^^
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 168, in __init__
    self.add_pages(None, default_page, self.ui.pages_tree)
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 113, in add_pages
    self.add_pages(page.NAME, default_page, item)
  File "/usr/local/lib64/python3.12/site-packages/picard/ui/options/dialog.py", line 104, in add_pages
    for foo, bar, page in sorted(pages):
                          ^^^^^^^^^^^^^
TypeError: '<' not supported between instances of 'ReplayGain2OptionsPage' and 'ReplayGain2OptionsPage'
[1]    2000396 IOT instruction  picard
```

hboetes:
For the rest it seems to be working. :+1:

picard/pluginmanager.py
```diff
@@ -553,3 +591,24 @@ def _display_update():
...
         else:
             _display_update()
+
+
+class PluginMetaPathFinder(MetaPathFinder):
+    def find_spec(self, fullname, path, target=None):
+        if not fullname.startswith(_PLUGIN_MODULE_PREFIX):
+            return None
+        plugin_name = fullname[len(_PLUGIN_MODULE_PREFIX):]
+        spec = None
```
zas, line 601: Instead of using `_plugin_dirs` global variable (as it is somehow "private"), we should use the generator wrapping it instead:

```python
paths = tuple(plugin_dirs())
```

and `s/_plugin_dirs/paths/` below.

zas (changes requested):
It looks good to me, I tested and didn't find any issue, good job!

picard/pluginmanager.py
phw: I refactored the class a bit in  https://github.com/metabrainz/picard/pull/2307/commits/70904c11e035c2ad850c2eb3d09509808e25aea9 . I think this makes it easier to understand and extend. It unifies handling of zipped and normal file plugins.

phw:
I have fixed the double execution in https://github.com/metabrainz/picard/pull/2307/commits/4db506c49f8c4e8dcf73ec9bfdba40010b626173 .

Essentially this happened for all plugins that themselves have relative imports. When running `spec.loader.exec_module` the module is not yet `sys.modules`. So if the module does a relative import on itself the module gets imported again. This change feels kind of like a hack, but also the existing behavior using the APIs seems slightly off. I'm not fully sure, but the solution / workaround works.

In the last commit I also changed a bit how plugins get loaded if they exist in multiple directories. The original implementation first tried to load from system plugin dir, but if it also existed in the user plugin dir it "unloaded" the previous plugin and loaded the new one. Now because running `exec_module` causes the module body to run this could cause double execution again. You cannot undo having run the code, and running it twice can have side effects. So now it first tries user dir and later on ignores the same plugin in other directories again.

Overall it just shows again that we need to separate plugin metadata from code for the next plugin system.

Overall I'm quite happy with how well this works now. This really should in practice work as before but using the new APIs.

Also it shows that importlib provides really good interface now for us to implement plugin loading and if done properly we can for the next iteration of the plugin system simplify the loading a lot.

picard/pluginmanager.py
```diff
@@ -156,37 +158,68 @@ def _plugin_name_from_path(path):
...
+
+
+def register_plugin_dir(path):
+    if path not in _plugin_dirs:
+        _plugin_dirs.append(path)
+
+
+def plugin_dir_for_path(path):
+    for plugin_dir in plugin_dirs():
+        if path.startswith(plugin_dir):
```
zas, line 218: Not sure this is correct, if path is `/a/bc/d` and plugin_dir is `/a/b`, this would not give the expected result I think.

https://docs.python.org/3/library/os.path.html#os.path.commonpath could help.

picard/pluginmanager.py
```diff
@@ -272,50 +306,57 @@ def _get_plugin_index_by_name(self, name):
...
                 self.plugin_error(name, errorfmt, params={
                     'plugin': name,
-                    'dirname': plugindir,
                 })
                 return None
 
+            module_pathname = spec.origin
+            if isinstance(spec.loader, zipimport.zipimporter):
+                manifest_data = load_zip_manifest(spec.loader.archive)
+            if module_pathname.endswith("__init__.py"):
```
zas, line 341: May be use https://docs.python.org/3/library/os.path.html#os.path.basename instead:

```suggestion
            if os.path.basename(module_pathname) == '__init__.py':
```

zas (changes requested):
Overall, it looks good, just 2 minor comments

picard/pluginmanager.py
phw: Done

picard/pluginmanager.py
phw: Done in 702a769561c21ff2c813bb8ae64384704ef62583

zas (approved):
LGTM

hboetes:
Thanks for the speedy fix. One note, there are various plugins that won't work with python-3.12, like the `replaygain` plugin. Perhaps a heads-up to the plugin developers is warranted.

phw:
@hboetes the replaygain plugin is actually supposed to work after the latest changes from this PR. At least it does so for me. Do you have any different issue now? 

If there is compatibility issue with one of the existing plugins we should of course fix it, but I'm currently not aware of one.

PR #2309, opened 2023-09-12
PICARD-2751: Restore the ability to load compiled .pyc plugins

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2751

With the changes in #2307 the plugin system's ability to load plugins that only exist as compiled bytecode (`.pyc`) files got lost.

# Solution

Consider all possible file extensions again when trying to load the spec.


phw:
This is probably one of those really rare use cases as we don't make use of this currently ourselves. Well possible nobody would have noticed :)

PR #2310, opened 2023-09-13
On plugin import error unset sys.modules again

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

pluginmanager sets plugin module in `sys.modules`, but this should not happen in case of error importing the module.


zas (approved):
LGTM

PR #2315, opened 2023-09-14
PICARD-1377: New plugin UI icons

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1377

Use the new icons designed by @Aerozol for the plugin UI.

# Solution

Set the icons. Also add dark UI detection and use a separate set of icons.

# Action

@Aerozol The plugins looks really clean and good. But I think the trash can is a bit small, what do you think? Shall we make that a bit larger?

grafik


rdswift:
Once this is implemented, we need to update the images in the documentation at https://picard-docs.musicbrainz.org/en/config/options_plugins.html

Aerozol:
Nice one @phw. How much padding should be in the icon itself was something I wasn’t sure about.
I uploaded bigger ‘trash’ icons [into the ticket](https://tickets.metabrainz.org/browse/PICARD-1377) for you to use :)

zas:
> I personally prefer the cross and would go with that (it's also what is currently in the PR). What do you think?

No strong preference, but I'm fine with cross

zas (approved):
LGTM

Aerozol:
Looks great! Visually I much prefer the cross, hopefully it doesn’t cause any confusion.

PR #2317, opened 2023-09-17
PICARD-2756: Windows long path support for UNC paths

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2756

For Windows UNC paths (e.g. for network shares) the long path support did not work and lead to invalid paths.

While for local paths the file path needs to be preficed with `\\?\`, e.g. `\\?\C:\somelongpath...`, for UNC paths in the form of `\\servername\path` the long path must be defined as `\\?\UNC\servername\somelongpath...`.

See also:

- https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
- https://en.wikipedia.org/wiki/Path_(computing)#Universal_Naming_Convention

# Solution

Extend `picard.util.win_prefix_longpath`  to handle UNC paths.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -267,8 +267,16 @@ def normpath(path):
 
 
 def win_prefix_longpath(path):
+    """
+    For paths longer then WIN_MAX_FILEPATH_LEN enable long path support by prefixing with WIN_LONGPATH_PREFIX.
+
+    See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+    """
     if len(path) > WIN_MAX_FILEPATH_LEN and not path.startswith(WIN_LONGPATH_PREFIX):
-        path = WIN_LONGPATH_PREFIX + path
+        if path.startswith('\\\\'):  # UNC path
+            path = WIN_LONGPATH_PREFIX + 'UNC' + path[1:]
+        else:
+            path = WIN_LONGPATH_PREFIX + path
     return path
 
 
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -273,7 +273,7 @@ def win_prefix_longpath(path):
     See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
     """
     if len(path) > WIN_MAX_FILEPATH_LEN and not path.startswith(WIN_LONGPATH_PREFIX):
-        if path.startswith('\\\\'):  # UNC path
+        if path.startswith(r'\\'):  # UNC path
             path = WIN_LONGPATH_PREFIX + 'UNC' + path[1:]
         else:
             path = WIN_LONGPATH_PREFIX + path
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -809,6 +809,10 @@ class WinPrefixLongpathTest(PicardTestCase):
         path = 'C:\\foo\\' + (252 * 'a')
         self.assertEqual(path, win_prefix_longpath(path))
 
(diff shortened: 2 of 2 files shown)


phw:
Don't merge this yet. While I think this is a correct implementation I have not yet fully tested this under Windows.

phw:
Tested now on Windows 10 / 11 with system wide long path support disabled. Interesting is that in most cases the path I got in Picard already had the `\\?\UNC` prefix added. This is the case for drag and drop, open file with from explorer or using the native windows dialogs.

The prefix is not available when either using the Qt dialogs (which happens if multi-directory selection is active) or when explicitly passing a `\\servername\longpath` path via command line.

With the changes here the proper prefix now gets added and loading / saving such files works regardless how they get added.

phw:
@zas I know you can't test, but I'd appreciate your look over the code itself.

picard/util/__init__.py
```diff
@@ -267,8 +267,16 @@ def normpath(path):
...
 
 def win_prefix_longpath(path):
+    """
+    For paths longer then WIN_MAX_FILEPATH_LEN enable long path support by prefixing with WIN_LONGPATH_PREFIX.
+
+    See https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
+    """
     if len(path) > WIN_MAX_FILEPATH_LEN and not path.startswith(WIN_LONGPATH_PREFIX):
-        path = WIN_LONGPATH_PREFIX + path
+        if path.startswith('\\\\'):  # UNC path
```
zas, line 276:
```suggestion
        if path.startswith(r'\\'):  # UNC path
```

May be use raw string to ease reading: `r'\\'` is the same as `'\\\\'`

test/test_utils.py
```diff
@@ -809,6 +809,10 @@ def test_win_prefix_longpath_is_short(self):
         path = 'C:\\foo\\' + (252 * 'a')
         self.assertEqual(path, win_prefix_longpath(path))
 
+    def test_win_prefix_longpath_unc(self):
+        path = '\\\\server\\' + (251 * 'a')
```
zas, line 813:
```suggestion
        path = r'\\server\' + (251 * 'a')
```

test/test_utils.py
```diff
@@ -809,6 +809,10 @@ def test_win_prefix_longpath_is_short(self):
...
+        self.assertEqual('\\\\?\\UNC' + path[1:], win_prefix_longpath(path))
```
zas, line 814:
```suggestion
        self.assertEqual(r'\\?\UNC' + path[1:], win_prefix_longpath(path))
```

zas (changes requested):
It looks correct to me, though I think we should definitively use raw strings for such Windows paths.

test/test_utils.py
phw: This one does not work because of the trailing backslash. `r'\\server\'`  is a syntax error as the trailing backslash is interpreted as escape character for the single quite..

zas: Ah yes, forgot about this stupid limitation (raw strings cannot end with a backslash).

phw:
@zas I changed to raw strings in b583ea4ab1aa17e2ec4d5dfa3485dd5396086745. Using template strings we can also get around the trailing backslash problem. Overall working with backslash paths is awkward. E.g. my editor (vscode) now renders some of the backlashes + following character like it was an escape sequence.

zas (approved):
LGTM

PR #2319, opened 2023-09-20
TagCoverArtImage(): properly set _repr() and _str() for this subclass

zas:
Also add matching tests

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -228,7 +228,6 @@ class CoverArtImage:
         return "%s(%s)" % (self.__class__.__name__, ", ".join(self._repr()))
 
     def _str(self):
-        yield 'Image'
         if self.url is not None:
             yield "from %s" % self.url.toString()
         if self.types:
@@ -237,7 +236,11 @@ class CoverArtImage:
             yield "and comment '%s'" % self.comment
 
     def __str__(self):
-        return ' '.join(self._str())
+        output = self.__class__.__name__
+        s = tuple(self._str())
+        if s:
+            output += ' ' + ' '.join(s)
+        return output
 
     def __eq__(self, other):
         if self and other:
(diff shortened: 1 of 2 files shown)


picard/coverart/image.py
```diff
@@ -469,19 +469,24 @@ def source(self):
...
+            yield 'types=%r' % self.types
         if self.is_front is not None:
-            p.append("is_front=%r" % self.is_front)
-        p.append('support_types=%r' % self.support_types)
+            yield 'is_front=%r' % self.is_front
+        yield 'support_types=%r' % self.support_types
         if self.comment:
-            p.append("comment=%r" % self.comment)
-        return "%s(%s)" % (self.__class__.__name__, ", ".join(p))
+            yield 'comment=%r' % self.comment
```
phw, line 482: The parent implementation has a different order for is _front and _support_types and also outputs support_multi_types. I wonder if we should instead call the parent implementation:

```python
def _repr(self):
    yield '%r' % self.sourcefile
    if self.tag is not None:
        yield 'tag=%r' % self.tag
    yield from super()._repr()
```

The only additional field handled by the parent function is `url`, but if that is empty it is skipped.

picard/coverart/image.py
```diff
@@ -469,19 +469,24 @@ def source(self):
...
+
+    def _str(self):
+        yield 'Image from %r' % self.sourcefile
+        if self.types:
+            yield "of type %s" % ','.join(self.types)
+        if self.comment:
+            yield "and comment '%s'" % self.comment
```
phw, line 489: Here as well:

```python
def _str(self):
    yield 'Image from %r' % self.sourcefile
    yield from super()._str()
```

phw (changes requested):

picard/coverart/image.py
zas: Done (and I did more or less the same for _str()) in [c8a41de](https://github.com/metabrainz/picard/pull/2319/commits/c8a41de4103e631164241fd81f00ee46f14e4022)

picard/coverart/image.py
```diff
@@ -228,7 +228,6 @@ def __repr__(self):
         return "%s(%s)" % (self.__class__.__name__, ", ".join(self._repr()))
 
     def _str(self):
-        yield 'Image'
```
phw, line 231: Not sure what the intention is, but having as string representation just "from /some/file/name" seems odd as well. What about outputting `self.__class__.__name__` here as well?

zas: Done in [6e0054d](https://github.com/metabrainz/picard/pull/2319/commits/6e0054d2423a8d06d2d32d35beca82cf24bd5e04)

phw (approved):

PR #2320, opened 2023-09-20
Caa tidyup 3

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/coverart/__init__.py
+++ b/picard/coverart/__init__.py
@@ -53,7 +53,7 @@ class CoverArt:
         self.front_image_found = False
 
     def __repr__(self):
-        return "CoverArt for %r" % (self.album)
+        return "%s for %r" % (self.__class__.__name__, self.album)
 
     def retrieve(self):
         """Retrieve available cover art images for the release"""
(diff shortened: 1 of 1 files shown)


picard/coverart/__init__.py
```diff
@@ -95,7 +93,7 @@ def _coverart_downloaded(self, coverartimage, data, http, error):
         self.album._requests -= 1
 
         if error:
-            self.album.error_append('Coverart error: %s' % http.errorString())
+            self.album.error_append("Coverart error: %s" % http.errorString())
```
phw, line 96: Just because I don't understand: Why did this get changed to double quote, while in lines 128 following strings are changed from double quote to single quote?

phw (approved):
Makes sense and improved readability. LGTM

Some question about the quoting style, but it's likely really just a question. No strong opinion either way, but I agree that there should be some consistency (at least in code close together)

picard/coverart/__init__.py
zas: Well, my idea is that strings that are English phrases are likely to change to include a single quote at any moment, but that identifiers and keys are very unlikely to contain one.

Example:

```python
f("It is red")
f("It's red")
d['key']
```

Changing from `"It is"` to `"It's"` doesn't require any change regarding quotes, while changing from `'It is'` does.

At least, that the "consistency" I'm trying to apply since a while.

phw: Understood. Thanks for the explanation. Makes sense to me, I'll try to follow this scheme as well.

zas: I tried to clarify this: https://github.com/metabrainz/picard/pull/2322

PR #2321, opened 2023-09-21
PICARD-2760: Timestamp the Windows binaries when codesigning

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

* JIRA ticket (_optional_): PICARD-2760

Enable timestamping when code signing for Windows. Signed binaries with a timestamp will remain valid even after the certificates have expired as long as they get signed before expiration of the certificates.

For details see:

- https://learn.microsoft.com/en-us/windows/win32/seccrypto/time-stamping-authenticode-signatures
- https://www.ssl.com/how-to/using-your-code-signing-certificate/#ftoc-heading-10

# Solution

Use the timestamp server provided by ssl.com with SignTool

### DIFF
--- a/.github/workflows/package.yml
+++ b/.github/workflows/package.yml
@@ -188,8 +188,10 @@ jobs:
     - name: Build Windows 10 signed app package
       if: matrix.type == 'signed-app' && env.CODESIGN == '1'
       run: |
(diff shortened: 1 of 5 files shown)


phw:
Tested with the signed binaries on Windows 10 and 11. The binaries now show the timestamp in the file properties:

grafik

PR #2325, opened 2023-09-22
PICARD-2762: Hide program update for Linux packaged version but leave plugins update check visible

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Show plugin update check option for Linux packaged version of Picard

# Problem

The Linux packaged version of Picard does not show the option to check for plugin updates on startup, even though this should be a valid option.

* JIRA ticket: PICARD-2762

# Solution

Hide only the items specific to the program update checking, and leave the groupbox and plugin update check option visible.

# Action

None.


phw (approved):
Good find. Plugin updates should of course still be checked.

PR #2326, opened 2023-09-23
Quotes consistency

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

# Solution

# Action


zas:
Not finished, but the idea is to go through all files.
Overall, it appears we were mainly following quoting guidelines (well, in fact guidelines were made after our practice), but there are still few inconsistencies.

I clarified quoting of URIs: basically a quote can theoretically appear in those (single quote is a reserved character for sub delimitation, a bit like `#` or `?`). So better double quote those.

rdswift (commented):
These look good.  The convention also makes sense.  Easy to understand, and it makes the code easier to read.

rdswift (approved):
This all looks good to me.  Now I just need to try to remember to follow it when submitting changes.  ;-)

phw (approved):

phw:
Merged, but I squashed all the commits that updated the quotes only

zas:
> Merged, but I squashed all the commits that updated the quotes only

In general, I wouldn't squash such big set of changes, it disables the possibility to bisect. I intentionally split changes in small sets.
In this case, that's mainly stylistic changes, and they shouldn't have impact on code execution though.

phw:
In this case it was really a huge amount of separate commits with essentially all the same commit message and with many but trivial changes that causes parser error if done wrong. There was not much point in keeping those.

I totally agree in doing changes in multiple commits and I'm not much a friend of squashing all PR as some projects practice it, as any PR that is a bit more complex easily consists of multiple separate steps of implementation. But if each commit is just an extension or fix of the previous one we can keep them combined.

PR #2329, opened 2023-09-25
Drop unused imports

zas:
I think it may break some plugins expecting those constants in `picard.const`.

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -120,10 +120,6 @@ for k, v in MB_ATTRIBUTES.items():
     elif k.startswith('DB:release_status/name:'):
         RELEASE_STATUS[v] = v
 
-# List of alias locales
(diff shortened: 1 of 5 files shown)


zas:
We could drop the imports from code using them, but keep them in `const` in order to maintain compatibility (in short, not apply changes to `picard/const/__init__.py`.

phw:
None of the plugins from the plugin repo uses any of those import. But yes, the save option would be to keep the imports in `const/__init__.py`.

But my proposal would be to release 2.10rc1 as is and drop the imports completely afterwards. I'd like to open the 2.10.x branch immediately after we have released rc1 and from there on only cherry pick changes if needed. So we could start with potentially breaking changes then.

phw:
Sorry, closed by accident.

zas:
I went to the middle road, keeping those in `__init__.py` for now, but still cleaning our source files.

See https://github.com/metabrainz/picard/pull/2330

PR #2333, opened 2023-10-02
Make `get_qt_enum()` a generator and extend tests

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -492,10 +492,46 @@ class GetQtEnum(PicardTestCase):
 
     def test_get_qt_enum(self):
         from PyQt5.QtCore import QStandardPaths
-        values = util.get_qt_enum(QStandardPaths, QStandardPaths.LocateOption)
-        self.assertIn('LocateFile', values)
-        self.assertIn('LocateDirectory', values)
-        self.assertNotIn('DesktopLocation', values)
+        keys = set(util.get_qt_enum(QStandardPaths, QStandardPaths.LocateOption))
+        self.assertIn('LocateFile', keys)
+        self.assertIn('LocateDirectory', keys)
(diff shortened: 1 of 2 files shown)


zas:
Since the method isn't specific to qt, it should likely be renamed to something more generic, any ideas?

test/test_utils.py
```diff
@@ -492,11 +492,35 @@ class GetQtEnum(PicardTestCase):
...
+
+        values = set(util.get_qt_enum(SomeOtherClass, SomeAttrClass))
+        self.assertIn('a', values)
+        self.assertIn('b', values)
+
+        obj = SomeOtherClass()
+        res = []
+        for attr in sorted(util.get_qt_enum(SomeOtherClass, list)):
+            res.append(getattr(obj, attr))
+        self.assertEqual(res, [[1, 2, 3], [4, 5, 6]])
```
phw, line 522: I'd put the test for the three classes into three separate test functions. That makes it easier to handle failing tests.

phw (changes requested):
Definitely the refactoring of the function we should put in.

About the renaming I'm unsure. Yes, in its current implementation it is generic and could be reused for something else. But currently we don't have a use case for it.

On the other hand let's say the implementation of Qt's enums changes. We could then update a `get_qt_enum` function to deal with that, as it is for this specific purpose. If we rename it to something specific we instead would need to introdice `get_qt_enum` again (and if we have not other use case remove the generic function).

So my vote is for keeping the function as is for now. Should we have a use case where we need the same functionality but for some other class I would say we rename the function then. Does this make some sense?

phw (approved):
LGTM

PR #2334, opened 2023-10-02
PICARD-2769: Show a file's original metadata in the panes for unset tags

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2769

If a tag is unset then the metadata for this is shown empty on the main panes' columns.

At least for the case where there is a file and the option "clear existing tags" is not set the original value for the file should be shown to better reflect what is being saved.

# Solution

In case a tag has been unset and "clear existing tags" is not enabled display a file's original value in the pane columns.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -816,7 +816,10 @@ class File(QtCore.QObject, Item):
             return self.base_filename
         elif column == 'covercount':
             return self.cover_art_description()
-        return m[column]
+        value = m[column]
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2338, opened 2023-10-17
PICARD-2777: Update metadata option screen to clarify relationship selections

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**: Update the metadata option screen to clarify the relationship selections.

# Problem

Enabling the "Use track relationships" option automatically includes the release relationships which are enabled with the "Use release relationships" option.  Thus if "Use track relationships" is enabled, the setting of the "Use release relationships" option is moot.  This is not clear in the user interface.

* JIRA ticket (_optional_): PICARD-2777

# Solution

Update the user interface for the Metadata Options screen such that when the "Use track relationships" option is checked, the "Use release relationships" option is automatically shown as checked and the selection for this option is disabled.

When the "Use track relationships" option is unchecked, the "Use release relationships" option is automatically set to the value that was saved before the "Use track relationships" option was checked.

Also revise the description of the "Use track relationships" option to clarify that it also includes the release relationships.

# Action

Once merged, the documentation will need to be updated with the new screen image and option description.


phw (approved):
Thanks, this really clarifies the relationship settings in the UI.

I was a bit unsure about including the full label "Use track relationships (includes release relationships)", as the UI labels sometimes already are pretty verbose and become very long and it probably would also work without this clarification. And in some translations this then becomes even longer. E.g. for German it'll be "Titelbeziehungen verwenden (beinhaltet Veröffentlichungsbeziehungen)", and I don't know Finnish, but it'll probably be something like "Käytä kappaleiden suhteiden tietoja (sisältää julkaisujen suhteiden tietoja)".

But I actually tried it, and it's not so bad in the UI as we have other long labels and enough room:

grafik

grafik

@zas Any opinion here?

rdswift:
> Can't we rewrite "Use track relationships (includes release relationships)" to "Use track and release relationships" ?

Duh, yeah. :stuck_out_tongue_winking_eye:   That makes a lot of sense.  Wish I had thought of it in the first place.

zas (approved):
LGTM

phw (approved):

PR #2341, opened 2023-10-26
PICARD-2781: fix releasetype weights

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2781, PICARD-2780

This PR fixes two issues with search result comparison for found recordings:

- If a release + release group had no type set at all the search score got really low as this was treated as a total mismatch with score 0.0
- If a recording had the video flag not set on MB, the API can return `"video": null` (not sure if always or only sometimes). The matching code only expected `true` or `false`, and hence non-video recordings always mismatched, lowering the score

See also discussion at https://community.metabrainz.org/t/similar-track-with-score-100-will-not-match/661856

# Solution

- If no release type is set apply the weight for "Other" type
- Fix the video flag check by explicitly converting to boolean

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -373,7 +373,7 @@ class Metadata(MutableMapping):
 
             if 'isvideo' in weights:
                 metadata_is_video = self['~video'] == '1'
-                track_is_video = track.get('video', False)
+                track_is_video = bool(track.get('video'))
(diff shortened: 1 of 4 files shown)


zas (approved):
LGTM

PR #2348, opened 2023-12-19
PICARD-2792: fix matching for AcoustID results without metadata and standalone-recordings

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2792

In PICARD-2792 and the forum discussions at https://community.metabrainz.org/t/worse-results-in-picard-2-10/663668/34 two issues with the metadata matching were identified:

1. For AcoustID recordings without metadata a high submission score led to disproportional high scoring. This happens for example if recordings get merged without the AcoustID server having applied this merge yet.

2. Generally matching recordings without any releases (standalone-recordings) scored higher on average. This is because as soon as release data gets compared it usually leads to lower score (as a 100% score is seldom).

# Solution
For 1) ignore AcoustID recordings without additional metadata when determining the submission count score. This intentionally does not drop those recordings completely, as they in some cases are the only result and still useful for standalone-recording matches. Also during metadata comparison those recordings naturally score lower due to missing data.

For 2) apply a weighted score in case there are no releases for the recording. The weight is based on the total weight of release specific weights, the score uses the "Other" release type by default.

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2017-2018 Sambhav Kothari
 # Copyright (C) 2018 Vishal Choudhary
 # Copyright (C) 2018-2021 Laurent Monin
-# Copyright (C) 2018-2022 Philipp Wolfer
+# Copyright (C) 2018-2023 Philipp Wolfer
 # Copyright (C) 2023 Bob Swift
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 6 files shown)


zas (approved):
Nice patch, definitively an improvement, LGTM.

phw:
Also cherry-picked to the 2.10.x branch

PR #2351, opened 2023-12-21
PICARD-2584: Load recording if AcoustId metadata is missing

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2584

AcoustID can sometimes contain no MB metadata for linked recordings. This happens if:

1. recordings have been merged and this merge was not yet handled by the AcoustID server
2. recordings have been deleted and the AcoustID server has not yet cleaned up

AFAIK both should be handled by AcoustID, but currently this does not seem to work correctly (e.g. https://github.com/acoustid/acoustid-server/issues/114). There can always be some time frame where this has not been done as there is some delay in AcoustID database synchronizing., though

In case of 1) this can lead to recording data not being available and files getting matched to standalone recordings instead. It also messes with submission counts.

Examples:

- https://acoustid.org/track/d10da0d8-8b13-4453-9ac5-b179958d01ca

In case of 2) this can lead to files being matched to a standalone-recording that does not load.

Examples:

- https://acoustid.org/track/cb127606-8e3d-4dcf-9902-69c7a9b59bc9
- https://acoustid.org/track/aed07515-2856-4265-afce-247c09e2b12a
- https://acoustid.org/track/47e0f233-9494-4197-b5ab-aa943ab54ee7
- https://acoustid.org/track/85c71305-a4e5-43cf-8a0f-1749cf10edde

# Solution

If AcoustID provides no MB metadata but only MB recording ID load the recording data from the MB web service. This is done by the new `RecordingResolver` in the picard.acoustid package.

If metadata is provided, the recording is used directly. If not a MB web service call to load the recording is initiated.

If the recording no longer exist (404) then the result is ignored. If it exists and was redirected to another recording MBID that is also linked to the same AcoustID, the submission counts get combined.

### DIFF
--- a/picard/acoustid/__init__.py
+++ b/picard/acoustid/__init__.py
@@ -34,10 +34,7 @@ import json
 from PyQt6 import QtCore
 
 from picard import log
-from picard.acoustid.json_helpers import (
-    max_source_count,
-    parse_recording,
-)
+from picard.acoustid.recordings import RecordingResolver
 from picard.config import get_config
 from picard.const import (
     DEFAULT_FPCALC_THREADS,
@@ -49,6 +46,7 @@ from picard.util import (
     find_executable,
     win_prefix_longpath,
 )
+from picard.webservice.api_helpers import AcoustIdAPIHelper
 
 
 def get_score(node):
@@ -76,7 +74,7 @@ AcoustIDTask = namedtuple('AcoustIDTask', ('file', 'next_func'))
 
 class AcoustIDClient(QtCore.QObject):
 
-    def __init__(self, acoustid_api):
+    def __init__(self, acoustid_api: AcoustIdAPIHelper):
         super().__init__()
         self._queue = deque()
         self._running = 0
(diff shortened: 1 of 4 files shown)


picard/acoustid/__init__.py
```diff
@@ -108,42 +106,16 @@ def _on_lookup_finished(self, task, document, http, error):
                 mparms,
                 echo=None
             )
+            task.next_func(doc, http, error)
```
zas, line 109: `doc` is always `{}`, expected?

phw: Yes, it's in the error case empty. It was the same before. But it also was confusing (and partially used wrong) with two similarly named variables `document` and `doc` with different purpose. I cleaned this up and removed the `doc` variable.

picard/acoustid/recordings.py
```diff
@@ -0,0 +1,174 @@
...
+        if error:
+            if error == QNetworkReply.NetworkError.ContentNotFoundError:
+                # Recording does not exist, ignore and move on
+                self._load_recordings()
+            else:
+                self._send_results(error)
+            return
+
+        mbid = mb_recording.get('id')
+        self._recording_cache[mbid] = mb_recording
```
zas, line 128: Do we still want to use `mbid` here if it's empty? It is tested just below (`if mbid:`)

picard/acoustid/recordings.py
```diff
@@ -0,0 +1,174 @@
...
+
+        mbid = self._missing_metadata[0].mbid
+        if mbid in self._recording_cache:
+            mb_recording = self._recording_cache[mbid]
+            self._recording_request_finished(mb_recording, None, None)
+        else:
+            self._mbapi.get_track_by_id(
+                self._missing_metadata[0].mbid,
+                self._recording_request_finished,
+                inc=['artists', 'release-groups', 'releases', 'media'],
```
zas, line 114: Just a minor thing, but it could be passed as a tuple, rather than a list.

picard/acoustid/recordings.py
phw: I moved it down into the if

phw:
@zas I made a few enhancements, and I think this is now ready for review. With the changes I don't get significantly longer loading times while still eliminating the standalone recording results for my 5k test files.

The changes allowing this are:

- If a recording without metadata is returned from AcoustID it is ignored unless it has more then 25% submission count compared to the recording with highest submission count.
- Loaded recording IDs are cached and only loaded if they are not in cache. This avoids loading the same recording twice. This cache includes redirects.
- A max. of 3 recordings without metadata are resolved for a single AcoustID. That means any AcoustID lookup never can trigger more then 3 additional lookups. The 3 recordings being looked up are always the three with the highest submission count.

zas (approved):
Very clean patch, good job! LGTM

PR #2352, opened 2023-12-22
PICARD-2805: Handle the None case for metadata compare

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2805

If a loaded track has no length (length is `None`) the metadata compare throws an exception.

# Solution

Handle the None case for metadata compare. Also handle the theoretical case that the `release-group`  might also miss for the releases loaded for a track.

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -206,6 +206,8 @@ class Metadata(MutableMapping):
 
     @staticmethod
     def length_score(a, b):
+        if a is None or b is None:
+            return 0.0
         return (1.0 - min(abs(a - b),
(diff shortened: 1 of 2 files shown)


zas (changes requested):
LGTM, but I suggest we also reformat `results` to introduce ending comma in order to avoid unrelated changes in diffs when adding or removing entries in the future.

See https://github.com/phw/picard/pull/20

zas (approved):
LGTM

PR #2336, opened 2023-10-10
PICARD-2774: Do not fail loading files with invalid ID3 image types

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem

* JIRA ticket (_optional_): PICARD-2774

Files containing embedded cover art with invalid ID3 image type fail to load.

# Solution

Load files even with invalid type. Assume type "other" in this case.

PR #2359, opened 2024-01-08
PICARD-2814: Fix exception on loading embedded cover with invalid ID3 type

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2814

This is a follow up to the fix for PICARD-2774 (#2336). While the type fix in general was correct an exception was thrown when doing the logging in https://github.com/phw/picard/blob/c612116a6342b0aab0e7a4af1862eb2b6fbe01e2/picard/coverart/image.py#L182 as the implementation of `TagCoverArtImage._repr` did access variables not yet initialized at this point.

# Solution

Set `self.sourcefile` and `self.tag` for `TagCoverArtImage` before calling the parent constructor. Both are needed for `_repr()`  and `_str()` working as expected.

Added additional test case to capture this specific issue.

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2007 Oliver Charles
 # Copyright (C) 2007, 2010-2011 Lukáš Lalinský
-# Copyright (C) 2007-2011, 2014, 2018-2023 Philipp Wolfer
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

PR #2360, opened 2024-01-08
Do PyPI deployment with Trusted Publisher Management

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

This eliminates the need to provide permanent access tokens for PyPI in the CI and instead uses temporary tokens issued by OpenID Connect (OIDC)

# Solution

The metabrainz/picard repository has been setup as a trusted publisher on the PyPI picard project. The Github action now uses pypa/gh-action-pypi-publish@release/v1 instead of unvoking twine.

# Action

Once this has been merged (including the 2.10.x branch) we could remove PYPI_UPLOAD_TOKEN from the Github Actions environment.

Generally this is difficult to test as we need to perform our first release with this to see how it goes. But the risk is low. If everything fails twine can be run locally.

### DIFF
--- a/.github/workflows/package-pypi.yml
+++ b/.github/workflows/package-pypi.yml
@@ -12,6 +12,11 @@ jobs:
     runs-on: ubuntu-latest
     env:
       CODESIGN: 0
+    environment:
+      name: pypi
+      url: https://pypi.org/p/picard
+    permissions:
+      id-token: write  # required for PyPI upload
 
     steps:
     - uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

phw:
@zas The workflow needs further restructuring. The publishing action only runs on Linux, but currently the macOS and Windows builds to their publishing with the same machine. That means we need to separate the builds from the publishing by gathering artifcats. Which isn't too bad, as we can publish all files (the sdist and all the bdist packages) in one go.

phw:
@zas I restructured it so there is a single job that downloads all previously build pypi distributions and does the upload. Except for the actual upload task I verified that this works on https://github.com/metabrainz/picard/actions/runs/7464647291/job/20312383875 . All artifacts are placed in `dist/` prior to running the upload task itself. So this should work.

With the latest commit the entire job is skipped now unless we are on a tag. So we see a bit different result.

zas (approved):
LGTM

PR #2361, opened 2024-01-10
PICARD-1685 : Add ~filesize variable

nullHawk:
# Summary

Add ~filesize variable

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem
Provide the file size in bytes as a hidden file variable ~filesize (similar to the existing file variables like e.g. ~bitrate and ~length). 

* JIRA ticket (_optional_): PICARD-1685

# Solution

I have added the "~filename" variable which stores the size of the file in bytes using "os.path.getsize" and passing the "self.filename" in it. also I have used this variable in infodialog.py to show Size. Modified test cases for '~filsize'. Add '~filesize' to metadata in wav.py since it doesn't uses mutagen. 

# Action
I don't know if you want to also add size option in itemviews.py (since it was mention in the old discussion "https://community.metabrainz.org/t/current-file-size/453338/5"). If yes, then please specifiy how to add it in UI, sicne the size is in bytes and in itemviews.py values are directly added from the list ("columns").

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -83,13 +83,13 @@ from picard.util import (
     any_exception_isinstance,
     decode_filename,
     emptydir,
+    encode_filename,
     find_best_match,
     format_time,
     is_absolute_path,
     normpath,
     thread,
     tracknum_and_title_from_filename,
-    encode_filename,
 )
 from picard.util.filenaming import (
     get_available_filename,
--- a/picard/file.py
+++ b/picard/file.py
@@ -762,7 +762,7 @@ class File(QtCore.QObject, Item):
 
     def _info(self, metadata, file):
         try:
-            metadata['~filesize'] = os.path.getsize(encode_filename(self.filename))
+            metadata['~filesize'] = os.path.getsize(encode_filename(file.filename))
         except BaseException:
             pass
         if hasattr(file.info, 'length'):
--- a/picard/file.py
+++ b/picard/file.py
@@ -762,8 +762,7 @@ class File(QtCore.QObject, Item):
 
     def _info(self, metadata, file):
         try:
-            size = os.path.getsize(encode_filename(self.filename))
-            metadata['~filesize'] = size
+            metadata['~filesize'] = os.path.getsize(encode_filename(self.filename))
         except BaseException:
             pass
         if hasattr(file.info, 'length'):
--- a/picard/file.py
+++ b/picard/file.py
@@ -89,6 +89,7 @@ from picard.util import (
     normpath,
     thread,
     tracknum_and_title_from_filename,
+    encode_filename,
 )
 from picard.util.filenaming import (
     get_available_filename,
@@ -132,7 +133,7 @@ class File(QtCore.QObject, Item):
 
     EXTENSIONS = []
 
-    FILE_INFO_TAGS = ('~bitrate', '~sample_rate', '~channels', '~bits_per_sample', '~format')
+    FILE_INFO_TAGS = ('~bitrate', '~sample_rate', '~channels', '~bits_per_sample', '~format', '~filesize')
 
     comparison_weights = {
         'title': 13,
@@ -760,6 +761,11 @@ class File(QtCore.QObject, Item):
         return True
 
     def _info(self, metadata, file):
+        try:
+            size = os.path.getsize(encode_filename(self.filename))
+            metadata['~filesize'] = size
+        except BaseException:
+            pass
         if hasattr(file.info, 'length'):
             metadata.length = int(file.info.length * 1000)
         if hasattr(file.info, 'bitrate') and file.info.bitrate:
--- a/picard/file.py
+++ b/picard/file.py
@@ -763,8 +763,8 @@ class File(QtCore.QObject, Item):
     def _info(self, metadata, file):
         try:
             metadata['~filesize'] = os.path.getsize(encode_filename(file.filename))
-        except BaseException:
-            pass
+        except BaseException as e:
+            log.error(e)
         if hasattr(file.info, 'length'):
             metadata.length = int(file.info.length * 1000)
         if hasattr(file.info, 'bitrate') and file.info.bitrate:
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -286,12 +286,10 @@ def format_file_info(file_):
     info.append((_("Filename:"), file_.filename))
     if '~format' in file_.orig_metadata:
         info.append((_("Format:"), file_.orig_metadata['~format']))
-    try:
-        size = os.path.getsize(encode_filename(file_.filename))
+    if '~filesize' in file_.orig_metadata:
+        size = file_.orig_metadata['~filesize']
         sizestr = "%s (%s)" % (bytes2human.decimal(size), bytes2human.binary(size))
         info.append((_("Size:"), sizestr))
-    except BaseException:
-        pass
     if file_.orig_metadata.length:
         info.append((_("Length:"), format_time(file_.orig_metadata.length)))
     if '~bitrate' in file_.orig_metadata:
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -33,7 +33,6 @@ from collections import (
     namedtuple,
 )
 from html import escape
-import os.path
 import re
 import traceback
 
--- a/picard/ui/infodialog.py
+++ b/picard/ui/infodialog.py
@@ -49,7 +49,6 @@ from picard.file import File
 from picard.track import Track
 from picard.util import (
     bytes2human,
-    encode_filename,
     format_time,
     open_local_path,
     union_sorted_lists,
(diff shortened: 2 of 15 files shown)


picard/ui/infodialog.py
```diff
@@ -286,12 +286,10 @@ def format_file_info(file_):
     info.append((_("Filename:"), file_.filename))
     if '~format' in file_.orig_metadata:
         info.append((_("Format:"), file_.orig_metadata['~format']))
-    try:
-        size = os.path.getsize(encode_filename(file_.filename))
+    if '~filesize' in file_.orig_metadata:
+        size = file_.orig_metadata['~filesize']
```
phw, line 290: That's a good catch to update this piece of code

test/formats/test_vorbis.py
```diff
@@ -208,6 +208,7 @@ class FLACTest(CommonVorbisTests.VorbisTestCase):
         '~channels': '2',
         '~sample_rate': '44100',
         '~format': 'FLAC',
+        '~filesize':'6546'
```
phw, line 211: Can you please add this check for other formats as well? As it is the responsibility of each format to call the `_info` method (and in some cases might not do so) this makes sure this works for all formats and does not break.

Which reminds me that getting the `~filesize` should also be added to https://github.com/metabrainz/picard/blob/master/picard/formats/wav.py#L237 . That's the legacy WAVE file support which does not use mutagen. While me might remove this at some point we should still set the filesize as long as we have the code still.

And minor coding style issue, but there should be a space after the colon.

phw (changes requested):
This looks really good, thanks a lot for for looking over the code with so much detail. Just some minor comments.

If you feel up to it you could also update the documentation at https://picard-docs.musicbrainz.org/en/variables/variables_file.html to have the `_filesize` variable listed there. The docs are managed in https://github.com/metabrainz/picard-docs

test/formats/test_vorbis.py
nullHawk: Sure, I will soon add this after probably 15 Jan (due to ongoing semester exams :( )

picard/ui/infodialog.py
nullHawk: Thanks

nullHawk:
I will soon add the rest of the requirements along with the documentations and code the formated according to the PEP 8 standards

picard/file.py
```diff
@@ -760,6 +761,11 @@ def can_view_info(self):
         return True
 
     def _info(self, metadata, file):
+        try:
+            size = os.path.getsize(encode_filename(self.filename))
+            metadata['~filesize'] = size
```
zas, line 766: `size` is not needed.

```python
 metadata['~filesize'] = os.path.getsize(encode_filename(self.filename))
```

zas (changes requested):

picard/file.py
nullHawk: Thank you for mentioning, I will change that...

nullHawk:
Added checks to other formates for '~filesize', also added '~filsize' to metadata in wav.py

nullHawk:
Also raised PR "https://github.com/metabrainz/picard-docs/pull/218"  for updating the _filesize variable as mentioned by @phw :)

phw (dismissed):
Thanks, this looks good. LGTM

picard/ui/infodialog.py
zas, line 290: Since you removed the call to `os.path.getsize()` importing `os.path` at top of the file isn't needed anymore.

zas (changes requested):

phw:
Not fully sure why https://github.com/metabrainz/picard/actions/runs/7540649415/job/20539127340?pr=2361 is failing. I first assumed it was because of mutagen==1.3.7, but testing with this mutagen version locally shows no errors.

One cause might be that the randomization of tests (with pytest-randomly) makes this issue happen only sometimes. Does some other test maybe write to the source file?

picard/file.py
```diff
@@ -760,6 +761,10 @@ def can_view_info(self):
         return True
 
     def _info(self, metadata, file):
+        try:
+            metadata['~filesize'] = os.path.getsize(encode_filename(self.filename))
```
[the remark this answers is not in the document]
nullHawk: I checked the `file` object it has the `filename` attribute itself. Instead of passing the filename as a parameter, can we use `file.filename` ?

nullHawk:
> Not fully sure why https://github.com/metabrainz/picard/actions/runs/7540649415/job/20539127340?pr=2361 is failing. I first assumed it was because of mutagen==1.3.7, but testing with this mutagen version locally shows no errors.
> 
> One cause might be that the randomization of tests (with pytest-randomly) makes this issue happen only sometimes. Does some other test maybe write to the source file?

Yes, I have confirmed this. When I tested with `mutagen` version 1.47 and `python 3.11`, all the tests worked fine. However, upon revisiting the requirements file and installing the correct version of `mutagen`, that is `mutagen 1.37` with `python 3.8`, that particular test case failed. So I modified the file size in the test case accordingly.

phw:
> However, upon revisiting the requirements file and installing the correct version of `mutagen`, that is `mutagen 1.37` with `python 3.8`, that particular test case failed. So I modified the file size in the test case accordingly.

Now it fails in other cases :( Something is going on here, but I don't yet understand what. mutagen 1.37 is an old version without WAVE support, so we are in the alternative code paths https://github.com/metabrainz/picard/blob/94250ff086486f10b691b8f9743d5d7504eb28ed/test/formats/test_wav.py#L148 and https://github.com/metabrainz/picard/blob/94250ff086486f10b691b8f9743d5d7504eb28ed/picard/formats/wav.py#L226 . But it only seems to happen in the combination of both Python 3.8 and old mutagen. Reading the file size however is not involving mutagen.

I don't see what the issue is right now. Maybe some Python 3.8 bug?

But I'd be happy with just disabling the filesize check for WAVE for now and merge this change.

@zas any ideas?

nullHawk:
I carefully checked the failed tests and found the following issues:
1. `mutagen 1.37` is not compatible with Python 3.11. When attempting to install `mutagen 1.37` with `python 3.11` as specified in requirements.txt, it automatically installs `mutagen 1.47` instead of `mutagen 1.37`
2. `mutagen 1.37` is compatible with `python 3.8`
3. and we have 2 cases now
- `mutagen 1.37` with `python 3.8` expects the filesize to be `14652`
- while `mutagen 1.47` with `python 3.11` expects the filesize to be `15706`.

I am not sure why it's behaving like this.

phw:
> As python 3.8 support is [set to end in 2024 ](https://devguide.python.org/versions/)and the current version is 3.12, are we planning to migrate from python 3.8 to a higher version? If so, I am happy to contribute to the migration process.

We'll drop 3.8 out of the CI builds and set 3.9 as min. supported version some time shortly after it has become EOL. No big migration really as there is full compatibility with newer versions.

The only thing that will be impacted is drop for Windows 7 support, but we have this with the Qt6 migration anyway. That's why the Windows builds in the master branch do build with Python 3.12 already.

phw:
@nullHawk I found the issue, that was tricky.  The issue was that the `test_info` test case was saving the data before performing the test, and hence as a side effect changing the actual test file. This resulted in different file size than the actual test files had. I only noticed because I compared file size on disk to file size in test.

See https://github.com/metabrainz/picard/pull/2361/commits/0332f29e79c26963ee2375446f04cd99e554aad7#diff-7d25cea7df3c99710f9ba35f5231579dcd76cebc8effb6fb0ea01dfa531e2b57L238

I took the freedom to commit this change directly to the PR in 0332f29e79c26963ee2375446f04cd99e554aad7

phw:
Now only some style issues remain. Can you check this locally? We use isort to organize the import statements, see https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md

nullHawk:
> @nullHawk I found the issue, that was tricky. The issue was that the `test_info` test case was saving the data before performing the test, and hence as a side effect changing the actual test file. This resulted in different file size than the actual test files had. I only noticed because I compared file size on disk to file size in test.
> 
> See [0332f29#diff-7d25cea7df3c99710f9ba35f5231579dcd76cebc8effb6fb0ea01dfa531e2b57L238](https://github.com/metabrainz/picard/commit/0332f29e79c26963ee2375446f04cd99e554aad7#diff-7d25cea7df3c99710f9ba35f5231579dcd76cebc8effb6fb0ea01dfa531e2b57L238)
> 
> I took the freedom to commit this change directly to the PR in [0332f29](https://github.com/metabrainz/picard/commit/0332f29e79c26963ee2375446f04cd99e554aad7)

Oh I see now...Thanks a lot!

nullHawk:
> Now only some style issues remain. Can you check this locally? We use isort to organize the import statements, see https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md

Sure :)

picard/file.py
```diff
@@ -760,6 +761,10 @@ def can_view_info(self):
         return True
 
     def _info(self, metadata, file):
+        try:
+            metadata['~filesize'] = os.path.getsize(encode_filename(file.filename))
+        except BaseException:
```
zas, line 766: Please log an error using ` log.error()` if this exception happens.

zas (changes requested):

zas (approved):
LGTM

phw (approved):

PR #2362, opened 2024-01-18
Update copyright headers

zas:
New Year, new headers


rdswift (approved):
It's too bad that this couldn't be automated somehow.  Perhaps something like putting a placeholder in the copyright line when the file is changed and then update it to the release year as part of creating the release.  I know it's not that simple, but it could save a lot of manual effort.

phw:
> Perhaps something like putting a placeholder in the copyright line when the file is changed and then update it to the release year as part of creating the release.

I think this could be a pre-commit hook. The update should happen when files get committed.

PR #2363, opened 2024-01-27
PICARD-2826: Update ~filesize after saving file

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2826

The `~filesize` variable added in #2361 does not update after saving.

# Solution

Move the filesize update into `File._add_path_to_metadata`, which is specifically called after file save. Also rename this to `File._update_filesystem_metadata` due to changed scope.

# Action

PICARD-2826 is broader. We have other info metadata that is not being updated properly. I think it is only about `~format`, but e.g. for ID3 this cane change from v2.3 to v2.4 (or vice versa). And for e.g. AC3 it indicates whether APE tags are present or not, which can also change after save.

Fixing this is more difficult, especially if we want to avoid re-reading the entire file (which would be a straightforward solution). I'll look into this separately.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -787,10 +787,7 @@ class File(QtCore.QObject, Item):
         filename_encoded = encode_filename(self.filename)
         try:
             metadata['~filesize'] = os.path.getsize(filename_encoded)
-        except OSError as ex:
-            log.error(f"File Size Error: {ex}")
 
-        try:
             created = os.path.getctime(filename_encoded)
             created_timestamp = time.strftime(DEFAULT_TIME_FORMAT, time.localtime(created))
             metadata['~file_created_timestamp'] = created_timestamp
@@ -799,7 +796,7 @@ class File(QtCore.QObject, Item):
             modified_timestamp = time.strftime(DEFAULT_TIME_FORMAT, time.localtime(modified))
             metadata['~file_modified_timestamp'] = modified_timestamp
         except OSError as ex:
-            log.error(f"File Timestamps Error: {ex}")
+            log.error(f"File access error: {ex}")
 
     @property
     def state(self):
--- a/picard/file.py
+++ b/picard/file.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2004 Robert Kaye
 # Copyright (C) 2006-2009, 2011-2013, 2017 Lukáš Lalinský
-# Copyright (C) 2007-2011, 2015, 2018-2023 Philipp Wolfer
+# Copyright (C) 2007-2011, 2015, 2018-2024 Philipp Wolfer
 # Copyright (C) 2008 Gary van der Merwe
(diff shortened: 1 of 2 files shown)


picard/file.py
```diff
@@ -780,20 +776,26 @@ def _info(self, metadata, file):
...
+        filename = encode_filename(self.filename)
+        metadata['~dirname'] = os.path.dirname(filename)
+        filename_no_ext, extension = os.path.splitext(os.path.basename(filename))
+        metadata['~filename'] = filename_no_ext
         metadata['~extension'] = extension.lower()[1:]
 
         try:
-            created = os.path.getctime(self.filename)
+            metadata['~filesize'] = os.path.getsize(filename)
+        except BaseException as e:
```
zas, line 790:
```suggestion
        except OSError as e:
```

picard/file.py
```diff
@@ -780,20 +776,26 @@ def _info(self, metadata, file):
...
+            log.error(e)
```
zas, line 791:
```suggestion
            log.error(f"File Size Error: {e}")
```

zas (changes requested):
LGTM, but while at it, the exception handling for `os.path.size()` can be slightly improved and sync with the one for ctime/mtime below. Check suggested changes.

picard/file.py
zas: Or, just move the filesize extraction inside the try/except block below, and make the error message a bit more generic

phw: I updated the exception handling, but kept it separate from the timestamp case

picard/file.py
```diff
@@ -780,20 +776,26 @@ def _info(self, metadata, file):
...
-        filename, extension = os.path.splitext(os.path.basename(self.filename))
-        metadata['~filename'] = filename
+        filename_no_ext, extension = os.path.splitext(os.path.basename(self.filename))
+        metadata['~filename'] = filename_no_ext
         metadata['~extension'] = extension.lower()[1:]
 
+        filename_encoded = encode_filename(self.filename)
+        try:
+            metadata['~filesize'] = os.path.getsize(filename_encoded)
+        except OSError as ex:
```
zas, line 790: As said in a previous comment, I think we can regroup this with ctime/mtime below and use the same try/except block since they all raise OSError.

zas (changes requested):

picard/file.py
zas: > I updated the exception handling, but kept it separate from the timestamp case

I don't see any good reason to keep it separated, because they all raise OSError, a more generic error message can do, plus if we can't get size, it is very likely we can't get ctime/mtime from the same file.

picard/file.py
phw: Ok, I combined it

zas (approved):
LGTM

PR #2364, opened 2024-01-27
PICARD-2832: Add warning when select multiple directories option is enabled

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds a warning dialog when the "Select multiple directories" configuration option is enabled.

# Problem

As discussed in https://tickets.metabrainz.org/browse/PICARD-2829 it is possible for Picard to not recognize external drives when the "Allow selection of multiple directories" option setting is enabled.  This is unexpected behavior for the user.

* JIRA ticket (_optional_): PICARD-2832

# Solution

This PR adds a warning dialog that is presented to the user to inform them of the potential consequence (external drives not being recognized by Picard) when the "Allow selection of multiple directories" option setting is enabled.

This should be considered an interim measure until the issue is resolved in the Qt library.

# Action

If this is acceptable, it might be appropriate to cherry pick the commit into the `2.x` branch in case a future release is required at some point.

### DIFF
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -143,10 +143,13 @@ class InterfaceOptionsPage(OptionsPage):
         if not OS_SUPPORTS_THEMES:
             self.ui.ui_theme_container.hide()
 
-        self.ui.toolbar_multiselect.clicked.connect(self.multi_selection_warning)
+        self.loading = False
+        self.warning_displayed = False
+        self.ui.toolbar_multiselect.stateChanged.connect(self.multi_selection_warning)
 
     def load(self):
         config = get_config()
+        self.loading = True
         self.ui.toolbar_show_labels.setChecked(config.setting['toolbar_show_labels'])
         self.ui.toolbar_multiselect.setChecked(config.setting['toolbar_multiselect'])
         self.ui.show_menu_icons.setChecked(config.setting['show_menu_icons'])
@@ -162,9 +165,13 @@ class InterfaceOptionsPage(OptionsPage):
         self.ui.starting_directory_path.setText(config.setting['starting_directory_path'])
         current_theme = UiTheme(config.setting['ui_theme'])
         self.ui.ui_theme.setCurrentIndex(self.ui.ui_theme.findData(current_theme))
+        self.loading = False
 
     def save(self):
         config = get_config()
+        # Only display warning dialog if it has not been displayed during this Options session.
+        if not self.warning_displayed:
+            self.multi_selection_warning()
         config.setting['toolbar_show_labels'] = self.ui.toolbar_show_labels.isChecked()
         config.setting['toolbar_multiselect'] = self.ui.toolbar_multiselect.isChecked()
         config.setting['show_menu_icons'] = self.ui.show_menu_icons.isChecked()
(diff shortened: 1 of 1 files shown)


Sophist-UK:
If this issue is O/S dependent (i.e. Mac), then the warning should also be O/S dependent.

That said I am as yet unclear whether this is a Qt bug or an O/S config/permissions issue, and whether it is just Macs or Windows and/or Linux also, and unclear as to why the user reports that this is only happening with the latest release and not all previous releases too.

picard/ui/options/interface.py
```diff
@@ -208,5 +210,19 @@ def starting_directory_browse(self):
             path = os.path.normpath(path)
             item.setText(path)
 
+    def multi_selection_warning(self):
+        if self.ui.toolbar_multiselect.isChecked():
+            dialog = QtWidgets.QMessageBox(
+                QtWidgets.QMessageBox.Icon.Warning,
+                _('Option Setting Warning'),
+                _(
+                    'Enabling the multiple directories option setting may result in external drives not being recognized by Picard.\n\n'
```
phw, line 219: I wouldn't make it so specific. It's not really that it does not show external drives per se.

I would rather write e.g.: "When enabling the multiple directories option setting Picard will no longer use the system file picker for selecting directories. This may result in reduced functionality."

picard/ui/options/interface.py
```diff
@@ -143,6 +143,8 @@ def fcmp(x):
         if not OS_SUPPORTS_THEMES:
             self.ui.ui_theme_container.hide()
 
+        self.ui.toolbar_multiselect.clicked.connect(self.multi_selection_warning)
```
phw, line 146: I would instead bind this to the `stateChanged`  signal. That way it does not only trigger on click but generally when the setting is toggled.

phw (changes requested):
Thanks for reacting on my ticket comment so quickly with a patch. That's awesome.

picard/ui/options/interface.py
rdswift: I originally had it that way, but during testing if you enabled the option, then disabled it from the dialog, then enabled it again the dialog was not triggered.  I'll do some more testing to see if I can switch to using `stateChanged`.

Also, should it open the warning dialog when you open the options page (and it loads the saved values) if the option had been saved as enabled?

phw: > Also, should it open the warning dialog when you open the options page (and it loads the saved values) if the option had been saved as enabled?

I think not. So the signal must be connected after the initial value has been set.

rdswift: Now changed to use the `stateChanged` signal.  I also revised the logic of when the dialog is displayed to:

- Do not display warning when loading settings with option enabled
- Display warning when option state has changed and option state is
enabled
- Display warning when saving if option state is enabled and warning has
not been displayed this Options session

rdswift: > So the signal must be connected after the initial value has been set.

Okay, that makes much more sense than the way I had done it (using a `self.loading` variable).  I'll make the change.

Also, after thinking about it again, I'll remove the warning trigger when saving the settings.  The user will have received the warning at the point that they enabled the option initially, so shouldn't be bothered with it each time they save unrelated settings.

I'll also include some logic to not display the warning when changing profile settings (which I had forgotten about).

zas (dismissed):
LGTM

picard/ui/options/interface.py
rdswift: I ended up connecting the signal in the initialization and then blocking it temporarily while loading values in order to avoid having the warning come up when enabling / disabling the option profiles.

phw (approved):

Sophist-UK:
Should this be reverted now that @phw Philipp has submitted and merged #2829 ?

zas:
> Should this be reverted now that @phw Philipp has submitted and merged #2829 ?

I don't think so, because #2829 probably doesn't cover all cases, that's definitely an improvement, but multi selector isn't really equivalent to native dialog, and users may experienced various issues with it.

phw:
@Sophist-UK I don't think so. The general topic remains: Toggling this setting completely switches out the file picker for "Add folder" and everything the native file picker provides is no longer available. That this happens is not obvious.

PR #2365, opened 2024-01-28
PICARD-2829: make volumes available in MultiDirsSelectDialog sidebar

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2829

As discovered in PICARD-2829 when using the "multi dir" directory selection dialog on macOS the user cannot select mounted volumes other than the main volume. That's because by default the dialog only shows the root volume and macOS hides the `/Volumes/` folder.

# Solution

Customize the sidebar of the Qt file dialog to list all mounted volumes. As this is generic and also improves the dialog on other platforms the implementation is not macOS specific. By default on all platforms there were only two entries in the sidebar: "Computer" and the user's home directory.

Extend this to always include:

- Root volume
- User's home
- Primary "Music" folder
- All additionally mounted volumes (on Linux this get filtered to include only those under `/media` or `/mnt`, otherwise all kinds of virtual filesystem end up there).

Here is how this looks on Windows and Linux (sorry, not macOS screenshot today, but essentially the same):

grafik

grafik

There is one caveat: By default the sidebar of Qt's file dialog shows an entry "Computer" which on Linux and macOS gives only access to the main volume, on Windows lists all drives.

I found not way to make this specific location show up in the sidebar if the sidebar gets customized. But it is still accessible via the path dropdown at the top, as can be seen in the screenshots. Also I believe on none of the platforms it was actually very useful and having the volumes in the sidebar is more convenient.

### DIFF
--- a/picard/ui/util.py
+++ b/picard/ui/util.py
@@ -8,7 +8,7 @@
 # Copyright (C) 2017 Frederik “Freso” S. Olesen
 # Copyright (C) 2017 Sophist-UK
 # Copyright (C) 2018 Vishal Choudhary
-# Copyright (C) 2019, 2021-2022 Philipp Wolfer
+# Copyright (C) 2019, 2021-2022, 2024 Philipp Wolfer
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -24,7 +24,6 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-
 from PyQt6 import (
     QtCore,
     QtGui,
@@ -33,6 +32,7 @@ from PyQt6 import (
 
 from picard.config import get_config
 from picard.const.sys import (
+    IS_LINUX,
     IS_MACOS,
     IS_WIN,
 )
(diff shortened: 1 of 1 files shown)


picard/ui/util.py
```diff
@@ -89,6 +89,29 @@ def __init__(self, *args):
...
+                path = volume.rootPath()
+                if volume.isRoot():
+                    root_volume = path
+                else:
+                    if not IS_LINUX or (path.startswith("/media/") or path.startswith("/mnt/")):
+                        volume_paths.append(path)
+        sidebar_urls = [
+            QtCore.QUrl.fromLocalFile(root_volume),
+            QtCore.QUrl.fromLocalFile(QtCore.QDir.homePath()),
+            QtCore.QUrl.fromLocalFile(QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.StandardLocation.MusicLocation))
```
zas, line 109: According to https://doc.qt.io/qt-6/qstandardpaths.html#writableLocation `QtCore.QStandardPaths.writableLocation` could return an empty string, though `QtCore.QStandardPaths.StandardLocation.MusicLocation` is said to default to a sensible location, perhaps safer to check the result and not add it if empty.

See https://github.com/phw/picard/pull/21

zas (changes requested):

zas (approved):
LGTM

PR #2367, opened 2024-01-28
Handle exceptions if ~filesize is not numeric

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem
This is a bit of an edge case, but should `~filesize` be non-numeric `bytes2human` would throw a `ValueError`. This makes sure Picard keeps functioning.


zas (approved):
LGTM

PR #2369, opened 2024-02-12
Drop python 3.8

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

By the time Picard 3 will be released official support for Python 3.8 will be already or nearly at its end. Also pygit2 dropped support for Python 3.8, and we might want to use that for the new plugin system.

This PR also removes some gettext workaround for missing pgettext in Python 3.7 and lower.


zas (approved):
LGTM

PR #2370, opened 2024-02-26
PICARD-2468: Properly handle unspecified languages in ID3v2 comments

Serial-ATA:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
Picard no longer creates a duplicate `COMM` frame with the incorrect language when using `lang='XXX' + desc=''`.
# Problem

Previously, a `COMM` frame with `lang='XXX' desc=''` would be saved as a separate frame with `lang='eng' desc='XXX'`. This is due to Picard supporting both `comment:desc` and `comment:lang:desc` syntax for comment tags.

* JIRA ticket (_optional_): PICARD-2468

# Solution

A special case was added for `lang='XXX', desc=''`. I could not think of a better solution for the problem, as Picard normalizes all metadata keys. It would be more elegant if trailing colons were preserved.

# Action

Possibly explore storing these comments as `comment:XXX:`, leaving the trailing colon as a marker for an empty description. This solution is not perfect. In the event that someone has `comment:eng:XXX`, all `eng` comments are simply stored as `comment:desc`. With this change, their language would be overwritten with `XXX`.

### DIFF
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -177,13 +177,22 @@ def parse_comment_tag(name):  # noqa: E302
     If language is not set ("comment:desc") "eng" is assumed as default.
     Returns a (lang, desc) tuple.
     """
-    try:
-        desc = name.split(':', 1)[1]
-    except IndexError:
-        desc = ''
     lang = 'eng'
+    desc = ''
+
+    split = name.split(':', 1)
+    if len(split) > 1:
+        desc = split[1]
+
     match = RE_COMMENT_LANG.match(desc)
     if match:
         lang = match.group(1)
         desc = desc[4:]
-    return (lang, desc)
+        return lang, desc
+
+    # Special case for unspecified language + empty description
+    if desc == 'XXX':
+        lang = 'XXX'
+        desc = ''
+
+    return lang, desc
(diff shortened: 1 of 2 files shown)


phw:
Thanks a lot for this. I think as a workaround this probably makes sense and we could include it even in the next Picard 2.x release.

Essentially this is issue happens because the language notation got introduced as an afterthought (https://github.com/metabrainz/picard/commit/4d85c2e31fc0fe9dfd71f9e2989bf4a699bfa7be). Similar issues plague the still open PR #1650 for doing the same to the lyrics tag.

I guess we should reconsider this and make a harder cut for Picard 3, making the language not optional. I' thinking of having both colons be present even if language is empty, such as `comment::desc`. @zas what do you think?

zas:
> I guess we should reconsider this and make a harder cut for Picard 3, making the language not optional. I' thinking of having both colons be present even if language is empty, such as `comment::desc`.

Yes, I agree.

phw (changes requested):
Ok, let's go with this. The code itself is good to go. Please only add a test for this case to https://github.com/metabrainz/picard/blob/master/test/test_util_tags.py#L37

Also please rebase this branch. CI is currently failing due to changes to the Github Action Windows image, and this has been fixed on the master branch.

Serial-ATA:
Hey @phw is there anything else I need to do for this?

zas (approved):
LGTM

PR #2372, opened 2024-03-12
PICARD-2716: Accept encodings other than utf-8 when opening CD extraction logs

ShubhamBhut:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2716

# Solution
Update  `detect_unicode_encoding()` function in `picard/util/__init__.py` to recognize more encodings. Since Python natively supports reading those encoding files by their names, no further changes are necessary.

# Action

modified `detect_unicode_encoding()` using `chardet` to detect other encoding formats.  Additionally, included a test log file of type windows-1251 to validate the functionality of the modified function


picard/util/__init__.py
```diff
@@ -1169,9 +1170,9 @@ def strxfrm(string):
 
 
 def detect_unicode_encoding(path):
```
phw, line 1172: I'd renamed this to `detect_file_encoding` (because it no longer is just for Unicode)

requirements.txt
```diff
@@ -11,3 +11,4 @@ PyQt6-Qt6~=6.5.3; sys_platform == 'darwin'
 python-dateutil~=2.7
 pywin32; sys_platform == 'win32'
 PyYAML>=5.1, <7
+chardet~=5.2.0
```
phw, line 14: Because of the way how we use separate requirements files for specific binary builds (on master this is currently `requirements-win.txt` and `requirements-macos-11.txt`) you need to add it there as well. Please do so with a fixed version:

```
chardet==5.2.0
```

But keep it as `chardet~=5.2.0` here in the generic requirements.txt.

test/test_utils.py
```diff
@@ -961,3 +964,14 @@ def test_detect_encoding(self):
...
+    def test_detect_encoding_file(self):
+        test_files = {
+            get_test_data_path('eac-utf16le.log'): 'utf-16-le',
+            # get_test_data_path('eac-utf8.log'): 'utf-8',
+            get_test_data_path('eac-windows1251.log'): 'windows-1251'
+        }
+
+        for file_path, expected_encoding in test_files.items():
+            self.assertEqual(expected_encoding, detect_unicode_encoding(file_path))
+
```
phw, line 977: blake8 is complaining about this extra blank line

test/test_utils.py
```diff
@@ -961,3 +964,14 @@ def test_detect_encoding(self):
             finally:
                 f.close()
                 os.remove(f.name)
+
+    def test_detect_encoding_file(self):
+        test_files = {
+            get_test_data_path('eac-utf16le.log'): 'utf-16-le',
+            # get_test_data_path('eac-utf8.log'): 'utf-8',
```
phw, line 971: Is there a problem with the utf8 file specifically?

picard/util/__init__.py
```diff
@@ -1183,4 +1184,10 @@ def detect_unicode_encoding(path):
         for bom, encoding in ENCODING_BOMS.items():
             if first_bytes.startswith(bom):
                 return encoding
-        return 'utf-8'
+        rawdata = f.read()
+        result = chardet.detect(rawdata)
```
phw, line 1188: Should we do `first_bytes + rawdata` here, just in case that we don't muss probably relevant bytes?

phw (commented):
Thanks a lot for the patch. This looks good. Only some minor comments and questions from my side.

phw (changes requested):
Thanks a lot for the patch. This looks good. Only some minor comments and questions from my side.

picard/util/__init__.py
```diff
@@ -1183,4 +1184,10 @@ def detect_unicode_encoding(path):
...
+        encoding = result.get('encoding')
+        if encoding is None:
```
zas, line 1190: I think we should check `confidence` too.

What about something like:

```python
try:
    result = chardet.detect(rawdata)
    if result['confidence'] > 0.5:
        return result['encoding'].lower()
except Exception as e:
    log.warning("Couldn't detect encoding for file %r: %s", path, e)
    pass
return 'utf-8'
```

picard/util/__init__.py
```diff
@@ -1169,9 +1170,9 @@ def strxfrm(string):
 
 
 def detect_unicode_encoding(path):
-    """Attempts to guess the unicode encoding of a file based on the BOM.
+    """Attempts to guess the unicode encoding of a file based on the BOM and using chadet.
```
zas, line 1173:
```suggestion
    """Attempts to guess the unicode encoding of a file based on the BOM and using chardet.
```

requirements.txt
zas, line 14: This extra dependency bothers me a bit, especially to solve such a corner case.

picard/util/__init__.py
zas, line 1172: After this patch, this name doesn't look appropriate, since it detects any encoding. May be rename it to `detect_file_encoding`?

picard/util/__init__.py
```diff
@@ -1183,4 +1184,10 @@ def detect_unicode_encoding(path):
         for bom, encoding in ENCODING_BOMS.items():
```
zas, line 1184: If we're going to use `chardet` this code is likely a bit redundant.

requirements-dev.txt
```diff
@@ -5,3 +5,4 @@ isort>=5.0
 pycodestyle
 pylint>=2.6.0
 wlc
+pytest~=7.4.3
```
zas, line 8: Needed??

zas (changes requested):

requirements.txt
phw: We did do the BOM detection to avoid extra dependency, and because all the examples we had where using some kind of Unicode encoding. But if we need to assume that essentially any text encoding is possible and we need to detect it I don't think we get around adding an additional dependency. The only alternative would be to reimplement chardet's functionality, and that is far worse.

requirements-dev.txt
ShubhamBhut: shall I remove this ? I thought of adding `pytest` as any developer would need to run tests.

test/test_utils.py
ShubhamBhut: oh right, I forgot to mention it. Actually the current detector and a couple of online detectors are considering this log file of type `ascii` and one strange detector considered the type `iso-8859-1`.

requirements.txt
zas: > that i18n is a corner case.

Not sure what you are talking about, this isn't about i18n but detection of historical charsets used by externally generated files that Picard may need to read, this PR is about solving a very specific use case.
This detection is only a guess, and may not work in all cases.

Unicode is around since more than 30 years.

> But if we need to assume that essentially any text encoding is possible and we need to detect it I don't think we get around adding an additional dependency. The only alternative would be to reimplement chardet's functionality, and that is far worse.

To me, the question is:
should we add support for that  (and the best way is clearly `chardet`) or just decide Picard doesn't support them and let user convert his files to Unicode before trying to load them in Picard.

Not a big deal here, I'm ok to support such files whenever possible but better think twice before adding any extra dependency, as we know it will always translate to extra work for us at some point (and it starts here).

zas: BTW, there are alternatives to `chardet` like https://pypi.org/project/charset-normalizer/

picard/util/__init__.py
ShubhamBhut: actually chardet doesn't seem to differentiate well between various `utf-16` and `utf-32` encodings. The BOM based test function failed as chardet considered all 4 to be of type utf-8:

            b'\xff\xfe': 'utf-16-le',
            b'\xfe\xff': 'utf-16-be',
            b'\00\00\xff\xfe': 'utf-32-le',
            b'\00\00\xfe\xff': 'utf-32-be',
            
   however it worked correctly for the log file

requirements.txt
ShubhamBhut: oh, didn't notice that, I will check them out and see if any one fixes some [issues](https://github.com/metabrainz/picard/pull/2372#discussion_r1521376049) we have with `chardet`

test/test_utils.py
phw: Ah, yes. I see. Actually they are all correct :) As the data contains only ASCII characters (and no BOM), all of a `ascii`, `iso-8859-1` and `utf-8` result in exactly the same bytes.

I think it's allright to leave this specific case out of the test. We could maybe replace it with a file where UTF-8 encoding actually makes a difference. But we don't need to do this as part of this PR.

picard/util/__init__.py
phw: > Also this page says that 'utf-32-le' has a BOM of `b'\xff\xfe\x00\x00'`. That said utf-32 is pretty rare as are the other encodings in the table on the Wikipedia page.

I think it's just a bug and the existing entry of `b'\00\00\xff\xfe'` needs to be updated.

picard/util/__init__.py
phw: That was also my suggestion above

picard/util/__init__.py
phw: > If we're going to use chardet this code is likely a bit redundant.

I probably would keep this. If we get a BOM I think it is fair to assume the corresponding encoding, and we avoid the messy stuff that character encoding guessing is. And the code is simple enough.

picard/util/__init__.py
```diff
@@ -1183,4 +1184,10 @@ def detect_unicode_encoding(path):
...
             if first_bytes.startswith(bom):
                 return encoding
-        return 'utf-8'
+        rawdata = f.read()
```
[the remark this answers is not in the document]
phw: It could be limited to a certain amount of bytes to avoid accidentally loading excessively huge files. Not sure how many, though. If you read not enough you risk reading a file with wrong encoding. Imagine a iso-8859-1 encoded file where all is ASCII but the last track title contains an french accent. If you miss reading the last track the guessed encoding could be ASCII or UTF-8, both resulting in problems reading the last title.

So I would limit reading to a byte count where we can assume a typical disc ID log file fits in. The largest log in our test data is currently 11kb. If we limit reading to a couple of MB maybe we would be safe that we get the data, but still small enough that there is no risk of extreme delays in reading the file or even causing major issue like full RAM and swapping.

picard/util/__init__.py
ShubhamBhut: I tried chardet alternative and the best one `chardet_normalizer` also struggles (more) with above 4 types. It also doesn't have give any confidence value, therefore [checking confidence](https://github.com/metabrainz/picard/pull/2372#discussion_r1521161792) won't be possible with it. However it is considered faster and more accurate model generally. Which package shall I go with for now ?

ShubhamBhut: > According to [Wikipedia](https://en.wikipedia.org/wiki/Byte_order_mark) (yes, I know that this is not necessarily definitive or correct), we should also check for `b'\xef\xbb\xbf'` as 'utf-8'.
> 
> Also this page says that 'utf-32-le' has a BOM of `b'\xff\xfe\x00\x00'`. That said utf-32 is pretty rare as are the other encodings in the table on the Wikipedia page.

- utf-8 checking is also available in current test function and it is detected perfectly.
- Also considering `chardet` struggles with these 4 encodings detection, so should I totally remove BOM based detection or keep it for now ?

zas: If we give up on checking confidence, we could support both libs, if we stick to https://charset-normalizer.readthedocs.io/en/latest/user/getstarted.html#backward-compatibility style.

Imho, it could be good in case one of those goes unmaintained.

I would choose `chardet_normalizer` over `chardet`, because it's smaller, faster, and detect more charsets.
@phw what's your take on this?

picard/util/__init__.py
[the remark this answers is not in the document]
zas: We could limit size to something like 256kb, as it is very unlikely a log file is ever bigger than this.
But if we go this way, we need to pass this limit (as optional, with default) to the method, since this function is not specific to log files (even though it is only used for those atm).

picard/util/__init__.py
phw: I agree, supporting both with the compatbility API, but preferring chardet_normalizer. Which means we should try to import chardet_normalizer and if that fail try chardet. But in requirements-*.txt files we use chardet_normalizer.

Makes sense?

> Also considering chardet struggles with these 4 encodings detection, so should I totally remove BOM based detection or keep it for now ?

I would still opt for keeping it. @zas? We could even make chardet/chardet_normalizer completely optional that way

picard/util/__init__.py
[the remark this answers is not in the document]
phw: Sounds good let's go with 256kb as default, but make it a parameter to `detect_file_encoding`

picard/util/__init__.py
zas: > > 
> 
> I would still opt for keeping it. @zas? We could even make chardet/chardet_normalizer completely optional that way

I agree.

picard/util/__init__.py
```diff
@@ -1183,4 +1185,10 @@ def detect_unicode_encoding(path):
         for bom, encoding in ENCODING_BOMS.items():
             if first_bytes.startswith(bom):
                 return encoding
-        return 'utf-8'
+        rawdata = f.read(read_limit)
```
zas, line 1188:
```python
if detect is None:
    return 'utf-8'

rawdata = f.read(read_limit)
....
```

zas (changes requested):

zas:
@ShubhamBhut Can you fix style errors? See failing checks.

ShubhamBhut:
> @ShubhamBhut Can you fix style errors? See failing checks.

Sincere apologies on these errors. I didn't had set up the suggested pre-commit hook previously.

zas:
> > 
> 
> Sincere apologies on these errors. I didn't had set up the suggested pre-commit hook previously.

No problem ;)

picard/util/__init__.py
```diff
@@ -1183,4 +1190,14 @@ def detect_unicode_encoding(path):
         for bom, encoding in ENCODING_BOMS.items():
             if first_bytes.startswith(bom):
                 return encoding
-        return 'utf-8'
+
+        if detect is None:
+            return 'utf-8'
+
+        rawdata = f.read(read_limit)
+        result = detect(first_bytes + rawdata)
```
zas, line 1198: Instead of creating a new string by concatenation, we should just use `seek()`:

```python
f.seek(0)
result = detect(f.read(read_limit))
```

`rawdata` intermediary variable isn't strictly needed either.

zas (changes requested):

test/test_utils.py
```diff
@@ -957,7 +960,16 @@ def test_detect_encoding(self):
...
                 f.close()
                 os.remove(f.name)
+
+    def test_detect_file_encoding_files(self):
+        test_files = {
+            get_test_data_path('eac-utf16le.log'): 'utf-16-le',
+            get_test_data_path('eac-windows1251.log'): 'windows-1251'
+        }
+
+        for file_path, expected_encoding in test_files.items():
```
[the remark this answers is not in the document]
zas: Hmmm, after some thoughts, we could even split this in two test functions.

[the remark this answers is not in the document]
ShubhamBhut: Got it, just wondering (as a beginner programmer) -  why 2 separate functions preferred over 1 ? considering more files might get added in the future.

zas (approved):
LGTM

test/test_utils.py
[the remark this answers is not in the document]
zas: @ShubhamBhut mainly because each test function can be executed independently, and it's easier to disable a test that is failing on one platform.

We often use loops within test functions, but it somehow goes against the principle of *unit* testing.
In this specific case, we only have 2 elements in the loop, so making 2 separate functions isn't not much a problem.

We could improve that (and still use loops) by using a feature that appeared in Python 3.4: subtests (see https://docs.python.org/3.4/library/unittest.html?highlight=subtest#distinguishing-test-iterations-using-subtests).

picard/util/__init__.py
```diff
@@ -1168,10 +1174,11 @@ def strxfrm(string):
...
-    """Attempts to guess the unicode encoding of a file based on the BOM.
+def detect_file_encoding(path, read_limit=1024*256):
+    """Attempts to guess the unicode encoding of a file based on the BOM, and
+    depending on avalibility, using a charset detection method.
 
-    Assumes UTF-8 by default if there is no BOM.
+    Assumes UTF-8 by default if no other encoding is detected.
 
     Args:
         path: The path to the file
```
zas, line 1184: Please document new `read_limit` here.

BTW, I wonder if a name likes `max_bytes_to_read` would be better or not. What do you think?

zas (changes requested):

picard/util/__init__.py
ShubhamBhut: I think `max_bytes_to_read` will be much better variable name. Adding it.

zas (approved):
Good job on this PR, thanks for your contribution.

LGTM

phw (approved):
This is now looking really good. Thanks a lot.

csaavedra:
Thanks for doing this!

PR #2373, opened 2024-03-15
PICARD-1092: Add support for preserving the SYLT tag

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
Allow to preserve the SYLT tag. Also add support for reading and writing the SYLT tag of ID3.

# Problem

* JIRA ticket (_optional_): [PICARD-1092](https://tickets.metabrainz.org/browse/PICARD-1092)

Like @phw said in the comments of the ticket, besides preserving the SYLT tag, it would be nice if Picard could both read and write this tag.

# Solution

As it was suggested, this solution converts the SYLT tag to the LRC format for display, and it converts the LRC format back into the SYLT tag when saving.

# Action

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -35,7 +35,6 @@
 
 from collections import Counter
 from enum import IntEnum
-from itertools import batched
 import re
 from urllib.parse import urlparse
 
@@ -63,8 +62,24 @@ from picard.util import (
     encode_filename,
     sanitize_date,
 )
-from picard.util.tags import parse_comment_tag
+from picard.util.tags import (
+    parse_comment_tag,
+    parse_lyrics_tag,
+)
 
+
+try:
+    from itertools import batched
+except ImportError:
+    # itertools.batched is only available in Python >= 3.12
+    from itertools import islice
+
+    def batched(iterable, n):
+        if n < 1:
+            raise ValueError('n must be at least one')
+        it = iter(iterable)
+        while batch := tuple(islice(it, n)):
+            yield batch
 
 UNSUPPORTED_TAGS = {'r128_album_gain', 'r128_track_gain'}
 
@@ -230,7 +245,8 @@ class ID3File(File):
         'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$')
     }
 
-    __lrc_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d])')
+    __lrc_line_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\])')
+    __lrc_syllable_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\]|<\d\d:\d\d\.\d\d>)')
 
     def __init__(self, filename):
         super().__init__(filename)
@@ -333,19 +349,17 @@ class ID3File(File):
                 if frame.desc:
                     name += ':%s' % frame.desc
                 metadata.add(name, frame.text)
-            elif frameid == 'SYLT':
+            elif frameid == 'SYLT' and frame.type == 1:
+                if frame.format == 0:
+                    raise NotImplementedError("SYLT format 0 is not supported")
                 name = 'syncedlyrics'
-                if frame.desc:
-                    name += ':%s' % frame.desc
-
-                lrc_lyrics = []
-                for lyrics, milliseconds in frame.text:
-                    minutes = milliseconds // (60 * 1000)
-                    seconds = (milliseconds % (60 * 1000)) // 1000
-                    hundredths = (milliseconds % 1000) // 10
-                    lrc_lyrics.append(f"[{minutes:02d}:{seconds:02d}.{hundredths:02d}]{lyrics.strip()}")
-                lrc_lyrics = "\n".join(lrc_lyrics)
-
+                if frame.lang:
+                    name += ':%s' % frame.lang
+                    if frame.desc:
+                        name += ':%s' % frame.desc
+                elif frame.desc:
+                    name += '::%s' % frame.desc
+                lrc_lyrics = self._parse_sylt_text(frame.text)
                 metadata.add(name, lrc_lyrics)
             elif frameid == 'UFID' and frame.owner == "http://musicbrainz.org":
                 metadata['musicbrainz_recordingid'] = frame.data.decode('ascii', 'ignore')
@@ -476,21 +490,13 @@ class ID3File(File):
                     desc = ''
                 for value in values:
                     tags.add(id3.USLT(encoding=encoding, desc=desc, text=value))
-            elif name.startswith('syncedlyrics:') or name == 'syncedlyrics':
-                if ':' in name:
-                    desc = name.split(':', 1)[1]
-                else:
-                    desc = ''
+            elif name == 'syncedlyrics' or name.startswith('syncedlyrics:'):
+                (lang, desc) = parse_lyrics_tag(name)
                 for value in values:
-
-                    sylt_lyrics = []
-                    timestamp_and_lyrics = batched(re.split(self.__lrc_re_parse, value)[1:], 2)
-                    for timestamp, lyrics in timestamp_and_lyrics:
-                        minutes, seconds, hundredths = timestamp[1:-1].replace(".", ":").split(':')
-                        milliseconds = int(minutes) * 60 * 1000 + int(seconds) * 1000 + int(hundredths) * 10
-                        sylt_lyrics.append((lyrics, milliseconds))
-
-                    tags.add(id3.SYLT(encoding=encoding, desc=desc, text=sylt_lyrics))
+                    sylt_lyrics = self._parse_lrc_text(value)
+                    # If the text does not contain any timestamps, the tag is not added
+                    if sylt_lyrics:
+                        tags.add(id3.SYLT(encoding=encoding, lang=lang, format=2, type=1, desc=desc, text=sylt_lyrics))
             elif name in self._rtipl_roles:
                 for value in values:
                     tipl.people.append([self._rtipl_roles[name], value])
@@ -612,13 +618,11 @@ class ID3File(File):
                     for key, frame in list(tags.items()):
                         if frame.FrameID == 'USLT' and frame.desc == desc:
                             del tags[key]
-                elif name.startswith('syncedlyrics:') or name == 'syncedlyrics':
-                    if ':' in name:
-                        desc = name.split(':', 1)[1]
-                    else:
-                        desc = ''
+                elif name == 'syncedlyrics' or name.startswith('syncedlyrics:'):
+                    (lang, desc) = parse_lyrics_tag(name)
                     for key, frame in list(tags.items()):
-                        if frame.FrameID == 'SYLT' and frame.desc == desc:
+                        if frame.FrameID == 'SYLT' and frame.desc == desc and frame.lang == lang \
+                                and frame.type == 1 and frame.format == 2:
                             del tags[key]
                 elif name in self._rtipl_roles:
                     role = self._rtipl_roles[name]
@@ -723,6 +727,31 @@ class ID3File(File):
             values = [join_with.join(values)]
 
         return values
+
+    def _parse_sylt_text(self, text):
+        lrc_lyrics = []
+        previous_line_ended = True
+        for lyrics, milliseconds in text:
+            minutes = milliseconds // (60 * 1000)
+            seconds = (milliseconds % (60 * 1000)) // 1000
+            hundredths = (milliseconds % 1000) // 10
+            if previous_line_ended:
+                lrc_lyrics.append(f"[{minutes:02d}:{seconds:02d}.{hundredths:02d}]{lyrics}")
+                previous_line_ended = False
+            else:
+                lrc_lyrics.append(f"<{minutes:02d}:{seconds:02d}.{hundredths:02d}>{lyrics}")
+            if '\n' in lyrics:
+                previous_line_ended = True
+        return "".join(lrc_lyrics)
+
+    def _parse_lrc_text(self, text):
+        sylt_lyrics = []
+        timestamp_and_lyrics = batched(self.__lrc_syllable_re_parse.split(text)[1:], 2)
+        for timestamp, lyrics in timestamp_and_lyrics:
+            minutes, seconds, hundredths = timestamp[1:-1].replace(".", ":").split(':')
+            milliseconds = int(minutes) * 60 * 1000 + int(seconds) * 1000 + int(hundredths) * 10
+            sylt_lyrics.append((lyrics, milliseconds))
+        return sylt_lyrics
 
 
 class MP3File(ID3File):
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -64,7 +64,7 @@ from picard.util import (
 )
 from picard.util.tags import (
     parse_comment_tag,
-    parse_lyrics_tag,
+    parse_subtag,
 )
 
 
@@ -245,8 +245,10 @@ class ID3File(File):
         'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$')
     }
 
-    __lrc_line_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\])')
-    __lrc_syllable_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\]|<\d\d:\d\d\.\d\d>)')
+    __lrc_line_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\])')
+    __lrc_syllable_re_parse = re.compile(r'(<\d\d:\d\d\.\d\d\d>)')
+    __lrc_syllable_at_the_end_re_parse = re.compile(r'(<\d\d:\d\d\.\d\d\d>$)')
+    __lrc_both_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\]|<\d\d:\d\d\.\d\d\d>)')
 
     def __init__(self, filename):
         super().__init__(filename)
@@ -491,7 +493,7 @@ class ID3File(File):
                 for value in values:
                     tags.add(id3.USLT(encoding=encoding, desc=desc, text=value))
             elif name == 'syncedlyrics' or name.startswith('syncedlyrics:'):
-                (lang, desc) = parse_lyrics_tag(name)
+                (lang, desc) = parse_subtag(name)
                 for value in values:
                     sylt_lyrics = self._parse_lrc_text(value)
                     # If the text does not contain any timestamps, the tag is not added
@@ -619,7 +621,7 @@ class ID3File(File):
                         if frame.FrameID == 'USLT' and frame.desc == desc:
                             del tags[key]
                 elif name == 'syncedlyrics' or name.startswith('syncedlyrics:'):
-                    (lang, desc) = parse_lyrics_tag(name)
+                    (lang, desc) = parse_subtag(name)
                     for key, frame in list(tags.items()):
                         if frame.FrameID == 'SYLT' and frame.desc == desc and frame.lang == lang \
                                 and frame.type == 1 and frame.format == 2:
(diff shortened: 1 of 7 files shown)


picard/formats/id3.py
```diff
@@ -330,6 +333,20 @@ def _load(self, filename):
...
+                if frame.desc:
+                    name += ':%s' % frame.desc
+
+                lrc_lyrics = []
+                for lyrics, milliseconds in frame.text:
+                    minutes = milliseconds // (60 * 1000)
+                    seconds = (milliseconds % (60 * 1000)) // 1000
+                    hundredths = (milliseconds % 1000) // 10
+                    lrc_lyrics.append(f"[{minutes:02d}:{seconds:02d}.{hundredths:02d}]{lyrics.strip()}")
+                lrc_lyrics = "\n".join(lrc_lyrics)
```
zas, line 347: We should move this code to a separate method in order to ease testing.

Something like:

```python
def parse_sylt_text(text):
    for lyrics, milliseconds in text:
        minutes = milliseconds // (60 * 1000)
        seconds = (milliseconds % (60 * 1000)) // 1000
        hundredths = (milliseconds % 1000) // 10
        yield f"[{minutes:02d}:{seconds:02d}.{hundredths:02d}]{lyrics.strip()}"
```

```python
lrc_lyrics = "\n".join(parse_sylt_text(frame.text))
```

This way, unit tests can be written for `parse_sylt_text()`

picard/formats/id3.py
```diff
@@ -229,6 +230,8 @@ class ID3File(File):
         'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$')
     }
 
+    __lrc_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d])')
```
zas, line 233: Not strictly needed, but:

```suggestion
    __lrc_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\])')
```

picard/formats/id3.py
```diff
@@ -459,6 +476,21 @@ def _save(self, filename, metadata):
...
+                else:
+                    desc = ''
+                for value in values:
+
+                    sylt_lyrics = []
+                    timestamp_and_lyrics = batched(re.split(self.__lrc_re_parse, value)[1:], 2)
+                    for timestamp, lyrics in timestamp_and_lyrics:
+                        minutes, seconds, hundredths = timestamp[1:-1].replace(".", ":").split(':')
+                        milliseconds = int(minutes) * 60 * 1000 + int(seconds) * 1000 + int(hundredths) * 10
+                        sylt_lyrics.append((lyrics, milliseconds))
```
zas, line 491: Same here, better move this code to a separate small method, and write tests for it.

picard/formats/id3.py
```diff
@@ -580,6 +612,14 @@ def _remove_deleted_tags(self, metadata, tags):
                     for key, frame in list(tags.items()):
                         if frame.FrameID == 'USLT' and frame.desc == desc:
                             del tags[key]
+                elif name.startswith('syncedlyrics:') or name == 'syncedlyrics':
```
zas, line 615: Which one of those both cases is the more likely?

test/formats/test_id3.py
```diff
@@ -370,6 +370,13 @@ def test_lyrics_with_description(self):
             loaded_metadata = save_and_load_metadata(self.filename, metadata)
             self.assertEqual(metadata['lyrics:foo'], loaded_metadata['lyrics:foo'])
 
+        @skipUnlessTestfile
+        def test_syncedlyrics_converting_to_and_from_lrc(self):
+            syncedlyrics = "[00:00.00]first bar\n[00:01.00]second bar"
```
zas, line 375: Invalid tag values should also be tested.

zas (changes requested):

picard/formats/id3.py
```diff
@@ -35,6 +35,7 @@
 
 from collections import Counter
 from enum import IntEnum
+from itertools import batched
```
zas, line 38: Only appeared in Python 3.12.

picard/formats/id3.py
```diff
@@ -330,6 +333,20 @@ def _load(self, filename):
                 if frame.desc:
                     name += ':%s' % frame.desc
                 metadata.add(name, frame.text)
+            elif frameid == 'SYLT':
```
Sophist-UK, line 336: IMO they are not synced lyrics unless also `frame.type== 1`. (See [SYLT](https://mutagen-specs.readthedocs.io/en/latest/id3/id3v2.4.0-frames.html#synchronised-lyrics-text).)

But if we are going to accept any SYLT type, then we probably need to preserve it since you might have multiple types.

We also absolutely need to preserve the lyrics language - because you might have several SYLT frames with different languages. How about naming the metadata `syncedlyrics:lang:desc`?

We may also need to handle (or raise an exception) if `frame.format == 0` (i.e. timestamp is in MPEG frames rather than millisecs).

I am unclear whether we need to preserve the `frame.encoding` value and/or e.g. convert the contents from ISO-8859-1 to UNICODE or vice-versa to standardise the metadata.

Sophist-UK (commented):
This seems pretty good in principle.

I appreciate that the intent of this PR is to preserve (and display) ID3 SYLT tags, however the same concept applies equally to other tagging formats so, in addition to the comments below, I would also like to see support for loading and saving the same `syncedlyrics` metadata from other formats.

picard/formats/id3.py
Sophist-UK, line 347: This is NOT my understanding of the SYLT tags - which are (supposedly) syllable based and have embedded "\n" to indicate a new line. These embedded newlines may well be preserved in addition to those inserted by the  later join.

The LRC format seems to me to be a subset whereby each line of an LRC corresponds to a line of the lyrics, and the lyrics have a single start time-stamp for each line.

This PR therefore needs to have a different approach to newlines than the one here if it is to preserve the SYLT tag exactly as-is. We either need to e.g. simplify the SYLT by converting to LRC and combining multiple syllables under a single time-stamp - in which case data will be lost when you save it - or use a modified LRC format whereby we allow multiple timestamps per line.

picard/formats/id3.py
```diff
@@ -459,6 +476,21 @@ def _save(self, filename, metadata):
...
                     tags.add(id3.USLT(encoding=encoding, desc=desc, text=value))
+            elif name.startswith('syncedlyrics:') or name == 'syncedlyrics':
+                if ':' in name:
+                    desc = name.split(':', 1)[1]
+                else:
+                    desc = ''
+                for value in values:
+
+                    sylt_lyrics = []
+                    timestamp_and_lyrics = batched(re.split(self.__lrc_re_parse, value)[1:], 2)
```
Sophist-UK, line 487: This appears to assume that the metadata is in valid LRC format. If the user edits the data or if the data has been provided by a plugin this may not be the case. Also we need to handle newlines as described in an earlier comment.

picard/formats/id3.py
phw: > IMO they are not synced lyrics unless also frame.type== 1. (See SYLT.)
>
> But if we are going to accept any SYLT type, then we probably need to preserve it since you might have multiple types.

I think it should only read / write type 1 lyrics. Otherwise the becomes too complex for the tag name. Other types should use other tag names if needed.

> We also absolutely need to preserve the lyrics language - because you might have several SYLT frames with different languages. How about naming the metadata syncedlyrics:lang:desc?

Absolutely. And we should from the beginning start with a syntax that avoids ambiguity between syncedlyrcis:lang and syncedlyrics:description. As I proposed in PICARD-2844 I would use `syncedlyrics::description`  for the case where language is empty but description is set.

> I am unclear whether we need to preserve the frame.encoding value and/or e.g. convert the contents from ISO-8859-1 to UNICODE or vice-versa to standardise the metadata.

Encoding is handled transparently already.

picard/formats/id3.py
```diff
@@ -330,6 +333,20 @@ def _load(self, filename):
                 if frame.desc:
                     name += ':%s' % frame.desc
                 metadata.add(name, frame.text)
+            elif frameid == 'SYLT':
+                name = 'syncedlyrics'
```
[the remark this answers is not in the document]
phw: An argument for syncedlyrics might be that unsyncedlyrics is commonly used for e.g. Vorbis/FLAC tags.

phw (changes requested):
Thank you, I appreciate your initiative very much. Adding synced lyrics support will need some work, so I hope you will stay with us to get this through. Not all need to be done as part of a single PR, but we at least need to identify what needs to be done next. 

Most has already been said in the other reviews. I just want to stress three points:

1. Converting LRC to/from SYLT really needs to go into separate functions with a set of tests that handles different cases, including invalid ones.

2. Other formats should be handled as well. I don't know how this is currently done in those formats. Does anyone have some experience with synced lyrics used in Vorbis? Would we just store the LRC format there? ASF has this not useful gem in its documentation for [WM/Lyrics_Synchronised](https://learn.microsoft.com/en-us/previous-versions/ms867702(v=msdn.10)#wmlyrics_synchronised):
    > For more information about how to use synchronized lyrics, see the documentation for the technology you are using.
    
    However, I also think we can handle separate formats in separate PRs, especially if they are more complex. But then we should probably make sure the syncedlyrics tag is not yet written to all other formats to avoid issues later on.

3. The `syncedlyrics` tag needs to handle language from the beginning. I would suggest to follow my proposal on PICARD-2844. We should get this right this time from the beginning and avoid the mistakes done with the comment and lyrics tags.

picard/formats/id3.py
twodoorcoupe: > Absolutely. And we should from the beginning start with a syntax that avoids ambiguity between syncedlyrcis:lang and syncedlyrics:description.

I followed your proposal, but have a small doubt. This makes specifying the language mandatory when accessing the metadata, or am I wrong? Or should `syncedlyrics` default to a specific language, like `syncedlyrics:eng`? Also, I made `eng` the default language like it already is for comments.

phw: No, it doesn't make specifying the language mandatory. We can still allow just `syncedlyrics` and default to `eng` for that.

The only thing it does is removing the ambiguity of `comment:xxx`. Is this a comment with language `abc` or a comment with description `abc`? In the proposal you have `comment:abc` (the language, with empty description) and `comment::abc` (the description, with default language).

twodoorcoupe:
Thank you all for the reviews!

I've made the tag preserve the language, which should be easy to extend to the `lyrics` tag. Also made it preserve the timestamps for syllables.

>  Does anyone have some experience with synced lyrics used in Vorbis? Would we just store the LRC format there?

When I was trying to put together a plugin to download synced lyrics for [this request on the forum](https://community.metabrainz.org/t/porting-to-picard-webservice/683480), I found out that most players just store the classic LRC format there. But I had a doubt about the mapping, because most players I've come across use `LYRICS` for lyrics with timestamp and `UNSYNCEDLYRICS` for normal lyrics. Picard has `LYRICS` already mapped for normal lyrics, so I was wondering what would be best for synced lyrics.

Sophist-UK (changes requested):
@twodoorcoupe Please keep going - we are getting there.

picard/util/tags.py
```diff
@@ -196,3 +197,23 @@ def parse_comment_tag(name):  # noqa: E302
         desc = ''
 
     return lang, desc
+
+
+def parse_lyrics_tag(name):
```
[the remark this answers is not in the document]
phw: Generally agree on this, with one limitation: This PR should only add and use the `parse_subtag` function, but leave the current `parse_comment_tag` alone. If we do the small syntax changes I proposed this will be incompatible with the current syntax, and we need to consider how to handle the update (and only do that for Picard 3 anyway). So this should all go into a separate PR

rdswift:
When this is all sorted out and merged, will we need to update the [Tag Mapping](https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html) information in the documentation?  If so, the updates should be entered in a PR updating https://github.com/metabrainz/picard-docs/blob/master/tag_mapping.py in the [metabrainz/picard-docs](https://github.com/metabrainz/picard-docs) project.

picard/formats/id3.py
```diff
@@ -684,6 +727,31 @@ def format_specific_metadata(self, metadata, tag, settings=None):
...
 
+    def _parse_sylt_text(self, text):
+        lrc_lyrics = []
+        previous_line_ended = True
+        for lyrics, milliseconds in text:
+            minutes = milliseconds // (60 * 1000)
+            seconds = (milliseconds % (60 * 1000)) // 1000
+            hundredths = (milliseconds % 1000) // 10
+            if previous_line_ended:
+                lrc_lyrics.append(f"[{minutes:02d}:{seconds:02d}.{hundredths:02d}]{lyrics}")
```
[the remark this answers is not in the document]
twodoorcoupe: Thank you for the thorough explanation!

I tried estimating the line-timestamp like you suggested by looking ahead. But I'm not sure how the estimation should be done. As of now I just take the time in the middle of two SYLT frames, but that does not make much sense, since all line-timestamps within a SYLT frame will show the same time.

Maybe it's best to evenly spread the difference between two SYLT frames among all line-timestamps. For example "<00:00.000>Ahhhhhhhhhhhhhhhhhhhh-\nhhh-\nhhh-\nhhhh<00:40.000>men" becomes:
```
[00:00.000]<00:00.000>Ahhhhhhhhhhhhhhhhhhhh-
[00:10.000]hhh-
[00:20.000]hhh-
[00:30.000]hhh-
[00:40.000]<00:40.000>men
```
Then there's also the issue of the last SYLT frame. Since there is no successive frame, it's hard to estimate the timestamps. As of now the code just forgoes the estimation for the last SYLT frame and copies the timestamp. As an example, "<00:10.000>Ahhhhhhhhhhhhhhhhhhhh-\nhhh-\nhhh-\nhhhh" becomes:
```
[00:10.000]<00:10.000>Ahhhhhhhhhhhhhhhhhhhh-
[00:10.000]hhh-
[00:10.000]hhh-
[00:10.000]hhh-
```
Hopefully I was clear enough!

picard/formats/id3.py
```diff
@@ -684,6 +727,31 @@ def format_specific_metadata(self, metadata, tag, settings=None):
...
+                previous_line_ended = False
+            else:
+                lrc_lyrics.append(f"<{minutes:02d}:{seconds:02d}.{hundredths:02d}>{lyrics}")
+            if '\n' in lyrics:
+                previous_line_ended = True
+        return "".join(lrc_lyrics)
+
+    def _parse_lrc_text(self, text):
+        sylt_lyrics = []
+        timestamp_and_lyrics = batched(self.__lrc_syllable_re_parse.split(text)[1:], 2)
```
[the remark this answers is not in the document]
twodoorcoupe: I followed your approach, but I have a question about this:
>4. If A2 Enhanced LRC, then if there is a phrase-timestamp at the very end of the line i.e. followed by nothing or white space, then remove that end-of-line phrase-timestamp.

When working with standard LRC files, I noticed that sometimes there would be line-timestamps with no text after. I think this is done to "clear the screen" when following along with the music. Maybe a phrase-timestamp and the end of a line could be used for the same purpose.

 Also, there could be some phrase-timestamps with no text in the middle of a line, like:
 ```
 <00:10.000>example<00:20.000><00:30.000>example
 ```
I think the expected behaviour would be to create an empty SYLT frame for those, or am I misinterpreting it?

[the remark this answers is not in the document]
Sophist-UK: I think you may be right - and I think you are asking whether there is a difference between:
```
[00:10.000] some stuff
[00:15.000]
[00:20.000] some stuff
```
where as you suggest I agree that we need to preserve the empty time stamp, and
```
[00:10.000] <00:10.000> some stuff <00:15.000>
[00:20.000] <00:20.000> some stuff
```
which is effectively the same and
```
[00:10.000] <00:10.000> some stuff <00:20.000>
[00:20.000] <00:20.000> some stuff
```
where (because the timestamps are the same) is a null subframe that should not be saved.

So when for A2 Enhanced versions you remove the line-timestamps, then if you get the exact same timestamp repeating (hopefully only ever getting this with nothing in between) then you skip it.

picard/formats/id3.py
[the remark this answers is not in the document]
Sophist-UK: I think the first example is better than the second, but even better:
```
[00:00.000]<00:00.000>Ahhhhhhhhhhhhhhhhhhhh- (23 chars in \n)
[00:xx.000]hhh- (5 chars inc \n)
[00:xx.000]hhh- (5 chars inc \n)
[00:xx.000]hhh- (4 chars)
[00:40.000]<00:40.000>men

So there are 37 chars spread across 40secs, so you can estimate this as 40/37 s/char and so 23*40/37 = 25s for the first line, then 5*40/37 = 5s for each of the remaining lines resulting in:
[00:00.000]<00:00.000>Ahhhhhhhhhhhhhhhhhhhh-
[00:25.000]hhh-
[00:30.000]hhh-
[00:35.000]hhh-
[00:40.000]<00:40.000>men 
```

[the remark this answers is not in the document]
Sophist-UK: Where the multi-line frame is the last frame and we have no end time on the lyrics, you can use the track time to estimate the length of the subframe. So in your example, if the track is 40s long you would get the same spread of times.

zas (approved):
LGTM

twodoorcoupe:
I made it so the tag is not written to other formats for now like @phw suggested.

picard/formats/id3.py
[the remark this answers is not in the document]
twodoorcoupe: Thank you for the idea, this makes much more sense now.

phw (dismissed):
@twodoorcoupe  This looks really good. I'm happy you tackled this issue. Let's merge this.

Would you be willing to update picard-docs as well, as @rdswift has outlined above?

picard/formats/id3.py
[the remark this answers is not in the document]
phw: @twodoorcoupe, @Sophist-UK  Is there anything open regarding the above?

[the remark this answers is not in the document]
twodoorcoupe: No, I made it so that the line timestamps are removed if there are also syllable timestamps in the text. Then SYLT frames with no lyrics and with a timestamp equal to the successive are removed.

twodoorcoupe:
>Would you be willing to update picard-docs as well, as @rdswift has outlined above?

Will do right away.

zas (approved):

Sophist-UK:
Do you want my review or not?

If you ask for it you might at least wait for me to do it before approving it.

phw:
@Sophist-UK Yes please, your review would be very welcomed here.

> If you ask for it you might at least wait for me to do it before approving it.

Approval is after reviewing. I think this pull requests really benefits from multiple reviews. We don't need a competition who approves first. Currently this is still showing open review from you, so I would wait with merging until this is done. But if you don't want to review again or don't have time for it, just let us know.

picard/formats/id3.py
```diff
@@ -580,6 +619,12 @@ def _remove_deleted_tags(self, metadata, tags):
                     for key, frame in list(tags.items()):
                         if frame.FrameID == 'USLT' and frame.desc == desc:
                             del tags[key]
+                elif name == 'syncedlyrics' or name.startswith('syncedlyrics:'):
+                    (lang, desc) = parse_subtag(name)
+                    for key, frame in list(tags.items()):
+                        if frame.FrameID == 'SYLT' and frame.desc == desc and frame.lang == lang \
+                                and frame.type == 1 and frame.format == 2:
```
Sophist-UK, line 626: This should probably exclude `frame.format == 2`.

picard/formats/id3.py
```diff
@@ -330,6 +350,18 @@ def _load(self, filename):
                 if frame.desc:
                     name += ':%s' % frame.desc
                 metadata.add(name, frame.text)
+            elif frameid == 'SYLT' and frame.type == 1:
+                if frame.format == 0:
+                    raise NotImplementedError("SYLT format 0 is not supported")
```
Sophist-UK, line 355: What is the user experience when they load a file with format == 0? Is an exception the right response or would a debug log warning message be more appropriate?

What if the format is something other than 0 or 2?

Sophist-UK (changes requested):
Overall excellent. Only a couple of quite minor points...

picard/formats/id3.py
twodoorcoupe: Thanks, I forgot about that. Now a warning is logged and the tag is not added to the metadata when the format is different than 2, which is "using milliseconds as unit".

zas (dismissed):

picard/formats/id3.py
```diff
@@ -351,8 +351,9 @@ def _load(self, filename):
                     name += ':%s' % frame.desc
                 metadata.add(name, frame.text)
             elif frameid == 'SYLT' and frame.type == 1:
-                if frame.format == 0:
-                    raise NotImplementedError("SYLT format 0 is not supported")
+                if frame.format != 2:
+                    log.warning("SYLT formats other than 2 are not supported")
```
Sophist-UK, line 355: Do we want to give any more detail to the user e.g. about which file has SYLT tags with invalid format?

Sophist-UK (changes requested):
Sorry.....

Sophist-UK (approved):
LGTM.

zas (approved):
LGTM

phw (approved):
Good catch about the exception when loading the file.

Thanks everyone for participating. I'll merge this now.

PR #2379, opened 2024-03-28
Simplify config upgrade code

zas:
- get rid of hook args, all are using `config`
- simplify `config._upgrade_hooks` making it a simple dict
- automatically register hooks based on their name, using suffix as version

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action


picard/config_upgrade.py
```diff
@@ -529,33 +551,18 @@ def rename_option(config, old_opt, new_opt, option_type, default):
...
+        """Check if passed function is an upgrade hook"""
+        return (
+            isfunction(f)
+            and f.__module__ == __name__
+            and f.__name__.startswith(UPGRADE_FUNCTION_PREFIX)
+        )
+
+    # Build a dict with version as key and function as value
+    config.run_upgrade_hooks({
+        Version.from_string(name[len(UPGRADE_FUNCTION_PREFIX):]): hook
```
phw, line 566: Can this fail and what happens if it does?I think one easy to make mistake with the new auto detection would be to use the proper prefix but not getting the version string right. As a developer I should notice this quickly.

Could we have a test that checks if all upgrade hooks that get detected have a proper version in the name?

phw (commented):
Great simplification. I only think we should make the code that detects the upgrade hooks itself more testsable.

picard/config_upgrade.py
zas: If it fails to convert version it should raise a VersionError and then a ConfigUpgradeError from that.

But  `Version.from_string()` can have unexpected behaviors:

```
  File "/home/zas/src/picard/tagger.py", line 17, in <module>
    main(os.path.join(basedir, 'locale'), True)
  File "/home/zas/src/picard/./picard/tagger.py", line 1567, in main
    tagger = Tagger(picard_args, localedir, autoupdate, pipe_handler=pipe_handler)
  File "/home/zas/src/picard/./picard/tagger.py", line 337, in __init__
    upgrade_config(config)
  File "/home/zas/src/picard/./picard/config_upgrade.py", line 572, in upgrade_config
    config.run_upgrade_hooks({
  File "/home/zas/src/picard/./picard/config.py", line 305, in run_upgrade_hooks
    raise ConfigUpgradeError(
picard.config.ConfigUpgradeError: Upgrade hook upgrade_to_v1_3_0xxx4 has version 4.0.0.final0 > Picard version 3.0.0.dev3
```

That's because the regex in https://github.com/metabrainz/picard/blob/1329c06bde5fe7ec5a0a67e0fe2438702cae39da/picard/version.py#L32 still matches the ending 4, that's by design.

We could have a strict mode for it, that looks for a full version string (not matching only the end, but the whole string).

What do you think?

picard/config_upgrade.py
```diff
@@ -529,33 +551,18 @@ def rename_option(config, old_opt, new_opt, option_type, default):
...
-    cfg.register_upgrade_hook(upgrade_to_v2_7_0_dev_4)
-    cfg.register_upgrade_hook(upgrade_to_v2_7_0_dev_5)
-    cfg.register_upgrade_hook(upgrade_to_v2_8_0_dev_2)
-    cfg.register_upgrade_hook(upgrade_to_v2_9_0_alpha_2)
-    cfg.register_upgrade_hook(upgrade_to_v3_0_0_dev_1)
-    cfg.register_upgrade_hook(upgrade_to_v3_0_0_dev_2)
-    cfg.run_upgrade_hooks(log.debug)
+    """Detect upgrade hooks methods, register and execute them"""
+
+    def is_upgrade_hook(f):
```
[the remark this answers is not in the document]
zas: Yes, I'll write tests for this stuff.

picard/config_upgrade.py
zas: Also the loop that executes all hooks also does few checks against versions; not bullet-proof but it catches few extra issues (like an upgrade hook version > Picard version).

Since writing upgrade hooks is mainly a dev thing, I don't want to add to much code to handle dev stupidity ;)

rdswift: > Since writing upgrade hooks is mainly a dev thing, I don't want to add to much code to handle dev stupidity ;)

Even mine?  :grinning:

zas: After some thoughts I think we can (and should) make `Version.from_string()` a bit stricter, matching the full string.
See https://github.com/metabrainz/picard/pull/2380

zas: Matching only end of the string also leads to this tricky situation: https://github.com/metabrainz/picard/pull/2379/commits/7e0b89ccd5cee6de2daa1890d88f30d95d2d9ba2

Basically that's possible to have more than one method matching the same version, with #2380 applied it would raise an error instead of passing this test, which does more sense.

picard/config_upgrade.py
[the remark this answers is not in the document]
zas: Done in https://github.com/metabrainz/picard/pull/2379/commits/9852b8b1ccf1bbfc68396adaff540498d94577f8

picard/config_upgrade.py
zas: Another weird case is handled in https://github.com/metabrainz/picard/pull/2379/commits/e87333eb034cadb3b3f5e587e6d947cda0c8e406

Nothing much to do against that, but the code will now raises an exception if it happens.

rdswift (approved):
This looks like a great improvement.  I like it.

rdswift (approved):
The changes look good to me.

phw (approved):
This looks good. LGTM

This will also fix PICARD-2849

PR #2380, opened 2024-03-30
Make `Version.from_string()` stricter, matching from start of string

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Old code leads to unexpected results with strings like `1.1.1xxx2`, as it matches only from end of string, it would consider this as a valid version, equivalent to ... `2.0.0final0`.

AFAIK, this "feature" isn't used anywhere anyway.

# Solution

Be stricter, and match the full string.

With this change, `1.1.1xxx2` will raise a `VersionError`.
Tests were updated accordingly.

# Action

### DIFF
--- a/picard/version.py
+++ b/picard/version.py
@@ -29,7 +29,7 @@ class VersionError(Exception):
 
 
 class Version(namedtuple('VersionBase', 'major minor patch identifier revision')):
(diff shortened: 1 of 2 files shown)


phw (approved):
This was an odd feature. Not sure why this existed at all. LGTM

rdswift (approved):
Two thumbs up from me.  :+1:  :+1:

PR #2382, opened 2024-04-01
Move `FILE_INFO_TAGS` from class `File` to module `picard.util.tags`

zas:
It reduces the size of objects, and it doesn't seem we ever needed to change it.

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action


phw (approved):
Makes sense. LGTM

PR #2384, opened 2024-04-02
Option class improvement

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -30,7 +30,6 @@
 
 
 from collections import defaultdict
-import inspect
 import os
 import shutil
 
@@ -375,17 +374,7 @@ class Option(QtCore.QObject):
     def __init__(self, section, name, default):
         key = (section, name)
         if key in self.registry:
-            stack = inspect.stack()
-            fmt = "Option %s/%s already declared"
-            args = [section, name]
-            if len(stack) > 1:
-                f = stack[1]
-                fmt += "\nat %s:%d: in %s"
-                args.extend((f.filename, f.lineno, f.function))
-                if f.code_context:
-                    fmt += "\n%s"
-                    args.append("\n".join(f.code_context).rstrip())
-            log.error(fmt, *args)
+            raise OptionError("Already declared", section, name)
         super().__init__()
         self.section = section
         self.name = name
--- a/picard/config.py
+++ b/picard/config.py
@@ -360,6 +360,11 @@ class Config(QtCore.QSettings):
         return self._save_backup(backup_path)
 
 
+class OptionError(Exception):
+    def __init__(self, message, section, name):
+        super().__init__("Option %s/%s: %s" % (section, name, message))
+
+
 class Option(QtCore.QObject):
 
     """Generic option."""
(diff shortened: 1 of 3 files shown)


picard/config.py
```diff
@@ -370,17 +374,7 @@ class Option(QtCore.QObject):
...
-            fmt = "Option %s/%s already declared"
-            args = [section, name]
-            if len(stack) > 1:
-                f = stack[1]
-                fmt += "\nat %s:%d: in %s"
-                args.extend((f.filename, f.lineno, f.function))
-                if f.code_context:
-                    fmt += "\n%s"
-                    args.append("\n".join(f.code_context).rstrip())
-            log.error(fmt, *args)
```
zas, line 383: Raising an exception will actually cause a stack trace to be dumped. This should not happen anyway.

zas: Why would a plugin redefine an option?

rdswift: > Why would a plugin redefine an option?

Perhaps two different plugins each define a new option, and coincidentally use the same key?

zas: Then in this case (2 modules defining same option) the conflict might cause issues anyway, not raising an exception will actually hide it.

Module options should have their own namespace anyway, this is something we need to address in pluginsV3.

rdswift: > In general I would agree that raising an exception here is better. But we should still make sure plugins cannot crash Picard just be redefining an option. Can we prevent this somehow?

Just brainstorming here, but can the exception be caught while loading the plugin and abort loading / enabling the plugin at that time (with an appropriate error message in the log)?

zas: I think exceptions from plugins are non-fatal to Picard, due to https://github.com/metabrainz/picard/blob/c16604153192f6140d4df9327ee2cffd25d4f67b/picard/pluginmanager.py#L269-L272

rdswift:
Another dumb question, but does this impact the use (or need) for the `add_if_missing()` method in the `Option()` class?

picard/config.py
rdswift: As long as the option redefinition is executed and caught during the `_load_plugin()` method.

zas:
> Another dumb question, but does this impact the use (or need) for the `add_if_missing()` method in the `Option()` class?

Well, `add_if_missing()` could try and handle the exception instead of using `Option.exists()` but to me it doesn't have to change.

picard/config.py
phw: If this does not lead to crashes due to (faulty) plugins I'm fine with merging. We should be strict with this in general.

> As long as the option redefinition is executed and caught during the `_load_plugin()` method.

Usually it should be, as options should be declared as definition. If not and they are triggered at runtime during some arbitrary code this should be considered a bug (and plugins can crash Picard in other ways as well).

As I remember we had this double definition happening with the option profiles, and rdswift added the add_if_missing to handle the case cleanly. But if the double definition happens as part of Picard's own code it is ok if it crashes on start and this should be fixed immediatelly.

phw (approved):

PR #2385, opened 2024-04-02
PICARD-2850: Fix evaluation of %_filename in examples

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2850

PICARD-2850 touches two issues:

- The variable `%_filename%` is not set in  the default file naming examples, hence using it in a file naming scripts gives bad previews
- If the basename of the filenaming script evaluates to an empty string the filename turns into `_ext`. A variation of this was previously fixed with PICARD-2219, but only for the case where the entire filenaming script is empty.

# Solution

- Extend the examples with more variables and tags. Tags where taken from actual data as retrieved for the used recording / release IDs
- If the basename of the filenaming script evaluates to an empty string, use the basename of the original filename. This generalizes the fix from PICARD-2219


zas (approved):
LGTM

PR #2386, opened 2024-04-02
Move cluster & file comparison weights at module level

zas:
- it reduces the size of Cluster and File objects
- those are constant

It doesn't seem any plugin uses this.


phw (approved):
LGTM

PR #2389, opened 2024-04-04
Collections cleanup

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/collection.py
+++ b/picard/collection.py
@@ -69,35 +69,48 @@ class Collection(QtCore.QObject):
 
     def _finished(self, kind, ids, callback, document, reply, error):
         self.pending -= ids
-        statusbar = self.tagger.window.set_statusbar_message
         if not error:
-            count = len(ids)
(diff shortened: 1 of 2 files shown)


picard/collection.py
```diff
@@ -36,77 +36,87 @@
 
 
 class Collection(QtCore.QObject):
-    COLLECTION_ADD = 1
-    COLLECTION_REMOVE = 2
 
-    def __init__(self, collection_id, name, size):
+    def __init__(self, collection_id):
```
phw, line 40: What do you think of passing the MBAPIHelper here instead of accessing the global one on the tagger object? That way we could make the class fully testable (with a mock API helper).

phw (approved):

picard/collection.py
phw: Will do

PR #2391, opened 2024-04-05
Add a way to select extra debug output

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Web service replies can be very long, and it makes debug output rather hard to read.

# Solution

A debug option mechanism is now added, it is controlled by a new class
`DebugOpt`.

A command-line option `--debug-opts` allows selection of extra debug
output, for now the sole option is `ws_replies`.

A matching menu was added in View Debug Log dialog.

It is important to note those aren't saved in config file, as they are
meant to be disabled by default and only enabled when needed.

# Action

### DIFF
--- a/picard/debug_opts.py
+++ b/picard/debug_opts.py
@@ -1,0 +1,80 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2024 Laurent Monin
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from enum import Enum
+
+
+class DebugOptEnum(int, Enum):
+    __registry__ = set()
+
+    def __new__(cls, value: int, title: str, description: str) -> None:
+        value = int(value)
+        obj = super().__new__(cls, value)
+        obj._value_ = value
+        obj.title = title
+        obj.description = description
+        return obj
+
+    @property
+    def optname(self):
+        return self.name.lower()
+
+    @property
+    def enabled(self):
+        return self in self.__registry__
+
+    @enabled.setter
+    def enabled(self, enable: bool):
+        if enable:
+            self.__registry__.add(self)
+        else:
+            self.__registry__.discard(self)
+
+    @classmethod
+    def opt_names(cls):
+        """Returns a comma-separated list of all possible debug options"""
+        return ','.join(sorted(o.optname for o in cls))
+
+    @classmethod
+    def from_string(cls, string: str):
+        """Parse command line argument, a string with comma-separated values,
+        and enable corresponding debug options"""
+        opts = {str(o).strip().lower() for o in string.split(',')}
+        for o in cls:
+            o.enabled = o.optname in opts
+
+    @classmethod
+    def to_string(cls):
+        """Returns a comma-separated list of all enabled debug options"""
+        return ','.join(sorted(o.optname for o in cls.__registry__))
+
+    @classmethod
+    def set_registry(cls, registry: set):
+        """Defines a new set to store enabled debug options"""
+        cls.__registry__ = registry
+
+    @classmethod
+    def get_registry(cls):
+        """Returns current storage for enabled debug options"""
+        return cls.__registry__
+
+
+class DebugOpt(DebugOptEnum):
+    WS_REPLIES = 1, N_('Web Service Replies'), N_('Log content of web service replies')
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -101,6 +101,7 @@ from picard.const.sys import (
     IS_WIN,
 )
 from picard.dataobj import DataObject
+from picard.debug_opts import DebugOpt
 from picard.disc import (
     Disc,
     dbpoweramplog,
@@ -258,6 +259,9 @@ class Tagger(QtWidgets.QApplication):
 
         if picard_args.audit:
             setup_audit(picard_args.audit)
+
+        if picard_args.debug_opts:
+            DebugOpt.from_string(picard_args.debug_opts)
 
         # Main thread pool used for most background tasks
         self.thread_pool = QtCore.QThreadPool(self)
@@ -1464,6 +1468,9 @@ If a new instance will not be spawned files/directories will be passed to the ex
                         help="do not load any plugins")
     parser.add_argument('--no-crash-dialog', action='store_true',
                         help="disable the crash dialog")
+    parser.add_argument('--debug-opts', action='store',
+                        default=None,
+                        help="comma-separated list of debug options to enable: %s" % DebugOpt.opt_names())
     parser.add_argument('-s', '--stand-alone-instance', action='store_true',
                         help="force Picard to create a new, stand-alone instance")
     parser.add_argument('-v', '--version', action='store_true',
--- a/picard/ui/logview.py
+++ b/picard/ui/logview.py
@@ -43,6 +43,7 @@ from picard.config import (
     IntOption,
     get_config,
 )
+from picard.debug_opts import DebugOpt
 from picard.util import (
     reconnect,
     wildcards_to_regex_pattern,
@@ -159,6 +160,23 @@ class VerbosityMenu(QtWidgets.QMenu):
         self.actions[level].setChecked(True)
 
 
+class DebugOptsMenu(QtWidgets.QMenu):
+
+    def __init__(self, parent=None):
+        super().__init__(parent=parent)
+
+        self.actions = {}
+        for debug_opt in DebugOpt:
+            action = QtGui.QAction(_(debug_opt.title), self, checkable=True, checked=debug_opt.enabled)
+            action.setToolTip(_(debug_opt.description))
+            action.triggered.connect(partial(self.debug_opt_changed, debug_opt))
+            self.addAction(action)
+            self.actions[debug_opt] = action
+
+    def debug_opt_changed(self, debug_opt, checked):
+        debug_opt.enabled = checked
+
+
 class LogView(LogViewCommon):
 
     options = [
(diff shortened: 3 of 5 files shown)


rdswift:
A reminder that when this is merged, it will need to be added to the documentation.

rdswift:
Are there any other log items that fall into this category?  If so, perhaps the option should be `--verbose` (although the `-v/-V` options are used for displaying the version number).

phw:
I agree with the general idea. The WS responses indeed can be very heavy and often make debugging difficult unless you actually need the output.

I'm not so sure if the command line parameter is the best solution, though. Beside adding some very specific switch it also is only possible to trigger from command line. This will mean we will miss important details when users use the log window to report issues.

What about adding a new log level below debug? There are two ways to handle this:

1. We could just define a new level name, which can be set in via command line or UI. But internally we treat it specially, for logging enabling the "debug" level and switch the ` debug_replies`  on the WebService class
2. Or we actually define a custom log level with lower priority then debug (https://docs.python.org/3/howto/logging.html#logging-levels)

Not sure about the naming. Some logging libraries have levels like "pedantic" or "verbose" that sits below debug.

Or we keep it as implemented here, but additionally make this available in the log window UI somehow.

rdswift:
> Not sure about the naming. Some logging libraries have levels like "pedantic" or "verbose" that sits below debug.

Or perhaps the current "debug" level remains as-is, and a new level "verbose" is added above that does not include the service replies?  That leaves "debug" as the setting that provides the most information, and is more appropriate to actually debugging a problem.  I agree that this new level should be available in the log window as well.

zas:
I tried a different approach in last commit.

I'm not for adding another log level, but rather having switchable debug outputs.
In the initial case, it is more about having a usable debug log file for most cases and being able to ask a user to enable more details if needed when we need to narrow down an issue.

I went for something much more generic and expandable.

picard/debug_opts.py
```diff
@@ -0,0 +1,79 @@
...
+    """Returns True if passed debug option is enabled"""
+    return debug_opt in debug_opts
+
+
+def debug_opt_set(debug_opt, state):
+    """Set the boolean state of each debug option, stored in debug_opts"""
+    if state:
+        debug_opts.add(debug_opt)
+    else:
+        debug_opts.remove(debug_opt)
```
rdswift, line 69: I really like this new approach, but under what cirrcumstance do you expect this (remove) to occur?  Or is this included to make the functionality complete to handle future changes when the command line could be used to disable options that are normally set by default?

zas: When we toggle those by UI.

rdswift: Okay, I think I'll have to wait and see the code to fully understand.  Are you saying that omitting an item on the command line will toggle the option off if it is otherwise set (via the UI and a persistent setting)?

rdswift: I just looked at it again, and it all now makes sense to me.  I have no idea what I was thinking about before.  Sorry for the confusion.  (And I still really like this approach.)

zas:
image

picard/debug_opts.py
```diff
@@ -0,0 +1,79 @@
...
+
+class DebugOpt(IntEnum):
+    WS_REPLIES = 1
+
+
+DebugOptDesc = namedtuple('DebugOptDesc', ('name', 'description'))
+
+debug_opts_descriptions = {
+    DebugOpt.WS_REPLIES: DebugOptDesc(name='ws_replies', description=N_('Log web service replies'))
+}
```
rdswift, line 33: Perhaps this could be modified slightly to include 'title' which would appear in the drop down list rather than the name that is used on the command line?  This title could also be translated in the drop down.

```suggestion
DebugOptDesc = namedtuple('DebugOptDesc', ('name', 'title', 'description'))

debug_opts_descriptions = {
    DebugOpt.WS_REPLIES: DebugOptDesc(name='ws_replies', title=N_('WS replies'), description=N_('Log web service replies'))
}
```

Then Line 174 in `logview.py` might be something like:

```
action = QtGui.QAction(_(desc.title), self, checkable=True, checked=debug_opt in debug_opts)
```

rdswift (approved):

picard/debug_opts.py
zas: Good idea, done in [6d15126](https://github.com/metabrainz/picard/pull/2391/commits/6d151260bd133e25b1998b304e2e0f10ca20b96f)

rdswift (dismissed):

picard/ui/logview.py
```diff
@@ -181,10 +199,18 @@ def __init__(self, parent=None):
         self.hbox.addWidget(self.verbosity_menu_button)
 
         self.verbosity_menu = VerbosityMenu()
-        self._set_verbosity(self.verbosity)
         self.verbosity_menu.verbosity_changed.connect(self._verbosity_changed)
         self.verbosity_menu_button.setMenu(self.verbosity_menu)
 
+        self.debug_opts_menu_button = QtWidgets.QPushButton(_("Debug Options"))
+        self.debug_opts_menu_button.setAccessibleName(_("Debug Options"))
```
rdswift, line 206: Would it make sense to create a temporary variable here to avoid having to translate the same string twice?  I ask because I'm still a bit fuzzy on how the i18n stuff works.

zas: It would make sense, but it also hides the source a bit to translators.
In this case, that's definitively not an issue, in a loop or in critical performance code caching would make sense.

picard/webservice/__init__.py
```diff
@@ -530,7 +531,8 @@ def _handle_reply(self, reply, request):
                 elif request.response_parser:
                     try:
                         document = request.response_parser(reply)
-                        log.debug("Response received: %s", document)
+                        if DebugOpt.WS_REPLIES.enabled:
+                            log.debug("Response received: %s", document)
```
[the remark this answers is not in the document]
zas: That's the whole point, it will only log if the matching DebugOpt is set, it can be very useful when we want to log something heavy in code that is critical performance-wise.
Here we avoid to flood the internal log memory (and/or log output) with huge web service answers.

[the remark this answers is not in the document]
rdswift: In that case, would it make sense to not allow it to be enabled in the log viewer if was not previously enabled (so as not to raise false expectations), or am getting way too picky?  :stuck_out_tongue:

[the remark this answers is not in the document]
rdswift: Actually, I am being way too picky.  It makes sense the way you have it done.  It's just that we need to explain in the documentation that the responses won't be saved while the option is disabled, so they won't be available for viewing in the log.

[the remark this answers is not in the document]
zas: > Also that disabling the option in the log viewer won't hide the responses already saved to the log.

This can't be done easily, and that's definitively not my goal here. I wanted to KISS, because that's definitively an end user thing.
It should be noted that we weren't displaying those ws replies before, and the debug log is totally unreadable (and far too huge) when they are in.
It makes sense to have those when narrowing an issue, but most of the time the query is enough (most of the times replies are the same, for non-authenticated queries).

rdswift (approved):
This looks good to me.

zas:
> Another possible change would be to not output the full path to a plugin by default when logging the plugin name, and then add a debug option to enable logging the full path.

Where is it in the code?

rdswift:
> Where is it in the code?

In the `name_filter()` method of the `TailLogger()` class in the `picard/log.py` module starting at Line 159.

https://github.com/metabrainz/picard/blob/0a36d1e0d5494cdc046b407854f0b8b3c0a8716a/picard/log.py#L159-L174

zas:
@rdswift This one is tricky, this code isn't only used for plugins. I would keep it for later.

rdswift:
> @rdswift This one is tricky, this code isn't only used for plugins. I would keep it for later.

Okay.  When the time comes, I was thinking of something like (rough code not optimized):

```
def name_filter(record):
    # In case the module exists within picard, remove the picard prefix
    # else, in case of something like a plugin, keep the path as it is.
    # It provides a significant but short name from the filepath of the module
    path = Path(record.pathname).with_suffix('')
    # PyInstaller paths are already relative
    # FIXME: With Python 3.9 this should better use
    # path.is_relative_to(picard_module_path.parent)
    # to avoid the exception handling.
    if path.is_absolute():
        try:
            path = path.resolve().relative_to(picard_module_path.parent)
        except ValueError:
            if not DebugOpt.FULL_PLUGIN_PATH.enabled:
                parts = tuple(reversed(path.parts))
                if 'plugins' in parts:
                    parts = parts[:parts.index('plugins') + 1]
                path = Path('/'.join(p for p in reversed(parts) if not p.endswith('.zip')))
    record.name = '/'.join(p for p in path.parts if p != '__init__')
    return True
```

zas:
> 
> Okay. When the time comes, I was thinking of something like (rough code not optimized):
> 

Yes, if this PR goes through, we can make another with such changes (btw keep in mind plugin system will go through a complete overall soon enough).

rdswift:
> keep in mind plugin system will go through a complete overall soon enough

Exactly why I wasn't sure about proceeding with this in the first place, but I haven't seen anything so far to suggest that the logging would be changed significantly.

This isn't critical by any streatch of the imagination, but I thought that it might be another potential use of the new functionality that you've introduced.  Besides, it has been one of the things that has always annoyed me about Picard's logging. :wink:

zas:
> Exactly why I wasn't sure about proceeding with this in the first place, but I haven't seen anything so far to suggest that the logging would be changed significantly.

At least for the `.zip` thing.

> Besides, it has been one of the things that has always annoyed me about Picard's logging. 😉

Understandable :)

BTW, I tried, and it works (though current version of Picard doesn't support old plugins, so I tested with hardcoded paths).
I would rewrite the code to:

```python
    if path.is_absolute():
        try:
            path = path.resolve().relative_to(picard_module_path.parent)
        except ValueError:
            if not DebugOpt.PLUGIN_FULLPATH.enabled:
                try:
                    parts = list(reversed(path.parts))
                    parts[0] = parts[0].rstrip('.zip')
                    parts = parts[:parts.index('plugins') + 1]
                    path = Path(*reversed(parts))
                except ValueError:
                    pass
    record.name = str(Path(*(p for p in path.parts if p != '__init__')))
```

Note: the change on last line should be done anyway, I noticed we get '//' in front of absolute paths:

```python
>>> from pathlib import Path
>>> p = Path('/a/b/c')
>>> p.parts
('/', 'a', 'b', 'c')
>>> '/'.join(p.parts)
'//a/b/c'
>>> str(Path(*p.parts))
'/a/b/c
 ```

Also we need tests for this code.
So definitively something to do outside this PR.

**Comment by zas** [comment]
@rdswift FYI, I added tests and modified `name_filter()` in https://github.com/metabrainz/picard/pull/2396

File: picard/tagger.py [hunk 5]
```diff
@@ -1464,6 +1468,9 @@ def process_picard_args():
                         help="do not load any plugins")
     parser.add_argument('--no-crash-dialog', action='store_true',
                         help="disable the crash dialog")
+    parser.add_argument('--debug-opts', action='store',
+                        default=None,
+                        help="Comma-separated list of debug options to enable: %s" % DebugOpt.opt_names())
```
phw, line 1473: That's really minor, but all other options start with lowercase letters. We should be consistent

picard/debug_opts.py
```diff
@@ -0,0 +1,80 @@
...
+        cls.__registry__ = registry
+
+    @classmethod
+    def get_registry(cls):
+        """Returns current storage for enabled debug options"""
+        return cls.__registry__
+
+
+class DebugOpt(DebugOptEnum):
+    WS_REPLIES = 1, N_('WS Replies'), N_('Log web service replies')
```
phw, line 80: I would suggest to actually call the label "Web service replies". Just WS might be unclear and make translations more difficult. I think we have enough space available in the menu.

phw (changes requested):
Looks good. The UI makes sense.

Just minor comments about the UI labels.

picard/debug_opts.py
zas: Done in (rebased) https://github.com/metabrainz/picard/pull/2391/commits/61dbae164b69efe3c63de1446b67106f7079b7b6

picard/tagger.py
zas: Done in (rebased) https://github.com/metabrainz/picard/pull/2391/commits/61dbae164b69efe3c63de1446b67106f7079b7b6

phw (approved):
LGTM

rdswift:
Added PICARD-2860 to update the documentation.

PR #2393, opened 2024-04-06
Configure cache size from environ

bitrate16:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Configure network cache size from env.

# Problem

When tagging large libraries, 100M cache is not enough. After tagging 25K files, cache size was about 8GB. In addition I suspect that after cache overflow picard silently crashes, so you can't just leave library lookup for a night and have tagged songs in the morning.

# Solution

Environmental variable `PICARD_CACHE_SIZE` to configure cache size, defaults to stock `100M`.

# Action

Check if it is a good approach to use env when also using env for cache location

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -34,6 +34,7 @@
 
 import builtins
 from collections import OrderedDict
+import os
 
 from picard import PICARD_VERSION
 from picard.const import appdirs
@@ -49,7 +50,11 @@ USER_DIR = appdirs.config_folder()
 USER_PLUGIN_DIR = appdirs.plugin_folder()
 
 # Network Cache default settings
-CACHE_SIZE_IN_BYTES = 100*1000*1000
+CACHE_SIZE_IN_BYTES = os.environ.get('PICARD_CACHE_SIZE', '').strip()
+if CACHE_SIZE_IN_BYTES.isdigit():
+    CACHE_SIZE_IN_BYTES = int(CACHE_SIZE_IN_BYTES)
+else:
+    CACHE_SIZE_IN_BYTES = 100*1000*1000
 
 # AcoustID client API key
 ACOUSTID_KEY = 'v8pQ6oyB'


Sophist-UK:
Over a decade ago I wrote a [PR](https://github.com/metabrainz/picard/pull/122) to address this. But there was no interest in merging it.

So I am generally in favour of increasing the cache size and making it configurable. However IMO this PR has various issues as follows:

1. There needs to be a default in the event that the environment variable is not set, and as far as I can see by looking at the code it is likely to break Picard if the environment variable is not set. 

2. Until now we have avoided using environment variables for these things. This is more normal in Linux, but far less normal in Windows environments. Instead we use Options.

(As an aside - WOW - my 2013 PR was number 122 and this one is almost 20x as many PRs (and issues) later!!!!)

zas:
Thanks for your contribution.

It would be better to add an option in the UI instead of (or in addition of) an env var.

That said, I'm not sure what you are trying to fix here, as the issue isn't well defined (did you open a ticket about this issue?).

> When tagging large libraries, 100M cache is not enough. After tagging 25K files, cache size was about 8GB.

First, explain about "tagging 25K files", what do you do exactly?
How do you measure a cache size of 8GB, what's the exact use ratio for this cache (https://github.com/metabrainz/picard/blob/58c1177b7d66d630ac9d7ccb4e9051c5e1609eee/picard/webservice/__init__.py#L519 actually logs cached answers)?

This constant controls size of QNetworkDiskCache.
It caches answers from the web service that are cacheable (it depends on headers).

In any case, at worse, Picard will just re-download data.
When tagging a lot of files, it will spare few requests (some are cached for few seconds on server side), only if those are repeated and server doesn't invalidate them.

> I suspect that after cache overflow picard silently crashes

Can you reproduce this issue?
It should be fairly easy to reduce the cache size to simulate that.
Of course providing details about your environment would help.

You make an assumption that network cache size is the cause here, without providing much infos.
Perhaps that's just Picard running out of memory or anything else. Does increasing the network cache size actually solve the issue for you?

Tagging 25K new files at once isn't recommended (https://picard-docs.musicbrainz.org/en/troubleshooting/stopped_working.html), but you don't tell if those were previously tagged or not.

zas:
It can be something like https://github.com/metabrainz/picard/pull/2395

zas:
https://github.com/metabrainz/picard/pull/2395 replaces this

PR #2394, opened 2024-04-07
PICARD-2580: Add keyboard shortcut for Artwork/Show more details

ShubhamBhut:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2580

# Solution

Added a new shortcut which connects to existing `show_cover_art_info()` method

# Action

### DIFF
--- a/picard/ui/coverartbox.py
+++ b/picard/ui/coverartbox.py
@@ -393,6 +393,7 @@ class CoverArtBox(QtWidgets.QGroupBox):
         self.orig_cover_art = CoverArtThumbnail(False, False, self.pixmap_cache, parent)
         self.orig_cover_art_label.setAlignment(QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter)
         self.show_details_button = QtWidgets.QPushButton(_('Show more details'), self)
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM.

We really should have a way to list and configure those shortcuts in Picard, discovering this one isn't easy.

phw (approved):
Thanks a lot, this makes sense to have. LGTM.

We must also not forget to update the documentation at https://picard-docs.musicbrainz.org/en/appendices/keyboard_shortcuts.html

Would you mind adding a PR for this in https://github.com/metabrainz/picard-docs/ ? I think we can add the new shortcut as part of the "Metadata view" section.

phw:
> We really should have a way to list and configure those shortcuts in Picard, discovering this one isn't easy.

I agree. I like the keyboard shortcuts overviews of GTK3 applications. Something like this would be great. Or actually editable shortcuts. That one is trickier, as shortcuts are used all over the place, and there are different types (global shortcuts, shortcuts depending on context in some widget, built-in shortcuts). But it would also help with the discovery. We might already have a ticket for this.

PR #2395, opened 2024-04-09
PICARD-2855: Let the user configure network cache size via Network Options

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Network cache size isn't user-configurable.

* JIRA ticket (_optional_): PICARD-2855
 

# Solution

Add an option defaulting to current fixed value.

# Action


picard/const/__init__.py
```diff
@@ -49,7 +49,8 @@
 USER_PLUGIN_DIR = appdirs.plugin_folder()
 
 # Network Cache default settings
-CACHE_SIZE_IN_BYTES = 100*1000*1000
+CACHE_SIZE_DISPLAY_UNIT = 1000*1000
```
Sophist-UK, line 52: I would personally prefer this  to be 1024x1024.

test/test_webservice.py
```diff
@@ -61,6 +61,7 @@
     "proxy_username": 'user',
     "proxy_password": 'password',
     "network_transfer_timeout_seconds": 30,
+    "network_cache_size_bytes": 100*1000*1000,
```
Sophist-UK, line 64: This would need to be changed too plus similar lines below.

zas: I think we already had this discussion (https://github.com/metabrainz/picard/pull/122/files#r4631209).

MB stands for Megabytes and SI prefixes defines Mega as 10⁶. MiB could be used of course, but I guess end user doesn't much about the difference between Mebibytes and Megabytes anyway.

https://en.wikipedia.org/wiki/Megabyte
https://physics.nist.gov/cuu/Units/binary.html
https://www.bipm.org/en/measurement-units/si-prefixes

My preference is to stick to widely adopted International System of Units and their prefixes.

picard/webservice/__init__.py
```diff
@@ -392,6 +392,9 @@ def _init_timers(self):
     def set_cache(self, cache_size_in_bytes=None):
         if cache_size_in_bytes is None:
             cache_size_in_bytes = CACHE_SIZE_IN_BYTES
+        if cache_size_in_bytes <= 0:
+            log.debug("NetworkDiskCache disabled")
+            return
```
phw, line 397: As I understand it is not necessarily disabled at this point, just keeps the previous setting. So when I change the cache settings and put it to 0 then it takes a restart to actually disable the cache. Would `manager.setCache`  allow to actually unset it?

phw (changes requested):
I personally am not quite convinced the network cache has a big influence on most workflows. If I scan a lot of files and load thousands of releases I get thousands of different requests, not much potential to caching. Also as zas poined out the request delay likely is the limiting factor anyway. Doesn't matter if a request takes a few hundred milliseconds longer if the next request has to wait afterwards anyway.

Regardless, the PR looks good and configuring the cache in general makes sense to give users control here. And if it is just for disabling caching completely or setting it very high to have a good feeling and because they have the disk space :D

Just a question regarding the behavior when setting cache to zero. Otherwise LGTM

picard/webservice/__init__.py
zas: A restart is needed anyway.

rdswift: I agree that this likely won't be used much.  My suggestion may be to add a note saying something like "Requires restart" to the option page where this is set.  The real difficulty will be notifying (reminding) the user that a restart is required if the setting is changed as part of enabling/disabling an option profile.

zas: I left it out on purpose, I think we need something generic for options that requires a restart. We already have such message for certain options.
Having a message listing settings that will only apply after restart would be great, and that's something we can define once per option, so basically when we declare them.

Sophist-UK:
> I personally am not quite convinced the network cache has a big influence on most workflows. If I scan a lot of files and load thousands of releases I get thousands of different requests, not much potential to caching. Also as zas poined out the request delay likely is the limiting factor anyway. Doesn't matter if a request takes a few hundred milliseconds longer if the next request has to wait afterwards anyway.

@phw Philipp - as you say if you load your music library into Picard literally once in a blue moon, the cache is not worth jack sh*t. But, I think there are some uses cases where it might give some small benefits and here are some examples off the top of my head:
* Sometimes I switch back and forth between related releases before settling on one.
* Sometimes I refresh a release because I have screwed up a script or (only very very occasionally because I don't believe in them) screwed up a manual edit.
* If I am developing some Picard functionality I may load the same test releases repeatedly.

BUT, equally there is a BIG BIG problem because I don't run Picard that frequently and the cache does not get cleared when I stop using it for several months. When the cache is 10MB in size, who cares. If it is 10GB is size, then perhaps I would care a lot.

So IMO we really need a "On close, prompt the user whether to clear the network cache or not" function.

zas:
> So IMO we really need a "On close, prompt the user whether to clear the network cache or not" function.

Can you please create a ticket for such feature?

picard/webservice/__init__.py
zas: @phw: what we could do in case of value change is to only call `QNetworkDiskCache.setMaximumCacheSize()` and unconditionally create the cache directory.

 It would avoid any need to restart as [`QNetworkDiskCache.setMaximumCacheSize()`](https://doc.qt.io/qt-6/qnetworkdiskcache.html#setMaximumCacheSize) "Sets the maximum size of the disk cache to be size.

If the new size is smaller then the current cache size then the cache will call [expire](https://doc.qt.io/qt-6/qnetworkdiskcache.html#expire)()."

phw: @zas Yes, that seems reasonable

> I left it out on purpose, I think we need something generic for options that requires a restart.

I totally agree with that.

zas: @phw I implemented changing cache size without restart in [839e02f](https://github.com/metabrainz/picard/pull/2395/commits/839e02f15436a7a275598037964a473efc9b4d1d)

phw (approved):
LGTM

rdswift:
Added PICARD-2861 to update the documentation.

PR #2397, opened 2024-04-12
PICARD-2856: Introduce DebugOpt.PLUGIN_FULLPATH and shorten plugin paths in logs 

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): [PICARD-2856](https://tickets.metabrainz.org/browse/PICARD-2856)

# Solution

# Action

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -45,6 +45,7 @@ from picard.const.sys import (
     FROZEN_TEMP_PATH,
     IS_FROZEN,
 )
+from picard.debug_opts import DebugOpt
 
 
 # Get the absolute path for the picard module
@@ -176,6 +177,12 @@ def name_filter(record):
             path = path.resolve().relative_to(picard_module_path)
         except ValueError:
             pass
+        else:
+            if not DebugOpt.PLUGIN_FULLPATH.enabled and 'plugins' in path.parts:
+                parts = list(reversed(path.parts))
+                parts = parts[:parts.index('plugins') + 1]
+                path = Path(*reversed(parts))
+
     parts = list(path.parts)
     if parts[-1] == '__init__':
         del parts[-1]
--- a/picard/debug_opts.py
+++ b/picard/debug_opts.py
@@ -78,3 +78,4 @@ class DebugOptEnum(int, Enum):
 
 class DebugOpt(DebugOptEnum):
     WS_REPLIES = 1, N_('Web Service Replies'), N_('Log content of web service replies')
+    PLUGIN_FULLPATH = 2, N_('Plugin Fullpath'), N_('Log plugin full paths')
--- a/test/test_log.py
+++ b/test/test_log.py
@@ -33,6 +33,7 @@ from unittest.mock import patch
 from test.picardtestcase import PicardTestCase
 
 from picard.const.sys import IS_WIN
+from picard.debug_opts import DebugOpt
 from picard.log import (
     _calculate_bounds,
     name_filter,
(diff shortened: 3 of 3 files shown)


picard/log.py
```diff
@@ -176,6 +177,12 @@ def name_filter(record):
             path = path.resolve().relative_to(picard_module_path)
         except ValueError:
             pass
+        else:
+            if not DebugOpt.PLUGIN_FULLPATH.enabled and 'plugins' in path.parts:
+                parts = list(reversed(path.parts))
+                parts = parts[:parts.index('plugins') + 1]
+                path = Path(*reversed(parts))
+
```
rdswift, line 185: Perhaps I'm missing something, but won't this only shorten the paths if the path is already relative to the `picard` module (which will not be the case for plugins)?  I'm wondering if Lines 179-180 should be removed to shorten the paths that are not relative to the `picard` module?

zas: Yes, it only shortens file paths relative to picard module, for now.

We can extend it to apply to all paths, but it would be more "risky", imagine someone having picard installed this way:
`/some/path/plugins/picard/module/file.py` it would lead to  `plugins/picard/module`.

Instead we should perhaps use actual plugin paths, instead of simply matching on `plugins` in the path.

rdswift: I really must be missing something because I don't understand the issue.

> Yes, it only shortens file paths relative to picard module, for now.

So it only (potentially) shortens paths that have already been shortened because they are relative to the picard module?  Since the installed plugins are in a directory somewhere under the user's home directory (even when installed manually) -- with the possible exception of the Windows portable version, then it is **extremely** unlikely that they would ever be marked as relative to the picard module and thus would never be shortened.  In that case this new option to display the full path for plugins would be irrelevant because the full path to the plugins would always be displayed by default.

> We can extend it to apply to all paths, but it would be more "risky", imagine someone having picard installed this way: `/some/path/plugins/picard/module/file.py` it would lead to `plugins/picard/module`.

That's why I suggested that the new plugin shortening code would only be appled when the path was **not** relative to the picard module (when the exception was triggered) and the new `PLUGIN_FULLPATH.enabled` was false.

I don't understand your comment about it being more "risky".  I don't believe that this `name_filter()` function is used anywhere other than in the logger.  If your concern is that the displayed path to the plugin might be somehow confusing because it has been shortened, then isn't that why we're adding the option to display the full path?

> Instead we should perhaps use actual plugin paths, instead of simply matching on `plugins` in the path.

Are you saying that we should be displaying the full path to the plugins by default?  In that case, there would be no need to have this new option.

---

Perhaps a more acceptable solution would be to have the function first check if the path is relative to the picard module and only display starting at that directory (as in the original code).  Then if it is not relative to the picard module, check if it is relative to the current plugins directory (`USER_PLUGIN_DIR` from `picard.const`) and then display starting from that directory unless `PLUGIN_FULLPATH.enabled` was true.

Something like (untested and unoptimized code):

```python
from picard.const import USER_PLUGIN_DIR


def name_filter(record):
    # In case the module exists within picard, remove the picard prefix
    # else, in case of something like a plugin, keep the path as it is.
    # It provides a significant but short name from the filepath of the module
    path = Path(record.pathname).with_suffix('')
    # PyInstaller paths are already relative
    # FIXME: With Python 3.9 this should better use
    # path.is_relative_to(picard_module_path.parent)
    # to avoid the exception handling.
    if path.is_absolute():
        try:
            path = path.resolve().relative_to(picard_module_path)
        except ValueError:
            pass

        try:
            path = path.resolve().relative_to(USER_PLUGIN_DIR)
            if not DebugOpt.PLUGIN_FULLPATH.enabled and 'plugins' in path.parts:
                parts = list(reversed(path.parts))
                parts = parts[:parts.index('plugins') + 1]
                path = Path(*reversed(parts))
        except ValueError:
            pass

    parts = list(path.parts)
    if parts[-1] == '__init__':
        del parts[-1]
    if parts[0] == path.anchor:
        del parts[0]
    record.name = str(PurePosixPath(*parts))
    return True
```

rdswift: Note that I cleaned up the code a bit in my previous comment so that the logic makes a bit more sense.

zas: @rdswift I agree, I'll merge this, then you can submit a PR implementing it

rdswift (approved):

rdswift:
Added PICARD-2860 to update the documentation.

PR #2399, opened 2024-04-12
Improve changes applied on restart dialog

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

- display all changes as a list (it fixes an issue where if you changed both theme and language only warning about theme was displayed)
- reformat using richtext
- use an actual warning dialog
- modifies the API so we can pass multiples warnings and notes
- make title and last sentence fixed
- since we detect changes in options not need to write to them if unchanged

Capture d’écran du 2024-04-12 14-53-24

# Action

### DIFF
--- a/picard/ui/options/interface.py
+++ b/picard/ui/options/interface.py
@@ -56,6 +56,7 @@ from picard.ui.theme import (
     UiTheme,
 )
 from picard.ui.ui_options_interface import Ui_InterfaceOptionsPage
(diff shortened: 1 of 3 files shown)


phw (approved):
This is pretty cool. Much easier to extend for future settings.

PR #2400, opened 2024-04-12
Move titles from profiles to Option declarations

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action


rdswift (dismissed):
I like this.  Having the title defined at the same time as the setting makes so much sense.  Just need to fix the failing test.

zas:
There's an issue with Option `selected_file_naming_script_id`, it is only declared in upgrade hooks, not sure why, any idea?

zas:
@rdswift I think we can somehow get rid of `UserProfileGroups`:

- add an optional property `profile_group` to Option (for example `interface`)
- add an optional property `profile_highlights` to Option (a list, for example `['ui_language', 'label']`)
- have a dict for group's titles (for example, `{'interface': N_("User Interface"), ...}`

And use that instead.

Main advantage: reduce duplication, easier to manage
What do you think?

rdswift:
> @rdswift I think we can somehow get rid of `UserProfileGroups`:
> 
> Main advantage: reduce duplication, easier to manage What do you think?

I think that's a great idea.  It indeed makes it a lot easier to manage.

zas:
> > @rdswift I think we can somehow get rid of `UserProfileGroups`:
> > Main advantage: reduce duplication, easier to manage What do you think?
> 
> I think that's a great idea. It indeed makes it a lot easier to manage.

I wonder about ordering though, currently SettingDesc are sorted in an OrderedDict, but the result is that translated titles appear in random order (that's options order) and that's a bit weird. I guess we can just sort them alphabetically.

Also we could just group those by OptionsPage, using their titles (https://github.com/metabrainz/picard/blob/93623cb8d34c563bfe09889ed88aba5a7b222c0c/picard/ui/options/general.py#L56), this way we don't even need `profile_group`.

The highlights thing is rather annoying, because it needs to be updated each time we rename an element which doesn't even appear in the source file. Is there a way to drop that without too much impacting the user experience?

If we can get rid of both, then we can just mark options to **exclude** from profiles. It would make things much simpler to maintain.

Whatever, I think we can merge this PR first, as it doesn't change behavior at all.

rdswift:
> I wonder about ordering though, currently SettingDesc are sorted in an OrderedDict, but the result is that translated titles appear in random order (that's options order) and that's a bit weird. I guess we can just sort them alphabetically.

I had them sorted so they displayed in the same order that they appeared on the options page, thus keeping related option settings together.  I suppose that isn't strictly necessary, so sorting alphabetically could work.

> Also we could just group those by OptionsPage, using their titles (
> 
> https://github.com/metabrainz/picard/blob/93623cb8d34c563bfe09889ed88aba5a7b222c0c/picard/ui/options/general.py#L56
> 
> ), this way we don't even need `profile_group`.

That should be fine.  Again, I had the groups sorted in the same order that they appear in the option page tree for consistency.  I suppose that isn't really necessary.

> The highlights thing is rather annoying, because it needs to be updated each time we rename an element which doesn't even appear in the source file. Is there a way to drop that without too much impacting the user experience?

Personally, I think the highlighting on the option pages is important, and would hate to see it disappear.

For the most part, the highlight items are actually the fields updated in the `load()` method and read in the `save()` method of the option page, so they generally appear in the same file where the options are defined.  The exceptions are infrequent, and are typically when a group box can be checked to enable/disable other option settings contained in the group box.  I've actually considered revising those to simplify and allow proper highlighting of the individual options, but it hasn't floated to the top of my "to do" list yet.  :wink: 

> If we can get rid of both, then we can just mark options to **exclude** from profiles. It would make things much simpler to maintain.

I like that idea of being able to mark an option to include/exclude from profiles.  You're right that would make maintenance so much simpler.  I'll think about it some more and see if I can come up with a way to do that but still keep the ability to highlight the appropriate settings on the option pages.

> Whatever, I think we can merge this PR first, as it doesn't change behavior at all.

Agreed.

rdswift (approved):

rdswift:
> If we can get rid of both, then we can just mark options to **exclude** from profiles. It would make things much simpler to maintain.

Actually, it just occurred to me that if an option doesn't have a title assigned then it can be considered as excluded from profiles.  No need for a separate attribute.

PR #2401, opened 2024-04-13
Add Persist & Setting classes, subclassing Option

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -413,36 +413,128 @@ class Option(QtCore.QObject):
         return type(self.default)(value)
 
 
-class TextOption(Option):
+class OptSect(Option):
+    section = None
 
+    def __init__(self, name, default, title=None):
+        if self.section is None:
+            raise OptionError("No section defined")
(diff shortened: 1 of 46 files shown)


zas:
Not sure about this, so that's mainly to open a discussion.

Having separated classes for persist & setting makes sense to me, especially after the changes to profiles.

- we can have different methods and properties (for example, title only makes sense for settings atm)
- it shortens declaration and reduces redundancy.

Basically `Persist` & `Setting` are instances of `Option`, but we don't have to repeat the section.
I had to introduce an intermediate class `OptSect` for override methods (it also checks section is defined).

The way I did it preserve existing classes and associated methods, so it doesn't break compatibility with plugins or existing Picard code.

zas:
This move has too many side-effects, I prefer to not do it now.

PR #2402, opened 2024-04-13
Update plugin path shortening logic

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Further to the discussion in https://github.com/metabrainz/picard/pull/2397 this PR modifies the way that plugin paths are displayed in the logs.

# Solution

Shorten the path by making it relative to the path stored in the `picard.const.USER_PLUGIN_DIR` variable if possible.

# Action

None.

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -184,6 +184,7 @@ def name_filter(record):
         try:
             path = path.resolve().relative_to(USER_PLUGIN_DIR)
             parts = list(p for p in path.parts if not p.endswith('.zip'))
+            parts.insert(0, 'plugins')
             path = Path(*parts)
         except ValueError:
             pass
--- a/picard/log.py
+++ b/picard/log.py
@@ -10,6 +10,7 @@
 # Copyright (C) 2017 Sophist-UK
 # Copyright (C) 2018 Wieland Hoffmann
 # Copyright (C) 2021 Gabriel Ferreira
+# Copyright (C) 2024 Bob Swift
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -41,6 +42,7 @@ from threading import Lock
 
 from PyQt6 import QtCore
 
+from picard.const import USER_PLUGIN_DIR
 from picard.const.sys import (
     FROZEN_TEMP_PATH,
     IS_FROZEN,
@@ -177,11 +179,15 @@ def name_filter(record):
             path = path.resolve().relative_to(picard_module_path)
         except ValueError:
             pass
-        else:
-            if not DebugOpt.PLUGIN_FULLPATH.enabled and 'plugins' in path.parts:
-                parts = list(reversed(path.parts))
-                parts = parts[:parts.index('plugins') + 1]
-                path = Path(*reversed(parts))
+
+    if path.is_absolute() and not DebugOpt.PLUGIN_FULLPATH.enabled:
+        try:
+            path = path.resolve().relative_to(USER_PLUGIN_DIR)
+            parts = list(p for p in path.parts if not p.endswith('.zip'))
+            parts.insert(0, 'plugins')
+            path = Path(*parts)
+        except ValueError:
+            pass
 
     parts = list(path.parts)
     if parts[-1] == '__init__':
--- a/picard/log.py
+++ b/picard/log.py
@@ -183,7 +183,7 @@ def name_filter(record):
     if path.is_absolute() and not DebugOpt.PLUGIN_FULLPATH.enabled:
         try:
             path = path.resolve().relative_to(USER_PLUGIN_DIR)
-            parts = list(p for p in path.parts if not p.endswith('.zip'))
+            parts = list(path.parts)
             parts.insert(0, 'plugins')
             path = Path(*parts)
         except ValueError:
@@ -194,6 +194,11 @@ def name_filter(record):
         del parts[-1]
     if parts[0] == path.anchor:
         parts[0] = '/'
+    # Remove the plugin module file if the file name is the same as
+    # the immediately preceeding plugin zip file name, similar to the
+    # way that the final `__init__.py` file is removed.
+    if len(parts) > 1 and parts[-1] + '.zip' == parts[-2]:
+        del parts[-1]
     record.name = str(PurePosixPath(*parts))
     return True
 
--- a/picard/log.py
+++ b/picard/log.py
@@ -184,7 +184,6 @@ def name_filter(record):
         try:
             path = path.resolve().relative_to(USER_PLUGIN_DIR)
             parts = list(p for p in path.parts if not p.endswith('.zip'))
-            parts.insert(0, 'plugins')
             path = Path(*parts)
         except ValueError:
             pass
@@ -193,7 +192,7 @@ def name_filter(record):
     if parts[-1] == '__init__':
         del parts[-1]
     if parts[0] == path.anchor:
-        del parts[0]
+        parts[0] = '/'
     record.name = str(PurePosixPath(*parts))
     return True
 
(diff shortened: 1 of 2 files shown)


picard/log.py
```diff
@@ -177,11 +179,15 @@ def name_filter(record):
...
-            if not DebugOpt.PLUGIN_FULLPATH.enabled and 'plugins' in path.parts:
-                parts = list(reversed(path.parts))
-                parts = parts[:parts.index('plugins') + 1]
-                path = Path(*reversed(parts))
+
+    if path.is_absolute() and not DebugOpt.PLUGIN_FULLPATH.enabled:
+        try:
+            path = path.resolve().relative_to(USER_PLUGIN_DIR)
+            parts = list(p for p in path.parts if not p.endswith('.zip'))
+            parts.insert(0, 'plugins')
```
zas, line 187: I don't understand this insert.


```python
>>> path = Path('/a/b/c/subdir/e/f.zip/g')
>>> USER_PLUGIN_DIR = Path('/a/b/c/subdir')
>>> path = path.resolve().relative_to(USER_PLUGIN_DIR)
>>> str(path)
'e/f.zip/g'
>>> parts = list(p for p in path.parts if not p.endswith('.zip'))
>>> parts
['e', 'g'] 
>>> parts.insert(0, 'plugins')
>>> parts
['plugins', 'e', 'g']
>>> path = Path(*parts)
>>> str(path)
'plugins/e/g'
```

Why would we introduce an element which isn't in any of existing paths?

test/test_log.py
```diff
@@ -162,21 +164,34 @@ def test_5(self):
...
+        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip')
         self.assertTrue(name_filter(record))
-        self.assertEqual(record.name, 'plugins/path3/plugins/plugin')
+        self.assertEqual(record.name, 'user/picard/plugins/plugin')
+
+    def test_plugin_path_long_2(self):
+        DebugOpt.PLUGIN_FULLPATH.enabled = True
+        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip/plugin')
+        self.assertTrue(name_filter(record))
+        self.assertEqual(record.name, 'user/picard/plugins/plugin.zip/plugin')
```
zas, line 177: Why are we ending with a path not starting with `/` here?

test/test_log.py
```diff
@@ -162,21 +164,34 @@ def test_5(self):
...
+        DebugOpt.PLUGIN_FULLPATH.enabled = False
+        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip')
+        self.assertTrue(name_filter(record))
+        self.assertEqual(record.name, 'plugins/plugin')
 
-    def test_plugin_path_short(self):
+    def test_plugin_path_short_2(self):
         DebugOpt.PLUGIN_FULLPATH.enabled = False
-        record = FakeRecord(name=None, pathname='/path1/path2/plugins/path3/plugins/plugin.zip')
+        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip/plugin')
```
zas, line 187: It would be better to use `'/user/picard/plugins/plugin.zip/xxx' here because we don't know which `plugin` was kept in the result.

picard/log.py
rdswift: > Why would we introduce an element which isn't in any of existing paths?

It sort of is in the original `USER_PLUGIN_DIR` in Picard, but gets stripped out when the path is resolved relative to `USER_PLUGIN_DIR`.  My thinking was to add it back in to clearly indicate in the log that this was in a plugin.

I'll remove it.

test/test_log.py
rdswift: I believe that's because of stripping out the anchor in the lines in `log.py`:

```python
    if parts[0] == path.anchor:
        del parts[0]
```

That's the same behavior we're seeing when the path isn't relative to the `picard` module.  For example, in `test_log.py`:

```python
@unittest.skipIf(IS_WIN, "Posix test")
@patch('picard.log.picard_module_path', PurePosixPath('/picard'))
class NameFilterTestAbs(PicardTestCase):


    def test_1(self):
        record = FakeRecord(name=None, pathname='/path/module/file.py')
        self.assertTrue(name_filter(record))
        self.assertEqual(record.name, 'path/module/file')
```

If you want the absolute (non-shortened) paths to start with a "/" (for both Windows and POSIX) the code could be changed to:

```python
    if parts[0] == path.anchor:
        parts[0] = '/'
```

I'll make the change, because it will make it clearer that the path is absolute.

test/test_log.py
```diff
@@ -168,25 +168,25 @@ def test_plugin_path_long_1(self):
...
-        self.assertEqual(record.name, 'plugins/plugin')
+        self.assertEqual(record.name, 'plugin')
 
     def test_plugin_path_short_2(self):
         DebugOpt.PLUGIN_FULLPATH.enabled = False
-        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip/plugin')
+        record = FakeRecord(name=None, pathname='/user/picard/plugins/plugin.zip/xxx')
         self.assertTrue(name_filter(record))
-        self.assertEqual(record.name, 'plugins/plugin')
+        self.assertEqual(record.name, 'xxx')
```
zas, line 189: In this case it looks weird to lose the plugin name (from `plugin.zip`), I guess having `plugin/xxx` as result would make much more sense.

rdswift: What do you think of prefacing all shortened plugin paths with "Plugin:" to clearly indicate that's what they are?  For example, "/home/user/picard/plugins/xxx.zip/yyy" would be logged as "Plugin: yyy" and "/home/user/picard/plugins/xxx.zip" would be logged as "Plugin: xxx" (assuming that `USER_PLUGIN_DIR` was "/home/user/picard/plugins").

Otherwise, I can just put it back to starting with "plugins/".  I'm okay either way.  Just let me know what you prefer.

zas: Hmmm, I now get why you added `plugins/` at first. And I guess that's the best option.

rdswift: Actually, in looking at the plugins, stripping out the zip file from the path might not be such a good idea because in some cases the file inside may be `__init__.py`.  Perhaps we should leave the zip file name in the path to help ensure that the plugin is properly identified.  That way, some examples might be:

* `Plugin: abbreviate_artistsort.zip/abbreviate_artistsort`
* `Plugin: add_to_collection.zip/__init__`
* `Plugin: extra/path/plugin.zip/__init__`

Or we could stop at the zip file and strip the ".zip", but that wouldn't show which file it was within the plugin (if there were multiple modules).  For example, the three above would be:

* `Plugin: abbreviate_artistsort`
* `Plugin: add_to_collection`
* `Plugin: extra/path/plugin`

Considering that we have the option to show the full path if needed, I'm actually tempted to do it this way.  What do you think?

rdswift: I'll put it back to just including the "plugin/" for now and we can change later if we decide we want to go a different route.

rdswift: Another option, and I think this might actually make the most sense, is to not add the "plugins" back in, but also not remove the zip file from the path.  This keeps it as a shortened (relative) path, and the presense of a zip file in the path clearly indicates that this is a plugin.  Some examples might be:

* `abbreviate_artistsort.zip/abbreviate_artistsort`
* `add_to_collection.zip/__init__`
* `some_plugin.zip/module_path/some_module`
* `extra/path/plugin.zip/__init__`

zas (approved):
LGTM

rdswift:
Revised the logic slightly to include the zip file in the shortened plugin path for clarity.  Also fixed some tests that I had screwed up regarding absolute path testing, and added additional tests.

zas (approved):
LGTM

PR #2403, opened 2024-04-15
Build SETTINGS_GROUPS from Options pages

zas:
Not perfect yet, but that's the idea.

I think:
- we can get rid of `UserProfileGroups` totally
- we need a multilevel tree for Profiles (like the one for option pages)
- we could reduce the need of highlight with option name = widget name and defaulting to that if not provided
- tests are broken, because not initialized in tests

@rdswift give it a try, and see if you can improve this

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -187,6 +187,8 @@ class SettingConfigSection(ConfigSection):
         opt = Option.get(self.__name, name)
         if opt is None:
             return None
+        if opt.title is not None and name not in UserProfileGroups.ALL_SETTINGS:
+            log.debug("Option %s has a title (%s), but it isn't in UserProfileGroups", opt.name, opt.title)
         return self.value(name, opt, opt.default)
 
     def __setitem__(self, name, value):
@@ -384,6 +386,12 @@ class Option(QtCore.QObject):
         self.registry[key] = self
 
     @classmethod
+    def getall(cls, section=None):
+        for opt in cls.registry.values():
+            if section is None or opt.section == section:
+                yield opt
+
+    @classmethod
     def get(cls, section, name):
         return cls.registry.get((section, name))
 
@@ -461,7 +469,7 @@ def setup_config(app, filename=None):
     setting = config.setting
     persist = config.persist
     profiles = config.profiles
-    UserProfileGroups.initialize()
+    UserProfileGroups.initialize(config)
 
 
 def get_config():
--- a/picard/config.py
+++ b/picard/config.py
@@ -41,7 +41,10 @@ from picard import (
     PICARD_VERSION,
     log,
 )
-from picard.profile import UserProfileGroups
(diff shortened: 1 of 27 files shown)


picard/config.py
```diff
@@ -371,14 +371,15 @@ class Option(QtCore.QObject):
...
+    def __init__(self, section, name, default, title=None, highlight=None):
         key = (section, name)
         if key in self.registry:
             raise OptionError("Already declared", section, name)
         super().__init__()
         self.section = section
         self.name = name
         self.default = default
+        self.highlight = highlight
         self.title = title
```
rdswift, line 383: If there is a title set, and highlight is None, perhaps we could set self.highlight to an empty list?  Something like:

```suggestion
        self.title = title
        self.highlight = highlight
        if self.title is not None and not self.highlight:
            self.highlight = []
```

That way we may be able to simply avoid setting the highlight attribute when adding an option to an option page, rather than having to include the setting `highlight=[]`.  Even though there aren't very many where this would apply, this might simplify entering an option on one of the pages.  If there is no highlight specified, then nothing will be highlighted.

rdswift:
> I think:
> 
> * we can get rid of `UserProfileGroups` totally

That would be great, and would simplify things for maintaining or adding a new option setting because everything would be in one place.

> * we need a multilevel tree for Profiles (like the one for option pages)

This sounds good, but I'm having trouble picturing what it would look like.

> * we could reduce the need of highlight with option name = widget name and defaulting to that if not provided

I like this, but I have to give it some more thought to see how it might actually work.  Anything we can do to automate and simplify will be an improvement over the way I set it up initially.

> * tests are broken, because not initialized in tests

Perfectly understandable for work in progress.  :wink: 

All in all, I really like this conceptually, and appreciate the considerable work that you've put into it already.

picard/tagger.py
```diff
@@ -244,6 +245,7 @@ def __init__(self, picard_args, localedir, autoupdate, pipe_handler=None):
         self.__class__.__instance = self
         setup_config(self, picard_args.config_file)
         config = get_config()
+        init_settings_groups()
```
zas, line 248: This method should be called from `setup_config()`, it has to be executed before any access to config, and after options were defined.

rdswift:
> I'm not to sure about this, the option `selected_file_naming_script_id` is a special case. What do you think?

I see that this and others are listed in a set called `OPTIONS_NOT_IN_PAGES` in the `picard.ui.options.maintenance` module.

The `selected_file_naming_script_id` option is actually defined in the `picard.ui.scripteditor` module.  I wonder if it would make more sense defining it in the `picard.ui.options.renaming` module instead, and only have an `add_if_missing()` definition in the `picard.ui.scripteditor` module?  I think that might help address the issue.

zas:
This move has too many side-effects, I prefer to not do it now.

PR #2404, opened 2024-04-16
PICARD-2857: exit with 0 code if no error when connecting to an exist…

zas:
…ing instance

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

We had EXIT_NO_NEW_INSTANCE = 30403 code in this case, but it's a non-zero code and that's not expected when everything was normal. Only 8 bits were preserved on actual exit code (195).

* JIRA ticket (_optional_): PICARD-2857

# Solution

Exit with code 0 instead.

# Action

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1506,8 +1506,6 @@ If a new instance will not be spawned files/directories will be passed to the ex
 
 
 def main(localedir=None, autoupdate=True):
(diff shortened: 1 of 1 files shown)


rdswift (approved):
This makes perfect sense to me.  I see no real need to exit with a non-zero code.

PR #2406, opened 2024-04-17
PICARD-2729: Allow disabling date sanitazation for APE and Vorbis tags

ShubhamBhut:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

Added a new config setting which is checked before sanitizing date for Vorbis and APE types. Config setting can be changed by a checkbox added in Metadata section of the Options 

# Action

Similar as Solution


picard/formats/apev2.py
```diff
@@ -160,7 +160,9 @@ def _load(self, filename):
                     name = name_lower
                     if name == 'year':
                         name = 'date'
-                        value = sanitize_date(value)
+                        config = get_config()
```
zas, line 163: It would be much better to set `config` outside loops.
In this case, I would even move `config.setting['disable_date_sanitize']` outside the loops and cache the value.

Something like:

```python
        config = get_config()
        date_sanitize =  not config.setting['disable_date_sanitize']
        if file.tags:
            for origname, values in file.tags.items():
...
                            if date_sanitize:
                                value = sanitize_date(value) 
```

picard/formats/vorbis.py
```diff
@@ -142,7 +142,9 @@ def _load(self, filename):
                 name = origname
                 if name in {'date', 'originaldate', 'releasedate'}:
                     # YYYY-00-00 => YYYY
-                    value = sanitize_date(value)
+                    config = get_config()
```
zas, line 145: Same here.

zas (changes requested):
Also tests have to be fixed.

picard/formats/apev2.py
```diff
@@ -135,6 +135,8 @@ def _load(self, filename):
         self.__casemap = {}
         file = self._File(encode_filename(filename))
         metadata = Metadata()
+        config = get_config()
+        date_sanitize = not config.setting['disable_date_sanitize']
```
zas, line 139: Since those variables are only used in `if file.tags` they should be moved inside the branch, just before the `for` loop.

zas (changes requested):

picard/formats/vorbis.py
```diff
@@ -136,13 +136,16 @@ def _load(self, filename):
...
+        date_sanitize = not config.setting['disable_date_sanitize']
         for origname, values in file.tags.items():
             for value in values:
                 value = value.rstrip('\0')
                 name = origname
                 if name in {'date', 'originaldate', 'releasedate'}:
                     # YYYY-00-00 => YYYY
-                    value = sanitize_date(value)
+                    if date_sanitize:
+                        value = sanitize_date(value)
```
rdswift, line 148: Since the variable `date_sanitize` is only used in one place, is it really necessary, or should the test in Line 147 simply be `if not config.setting['disable_date_sanitize']:` similar to the way it is done later in Line 286?

rdswift:
Also a reminder that the documentation will need to be updated once this is merged.

picard/formats/vorbis.py
zas: > Since the variable `date_sanitize` is only used in one place, is it really necessary, or should the test in Line 147 simply be `if not config.setting['disable_date_sanitize']:` similar to the way it is done later in Line 286?

It is better to cache config variables used in loops. Getting their values imply quite a lot of code.

rdswift:
Perhaps this is a dumb question, but would there ever be a case where you might want to only apply this to one of the formats?  If so, would it make sense to have separate settings for APE and Vorbis?

picard/formats/vorbis.py
rdswift: Right.  I completely missed that.  Sorry.

zas:
> Perhaps this is a dumb question, but would there ever be a case where you might want to only apply this to one of the formats? If so, would it make sense to have separate settings for APE and Vorbis?

Actually we could have an option that isn't a boolean, but rather a list of formats, so we can apply this to any chosen format instead.

picard/ui/options/metadata.py
```diff
@@ -92,6 +92,7 @@ class MetadataOptionsPage(OptionsPage):
         ListOption('setting', 'script_exceptions', [], title=N_("Translation script exceptions")),
         BoolOption('setting', 'release_ars', True, title=N_("Use release relationships")),
         BoolOption('setting', 'track_ars', False, title=N_("Use track and release relationships")),
+        BoolOption('setting', 'disable_date_sanitize', False, title=N_("Disable date sanitization for APE and Vorbis tags")),
```
zas, line 95: As discussed, it would be better to have a list of formats for which we want to disable date sanitization.
It can be any format NAME has reported by https://github.com/metabrainz/picard/blob/16b2ddc9472a294ad636fd7858f731d724eb2af7/picard/formats/util.py#L43

And this `BoolOption` can be converted to a set:

```python
Option('setting', 'disable_date_sanitize', set(), title=N_("Disable date sanitization for specified formats")),
```

Of course a selector should let the user select multiple formats.

We could use a multiselect combobox (see, for example, https://stackoverflow.com/a/77755095), and add the widget to https://github.com/metabrainz/picard/tree/master/picard/ui/widgets for reuse.

zas (changes requested):

rdswift:
Added PICARD-2862 to update the documentation.

zas:
@ShubhamBhut what do you think about suggested approach?

ShubhamBhut:
> @ShubhamBhut what do you think about suggested approach?

I understood it, except that there are only 3 tags vorbis, mp3/id3 and APE which uses sanitize_date so I think it would be implemented for only these 3 tags ?

Also apologies for late update, it's just that I have been tightly occupied for last couple of days. I will implement it tomorrow.

zas:
> I understood it, except that there are only 3 tags vorbis, mp3/id3 and APE which uses sanitize_date so I think it would be implemented for only these 3 tags ?

I'd add a method to those classes that wraps `util.sanitize_date()` so we can test if this method exists and add the matching class to the selector.

```python
from picard.util import sanitize_date

class APEv2File(File):

    sanitize_date = sanitize_date
....
                    if name == 'year':
                        name = 'date'
                        value = self.sanitize_date(value)
```

In [formats/util.py](https://github.com/metabrainz/picard/blob/05ddd5fc461d49f83e90cfab9bd1b2160e25ac0b/picard/formats/util.py) we need to add a method returning all known format classes having this method:

```python
def formats_with_sanitize_date():
    for fmt in _formats:
        if hasattr(fmt, 'sanitize_date'):
             yield fmt
```

In selector code, you can iterate those classes:

```python
from picard.formats.util import formats_with_sanitize_date

for fmt in formats_with_sanitize_date():
    do_something_with(fmt)
```

Totally untested, but you get the idea I guess.

picard/formats/util.py
```diff
@@ -45,6 +45,10 @@ def supported_formats():
     return [(file_format.EXTENSIONS, file_format.NAME) for file_format in _formats]
 
 
+def formats_with_sanitize_date():
+    return [f"{fmt.EXTENSIONS, fmt.NAME}" for fmt in _formats if hasattr(fmt, 'sanitize_date')]
```
zas, line 49: It should return `fmt` classes only and we could make it a generator:

```python
def formats_with_sanitize_date():
    for fmt in _format:
        if hasattr(fmt, 'sanitize_date'):
            yield fmt
```

To get the sorted list of names:

```python
fmt_names =  sorted(fmt.NAME for fmt in formats_with_sanitize_date())
```

picard/formats/apev2.py
```diff
@@ -136,6 +137,8 @@ def _load(self, filename):
         file = self._File(encode_filename(filename))
         metadata = Metadata()
         if file.tags:
+            config = get_config()
+            date_sanitize = not config.setting['disable_date_sanitize'] and not any('.ape' in fmt[0] for fmt in config.setting['formats_to_disable_date_sanitize'])
```
zas, line 141: I don't think we still need `not config.setting['disable_date_sanitize']`.

Also I think it should apply to **file format** not **file extensions** (one format, multiple extensions).

The test should be  as simple as:

```python
date_sanitize = self.NAME in config.setting['formats_to_disable_date_sanitize'])
```

See https://github.com/metabrainz/picard/blob/226ebef845151cdb3a89698e5802340b29edfc57/picard/formats/apev2.py#L354-L373 , basically all formats have a unique name (but each can match multiple file extensions)

zas (changes requested):

picard/ui/options/metadata.py
```diff
@@ -92,6 +92,8 @@ class MetadataOptionsPage(OptionsPage):
...
+        ListOption('setting', 'formats_to_disable_date_sanitize', [], title=N_("Formats to disable date sanitize")),
```
zas, line 96: Since each format should appear only once I guess a set() is a better type for this option.

picard/ui/widgets/multicombobox.py
```diff
@@ -0,0 +1,73 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2019-2020, 2022 Philipp Wolfer
+# Copyright (C) 2020-2021, 2024 Laurent Monin
```
zas, line 6: This should be updated, I guess it was copied over.
Add your name, and current year and remove others (we didn't edit this file yet).

picard/ui/widgets/multicombobox.py
```diff
@@ -0,0 +1,73 @@
...
+        self.model().appendRow(item)
+
+    def addItems(self, items_list: list):
+        for text in items_list:
+            self.addItem(text)
+
+    def updateText(self):
+        selected_items = [self.model().item(i).text() for i in range(self.model().rowCount())
+                          if self.model().item(i).checkState() == Qt.CheckState.Checked]
+        self.lineEdit().setText(", ".join(selected_items))
```
zas, line 52: Not sure about this, what happens if item text contains ', '?

I would expect we end with a list

zas (changes requested):

ShubhamBhut:
Hey @zas, actually this is not complete yet. For some reason, the marked formats of multicombobox are not saved. I have yet to figure it out.

picard/ui/widgets/multicombobox.py
ShubhamBhut: comma (,) of item text doesn't seem to impact the output, I have added a new method which returns the list containing marked formats

zas:
> Hey @zas, actually this is not complete yet. For some reason, the marked formats of multicombobox are not saved. I have yet to figure it out :(

See https://github.com/ShubhamBhut/picard/pull/1

- I don't think we need 2 options, if the set is empty that's enough
- We don't usually add our own widgets classes to .ui, we just override a dummy widget or just insert it in the layout.

picard/ui/ui_options_metadata.py
```diff
@@ -2,16 +2,11 @@
 #
 # Created by: PyQt6 UI code generator 6.6.1
 #
-# Automatically generated - do not edit.
-# Use `python setup.py build_ui` to update it.
+# WARNING: Any manual changes made to this file will be lost when pyuic6 is
+# run again.  Do not edit this file unless you know what you are doing.
```
zas, line 6: This file has to be generated by running `python setup.py build_ui`

zas (changes requested):
It looks much better.
BTW, you shouldn't rebase all the time,Rebasing is useful, but commit history is too.

You didn't merge my PR?

ShubhamBhut:
> It looks much better. BTW, you shouldn't rebase all the time,Rebasing is useful, but commit history is too.
> 
> You didn't merge my PR?

Apologies, I will keep that in mind. Actually I was using jujutsu for version control and it doesn't have pre-commit hooks support yet. I will stick to vanilla git from now on.

picard/ui/ui_options_metadata.py
ShubhamBhut:
```
# WARNING: Any manual changes made to this file will be lost when pyuic6 is
# run again.  Do not edit this file unless you know what you are doing.
```

is to be manually added, right ? running `python setup.py build_ui` doesn't seem to add that

zas:
@ShubhamBhut what's the status of this PR?

PR #2411, opened 2024-04-17
Scripteditor cleanup

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/scripteditor.py
+++ b/picard/ui/scripteditor.py
@@ -843,8 +843,7 @@ class ScriptEditorDialog(PicardDialog, SingletonDialog):
     def update_script_text(self):
         """Updates the combo box item with changes to the current script.
         """
-        selected = self.ui.preset_naming_scripts.currentIndex()
-        script_item = self.ui.preset_naming_scripts.itemData(selected)
+        selected, script_item = self.get_selected_index_and_item()
         script_item['script'] = self.get_script()
         self.update_combo_box_item(selected, script_item)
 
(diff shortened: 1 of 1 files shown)


rdswift (approved):
This looks fine to me.

The reason I initially set the keys to constants was to avoid the possibility of a problem if a key name was mis-spelled.  Also a habit I developed a long time ago to save space by avoiding duplicating strings when programming in C.  Not as relevant in Python, but old habits die hard.  :wink:

zas:
> This looks fine to me.
> 
> The reason I initially set the keys to constants was to avoid the possibility of a problem if a key name was mis-spelled. Also a habit I developed a long time ago to save space by avoiding duplicating strings when programming in C. Not as relevant in Python, but old habits die hard. 😉

Actually that's not a bad habit, and in general actually better. But for config keys it makes things harder to follow, and Python will internally use the same object (using an intermediate variable will actually use more memory, because it will create an object for it in addition of the string itself):

```python
>>> s = 'abc'
>>> a = 'abc'
>>> id(s)
140000717587760
>>> id(a)
140000717587760
>>> a = 'def'
>>> id(a)
140000717992048
>>> print(id('abc'))
140000717587760
```

As you can see above, the string 'abc' has always the same id.

Also those strings are unlikely to change a lot (they were defined as "constants"), and searching/replacing them is a non-issue with good tools.

Also for Picard config keys (and dict keys in general) I prefer we keep them as single quoted strings whenever possible (rather than mixing up single & double quotes, it only makes search/replace more complicated).

PR #2414, opened 2024-04-19
PICARD-2863: Add a tooltip with additional disambiguation to release versions menu

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2863

Disambiguation is lacking when releases versions are very similar (like 3 digital releases over 3 years).
But we are limited by the width of the menu to display everything needed.

# Solution

In addition of current output, add a tooltip to each menu item showing the extra infos needed.

versions_tooltip

# Action

Additional actions required:


phw (dismissed):
Makes sense. Better than to try to show everything in the menu itself.

The comments on the ticket also I dictate there might be a bug with the selection on which fields to show in the menu.

zas:
> The comments on the ticket also I dictate there might be a bug with the selection on which fields to show in the menu.

According to my tests, everything works as it should. But I may have missed something.

phw (approved):

PR #2418, opened 2024-04-20
PICARD-2869: Register a global exception handler

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2869

Register a handler on sys.excepthook and have it call the global crash_handler(). This allows to show the exception dialog in otherwise unhandled exceptions, e.g. has happened in PICARD-2868

grafik

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -83,8 +83,11 @@ def crash_handler(exc: Exception = None):
     from tempfile import NamedTemporaryFile
     import traceback
     if exc:
-        # Since Python 3.10 calling only traceback.format_exception(exc) is possible
-        trace = "\n".join(traceback.format_exception(type(exc), exc, exc.__traceback__))
+        if sys.version_info < (3, 10):
+            trace_list = traceback.format_exception(None, exc, exc.__traceback__)
+        else:
+            trace_list = traceback.format_exception(exc)  # pylint: disable=no-value-for-parameter
+        trace = "".join(trace_list)
     else:
         trace = traceback.format_exc()
     logfile = None
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006-2008, 2011-2014 Lukáš Lalinský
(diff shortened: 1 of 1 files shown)


picard/__init__.py
```diff
@@ -83,7 +82,11 @@ def crash_handler():
     # with minimum chance to fail.
     from tempfile import NamedTemporaryFile
     import traceback
-    trace = traceback.format_exc()
+    if exc:
+        # Since Python 3.10 calling only traceback.format_exception(exc) is possible
```
zas, line 86: I think we should test Python version, this is likely to be deprecated at some point, it'll make easier to spot it and remove the need of a comment :

```python
        if sys.version_info < (3, 10):
            trace_list = traceback.format_exception(None, exc, exc.__traceback__)
        else:
            trace_list = traceback.format_exception(exc)
```

Also the first argument is ignored since Python 3.5 (hence None).

By the way, to maintain compatibility they wrote rather hacky code: https://github.com/python/cpython/blob/812245ecce2d8344c3748228047bab456816180a/Lib/traceback.py#L93-L101

picard/__init__.py
```diff
@@ -83,7 +82,11 @@ def crash_handler():
...
+        trace = "\n".join(traceback.format_exception(type(exc), exc, exc.__traceback__))
```
zas, line 87: `"\n"` here causes an extra newline, https://docs.python.org/3.10/library/traceback.html#traceback.format_exception says: "The return value is a list of strings, each ending in a newline and some containing internal newlines."

So I think we can just `"".join()` here.

zas (changes requested):

picard/__init__.py
phw: You are right. This is also what `traceback.format_exc`  does internally.

picard/__init__.py
phw: Done in be432d0701dd7ee993a86dcc0b104a60b6b4147a

zas (approved):
LGTM

PR #2419, opened 2024-04-20
WIP: Plugins v3

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-1861

This is the work in progress of a proposed rework of the plugin system.

The main part here is an initial specification of the plugin API in [PLUGINS.md](https://github.com/phw/picard/blob/plugins-v3/PLUGINS.md). The document builds upon the extended discussion of requirements for a new plugin system on [the wiki](https://github.com/rdswift/picard-plugins/wiki/Picard-Plugins-System-Proposal). But currently the focus is on the plugin API and loading inside Picard, not so much on the distributions side. But @zas had some good ideas here of using git, and parts of this are documented in the spec

Beside the document there is only very rough prototyping to evaluate the ideas a bit. The very last commit does do some initial plugin loading, but this was committed as current state very much in the middle of the work.

This is IMHO not yet ready for review, but as I had little time the recent weeks let's have this up here to build upon. So far it was mostly @zas and myself discussing some ideas. I think the general direction is solid, but we should put this up for a wider discussion.

# Solution

# Action

Additional actions required:

### DIFF
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -61,10 +61,6 @@ example/
   MANIFEST.toml
 ```
 
-The plugin package MAY be put into a ZIP archive. In this case the filename
-must be the same as the plugin package name followed by `.picard.zip`, e.g.
-`example.picard.zip`.
-
 
 ### File system locations
 TBD
@@ -128,7 +124,6 @@ The file MAY define any of the following optional fields:
 
 | Field name     | Type   | Description                                                      |
 |----------------|--------|------------------------------------------------------------------|
-| extract        | bool   | If set to `true` the plugin must be extracted on installation (to be discussed)   |
 | license-url    | string | URL to the full license text                                     |
 | user-guide-url | string | URL to the plugin's documentation                                |
 
@@ -144,7 +139,6 @@ This is an example plugin showcasing the new **Picard 3 plugin** API.
 You can use [Markdown](https://daringfireball.net/projects/markdown/) for formatting."""
 version     = "1.0.0"
 api         = ["3.0", "3.1"]
-extract     = true
 license     = "CC0-1.0"
 license-url = "https://creativecommons.org/publicdomain/zero/1.0/"
 user-guide-url = "https://example.com/"
@@ -281,16 +275,55 @@ class PluginApi:
 TBD
 
 
-### Plugin live cycle
+### Plugin life cycle
 TBD
 
 
 ### To be discussed
 
-- Localization?
-- Categorization? [PW-12](https://tickets.metabrainz.org/browse/PW-12)
-- Extra data files?
-- Additional extension points?
+#### Localization
+Existing plugins in Picard 2 cannot be localized. The new plugin system should
+allow plugins to provide translations for user facing strings.
+
+Plugins could provide gettext `.mo` files that will be loaded under a plugin
+specific translation domain.
+
+Also the description from `MANIFEST.json` should be localizable.
+
+
+#### Categorization
+See [PW-12](https://tickets.metabrainz.org/browse/PW-12)
+
+
+#### Extra data files
+Does the Plugin API need to expose functions to allow plugins to easily load
+additional data files shipped as part of the plugins? E.g. for loading
+configuration from JSON files.
+
+
+#### Additional extension points
+Which additional extension points should be supported?
+
+
+#### Support for ZIP compressed plugins:
+As before plugins in a single ZIP archive could also be supported. The "Format"
+section above could be extended with:
+
+> The plugin package MAY be put into a ZIP archive. In this case the filename
+> must be the same as the plugin package name followed by `.picard.zip`, e.g.
+> `example.picard.zip`.
+
+It needs to be discussed whether such plugins should be extracted by default
+or whether module loading from ZIP should be retained.
+
+The advantage of loading directly from ZIP is the simplicity of plugin handling,
+as the user can move around a single plugin file.
+
+Disadvantages are:
+
+- Additional complexity in the module loader
+- Inability of accessing shared libraries shipped as part of the plugin
+- No bytecode caching
 
 
 ## Implementation considerations
--- a/PLUGINS.md
+++ b/PLUGINS.md
@@ -1,0 +1,302 @@
+# Picard Plugin API v3 (Proposal)
+
+## Introduction / Motivation
+TBD
+
+
+## Scope
+
+This document only discusses the structure and API for Picard plugins. It does not
+discuss distribution of plugins or the maintenance of a plugins repository. See
+[the wiki](https://github.com/rdswift/picard-plugins/wiki/Picard-Plugins-System-Proposal)
+for and extended discussion of distribution and maintenance.
+
+
+## Limitations of the old plugin system
+
+- **No separation of metadata and code:** As the metadata, such as plugin name
+  and description, but also supported API versions, is part of the Python code,
+  each installed plugin's code was executed regardless of whether the plugin
+  is enabled or even compatible with the current Picard version.
+
+- **No defined API:** Apart from a few methods to register plugin hooks there
+  is no actual API provided. This makes it both difficult for plugin developers
+  to decide what parts of Picard can be safely used as well for Picard developers
+  to decide which internal change should be considered a breaking change for plugins.
+
+- **Imports scattered over the codebase:** The different functions for registering
+  plugin hooks as well as the objects provided by Picard that are actually useful
+  for plugins are all scattered over the Picard code base. While this follows a
+  system that is logical if you are familiar with Picard's code base, this is not
+  transparent to plugin developers.
+
+- **Many supported plugin formats:** The old system allowed multiple ways how a
+  plugin can be structured. The following formats where supported:
+  - A single Python module (`example.py`)
+  - A Python package (`example/__init__.py`)
+  - A ZIP archive (`example.zip`) containing a single Python module
+  - A ZIP archive (`example.zip`) containing a Python package
+  - A ZIP archive (`example.picard.zip`) with either a Python module or package
+    and an additional metadata file `MANIFEST.json`.
+  This variation leeds to extra complexity in the implementation and increases
+  maintenance and testing effort. It also increased complexity for users, as they
+  needed to decide whether a plugin file needs to be placed at the top level
+  or inside a directory.
+
+
+## Format
+
+A Picard plugin MUST be a Python package, that is a directory with a valid Python
+package name containing at least a single `__init__.py` file. The package directory
+MUST also contain a manifest file which provides metadata about the plugin.
+
+The package directory MAY contain additional files, such as Python modules to load.
+
+A basic plugin `example` could have the following structure:
+
+
+```
+example/
+  __init__.py
+  MANIFEST.toml
+```
+
+The plugin package MAY be put into a ZIP archive. In this case the filename
+must be the same as the plugin package name followed by `.picard.zip`, e.g.
+`example.picard.zip`.
+
+
+### File system locations
+TBD
+
+
+### Package structure and implemented API
+
+The package MUST define the following top-level functions:
+
+- `enable` gets called when the plugin gets enabled. This happens on startup for
+  all enabled plugins and also if the user enables a previously disabled plugin.
+  The function gets passed an instance of `picard.plugin.PluginApi`, which
+  provides access to Picard's official plugin API and allows to register plugin hooks.
+
+The package MAY define the following top-level functions:
+
+- `disable` gets called when the plugin gets disabled. The plugin should stop all
+  processing and free required resources. The plugin does not need to de-register
+  plugin hooks, as those get disabled automatically.
+  After being disabled the plugin can always be enabled again (the `enable`
+  function gets called).
+
+> ***Discussion:** Are `install` and `uninstall` hooks needed?*
+
+
+A basic plugin structure could be:
+
+```python
+from picard.plugin import PluginApi
+
+def enable(api: PluginApi) -> None:
+    # api can be used to register plugin hooks and to access essential Picard APIs.
+    pass
+
+def disable() -> None:
+    pass
+```
+
+The plugin MUST NOT perform any actual work, apart from defining types and
+functions, on import. All actual processing must be performed only as part of
+the `enable` and `disable` functions and any plugin hooks registered in `enable`.
+
+
+### Manifest format
+The plugin's package directory MUST contain a file `MANIFEST.toml`.
+
+> ***Discussion:** Is TOML the proper format, or should something like JSON or YAML be preferred?*
+
+The file MUST define the following mandatory metadata fields:
+
+| Field name     | Type   | Description                                                      |
+|----------------|--------|------------------------------------------------------------------|
+| name           | string | The plugin's full name                                           |
+| author         | string | The plugin author                                                |
+| description    | string | Detailed description of the plugin. Supports Markdown formatting |
+| version        | string | Plugin version. Use semantic versioning in the format "x.y.z"    |
+| api            | list   | The Picard API versions supported by the plugin                  |
+| license        | string | License, should be a [SPDX license name](https://spdx.org/licenses/) and GPLv2 compatible |
+
+The file MAY define any of the following optional fields:
+
+| Field name     | Type   | Description                                                      |
+|----------------|--------|------------------------------------------------------------------|
+| extract        | bool   | If set to `true` the plugin must be extracted on installation    |
+| license-url    | string | URL to the full license text                                     |
+| user-guide-url | string | URL to the plugin's documentation                                |
+
+
+Example `MANIFEST.toml`:
+
+```toml
+name        = "Example plugin"
+author      = "Philipp Wolfer"
+description = """
+This is an example plugin showcasing the new **Picard 3 plugin** API.
+
+You can use [Markdown](https://daringfireball.net/projects/markdown/) for formatting."""
+version     = "1.0.0"
+api         = ["3.0", "3.1"]
+extract     = true
+license     = "CC0-1.0"
+license-url = "https://creativecommons.org/publicdomain/zero/1.0/"
+user-guide-url = "https://example.com/"
+```
+
+
+### Picard Plugin API
+
+As described above the plugin's `enable` function gets called with an instance
+of `picard.plugin.PluginApi`. `PluginApi` provides access to essential Picard
+APIs and also allows registering plugin hooks.
+
+`PluginApi` implements the interface below:
+
+```python
+from typing import (
+    Callable,
+    Type,
+)
+from logging import Logger
+
+from picard.config import (
+    Config,
+    ConfigSection,
+)
+from picard.coverart.providers import CoverArtProvider
+from picard.file import File
+from picard.plugin import PluginPriority
+from picard.webservice import WebService
+from picard.webservice.api_helpers import MBAPIHelper
+
+from picard.ui.itemviews import BaseAction
+from picard.ui.options import OptionsPage
+
+class PluginApi:
+    @property
+    def web_service(self) -> WebService:
+        pass
+
+    @property
+    def mb_api(self) -> MBAPIHelper:
+        pass
+
+    @property
+    def logger(self) -> Logger:
+        pass
+
+    @property
+    def global_config(self) -> Config:
+        pass
+
+    @property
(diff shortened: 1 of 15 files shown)


picard/plugin3/api.py
```diff
@@ -0,0 +1,170 @@
...
+        return self._logger
+
+    @property
+    def global_config(self) -> Config:
+        return get_config()
+
+    @property
+    def plugin_config(self) -> ConfigSection:
+        """Configuration private to the plugin"""
+        return self._api_config
```
zas, line 100: @rdswift ^^ (about plugins options)

PLUGINS.md
```diff
@@ -0,0 +1,447 @@
...
+A Picard plugin MUST be a Python package, that is a directory with a valid Python
+package name containing at least a single `__init__.py` file. The package directory
+MUST also contain a manifest file which provides metadata about the plugin.
+
+The package directory MAY contain additional files, such as Python modules to load.
+
+A basic plugin `example` could have the following structure:
+
+
+```
```
zas, line 87:
```suggestion
```text
```

https://github.com/DavidAnson/markdownlint/blob/main/doc/Rules.md#md040---fenced-code-blocks-should-have-a-language-specified

PLUGINS.md
```diff
@@ -0,0 +1,447 @@
...
+Picard must provide a user interface for installing third-party plugins which
+are not provided in the official plugin list. The user needs to enter the
+plugin's git URL and Picard will verify the manifest and offer to install and
+activate the plugin. The UI must make it clear that the user is installing the
+plugin at their own risk and that the plugin can execute arbitrary code.
+
+
+## To be discussed
+
+### Localization
```
zas, line 391:
```suggestion
### Localization (l10n) and internationalization (i18n)
```

PLUGINS.md
```diff
@@ -0,0 +1,447 @@
...
+
+Existing plugins in Picard 2 cannot be localized. The new plugin system should
+allow plugins to provide translations for user facing strings.
+
+Plugins could provide gettext `.mo` files that will be loaded under a plugin
+specific translation domain.
```
zas, line 397: How would that work?

1. Plugins have their own domain, but cannot access picard domains, this may be a problem for constants/attributes from database
2. Plugins have their own domain, a set of methods from plugin API to use it (_(), N_() are those from Picard, P_() and PN_() for plugins)
3. We append plugin domain to the main one (chaining fallbacks https://docs.python.org/3/library/gettext.html#gettext.NullTranslations.add_fallback), basically Picard main domain is used, but plugin "extends" it, standard picard/i18n are used
4. ?

We need to test each approach to find the best.
Loading plugin translations can be either automatic (we expect .mo files to be somewhere identical for all plugins), or by a specific call from the plugin to the API to indicate where to find translations  (decided by plugin author). I tend to prefer the first approach, as it is simple, safer, and "normalize" plugins.

zas (changes requested):

PLUGINS.md
phw: My preference would also be 1. I think plugins should have their own translation generally. For thing like attribute translations we have separate domains anyway. If this becomes a requirement we could consider exposing this via the API as well.

Opiton 3. sounds interesting though, we should probably try how well this works.

zas: Actually using fallbacks isn't a good idea, as it may leads to conflicts between strings (if we search main domain first and plugin has a different translation for the same string, it will not appear, and the reverse searching plugin domain first and falling back on main, would let plugin translations to override main translations).

Also `.mo` aren't supposed to be portable between systems.
Though GNU gettext is able to take care of endianess at runtime it seems, it is usually preferable to compile .po to .mo on target system.
An interesting thread about that: https://stackoverflow.com/questions/53285634/is-there-a-portable-way-to-provide-localization-of-a-package-distributed-on-pypi

zas:
@phw, I rebased onto master and added fixes for imports related to extension_points changes

zas:
@phw : I fixed dependencies in https://github.com/metabrainz/picard/pull/2419/commits/8951ddf4158d63e98a97d811e45ccbfac96200d2 though I'm not sure that's the correct way of doing it for those, please check.

Sophist-UK:
In `PLUGINS.md` you ask whether `toml` is the best format for the manifest file. I have had a quick look and I think that (because it has a lot of similarities with windows ancient-style `ini` files, that it is more easily understood and edited than `json` or `yaml` files. 

I am surprised that you guys have decided to adopt a one-plugin-per-repository model in order to support external 3rd-party (i.e. dubious security) git repos. I know that this can potentially be controlled by limiting the repos to metabrainz owned repos, where you guys can set the security permissions etc., but in practice this will become very unwieldy and become very insecure very quickly. It is also likely that 3rd party repos will have default security (i.e. insecure) which allow merges by any Github user. **I urge you to reconsider this approach**, because I consider it extremely insecure and therefore likely to eventually be used for malicious purposes and become a brand-image issue. (I really really really do NOT want to be saying "Told you so" in a few years (or months) time.)

I see nothing in the pygit2 calling code to verify (by a hash or other means) that the local clone matches the git main source version (and has not been maliciously modified locally).

I see nothing in the code which asks the user explicitly to authorise manually installed and locally modified plugins - the hashes for these should be stored locally and reauthorisation requested if the local source code changes.

I see no changes to the options page for plugins to distinguish between valid (i.e. has matching hashes) centrally downloaded plugins, manually installed plugins, and plugins that were downloaded from a central repo but have since been locally modified. IMO this should be done by having several sections to the plugin list (manually installed and locally modified sections only being shown when needed - so for most users they will continue to see only a single section).

When using git cloning, I am unclear how you tell it to download a release rather than the current commit (which may be a WIP). I am also unclear how the plugins website / API will know when a new release is published.

Future plugin upgrades e.g. from v3 to v4: By allowing external plugin repos you are going to make it MUCH more difficult to migrate to a newer incompatible api. At present you can create a v3 branch in the musicbrainz/picard-plugins repo and edit each plugin to meet the new standard. How will you do this when there are hundreds of separate external repos?

**@zas / @phw Zas / Philipp - I urge you to reconsider this approach.**

zas:
@Sophist-UK We'll continue to support "official" plugins, and there will be a list of them from an official website. Nothing changes here.
We go for one repo per plugin to ease maintenance and contributions, to separate commit history of each plugin, and to allow easy forking.

User will get a warning about installing "third-party" plugins.


> I see no changes to the options page for plugins

Yes, that's why that's a Draft Pull Request, with WIP in the title... we hardly started to implement this.

> I urge you to reconsider this approach.

And what do you suggest exactly?

phw:
@Sophist-UK This PR is all WIP right now. The primary goal of the existing code was to validate and prototype the concepts we had in mind for the new plugin. Also the primary focus so far has been on the plugin structure and API, and ideas about distribution are just coming together.

But having said this, having the ability to allow third-party plugin repositories is definitely one of the goals. We have seen that the single repo controlled by Picard developers doesn't scale. This leads to situations like the classical extras.

We can (and will) still have official repositories. And we can war users about using third party plugins. But in the end the users can already now add plugins from other places. I don't see how the repository system is much different. But it definitely makes it easier for both users and developers.

rdswift:
Sorry I haven't had much chance to go over this yet.

So each (third-party) plugin will have to be in a separate repo, right?  I can work with that, but it means that I'll need to create a bunch of new repos :grinning: and it won't be quite as easy to get a full list of all of my plugins.  Not a big deal, and I'm sure I can work around that somehow.

Am I understanding correctly that the third-party plugins will still be installed the same way (by downloading and installing a zip file), and that active plugins will be checked for updates (by querying the plugin repo) during Picard startup?  Sort of like subscribing to a plugin?

I'm wondering if a potential enhancement might be to allow the user to enter (paste) a url to the plugin repo and Picard could do the download and install?

Another possible enhancement might be to allow the user to subscribe to a plugin provider so they can see all plugins available from that provider (similar to the way that they can see all the plugins available from picard.musicbrainz.org/plugins).

zas:
> I'm wondering if a potential enhancement might be to allow the user to enter (paste) a url to the plugin repo and Picard could do the download and install?

There will be 2 main sources of plugins:

- local directories (no upgrade procedure, what's in the directory is the plugin, it will allow to use no or whatever source control system)
- local and remote git repositories: a local repository accessible on local medias, or a remote repository, installation by cloning, upgrading by updating working tree

There will be 2 ways to add plugins:
- adding a local directory (with .git or not)
- adding a remote repository specified by url & ref

There will be an "official" plugin selector:
- a list is provided via Picard website API, listing url & ref for known plugins
- optionally we can support lists from other sources, but that's not planned yet

The plugin UI will provide means to:
- select a local directory (manual git repo or simple directory)
- select a remote url + ref (manual git repo)
- select a plugin from a list (automatic download, from remote git repo)
- select a different ref for a git installed plugin
- if commits are signed, show an icon indicating the signature was validated.

A way for the user to change the ref used by a git-based plugin: it will let a dev to run its beta version, or a user to select a certain version of the plugin. If ref doesn't match "official" list then it's considered a user-installed plugin, and it will not be upgraded automatically.


Using git will provide:
- hashing (a commit id is a hash, https://alembic.com.au/blog/the-identity-of-a-git-commit-explained)
- optional commit signing (https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits)
- easy upgrade and downgrade
- a way for devs to fork an existing plugin
- a full history of changes via commit messages

Every plugin has 3 mandatory things:
- a MANIFEST file
- a Python module structure (that's at least __init__.py file)
- a directory containing both

So we drop support for single file plugins (because we want to separate metadata from plugin code, hence MANIFEST file).
We also drop support for zipped plugins (nothing prevents to share a plugin as a zipped directory, and unzip it locally), it will simplify plugin code.


That's more or less our thoughts about this at this point. But things can evolved along the way, that's work in progress.

Sophist-UK:
@phw Philipp - so not so much WIP as an experiment? Nothing wrong with that, but it does mean different things than WIP.

I don't have any issues for e.g. a well supported plugin by a reputable known developer having a separate source (and Classical Extras almost certainly fits this profile) - and we can provide guidelines to ensure that the repo is secure. 

But that is wildly different from A) switching to Git without ensuring that it will be secure; and B) insisting that every plugin has its own repo (which IMO might be unmanageable in practice - there isn't a huge influx of new plugins, so more of a big one-off change than an ongoing overhead, but still more likely to have human errors in security settings, and any churn on the change approval team will be a big overhead too).

Also, git does NOT have a concept of releases (and no concept of beta / full releases either) - these are Github enhancements. Git does have a concept of Tags which you could use (if you insist on being aligned to semantic-versioning).

I would suggest that we have 5 tiers of plugins:

1. Primary `picard-plugins` repo, v3 branch.
2. Primary external repos - for special cases like Classical Extras which are somewhere between 1. and 3. in terms of MusicBrainz brand support.
3. Secondary `picard-community-plugins`, v3 branch - just like 1. except the requirements for plugins to be added and updated are less stringent and less supported (i.e. the older, less supported, less frequently used existing plugins that the core Picard team doesn't want to support). (Classical Extras might fit here or might not.)

All the above would be in the Picard plugin catalogue.

4. Other external repos - specified by git URL **including a commit hash** - you cannot add a repo without specifying a hash. Picard provides upgrade support by the user supplying a new URL with a different commit hash.
5. Manually installed plugins.

I have no real issue with using a git clone as a means of downloading so long as there is good central and local security control for any plugins in the catalogue and good local security controls and warnings for any plugins not in the catalogue.

zas:
Supporting one repo with multiple plugins isn't excluded. But it shouldn't be mandatory either.

So we could support 2 structures:
topdir/MANIFEST and topdir/plugin/MANIFEST

For plugin "release" version, we can rely on version from MANIFEST: we update the plugin code only when this version is increased. Then we need to separate code repository from code of the installed plugin. Also what happens when code was changed but version not updated, or the reverse. It would be much more reliable to use git tags as "releases" imho.
But then what about non-git directories containing plugins...
Not clear yet.

Repos are resynced on regular basis (automatically or manually). But installing a new version only happens if the version from MANIFEST changed.

Actually it would be convenient to manage lists of plugins in a git repo too, so we can easily update it without redeploying Picard website.

Each listed plugin has simple metadata: a list of plugin UUIDs, a git URL with a ref (let's called this "Official Plugin List")

Each plugin has a MANIFEST listing a version, an UUID (matching the list above), and the rest of metadata (name, description, etc)

On first sync, Picard clones official repo containing "Official Plugin List", loads it, and for each of entries, clone matching repositories locally (let's call that "Installable plugin repos").

Then it looks for all MANIFEST files in "Installable plugin repos/repo_name/" or  "Installable plugin repos/repo_name/subdir".
repo_name has to be unique, it is based on URL + ref.
If MANIFEST contains one UUID that's listed in "Official Plugin List" then it extracts plugin metadata (version, name, description, etc...)

Now the user can choose to install one plugin:
- he selects one of the available plugins
- the plugin code is copied to "Installed Plugins/UUID" and the module can then be loaded

Now let's explore the evolution over time:

- user uninstalls a plugin: "Installed Plugins/UUID" is totally removed
- use disables a plugin: "Installed Plugins/UUID" is left untouched, but the plugin is marked as disabled, and the module is not loaded (or unloaded if it was)
- the plugin disappears from "Official Plugin List" (=its UUID doesn't appear anymore): the plugin is left untouched but a warning informs the user, and asks if he wants to keep it, disable it, uninstall it
- the plugin is moved to another repo (UUID is preserved, but URL + ref change): the plugin is left untouched but a warning informs the user, and asks if he wants to keep it, disable it, uninstall it, or resync from new repo
- a new plugin is added to "Official Plugin List": it just appears as installable (as above)
- user wants to use a different version of the plugin (ie. the current version is bugged, the previous one was working): we go through MANIFEST file history (a command like `git log --source -S 'version = ' --patch -- MANIFEST.toml` could detect changes of versions) to list previous "released" versions, or let the user selects a specific ref
- If the "Installed Plugin" is set to a different ref, upgrade is disabled until manually re-enabled.

Installing a plugin is basically: cloning to a temp dir (in the destination), with a certain revision, removing .git directory, renaming temp dir to final dir, unloading/reloading the plugin if it was loaded.


I really think we need an UUID for each plugin to handle changes of sources (author moves the git repo elsewhere, but the plugin is still the same). We could use an "unique name" (org.metabrainz.picard.plugins.pluginA) instead of an UUID perhaps, not sure about this.

Still many questions.

Sophist-UK:
1. We probably cannot use Git Clone for downloading plugins which are in a multiple-plugin repo because cloning is (AFAIK) an all-or-nothing thing at a particular commit. The current download of ZIP files seems to me to work pretty well, but what do I know of the difficulties you guys face on a day-to-day basis, but if not this, then what?

2. I agree that we probably need to handle moves from one repository to another. I am not sure that any GUIDs are the way to go because they might introduce a security vulnerability by someone spoofing it. The only means I can see is to retain a global catalogue of MB-supported and community-supported plugins - and Picard has to consult that directory to get the list of the plugins, their versions, their hashes and their download locations.

3. However we do need to ensure that each plugin has a unique directory - and I don't see any security issues with using a GUID or (better) a unique directory name per plugin (and we should probably insist that the unique name is **never** changed).

zas:
Few notes after discussion we had with @outsidecontext

First steps:

- command line plugin management: `picard plugin <subcmd>`
- initial subcommands: install/uninstall/enable/disable/info
- info -> show author, name, short description

List of plugins available from a known (team-managed) repository:
- official list of plugins as toml with:
- url (git/https), category
- category: official team-supported, official non-team supported, third-party, blacklisted
- url is splitted internally as domain + path (https://github.com/user/path or git://github.com:user/path -> domain=github.com, path=user/path)

After, UI, from scratch:
- 1 dialog: installed plugins, enable/disable, settings + button "Install new plugins" / search (freetext)
- 1 dialog: new plugins, showing name,author,short description + install / search (freetext)
- 1 dialog for each plugin: showing info, uninstall, optional settings (plugin-defined)

Picard will not install a blacklisted plugin, match domain / path

rdswift:
Just so I understand correctly, the toml file will be retrieved from the Picard website each time Picard is started, and it will also contain the list of blacklisted sites?

If each plugin is going to be contained in a separate repo, does that mean MetaBrainz will be adding a bunch of new repos for the "team supported" plugins?

phw:
> Just so I understand correctly, the toml file will be retrieved from the Picard website each time Picard is started, and it will also contain the list of blacklisted sites?

Yes, in order to have the blocked plugins effectively disabled in case of security issues, especially those by plugins done in bad faith, the list needs to be loaded early. However, I think we need to inform the user about this. But generally we should have some consent for performing update checks as well.

> If each plugin is going to be contained in a separate repo, does that mean MetaBrainz will be adding a bunch of new repos for the "team supported" plugins?

The idea is to allow multiple Plugins in one repository as wellence the "path" in zas' description above. Authors can use this to have multiple smaller Plugins in a single repo. We would use that for the official plugins as well.

PLUGINS.md
```diff
@@ -0,0 +1,476 @@
...
+This is an example plugin showcasing the new **Picard 3** plugin API.
+
+You can use [Markdown](https://daringfireball.net/projects/markdown/) for formatting."""
+de = """
+Dies ist ein Beispiel-Plugin, das die neue **Picard 3** Plugin-API vorstellt.
+
+Du kannst [Markdown](https://daringfireball.net/projects/markdown/) für die Formatierung verwenden.
+"""
+fr = """
+Ceci est un exemple de plugin présentant la nouvelle API de plugin **Picard 3**.
```
zas, line 197:
```suggestion
Ceci est un exemple de greffon démontrant la nouvelle API de plugin **Picard 3**.
```

PLUGINS.md
```diff
@@ -0,0 +1,476 @@
...
+In order to both simplify third-party development of plugins and distribute
+the maintenance work there will no longer be a single plugin repository. Instead
+each plugin SHOULD be provided in a separate git repository. Plugins also CAN
+be installed locally without a git repository by placing the plugin package
+inside the plugin directory.
+
+
+### Repository structure
+
+A Picard plugin repository MUST be a git repository containing exactly one
```
zas, line 351: Not exactly one, as discussed previously, we'll support one ore more plugins per repo for convenience.

test/data/testplugins3/example/MANIFEST.toml
```diff
@@ -0,0 +1,12 @@
...
+version     = "1.0.0"
+api         = ["3.0", "3.1"]
+license     = "CC0-1.0"
+license-url = "https://creativecommons.org/publicdomain/zero/1.0/"
+user-guide-url = "https://example.com/"
+
+[description]
+en = "This is an example plugin"
+de = "Dies ist ein Beispiel-Plugin"
+fr = "Ceci est un exemple de plugin"
```
zas, line 12:
```suggestion
fr = "Ceci est un exemple de greffon"
```

PLUGINS.md
rdswift: > My preference would also be 1. I think plugins should have their own translation generally. For thing like attribute translations we have separate domains anyway. If this becomes a requirement we could consider exposing this via the API as well.

If I understand correctly, this should accommodate translations for any new options pages and such associated with the plugin, but I'm not clear as to whether this covers translations for text in pop-up dialog boxes such as found in the Submit ISRCs plugin (but I think it might).

Where this **might** become an issue is descriptions of functions added by a plugin (will they show up properly in the help system), and possibly variable descriptions if we extend the help system to allow including variables created by plugins (as discussed in https://github.com/metabrainz/picard/pull/2649).

If I understand the description in https://github.com/phw/picard/blob/plugins-v3/PLUGINS.md the translations will be in the MANIFEST toml file.  Won't this make it a bit clumsy to use something like Weblate to provide the translations?  This is probably a minor inconvenience, but I'm not really seeing a clean path to crowd sourcing translations for plugins.  Then again, should that really be a Picard issue?  Probably not.

I still don't have a great understanding of the whole translation topic, but I wouldn't be at all surprised if the level of integration with separate plugin translations could become a problem.

knguyen1:
Implemented 3 outstanding TODO's: https://github.com/phw/picard/pull/24

knguyen1:
Localisation support and multi-plugin repo documentation and tests: https://github.com/phw/picard/pull/25

knguyen1:
UI to manage plugins: https://github.com/phw/picard/pull/26 test with: https://github.com/knguyen1/every-cover-art-is-homer-simpsons

image

PR #2420, opened 2024-04-20
Build with Qt 6.7.0 on windows and macos

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Build with PyQt 6.6.1 and Qt 6.6.3 on both Windows and macOS.

Previously there have been issues running 6.6 on macOS, hence the version was pinned to 6.5.3 there. This change attempts to unify the version again between the plaforms.


phw:
I don't really understand why the tests are segfaulting on macOS. I can't reproduce this here at all.

phw:
I'm closing this as there are currently too many issues with Qt 6.7 overall. We will get back to this, but likely only after we have moved to PySide6 and there is at least a 6.7.1 release available on pypi..

PR #2421, opened 2024-04-21
WIP: Get rid of gettext builtins

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


zas:
_ is missing from generated ui files, not sure about how to handle it

rdswift:
> _ is missing from generated ui files, not sure about how to handle it

Not the greatest solution, but could you perhaps include the necessary in the comment text added when the ui is built using the `setup.py` script?

https://github.com/metabrainz/picard/blob/08588465639bc0136a0310816d69cf53ac932391/setup.py#L380-L394

Another "hack" might be to add the necessary include if it is missing when building the application.

Unfortunately, neither of these will help if the ui was generated without `setup.py` when running locally from source.

rdswift (dismissed):
I believe that this is a real improvement.  Thanks for all the refactoring changes to implement it.  This looks great to me.

zas:
@phw as discussed, `from picard.i18n import gettext as _` is now possible, but I'm not sure about this move, as we use `_()` everywhere, I don't think it adds anything (but perhaps let use _ for something else, though that will be confusing).
See https://github.com/metabrainz/picard/pull/2421/commits/7db3d007cb3ce4c270ebbca3b9c6f10a1063a2fb (revert is possible)

rdswift:
> @rdswift what's your opinion on this?

I don't think it hurts anything by having the extra flexibility, but as you say it could be confusing.  I do think this makes it cleaner though.

zas:
> I don't think it hurts anything by having the extra flexibility, but as you say it could be confusing to use `_` for something else. I do think this makes it cleaner though.

It was more about https://github.com/metabrainz/picard/commit/7db3d007cb3ce4c270ebbca3b9c6f10a1063a2fb which spreads `from picard.i18n import gettext as _` to replace equivalent `from picard.i18n import _`.
As said, `picard.i18n.gettext` is here to stay anyway (though it could conflict with module `gettext`, as it does in `i18n.py`, hence `import gettext as module_gettext`).

rdswift:
> It was more about [7db3d00](https://github.com/metabrainz/picard/commit/7db3d007cb3ce4c270ebbca3b9c6f10a1063a2fb) which spreads `from picard.i18n import gettext as _` to replace equivalent `from picard.i18n import _`.

Sorry, I misunderstood.  I don't really have any issues with it either way.

setup.py
```diff
@@ -369,29 +370,47 @@ def finalize_options(self):
...
+            command = _get_option_name(self)
+            new_header = f"""
+# Automatically generated - do not edit.
+# Use `python setup.py {command}` to update it.
+
+from PyQt6 import (
+    QtCore,
+    QtGui,
+    QtWidgets,
+)
```
phw, line 404: If I understand correctly this now hard codes the imports for Qt to the above three, while previously the import would be kept as generated. There is a potential this will break if a UI file would cause other imports for some reason. Not sure if that's a thing, so.

phw (approved):
Huge changelist, but looks good.

Let's also update the translations once this got merged.

setup.py
zas: Yes, I thought about it, but for now, only those 3 imports are done everywhere, so that's not an issue. Though the change in the generated comment actually caused an issue (fixed in this PR), so I guess we just need to adapt if something arises, at least any change will be easy to spot since it will break quite obviously. Having imports ordered isn't really important, since those are ignored due to https://github.com/metabrainz/picard/blob/ebfd011a9981ec23845fb5213a2b70782b415a3e/pyproject.toml#L18
I'd say if any issue arises, it will be easy to fix anyway.

PR #2424, opened 2024-04-22
PICARD-2871: Get rid of `QtCore.QObject.tagger`

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

Using `QtCore.QObject.tagger` is a bit problematic and hacky.
It was kept around for a long time, but it is now time to get rid of it.

* JIRA ticket (_optional_): [PICARD-2871](https://tickets.metabrainz.org/browse/PICARD-2871)

# Solution

Use `QtCore.QCoreApplication.instance()` as Tagger class inherit from it anyway.

# Action

Additional actions required:

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -451,7 +451,9 @@ persist = None
 profiles = None
 
 
-def setup_config(app, filename=None):
+def setup_config(app=None, filename=None):
+    if app is None:
+        app = QtCore.QCoreApplication.instance()
     global config, setting, persist, profiles
     if filename is None:
         config = Config.from_app(app)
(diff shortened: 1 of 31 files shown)


picard/config.py
```diff
@@ -451,7 +451,8 @@ def convert(self, value):
 profiles = None
 
 
-def setup_config(app, filename=None):
+def setup_config(filename=None):
+    app = QtCore.QCoreApplication.instance()
```
phw, line 455: I think this is one place where we could better keep the parameter. This function is straight forward be supposed to be called from the instantiated app, where this variable is available.

zas: See last changes in https://github.com/metabrainz/picard/pull/2424/commits/afa49dae4fab1e49cd3b51c4825f62e293a716a0

phw (approved):
LGTM

PR #2425, opened 2024-04-22
Fixed calls to QFileDialog options

phw:
The code is invalid in PyQt6 and also just was supposed to set the standard options being set anyway.

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Invalid call to get default QFileDialog options. This is unneeded anyway, as the dialog will use the default options regardless.


zas (approved):
Tested, LGTM

PR #2426, opened 2024-04-22
metainfo: fix recommended display length definition

Lctrs:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Currently, flathub's build are failing as they don't allow releasing builds with invalid appdata (even if it's just a warning). Running `appstreamcli validate` do raise this warning: `W: org.musicbrainz.Picard:318: relation-display-length-value-invalid medium` currently.

According to https://www.freedesktop.org/software/appstream/docs/chap-Metadata.html#tag-relations-display_length, display length's value must be an integer value corresponding to logical pixels.

# Solution

I took `768`px as value because it corresponds to the recommended referenced size for laptops and tablets, which should match imho the previous incorrect `medium` value.

# Action

Additional actions required:

### DIFF
--- a/org.musicbrainz.Picard.appdata.xml.in
+++ b/org.musicbrainz.Picard.appdata.xml.in
@@ -76,7 +76,7 @@
   <recommends>
     <control>keyboard</control>
     <control>pointing</control>
-    <display_length compare="ge">medium</display_length>
+    <display_length compare="ge">768</display_length>
     <internet>always</internet>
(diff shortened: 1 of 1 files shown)


phw (approved):
Thanks a lot, looks good.

"medium" must have been an allowed value at some time in the past.

PR #2427, opened 2024-04-23
PICARD-2873: Fetch Search dialog cover art images after it was resized

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

In Album Search Dialog, scrolling table causes missing cover art images to be fetched, but it doesn't work if the dialog is resized and those cells made visible.
No signal is associated with this event, and therefore no code leading to fetching missing images is executed when it happens

* JIRA ticket (_optional_): PICARD-2873

# Solution

- create a new `resized` signal emitted every second
- call `AlbumSearchDialog.fetch_coverarts()` when it happens

# Action

Additional actions required:

### DIFF
--- a/picard/ui/tablebaseddialog.py
+++ b/picard/ui/tablebaseddialog.py
@@ -50,9 +50,9 @@ from picard.ui.theme import theme
 
 class ResultTable(QtWidgets.QTableWidget):
 
-    def __init__(self, parent):
-        super().__init__(parent)
-        self._parent_dialog = parent
+    def __init__(self, parent=None, parent_dialog=None):
+        super().__init__(parent=parent)
+        self.parent_dialog = parent_dialog
         self.setSelectionMode(QtWidgets.QAbstractItemView.SelectionMode.ExtendedSelection)
         self.setSelectionBehavior(QtWidgets.QAbstractItemView.SelectionBehavior.SelectRows)
         self.setEditTriggers(QtWidgets.QAbstractItemView.EditTrigger.NoEditTriggers)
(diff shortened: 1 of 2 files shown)


picard/ui/tablebaseddialog.py
```diff
@@ -52,6 +52,7 @@ class ResultTable(QtWidgets.QTableWidget):
 
     def __init__(self, parent):
         super().__init__(parent)
+        self._parent_dialog = parent
```
phw, line 55: Parent is available via `self.parent()`, so does not need to be stored separately.

picard/ui/tablebaseddialog.py
```diff
@@ -73,6 +74,14 @@ def prepare(self, headers):
         self.setRowCount(0)
         self.setSortingEnabled(False)
 
+    @throttle(1000)  # only emit resized signal once per second
+    def emit_resized(self):
+        self._parent_dialog.resized.emit()
+
+    def resizeEvent(self, event):
```
phw, line 81: Wouldn't it be easier to implement the `resizeEvent` on the `AlbumSearchDialog` directly?

phw (changes requested):

picard/ui/tablebaseddialog.py
zas: They differ, `self.parent()` is the parent widget, not the parent dialog

picard/ui/tablebaseddialog.py
zas: Well, since there was a `scrolled` signal already, I made both here since they are pretty similar (at it works as is).

phw (approved):

PR #2428, opened 2024-04-23
Only register sys.excepthook if the app actually started

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Without this patch, when running tests the crash message box may be displayed. 
The issue appeared in https://github.com/metabrainz/picard/commit/d796b1ec295b9c2030510e83e3c1a9db4f9192fd

# Solution

With those changes, it will be displayed only if the app was actually started.

# Action

Additional actions required:

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -132,10 +132,10 @@ def crash_handler(exc: Exception = None):
     app.quit()
 
 
-def _global_exception_handler(exctype, value, traceback):
-    from picard import crash_handler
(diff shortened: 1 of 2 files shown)


phw (approved):
Good move, LGTM

Now we only need.to figure out why the crash dialog sometimes shows up twice. I had this happening, too.

PR #2433, opened 2024-04-25
Move option declarations to picard/options.py and introduce const.defaults

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


picard/options.py
```diff
@@ -0,0 +1,340 @@
...
+    DEFAULT_TOOLBAR_LAYOUT,
+    DEFAULT_TOP_TAGS,
+    DEFAULT_WIN_COMPAT_REPLACEMENTS,
+)
+from picard.i18n import N_
+
+from picard.ui.colors import InterfaceColors
+
+
+class PicardOptions:
```
phw, line 44: Why the class? Wouldn't it be enought to declare all options on top level here?

picard/ui/options/advanced.py
```diff
@@ -49,16 +42,6 @@ class AdvancedOptionsPage(OptionsPage):
     HELP_URL = "/config/options_advanced.html"
 
     options = [
```
phw, line 44: Can the options array be removed completely then?

phw (commented):
Only a few quick comments, but this looks promising.

picard/options.py
zas: Because if we import without doing anything with it linters will complain, but that's not definitive, WIP

rdswift (commented):
I really like this.  I expect it will simplify maintenance significantly by moving all of the options into one place (and the same with the defaults).  I haven't thought this through, but I wonder if there would be any (future) value in having the 'setting' and 'persist' options in separate lists?

zas:
This move will break per-page Restore Defaults:
https://github.com/metabrainz/picard/blob/92ca2d86641f409562569dd2d259eb91fbdf28e3/picard/ui/options/__init__.py#L82-L86

zas:
> I really like this. I expect it will simplify maintenance significantly by moving all of the options into one place (and the same with the defaults). I haven't thought this through, but I wonder if there would be any (future) value in having the 'setting' and 'persist' options in separate lists?

Yes, I thought about it too.
For now, that's very experimental, and few issues will arise for sure.

zas:
@rdswift @phw please look at last changes, I think we are on the right track with the introduction of `OptionPage.register_setting()`

- we need to link pages (where user can edit options) and options somehow
- we already add a link for profile highlights but it was very specific
- we need to be able to restore page defaults, it only makes sense for settings (not persist).
- we want minimal work in Option Pages when adding an option
- an option doesn't have to be editable via Option Pages, not it has to be declared there
- all options are in a central place (`options.py`) 

There is more work to do, I'm not happy with `UserProfileGroups` class, and we can simplify code even more with this move, more to come.

Please give feedback along.

picard/options.py
zas: Addressed in https://github.com/metabrainz/picard/pull/2433/commits/9bf11ba0e0988acacfb03868fc5c09840bfe497f

picard/const/defaults.py
```diff
@@ -0,0 +1,127 @@
...
+            if root_volume:
+                if path.startswith("/"):
+                    path = path[1:]
+                path = os.path.join(root_volume, path)
+        return path
+
+    DEFAULT_CURRENT_BROWSER_PATH = _macos_extend_root_volume_path(DEFAULT_CURRENT_BROWSER_PATH)
+
+# Default query limit
+DEFAULT_QUERY_LIMIT = 50
```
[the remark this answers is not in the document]
zas: Yes, actually they are few more in `picard/const/__init__.py` that should move here, but it will come later.

picard/ui/options/__init__.py
```diff
@@ -125,6 +129,16 @@ def live_checker(text):
 
         regex_edit.textChanged.connect(live_checker)
 
+    def register_setting(self, name, highlights=None):
```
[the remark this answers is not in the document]
zas: Basically, each option "editor" (an option page) registers which settings it edits, but it doesn't declare options anymore.

picard/ui/options/dialog.py
```diff
@@ -206,21 +200,13 @@ def load_all_pages(self):
...
-        return None
-
     def show_attached_profiles_dialog(self):
         window_title = _("Profiles Attached to Options")
         items = self.ui.pages_tree.selectedItems()
         if not items:
             return
         page = self.item_to_page[items[0]]
-        option_group = self.page_option_group(page)
+        option_group = UserProfileGroups.group_from_page(page)
```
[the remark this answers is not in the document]
zas: Yes, but I have the feeling `UserProfileGroups` can be totally removed and just use module level methods:

```python
def user_profile_group_from_page(page):
```

Instead of:
```python
class UserProfileGroups:

    @classmethod
    def group_from_page(cls, page):
```

This way we can only import methods we actually use:

```python
from picard.profile import user_profile_group_from_page
```

Current way is more or less equivalent to:

```python
from picard.profile import *
```



What do you think?

picard/ui/theme.py
```diff
@@ -54,7 +55,7 @@
 
 # Those are values stored in config file:
 class UiTheme(Enum):
-    DEFAULT = 'default'
+    DEFAULT = DEFAULT_THEME_NAME
```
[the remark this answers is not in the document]
zas: Yep, actually it was like this before, not sure why I changed it, fixed in https://github.com/metabrainz/picard/pull/2433/commits/baf516fa144ef377100478274e6445c26acf320e

picard/ui/options/dialog.py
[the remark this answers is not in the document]
phw: Yes, I agree that's the case and would provide the nicer API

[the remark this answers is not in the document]
zas: See https://github.com/metabrainz/picard/pull/2433/commits/34b046fe31b55418352044b45c691ae6f53f3d78

phw (approved):

rdswift:
> @rdswift does it break loading old profiles??

Just checked it here, and it seems to work perfectly.  Well done!

rdswift (approved):
This looks great to me.  A very elegant way of handling options and option profiles that should simplify maintenance immensely.

One thing that I suggest is that you consider adding a comment near the top of `picard/options.py` to explain the process of adding an option, to help make it clear what is required (since it also involves registering the setting in the appropriate options page).  I'll try to come up with something.

Another thing that just occurred to me...  How will this impact an option that is added as part of a plugin?  Obviously the plugin won't be modifying `picard/options.py`, so in that case would it still be defined in the plugin (similar to the way it was before, but directly rather than in the `options` array) with the addition of the call to `self.register_setting()`?

phw (approved):

picard/options.py
```diff
@@ -0,0 +1,350 @@
...
+    DEFAULT_TOP_TAGS,
+    DEFAULT_WIN_COMPAT_REPLACEMENTS,
+)
+from picard.i18n import N_
+
+from picard.ui.colors import InterfaceColors
+
+
+# picard/coverart/providers/caa.py
+# Cover Art Archive Cover Art Archive: Release
```
rdswift, line 45: Possible wording for instructions.

```suggestion
from picard.ui.colors import InterfaceColors


# Note: There are two steps to adding an option to an option page:
#
#   1. The option is added to the appropriate section below.
#
#   2. The option is registered in the __init__ method of the option
#      page with a call to the page's `register_setting()` method.

# picard/coverart/providers/caa.py
# Cover Art Archive Cover Art Archive: Release
```

zas: On the second point, I would rephrase it a bit since that's an optional step, only if a Option in section 'setting' has a matching way to edit it.

rdswift: Good point.  I've updated my suggestion to try to cover that.

rdswift:
> To me, it should work from the new plugin API only, plugin options being stored in their own section to avoid any conflict, and accessed by new plugin API methods.

Right.  I forgot that the intent was to have the plugin options stored in their own section.  That makes perfect sense.

zas:
> to help make it clear what is required

Actually that's not required.
`register_setting()` is here to associate an UI and the setting options it edits. If not called, Restore Defaults will not work for this option, and eventually the option cannot be overridden in profiles. But that's all, it isn't required, and we don't call it for `persist` options (because those aren't edited directly).

For an option to exist, the only requirement is to call `Option()` or its subclasses somewhere, ideally in `options.py`.

picard/options.py
zas: @rdswift let's work on that after I merge this elephant PR.

rdswift:
> Okay, so no clarification note then I guess.

Oops.  Sorry, I didn't see your note above.

zas:
> > Okay, so no clarification note then I guess.
> 
> Oops. Sorry, I didn't see your note above.

See https://github.com/metabrainz/picard/pull/2435

PR #2436, opened 2024-04-27
Better handling of exceptions while loading option pages

zas:
- without this patch, Picard crashes, and/or the options display gets corrupted
- with this patch, a fake page is loaded instead, displaying the message, but avoiding a crash
- it will be important for pages created by plugins

# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -296,16 +296,15 @@ class OptionsDialog(PicardDialog, SingletonDialog):
             self.highlight_enabled_profile_options(load_settings=True)
 
     def get_working_profile_data(self):
-        profile_page = self.get_page('profiles')
-        if not profile_page.loaded:
-            return
-        working_profiles = profile_page._clean_and_get_all_profiles()
+        working_profiles = self.profile_page._clean_and_get_all_profiles()
         if working_profiles is None:
             working_profiles = []
-        working_settings = profile_page.profile_settings
+        working_settings = self.profile_page.profile_settings
         return working_profiles, working_settings
 
     def highlight_enabled_profile_options(self, load_settings=False):
+        if not self.profile_page.loaded:
+            return
         working_profiles, working_settings = self.get_working_profile_data()
         from picard.ui.colors import interface_colors as colors
         fg_color = colors.get_color('profile_hl_fg')
@@ -361,7 +360,7 @@ class OptionsDialog(PicardDialog, SingletonDialog):
         return self.item_to_page[self.page_to_item[name]]
 
     def page_has_attached_profiles(self, page, enabled_profiles_only=False):
-        if not page.loaded:
+        if not page.loaded or not self.profile_page.loaded:
             return False
         option_group = profile_groups_group_from_page(page)
         if not option_group:
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -321,9 +321,6 @@ class OptionsDialog(PicardDialog, SingletonDialog):
         return working_profiles, working_settings
 
     def highlight_enabled_profile_options(self, load_settings=False):
-        profile_page = self.get_page('profiles')
-        if not profile_page.loaded:
-            return
         working_profiles, working_settings = self.get_working_profile_data()
         from picard.ui.colors import interface_colors as colors
         fg_color = colors.get_color('profile_hl_fg')
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -149,13 +149,9 @@ class OptionsDialog(PicardDialog, SingletonDialog):
     suspend_signals = False
 
     def add_pages(self, parent_pagename, default_pagename, parent_item):
-        def pages():
-            for p in self.pages:
-                if p.PARENT == parent_pagename:
-                    yield p
-
+        pages = (p for p in self.pages if p.PARENT == parent_pagename)
         items = []
-        for page in sorted(pages(), key=lambda p: (p.SORT_ORDER, p.NAME)):
+        for page in sorted(pages, key=lambda p: (p.SORT_ORDER, p.NAME)):
             item = HashableTreeWidgetItem(parent_item)
             if not page.initialized:
                 title = _("%s (error)") % _(page.TITLE)
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -425,13 +425,9 @@ class OptionsDialog(PicardDialog, SingletonDialog):
                 self._show_page_error(page, e)
                 return
 
-        if self.profile_page.loaded:
-            self.profile_page.save()
-
         for page in self.loaded_pages:
             try:
-                if page != self.profile_page:
-                    page.save()
+                page.save()
             except Exception as e:
                 log.exception("Failed saving options page %r", page)
                 self._show_page_error(page, e)
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -148,10 +148,10 @@ class OptionsDialog(PicardDialog, SingletonDialog):
 
     suspend_signals = False
 
-    def add_pages(self, parent, default_page, parent_item):
+    def add_pages(self, parent_pagename, default_pagename, parent_item):
         def pages():
             for p in self.pages:
-                if p.PARENT == parent:
+                if p.PARENT == parent_pagename:
                     yield p
 
         items = []
@@ -168,11 +168,11 @@ class OptionsDialog(PicardDialog, SingletonDialog):
                 profile_groups_order(page.NAME)
             else:
                 item.setFlags(QtCore.Qt.ItemFlag.ItemIsEnabled)
-            self.add_pages(page.NAME, default_page, item)
-            if page.NAME == default_page:
+            self.add_pages(page.NAME, default_pagename, item)
+            if page.NAME == default_pagename:
                 self.default_item = item
             items.append(item)
-        if not self.default_item and not parent:
+        if not self.default_item and not parent_pagename:
             self.default_item = items[0]
 
     def __init__(self, default_page=None, parent=None):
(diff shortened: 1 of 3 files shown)


zas:
To simulate the problem, change the name of an option in register_setting() to a non-existing one.

zas:
image

zas:
Without this patch, Picard crashed due to another exception. The initial exception doesn't cause the crash (because it is catched in the loop creating pages).

```
  File "/home/zas/src/picard/./picard/ui/options/dialog.py", line 158, in __init__
    page = Page(self.ui.pages_stack)
  File "/home/zas/src/picard/./picard/ui/options/advanced.py", line 50, in __init__
    self.register_setting('ignore_regexxx', ['ignore_regex'])
  File "/home/zas/src/picard/./picard/ui/options/__init__.py", line 137, in register_setting
    raise Exception(f"Cannot register setting for non-existing option {name}")
Exception: Cannot register setting for non-existing option ignore_regexxx
D: 17:57:45,061 webservice/ratecontrol.get_delay_to_next_request:122: ('picard.musicbrainz.org', 443): First request
D: 17:57:45,062 webservice/ratecontrol.increment_requests:147: ('picard.musicbrainz.org', 443): Incrementing requests to: 1
D: 17:57:45,214 webservice/ratecontrol.decrement_requests:155: ('picard.musicbrainz.org', 443): Decrementing requests to: 0
D: 17:57:45,216 webservice._handle_reply:543: Received reply for https://picard.musicbrainz.org/api/v2/plugins/ -> HTTP2 200 () 
D: 17:57:45,227 webservice/ratecontrol._out_of_backoff:231: ('picard.musicbrainz.org', 443): oobackoff; delay: 1000ms -> 1000ms; slow start; window size 1.000 -> 2.000
D: 17:57:54,345 tagger.exit:728: Picard stopping
D: 17:57:54,347 util/pipe.stop:279: Stopping pipe
I: 17:57:54,350 browser/browser.stop:134: Stopping the browser integration
D: 17:57:54,606 tagger._pipe_server_finished:403: pipe server stopped
  File "/home/zas/src/picard/./picard/ui/mainwindow.py", line 1302, in show_options
    options_dialog = OptionsDialog.show_instance(page, self)
  File "/home/zas/src/picard/./picard/ui/__init__.py", line 156, in show_instance
  File "/home/zas/src/picard/./picard/ui/__init__.py", line 150, in get_instance
  File "/home/zas/src/picard/./picard/ui/options/dialog.py", line 188, in __init__
    self.maintenance_page = self.get_page('maintenance')
  File "/home/zas/src/picard/./picard/ui/options/dialog.py", line 299, in get_page
    return self.item_to_page[self.page_to_item[name]]
KeyError: 'maintenance'

```

picard/ui/options/dialog.py
```diff
@@ -93,6 +94,30 @@
...
+
+        super().__init__(parent)
+
+        msg = _(
+            "Error while loading option page '%s':\n"
+            "\n"
+            "%s\n"
+            "\n"
+            "Please report this issue."
+        ) % (_(from_cls.TITLE), errmsg)
```
rdswift, line 116: What do you think of including a link to the ticket tracker here, similar to the link under "Help" > "Report a bug..." from the menu bar?

rdswift (dismissed):
This is a nice improvement.

zas:
> This is a nice improvement.

Yes, sorry, I added a bit more code o ensure Picard doesn't crash too easily in case of option page failure.

I simulated raising an exception in `__init__()` and/or `load()` of all options pages, and now Picard doesn't crash anymore (though it makes the whole Options quite useless).

picard/ui/options/dialog.py
zas: image

picard/ui/options/dialog.py
```diff
@@ -93,6 +95,44 @@
...
+        widget = QtWidgets.QLabel()
+        widget.setTextFormat(QtCore.Qt.TextFormat.PlainText)
+        widget.setText(msg)
+        widget.setWordWrap(True)
+        layout.addWidget(widget)
+
+        widget = QtWidgets.QLabel(
+            _('Please report this issue on the <a href="%s">Picard bug tracker</a>.')
+            % PICARD_URLS['bug_tracker']
+        )
```
rdswift, line 128: I know that I initially suggested this link, but now I'm wondering if it might be more appropriate to direct the user to the "Troubleshooting" section of the documentation (`webbrowser2.open('troubleshooting')`) or the "Reporting a Bug" section of that page at https://picard-docs.musicbrainz.org/troubleshooting/troubleshooting.html#reporting-a-bug instead, because that would provide instructions about including logs and such.  I'm okay with whatever you decide, but I thought I would mention the alternative for consideration.

rdswift (approved):
All in all, I like this because it makes the Options page displays extremely resilent.

rdswift:
> I simulated raising an exception in `__init__()` and/or `load()` of all options pages, and now Picard doesn't crash anymore (though it makes the whole Options quite useless).

I tried something similar here as well, and you're right.  :grinning:

picard/ui/options/dialog.py
zas: Done in https://github.com/metabrainz/picard/pull/2436/commits/4346b771528ac5f5a243f3998a4206358faa1e73

zas:
> > I simulated raising an exception in `__init__()` and/or `load()` of all options pages, and now Picard doesn't crash anymore (though it makes the whole Options quite useless).
> 
> I tried something similar here as well, and you're right. 😀

Well, test the new changes, I fixed some issues introduced in previous patch, and make things even  more solid.

zas:
@rdswift sorry I dismissed your review again, but I keep finding weak points.

zas:
image

rdswift:
> sorry I dismissed your review again, but I keep finding weak points.

Not a problem.  I will happily review as many times as necessary if it is to accommodate improvements.  After all, it's you that's doing all the heavy lifting.  :wink:

picard/ui/options/dialog.py
```diff
@@ -149,9 +149,13 @@ class OptionsDialog(PicardDialog, SingletonDialog):
     suspend_signals = False
 
     def add_pages(self, parent, default_page, parent_item):
-        pages = [(p.SORT_ORDER, p.NAME, p) for p in self.pages if p.PARENT == parent]
+        def pages():
+            for p in self.pages:
+                if p.PARENT == parent:
+                    yield p
+
```
phw, line 156: Why the inline function instead of the comprehension?

```python
pages = (p for p in self.pages if p.PARENT == parent)
``` 

Seems more elegant and clear to me.

phw (commented):
This is a very good addition, much cleaner error handling. I tested, and it works well. A minor point for discussion regarding the code.

picard/ui/options/dialog.py
zas: Well, that's a tuple vs a generator, so not totally the same thing.
But in this case we can prefer "elegant" I guess, done in [25964df](https://github.com/metabrainz/picard/pull/2436/commits/25964df99a633cd9601b59ce3b5cee3954040b1e)

zas:
@rdswift @phw please test heavily, I think it is ok to merge now, but as usual better check twice ;)

picard/ui/options/dialog.py
phw: That's a [generator expression](https://peps.python.org/pep-0289/), so both are generators.

rdswift:
@zas, any thoughts on my idea to add a "Copy to clipboard" button (https://github.com/zas/picard/pull/14)?  If we're not going to include something like that, then can you please close the PR as "won't do", and I'll remove the branch from my repo.  Thanks.

zas:
> @zas, any thoughts on my idea to add a "Copy to clipboard" button ([zas#14](https://github.com/zas/picard/pull/14))? If we're not going to include something like that, then can you please close the PR as "won't do", and I'll remove the branch from my repo. Thanks.

I'm not sure it is of any use since one can just select & copy the text (right click is showing a context menu, and CTRL+C is working).

image

If we want to include the page name (or any other information) I'd rather add it to the error message.
I really don't see the need of a button to copy text, but may the need to improve the display or/and error message.

rdswift:
Okay, I'll cancel my PR then.  Thanks.

picard/ui/options/dialog.py
```diff
@@ -93,28 +95,80 @@
...
+        error_widget.setTextFormat(QtCore.Qt.TextFormat.PlainText)
+        error_widget.setText(errmsg)
+        error_widget.setWordWrap(True)
+        error_widget.setFrameShape(QtWidgets.QFrame.Shape.StyledPanel)
+        error_widget.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        error_widget.setLineWidth(1)
+        error_widget.setTextInteractionFlags(
+            QtCore.Qt.TextInteractionFlag.TextSelectableByKeyboard
+            | QtCore.Qt.TextInteractionFlag.TextSelectableByMouse
+        )
```
phw, line 128: I wonder whether it would be good to use a readonly QTextEdit here. This already handles text selection and text copy. See https://github.com/zas/picard/pull/15

picard/ui/options/dialog.py
```diff
@@ -93,28 +95,80 @@
...
+
+        report_bug_widget = QtWidgets.QLabel(
+            _('Please see <a href="%s">Troubleshooting documentation</a>'
+              ' and eventually report a bug.')
+            % PICARD_URLS['troubleshooting']
+        )
```
phw, line 134: Just a side note: I like the way this links to the documentation for error reporting. We should use the very same text in the crash handler dialog.

phw (commented):
In my tests this has been working well. Great work

picard/ui/options/dialog.py
phw: Never mind, I just noticed that the context menu is also available in the text label when the text is selectable. This is nicer, as it adjusts to the text content height

phw (approved):

picard/ui/options/dialog.py
zas: Ah yes, my bad :)

PR #2437, opened 2024-04-29
PICARD-2869: Avoid double crash dialog

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2869

For some exceptions the crash dialog got shown twice. This happens because we already catch exceptions around the call of ` main` in the startup scripts. Exceptions catched there would show the crash dialog. Then the exception got re-raised (intentionally, to trigger normal exception handling, including logging and final exit) and catched by the new excepthook introduced in #2436

# Solution

Run register_excepthook early, don't separately catch exceptions around main.

This prevents the crash dialog to be shown twice for exceptions which are catched by the exception handler already. Early registration is required so that errors early during startup (e.g. import errors) are being catched.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -73,7 +73,6 @@ from picard import (
     PICARD_ORG_NAME,
     acoustid,
     log,
-    register_excepthook,
 )
 from picard.acoustid.manager import AcoustIDManager
 from picard.album import (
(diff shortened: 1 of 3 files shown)


zas (approved):
Tested in various cases, it seems to work well now.

PR #2438, opened 2024-04-29
Cleanup defaults

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/util/macos.py
+++ b/picard/util/macos.py
@@ -1,0 +1,43 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2024 Laurent Monin
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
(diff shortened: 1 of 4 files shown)


zas:
@phw can you test on macos (in case of)?

Also shouldn't [`DEFAULT_STARTING_DIR`](https://github.com/metabrainz/picard/blob/4af62cc88baccebcdc5627d2663717c10b31baec/picard/const/defaults.py#L127) be the same as [`DEFAULT_CURRENT_BROWSER_PATH`](https://github.com/metabrainz/picard/blob/4af62cc88baccebcdc5627d2663717c10b31baec/picard/const/defaults.py#L80-L101) ? Current code makes them different on MacOs.

picard/util/macos.py
```diff
@@ -0,0 +1,43 @@
...
+    try:
+        for entry in os.scandir("/Volumes/"):
+            if entry.is_symlink() and os.path.realpath(entry.path) == "/":
+                return entry.path
+    except OSError:
+        log.warning("Could not detect macOS boot volume", exc_info=True)
+    return None
+
+
+def extend_root_volume_path(path):
```
phw, line 36: We could have basic tests for this function, which run only on macOS. We cannot test for the exact volume path, but:

- Any path passed to the function starting with `/Volumes/` must be returned unchanged
- In any other case the resulting path must:
  - start with `/Volumes/` 
  - end with the original `path`

zas: Tests added in https://github.com/metabrainz/picard/pull/2438/commits/fbab1f1884e6d39f687b4c1055a66e9299555b34

phw (approved):
Tested on macOS, works as before. So let's merge it.

The handling of the default paths should indeed be changed a bit, I'll open a separate PR for this.

PR #2440, opened 2024-04-30
PICARD-2879: macOS: Extend all paths in filebrowser with /Volumes/

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2879

If the user has selected a starting directory in Opions > User interface that is not under the /Volumes/ path, this starting directory is not being used in the file browser on startup.

# Solution

Extend all paths in filebrowser with /Volumes/

This ensures that selected starting directory gets resolved correctly, as the filebrowser only supports paths under /Volumes/, not directly under /. On saving drop the /Volumes/ prefix for paths on the root volume. This gives the nicer shorter paths the user is more familiar with (e.g. `/Users/phw/Music`  instead of the long `/Volumes/System/Users/phw/Music`).

### DIFF
--- a/picard/const/defaults.py
+++ b/picard/const/defaults.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2007, 2014, 2016 Lukáš Lalinský
-# Copyright (C) 2014, 2019-2022 Philipp Wolfer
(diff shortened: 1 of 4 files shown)


zas (approved):
LGTM

PR #2442, opened 2024-04-30
Replace deprecated datetime.utcnow() with datetime.now(datetime.UTC)

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

datetime.utcnow is deprecated since Python 3.12, see https://docs.python.org/3/library/datetime.html#datetime.datetime.utcnow


zas (approved):
LGTM

PR #2447, opened 2024-05-03
PICARD-2884: Show tooltips for more file / track icon types

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2883, PICARD-2884

This enhances the tooltips for file / track icons:

- Show match status icon tooltips also for files, not only for tracks. Fixes PICARD-2883
- Show tooltips for more icons, specifically for unmatched tracks, tracks matched to multiple files and file errors (PICARD-2884)

# Solution

Refactor determining icon and tooltip for file items by combining both in the same function (similar to how the icon / tooltip for the fingerprint column is implemented).

Add additional tooltips.

### DIFF
--- a/picard/ui/itemviews.py
+++ b/picard/ui/itemviews.py
@@ -1088,8 +1088,8 @@ class TrackItem(TreeItem):
             file.item = self
             color = TrackItem.track_colors[file.state]
             bgcolor = get_match_color(file.similarity, TreeItem.base_color)
-            icon = FileItem.decide_file_icon(file)
-            self.setToolTip(MainPanel.TITLE_COLUMN, _(FileItem.decide_file_icon_info(file)))
+            icon, icon_tooltip = FileItem.decide_file_icon_info(file)
+            self.setToolTip(MainPanel.TITLE_COLUMN, icon_tooltip)
             self.takeChildren()
             self.setExpanded(False)
             fingerprint_icon, fingerprint_tooltip = FileItem.decide_fingerprint_icon_info(file)
@@ -1150,8 +1150,9 @@ class FileItem(TreeItem):
 
     def update(self, update_track=True, update_selection=True):
         file = self.obj
-        self.setIcon(MainPanel.TITLE_COLUMN, FileItem.decide_file_icon(file))
-        self.setToolTip(MainPanel.TITLE_COLUMN, _(FileItem.decide_file_icon_info(file)))
+        icon, icon_tooltip = FileItem.decide_file_icon_info(file)
+        self.setIcon(MainPanel.TITLE_COLUMN, icon)
+        self.setToolTip(MainPanel.TITLE_COLUMN, icon_tooltip)
         fingerprint_icon, fingerprint_tooltip = FileItem.decide_fingerprint_icon_info(file)
         self.setToolTip(MainPanel.FINGERPRINT_COLUMN, fingerprint_tooltip)
         self.setIcon(MainPanel.FINGERPRINT_COLUMN, fingerprint_icon)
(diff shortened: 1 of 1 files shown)


picard/ui/itemviews.py
```diff
@@ -1160,47 +1168,41 @@ def update(self, update_track=True, update_selection=True):
...
+                icon = FileItem.icon_saved
+                tooltip = _("Track saved")
             elif file.state == File.PENDING:
-                return FileItem.match_pending_icons[int(file.similarity * 5 + 0.5)]
+                icon = FileItem.match_pending_icons[int(file.similarity * 5 + 0.5)]
+                tooltip = _("Pending")
             else:
-                return FileItem.match_icons[int(file.similarity * 5 + 0.5)]
+                icon = FileItem.match_icons[int(file.similarity * 5 + 0.5)]
+                tooltip = _(FileItem.match_icons_info[int(file.similarity * 5 + 0.5)])
```
zas, line 1199: This calculation is repeated 3 times, maybe we could move it to a separate function.

Also the matching is a bit weird, as a `similarity` of 0 selects the icon 50.

Similarity, index, icon:

```
0.00 0  50
0.05 0  50
0.10 1  60
0.15 1  60
0.20 1  60
0.25 1  60
0.30 2  70
0.35 2  70
0.40 2  70
0.45 2  70
0.50 3  80
0.55 3  80
0.60 3  80
0.65 3  80
0.70 4  90
0.75 4  90
0.80 4  90
0.85 4  90
0.90 5 100
0.95 5 100
1.00 5 100
```

zas (approved):
LGTM, just a comment about repeated code & clarity

picard/ui/itemviews.py
```diff
@@ -1160,47 +1168,46 @@ def update(self, update_track=True, update_selection=True):
...
+                icon = FileItem.icon_error_not_found
+                tooltip = _("File not found")
             elif file.error_type == FileErrorType.NOACCESS:
-                return FileItem.icon_error_no_access
+                icon = FileItem.icon_error_no_access
+                tooltip = _("File permission error")
             else:
-                return FileItem.icon_error
+                icon = FileItem.icon_error
+                tooltip = _("Processing error(s): See the Errors tab in the File Info dialog")
```
rdswift, line 1189: I would be tempted to set the text to "Processing error: See the Errors tab in the File Info dialog", only because I don't know how well "error(s)" translates to other languages.

picard/ui/itemviews.py
phw: @zas I addressed the code duplication in d753c69f690b69cfe32f5f7801cb04f299f5b995

> Also the matching is a bit weird, as a similarity of 0 selects the icon 50.

That's because we have 5 different quality match icons, with ` match-50.png`  being the lowest. The icons are not a direct visualization of the matching score. The score itself has very little meaning in itself anyway, it's mostly useful when being compared with other scores.

I wouldn't change this really. I don't think we would gain much from introducing for example 5 additional icons with even smaller bars. For one those small bars become essentially indistinguishably (at some point even the color does not matter anymore), but also for existing users the icons would just start to appear as worse matches then in previous Picard versions.

picard/ui/itemviews.py
phw: I don't think that has much impact on translation. If that's uncommon in the target language or the target language has no plural forms then they would just not have any special form for this. That's part of the translation.

rdswift: Okay, that makes sense.  Thanks for clarifying.

zas (approved):
LGTM

PR #2449, opened 2024-05-03
PICARD-2886: OAuth2 token revocation

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2886

When logging out Picard so far was only removing the locally stored OAuth2 tokens (access and refresh tokens). This removed the tokens from the local system, but the tokens themselves were still present on the MB server and could potentially be used.

MB's OAuth2 implementation supports token revocation, see https://musicbrainz.org/doc/Development/OAuth2#Revoking_a_token. This PR makes use of this to revoke the tokens on the server.

# Solution

Implement support for calling `POST /oauth2/revoke` in `OAuthManager.revoke_tokens`. Refactor this method to work asynchronously with a callback similar to  the login process.

In Options > General implement a dialog that is shown in case of an error during logout. As there can be network errors other technical issues the user could theoretically be stuck with invalid local credentials without the possibility to logout and login again. The dialog allows the user to cancel, retry or just removing the local credentials without revoking access on the server (this is identical to Picard's old behavior).

image

# Action

Additional actions required:

Please test carefully. After login the "MusicBrainz Picard" application should be visible in your MB profile on https://musicbrainz.org/account/applications . After logout with this patch this entry should be removed again, if no error occurred. With older versions of Picard the entry will stay.

### DIFF
--- a/picard/ui/options/general.py
+++ b/picard/ui/options/general.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006-2007, 2014 Lukáš Lalinský
-# Copyright (C) 2008, 2018-2023 Philipp Wolfer
+# Copyright (C) 2008, 2018-2024 Philipp Wolfer
 # Copyright (C) 2011, 2013 Michael Wiencek
 # Copyright (C) 2011, 2019 Wieland Hoffmann
 # Copyright (C) 2013-2014 Sophist-UK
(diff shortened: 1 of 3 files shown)


picard/ui/options/general.py
```diff
@@ -158,17 +161,36 @@ def update_login_logout(self, error_msg=None):
...
-        self.tagger.mb_logout()
+        self.tagger.mb_logout(self.on_logout_finished)
+
+    def on_logout_finished(self, successful, error_msg=None):
+        if not successful:
+            msg = QtWidgets.QMessageBox(self)
+            msg.setIcon(QtWidgets.QMessageBox.Icon.Warning)
+            msg.setWindowTitle(_("Logout error"))
+            msg.setText(_(
+                "A server error occurred while revoking access to the MusicBrainz server: %s\n\nRemove locally stored credentials anyway?") % error_msg)
```
zas, line 176: Just a minor thing, since some time I started to split those long strings with newlines at newlines:

```python
         msg.setText(_(
                "A server error occurred while revoking access to the MusicBrainz server: %s\n"
                "\n"
                "Remove locally stored credentials anyway?") % error_msg)
```

It is easier to read as it is more or less formatted as it will be written, and it doesn't change anything to the actual string

zas (dismissed):
Good addition, it looks good to me.

picard/ui/options/general.py
phw: That makes sense, looks good. I'll update this PR prior to merging.

phw: Changed as suggested

PR #2450, opened 2024-05-03
Mainwindow actions: move to their own file, and overall cleanup

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


zas:
The main (and big) change in this patch is that actions are no more set as properties of `MainWindow` but rather as values of the a new property called `MainWindow.actions`.

It clearly separates actions from `MainWindow` methods (there are a lot of those already).
It helps to simplify parts of this code.

phw (dismissed):
I haven't run the code, but this definitely looks like a good improvement. Our main classes Tagger and Main window tend to grow over time all kinds of methods :)

My only concern would be that the change from method calls to action names as string make it more likely to have bugs due to misspelling the action, and discovering the valid action names is more difficult. Would it make sense to have an enum to cover at least the built-in action names? Or is this overcomplicating things again?

zas:
> My only concern would be that the change from method calls to action names as string make it more likely to have bugs due to misspelling the action, and discovering the valid action names is more difficult. 

I'm not sure using dict keys over class attributes/methods is actually making any difference:
- if you try to use a non-existing method, you get an error, as if you try to use a non-existing key
- discovering valid action names is even easier, as one can just list all dict keys, it was much more complicated before

That said, I'm not totally happy about what we have now, so I'll continue to work on this a bit.
Any suggestion is welcome.

> Would it make sense to have an enum to cover at least the built-in action names? Or is this overcomplicating things again?

The initial idea was to separate actions from Mainwindow, and to be able to change things more easily from that.
An enum will not be much better than a string key imho, It will require to list them in the enum declaration and to link them to matching methods; also enum will need to be imported where it is used, so basically it makes things more complicated, and I'm not yet convinced it has any significant advantage in our case.

As said above, I will try to improve this anyway, I'm not sure yet about all this. I wanted to reduce the size of MainWindow class, and to clearly separate action methods (those were only called for actions creation anyway).
I also tried to mark local methods private so we have a better view about what needs to be exported or not.

zas:
@phw See https://github.com/metabrainz/picard/pull/2450/commits/909396a45690887e864fc2afbb7611c79cc997b8 for a move to enum, I think it's not too bad, but I had to make quite of lot changes to preserve compatibility (mainly `toolbar_layout` option).

So please test test test.

picard/ui/mainwindow/__init__.py
```diff
@@ -961,111 +541,114 @@ def _update_cd_lookup_button(self):
...
-    def create_menus(self):
+    def _create_menus(self):
+        def add_action(menu, action_id):
+            menu.addAction(self.actions[action_id])
+
         menu = self.menuBar().addMenu(_("&File"))
-        menu.addAction(self.add_directory_action)
-        menu.addAction(self.add_files_action)
+        add_action(menu, MainAction.ADD_DIRECTORY)
+        add_action(menu, MainAction.ADD_FILES)
```
rdswift, line 577: Brainstorming here but I'm wondering if it would work (and possibly shorten the code) to somehow iterate through the `MainAction` attributes rather than have separate calls to `add_action()` for each.  I haven't looked at this in detail, but I suspect that it might require additional (conditional) information on each of the attributes.  Or would that actually complicate things again?

zas: The thing is that iterate will not give the hierarchy of menus. Nothing much to improve here imho.

rdswift: Okay.  As I said, I was just brainstorming.

zas: @rdswift have a look at https://github.com/metabrainz/picard/pull/2450/commits/5429085d7570fa2623c7ae8243a1be5ba2073600
It's based on your idea, what do you think?

zas: > Okay. As I said, I was just brainstorming.

No problem, that's welcome, actually it led me to the code you can see in latest commits. It makes menu building much clearer and hides the complexity in a small helper method that replaces previous one.

rdswift: > It's based on your idea, what do you think?

I really like it.  It really simplifies building the menus because it's sort of "what you see is what you get".  Well done!

zas: I applied the same medicine (more or less) to itemviews context menu in https://github.com/metabrainz/picard/pull/2450/commits/4ec3b91318fdaa63da816d7de43c250065063e00

rdswift (approved):
I'm not sure I've tested all of the menus, but everything that I have checked has worked properly.  This should make maintenance or menu changes easier.  Nice improvement.

picard/ui/mainwindow/__init__.py
```diff
@@ -1574,29 +1170,30 @@ def update_actions(self):
...
+        self.action_enabled(MainAction.SUBMIT_CLUSTER, can_submit)
+        self.action_enabled(MainAction.SUBMIT_FILE_AS_RECORDING, have_files)
+        self.action_enabled(MainAction.SUBMIT_FILE_AS_RELEASE, have_files)
+        self.action_enabled(MainAction.TAGS_FROM_FILENAMES, self._get_selected_or_unmatched_files())
+        self.action_enabled(MainAction.SIMILAR_ITEMS_SEARCH, is_file or is_cluster)
+        self.action_enabled(MainAction.TRACK_SEARCH, is_file)
+        self.action_enabled(MainAction.ALBUM_SEARCH, is_cluster)
+        self.action_enabled(MainAction.ALBUM_OTHER_VERSIONS, is_album)
+
+    def action_enabled(self, action_id, enabled):
```
phw, line 1194: Would `enable_action` be better? `action_enabled` to me sounds like you would use it to query the enabled state of an action.

phw (commented):
Yo are right, the enum works better than expected. I was myself unsure, and after your first response to my question I was more leaning toward keeping the strings. But I'm glad you tried it, I think it works well.

I also like that you marked private mainwindow methods with a leading underscore. We should be more consistent with this, and a) always define internal methods with an underscore and b) avoid accessing underscore methods (or properties) from elsewhere (or if needed remove the underscore).

For this PR I don't see any issues. Code runs well, changes make sense.

picard/ui/mainwindow/__init__.py
zas: Yes, and shorter by one character, done in https://github.com/metabrainz/picard/pull/2450/commits/048c22c28c99fddf1e5921937d180b9a8d65028a

zas:
@phw I made few more adjustements, main change is to move MainAction to a separate file (ui/enums.py) to prevent circular import issues (future enums can be added there too).

rdswift (approved):
This all looks good to me.  Much cleaner and easier to use.

phw (commented):
This is looking good to merge. Small concern on whether we need a config migration, but after that has been discussed LGTM

picard/const/defaults.py
```diff
@@ -101,24 +102,24 @@
...
-    'remove_action',
-    'separator',
-    'cd_lookup_action',
-    'separator',
-    'submit_acoustid_action',
-]
+DEFAULT_TOOLBAR_LAYOUT = (
+    MainAction.ADD_DIRECTORY,
+    MainAction.ADD_FILES,
+    '-',
```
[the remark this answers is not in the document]
zas: We could, but I kept support for both, and, on write, it will use '-' from now.
Since that's quite simple, I don't think we need an upgrade hook (or we can do next time we add an upgrade hook).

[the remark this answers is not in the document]
phw: I see, that works

PR #2451, opened 2024-05-05
PICARD-736: OAuth2 login with local callback URL

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-736

So far Picard supported the OAuth2 login to MusicBrainz only via the out-of-band flow for token exchange, which requires the user to manually copy the code used to exchange the access token.

This PR updates the implementation to use a callback URL on localhost, making use of Picard's built-in web server.

# Solution

The built-in webserver provides a new `/auth` endpoint to be used as a callback URL. If the built-in webserver is running, this is used on localhost as the redirect URI for the OAuth2 authorization URL, and the auth endpoint handles exchanging the provided code for the access token.

The old OOB workflow is used as a fallback in case the browser integration is turned off.

Note: This will also ease transition to the future central MeB OAuth endpoint, but this will need further changes, some of them still up for discussion.

# Action

Additional actions required:

### DIFF
--- a/picard/browser/browser.py
+++ b/picard/browser/browser.py
@@ -5,7 +5,7 @@
 # Copyright (C) 2006-2007, 2011 Lukáš Lalinský
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
-# Copyright (C) 2012-2013, 2018, 2021-2022 Philipp Wolfer
+# Copyright (C) 2012-2013, 2018, 2021-2022, 2024 Philipp Wolfer
 # Copyright (C) 2013, 2018, 2020-2021, 2024 Laurent Monin
 # Copyright (C) 2016 Suhas
 # Copyright (C) 2016-2017 Sambhav Kothari
@@ -24,7 +24,6 @@
 # You should have received a copy of the GNU General Public License
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
-
 
 from http.server import (
     BaseHTTPRequestHandler,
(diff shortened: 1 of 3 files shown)


picard/oauth.py
```diff
@@ -45,10 +45,35 @@
 )
 
 
+OOB_URI = 'urn:ietf:wg:oauth:2.0:oob'
+
+
+class InvalidStateError(Exception):
```
zas, line 51: I would prefix this class name.

```suggestion
class OAuthInvalidStateError(Exception):
```

zas (changes requested):
Apart the comment about `InvalidStateError`, it looks good to me.

picard/oauth.py
phw: I renamed it and rebased

picard/browser/browser.py
```diff
@@ -211,6 +213,26 @@ def _add_release(self, args):
...
+            except OAuthInvalidStateError:
+                self._response(400, 'Invalid "state" parameter.')
+                return
+            to_main(
+                oauth_manager.exchange_authorization_code,
+                authorization_code=args['code'][0],
+                scopes='profile tag rating collection submit_isrc submit_barcode',
+                callback=callback,
+            )
+            self._response(200, "Authentication successful, you can close this window now.", 'text/html')
```
zas, line 232: Since this response is displayed in browser, it might be good to indicate it's Picard. But this can be improved in another PR.

zas (approved):
Tested, LGTM.

picard/browser/browser.py
phw: We could also generally improve the output by Picard a bit. There is not many user visible output in browser by Picard, but I think it would be useful and not much work to show a bit nicer pages with at least some minimal layout :)

PR #2454, opened 2024-05-06
Itemviews refactoring

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


picard/ui/itemviews.py
```diff
@@ -337,64 +451,47 @@ def select_object(self, obj):
                 break
 
 
-def paint_column_icon(painter, rect, icon):
+def paint_column_icon(painter, rect, column):
```
phw, line 454: The history of this function is that it originally was used for painting both the header and the content of the cells, but it was using different icons (gray in the header, gray or colored in the columns). This was later changed to use `TreeItem.setIcon()` in the content cells for performance reason (even though it causes a slight misalignment, but it is much faster than a custom paint).

Anyway, my suggestion would be to move the `paint_column_icon` to `IconColumn.paint_header(painter, rect)`, as it now is tightly couple to the IconColumn class anyway and also is only about painting the header. Maybe also rename `IconColumn.icon` to `IconColumn.header_icon`.

zas: Done in https://github.com/metabrainz/picard/pull/2454/commits/5fb55d30e0fccafbfc58f11dc926a0ae9ed35e65 and https://github.com/metabrainz/picard/pull/2454/commits/bc99b478b8a2715a1db50d78b810a864108c4df9

picard/extension_points/__init__.py
```diff
@@ -0,0 +1,95 @@
+# -*- coding: utf-8 -*-
```
phw, line 1: What about putting this into a separate file `extension_points/item_actions.py`? I fear if we start putting this into `__init__.py` then we will end up with all different extension points being shoved into this file.

zas: Done in https://github.com/metabrainz/picard/pull/2454/commits/e11d2a6e68fd5233032137b054d85969bcd5f12b (and rebased)

zas:
Obsolete PR, most changes were introduced in smaller PRs.

See:
https://github.com/metabrainz/picard/pull/2461
https://github.com/metabrainz/picard/pull/2462
https://github.com/metabrainz/picard/pull/2463
https://github.com/metabrainz/picard/pull/2464
https://github.com/metabrainz/picard/pull/2466

PR #2459, opened 2024-05-08
PICARD-2891: When clustering files attached to a Track use orig_metadata

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2891

When dragging files already matched to tracks on the right pane directly back to the left pane "Clusters" item the files get clustered by the new metadata from the matches releases. This is confusing, as the user likely wants to drag mismatched files back and retry with new clustering.

After clustering the files get their metadata reset to original anyway (because this happens when moving away a file from a track), which makes the cluster result even less understandable. 

Note that using the new metadata is actually wanted when clustering files already on the left pane.

# Solution

During clustering check the parent of the file. If it is a track, then use the original metadata for clustering.

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -6,7 +6,7 @@
 # Copyright (C) 2006-2008, 2011 Lukáš Lalinský
 # Copyright (C) 2008 Hendrik van Antwerpen
 # Copyright (C) 2008 Will
-# Copyright (C) 2010-2011, 2014, 2018-2023 Philipp Wolfer
+# Copyright (C) 2010-2011, 2014, 2018-2024 Philipp Wolfer
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
 # Copyright (C) 2012 Wieland Hoffmann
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

picard/cluster.py
```diff
@@ -315,8 +316,14 @@ def cluster(files):
 
         cluster_list = defaultdict(FileCluster)
         for file in files:
-            artist = file.metadata['albumartist'] or file.metadata['artist']
-            album = file.metadata['album']
+            metadata = file.metadata
+            # If the file is attached to a track we should use the original
+            # metadata for clustering. This is often used by users when moving
+            # mismatched files back from the right pane to the left.
+            if isinstance(file.parent, Track):
```
eharris, line 323: Rather than setting and then re-setting metadata, wouldn't it be better to do an else instead?

phw: I changed it to if-else

PR #2466, opened 2024-05-12
Itemviews: Introduce new classes Columns & Column, simplfify code

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


picard/ui/itemviews/columns.py
```diff
@@ -0,0 +1,222 @@
...
+
+    def __repr__(self):
+        cls_name = self.__class__.__name__
+        return f'{cls_name}.{self.name}'
+
+
+class ColumnSortType(IntEnum):
+    NONE = 0
+    NAT = 1
+    DEFAULT = 2
```
phw, line 67: Just a minor thing, but maybe either move DEFAULT to index 1, because NAT is the more special case and/or rename it from default to TEXT, because it is just normal alphabetical text order

phw (approved):
Good work, this makes the code much more readable and maintainable. Especially with getting rid of all those special cases for certain columns inside the various itemviews functions.

A small question / discussion about the ColumnSortType enum, but otherwise I think this is good to merge.

picard/ui/itemviews/columns.py
zas: Done in last iteration, rebased.
I also added a check for sortkey when sort_type is SORTKEY

phw (approved):

PR #2471, opened 2024-05-14
PICARD-2897: Load supported file formats with unknown / uncommon extension

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2897

Picard currently tries load files based on the file extension. Only if the file has no extension or if loading causes an error it falls back to guess format by file content.

Picard should also guess format by file content if the extension was unknown.

# Solution

- If detection of file format by extension fails, try guessing by content instead of returning `None`
- Added tests for `format.open_`
- Refactor `format.open_` to simplify the code. This also changes `extension_points.formats.ext_to_format`, which now expects the extension with leading dot. This avoid the need to remove the dot in every place where `ext_to_format` got called.

### DIFF
--- a/picard/extension_points/formats.py
+++ b/picard/extension_points/formats.py
@@ -4,7 +4,7 @@
 #
 # Copyright (C) 2006-2008, 2012 Lukáš Lalinský
 # Copyright (C) 2008 Will
-# Copyright (C) 2010, 2014, 2018-2020, 2023 Philipp Wolfer
+# Copyright (C) 2010, 2014, 2018-2020, 2023-2024 Philipp Wolfer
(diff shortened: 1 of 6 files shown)


zas (approved):
Good improvement and simplification.

rdswift:
This may sound strange, but what do you think of having an option setting to change the file extension from an invalid one to a valid one, or add a valid extension if it is missing?  In the case of replacing invalid, it might actually make more sense to simply add the correct extension in case there is a period in the original file name which is not intended to separate out an extension.

PR #2474, opened 2024-05-16
Fixed callback for click signal in CAATypesSelectorDialog

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

The `QListWidget.Clicked` signal expects a slot which accepts a single index parameter. In `CAATypesSelectorDialog` the signal gets connected to a function without parameters. See https://doc.qt.io/qt-6/qabstractitemview.html#clicked

While PyQt does handle this, we should still provide callbacks following the expected signature. I came across this while testing with PySide (which is less forgiving).

I chose to set the parameter to `*argc`  here, as we don't care about the actually passed parameter.

### DIFF
--- a/picard/ui/caa_types_selector.py
+++ b/picard/ui/caa_types_selector.py
@@ -325,7 +325,7 @@ class CAATypesSelectorDialog(PicardDialog):
     def excluded(self):
         return list(self.list_exclude.all_items_data()) or ['none']
 
-    def clear_focus(self, lists):
+    def clear_focus(self, lists, *args):
         for temp_list in lists:
             temp_list.clearSelection()
         self.set_buttons_enabled_state()
--- a/picard/ui/caa_types_selector.py
+++ b/picard/ui/caa_types_selector.py
@@ -190,9 +190,9 @@ class CAATypesSelectorDialog(PicardDialog):
         self.fill_lists(types_include, types_exclude)
 
         # Set triggers when the lists receive the current focus
-        self.list_include.clicked.connect(partial(self.clear_focus, [self.list_ignore, self.list_exclude]))
-        self.list_exclude.clicked.connect(partial(self.clear_focus, [self.list_ignore, self.list_include]))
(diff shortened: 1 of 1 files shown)


picard/ui/caa_types_selector.py
```diff
@@ -325,7 +325,7 @@ def included(self):
     def excluded(self):
         return list(self.list_exclude.all_items_data()) or ['none']
 
-    def clear_focus(self, lists):
+    def clear_focus(self, lists, *argc):
```
zas, line 328: If it's one parameter, why using `*argc` ? (and why `*argc` instead of more common`*args`)

I would go for an even more explicit variable name to match https://doc.qt.io/qt-6/qabstractitemview.html#clicked

```python
def clear_focus(self, lists, qmodelindex):
```

phw: argc is really just a typo, it should be args of course.

But I chose `*args`  because we really don't care about any extra arguments here, and the values passed don't have any further meaning for the function. I first had it as `clear_focus(self, lists, index)` , but it seemed confusing why the `clear_focus` function would have an index parameter and why it needs it. It only makes sense if you know that this is a "clicked" slot.

We could also maybe just rename this method to e.g. `def _on_list_clicked(self, lists, index):` to make things clearer

phw: I did the renaming in 4f15d262e9b937139e398784419533213de01274

zas: > We could also maybe just rename this method to e.g. `def _on_list_clicked(self, lists, index):` to make things clearer

The problem I have with `index` is that's not very clear that's a QModelIndex object as we use it a lot for integer indexes, but no strong opinion about this, especially since in this case this parameter is unused.
But I agree with the function's name, using `_on` (or `on_`) prefix makes it clearer that's a signal handler.

Maybe:

```python
def def _on_list_clicked(self, lists, *args):
```

zas (approved):
LGTM

PR #2492, opened 2024-05-24
PICARD-2912: Script editor syntax highlighting fixes and improvements

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Capture d’écran du 2024-05-24 17-26-14

* JIRA ticket (_optional_): PICARD-2912

# Solution

Capture d’écran du 2024-05-24 19-44-13

# Action

Additional actions required:

### DIFF
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -92,8 +92,11 @@ EXTRA_VARIABLES = (
 
 
 def find_regex_index(regex, text, start=0):
-    match = regex.search(text[start:])
-    return start + match.start() if match else -1
+    m = regex.search(text[start:])
+    if m:
+        return start + m.start()
+    else:
+        return -1
 
 
 class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -107,55 +107,47 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
 
     def __init__(self, document):
         super().__init__(document)
-        syntax_theme = theme.syntax_theme
+        self.syntax_theme = theme.syntax_theme
 
-        self.unknown_func_re = re.compile(r"\$(?!noop)[_a-zA-Z0-9]*\(")
-        self.unknown_func_fmt = QtGui.QTextCharFormat()
-        self.unknown_func_fmt.setFontItalic(True)
-        self.unknown_func_fmt.setForeground(syntax_theme.special)
+        unknown_func_re = re.compile(r"\$(?!noop)[_a-zA-Z0-9]*\(")
+        unknown_func_fmt = QtGui.QTextCharFormat()
+        unknown_func_fmt.setFontItalic(True)
+        unknown_func_fmt.setForeground(self.syntax_theme.special)
 
-        self.func_fmt = QtGui.QTextCharFormat()
-        self.func_fmt.setFontWeight(QtGui.QFont.Weight.Bold)
-        self.func_fmt.setForeground(syntax_theme.func)
+        var_re = re.compile(r"%[_a-zA-Z0-9:]*%")
+        var_fmt = QtGui.QTextCharFormat()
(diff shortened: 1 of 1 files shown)


rdswift (dismissed):

picard/ui/widgets/scripttextedit.py
```diff
@@ -101,44 +101,66 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
...
         self.noop_fmt = QtGui.QTextCharFormat()
         self.noop_fmt.setFontWeight(QtGui.QFont.Weight.Bold)
         self.noop_fmt.setFontItalic(True)
         self.noop_fmt.setForeground(syntax_theme.noop)
-        self.rules = [
-            (self.func_re, self.func_fmt, 0, -1),
+
+        self.rules = list(self.func_rules())
+        self.rules.extend([
+            (self.unknown_func_re, self.unknown_func_fmt, 0, -1),
```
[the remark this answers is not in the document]
zas: @rdswift I went for classes instead of namedtuple finally, check last commits

rdswift (dismissed):
Looks good to me.  Thanks for using a class to clarify the rules items.  That makes it a lot easier to understand.

picard/ui/theme.py
```diff
@@ -87,7 +87,7 @@ def _missing_(cls, value):
     func=QtGui.QColor(255, 87, 160, 255),  # magenta
     var=QtGui.QColor(252, 187, 81, 255),  # orange
     escape=QtGui.QColor(75, 239, 31, 255),  # green
-    special=QtGui.QColor(255, 87, 160, 255),  # blue
+    special=QtGui.QColor(150, 150, 239),  # blue
     noop=QtGui.QColor(4, 231, 213, 255),  # cyan
```
phw, line 91: I'm not entirely sure about this change. I find the blueish color hard to read and don't really see why we need to use a separate color for the parentheses and commas. But if we change it, shouldn't we also change it in the light theme? Maybe I also don't fully understand what this intents to fix.

If it is only about highlighting invalid functions I would suggest to introduce a separate "error" color just for this.

Small side note: For consistency I would either remove the alpha from all values or keep it in this value as well.

phw (changes requested):

phw:
Good move with highlighting and refactoring it with the classes. I only don't quite get the color change, especially as it is in dark mod only.

I also added PICARD-2912 to track the change of highlighting missing functions.

picard/ui/theme.py
zas: Perhaps you missed the point: it has a different color, but in light mode only. I see no reason to have same color in dark mode, while they differ in light mode.

zas: > If it is only about highlighting invalid functions

It was here before, not for invalid functions only.

> I would suggest to introduce a separate "error" color just for this.

Yes, I thought about that too, but first I'd like to make those colors configurable.

phw: > Perhaps you missed the point: it has a different color, but in light mode only.

Not it hasn't, it's `QtCore.Qt.GlobalColor.blue` for both function names and the parentheses / commas. That's why this change confused me.

zas: Ah damn, I was confused by the comment on https://github.com/metabrainz/picard/blob/e4168b53859ab8f4c2b5fa205f292349770a3d3c/picard/ui/theme.py#L90 (vs https://github.com/metabrainz/picard/blob/e4168b53859ab8f4c2b5fa205f292349770a3d3c/picard/ui/theme.py#L87)

Ok, so we have 2 identical colors, I'll revert this change.

zas: Done & rebased

phw: I see. I think that's an example of a comment not following code changes. I likely added the comments first before replacing the colors, and as the light theme uses "blue" here the comment was blue. When changing the actual color to magenta only the comment for the first use was updated. So we should fix the comment here, I did so in f78d02e415d89b608735548d8426e5d69dccb3c9

phw (approved):
LGTM

PR #2494, opened 2024-05-25
match -> match_: match is a statement in Python 3.10+

zas:
It doesn't really conflict with variable name but it confuses syntax highlighters (if they are Python 3.10+ aware ofc)

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/installer/i18n/nshutil.py
+++ b/installer/i18n/nshutil.py
@@ -135,11 +135,11 @@ def unescape_string(text):
 
 
 def parse_langstring(line):
-    match = RE_LANGSTRING_LINE.match(line)
-    if match:
+    match_ = RE_LANGSTRING_LINE.match(line)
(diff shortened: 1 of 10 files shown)


phw (approved):
Let's change this.

But it's a bit annoying, as we so commonly use it and it reads good. Maybe we can come up with better names in future code (but we don't have to overthink this).

rdswift:
> But it's a bit annoying, as we so commonly use it and it reads good. Maybe we can come up with better names in future code (but we don't have to overthink this).

Maybe something like "re_match"?  Descriptive, but shouldn't trigger syntax highlighting.

PR #2496, opened 2024-05-25
Introduce TaggingScriptSetting and few associated methods

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/script/__init__.py
+++ b/picard/script/__init__.py
@@ -136,13 +136,15 @@ def script_function_documentation_all(fmt='markdown', pre='',
     return "\n".join(doc_elements)
 
 
-def enabled_tagger_scripts_texts():
-    """Returns an iterator over the enabled tagger scripts.
-    For each script, you'll get a tuple consisting of the script name and text"""
-    config = get_config()
+def iter_active_tagging_scripts(config=None):
+    """Returns an iterator over the enabled and not empty tagging scripts."""
+    if config is None:
+        config = get_config()
     if not config.setting['enable_tagger_scripts']:
-        return []
-    return [(s.name, s.content) for s in iter_tagging_scripts_from_config(config=config) if s.enabled and s.content]
+        return
+    for script in iter_tagging_scripts_from_config(config=config):
+        if script.enabled and script.content:
+            yield script
 
 
 def get_file_naming_script(settings):
--- a/picard/script/__init__.py
+++ b/picard/script/__init__.py
@@ -63,7 +63,7 @@ from picard.script.parser import (  # noqa: F401 # pylint: disable=unused-import
     ScriptUnknownFunction,
     ScriptVariable,
 )
-from picard.script.serializer import PicardFileNamingScript
+from picard.script.serializer import FileNamingScriptInfo
 
 
 class TaggingScriptSetting:
@@ -173,7 +173,7 @@ def get_file_naming_script_presets():
     """Generator of preset example file naming script objects.
 
     Yields:
-        PicardFileNamingScript: the next example PicardFileNamingScript object
+        FileNamingScriptInfo: the next example FileNamingScriptInfo object
     """
     AUTHOR = "MusicBrainz Picard Development Team"
     DESCRIPTION = _("This preset example file naming script does not require any special settings, tagging scripts or plugins.")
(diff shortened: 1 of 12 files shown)


rdswift (approved):
Looks okay to me.

picard/script/__init__.py
```diff
@@ -109,7 +140,7 @@ def enabled_tagger_scripts_texts():
     config = get_config()
     if not config.setting['enable_tagger_scripts']:
         return []
-    return [(s_name, s_text) for _s_pos, s_name, s_enabled, s_text in config.setting['list_of_scripts'] if s_enabled and s_text]
+    return [(s.name, s.content) for s in iter_tagging_scripts_from_config(config=config) if s.enabled and s.content]
```
phw, line 143: Not directly related to this PR, but the `enabled_tagger_scripts_texts` function could be a generator. It only is used for iterating (as the doc comment already indicates).

picard/ui/itemviews/basetreeview.py
```diff
@@ -424,7 +426,8 @@ def add_actions(*args):
                         action_menu = plugin_menus[key] = action_menu.addMenu(key[-1])
                 action_menu.addAction(action)
 
-        if scripts:
+        if has_tagging_scripts(config=config):
+            scripts = iter_tagging_scripts_from_config(config=config)
```
phw, line 430: Not sure about that. We now have to read `config.setting['list_of_scripts']` twice, and config access is kind of expensive. Doesn't probably matter much in this function, but still.

What about using `iter_tagging_scripts_from_tuples` here instead?

```python
scripts = config.setting['list_of_scripts']
if scripts:
    scripts_menu = ScriptsMenu(iter_tagging_scripts_from_tuples(scripts), _("&Run scripts"), menu)
    ....
```

zas: The idea was to regroup calls to `config.setting['list_of_scripts']` in one place, but I agree that's not optimal, still WIP

zas: Done and rebased

picard/script/__init__.py
zas: I did that in https://github.com/metabrainz/picard/pull/2496/commits/a00454afdf5e241c26210f3101da81412c92637b

I changed name to `iter_active_tagging_scripts()` as it yields enabled and non-empty TaggingScriptSetting objects now.

picard/script/serializer.py
```diff
@@ -57,16 +57,28 @@ class PicardScriptType(IntEnum):
     FILENAMING = 2
 
 
-class ScriptImportExportError(Exception):
+class PicardScriptError(Exception):
```
phw, line 60: Do we really want to prefix all class names with "Picard" in this file? Maybe if there is confusion with other script related classes  it would be better to include the purpose about the classes in this file. Those are all classes to assist with (de)serializing the script.

E.g. instead of `PicardScript` use `ScriptSerializer`. Or if it is more about holding the info / metadata of the script call it `ScriptInfo` (I would avoid `ScriptMetadata` to avoid confusion with other metadata).

The naming here is a bit difficult, but I don't see adding the "Picard" prefix is improving the situation. It doesn't make it clear that these classes hold the script + metadata and allow writing it to yaml files.

zas: It was more a first step to ease renaming. I'm not happy with current names either.

So if we renamed PicardScript to ScriptSerializer (and related exceptions), then PicardTaggingScript to TaggingScriptInfo (and we have TaggingScriptSetting for now), do you think it could work?

phw: Yes, I think that would work. It reads much better IMHO.

zas: Done in latest commits

picard/script/__init__.py
```diff
@@ -103,13 +128,15 @@ def script_function_documentation_all(fmt='markdown', pre='',
...
+    """Returns an iterator over the enabled and not empty tagging scripts."""
+    if config is None:
+        config = get_config()
     if not config.setting['enable_tagger_scripts']:
-        return []
-    return [(s_name, s_text) for _s_pos, s_name, s_enabled, s_text in config.setting['list_of_scripts'] if s_enabled and s_text]
+        return
+    for script in iter_tagging_scripts_from_config(config=config):
+        if script.enabled and script.content:
+            yield script
```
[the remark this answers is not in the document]
zas: Done in https://github.com/metabrainz/picard/pull/2496/commits/b1f327252d6f71f30272427c28c084a26c101efc

phw (approved):

PR #2498, opened 2024-05-26
PICARD-2913: Let user configure syntax highlighting colors

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Script syntax highlighting colors are hardcoded.

* JIRA ticket (_optional_): [PICARD-2913](https://tickets.metabrainz.org/browse/PICARD-2913)

# Solution

Let the user configure them in Options > Interface > Colors

image

# Action

Additional actions required:

### DIFF
--- a/picard/ui/colors.py
+++ b/picard/ui/colors.py
@@ -52,6 +52,13 @@ _COLOR_DESCRIPTIONS = {
     'tagstatus_added': N_("Tag added"),
     'tagstatus_changed': N_("Tag changed"),
     'tagstatus_removed': N_("Tag removed"),
+    'syntax_hl_error': N_("Error syntax highlight"),
+    'syntax_hl_escape': N_("Escape syntax highlight"),
+    'syntax_hl_func': N_("Function syntax highlight"),
+    'syntax_hl_noop': N_("Noop syntax highlight"),
+    'syntax_hl_special': N_("Special syntax highlight"),
+    'syntax_hl_unicode': N_("Unicode syntax highlight"),
+    'syntax_hl_var': N_("Variable syntax highlight"),
 }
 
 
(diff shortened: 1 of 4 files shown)


picard/ui/widgets/scripttextedit.py
```diff
@@ -125,16 +125,15 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
...
-            'noop': HighlightFormat(fg_color=syntax_theme.noop, bold=True, italic=True),
-            'special': HighlightFormat(fg_color=syntax_theme.special),
-            'unicode': HighlightFormat(fg_color=syntax_theme.escape, italic=True),
-            'unknown_func': HighlightFormat(fg_color=syntax_theme.special, italic=True),
-            'var': HighlightFormat(fg_color=syntax_theme.var),
+            'escape': HighlightFormat(fg_color='syntax_hl_escape'),
+            'func': HighlightFormat(fg_color='syntax_hl_func', bold=True),
+            'noop': HighlightFormat(fg_color='syntax_hl_noop', bold=True, italic=True),
+            'special': HighlightFormat(fg_color='syntax_hl_special'),
+            'unicode': HighlightFormat(fg_color='syntax_hl_unicode', italic=True),
```
phw, line 134: Minor thing I noticed during testing: The unicode escapes have become italic in #2492 . This is looking odd, especially due to the backslash getting tilted. Maybe you can change this before merging this PR, or we change it after merging.

phw (approved):
LGTM

picard/ui/colors.py
```diff
@@ -97,6 +104,22 @@ def register_color(themes, name, value):
 register_color(_LIGHT, 'first_cover_hl', 'darkgoldenrod')
 register_color(_DARK, 'first_cover_hl', 'orange')
 
+# syntax highlighting colors
+register_color(_LIGHT, 'syntax_hl_error', 'blue')
```
phw, line 108: Should we give the error a separate color? Maybe red.

picard/ui/widgets/scripttextedit.py
zas: I set it to italic because it was using the same color as 'escape', but now we can have different colors so I guess we can remove it.
I'd better adjust overall style after we merged this, we can define slightly different colors for each case.

picard/ui/colors.py
zas: Yes, but I prefer to do tuning after we merged this. I intentionnally didn't change colors in this PR.

phw: I don't think we need different colors. The cases where the colors are currently the same makes. It's great that the colors can be configured separately, but I don't see why we need to have different colors by default.

PR #2500, opened 2024-05-27
Introduce MetadataItem.suspend_metadata_images_update context manager

zas:
It just enable/disable update_metadata_images_enabled

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/item.py
+++ b/picard/item.py
@@ -30,6 +30,7 @@ from PyQt6 import QtCore
 from picard import log
 from picard.i18n import ngettext
 from picard.metadata import Metadata
+from picard.util import IgnoreUpdatesContext
 
 
 class Item(object):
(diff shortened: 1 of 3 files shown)


picard/item.py
```diff
@@ -164,9 +167,13 @@ def __init__(self):
...
         self.metadata = Metadata()
         self.orig_metadata = Metadata()
-        self.update_metadata_images_enabled = True
+        self.enable_update_metadata_images(True)
         self.update_children_metadata_attrs = {}
         self.iter_children_items_metadata_ignore_attrs = {}
+        self.suspend_metadata_images_update = IgnoreUpdatesContext(
+            on_first_enter=partial(self.enable_update_metadata_images, False),
+            on_last_exit=partial(self.enable_update_metadata_images, True),
+        )
```
phw, line 176: We don't need to function. Instead we could do directly

```IgnoreUpdatesContext
self.update_metadata_images_enabled = IgnoreUpdatesContext()
```

`IgnoreUpdatesContext` is designed to be treated as a boolean to allow this kind of usage. So you can still do:

```python
if self.update_metadata_images_enabled:
   ...
```

phw: Sorry, it is the other way around. We would need to call it `suspend_metadata_images_update` like you did, so we would need to turn around all current if checks (`if not self.suspend_metadata_images_update`  instead of `if self.update_metadata_images_enabled`).

But I still think we should do it :)

zas: It should be done in latest iteration, please check.

phw (approved):
Looks good. I'm glad we make more use of this context class now

PR #2502, opened 2024-05-28
Review and pass proper arguments to parent classes (mainly Qt*)

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/util.py
+++ b/picard/ui/util.py
@@ -88,8 +88,8 @@ class MultiDirsSelectDialog(QtWidgets.QFileDialog):
     Depending on the platform, dialog may fallback on non-native.
     """
 
-    def __init__(self, *args):
-        super().__init__(*args)
+    def __init__(self, parent=None, caption="", directory="", filter=""):
+        super().__init__(parent=parent, caption=caption, directory=directory, filter=filter)
         self.setFileMode(QtWidgets.QFileDialog.FileMode.Directory)
         self.setOption(QtWidgets.QFileDialog.Option.ShowDirsOnly)
         # The native dialog doesn't allow selecting >1 directory
--- a/picard/ui/widgets/scriptdocumentation.py
+++ b/picard/ui/widgets/scriptdocumentation.py
@@ -62,14 +62,19 @@ code {
 class ScriptingDocumentationWidget(QtWidgets.QWidget):
     """Custom widget to display the scripting documentation.
     """
-    def __init__(self, parent, include_link=True, *args, **kwargs):
+    def __init__(self, include_link=True, parent=None):
         """Custom widget to display the scripting documentation.
 
         Args:
-            parent (QWidget): Parent screen to check layoutDirection()
             include_link (bool): Indicates whether the web link should be included
+            parent (QWidget): Parent screen to check layoutDirection()
         """
-        super().__init__(*args, **kwargs)
+        super().__init__(parent=parent)
+
+        if self.layoutDirection() == QtCore.Qt.LayoutDirection.RightToLeft:
+            text_direction = 'rtl'
+        else:
+            text_direction = 'ltr'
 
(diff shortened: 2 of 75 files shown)


phw (approved):

phw (approved):
Look good

picard/ui/util.py
```diff
@@ -81,15 +81,71 @@ def find_starting_directory():
...
+    def __init__(self, parent=None, caption="", directory="", filter=""):
+        if not caption:
+            caption = _("Select a file or a directory")
+        caption = picardize_caption(caption)
+        super().__init__(parent=parent, caption=caption, directory=directory, filter=filter)
+
+    @staticmethod
+    def getSaveFileName(**kwargs):
+        convert_filedialog_params(kwargs, _("Select a target file"))
+        return QtWidgets.QFileDialog.getSaveFileName(**kwargs)
```
phw, line 122: Just something we need to be aware of with this implementation: `FileDialog.getSaveFileName` gives the caller a `QFileDialog`, not a `FileDialog`.

This might become unexpected should we for example replace existing uses of `QFileDialog` with `FileDialog` and add additional default handling to this class.

We currently have one direct instantiation of `QFileDialog` at https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L1204 . Probably this should also be changed to the static helper functions instead?

I think I avoided the helper functions when implementing this because I found the many arguments to a single static function harder to use then just instantiating the class itself and setting properties on it.

Another thing is especially with these functions with many parameters but only using `**kwargs` that I need to lookup the possible parameters and code completion cannot assist me. Currently I would get something like this:

grafik

That would definitely motivate me to avoid the wrapper and use the class directly. Maybe we should just have our very own specific wrappers that open the dialog as we actually need it right now so we can easily use it and have consistent behavior.

Another point is how we dealt with the separate `MultiDirsSelectDialog`. Should we make this a subclass of `FileDialog`? We should also use consistent caption there at least.

picard/ui/util.py
```diff
@@ -81,15 +81,71 @@ def find_starting_directory():
...
+
+    if 'selectedFilter' in kwargs:
+        # TODO: remove with PySide
+        # pyside & qt use `selectedFilter`, pyqt uses `initialFilter`
+        kwargs['initialFilter'] = kwargs['selectedFilter']
+        del kwargs['selectedFilter']
+
+    # Ensure we have a caption in those dialogs
+    if 'caption' not in kwargs or not kwargs['caption']:
+        kwargs['caption'] = default_caption
```
phw, line 104: I think I never noticed this because Gtk does not show titles for file chooser dialogs. I checked and Windows uses the caption. Not sure about macOS, I think it might not have titles as well, but that needs to be checked.

Anyway, consistently setting a title makes sense.

phw (changes requested):
The parent changes look good, much more consistency and solving potential side effects.

But I would move the file chooser changes to a separate PR. I think there is more discussion needed on that side.

picard/ui/util.py
zas: I moved those changes outside this branch, I'll submit a PR after this one is merged.

picard/ui/widgets/scriptdocumentation.py
```diff
@@ -90,7 +90,8 @@ def process_html(html, function):
             postprocessor=process_html,
         )
 
-        if parent.layoutDirection() == QtCore.Qt.LayoutDirection.RightToLeft:
+        # FIXME: A QWidget doesn't always have a parent at this point
```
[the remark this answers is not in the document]
phw: Cool

phw (approved):
LGTM

PR #2503, opened 2024-05-29
Filedialog wrappers

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/script/serializer.py
+++ b/picard/script/serializer.py
@@ -47,6 +47,8 @@ from picard.i18n import (
 )
 from picard.util import make_filename_from_title
 
+from picard.ui.util import FileDialog
+
 
 @unique
 class ScriptSerializerType(IntEnum):
@@ -222,10 +224,10 @@ class ScriptSerializer():
 
         dialog_title = _("Export Script File")
         dialog_file_types = self._get_dialog_filetypes()
-        filename, file_type = QtWidgets.QFileDialog.getSaveFileName(
+        filename, file_type = FileDialog.getSaveFileName(
             parent=parent,
             caption=dialog_title,
-            directory=default_path,
+            dir=default_path,
             filter=dialog_file_types,
         )
         if not filename:
@@ -264,10 +266,10 @@ class ScriptSerializer():
         dialog_title = _("Import Script File")
         dialog_file_types = cls._get_dialog_filetypes()
         default_script_directory = os.path.normpath(QtCore.QStandardPaths.writableLocation(QtCore.QStandardPaths.StandardLocation.DocumentsLocation))
-        filename, file_type = QtWidgets.QFileDialog.getOpenFileName(
+        filename, file_type = FileDialog.getOpenFileName(
             parent=parent,
             caption=dialog_title,
-            directory=default_script_directory,
+            dir=default_script_directory,
             filter=dialog_file_types,
         )
         if not filename:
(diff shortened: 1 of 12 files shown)


zas:
For reference https://github.com/metabrainz/picard/pull/2502#discussion_r1618311678

phw:
Copying my comment from #2502:

Just something we need to be aware of with this implementation: `FileDialog.getSaveFileName` gives the caller a `QFileDialog`, not a `FileDialog`.

This might become unexpected should we for example replace existing uses of `QFileDialog` with `FileDialog` and add additional default handling to this class.

We currently have one direct instantiation of `QFileDialog` at https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L1204 . Probably this should also be changed to the static helper functions instead?

I think I avoided the helper functions when implementing this because I found the many arguments to a single static function harder to use then just instantiating the class itself and setting properties on it.

Another thing is especially with these functions with many parameters but only using `**kwargs` that I need to lookup the possible parameters and code completion cannot assist me. Currently I would get something like this:

grafik

That would definitely motivate me to avoid the wrapper and use the class directly. Maybe we should just have our very own specific wrappers that open the dialog as we actually need it right now so we can easily use it and have consistent behavior.

Another point is how we dealt with the separate `MultiDirsSelectDialog`. Should we make this a subclass of `FileDialog`? We should also use consistent caption there at least.

zas:
> Just something we need to be aware of with this implementation: `FileDialog.getSaveFileName` gives the caller a `QFileDialog`, not a `FileDialog`.

I'm not sure to understand, what do you mean by "gives the caller a `QFileDialog`, not a `FileDialog`." ?



> We currently have one direct instantiation of `QFileDialog` at https://github.com/metabrainz/picard/blob/master/picard/tagger.py#L1204 . Probably this should also be changed to the static helper functions instead?

https://github.com/metabrainz/picard/pull/2503/files#diff-754164b29034aa27e27d501630b16fe491c78108500b10caeab6b272bc240fbdR1205 changes it to use `FileDialog`, the only difference here is that `file_chooser` becomes a `FileDialog`, but that's still an instance of `QFileDialog`.
Again, not sure to understand what's the problem here.



> Another thing is especially with these functions with many parameters but only using `**kwargs` that I need to lookup the possible parameters and code completion cannot assist me.

We could define arguments in static methods instead (in fact, that what I did first), but I went this way because I had issue with `options` default. It doesn't seem `QFileDialog.Options()` works (pyqt?), so I used `QFileDialog.Option(0)`, which provides an empty enum flag as default for options parameter.



> Another point is how we dealt with the separate `MultiDirsSelectDialog`. Should we make this a subclass of `FileDialog`? We should also use consistent caption there at least.

Or just a new static method of `FileDialog`.

zas:
> Or just a new static method of `FileDialog`.

I did that in https://github.com/metabrainz/picard/pull/2503/commits/13dfe589b96784c3321614cbe2b0ab6e913a9810

zas:
> 
> We could define arguments in static methods instead

See https://github.com/metabrainz/picard/pull/2503/commits/672c37eef62d0416bf424000e8a47658cd8ad187

phw (approved):

PR #2506, opened 2024-05-30
PICARD-2926: Add option to filter out images below a given size

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Adds a setting to discard downloaded cover art images if they are below a given width and height.
# Problem

* JIRA ticket (_optional_): PICARD-2926

# Solution

The options page looks like this:
image
I plan to add settings for resizing and converting formats also to this page.

I created an extension point for adding image filters. I'm aware this is bound to change for Picard 3, but I was thinking that I can use this for now, until the new plugin apis are defined, and change it later in the last couple of weeks of gsoc.

I'm still working on adding another extension point for filtering functions that use the image metadata, since outsidecontext suggested I should add a way for providers to filter images before actually downloading them, if they can.

# Action

Additional actions required:

### DIFF
--- a/picard/coverart/processing/filters.py
+++ b/picard/coverart/processing/filters.py
@@ -1,0 +1,47 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2024 Giorgio Fontanive
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from PyQt6.QtGui import QImage
+
+from picard import log
+from picard.config import get_config
+from picard.extension_points.cover_art_filters import register_cover_art_filter
+
+
+def size_filter(data):
+    config = get_config()
+    if not config.setting['filter_cover_by_size']:
+        return True
+    image = QImage.fromData(data)
+    width = config.setting['cover_width_threshold']
+    height = config.setting['cover_height_threshold']
+    if not (image.width() >= width and image.height() >= height):
+        log.debug(
+            "Discarding cover art due to size. Image size: %d x %d. Minimum: %d x %d",
+            image.width(),
+            image.height(),
+            width,
+            height
+        )
+        return False
+    return True
+
+
+register_cover_art_filter(size_filter)
--- a/picard/coverart/processing/filters.py
+++ b/picard/coverart/processing/filters.py
@@ -28,42 +28,33 @@ from picard.extension_points.cover_art_filters import (
 )
 
 
-def size_filter(data):
+def _check_threshold_size(width, height):
     config = get_config()
     if not config.setting['filter_cover_by_size']:
         return True
-    image = QImage.fromData(data)
-    width = config.setting['cover_width_threshold']
-    height = config.setting['cover_height_threshold']
-    if not (image.width() >= width and image.height() >= height):
+    min_width = config.setting['cover_width_threshold']
+    min_height = config.setting['cover_height_threshold']
+    if width < min_width or height < min_height:
         log.debug(
             "Discarding cover art due to size. Image size: %d x %d. Minimum: %d x %d",
-            image.width(),
-            image.height(),
             width,
-            height
+            height,
+            min_width,
+            min_height
         )
         return False
     return True
+
+
+def size_filter(data):
+    image = QImage.fromData(data)
+    return _check_threshold_size(image.width(), image.height())
 
 
 def size_metadata_filter(metadata):
-    config = get_config()
-    if (not config.setting['filter_cover_by_size']
-            or 'width' not in metadata or 'height' not in metadata):
+    if 'width' not in metadata or 'height' not in metadata:
         return True
-    width = config.setting['cover_width_threshold']
-    height = config.setting['cover_height_threshold']
-    if not (metadata['width'] >= width and metadata['height'] >= height):
-        log.debug(
-            "Avoiding download of cover art due to size. Image size: %d x %d. Minimum: %d x %d",
-            metadata['width'],
-            metadata['height'],
-            width,
-            height
-        )
-        return False
-    return True
+    return _check_threshold_size(metadata['width'], metadata['height'])
 
 
 register_cover_art_filter(size_filter)
--- a/picard/coverart/processing/filters.py
+++ b/picard/coverart/processing/filters.py
@@ -22,7 +22,10 @@ from PyQt6.QtGui import QImage
 
 from picard import log
 from picard.config import get_config
-from picard.extension_points.cover_art_filters import register_cover_art_filter
+from picard.extension_points.cover_art_filters import (
+    register_cover_art_filter,
+    register_cover_art_metadata_filter,
+)
 
 
 def size_filter(data):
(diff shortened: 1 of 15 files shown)


zas:
It looks good, but you have to be aware of something: https://github.com/metabrainz/picard/blob/7492d608bc0520eba5d4f94ad1b41356b20646b0/picard/coverart/providers/caa.py#L326-L338

Basically for PDFs, we download the "thumbnail" image, which is likely small, but it's only for display in interface, the real image being the PDF itself. We do not want to filter out such thumbnail image.

twodoorcoupe:
Thank you zas, I did not notice that.

I was wondering if it made sense for the caa provider to try and filter out images by size before downloading them, since there's already an option for choosing the size of caa images.

Also, I don't understand why the check  is failing. There seems to be an issue with PyQt6, but I can't see how it's related to my last 2 commits, since the first one in the pr was fine.

picard/coverart/processing/filters.py
```diff
@@ -0,0 +1,70 @@
...
+
+
+def size_filter(data):
+    config = get_config()
+    if not config.setting['filter_cover_by_size']:
+        return True
+    image = QImage.fromData(data)
+    width = config.setting['cover_width_threshold']
+    height = config.setting['cover_height_threshold']
+    if not (image.width() >= width and image.height() >= height):
```
zas, line 38: I think that's easier to read if written:

```suggestion
    if image.width() < width and image.height() < height:
```

picard/coverart/processing/filters.py
```diff
@@ -0,0 +1,70 @@
...
+            height
+        )
+        return False
+    return True
+
+
+def size_metadata_filter(metadata):
+    config = get_config()
+    if (not config.setting['filter_cover_by_size']
+            or 'width' not in metadata or 'height' not in metadata):
```
zas, line 53: Minor thing, accessing config is costly, so better first test metadata and return, I think it's also easier to read.

```python
def size_metadata_filter(metadata):
    if width' not in metadata or 'height' not in metadata:
        return True
    config = get_config()
    if not config.setting['filter_cover_by_size']:
        return True
``

File: picard/coverart/processing/filters.py [hunk 3]
```diff
@@ -0,0 +1,70 @@
...
+        return True
+    width = config.setting['cover_width_threshold']
+    height = config.setting['cover_height_threshold']
+    if not (metadata['width'] >= width and metadata['height'] >= height):
```
zas, line 57: The code here is almost identical to the one above, it would be a good idea to make it common.

zas (changes requested):

zas:
> Also, I don't understand why the check is failing. There seems to be an issue with PyQt6, but I can't see how it's related to my last 2 commits, since the first one in the pr was fine.

It was Github Actions bugging, it happens quite a lot those times, restarting the jobs was enough to fix it.

zas:
> I was wondering if it made sense for the caa provider to try and filter out images by size before downloading them, since there's already an option for choosing the size of caa images.

It makes sense for "original" CAA images.
CAA provides thumbnails(those are generated on Internet Archives servers)

See this example from https://archive.org/download/mbid-81b20fe4-e57e-44fd-8300-40f80b002dd2

`index.json` contains:

```json
    {
      "types": [
        "Front"
      ],
      "front": true,
      "back": false,
      "edit": 20306152,
      "image": "http://coverartarchive.org/release/81b20fe4-e57e-44fd-8300-40f80b002dd2/2928750079.jpg",
      "comment": "",
      "approved": true,
      "thumbnails": {
        "large": "http://coverartarchive.org/release/81b20fe4-e57e-44fd-8300-40f80b002dd2/2928750079-500.jpg",
        "small": "http://coverartarchive.org/release/81b20fe4-e57e-44fd-8300-40f80b002dd2/2928750079-250.jpg"
      },
      "id": "2928750079"
    },
```
In this case, `index.json` (which is used by CAA provider in Picard) lists 2 thumbnails (a "small" and a "large"; names not size in pixels), it should be noted there's also a 1200px thumbnail (those we added more recently) but in this case `index.json` wasn't yet regenerated; so it doesn't list it. Picard relies on `index.json`.

So we can use thumbnails or original image (`image` key) depending on cases.

We could use new filter by size settings to decide which image to download and drop CAA size selector (of course it means selecting a maximum size). Also it should be noted the thumbnail size on CAA (500px, etc.) is the width of the image.

twodoorcoupe:
Thanks! I'll look into this.

twodoorcoupe:
> We could use new filter by size settings to decide which image to download and drop CAA size selector (of course it means selecting a maximum size).

If I understand this correctly, the CAA provider should download the image closest to the maximum size, only if it is bigger than the size in the filter (if it is active)? 

In this case, in the CAA options page, the setting "Only use images of the following size" should become something like "Use images at most of the following size"?

Also, in the `index.json` there's no information about the size of the original CAA image, only of the thumbnails, correct?

zas:
> If I understand this correctly, the CAA provider should download the image closest to the maximum size, only if it is bigger than the size in the filter?
> 
> In this case, in the CAA options page, the setting "Only use images of the following size" should become something like "Use images at most of the following size"?
> 

Yes, something like that, though that's a bit tricky.

Basically the size in CAA defines the max width of images (we don't know anything about height at this point), with (currently) 3 values: 250, 500, 1200 (let's called that CAAW)

So if user wants an image no smaller than WxH we clearly need to select something with CAAW > W, but we have no guarantee about actual height being > H.

The reason we download thumbnails is mainly to speed up downloads. Original images can be very heavy, like > 100MB per image, so downloading them by default isn't really an option.


> Also, in the `index.json` there's no information about the size of the original CAA image, only of the thumbnails, correct?

Unfortunately no. And not even for thumbnails, since we only know the width because those are scaled up/down to this width, it  appears as key (either small/large (historical), or 250, 500, 1200 for more recent format), and in the name of the file (`-500.jpg` suffix for example).

See https://ia800400.us.archive.org/10/items/mbid-dfe84843-586d-4391-8a03-c0757e6073d8/index.json

phw:
> I was wondering if it made sense for the caa provider to try and filter out images by size before downloading them, since there's already an option for choosing the size of caa images.

I think it definitely makes sense to allow providers to abort early if they know there is no suitable artwork available, but the details where to check this can vary from provider to provider. And I fear we currently don't have a built-in provider that can provide the info.

For the CAA we only know whether cover art is available for a release. This is already checked in `CoverArtProviderCAA.enabled`, or rather in the `_has_suitable_artwork` method. But it doesn't tell us sizes.

Next the CAA provider downloads JSON with metadata like this:

https://coverartarchive.org/release/208a416f-7ecc-4726-a747-6829946d8e47

This might look like it can indicate the sizes, but it doesn't. E.g. take the following:

https://coverartarchive.org/release/04cf534c-ecc0-4bfe-9843-0648159168af

First there are no 250, 500 and 1200 keys. But this is just because this is an old entry, and the old names where "small" and "large" for "250" and "500". The image might well be over 1200 pixels in size, but since it is an old entry it does not provide the number indexes. Second issue is that if the cover art is smaller the CAA JSON still offers the 250, 500, and 1200 versions. The resulting image will be smaller, though.

picard/coverart/processing/filters.py
```diff
@@ -0,0 +1,62 @@
...
+    register_cover_art_metadata_filter,
+)
+
+
+def _check_threshold_size(width, height):
+    config = get_config()
+    if not config.setting['filter_cover_by_size']:
+        return True
+    # If the given width or height is -1, that dimension is not considered
+    min_width = config.setting['cover_width_threshold'] if width != -1 else -1
```
zas, line 36: `cover_width_threshold` isn't really clear, I think something like `cover_minimum_width` would be more explicit.

picard/coverart/providers/caa.py
```diff
@@ -109,7 +111,8 @@ def caa_url_fallback_list(desired_size, thumbnails):
                 url = thumbnails.get(size_alias, None)
         if url is not None:
             urls.append(url)
-    return urls
+            widths.append(item_id)
+    return urls, widths
```
zas, line 115: Here we should rather return a list of named tuples:

```python
CAAThumbnailListItem = namedtuple('CAAThumbnailListItem', ('url', 'width')

...

thumbnail_list = []

...
    thumbnail_list.append(CAAThumbnailListItem(url=url, width=item_id))
return thumbnail_list
```

BTW, I wonder also about `item_id` here, not really a great name (but that's not your fault;).

And below:

```python
                        thumbnail_list = caa_url_fallback_list(config.setting['caa_image_size'], image['thumbnails'])
                        if not thumbnail_list or is_pdf:
                            url = image['image']
                        else:
                            image_data = {'width': thumbnail_list[0].width, 'height': -1}
...
```

zas (changes requested):

zas:
@twodoorcoupe it looks good to me now :) good job

zas (approved):

phw (approved):
@twodoorcoupe Yes, this looks really good. Great, first step done!

PR #2508, opened 2024-06-03
Refactor dataobj into item

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

This further refactors the `Item` class hierarchy by:

1. Merging the old `DataObj` class with `MetadataItem`. `DataObj` had basically no functionality on its own. This also means `MetadataItem` now inherits from `QObject`, which it should anyway as it is providing the `metadata_images_changed` Qt signal.
2. Renamed the old `DataObj.item` attribute to `ui_item` and moved it to the `Item` class. This makes the purpose more clear and avoids confusion. This attribute is meant to hold item classes from `ui.itemviews` if the object is shown in the UI.
3. Renamed `File.parent` to `File.parent_item`. Because `File` inherits from `QObject` this was another case of hiding `QObject.parent`. Because the QObject parents are not being used in our code this worked, but hiding inherited functions or properties always has the potential to cause side effects.


phw:
@zas There likely will be another PR, I'm looking into some changes to the file parent handling. No 100% sure yet, but I think there will be some additional refactoring before I finally do the fix for the issue with `suspend_metadata_images_update` context in coverartbox.

picard/ui/itemviews/__init__.py
```diff
@@ -512,13 +512,13 @@ def update(self, update_album=True, update_files=True, update_selection=True):
...
                     for i in range(oldnum - newnum):
-                        self.takeChild(newnum - 1).obj.item = None
+                        self.takeChild(newnum - 1).obj.ui_item = None
                     oldnum = newnum
                 for i in range(oldnum):  # update existing items
                     item = self.child(i)
                     file = track.files[i]
                     item.obj = file
-                    file.item = item
+                    file.ui_item = item
```
zas, line 521: I think we can get rid of intermediate `item` variable here:

```python
    file = track.files[i]
    file.ui_item = self.child(i)
    file.ui_item.obj = file                 
```

picard/file.py
```diff
@@ -627,36 +627,36 @@ def _apply_additional_files_moves(self, moves):
...
+            log.debug("Removing %r from %r", self, self.parent_item)
+            self.parent_item.remove_file(self)
         self.tagger.acoustidmanager.remove(self)
         self.state = File.REMOVED
 
     def move(self, parent):
         # To be able to move a file the target must implement add_file(file)
-        if hasattr(parent, 'add_file') and parent != self.parent:
-            log.debug("Moving %r from %r to %r", self, self.parent, parent)
+        if hasattr(parent, 'add_file') and parent != self.parent_item:
```
zas, line 638: For consistency, we should also rename `parent` argument, I suggest indicating direction too for all those methods: `to_parent_item` , `from_parent_item`

zas (dismissed):
Good move, getting rid of `DataObj` is welcome.
Renaming overused `item` and `parent` will help.

picard/ui/itemviews/__init__.py
phw: I took a bit different approach in 402f8463b62a9327e03e935c28b74e39ca1c268d

picard/file.py
phw: See 2369431e4713d6c1acff44e6342cfe992daabaf7

zas (approved):
LGTM

PR #2510, opened 2024-06-07
PICARD-1241: Add option to resize images above a given maximum size

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
Adds an option resize downloaded images down to a given width and height, if they are above the given size.
# Problem

* JIRA ticket (_optional_): PICARD-1241

Also these tickets are similar PICARD-237, PICARD-1663, PICARD-2783

# Solution

The feature is fully functional, I was planning to add multi-threading and other optimizations later.

The processing options page looks like this as of now, let me know if it's too cluttered. 

image

The image processors extension point is a bit more complicated than the filters one, because we need to know if it's running for either embedded images or ones saved to an external file. We also need to know whether it's running for both with the same settings, in order to avoid running the same processor twice.

My idea was for new image processors to inherit from the `ImageProcessor` class and redefine its methods, similar to how new `OptionPage`s are handled.

Also, the `coverartimage` now has a `external_file_coverart` to store the image meant to be saved to an external file. This is necessary when the processing options for the two targets are different.

# Action

Additional actions required:


twodoorcoupe:
I still have a doubt about image quality. When getting the image data from a `QImage`, there's a field to set the quality level that goes from 0 to 100. This quality slider is independent of image formats. As of now, I just put 100, but this result in  images large in size. 

For example, when I download a 1200px jpeg image from CAA and resize it to 500px, the 500px is actually larger in file size when quality is set to 100.

Should I just set a fixed reasonable quality level? While playing around, I found that anywhere from 85 to 90 was similar to CAA jpeg pictures of the same size.

rdswift:
Welcome to the world of JPEG compression.  :wink: 

Here are a couple of links with a good technical explanation of what's happening:

- https://www.baeldung.com/cs/jpeg-compression
- https://www.techspot.com/article/2792-jpeg-explained/

Unfortunately JPEG (lossy) compression can often result in [compression artifacts](https://en.wikipedia.org/wiki/Compression_artifact) if not used properly.  The compression / quality setting that yields an acceptable image can vary significantly.  For a reasonable explanation of JPEG compression artifacts and how they can be addressed, have a look at https://www.widsmob.com/tips/jpeg-artifacts.html

rdswift:
I also checked the "documentation" box in the ***Action*** section above.  Since this is a new feature, we will definitely have to update the documentation.

picard/coverart/processing/__init__.py
```diff
@@ -41,5 +56,19 @@ def run_image_metadata_filters(metadata):
     return True
 
 
-# def run_image_processors(data):
-#     pass
+def run_image_processors(data, coverartimage):
+    try:
+        info = ImageInfo(*imageinfo.identify(data))
```
zas, line 61: Instead `imageinfo.identify()` could directly return ImageInfo tuple and move the ImageInfo declaration there.

https://github.com/metabrainz/picard/blob/cc3801059734b4d6cca0a3fd51b07666d0c5309f/picard/util/imageinfo.py#L63

picard/extension_points/cover_art_processors.py
```diff
@@ -0,0 +1,65 @@
...
+    queue_both, queue_tags, queue_file = [], [], []
+    for processor in ext_point_cover_art_processors:
+        if processor.same_processing():
+            queue_both.append(processor)
+        else:
+            if processor.save_to_tags():
+                queue_tags.append(processor)
+            if processor.save_to_file():
+                queue_file.append(processor)
+    return queue_both, queue_tags, queue_file
```
zas, line 60: We could perhaps make this a generator and yield `ProcessingTarget` along:

```python
def get_cover_art_processors():
    for processor in ext_point_cover_art_processors:
        if processor.same_processing():
            yield (processor, ProcessingTarget.BOTH)
        else:
            if processor.save_to_tags():
                yield (processor, ProcessingTarget.TAGS)
            if processor.save_to_file():
                yield (processor, ProcessingTarget.FILE)
  ```

And in the code calling this method, something like:

```python
        processed_data = dict()
        for processor, target in get_cover_art_processors():
            processed_data[target] = processor.run(data, info, target)
```

What do you think?

picard/options.py
```diff
@@ -173,6 +173,12 @@
 BoolOption('setting', 'filter_cover_by_size', False)
 IntOption('setting', 'cover_minimum_width', 250)
 IntOption('setting', 'cover_minimum_height', 250)
+BoolOption('setting', 'resize_images_saved_to_tags', False)
+IntOption('setting', 'cover_tags_maximum_width', 1000)
```
zas, line 177: Those numbers should perhaps be stored in https://github.com/metabrainz/picard/blob/master/picard/const/defaults.py :

```python
DEFAULT_COVER_MIN_SIZE = 250
DEFAULT_COVER_MAX_SIZE = 1000
```

```python
IntOption('setting', 'cover_minimum_width', DEFAULT_COVER_MIN_SIZE)
IntOption('setting', 'cover_minimum_height', DEFAULT_COVER_MIN_SIZE)
IntOption('setting', 'cover_tags_maximum_width', DEFAULT_COVER_MAX_SIZE)
IntOption('setting', 'cover_tags_maximum_height', DEFAULT_COVER_MAX_SIZE)
IntOption('setting', 'cover_file_maximum_width', DEFAULT_COVER_MAX_SIZE)
IntOption('setting', 'cover_file_maximum_height', DEFAULT_COVER_MAX_SIZE)
```

zas:
> JPEG (lossy) compression

Also, be sure internal processing is done using non-lossy format especially if we intend to chain processors. Conversion from JPEG should be done at entry, and once at exit.

Usually a good compression value for JPEG is around 90-95 if quality is important, 80-90 if less important, above 95 usually increases the size of files for nothing, and under 80 is very likely to have visible artifacts.

While we can save any format to disk, that's not the case for tags.

phw (commented):
This is looking like a very good next step.

> Also, be sure internal processing is done using non-lossy format especially if we intend to chain processors. Conversion from JPEG should be done at entry, and once at exit.

Yes, I agree. Maybe this could be achieved by loading the image as a QImage in `run_image_processors` once and pass the `QImage` instance around from processor to processor. This would also eliminate the overhead of decoding / encoding the image multiple times.

phw:
Another thing we should keep in mind for this project if there is time at the end is to also adapt the cover art preview UI. Maybe we need there an additional column to separate tag and file image.

phw:
> I still have a doubt about image quality. When getting the image data from a `QImage`, there's a field to set the quality level that goes from 0 to 100. This quality slider is independent of image formats. As of now, I just put 100, but this result in images large in size.

I strongly assume that the 0-100 quality setting of Qt only applies to JPEG, at least this is the format where it comes from and where it makes sense. So we totally could add a setting "JPEG quality" for this on the provider option, with a default to maybe 90. Then users can adjust this if needed.

twodoorcoupe:
>Also, be sure internal processing is done using non-lossy format especially if we intend to chain processors. Conversion from JPEG should be done at entry, and once at exit.

>Maybe this could be achieved by loading the image as a QImage in run_image_processors once and pass the QImage instance around from processor to processor. This would also eliminate the overhead of decoding / encoding the image multiple times.

I had thought about passing the QImage in the processor chain, but that would force plugins to use qt for their processing and make it more complicated if in the future another image processing library is added. For now, I made internal processing use the bmp format as it is uncompressed, but let me know if you prefer passing the QImage instead.

I added the processor for converting images, for now it runs after resizing and converts the image back to its original format. I was planning to add the option to convert to a different format and its UI in a different pr, since I think it's a different feature. If you prefer to see it added to this pr instead, that's ok with me.

picard/extension_points/cover_art_processors.py
twodoorcoupe: The issue is that we need to execute all processors with `same_processing = True` first. Then once all of those are finished we run the two other target specific queues with the data obtained from the first. I think using a generator like this does not ensure the order of the different queues, or am I wrong? 

Once I start working on adding multi-threading, I will have to change the `run_image_processors` function anyways, because I was planning on having the processing chain for each target run in a separate thread.

zas: > The issue is that we need to execute all processors with `same_processing = True` first. Then once all of those are finished we run the two other target specific queues with the data obtained from the first. I think using a generator like this does not ensure the order of the different queues, or am I wrong?

Hmmm, right.
We can still use a generator, but explode the loops:

```python
def get_cover_art_processors():
    not_same_processing = []
    for processor in ext_point_cover_art_processors:
        if processor.same_processing():
            yield (processor, ProcessingTarget.BOTH)
        else:
            not_same_processing.append(processor)
    for processor in not_same_processing:
        if processor.save_to_tags():
            yield (processor, ProcessingTarget.TAGS)
        if processor.save_to_file():
            yield (processor, ProcessingTarget.FILE)
```

zas:
> bmp format

I would avoid this format, for internal lossless processing we should instead use PNG which has advantages in all fields (open standard, transparency support, file size limitations, and compressed file size on disk).

picard/coverart/processing/processors.py
```diff
@@ -0,0 +1,99 @@
...
+        if target == ProcessingTarget.TAGS:
+            max_width = config.setting['cover_tags_maximum_width']
+            max_height = config.setting['cover_tags_maximum_height']
+        else:
+            max_width = config.setting['cover_file_maximum_width']
+            max_height = config.setting['cover_file_maximum_height']
+        if info.width <= max_width and info.height <= max_height:
+            return data
+        image = QImage.fromData(data)
+        scaled_image = image.scaled(max_width, max_height, Qt.AspectRatioMode.KeepAspectRatio)
```
zas, line 70: It would be useful to collect processing time (here and everywhere we do image processing) and display it in debug output.

twodoorcoupe:
>I would avoid this format, for internal lossless processing we should instead use PNG

Makes sense, thank you! I think we can use png for now, then once I finish adding multi-threading to the processing, I can check the performance as a whole and see if it's best to pass the `QImage` in the chain instead.

picard/coverart/processing/__init__.py
```diff
@@ -53,17 +56,21 @@ def run_image_metadata_filters(metadata):
...
-        coverartimage.set_tags_data(tags_data)
-        coverartimage.set_external_file_data(file_data)
+        processed_data = {ProcessingTarget.BOTH: data}
+        for processor, target in get_cover_art_processors():
+            if target == ProcessingTarget.SKIP:
+                common_data = processed_data[ProcessingTarget.BOTH]
+                processed_data = {ProcessingTarget.TAGS: common_data, ProcessingTarget.FILE: common_data}
+                continue
+            processed_data[target] = processor.run(processed_data[target], info, target)
+        coverartimage.set_tags_data(processed_data[ProcessingTarget.TAGS])
```
zas, line 68: Ok, it seems having a generator leads to more complex code than I expected.
Also I suspect `processed_data[ProcessingTarget.TAGS]` may not be always defined.

If you prefer, go back to lists, the initial goal of my suggestion was to have simpler code, but it doesn't seem to work.

zas (changes requested):

picard/coverart/processing/__init__.py
twodoorcoupe: I went back to lists, sorry about that.  I assume I will have to change this part anyways to make each queue run in a different thread.

picard/coverart/processing/__init__.py
```diff
@@ -41,5 +54,25 @@ def run_image_metadata_filters(metadata):
...
+    try:
+        start_time = time.time()
+        info = imageinfo.identify(data)
+        both, tags, file = get_cover_art_processors()
+        for processor in both:
+            data = processor.run(data, info, ProcessingTarget.BOTH)
+        tags_data = data
+        file_data = data
+        for processor in tags:
+            tags_data = processor.run(tags_data, info, ProcessingTarget.TAGS)
```
zas, line 67: `info` is not modified by processor, so it keeps initial values even though the processor changes those (for example, a processor changed dimensions, the next processor will get original dimensions, not the actual ones).

zas (changes requested):
Overall it looks good, I tested and it works as expected.
Though I think there are few issues:

- chained processors execution needs some refinements
- info is passed but never modified, so basically the next processor may get and/or use incorrect information
- errors / exceptions handling needs some attention

picard/coverart/processing/processors.py
```diff
@@ -0,0 +1,104 @@
...
+from picard.extension_points.cover_art_processors import (
+    ImageProcessor,
+    ProcessingTarget,
+    register_cover_art_processor,
+)
+
+
+def _get_image_data(image, image_format):
+    buffer = QBuffer()
+    image.save(buffer, image_format, quality=90)
```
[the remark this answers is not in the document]
twodoorcoupe: If I understand correctly: when the `image.save` function returns false, we should still set whatever data we already have to the `coverartimage`. Then we could append an error to the `Album`, that way it's clearer to the user that the processing did not run correctly, instead of just writing to the log.

picard/coverart/processing/__init__.py
twodoorcoupe: You're right, sorry for missing that. I was thinking whether it makes more sense to use a `dataclass` then? Since they are mutable, each image processor does not have to import and create a new `ImageInfo` named tuple, but can instead modify whatever field it has changed. Does it make sense?

zas: Yes, it makes sense.

twodoorcoupe:
To avoid converting the image for each processor, I thought about a slightly different approach than what you suggested @zas. 

I created the `ProcessingImage` class that gets passed in the chain. That way each processor calls `image.get_result()` to get the `QImage`, with the option of specifying a format if encoding is necessary. Then each processor can do `image.set_result(image)` and can pass either a `QImage` or data directly. 

This way should also avoid concurrency issues later down the line. I also moved handling of `info` inside this class to simplify code a little bit. If you prefer I can try to follow what you suggested instead.

picard/util/imageinfo.py
```diff
@@ -60,7 +74,7 @@ def read(self):
         return self._result()
 
     def _result(self):
-        return (int(self.w), int(self.h), self.mime, self.extension, self.datalen)
+        return ImageInfo(int(self.w), int(self.h), self.mime, self.extension, self.datalen)
```
zas, line 77: Just a minor thing, but better use keyword arguments, it eases changes:

```python
    return ImageInfo(
        width=int(self.w),
        height=int(self.h),
        mime=self.mime,
        extension=self.extension,
        datalen=self.datalen,
    )
``

File: test/test_util_imageinfo.py [hunk 9]
```diff
@@ -36,7 +36,7 @@ def test_gif(self):
         with open(file, 'rb') as f:
             self.assertEqual(
                 imageinfo.identify(f.read()),
-                (140, 96, 'image/gif', '.gif', 5806)
+                imageinfo.ImageInfo(140, 96, 'image/gif', '.gif', 5806)
```
zas, line 39: We can also use keyword arguments here (and below) now.

picard/coverart/image.py
```diff
@@ -307,7 +310,7 @@ def set_data(self, data):
 
         try:
             (self.width, self.height, self.mimetype, self.extension,
-             self.datalength) = imageinfo.identify(data)
+             self.datalength) = astuple(imageinfo.identify(data))
```
zas, line 313: Instead of using tuples we could just use attributes:

```python
   info = imageinfo.identify(data)
   self.width, self.height = info.width, info.height
   self.mimetype = info.mime
   self.extension = info.extension
   self.datalength = info.datalen
```

I know it's longer but experience shows the code is easier to manage if we don't have too much positional arguments and unnamed tuples.

zas (changes requested):
It looks much better, just few stylistic comments.

I'll give it a full test later.

zas:
I tested, it works as expected.

Though I think we should improve over this:

- let the user set a maximum width OR maximum height (for example, max width = 1000 px, with an original image of 2000 (w)  x 4000 (h) will result in a scaled image of 1000 (w) x 2000 (h), aspect preserved ofc. Current UI doesn't let ignore one of the dimensions.
- let the user scales up: if the original image is 500x1000 and minimum resize width is 1000, resulting image will be 1000x2000. Current code only considers maximum dimensions.
- for CAA provider, it isn't very clear how this interacts with thumbnail size, perhaps a message should be displayed when user selects for example a maximum width > CAA width (in this case, it has no effect).

Perhaps you planned this for another PR, if that's the case, just ignore this comment.

twodoorcoupe:
I can add these features to this pr.

Regarding your first two points, how does this UI look? For now I can't come up with anything simpler. But maybe, instead of the checkboxes next to the spinboxes, we could let the user put 0 or -1 to indicate that the dimension should not matter.

image

Regarding your third point, just so we are on the same page: if the CAA thumbnail size setting is lower than the scale down width setting for both targets, a message is displayed to the user saying the resize has no effect? And the message should just be a label displayed in the cover art archive options page?

I guess for scaling up this does not apply because we don't know what size we are actually getting from CAA. Sometimes images would get scaled up even when the scale up width setting is lower than the CAA thumbnail size setting.

zas:
> Regarding your first two points, how does this UI look? For now I can't come up with anything simpler. But maybe, instead of the checkboxes next to the spinboxes, we could let the user put 0 or -1 to indicate that the dimension should not matter.

IMHO checkboxes are better than 0/-1 values, the UI you suggested look good to me, we'll have to test and see if it works well enough, but  it covers all cases I think.
Just perhaps move the checkbox for width/height before the label ([x] Width       1000 px instead of Width      [x] 1000 px).


> if the CAA thumbnail size setting is lower than the scale down width setting for both targets, a message is displayed to the user saying the resize has no effect? And the message should just be a label displayed in the cover art archive options page?

For the first one it seems difficult to manage if there are multiple providers, so I guess the best option is to add a warning on the CAA options page about impact of CAA image size in terms of postprocessing and download time (original might be too heavy, 250px too small if image postprocessing is used). We can do that later if you prefer to focus on processing.

phw (commented):
I like the latest changes with the ProcessingImage class. This really streamlines the pipeline. Should we actually swap out the image processing in the future this will also reduce the pain. Much better than just passing image around directly.

I haven't run the latest changes yet, but code looks good to me.

zas (approved):
LGTM.


Perhaps we should merge this PR, so you can continue on another one. This one works.

twodoorcoupe:
>Perhaps we should merge this PR, so you can continue on another one.

Ok, that's fine with me. Thank you both.

phw (approved):
Agreed. Good work again!

picard/const/defaults.py
```diff
@@ -150,3 +150,6 @@
 DEFAULT_NAMING_PRESET_ID = "Preset 1"
 
 DEFAULT_TIME_FORMAT = '%Y-%m-%d %H:%M:%S'
+
+DEFAULT_COVER_MIN_SIZE = 250
+DEFAULT_COVER_MAX_SIZE = 1000
```
Sophist-UK, line 155: 1. I would prefer constants and variables like these to include the UoM in their name i.e. `DEFAULT_COVER_MIN_SIZE_PX`.
2. I wonder whether it is more normal for images to come in binary powers i.e. should these be 256 and 1024?

picard/coverart/processing/processors.py
```diff
@@ -0,0 +1,77 @@
...
+        config = get_config()
+        return config.setting['resize_images_saved_to_file']
+
+    def save_to_tags(self):
+        config = get_config()
+        return config.setting['resize_images_saved_to_tags']
+
+    def same_processing(self):
+        config = get_config()
+        same_width = config.setting['cover_tags_maximum_width'] == config.setting['cover_file_maximum_width']
```
Sophist-UK, line 46: See above e.g. `cover_tags_maximum_width_px` etc.?

picard/options.py
```diff
@@ -171,8 +173,14 @@
...
-IntOption('setting', 'cover_minimum_width', 250)
-IntOption('setting', 'cover_minimum_height', 250)
+IntOption('setting', 'cover_minimum_width', DEFAULT_COVER_MIN_SIZE)
+IntOption('setting', 'cover_minimum_height', DEFAULT_COVER_MIN_SIZE)
+BoolOption('setting', 'resize_images_saved_to_tags', False)
+IntOption('setting', 'cover_tags_maximum_width', DEFAULT_COVER_MAX_SIZE)
+IntOption('setting', 'cover_tags_maximum_height', DEFAULT_COVER_MAX_SIZE)
+BoolOption('setting', 'resize_images_saved_to_file', False)
+IntOption('setting', 'cover_file_maximum_width', DEFAULT_COVER_MAX_SIZE)
+IntOption('setting', 'cover_file_maximum_height', DEFAULT_COVER_MAX_SIZE)
```
Sophist-UK, line 183: 1. I am unclear why files and tags have separate maximum sizes but not separate minimum sizes?
2. Unclear why `cover_tags_maximum_width` is plural `tags` but `cover_file_maximum_width` is singular `file`?

Sophist-UK (commented):
Apologies for late comments - my wife had a hip replacement op and I have been caring for her. Since there will be follow-on PRs, hopefully these comments are not too late.

> I would avoid this format, for internal lossless processing we should instead use PNG which has advantages in all fields (open standard, transparency support, file size limitations, and compressed file size on disk).

I think we should let the user have an option for cover art file type, and image processing should convert files as needed.

Also, is this an opportunity to fix the cover art comparison (existing cover art vs. downloaded cover art) so that we don't get false-positives for album metadata having been changed even though both metadata and images are the same.

twodoorcoupe:
Thank you @Sophist-UK for your review!

> I think we should let the user have an option for cover art file type, and image processing should convert files as needed.

I plan on adding this very soon, after I'm finished with zas' suggested improvements.

>  so that we don't get false-positives for album metadata having been changed even though both metadata and images are the same.

This is the first time I hear of this issue, but I'd be happy to look into it. I couldn't find a ticket on jira or a topic on the forum, could you please show me some examples? 

> I would prefer constants and variables like these to include the UoM in their name

I just followed what the other constants do, for example `DEFAULT_CAA_IMAGE_SIZE`

> I am unclear why files and tags have separate maximum sizes but not separate minimum sizes?

The minimum size is intended to discard providers' images and download the next ones in the queue, whereas the maximum sizes are used to scale down the downloaded images. 

The idea is that the same downloaded image is processed into two different ones for tags and files. I didn't want to overcomplicate things, and I think downloading two separate images is slower than just one and resizing it in most cases, but I might be wrong on that one.

Also, as zas suggested I will be adding the option to scale up downloaded images as well. So all the "maximum" settings will become something more like `target_width` and `target_height`.

> Unclear why cover_tags_maximum_width is plural tags but cover_file_maximum_width is singular file?

In my head it made sense because I say "I embed one image into the tags" and "I save one image as a separate file". But I don't feel strongly either way, I can change it.

phw:
> I think we should let the user have an option for cover art file type, and image processing should convert files as needed.

Format conversion is on the list for image processing options. The format discussed here is unrelated. The discussion here was solely about how to pass images internally between processors while processing.

> Also, is this an opportunity to fix the cover art comparison (existing cover art vs. downloaded cover art) so that we don't get false-positives for album metadata having been changed even though both metadata and images are the same.

I assume you mean PICARD-1001 and related tickets. Currently there are two known issues: 1. image order not being stable in ID3 tags (needs to be fixed in mutagen) and 2. issues with comparing images in combination with multiple image types and the local image priovider. Both are out-of-scope for and unrelated to the image processing twodoorcoupe is implementing here.

PR #2511, opened 2024-06-11
Move PluginPriority to extension_points

zas:
- it is used mainly as an optional parameter by register_*() extension methods
- make it an IntEnum

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/extension_points/enums.py
+++ b/picard/extension_points/enums.py
@@ -1,0 +1,48 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2007-2008 Lukáš Lalinský
+# Copyright (C) 2009 Carlin Mangar
+# Copyright (C) 2009, 2014, 2017-2021, 2023 Philipp Wolfer
+# Copyright (C) 2011 johnny64
+# Copyright (C) 2011-2013 Michael Wiencek
+# Copyright (C) 2013 Sebastian Ramacher
+# Copyright (C) 2013 Wieland Hoffmann
+# Copyright (C) 2013 brainz34
+# Copyright (C) 2013-2014 Sophist-UK
+# Copyright (C) 2014 Johannes Dewender
+# Copyright (C) 2014 Shadab Zafar
+# Copyright (C) 2014-2015, 2018-2021, 2023-2024 Laurent Monin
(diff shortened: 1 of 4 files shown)


phw:
@zas No strong opinion. It makes some sense, but it also makes sense when imported from `plugins`. `from picard.plugin import PluginPriority` seems fine to me, also from the perspective of plugins. The other question is whether this option stays as it is. So I'd probably leave this alone for now.

But anyway, if you prefer it to be moved I'm fine with that as well.

zas:
Perhaps a better move would be to get rid of the enum and just use an integer: 0 being normal priority, positive number increased priority, negative decreased priority.

The code only use NORMAL priority, and it was meant to let one run plugins in a defined order (it can be handy in some cases). That's perhaps something that should be configurable by UI, because it really depends on the use.

phw:
> That's perhaps something that should be configurable by UI, because it really depends on the use.

That. Organizing the order would be more helpful. And not by ordering plugins, but by ordering registered hooks.

> Perhaps a better move would be to get rid of the enum and just use an integer: 0 being normal priority, positive number increased priority, negative decreased priority.

I think the enum is helpful for the public interface (should we keep this mechanism). It makes it clear what the priority is. With 0 and 1 you get people asking what runs first.

> The code only use NORMAL priority, and it was meant to let one run plugins in a defined order (it can be handy in some cases).

Plugins also use LOW and HIGH, and there have been some constellations where this was required to be set for certain plugins to work together. But it is overall not a good mechanism and only works (more or less) if you know and control all plugins.

zas:
I tried to drop the enum, I think it makes things simpler finally.
Let's say someone has 2 plugins, running on prio HIGH and NORMAL, and wants to run a third plugin between HIGH and NORMAL -> possible, because Python doesn't check types and one can pass a value of 50 (between 0=NORMAL and 100=HIGH), but that's hacky, and defeat the initial purpose of enum.

So I think we should just let the user passes any number, and just sort (reversed) those, not bothering with enums in this case (btw, I think that's me who added them and they were numbers before..., so I basically revert this).

See code in https://github.com/metabrainz/picard/pull/2512

PR #2512, opened 2024-06-11
Drop PluginPriority in favor of a simple integer and add tests

zas:
- Only default value was used in the code
- having an enum for that is rather limiting its use
- add tests for PluginFunctions class

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/extension_points/event_hooks.py
+++ b/picard/extension_points/event_hooks.py
@@ -57,62 +57,61 @@ from picard.file import (
     file_post_removal_to_track_processors,
     file_post_save_processors,
 )
-from picard.plugin import PluginPriority
 
 
-def register_album_post_removal_processor(function, priority=PluginPriority.NORMAL):
+def register_album_post_removal_processor(function, priority=0):
     """Registers an album-removed processor.
     Args:
         function: function to call after album removal, it will be passed the album object
-        priority: optional, PluginPriority.NORMAL by default
+        priority: optional, 0 by default
     Returns:
         None
     """
     album_post_removal_processors.register(function.__module__, function, priority)
 
 
-def register_file_post_load_processor(function, priority=PluginPriority.NORMAL):
+def register_file_post_load_processor(function, priority=0):
     """Registers a file-loaded processor.
 
     Args:
         function: function to call after file has been loaded, it will be passed the file object
-        priority: optional, PluginPriority.NORMAL by default
+        priority: optional, 0 by default
     Returns:
         None
     """
     file_post_load_processors.register(function.__module__, function, priority)
 
 
-def register_file_post_addition_to_track_processor(function, priority=PluginPriority.NORMAL):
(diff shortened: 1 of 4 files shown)


phw (commented):
I still have no strong opinion either way. Generally changes look fine to me, and it is great there are tests now.

rdswift:
I seem to recall seeing some question as to whether this priority setting was actually being used.  If I recall, I've set the priority to LOW in the Format Performer Tags plugin to ensure that any performer changes from other plugins were processed before the final formatting of the performer tags.

I really don't have any concern changing from an enum to an integer value, so long as we make it clear in the documentation.  The one question that I have is what is the available range for the priority setting?  The general discussion seems to be that a range of -100 to 100 (with zero as the default value), but what happens if someone uses a value outside of this range?  Should it be forced to the appropriate min/max value of the range and/or reported as an error, or is the acceptable range actually the range for a signed integer?  Again, I have no strong opinions either way, just so long as we make it clear in the documentation.

zas:
> The one question that I have is what is the available range for the priority setting?

Using a number (real or integer) leads to the acceptable range for those (defined by Python), and permits the use of decimals. The problem with the enum is that it fixes only 3 values, and doesn't help if one wants for order 2 plugins having lower or higher priority than normal.

IMHO this approach is a bit better, because it's easier to "insert" a plugin: let's say we have 2 plugins, one (P0) is on default priority (0) and the other (P1) on priority 1, it's possible to add a plugin P2 and set its priority to 0.5 for example, to ensure this run order: P1 P2 P0.
It's also easier to shift all priorities or scale priorities. So in general I think that's better approach if we want to let the user control the order of plugin execution.

For plugin metadata, we could have something a bit more complex, like "run before <list of plugins>" or "run after" as hints, and from that calculate a priority order (since it will be easier to "insert" a plugin in the order). But that's not the point here.

Also if we want to use UI to order plugin execution we could simply using positions in a list, and center that in the middle of the list.
For example, let's say we have 3 plugins, initially with same default priority (0): P1 P2 P3

And the UI let the user order those, this way: P2 runs before P1 runs before P3 -> we number from lowest, P3=0, P1=1, P2=2, and center on middle, P3=1, P1=0, P2=-1
If a new plugin is enabled, with default prio, the list becomes P3=1, P1=0, P4=0, P2=-1
If it doesn't have default prio, in all depends on the value but a very positive or very negative value is likely to do the right thing (it all depends on current list).

We shouldn't make this overcomplicated though, but the use of an enum was likely a bad idea, because what we really need is a way to run one plugin before others, mainly (or the reverse).

Sophist-UK:
Just to register a request that the documentation clearly explains when explicitly to choose a priority rather than use the default and how to determine the priority that the plugin should use.

PR #2514, opened 2024-06-15
Add new column for external images to artwork infodialog

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

Now the table looks like this:

(Both embedding images to tags and saving to external file)
existing_external

(Same as above but with no pre-existing cover art)
both_new

(Saving only to an external file, not embedding to tags)
only_external

I modified the `run_image_processors` function so that the tags and file queues run only if their result is actually used to either embed into tags or save to external file.

Also now the external `CoverArtImage` object is created only if the option to save to an external file is selected. This way it's not shown in the `infodialog`, so the resulting table in this case is the same as before.

# Action

Additional actions required:

### DIFF
--- a/picard/coverart/processing/__init__.py
+++ b/picard/coverart/processing/__init__.py
@@ -21,6 +21,7 @@
 import time
 
 from picard import log
+from picard.config import get_config
 from picard.coverart.processing import (  # noqa: F401 # pylint: disable=unused-import
     filters,
     processors,
@@ -53,6 +54,7 @@ def run_image_metadata_filters(metadata):
 
 
 def run_image_processors(data, coverartimage):
+    config = get_config()
     tags_data = data
     file_data = data
     try:
@@ -61,16 +63,18 @@ def run_image_processors(data, coverartimage):
         both_queue, tags_queue, file_queue = get_cover_art_processors()
         for processor in both_queue:
             processor.run(image, ProcessingTarget.BOTH)
-        tags_image = image.copy()
-        for processor in tags_queue:
-            processor.run(tags_image, ProcessingTarget.TAGS)
-        tags_data = tags_image.get_result(default_format=True)
+        if config.setting['save_images_to_tags']:
+            tags_image = image.copy()
+            for processor in tags_queue:
+                processor.run(tags_image, ProcessingTarget.TAGS)
+            tags_data = tags_image.get_result(default_format=True)
         coverartimage.set_tags_data(tags_data)
-        file_image = image.copy()
-        for processor in file_queue:
-            processor.run(file_image, ProcessingTarget.FILE)
-        file_data = file_image.get_result(default_format=True)
-        coverartimage.set_external_file_data(file_data)
+        if config.setting['save_images_to_files']:
+            file_image = image.copy()
+            for processor in file_queue:
+                processor.run(file_image, ProcessingTarget.FILE)
+            file_data = file_image.get_result(default_format=True)
+            coverartimage.set_external_file_data(file_data)
         log.debug(
             "Image processing for %s finished in %d ms",
             coverartimage,
@@ -80,5 +84,6 @@ def run_image_processors(data, coverartimage):
         raise CoverArtProcessingError(e)
     except CoverArtProcessingError as e:
         coverartimage.set_tags_data(tags_data)
-        coverartimage.set_external_file_data(file_data)
+        if config.setting['save_images_to_files']:
+            coverartimage.set_external_file_data(file_data)
         raise e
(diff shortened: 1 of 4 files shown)


picard/coverart/processing/__init__.py
```diff
@@ -61,16 +63,18 @@ def run_image_processors(data, coverartimage):
         both_queue, tags_queue, file_queue = get_cover_art_processors()
         for processor in both_queue:
             processor.run(image, ProcessingTarget.BOTH)
-        tags_image = image.copy()
-        for processor in tags_queue:
-            processor.run(tags_image, ProcessingTarget.TAGS)
-        tags_data = tags_image.get_result(default_format=True)
+        if config.setting['save_images_to_tags']:
```
zas, line 66: If "Embed cover images into tags" is disabled, we still have the image shown in New Cover (unprocessed), that's a bit misleading I think.

Capture d’écran du 2024-06-15 12-18-53

Capture d’écran du 2024-06-15 12-19-31

test/test_coverart_processing.py
```diff
@@ -115,7 +115,9 @@ def test_image_processors(self):
             'cover_tags_maximum_height': 500,
             'resize_images_saved_to_file': True,
             'cover_file_maximum_width': 750,
-            'cover_file_maximum_height': 750
+            'cover_file_maximum_height': 750,
+            'save_images_to_tags': True,
+            'save_images_to_files': True,
```
zas, line 120: Others combinations of this two options should be tested too.

picard/ui/infodialog.py
```diff
@@ -150,11 +163,28 @@ class ArtworkTableExisting(ArtworkTable):
...
+    NUM_COLS = 4
+
+    _columns = {
+        'orig': 0,
+        'type': 1,
+        'new': 2,
+        'external': 3,
+    }
+
+    _labels = (_("Existing Cover"), _("Type"), _("New Cover"), _("New External Cover"),)
```
zas, line 176: We could improve column titles and/or add a tool tip to clarify what are those exactly.

phw: Yes. I think we need better naming, but not sure about the exact wording. As we have "External Cover" the other should probably be "Tag Cover". It's all not really self-explaining, though. But yes, a tooltip could help with that.

picard/ui/infodialog.py
```diff
@@ -150,11 +163,28 @@ class ArtworkTableExisting(ArtworkTable):
     artwork_columns = ('orig', 'new',)
 
 
+class ArtworkTableExistingExternal(ArtworkTable):
```
phw, line 166: Do we really need separate classes for each case? As zas noted above it should also consider "Embed cover images into tags" and hide that column. Implementing this with yet another class will increase the combinations and the number of classes needed.

Maybe the better approach is then to hide no needed columns with https://doc.qt.io/qt-6/qtableview.html#setColumnHidden

phw (changes requested):
This is great. Worked far in my testing. Only some comments about how to handle the different table views.

twodoorcoupe:
> Maybe the better approach is then to hide no needed columns

I removed the extra classes and made it this way, thank you for the tip!

Also, let me know if the column titles and tooltips make sense now, I wasn't really sure about those.

zas:
When files aren't yet in the right column, the display is a bit misleading:

image

phw (approved):
This looks good now, thanks for the update

> Also, let me know if the column titles and tooltips make sense now, I wasn't really sure about those.

The tooltips I think explain this very well. About the columns headers to be honest I'm unsure myself. I find both "New Tags Cover" and "New External Cover" a bit awkward. But I also have no better idea, and I think together with the tooltips it works well enough.

@rdswift what is your opinion on this?

zas:
> I find both "New Tags Cover" and "New External Cover" a bit awkward. But I also have no better idea,

Me too. Also we use same column title for different cases. For example, when a file was just loaded but not matched.

rdswift:
Maybe something like "New Cover Art Embedded" and "New Cover Art File Export"?

twodoorcoupe:
Thank you @rdswift for the names suggestions!

> When files aren't yet in the right column, the display is a bit misleading

> Also we use same column title for different cases. For example, when a file was just loaded but not matched.

Given this, I think there are 3 cases for the `ArtworkTable`.

1. Pre-existing cover with a new one downloaded
file_with_new_cover

2. No pre-existing cover with a new one downloaded
file_without_cover

3. Either no new cover art downloaded, or the new cover art was the same as the previous (like for golden albums) and it's not being saved to a separate file
file_with_cover
golden_album

Then for cases 1 and 2 whichever target is not used, its column gets hidden.

zas (approved):
LGTM

phw (approved):
Looks great

PR #2517, opened 2024-06-25
PICARD-2935: apply genre filters and threshold before selecting minimal usage

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2935

The genre filter options can give unexpected results, where less genres are used then the maximum number of genres configured, even though more genres are available that would match the criteria. There are two distinct issues:

1. The genre list is reduced to the maximum allowed number of items before filters are applied.
2. The max. count (used to filter out genres by minusage threshold) is taken over the entire list. It better should ignore filtered out genres.

**Examples case 1:**

You have the following tags with counts:<br>
Pop: 5, Blues: 7, Rock: 5, Psychedelic: 4
You have a filter "-pop"<br>
Number of genres is limited to max. 3.

Picard takes the first most used tags (Blues, Pop, Rock), then applies the filters.

Result: *Blues, Rock*

Better would be to ignore the filtered genre and use: *Blues, Rock, Psychedelic*

 
**Example case 2**

You have the following tags with counts:<br>
Pop: 5, Blues: 11, Rock: 20, Psychedelic: 9
You have a filter "-rock"<br>
The threshold is set to 50%.

Picard takes the first most used tags (Rock, Blues, Pop), then applies the filters and the threshold. For the threshold the most used genre (Rock=20) is used as the reference max. count, even if it gets removed.

Result:  *Blues*

Better would be to ignore the filtered genre. Then the max. count would be Blues=11 and the result be: *Blues, Psychedelic*

# Solution

Apply the filters first, then the minusage threshold on the filtered result. Take max. no. of genres from the resulting list.

# Action

Additional actions required:

### DIFF
--- a/picard/track.py
+++ b/picard/track.py
@@ -110,10 +110,17 @@ class TagGenreFilter:
                 return True
         return False
 
-    def filter(self, counter):
-        for name, count in counter:
+    def filter(self, counter: Counter, minusage=0) -> Counter:
+        result = Counter()
+        for name, count in counter.items():
             if not self.skip(name):
-                yield (name, count)
+                result[name] = count
+        topcount = result.most_common(1)[0][1]
+        for name, count in counter.items():
+            percent = 100 * count // topcount
+            if percent < minusage:
+                del result[name]
+        return result
 
     def format_errors(self):
         fmt = _("Error line %(lineno)d: %(error)s")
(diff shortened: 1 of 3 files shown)


zas (approved):
Good fix! LGTM

phw:
It looses a bit the elegance of the generator implementation and I think we can't avoid iterating the list multiple times now. But this logic works much better :) I now need to retag some files.

PR #2520, opened 2024-06-28
Improve resizing options page UI

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
Improved the image resizing options page ui as suggested by @Aerozol in #2515.

# Problem

# Solution

Now the new ui looks like this:

image

# Action

Additional actions required:


rdswift (commented):
Overall, this is a nice improvement.

I still have a concern that we are inconsistent in the way that missing / unselected scaling dimensions are handled between the various processing options.  For example, with the "Crop to fit" and "Stretch to fit" methods an unspecified dimension is taken to be that dimension from the source image, while an unspecified dimension in the "Maintain aspect ratio" method is taken to be an infinitely large number.  This inconsistency could lead to confusion and unexpected results, and thus needs to be made very clear to the user (in my opinion).  I'm not sure that the examples in the processing option tooltips make that inconsistency sufficiently clear.

Personally, I think having to specify target dimensions for both directions in all cases is much cleaner.  In each processing case, the final image would fit within the target dimensions specified.

rdswift:
Just to clarify my earlier comment a bit, typically when a program asks for target dimensions for an image it is both dimensions (at least in my experience) and specifies the (maximum) canvas size into which the final image is placed.  Depending on the scaling process used, one (or both) of the final dimensions of the resulting image may be less than the corresponding target dimension.  To only specify one dimension is unusual.

rdswift:
> I understand. I guess I should probably wait for @zas' opinion on this, since he was the one who initially proposed allowing to ignore one of the dimensions.

I'll support whatever the consensus is.  I was just expressing my opinion.  I think that the functionality that you've added to Picard in terms of image processing is a great improvement, and will be well received by the community.  I want to let you know that, to help offset some of the frustration you might be feeling with my UI comments.  Thanks for all your work on this.

rdswift:
@twodoorcoupe,

I hacked together a few suggestions regarding my comment above (and a few other things) into a commit on my fork of the repository: https://github.com/metabrainz/picard/commit/7236a8e4c3f9267f0a36e6ae390e4c85b2284905

Feel free to use as much or as little as you want from it.  I actually tried to submit it as a pull request to your fork of the repository, but for some unknown (to me) reason, your fork doesn't appear in the list of repositories on the pull request form.

Aerozol:
Great work!
I think there are two changes to be made, the first of which may address one of @rdswift's concerns;
- Width and height should be called 'Maximum width' and 'Maximum height', and should be radio buttons, not checkboxes (only one should be selectable, if the ratio is being maintained*)
- Rather than 'don't enlarge', let's go with 'Enlarge' and default to unchecked (Currently it can be translated to "Do don't enlarge" haha)
*When 'stretch' or 'crop' is selected in the drop down, then this menu is valid (width and height should both be set).
Hopefully I've imagined everything correctly!

rdswift:
I agree that perhaps identifying them as "Maximum" dimensions hight help clarify.  In addition, it may be as simple as including a statement on the settings page that says something like:

---

***Note that if a maximum size is not specified for a horizontal or vertical dimension:***
- ***in the case of the "Crop to fit" and "Stretch to fit" modes, the size of the source image for that dimension will be used;***
- ***in the case of the "Maintain aspect ratio" mode, resizing will not be constrained in that dimension.***

---

zas:
If having one dimension set is confusing, I have no issue with getting rid of it.
Combination of source image proportions and current options leads to a high number of cases.

Changes suggested by @rdswift in https://github.com/metabrainz/picard/commit/337e80e3b73d2cc9dea56ba3ead5188df2aa77e5 look good to me.

Aerozol:
I suspect that only implementing part of my suggestion (the drop downs), without changing the menus, has ended up not helping all that much. This is a full mockup with all three menus, where I feel the UI is clear:

image

Now that I think about it though, it is could be useful to specify max height _or_ width (I wonder why Adobe has just gone with one):

image

I don't fully follow the discussion re. "taking the size from the source image". I assume my mockups here aren't taking that into account - I would expect both sizes to have to be filled out, for crop and stretch.

rdswift:
> I don't fully follow the discussion re. "taking the size from the source image". I assume my mockups here aren't taking that into account - I would expect both sizes to have to be filled out, for crop and stretch.

That was my main point.  The current UI only requires you to specify one dimension (optionally leaving the other dimension unspecified) for all three processing cases ("crop", "stretch", and "maintain aspect ratio").  As I understand it, if only one dimension is specified, the other dimension for the "crop" and "shrink" options is taken to be that dimension from the source image.  To illustrate, if you have a source image that is 400x200 and the user has specified either "crop" or "stretch" processing but has only provided a target horizontal size as 100, the current processing will take the vertical size of the source image (200) as the target vertical size.  With the original image being:

image

the resulting 100x200 target image for the "crop" processing would be:

image

Using the same settings and source image but changing the processing to "stretch", the resulting 100x200 target image would be:

image

If the scaling option is "maintain aspect ratio" and only one target dimension is specified, the other target dimension is ignored (as an infinitely large number).  This is using a different default for a missing dimension than that used for the "crop" and "stretch" cases, but is really only evident/relevant if an image is being scaled up.  Using our original 400x200 image and only specifying a horizontal target dimension of 100, the resulting 100x(infinity) image would look like:

image

My thinking is that the user needs to understand what's happening and what defaults are being applied for missing dimensions, so that they don't end up with unexpected results.  Another option is to always require both target dimensions to be specified if the "crop" or "stretch" processing option is selected.

Aerozol:
Thanks @rdswift. I _think_ the confusion is solved in my full mockups?

e.g. Makes it clear that only one dimension can be followed with 'maintain aspect ratio', and does away with setting a dimension to null for 'crop' and 'stretch'. I don't know if the functionality that is lost by removing the null dimension (thanks for the screenshots!) has a use case.

rdswift:
> Makes it clear that only one dimension can be followed with 'maintain aspect ratio',

But you ***can*** currently set targets for both dimensions for "maintain aspect ratio", and it can make sense to do so.  For example, if you want your final image to keep the same aspect ratio, and to be no larger than 1000 in either direction, but ensure that at least one of the dimensions is no smaller than 1000:

| Source | Target H | Target V | Final Image | Comment|
| :-: | :-: | :-: | :-: | :-: |
| 2000x1000 | 1000 | Not Set | 1000x500 | Okay |
| 2000x1000 | Not Set | 1000 | 2000x1000 | Too large horizontally |
| 2000x1000 | 1000 | 1000 | 1000x500 | Okay |
| 1000x2000 | 1000 | Not Set | 2000x1000 | Too large vertically |
| 1000x2000 | Not Set | 1000 | 500x1000 | Okay |
| 1000x2000 | 1000 | 1000 | 500x1000 | Okay |

This could be the use case that makes sense of allowing both dimensions to be specified for the "maintain aspect ratio" option, as @twodoorcoupe has it currently implemented.  I think it would be a shame to remove this.

The way I look at this is that the target dimensions specified define the (maximum) size of the frame into which the output image must fit.  If the mode is "maintain aspect ratio" it is okay to remove one of the bounds because the other bound limits the size.  In both the "crop" and "shrink" cases, both dimensions need to be defined because they don't have the benefit of a constant aspect ratio to determine the missing dimension.

My point is that if the user isn't specifying these dimensions, they need to clearly understand how the missing dimension will be determined.  Currently the missing dimensions are determined by using the corresponding dimension from the source image, over which the user generally has no control, so they may result in an unexpected final product.  Personally, I'd like to force the user to specify both dimensions when using the "crop" and "shrink" modes, rather than automatically selecting a default value outside of the user's control.

At the same time, I can understand the simplicity of only specifying one dimension in the "maintain aspect ratio" scenario if the the user doesn't care about the final size of the other dimension (which could be the case for an image of booklet pages where perhaps only a height is specified to accommodate images that are different width such as one page vs two or more pages in the same image).

All of this tremendous flexibility already exists in @twodoorcoupe's code.  I would just like to find a way to make it clear to the user how the settings are used (or derived) so that they can make intellegent choices and not end up with unexpected results.

At the very least this is looking like it's going to fill an entire section of the Picard User Guide, which unfortunately a very small percentage of the users actually read.

rdswift:
Another thought...  If we want to keep the option of not specifying one of the dimensions, perhaps when that setting is unchecked the spinbox can be hidden and replaced with a label stating something like "***Dimension not considered***" for the "maintain aspect ratio" option, and "***Taken from source image***" for the "crop" and "stretch" options.  That still allows the flexibility, but makes it clear to the user what to expect.

twodoorcoupe:
Thank you so much for your help folks, especially @rdswift!

Regarding the optional dimensions, if we want to allow to ignore a dimension only when aspect ratio is being kept, maybe it's more intuitive to consider them as separate resize modes, like "Scale to width" and "Scale to height".

I think this simplifies both the ui and the code. Here's an example of what I mean:

Pasted image

Then, when either scale to width or scale to height is selected, the ignored dimension is disabled:

image

Otherwise, I can try @Aerozol's suggestion and remove the width and height checkboxes when the resize mode is crop or stretch.

Aerozol:
I think having separate options for 'scale to width' and 'scale to height' is quite an elegant solution - anyone confused by the complexity of the 'maintain aspect ratio' setting can select those, and rest easy.

Leaving the greyed out options there presumably gives quite a smooth transition between settings, as well? e.g. no change in positioning for the options. Probably pointing out the obvious, but for the greyed out options I would replace '1000' with nothing, or maybe a dash.

twodoorcoupe:
> Leaving the greyed out options there presumably gives quite a smooth transition between settings, as well? e.g. no change in positioning for the options

Yes this is the case

> Probably pointing out the obvious, but for the greyed out options I would replace '1000' with nothing, or maybe a dash.

Done. I preferred making them show nothing as the code was much easier

zas (changes requested):
It looks good, just few comments.

twodoorcoupe:
> The only (minor) thing that I throw out there for consideration is to possibly show the "Scale to width" and "Scale to height" options indented or proceeded with a hyphen or something to indicate that they are subset options of the "Maintain aspect ratio" option.

Not sure how to do this properly, I think adding spaces or tab characters might not always align the indented options correctly due to the font not being mono spaced. Maybe I could use an em space with `\u2003` instead? 

This is what it looks like as of now with space, hyphen, space.

image

zas:
IMHO, formatting / indenting to indicate the option preserves aspect ratio is a bad idea, and it may not translate very well.

What about using a radio button instead and changing the list:

[x] Preserve aspect ratio
[Scale / Scale to width / Scale to height]

[ ]  Preserve aspect ratio
[Crop to fit / Stretch to fit]

Or even radio buttons in one group but displayed in 2 sets:

Aspect ratio preserved:
 [ ] Scale    [ ] Scale to width     [X] Scale to height

Aspect ratio not preserved:
 [ ] Crop to fit   [ ] Stretch to fit


Dunno, the indentation approach doesn't look right to me.
@Aerozol what do you think?

twodoorcoupe:
> IMHO, formatting / indenting to indicate the option preserves aspect ratio is a bad idea, and it may not translate very well.

Maybe it's best to just keep the drop down list and not indent anything.

If it's not clear enough that those two options keep the aspect ratio, maybe we can find different names like "scale width (keep ratio)" or something more explicit. There's already the tooltip that explains what the option does as well.

zas:
> Maybe it's best to just keep the drop down list and not indent anything.

Yes, likely.

zas (approved):
LGTM

rdswift:
> Maybe it's best to just keep the drop down list and not indent anything.

I'm okay with that.  It was only a minor suggestion.  Certainly not a show-stopper.  For most people it will be a "set it once and forget it" option anyway, and we can explain it thoroughly in the documentation (that nobody reads).

phw (approved):
@twodoorcoupe  This UI is a great improvement. This addresses my concerns about the complexity, while allowing even more options.

Thanks @rdswift and @Aerozol also.

PR #2522, opened 2024-07-08
Update caa types selector to allow replacement instructions text

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: This change allows the default instructions text to be overridden in the CAA Types Selector dialog.  This allows the dialog to be used in other cases such as the new cover art processing settings.

# Problem

The original CAA Types Selector dialog has fixed instructions, and as such cannot be reused for other (similar) cases.

# Solution

This change allows the default instructions text to be overridden in the CAA Types Selector dialog.  This allows the dialog to be used in other cases such as the new cover art processing settings.

# Action

Additional actions required:

### DIFF
--- a/picard/ui/caa_types_selector.py
+++ b/picard/ui/caa_types_selector.py
@@ -11,7 +11,7 @@
 # Copyright (C) 2015-2016 Rahul Raturi
 # Copyright (C) 2016-2017 Sambhav Kothari
 # Copyright (C) 2017 Frederik “Freso” S. Olesen
-# Copyright (C) 2018 Bob Swift
+# Copyright (C) 2018, 2024 Bob Swift
 # Copyright (C) 2018 Vishal Choudhary
 #
 # This program is free software; you can redistribute it and/or
@@ -164,6 +164,8 @@ class CAATypesSelectorDialog(PicardDialog):
         types_include {[string]} -- List of CAA image types to include (default: {None})
         types_exclude {[string]} -- List of CAA image types to exclude (default: {None})
         parent {[type]} -- Parent of the QDialog object being created (default: {None})
+        instructions_top {string} -- Replacement for the default instruction text to display above the list boxes (default: {None})
+        instructions_bottom {string} -- Replacement for the instruction text to display between the list boxes and the button bar (default: {None})
     """
 
     help_url = 'doc_cover_art_types'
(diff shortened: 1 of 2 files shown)


rdswift:
@zas,

If we decide to incorporate these changes, should I move the instructions text to be part of the call in `/picard/coverart/providers/caa.py` and make the arguments non-optional in `/picard/ui/caa_types_selector.py`?  I like the idea of having the defaults as I've shown it, but I'm okay with changing it if you think it's more appropriate.

picard/ui/caa_types_selector.py
```diff
@@ -201,7 +205,10 @@ def __init__(
 
         # Add instructions to the dialog box
         instructions = QtWidgets.QLabel()
-        instructions.setText(_("Please select the contents of the image type 'Include' and 'Exclude' lists."))
+        if instructions_top:
+            instructions.setText(_(instructions_top))
+        else:
+            instructions.setText(_("Please select the contents of the image type 'Include' and 'Exclude' lists."))
```
zas, line 211: Just a minor thing, but instead repeating calls we may just do:

```python
        if instructions_top is None:
           instructions_top = N_("Please select the contents of the image type 'Include' and 'Exclude' lists.")
        instructions.setText(_(instructions_top))
```

picard/ui/caa_types_selector.py
```diff
@@ -244,13 +251,16 @@ def __init__(
...
+        if instructions_bottom:
+            instructions.setText(_(instructions_bottom))
+        else:
+            instructions.setText(_(
+                "CAA images with an image type found in the 'Include' list will be downloaded and used "
+                "UNLESS they also have an image type found in the 'Exclude' list. Images with types "
+                "found in the 'Exclude' list will NEVER be used. Image types not appearing in the 'Include' "
+                "or 'Exclude' lists will not be considered when determining whether or not to download and "
+                "use a CAA image.\n")
+            )
```
zas, line 263: Same comment as above.

zas (changes requested):
The move looks good to me, just minor stylistic comment.

picard/ui/caa_types_selector.py
rdswift: Great suggestion.  Thanks.

zas (approved):
LGTM

PR #2525, opened 2024-07-15
Make image processing multithreaded

twodoorcoupe:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
Makes Picard run image processors in different threads.

# Problem

# Solution

The main advantage of this is that when there are multiple images in the `CoverArt`'s queue, the next image's download can start while the previous is still processing.

Then once all images have been downloaded, the `Album` waits for all images to finish processing before finalizing loading.

I'm still working to make sure exceptions are handled correctly in the threads.

# Action

Additional actions required:

### DIFF
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -328,10 +328,10 @@ class CoverArtImage:
     def has_finished_processing(self):
         if not self.datahash:
             return False
-        config = get_config()
-        if (config.setting['save_images_to_files'] and not
-                (self.external_file_coverart and self.external_file_coverart.datahash)):
-            return False
+        if not (self.external_file_coverart and self.external_file_coverart.datahash):
+            config = get_config()
+            if config.setting['save_images_to_files']:
+                return False
         return True
 
     @property
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -325,16 +325,6 @@ class CoverArtImage:
         self.external_file_coverart = CoverArtImage(data=data, url=self.url)
 
     @property
-    def has_finished_processing(self):
-        if not self.datahash:
-            return False
-        if not (self.external_file_coverart and self.external_file_coverart.datahash):
-            config = get_config()
-            if config.setting['save_images_to_files']:
-                return False
-        return True
-
-    @property
     def maintype(self):
         """Returns one type only, even for images having more than one type set.
         This is mostly used when saving cover art to tags because most formats
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -325,6 +325,16 @@ class CoverArtImage:
         self.external_file_coverart = CoverArtImage(data=data, url=self.url)
 
     @property
+    def has_finished_processing(self):
+        if not self.datahash:
+            return False
+        config = get_config()
+        if (config.setting['save_images_to_files'] and not
+                (self.external_file_coverart and self.external_file_coverart.datahash)):
+            return False
+        return True
+
+    @property
     def maintype(self):
         """Returns one type only, even for images having more than one type set.
         This is mostly used when saving cover art to tags because most formats
--- a/picard/util/imagelist.py
+++ b/picard/util/imagelist.py
@@ -24,7 +24,6 @@
 
 
 from collections.abc import MutableSequence
-import time
 
 from picard.config import get_config
 
@@ -113,7 +112,3 @@ class ImageList(MutableSequence):
                     continue
             types_dict[image_types] = image
         return types_dict
-
-    def wait_for_image_processing(self):
-        while not all(image.has_finished_processing for image in self._images):
-            time.sleep(0.1)
--- a/picard/util/imagelist.py
+++ b/picard/util/imagelist.py
@@ -24,6 +24,7 @@
 
 
 from collections.abc import MutableSequence
+import time
 
 from picard.config import get_config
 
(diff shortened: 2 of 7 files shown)


picard/coverart/image.py
```diff
@@ -324,6 +324,16 @@ def set_tags_data(self, data):
     def set_external_file_data(self, data):
         self.external_file_coverart = CoverArtImage(data=data, url=self.url)
 
+    @property
+    def has_finished_processing(self):
+        if not self.datahash:
+            return False
+        config = get_config()
+        if (config.setting['save_images_to_files'] and not
+                (self.external_file_coverart and self.external_file_coverart.datahash)):
```
zas, line 333: Perhaps it would be better to use call `get_config()` only if strictly needed and do:

```python
if not (self.external_file_coverart and self.external_file_coverart.datahash):
    config = get_config()
    if config.setting['save_images_to_files']:
        return False
return True
```

twodoorcoupe:
Still not ready, as I'm scratching my head on how to deal with the exceptions raised by `coverartimage.set_tags_data()` in a nice way

picard/util/imagelist.py
```diff
@@ -112,3 +113,7 @@ def get_types_dict(self):
                     continue
             types_dict[image_types] = image
         return types_dict
+
+    def wait_for_image_processing(self):
```
phw, line 117: Ideally this should be callback based and not wait with polling.

But currently the threading tools provided by Picard do not make that easy. The most common case in Picard is either running a single thread then callback or in some cases a sequential queue of things to run in a thread one after another.

I think we need some general way to run code after a bunch of threads has finished. @zas, what do you think?

In any way we could do it with the wait and status polling as part of this PR and refactor to a different mechanism later.

Sophist-UK: IMO there are a bunch of things that should be provided as general methods, and an improved threading approach is one of them, generalised caching being another.

zas: > Unless it is impossible to achieve, polling should be avoided at all costs IMO.

Can you elaborate on this?
What alternatives do you suggest?

Sophist-UK: Sorry - bit it is so long since I looked at the multi-threading code that I am unable to make any suggestions.

From memory of c. a decade and a half ago, I felt that the way that stuff was passed to and from a thread was wrong and that queues (and possibly signals) would be a better architecture than the current one (or rather the one at that time).

Plus queues gives a common approach for both multithreading **and** multi-processing.

As for generic methods, examining the code of existing plugins for stuff that is plumbing (that is making an individual plugin much more complicated than it really should need to be and / or which is duplicated across >= 2 plugins) rather than direct functionality might well be a decent way to establish what generic code is needed.

twodoorcoupe:
> Ideally this should be callback based and not wait with polling.

> Unless it is impossible to achieve, polling should be avoided at all costs IMO.

I understand, for now I changed it so that each `CoverArt` class has its own thread for image processing, and `album.finalize_loading()` is called only after each processor has finished, without needing to poll.

My goal is to eventually have the same `QThreadPool` be shared among all `CoverArt`s, and not having one thread for each. I think I can get something to work using callbacks, I can work on making a general implementation in a later pr.

---

Regarding exceptions, at first I tried using callbacks through the `error` argument, but this created a race condition when errors needed to be appended to albums. Since callbacks run on the main thread, `QThreadPool.waitForDone()` does not consider them.

For now there's a decorator function handling these exceptions, but once I settle the "wait for a set of threads to finish" business, I can replace this.

Last thing. Before, when an image could not be saved to the temporary folder, `CoverArt` would not call `next_in_queue()` any longer. Since now images are saved in different threads, I can't catch this exception as before.

The only way I could think of for achieving a similar behaviour is to stop if the album has already finished loading when `next_in_queue()` is called, because when an image can't be saved to the temporary folder, the album finalizes loading with `error=True`. But, I'm still not sure this is the best approach.

test/test_coverart_processing.py
```diff
@@ -300,8 +303,22 @@ def test_format_conversion(self):
...
-        image, info = create_fake_image(0, 0, 'jpg')
+    def _check_processing_error(self, image, info):
+        self.set_config_values(self.settings)
         coverartimage = CoverArtImage()
-        with self.assertRaises(CoverArtProcessingError):
-            run_image_processors(coverartimage, image, info)
+        album = Album(None)
+        image_processing = CoverArtImageProcessing(album)
+        image_processing.run_image_processors(coverartimage, image, info)
+        image_processing.threadpool.waitForDone()
```
zas, line 312: Perhaps wrap this in a `CoverArtImageProcessing` method?

Something like:

```python
        image_processing = CoverArtImageProcessing(album)
        image_processing.run_image_processors(coverartimage, image, info)
        image_processing.wait_for_done()
```

zas (approved):
The patch looks good to me, I tested and didn't find any showstopper.
There's perhaps room for improvements, of course, but the current code looks functional enough to me.

phw (approved):
Tested this. Patch looks good for what it does and works as expected. LGTM

Sophist-UK:
I haven't had time to read and digest this code, and AFAICS the image processing code itself is in a separate plugin (and so outside the control of this code), so I have no idea whether the image processing code is going to be CPU intensive (or CPU only) or whether it will be I/O bound.

The reason I mention this is that the GIL implementing in cPython does NOT handle multiple CPU-bound threads (or sometimes even one such thread) particularly well, and this can lead to GIL starvation of the main thread and thus GUI unresponsiveness.

You may therefore wish to add a microsecond sleep to the iterative code so that the GIL is released for a very short period to allow the UI to get some CPU.

(I can reasonably claim some expertise in this area - see https://github.com/faster-cpython/ideas/discussions/328.)

PR #2532, opened 2024-08-04
PICARD-2914: Workaround for segfaults on macOS with Spanish locale

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2914

Picard on macOS segfaults semi-randomly when sorting the release countries / formats locale aware in the preferred releases options. The segfault happens when the key-function used for sorting is calling `strxfrm`.

I could not narrow it down to any specific value. When reducing the number of items sorted often the crash could be avoided. Calling strxfrm on all individual values on itself is fine. With some luck the segfault did not happen at all.

I could not reproduce the crash with other locales like en, fr or de, but I haven't tested all.

# Solution

As a workaround avoid calling strxfrm and use strcoll for locale aware sorting. In my testing this never triggered the segfault. But this should be confirmed by others.

The PR also applies some refactoring in 4af20ad4af40b35cd057fb999e2b05b67485e48b to simplify the code.

### DIFF
--- a/picard/ui/options/releases.py
+++ b/picard/ui/options/releases.py
@@ -24,7 +24,11 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from functools import partial
+from functools import (
+    cmp_to_key,
+    partial,
+)
+from locale import strcoll
 
 from PyQt6 import (
     QtCore,
@@ -47,7 +51,6 @@ from picard.i18n import (
     gettext_countries,
     pgettext_attributes,
 )
-from picard.util import strxfrm
 
 from picard.ui.forms.ui_options_releases import Ui_ReleasesOptionsPage
 from picard.ui.options import OptionsPage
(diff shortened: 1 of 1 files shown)


phw:
I created a 2.12.1 test build with the patch attached and asked affected users for testing at https://community.metabrainz.org/t/picard-crashes-when-open-preferences/564794/11

phw:
If I understand the discussion at https://community.metabrainz.org/t/picard-crashes-when-open-preferences/564794/12 correctly the issue seems not be solved for the user there.

I can check whether there are other uses of strxfrm that might cause this and replace them as well. Or this entire thing is a red herring and crashes are happening due to different reasons. But the above and the workaround is all I was able to reproduce on my own system.

phw:
According to the forums discussions this didn't work. But avoiding strxfrm by using PyICU did. The test results are interesting, I'm working on a new patch with a completely different approach.

phw:
This was resolved by the better approach in #2533. I'll resubmit the refactoring from the first commit as a spearate PR

PR #2533, opened 2024-08-06
PICARD-2914: Replace strxfrm with QCollator

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2914

On macOS Picard is experiencing segfaults related to Pythons locale.strxfrm function. We had similar issues in the past.

Testing on https://community.metabrainz.org/t/picard-crashes-when-open-preferences-with-spanish-ui-language-on-macos/564794/ showed that replacing strxfrm with PyICU did resolve the crashes.

# Solution

Replace all uses of `locale.strxfrm` with Qt's `QCollator.sortKey`. Introduce a new function `picard.18n.sort_key` for wrapping this.

Qt is already using ICU and provides own wrappers around this. This allows us to use the ICU methods without the overhead of importing and packaging PyICU additionally.

As a result the `picard.util.natsort` module became obsolete and has been removed.

# Actions

This PR is for the 2.x branch (as I tested close to current release builds). Port over for Picard 3.


phw:
Fixed Windows, but still need to check what is going wrong on macOS.

zas (approved):
LGTM

phw:
@zas This is I think now ready for review. The change totally gets rid of strxfrm. I think one upside of having a `i18n.sort_key` function is also that we can change the implementation at any time.

A couple of additional tweaks / workarounds were necessary in order to maintain full functionality across all operating systems. I really wanted the sorting to continue sorting decimal numbers from e.g. Arabic script properly as well, so Arabic four "٤" and Latin variant "4" sort properly. On Linux this worked by default, on macOS and Windows it needs some help converting the numbers.

Also on macOS there is something that looks like a bug: Numeric natural sorting works in general, so e.g. titles like "1 Track" and "002 Track" sort accordingly. But if the string consists only of numeric characters the it breaks, e.g. "002" sorts then before "1". The simple workaround is to always prepend an alphabetic character. I chose to prepend an "a".

I tested on all Linux, Windows and macOS. Sorting works as expected for me and no crashes anymore on macOS.

PR #2541, opened 2024-09-01
PICARD-2960: Prevent dir separator replacement being set to dir separator

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2960

It must not be possible to configure the replacement for directory separators to be a separator itself.

# Solution

Add a validator to the dir separator option input to disallow `os.sep` and `os.altsep`. Also validate on form close. While theoretically users can no longer configure this it can happen on upgrade and also when moving config from a platform without `os.altsep`  (e.g. Linux) to one with (Windows where altsep is a backslash), or in general if separators differ between platforms.


phw:
I added a config upgrade hook to sanitize the option

zas (approved):
LGTM

PR #2544, opened 2024-10-09
PICARD-2989: fix settings override recursion

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2989

The `ScriptEditorDialog` uses `SettingsOverride` for locally changed settings. On repeated example file updates the `SettingsOverride` would recursively reference itself. As there are many example updates on selection change this could lead to a deep recursion depth until max recursion would trigger an error. 

# Solution

Fix SettingsOverride always referencing base settings only.

Also avoid example update calls if the script editor dialog is not visible. The edittor will update anyway once it becomes visible or gains focus.


zas (approved):
LGTM

PR #2545, opened 2024-10-19
PICARD-2607: Add a “_genre” variable

rakim0:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
Adds a custom field for folksonomy tags and genre.
# Problem
There was no custom field for folksonomy tags. User was forced to either set all genre as genre tags or all folksonomy tags as Genre. 
This PR aims to resolve it by introducing two new tags _folksonomy_tag and _genre.  

* JIRA ticket (_optional_): PICARD-2607

# Solution

To get both _genres and _folksonomy tags from the MusicBrainz server, I removed the 'use_folksonomy` check in picard/items.py and stored the tags in respective variables. Then for writing the metadata, I added two new functions in picard/track.py that would update the metadata with _genre and _folksonomy_tag. And before writing genre into the file, check whether genre should be set to _folksonomy_tag or _genre itself.   

Check for 'Use only my genre' option is still done in picard/items.py
# Action

Additional actions required:

### DIFF
--- a/picard/item.py
+++ b/picard/item.py
@@ -200,6 +200,7 @@ class MetadataItem(QtCore.QObject, Item):
         self.iter_children_items_metadata_ignore_attrs = {}
         self.suspend_metadata_images_update = IgnoreUpdatesContext()
         self._genres = Counter()
+        self._folksonomy_tags = Counter()
 
     @property
     def tagger(self):
@@ -314,13 +315,20 @@ class MetadataItem(QtCore.QObject, Item):
         require_authentication = False
         config = config or get_config()
         if config.setting['use_genres']:
-            use_folksonomy = config.setting['folksonomy_tags']
             if config.setting['only_my_genres']:
                 require_authentication = True
-                inc |= {'user-tags'} if use_folksonomy else {'user-genres'}
+                inc |= {'user-tags', 'user-genres'}
             else:
-                inc |= {'tags'} if use_folksonomy else {'genres'}
+                inc |= {'tags', 'genres'}
         return require_authentication
+
+    @property
+    def folksonomy_tags(self):
+        return self._folksonomy_tags
+
+    def add_folksonomy_tag(self, name, count):
+        if name:
+            self._folksonomy_tags[name] += count
 
 
 class FileListItem(MetadataItem):
--- a/picard/item.py
+++ b/picard/item.py
@@ -310,6 +310,14 @@ class MetadataItem(QtCore.QObject, Item):
         if name:
             self._genres[name] += count
 
+    @property
+    def folksonomy_tags(self):
+        return self._folksonomy_tags
+
+    def add_folksonomy_tag(self, name, count):
+        if name:
+            self._folksonomy_tags[name] += count
+
     @staticmethod
     def set_genre_inc_params(inc, config=None):
         require_authentication = False
@@ -321,14 +329,6 @@ class MetadataItem(QtCore.QObject, Item):
             else:
                 inc |= {'tags', 'genres'}
         return require_authentication
-
-    @property
-    def folksonomy_tags(self):
-        return self._folksonomy_tags
-
-    def add_folksonomy_tag(self, name, count):
-        if name:
-            self._folksonomy_tags[name] += count
 
 
 class FileListItem(MetadataItem):
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -636,16 +636,14 @@ def add_secondary_release_types(node, m):
 def add_genres_from_node(node, obj):
     if obj is None:
         return
-    config = get_config()
-    if config.setting['use_genres']:
-        if 'tags' in node:
-            add_tags(node['tags'], obj)
-        if 'user-tags' in node:
-            add_user_tags(node['user-tags'], obj)
-        if 'genres' in node:
-            add_genres(node['genres'], obj)
-        if 'user-genres' in node:
-            add_user_genres(node['user-genres'], obj)
+    if 'tags' in node:
+        add_tags(node['tags'], obj)
+    if 'user-tags' in node:
+        add_user_tags(node['user-tags'], obj)
+    if 'genres' in node:
+        add_genres(node['genres'], obj)
+    if 'user-genres' in node:
+        add_user_genres(node['user-genres'], obj)
 
 
 def add_genres(node, obj):
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -636,14 +636,16 @@ def add_secondary_release_types(node, m):
 def add_genres_from_node(node, obj):
     if obj is None:
         return
-    if 'genres' in node:
-        add_genres(node['genres'], obj)
-    if 'tags' in node:
-        add_genres(node['tags'], obj)
-    if 'user-genres' in node:
-        add_user_genres(node['user-genres'], obj)
-    if 'user-tags' in node:
-        add_user_genres(node['user-tags'], obj)
+    config = get_config()
+    if config.setting['use_genres']:
+        if 'tags' in node:
+            add_tags(node['tags'], obj)
+        if 'user-tags' in node:
+            add_user_tags(node['user-tags'], obj)
+        if 'genres' in node:
+            add_genres(node['genres'], obj)
+        if 'user-genres' in node:
+            add_user_genres(node['user-genres'], obj)
 
 
 def add_genres(node, obj):
@@ -654,6 +656,16 @@ def add_genres(node, obj):
 def add_user_genres(node, obj):
     for tag in node:
         obj.add_genre(tag['name'], 1)
+
+
+def add_tags(node, obj):
+    for tag in node:
+        obj.add_folksonomy_tag(tag['name'], tag['count'])
+
+
+def add_user_tags(node, obj):
+    for tag in node:
+        obj.add_folksonomy_tag(tag['name'], 1)
 
 
 def add_isrcs_to_metadata(node, metadata):
--- a/picard/track.py
+++ b/picard/track.py
@@ -313,6 +313,8 @@ class Track(FileListItem):
             tm['~silence'] = '1'
 
         if config.setting['use_genres']:
+            self.add_folksonomy_tags()
+            self.add_genres()
             self._convert_folksonomy_tags_to_genre()
 
         # Convert Unicode punctuation
@@ -367,6 +369,18 @@ class Track(FileListItem):
             filters=config.setting['genres_filter'],
             join_with=config.setting['join_genres']
         )
+
+    def add_folksonomy_tags(self):
+        config = get_config()
+        tags = Counter(self._folksonomy_tags)
+        tags += self.album._folksonomy_tags
+        self.metadata['_folksonomy_tags'] = self._genres_to_metadata(tags, join_with=config.setting['join_genres'])
+
+    def add_genres(self):
+        config = get_config()
+        _genres = Counter(self._genres)
+        _genres += self.album._genres
+        self.metadata['_genres'] = self._genres_to_metadata(_genres, join_with=config.setting['join_genres'])
 
 
 class NonAlbumTrack(Track):
--- a/picard/track.py
+++ b/picard/track.py
@@ -314,7 +314,7 @@ class Track(FileListItem):
 
         if config.setting['use_genres']:
             self._convert_folksonomy_tags_to_genre()
-            self._delete_genres_from_tags()
+            self.album.delete_genres_from_tags()
             self._add_folksonomy_tags()
             self._add_genres()
 
@@ -387,10 +387,6 @@ class Track(FileListItem):
         genre = Counter(self.genres)
         genre += self.album.genres
         self._add_tags(genre, '_genres')
-
-    def _delete_genres_from_tags(self):
-        for genre in self.album.genres:
-            del self.album.folksonomy_tags[genre]
 
 
 class NonAlbumTrack(Track):
--- a/picard/track.py
+++ b/picard/track.py
@@ -313,9 +313,8 @@ class Track(FileListItem):
             tm['~silence'] = '1'
 
         if config.setting['use_genres']:
-            config = get_config()
-            self._add_folksonomy_tags(config)
-            self._add_genres(config)
+            self._add_folksonomy_tags()
+            self._add_genres()
             self._convert_folksonomy_tags_to_genre()
 
         # Convert Unicode punctuation
@@ -375,19 +374,18 @@ class Track(FileListItem):
             join_with=config.setting['join_genres']
         )
 
-    def _add_tags(self, tags, name, config=None):
-        config = config or get_config()
+    def _add_tags(self, tags, name):
         self.metadata[name] = tags.keys()
 
-    def _add_folksonomy_tags(self, config=None):
+    def _add_folksonomy_tags(self):
         tags = Counter(self._folksonomy_tags)
         tags += self.album._folksonomy_tags
-        self._add_tags(tags, '_folksonomy_tags', config)
+        self._add_tags(tags, '_folksonomy_tags')
 
-    def _add_genres(self, config=None):
+    def _add_genres(self):
         genre = Counter(self._genres)
         genre += self.album._genres
-        self._add_tags(genre, '_genres', config)
+        self._add_tags(genre, '_genres')
 
 
 class NonAlbumTrack(Track):
--- a/picard/track.py
+++ b/picard/track.py
@@ -389,8 +389,8 @@ class Track(FileListItem):
         self._add_tags(genre, '_genres')
 
     def _delete_genres_from_tags(self):
-        for genres in self.album.genres:
-            del self.album._folksonomy_tags[genres]
+        for genre in self.album.genres:
+            del self.album._folksonomy_tags[genre]
 
 
 class NonAlbumTrack(Track):
--- a/picard/track.py
+++ b/picard/track.py
@@ -352,9 +352,9 @@ class Track(FileListItem):
         genres = Counter(self.genres)
         use_folksonomy = config.setting['folksonomy_tags']
         if use_folksonomy:
-            genres += self.album._folksonomy_tags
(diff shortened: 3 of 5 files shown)


rakim-0:
I tried to follow the Contribution Guidelines as much as possible but was a bit unsure about having multiple commits but couldn't figure a good commit message so broke the commit into 4 parts.

Sophist-UK:
I am really unclear whether this is actually a bug fix. And if it is an enhancement instead, then if it is not going to do something much more generalised, then it least needs to make sure it will be compatible with a future generalised function.

The issue with Genres is that there are so many competing sources of variable quality - many are free-form rather than a fixed list, those that are fixed lists have different fixed lists, and even when there are identical names, the definitions may be different or the allocation subjective. So if you are trying to fill a genre tag from multiple sources, you need to make sense of the mess you may often be given. In addition, there really needs to be needs to be a hierarchy i.e. there are many sub-forms of "Rock", and indeed these need to be many-to-many because there are also fusion genres which are sub-categories of more than one parent category.

So Picard IMO really needs to provide several generalised functions...

1. Putting downloaded genres into separate metadata variables
2. Providing functionality to clean-up these multiple lists of genres
3. Providing functionality to consolidate them down into a single genre list for a track.
4. Providing a plugin category for genre downloading - various types (artist, album, track) from various sources (MB being only one possible source of genre data).

I therefore encourage the authors and reviewers of this PR to think about the above and do their best to at least try to guide this PR into a form that is in the same general direction and which will at a minimum not make delivery of comprehensive genre functionality more difficult and ideally at least takes us a step in that direction.

(See also: Lyric handling which has similar (but simpler) plugin-based requirements.)

picard/track.py
```diff
@@ -368,6 +374,18 @@ def _convert_folksonomy_tags_to_genre(self):
             join_with=config.setting['join_genres']
         )
 
+    def add_folksonomy_tags(self):
+        config = get_config()
+        tags = Counter(self._folksonomy_tags)
+        tags += self.album._folksonomy_tags
+        self.metadata['_folksonomy_tags'] = self._genres_to_metadata(tags, join_with=config.setting['join_genres'])
```
phw, line 381: To my understanding we can do this simpler and don't need to filter the values for `_folksonomy_tags` and `_genres`. We just include all the loaded values in either variable. Only the final `genre` variable gets filtered.

But I asked UltimateRiff on PICARD-2607 for clarification also.

What I'm also unsure about if `_folksonomy_tags` should strictly only contain the non-genre tags. In that case we would filter out everything that is in `_genres`.

picard/track.py
```diff
@@ -368,6 +374,18 @@ def _convert_folksonomy_tags_to_genre(self):
...
+
+    def add_genres(self):
+        config = get_config()
+        _genres = Counter(self._genres)
+        _genres += self.album._genres
+        self.metadata['_genres'] = self._genres_to_metadata(_genres, join_with=config.setting['join_genres'])
```
phw, line 387: As discussed on the chat the two methods here can be simplified. The suggestion you made in the chat seemed fine, just also consider my above comment as there is I think no need to run the lists through the `_genres_to_metadata` filtering.

Also please prefix those methods with an underscore, as I would consider them private for use inside the  `Track` class only.

phw (changes requested):
Thanks a lot for this PR. I initially thought we can handle it more simple inside `mbjson` and directly write the values for `_genres` and `_folksonomy_tags` in the metadadata object there. But your implementation of storing the list with counts is more complex but needed to run the proper filtering later on.

Also tests should be fixed.

phw:
> I therefore encourage the authors and reviewers of this PR to think about the above and do their best to at least try to guide this PR into a form that is in the same general direction and which will at a minimum not make delivery of comprehensive genre functionality more difficult and ideally at least takes us a step in that direction.

@rakim-0 @Sophist-UK We have a separate ticket PICARD-2054 for this, but this needs additional considerations. But it has no effect on this story, which is only about adding the `_genres` and `_folksonomy_tags` variables with data loaded from MB.

Genres for separate sources would be handled by the genre handling functions in `item.py`. Generally we have most functionality available for this, and probably we could change plugins to use `MetadataItem.add_genre` / `MetadataItem.add_folksonomy` tags feature. But for full implementation we should have genre providers similar to cover art providers that can be enabled / disabled. If we do this correctly we could also solve additional tickets, e.g. we could have providers separately for recordings, releases, artist etc. and users could choose.

It's not unlikely that if we implement this the code added here will need some refactoring as well, but that needs to be done when PICARD-2054  actually gets implemented.

picard/mbjson.py
```diff
@@ -636,14 +636,16 @@ def add_secondary_release_types(node, m):
...
-    if 'genres' in node:
-        add_genres(node['genres'], obj)
-    if 'tags' in node:
-        add_genres(node['tags'], obj)
-    if 'user-genres' in node:
-        add_user_genres(node['user-genres'], obj)
-    if 'user-tags' in node:
-        add_user_genres(node['user-tags'], obj)
+    config = get_config()
+    if config.setting['use_genres']:
```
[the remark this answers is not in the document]
rakim-0: Ah ok, got it. Fixed it

picard/item.py
```diff
@@ -314,14 +315,21 @@ def set_genre_inc_params(inc, config=None):
...
         return require_authentication
 
+    @property
+    def folksonomy_tags(self):
+        return self._folksonomy_tags
+
+    def add_folksonomy_tag(self, name, count):
+        if name:
+            self._folksonomy_tags[name] += count
+
```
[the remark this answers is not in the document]
rakim-0: Fixed in cf267f6dde6abb34784c0d4eed14e7af0dc8e674

picard/track.py
rakim-0: Yep, I moved the redundant code to a new function and directly assigned the _folksonomy_tags and _genres
fixed in commit: d543ec0b1277b5e69111195cf19b604dcd789f8a

picard/track.py
rakim-0: Hmm I think _folksonomy_tags should only have the non-genre tags. But well I'll just to see what UltimateRiff expects before I implement it.

For the implementation though, is iterating through _genres.keys() and deleting them from _folksonomy_tags going to be good enough or is there a more efficient method?

rakim-0:
I'll squash the smaller commits after I fix the tests.

rakim-0:
Fixed tests and resolved most comments. 
I'm just unsure about this one: https://github.com/metabrainz/picard/pull/2545#discussion_r1807834324 (filtering out genres from the folksonomy tags).

picard/track.py
rakim-0: @phw [UltimateRiff replied](https://tickets.metabrainz.org/browse/PICARD-2607?focusedId=73844&page=com.atlassian.jira.plugin.system.issuetabpanels%3Acomment-tabpanel#comment-73844):
> on the first expectation, I'm honestly not sure what my preference would be
on the second tho, I think I'd expect _folksonomy_tags to be only non-genre tags


I think for first, including everything is the right way to go for now.    
For the deleting genres from folksonomy tags, is my earlier approach fine? 
> Is iterating through _genres.keys() and deleting them from _folksonomy_tags going to be good enough or is there a more efficient method?

picard/track.py
```diff
@@ -313,6 +313,9 @@ def _customize_metadata(self):
             tm['~silence'] = '1'
 
         if config.setting['use_genres']:
+            config = get_config()
```
zas, line 316: No need to set `config` here, it is already at start of this method.

zas (changes requested):

picard/track.py
rakim-0: fixed in 5b50798

picard/track.py
```diff
@@ -368,6 +375,20 @@ def _convert_folksonomy_tags_to_genre(self):
             join_with=config.setting['join_genres']
         )
 
+    def _add_tags(self, tags, name, config=None):
+        config = config or get_config()
```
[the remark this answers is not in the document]
rakim-0: Nope I don't think so. I initially had it since I was filtering _genres and _folksonomy_tags and had forgotten to remove it.
fixed in 5b50798

phw (dismissed):
Thanks a lot for the updates. This looks good to me and should work.

zas (dismissed):
LGTM, thanks for your contribution.

rdswift:
If you don't mind, can you please enter a ticket to update the Picard documentation to include this new variable?  I suspect that the page that needs to be updated is https://github.com/metabrainz/picard-docs/blob/master/variables/variables_basic.rst

Thanks.

picard/track.py
rakim-0: Implemented deleting tags from genres in commit: 58286b778343fd23f02df682ca3ee2eea42044c6

picard/track.py
```diff
@@ -368,6 +375,23 @@ def _convert_folksonomy_tags_to_genre(self):
...
+        self._add_tags(tags, '_folksonomy_tags')
+
+    def _add_genres(self):
+        genre = Counter(self._genres)
+        genre += self.album._genres
+        self._add_tags(genre, '_genres')
+
+    def _delete_genres_from_tags(self):
+        for genres in self.album.genres:
+            del self.album._folksonomy_tags[genres]
```
rdswift, line 393: This is a really minor nit-picking comment, but should the variable be `genre` rather than `genres`?

```suggestion
    def _delete_genres_from_tags(self):
        for genre in self.album.genres:
            del self.album._folksonomy_tags[genre]
```

rakim-0: Oops. Thanks for pointing it out! Fixed in commit: 3f0a795d8bbceba4d239480a9d38d64cc1bd237e

phw (dismissed):
Thank you again. For me this looks great. Waiting for last review from @zas

picard/track.py
```diff
@@ -346,9 +349,13 @@ def _genres_to_metadata(genres, limit=None, minusage=0, filters='', join_with=No
 
     def _convert_folksonomy_tags_to_genre(self):
         config = get_config()
-        # Combine release and track genres
         genres = Counter(self.genres)
-        genres += self.album.genres
+        use_folksonomy = config.setting['folksonomy_tags']
+        if use_folksonomy:
+            genres += self.album._folksonomy_tags
```
zas, line 355:
```suggestion
            genres += self.album.folksonomy_tags
```

picard/track.py
```diff
@@ -346,9 +349,13 @@ def _genres_to_metadata(genres, limit=None, minusage=0, filters='', join_with=No
...
+        else:
+            genres += self.album._genres
```
zas, line 357:
```suggestion
            genres += self.album.genres
```

test/test_mbjson.py
```diff
@@ -135,11 +135,14 @@ def test_release(self):
         self.assertEqual(m['~release_seriesid'], '7421b602-a413-4151-bcf4-d831debc3f27')
         self.assertEqual(m['~release_seriescomment'], 'Pink Floyed special editions')
         self.assertEqual(m['~release_seriesnumber'], '')
-        self.assertEqual(a.genres, {
-            'genre1': 6, 'genre2': 3,
-            'tag1': 6, 'tag2': 3})
+        self.assertEqual(a._genres, {
```
zas, line 138:
```suggestion
        self.assertEqual(a.genres, {
```

test/test_mbjson.py
```diff
@@ -135,11 +135,14 @@ def test_release(self):
...
+            'genre1': 6, 'genre2': 3
+        })
+        self.assertEqual(a._folksonomy_tags, {
```
zas, line 141:
```suggestion
        self.assertEqual(a.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -135,11 +135,14 @@ def test_release(self):
...
+            'tag1': 6, 'tag2': 3
+        })
         for artist in a._album_artists:
-            self.assertEqual(artist.genres, {
+            self.assertEqual(artist._folksonomy_tags, {
```
zas, line 145:
```suggestion
            self.assertEqual(artist.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -166,11 +169,14 @@ def test_release_without_release_relationships(self):
         self.assertEqual(m['~albumartists_sort'], 'Pink Floyd')
         self.assertEqual(m['~releaselanguage'], 'eng')
         self.assertEqual(m.getall('~releasecountries'), ['GB', 'NZ'])
-        self.assertEqual(a.genres, {
-            'genre1': 6, 'genre2': 3,
-            'tag1': 6, 'tag2': 3})
+        self.assertEqual(a._genres, {
```
zas, line 172:
```suggestion
        self.assertEqual(a.genres, {
```

test/test_mbjson.py
```diff
@@ -166,11 +169,14 @@ def test_release_without_release_relationships(self):
...
+            'genre1': 6, 'genre2': 3
+        })
+        self.assertEqual(a._folksonomy_tags, {
```
zas, line 175:
```suggestion
        self.assertEqual(a.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -166,11 +169,14 @@ def test_release_without_release_relationships(self):
...
+            'tag1': 6, 'tag2': 3
+        })
         for artist in a._album_artists:
-            self.assertEqual(artist.genres, {
+            self.assertEqual(artist._folksonomy_tags, {
```
zas, line 179:
```suggestion
            self.assertEqual(artist.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -293,11 +299,11 @@ def test_recording(self):
         self.assertEqual(m['~video'], '')
         self.assertNotIn('originaldate', m)
         self.assertNotIn('originalyear', m)
-        self.assertEqual(t.genres, {
+        self.assertEqual(t._folksonomy_tags, {
```
zas, line 302:
```suggestion
        self.assertEqual(t.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -293,11 +299,11 @@ def test_recording(self):
...
             'blue-eyed soul': 1,
             'pop': 3})
         for artist in t._track_artists:
-            self.assertEqual(artist.genres, {
+            self.assertEqual(artist._folksonomy_tags, {
```
zas, line 306:
```suggestion
            self.assertEqual(artist.folksonomy_tags, {
```

test/test_mbjson.py
```diff
@@ -770,7 +776,7 @@ def test_release_group(self):
         self.assertEqual(m['releasetype'], 'album')
         self.assertEqual(m['~primaryreleasetype'], 'album')
         self.assertEqual(m['~releasegroup'], 'The Dark Side of the Moon')
-        self.assertEqual(r.genres, {'test2': 3, 'test': 6})
+        self.assertEqual(r._folksonomy_tags, {'test2': 3, 'test': 6})
```
zas, line 779:
```suggestion
        self.assertEqual(r.folksonomy_tags, {'test2': 3, 'test': 6})
```

picard/track.py
```diff
@@ -368,6 +375,23 @@ def _convert_folksonomy_tags_to_genre(self):
             join_with=config.setting['join_genres']
         )
 
+    def _add_tags(self, tags, name):
+        self.metadata[name] = tags.keys()
+
+    def _add_folksonomy_tags(self):
+        tags = Counter(self._folksonomy_tags)
+        tags += self.album._folksonomy_tags
```
zas, line 383:
```suggestion
        tags += self.album.folksonomy_tags
```

picard/track.py
```diff
@@ -368,6 +375,23 @@ def _convert_folksonomy_tags_to_genre(self):
...
+        self._add_tags(tags, '_folksonomy_tags')
+
+    def _add_genres(self):
+        genre = Counter(self._genres)
+        genre += self.album._genres
```
zas, line 388:
```suggestion
        genre += self.album.genres
```

picard/track.py
```diff
@@ -368,6 +375,23 @@ def _convert_folksonomy_tags_to_genre(self):
...
+        self._add_tags(genre, '_genres')
+
+    def _delete_genres_from_tags(self):
```
zas, line 391: Here we're in `Track` and delete stuff in `Album`, what about moving this method to `Album` instead and calling it from here?

zas (changes requested):
Declaring attributes as "protected" (prefixing them with underscore) is good, but then they aren't meant to be used outside the class, apart its subclasses.

For `_genres` and `_folksonomy_tags`, please check carefully, and use the getter method if not in the class itself or a subclass.

Sophist-UK:
> Declaring attributes as "protected" (prefixing them with underscore) is good, but then they aren't meant to be used outside the class, apart its subclasses.

Aside from in tests, are these used outside the class?

Because I wonder whether using protected attributes in tests is actually a good thing or not. (And funny enough about 20mins before reading this I was considering exactly this question but for PHP - except in PHP protected attributes are not actually accessible outside a class except through reflection manipulation).

The reason I wonder is because the idea behind protecting them is to prevent their use in genuine application code that shouldn't have access - and tests are IMO a special case. So on the one hand, making them non-protected fits the language intent and avoids e.g. codacy flags, but then makes them available to other genuine code when they probably should be hidden - and on the other hand using protected attributes outside the class in tests may lead to issues if Python ever decides that protected really means protected.

On the whole, my personal opinion leans towards continuing to have them as protected attributes and using them outside the class only in tests - and I am wondering how to achieve this in php tests using reflection.

picard/track.py
rakim0: Fixed in: 539e104fb2deec4b10c87359b09020ba51b88e84

rakim0:
@zas Made the changes

zas (approved):
LGTM

phw (approved):
Thanks a lot for the contribution and for following up on all the discussions and change requests. LGTM

PR #2548, opened 2024-11-05
PICARD-3000: Children's Music is shown as "Children'S Music" in Picard

zytact:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3000

Genre `Children's Music` is shown as `Children'S Music` using Python's title method.

# Solution

Create a custom function to handle making the genre title case which also handles the apostrophe.


phw (changes requested):
Thanks for this patch. For this use case it should work. But I wonder if we should use a better function to begin with.

There is an existing title case function as scripting at https://github.com/metabrainz/picard/blob/master/picard/script/functions.py#L964

I think we should factor out the title casing functionality into a function in `picard.util`, e.g. just `def titlecase(s)`. Then this function can be used both here for handling the genres and in the `func_title` scripting function.

This also allows to add tests for the title case function.

zytact:
Thank you for looking into it. I think that would be best way to do it. Should I do all this in one PR or create another PR?

phw:
@zytact Would be allright to do this as part of this PR

zas (approved):
Nice fix & code improvement, LGTM.

picard/track.py
```diff
@@ -321,6 +324,29 @@ def _customize_metadata(self):
...
+            Examples:
+                >>> titlize("children's music")
+                'Children's Music'
+                >>> titlize("blues")
+                "Blues"
+            """
+            return re.sub(r"[A-Za-z]+('[A-Za-z]+)?",
+                        lambda m: m.group(0)[0].upper() + m.group(0)[1:].lower(),
+                        text)
+
```
phw, line 349: The `titlize` function here can be removed now that the code form util is being used.

phw (changes requested):
Thanks a lot for this contribution. Looks great. Just the initial titlize function added to track.py can now be removed. Then this is good to merge

picard/track.py
zytact: Apologies for the oversight! Thank you for pointing that.

zas (approved):

PR #2552, opened 2024-11-24
PICARD-2025: Add file path to metadata view

zytact:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Adding File Path to metadata view

* JIRA ticket (_optional_): PICARD-2025

# Solution

Added a `Path` key to the metadata dictionary which gets the file name from `~new_path` and `~old_path` key .


Sophist-UK (commented):
Aside from the pylint issue, the code looks fine to me. 

(With the benefit of hindsight it might have been helpful to split the changes into two commits (one for the non-functional formatting changes, and one for the actual changes for this PR, but it was still possible to review this without these being separated.)

Before this gets merged and sets the metadata names in stone to preserve backwards compatibility, I do wonder whether the variables should be called `["path_old", "~path_saved"]` rather than `["~old_path", "~new_path"]`.

Finally there is nothing wrong with this requirement, but these are hidden variables.
1. I am not sure that this PR should have "metadata view" in the name if the variables are not viewable in the UI.
2. A separate requirement (and so not for this PR) is IMO to have both the current file location / filename displayed in the old metadata and where it will be saved if you do a save (i.e. taking into account the Move and Rename options) in the new metadata (with the standard value-changed colouring and album icons applied).

zas (changes requested):
There are too many changes that are totally unrelated to the feature you want to implement.
Please do not change string quotes without a good reason to do it, read https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#strings-quoting-single-or-double-quotes

Also, re-indentation, line wrapping, etc, should be done in a separate PR (if really needed).

Sophist-UK:
@zas I guessed that you would probably but understandably say this. In @zytact's defense, it was probably his code editor settings doing it automatically. But of course, had he put these changes in a separate commit, rebasing to remove the commit would have been easy-peasy.

zytact:
Yes, that was a mistake on my part. I mistakenly turned on format on save and didn't notice.  I apologize I will properly do a new one.

zytact:
Yes, that was a mistake on my part. I mistakenly turned on format on save and didn't notice. I apologize I will properly do a new one.

zas:
> somehow closed the pull request, should I properly do a new one?

You can just push your commits on your `path-metadata` and we can re-open this PR.

Sophist-UK:
The force push you just made removed every single change. Is that what you intended?

zytact:
Yes that was intended. I wanted to clear the git history. I just pushed the new changes.

picard/file.py
```diff
@@ -367,6 +367,7 @@ def _preserve_times(self, filename, func):
 
     def _save_and_rename(self, old_filename, metadata):
         """Save the metadata."""
+        metadata["path_old"] = old_filename
```
Sophist-UK, line 370: "~" missing.

picard/ui/metadatabox.py
```diff
@@ -561,9 +561,24 @@ def update(self, drop_album_caches=False):
...
-        tracks = self.tracks
-        self.selection_mutex.unlock()
+        try:
+            files = self.files
+            tracks = self.tracks
+
+            if not (files or tracks):
+                return None
+            for file in files:
+                old_path = file.metadata.get('path_old', file.filename)
```
Sophist-UK, line 571: Missing "~".

picard/ui/metadatabox.py
```diff
@@ -561,9 +561,24 @@ def update(self, drop_album_caches=False):
 
     def _update_tags(self, new_selection=True, drop_album_caches=False):
         self.selection_mutex.lock()
-        files = self.files
-        tracks = self.tracks
-        self.selection_mutex.unlock()
+        try:
```
Sophist-UK, line 564: Why have you introduced a try block here? What exceptions are you expecting?

There is no `except` statement - so what happens if there is an exception in the `try` section?

picard/ui/metadatabox.py
```diff
@@ -561,9 +561,24 @@ def update(self, drop_album_caches=False):
...
+                return None
+            for file in files:
+                old_path = file.metadata.get('path_old', file.filename)
+                file.orig_metadata['Path'] = old_path
+
+                new_path = file.metadata.get('~path_saved', file.filename)
+                file.metadata['Path'] = new_path
+            # Update album cache check
+            if new_selection or drop_album_caches:
+                self._single_file_album = len({file.metadata['album'] for file in files}) == 1
```
Sophist-UK, line 578: This seems to have been included in the `try` block in error since it still exists a few lines further down.

picard/ui/metadatabox.py
```diff
@@ -561,9 +561,24 @@ def update(self, drop_album_caches=False):
...
+            tracks = self.tracks
+
+            if not (files or tracks):
+                return None
+            for file in files:
+                old_path = file.metadata.get('path_old', file.filename)
+                file.orig_metadata['Path'] = old_path
+
+                new_path = file.metadata.get('~path_saved', file.filename)
+                file.metadata['Path'] = new_path
```
Sophist-UK, line 575: You have introduced a new metadata tag here and there are various issues with this:

1. I think you should split this into path and filename so that we can compare the two separately - and so that they relate directly to the Move and Rename options.
2. This is NOT a tag that should be saved to the file, so it should be `~path` and `~filename`. `Path` may be saved as a miscellaneous text tag in some formats that support this (e.g. ID3).
3. You will need to make sure that these metadata variables are recalculated if the user changes Move or Rename options.
4. You will need to change the metadata table generation to add rows for `~path` and `~filename` and ensure that they are used when comparing old and new metadata and when summarising metadata over several albums or tracks.
5. You will need to add definitions for Path and Filename metadata elsewhere so that the names get translated.
6. You may need to write additional tests.

None of the above are major pieces of work, but they do need to be done.

rdswift:
Assuming this is eventually going to be accepted and merged, please add a ticket to update the Picard documentation to include the new tags/variables.  When adding the new ticket, please reference the ticket that you're addressing in this PR so that they are linked.  Thanks.

picard/file.py
```diff
@@ -412,6 +413,10 @@ def _save_and_rename(self, old_filename, metadata):
         return new_filename
 
     def _saving_finished(self, result=None, error=None):
+        if error is None and result is not None:
+            new_filename = result
+            # Add new path info after successful rename
+            self.metadata["~path_saved"] = new_filename
```
rdswift, line 419: I must be missing something because it appears that the `"~path_saved"` variable is being set after the file has been saved.  Can you provide an example of how this would be used?

Sophist-UK: Now that Bob has prompted me to think further, it seems to me that we need to know:

1. The current path i.e. where the file actually is - displayed in Old metadata.
2. The future path if we were to save the file now - displayed in New metadata.
3. The previous path if we have saved the file since it has been loaded.

To make life easiest for scripters, once a file is loaded and before it is saved we need to decide whether the previous path is set to the current path or left empty / non-existent.

And we should decide whether to save the entire path (directory path and filename as a string) or the directory path and the filename separately or possibly both. (Scripters may want directory and filename separate or they may want it together. UI needs them separate.)

rdswift:
If I understand PICARD-2025 correctly, isn't the intent to display the same information that is displayed at the bottom of the File Naming Script Editor window (with the addition of the source path to the source filename), perhaps separated into separate lines for path and filename in the Metadata pane?

image

If so, then I don't believe it's as simple as creating new tags/variables, especially **after** the files have been saved.  It would need to apply the code that calculates the target path and filename to get the values **before** the files are saved and include them as special lines in the table in the Metadata pane (because if they were tags they would be saved in the resulting files).  You could include them as variables in the track metadata (so the user has the option of copying them to a tag for inclusion in the file), and do something special to display those variables in the table.  The values for these new variables would need to be updated after every change made to the metadata, including changes made in the table manually.  Also, if the variables themselves were used in a tagging script to modify tags/variables used in the file naming script you could end up with a circular reference situation.

My suggestion is to **not** make the (new) values available in scripts, and add them as separate read-only lines in the table in the Metadata pane.  That way the user can benefit from the comparison functionality in the table, but not screw up the metadata or file naming.

rdswift:
Further to my earlier comment, note that the existing file's path, filename and extension are already available as file metadata variables `~dirname`, `~filename` and `~extension`.  Only the new values will need to be calculated, and both old and new displayed in the table.  Also, I suggest that they only be displayed if the user has the "Move files when saving" and/or "Rename files when saving" options enabled, otherwise the information will be meaningless.  In fact, unless at least one of these options is enabled, there is no need to actually calculate the values.  This means that you will also need to track any changes to these option settings and calculate and adjust the display of the metadata table accordingly.

picard/file.py
rdswift: > 3\. The previous path if we have saved the file since it has been loaded.

Do we need this?  When the file is saved, I think the file variables are automatically updated to the new path and filename, so I'm not sure what the use case would be for the original path and filename.

> And we should decide whether to save the entire path (directory path and filename as a string) **_or_** the directory path and the filename separately **_or_** possibly both.

My suggestion is to keep the path and filename as separate variables for the maximum flexibility of use.  In Picard's scripting, it's easier to combine the two than split them out from a combined value.

Another consideration is how path separators are displayed and saved.  The current file naming script preview displays the drive and path as used for the file system (using a colon and backslashes for windows paths).  I think this is fine (and appropriate) for the display in the table in the Metadata pane.  It's probably okay for the values stored in the variables as well, but we may need to be careful because of the use of the backslashes in paths on windows systems.  I assume they would need to be escaped for the tags/variables, although I really don't know the details of how they are saved in the file metadata (if the user chooses to copy the variable to a tag so it is saved).

> **Motivation:** We should try to get this exactly right from the start before we create a backwards compatibility problem for scripters.

I agree, although I don't think we're actually changing any of the existing tags/variables so we **shouldn't** have any backwards compatibility issues.  [Famous last words...  😜]

picard/ui/metadatabox.py
zytact: Thank you for pointing that out, I forgot to add the except block but I simply couldn't get it to work on my machine without using a try block. I'll look into it properly.

Sophist-UK:
> because if they were tags they would be saved in the resulting files

Yes - I have previously commented that these metadata variables must start with `~` so that they are NOT saved as tags in the file.

> The values for these new variables would need to be updated after every change made to the metadata, including changes made in the table manually. Also, if the variables themselves were used in a tagging script to modify tags/variables used in the file naming script you could end up with a circular reference situation.

These are _**very**_ good points indeed. The possibility of circular references in scripts is an existing possibility (e.g. `$set(a, %b%a)$set(b,%a%b)`), and but because Tagging scripts run only once after metadata load, and they run sequentially, although there may be a circular case, it does not result in any looping. Similarly the filenaming script runs once when you do a save and doesn't create or change any metadata.

But if we allow the filenaming script to modify metadata implicitly (e.g. `~filename_proposed`), then this changes the situation drastically. Without analysing this extensively, I believe that this would only be useful if it is part of an extensive change in script processing whereby to maintain backwards compatibility:

1. Once any metadata has been downloaded from MusicBrainz, the pre-script new metadata is saved for reuse by scripts.
2. The filenaming script is run to get the `~*_proposed` variables.
3. The Tagging script is run against the original metadata.
4. The filenaming script is run again, and the `*_proposed` variables are compared to the last version. And if any of these are different, then steps 3 and 4 are run again until the values are stable. To prevent infinite loops, this would have to have an upper limit of iterations and give an error if it never stabilises.
5. We would need to run steps 2. to 4. again every time any metadata changed.
6. And to prevent any manual edits or changes from manually run scripts being overwritten, we would have to keep track of these and preserve them (and this wouldn't handle the situation where a manually run script might need to be rerun).

Preserving another copy of the metadata for the lifetime of a track/file being loaded seems likely to increase memory usage by 50% (because we will be holding original new in addition to the current original and new. Of course, we could probably save memory by enhancing the metadata class to track the changes to the new and so use memory only for those variables that have changed (either during original automated tagging scripts or manually) - but that would be a major increase in complication.

Consequently, I am forced to come to the same conclusion as Bob, that having new variables for `~directory` and `~filename` to hold the proposed location after a save is going to need major effort to get it to be useful.

Reluctantly, I think we need to thank Arnab for all his efforts so far, and (if he is willing) ask him to start afresh and focus on making the proposed path and filename visible in the Metadata pane and NOT to create new hidden `~*` metadata from the filenaming script.

(@zytact - Let me be the first to apologise for my lack of analytical forethought being part of having caused you to waste time on the current false start. If others agree with this analysis, then... if you still wish to create new metadata variables for your own use or for minority use, then you will need to use a plugin to do it rather than have it as part of Picard.)

phw (changes requested):
Thanks for trying to tackle this. I think this change as it is misses the point and is too simplistic, though.

It introduces `path_old`, which is only available after saving. The new path is then the current path. So with this patch you have the current path on the right and any path before saving on the left. 

What would actually be needed is showing the current path on the left, and having the new path as it is expected to be after saving shown on the right. This is the tricky part and requires running the renaming script on the metadata. In the ticket I also mentioned how this can be a performance issue. But if this is only done for single file selection on display this might be fine.

Also in order to show an entry in the tag list it would not be appropriate to modify the file.metadata or file.orig_metadata, as this will create tags or variables. Another solution needs to be found to add this information to the data displayed in metadatabox.py.

Sophist-UK:
Since the tickets requesting this date back to 2019/2020, I don't think a few weeks delay for you to complete your exams will be a major problem for anyone. 😃  And relatively minor Picard enhancements have got to come a long way down anyone's priority list compared to exams that will decide your future life!!

phw:
> I hope this delay isn’t too inconvenient, and I sincerely apologize. Thank you for your understanding, and I’ll make sure to follow up as soon as I’m able.

Quite the opposite, if you are interested in making the necessary changes you are more then welcome and I think we will all be happy to guide you through the task. This is really one of those issues that should be solved.

That it is more complex than it looks like on first sight is the main reason why this ticket is still open. On the other hand I think it is also not as complicated as we maybe first thought when looking at the ticket.

zytact:
Thank you both for your kind words. I'm definitely still interested in working on it and greatly appreciate your willingness to guide me through the process. I'll make sure to revisit it with fresh focus once my exams are over! 😊

rdswift:
@zytact,

When you pick this up again, here are a couple of links to existing code that may be of interest:

Function to produce the before and after file naming example tuple for the specified file (used in the File Naming Script Editor to display examples): https://github.com/metabrainz/picard/blob/15a8fc87144d25e9a94ff4310971aa430ead896b/picard/ui/scripteditor.py#L152

The metadata table widget (where you would be adding the before and after path and filename items for display): https://github.com/metabrainz/picard/blob/15a8fc87144d25e9a94ff4310971aa430ead896b/picard/ui/metadatabox.py#L213

zytact:
Thank you, I'll be sure to check these out.

zytact:
Hello, my exams are over. I would to like to work on this again.

> Another solution needs to be found to add this information to the data displayed in metadatabox.py

> I think we need to thank Arnab for all his efforts so far, and (if he is willing) ask him to start afresh and focus on making the proposed path and filename visible in the Metadata pane and NOT to create new hidden ~* metadata from the filenaming script.

 I read what you all said and it seems like this could be slightly tricky. It seems the original is commit in not what we are looking for.

> When you pick this up again, here are a couple of links to existing code that may be of interest:

 I also looked at the code that Bob linked to. 

Could you let me know the best way to move forward with this?

phw:
@zytact Thanks for coming back to this. To move forward with this the places Bob linked to seem really relevant. First issue is we want a new row added to the metadata table. Have a look at https://github.com/metabrainz/picard/blob/15a8fc87144d25e9a94ff4310971aa430ead896b/picard/ui/metadatabox.py#L625 how a new entry for the track length is added, even though this is not part of the metadata object.

I think we can do the same. But for performance reasons let's do this only if there is exactly one file (`len(files) == 1`). This new row should have the current file name on the left, and the new filename on the right. For the existing filename you can use `file.filename`. For the expected new filename you need to call `file.make_filename` (https://github.com/metabrainz/picard/blob/master/picard/file.py#L529).

zytact:
Could you please check if [this commit](https://github.com/metabrainz/picard/pull/2552/commits/fd9990ce2b408f51cce67a7f2fd953d645c000aa) is what we want? Also I did a rebase to remove the old changes. I am not sure if this has done something unexpected since I can see all the other commits since last month too.

picard/ui/metadatabox.py
```diff
@@ -624,6 +624,9 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
+            if (len(files) == 1):
```
Sophist-UK, line 627: This code snippet looks good to me. However you can combine the two if statements as follows:
 ```python
 if (len(files) == 1) and (settings['rename_files'] or settings['move_files']):
```

Sophist-UK:
Sorry - but looking at the 19 files that have changed I fear that your merging and rebasing has resulted in other changes being mixed in with this PR.

I think that the simplest thing to do would be to close this PR, and submit your changes as part of a new PR.

picard/ui/metadatabox.py
zytact: Ok 👍

zytact:
> Sorry - but looking at the 19 files that have changed I fear that your merging and rebasing has resulted in other changes being mixed in with this PR - and I think this will potentially cause issues in the code base as well as making it nearly impossible to review.
> 
> I think that the simplest thing to do would be to close this PR, and submit your changes as part of a new PR.

Okay, will do. Sorry for the troubles again.

PR #2555, opened 2024-12-01
PICARD-2991: Handle statvfs errors on file rename

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2991

`make_short_filename` runs `_get_filename_limit` on the basedir, which in turn tries to get filename limits by calling `statvfs`. `statvfs` can fail with `FileNotFoundError` or `PermissionError`.

To reproduce:

- In Options > File naming set the target dir to something on which `os.statvfs` fails. On my Linux system this is e.g. `/run/user/1000/doc`
- Save options and reopen the options dialog

=> File naming failed to load, exception is logged

# Solution

1. If in `_get_filename_limit` the call to `os.statvfs` fails, retry on the parent path
2. Do not fail to load the options dialog on `FileNotFoundError` or `PermissionError` when updating the examples

### DIFF
--- a/picard/util/filenaming.py
+++ b/picard/util/filenaming.py
@@ -29,6 +29,7 @@
 from enum import IntEnum
 import math
 import os
+from pathlib import Path
 import re
 import shutil
 import struct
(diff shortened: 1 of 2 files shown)


picard/util/filenaming.py
```diff
@@ -326,14 +327,21 @@ def _get_filename_limit(target):
         limit = limits[target]
     except KeyError:
         # we need to call statvfs on an existing target
-        d = target
```
zas, line 329: There's a comment just before the start of this function about Python statvfs and f_fsid not being supported though it changed in version 3.7: Added the f_fsid attribute.

I wonder if the suggested merge would make sense now.

zas (approved):
LGTM

picard/util/filenaming.py
phw: I'll look into it

PR #2560, opened 2024-12-22
Mark Python 3.13 as supported

C0rn3j:
~~Dropped utf-8 compat line for Python 2.~~

Added 3.13 Python as explicitly supported.

~~Converted the file description to an appropriate docstring.~~

---

PR irrelevant side note - It would be nice if there could be a new release, as the lack of [this change](https://github.com/metabrainz/picard/commit/cbbf53dee4c7e0de201daddc096941a4c822d5fc) breaks installation for us.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,8 @@
 #!/usr/bin/env python3
-"""Picard, the next-generation MusicBrainz tagger"""
-
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
 # Copyright (C) 2006-2008, 2011-2014, 2017 Lukáš Lalinský
 # Copyright (C) 2007 Santiago M. Mola
 # Copyright (C) 2008 Robert Kaye
--- a/setup.py
+++ b/setup.py
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
-#
-# Picard, the next-generation MusicBrainz tagger
-#
+"""Picard, the next-generation MusicBrainz tagger"""
+
 # Copyright (C) 2006-2008, 2011-2014, 2017 Lukáš Lalinský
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -1,8 +1,6 @@
 #!/usr/bin/env python3
-# -*- coding: utf-8 -*-
-#
-# Picard, the next-generation MusicBrainz tagger
-#
+"""Picard, the next-generation MusicBrainz tagger"""
+
```
phw, line 3: The header is written by [scripts/tools/fix-header.py](https://github.com/metabrainz/picard/blob/master/scripts/tools/fix-header.py). If we want to remove the coding part should be done there and the script rerun to update the whole project.

Also "Picard, the next-generation MusicBrainz tagger" should stay as a comment. This belongs to the copyright lines following it. Some files in the repo do have an explicit doc comment that is separate from the copyright comment. I'm not sure whether a doc comment for setup.py is any use, but "Picard, the next-generation MusicBrainz tagger" doesn't appear to me of being of much use.

C0rn3j: I've only left the version change in this PR then

phw (approved):

phw:
Thanks, let's merge this. For some reason  couldn't approve the pipeline to run. Just bypassed this and merged directly, bit really relevant here anyway.

As for a new release: I guess we will do one some time in Januar.

zas:
No strong opinion on this, but I don't feel there's any emergency for this move.

@phw I gave my opinion, but if you prefer to remove those, just make the changes.

phw:
Really no strong opinion on this. IMHO the lines do no harm, so there is no real need to remove them. On the other hand I wouldn't add them to a new project in the first place.

As we can apply this change across the project by changing and running `fix-header.py` I think it's ok to do so. I'd merge a PR doing so, but right now it's not something I  need to bother with myself.

PR #2563, opened 2024-12-28
PICARD-2025: Add file path to metadatabox

zytact:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Adding File Path to metadata view

* JIRA ticket (_optional_): PICARD-2025
* JIRA ticket to update documentation: PICARD-3017

# Solution

Add a Path tag in the metadatabox for a single file (for efficiency reasons) with original value from `file.filename`, new value from `file.make_filename` only when `move_files` or `rename_files` is enabled.

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -624,6 +624,8 @@ class MetadataBox(QtWidgets.QTableWidget):
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
+            if (len(files) == 1) and (settings['rename_files'] or settings['move_files']):
+                tag_diff.add('Path', [file.filename], [file.make_filename(file.filename, orig_metadata)], removable=False, readonly=True)
 
         for track in tracks:
             if track.num_linked_files == 0:
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -282,8 +282,21 @@ class MetadataBox(QtWidgets.QTableWidget):
         self.ignore_updates = IgnoreUpdatesContext(on_exit=self.update)
         self.tagger.clipboard().dataChanged.connect(self._update_clipboard)
 
-    def _on_setting_changed(self, key, old_value, new_value):
-        if key in ('rename_files', 'move_files'):
+    def _on_setting_changed(self, name, old_value, new_value):
+        settings_to_watch = {
+            "enabled_plugins",
+            "move_files",
+            "move_files_to",
+            "rename_files",
+            "standardize_artists",
+            "va_name",
+            "windows_compatibility",
+            "selected_file_naming_script_id",
+            "file_naming_scripts",
+            "user_profiles",
+            "user_profile_settings"
+        }
+        if name in settings_to_watch:
             self.update(drop_album_caches=False)
 
     def _get_file_lookup(self):
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -624,8 +624,12 @@ class MetadataBox(QtWidgets.QTableWidget):
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
-            if len(files) == 1 and (settings['rename_files'] or settings['move_files']):
-                tag_diff.add('~filepath', [file.filename], [file.make_filename(file.filename, orig_metadata)], removable=False, readonly=True)
+            if len(files) == 1:
+                if settings['rename_files'] or settings['move_files']:
+                    new_filename = file.make_filename(file.filename, new_metadata)
+                else:
+                    new_filename = file.filename
+                tag_diff.add('~filepath', [file.filename], [new_filename], removable=False, readonly=True)
 
         for track in tracks:
             if track.num_linked_files == 0:
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -238,6 +238,7 @@ class MetadataBox(QtWidgets.QTableWidget):
         super().__init__(parent=parent)
         self.tagger = QtCore.QCoreApplication.instance()
         config = get_config()
+        config.setting.setting_changed_signal.connect(self._on_setting_changed)
         self.setAccessibleName(_("metadata view"))
         self.setAccessibleDescription(_("Displays original and new tags for the selected files"))
         self.setColumnCount(3)
@@ -280,6 +281,10 @@ class MetadataBox(QtWidgets.QTableWidget):
         self._single_track_album = False
         self.ignore_updates = IgnoreUpdatesContext(on_exit=self.update)
         self.tagger.clipboard().dataChanged.connect(self._update_clipboard)
+
+    def _on_setting_changed(self, key, old_value, new_value):
+        if key in ('rename_files', 'move_files'):
+            self.update(drop_album_caches=False)
 
     def _get_file_lookup(self):
         """Return a FileLookup object."""
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -149,6 +149,9 @@ class TagDiff:
         else:
             return orig != new
 
+    def is_readonly(self, tag):
+        return bool(self.status[tag] & TagStatus.READONLY)
+
     def add(self, tag, orig_values, new_values, removable, removed=False, readonly=False, top_tags=None):
         if orig_values:
             self.orig.add(tag, orig_values)
@@ -714,7 +717,7 @@ class MetadataBox(QtWidgets.QTableWidget):
             if not new_item:
                 new_item = QtWidgets.QTableWidgetItem()
                 new_item.setTextAlignment(alignment)
-                if tag == '~length':
+                if self.tag_diff.is_readonly(tag):
                     new_item.setFlags(orig_flags)
                 else:
                     new_item.setFlags(new_flags)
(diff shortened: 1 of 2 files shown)


picard/ui/metadatabox.py
```diff
@@ -624,6 +624,8 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
+            if (len(files) == 1) and (settings['rename_files'] or settings['move_files']):
```
phw, line 627: Minor, but I'd remove the round brackets around `len(files) == 1`. Not really needed here.

picard/ui/metadatabox.py
```diff
@@ -624,6 +624,8 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
...
+                tag_diff.add('Path', [file.filename], [file.make_filename(file.filename, orig_metadata)], removable=False, readonly=True)
```
phw, line 628: "Path" should be translatable. Let's handle this how `~length` is handled, as a special "tag". Let's call it `~filepath` (we already have `~filename`) and then add an entry in `picard/util/tags.py` to `TAG_NAMES`:

```python
    '~filepath': N_('File Path'),
```

phw (changes requested):
Thanks. This is one of those cases where the more we talked about it the simpler the code changes became :D

Looks good to me except for some minor stylistic comment and making the name translatable. I think otherwise this change does what we need.

But I'd like to test run this to see whether it does what we need. I'll need to leave in a few minutes, so will look at this later again.

Also let's see what others have to say.

picard/ui/metadatabox.py
zytact: Okay!

picard/ui/metadatabox.py
zytact: Done.

zytact:
> Thanks. This is one of those cases where the more we talked about it the simpler the code changes became :D
> 
> Looks good to me except for some minor stylistic comment and making the name translatable. I think otherwise this change does what we need.
> 
> But I'd like to test run this to see whether it does what we need. I'll need to leave in a few minutes, so will look at this later again.
> 
> Also let's see what others have to say.

Yes, it did seem daunting at first. Thank you for the patience you and everyone else showed. I am still fairly new to this codebase and to contributing to open source.

Sophist-UK:
Can we please stop using force push?

rdswift:
A couple of things...

1. Can you please make sure that there is also a ticket to update the Picard documentation, and note the ticket in PICARD-2025 and here so that it doesn't get lost?  Thanks.

2. Does there need to be some way to update the new value automatically whenever one of the other new tag values is changed (because the tag value being updated **may** be used in the renaming script?

Sophist-UK:
@rdswift Bob - you make a very good point about updating the file path when any metadata used in the file naming script gets changed.

Also, we need to be mindful of the metadata summarisation and check that this works correctly. And for this I think that the path and file name should probably be shown separately - so that e.g. if you select all the tracks in an album, then you can see that the path is the same, but the filenames are different.

picard/ui/metadatabox.py
```diff
@@ -624,6 +624,8 @@ def _update_tags(self, new_selection=True, drop_album_caches=False):
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
+            if len(files) == 1 and (settings['rename_files'] or settings['move_files']):
+                tag_diff.add('~filepath', [file.filename], [file.make_filename(file.filename, orig_metadata)], removable=False, readonly=True)
```
phw, line 628: The new filename must be made from `new_metadata`, not `orig_metadata`.

phw:
> Does there need to be some way to update the new value automatically whenever one of the other new tag values is changed (because the tag value being updated may be used in the renaming script?

That's a good point. I tested, and this is already happening.

But a separate thing is that the metadata view now also should update if file naming options are changed (all of rename files, move files, the selected script or the script itself). Not sure how to best do this, we currently don't have a good generic solution for this. Some option dialogs trigger UI updates explicitly for their changes. But the file naming options can be toggled in multiple places (file naming options, script editor dialog, options menu quick settings).

> Also, we need to be mindful of the metadata summarisation and check that this works correctly. And for this I think that the path and file name should probably be shown separately - so that e.g. if you select all the tracks in an album, then you can see that the path is the same, but the filenames are different.

I asked @zytact to implement this only for single files selected. I don't think running the script on a huge number of files is a good idea. We already need to be careful with metadatabox updates if multiple files are selected. It's also seldom useful, as every file loaded has a different path.

If this is a wanted feature this can be addressed separately, but it will be significantly more complicated. We could for example pre-calculate the path for all files and update it in the background if it changes for any

picard/ui/metadatabox.py
phw: After testing this I think the path should also be shown in case the settings `rename_files` and `move_files` are disabled. Then it is just easier and both the old and new value are `file.filename`. I'd do something like this:

```python
            if len(files) == 1:
                if settings['rename_files'] or settings['move_files']:
                    new_filename = file.make_filename(file.filename, new_metadata)
                else:
                    new_filename = file.filename
                tag_diff.add('~filepath', [file.filename], [new_filename],
                             removable=False, readonly=True)
```

phw (changes requested):
Tested this and I have a few more comments, see below.

Also the new row is not properly readonly, setting the flag on the `TagDiff.add` call is currently not enough. There is some special handling for `~length` in https://github.com/metabrainz/picard/blob/4fefe6ca81344790691506936a2dbabaee6cb36f/picard/ui/metadatabox.py#L713-L714 , which makes sure the field does not get the `Qt.ItemFlag.ItemIsEditable` set.

We can improve this. My suggestion is to add a new helper method to `TagDiff` to check the readonly state of a tag:

```python
class TagDiff:
    # ...

    def is_readonly(self, tag):
        return bool(self.status[tag] & TagStatus.READONLY)
```

Then use this in https://github.com/metabrainz/picard/blob/4fefe6ca81344790691506936a2dbabaee6cb36f/picard/ui/metadatabox.py#L713-L714

```python
                if self.tag_diff.is_readonly(tag):
                    new_item.setFlags(orig_flags)
```

zytact:
Does this work?

rdswift:
> But a separate thing is that the metadata view now also should update if file naming options are changed (all of rename files, move files, the selected script or the script itself). Not sure how to best do this, we currently don't have a good generic solution for this. Some option dialogs trigger UI updates explicitly for their changes. But the file naming options can be toggled in multiple places (file naming options, script editor dialog, options menu quick settings).

Maybe we can detect this in the module that actually updates the settings and emit a signal from there?  All the (related) changes from the various locations all pass through this module (including changes to the enabled profiles), so this might be a simple way to catch them.

https://github.com/metabrainz/picard/blob/241aa6690ebfa7dbca13643e3e57dbc0f5196edc/picard/config.py#L192-L201

We could create a set containing the settings that would trigger a change to the file naming behavior and emit a signal if the setting being changed was contained in the set.  The suggested settings to trigger the signal might include: "enabled_plugins", "move_files", "move_files_to", "rename_files", "standardize_artists", "va_name", "windows_compatibility", "selected_file_naming_script_id", "file_naming_scripts", "user_profiles" and "user_profile_settings".

phw:
@zytact Your newest changes look good to me. What we could do is to wait for rdswift's changes from #2567 to be merged, then the metadatabox could also implement listening to config changes and run an update if file naming related options change.

zytact:
> @zytact Your newest changes look good to me. What we could do is to wait for rdswift's changes from #2567 to be merged, then the metadatabox could also implement listening to config changes and run an update if file naming related options change.

Of course.

zas:
> > @zytact Your newest changes look good to me. What we could do is to wait for rdswift's changes from #2567 to be merged, then the metadatabox could also implement listening to config changes and run an update if file naming related options change.
> 
> Of course.
> 

#2567 was merged

zytact:
Is this what we are looking for?

rdswift:
> Is this what we are looking for?

I think @phw was also looking for a way to indicate that the path and filename need to be recalculated if any of the settings that impact the file naming have changed.  Settings like: "enabled_plugins", "move_files", "move_files_to", "rename_files", "standardize_artists", "va_name", "windows_compatibility", "selected_file_naming_script_id", "file_naming_scripts", "user_profiles" and "user_profile_settings".  Note that some of these may not impact the file renaming, but they ***could***.

zytact:
> > Is this what we are looking for?
> 
> I think @phw was also looking for a way to indicate that the path and filename need to be recalculated if any of the settings that impact the file naming have changed.  Settings like: "enabled_plugins", "move_files", "move_files_to", "rename_files", "standardize_artists", "va_name", "windows_compatibility", "selected_file_naming_script_id", "file_naming_scripts", "user_profiles" and "user_profile_settings".  Note that some of these may not impact the file renaming, but they ***could***.

I see. OK. I'll do that.

phw:
> Is this what we are looking for?

Yes, but rdswift compiled a much more complete list of setting keys that might affect it.

picard/ui/metadatabox.py
```diff
@@ -278,6 +282,10 @@ def __init__(self, parent=None):
         self.ignore_updates = IgnoreUpdatesContext(on_exit=self.update)
         self.tagger.clipboard().dataChanged.connect(self._update_clipboard)
 
+    def _on_setting_changed(self, key, old_value, new_value):
+        if key in ('rename_files', 'move_files'):
```
rdswift, line 286: It may not make much of a difference but this might be more efficient if the keys to check were in a set rather than a tuple.

zytact: Done.

phw (approved):
Thanks a lot. Works great for me. Let's merge this.

zytact:
Thanks for helping me and the all the patience. 🙂

PR #2564, opened 2024-12-28
Run CI tests and package builds with Python 3.13

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Test with Python 3.13 in CI builds, package for Windows and macOS using Python 3.13

This also contains a fix for PICARD-3016 to fix relative plugin path logging on Windows. This being broken was already the case, but now exposed by the CI tests failing with Python 3.13 due to differences in absolute path handling.

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -11,8 +11,8 @@ jobs:
       matrix:
         setup:
         - macos-deployment-version: '11.0'
-          python-version: 3.12.1-macos11
-          python-sha256sum: 6178e42679eb83196240fc58b1438f481c32c2b0557f28ccf43aa7b1b80b7c4a
+          python-version: 3.13.1-macos11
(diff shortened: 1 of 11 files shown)


phw:
Interesting how file path tests are failing with Python 3.13 on Windows. Not yet sure what exactly changed in Python, will investigate.

phw:
Python 3.13 has the following change:

```
On Windows, [isabs()](https://docs.python.org/3/library/os.path.html#os.path.isabs) no longer considers paths starting with exactly one slash (\ or /) to be absolute. (Contributed by Barney Gale and Jon Foster in [gh-44626](https://github.com/python/cpython/issues/44626).)
```

See https://docs.python.org/3/whatsnew/3.13.html

zas (approved):
LGTM

PR #2566, opened 2024-12-29
Support more LRC timestamps

YohayAiTe:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

When saving an ID3 file with LRC synced lyrics, it only accepted a very narrow definition of timestamps (two-digit minutes, two-digit seconds, and three-digit milliseconds). While this is fine for preserving the field (as in PICARD-1092), some plugins may import or download LRCs with different timestamps, and people can edit the field manually and change as well.

No JIRA ticket was open for this problem, and there was no LRC support at all in the previous release.

# Solution

I changed the regular expressions that are used for LRC decoding (`ID3File.__lrc_*_re_parse`), and the corresponding timestamp parsing at `ID3File._parse_lrc_text`. I also added new test cases for the new timestamps.

# Action

Additional actions required:

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -245,9 +245,11 @@ class ID3File(File):
         'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$')
     }
 
-    __lrc_line_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\])')
-    __lrc_syllable_re_parse = re.compile(r'(<\d\d:\d\d\.\d\d\d>)')
-    __lrc_both_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\]|<\d\d:\d\d\.\d\d\d>)')
+
+    __lrc_time_format_re = r'\d+:\d{1,2}(?:.\d+)?'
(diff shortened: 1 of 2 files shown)


phw (approved):
Thanks a lot for this fix. Looks good to me

picard/formats/id3.py
```diff
@@ -245,9 +245,11 @@ class ID3File(File):
         'MVIN': re.compile(r'^(?P<movementnumber>\d+)(?:/(?P<movementtotal>\d+))?$')
     }
 
-    __lrc_line_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\])')
-    __lrc_syllable_re_parse = re.compile(r'(<\d\d:\d\d\.\d\d\d>)')
-    __lrc_both_re_parse = re.compile(r'(\[\d\d:\d\d\.\d\d\d\]|<\d\d:\d\d\.\d\d\d>)')
+
```
phw, line 248: Minor thing, but flake8 is complaining about the additional blank line here

phw (changes requested):

picard/formats/id3.py
YohayAiTe: Removed the extra blank line, should be fine now

phw (approved):
Thanks again

PR #2567, opened 2024-12-29
Add signal when file naming related settings are changed

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds a signal when file naming related settings are changed

# Problem

File naming related settings can be changed in a number of places, and there is no consolidated way to capture whether a setting has been changed.  Something like this is required for https://github.com/metabrainz/picard/pull/2563

# Solution

This PR adds a signal to the config settings class to determine when a file naming related setting has been updated.

# Action

Additional actions required:

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -136,21 +136,8 @@ class SettingConfigSection(ConfigSection):
     PROFILES_KEY = 'user_profiles'
     SETTINGS_KEY = 'user_profile_settings'
 
-    # Set containing settings that can impact the result of file renaming (path or filename)
-    FILE_NAMING_SETTINGS = {
-        'enabled_plugins',
-        'move_files',
-        'move_files_to',
-        'rename_files',
-        'standardize_artists',
-        'va_name',
-        'windows_compatibility',
-        'selected_file_naming_script_id',
-        'file_naming_scripts',
-        'user_profiles',
-        'user_profile_settings',
-    }
-    naming_settings_changed_signal = QtCore.pyqtSignal(str)
+    # Signal emitted when the value of a setting has changed.
+    setting_changed_signal = QtCore.pyqtSignal(str)
 
     @classmethod
     def init_profile_options(cls):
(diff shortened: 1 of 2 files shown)


rdswift:
To make this signal more "generic", we could include a function in the `SettingConfigSection` class to provide the setting names set to check for triggering the signal.  That way it could be used to signal on different settings as required for different applications.  I can make that change if you think there might be some value in having this flexibility.

phw (commented):
Generally this is a good approach. But I think we should avoid adding this very specific logic to the settings module. I'd instead have the settings emit a generic signal when settings have changed and have the update logic in e.g. the metadata box handle the specifics.

We have multiple places where we could make use of this. E.g. syncing the rename andove file settings with the menu and the script editor dialog.

rdswift:
> Generally this is a good approach. But I think we should avoid adding this very specific logic to the settings module. I'd instead have the settings emit a generic signal when settings have changed and have the update logic in e.g. the metadata box handle the specifics.

Yeah, that makes sense.  I'll modify it to emit a signal when a setting value has changed, with the signal providing the name of the setting as an argument.  If the new value for the setting is the same as the old value then the signal will not be emitted.  This should avoid the situation of emitting signals for every setting when only one setting on an option page has changed (but all settings on the page are saved).

zas (approved):
Good move, it looks good to me.

phw (dismissed):
Looks good to me.

Just an idea, but no blocker: should the signal also emit the new value and old value of the setting? Not sure we have a use case for this, but it would avoid querying the settings again.

rdswift:
> Just an idea, but no blocker: should the signal also emit the new value and old value of the setting? Not sure we have a use case for this, but it would avoid querying the settings again.

That would make it more complete for sure.  I'll have a look.

phw (approved):
Great, thanks.

zas (approved):
Even better with old & new values passed to signal, LGTM.

PR #2569, opened 2024-12-31
Use settings changed signal to update file move / rename toggles

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

This makes use of @rdswift's recent settings changed signal (#2567) to handle updating the filenaming toggles in Options menu

# Solution

# Action

There are additionally issues with the profiles. When switching profiles via menu the toggles don't update accordingly. This issue is also present on the current 2.x release.

I started looking into this, but encountered additional profile issues on the master branch:

1. Profile specific settings are not available directly after startup, hence the toggles don't show the settings properly. The profile groups are only properly initialized after the option dialog has been opened the first time. This is due to #2433 moving the initialization of profile groups to the option pages constructors.
2. When opening the option dialog multiple times the settings in the profile configuration tree view duplicate each time.

I'll look into this in more detail in the new year. Because this is more complex I would like to keep this out of this more simple PR.

### DIFF
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -7,7 +7,7 @@
 # Copyright (C) 2008 Gary van der Merwe
 # Copyright (C) 2008 Robert Kaye
 # Copyright (C) 2008 Will
-# Copyright (C) 2008-2010, 2015, 2018-2023 Philipp Wolfer
+# Copyright (C) 2008-2010, 2015, 2018-2024 Philipp Wolfer
 # Copyright (C) 2009 Carlin Mangar
 # Copyright (C) 2009 David Hilton
 # Copyright (C) 2011-2012 Chad Wilson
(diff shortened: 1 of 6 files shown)


picard/ui/mainwindow/__init__.py
```diff
@@ -255,6 +255,20 @@ def setupUi(self):
...
+    def handle_settings_changed(self, name, old_value, new_value):
+        if name == 'rename_files':
+            self.actions[MainAction.ENABLE_RENAMING].setChecked(new_value)
+        elif name == 'move_files':
+            self.actions[MainAction.ENABLE_MOVING].setChecked(new_value)
+        elif name == 'dont_write_tags':
+            self.actions[MainAction.ENABLE_TAG_SAVING].setChecked(not new_value)
+        elif name == 'clear_existing_tags':
+            self.metadata_box.update()
+        elif name in ('file_renaming_scripts', 'selected_file_naming_script_id'):
```
rdswift, line 269: I don't know if it makes any difference, but should this be a set rather than a tuple?

zas: In this context it doesn't make any difference though a set is more correct.

rdswift: Okay, thanks for clarifying.  You've caught me on that so many times in the past that I use sets whenever ordering is not required out of habit now.  😜

zas (approved):
LGTM

picard/ui/mainwindow/__init__.py
phw: Yes, I think we agreed in the past to use sets here. I'll change it

zas (approved):
LGTM

PR #2570, opened 2025-01-01
Fix option profiles

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3019

This addresses several issues with option profiles.

1. PICARD-3019: Changing profile using the menu is not persisted and does not update the other quick setting toggles in the Option menu. This issue is also present in the current release version 2.13.3.
2. Option profiles are not initialized properly on startup until the Options dialog got opened the first time, as settings being part of profiles are registered late in the profile page constructors since https://github.com/metabrainz/picard/pull/2433

See also comment in https://github.com/metabrainz/picard/pull/2569

# Solution

1. The first issue is addressed by generalizing the the new `setting_changed` signal and using it to listen to profile changes. If profiles are changed rebuild the full option menu and the CD lookup config.
2. For fixing the option profile registration move the registration back to initialization by making `register_setting` static and have each option page define an `OPTIONS` tuple again. It's a bit similar to old 2.x code, but the OPTIONS tuple is simplified and just consists of the name and the highlights.

### DIFF
--- a/picard/config.py
+++ b/picard/config.py
@@ -59,6 +59,9 @@ class ConfigSection(QtCore.QObject):
 
     """Configuration section."""
 
+    # Signal emitted when the value of a setting has changed.
+    setting_changed = QtCore.pyqtSignal(str, object, object)
+
     def __init__(self, config, name):
         super().__init__()
         self.__qt_config = config
@@ -76,9 +79,12 @@ class ConfigSection(QtCore.QObject):
         return self.value(name, opt, opt.default)
 
     def __setitem__(self, name, value):
+        old_value = self.__getitem__(name)
         key = self.key(name)
         self.__qt_config.setValue(key, value)
         self._memoization[key].dirty = True
+        if value != old_value:
+            self.setting_changed.emit(name, old_value, value)
 
     def __contains__(self, name):
         return self.__qt_config.contains(self.key(name))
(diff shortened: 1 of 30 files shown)


phw:
@rdswift One detail I was unsure about was whether not persisting the profile when changed via menu was intentional or not. But IMHO it should be persisted. Otherwise it also is not properly applied.

rdswift:
> @rdswift One detail I was unsure about was whether not persisting the profile when changed via menu was intentional or not. But IMHO it should be persisted. Otherwise it also is not properly applied.

Not intentional.  I agree with you that it should be persisted.

rdswift:
I'm seeing a duplication of the changes that added the setting changed signal.  Does this need to be rebased on the master branch?

rdswift (approved):
Looks good to me.  The only thing I noticed was that you didn't update the copyright notice in most of the files.  :wink:

zas (approved):
Good fix, and moving options back to OPTIONS makes things cleaner. LGTM.

phw:
> The only thing I noticed was that you didn't update the copyright notice in most of the files. 😉

I thought the changes were really minor, and also we should probably rerun our fix-headers script again.

PR #2573, opened 2025-01-08
Fix loading of folksonomy tags / genres

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2607, PICARD-489

This is a follow up for #2545, which added the genre and folksonomy tags variables. During testing I found some issues with the behavior. The new variables were not properly created as hidden variables. Also the merging of variables from different levels was not applied consistently and setting the `genre` tag from the genre and folksonomy values was only fully applied on album reload.

# Solution

- Save `~folksonomy_tags` and `~genres` as hidden variables
- Fix merging tags from different levels
- Fix excluding genre tags from `~folksonomy_tags`

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -858,10 +858,6 @@ class Album(MetadataItem):
             yield from track.files
         yield from self.unmatched_files.files
 
-    def delete_genres_from_tags(self):
-        for genre in self.genres:
-            del self.folksonomy_tags[genre]
-
 
 class NatAlbum(Album):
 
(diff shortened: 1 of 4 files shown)


zas (changes requested):
BTW, tests didn't catch the issues you reported, those might need to be improved.
The new code looks much better to me :)

A `print()` was forgotten.

phw:
@zas I removed the print statement and added extended tests for the track genre handling.

zas (approved):
LGTM

PR #2577, opened 2025-02-06
PICARD-3030: Fix crash when merging original values for multiple tags

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3030

This refactors the change from PICARD-3021 to address a crash when selecting more then one changed tag in the metadatabox and then open the context menu.

# Solution

Avoid the crash by implementing merge originals similar to the existing use originals.

Also refactor use original values a bit to reduce code duplication and harmonize the implementation of both merge and use originals.

### DIFF
--- a/picard/ui/metadatabox.py
+++ b/picard/ui/metadatabox.py
@@ -419,6 +419,7 @@ class MetadataBox(QtWidgets.QTableWidget):
                     menu.addAction(remove_from_preserved_tags_action)
             removals = []
             useorigs = []
+            mergeorigs = []
             item = self.currentItem()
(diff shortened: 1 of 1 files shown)


zas (dismissed):
It looks good to me, perhaps just merge https://github.com/phw/picard/pull/23 which slightly reduce code redundancy.

Also I think we should rename `useorigs` and `mergeorigs` lists but I didn't come up with a good name.

PR #2578, opened 2025-02-15
PICARD-3034: Fix win-build-fixes.ps1 not being called

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3034

This fixes packaging issues on Windows, where system wide Qt and libssl related libraries would be loaded over the bundled ones.

This was originally fixed in PICARD-2736. But the change in PICARD-3002 moved the cleanup code to a separate helper script `win-build-fixes.ps1`, which was forgotten to be called.

# Solution

Call `win-build-fixes.ps1`


phw:
Tested the installer. The changes work as expected and all DLLs are in the root folder again.

PR #2579, opened 2025-02-16
PICARD-2192: For macOS build separate x86_64 and arm64 packages

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2192

This is an alternative attempt to provide ARM64 packages for macOS. 

# Solution

Instead of building a single universal2, as attempted in #2423, this builds two separate packages for the x86_64 and arm64 architectures.

The reason for now is that the PyQt6-Qt6 package is only available as separate packages as well. 

While it would be nicer to have a single package, this approach will solve PICARD-2192 as well. If we want to provide a combined package in the future there would be two approaches:

- Download the separate binary packages of PyQt6-Qt6 and have the individual dynamic libraries from each package combined in a single universal2 one by a script.
- Build Qt6 ourselves as part of the build pipeline.

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -11,6 +11,11 @@ jobs:
       matrix:
         setup:
         - macos-deployment-version: '11.0'
+          target-arch: 'x86_64'
+          python-version: 3.13.1-macos11
+          python-sha256sum: 67c6f0a3190851e0013214d5abd725a42ec398ff1b50eec47826820fd052d86b
+        - macos-deployment-version: '11.0'
+          target-arch: 'arm64'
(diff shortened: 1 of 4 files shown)


phw:
As a side note, this could also be backported to the 2.x branch with Qt5, as the PyQt5-Qt5 is available as arm64 as well (which originally was a blocker to this).

phw:
@zas This should be fine. I can't test this no better than you, though. I think we should merge it, then I can ask users for help with testing the builds.

PR #2580, opened 2025-02-16
PICARD-2909: Do not write image width / height for Opus files

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

libopus expects width, height and depth to all be zero or all be non-zero

# Solution

As depth is not easily available, do not set width and height either


zas (approved):
LGTM

PR #2582, opened 2025-02-22
Codacy fixes

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/.prospector.yml
+++ b/.prospector.yml
@@ -1,4 +1,12 @@
 # Configuration for prospector, mainly used by Codacy.
+output-format: grouped
+
+max-line-length: 120
+ignore-patterns:
+  - ^picard/ui/forms/ui_.*\.py$
+
+mccabe:
+  run: false
 
 pep8:
   # Please see comments in setup.cfg as to why we disable the below checks.
--- a/picard/coverart/__init__.py
+++ b/picard/coverart/__init__.py
@@ -109,7 +109,8 @@ class CoverArt:
                     filters_result = run_image_filters(data, image_info, self.album, coverartimage)
                 if filters_result:
                     self._set_metadata(coverartimage, data, image_info)
-            except imageinfo.IdentificationError:
+            except imageinfo.IdentificationError as e:
+                log.warning(f"Couldn't identify image {coverartimage}: {e}")
                 return
 
         self.next_in_queue()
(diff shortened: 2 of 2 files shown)


zas:
The first commit is clearly a bug fix, though it only concerns Picard 3 branch (master)

rdswift:
Minor comment (and perhaps a dumb question), but isn't it (a tiny bit) more efficient to use a string with parameters than an f-string for log entries because the replacement doesn't need to be done if the log entry isn't saved?  (Lines 113 and 180 in your revised `picard/coverart/__init__.py` file.)

phw (approved):
Looks good. Was the missing image info an actual bug with consequences? I wonder why this was not discovered earlier and whether we should fix it in 2.x

zas:
> Minor comment (and perhaps a dumb question), but isn't it (a tiny bit) more efficient to use a string with parameters than an f-string for log entries because the replacement doesn't need to be done if the log entry isn't saved? (Lines 113 and 180 in your revised `picard/coverart/__init__.py` file.)

You're right, I just overlook those and should have use % style.
That's the usual style we use for logging in Picard. Performance-wise that's not critical.
It's fixed in https://github.com/metabrainz/picard/pull/2582/commits/973593faab2187423fe69834d6caea1c8286deab

But it opens a good discussion about `logging` module and f-strings. I think we would benefit using f-strings more in the code, but those don't play well with logging and variable arguments.

For reference see those discussions:
- https://discuss.python.org/t/safer-logging-methods-for-f-strings-and-new-style-formatting/13802
- https://stackoverflow.com/questions/13131400/logging-variable-data-with-new-format-string

zas:
> I wonder why this was not discovered earlier and whether we should fix it in 2.x

Was the patch https://github.com/metabrainz/picard/commit/e8e3bc7188a2b4b12e3cabe512ea79ddbe223371 merged to 2.X ??

zas:
> Was the missing image info an actual bug with consequences?

Yes, likely raising an exception and bye bye. But the case is rare, because it is triggered by the use of local files as cover art images only. Basically the function call is missing an argument, so I expect Python to complain at runtime.

phw:
> Was the patch e8e3bc7 merged to 2.X ??

No

PR #2583, opened 2025-02-23
Add Ruff config

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1203,7 +1203,7 @@ def func_datetime(parser, format=None):
         return datetime.datetime.now(tz=local_tz).strftime(format)
     except ValueError:
         stackitem = parser._function_stack.get()
-        raise ScriptRuntimeError(stackitem, "Unsupported format code")
+        raise ScriptRuntimeError(stackitem, "Unsupported format code") from None
 
 
 @script_function(eval_args=False, documentation=N_(
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -1201,9 +1201,9 @@ def func_datetime(parser, format=None):
         format = '%Y-%m-%d %H:%M:%S'
     try:
         return datetime.datetime.now(tz=local_tz).strftime(format)
-    except ValueError:
+    except ValueError as e:
         stackitem = parser._function_stack.get()
-        raise ScriptRuntimeError(stackitem, "Unsupported format code") from None
+        raise ScriptRuntimeError(stackitem, "Unsupported format code") from e
 
 
 @script_function(eval_args=False, documentation=N_(
--- a/picard/script/serializer.py
+++ b/picard/script/serializer.py
@@ -245,7 +245,7 @@ class ScriptSerializer():
             with open(filename, 'w', encoding='utf-8') as o_file:
                 o_file.write(script_text)
         except OSError as error:
-            raise ScriptSerializerExportError(format=FILE_ERROR_EXPORT, filename=filename, error_msg=error.strerror)
+            raise ScriptSerializerExportError(format=FILE_ERROR_EXPORT, filename=filename, error_msg=error.strerror) from None
         dialog = QtWidgets.QMessageBox(
             QtWidgets.QMessageBox.Icon.Information,
             _("Export Script"),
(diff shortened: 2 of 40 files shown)


zas:
@phw ruff could replace isort, pylint, and flake8 we have configs for, so that's a first step.

I created a separated workflow for now, but we may want to actually replace pylint in existing workflows.
Also I fixed files according to extra warnings I was getting locally, some are quite interesting (like missing @staticmethod I detected because my precommit hook is running flake8, but those weren't catched by ruff afaik), most are cosmetic only.

Those related to raise ... from are trickier, most of times we want to use from parent exception, but in case we actually handle it, so I tried to stick to some ... logic. A review is very needed though ;)

picard/script/functions.py
```diff
@@ -1203,7 +1203,7 @@ def func_datetime(parser, format=None):
         return datetime.datetime.now(tz=local_tz).strftime(format)
     except ValueError:
         stackitem = parser._function_stack.get()
-        raise ScriptRuntimeError(stackitem, "Unsupported format code")
+        raise ScriptRuntimeError(stackitem, "Unsupported format code") from None
```
phw, line 1206: Wouldn't the source `ValueError` be interesting here, because it could hint into what error the format string had? Also I'm not sure what actually causes a `ValueError` here.

picard/script/serializer.py
```diff
@@ -245,7 +245,7 @@ def export_script(self, parent=None):
             with open(filename, 'w', encoding='utf-8') as o_file:
                 o_file.write(script_text)
         except OSError as error:
-            raise ScriptSerializerExportError(format=FILE_ERROR_EXPORT, filename=filename, error_msg=error.strerror)
+            raise ScriptSerializerExportError(format=FILE_ERROR_EXPORT, filename=filename, error_msg=error.strerror) from None
```
phw, line 248: Should this better raise `from error`, to not hide the file access issue? Could be important for debugging.

picard/ui/widgets/editablelistview.py
```diff
@@ -209,7 +209,7 @@ def sort(self, column, order=QtCore.Qt.SortOrder.AscendingOrder):
     def get_display_name(self, item):  # pylint: disable=no-self-use
         return item
 
-    def rowCount(self, parent=QtCore.QModelIndex()):
+    def rowCount(self, parent=QtCore.QModelIndex()):  # noqa: B008
```
phw, line 212: I think this really should be rewritten to `parent=None` and doing a 

```
if parent is None:
    parent=QtCore.QModelIndex()
```

inside the function. I know https://doc.qt.io/qt-6/qabstractitemmodel.html#rowCount shows it that way it is defined here, but that's C++. And there the function providing the default argument does get invoked when the actual function is called (without the argument provided).

Same for `insertRows` and `removeRows` below.

pyproject.toml
```diff
@@ -1,3 +1,6 @@
+#[project]
+#requires-python = ">=3.8"
```
phw, line 2: Not sure about this, this could have influenced the checks codacy and/or pylint are performing. We should double check.

But definitely needs to be updated to 3.9

pyproject.toml
```diff
@@ -26,3 +29,62 @@ lines_after_imports = 2
...
+    "site-packages",
+    "venv",
+    "picard/ui/forms/ui_*.py",
+    "picard/resources.py",
+]
+
+line-length = 120
+indent-width = 4
+
+target-version = "py38"
```
phw, line 68: "py39"

test/test_bytes2human.py
```diff
@@ -81,7 +81,7 @@ def _create_testlist():
         values = [0, 1]
         for n in (1000, 1024):
             p = 1
-            for e in range(0, 6):
+            for _e in range(0, 6):
```
phw, line 84: Maybe better `_i` for consistency. I don't directly understand where the "e" is coming from, but "i" is typically a iterated counter.

setup.cfg
```diff
@@ -6,7 +6,7 @@
 # E241: multiple spaces after ','
 # E501: line too long (xx > 79 characters)
 # W503: line break occurred before a binary operator
-ignore = E127,E128,E129,E226,E241,E501,W503
+ignore = E127,E128,E129,E226,E241,E501,W503,N801,N802,N806
```
phw, line 9: `N802` and `N806` make sense to me as we sometimes need to follow Qt's camel case style.

But could we have "N801: class names should use CapWords convention", or are there important exceptions to this rule in our code base?

phw (changes requested):
I didn't know about ruff, but this definitely looks like a simplification. Would of course be interesting to have this integrated into codacy (https://github.com/astral-sh/ruff/issues/13391). Then we could really have only one tool.

It already found quite some things. Especially the missing `raise ... from ...` are already an annoyance when looking at some error logs with the "during handling of the error..." messages.

I added `astral-sh/ruff-action@v3` to the allowed actions, so we can see how this performs here on the next run of this branch.

phw:
> Question: should'nt we write a proper pyproject.toml with a project section now?

@zas Generally yes. But we also still need `setup.py` for our custom build stuff. And ideally we should not duplicate metadata.

Would be great if all the metadata can stay in `pyproject.toml`, though. At least https://packaging.python.org/en/latest/guides/modernize-setup-py-project/#how-to-handle-dynamic-metadata suggests that with modern setuptools both metadata from pyproject.toml and setup.py can be combined. So we can have at least the static data in pyproject.toml, while still have everything that is dynamic generated in setup.py.

setup.cfg
zas: Those are for flake8 and enabling them cause a lot of noise, I'd keep them as is for now. But feel free to experiment.

zas:
> But we also still need `setup.py` for our custom build stuff. And ideally we should not duplicate metadata.

About that, I think we should move to `uv`

PR #2584, opened 2025-02-24
Define project section of pyproject and use setuptools

zas:
python -m build seems to work at least

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,22 @@
-#[project]
-#requires-python = ">=3.9"
-
 [build-system]
-requires = ["setuptools>=62.4.0"]
+requires = [
+    "Babel>=2.10.0",
+    "PyInstaller==6.12.0",
+    "pytest>=8.1",
+    "setuptools>=62.4.0",
+]
 build-backend = "setuptools.build_meta"
+
+[project]
+name = "musicbrainz-picard"
+description = "MusicBrainz Picard"
+authors = [{name = "The MusicBrainz Team"}]
+license = {text = "GPL-2.0-or-later"}
+keywords = ["musicbrainz", "tagging", "audio"]
+dynamic = ["version", "readme", "classifiers", "requires-python", "dependencies"]
+
+[tool.setuptools.dynamic]
+version = {attr = "picard.PICARD_VERSION_STR_SHORT"}
 
 [tool.isort]
 sections = [


pyproject.toml
```diff
@@ -1,10 +1,23 @@
...
+requires = [
+    "Babel>=2.10.0",
+    "PyInstaller==6.12.0",
+    "pytest>=8.1",
+    "setuptools>=62.4.0",
+]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "musicbrainz-picard"
```
phw, line 11: This would change the package name, which currently is just "picard"

pyproject.toml
```diff
@@ -1,10 +1,23 @@
-#[project]
-#requires-python = ">=3.9"
```
phw, line 2: `requires-python` should be kept and added to the `[project]` section below.

pyproject.toml
```diff
@@ -1,10 +1,23 @@
...
+    "Babel>=2.10.0",
+    "PyInstaller==6.12.0",
+    "pytest>=8.1",
+    "setuptools>=62.4.0",
+]
 build-backend = "setuptools.build_meta"
 
+[project]
+name = "musicbrainz-picard"
+description = "MusicBrainz Picard"
```
phw, line 12: In current setup.py the `summary` corresponds to `description` and is currently "The next generation MusicBrainz tagger". Not saying we shouldn't change it, but maybe set it to "The MusicBrainz audio tagger" or "MusicBrainz Picard audio tagger" or such.

pyproject.toml
```diff
@@ -1,10 +1,23 @@
...
+authors = [{name = "The MusicBrainz Team"}]
+license = {text = "GPL-2.0-or-later"}
+keywords = ["musicbrainz", "tagging", "audio"]
+dynamic = ["version", "readme", "classifiers", "requires-python", "dependencies"]
```
phw, line 16: I think for this only "version", "readme" and for now "dependencies" should be listed.

"readme" can then be set in `tool.setuptools.dynamic` as 

```toml
readme = {file = ["README.md"], content-type = "text/markdown"}
```

and the corresponding code be removed from `setup.py`.

"requires-python" and "classifiers" should be set above as static data (and removed from setup.py.

pyproject.toml
```diff
@@ -1,10 +1,23 @@
...
+
```
phw, line 17: We can also add a section `project.urls` and remove the `url` from setup.py:

```toml
[project.urls]
Homepage = "https://picard.musicbrainz.org/"
Documentation = "https://picard-docs.musicbrainz.org/"
Repository = "https://github.com/metabrainz/picard.git"
Issues = "https://tickets.metabrainz.org/projects/PICARD/issues/"
Changelog = "https://picard.musicbrainz.org/changelog/"
```

pyproject.toml
```diff
@@ -1,10 +1,23 @@
-#[project]
-#requires-python = ">=3.9"
-
 [build-system]
-requires = ["setuptools>=62.4.0"]
+requires = [
+    "Babel>=2.10.0",
+    "PyInstaller==6.12.0",
+    "pytest>=8.1",
+    "setuptools>=62.4.0",
```
phw, line 6: In my test I had to use at least `setuptools>=setuptools>=66.1.0` for this to properly process.

Also not sure about the rest of the dependencies. They are all not strictly needed for a `pyhon -m build`. We only need them for a more complete build. Shall we keep them out of this PR and add them as needed later?

phw (changes requested):
Yes, we should do this. And I tested again and setuptools actually does properly combine the metadata here with the args set in setup.py in `setup(args)`.

But this also means when we move the data over all the corresponding fields in `setup.py` args should be removed. We won't support older setuptools without support for `pyproject.toml`.

pyproject.toml
phw: I just found out that we can also set `dependencies` in `tool.setuptools.dynamic` as 

```
dependencies = {file = ["requirements.txt"]}
```

This allows us to also remove `install_requires` from `setup.py`.

phw: But I guess we will rework dependencies when we move to e.g. ruff, so we can keep it in `setup.py` or move it here for now. Doesn't matter much.

pyproject.toml
phw: In `org.musicbrainz.Picard.appdata.xml.in` the short description is currently "MusicBrainz's music tagger". We really should settle on one being used everywhere :)

PR #2587, opened 2025-02-25
Use pytest and PEP 517 build

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

# Problem

Attempt to reduce the direct calls to `setup.py`.

# Solution

Remove `setup.py test` completely and replace the calls with `pytest`.

Build sdist / bdist packages with `python -m build`.

**Note:** As a side effect this removes the source builds in zip format, as PEP 517 only specifies .tar.gz as format. Do we need the zip source distribution? I think with public availability of the source in git and wide availability of extraction tools for .tar.gz even on Windows this is not really needed.

If we want to provide it we could add a CI step to also create a zip file.

### DIFF
--- a/scripts/package/run-sdist-test.sh
+++ b/scripts/package/run-sdist-test.sh
@@ -5,7 +5,7 @@
 set -e
 
 rm -rf dist
-python3 setup.py sdist
+python -m build --sdist
 cd dist
 SDIST_ARCHIVE=$(echo picard-*.tar.gz)
 tar xvf "$SDIST_ARCHIVE"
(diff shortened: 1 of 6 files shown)


scripts/package/run-sdist-test.sh
```diff
@@ -5,9 +5,9 @@
 set -e
 
 rm -rf dist
-python3 setup.py sdist
+python3 -m build --sdist
```
zas, line 8: I saw we have multiple occurrences of `python3`, since we require Python >= 3 anyway, I guess we can just renamed those.
What do you think?

phw: Yes, I think this is true. We have this a lot on the mac build also. Not sure there. But my preferred way would be to get rid of all the python3 calls. I changed it to just "python" here in this file now. Let's do this more generally in a separate PR.

zas (approved):
LGTM

PR #2588, opened 2025-02-25
Pyproject metadata

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

# Problem

This is one step in the goal to nove package metadata into `pyproject.toml` and reduce code in `setup.py`.

# Solution

- Define `[project]` section in `pyproject.toml` and declare package metadata there. Remove the corresponding data from `setup.py`
- Remove obsolete code from `setup.py`
- Define setuptools package and extension in `pyproject.toml` and remove custom code from `setup.py`
- Unified the short app description to be the same in Python package metadata and appstream metadata.

Not handled here:

- Requirements are still defined in `requirements.txt` and read from there. This can be changed once we move to e.g. ruff
- Entry points: This is currently highly system dependent and for now handled dynamically by `setup.py`.
- Data files: Handling of locale translation files will be changed and done in a separate PR. The additional handling for system wide assets (which are not directly loaded by the application) on Unix systems (.desktop file, app icon, appstream metadata) will most likely stay in `setup.py` as part of the packaging process.
- Custom commands in `setup.py`: In he future those could be moved to a separate task handler, at least the ones that are not invoked during build. Custom commands that are invoked during build likely better stay in `setup.py` to avoid additional build-time dependencies.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,10 +38,18 @@ Repository = "https://github.com/metabrainz/picard.git"
 Issues = "https://tickets.metabrainz.org/projects/PICARD/issues/"
 Changelog = "https://picard.musicbrainz.org/changelog/"
 
+[tool.setuptools]
+ext-modules = [
+    {name = "picard.util._astrcmp", sources = ["picard/util/_astrcmp.c"]},
+]
+
 [tool.setuptools.dynamic]
 version = {attr = "picard.PICARD_VERSION_STR_SHORT"}
 readme = {file = ["README.md"], content-type = "text/markdown"}
 dependencies = {file = ["requirements.txt"]}
+
+[tool.setuptools.packages.find]
+include = ["picard*"]
 
 [tool.isort]
 sections = [
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -1,9 +1,47 @@
-#[project]
-#requires-python = ">=3.9"
-
 [build-system]
-requires = ["setuptools>=62.4.0"]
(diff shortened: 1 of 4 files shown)


pyproject.toml
```diff
@@ -1,10 +1,56 @@
...
+
+[tool.setuptools]
+ext-modules = [
+    {name = "picard.util._astrcmp", sources = ["picard/util/_astrcmp.c"]},
+]
+
+[tool.setuptools.dynamic]
+version = {attr = "picard.PICARD_VERSION_STR_SHORT"}
+readme = {file = ["README.md"], content-type = "text/markdown"}
+dependencies = {file = ["requirements.txt"]}
```
zas, line 49: According to https://stackoverflow.com/questions/76722680/what-is-the-best-way-to-combine-conda-with-standard-python-packaging-tools-e-g we can even add optional dependencies:

```toml
[project]
dynamic = ["version", "readme", "dependencies", "optional-dependencies"]
```

```toml
[tool.setuptools.dynamic.optional-dependencies]
dev = { file = ["requirements-dev.txt"] }
build = { file = ["requirements-build.txt"] }
```

zas (approved):
LGTM

pyproject.toml
phw: Yes, but let's do this once we know more about how we handle packages in the future. Right now we don't benefit from this much :)

PR #2590, opened 2025-02-25
Refactor id3._load()

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -285,21 +285,7 @@ class ID3File(File):
         for frame in tags.values():
             frameid = frame.FrameID
             if frameid in self.__translate:
-                name = self.__translate[frameid]
-                if frameid.startswith('T') or frameid in {'GRP1', 'MVNM'}:
-                    for text in frame.text:
-                        if text:
-                            metadata.add(name, text)
-                elif frameid == 'COMM':
-                    for text in frame.text:
-                        if text:
-                            if frame.lang == 'eng':
-                                name = '%s:%s' % (name, frame.desc)
-                            else:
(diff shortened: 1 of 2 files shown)


picard/formats/id3.py
```diff
@@ -269,142 +291,209 @@ def build_TXXX(self, encoding, desc, values):
...
+    def _process_frame(self, frame, metadata, config_params):
+        """Process an ID3 frame and add its data to the metadata."""
+        frameid = frame.FrameID
+
+        # Get appropriate processor from dictionary
+        processor = self._frame_processors.get(frameid)
+        if processor:
+            processor(frame, metadata, config_params)
+
+    def _init_load(self, filename):
```
phw, line 317: Where is this called?

phw (commented):
This is definitely a good approach to refactor those huge load/save functions.

If we think about also having this for save we could also encapsulate the functionality into classes with a load and save method. We would have one frame processor class per general frame type we need to handle.

picard/formats/id3.py
zas: https://github.com/metabrainz/picard/pull/2590/files#diff-1bde7f3f6de30b8d739c17c9e528a07ee6c5f511ac73846855fdfcb36dd39cb2R294

rdswift (approved):
Very nice!

phw (approved):
Much needed improvement, will make changing / adding tags much easier. Let's continue with this approach

PR #2591, opened 2025-02-25
PICARD-2003: Include locales as package data inside the main package.

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2003

If picard got installed by `pip install picard` the translations where unavailable. That is because the locale files got included as a data files outside the package folder, and there is no default way to get this data folder.

This PR ensures locales are available also in pip installs by moving the locale as data files into the picard package.

# Solution

Move the locale directory holding the *.mo files into the picard package and include them as data files in the source package and wheel. Adjust the loading to use this locale dir by default.

This simplifies the code to build the locales, as there are no longer separate build target directories. This also removes the need to run `setup.py build_locales -i` after `setup.py build` for local development.

Separate handling is needed for pyinstaller "frozen" install. Oherwise the location of the locales is now always the same.

See also https://setuptools.pypa.io/en/latest/userguide/datafiles.html#interplay-between-these-keywords

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -50,7 +50,7 @@ jobs:
       if: always()
       timeout-minutes: 30
       run: |
-        python setup.py build_locales -i
+        python setup.py build_locales
         pip install pytest pytest-randomly pytest-cov
         pytest --verbose --cov=picard --cov-report xml:coverage.xml test
     - name: Submit code coverage to Codacy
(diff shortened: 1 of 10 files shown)


phw:
@zas I hope we can keep it this way. Should someone really insist on having a way to install locales in `/use/share/` again we will need to add something back, though. But I hope we don't need that, and if we do I think we can have it simpler than before.

PR #2593, opened 2025-02-27
Refactor id3 save

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -540,7 +540,7 @@ class ID3File(File):
                 for value in values:
                     tipl.people.append([self._rtipl_roles[name], value])
             elif name == 'musicbrainz_recordingid':
-                tags.add(id3.UFID(owner="http://musicbrainz.org", data=bytes(values[0], 'ascii')))
+                self._save_musicbrainz_recording_id(tags, values)
             elif name == '~rating':
                 rating_user_email = id3_rating_user_email(config)
                 # Search for an existing POPM frame to get the current playcount
@@ -886,6 +886,10 @@ class ID3File(File):
             # If the text does not contain any timestamps, the tag is not added
             if sylt_lyrics:
                 tags.add(id3.SYLT(encoding=encoding, lang=lang, format=2, type=1, desc=desc, text=sylt_lyrics))
+
+    def _save_musicbrainz_recording_id(self, tags, values):
+        """Save MusicBrainz recording ID to UFID frame."""
+        tags.add(id3.UFID(owner="http://musicbrainz.org", data=bytes(values[0], 'ascii')))
 
 
 class MP3File(ID3File):
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -507,30 +507,11 @@ class ID3File(File):
             tags.delall('APIC')
 
         encoding = Id3Encoding.from_config(config.setting['id3v2_encoding'])
-
-        # This is necessary because mutagens HashKey for APIC frames only
-        # includes the FrameID (APIC) and description - it's basically
-        # impossible to save two images, even of different types, without
-        # any description.
-        counters = Counter()
-        for image in images_to_save:
-            desc = desctag = image.comment
-            if counters[desc] > 0:
-                if desc:
-                    desctag = "%s (%i)" % (desc, counters[desc])
-                else:
-                    desctag = "(%i)" % counters[desc]
-            counters[desc] += 1
-            tags.add(id3.APIC(encoding=Id3Encoding.LATIN1,
-                              mime=image.mimetype,
-                              type=image.id3_type,
-                              desc=id3text(desctag, Id3Encoding.LATIN1),
-                              data=image.data))
-
         tmcl = mutagen.id3.TMCL(encoding=encoding, people=[])
         tipl = mutagen.id3.TIPL(encoding=encoding, people=[])
 
         self._save_track_disc_movement_numbers(tags, metadata)
+        self._save_images(tags, images_to_save)
         for name, values in metadata.rawitems():
             values = [id3text(v, encoding) for v in values]
             name = id3text(name, encoding)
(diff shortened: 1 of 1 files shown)


zas:
Please review carefully.
Not finished yet, but early reviews can help to avoid future problems.
Note that tests should pass on each step.

picard/formats/id3.py
```diff
@@ -869,6 +691,304 @@ def _parse_lrc_text(self, text):
...
+        (lang, desc) = parse_subtag(name)
+        for value in values:
+            sylt_lyrics = self._parse_lrc_text(value)
+            # If the text does not contain any timestamps, the tag is not added
+            if sylt_lyrics:
+                tags.add(id3.SYLT(encoding=encoding, lang=lang, format=2, type=1, desc=desc, text=sylt_lyrics))
+
+    def _save_musicbrainz_recording_id(self, tags, values):
+        """Save MusicBrainz recording ID to UFID frame."""
+        tags.add(id3.UFID(owner="http://musicbrainz.org", data=bytes(values[0], 'ascii')))
```
rdswift, line 766: I'm wondering if it makes sense to store the "http://musicbrainz.org" as a constant and serve it from there?  I'm pretty sure it (or the https:// version) is used elsewhere in the code.  If so, it may be worthy of a separate PR because it isn't directly related to what you're doing here.

rdswift (commented):
I really like the way this is looking.  It is very easy to follow and should greatly simplify maintenance.

picard/formats/id3.py
zas: Done in https://github.com/metabrainz/picard/commit/1d95d9f087b923907b0012e55fd1c6a47d027423

zas:
@phw can you double check https://github.com/metabrainz/picard/pull/2593/commits/592dd46fcf91f1e8f8a81899c9e05132cb73177b ?

I *think* we don't need real name in this method, it simplifies code but I'm not totally sure about this move (tests are passing though).

I also wonder about https://github.com/metabrainz/picard/blob/592dd46fcf91f1e8f8a81899c9e05132cb73177b/picard/formats/id3.py#L535 vs https://github.com/metabrainz/picard/blob/592dd46fcf91f1e8f8a81899c9e05132cb73177b/picard/formats/id3.py#L922 

In one case we match 'performer' or 'performer:' but not in the other, not sure if this is intended.

phw (commented):
@zas Same as for loading this improves the code quite a bit

> In one case we match 'performer' or 'performer:' but not in the other, not sure if this is intended.

This actually looks like a bug to me, where deleting without role would cause problems. We should test this. It looks like we save just `performer` as TIPL with role being set to "performer".  But deleting just `performer` does not seem to get handled.

> EDIT: Actually that's not a 1-1 relationships due to 

Yes, that's a problem. Because of the complexity and data getting written to multiple fields, or read from multiple fields etc. it is not a 1:1 always.

zas:
> 
> This actually looks like a bug to me

Actually my feelings too, and likely not the only in this part of the code, I'll avoid to make any behavior changes for now, but we'll need to carefully review this part (and likely improve unit tests for it).

zas:
> This actually looks like a bug to me, where deleting without role would cause problems. We should test this. It looks like we save just `performer` as TIPL with role being set to "performer". But deleting just `performer` does not seem to get handled.

@phw Can you check if you can manage to reproduce the potential bug? This tag removal code looks fragile to me, but I'm not sure how to produce good unit tests for this at this point. Help welcome.

zas:
I rebased, dropped reverted commits, I think there's enough changes in this PR already.
The changes made shouldn't change any behavior and tests are all passing; but a careful review is still welcome.

picard/formats/id3.py
```diff
@@ -499,259 +500,86 @@ def _load_popm_frame(self, frame, metadata, config_params):
     def _save(self, filename, metadata):
         """Save metadata to the file."""
         log.debug("Saving file %r", filename)
+
+        # TODO: check _get_tags vs encode_filename(), not sure if we can pass it directly using
+        # encoded_filename below
```
phw, line 505: That part is a bit blurry. We don't use `encode_filename` consistently. E.g. `_get_tags` is implemented to call `encode_filename`, but `_get_file` isn't and instead `encode_filename` is done by the caller.

I'm not fully sure what your TODO comment aims at. But we could make this consistent.

phw: Also I'm not sure we are using `encode_filename` consistently everywhere where a file is accessed, and I'm not even sure it is needed. The implementation of `encode_filename` suggest it is always needed for macOS, but I'm not aware of any errors on macOS due to a missing call.

E.g. just in id3.py the implementation of `NonCompatID3File._get_tags` doesn't call `encode_filename`. I think Picard is generally not well tested on file systems not using Unicode encoding, and there are for sure some bugs about this. And actually we have some bug reports about file name encoding, e.g. PICARD-233, PICARD-644, PICARD-1287, PICARD-2055. Some of them are rather old and it is unclear whether this still applies. But at least PICARD-2055 indicates the issue is happening even despite or probably because `encode_filename` is being used.

phw (approved):
Good improvement, LGTM

picard/formats/id3.py
zas: My TODO is because I found inconsistencies while refactoring, to me code needs to be audited. And what you're writing above clearly shows that too ;)
But that's another PR anyway.

We should add a test env (in our action matrix) specifically made to simulate non-unicode env, I guess just running current tests in such env would reveal some issues.
Perhaps we should even build a non-unicode loopback device (typical vfat cp1252 for example) and read/write files we use in tests from there.
Dunno exactly but I totally agree Picard lacks of testing in such env.

PR #2595, opened 2025-03-05
Packaging fixes

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

This fixes two issues in the PyPI packages recently introduced with #2588

1. Locales where not included in the wheel if building with e.g. "python -m build". As this is building in a clean environment the locales did not end up in the local source tree and where not used. Locales need to be put into the build-lib dir. For local source builds running `python setup.py build_locales` still works and writes the locales to `picard/locale` as expected.
2. On Windows setuptools complained that "gui-scripts" in pyproject.toml needs to be dynamic. As gui-scripts currently is only used on Windows this was not obvious during testing.

### DIFF
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -71,8 +71,9 @@ jobs:
         python setup.py clean
     - name: Build
       run: |
-        python setup.py build --build-number=$BuildNumber
-        python setup.py build_ext
+        python setup.py build --build-number=$BuildNumber --disable-locales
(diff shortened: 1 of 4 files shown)


phw:
@zas This is now properly packaging the locales in both wheels and PyInstaller created packages (the latter is running on local source tree).

zas (approved):
LGTM

PR #2600, opened 2025-03-10
PICARD-3038: Fix crashes and editing of wrong indexes in edit tag dialog

phw:
The visual representation could differ from the underlying storage. This could cause index out of range exceptions. Also empty values were omitted in display, causing the wrong indexes to update when editing.

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3038

This change addresses crashes and editing issues in the edit tag dialog. Specifically this fixes two issues:

- When removing all values from a tag, then switching to another tag and back to the original, editing or removing the shown empty row causes an "index out of range" exception.
- When clearing values (empty value, but keeping the row, then switching to another tag and back to the original, editing would update the wrong indexes of the underlying stored data (because empty values were not displayed but still present in the underlying data).

See my comment in PICARD-3038 for more details on how to reproduce both issues.

# Solution

1. Keep empty rows when switching between tags. This keeps the same display if user switches between tags and avoids the displayed data being stored differently from the data. Empty rows will still be removed on saving.
2. Do not add an empty `[""]` when preparing the displayed data. This was the cause for the crash after removing all values, as the underlying data was an empty list. Also this is not needed, as `add_or_edit_value()` takes care of adding and displaying at least one empty row.

For 1. an alternative way would be to cleanup empty rows in stored data when switching tags. But my reasoning is that the user, if they emptied a row and just switch back and forth between tags, should still see the same view.

I could not reproduce the exact issue as originally reported in PICARD-3038, where there is an unexpected empty row added, so I cannot verify it is solved. As the exception is also about "index out of range" it clearly is also caused by differences in data storage in `modified_tags`  and the visual representation. I think there are good chances this is hence also fixed by these changes, as it eliminates the two reasons I could find for differences in data.

### DIFF
--- a/picard/ui/edittagdialog.py
+++ b/picard/ui/edittagdialog.py
@@ -233,6 +233,7 @@ class EditTagDialog(PicardDialog):
         model = self.ui.value_list.model()
         model.rowsInserted.connect(self.on_rows_inserted)
         model.rowsRemoved.connect(self.on_rows_removed)
+        model.rowsMoved.connect(self.on_rows_moved)
 
     def keyPressEvent(self, event):
         if (event.modifiers() == QtCore.Qt.KeyboardModifier.NoModifier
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2601, opened 2025-03-10
Simplify modified tags cleanup when exiting the dialog

zas:
- use a generator for modified tags and their non-empty values
- drop useless intermediate methods

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/edittagdialog.py
+++ b/picard/ui/edittagdialog.py
@@ -514,30 +514,18 @@ class EditTagDialog(PicardDialog):
             list(self.metadata_box.tag_diff.new[self.tag]) or [""]
         )
 
-    def _cleanup_modified_tags(self):
-        """Remove empty values from modified tags."""
+    def _modified_tags_without_empty_values(self):
+        """Generate each modified tag and its non-empty values."""
         for tag, values in self.modified_tags.items():
-            self.modified_tags[tag] = [v for v in values if v]
-
-    def _update_object_metadata(self, obj, modified_tags):
-        """Update the metadata for a single object with the modified tags.
-
-        Args:
-            obj: The object to update
-            modified_tags: Iterable of (tag, values) pairs to update
-        """
-        for tag, values in modified_tags:
-            obj.metadata[tag] = list(values)
-        obj.update()
+            yield (tag, [v for v in values if v])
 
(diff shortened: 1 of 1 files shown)


picard/ui/edittagdialog.py
```diff
@@ -514,30 +514,18 @@ def _modified_tag(self):
...
-        """Update the metadata for a single object with the modified tags.
-
-        Args:
-            obj: The object to update
-            modified_tags: Iterable of (tag, values) pairs to update
-        """
-        for tag, values in modified_tags:
-            obj.metadata[tag] = list(values)
-        obj.update()
+            yield (tag, list(v for v in values if v))
```
phw, line 520: If it's supposed to be a map I'd use the list comprehension directly, instead of iterator and `list()`:

```
yield (tag, [v for v in values if v])
```

picard/ui/edittagdialog.py
```diff
@@ -514,30 +514,18 @@ def _modified_tag(self):
             list(self.metadata_box.tag_diff.new[self.tag]) or [""]
         )
 
-    def _cleanup_modified_tags(self):
-        """Remove empty values from modified tags."""
+    def _modified_tags(self):
```
phw, line 517: The naming is a bit unclear I think. Calling this `_modified_tag` is very similar to `_modified_tags`, but they do behave differently. While one gives you all the values for the currently active tag for editing, the other gives you all tags to be saved.

Not sure how to best rename them. Maybe the original `_modified_tag` should be renamed to something like `_current_tag` or `_current_tag_values`.

phw (changes requested):
That's a great catch that `Metadata.update` can be used directly.

picard/ui/edittagdialog.py
zas: Done, rebased

picard/ui/edittagdialog.py
zas: For now, I'll renamed it to _modified_tags_without_empty_values() so the patch doesn't change lines you're working on, but I agree we should rethink names in this file.

phw (approved):
LGTM

PR #2602, opened 2025-03-11
Fix empty tags being removed from tag selection

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

The tag editor removes tags which are not standard and store only a single empty value from the tag selection list. But this does not take into account tags where all values have been removed or where there are multiple empty values (which still counts as an empty tag in the end).

# Solution

Check if the tag values are fully empty. Also some more refactoring.


zas (approved):
LGTM

PR #2603, opened 2025-03-11
PICARD-2192: Fix macos arm builds

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2192

# Solution

Actually build on an arm64 runner. Previously the build pulled the x86_64 binaries for Qt6, which prevented the app to run. Also this way the code to test running the executable can be actually be run during build.

This also changes the install to use Python via actions/setup-python. This is consistent with other builds we have, and it fixes an issue with the system wide installation specifically on ARM refusing to install PyPI packages.

The builds actually should use a venv, but that's another story :D

### DIFF
--- a/.github/workflows/package-macos.yml
+++ b/.github/workflows/package-macos.yml
@@ -13,20 +13,14 @@ jobs:
         - macos-deployment-version: '11.0'
           runner: 'macos-13'
           target-arch: 'x86_64'
-          python-version: 3.13.1-macos11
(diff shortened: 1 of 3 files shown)


phw:
This time I actually was able to test run the build on an actual ARM64 mac. It works (and as expected launches faster than the Intel build)

zas (approved):
Good job!
LGTM

PR #2607, opened 2025-03-13
PICARD-3044: Fix ~filepath sometimes being editable in metadata list

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3044

# Solution

Always set the QTableWidgetItem flags, as items can be reused. If a previously readable QTableWidgetItem was reused the field meant to be readonly was suddenly writable.


zas (approved):
Tested, works as expected.
Thanks for the fix.

PR #2608, opened 2025-03-13
Refactor MetadataBox

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -575,7 +575,7 @@ class MetadataBox(QtWidgets.QTableWidget):
                     new_values = list(orig_values or [""])
 
                 removed = tag in new_metadata.deleted_tags
-                tag_diff.add(tag, orig_values, new_values, removed=removed, top_tags=top_tags_set)
+                tag_diff.add(tag, old=orig_values, new=new_values, removed=removed, top_tags=top_tags_set)
 
             tag_diff.add('~length', str(orig_metadata.length), str(new_metadata.length),
                          removable=False, readonly=True)
@@ -584,7 +584,7 @@ class MetadataBox(QtWidgets.QTableWidget):
                     new_filename = file.make_filename(file.filename, new_metadata)
                 else:
                     new_filename = file.filename
-                tag_diff.add('~filepath', [file.filename], [new_filename], removable=False, readonly=True)
+                tag_diff.add('~filepath', old=[file.filename], new=[new_filename], removable=False, readonly=True)
 
         for track in tracks:
             if track.num_linked_files == 0:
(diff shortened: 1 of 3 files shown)


picard/ui/metadatabox/__init__.py
```diff
@@ -647,6 +640,12 @@ def _update_items(self, result=None, error=None):
         readonly_item_flags = QtCore.Qt.ItemFlag.ItemIsSelectable | QtCore.Qt.ItemFlag.ItemIsEnabled
         editable_item_flags = readonly_item_flags | QtCore.Qt.ItemFlag.ItemIsEditable
         alignment = QtCore.Qt.AlignmentFlag.AlignLeft | QtCore.Qt.AlignmentFlag.AlignTop
+        colors = {
+            TagStatus.UNCHANGED: self.palette().color(QtGui.QPalette.ColorRole.Text),
+            TagStatus.REMOVED: QtGui.QBrush(interface_colors.get_qcolor('tagstatus_removed')),
+            TagStatus.ADDED: QtGui.QBrush(interface_colors.get_qcolor('tagstatus_added')),
+            TagStatus.CHANGED: QtGui.QBrush(interface_colors.get_qcolor('tagstatus_changed'))
+        }
```
phw, line 648: Setting up the colors list looks like quite some work. Given that the update of the metadata is a performance critical part, wouldn't it be better to keep this outside the function?

If we need to react on config changes to use the new colors we could have `InterfaceColors` emit a signal whenever colors change, and MetadataBox could update the colors (and even refresh the view).

picard/ui/metadatabox/tagdiff.py
```diff
@@ -241,39 +252,39 @@ def is_readonly(self, tag):
...
+            old: The old tag value(s).
+            new: The new tag value(s).
             removable (bool): Whether the tag can be removed.
             removed (bool): Whether the tag was marked as removed.
             readonly (bool): Whether the tag is read-only.
             top_tags (set): Set of top level tags
         """
-        if orig_values:
-            self.orig.add(tag, orig_values)
+        if old:
```
phw, line 268: We call it "original" elsewhere in the code base, and also in the UI. I would stick with that.

phw (commented):
Looks good. I'm just unsure about the renaming of "original" to "old". This both is inconsistent with other places in the code base, and I find it also less clear / too generic.

picard/ui/metadatabox/tagdiff.py
zas: "old"/"new" are very common when it comes to "diff".
I agree we use orig/original in the rest of the code, but in the context of TagDiff it doesn't make much sense because we can compare any set of tags (not only "original tags" vs "new tags").
For any coder, old to new direction is obvious, while orig to new is only obvious for one familiar with Picard code and the meaning of "orig".

phw: Ok, that's convincing

phw (approved):

PR #2609, opened 2025-03-13
PICARD-3045: Fail safe if the parameter passed to format_time cannot be converted to a float

StevilKnevil:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
Added exception handling for passing bad data to format_time()

# Problem
Calls to `format_time()` can pass data that cannot be converted to a float, causing a crash.
* JIRA ticket: PICARD-3045

# Solution
Wrapping
```ms = float(ms)```
With exception handling means that callers can pass non-numerical strings or unexpected datatypes without causing a crash

### DIFF
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -274,8 +274,14 @@ class MetadataBox(QtWidgets.QTableWidget):
                 value = self.tag_diff.orig[tag]
             elif column == self.COLUMN_NEW:
                 value = self.tag_diff.new[tag]
+
             if tag == '~length':
-                value = [format_time(value or 0), ]
+                try:
+                    value = float(value)
+                    value = [format_time(value or 0), ]
+                except (TypeError, ValueError):
+                    value = ['']
+                
             if value is not None:
                 self.tagger.clipboard().setText(MULTI_VALUED_JOINER.join(value))
                 self.clipboard = value
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -278,7 +278,8 @@ class MetadataBox(QtWidgets.QTableWidget):
             if tag == '~length':
                 try:
                     value = [format_time(value or 0), ]
-                except (TypeError, ValueError):
+                except (TypeError, ValueError) as why:
+                    log.warning(why)
                     value = ['']
 
             if value is not None:
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -280,7 +280,7 @@ class MetadataBox(QtWidgets.QTableWidget):
                     value = [format_time(value or 0), ]
                 except (TypeError, ValueError):
                     value = ['']
-                
+
             if value is not None:
                 self.tagger.clipboard().setText(MULTI_VALUED_JOINER.join(value))
                 self.clipboard = value
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -40,6 +40,7 @@ from PyQt6 import (
     QtWidgets,
 )
 
+from picard import log
 from picard.album import Album
 from picard.browser.filelookup import FileLookup
 from picard.cluster import Cluster
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -277,7 +277,6 @@ class MetadataBox(QtWidgets.QTableWidget):
 
             if tag == '~length':
                 try:
-                    value = float(value)
                     value = [format_time(value or 0), ]
                 except (TypeError, ValueError):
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

File: picard/util/__init__.py [hunk 1]
```diff
@@ -301,7 +301,12 @@ def samefile(path1, path2):
 
 def format_time(ms, display_zero=False):
     """Formats time in milliseconds to a string representation."""
-    ms = float(ms)
+    # Ensure that the input represents a number
+    try:
+        ms = float(ms)
+    except (TypeError, ValueError):
```
zas, line 307: I don't think hiding the exception is the way to fix this issue.

StevilKnevil: Should we throw the exception instead, and make all callers handle it?

zas: How do you reproduce it? Following indications on PICARD-3045 doesn't cause any crash here with current master. Provide the actual trace.

StevilKnevil: Updated the ticket

StevilKnevil:
> This approach isn't the correct one to fix the issue, we should either handle copying multiple values (it could be convenient sometimes), or not allowing at all. We should definitively let `format_time()` raises exceptions though.

Have updated to handle this case in the calling code

picard/ui/metadatabox/__init__.py
```diff
@@ -274,8 +274,14 @@ def _copy_value(self):
                 value = self.tag_diff.orig[tag]
             elif column == self.COLUMN_NEW:
                 value = self.tag_diff.new[tag]
+
             if tag == '~length':
-                value = [format_time(value or 0), ]
+                try:
+                    value = float(value)
```
phw, line 280: No need for the float conversion here, `format_time` takes care of this

picard/ui/metadatabox/__init__.py
```diff
@@ -274,8 +274,14 @@ def _copy_value(self):
...
+                    value = [format_time(value or 0), ]
+                except (TypeError, ValueError):
+                    value = ['']
```
phw, line 283: This is an interesting issue, and I think it runs even deeper. `value` nearly always is a list of strings. If you copy other values than length it is a list (no matter if the tag contains one or multiple entries). If you use copy on multiple selected files for a tag which has different values in the selected items, than `value` is also a list with a single empty string (`[""]`).

Only when copying the length it isn't a list but a single string. This already is odd, not sure why it isn't consistently a list of strings. @zas any idea? Should we change this?

Also yes, we could extend this to also handle copying of multiple distinct values on multiple selected items. But that's entirely separate from the bugfix :)

phw (commented):
Thanks. I think exception handling at this place is the right approach. Patch seems fine.

phw (approved):

phw (changes requested):
Just the unneeded call to `float` conversion should be removed.

picard/ui/metadatabox/__init__.py
zas: > Only when copying the length it isn't a list but a single string. This already is odd, not sure why it isn't consistently a list of strings. @zas any idea? Should we change this?

Yes, this is something weird in the code and we have many places where it leads to special handling. It isn't very logical and I'm not sure from where it comes.
Changing it will need a lot of care, but IMHO this is something we should do.



> Also yes, we could extend this to also handle copying of multiple distinct values on multiple selected items.

Yes, separate PR. But the question above is related, one tag value is a list of strings, so multiple objects tag value is a list of lists of strings (and no exception for `~length`).

picard/ui/metadatabox/__init__.py
```diff
@@ -274,8 +274,13 @@ def _copy_value(self):
                 value = self.tag_diff.orig[tag]
             elif column == self.COLUMN_NEW:
                 value = self.tag_diff.new[tag]
+
             if tag == '~length':
-                value = [format_time(value or 0), ]
+                try:
+                    value = [format_time(value or 0), ]
+                except (TypeError, ValueError):
```
zas, line 281: We should not hide those exceptions totally, especially if we want to change how `~length` works (see below).
At least log a warning in case of exception:


```suggestion
                except (TypeError, ValueError) as why:
                    log.warning(why)
```

Also, we expect `value` to contain a positive integer number (time in milliseconds), so the actual conversion should use `int()` and check it is a positive number (because it will happily convert -1000 without raising any error). This can be done in another PR, and matching tests will have to be added (test_utils.py has tests for `format_time()` already.

To me, `format_time()` should only be called with values that can actually be converted, so we should determine if we have one or more than one item in the selection and act accordingly (call `format_time()` for each value).

StevilKnevil: Will raise a new PR for making `format_time()` expect positive ints

phw (approved):

PR #2610, opened 2025-03-14
Ensure that format_time() only expects integer values for ms

StevilKnevil:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [X] Refactoring
  * [ ] Other

# Problem
`format_time()` should only expect to work for positive numbers, however it currently handles negative values and returns negative times.

# Solution
Raise a ValueError if the passed value is a negative.

Added function documentation and tests.

### DIFF
--- a/test/test_utils.py
+++ b/test/test_utils.py
@@ -258,10 +258,17 @@ class FormatTimeTest(PicardTestCase):
         self.assertEqual("0:00", util.format_time(0, display_zero=True))
         self.assertEqual("3:00", util.format_time(179750))
         self.assertEqual("3:00", util.format_time(179500))
+        self.assertEqual("3:00", util.format_time(179500.5))
         self.assertEqual("2:59", util.format_time(179499))
         self.assertEqual("59:59", util.format_time(3599499))
         self.assertEqual("1:00:00", util.format_time(3599500))
         self.assertEqual("1:02:59", util.format_time(3779499))
+        with self.assertRaises(ValueError):
+            util.format_time(-40000)
+        with self.assertRaises(ValueError):
+            util.format_time("Hello World")
+        with self.assertRaises(TypeError):
+            util.format_time(list())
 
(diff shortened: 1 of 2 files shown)


test/test_utils.py
```diff
@@ -258,10 +258,13 @@ def test(self):
         self.assertEqual("0:00", util.format_time(0, display_zero=True))
         self.assertEqual("3:00", util.format_time(179750))
         self.assertEqual("3:00", util.format_time(179500))
+        self.assertEqual("3:00", util.format_time(179500.5))
         self.assertEqual("2:59", util.format_time(179499))
         self.assertEqual("59:59", util.format_time(3599499))
         self.assertEqual("1:00:00", util.format_time(3599500))
         self.assertEqual("1:02:59", util.format_time(3779499))
+        with self.assertRaises(ValueError):
+            util.format_time(-40000)
```
zas, line 267: Please also test it raises `TypeError` if called with a value than can't be converted to `float`.
Else the PR looks good, maybe squash your commits ;)

zas (changes requested):

test/test_utils.py
StevilKnevil: Settled on float being a valid input (and added a test for that above), no reason not to handle fractional milliseconds any differently to a small number of milliseconds, they all get rounded below 1 second anyway.

Will do on the squash.

zas: The test for `TypeError` is still missing:

 ```python
with self.assertRaises(TypeError):
    util.format_time(list())
```

zas (changes requested):

StevilKnevil:
Oh yes! I was just thinking about float! On it

test/test_utils.py
StevilKnevil: Fixed!

PR #2613, opened 2025-03-14
PICARD-2287: Copy and Paste multiple Tags

StevilKnevil:
# Summary

* This is a…
  * [ ] Bug fix
  * [X] Feature addition
  * [ ] Refactoring
  * [ ] Other

Multiple tags can be selected, and then copied and pasted between files.

PICARD-2287

# Solution

When multiple tags are selected, copying them to the clipboard turns them into a JSON string with mimetype `application/json;contents=picard-tags`. Pasting this mimetype will iterate over all the tags in the JSON, and will apply the `new_value` to the current file. If `new_value` doesn't exist for the given tag in the clipboard then it will use `original_value`

Example of the JSON on the clipboard:

```
{
	"album": {
		"new_value": ["test"],
		"original_value": [""]
	}, 
	"title": {
		"new_value": ["My Song"], 
		"original_value": ["My Song"]
	},
	"~length": {
		"new_value": ["3:18"], 
		"original_value": ["3:18"]
	}
}
```

This means that Copy and Paste now need to be added to the context menu when there is more than one cell selected.

The behaviour when a single cell is selected is unchanged, i.e. copy the value as plain text.


zas:
Apart coding style issues, please create a matching ticket, and be sure to prefix PR title with PICARD-xxxx (it creates a link between jira tickets & github PRs).
In the ticket, explain why this feature is interesting, and how it works. Title should be clear (think it will appear in Changelogs).

rdswift:
Also, is this something that should be included / documented in the Picard User Guide?  If so, please add a documentation ticket with a reference to this PR.  (I noticed that you removed that part of the PR template for this PR.)  Thanks.

StevilKnevil:
Sorry - I meant to mark this as draft, while garnering feedback!

Sophist-UK:
The current code (at the time of this comment) appears to be designed to meet the individual use case of @StevilKnevil, and as such is great as a proof of concept but it is not IMO generically applicable and should not be merged into Picard in its current form.

I have copied the following comments I made from the Matrix / IRC discussion with my concerns:

1. I can see why you might want to put the original value in the JSON in order to confirm that you are replacing the correct value, but that would be a very specific use case and not the general one where only the new values are needed. The JSON should be simplified to just be \`{"title": "Waterloo", "artist": "ABBA"}.

2. I think you need to handle values as both arrays as in your original JSON and plain strings as in point 1.

3. This JSON is limited to a single track which makes sense for pasting into the metadata area of a single selected track. But do we want to extend the paste to copying to multiple selected tracks, and if so then A) should this be limited to album-related metadata and B) do we want to extend the JSON schema to allow for several tracks specific data to be specified using e.g. some sort of MBIDs (which IMO will be complicated and shouldn't be attempted at this point - but you should make sure now that the JSON schema can be extended later).

4. I am not sure that I agree with @outsidecontext's suggestion to use QT's special clipboard functionality - users **will** want to 
copy and paste between applications, so you want to have compatibility here - and that means either using a pre-defined JSON clipboard type or even thinking about implementing several common clipboard formats e.g. plain text, JSON, perhaps a spreadsheet-specific format or XML too or perhaps there is a key/value pair format.

5. Preserving Copy backwards compatibility probably should mean **NOT** copying JSON to the plain clipboard, but instead only by default to the JSON specific clipboard. That will also likely mean that we will need a context menu to copy JSON to the plain clipboard.

6. We will also need to preserve Paste backwards compatibility and deal with several different types of Paste contents. So you probably need to analyse what tracks are selected to be displayed in the metadata area, what metadata rows are selected and whether there is an edit box open and combine that with analysis of the possibly several types of paste data available on  the clipboard before determining which data and what type of paste to do - and you probably want to warn the user if they are about to paste album data across several albums or track data across several tracks.

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
...
-                self.tagger.clipboard().setText(MULTI_VALUED_JOINER.join(value))
-                self.clipboard = value
+            return value
+
+        #TODO: Support  copying data from multiple files
+
+        items = self.selectedItems()
+        if len(items) > 1:
+            # We have multiple tags selected, so copy them as a JSON string
+            data = {}
```
phw, line 295: I'd move generating the JSON data structure into a separate helper function.

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
...
+
+                fieldnames = {self.COLUMN_TAG: 'tag_name', self.COLUMN_ORIG: 'original_value', self.COLUMN_NEW: 'new_value'}
+                fieldname = fieldnames[item.column()]
+
+                value = get_value_as_string(item)
+                if value and fieldname:
+                    data[key] = { fieldname: value } | data[key]
+            mimedata = QtCore.QMimeData()
+            mimedata.setData("application/json;contents=picard-tags", json.dumps(data).encode('utf-8'))
+            mimedata.setText(json.dumps(data))
```
phw, line 309: Maybe we should choose a different representation for plaintext. Something that is closer to just text. E.g.:

```
tagname1: value1
tagname2: value1; value2; value3
```

This at least can be easily reused by the user in a text editor. If another application would want to support this it would need to understand the structure anyway, at which point it could also use the `application/json;contents=picard-tags` data.

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
...
+            data = json.loads(text)
+            for tag, values in data.items():
+                if self._tag_is_editable(tag):
+                    if 'new_value' in values:
+                        self._set_tag_values(tag, values['new_value'])
+                    else:
+                        self._set_tag_values(tag, values['original_value'])
+            return
+        else:
+            # Cliboard contents doesn't contain our JSON, so trat it as text.
```
phw, line 335: Small typo: "treat"

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
...
+                    data[key] = {}
+
+                fieldnames = {self.COLUMN_TAG: 'tag_name', self.COLUMN_ORIG: 'original_value', self.COLUMN_NEW: 'new_value'}
+                fieldname = fieldnames[item.column()]
+
+                value = get_value_as_string(item)
+                if value and fieldname:
+                    data[key] = { fieldname: value } | data[key]
+            mimedata = QtCore.QMimeData()
+            mimedata.setData("application/json;contents=picard-tags", json.dumps(data).encode('utf-8'))
```
phw, line 308: Let's put `"application/json;contents=picard-tags"` into a constant and use it here and below in the pasting code.

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
                 except (TypeError, ValueError) as why:
                     log.warning(why)
                     value = ['']
-
-            if value is not None:
-                self.tagger.clipboard().setText(MULTI_VALUED_JOINER.join(value))
-                self.clipboard = value
+            return value
+
+        #TODO: Support  copying data from multiple files
```
phw, line 290: I'm not even sure this should be implemented. What would be the semantics when pasting the data? Right now it would skip values that are different, and I think it is fine. The only other way would be to combine all values into a list. But not sure if this is even a usecase. If, then maybe as an explicit action on those tags to copy all values from all files.

picard/ui/metadatabox/__init__.py
```diff
@@ -282,19 +285,61 @@ def _copy_value(self):
...
+        items = self.selectedItems()
+        if len(items) > 1:
+            # We have multiple tags selected, so copy them as a JSON string
+            data = {}
+            for item in items:
+                key = self.tag_diff.tag_names[item.row()]
+                if key not in data:
+                    data[key] = {}
+
+                fieldnames = {self.COLUMN_TAG: 'tag_name', self.COLUMN_ORIG: 'original_value', self.COLUMN_NEW: 'new_value'}
```
phw, line 301: Do we actually need the `original_value` and `new_value` fields? Wouldn't a simpler structure like this work?

```json
{
  "genre": ["value1", "value2"],
  "artist": ["value1"]
}
```

As I see it the pasting code prefers new anyway, and copied valued depend on what was selected.

Btw, what I really like about the implementation is how it is possible to CTRL select some old and some new values and have them copied properly to another file.

phw (changes requested):
Thanks for this PR. First off this works really well, and it is a very welcomed feature. I noticed I had even already created a ticket for this a long time ago. I updated the PR to link to PICARD-2287 instead and closed the newer one as a duplicate.

I have a few coding comments. But my main concern about right now is use experience when pasting. While this is a very powerful feature, it also can lead to some unwanted surprise if a paste suddenly overwrites a large number of tags, where the user maybe expected just to paste a value. This is especially bad as Picard does not yet support any undo functionality.

I see two possible solutions:

1. Make copying the full tags an explicit option. At least then the user is aware they copied multiple values. But copying is less convenient.
2. When pasting full tags, show a warning dialog that multiple tags would be overwritten. We could have an option to not show this dialog, which would allow advanced users who use this often to skip it. Ideally the dialog would also show which tags will get overwritten or even show a preview of the changes.

Even though more complex, I think the second option would work best.

picard/ui/metadatabox/__init__.py
phw: Actually better suggesion: Use tab separated values. This would be compatible with spreadsheets and pate nicely there. So the above would be:

```
tagname1	value1
tagname2	value1	value2	value3
```

And when pasted to Google Sheets it is:

grafik

This should work cross-platform for all common spreadsheet programs.

phw:
> I am not sure that I agree with @outsidecontext's suggestion to use QT's special clipboard functionality - users will want to copy and paste between applications, so you want to have compatibility here 

@Sophist-UK You can store multiple representations of the data in the clipboard. So the JSON would be for internal use. The code can also fill in a plain text representation that by default would be pasted in other applications. Right now in the implementation this also contains the JSON structure, but I think it shouldn't. See my comment above.

> perhaps a spreadsheet-specific format

That's actually rather easy, as the spreadsheet tools all support tab separated values. And this also makes for a nice default text representation of multiple tags, as pasting this to a text editor is also nice. 

Supporting a proper plain text representation is definitely something that needs to be considered here. Other formats can of course be added, but this is definitely outside the scope of this PR. Let's focus on what this PR tries to enable: Copy and pasting multiple tags between files inside Picard.

Sophist-UK:
@phw Philipp

Steve and I have been discussing this privately on Discord, the gist of which was very similar to your comment - except we were thinking CSV rather than TSV.

And we were thinking monospaced tabular as a default text representation rather than TSV - but your idea might work better.

picard/ui/metadatabox/__init__.py
phw: Another thought: If we consider your original idea of having both original and new values and also look at the display in Picard, maybe having both original and new as separate column, and in each column the multiple values separated by `MULTI_VALUED_JOINER`, makes even more sense. Something like:

```
tagname1	old1; old2	new1; new2
tagname2	old1	new1; new2; new3
```

When copied over to a spreadsheet this would lead to a display very similar to what is inside Picard.

At this point I don't have a strong preference for this or my previous suggestion. @StevilKnevil @zas @Sophist-UK What are your thoughts?

phw:
I looked a bit what other tagging tools are doing. A lot don't really support a table view like Picard and just have some text fields.

MP3Tag generally shows all files with their tags in a spreadsheet like view, but doesn't allow selecting multiple specific values. It kind of supports  copying the tags from one file to another internally, but to the system clipboard it only writes a plain/text list of compact info about the files:

```
1. Nine Inch Nails - [Ghosts I-IV CD1/1 #02] 2 Ghosts I
2. Paradise Lost - [Symbol of Life CD1 #03] Two Worlds
```

Kid3 does have a spreadsheet like view for the metadata, but does not allow selecting multiple rows or fields. You can only edit a single field and copy data from there.

Puddletag is interesting, as it writes a JSON structure to the normal text clipboard. See the selection in below screenshot.

grafik

This gives the following structure:

```json
[
    {
        "title": [
            "Hells Bells"
        ],
        "album": [
            "Back in Black"
        ],
        "track": [
            "1/10"
        ],
        "__length": "05:12"
    },
    {
        "title": [
            "Shoot to Thrill"
        ],
        "album": [
            "Back in Black"
        ],
        "track": [
            "2/10"
        ],
        "__length": "05:17"
    }
]
```

Very similar to ours. But I still think the approach from this PR with custom mime data and as discussed a separate text representation with spreadsheet compatibility is the better solution. App specific JSON data in plaintext clipboard is not that useful.

StevilKnevil:
This is all really good input - thanks everyone, it's what I'd hoped to get from raising this PR. I'm going to iterate again taking the above into account. I reckon I'm about 20% done so far :)

My focus is to get copy and paste working inside Picard, and for Copy to other apps. There's definitely a follow up PR for paste from other apps.

picard/ui/metadatabox/__init__.py
phw: > As usual, we need to ensure strings don't contain separators or we need to escape them somehow.

That's a problem we already have at several places, including the existing copy and paste (pasting just splits at the joiner). It's probably outside the scope of this PR to fully address this.

picard/ui/metadatabox/__init__.py
StevilKnevil: Agreed

picard/ui/metadatabox/__init__.py
StevilKnevil: I've now implemented plain text as TSV. This works well in both Excel and Notepad++.

MOST importantly: I added some tests for this, so that I can iterate on it and keep my sanity :) Very much lowest bar for the testing at the moment, but something to build on. Would value any input on how "pythonic" my implementation is :)

picard/ui/metadatabox/__init__.py
StevilKnevil: In terms of copying, I don't think we should add logic to what we "think" the user intends, lets just put everything in the clipboard in case they want to see before and after. If they do only want the new values then they can just select and copy those.

zas:
@StevilKnevil please fix style issues, it would ease reviews

rdswift:
> I think the docs would need a separate section for manually editing tags in the metadata view. Could be placed under Using Picard > General Usage. The section would explain the view, explain how values can be edited (both inline and in the tag editor dialog), and explain the handling of multi-value tags. This doc then also could provide documentation about the copy and paste functionality.

Perfect!  Thanks for creating the ticket.

zas (commented):
Just few stylistic comments at this point, please respect code style on each commit.
Tests should pass after each of your commits.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,36 +266,107 @@ def keyPressEvent(self, event):
...
+        if tag == '~length':
+            try:
+                value = [format_time(value or 0), ]
+            except (TypeError, ValueError) as why:
+                log.warning(why)
+                value = ['']
+        return value
+
+    def get_selected_tag_data(self, items):
+        fieldnames = {self.COLUMN_TAG: 'tag_name', self.COLUMN_ORIG: 'original_value', self.COLUMN_NEW: 'new_value'}
```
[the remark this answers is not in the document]
StevilKnevil: Yeah - I'm definitely going to move that, but it may end up in tag_diff, just mulling that at the moment

picard/ui/metadatabox/__init__.py
```diff
@@ -265,36 +266,107 @@ def keyPressEvent(self, event):
...
+
+        result = {}
+        for item in items:
+            key = self.tag_diff.tag_names[item.row()]
+            if key not in result:
+                result[key] = {}
```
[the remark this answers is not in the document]
StevilKnevil: Ooo - nice!

picard/ui/metadatabox/__init__.py
```diff
@@ -265,36 +266,107 @@ def keyPressEvent(self, event):
...
+            if 'original_value' in values:
+                result += MULTI_VALUED_JOINER.join(values['original_value']) + "\t"
+            else:
+                result += "\t\t"
+            if 'new_value' in values:
+                result += MULTI_VALUED_JOINER.join(values['new_value'])
+            else:
+                result += "\t"
+            result = result.rstrip() + "\n"
+        return result.rstrip()
```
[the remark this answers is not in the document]
StevilKnevil: Done

picard/ui/metadatabox/__init__.py
```diff
@@ -72,6 +71,7 @@
     TagDiff,
     TagStatus,
 )
+import dill
```
zas, line 74: This is an extra dependency, what is it needed for?

StevilKnevil: Rather than writing a custom tag diff to json serialisation, I figured that it may be easier to just use something off the shelf.

Not sure it is worth the extra dependency though, unless it's useful anywhere else!

phw: Let's go without this dependency then. We are trying to avoid external dependencies if possible, or keep it optional. I think the use case here does not really warrant that. You need to build the data structure anyway, JSON seemed fine for serializing this.

Having this in JSON format means some other application could potentially implement support for this mimetype in clipboard easily (without needing to understand the internals of Python dill or pickle).

zas:
I noticed an issue that don't happen on master branch:

I select a grouped tag, open context menu in metadata box, click Copy:

Capture d’écran du 2025-03-23 10-48-22



```
  File "/home/zas/src/picard/picard/ui/metadatabox/__init__.py", line 327, in _copy_value
    value = self.get_selected_tags(item)
            ^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/zas/src/picard/picard/ui/metadatabox/__init__.py", line 293, in get_selected_tags
    for item in items:
TypeError: 'QTableWidgetItem' object is not iterable
```

StevilKnevil:
Going to convert this back to draft for a bit. I need to do a bunch of testing around non-happy paths and also need to  think about escaping copied values!

zas (changes requested):

picard/ui/metadatabox/tagdiff.py
```diff
@@ -342,3 +343,29 @@ def update_tag_names(self, changes_first=False, top_tags=None):
...
+        COL_SEP = "\t"
+        LINE_SEP = "\n"
+        lines = []
+        for tag in self.tag_names:
+            columns = [
+                tag,
+                MULTI_VALUED_JOINER.join(self.old[tag]),
+                MULTI_VALUED_JOINER.join(self.new[tag])
+            ]
+            line = COL_SEP.join(columns).rstrip()
```
zas, line 368: Why `rstrip()` ? I don't think we need it.

zas (changes requested):

picard/ui/metadatabox/tagdiff.py
StevilKnevil: I've changed to using `csv.writer()` to handle all the secpaing. This leaves a trailing `\r\n` at the end of the string, which I don't like so I've kept one `rstrip()` right at the end

picard/ui/metadatabox/tagdiff.py
```diff
@@ -342,3 +345,28 @@ def update_tag_names(self, changes_first=False, top_tags=None):
...
+            if tag in self.new:
+                result[tag]['new'] = self.new[tag]
+
+        return json.dumps(result)
+
+    def to_tsv(self):
+        COL_SEP = "\t"
+
+        f = io.StringIO()
+        writer = csv.writer(f, delimiter=COL_SEP)
```
zas, line 364: Then you should use a dialect to make things clear, see `excel-tab` dialect (https://docs.python.org/3/library/csv.html#csv.excel_tab).


```python
import io
import csv

csvdata = ["1", "2\t3", "4\n5", '6"7', '']

dialect = 'excel-tab'
target = io.StringIO()
writer = csv.writer(target, dialect=dialect)
writer.writerow(csvdata)
csvresult = target.getvalue()
print(csvresult)

source = io.StringIO(csvresult)
reader = csv.reader(source, dialect=dialect)
for row in reader:
    print(row)
    print(row == csvdata)

```

@phw: do we need to handle encoding somehow?

picard/ui/metadatabox/tagdiff.py
```diff
@@ -342,3 +345,28 @@ def update_tag_names(self, changes_first=False, top_tags=None):
...
+
+        for tag in self.tag_names:
+            writer.writerow([
+                tag,
+                MULTI_VALUED_JOINER.join(self.old[tag]),
+                MULTI_VALUED_JOINER.join(self.new[tag])])
+
+        return f.getvalue().rstrip()
```
zas, line 372: rstrip() can be a problem if last field contains whitespaces, so better use [`removesuffix()`](https://docs.python.org/3/library/stdtypes.html#str.removesuffix)

test/test_tagdiff.py
```diff
@@ -272,3 +272,71 @@ def test_special_handler_unchanged(self):
...
+        self.tag_diff.update_tag_names()
+        tags = self.tag_diff.to_tsv()
+        self.assertEqual(tags, 'artist\tArtist 1\tArtist 2\r\n')
+
+    def test_multiple_tags_to_tsv(self):
+        self.tag_diff.add("artist", ["Artist 1"], ["Artist 2"])
+        self.tag_diff.add("album", ["Album 1"], ["Album 2"])
+        self.tag_diff.update_tag_names()
+        tags = self.tag_diff.to_tsv()
+        self.assertEqual(tags, 'album\tAlbum 1\tAlbum 2\r\nartist\tArtist 1\tArtist 2\r\n')
```
zas, line 330: I think we should also have tests for artist names containing weird stuff: tab chars, double quotes, newlines, cr + newlines,  or having only whitespaces. Even though those are unlikely in real life, there's no reason for the conversion to TSV do not handle those.
Also it would be interesting to test the conversion in both directions (to and from tsv).

picard/ui/metadatabox/__init__.py
```diff
@@ -316,76 +292,95 @@ def get_value_as_string(self, item):
...
-                    tag=tag,
-                    new=value,
-                    removable=self.tag_diff.status[tag] != TagStatus.NOTREMOVABLE,
-                    readonly=self.tag_diff.status[tag] == TagStatus.READONLY
-                )
+            tag, value = self.get_row_info(item.row())
+            result.add(
+                tag=tag,
+                old=value[self.COLUMN_OLD] if item.column() == self.COLUMN_OLD else None,
+                new=value[self.COLUMN_NEW] if item.column() == self.COLUMN_NEW else None,
```
zas, line 302: Just a minor thing, but caching the value of `item.column()` wouldn't hurt and make lines a bit shorter:

```python
            column = item.column()
            result.add(
                tag=tag,
                old=value[self.COLUMN_OLD] if column == self.COLUMN_OLD else None,
                new=value[self.COLUMN_NEW] if column == self.COLUMN_NEW else None,
```

zas (changes requested):

zas:
@StevilKnevil it would be useful to have more debug logging when copying/pasting, at the moment nothing appears in the log if copy and/or paste fails.
The reason a copy or a paste doesn't work could be displayed when the action is supposed to take place. Something like "Cannot copy....", "Copying ... to ...", "Cannot paste to ..", etc... would help a lot.

picard/ui/metadatabox/tagdiff.py
```diff
@@ -342,3 +349,41 @@ def update_tag_names(self, changes_first=False, top_tags=None):
...
+
+    def to_tsv(self, prettify_times=True):
+        f = io.StringIO()
+        writer = csv.writer(f, dialect='excel-tab')
+
+        for tag in self.tag_names:
+            # Take care with the special case of '~length', as that is not wrapped in a list it's a plain string
+            old = self.old[tag]
+            new = self.new[tag]
+            if (tag == "~length"):
```
zas, line 380: Parenthesis can be removed here.

picard/ui/metadatabox/tagdiff.py
```diff
@@ -342,3 +349,41 @@ def update_tag_names(self, changes_first=False, top_tags=None):
...
+            return ['']
+
+    def to_tsv(self, prettify_times=True):
+        f = io.StringIO()
+        writer = csv.writer(f, dialect='excel-tab')
+
+        for tag in self.tag_names:
+            # Take care with the special case of '~length', as that is not wrapped in a list it's a plain string
+            old = self.old[tag]
+            new = self.new[tag]
```
zas, line 379: Move those statements to an `else` branch below

zas (changes requested):

picard/ui/metadatabox/tagdiff.py
StevilKnevil: I'm not sure about that, I find this more readable. This highlights the special case (rather than 'alternative' case) for lengths. To put it another way, if we have a future PR that returns `~length` as an array with a single element (to match other tags) then the three lines  for `~length` would just be removed

zas:
I noticed an issue:

image

I get `D: 21:46:16,774 ui/metadatabox._copy_value:300: Unable to copy when multiple files are selected, clearing the clipboard` though I have only one tag selected, one file selected in the right pane

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +270,104 @@ def keyPressEvent(self, event):
...
-        item = self.currentItem()
-        if item:
-            column = item.column()
-            tag = self.tag_diff.tag_names[item.row()]
-            value = None
-            if column == self.COLUMN_ORIG:
-                value = self.tag_diff.old[tag]
-            elif column == self.COLUMN_NEW:
-                value = self.tag_diff.new[tag]
+        if len(self.objects) > 1:
```
StevilKnevil, line 298: Is this the right check to see if there are multiple files selected? @zas is reporting that this check isn't working correctly they only have 1 file selected, I'm unable to reproduce.

zas: Clearly not, because it contains 2 objects (a Track and a File) since it was tagged already:

```
D: 10:26:22,572 ui/metadatabox._copy_value:298: {<FLACFile '01 Ummon.flac'>, <Track 38851005-c0f5-4a59-a623-3589ae86db11 'Ummon'>}
D: 10:26:22,572 ui/metadatabox._copy_value:301: Unable to copy when multiple files are selected, clearing the clipboard
```

StevilKnevil: Oh, OK, I'll go and do some reverse engineering to see what the best approach is

picard/ui/metadatabox/__init__.py
```diff
@@ -295,9 +295,9 @@ def get_row_info(self, row):
         return tag, value
 
     def _copy_value(self):
-        if len(self.objects) > 1:
+        if len(self.tracks) > 1 or len(self.files) > 1:
             # We cannot copy from multiple files at the same time, clear the clipboard to avoid confusion
-            log.debug("Unable to copy when multiple files are selected, clearing the clipboard")
+            log.debug("Unable to copy when multiple tracks or files are selected, clearing the clipboard")
```
StevilKnevil, line 300: I'm not sure if this is the best logic to detect multiple selection, as I only have a vague understanding of how `tracks` and `files` relate to each other. I think that if you have a single file and single track then they are guaranteed to be related to each other, so just single selection?

zas: So, after last patch, I can copy tags from one track IF one file is linked to it, but not if more are linked to it, even if the tag is the same for all (I expected to not be able to copy tags that are grouped, but I expect to be able to copy those that aren't).

StevilKnevil: I'm inclined to leave copying from multiple items (even if they have the same value) for a future PR.

picard/ui/metadatabox/__init__.py
```diff
@@ -137,8 +137,15 @@ def get_tag_name(self, index):
...
+    MIMETYPE_PICARD_TAGS = "application/vdr.picard"
+    MIMETYPE_TEXT = 'text/plain'
+    mimedatas = {
+        MIMETYPE_PICARD_TAGS: 'to_json',
+        MIMETYPE_TEXT: 'to_tsv',
+    }
+
     COLUMN_TAG = 0
-    COLUMN_ORIG = 1
+    COLUMN_OLD = 1
```
zas, line 148: This rename wasn't strictly needed.
My recent patch to tagdiff leads to this confusion: tagdiff uses old/new because it isn't really tied to anything. but this column is named after what it contains and its title (Original Tag).

StevilKnevil: rgr that - I was following your lead, but you raise a good point re column names!

zas (changes requested):
The patch now looks almost good to go to me, just few minor comments now.
I tested it, and it seems to work in all cases it is supposed to.

There will be some user-friendly improvements to make in future PRs but the overall patch is already a nice improvement.

picard/ui/metadatabox/__init__.py
```diff
@@ -395,17 +463,16 @@ def contextMenuEvent(self, event):
...
-                    paste_action.setShortcut(QtGui.QKeySequence.StandardKey.Paste)
-                    paste_action.setEnabled(editable)
-                    menu.addAction(paste_action)
+
+                menu.addSeparator()
+                copy_action = QtGui.QAction(icontheme.lookup('edit-copy', icontheme.ICON_SIZE_MENU), _("&Copy"), self)
+                copy_action.triggered.connect(self._copy_value)
+                copy_action.setShortcut(QtGui.QKeySequence.StandardKey.Copy)
+                menu.addAction(copy_action)
+                paste_action = QtGui.QAction(icontheme.lookup('edit-paste', icontheme.ICON_SIZE_MENU), _("&Paste"), self)
```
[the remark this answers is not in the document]
Sophist-UK: Actually, can I suggest that this needs to have a little more complex logic than this:

1. There is no point in enabling the paste button if the clipboard isn't empty but what is on the clipboard cannot be pasted.

2. If you disable the paste button, then you will need to watch for changes on the clipboard so that you can enable it when something appears there that is now pasteable.

picard/ui/metadatabox/__init__.py
```diff
@@ -137,6 +137,13 @@ def get_tag_name(self, index):
 
 class MetadataBox(QtWidgets.QTableWidget):
 
+    MIMETYPE_PICARD_TAGS = "application/vdr.picard"
```
Sophist-UK, line 140: The problem I have with copying only to these two types is that `vdr.picard` is by definition only going to be understandable by picard, and there is therefore no structured form of data that can be used for pasting by e.g. a spreadsheet or a database.

I would personally like to see a Copy being sent to the clipboard as mimetypes for JSON, for TSV and for CSV (perhaps with and without headers if there are different standard mimetypes for these), so that receiving applications have a wider choice of structured data to choose from.

Sophist-UK (commented):
Firstly, apologies for the delay in reviewing, but we have had a week long internet outage following a massive storm that sent several bolts of lightening through our ISP's communications tower and brought down loads of telegraph poles carrying their fibre trunks.

Secondly I am presuming that @StevilKnevil has done some empirical testing of how e.g. spreadsheet apps will handle various standard mime types to see if they are worth handling on both copy and paste. My guess is that if they cannot handle TSV, CSV that a generic JSON mimetype would still be useful to have.

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: Yeah, I think maybe a future PR though, because it should be refactored so that keyboard shortcuts use the same logic (currently we hide the context menu option but the keyboard shortcuts still functions)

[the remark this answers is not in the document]
Sophist-UK: My understanding of modern development is that each and every PR should leave the code base in a state that you will be happy to publish a new release. 

But Picard doesn't do releases like that and it is entirely up to Laurent and Philipp et al. to decide whether to accept a PR that isn't perfect enough for a release and then follow-up with another one.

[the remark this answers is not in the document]
zas: There's no doubt the implemented feature is useful, and that current implementation can be improved.
My experience shows that asking too much from a new contributor in one PR is rather discouraging, and lead to unfinished code and abandoned PRs.
This PR isn't approved yet anyway.

picard/ui/metadatabox/__init__.py
zas: > The problem I have with copying only to these two types is that `vdr.picard` is by definition only going to be understandable by picard, and there is therefore no structured form of data that can be used for pasting by e.g. a spreadsheet or a database.

Can you give exact use cases, application names and mime-types?
We can easily add more mime-types if needed, now or in a later PR.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +270,104 @@ def keyPressEvent(self, event):
...
+            item = self.currentItem()
+            if item:
+                column_is_editable = (item.column() == self.COLUMN_NEW)
+                tag = self.tag_diff.tag_names[item.row()]
+                if column_is_editable and self._tag_is_editable(tag):
+                    value = self.tagger.clipboard().text()
+                    if value:
+                        log.info("Pasting %s from text clipboard to tag %s", value, tag)
+                        self._set_tag_values(tag, value.split(MULTI_VALUED_JOINER))
+                        self.update()
```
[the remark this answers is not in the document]
zas: While I partly agree, previous code didn't do much more "analysis".
At this point we need to ensure copying anything doesn't lead to a crash or anything bad.
And that copying and pasting multiple tags works (that's the purpose of this PR), does it for you, on which platform did you test the code?

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: For awareness, the behaviour of copying in the current release is:
* If multiple files are selected, the copy menu option is displayed, and the keyboard shortcut still works
* If multiple tags are selected, the copy menu option isn't displayed, but the keyboard shortcut still works

* If multiple files are selected, and the user copies a tag value to the clipboard through the menu or keyboard shortcut, then Picard crashes
* If multiple tags are selected and the user copies via the keyboard shortcut then one of the tags will be copied to the clipboard (not sure what the logic is for which one of the tags is copied however

I would suggest that the issue is in the logic of the copy, not whether a menu item is displayed or not, given that a keyboard shortcut will circumvent any logic we put in the menu.

We could instruct the OS to play the "invalid operation" noise when trying to copy something invalid - a message box feels too invasive. Not sure how to get QT to ask the OS to emit a sound though.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +270,104 @@ def keyPressEvent(self, event):
...
+                if self._tag_is_editable(tag):
+                    # Prefer 'new' values, but fall back to 'old' if not available
+                    value = data[tag].get(TagDiff.NEW_VALUE) or data[tag].get(TagDiff.OLD_VALUE)
+                    if value:
+                        if isinstance(value, list):
+                            # There are multiple values for the tag
+                            value = MULTI_VALUED_JOINER.join(value)
+                        # each value may also represent multiple values
+                        log.info("Pasting '%s' from JSON clipboard to tag '%s'", value, tag)
+                        value = value.split(MULTI_VALUED_JOINER)
```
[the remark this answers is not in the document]
StevilKnevil: Yep - that makes sense!

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +270,104 @@ def keyPressEvent(self, event):
...
+        else:
+            # Just copy the current item as a string
+            item = self.currentItem()
+            if item:
+                tag, value = self.get_row_info(item.row())
+                value = value[item.column()]
+                if tag == '~length':
+                    value = self.tag_diff.handle_length(value, prettify_times=True)
+                if value is not None:
+                    log.debug("Copying '%s' to clipboard", value)
```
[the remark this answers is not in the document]
StevilKnevil: We can, though I think that this may be misleading. This code path will only copy a single string from a single cell, so the tag it relates to is less relevant, it's just a plain text copy.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +270,104 @@ def keyPressEvent(self, event):
...
-        if item:
-            column = item.column()
-            tag = self.tag_diff.tag_names[item.row()]
-            value = None
-            if column == self.COLUMN_ORIG:
-                value = self.tag_diff.old[tag]
-            elif column == self.COLUMN_NEW:
-                value = self.tag_diff.new[tag]
+        if len(self.tracks) > 1 or len(self.files) > 1:
+            # We cannot copy from multiple files at the same time, clear the clipboard to avoid confusion
```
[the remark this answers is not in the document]
StevilKnevil: See above about the issues with the current implementation of copy that are improved, though not perfected, by this PR

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: To be clear, the current release does:
```
        item = self.currentItem()
        if item:
            column = item.column()
            tag = self.tag_diff.tag_names[item.row()]
            if column == self.COLUMN_NEW and self._tag_is_editable(tag):
                self._set_tag_values(tag, self.clipboard)
                self.update()
```
So the change I've made to the existing codepath is to add some logging and to add a defensive if statement in case the clipboard contains something other than text.

I could just revert back to the existing code if that's preferable?

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
zas: I don't get your point here, isn't the value copied from the tag row? That's debug output, I think it is useful to clarify from where the value is copied.

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
zas: But the code I suggest should be equivalent or do I miss something?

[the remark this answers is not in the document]
zas: Ah, got it. Ok.

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: Valid

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
zas: > We could instruct the OS to play the "invalid operation" noise when trying to copy something invalid - a message box feels too invasive. Not sure how to get QT to ask the OS to emit a sound though.

I don't think emitting a sound is a good approach (we can't assume there is any sound device enabled, and system beep isn't really made for this, even if it works).

A warning box wouldn't that much a problem, since that will be triggered by user action anyway, at least we would be able to explain what is going on. If we add a 'Do not show again' checkbox to this message box the user can choose to silence the warning.

Additionally, we could display something in the status bar (bottom of Picard main window, as we do for certain operations), and this is logged by default.

For sure, it would be better than having a silent failure.

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
zas: > I could just revert back to the existing code if that's preferable?

For my part, I'm ok with your changes, especially dropping `self.clipboard` (clipboard content are stored anyway as you said).

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: Oh yes, there are drawbacks of relying on sounds - though if it's an OS operation it's better (as accessibility options will be covered by the OS). Any warning box that has a "do not show this again" rings alarm bells to me. They're not inherently bad, but they can be often used as a "get out of jail free card"...

Do we have toasts already in Picard? That might be my favourite option. Status bar message is also good with me (if you can point me to something to copy. Or: how about just populating the clipboard with the message "Picard is unable to copy content when multiple tags or files are selected" 😜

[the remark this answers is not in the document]
zas: We don't have toasts yet. And the main reason is there's no native PyQt support for them, though there's a Python module that implements those (https://pypi.org/project/pyqt-toast-notification/). Not sure we want to add another dependency right now.

For now, we use normal messages, and I would stick to that until we move to something better.

Populating the clipboard is a bad idea. Actually I saw you clear it in https://github.com/metabrainz/picard/pull/2613/files#diff-8dae8dfb0d9d5159ab315c5d2c081f3abb6e77fba08d861aee6f70b5a8aa1592R303 and I also think that's a bad idea (we can just return, and let the clipboard as is).

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
phw: I agree, as part of this PR it is sufficient if paste of plain text keeps working as before. Any analysis of plain text will become quickly very complex and error prone. It is easy to copy and paste a TSV to a spreadsheet, as the spreadsheet does not further care about the content and just distributes it over a table. The backward copy and paste some random TSV to Picard is not clear at all. The data can have any structure.

Let's keep this out of this PR. Let's paste plain text content from the clipboard just as text as before.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +272,107 @@ def keyPressEvent(self, event):
...
+            log.debug("Unable to copy when multiple tracks or files are selected, clearing the clipboard")
```
phw, line 302: Why this restriction? In the current code copying is possible.

I think we shouldn't restrict this here. When copying multiple selected tags it should instead ignore the tags that differ between the files, but copy all tags that are the same.

When copying a single value we could either retain the current behavior (which copies an empty string in case of differing values) or not copy anything.

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
phw: You can set a message on the statusbar with `tagger.window.set_statusbar_message`.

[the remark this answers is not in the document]
rdswift: I use the following in one of my personal projects.  Not exactly a "toast", but it also doesn't introduce any additional dependencies.

```python
class AutoCloseMessageBox(QMessageBox):
    """Automatically closing message box.
    """
        super().__init__(parent=parent)

    @staticmethod
    def showWithTimeout(title: str = None, text: str = None, show_button: bool = False, timeout: float = 2.5,
                        icon: QMessageBox.icon = QMessageBox.NoIcon, parent=None):
        """Show dialog box with timeout.

        Args:
            title (str, optional): Title of the displayed dialog. Defaults to None.
            text (str, optional): Text displayed in the dialog. Defaults to None.
            show_button (bool, optional): Show an Okay button. Defaults to False.
            timeout (float, optional): Seconds to wait before auto closing the dialog. Defaults to 2.5 (maximum 10).
            icon (QMessageBox.icon, optional): Icon to display. Defaults to QMessageBox.NoIcon.
            parent (QWidget, optional): Parent widget. Defaults to None.
        """
        # pylint: disable=too-many-arguments
        w = QMessageBox(parent=parent)
        timeout = min(int(timeout * 1000), 10000)
        w.setEscapeButton(QMessageBox.Close)
        if title is not None and title.strip():
            w.setWindowTitle(title.strip())
        else:
            w.setWindowTitle(' ')
        if text is None:
            text = ''
        w.setText(text.strip() or 'No message provided')
        w.setIcon(icon)
        w.setStandardButtons(QMessageBox.Close)
        w.button(QMessageBox.Close).animateClick(timeout)
        if not show_button:
            w.button(QMessageBox.Close).setVisible(False)
        w.exec()
```

picard/ui/metadatabox/__init__.py
StevilKnevil: Whilst we don't stop people copying with multiple files selected currently, it does crash Picard (at least in my experience!) So rather than trying to amalgamate a bugfix for that into this PR I figured adding this failsafe for now and fix the underlying issue in a different PR.

phw: It's working fine for me. The only crash I'm aware of is the one you already fixed in https://github.com/metabrainz/picard/pull/2609 .

So we should be good to allow this. At a minimum as part of this PR I would expect that the copying of a single value keeps working as before. Having multiple tags work across multiple files we could do in a separate PR, if it complicates things too much.

If there are still some crash conditions in existing code, I agree we can also fix them separate from this PR.

StevilKnevil: > It's working fine for me. The only crash I'm aware of is the one you already fixed in https://github.com/metabrainz/picard/pull/2609

Wow that is embarrassing - I'm rediscovering bugs that I've already fixed! 🫣

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: OK - Have added status bar message for copying and pasting to/from invalid targets

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +272,107 @@ def keyPressEvent(self, event):
...
+        # Prefer to paste the Picard JSON data if available, otherwise text
+        mimedata = self.tagger.clipboard().mimeData()
+        if mimedata.hasFormat(self.MIMETYPE_PICARD_TAGS):
+            text = mimedata.data(self.MIMETYPE_PICARD_TAGS).data()
+            try:
+                data = json.loads(text)
+            except json.JSONDecodeError as e:
+                log.error("Failed to decode JSON data from clipboard: %r", e)
+                return
+
```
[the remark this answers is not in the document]
StevilKnevil: I like it! But not for for this PR 😉 https://tickets.metabrainz.org/browse/PICARD-3052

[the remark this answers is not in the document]
zas: Of course, not in this PR, though what I was suggesting in my comments is to prepare for such move by moving code in `if/elif` branches to separate functions for them to be reused. We could have `_parse_value_single(text)` and  `_parse_value_multiple(data)` or something like that.

zas:
@StevilKnevil @phw I noticed we are updating File objects more than once when pasting multiple values. In order to avoid this I wrote https://github.com/StevilKnevil/picard/pull/1

Basically I split the code that update objects so we can delay the object.update() calls as well as ensuring we call that only once per object.
So the code that paste multiple values is modified so we get a set of objects to update at the end.

Please have a look and tell me what you think.

StevilKnevil:
> @StevilKnevil @phw I noticed we are updating File objects more than once when pasting multiple values. In order to avoid this I wrote [StevilKnevil#1](https://github.com/StevilKnevil/picard/pull/1)
> Please have a look and tell me what you think.

I see (and accept) your suggestion, and have a counter proposal: https://github.com/StevilKnevil/picard/pull/2. Essentially extend the mime data handling to have both a copy to mimedata from TagDiff _and_ a populate metadatabox from mimedata function. That way the paste function just goes through the list (in order) and the first entry in that list that can handle one of the mimedatas on the clipboard is used to paste from the clipboard

picard/ui/metadatabox/__init__.py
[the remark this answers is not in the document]
StevilKnevil: Done

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +295,134 @@ def keyPressEvent(self, event):
...
+        items = self.selectedItems()
+        if len(items) > 1:
+            selected_data = self.get_selected_tags(items)
+            # Build the mimedata to use for the clipboard
+            mimedata = QtCore.QMimeData()
+            converted_data_cache = {}
+            for mimetype, converters in self.mimedatas.items():
+                if converters.encode_func:
+                    try:
+                        encoded_data = converted_data_cache.setdefault(mimetype, converters.encode_func(selected_data))
```
zas, line 340: Using `setdefault()` forces evaluation of `converters.encode_func(selected_data)`, do we really need a "cache" mechanism here?

StevilKnevil: Data is converted to TSV for bother the TSV mimetype and the text mimetype. This caches is supposed to avoid that happening twice if the same encoding function is used for multiple mime-types.

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +295,134 @@ def keyPressEvent(self, event):
...
+            log.info("Pasting %s from text clipboard to tag %s", value, tag)
+            value = value.split(MULTI_VALUED_JOINER)
+            yield from self._set_tag_values_delayed_updates(tag, value)
+
+    def _can_paste(self):
+        has_valid_mime_data = False
+        mimedata = self.tagger.clipboard().mimeData()
+        for mimetype, converters in self.mimedatas.items():
+            if mimedata.hasFormat(mimetype) and converters.decode_func:
+                has_valid_mime_data = True
```
zas, line 402: `break` here, no need to continue looping

picard/ui/metadatabox/__init__.py
```diff
@@ -265,41 +295,134 @@ def keyPressEvent(self, event):
...
 
-    def _update_clipboard(self):
-        clipboard = self.tagger.clipboard().text().split(MULTI_VALUED_JOINER)
-        if clipboard:
-            self.clipboard = clipboard
+        objects_to_update = set()
+        mimedata = self.tagger.clipboard().mimeData()
+
+        for mimetype, converters in self.mimedatas.items():
+            if mimedata.hasFormat(mimetype) and converters.decode_func:
```
zas, line 416: We have same code above, we can convert this to a generator:

```python

def decode_funcs(self, mimedata):
    for mimetype, converters in self.mimedatas.items():
        if mimedata.hasFormat(mimetype) and converters.decode_func:
             yield converters.decode_func
```

```python
for decode_func in decode_funcs(mimedata):
    objs = decode_func(self, mimedata)
```


And above:

```python
has_valid_mime_data = any(decode_funcs(mimedata))
```

zas (changes requested):

picard/ui/metadatabox/__init__.py
StevilKnevil: I'm thinking of breaking this out into a TagDiffMimeRegistry class in the TagDiff file. Can add tests around it, and avoids bloating the metadata box class any more!

StevilKnevil: Something like this? https://github.com/StevilKnevil/picard/pull/3

zas (approved):
LGTM

phw (changes requested):
@StevilKnevil This works now great and code looks good to me. I like how it also improves the general copy experience.

There is an accidentally committed bytecode file, but otherwise this is good to go.

test/__pycache__/test_mimedatahelper.cpython-311-pytest-8.3.5.pyc.36376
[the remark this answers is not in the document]
zas: @phw good catch, I didn't see it :(

And yes, we should, see https://github.com/metabrainz/picard/pull/2625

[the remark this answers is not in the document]
zas: @StevilKnevil please remove this bytecode file and rebase

zas (approved):
LGTM

phw (approved):
Great work, thanks a lot

PR #2616, opened 2025-03-18
CI: Move code style checks with flake8 and isort to separate jobs

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Move code style checks with flake8 and isort to separate jobs

This avoids the need to run the checks multiple times and allows tests to execute independently. In recent PRs the old behavior was annoying, as code style issues hid the fact whether tests would pass. Also running the style checks 21 times on each push seems wasteful.

@zas not sure if we maybe just could fully rely on ruff instead. Does the ruff check already cover flake8 fully? What is needed to replace isort?


zas:
> not sure if we maybe just could fully rely on ruff instead. Does the ruff check already cover flake8 fully? What is needed to replace isort?

I run ruff check and a pre-commit hook with usual flake8 & isort:
- I never get flake8 warnings that weren't already in ruff output
- I sometimes get isort warnings that ruff didn't output (so the config isn't exactly the same for now)

So, at the moment, I think we can totally get rid of flake8, but we still need isort.

zas (approved):
LGTM

PR #2617, opened 2025-03-18
Use PyInstaller code signing on macOS

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

# Problem

Current versions of PyInstaller support code signing on macOS. Use this functionality

# Solution

-  Replace custom code signing code with proper PyInstaller configuration
- Always perform the notarization if app got signed (we no longer support older versions of macOS that would not require notarization)
- Always test run the built executables. This is now possible as PyInstaller will ad-hoc sign the app bundle even if we didn't explicitly specify a `CODESIGN_IDENTITY` (which is still optional, as not all CI builds have the required secrets)
- Additionally remove unused upload code in the macos packaging script.

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -113,7 +113,10 @@ else:
               upx=False,
               icon='picard.ico',
               version='win-version-info.txt',
-              console=False)
+              console=False,
+              # macOS code signing
+              codesign_identity=os.environ.get('CODESIGN_IDENTITY', None),
+              entitlements_file='./scripts/package/entitlements.plist')
 
 
     coll = COLLECT(exe,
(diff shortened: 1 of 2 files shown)


phw:
A successful run with full code signing is in my repo at https://github.com/phw/picard/actions/runs/13918238439/job/38945260305

When run as part of a PR (no certificate secrets available) the app is ad-hoc signed and notarization is not performed. Still the bundled executables are tested (this wasn't done before). Example run: https://github.com/metabrainz/picard/actions/runs/13918312533/job/38945491001

PR #2622, opened 2025-03-21
Added utility function to update tag_names in TagDiff

StevilKnevil:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [X] Refactoring
  * [ ] Other

# Problem

`TagDiff` has a `tag_names` member to represent the order that the tags should be displayed. It has some quite involved logic about what order they should be sorted in, though it is deterministic based on the contents of `new` and `old` members of the `TagDiff`

# Solution

Moved the ordering logic into the `TagDiff` class, so `MetaDataBox.updateTags()` populates the contents of the `TagDiff` and lets the `TagDiff` populate its `tag_names` member.

Added tests for this functionality.

# Action

### DIFF
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -608,25 +608,7 @@ class MetadataBox(QtWidgets.QTableWidget):
 
                 tag_diff.objects += 1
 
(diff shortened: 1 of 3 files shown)


zas (approved):
LGTM

PR #2624, opened 2025-03-31
PICARD-3029: Incorrect mapping of case-variant tags in MP4 files

jpmsousa03:
# Summary  

* This is a…  
  * [X] Bug fix  
  * [ ] Feature addition  
  * [ ] Refactoring  
  * [ ] Minor/simple change (e.g., a typo)  
  * [ ] Other  

* **Description (1-2 sentences):**  
  A change has been applied to `mp4.py`, ensuring that `_casemap` is checked before applying pre-mappings. This prevents duplicate tags from being saved in the file.  

---  

# Problem  

Although `_casemap` correctly stores the capitalization of a tag when loading, an issue arises if a stored tag (e.g., `----:com.apple.iTunes:Label`) shares the same internal Picard tag as one of the pre-mapped cases (e.g., `label`). In such cases, the internal tag would be converted to the pre-mapped format (e.g., `----:com.apple.iTunes:LABEL`) when saving. This leads to duplicate tags in the saved file.  

* JIRA ticket (_optional_): [PICARD-3029](https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-3029)  

---  

# Solution  

The fix prioritizes maintaining the original tag capitalization. When saving or deleting a tag, `_casemap` is checked before applying the pre-mapped cases.  

---  

# Action  

Additional actions required:  
* [ ] Update Picard [documentation](https://github.com/metabrainz/picard-docs) 

### DIFF
--- a/picard/formats/mp4.py
+++ b/picard/formats/mp4.py
@@ -265,7 +265,19 @@ class MP4File(File):
                 name = 'lyrics'
             if name == 'comment:':
                 name = 'comment'
-            if name in self.__r_text_tags:
+            if name in self.__r_freeform_tags_ci:
+                values = [v.encode('utf-8') for v in values]
+                delall_ci(tags, self.__r_freeform_tags_ci[name])
+                if name in self.__casemap:
+                    name = self.__casemap[name]
+                else:
+                    name = self.__r_freeform_tags_ci[name]
+                tags[name] = values
+            elif name in self.__casemap:
+                values = [v.encode('utf-8') for v in values]
+                name = self.__casemap.get(name, name)
+                tags['----:com.apple.iTunes:' + name] = values
+            elif name in self.__r_text_tags:
                 tags[self.__r_text_tags[name]] = values
             elif name in self.__r_bool_tags:
                 tags[self.__r_bool_tags[name]] = (values[0] == '1')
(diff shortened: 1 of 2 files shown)


phw (commented):
Thanks for the bugfix. This looks sensible and should work.

But I'd like to test this some more and make sure the change does not have other side effects. Changing the order of tag processing could easily break some other case.

I'm currently low on time, hence it will take a couple of days until I get to testing this change fully.

jpmsousa03:
Sure! Let me know if I can help with anything. I'll also keep testing to check if there are any case where the processing might break.

test/formats/test_mp4.py
```diff
@@ -81,7 +81,7 @@ def test_replaygain_tags_case_insensitive(self):
         def test_ci_tags_preserve_case(self):
             # Ensure values are not duplicated on repeated save and are saved
             # case preserving.
-            for name in ('Replaygain_Album_Peak', 'Custom', 'äöüéß\0'):
+            for name in ('Replaygain_Album_Peak', "Label", 'Custom', 'äöüéß\0'):
```
phw, line 84: Minor thing, but please update this to use single quotation marks for consistency.

picard/formats/mp4.py
```diff
@@ -353,16 +357,19 @@ def supports_tag(cls, name):
     def _get_tag_name(self, name):
         if name.startswith('lyrics:'):
             name = 'lyrics'
-        if name in self.__r_text_tags:
+        if name in self.__r_freeform_tags_ci:
+            return self.__r_freeform_tags_ci[name]
+        if name in self.__casemap:
```
phw, line 362: Use `elif` here for clarity

phw (changes requested):
Thanks a lot. I checked this, and I can't see that it would break anything. General change is good o go, I just have two minor stylistic comments.

And sorry for the longer delay, last two weeks didn't exactly turn out as expected for me.

phw (approved):
Looks good, thanks a lot. LGTM

As a side note I wonder whether we need something similar for ID3. I haven't checked the code yet, but it might have the same issue.

jpmsousa03:
Thanks! I'm not sure if it happens in ID3 as well, I haven’t looked at the ID3 code yet, but I might take a look when I get some time.

PR #2627, opened 2025-04-03
PICARD-3054: Add new artist country variables

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

* **Describe this change in 1-2 sentences**:  Add new artist country variables.

# Problem

Changes have been made to the MusicBrainz webservice to include country information for all artists.  This is useful information that some users would like to be able to access from scripting within Picard.

* JIRA ticket (_optional_): PICARD-3054

# Solution

Add the following new variables:

~~**\_albumartistcountry**: A string showing a unique list of the country codes for all album artists.  If a country code is not provided by the webservice the code "XX" will be used.  Examples are "GB" (where all artists are from Great Britain), "GB & US" (where one or more artists are from Great Britain and one or more artists are from the United States), or "GB, US & XX" (where one or more artists are from Great Britain, one or more artists are from the United States, and one or more artists have no country code assigned).~~

**\_albumartists_countries**: A multi-value variable listing the country codes for all of the credited album artists, in the same order as the artists.  Duplicate country codes will be shown if there are more than one artist from the same country.  If a country code is not provided by the webservice the code "XX" will be used.  For example, if the first credited artist is from Great Britain and there are two other credited artists from Canada, the value would be "GB; CA; CA".

~~**\_artistcountry**: A string showing a unique list of the country codes for all track artists.  If a country code is not provided by the webservice the code "XX" will be used.  Examples are "GB" (where all artists are from Great Britain), "GB & US" (where one or more artists are from Great Britain and one or more artists are from the United States), or "GB, US & XX" (where one or more artists are from Great Britain, one or more artists are from the United States, and one or more artists have no country code assigned).~~

**\_artists_countries**: A multi-value variable listing the country codes for all of the credited track artists, in the same order as the artists.  Duplicate country codes will be shown if there are more than one artist from the same country.  If a country code is not provided by the webservice the code "XX" will be used.  For example, if the first credited artist is from Great Britain and there are two other credited artists from Canada, the value would be "GB; CA; CA".

# Action

Additional actions required:

We need to decide whether we want all four of these new variables, and if we want to use different variable names from those initially assigned.  **[Complete: Keep multi-value variables only.]**

We need to confirm the format of the values for the **\_albumartistcountry** and **\_artistcountry** variables, if we decide to keep them.  **[Complete: Variables removed.]**

We need to confirm how we want to handle artists with no country codes provided.  The initial code submission shows them with an Unknown Country code "XX".  This was done to ensure that the multi-value variables contain the same number of entries as there are credited artists to allow individual matching of country to artist.

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -28,7 +28,6 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-from collections import OrderedDict
 from types import SimpleNamespace
 
 from picard import log
@@ -43,6 +42,7 @@ from picard.util import (
     linear_combination_of_weights,
     parse_amazon_url,
     translate_from_sortname,
+    uniqify,
 )
 from picard.util.script_detector_weighted import detect_script_weighted
 
@@ -388,16 +388,6 @@ def artist_credit_from_node(node):
     return (artist_name, artist_sort_name, artist_names, artist_sort_names, artist_countries)
 
 
-def _format_artist_countries(countries: list):
-    if not countries:
-        return ''
-    # Use an OrderedDict to get a list of unique countries in the original countries order.
-    unique_countries = OrderedDict()
-    for country in countries:
-        unique_countries[country] = None
-    return ', '.join(list(unique_countries.keys()))
-
-
 def artist_credit_to_metadata(node, m, release=False):
     ids = [n['artist']['id'] for n in node]
     artist_name, artist_sort_name, artist_names, artist_sort_names, artist_countries = artist_credit_from_node(node)
@@ -407,7 +397,7 @@ def artist_credit_to_metadata(node, m, release=False):
         m['albumartistsort'] = artist_sort_name
         m['~albumartists'] = artist_names
         m['~albumartists_sort'] = artist_sort_names
-        m['~albumartistcountry'] = _format_artist_countries(artist_countries)
+        m['~albumartistcountry'] = ', '.join(uniqify(artist_countries))
         m['~albumartists_countries'] = artist_countries
     else:
         m['musicbrainz_artistid'] = ids
@@ -415,7 +405,7 @@ def artist_credit_to_metadata(node, m, release=False):
         m['artistsort'] = artist_sort_name
         m['artists'] = artist_names
         m['~artists_sort'] = artist_sort_names
-        m['~artistcountry'] = _format_artist_countries(artist_countries)
+        m['~artistcountry'] = ', '.join(uniqify(artist_countries))
         m['~artists_countries'] = artist_countries
 
 
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -369,7 +369,7 @@ def artist_credit_from_node(node):
         artist = artist_info['artist']
         if artist and 'id' in artist and artist['id']:
             # Add artist's country code if specified, otherwise 'XX' (Unknown Country)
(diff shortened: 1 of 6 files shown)


test/test_mbjson.py
```diff
@@ -317,6 +321,34 @@ def test_recording_instrument_credits(self):
...
+    """Test multiple artists with one unknown (missing) country.
+    """
+
+    filename = 'recording_multi_artists_2.json'
+
+    def test_recording_multi_artists_2(self):
+        m = Metadata()
+        t = Track('1')
+        recording_to_metadata(self.json_doc, m, t)
+        self.assertEqual(m['~artistcountry'], 'GB, US & XX')
```
zas, line 348: I find it weird we use different formats for both lists, what about using a comma + whitespace only?

```python
self.assertEqual(m['~artistcountry'], 'GB, US,  XX')
self.assertEqual(m['~artists_countries'], 'GB, US, XX')
```

Naming isn't very clear, one needs to check documentation to understand the difference between both.

rdswift: > I find it weird we use different formats for both lists, what about using a comma + whitespace only?

I'm okay with comma and whitespace only for separators.  That will actually simplify the code a bit.

> Naming isn't very clear, one needs to check documentation to understand the difference between both.

I actually tried to kind of follow the existing naming structure ("albumartist" and "albumartistsort" for the strings, and "\_albumartists" and "\_albumartists_sort" for the multi-values), but I'm open to alternate suggestions, especially if it will make it clearer for the users.

rdswift: Thinking about this a bit more, we can probably get by with only providing two multi-value variables (**\_albumartist_countries** and **\_artist_countries**), each containing the full list of country codes (including duplicates) for all credited artists.  The user can then use the **$unique()** function to get a consolidated list if they want to get rid of the duplicates.

picard/mbjson.py
```diff
@@ -380,24 +385,38 @@ def artist_credit_from_node(node):
...
+
+
+def _format_artist_countries(countries: list):
+    if not countries:
+        return ''
+    # Use an OrderedDict to get a list of unique countries in the original countries order.
+    unique_countries = OrderedDict()
+    for country in countries:
+        unique_countries[country] = None
+    return ', '.join(list(unique_countries.keys()))
```
zas, line 398: You can even simplify more using https://github.com/metabrainz/picard/blob/87b27f772d626addfa16f6217893285579330deb/picard/util/__init__.py#L590

zas (changes requested):

picard/mbjson.py
rdswift: Nice!  Changed in [9e16ee0](https://github.com/metabrainz/picard/pull/2627/commits/9e16ee0394b1cee5d0939eed731f3de7b253bb1d).

test/test_mbjson.py
rdswift: Removed the unique **\_albumartistcountry** and **\_artistcountry** variables in [cfff376](https://github.com/metabrainz/picard/pull/2627/commits/cfff37667d2b4c5beceff1d1d6f3471b3b35ab21).  This leaves only **\_albumartists_countries** and **\_artists_countries** multi-value variables, which contain the full artists country lists.  These can easily be converted to unique lists using the **$unique()** scripting function.

zas (approved):
LGTM, much simpler patch, and still has all features (with a little scripting).

phw (approved):

PR #2628, opened 2025-04-04
PICARD-3054: Add new artist country variables

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

* **Describe this change in 1-2 sentences**:  Add new artist country variables.

# Problem

Changes have been made to the MusicBrainz webservice to include country information for all artists.  This is useful information that some users would like to be able to access from scripting within Picard.

* JIRA ticket (_optional_): PICARD-3054

# Solution

Add the following new variables:

**\_albumartists_countries**: A multi-value variable containing the country codes for all of the credited album artists, in the same order as the artists.  Duplicate country codes will be shown if there are more than one artist from the same country.  If a country code is not provided by the webservice the code "XX" will be used.  For example, if the first credited artist is from Great Britain and there are two other credited artists from Canada, the value would be "GB; CA; CA".

**\_artists_countries**: A multi-value variable containing the country codes for all of the credited track artists, in the same order as the artists.  Duplicate country codes will be shown if there are more than one artist from the same country.  If a country code is not provided by the webservice the code "XX" will be used.  For example, if the first credited artist is from Great Britain and there are two other credited artists from Canada, the value would be "GB; CA; CA".

# Action

Additional actions required:

### DIFF
--- a/picard/mbjson.py
+++ b/picard/mbjson.py
@@ -9,7 +9,7 @@
 # Copyright (C) 2019 Michael Wiencek
 # Copyright (C) 2020 dukeyin
 # Copyright (C) 2020, 2023 David Kellner
-# Copyright (C) 2021 Bob Swift
+# Copyright (C) 2021, 2025 Bob Swift
 # Copyright (C) 2021 Vladislav Karbovskii
 #
 # This program is free software; you can redistribute it and/or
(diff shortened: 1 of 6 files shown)


phw (approved):
Thanks a lot

rdswift:
It appears that some of the tests have not been initiated.  Is there a problem with the github actions on this branch not matching the master branch?

phw:
@rdswift it's because we have changed the CI on master a lot, and it is expecting CI run names that aren't available on this branch. I guess we need to setup separate rules for this branch.

But for now we can ignore this. Important is that none of the tasks failed

zas (approved):
LGTM

PR #2630, opened 2025-04-06
PICARD-2750: Handle invalid LINK frames when saving MP3 files

francescouteiro:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
  
* **Description**: MP3 files with LINK frames that do not follow the ID3 specification cause mutagen exceptions when saving with ID3 v2.3 format. This fix handles invalid LINK frames to prevent errors during save operations.

# Problem

When processing MP3 files containing LINK frames that don't follow the ID3 specification, Picard returns an error of "Invalid frame ID" exception when saving with ID3 v2.3 format enabled.

Files from archive.org (like https://archive.org/details/yallcomesoundrec00page) store a null byte followed by a URL in the LINK frame instead of the required frame ID followed by URL. This invalid format causes mutagen to raise an exception during the ID3 v2.3 downgrade process.

Current workarounds include setting ID3 format to v2.4 or enabling "Clear existing tags" to remove the problematic frames. This fix proposes a handle for malformed LINK frames, preventing exceptions when saving with ID3 v2.3.

* JIRA ticket: PICARD-2750

# Solution

This fix addresses the exception by detecting and handling malformed LINK frames in MP3 files during the ID3 v2.3 saving process. When an invalid LINK frame is encountered (containing a null byte followed by URL instead of proper frame ID), it's transformed into a WXXX frame (user-defined website metadata/label).

This approach preserves the URL data from the malformed LINK frames by converting them to the appropriate frame type for URLs, rather than simply discarding the information. The conversion only happens when saving in ID3 v2.3 format, which would otherwise cause errors with these malformed frames.

While not the most comprehensive solution, this approach specifically targets the common issue with archive.org files without requiring UI changes or affecting performance. A more robust solution as suggested in the [community discussion](https://community.metabrainz.org/t/why-does-picard-refuse-to-save-these-mp3s/653191/11) could involve user prompts to review and decide on invalid tag data, but that would be a larger feature enhancement rather than a focused bug fix.

This implementation prioritizes stability for users of archive.org files while intelligently preserving the URL data in a proper ID3 frame format.

# Action

Additional actions required:

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -170,6 +170,7 @@ class ID3File(File):
         'TSOT': 'titlesort',
         'WCOP': 'license',
         'WOAR': 'website',
+        'WXXX': 'user_website',
         'COMM': 'comment',
         'TOAL': 'originalalbum',
         'TOPE': 'originalartist',
@@ -299,6 +300,8 @@ class ID3File(File):
         metadata = Metadata()
         for frame in tags.values():
             self._process_frame(frame, metadata, config_params)
+
+        self._process_link_frames_on_load(tags, metadata)
 
         if 'date' in metadata:
             self._sanitize_date(metadata)
@@ -614,6 +617,12 @@ class ID3File(File):
 
     def _save_tags(self, tags, filename):
         config = get_config()
+
+        try:
+            self._sanitize_id3_frames(tags)
+        except Exception as e:
+            log.error("Error sanitizing ID3 frames: %s", e)
+
         if config.setting['write_id3v1']:
             v1 = 2
         else:
@@ -622,10 +631,128 @@ class ID3File(File):
         if config.setting['write_id3v23']:
             tags.update_to_v23()
             separator = config.setting['id3v23_join_with']
-            tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
+            try:
+                tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
+            except ValueError:
+                for frame_id in list(tags.keys()):
+                    if not (frame_id.isalnum() and frame_id.isupper()):
+                        log.warning("Removing invalid frame ID: %r", frame_id)
+                        del tags[frame_id]
+                tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
         else:
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
+
+    def _sanitize_id3_frames(self, tags):
+        """This method attempts to fix various issues with ID3 frames:
+        1. Handle invalid LINK frames (convert to WXXX if possible)
+        2. Remove null bytes and other illegal characters from frame IDs
+        Note: In the future, we might consider prompting users about invalid frames."""
+        to_remove = []
+        to_add = []
+        confirmed_problematic_frames = []
+        extracted_urls = []
+        for frame_id, frame in list(tags.items()):
+            try:
+                invalid_id = False
+                if '\x00' in frame_id:
+                    invalid_id = True
+                if self._is_malformed_frame(frame_id, frame):
+                    invalid_id = True
+                    for url in self._extract_and_convert_link_frame(frame, frame_id):
+                        log.debug("Successfully converted malformed LINK frame %r to URL: %r", frame_id, url)
+                        extracted_urls.append(url)
+                        wxxx_frame = mutagen.id3.WXXX(encoding=Id3Encoding.LATIN1, desc="URL from malformed LINK frame", url=url)
+                        to_add.append(wxxx_frame)
+
+                if invalid_id:
+                    confirmed_problematic_frames.append(frame_id)
+                    to_remove.append(frame_id)
+
+            except Exception as e:
+                log.error("Error processing frame %r: %s", frame_id, e)
+
+        for frame_id in to_remove:
+            del tags[frame_id]
+        for frame in to_add:
+            tags.add(frame)
+
+        if extracted_urls and hasattr(self, 'metadata') and hasattr(self, 'orig_metadata'):
+            if 'user_website' not in self.metadata:
+                self.metadata['user_website'] = extracted_urls
+            if 'user_website' not in self.orig_metadata:
+                self.orig_metadata['user_website'] = extracted_urls
+
+        return confirmed_problematic_frames
+
+    def _extract_and_convert_link_frame(self, frame, frame_id):
+        """Extract URL from a malformed LINK frame and convert it to a WXXX frame. Handles cases where URLs
+        are split between frame ID and data or separated by null bytes (common for archive.org)"""
+
+        try:
+            url_parts = []
+            if hasattr(frame, 'data') and frame.data:
+                frame_data = frame.data.decode('latin1', errors='ignore')
+                if '\x00' in frame_data:
+                    parts = frame_data.split('\x00', 1)
+                    # Likely valid LINK frame if the first part looks like a URL and second part doesn't contain URL patterns
+                    if (parts[0] and self._contains_url_pattern(parts[0])
+                            and len(parts) > 1 and not self._contains_url_pattern(parts[1])):
+                        log.debug("Skipping valid LINK frame with URL %r", parts[0])
+                        return
+
+            if ':' in frame_id:
+                frame_id_parts = frame_id.split(':', 1)
+                if frame_id_parts[0] == 'LINK':
+                    # Standard case where everything after LINK: is part of the URL
+                    url_parts.append(frame_id_parts[1])
+                elif any(proto in frame_id_parts[0] for proto in ('LINKhttp', 'LINKhtt', 'LINKwww')):
+                    # Case where part of the URL protocol is in the frame_id
+                    prefix = 'LINK'
+                    protocol_part = frame_id_parts[0][len(prefix):]
+                    url_parts.append(protocol_part + ':' + frame_id_parts[1])
+
+            if hasattr(frame, 'data') and frame.data:
+                frame_data = frame.data.decode('latin1', errors='ignore').strip()
+                if frame_data:
+                    if '\x00' in frame_data:
+                        parts = frame_data.split('\x00')
+                        for part in parts:
+                            if part.strip() and self._contains_url_pattern(part):
+                                url_parts.append(part.strip())
+                    elif self._contains_url_pattern(frame_data):
+                        url_parts.append(frame_data)
+
+            combined_url = ''.join(url_parts).strip(':;,. \t\n\r')
+
+            if combined_url:
+                url_match = re.search(r'(?:https?:?(?:/+|\\+)|www\.)[a-zA-Z0-9][-a-zA-Z0-9\.]+\.[a-zA-Z]{2,}(?:/[^\s:;,]*)?', combined_url)
+                if url_match:
+                    raw_url = url_match.group(0)
+
+                    if raw_url.startswith('www.'):
+                        final_url = 'http://' + raw_url
+                    elif raw_url.startswith(('http://', 'https://')):
+                        final_url = raw_url
+                    else:
+                        final_url = re.sub(r'^htt:?p', 'http', raw_url)
+                        if not final_url.startswith(('http://', 'https://')):
+                            if final_url.startswith(':'):
+                                final_url = 'http' + final_url
+                            elif final_url.startswith('//'):
+                                final_url = 'http:' + final_url
+                            elif final_url.startswith('p://'):
+                                final_url = 'http://' + final_url[4:]
+                            else:
+                                final_url = 'http://' + final_url
+
+                    yield final_url
+                    return
+
+            log.warning("Could not extract URL from malformed LINK frame %r", frame_id)
+
+        except Exception as e:
+            log.error("Failed to process LINK frame %r: %s", frame_id, e)
 
     def format_specific_metadata(self, metadata, tag, settings=None):
         if not settings:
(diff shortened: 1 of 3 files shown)


picard/formats/id3.py
```diff
@@ -622,11 +631,153 @@ def _save_tags(self, tags, filename):
...
+        confirmed_problematic_frames = []
+        extracted_urls = []
+        for frame_id, frame in list(tags.items()):
+            try: 
+                invalid_id = False
+                if '\x00' in frame_id:
+                    invalid_id = True
+                if frame_id.startswith('LINK'):
+                    is_malformed = False
+                    if ':' in frame_id and any(term in frame_id for term in ('http', 'www.', '://')):
```
zas, line 665: `any(term in frame_id for term in ('http', 'www.', '://'))` expression (or similar) is used a lot below, move it to a function.

picard/formats/id3.py
```diff
@@ -1035,6 +1186,49 @@ def _remove_other_supported_tag(self, tags, real_name):
...
+
+                if is_potentially_malformed:
+                    link_frames.append((frame_id, frame))
+        
+        if not link_frames:
+            return
+
+        to_add = []
+        for frame_id, frame in link_frames:
+            if self._extract_and_convert_link_frame(frame, frame_id, to_add):
```
zas, line 1225: `_extract_and_convert_link_frame()` could be made a generator (using `yield`), instead of passing `to_add`.

zas (changes requested):
Overall it should work, but I think the code can be simplified (too much code redundancy, too long methods).

Also please fix linting errors.

francescouteiro:
Thanks for your insights, @zas! I think that I've addressed your concerns for redundancy, as well fixed linting errors!

picard/formats/id3.py
```diff
@@ -1035,6 +1162,53 @@ def _remove_other_supported_tag(self, tags, real_name):
...
+            if self._is_malformed_frame(frame_id, frame):
+                malformed_link_frames.append((frame_id, frame))
+
+        if not malformed_link_frames:
+            return
+
+        for frame_id, frame in malformed_link_frames:
+            for url in self._extract_and_convert_link_frame(frame, frame_id):
+                metadata.add('user_website', url)
+
```
zas, line 1211: Perhaps I miss something but can't this method be simplified even more:

```python

    def _process_link_frames_on_load(self, tags, metadata):
        """Process malformed LINK frames during initial file load and add them as website URLs."""

        for frame_id, frame in list(tags.items()):
            if self._is_malformed_frame(frame_id, frame):
                for url in self._extract_and_convert_link_frame(frame, frame_id):
                    metadata.add('user_website', url)
```

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+        if extracted_urls and hasattr(self, 'metadata') and hasattr(self, 'orig_metadata'):
+            if 'user_website' not in self.metadata:
+                self.metadata['user_website'] = extracted_urls
+            if 'user_website' not in self.orig_metadata:
+                self.orig_metadata['user_website'] = extracted_urls
+
+        return confirmed_problematic_frames
+
+    def _extract_and_convert_link_frame(self, frame, frame_id):
```
zas, line 688: Just a minor thing, but everywhere we have `frame_id, frame` but here, perhaps swap arguments.

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+    def _sanitize_id3_frames(self, tags):
+        """This method attempts to fix various issues with ID3 frames:
+        1. Handle invalid LINK frames (convert to WXXX if possible)
+        2. Remove null bytes and other illegal characters from frame IDs
+        Note: In the future, we might consider prompting users about invalid frames."""
+        to_remove = []
+        to_add = []
+        confirmed_problematic_frames = []
+        extracted_urls = []
+        for frame_id, frame in list(tags.items()):
```
zas, line 655: Is there a reason to convert `tags.items()` to `list` here?
Elsewhere `tags` are modified during the loop, but not here.

picard/formats/id3.py
```diff
@@ -1035,6 +1162,53 @@ def _remove_other_supported_tag(self, tags, real_name):
         """Remove other supported tag from ID3 frames."""
         del tags[real_name]
 
+    def _contains_url_pattern(self, text):
```
zas, line 1165: Since it doesn't depend on `self` it could be made a simple utility function at module level:

```python
def _contains_url_pattern(text):
     """Check if the given text contains URL-like patterns."""
    return any(term in text for term in ('http', 'www.', '://'))
```

It would shorten calls:

```python
if part.strip() and self._contains_url_pattern(part):
```

would become:

```python
if part.strip() and _contains_url_pattern(part):
```


If you prefer to keep it at class level, make it a static method at least (using `@staticmethod` decoration), but I think the code would be more readable with shorter calls.

zas (changes requested):
Code looks much better after recent changes, few more comments though.

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+                                final_url = 'http://' + final_url[4:]
+                            else:
+                                final_url = 'http://' + final_url
+
+                    yield final_url
+                    return
+
+            log.warning("Could not extract URL from malformed LINK frame %r", frame_id)
+
+        except Exception as e:
```
rdswift, line 754: This is a pretty broad exception capture.  What exceptions do you anticipate could occur (and where), and could they be specifically captured?

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+                        for part in parts:
+                            if part.strip() and self._contains_url_pattern(part):
+                                url_parts.append(part.strip())
+                    elif self._contains_url_pattern(frame_data):
+                        url_parts.append(frame_data)
+
+            combined_url = ''.join(url_parts).strip(':;,. \t\n\r')
+
+            if combined_url:
+                url_match = re.search(r'(?:https?:?(?:/+|\\+)|www\.)[a-zA-Z0-9][-a-zA-Z0-9\.]+\.[a-zA-Z]{2,}(?:/[^\s:;,]*)?', combined_url)
```
rdswift, line 729: Just a question because I'm really not sure, but would it be more efficient if the search/match string was pre-compiled as a constant?

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+                    prefix = 'LINK'
+                    protocol_part = frame_id_parts[0][len(prefix):]
+                    url_parts.append(protocol_part + ':' + frame_id_parts[1])
+
+            if hasattr(frame, 'data') and frame.data:
+                frame_data = frame.data.decode('latin1', errors='ignore').strip()
+                if frame_data:
+                    if '\x00' in frame_data:
+                        parts = frame_data.split('\x00')
+                        for part in parts:
```
rdswift, line 720: Minor comment:  This is likely clearer to read and understand, but since you only use `parts` in the `for` line, it could be shortened to `for part in frame_data.split('\x00'):`.

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+            if 'user_website' not in self.orig_metadata:
+                self.orig_metadata['user_website'] = extracted_urls
+
+        return confirmed_problematic_frames
+
+    def _extract_and_convert_link_frame(self, frame, frame_id):
+        """Extract URL from a malformed LINK frame and convert it to a WXXX frame. Handles cases where URLs
+        are split between frame ID and data or separated by null bytes (common for archive.org)"""
+
+        try:
```
rdswift, line 692: Does everything need to be included within the `try:` block?

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+        extracted_urls = []
+        for frame_id, frame in list(tags.items()):
+            try:
+                invalid_id = False
+                if '\x00' in frame_id:
+                    invalid_id = True
+                if self._is_malformed_frame(frame_id, frame):
+                    invalid_id = True
+                    for url in self._extract_and_convert_link_frame(frame, frame_id):
+                        log.debug("Successfully converted malformed LINK frame %r to URL: %r", frame_id, url)
```
rdswift, line 663: Is this an unusual enough situation that it might warrant logging at one level above `debug` (i.e. `info`)?

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+        Note: In the future, we might consider prompting users about invalid frames."""
+        to_remove = []
+        to_add = []
+        confirmed_problematic_frames = []
+        extracted_urls = []
+        for frame_id, frame in list(tags.items()):
+            try:
+                invalid_id = False
+                if '\x00' in frame_id:
+                    invalid_id = True
```
rdswift, line 659: I think this could be shortened to:

```suggestion
                invalid_id = '\x00' in frame_id
```

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
             tags.update_to_v23()
             separator = config.setting['id3v23_join_with']
-            tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
+            try:
+                tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
+            except ValueError:
+                for frame_id in list(tags.keys()):
+                    if not (frame_id.isalnum() and frame_id.isupper()):
+                        log.warning("Removing invalid frame ID: %r", frame_id)
+                        del tags[frame_id]
```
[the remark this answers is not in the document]
rdswift: An option seems a bit cleaner to me so that the user has some measure of control, but I don't have a feel for how common this situation is encountered.  It may not be worth having another setting (that most users won't understand).

picard/formats/id3.py
```diff
@@ -170,6 +170,7 @@ class ID3File(File):
         'TSOT': 'titlesort',
         'WCOP': 'license',
         'WOAR': 'website',
+        'WXXX': 'user_website',
```
rdswift, line 173: If we're going to standardize on the use of this tag, is it something that should be documented in https://picard-docs.musicbrainz.org/en/appendices/tag_mapping.html and/or https://picard-docs.musicbrainz.org/en/variables/variables_other_information.html or should we remain silent about it?

francescouteiro:
Thanks a lot for the insightful observations, @zas & @rdswift ! I’ll take some time to work through these suggestions and update the PR accordingly as soon as I can!

picard/util/tags.py
```diff
@@ -136,6 +136,7 @@
     'totaltracks': N_('Total Tracks'),
     'tracknumber': N_('Track Number'),
     'website': N_('Artist Website'),
+    'user_website': N_('User Defined Website'),
```
Sophist-UK, line 139: What about all the other non-ID3 formats - how will they handle this completely new tag?????

picard/formats/id3.py
```diff
@@ -622,11 +631,129 @@ def _save_tags(self, tags, filename):
...
+                        log.warning("Removing invalid frame ID: %r", frame_id)
+                        del tags[frame_id]
+                tags.save(filename, v2_version=3, v1=v1, v23_sep=separator)
         else:
             tags.update_to_v24()
             tags.save(filename, v2_version=4, v1=v1)
 
+    def _sanitize_id3_frames(self, tags):
+        """This method attempts to fix various issues with ID3 frames:
+        1. Handle invalid LINK frames (convert to WXXX if possible)
```
Sophist-UK, line 648: Perhaps I don't understand exactly what cases this is supposed to fix - so it might be helpful if it was spelled out in the comments here. 

Perhaps the most obvious confusion is that this PR talks specifically about issues with the LINK tag (see section 4.20 of the [ID3 2.4 spec](https://id3.org/id3v2.4.0-frames) but appears to be looking at all tags and not just LINK tags (and not even just tags starting with W (see section 4.3 of the [ID3 2.4 spec](https://id3.org/id3v2.4.0-frames) which might potentially have this issue). 

I would like to see the specific tags to be sanitized listed here and the specific issues and fixes spelled out.

Sophist-UK (changes requested):
This PR is problematic for me for several reasons:

1. Perhaps I don't understand exactly what cases this is supposed to fix. Is it just the LINK tag, or all URL tags (W000-WZZZ + LINK according to the ID3 v2.4 spec)? Why does it process every tag and not just these?

2. Is this simply an issue over the misplacing of a single 0x00 byte or something more complicated. I cannot understand how it needs 200 lines of new code (which is c. 15% of the size of the existing file) just to handle a simple issue over a mis-placed 0x00 byte.

3. I have a real problem about the use of exceptions to handle failures in the new code rather than avoiding them by checks in the code, leaving exceptions to indicate that something unexpected has happened as a genuine exception to what is expected to happen. But if you are going to use exception handling, it needs to be confined to as few lines as possible and catch THE specific exception that is expected and not just any old exception that might happen.

I cannot really review the bulk of the code until points 1. and 2. are addressed - because this seems to be a massive sledgehammer for a very tiny nut.

picard/formats/id3.py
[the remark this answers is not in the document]
phw: I agree that this could be an option. The question is however what we do, if the user disables the removal? We can't keep it as it is, as it will fail saving.

I actually also don't quite see why this code is part of the PR, it seems to try to solve an entirely different problem. Are there cases where the frame IDs are invalid like this?

picard/formats/id3.py
phw: Yes, I agree. I'd also prefer if this would fix the specific case we have.

Also something is strange here: For some reason this issue seems only to happen when saving ID3 v2.3, but not witth v2.4 (at least if what I wrote on the ticket is still true). Looking at the specs for "Linked information" in  both ID3 v2.3 and v2.4 I don't really understand why it should be. The specs are very similar. According to the problem description (files containing a LINK frame without frame ID, only the null byte) the LINK frame would be invalid for both.

So maybe mutagen does something different here, but for some reason can successfully save it to ID3 v2.4.

picard/util/tags.py
phw: Yes, this needs to be defined. We shouldn't just write `USER_WEBSITE` to Vorbis tags, this seems odd. I actually wouldn't expect a fix for the LINK frame issue to introduce a new tag. If we add a new tag it should be well handled from the start. If in doubt I would rather not write a new tag to some formats initially until we have figured out a proper mapping.

phw (changes requested):

picard/formats/id3.py
zas: Do we really need this? I mean it creates complications and the best would to avoid it totally.

zas:
@francescouteiro any update on this?

francescouteiro:
> @francescouteiro any update on this?

Very sorry for the delay! Last week was my exams' week, so I just started looking into all the things mentioned here recently! I'll address them between today and this Friday!

francescouteiro:
I'm again very sorry for the delay! I've been out of Wi-Fi at my address since Apr. 24 (plus no mobile data remaining), and with the recent power outage in Portugal it's still not solved. Thankfully, my data plan renewed today, so I can get back to this at 100%! TL;DR: This issue has not been forgotten or abandoned; I am still working on it!

zas:
Closing this stale PR, feel free to re-open if needed.

PR #2631, opened 2025-04-13
PICARD-3056: Set cover art local file chooser to FileMode.ExistingFiles

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3056

On macOS the cover art local file picker can crash. By default it seems to be allowing the selection of directories.

# Solution

Ensure the cover art file picker can only select existing files, and not directories or new files, by setting `FileMode.ExistingFiles`.

Also I think there was no reason to limit this to the first selected file, allow adding all selected files as cover art.

### DIFF
--- a/picard/ui/coverartbox/__init__.py
+++ b/picard/ui/coverartbox/__init__.py
@@ -275,8 +275,8 @@ class CoverArtBox(QtWidgets.QGroupBox):
         ])
         if file_chooser.exec():
             file_urls = file_chooser.selectedUrls()
-            if file_urls:
-                self.fetch_remote_image(file_urls[0])
+            for url in file_urls:
+                self.fetch_remote_image(url)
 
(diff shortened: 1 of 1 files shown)


phw:
Closing again. Fix is probably not fully correct. The code is currently using `picard.ui.util.FileDialog`, but likely shouldn't. Or if it does, this class needs to be fixed to handle all cases.

We primarily created this dialog for wrapping the case with our custom multi dir selection. I need to investigate in more detail. Also need to check the ripper log file picker, might be affected in a similar way.

As we refactored this quite heavily it behaves yet again differently in the 2.x branch, where we are using QFileDialog directly. There the above patch is likely the right approach.

PR #2635, opened 2025-04-15
Script completer minor refactoring

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -176,8 +176,8 @@ class ScriptCompleter(QCompleter):
     def __init__(self, parent=None):
         super().__init__(sorted(self.choices), parent)
         self.setCompletionMode(QCompleter.CompletionMode.UnfilteredPopupCompletion)
-        self.highlighted.connect(self.set_highlighted)
         self.last_selected = ''
+        self.highlighted.connect(self.set_highlighted)
 
     @property
     def choices(self):
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -312,7 +312,7 @@ class ScriptTextEdit(QTextEdit):
         super().__init__(parent)
         config = get_config()
         self.highlighter = TaggerScriptSyntaxHighlighter(self.document())
-        self.enable_completer()
+        self.initialize_completer()
         self.setFontFamily(FONT_FAMILY_MONOSPACE)
         self.setMouseTracking(True)
         self.setAcceptRichText(False)
@@ -414,7 +414,7 @@ class ScriptTextEdit(QTextEdit):
             QToolTip.hideText()
             self.setToolTip('')
 
-    def enable_completer(self):
+    def initialize_completer(self):
         self.completer = ScriptCompleter()
         self.completer.setWidget(self)
         self.completer.activated.connect(self.insert_completion)
--- a/picard/ui/widgets/scripttextedit.py
+++ b/picard/ui/widgets/scripttextedit.py
@@ -174,15 +174,15 @@ class TaggerScriptSyntaxHighlighter(QtGui.QSyntaxHighlighter):
 
 class ScriptCompleter(QCompleter):
     def __init__(self, parent=None):
-        super().__init__(sorted(self.choices), parent)
+        super().__init__(self.choices, parent)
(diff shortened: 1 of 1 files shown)


zas:
Note: https://github.com/metabrainz/picard/pull/2635/commits/4eb1e2f8535aace7e4eb28ee97592a12e20cc30e changes the completer behavior, was there a good reason to use Unfiltered mode?

@phw @rdswift Please test and tell me what you think

rdswift (dismissed):
Looks good to me.  Everything seems to work as expected in my testing.  Is this something that we would also consider porting to the `2.x` branch?

rdswift:
Maybe a dumb question, but is there a reason that we don't support tooltips in the script editor for `EXTRA_VARIABLES` and `PRESERVED_TAGS` like we do for `TAG_NAMES`?  Would we be open to a change to support these lists as well?

zas:
> Maybe a dumb question, but is there a reason that we don't support tooltips in the script editor for `EXTRA_VARIABLES` and `PRESERVED_TAGS` like we do for `TAG_NAMES`? Would we be open to a change to support these lists as well?

I see no good reason, it may help users, feel free to open a PR

zas:
> Looks good to me. Everything seems to work as expected in my testing. Is this something that we would also consider porting to the `2.x` branch?

Did you compare the behavior between Unfiltered mode (current master) and normal mode (this PR) for the the popup? I wonder if there are cases where previous behavior is better, and why unfiltered mode was chosen to start with.

rdswift:
> Did you compare the behavior between Unfiltered mode (current master) and normal mode (this PR) for the the popup?

I'll check again, but when I originally tested I didn't see any difference.

rdswift:
> > Did you compare the behavior between Unfiltered mode (current master) and normal mode (this PR) for the the popup?
> 
> I'll check again, but when I originally tested I didn't see any difference.

Oops.  I just checked again and I see that the selection list in the original version showed **everything** all the time but only moved the selected item highlight as you typed, and your version actually reduces the selection list.  I'm not sure why it was set up initially as unfiltered, but I prefer the way that you have it set up as filtered.

phw:
> Note: [4eb1e2f](https://github.com/metabrainz/picard/commit/4eb1e2f8535aace7e4eb28ee97592a12e20cc30e) changes the completer behavior, was there a good reason to use Unfiltered mode?

The behavior is different. In unfiltered mode it jumps to the suggestion, but shows all the possible completions. As the list is sorted alphabetically and matching looks at the start I found this more convenient to be able to look through all possible functions. But that's probably just my personal preference.

> Maybe a dumb question, but is there a reason that we don't support tooltips in the script editor for EXTRA_VARIABLES and PRESERVED_TAGS like we do for TAG_NAMES? Would we be open to a change to support these lists as well?

The tooltips work on all variables. Actually the code to show the tooltip just searches for the percentage sign and valid characters afterwards. So it even works for any user defined variable. Only that for variables we don't have much to show, so it only shows the name. This is different for functions, where we show the function documentation.

picard/ui/widgets/scripttextedit.py
```diff
@@ -176,15 +176,16 @@ def highlightBlock(self, text):
 
 class ScriptCompleter(QCompleter):
     def __init__(self, parent=None):
-        super().__init__(sorted(self.choices), parent)
-        self.setCompletionMode(QCompleter.CompletionMode.UnfilteredPopupCompletion)
-        self.highlighted.connect(self.set_highlighted)
+        super().__init__(self.choices, parent)
+        self.setCompletionMode(QCompleter.CompletionMode.PopupCompletion)
```
phw, line 180: If we change this to `PopupCompletion` we can also remove this line, as `PopupCompletion` is the default.

rdswift:
> I found this more convenient to be able to look through all possible functions. But that's probably just my personal preference.

I can live with it either way.

picard/ui/widgets/scripttextedit.py
zas: Done in https://github.com/metabrainz/picard/pull/2635/commits/ee2c11dc95251861816e9c078bc89ac92cd0b3cd

phw (approved):

phw (dismissed):

phw (approved):
LGTM

PR #2636, opened 2025-04-16
PICARD-3061: Add long descriptions to tags, display them in tooltips

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**:  Add tooltip descriptions to `PRESERVED_TAGS` and `EXTRA_VARIABLES`

# Problem

The tooltips shown for the tags and variables in the script editor only showed information for the `TAG_NAMES`.  There were no tooltip descriptions for `PRESERVED_TAGS` and `EXTRA_VARIABLES`.

* JIRA ticket (_optional_): PICARD-3061

# Solution

Add tooltip descriptions to `PRESERVED_TAGS` and `EXTRA_VARIABLES`.  Add test for tooltips on tags and variables.

# Action

Additional actions required:

Since this change will result in some new and changed string for translation, please review the new and existing tag description strings to ensure that any desired changes are captured prior to merging this PR.

### DIFF
--- a/picard/options.py
+++ b/picard/options.py
@@ -439,8 +439,11 @@ def init_options():
     pass
 
 
-def get_option_title(text):
-    key = ('setting', text)
-    if key not in Option.registry or not Option.registry[key].title:
-        return N_('No option title available')
-    return Option.registry[key].title
+def get_option_title(name):
+    key = ('setting', name)
+    if key not in Option.registry:
+        return None
+    title = Option.registry[key].title
+    if title:
+        return title
+    return N_("No title for setting '%s'") % name
--- a/picard/options.py
+++ b/picard/options.py
@@ -5,6 +5,7 @@
 # Copyright (C) 2024 Giorgio Fontanive
 # Copyright (C) 2024 Laurent Monin
 # Copyright (C) 2025 Philipp Wolfer
+# Copyright (C) 2025 Bob Swift
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -436,3 +437,10 @@ BoolOption('persist', 'script_editor_wordwrap', False)
 
 def init_options():
     pass
+
+
+def get_option_title(text):
+    key = ('setting', text)
+    if key not in Option.registry or not Option.registry[key].title:
+        return N_('No option title available')
+    return Option.registry[key].title
--- a/picard/options.py
+++ b/picard/options.py
@@ -208,10 +208,10 @@ ListOption('persist', 'options_pages_tree_state', [])
 # Fingerprinting
 TextOption('setting', 'acoustid_apikey', '')
 TextOption('setting', 'acoustid_fpcalc', '')
-TextOption('setting', 'fingerprinting_system', 'acoustid')
+TextOption('setting', 'fingerprinting_system', 'acoustid', title=N_('Use AcoustID audio fingerprinting'))
 IntOption('setting', 'fpcalc_threads', DEFAULT_FPCALC_THREADS)
 BoolOption('setting', 'ignore_existing_acoustid_fingerprints', False)
-BoolOption('setting', 'save_acoustid_fingerprints', False)
+BoolOption('setting', 'save_acoustid_fingerprints', False, title=N_('Save AcoustID fingerprints'))
 
 # picard/ui/options/general.py
 # General
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -102,12 +102,12 @@ class TagVar:
                         contain markdown.
         is_preserved: the tag is preserved (boolean, default: False)
         is_hidden: the tag is "hidden", name will be prefixed with "~" (boolean, default: False)
-        is_script_variable: the tag cannot be used as script variable (boolean, default: True)
+        is_script_variable: the tag can be used as script variable (boolean, default: True)
         is_tag: the tag is an actual tag (not a calculated or derived one) (boolean, default: True)
         is_calculated: the tag is obtained by external calculation (boolean, default: False)
         is_file_info: the tag is a file information, displayed in file info box (boolean, default: False)
         is_from_mb: the tag information is provided from the MusicBrainz database (boolean, default: True)
-        is_populated_by_picard: the tag information is not populated by stock Picard (boolean, default: False)
+        is_populated_by_picard: the tag information is populated by stock Picard (boolean, default: True)
         see_also: an iterable containing ids of related tags
         related_options: an iterable containing the related option settings (see picard/options.py)
         doc_links: an iterable containing links to external documentation (DocumentLink tuples)
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -8,7 +8,7 @@
 # Copyright (C) 2011-2013 Michael Wiencek
 # Copyright (C) 2012 Chad Wilson
 # Copyright (C) 2013 Calvin Walton
-# Copyright (C) 2013-2014, 2019-2021, 2023-2024 Laurent Monin
+# Copyright (C) 2013-2014, 2019-2021, 2023-2025 Laurent Monin
 # Copyright (C) 2013-2015, 2017 Sophist-UK
 # Copyright (C) 2019 Zenara Daley
 # Copyright (C) 2023, 2025 Bob Swift
@@ -32,7 +32,10 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-from collections import OrderedDict
+from collections import (
+    OrderedDict,
+    namedtuple,
+)
 from collections.abc import MutableSequence
 import re
 
@@ -42,13 +45,19 @@ try:
 except ImportError:
     markdown = None
 
+from picard.const import PICARD_URLS
 from picard.i18n import (
     N_,
     gettext as _,
 )
+from picard.profile import profile_setting_title
 
+
+DocumentLink = namedtuple('DocumentLink', ('title', 'link'))
 
 TEXT_NOTES = N_('Notes:')
+TEXT_SETTINGS = N_('Option Settings:')
+TEXT_LINKS = N_('Links:')
 TEXT_NO_DESCRIPTION = N_('No description available.')
 
 ATTRIB2NOTE = OrderedDict(
@@ -57,15 +66,16 @@ ATTRIB2NOTE = OrderedDict(
     is_calculated=N_('calculated'),
     is_file_info=N_('info from audio file'),
     not_from_mb=N_('not provided from MusicBrainz data'),
+    not_populated_by_picard=N_('not populated by stock Picard'),
 )
 
 
 class TagVar:
     def __init__(
         self, name, shortdesc=None, longdesc=None,
-        is_preserved=False, is_hidden=False, not_script_variable=False,
-        is_tag=True, is_calculated=False, is_file_info=False,
-        not_from_mb=False
+        is_preserved=False, is_hidden=False, is_script_variable=True, is_tag=True, is_calculated=False,
+        is_file_info=False, is_from_mb=True, is_populated_by_picard=True, see_also=None, related_options=None,
+        doc_links=None
     ):
         """
         shortdesc: Short description (typically one or two words) in title case that is suitable
@@ -76,22 +86,30 @@ class TagVar:
                   contain markup.
         is_preserved: the tag is preserved (boolean, default: False)
         is_hidden: the tag is "hidden", name will be prefixed with "~" (boolean, default: False)
-        not_script_variable: the tag cannot be used as script variable (boolean, default: False)
+        is_script_variable: the tag cannot be used as script variable (boolean, default: True)
         is_tag: the tag is an actual tag (not a calculated or derived one) (boolean, default: True)
         is_calculated: the tag is obtained by external calculation (boolean, default: False)
         is_file_info: the tag is a file information, displayed in file info box (boolean, default: False)
-        not_from_mb: the tag information is not provided from the MusicBrainz database (boolean, default: False)
+        is_from_mb: the tag information is provided from the MusicBrainz database (boolean, default: True)
+        is_populated_by_picard: the tag information is not populated by stock Picard (boolean, default: False)
+        see_also: an iterable containing ids of related tags
+        related_options: an iterable containing the related option settings (see picard/options.py)
+        doc_links: an iterable containing links to external documentation (DocumentLink tuples)
         """
         self._name = name
         self._shortdesc = shortdesc
         self._longdesc = longdesc
         self.is_preserved = is_preserved
         self.is_hidden = is_hidden
-        self.not_script_variable = not_script_variable
+        self.is_script_variable = is_script_variable
         self.is_tag = is_tag
         self.is_calculated = is_calculated
         self.is_file_info = is_file_info
-        self.not_from_mb = not_from_mb
+        self.is_from_mb = is_from_mb
+        self.is_populated_by_picard = is_populated_by_picard
+        self.see_also = see_also
+        self.related_options = related_options
+        self.doc_links = doc_links
 
     @property
     def shortdesc(self):
@@ -106,6 +124,18 @@ class TagVar:
         if self._longdesc:
             return self._longdesc.strip()
         return self.shortdesc
+
+    @property
+    def not_from_mb(self):
+        return not self.is_from_mb
+
+    @property
+    def not_script_variable(self):
+        return not self.is_script_variable
+
+    @property
+    def not_populated_by_picard(self):
+        return not self.is_populated_by_picard
 
     def __str__(self):
         """hidden marked with a prefix"""
@@ -125,6 +155,20 @@ class TagVar:
         for attrib, note in ATTRIB2NOTE.items():
             if getattr(self, attrib):
                 yield note
+
+    def settings(self):
+        if not self.related_options:
+            return None
+        for setting in self.related_options:
+            title = profile_setting_title(setting)
+            if title:
+                yield _(title)
+
+    def links(self):
+        if not self.doc_links:
+            return None
+        for doclink in self.doc_links:
+            yield f"<a href='{doclink.link}'>{doclink.title}</a>"
 
 
 class TagVars(MutableSequence):
@@ -198,19 +242,50 @@ class TagVars(MutableSequence):
         else:
             return title
 
+    @staticmethod
+    def _add_section(title, text):
+        return f"<p><strong>{title}</strong> {'; '.join(text)}.</p>"
+
     def display_tooltip(self, name):
         name, tagdesc, _search_name, item = self.item_from_name(name)
 
         title = _(item.longdesc) if item and item.longdesc else _(TEXT_NO_DESCRIPTION)
 
+        if markdown is None:
+            title = '<p>' + title.replace('\n', '<br />') + '</p>'
+        else:
+            title = markdown(title)
+
         notes = tuple(item.notes()) if item else tuple()
         if notes:
-            title += f"\n\n**{_(TEXT_NOTES)}** {'; '.join(notes)}."
+            title += self._add_section(_(TEXT_NOTES), notes)
+
+        if tagdesc:
+            return f"<p><em>%{name}%</em> [{tagdesc}]</p>{title}"
+        else:
+            return f"<p><em>%{name}%</em></p>{title}"
+
+    def display_full_description(self, name):
+        name, tagdesc, _search_name, item = self.item_from_name(name)
+
+        title = _(item.longdesc) if item and item.longdesc else _(TEXT_NO_DESCRIPTION)
 
         if markdown is None:
             title = '<p>' + title.replace('\n', '<br />') + '</p>'
         else:
             title = markdown(title)
+
+        notes = tuple(item.notes()) if item else tuple()
+        if notes:
+            title += self._add_section(_(TEXT_NOTES), notes)
+
+        settings = tuple(item.settings()) if item else tuple()
+        if settings:
+            title += self._add_section(_(TEXT_SETTINGS), settings)
+
+        links = tuple(item.links()) if item else tuple()
+        if links:
+            title += self._add_section(_(TEXT_LINKS), links)
 
         if tagdesc:
             return f"<p><em>%{name}%</em> [{tagdesc}]</p>{title}"
@@ -242,7 +317,7 @@ ALL_TAGS = TagVars(
             'found in a file, and is calculated using the Chromaprint software.'
         ),
         is_calculated=True,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'acoustid_id',
@@ -253,7 +328,7 @@ ALL_TAGS = TagVars(
             'if the fingerprints are similar enough.'
         ),
         is_calculated=True,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'albumartist',
@@ -282,9 +357,7 @@ ALL_TAGS = TagVars(
             'A multi-value variable containing the country codes for all of the credited album artists, '
             'in the same order as the artists. Duplicate country codes will be shown if there are more '
             'than one artist from the same country. If a country code is not provided by the webservice '
-            'the code "XX" will be used to indicate an unknown country. For example, if the first '
-            'credited artist is from Great Britain and there are two other credited artists from Canada, '
-            'the value would be "GB; CA; CA".'
+            'the code "XX" will be used to indicate an unknown country.'
         ),
         is_hidden=True,
         is_tag=False,
@@ -312,7 +385,7 @@ ALL_TAGS = TagVars(
         'albumsort',
         shortdesc=N_('Album Sort Order'),
         longdesc=N_('The sort name of the title of the release.'),
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'arranger',
@@ -347,9 +420,7 @@ ALL_TAGS = TagVars(
             'A multi-value variable containing the country codes for all of the credited track artists, '
             'in the same order as the artists. Duplicate country codes will be shown if there are more '
             'than one artist from the same country. If a country code is not provided by the webservice '
-            'the code "XX" will be used to indicate an unknown country. For example, if the first '
-            'credited artist is from Great Britain and there are two other credited artists from Canada, '
-            'the value would be "GB; CA; CA".'
+            'the code "XX" will be used to indicate an unknown country.'
         ),
         is_hidden=True,
         is_tag=False,
@@ -375,12 +446,16 @@ ALL_TAGS = TagVars(
         'barcode',
         shortdesc=N_('Barcode'),
         longdesc=N_('The barcode assigned to the release.'),
+        doc_links=(
+            DocumentLink(N_('Barcode in MusicBrainz documentation'), PICARD_URLS['mb_doc'] + 'Barcode'),
+            DocumentLink(N_('Barcode mapping in Picard documentation'),  PICARD_URLS['documentation'] + 'appendices/tag_mapping.html#id6'),
+        ),
     ),
     TagVar(
         'bpm',
         shortdesc=N_('BPM'),
         longdesc=N_('Beats per minute of the track.'),
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'bitrate',
@@ -390,7 +465,7 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'bits_per_sample',
@@ -400,7 +475,7 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'catalognumber',
@@ -419,7 +494,7 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
 
     TagVar(
@@ -427,10 +502,10 @@ ALL_TAGS = TagVars(
         shortdesc=N_('Comment'),
         longdesc=N_(
             'The disambiguation comment entered to help distinguish one release from another (e.g.: '
-            'Deluxe version with 2 bonus tracks). This is not populated by stock Picard. It may be '
-            'used and populated by certain plugins. Picard stores this information in the '
-            '`%_releasecomment%` variable.'
+            'Deluxe version with 2 bonus tracks).'
         ),
+        is_populated_by_picard=False,
+        see_also=('releasecomment', ),
     ),
     TagVar(
         'compilation',
@@ -462,7 +537,7 @@ ALL_TAGS = TagVars(
             'The copyright message for the copyright holder of the original sound, beginning with a '
             'year and a space character.'
         ),
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'datatrack',
@@ -474,7 +549,7 @@ ALL_TAGS = TagVars(
     TagVar(
         'date',
         shortdesc=N_('Date'),
-        longdesc=N_('The date that the release (album) was issued, in the format YYYY-MM-DD.'),
+        longdesc=N_('The date that the release (album) was issued, in the format `YYYY-MM-DD`.'),
     ),
     TagVar(
         'director',
@@ -491,13 +566,14 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         # TODO: Check if this actually exists or if it should be %_musicbrainz_discid%
         'discid',
         shortdesc=N_('Disc Id'),
         longdesc=N_(''),
+        see_also=('musicbrainz_discid', ),
     ),
     TagVar(
         'discnumber',
@@ -510,7 +586,7 @@ ALL_TAGS = TagVars(
         longdesc=N_('Set to 1 if the disc the track is on has a "pregap track", otherwise empty.'),
         is_hidden=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'discsubtitle',
@@ -526,13 +602,13 @@ ALL_TAGS = TagVars(
         'encodedby',
         shortdesc=N_('Encoded By'),
         longdesc=N_('The person or organization that encoded the track.'),
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'encodersettings',
         shortdesc=N_('Encoder Settings'),
         longdesc=N_('The settings used when encoding the track.'),
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'engineer',
@@ -546,25 +622,25 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'file_created_timestamp',
         shortdesc=N_('File created timestamp'),
-        longdesc=N_('The file creation timestamp in the form "YYYY-MM-DD HH:MM:SS" as reported by the file system.'),
+        longdesc=N_('The file creation timestamp in the form `YYYY-MM-DD HH:MM:SS` as reported by the file system.'),
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'file_modified_timestamp',
         shortdesc=N_('File modified timestamp'),
-        longdesc=N_('The file modification timestamp in the form "YYYY-MM-DD HH:MM:SS" as reported by the file system.'),
+        longdesc=N_('The file modification timestamp in the form `YYYY-MM-DD HH:MM:SS` as reported by the file system.'),
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'filename',
@@ -573,7 +649,7 @@ ALL_TAGS = TagVars(
         is_hidden=True,
         is_preserved=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'filepath',
@@ -581,7 +657,7 @@ ALL_TAGS = TagVars(
         longdesc=N_('Full path and name of the file.'),
         is_hidden=True,
         is_tag=False,
-        not_from_mb=True,
+        is_from_mb=False,
     ),
     TagVar(
         'filesize',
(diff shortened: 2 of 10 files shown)


rdswift:
Since this change will result in some new and changed string for translation, please review the new and existing tag description strings to ensure that any desired changes are captured prior to merging this PR.

Also, I 'm not sure if `test/test_script.py` is the right place for the tag tooltip tests.  Is there a better place for these tests?

phw (changes requested):
Generally I agree with the goal of providing more context for the variables. But I don't think the current approach works, as it is repurposing the tag name feature. Instead there should be a separate longer description. See my more detailed comments on the code.

As a side note I think the strings used for description should ideally be the same as used in the documentation. Maybe not always the full text. But the more similar they are (ideally identical) the easier it is to reuse the translations on Weblate.

picard/util/tags.py
```diff
@@ -72,10 +72,10 @@
     'engineer': N_('Engineer'),
     '~filepath': N_('File Path'),
     'gapless': N_('Gapless Playback'),
-    'genre': N_('Genre'),
+    'genre': N_('Genre (multi)'),
     'grouping': N_('Grouping'),
-    'isrc': N_('ISRC'),
-    'key': N_('Key'),
+    'isrc': N_('International Standard Recording Code (ISRC)'),
+    'key': N_('Musical key of the track'),
```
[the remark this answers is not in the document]
zas: I think we should redesign this, it would be more sense to have an object, I was thinking about something like (just a draft):

```python

class TagVar:
     def __init__(self, name, shortdesc=None, longdesc=None, is_preserved=False, is_hidden=False, is_script_variable=True, is_tag=True):
          self._name = name
          self._shortdesc = shortdesc
          self._longdesc= longdesc
          self.is_preserved = is_preserved
          self.is_hidden = is_hidden
          self.is_script_variable = is_script_variable
          self.is_tag = is_tag

    @property
    def shortdesc(self):
        """default to name""""
        if self._shortdesc:
            return self._shortdesc
        return self.name

    @property
    def longdesc(self):
        """default to shortdesc"""
        if self._longdesc:
            return self._longdesc
        return self.shortdesc

    def __str__(self):
        """"hidden marked with a prefix""""
        if self.is_hidden:
             return '~' + self._name
        else:
             return self._name

    def __hash__(self):
         """"_name should be unique (filepath == ~filepath)"""
         return self._name

ALL_TAG_VARS = {
    TagVar('xxxx', is_script_variable=False),
    TagVar('engineer', shortdesc=N_('Engineer')),
    TagVar('filepath', shortdesc=N_('File Path'), is_hidden=True),
    TagVar('bitrate', shortdesc=N_("Bitrate"), longdesc=N_('Approximate bitrate in kbps'), is_hidden=True, is_preserved=True),
    ....
}

TAG_NAMES = dict(str(tagvar): tagvar.shortdesc for tagvar in ALL_TAG_VARS if tagvar.is_tag)
PRESERVED_TAGS = list(str(tagvar) for tagvar in ALL_TAG_VARS if tagvar.is_preserved)
SCRIPT_VARIABLES = list(tagvar for tagvar in ALL_TAG_VARS if tagvar.is_script_variable)          # contains all script variables, replace (TAG_NAMES, PRESERVED_TAGS, EXTRA_VARIABLES)
```

@phw @rdswift It would simplify code I think, because all variables/tags are in one unique set (and __hash__ () ensures there are uniques). We can more easily switch a variable to hidden/preserved/etc.. without moving lines. What do you think?

picard/util/tags.py
```diff
@@ -182,44 +182,55 @@
...
+    '~totalalbumtracks': N_('Total number of tracks across all discs'),
+    '~video': N_('File is a video'),
+}
 
 
-def display_tag_name(name):
+def display_tag_name(name: str):
+    if not name.strip():
+        return ''
+    na = N_("No help description available")
```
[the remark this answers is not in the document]
rdswift: That's a matter of debate.  I added it in for completeness, but I can go either way.

rdswift:
I'm not sure why, but it appears that the Ubuntu `test-requirements` isn't loading the Markdown module from `requirements.txt`.

rdswift:
Rebased and pushed with significant changes from the original PR.  Please review the descriptions and tooltip code, along with identified `TODO:` entries.  We should remove all of the `TODO:` items before merging.

zas:
@rdswift what do you think about displaying the tooltip on tag's titles in metadata box, as done in https://github.com/rdswift/picard/pull/19 ?

zas (changes requested):

rdswift:
> what do you think about displaying the tooltip on tag's titles in metadata box, as done in [rdswift#19](https://github.com/rdswift/picard/pull/19) ?

Great idea.  I merged your changes.  Thanks.

picard/util/tags.py
```diff
@@ -156,6 +176,46 @@ def display_name(self, name):
...
+        if item:
+            if item.is_preserved:
+                notes.append(_(TEXT_READ_ONLY))
+            if item.is_calculated:
+                notes.append(_(TEXT_CALCULATED))
+            if item.is_file_info:
+                notes.append(_(TEXT_FROM_FILE))
+            if not item.is_script_variable:
+                notes.append(_(TEXT_NOT_FOR_SCRIPTING))
+            if not item.is_from_mb and not item.is_file_info:
```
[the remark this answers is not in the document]
rdswift: Great suggestion.  Changes made in https://github.com/metabrainz/picard/pull/2636/commits/02e5f280e8b68d00e6e29031984c226e1b6ddddd.

picard/util/tags.py
```diff
@@ -156,6 +176,46 @@ def display_name(self, name):
...
 
+    def display_tooltip(self, name):
+        tagdesc = None
+
+        if name and name.startswith('_'):
+            name = name.replace('_', '~', 1)
+
+        if ':' in name:
+            name, tagdesc = name.split(':', 1)
+
```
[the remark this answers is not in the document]
rdswift: Great suggestion.  Changes made in https://github.com/metabrainz/picard/pull/2636/commits/02e5f280e8b68d00e6e29031984c226e1b6ddddd.

rdswift:
Also, just a "heads up" that it looks like the Ubuntu `test-requirements` isn't loading the `markdown` module, so I ended up checking for that and providing alternate test results for some of the tests in `test_util_tags.py` in order to make the CI process happy.  The macOS and Windows versions seemed to load okay and passed without the alternative results.

zas:
> Also, just a "heads up" that it looks like the Ubuntu `test-requirements` isn't loading the `markdown` module, so I ended up checking for that and providing alternate test results for some of the tests in `test_util_tags.py` in order to make the CI process happy. The macOS and Windows versions seemed to load okay and passed without the alternative results.

`markdown` is optional, so we should improve matrix testing so we test with and without it.

rdswift:
> `markdown` is optional, so we should improve matrix testing so we test with and without it.

In that case, the upside is that I've covered both conditions in the testing for this PR.  😉

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
         'lyrics',
         shortdesc=N_('Lyrics'),
+        longdesc=N_('The lyrics for the track.'),
+        not_from_mb=True,
     ),
     TagVar(
         'media',
         shortdesc=N_('Media'),
+        # TODO: Confirm this is a multi-value.
+        longdesc=N_('A multi-value tag containing the media on which the release was distributed (e.g.: CD).'),
```
phw, line 683: This isn't multi-value. This is set per track, and a track only has a single media type

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
     ),
     TagVar(
         'mixer',
         shortdesc=N_('Mixer'),
+        longdesc=N_('The names of the "*Mixed By*" engineers associated with the track.'),
     ),
     TagVar(
+        # TODO: Check if this actually exists or if it was provided by the last.fm plugin.
```
phw, line 691: "Mood" is one of the defined and mapped tags

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
+            'release in the release group. This can provide, for example, the release date of the vinyl '
+            'version of what you have on CD.\n\nThis is the same information provided in the '
+            '`%_releasegroup_firstreleasedate%` variable, and is consistent across all tracks in the '
+            'release. If you prefer to have this tag populated with the date of the earliest recording of '
+            'the track in the database, which will likely be different for each track in the release, '
+            'this can be achieved by enabling a one-line tagging script as:\n\n'
+            '`$set(originaldate,%_recording_firstreleasedate%)`\n\nBe aware that setting this can cause a '
+            'release to be scattered across multiple directories if you use `%originaldate%` as part of '
+            'the path portion of your file naming script.\n\n**Note:** If you are storing tags in MP3 files as '
+            'ID3v2.3 then the original date can only be stored as a year.'
```
phw, line 852: I haven't seen this in action yet, but it seems to me this might be a bit too verbose for a tooltip, with a lot of very specific technical details.

phw:
> `markdown` is optional, so we should improve matrix testing so we test with and without it.

We already have this, the `test-requirements` tests are specifically for this. They test with reduced or older dependencies. E.g. the first test it runs does install the oldest versions we claim to support, and does not include any optional dependency. The other tests include specific versions of one of the optional dependencies. Currently this is now mostly used to test for alternative packages we support (discid vs. libdiscid, chardet vs. charset-normalizer).

So the purpose here is exactly to catch cases where tests fail with optional dependencies missing. Worked as intended :) 

The "normal" tests run with `requirements.txt` applied fully.

zas:
> We already have this, the `test-requirements` tests are specifically for this. 

Thanks for clarifying, I thought something wasn't working, but actually that's the reverse, it is working perfectly :)
@rdswift : since `markdown` is optional, you have to handle both cases in the code (as you discovered), that's not a problem with tests (a contrario).

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
         shortdesc=N_('Gapless Playback'),
+        longdesc=N_('Indicates whether or not there are gaps between the recordings on the release.'),
+        not_from_mb=True,
     ),
     TagVar(
         'genre',
         shortdesc=N_('Genre'),
+        longdesc=N_(
+            'A multi-value tag containing the specified genre information from MusicBrainz. Requires the '
+            '"Use genres from MusicBrainz" option setting be enabled.'
```
zas, line 616: I think we should list required options by name and use title property to link them to the tag. See https://github.com/metabrainz/picard/blob/36014f0a6e93ab80edf3ab221e0bea5a5209770f/picard/options.py#L243

If a tag depends on one (or more) options, it could be convenient to be able to make the link programmatically, and it limits need to re-translate strings if an option title changes.


```python

TagVar(

   related_options = ('use_genres', ),
```

picard/util/tags.py
zas: Yes, that's far too long for a tool tip, if we really need a long documentation we could add an attribute to store it (`extralongdesc` or whatever) but it shouldn't be displayed in the tooltip (it could in an internally generated documentation page for example).

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
         'originalfilename',
         shortdesc=N_('Original Filename'),
+        longdesc=N_('The original name of the audio file.'),
+        not_from_mb=True,
     ),
     TagVar(
         'originalyear',
         shortdesc=N_('Original Year'),
+        longdesc=N_(
+            'The year of the original release date in the format YYYY. By default this is set to the earliest '
```
zas, line 865:
```suggestion
            'The year of the original release date in the format `YYYY`. By default this is set to the earliest '
```

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
+            'The two-character code for the country in which the release was issued. If more than one release country '
+            'was specified, this tag will contain the first one in the list.'
+        ),
+    ),
+    TagVar(
+        'releasedate',
+        shortdesc=N_('Release Date'),
+        # TODO: Confirm long description and that this is not provided from the MB data.
+        longdesc=N_(
+            'The date that the release (album) was issued, in the format YYYY-MM-DD.'
```
zas, line 1023:
```suggestion
            'The date that the release (album) was issued, in the format `YYYY-MM-DD`.'
```

Not sure about that, but I think date formats should be enclosed in back quotes.

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
         'replaygain_reference_loudness',
         shortdesc=N_('ReplayGain Reference Loudness'),
+        longdesc=N_(''),
         is_calculated=True,
+        not_from_mb=True,
     ),
     TagVar(
         'replaygain_track_gain',
         shortdesc=N_('ReplayGain Track Gain'),
+        longdesc=N_(''),
```
zas, line 1164: `longdesc` is optional, no need to set it to empty string, but I guess you added them in order to fill them.

zas (changes requested):

picard/util/tags.py
rdswift: That was exactly it.  Before we merge I'll remove any blank ones.

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
         is_tag=False,
+        not_from_mb=True,
     ),
 
     TagVar(
         'comment',
         shortdesc=N_('Comment'),
+        longdesc=N_(
+            'The disambiguation comment entered to help distinguish one release from another (e.g.: '
+            'Deluxe version with 2 bonus tracks). This is not populated by stock Picard. It may be '
```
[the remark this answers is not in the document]
rdswift: I agree.  As an initial starting point, I just copied the descriptions from the Picard User Guide, which also explains why some are excessively long.

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
+        ),
         is_hidden=True,
         is_tag=False,
     ),
     TagVar(
-        'script',
-        shortdesc=N_('Script'),
-    ),
-    TagVar(
+        # TODO: Does this actually exist and does it come from MusicBrainz?  Is it an indicator if a track is silence?
```
zas, line 1211: It is set by https://github.com/metabrainz/picard/blob/36014f0a6e93ab80edf3ab221e0bea5a5209770f/picard/track.py#L319-L320

According to https://github.com/metabrainz/picard/blob/36014f0a6e93ab80edf3ab221e0bea5a5209770f/picard/const/__init__.py#L100 it is when the title matches `[silence]`.

So that's set by Picard, but doesn't come directly from MusicBrainz database.

picard/util/tags.py
```diff
@@ -166,652 +227,1128 @@ def names(self, selector=None):
...
+        shortdesc=N_('Release Country'),
+        longdesc=N_(
+            'The two-character code for the country in which the release was issued. If more than one release country '
+            'was specified, this tag will contain the first one in the list.'
+        ),
+    ),
+    TagVar(
+        'releasedate',
+        shortdesc=N_('Release Date'),
+        # TODO: Confirm long description and that this is not provided from the MB data.
```
zas, line 1021: I think it comes only from file tags, see https://tickets.metabrainz.org/browse/PICARD-2616

picard/util/tags.py
```diff
@@ -32,56 +32,79 @@
...
                    for a column header.
         longdesc: Brief description in sentence case describing the tag/variable.  This should
                   be similar (within reasonable length constraints) to the description in the Picard User
-                  Guide documentation, and could be used as a tooltip when reviewing a script.
+                  Guide documentation, and could be used as a tooltip when reviewing a script.  May
+                  contain markup.
         is_preserved: the tag is preserved (boolean, default: False)
         is_hidden: the tag is "hidden", name will be prefixed with "~" (boolean, default: False)
-        is_script_variable: the tag can be used as script variable (boolean, default: True)
+        not_script_variable: the tag cannot be used as script variable (boolean, default: False)
```
zas, line 79: Actually I think we should keep `is_*` for parameters, it would be more consistent, but we can add derived properties for `ATTRIB2NOTE` purpose:

```python
class TagVar:
def __init__(
        self, name, shortdesc=None, longdesc=None,
        is_preserved=False, is_hidden=False, is_script_variable=True,
        is_tag=True, is_calculated=False, is_file_info=False,
        is_from_mb=True,
    ):
    ....

    @property
    def not_from_mb(self):
        return not self.is_from_mb

     @property
     def not_script_variable(self):
        return not self.is_script_variable

```

zas (changes requested):

rdswift:
A major bit of cleanup and added suggested improvements.  Shorten tooltip descriptions. Add settings and links sections:

- Add a constant to PICARD_URLS for `https://musicbrainz.org/doc/`
- Add a helper to get a title for an option setting in `picard/profile.py`
- Only include the long description and notes in tag tooltips.
- Add a new method to get a tag's full description including the long description, notes, related settings and links.  This could be used to provide tag documentation similar to scripting function documentation.
- Change tag options to `is_*` for consistency. Use properties `not_*` for use with `ATTRIB2NOTE` as proposed by @zas.
- Update tests.

picard/util/tags.py
```diff
@@ -844,14 +1463,22 @@ def script_variable_tag_names():
     yield from (
         tagvar.script_name()
         for tagvar in ALL_TAGS
-        if tagvar.is_script_variable
+        if not tagvar.not_script_variable
```
zas, line 1466: This change can be reverted since we keep `is_script_variable`.

test/test_util_tags.py
```diff
@@ -45,6 +50,10 @@ def setUp(self):
         self.tagvar_hidden_sd = TagVar('hidden_sd', is_hidden=True, shortdesc='hidden_sd_shortdesc')
         self.tagvar_notag = TagVar('notag', is_tag=False)
         self.tagvar_nodesc = TagVar('nodesc')
+        self.tagvar_notes1 = TagVar('notes1', shortdesc='notes1_sd', longdesc='notes1_ld', is_preserved=True, is_calculated=True,
+                                   is_file_info=True, is_hidden=True, is_script_variable=False)
+        self.tagvar_notes2 = TagVar('notes2', shortdesc='notes2_sd', longdesc='notes2_ld', is_file_info=True, is_from_mb=False)
+        self.tagvar_notes3 = TagVar('notes3', shortdesc='notes3_sd', longdesc='notes3_ld', is_from_mb=False)
```
zas, line 56: Just a reminder, tests need to be extended to handle new methods/attributes.

picard/util/tags.py
```diff
@@ -98,6 +152,31 @@ def script_name(self):
         else:
             return self._name
 
+    def notes(self):
+        for attrib, note in ATTRIB2NOTE.items():
+            if getattr(self, attrib):
+                yield note
+
+    def settings(self):
```
zas, line 160: Perhaps it would be clearer if we used `related_options_titles`, `settings` is used elsewhere already and it can be misleading.

picard/util/tags.py
```diff
@@ -140,22 +219,96 @@ def insert(self, index, tagvar):
...
+
+        if tagdesc:
+            return f"<p><em>%{name}%</em> [{tagdesc}]</p>{title}"
+        else:
+            return f"<p><em>%{name}%</em></p>{title}"
+
+    def display_full_description(self, name):
+        name, tagdesc, _search_name, item = self.item_from_name(name)
+
+        title = _(item.longdesc) if item and item.longdesc else _(TEXT_NO_DESCRIPTION)
```
zas, line 278: `title` isn't the best name for that imho, as it contains a whole block of text, what about `text`, `content` or something like that instead?

picard/util/tags.py
```diff
@@ -140,22 +219,96 @@ def insert(self, index, tagvar):
...
+        if links:
+            title += self._add_section(_(TEXT_LINKS), links)
+
+        alsos = tuple(item.see_alsos()) if item else tuple()
+        if alsos:
+            temp = set()
+            for also_name in alsos:
+                also_name, _also_desc, _also_search_name, also_item = self.item_from_name(also_name)
+                if also_item:
+                    temp.add('%' + also_name + '%')
```
zas, line 303: This could be done in `see_alsos()` (getting the name and building the variable string), and we could then simplify code.
The same pattern is repeated:

```python
        links = tuple(item.links()) if item else tuple()
        if links:
            title += self._add_section(_(TEXT_LINKS), links)

        alsos = tuple(item.see_alsos()) if item else tuple()
        if alsos:
            title += self._add_section(_(TEXT_SEE_ALSO), alsos)
```

I was thinking about something like (untested, but that's the idea):

```python
    def _add_section(title, values):
          strings = tuple(values)
          if strings:
              return f"<p><strong>{_(title)}</strong> {'; '.join(strings)}.</p>"
          else:
              return ''

     if item:
        sections = {
           TEXT_NOTES: item.notes(),
           TEXT_SETTINGS: item.settings(),
           TEXT_LINKS: item.links(),
           TEXT_SEE_ALSO: item.alsos(),
        }
        for title, values in sections.items():
            title += self._add_section(title, values)

```

zas (changes requested):
It really comes along, nice improvements, few more suggestions (ofc) ;)

picard/util/tags.py
rdswift: Good catch!  I completely missed that.

picard/util/tags.py
rdswift: Good point.  I'll change it to `related_options_titles` as you suggested.

picard/util/tags.py
rdswift: Changed to `content` which better describes the variable.

picard/util/tags.py
rdswift: Nice!  I think I've implemented what you had in mind.  Also did some refactoring and reduced a bit of code duplication.

zas: I went a bit further in https://github.com/rdswift/picard/pull/20

picard/util/tags.py
```diff
@@ -166,652 +344,1118 @@ def names(self, selector=None):
...
+            'The list provided for each medium only includes the disc ids attached to that medium. For example, '
+            'the list provided for Disc 1 of a three CD set will not include the disc ids attached to discs 2 '
+            'and 3 of the set.'
+        ),
         is_hidden=True,
         is_tag=False,
     ),
     TagVar(
         'musicbrainz_originalalbumid',
         shortdesc=N_('MusicBrainz Original Release Id'),
```
zas, line 881: For those, I wonder if something likes:

`Original Release MBID`
`Original Artist MBID`
`Recording MBID`

wouldn't be better (there are shorter).
But it will change order, not sure if that's an issue.

@rdswift @phw what do you think?

picard/util/tags.py
```diff
@@ -166,652 +344,1118 @@ def names(self, selector=None):
...
+        longdesc=N_('The MusicBrainz Identifier (MBID) for the release group.'),
     ),
     TagVar(
         'musicbrainz_trackid',
         shortdesc=N_('MusicBrainz Track Id'),
+        longdesc=N_('The MusicBrainz Identifier (MBID) for the track.'),
     ),
     TagVar(
         'musicbrainz_tracknumber',
         shortdesc=N_('FIXME:musicbrainz_tracknumber'),
```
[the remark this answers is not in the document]
rdswift: Yes.  Working on the description cleanup now.

picard/util/tags.py
```diff
@@ -1207,7 +1207,7 @@ def names(self, selector=None):
     ),
     TagVar(
         'release_seriesid',
-        shortdesc=N_('Release Series Id'),
+        shortdesc=N_('Release Series ID'),
```
zas, line 1210: Longdesc says those are MBIDs

zas, line 1210: Longdesc says those are MBIDs

picard/util/tags.py
```diff
@@ -1241,55 +1241,55 @@ def names(self, selector=None):
...
         shortdesc=N_('ReplayGain Track Range'),
-        longdesc=N_(''),
+        longdesc=N_('Track range setting resulting from ReplayGain analysis of the track.'),
         is_calculated=True,
         is_from_mb=False,
     ),
     TagVar(
         'sample_rate',
-        shortdesc=N_('File sample rate'),
+        shortdesc=N_('File Sample Rate'),
```
zas, line 1292: Maybe just "Sample Rate"

zas (commented):
Just few comments

rdswift:
> Just few comments

That will teach me for working on this with only 3 hours sleep last night.  :stuck_out_tongue_winking_eye:

picard/util/tags.py
```diff
@@ -32,58 +32,134 @@
...
                   be similar (within reasonable length constraints) to the description in the Picard User
-                  Guide documentation, and could be used as a tooltip when reviewing a script.
+                  Guide documentation, and will be used as a tooltip when reviewing a script.  May
+                  contain markdown.
+        additionaldesc: Additional description which might include more details or examples.  May
+                        contain markdown.
         is_preserved: the tag is preserved (boolean, default: False)
         is_hidden: the tag is "hidden", name will be prefixed with "~" (boolean, default: False)
-        is_script_variable: the tag can be used as script variable (boolean, default: True)
+        is_script_variable: the tag cannot be used as script variable (boolean, default: True)
```
zas, line 105: The change in this comment can be reverted.

zas (changes requested):

picard/util/tags.py
```diff
@@ -165,653 +343,1119 @@ def names(self, selector=None):
...
         'gapless',
         shortdesc=N_('Gapless Playback'),
+        longdesc=N_('Indicates whether or not there are gaps between the recordings on the release.'),
+        is_from_mb=False,
     ),
     TagVar(
         'genre',
         shortdesc=N_('Genre'),
+        longdesc=N_(
+            'A multi-value tag containing the specified genre information from MusicBrainz.'
```
zas, line 733: I wonder, while we are at it, if it would be convenient to have a `TagVar.is_multi_value` flag.

picard/util/tags.py
```diff
@@ -165,653 +343,1119 @@ def names(self, selector=None):
...
         'length',
         shortdesc=N_('Length'),
+        longdesc=N_('The length of the track in format mins:secs.'),
         is_hidden=True,
         is_tag=False,
     ),
     TagVar(
         'license',
         shortdesc=N_('License'),
+        longdesc=N_('The licenses associated with the track, either through the release or recording relationships.'),
```
zas, line 778: Just to be sure `licenses` with a `s` is intentional here?

picard/util/tags.py
```diff
@@ -165,653 +343,1119 @@ def names(self, selector=None):
...
     TagVar(
         'podcast',
         shortdesc=N_('Podcast'),
+        longdesc=N_('Indicates whether or not the recording is a podcast.'),
+        is_from_mb=False,
     ),
     TagVar(
         'podcasturl',
         shortdesc=N_('Podcast URL'),
+        longdesc=N_('The associated url if the recording is a podcast.'),
```
zas, line 1009:
```suggestion
        longdesc=N_('The associated URL if the recording is a podcast.'),
```

That's an acronym.

picard/util/tags.py
```diff
@@ -165,653 +343,1119 @@ def names(self, selector=None):
...
+        ),
     ),
     TagVar(
         'secondaryreleasetype',
-        shortdesc=N_('FIXME:secondaryreleasetype'),
+        shortdesc=N_('Secondary Release Type'),
+        longdesc=N_(
+            'Zero or more secondary types (i.e.: *audiobook*, *compilation*, *dj-mix*, *interview*, '
+            '*live*, *mixtape/street*, *remix*, *soundtrack*, or *spokenword*) for the release group.'
+        ),
```
zas, line 1314: A good candidate for a link to MB doc about secondary types (https://musicbrainz.org/doc/Release_Group/Type)

picard/util/tags.py
```diff
@@ -165,653 +343,1119 @@ def names(self, selector=None):
...
+        longdesc=N_(
+            'An indicator of the "official" status of the release. Possible values include "*official*", '
+            '"*promotional*", "*bootleg*", and "*pseudo-release*".'
+        ),
     ),
     TagVar(
         'releasetype',
         shortdesc=N_('Release Type'),
+        longdesc=N_(
+            'A multi-value tag containing the types of release assigned to the release group.'
```
zas, line 1232: https://musicbrainz.org/doc/Release_Group/Type

zas (changes requested):

picard/util/tags.py
rdswift: Not intentional.  I just copied what was written in the docs.  It probably makes more sense shown as singular.

picard/util/tags.py
rdswift: I think that's a good idea.  I'll add it in.

zas:
https://github.com/metabrainz/picard/pull/2636/commits/18aef96b064a08319df2aa136a3495f2bfd17a2b looks good, think about test for `is_multi_value` in `TagVarTest::test_basic_properties`.

picard/options.py
```diff
@@ -436,3 +437,10 @@ def make_default_toolbar_layout():
 
 def init_options():
     pass
+
+
+def get_option_title(text):
+    key = ('setting', text)
+    if key not in Option.registry or not Option.registry[key].title:
```
zas, line 444: If no key, I think we should return `None` (or even raise an exception and catch it in the caller?).
When there's no title, I would rather display the option name (something like "No title for setting <name>").

```python

def get_option_title(name):
    key = ('setting', name)
    if key not in Option.registry:
        return None
    title = Option.registry[key].title
    if title:
        return title
    return N_("No title for setting '%s'") % name
```

Minor thing, but option name (rather than text).

zas (changes requested):

picard/options.py
rdswift: I like it.  Updated in https://github.com/metabrainz/picard/pull/2636/commits/648556d072d221626577346247a35c2986a8bb5d

picard/util/tags.py
```diff
@@ -165,653 +346,1176 @@ def names(self, selector=None):
...
         'musicbrainz_artistid',
-        shortdesc=N_('MusicBrainz Artist Id'),
+        shortdesc=N_('Artist MBID'),
+        longdesc=N_('The MusicBrainz Identifiers (MBIDs) for the track artists.'),
+        is_multi_value=True,
     ),
     TagVar(
         'musicbrainz_discid',
-        shortdesc=N_('MusicBrainz Disc Id'),
+        shortdesc=N_('Disc MBID'),
```
zas, line 883: There is a little confusion here, that's not an MBID, but a MusicBrainz discid as per https://musicbrainz.org/doc/Disc_ID_Calculation.
This one is a 20 byte SHA-1 signature is converted into a Base64 encoded string of printable ASCII characters, so 28 characters-long string

Also below, we have the discid, as used per FreeDB & CDDB in the past, and still calculated by certain programs:
https://fr.wikipedia.org/wiki/DiscId
This one is 8 hex chars.

image

Perhaps we should rename this one "FreeDB discid" and the other "MusicBrainz DiscID".

zas (changes requested):

picard/util/tags.py
```diff
@@ -880,12 +892,15 @@ def names(self, selector=None):
...
-        shortdesc=N_('Disc MBID'),
+        shortdesc=N_('MusicBrainz DiscID'),
         longdesc=N_(
             'The Disc ID is the code number which MusicBrainz uses to link a physical CD to a release listing. '
             'This is based on the table of contents (TOC) information read from the disc. This tag contains the '
             'Disc ID if the album information was retrieved using `Tools > Lookup CD` from the menu.'
         ),
+        see_also=('discid', ),
+        is_calculated=True,
+        doc_links=(DocumentLink(N_('Disc ID Calculations'), PICARD_URLS['mb_doc'] + 'Disc_ID_Calculation'), ),
```
zas, line 903:
```suggestion
        doc_links=(DocumentLink(N_('Disc ID Calculation'), PICARD_URLS['mb_doc'] + 'Disc_ID_Calculation'), ),
```

picard/util/tags.py
```diff
@@ -165,653 +354,1183 @@ def names(self, selector=None):
...
     TagVar(
         'discid',
-        shortdesc=N_('Disc Id'),
+        shortdesc=N_('FreeDB Disc ID'),
+        longdesc=N_('The identification number of the disc in the FreeDB database.'),
+        see_also=('musicbrainz_discid', ),
+        doc_links=(
+            DocumentLink(N_('FreeDB'), 'https://wikipedia.org/wiki/Freedb'),
+            DocumentLink(N_('FreeDB DiscID Calculation'), 'https://wikipedia.org/wiki/CDDB#Example_calculation_of_a_CDDB1_(FreeDB)_disc_ID'),
+            DocumentLink(N_('FreeDB DiscID including Calculation Example (French)'), 'https://fr.wikipedia.org/wiki/DiscId'),
```
zas, line 642: While it makes total sense to mark the title for translation, it isn't translated in `TagVar.links()`, since `_()` isn't called.

```python
        for doclink in self.doc_links:
            translated_title = _(doclink.title)
            yield f"<a href='{doclink.link}'>{translated_title}</a>"
```

picard/util/tags.py
```diff
@@ -98,6 +177,45 @@ def script_name(self):
...
+        if not self.related_options:
+            return None
+        for setting in self.related_options:
+            title = get_option_title(setting)
+            if title:
+                yield _(title)
+
+    def links(self):
+        if not self.doc_links:
+            return None
```
zas, line 195:
```suggestion
            return
```

picard/util/tags.py
```diff
@@ -98,6 +177,45 @@ def script_name(self):
...
             return self._name
 
+    def notes(self):
+        for attrib, note in ATTRIB2NOTE.items():
+            if getattr(self, attrib):
+                yield note
+
+    def related_options_titles(self):
+        if not self.related_options:
+            return None
```
zas, line 187:
```suggestion
            return
```

picard/util/tags.py
```diff
@@ -98,6 +177,45 @@ def script_name(self):
...
+
+    def links(self):
+        if not self.doc_links:
+            return None
+        for doclink in self.doc_links:
+            yield f"<a href='{doclink.link}'>{doclink.title}</a>"
+
+    def see_alsos(self):
+        if not self.see_also:
+            return None
```
zas, line 201: Actually a simple `return` can do.

Actually that's equivalent, since returning None is the default for Python functions.

```python
>>> def f():
...     return
... 
>>> f() is None
True
```

```suggestion
            return
```

picard/util/tags.py
```diff
@@ -140,22 +258,93 @@ def insert(self, index, tagvar):
...
+            return '<p>' + text.replace('\n', '<br />') + '</p>'
+        return markdown(text)
+
+    def _base_description(self, tagname):
+        name, tagdesc, _search_name, item = self.item_from_name(tagname)
+        content = self._markdown(_(item.longdesc) if item and item.longdesc else _(TEXT_NO_DESCRIPTION))
+        if tagdesc:
+            return f"<p><em>%{name}%</em> [{tagdesc}]</p>{content}"
+        else:
+            return f"<p><em>%{name}%</em></p>{content}"
```
zas, line 309: We should have formats in a translatable string (in case it varies with language).

```python
        fmt_tagdesc = _("<p><em>%{name}%</em> [{tagdesc}]</p>{content}")
        fmt_normal = _("<p><em>%{name}%</em></p>{content}")
        fmt = fmt_tagdesc if tagdesc else fmt_normal
        return fmt.format(name=name, content=content, tagdesc=tagdesc)
```

zas (changes requested):

picard/util/tags.py
rdswift: Changed in https://github.com/metabrainz/picard/pull/2636/commits/48158eb412dd68c3b7906bedeef8e3be9bfb3ec1

picard/util/tags.py
rdswift: Changed in https://github.com/metabrainz/picard/pull/2636/commits/48158eb412dd68c3b7906bedeef8e3be9bfb3ec1

picard/util/tags.py
```diff
@@ -140,22 +259,93 @@ def insert(self, index, tagvar):
...
+
+        item: TagVar = self._name2item.get(search_name, None)
+
+        return name, tagdesc, search_name, item
+
+    def script_name_from_name(self, name):
+        tagname, tagdesc, search_name, item = self.item_from_name(name)
+        if item:
+            return str(item)
+        return
```
zas, line 284: Other functions were not expected to return anything, but this one returns a string (or None).
Even though None is the default we usually explicitly return None in such functions.
In this specific case, we could even just not call return at all since that's the end of the function, but let's be explicit in this case and keep `return None`. 

```suggestion
        return None
```

zas (dismissed):
LGTM

zas:
@rdswift do you have more changes to add to this PR? I think it is now (very) good, so I approved it.
I'd like @phw to review it too before any merge, there are more than expect changes (for the good imho) in this PR so extensive review is needed.

rdswift:
> @rdswift do you have more changes to add to this PR? I think it is now (very) good, so I approved it.

Nothing more from me on this PR.  I may have some more description changes, or link / see also / option items to add, but I will start a new PR for that.

> I'd like @phw to review it too before any merge, there are more than expect changes (for the good imho) in this PR so extensive review is needed.

I'd like to have him review it again as well.  One thing we still didn't finalize is whether or not to use "No description provided" or just repeat the tag name as the tooltip if there is no longdesc or shortdesc or the tag is not found in `TagVars`.

picard/util/tags.py
```diff
@@ -98,6 +177,46 @@ def script_name(self):
         else:
             return self._name
 
+    def notes(self):
+        for attrib, note in ATTRIB2NOTE.items():
+            if getattr(self, attrib):
+                yield note
```
rdswift, line 183: Should this be `yield _(note)`?

zas: Yes, absolutely, good catch.

It also shows we lack some tests.

I fixed this issue and added tests in https://github.com/rdswift/picard/pull/22

zas (approved):
LGTM

phw (approved):
Wow, that turned out to be much more than expected. Code looks good to me, and I test ran this and it feels really good with detailed information on the tooltips.

I assume there is a plan on how to use `display_tag_full_description`?

Only thing I would consider is moving `ALL_TAGS` to `picard.const.tags`. The `picard.util.tags` module has become much bigger now with a lot of utility code. I think it would be beneficial to keep the tag definition itself separate. But as I understand there is more planned here, so we can also do this separately.

The tag descriptions are looking. We might have some adjustments once we update the translations and start translating it in Weblate. 

> l. One thing we still didn't finalize is whether or not to use "No description provided" or just repeat the tag name as the tooltip if there is no longdesc or shortdesc or the tag is not found in `TagVars`.

The "*No description available*" looks good to me in the tooltip. It makes it quite clear and handles expectations.

Anyway, good to merge to me. Awesome work, thanks to both of you!

zas:
> I assume there is a plan on how to use `display_tag_full_description`?

Yes, @rdswift already works on an internal documentation page that will make use of it.

> 
> Only thing I would consider is moving `ALL_TAGS` to `picard.const.tags`. The `picard.util.tags` module has become much bigger now with a lot of utility code. I think it would be beneficial to keep the tag definition itself separate. But as I understand there is more planned here, so we can also do this separately.

I plan to restructure code a bit, perhaps even moving from `util/tags` to `tags/` and split code in smaller files. I didn't want to move code in this PR.

Thanks for the review.

PR #2640, opened 2025-05-01
PICARD-3062: Add tag documentation to script editor help

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds tag and variable documentation to the script editors help system

# Problem

The script editors (Options > Script and File Naming Script Editor) both provide help documentation regarding available functions (one through a new window and the other as a sidebar).  The available tags and variables help should be added to this.

* JIRA ticket (_optional_): PICARD-3062

# Solution

Add the full descriptions for the tags and variables to a new panel in the existing help display widget, and provide a button to toggle between the function help and the tags help.

Consider adding a section to the Picard User Guide showing the File Naming Script Editor and how it is used.

# Action

Additional actions required:

### DIFF
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -217,6 +217,35 @@ class TagVar:
                 values='; '.join(values),
             )
 
+    @staticmethod
+    def _markdown(text: str):
+        if markdown is None:
+            return '<p>' + text.replace('\n', '<br />') + '</p>'
+        return markdown(text)
+
+    def _add_sections(self, include_sections):
+        # Note: format has to be translatable, for languages not using left-to-right for example
+        fmt = _("<p><strong>{title}:</strong> {values}.</p>")
+        return ''.join(self._gen_sections(fmt, include_sections))
+
+    def full_description(self):
+        content = self._markdown(_(self.longdesc) if self.longdesc else _(TEXT_NO_DESCRIPTION))
+
+        # Append additional description
+        if self.additionaldesc:
+            content += self._markdown(_(self.additionaldesc))
+
+        # Append additional sections as required
+        include_sections = (
+            Section.notes,
+            Section.options,
+            Section.links,
+            Section.see_also,
+        )
+        content += self._add_sections(include_sections)
+
+        return content
+
 
 class TagVars(MutableSequence):
     """Mutable sequence for TagVar items
--- a/picard/util/tags.py
+++ b/picard/util/tags.py
@@ -38,6 +38,7 @@ from collections import (
 )
 from collections.abc import MutableSequence
 from enum import IntEnum
+import html
 import re
 
 
@@ -58,6 +59,7 @@ DocumentLink = namedtuple('DocumentLink', ('title', 'link'))
 
 
 def _markdown(text: str):
+    text = html.escape(text)
     if markdown is None:
         return '<p>' + text.replace('\n', '<br />') + '</p>'
     return markdown(text)
@@ -186,7 +188,7 @@ class TagVar:
     def notes(self):
         for attrib, note in ATTRIB2NOTE.items():
             if getattr(self, attrib):
-                yield _(note)
+                yield html.escape(_(note))
 
     def related_options_titles(self):
         if not self.related_options:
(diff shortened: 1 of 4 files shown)


zas (approved):
It looks to me.

I have an issue with links color on my system as shown below:
image

But that's unrelated to this PR (@phw any idea?).

A feature we may add is linking See Also tags so we can go from one to the other, but it can be done later (if not too complex).

rdswift:
> A feature we may add is linking See Also tags so we can go from one to the other, but it can be done later (if not too complex).

Great idea!  It was actually pretty easy, so I included it here.

zas:
@rdswift I'm bothered by the button that toggles between Functions/Tags at top right. IMHO we should use a QTabWidget instead and have 2 tabs. It would be easier to expand (adding more tabs), and it's more common UI-wise.
It would also simplify the code (perhaps).

What do you think?

rdswift:
> What do you think?

I like it, but I can't seem to get it to work in the File Naming Script Editor (even though it works fine in the separate documentation window opened from Options > Scripting window).  It looks like it has something to do with the splitter causing something else (the "Title:" text box?) to block access to anything on the left side of the top line of the documentation frame.  I've tried everything I can think of to correct this with no success.  This is the reason I went with the button on the right side of the frame in the first place.

Any suggestions?

phw:
> But that's unrelated to this PR (@phw any idea?).

In theme.py we already set the link color based on the accent color, see `BaseTheme.update_palette` which defines a color for `QtGui.QPalette.ColorRole.Link`. Theoretically this should also apply here, not sure. Maybe it needs to be set by stylesheet explicitly for this view. But yes, separate issue to solve elsewhere.

picard/util/tags.py
```diff
@@ -57,6 +57,12 @@
 DocumentLink = namedtuple('DocumentLink', ('title', 'link'))
 
 
+def _markdown(text: str):
+    if markdown is None:
+        return '<p>' + text.replace('\n', '<br />') + '</p>'
```
phw, line 62: We should have HTML escaping here to secure against accidental HTML reserved characters in `text`.

picard/util/tags.py
```diff
@@ -217,6 +226,38 @@ def _gen_sections(self, fmt, include_sections):
...
+            Section.links,
+            Section.see_also,
+        )
+        content += self._add_sections(include_sections)
+
+        return content
+
+    def full_description(self):
+        content = self.full_description_content()
+        return _("<p><em>%{name}%</em></p>{content}").format(name=self.script_name(), content=content)
```
phw, line 259: Also HTML escpaing

phw (changes requested):
This looks great. Using tabs would also be what I would have suggested, and with tabs its a really clean interface.

Overall not much to say here, changes look good. But in the places where text gets added into HTML formatted text we should use `html.escape`.

zas:
@rdswift @phw I rewrote code to be more object-oriented in https://github.com/rdswift/picard/pull/25

I agree with escaping html.

picard/util/tags.py
```diff
@@ -58,6 +59,7 @@
 
 
 def _markdown(text: str):
+    text = html.escape(text)
```
zas, line 62: Shouldn't we escape only when using html (and not markdown)?

phw: Not sure, as you can actually have HTML in Markdown it probably is correct that way.

rdswift: I'm confused.  Don't we want to escape any reserved HTML characters in the text before we pass it to `markdown` to turn the markdown text into html?  If we did the `html.escape()` after the markdown processing, wouldn't that destroy the resulting html text produced by `markdown`?

rdswift: My previous comment applies to the tags.  With respect to the function documentation, I'll need to go back into the code to see how it is generated.  I assumed (and I should know better) that the proper encoding and checks had already been addressed, so I just used the existing code for the content of the "Functions" tab.

zas (approved):
LGTM

phw (approved):

PR #2643, opened 2025-05-03
PICARD-3064: Fix ID3File hiding File._save_images

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3064

After recent refactorings the ID3File class overwrote the File._save_images method, triggering an exception on saving.

The issue was introduced in c36e4a47b66235d2ce4e4bd76a9017a19b564152

# Solution

Rename `ID3File._save_images` to `ID3File._save_images_to_tags`.

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -3,7 +3,7 @@
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006-2009, 2011-2012 Lukáš Lalinský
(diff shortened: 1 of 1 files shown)


zas (approved):
Good catch!

Obvious error now we see it, thanks for the fix.

PR #2644, opened 2025-05-03
PICARD-3066: Fix theme initialization and set generic accent color in dark mode

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3066

There are actually two changes in this PR:

1. The first commit is a straight forward fix to the theme being initialized too late since https://github.com/metabrainz/picard/commit/a2dd9df9ee75332958ed63e65ae90eaba31f7c4d . This is fixed by moving the theme setup as early as possible, but after all logging setup.
2. It attempts to have a solution for the link color in dark mode not being readable on Linux on some systems. The color in this case is system defined, so results varies. This change assumes the color for highlighting active UI elements is a good choice as the accent color (because if there was such an accent color defined by the system it would be used as the highlight color).

The latter change has no effect on Windows and macOS. In both cases there is a defined accent color, and it already gets used as the link color as well.

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -2,7 +2,7 @@
 #
 # Picard, the next-generation MusicBrainz tagger
 #
-# Copyright (C) 2019-2022, 2024 Philipp Wolfer
+# Copyright (C) 2019-2022, 2024-2025 Philipp Wolfer
 # Copyright (C) 2020-2021 Gabriel Ferreira
 # Copyright (C) 2021-2024 Laurent Monin
 #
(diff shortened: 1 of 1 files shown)


phw:
@zas Can you test this on your system? The first commit is I think a nobrainer, we can merge that to master in any case. The second is more experimental.

zas:
image

It is much more readable for sure :)

zas:
@phw update PR title & description (since first commit was removed)

picard/ui/theme.py
```diff
@@ -141,7 +143,7 @@ def is_dark_theme(self):
 
     @property
     def accent_color(self):  # pylint: disable=no-self-use
-        return None
+        return self._accent_color
```
zas, line 146: It should still be defined to None if not dark theme.

zas (changes requested):
It works for me.

picard/ui/theme.py
phw: Right, fixed this

zas: Did you push the change? ;)

phw: Now yes :)

zas (approved):

PR #2646, opened 2025-05-03
Allow up to 3 chars for Windows file name compatibility replacements

jpauwels:
# Summary

* This is a…
  * [] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
Allow up to 3 chars for Windows file name compatibility replacements.

# Problem

Only allowing one-for-one replacements of problematic characters on Windows limits the flexibility in rewriting file names. Concretely, I'd like to replace a colon with a dash with additional whitespace, such that "Title: Subtitle" becomes "Title - Subtitle" instead of "Title- Subtitle".

# Solution

Adapt the regex conformity check to allow up to three characters replacement for a single problematic character.

# Action

Additional actions required:

### DIFF
--- a/picard/ui/options/renaming_compat.py
+++ b/picard/ui/options/renaming_compat.py
@@ -172,7 +172,7 @@ class NoDirectorySeparatorValidator(QtGui.QValidator):
 
 
 class WinCompatReplacementValidator(QtGui.QValidator):
-    _re_valid_win_replacement = re.compile(r'^[^"*:<>?|/\\\s]?$')
+    _re_valid_win_replacement = re.compile(r'^[^"*:<>?|/\\\s]{0,3}$')
 
     def validate(self, text: str, pos):
         if self._re_valid_win_replacement.match(text):


phw (changes requested):
This makes sense. But the patch alone will not fix this. The UI limits the `maxLength` for the inputs to 1 character only. Also the inputs themselves then should get a wider width.

If you want to give this a try you can update the `ui/win_compat_dialog.ui` file, which should be done using the Qt6 designer tool.

phw:
As this is a functional change that needs to be added to the changelog we also need a ticket on https://tickets.metabrainz.org/ before we can merge this.

phw:
I added PICARD-3084 to track this. But the PR still needs more work before it is mergeable. The UI must be adjusted to allow more characters as well.

rdswift:
I'm not against this change, but couldn't the desired behavior be produced using a tagger script?  Something like:

```
$set(_new_title,$rreplace($replace(%title%,:, - ),\\s+, ))
```

and then use `%_new_title%` in the file naming script instead of `%title%`?

PR #2648, opened 2025-05-03
Added a condition to strip dots after the tracknumber in filenames

Marethyun:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

A lot of music files are named following the template `<tracknumber>. <title>.<extension>`, Picard recognizes this and interprets the tracknumber, but takes the whole rest as the title, resulting in titles beginning by a dot and then a space. 

Real world example right after importing a folder of yet-untagged music :

image

# Solution

I added an offset to skip the dot when it is present, and the following space is stripped like it would have if the filename started with some whitespace.

There could be (extreme) edge cases where titles actually start with a dot altogether with specific file naming (for example `715.TOTO.flac` could mean the title is `.TOTO` or `TOTO` and there would be no way of programmatically telling), so I thought it could be useful to add a configuration option for this, but I don't know the software nor the codebase enough to do that properly.

### DIFF
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -647,7 +647,9 @@ def tracknum_and_title_from_filename(base_filename):
         stripped_filename = filename.lstrip('0')
         tnlen = len(tracknumber)
         if stripped_filename[:tnlen] == tracknumber:
-            title = stripped_filename[tnlen:].lstrip()
+            # Strip the dot in front of the tracknumber, if present
+            dot_offset = 1 if stripped_filename[tnlen:][0] == '.' else 0
+            title = stripped_filename[tnlen + dot_offset:].lstrip()
 
     return GuessedFromFilename(tracknumber, title)
 


picard/util/__init__.py
```diff
@@ -647,7 +647,9 @@ def tracknum_and_title_from_filename(base_filename):
         stripped_filename = filename.lstrip('0')
         tnlen = len(tracknumber)
         if stripped_filename[:tnlen] == tracknumber:
-            title = stripped_filename[tnlen:].lstrip()
+            # Strip the dot in front of the tracknumber, if present
+            dot_offset = 1 if stripped_filename[tnlen:][0] == '.' else 0
+            title = stripped_filename[tnlen + dot_offset:].lstrip()
```
zas, line 652: If we were to strip dots, I think it should be done before we strip zeros, in order to handle '.01' for example.

Also, please add matching tests to https://github.com/metabrainz/picard/blob/50a29a166bf99c2a6919e7020818380871271c2c/test/test_utils.py#L636-L654

It would be great to create a matching ticket for this feature on https://tickets.metabrainz.org/projects/PICARD and prefix PR title with PICARD-XXX:

zas (changes requested):

picard/util/__init__.py
Marethyun: Didn't see the test case, I'm expanding it now.

> EDIT: In your case, you want to strip dot after the number, but before the title, it looks like a very specific case to me.

The case I addressed is the only one also addressed in regexes earlier in the parsing (trackname ending in a dot) :
https://github.com/metabrainz/picard/blob/8a062c9a7c103e1abed485e62f75bdc332753311/picard/util/__init__.py#L605

So I think the case isn't that specific, in general it applies to file names formatted as bullet points, I've seen the following cases in the real world:
* `10. XXX.flac`
* `10- XXX.flac`
* `10) XXX.flac` (extremely rarely, especially on non-POSIX platforms)

So if the behavior needs to be extended I think it's in this regard, and I may do that if you agree.

> If we were to strip dots, I think it should be done before we strip zeros, in order to handle '.01' for example.

On the other hand I've never seen dots (as that would also make hidden files in many platforms) or any other non-digit before the tracknumber, maybe with the exception of files like `(10) XXX.flac`, but that's out of scope I think.

phw: Filenames with track number followed by do are rather common, I don't think this case is that specific. And the track number extraction specifically tests for this case. Hence the patch with the test case makes sense to me.

Track numbers with leading dots look uncommon. One exception would be intentionally hidden files on posix systems, as you mentioned. But I would actually not consider this edge case.

In other cases I would actually not consider a number with a leading zero the track number. E.g. ".38"  well could refer to caliber 38 of a gun or such.

phw (approved):

test/test_utils.py
```diff
@@ -679,6 +679,8 @@ def test_returns_expected_tracknumber(self):
             (('99', 'Foo'), '99 Foo.mp3'),
             (('42', 'Foo'), '0000042 Foo.mp3'),
             (('2', 'Foo'), '0000002 Foo.mp3'),
+            # Tracknumber dot not in title
+            (('77', 'Foo.mp3'), '77. Foo.mp3'),
```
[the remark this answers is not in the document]
Marethyun: That was dumb, I fixed it

zas (approved):
LGTM

phw (approved):

chewi:
This broke things for me when the file is just called `1.opus`, which is how I name tracks between ripping and tagging.

```
Picard terminated unexpectedly
  File "/usr/lib/python3.13/site-packages/picard/tagger.py", line 790, in event
    event.run()
    ~~~~~~~~~^^
  File "/usr/lib/python3.13/site-packages/picard/util/thread.py", line 54, in run
    self.func(*self.args, **self.kwargs)
    ~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/picard/file.py", line 299, in _loading_finished
    self._copy_loaded_metadata(result, postprocessors)
    ~~~~~~~~~~~~~~~~~~~~~~~~~~^^^^^^^^^^^^^^^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/picard/file.py", line 313, in _copy_loaded_metadata
    processor(metadata)
    ~~~~~~~~~^^^^^^^^^^
  File "/usr/lib/python3.13/site-packages/picard/file.py", line 320, in _guess_tracknumber_and_title
    guessed = tracknum_and_title_from_filename(self.base_filename)
  File "/usr/lib/python3.13/site-packages/picard/util/__init__.py", line 658, in tracknum_and_title_from_filename
    dot_offset = 1 if stripped_filename[tnlen:][0] == '.' else 0
                      ~~~~~~~~~~~~~~~~~~~~~~~~~^^^
IndexError: string index out of range
```

zas:
> This broke things for me when the file is just called `1.opus`, which is how I name tracks between ripping and tagging.
>

Ticket created as https://tickets.metabrainz.org/browse/PICARD-3149

zas:
Fix: https://github.com/metabrainz/picard/pull/2902

PR #2649, opened 2025-05-04
Add provision for adding/removing plugin related tags

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds provision to register plugin created tags & variables

# Problem

Tags and variables created by plugins do not appear in the autocomplete list or scripting documentation.

# Solution

This PR provides the ability to register plugin-related tags and variables (when a plugin is enabled), and automatically removes (de-registers) the tags & variables when a plugin is no longer enabled.

# Action

Additional actions required:


rdswift:
Need some input as to when to register the tags & variables (when a plugin is loaded or enabled), and whether they should be removed from `ALL_TAGS` when the plugin is disabled or when the plugin is removed.  If not removed from `ALL_TAGS` until the plugin is removed, then there will need to be some filtering so that tags from installed but not enabled plugins do not appear in the autocomplete list or documentation.

rdswift:
I also envision adding the name of the plugin to the full documentation display, and possibly to the tooltip display.

zas:
I like it. Not sure if it should be merged yet though (we can merge and modify).
New plugin system is the thing we need to work on now, the idea of having a proper way to register variables is definitively the way to go.

phw:
Yes, we should integrate this once the new plugin system is in place. I'll get the current PR into a shape were we can continue from there.

ig-raiderler:
I'm impressed by the modularity of this codebase. Makes it very maintainable.

phw:
I think this is solved in the new plugin system by the `PluginApi.register_script_variable` API. This is now also used by tooltips and documentation.

PR #2653, opened 2025-05-12
Rework setting log verbosity

zas:
- handle incorrect value (possible if from config file)
- use effective level for comparaison & storage
- introduce `log.set_verbosity()`, which can optionally save it to config
- drop `tagger._debug` and use new method `log.is_debug()` instead
- drop `tagger.set_log_level()` in favor of `log.set_verbosity()`

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/log.py
+++ b/picard/log.py
@@ -62,6 +62,8 @@ if not picard_module_path.is_dir():
 
 _MAX_TAIL_LEN = 10**6
 
+_DEFAULT_LOG_LEVEL = logging.INFO
+
 
 def set_level(level):
     main_logger.setLevel(level)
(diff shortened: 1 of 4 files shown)


picard/log.py
```diff
@@ -71,6 +73,24 @@ def get_effective_level():
     return main_logger.getEffectiveLevel()
 
 
+def set_verbosity(level, save_to_config=False):
```
phw, line 76: I think we don't need both `set_level` and `set_verbosity`. Having both makes for a confusing API where it is unclear which one to use.

The change here to extend the function to handle errors and allow config saving makes sense. But I think we could either extend `set_level` with the functionality or replace it with `set_verbosity` if you prefer that name. No strong opinion either way, maybe I slightly prefer `set_verbosity` to be slightly more specific.

phw (changes requested):

rdswift (commented):
Generally looks good to me.  I agree with @phw's comment about likely not needing both `set_level` and `set_verbosity`.  I suggest sticking with `set_verbosity`.

picard/log.py
zas: I got rid of `set_level()` in https://github.com/metabrainz/picard/pull/2653/commits/c1e7b26d3bf95e93e79023330da043272c001417

rdswift (approved):
Looks good to me.

phw (approved):
LGTM

PR #2656, opened 2025-05-13
Browser integration: code cleanup and logging improvements

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:


picard/browser/browser.py
```diff
@@ -107,16 +112,26 @@ def start(self):
...
+        class OurHTTPServer(ThreadingHTTPServer):
+            address_family = ADDRESS_FAMILY
+
         config = get_config()
         if config.setting["browser_integration_localhost_only"]:
-            host_address = '127.0.0.1'
+            host_address = BROWSER_INTEGRATION_LOCALIP
         else:
-            host_address = '0.0.0.0'  # nosec
+            host_address = LISTEN_ALL  # nosec
```
phw, line 129: The `# nosec` can go here. It's well possible that the bandit tool will complain on the new line above where `'0.0.0.0'` is used. We'll see.

picard/const/__init__.py
```diff
@@ -163,3 +163,11 @@
 )
 
 SCRIPT_LANGUAGE_VERSION = '1.1'
+
+BROWSER_INTEGRATION_IPV6 = False
```
phw, line 167: If we want to provide this as an experimental feature we could have it controlled by an environment variable or such.

phw (commented):
Generally the code looks good. The question is, whether we actually want to merge the preparations for IPv6 already, even though they are of little practical use. Or whether we only merge the refactoring parts (new constants, logging changes etc.).

Also one minor thing I noticed: With IPv6 the log shows:

```
Starting the browser integration (::1:8001)
```

Would be nice if it instead would use the IPv6 URL syntax `[::1]:8001` to make this less confusing.

picard/browser/browser.py
```diff
@@ -166,11 +174,17 @@ def do_GET(self):
             log.error('Browser integration failed handling request', exc_info=True)
             self._response(500, 'Unexpected request error')
 
+    def _fmt_message(self, format, args):
+        message = format % args
+        return "%s %s" % (self.address_string(), message.translate(self._control_char_table))
```
phw, line 179: Where is `_control_char_table` defined?

picard/browser/browser.py
```diff
@@ -166,11 +174,17 @@ def do_GET(self):
...
+
     def log_error(self, format, *args):
-        log.error(format, *args)
+        message = self._fmt_message(format, args)
+        log.error(message)
```
phw, line 183: There is a minor change in behavior here. `log.error` only makes the argument substitution, if the message will get actually logged. Now the message gets processed regardless of log level.

Not sure, it's probably not really relevant in this context.

zas:
> Would be nice if it instead would use the IPv6 URL syntax `[::1]:8001` to make this less confusing.

I worked around this in https://github.com/metabrainz/picard/pull/2656/commits/0a2d5b71f69547501d2c96712e60812bc5953a5a

Log messages now look like:
`
I: 18:58:29,848 browser/browser.start:134: Browser Integration: Starting, listening on address 0.0.0.0 and port 8000
I: 18:58:55,647 browser/browser.log_message:191: Browser Integration: 127.0.0.1 "GET /openalbum?id=72056637-660a-4b0e-b1f8-f608c40daaa8 HTTP/1.1" 200 -
I: 18:59:14,901 browser/browser.stop:147: Browser Integration: Stopping
`

`
E: 19:08:23,598 browser/browser.start:140: Browser Integration: Failed to find an available port in range 8000-65535 on address 0.0.0.0
D: 19:08:23,598 browser/browser.stop:156: Browser Integration: inactive, no need to stop
`

picard/browser/browser.py
zas: In parent class, BaseHTTPRequestHandler,, perhaps we shouldn't rely on it, and use our own code. Not really sure it is needed, but it looks safer.

picard/browser/browser.py
zas: Yes, but it reduces code redundancy, not much impact as it is just when queries are handled, unlikely critical.

picard/const/__init__.py
zas: I don't think that's necessary at this point, I removed IPv6 specific code anyway (it was more a proof of concept than anything).

picard/browser/browser.py
phw: It's internal, so there is a small risk it will break between Python versions. But I see how the parent class uses it, so when overriding the log methods it makes sense to apply the same handling.

phw: This line is a bit long, and there is a lot going on for each parameter. I'd break this up over multiple lines

```python
        return "%s: %s %s" % (LOG_PREFIX,
                              self.address_string(),
                              message.translate(self._control_char_table))
```

picard/browser/browser.py
```diff
@@ -163,14 +175,20 @@ def do_GET(self):
         try:
             self._handle_get()
         except Exception:
-            log.error('Browser integration failed handling request', exc_info=True)
+            log.error('%s: failed handling request', LOG_PREFIX, exc_info=True)
             self._response(500, 'Unexpected request error')
 
+    def _fmt_message(self, format, args):
+        message = format % args
+        return "%s: %s %s" % (LOG_PREFIX, self.address_string(), message.translate(self._control_char_table))
```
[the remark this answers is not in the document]
zas: Addressed in https://github.com/metabrainz/picard/pull/2656/commits/31945b31ac8e6360763a369dd05e443970a10a87 as well as double evaluation

picard/browser/browser.py
zas: See https://github.com/metabrainz/picard/commit/31945b31ac8e6360763a369dd05e443970a10a87

picard/browser/browser.py
zas: Addressed in https://github.com/metabrainz/picard/commit/31945b31ac8e6360763a369dd05e443970a10a87 (better not rely on parent class private property).

zas:
@phw the file is named `browser.py` but it is actually the (local) server code, we have `browser/browser.py`.
I think we should rename those, what about `browser_integration/server.py` (`browser` is misleading I think, and `browser_integration` a bit long, ...).

@rdswift any ideas?

rdswift:
> @rdswift any ideas?

Not really.  I suppose if you didn't mind keeping it under `browser`, you could perhaps use `browser/server.py` or `browser/integration.py`.

phw (dismissed):
Looks good. LGTM

phw:
@rdswift @zas I agree, naming is not quite fitting, but other names are also difficult. I'd go with Bob's suggestion of using `browser/server.py`. It's still about the browser integration, but makes it clear this holds the server code.

picard/browser/browser.py
zas: I saw no complaints, after https://github.com/metabrainz/picard/pull/2656/commits/4a60bb5692d698d0299007e5610772f8529dc1ce so I guess we can just drop it.

phw (approved):

PR #2658, opened 2025-05-18
Columns & Search Dialog improvements & tidy up

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -43,6 +43,7 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
+from PyQt6 import QtCore
 
 from picard.i18n import N_
 from picard.util import icontheme
@@ -53,7 +54,7 @@ from picard.ui.columns import (
     Columns,
     ColumnSortType,
     DefaultColumn,
-    IconColumn,
+    ImageColumn,
 )
 
 
@@ -66,6 +67,42 @@ def _sortkey_filesize(obj):
         return int(obj.metadata['~filesize'] or obj.orig_metadata['~filesize'])
     except ValueError:
         return 0
+
+
+class IconColumn(ImageColumn):
+    _header_icon = None
+    header_icon_func = None
+    header_icon_size = QtCore.QSize(0, 0)
+    header_icon_border = 0
+
+    @property
+    def header_icon(self):
+        # icon cannot be set before QApplication is created
+        # so create it during runtime and cache it
+        # Avoid error: QPixmap: Must construct a QGuiApplication before a QPixmap
+        if self._header_icon is None:
+            self._header_icon = self.header_icon_func()
+        return self._header_icon
+
+    def set_header_icon_size(self, width, height, border):
+        self.header_icon_size = QtCore.QSize(width, height)
+        self.header_icon_border = border
+        self.size = QtCore.QSize(width + 2*border, height + 2*border)
+        self.width = self.size.width()
+
+    def paint(self, painter, rect):
+        icon = self.header_icon
+        if not icon:
+            return
+        h = self.header_icon_size.height()
+        w = self.header_icon_size.width()
+        border = self.header_icon_border
+        padding_v = (rect.height() - h) // 2
+        target_rect = QtCore.QRect(
+            rect.x() + border, rect.y() + padding_v,
+            w, h
+        )
+        painter.drawPixmap(target_rect, icon.pixmap(self.header_icon_size))
 
 
 _fingerprint_column = IconColumn(N_("Fingerprint status"), '~fingerprint')
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -74,9 +74,9 @@ _fingerprint_column.set_header_icon_size(16, 16, 1)
 
 
 ITEMVIEW_COLUMNS = Columns((
-    DefaultColumn(N_("Title"), 'title', sort_type=ColumnSortType.NAT, size=250, always_visible=True, status_icon=True),
-    DefaultColumn(N_("Length"), '~length', align=ColumnAlign.RIGHT, sort_type=ColumnSortType.SORTKEY, sortkey=_sortkey_length, size=50),
-    DefaultColumn(N_("Artist"), 'artist', size=200),
+    DefaultColumn(N_("Title"), 'title', sort_type=ColumnSortType.NAT, width=250, always_visible=True, status_icon=True),
+    DefaultColumn(N_("Length"), '~length', align=ColumnAlign.RIGHT, sort_type=ColumnSortType.SORTKEY, sortkey=_sortkey_length, width=50),
+    DefaultColumn(N_("Artist"), 'artist', width=200),
     Column(N_("Album Artist"), 'albumartist'),
     Column(N_("Composer"), 'composer'),
     Column(N_("Album"), 'album', sort_type=ColumnSortType.NAT),
@@ -94,4 +94,4 @@ ITEMVIEW_COLUMNS = Columns((
     Column(N_("Release Date"), 'releasedate'),
     Column(N_("Cover"), 'covercount'),
     Column(N_("Cover Dimensions"), 'coverdimensions')
-))
+), default_width=100)
--- a/picard/ui/searchdialog/album.py
+++ b/picard/ui/searchdialog/album.py
@@ -152,10 +152,10 @@ class AlbumSearchDialog(SearchDialog):
         self.existing_album = existing_album
         self.setWindowTitle(_("Album Search Results"))
         self.columns = Columns((
-            Column(N_("Name"), 'album'),
+            Column(N_("Name"), 'album', sort_type=ColumnSortType.NAT),
             Column(N_("Artist"), 'albumartist'),
             Column(N_("Format"), 'format'),
-            Column(N_("Tracks"), 'tracks', sort_type=ColumnSortType.NAT),
+            Column(N_("Tracks"), 'tracks', sort_type=ColumnSortType.NAT, align=ColumnAlign.RIGHT),
             Column(N_("Date"), 'date'),
             Column(N_("Country"), 'country'),
             Column(N_("Labels"), 'label'),
--- a/picard/ui/searchdialog/album.py
+++ b/picard/ui/searchdialog/album.py
@@ -110,15 +110,14 @@ class CoverWidget(QtWidgets.QWidget):
 
 class CoverCell:
 
-    def __init__(self, table, release, row, column, size, on_show=None):
+    def __init__(self, release, size, on_show=None):
         self.release = release
         self.fetched = False
         self.fetch_task = None
-        self.widget = widget = CoverWidget(size, parent=table)
+        self.widget = CoverWidget(size)
         self.widget.destroyed.connect(self.invalidate)
         if on_show is not None:
-            widget.shown.connect(partial(on_show, self))
-        table.setCellWidget(row, column, widget)
+            self.widget.shown.connect(partial(on_show, self))
 
     def is_visible(self):
         if self.widget:
@@ -343,16 +342,20 @@ class AlbumSearchDialog(SearchDialog):
         self.prepare_table()
         self.cover_cells = []
         cover_pos = self.columns.pos('cover')
+        cover_size = self.columns[cover_pos].width
         for row, release in enumerate(self.search_results):
             self.table.insertRow(row)
-            for c in self.columns:
-                value = release.get(c.key, "")
-                self.set_table_item_value(row, c.key, value)
-            self.cover_cells.append(CoverCell(
-                self.table, release, row, cover_pos,
-                self.columns[cover_pos].width,
-                on_show=self.fetch_coverart
-            ))
+            for pos, c in enumerate(self.columns):
+                if isinstance(c, CoverColumn):
+                    cover_cell = CoverCell(
+                        release,
+                        cover_size,
+                        on_show=self.fetch_coverart
+                    )
+                    self.cover_cells.append(cover_cell)
+                    self.table.setCellWidget(row, pos, cover_cell.widget)
+                else:
+                    self.set_table_item_value(row, pos, c, release)
             if self.existing_album and release['musicbrainz_albumid'] == self.existing_album.id:
                 self.highlight_row(row)
         self.show_table(sort_column='score')
--- a/picard/ui/searchdialog/album.py
+++ b/picard/ui/searchdialog/album.py
@@ -33,10 +33,7 @@ from PyQt6.QtCore import pyqtSignal
 from picard import log
 from picard.config import get_config
 from picard.const import CAA_URL
-from picard.i18n import (
-    N_,
-    gettext as _,
-)
+from picard.i18n import N_
 from picard.mbjson import (
     countries_from_node,
     media_formats_from_node,
@@ -147,14 +144,6 @@ class AlbumSearchDialog(SearchDialog):
     dialog_header_state = 'albumsearchdialog_header_state'
 
     def __init__(self, parent, force_advanced_search=None, existing_album=None):
-        super().__init__(
-            parent,
-            accept_button_title=_("Load into Picard"),
-            search_type='album',
-            force_advanced_search=force_advanced_search)
-        self.cluster = None
-        self.existing_album = existing_album
-        self.setWindowTitle(_("Album Search Results"))
         self.columns = Columns((
             Column(N_("Name"), 'album', sort_type=ColumnSortType.NAT, width=150),
             Column(N_("Artist"), 'albumartist'),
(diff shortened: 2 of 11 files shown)


picard/ui/tablebaseddialog.py
```diff
@@ -113,27 +132,15 @@ def __init__(self, parent):
...
-
-    def set_table_item_val(self, row, colname, value, sortkey=None):
-        # TODO: use Column.sortkey & align
-        # QVariant remembers the original type of the data
-        # matching comparison operator will be used when sorting
-        # get() will return a string, force conversion if asked to
-        if sortkey is None:
-            sortkey = sort_key(value, numeric=True)
-        item = SortableTableWidgetItem(sortkey)
-        item.setData(QtCore.Qt.ItemDataRole.DisplayRole, value)
```
zas, line 132: @phw not sure if there was a special reason to use `setData(DisplayRole)` here over `setText()` but I did the change in this PR.

picard/ui/tablebaseddialog.py
```diff
@@ -91,12 +95,27 @@ def resizeEvent(self, event):
...
 class SortableTableWidgetItem(QtWidgets.QTableWidgetItem):
 
-    def __init__(self, sort_key):
+    def __init__(self, obj=None):
         super().__init__()
-        self.sort_key = sort_key
+        self._obj = obj
+
+    def setText(self, text):
+        return super().setText(text)
```
phw, line 103: What's the point of overwriting `setText` without any change here?

picard/ui/tablebaseddialog.py
phw: Using `setText` should be fine and should make no difference. `text()` and `setText()` are also just accessing the data with `DisplayRole`.

picard/ui/tablebaseddialog.py
zas: Just a leftover from previous experiment, removed.

zas:
@phw @rdswift  I'm not totally happy yet with this, but I think it is going in the right direction.  I made a lot of small changes and it would be really useful if you could test in various cases (changes are mainly to Search Dialog album/artist/track but some changes were also made in common code with itemviews, mainly Columns/Column).

rdswift:
Everything I've tested here seems to work fine, but I did notice something that looks a bit odd (although I don't think it has anything to do with your changes).  In the Cluster and Album panes, if you click on a column title it switches between ascending, descending and unselected for sorting.  In the Search dialog it only switches between ascending and descending.  It's not a big deal, but I thought I would mention it in case it's something that's easy to make consistent.

The other thing I noticed was that in the Search dialog, the album cover download was sporadic and inconsistent, although I suspect that has more to do with the Internet Archive than with Picard.  The logs showed a lot of 200 (CACHED) responses but the covers sometimes displayed and other times not.  I was using an album search of "2112" for my testing.

zas:
> Everything I've tested here seems to work fine, but I did notice something that looks a bit odd (although I don't think it has anything to do with your changes). In the Cluster and Album panes, if you click on a column title it switches between ascending, descending and unselected for sorting. In the Search dialog it only switches between ascending and descending. It's not a big deal, but I thought I would mention it in case it's something that's easy to make consistent.
> 

That's because the item view work a bit differently for now, as it has the tristate sorting feature (https://github.com/metabrainz/picard/blob/ab0cc4be41622bd10651e629e2730a36a0b0f101/picard/ui/itemviews/basetreeview.py#L101) it also has a mean to select shown columns via context menu I would like to have on search tables too.

> The other thing I noticed was that in the Search dialog, the album cover download was sporadic and inconsistent, although I suspect that has more to do with the Internet Archive than with Picard. The logs showed a lot of 200 (CACHED) responses but the covers sometimes displayed and other times not. I was using an album search of "2112" for my testing

I broke the code with last 2 commits (I know it is broken, but not how to fix it yet), this patch is not ready yet.

Thanks for testing.

rdswift:
> I broke the code with last 2 commits (I know it is broken, but not how to fix it yet), this patch is not ready yet.

Actually, I think it is the same as the current `master` branch so I don't think your commits are the cause.

zas:
@phw @rdswift: preserving column width should now work (it doesn't work in master), but not sort column (the current code forces the sort column to 'score' after each result)

Please give it a run, and tell me if behavior looks ok to you.

rdswift:
> @phw @rdswift: preserving column width should now work (it doesn't work in master), but not sort column (the current code forces the sort column to 'score' after each result)
> 
> Please give it a run, and tell me if behavior looks ok to you.

The column width and sorting looks fine.  Even the cover art seems to be working better today (which makes me think that it is more IA related than code related).

I did spot something strange that I never tested before though.  Using my example test of an album search for "2112", the fresh search (from the main window of Picard) yields a list of 50 items.  From the search results window, if you click the magnifying glass icon without changing the search criteria the list shows 100 items with the initial 50 items duplicated.  Clicking the magnifying glass again shows 150 items with the initial 50 items triplicated.  It appears that the results of each successive search are being appended to the initial list rather than starting a new list.  This is different from the current release version of Picard, where each search clears the previous results.

zas:
> This is different from the current release version of Picard, where each search clears the previous results.

Yes, good catch, the result table wasn't reset anymore between searches.
Try https://github.com/metabrainz/picard/pull/2658/commits/fe93d24cddc3e220c67a32d5e6fddfd9c22fabf1 it should fix the issue (and I hope do not create another one).

picard/ui/itemviews/columns.py
```diff
@@ -68,15 +69,51 @@ def _sortkey_filesize(obj):
         return 0
 
 
+class IconColumn(ImageColumn):
```
phw, line 72: I'm not sure about the need of the `ImageColumn` base class and moving `IconColumn` back to itemviews. Isn't this rather generic and could be useful for other cases as well?

picard/ui/searchdialog/track.py
```diff
@@ -44,36 +44,42 @@
...
+
+
 class TrackSearchDialog(SearchDialog):
 
     dialog_header_state = 'tracksearchdialog_header_state'
 
     def __init__(self, parent, force_advanced_search=None):
+        self.columns = Columns((
+            Column(N_("Name"), 'title', width=150),
+            Column(N_("Length"), '~length', sort_type=ColumnSortType.SORTKEY, sortkey=_track_length_sortkey, align=ColumnAlign.RIGHT, width=50),
```
phw, line 68: Instead of `_track_length_sortkey` I think this could use `attrgetter`

rdswift:
> Try [8d3924f](https://github.com/metabrainz/picard/pull/2658/commits/8d3924fac310461d700352e61dfb351d3781bf83) it should fix the issue (and I hope does'nt create another one).

It all looks to be working as expected now.  Thanks.

picard/ui/itemviews/columns.py
zas: Not sure yet about this, IconColumn is only used in itemviews for now, and the base ImageColumn is an abstract class for stuff common to such column.

But I open to improvements in this field, navigating by sight right now.

picard/ui/searchdialog/track.py
zas: Done in https://github.com/metabrainz/picard/pull/2658/commits/e127d472da14aff6bfba3d62feb61cbd956c7be2

zas:
@rdswift I think I solved the issue you had with partial data in table, it was due to sorting (it was enabled during table creation, and it doesn't work well). Please retest with latest changes (including preserving sort column).

zas:
@phw @rdswift I'm not sure about preserving sort column, the previous behavior was to reset search column to score, but if the user selected another sort column/order it was lost between searches (and restarts), now it is preserved.

rdswift:
> @rdswift I think I solved the issue you had with partial data in table, it was due to sorting (it was enabled during table creation, and it doesn't work well). Please retest with latest changes (including preserving sort column).

Seems to work pretty well.  Thanks.

Now I hate to do this to you, but in my testing I found a couple more things that seem odd.

1. When initiating a search (again using my "2112" album search example) from the main Picard window, the website calls for cover art only happen when the line is scrolled into the view port.  When initiating the search (new criteria or unchanged) from the search results dialog, it appears that all of the cover art web queries are queued.

2. When the search dialog is displayed, you cannot click in the search criteria text box to edit the search text.  You must double-click to select the whole text and then you can move the cursor to make your edits.  The same thing happens if you leave the text box by clicking on a row or the advanced search option.

Neither of these are critical, but they just seemed odd to me.  Having to double clisk the text box to edit the text really caught me by surprise.  By the way, these have nothing to do with the changes that you've made.  The behavior is the same in the current release version of Picard (both the Windows and Linux versions).

The fact that I've never noticed these items before is primarily due to my normal workflow, which is to manually look up the release on the website and then load it into Picard from the website integration "tagger" link.  I rarely use the built in search.

rdswift:
> but if the user selected another sort column/order it was lost between searches (and restarts), now it is preserved.

I think the way you have it now works well, although I'm not sure that it makes sense to have the Cover Art column sortable.  What is it using for the sort criteria?  File name?  I couldn't really tell if selecting that column for sorting had any impact, other than to display the up/down indicators in the column header.  I don't know if it can be disabled for just the one column, but if so that would be my suggestion.

zas:
> I think the way you have it now works well, although I'm not sure that it makes sense to have the Cover Art column sortable. What is it using for the sort criteria? File name? I couldn't really tell if selecting that column for sorting had any impact, other than to display the up/down indicators in the column header. I don't know if it can be disabled for just the one column, but if so that would be my suggestion.

There's no easy way to disable indicator per column, it does nothing because that's not a sortable object.
See https://bugreports.qt.io/browse/QTBUG-2796 (not sure why all duplicates are closed though).
I'll see if there's a mean to improve that.

zas:
> * When initiating a search (again using my "2112" album search example) from the main Picard window, the website calls for cover art only happen when the line is scrolled into the view port.  When initiating the search (new criteria or unchanged) from the search results dialog, it appears that all of the cover art web queries are queued.
> 
> * When the search dialog is displayed, you cannot click in the search criteria text box to edit the search text.  You must double-click to select the whole text and then you can move the cursor to make your edits.  The same thing happens if you leave the text box by clicking on a row or the advanced search option.

Weird, I can't reproduce, perhaps platform-specific.

rdswift:
> Weird, I can't reproduce, perhaps platform-specific.

That is weird, because it is 100% reproducible here on two different computers, one running Windows 10 and the other running Linux Mint 21.3 (Virginia) with Ubuntu "jammy".  Moreover, it is 100% reproducible running from source from the `metabrainz/master` branch or from the `zas/columns2` branch, or from the installed versions of the v2.13.3 packages.

Steps to reproduce:

1. Start Picard.

2. Enter "2112" in the search textbox and select an "**Album**" search.  Press `Enter` or click the magnifying glass icon to initiate the search.  The "**Album Search Results**" dialog opens.  Note that the status bar on the main Picard window shows no pending downloads and there is no network activity.

3. While watching the Picard main window status bar, move the bottom scroll bar in the search results dialog to the right to bring the "**Cover**" column into the view port.  Note that as soon as the column is visible, downloads are queued in the status bar and begin executing.  Once they have completed and the cover images have been downloaded and appear in the column, scroll the search results down and note that new downloads are queued and begin executing as new rows are visible in the view port.  This demonstrates my first finding.

4. Now position your cursor over the search textbox in the search results dialog.  Note that the pointer changes to an I-beam.  Click the left mouse button once.  Note that the textbox is **not** shown as selected as the current widget, and there is no cursor shown in the textbox.  However, the textbox **is** the current widget and there **is** a cursor (although not visible) because `Backspace` will delete characters, text can be selected using `Shift` with the cursor keys, and text can be entered from the keyboard.

5. The cursor is not visible until some text has been selected, either by holding the left button down and dragging the mouse across the text, using `Shift` with the cursor keys, or double-clicking the textbox.  Once text has been selected, the cursor appears when the text is deselected either by single-clicking anywhere in the textbox or by using the cursor keys (without `Shift`) to reposition the cursor.

If you're not seeing this behavior on your system, then there there must be something different in the platforms that is causing the different behavior (as you mentioned earlier).  As I said earlier, this isn't critical but it is unusual.

picard/ui/tablebaseddialog.py
```diff
@@ -175,20 +179,62 @@ def highlight_row(self, row):
...
+                    pos = self.columns.pos(sort_column)
+                # It will trigger a call to sort_indicator_changed()
+                # This will set _sort_column_index and _sort_order to non-None values
+                header.setSortIndicator(pos, sort_order)
+        else:
+            # no indicator
+            # https://doc.qt.io/qt-6/qheaderview.html#setSortIndicator
+            header.setSortIndicator(-1, sort_order)
+
+        # Enabling sorting will sort using current sort colum & order
```
rdswift, line 235: Missing the 'n' in 'column'.

zas: Fixed in [a182160](https://github.com/metabrainz/picard/pull/2658/commits/a1821603d299813e231028d316ffd6f780b376a7)

zas:
> If you're not seeing this behavior on your system, then there there must be something different in the platforms that is causing the different behavior (as you mentioned earlier). 

I do actually, and I confirm those are unexpected behaviors.
I'll debug this, and try to provide fixes.

zas:
@rdswift about the second search and fetching all cover art it is due to https://github.com/metabrainz/picard/blob/ab0cc4be41622bd10651e629e2730a36a0b0f101/picard/ui/searchdialog/album.py#L122 which is working as expected just after table widget creation but not anymore after. 
Basically the region is empty first for all cells that aren't visible, but after the next search, I think because of size hints this is never empty, even if the cell isn't visible in the displayed part.
Not sure how to fix it though, this approach isn't reliable enough.

zas:
@rdswift can you create 2 tickets for those issues?

I think I found a fix that works for problem 1 -> please test https://github.com/metabrainz/picard/pull/2658/commits/ec16a82223071c87ec0280e97256ac5632e1ebc5

zas:
> > @rdswift can you create 2 tickets for those issues?
> 
> Created [PICARD-3073](https://tickets.metabrainz.org/browse/PICARD-3073) and [PICARD-3074](https://tickets.metabrainz.org/browse/PICARD-3074) for the two issues.

Thanks.


Actually PICARD-3073 describes the normal & expected behavior (cover art is only downloaded if the cell is displayed, that's to prevent excessive number of queries).
The buggy part of it is " When initiating the search (new criteria or unchanged) from the search results dialog, it appears that **all of the cover art web queries are queued.**"
Also the title isn't reflecting the issue.

rdswift:
I just tested both fixes, and confirm that they work consistently and as expected.  Well done!

From a personal preference perspective, I might have set the cover image downloading to queue all requests immediately so that they will work in the background and the image is more likely to be available when I scroll to a line rather than waiting for it after scrolling.  There are advantages and disadvantages with either approach, so the current method is fine.  The main thing is that the behavior is consistent -- which it is now.

rdswift:
> Actually [PICARD-3073](https://tickets.metabrainz.org/browse/PICARD-3073) describes the normal & expected behavior (cover art is only downloaded if the cell is displayed, that's to prevent excessive number of queries).
> The buggy part of it is " When initiating the search (new criteria or unchanged) from the search results dialog, it appears that **all of the cover art web queries are queued.**"
> Also the title isn't reflecting the issue.

Sorry about that.  Updated now.  Please let me know if I've still missed the mark.

picard/ui/searchdialog/album.py
```diff
@@ -135,36 +139,43 @@ def invalidate(self):
         if self.widget:
             self.widget = None
 
+    def __repr__(self):
+        return "CoverCell(%r, %r, %r, %r, %r, on_show=%r)" % (self.table, self.row, self.column, self.mbid, self.size, self.on_show)
```
rdswift, line 143: As you have this currently it's fine (perhaps even better), but I probably would have done this as an f-string as:

```python
return f"CoverCell({self.table!r}, {self.row!r}, {self.column!r}, {self.mbid!r}, {self.size!r}, on_show={self.on_show!r})"
```

I would have probably used the f-string because I try to use them by default unless there is a reason not to (such as in log entries where the resulting string might not be used).  Is there any advantage or disadvantage to using one or the other format in this case, so that I can make better usage decisions in the future?

rdswift (dismissed):
Nice improvement.  Thanks.

picard/ui/searchdialog/album.py
rdswift: Maybe I just answered my own question after reviewing https://peps.python.org/pep-0498/#s-r-and-a-are-redundant.

zas: I went for a slightly different approach in https://github.com/metabrainz/picard/pull/2658/commits/6a1f4f5daf5a6db78b13af8357690dd43898e71e

zas:
I rebased, I think I'll take last 2 bug fixes out of this PR and keep them as separated PRs (based on this one though)

rdswift (approved):
Looks good.  Thanks.

phw (approved):
LGTM

PR #2668, opened 2025-06-03
PICARD-150: Support for internal search/filter of the main two panes

jpmsousa03:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
 Added filtering and search functionality for the main panels.

# Problem

Picard lacked filtering and search functionality in its main panels, making it difficult for users to manage large file sets efficiently.
This commit introduces a find bar that supports filtering by tags, filenames, and file paths, allowing users to quickly narrow down displayed results.

The new feature enhances navigation and organization during the tagging process, significantly improving usability for users working with large music libraries.

* JIRA ticket (_optional_): PICARD-150

# Solution

We’ve added new find bars to the itemviews UI, which can be activated using the Ctrl+F shortcut (previously mapped to the search functionality). When the user types into the find bar, items in the basetree view are searched recursively, and only matching results are displayed.
Additionally, a filter feature allows the search query to be restricted to specific fields such as tags, filename, or filepath. The UI for the filter dialog is implemented in the find.py file.
We also added tests to this new functionality (test_ui_find.py)
# Action

Additional actions required:
* [x] Other
The current implementation has been tested in the following environments:
- macOS (M3, ARM64) using a Python virtual environment managed with Conda 24.11.3
- Windows 11 Version 23H2 (64-bit) — KB5058405
- WSL2 Version 2.4.11 running Ubuntu 22.04.5 LTS

Testing has been more extensive with .mp3 files; additional testing would be appreciated.

We are also uncertain if the current code structure is optimal — while it functions correctly, some parts might be better organized or placed elsewhere for clarity and maintainability. Please let us know if you spot anything that seems out of place.

The code includes comments intended to improve understanding of the new additions, but these can be removed prior to merging.

### DIFF
--- a/picard/ui/find.py
+++ b/picard/ui/find.py
@@ -1,0 +1,126 @@
+from PyQt6 import (
+    QtCore,
+    QtWidgets,
+)
+
+from picard.util.tags import tag_names
+
+
+class FindBox(QtWidgets.QWidget):
+    findChanged = QtCore.pyqtSignal(str, list)
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        layout = QtWidgets.QHBoxLayout(self)
+        layout.setContentsMargins(2, 2, 2, 2)
+
+        # find icon
+        self.find_icon = QtWidgets.QLabel()
+        find_icon = self.style().standardIcon(QtWidgets.QStyle.StandardPixmap.SP_FileDialogContentsView)
+        self.find_icon.setPixmap(find_icon.pixmap(16, 16))
+        layout.addWidget(self.find_icon)
+
+        self.filter_button = QtWidgets.QPushButton("Filters", self)
+        self.filter_button.setMaximumWidth(120)
+        self.filter_button.clicked.connect(self._show_filter_dialog)
+        layout.addWidget(self.filter_button)
+
+        self.valid_tags = set(tag_names())
+        self.selected_filters = []  # Start with "All" selected
+
+        # find input
+        self.find_query_box = QtWidgets.QLineEdit(self)
+        self.find_query_box.setPlaceholderText("Find")
+        self.find_query_box.setClearButtonEnabled(True)
+        self.find_query_box.textChanged.connect(self._query_changed)
+        layout.addWidget(self.find_query_box)
+
+    def _show_filter_dialog(self):
+        """Show dialog to select multiple filters"""
+        dialog = QtWidgets.QDialog(self)
+        dialog.setWindowTitle("Select Filters")
+        dialog.setMinimumWidth(300)
+
+        layout = QtWidgets.QVBoxLayout(dialog)
+
+        # Scroll area for tags
+        scroll = QtWidgets.QScrollArea(dialog)
+        scroll.setWidgetResizable(True)
+        scroll_content = QtWidgets.QWidget(scroll)
+        scroll_layout = QtWidgets.QVBoxLayout(scroll_content)
+
+        label = QtWidgets.QLabel("File Filters", scroll_content)
+        scroll_layout.addWidget(label)
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        checkboxes = {}
+        for file_filter in ["filename", "filepath"]:
+            checkbox = QtWidgets.QCheckBox(file_filter, scroll_content)
+            checkbox.setChecked(file_filter in self.selected_filters)
+            scroll_layout.addWidget(checkbox)
+            checkboxes[file_filter] = checkbox
+
+        scroll_layout.addItem(QtWidgets.QSpacerItem(0, 10, QtWidgets.QSizePolicy.Policy.Minimum, QtWidgets.QSizePolicy.Policy.Fixed))
+
+        label = QtWidgets.QLabel("Tag Filters", scroll_content)
+        scroll_layout.addWidget(label)
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        # Add checkboxes for all tags
+        for tag in sorted(self.valid_tags):
+            checkbox = QtWidgets.QCheckBox(tag, scroll_content)
+            checkbox.setChecked(tag in self.selected_filters)
+            scroll_layout.addWidget(checkbox)
+            checkboxes[tag] = checkbox
+
+        scroll_content.setLayout(scroll_layout)
+        scroll.setWidget(scroll_content)
+        layout.addWidget(scroll)
+
+        # Buttons
+        button_box = QtWidgets.QDialogButtonBox(
+            QtWidgets.QDialogButtonBox.StandardButton.Ok | QtWidgets.QDialogButtonBox.StandardButton.Cancel)
+        button_box.accepted.connect(dialog.accept)
+        button_box.rejected.connect(dialog.reject)
+        layout.addWidget(button_box)
+
+        # Show dialog and process result
+        if dialog.exec() == QtWidgets.QDialog.DialogCode.Accepted:
+            self.selected_filters = []
+            for tag, checkbox in checkboxes.items():
+                if checkbox.isChecked():
+                    self.selected_filters.append(tag)
+
+            # Update button text
+            if not self.selected_filters or self.selected_filters == []:
+                self.selected_filters = []
+                self.filter_button.setText("Filters")
+
+            elif len(self.selected_filters) <= 2:
+                self.filter_button.setText(", ".join(self.selected_filters))
+            else:
+                self.filter_button.setText(f"{len(self.selected_filters)} filters")
+
+            # Update find with new filters
+            self._query_changed(self.find_query_box.text())
+
+    def _query_changed(self, text):
+        self.findChanged.emit(text, self.selected_filters)
+
+    def clear(self):
+        self.find_query_box.clear()
+        self.selected_filters = []
+        self.filter_button.setText("Filters")
+
+    def set_focus(self):
+        self.find_query_box.setFocus()
--- a/picard/ui/itemviews/__init__.py
+++ b/picard/ui/itemviews/__init__.py
@@ -99,6 +99,23 @@ class MainPanel(QtWidgets.QSplitter):
         self._ignore_selection_changes = False
         self._sort_enabled = None  # None at start, bool once set_sorting is called
 
+        # Create a layout for each view to include the find box
+        for view in self._views:
+            container = QtWidgets.QWidget(self)
+            layout = QtWidgets.QVBoxLayout(container)
+            layout.setContentsMargins(0, 0, 0, 0)
+            layout.setSpacing(0)
+
+            # Create and add find box
+            find_box = view.setup_find_box()
+            layout.addWidget(find_box)
+
+            # Add view
+            layout.addWidget(view)
+
+            # Replace the view in the splitter with the container
+            self.replaceWidget(self.indexOf(view), container)
+
         def _view_update_selection(view):
             if not self._ignore_selection_changes:
                 self._ignore_selection_changes = True
@@ -226,6 +243,16 @@ class MainPanel(QtWidgets.QSplitter):
                 view.setCurrentItem(item)
                 self._update_selection(view)
                 break
+
+    def toggle_find_boxes(self):
+        """Toggle visibility of find boxes in both views."""
+        visible = not self._views[0].find_box.isVisible()
+        for view in self._views:
+            view.find_box.setVisible(visible)
+            if visible:
+                view.find_box.set_focus()
+            else:
+                view.find_box.clear()
 
 
 class FileTreeView(BaseTreeView):
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -89,6 +89,7 @@ from picard.util import (
 
 from picard.ui.collectionmenu import CollectionMenu
 from picard.ui.enums import MainAction
+from picard.ui.find import FindBox
 from picard.ui.itemviews.columns import (
     DEFAULT_COLUMNS,
     ITEM_ICON_COLUMN,
@@ -650,3 +651,82 @@ class BaseTreeView(QtWidgets.QTreeWidget):
     @property
     def default_drop_target(self):
         return None
+
+    def setup_find_box(self):
+        self.find_box = FindBox(self)
+        self.find_box.findChanged.connect(self.filter_items)
+
+        self.find_box.hide()  # Hide the find box initially
+
+        return self.find_box
+
+    def filter_items(self, text, filters):
+        if not text:  # When text is empty, show all items
+            self._restore_all_items()
+            return
+
+        text = text.lower()
+        self._filter_tree_items(self.invisibleRootItem(), text, filters)
+
+    def _filter_tree_items(self, parent, text, filters):
+        """Recursively filter tree items based on find text."""
+
+        match_found = False
+
+        for i in range(parent.childCount()):
+            child = parent.child(i)
+            child_match = False
+
+            # Special handling for different object types
+            if hasattr(child, 'obj'):
+                obj = child.obj
+
+                # Handle Clusters
+                if filters == [] or "filename" in filters:
+                    # Handle Tracks with files
+                    if hasattr(obj, 'iterfiles'):
+                        for file_ in obj.iterfiles():
+                            if filters == [] or "filename" in filters:
+                                if text in file_.base_filename.lower():
+                                    child_match = True
+                                    break
+                            if filters == [] or "filepath" in filters:
+                                if text in file_.filename.lower():
+                                    child_match = True
+                                    break
+
+                # Handle metadata in Files
+                if hasattr(obj, 'metadata'):
+                    for tag, values in obj.metadata.rawitems():
+                        if isinstance(values, list):
+                            for value in values:
+                                if filters == [] or tag.lower() in filters:
+                                    if text in str(value).lower():
+                                        child_match = True
+                                        break
+                        elif text in str(values).lower():
+                            child_match = True
+                            break
+
+            # Recursively check children
+            if child.childCount() > 0:
+                child_matches = self._filter_tree_items(child, text, filters)
+                child_match |= child_matches
+
+            # Hide/show based on match
+            child.setHidden(not child_match)
+            match_found |= child_match
+
+        return match_found
+
+    def _restore_all_items(self):
+        """Show all items in the tree."""
+        self._restore_tree_items(self.invisibleRootItem())
+
+    def _restore_tree_items(self, parent):
+        """Recursively show all items."""
+        for i in range(parent.childCount()):
+            child = parent.child(i)
+            child.setHidden(False)
+            if child.childCount() > 0:
+                self._restore_tree_items(child)
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -288,7 +288,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             else:
                 self.remove_selected_objects()
         elif event.matches(QtGui.QKeySequence.StandardKey.Find):
-            self.search_edit.setFocus(QtCore.Qt.FocusReason.ShortcutFocusReason)
+            self.toggle_find_boxes()
         else:
             super().keyPressEvent(event)
 
@@ -1631,6 +1631,10 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
     def show_plugins_options_page(self):
         self.show_options(page='plugins')
+
+    def toggle_find_boxes(self):
+        """Toggle the find boxes in the main panel."""
+        self.panel.toggle_find_boxes()
 
 
 def update_last_check_date(is_success):
--- a/test/test_ui_find.py
+++ b/test/test_ui_find.py
@@ -1,0 +1,202 @@
+from test.picardtestcase import PicardTestCase
+
+from picard.metadata import Metadata
+from picard.util.tags import tag_names
+
+from picard.ui.find import FindBox
+from picard.ui.itemviews import MainPanel
+from picard.ui.itemviews.basetreeview import BaseTreeView
+
+
+class FindBoxTest(PicardTestCase):
+    """Test the FindBox widget functionality"""
+
+    def test_tag_names_available(self):
+        """Test that tag_names() returns valid tags for filtering"""
+        tags = set(tag_names())
+        self.assertIsInstance(tags, set)
+        self.assertGreater(len(tags), 0)
+        self.assertIn('artist', tags)
+        self.assertIn('album', tags)
+        self.assertIn('title', tags)
+
+    def test_findbox_class_exists(self):
+        """Test that FindBox class can be imported and has required attributes"""
+        self.assertTrue(hasattr(FindBox, 'findChanged'))
+        self.assertTrue(hasattr(FindBox, '_query_changed'))
+        self.assertTrue(hasattr(FindBox, 'clear'))
+        self.assertTrue(hasattr(FindBox, 'set_focus'))
+
+    def test_filter_button_text_logic(self):
+        """Test the logic for updating filter button text from FindBox._show_filter_dialog"""
+        test_cases = [
+            ([], "Filters"),
+            (["filename"], "filename"),
+            (["filename", "artist"], "filename, artist"),
+            (["filename", "artist", "album", "title"], "4 filters"),
+        ]
+
+        for selected_filters, expected_text in test_cases:
+            if not selected_filters or selected_filters == []:
+                button_text = "Filters"
+            elif len(selected_filters) <= 2:
+                button_text = ", ".join(selected_filters)
+            else:
+                button_text = f"{len(selected_filters)} filters"
+
+            self.assertEqual(button_text, expected_text,
+                           f"Filter list {selected_filters} should produce text '{expected_text}'")
+
+    def test_filename_filtering_patterns(self):
+        """Test filename filtering patterns used in the implementation"""
+        filename = "Test_Song.mp3"
+        base_filename = filename.lower().rsplit('.', 1)[0]
+
+        self.assertTrue("test" in base_filename)
+        self.assertTrue("song" in base_filename)
+        self.assertFalse("album" in base_filename)
+
+    def test_filepath_filtering_patterns(self):
+        """Test filepath filtering patterns used in the implementation"""
+        filepath = "/home/user/Music/Artist/Album/01-Song.mp3"
+        filepath_lower = filepath.lower()
+
+        self.assertTrue("music" in filepath_lower)
+        self.assertTrue("artist" in filepath_lower)
+        self.assertTrue("album" in filepath_lower)
+        self.assertTrue("song" in filepath_lower)
+
+
+class BaseTreeViewFilteringTest(PicardTestCase):
+    """Test filtering functionality in BaseTreeView"""
+
+    def test_filter_methods_exist(self):
+        """Test that BaseTreeView has the required filtering methods"""
+        self.assertTrue(hasattr(BaseTreeView, 'setup_find_box'))
+        self.assertTrue(hasattr(BaseTreeView, 'filter_items'))
+        self.assertTrue(hasattr(BaseTreeView, '_filter_tree_items'))
+        self.assertTrue(hasattr(BaseTreeView, '_restore_all_items'))
+
+    def test_metadata_filtering_logic(self):
+        """Test metadata filtering logic used in BaseTreeView._filter_tree_items"""
+        metadata = Metadata()
+        metadata['artist'] = 'The Beatles'
+        metadata['album'] = 'Abbey Road'
+        metadata['title'] = 'Come Together'
+
+        text = "beatles"
+        filters = ["artist"]
+
+        child_match = False
+        for tag, values in metadata.rawitems():
+            if isinstance(values, list):
+                for value in values:
+                    if filters == [] or tag.lower() in filters:
+                        if text in str(value).lower():
+                            child_match = True
+                            break
+            elif filters == [] or tag.lower() in filters:
+                if text in str(values).lower():
+                    child_match = True
+                    break
+
+        self.assertTrue(child_match)
+
+        child_match = False
+        filters = ["album"]
+        for tag, values in metadata.rawitems():
+            if isinstance(values, list):
+                for value in values:
+                    if filters == [] or tag.lower() in filters:
+                        if text in str(value).lower():
+                            child_match = True
+                            break
+            elif filters == [] or tag.lower() in filters:
+                if text in str(values).lower():
+                    child_match = True
+                    break
+
+        self.assertFalse(child_match)
+
+    def test_empty_filter_searches_all(self):
+        """Test that empty filter list searches all fields"""
+        metadata = Metadata()
+        metadata['artist'] = 'Test Artist'
+        metadata['album'] = 'Different Album'
+
+        text = "test"
+        filters = []
+
+        metadata_match = False
+        for tag, values in metadata.rawitems():
+            if isinstance(values, list):
+                for value in values:
+                    if filters == [] or tag.lower() in filters:
+                        if text in str(value).lower():
+                            metadata_match = True
+                            break
+            elif filters == [] or tag.lower() in filters:
+                if text in str(values).lower():
+                    metadata_match = True
+                    break
+
+        self.assertTrue(metadata_match)
+
+    def test_filter_items_text_processing(self):
+        """Test that BaseTreeView.filter_items processes text correctly"""
+        test_text = "  TeSt  "
+        processed = test_text.strip().lower()
+
+        self.assertEqual(processed, "test")
+
+    def test_filename_iteration_logic(self):
+        """Test the logic for iterating through files for filename matching"""
+        class MockFile:
+            def __init__(self, filename):
+                self.filename = filename
+                self.base_filename = filename.split('/')[-1].rsplit('.', 1)[0]
+
+        class MockCluster:
+            def __init__(self, files):
+                self.files = files
+
+            def iterfiles(self):
+                return self.files
+
+        files = [
+            MockFile("test_song.mp3"),
+            MockFile("another_track.mp3")
+        ]
+        cluster = MockCluster(files)
+
+        text = "test"
+        filters = ["filename"]
+
+        child_match = False
+        if hasattr(cluster, 'iterfiles'):
+            for file_ in cluster.iterfiles():
+                if filters == [] or "filename" in filters:
+                    if text in file_.base_filename.lower():
+                        child_match = True
+                        break
+
+        self.assertTrue(child_match)
+
+
+class MainPanelFindTest(PicardTestCase):
+    """Test find functionality integration in MainPanel"""
+
+    def test_main_panel_has_toggle_method(self):
+        """Test that MainPanel has the toggle_find_boxes method"""
+        self.assertTrue(hasattr(MainPanel, 'toggle_find_boxes'))
+
+    def test_toggle_logic(self):
+        """Test the core toggle logic used in MainPanel.toggle_find_boxes"""
+        find_box_visible = False
+
+        new_visible_state = not find_box_visible
+        self.assertTrue(new_visible_state)
+
+        find_box_visible = True
+        new_visible_state = not find_box_visible
+        self.assertFalse(new_visible_state)


picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+        layout = QtWidgets.QHBoxLayout(self)
+        layout.setContentsMargins(2, 2, 2, 2)
+
+        # find icon
+        self.find_icon = QtWidgets.QLabel()
+        find_icon = self.style().standardIcon(QtWidgets.QStyle.StandardPixmap.SP_FileDialogContentsView)
+        self.find_icon.setPixmap(find_icon.pixmap(16, 16))
+        layout.addWidget(self.find_icon)
+
+        self.filter_button = QtWidgets.QPushButton("Filters", self)
```
zas: Strings should be translatable, add:

```python
from picard.i18n import (
    N_,
    gettext as _,
)
```

```suggestion
        self.filter_button = QtWidgets.QPushButton(_("Filters"), self)
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
+from PyQt6 import (
```
zas: Add proper header like in other files:

```python
# -*- coding: utf-8 -*-
#
# Picard, the next-generation MusicBrainz tagger
#
# Copyright (C) ....
....
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+        self.filter_button.setMaximumWidth(120)
+        self.filter_button.clicked.connect(self._show_filter_dialog)
+        layout.addWidget(self.filter_button)
+
+        self.valid_tags = set(tag_names())
+        self.selected_filters = []  # Start with "All" selected
+
+        # find input
+        self.find_query_box = QtWidgets.QLineEdit(self)
+        self.find_query_box.setPlaceholderText("Find")
```
zas:
```suggestion
        self.find_query_box.setPlaceholderText(_("Find"))
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+        self.find_query_box.setClearButtonEnabled(True)
+        self.find_query_box.textChanged.connect(self._query_changed)
+        layout.addWidget(self.find_query_box)
+
+    def _show_filter_dialog(self):
+        """Show dialog to select multiple filters"""
+        dialog = QtWidgets.QDialog(self)
+        dialog.setWindowTitle("Select Filters")
```
zas:
```suggestion
        dialog.setWindowTitle(_("Select Filters"))
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+
+        layout = QtWidgets.QVBoxLayout(dialog)
+
+        # Scroll area for tags
+        scroll = QtWidgets.QScrollArea(dialog)
+        scroll.setWidgetResizable(True)
+        scroll_content = QtWidgets.QWidget(scroll)
+        scroll_layout = QtWidgets.QVBoxLayout(scroll_content)
+
+        label = QtWidgets.QLabel("File Filters", scroll_content)
```
zas:
```suggestion
        label = QtWidgets.QLabel(_("File Filters"), scroll_content)
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        checkboxes = {}
+        for file_filter in ["filename", "filepath"]:
+            checkbox = QtWidgets.QCheckBox(file_filter, scroll_content)
```
zas: Label and key should be different:
```python
    file_filters = {
        'filename': N_("Filename"),
        'filepath': N_("Filepath"),
    }
    checkboxes = {}
        for file_filter in ["filename", "filepath"]:
            checkbox = QtWidgets.QCheckBox(_(file_filters[file_filter]), scroll_content)
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+            checkbox.setChecked(file_filter in self.selected_filters)
+            scroll_layout.addWidget(checkbox)
+            checkboxes[file_filter] = checkbox
+
+        scroll_layout.addItem(QtWidgets.QSpacerItem(0, 10, QtWidgets.QSizePolicy.Policy.Minimum, QtWidgets.QSizePolicy.Policy.Fixed))
+
+        label = QtWidgets.QLabel("Tag Filters", scroll_content)
```
zas:
```suggestion
        label = QtWidgets.QLabel(_("Tag Filters"), scroll_content)
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+        scroll_layout.addWidget(label)
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        checkboxes = {}
+        for file_filter in ["filename", "filepath"]:
```
zas: Perhaps we should improve that, adding a property to https://github.com/metabrainz/picard/blob/176762cf21ca80a3dc8be90dcdfcc595a777c1da/picard/tags/tagvar.py#L81 so we can toggle "filterable", it could be `True` by default, and set to `False` for TagVar we don't want to appear in filters.

Also make use of `ALL_VARS` and `TagVars.display_name()`.

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+        if dialog.exec() == QtWidgets.QDialog.DialogCode.Accepted:
+            self.selected_filters = []
+            for tag, checkbox in checkboxes.items():
+                if checkbox.isChecked():
+                    self.selected_filters.append(tag)
+
+            # Update button text
+            if not self.selected_filters or self.selected_filters == []:
+                self.selected_filters = []
+                self.filter_button.setText("Filters")
```
zas:
```suggestion
                self.filter_button.setText(_("Filters"))
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+
+            elif len(self.selected_filters) <= 2:
+                self.filter_button.setText(", ".join(self.selected_filters))
```
zas: This should be translated.

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+            else:
+                self.filter_button.setText(f"{len(self.selected_filters)} filters")
```
zas: For i18n, it should be something like:

```python
_("{num} filters").format(num=len(self.selected_filters))
```

picard/ui/find.py
```diff
@@ -0,0 +1,126 @@
...
+            # Update find with new filters
+            self._query_changed(self.find_query_box.text())
+
+    def _query_changed(self, text):
+        self.findChanged.emit(text, self.selected_filters)
+
+    def clear(self):
+        self.find_query_box.clear()
+        self.selected_filters = []
+        self.filter_button.setText("Filters")
```
zas:
```suggestion
        self.filter_button.setText(_("Filters"))
```

test/test_ui_find.py
```diff
@@ -0,0 +1,202 @@
+from test.picardtestcase import PicardTestCase
```
zas: Add header

zas:
Please rebase on current `master` branch.

picard/ui/mainwindow/__init__.py
```diff
@@ -288,7 +288,7 @@ def keyPressEvent(self, event):
             else:
                 self.remove_selected_objects()
         elif event.matches(QtGui.QKeySequence.StandardKey.Find):
-            self.search_edit.setFocus(QtCore.Qt.FocusReason.ShortcutFocusReason)
+            self.toggle_find_boxes()
```
phw: I think using the search shortcut for the filters by default is useful, and especially with a version 3 release this change can be well reasoned for.

But maybe we should keep a shortcut for the search box as well. Maybe something like `Ctrl+Shift+F`. @zas what do you think?

phw (changes requested):
This is seriously so cool, thank you very much for this implementation. I like the UI. That part worried me the most about this story, and I feared that a UI would take too much vertical screen space (which already is limited). But the toggle and general narrow design work really well.

The code also looks clean and good. zas already covered the changes needed for internationalization in his comments. A few more thoughts:

1. The search UI is not very discoverable. Unless you know about the search shortcut this is a hidden feature. My suggestion would be to add this toggle also to the "View" menu as "Search filters".
2. Filtering by filepath alone doesn't work. filepath filtering seems to be only used if filename is selected as well. I think filtering by path should be possible independently.

picard/ui/mainwindow/__init__.py
zas: Yes, I agree.

jpmsousa03:
Thanks for the feedback. We'll implement the required changes.

jpmsousa03:
> I noticed a warning from Qt with this patch, that needs to be investigated:
> 
> ```
> QSplitter::replaceWidget: Index -1 out of range
> QSplitter::replaceWidget: Index -1 out of range
> ```
> 
> It doesn't happen with `master` branch.

I haven’t seen this warning myself, but @francescouteiro did in his testing, maybe a enviroment thing? Not sure. Can you check it out, @francescouteiro?

picard/ui/find.py
francescouteiro: At "Copyright (C) ..." you're expecting us to list the most recent list of Picard's maintainers? Is there one available? They seem different between files. Or should we write ourselves in? Thanks.

zas: Add your names

Sophist-UK:
Any chance of a screenshot of the new UI elements?

francescouteiro:
> Any chance of a screenshot of the new UI elements?

Here's the demo: https://youtu.be/PJ52SwYDVek?si=UiAq_IsPl7v3JD-T 😉

Sophist-UK:
Nice!

francescouteiro:
Sorry for the spam! 

This commit rebases against the latest master branch. Due to structural changes in the codebase, we've made some adjustments to maintain compatibility.

Regarding the warning:

```
QSplitter::replaceWidget: Index -1 out of range
QSplitter::replaceWidget: Index -1 out of range
```

@jpmsousa03 and I have identified the root cause and implemented a fix. We're currently running tests to ensure the solution fully resolves the issue without introducing any regressions.

picard/ui/itemviews/__init__.py
```diff
@@ -104,6 +104,23 @@ def __init__(self, window, parent=None):
...
+
+            # Create and add find box
+            find_box = view.setup_find_box()
+            layout.addWidget(find_box)
+
+            # Add view
+            layout.addWidget(view)
+
+            # Replace the view in the splitter with the container
+            self.replaceWidget(self.indexOf(view), container)
```
zas: `QSplitter::replaceWidget: Index -1 out of range` comes from this line, `self.indexOf(view) == -1`

picard/ui/find.py
```diff
@@ -0,0 +1,147 @@
...
+        scroll_layout.addWidget(label)
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        # Add checkboxes for all tags
+        for tag in sorted(self.valid_tags):
```
zas: Likely, it should be sorted by translated tag display name.

rdswift:
> 1. The search UI is not very discoverable. Unless you know about the search shortcut this is a hidden feature. My suggestion would be to add this toggle also to the "View" menu as "Search filters".

And it should be added to the documentation.  I added PICARD-3077 as a sub-task so that doesn't get forgotten.

jpmsousa03:
I’m aware that the code I committed doesn’t translate the tag names in the filter logic, as I would like to take a more carefull look at that. For now, I’ve added the Find option to the View menu, along with the Ctrl+Shift+F shortcut for Search.

picard/ui/find.py
```diff
@@ -0,0 +1,156 @@
...
+        if dialog.exec() == QtWidgets.QDialog.DialogCode.Accepted:
+            self.selected_filters = []
+            for tag, checkbox in checkboxes.items():
+                if checkbox.isChecked():
+                    self.selected_filters.append(tag)
+
+            # Update button text
+            if not self.selected_filters or self.selected_filters == []:
+                self.selected_filters = []
+                self.filter_button.setText(_("Filters"))
```
rdswift: Should this default to something like "Select Filters" (or even just "Select") instead?  I'm okay with it as you have it now, but throwing this out in case it makes the operation clearer and easier for users to understand.

picard/ui/find.py
```diff
@@ -0,0 +1,156 @@
...
+
+            elif len(self.selected_filters) <= 2:
+                self.filter_button.setText(", ".join(_(filter) for filter in self.selected_filters))
+            else:
+                self.filter_button.setText(_("{num} filters").format(num=len(self.selected_filters)))
```
rdswift: This looks fine, but if listing two filters ends up becoming too long a string (depending on which two filters are selected), perhaps it could just use the "{num} filters" if there are 2 or more filters selected.

picard/ui/find.py
rdswift: Perhaps you could call `self.clear()` to avoid some code duplication.

rdswift (commented):
Do we want the visibility state of the filter boxes to be stored in a persistent variable?  @zas @phw what do you think?

phw:
I noticed that filtering also filters the "Unclustered Files" and "Clusters" special items as well. Maybe those should stay visible, as they are part of the default UI and usually always visible.

The items are `ClusterItem` items, which hold a reference to a `Cluster` object (or subclass of that. The object has a `special` attribute that indicates their special role. E.g. it causes them to not be removable, and not be sortable. Maybe the filtering logic can exclude those too. If needed, this could also set a special flag on the `TreeItem` object instead, similar to how it is done for the `sortable` in https://github.com/metabrainz/picard/blob/master/picard/ui/itemviews/basetreeview.py#L553

phw:
> Do we want the visibility state of the filter boxes to be stored in a persistent variable? @zas @phw what do you think?

I think yes

test/test_ui_find.py
```diff
@@ -0,0 +1,223 @@
...
+    """Test the FindBox widget functionality"""
+
+    def test_tag_names_available(self):
+        """Test that tag_names() returns valid tags for filtering"""
+        tags = set(preserved_tag_names())
+        self.assertIsInstance(tags, set)
+        self.assertGreater(len(tags), 0)
+        self.assertIn('artist', tags)
+        self.assertIn('album', tags)
+        self.assertIn('title', tags)
```
rdswift: `artist`, `album` and `title` are not preserved tags, so this test will fail.  Currently the only preserved tags are: `~bitrate`, `~bits_per_sample`, `~channels`, `~dirname`, `~extension`, `~filename`, `~file_created_timestamp`, `~file_modified_timestamp`, `~format`, `~sample_rate`, and `~video`.

Did you mean to set `tags = set(tag_names())` instead?

picard/ui/find.py
jpmsousa03: The self.clear() also clears the text, in this case I believe the text should not be cleared.

jpmsousa03: Maybe there could be separate clear_text and clear_filters functions, which clear() could call, as well as the code you commented on. What do you think?

picard/ui/find.py
jpmsousa03: I agree, I will change.

picard/ui/find.py
rdswift: Sorry, I missed that it also cleared the filter text.  Your suggestion will work, but I'm not sure that it is necessary because it doesn't really reduce the code.  I'm not sure my original comment makes sense any more.

test/test_ui_find.py
```diff
@@ -0,0 +1,223 @@
...
+        self.assertTrue(hasattr(FindBox, '_query_changed'))
+        self.assertTrue(hasattr(FindBox, 'clear'))
+        self.assertTrue(hasattr(FindBox, 'set_focus'))
+
+    def test_filter_button_text_logic(self):
+        """Test the logic for updating filter button text from FindBox._show_filter_dialog"""
+        test_cases = [
+            ([], "Filters"),
+            (["filename"], "filename"),
+            (["filename", "artist"], "filename, artist"),
```
rdswift: With the recent change in https://github.com/metabrainz/picard/pull/2668/commits/95c2e94056c4286c162efe6b6878b8f4eabfdfad this should be changed to `(['filename', 'artist'], "2 filters"),`.

rdswift: Also, a really minor point but in general we use single quotes for keys and such.  See https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md#strings-quoting-single-or-double-quotes for more information.

picard/ui/find.py
```diff
@@ -0,0 +1,160 @@
...
+)
+
+from picard.i18n import (
+    N_,
+    gettext as _,
+)
+
+from picard.tags.tagvar import (
+    TagVars,
+)
```
rdswift: Do you need this import.  It doesn't appear to be used anywhere.

jpmsousa03: True. I left that in by mistake. Apologies.

jpmsousa03:
I updated find.py to use ALL_TAGS and display_names, adding support for translations as well. I tested it in my native language, and it seems to be working.

jpmsousa03:
Sorry, I got distracted and commited a change I was testing by mistake (the one causing indexOf == -1 warning). Maybe my sign to take a code break. I'll continue later, in the meantime please let me know if there is something to fix on any of the new code.

picard/ui/find.py
```diff
@@ -0,0 +1,156 @@
...
+        find_icon = self.style().standardIcon(QtWidgets.QStyle.StandardPixmap.SP_FileDialogContentsView)
+        self.find_icon.setPixmap(find_icon.pixmap(16, 16))
+        layout.addWidget(self.find_icon)
+
+        self.filter_button = QtWidgets.QPushButton(_("Filters"), self)
+        self.filter_button.setMaximumWidth(120)
+        self.filter_button.clicked.connect(self._show_filter_dialog)
+        layout.addWidget(self.filter_button)
+
+        self.valid_tags = set(ALL_TAGS)
```
rdswift: Do we want to include **all** tags in the list, or should we remove hidden tags (and possibly preserved tags)?

picard/ui/find.py
```diff
@@ -0,0 +1,156 @@
...
+        self.selected_filters = []  # Start with "All" selected
```
rdswift: Is this "All" selected, or is it "None" selected?

test/test_ui_find.py
rdswift: Actually, I'm not sure what this is actually intended to test.  It doesn't seem to have anything to do with the `FindBox` widget.

picard/ui/find.py
zas: What  I suggested is to add a property to TagVar so we can define which one is filterable or not.

Also, perhaps it would make sense to not list tags that aren't present in at least one entity.

rdswift:
@jpmsousa03,

Sorry for pushing the recent commits.  I meant to send them as a pull request to your repo.

picard/ui/find.py
rdswift: > What I suggested is to add a property to TagVar so we can define which one is filterable or not.

Property and generator added in https://github.com/metabrainz/picard/pull/2669.  Properties need to be updated to ~False~ True as appropriate for TagVar items in ALL_TAGS.

picard/ui/find.py
```diff
@@ -0,0 +1,161 @@
...
+    def __init__(self, parent=None):
+        super().__init__(parent)
+        layout = QtWidgets.QHBoxLayout(self)
+        layout.setContentsMargins(2, 2, 2, 2)
+
+        # find icon
+        self.find_icon = QtWidgets.QLabel()
+        find_icon = self.style().standardIcon(QtWidgets.QStyle.StandardPixmap.SP_FileDialogContentsView)
+        self.find_icon.setPixmap(find_icon.pixmap(16, 16))
+        layout.addWidget(self.find_icon)
```
rdswift: This is not a show-stopper by any means, but I just realized that this is a label only to indicate that filtering is available.  I'm used to seeing this as an "apply filter" button and located to the right if the filter selection criteria, although I realize that it's not required as such because you monitor changes to the selected filters and filter text and update automatically.  I'm not sure that this icon is required since clicking it has no effect (and it just confused me).

Whether or not this icon is kept, we might want to consider adding a button or icon (to the right of the search criteria) to clear the search, and link it to the `clear()` method in the widget.

**EDIT:**  Another thought...  If I understand this correctly, there will be no filtering if there are no filter tags selected, even if there is text in the filter text box.  Perhaps the current icon can be used to indicate that there is a filter active, and only made visible at that time.

picard/ui/find.py
jpmsousa03: It used to be "All" in an earlier version of the code, but since no filter just means showing everything, it is easier to treat no filter as the same as "all". The comment could be updated tho.

picard/ui/find.py
jpmsousa03: The icon was mainly there for aesthetic reasons, but it can be removed if it's causing confusion. There's already an "x" button on the text bar when text is entered (though it doesn't clear the filters).

Regarding the filtering, sorry if the code comments were confusing. There used to be an "All" option in the code to indicate that matches should be looked up across all tags, but we removed it, since we thought that's actually what should happen when no filter is active. So just to clarify: having no filter simply means it searches for matches across all tags.

jpmsousa03:
Changed filter to not include hidden tags, this could be later changed to just include the "filterable" ones.

jpmsousa03:
> 2\. Filtering by filepath alone doesn't work. filepath filtering seems to be only used if filename is selected as well. I think filtering by path should be possible independently.

This last commit addresses this as well.

picard/ui/find.py
```diff
@@ -0,0 +1,164 @@
...
+            self._query_changed(self.find_query_box.text())
+
+    @classmethod
+    def make_button_text(cls, selected_filters):
+        if not selected_filters:
+            return _("Filters")
+
+        if len(selected_filters) == 1:
+            if selected_filters[0] not in cls.file_filters.keys():
+                return _(ALL_TAGS.display_name(selected_filters[0]))
```
zas: `display_name()` takes care of translation, so no need to wrap in `_()`, see https://github.com/metabrainz/picard/blob/1ca26422a4bf41165e61ea0e47af47ba9b6d17d4/picard/tags/tagvar.py#L239

picard/ui/find.py
```diff
@@ -0,0 +1,164 @@
...
+
+        return _("{num} filters").format(num=len(selected_filters))
+
+    def _query_changed(self, text):
+        self.findChanged.emit(text, self.selected_filters)
+
+    def clear(self):
+        self.find_query_box.clear()
+        self.selected_filters = []
+        self.filter_button.setText(_("Filters"))
```
zas: Maybe store default button label in a property:

At class level, something like:

```python
    default_filter_button_label = N_("Filters")

    def set_filter_button_label(self, label=None):
        if label is None:
            label = _(default_filter_button_label)
        self.filter_button.setText(label)

    def clear(self):
        self.find_query_box.clear()
        self.selected_filters = []
        self.set_filter_button_label()
```

`make_button_text()` can then return `None` instead of `_("Filters")`

picard/ui/itemviews/basetreeview.py
```diff
@@ -567,3 +568,82 @@ def moveCursor(self, action, modifiers):
...
+
+        return self.find_box
+
+    def filter_items(self, text, filters):
+        if not text:  # When text is empty, show all items
+            self._restore_all_items()
+            return
+
+        text = text.lower()
+        self._filter_tree_items(self.invisibleRootItem(), text, filters)
```
zas, line 585: Instead of passing `text.lower()` to `_filter_tree_items()`, pass `text` and do the conversion inside  `_filter_tree_items()`.

picard/ui/mainwindow/actions.py
```diff
@@ -347,6 +347,15 @@ def _create_show_toolbar_action(parent):
     return action
 
 
+@add_action(MainAction.SHOW_FINDBAR)
+def _create_find_toolbar_action(parent):
+    action = QtGui.QAction(_("Find"), parent)
+    action.setCheckable(True)
+    action.setShortcut(QtGui.QKeySequence(_("Ctrl+F")))
```
zas: As said in https://github.com/metabrainz/picard/pull/2668#discussion_r2126031760 this shortcut conflicts with existing one (used for entity searches), so use `Ctrl+Shift+F` instead.

zas (changes requested):

picard/ui/mainwindow/actions.py
jpmsousa03: > As said in [#2668 (comment)](https://github.com/metabrainz/picard/pull/2668#discussion_r2126031760) this shortcut conflicts with existing one (used for entity searches), so use Ctrl+Shift+F instead.

I may be misunderstanding comment #2668, I thought that Ctrl+F was supposed to toggle the Find Bar, and Ctrl+Shift+F put focus on the Search bar. (Thats how I implemented) Is supposed to be the other way around or am I missing something? I appreciate If you could clarify.

zas: Not sure about that, you got it right though, my bad. I'm not sure we should change the pre-existing shortcut to something else, perhaps not that important but users that actually use `Ctrl+F` currently will have to adapt.

@phw @rdswift : so, `Ctrl+F` for Filters & `Shift+Ctrl+F` for search focus (this patch), or the reverse (keep current master shortcut) ??

picard/ui/find.py
zas: I vote to remove the icon, to maximize available space for button + edit widget.
I even wonder about the button, a context menu could do too (though harder to find perhaps),  I did a quick experiment:


```python
class CheckableMenuItem:
    def __init__(self, key, label=None, checked=False):
        self.key = key
        self.label = label or key
        self.checked = checked


class CheckableMenuLineEdit(QLineEdit):
    choicesChanged = pyqtSignal()

    def __init__(self, checkablemenuitems=None, parent=None):
        super().__init__(parent)
        self._items = {}
        self._actions = []

        if checkablemenuitems:
            self.set_items(checkablemenuitems)

        self.update_placeholder()

    def contextMenuEvent(self, event):
        menu = QMenu(self)
        for action in self._actions:
            menu.addAction(action)
        menu.exec(event.globalPos())

    def _on_action_toggled(self, checked):
        action = self.sender()
        if isinstance(action, QAction):
            key = action.data()
            if key in self._items:
                self._items[key].checked = checked
                self.update_placeholder()
                self.choicesChanged.emit()

    def update_placeholder(self):
        selected_items = self.get_checked()
        if selected_items:
             self.setPlaceholderText(_("Find in %s" % ", ".join(i.label for i in selected_items)))
        else:
            self.setPlaceholderText("Find...")

    def get_checked(self):
        return tuple(item for item in self._items.values() if item.checked)

    def set_items(self, checkablemenuitems):
        for action in self._actions:
            action.toggled.disconnect(self._on_action_toggled)
        self._items.clear()
        self._actions.clear()

        for item in checkablemenuitems:
            self._items[item.key] = item
            action = QAction(item.label, self)
            action.setCheckable(True)
            action.setChecked(item.checked)
            action.setData(item.key)
            action.toggled.connect(self._on_action_toggled)
            self._actions.append(action)

        self.update_placeholder()
        self.choicesChanged.emit()


class FindBox(QtWidgets.QWidget):
    findChanged = QtCore.pyqtSignal(str, list)

        super().__init__(parent)
        layout = QtWidgets.QHBoxLayout(self)
        layout.setContentsMargins(2, 2, 2, 2)

        self.valid_tags = set(filter(lambda t: not t.is_hidden, ALL_TAGS))
        self.selected_filters = []  # Start with "All" selected

       # Add checkboxes for all tags
        items = []
        for tag in sorted(self.valid_tags, key=lambda t: ALL_TAGS.display_name(str(t)).lower()):
            name = str(tag)
            items.append(CheckableMenuItem(name, label=ALL_TAGS.display_name(name), checked=name in self.selected_filters))

        # find input
        self.find_query_box = CheckableMenuLineEdit(parent=self, checkablemenuitems=items)
        self.find_query_box.setPlaceholderText(_("Find"))
        self.find_query_box.setClearButtonEnabled(True)
        self.find_query_box.textChanged.connect(self._query_changed)
        layout.addWidget(self.find_query_box)

```

(and remove filter_button and _show_filter_dialog)

image


Context menu isn't persistent in this experiment so it closes once user toggled one option, and I didn't implement everything, but that's the idea.

picard/ui/find.py
```diff
@@ -0,0 +1,164 @@
...
+        scroll_layout.addWidget(label)
+
+        # Add a horizontal separator
+        line = QtWidgets.QFrame(scroll_content)
+        line.setFrameShape(QtWidgets.QFrame.Shape.HLine)
+        line.setFrameShadow(QtWidgets.QFrame.Shadow.Sunken)
+        scroll_layout.addWidget(line)
+
+        # Add checkboxes for all tags
+        for tag in sorted(self.valid_tags, key=lambda t: ALL_TAGS.display_name(t._name).lower()):
```
zas: Here, don't use `TagVar._name`, but `str()`: `str(tag)` is the unique name of the tag, with '~" prefix if hidden.

picard/ui/find.py
rdswift: Okay, I misinterpreted the comment as meaning all filters were selected.

picard/ui/mainwindow/actions.py
rdswift: In general my thinking is to avoid changing the current shortcut because it can cause confusion for people that use it if it suddenly performs a different function.

picard/ui/find.py
rdswift: I ***think*** I like the idea of showing the selected filters in the placeholder text to save as you've done, but I still like having a button to bring up the filter selection dialog.  Perhaps this could be the icon that we're talking about removing (or something similar), with a tooltip explaining that it is to select the fields upon which to filter.

Also, I suggest we avoid using "Find in" and rather use "Filter on".  In my opinion, this better describes what is actually happening, and avoids any confusion between this and the "Search" functionality.  Similarly, I suggest changing the item in the toolbar from "Find" to "Filter items" for clarity (and perhaps even add a tooltip indicating that it will toggle the filters).

picard/ui/mainwindow/actions.py
rdswift: Can we change the action item text from "Find" to "Filter items" (or something like that) to better describe the functionality (and possibly add a tooltip explaining that this will toggle the filter functionality)?  This may seem like a minor point (which it is), but I associate "find" as part of the "search" process and it seems a bit confusing to refer to filtering a list as "find".

rdswift:
As a general comment with respect to the testing, please have the tests use the actual code in the module being tested (through imports) rather than duplicating the code from the module in the test file.  That way the actual production code will be tested, and not a copy which may not be the same if changes are made to the production code.  I'll see if I can update the `test/test_ui_find.py` file to better illustrate what I mean.

picard/ui/mainwindow/actions.py
zas: @rdswift I agree, we should be more consistent here, I suggest we use filter rather than find everywhere (including in filenames & class names).

@jpmsousa03 I think we don't need to have separate handling for filepaths & filenames, instead we'll introduce a filterable property (see https://github.com/metabrainz/picard/pull/2669), it will simplify code.

zas:
@jpmsousa03 please ensure each commit passes tests

rdswift:
> I'll see if I can update the `test/test_ui_find.py` file to better illustrate what I mean.

@jpmsousa03,

I just submitted a pull request with updated tests.  See https://github.com/jpmsousa03/picard/pull/1.

jpmsousa03:
Sorry I haven’t really paid much attention to testing, as @francescouteiro was more involved with the testing code than I was. I’ll make sure to pay more attention to it from now on.

picard/ui/mainwindow/actions.py
jpmsousa03: > @jpmsousa03 I think we don't need to have separate handling for filepaths & filenames, instead we'll introduce a filterable property (see #2669), it will simplify code.

So these would be treated as tags from metadata?

zas: Yes, since both are available (see https://github.com/metabrainz/picard/blob/1df0a511f27cf044296f2e55e02fa2f62c9b8f47/picard/const/tags.py#L391-L407), or is there a reason for a specific handling?

jpmsousa03:
I'm not particularly good with wording (especially in English), so there might be better ways to phrase things. Just let me know what you think.

Also changed: Ctrl+F -> Search as it was, Ctrl+Shift+F -> Toggle New Find Functionality

picard/ui/mainwindow/actions.py
jpmsousa03: I might be mistaken, but in my testing, I didn’t seem to find those tags on the right panel (the left panel shows '~filename' and 'dirpath'). 

https://github.com/jpmsousa03/picard/blob/a4aab0d61ba9dc94d6fe00532895bd6e1dd5a4a6/picard/ui/itemviews/basetreeview.py#L615-L626

It was supposed to be inside metadata, right?

rdswift:
> I'm not particularly good with wording (especially in English), so there might be better ways to phrase things. Just let me know what you think.

What you have now looks fine to me.  Thanks.

picard/ui/mainwindow/actions.py
```diff
@@ -348,10 +348,11 @@ def _create_show_toolbar_action(parent):
 
 
 @add_action(MainAction.SHOW_FINDBAR)
-def _create_find_toolbar_action(parent):
-    action = QtGui.QAction(_("Find"), parent)
+def _create_find_bar_action(parent):
```
zas: To be consistent, it should be named `_create_filter_bar_action()` (see my comment about find -> filter).

zas (changes requested):

jpmsousa03:
Since it seems to be more intuitive, I'll change find -> filter in the code. Tomorrow I might not be able to take a look at this (As I would like to have proper time and care to not miss any comment or function), but I'll be able to do it on the weekend.

picard/ui/find.py
jpmsousa03: My thoughts on this:

I think the context menu looks good, and it should be intuitive enough if we include something like “(Right-click to select filters)” in the placeholder text. My only concern is that this might interfere with or make it more confusing to use right-click for copy/paste (maybe it will not?), which I often used during testing, especially between the two bars.

The idea of showing active filters in the placeholder also makes sense, but I don’t think it’s sufficient to fully replace the labeled button without some explicit indication of which filters are active when there’s text entered (Which would also take some space).

picard/ui/mainwindow/actions.py
rdswift: > or is there a reason for a specific handling?

I think the issue is that the filtering needs to be done on the attributes `filename` and `filepath` but the tags are hidden ones so they are set as `~filename` and `~filepath` when retrieved from `ALL_TAGS`.

zas:
@jpmsousa03 we merged https://github.com/metabrainz/picard/pull/2670 and https://github.com/metabrainz/picard/pull/2669 so you can use `TagVar.is_filterable` and `filterable_tag_names()` to get the list of tags to use in filters.
Merge master branch (or rebase).

picard/ui/mainwindow/actions.py
zas: > > or is there a reason for a specific handling?
> 
> I think the issue is that the filtering needs to be done on the attributes `filename` and `filepath` but the tags are hidden ones so they are set as `~filename` and `~filepath` when retrieved from `ALL_TAGS`.

But  this shouldn't be an issue anymore if `TagVar.is_filterable` is used, right?

jpmsousa03:
I changed all "find" for "filter", as previouly discussed and removed the "Box" from the class name, since it also contained logic for the select filter dialog, not just the bar and button "box".

jpmsousa03:
> I noticed that filtering also filters the "Unclustered Files" and "Clusters" special items as well. Maybe those should stay visible, as they are part of the default UI and usually always visible.
> 
> The items are `ClusterItem` items, which hold a reference to a `Cluster` object (or subclass of that. The object has a `special` attribute that indicates their special role. E.g. it causes them to not be removable, and not be sortable. Maybe the filtering logic can exclude those too. If needed, this could also set a special flag on the `TreeItem` object instead, similar to how it is done for the `sortable` in https://github.com/metabrainz/picard/blob/master/picard/ui/itemviews/basetreeview.py#L553

@phw I was experimenting with making special items always visible, but it doesn’t look right , as clusters with no matches would still appear (as shown in the image). 

 What do you think?
image

picard/ui/mainwindow/actions.py
rdswift: It still seems to be a problem because `~filename` and `~filepath` are identified as not being tags in `ALL_TAGS` so are not being yielded in the `filterable_tag_names()` generator.  I tested here by identifying them as tags so that they would appear in the filter list, but the filtering did not work properly.  I suspect it's because of the leading "~" in the tag names not actually matching a tag in the metadata.

It looks like there will need to be some specific handling required in any case, so my suggestion (at least for now) is to handle them in the `filter` module as currently in this PR.  At least I know that works as expected in my testing.

rdswift:
I just noticed something in my testing that might be worth discussing.  If you set a filter item that doesn't actually exist in the metadata all items are hidden.  For example, if you set a filter to "genre" and there is no "genre" tag in the metadata, all of the items will be hidden.

This sort of makes sense, but I wonder if it would be better to ignore the filtering for selected filter tags that don't actually exist in the metadata.  I think @zas brought this up earlier, but I'm raising it again for discussion and to come to some clear consensus.  My thinking is this should be applied on an item-by-item basis when actually applying the filters, and not impact the list of available filters displayed for selection.

jpmsousa03:
> I just noticed something in my testing that might be worth discussing. If you set a filter item that doesn't actually exist in the metadata all items are hidden. For example, if you set a filter to "genre" and there is no "genre" tag in the metadata, all of the items will be hidden.
> 
I believe this is the expected behavior, and changing it might actually confuse users. For example, suppose someone wants to check whether the "Release Date" tag was applied to their music after autotagging. They might try to filter by "Release Date". If no tracks have that tag, I still think the user should still be able to select that filter, and when they type an expected value like "2002" and see no results, it becomes clear that the tag wasn’t added.

If, on the other hand, the "Release Date" filter isn’t shown at all (as @zas suggested), users might end up searching for the filter on the list and become frustrated when they can’t find it. (As it might not be obvious that the filter is missing simply because no music has that tag).

picard/ui/mainwindow/actions.py
rdswift: @zas,

Thinking about this a bit more, we ***could*** use the `filename` and `filepath` from `ALL_TAGS` but we would need to change the `is_hidden` attribute to `False` on them.  I'm not sure if this would have other unintended consequences.  They are already marked as `is_tag = False`.  As such, we would also have to change the `filterable_tag_names()` generator to not filter on `is_tag` (which probably makes sense anyway).

I just did a quick test and that seems to work fine for the filtering.  I don't think it will affect anything else, but I haven't done thorough testing to see if there are any other unintended consequences.

If you like, I can open a PR to make the necessary changes in the `picard/tags/__init__.py` and `picard/const/tags.py` files, and submit a PR to @jpmsousa03 for the changes in the `picard/ui/filter.py` and `test/test_ui_filter.py` files (not to be merged until the tags PR is merged into `master` and `search_feature` is rebased).

**EDIT:**  Turns out it **does** have a consequence in that the titles shown in the metadata lists are displayed as "~filename" and "~filepath" rather than "File Name" and "File Path".  Looks like there will be some special processing required no matter how we proceed.  One way would be to introduce (yet another) `TagVar` attribute/property called `filter_key` that would default to the `__str__()` value if not set.  That should only be needed for this filering stuff, and should be really easy to implement.

zas: Filterable should be independent from is_tag or is_hidden. Those shouldn't be changed in this context.

rdswift: > Filterable should be independent from is_tag or is_hidden.

I agree, and I'll create a PR to fix that.  The problem in this case is that the filter key for `filename` and `filepath` must be `filename` and `filepath`, and not `~filename` and `~filepath` which is what the `__str__()` function returns because the tags are marked as hidden.  That's why I was suggesting adding a `filter_key` attribute, so that the correct keys will be returned for the filtering code.  I'll add this into the PR so you can see what I'm getting at.  I'll put it in a separate commit so that we can easily revert it if we decide not to go that route.

zas: > The problem in this case is that the filter key for filename and filepath must be filename and filepath, and not ~filename and ~filepath which is what the __str__() function returns because the tags are marked as hidden

I don't get it, isn't it possible to use ~filename as key?

rdswift: I added https://github.com/metabrainz/picard/pull/2671 to remove the additional filtering from the `filterable_tag_names()` generator.

In looking at this whole thing a bit closer, along with the correction to the `filterable_tag_names()` generator I think the simplest solution is to change lines 602-613 of `picard/ui/itemviews/basetreeview.py` to use `~filename` and `~filepath` instead of `filename` and `filepath`.  That will allow selection of those tags from `ALL_TAGS` just like any other tags returned from the `filterable_tag_names()` generator, and avoid the need for the special "File Filters" section of the filter selection dialog.  I tested it here and it worked as expected.

That way there is no need for the extra `filter_key` attribute in the `TagVar` class, or any of the related code to return the filter key property, and that commit should be removed (or reverted) in my PR.

phw:
> I was experimenting with making special items always visible, but it doesn’t look right , as clusters with no matches would still appear (as shown in the image).
> 
> What do you think?

This doesn't look right indeed. But it's strange, because normal `Cluster` instances have `special = False`. Maybe we need a new attribute on the items to indicate they should not be hidden.

> I believe this is the expected behavior, and changing it might actually confuse users.

I agree. If I filter for a specific genre I wouldn't expect to find entries without a genre. For me the current behavior seems most useful.

zas:
> I agree. If I filter for a specific genre I wouldn't expect to find entries without a genre. For me the current behavior seems most useful.

Yes, after some thoughts, I agree.

phw:
@jpmsousa03 and others, I'd like your feedback on https://github.com/jpmsousa03/picard/pull/4 . It introduces keeping the "Unmatched files" and "Clusters" special tree items visible even when filtering.

jpmsousa03:
> @jpmsousa03 and others, I'd like your feedback on [jpmsousa03#4](https://github.com/jpmsousa03/picard/pull/4) . It introduces keeping the "Unmatched files" and "Clusters" special tree items visible even when filtering.

For me, I dont think it makes much of a difference. Might be a bit more explicit for the "searching for something that has no match" scenario, which is nice.

rdswift:
@phw,

>> I'd like your feedback on [jpmsousa03#4](https://github.com/jpmsousa03/picard/pull/4) . It introduces keeping the "Unmatched files" and "Clusters" special tree items visible even when filtering.

My comments added on the PR, and copied here to keep the discussion in one place.

> On the one hand I think it is good to keep those entries, as they are otherwise also always visible and not removable.

I agree, because consistent behavior is (almost) always the best choice.

> But filtering would still also show the counts, which might be confusing especially with empty result?

I'm less concerned with this, because I think it's easily explained.  The number of clusters and unclustered files hasn't changed.  It's only the visibility of the details that changes based on the filter.

I would be more concerned if, for example, the filtered view only showed one cluster when there are actually 21, and the user performed a "Lookup" which would trigger for all 21 clusters (not just the one displayed).

zas:
> The number of clusters and unclustered files hasn't changed. It's only the visibility of the details that changes based on the filter.

Likely more complicated, but we could have a display like (2/10) when 2 elements are shown over 10. It would require some changes in the current code that displays this number.

rdswift:
Any thoughts or comments would be appreciated on https://github.com/jpmsousa03/picard/pull/5.

zas:
> Any thoughts or comments would be appreciated on https://github.com/jpmsousa03/picard/pull/5.

It looks good to me, code is much simpler and unified.

picard/ui/filter.py
```diff
@@ -0,0 +1,176 @@
...
+
+        self.default_filter_button_label = N_("Filters")
+
+        # filter button
+        self.filter_button = QtWidgets.QPushButton(self.default_filter_button_label, self)
+        self.filter_button.setMaximumWidth(120)
+        self.filter_button.clicked.connect(self._show_filter_dialog)
+        layout.addWidget(self.filter_button)
+
+        self.filterable_tags = self.get_filterable_tags()
```
rdswift: Since the list of filterable tags shouldn't change, I wonder if it can be stored as a class variable and only updated by `get_filterable_tags()` if it is empty.  Currently the set is built each time the `Filter` class is instantiated, which can happen multiple times during a session.

jpmsousa03:
> Any thoughts or comments would be appreciated on [jpmsousa03#5](https://github.com/jpmsousa03/picard/pull/5).

Looks good to me too, ready to approve the merge If everyone agrees.

picard/ui/filter.py
jpmsousa03: The only scenario I can think of where this might be annoying is if a user downloads a plugin that introduces new tags (which I assume pluggins would be able to make them filterable), and the only way to see those new tags on the filter options would be to restart Picard. I'm not sure how relevant or common that scenario is, though.

rdswift: We have been in discussions about allowing plugins to register tags and variables, so that is a valid consideration.

A possible solution is to provide something in the plugin tag registration (or deregistration) process to trigger a reload of the filterable tags set by calling a class method of the `Filter` class.  This `Filter` class method could possibly accept an optional argument to force the (re)generation of the set.

jpmsousa03: Yes, maybe there could be a update_filterable_items_set() method on the class that gets called whenever there's a change in a plugin. I'm not too familiar with the plugin or configuration code, so I'm not sure how or where would this call happen (to deal with installs, uninstalls, updates) and how complicated it would be.

rdswift: @phw is still working on the revised plugin system for Picard v3, so all of that is still to be finalized.  The draft PR is https://github.com/metabrainz/picard/pull/2419.

I think that all we need to do for now is make sure that there is a class method in `Filter` that can be called to regenerate the filter set.

**UPDATE:**  I just added https://github.com/jpmsousa03/picard/pull/6 to help illustrate what I was thinking.

picard/ui/filter.py
```diff
@@ -0,0 +1,142 @@
...
+        layout.addWidget(button_box)
+
+        # Show dialog and process result
+        if dialog.exec() == QtWidgets.QDialog.DialogCode.Accepted:
+            self.selected_filters = []
+            for tag, checkbox in checkboxes.items():
+                if checkbox.isChecked():
+                    self.selected_filters.append(tag)
+
+            if not self.selected_filters:
```
zas: Those 2 lines aren't needed.

picard/ui/itemviews/basetreeview.py
```diff
@@ -567,3 +568,82 @@ def moveCursor(self, action, modifiers):
     @property
     def default_drop_target(self):
         return None
+
+    def setup_filter_box(self):
+        self.filter_box = Filter(self)
```
zas, line 573:
```suggestion
        self.filter_box = Filter(parent=self)
```

Better use keyword arguments, it eases future changes (like adding new parameters), and ease reading (no need to check the declaration to know what's passed).

picard/ui/itemviews/basetreeview.py
```diff
@@ -567,3 +568,82 @@ def moveCursor(self, action, modifiers):
...
+        return self.filter_box
+
+    def filter_items(self, text, filters):
+        if not text:  # When text is empty, show all items
+            self._restore_all_items()
+            return
+
+        self._filter_tree_items(self.invisibleRootItem(), text, filters)
+
+    def _filter_tree_items(self, parent, text, filters):
```
zas, line 587: I think this method can be simplified and splitted in smaller bits.

Totally untested, but I'm thinking about something like:

```python
    def _filter_tree_items(self, parent, text, filters):
        text = text.lower()
        match_found = False

        for i in range(parent.childCount()):
            child = parent.child(i)
            child_match = False

            if hasattr(child, 'obj'):
                obj = child.obj

                if self._matches_file_properties(obj, text, filters):
                    child_match = True

                if self._matches_metadata(obj, text, filters):
                    child_match = True

            if child.childCount() > 0:
                child_match |= self._filter_tree_items(child, text, filters)

            child.setHidden(not child_match)
            match_found |= child_match

        return match_found

    def _matches_file_properties(self, obj, text, filters):
        if hasattr(obj, 'iterfiles') and (not filters or "~filename" in filters or "~filepath" in filters):
            for file_ in obj.iterfiles():
                if (not filters or "~filename" in filters) and text in file_.base_filename.lower():
                    return True
                if (not filters or "~filepath" in filters) and text in file_.filename.lower():
                    return True
        return False

    def _matches_metadata(self, obj, text, filters):
        if hasattr(obj, 'metadata'):
            for tag, values in obj.metadata.rawitems():
                if isinstance(values, list):
                    for value in values:
                        if (not filters or tag.lower() in filters) and text in str(value).lower():
                            return True
                else:
                    if (not filters or tag.lower() in filters) and text in str(values).lower():
                        return True
        return False
```

We could also reduce the number of calls to `lower()` but that's not critical.

zas (changes requested):
Overall it looks good, I'll do further testing tomorrow.
Just few comments.

picard/ui/filter.py
rdswift: @zas / @phw,

Can you please have a look at https://github.com/jpmsousa03/picard/pull/6 to see if it makes sense before @jpmsousa03 merges it.  This is the PR to provide a class method that can be called by the (future) plugins code to reload the list of available filter tags if tags have been added or removed as a result of plugin registration or deregistration.  It also attempts to avoid rebuilding the filter selection dialog each time it is displayed.  Thanks.

picard/ui/itemviews/basetreeview.py
rdswift: > I think this method can be simplified and splitted in smaller bits.

That should also make it easier to write some tests for the code.

rdswift:
I was just doing some more testing and found something that looks odd.  Steps to reproduce are:

1. Start Picard.  For the purpose of this test ensure the following General option settings:
    - "Automatically scan all new files": `Disabled`
    - "Automatically cluster all new files": `Disabled`
    - "Ignore MBIDs when loading new files": `Enabled`

3. Enable filters view if not already visible from a persistent setting.

4. Enable the "File Path" filter in the clustering pane filter dialog, and set the filter text to something that will pass with your test file.  The "Unclustered Files" and "Clusters" directories will both be hidden.  I understand that this behavior is already being discussed, so I won't dwell on it here.

5. Drop a test file that meets the filter criteria onto the clustering pane.  Even though the file meets the filter criteria, it is not displayed.  This seems wrong to me.

6. Clear the filter text.  The file now appears in the list.

7. Re-add the filter test that you used in Step 3.  The file remains visible.

Alternately, replace Steps 5 and 6 with:

5. Open the filter selection dialog for the clustering pane.

6. Click "Ok" in the filter selection changes without making any changes.  The file now appears in the list.

The unusual behavior in question is in Step 4.  Is anybody else seeing this?

As I guess, I'm wondering if the filtering process is not being triggered when a file is dropped on the clustering pane, although I'm not sure that would explain why the newly dropped files aren't being displayed (unless it's because the parent "Unclustered Files" node is hidden).

**EDIT:** That appears to be exactly what's happening (the filtering is not being called when a file is dropped on the clustering pane).  If I have a previous file visible so that the "Unclustered Files" node is visible, the newly dropped file is displayed, even if it doesn't match the filter criteria.

rdswift:
> **EDIT:** That appears to be exactly what's happening (the filtering is not being called when a file is dropped on the clustering pane). If I have a previous file visible so that the "Unclustered Files" node is visible, the newly dropped file is displayed, even if it doesn't match the filter criteria.

@jpmsousa03,

To give you a starting point as to how this might be solved, have a look at https://github.com/rdswift/picard/commit/a5313f365a16ff0ed6f867b7d3b2cabd7df068dc.  This is built on the latest commit to https://github.com/jpmsousa03/picard/pull/6.  It seems to work, but it is not the most efficient because it will reapply the filtering after each file is loaded.  If you drag and drop an album with 10 tracks it will reapply the filter 10 times.  Most of the time that wouldn't be a big deal, unless the user drags a directory tree with a large number of albums or files.  Then it becomes a much larger problem.  Ideally we would wait until all files are loaded before reapplying the filter.

zas:
> t seems to work, but it is not the most efficient because it will reapply the filtering after each file is loaded

We have more or less the same issue with sorting and metadata updates, we have context managers to disable them during some operations, see https://github.com/rdswift/picard/blob/9be530fc286bd997a709b0a29b6475977c928856/picard/tagger.py#L727

rdswift:
> > t seems to work, but it is not the most efficient because it will reapply the filtering after each file is loaded
> 
> We have more or less the same issue with sorting and metadata updates, we have context managers to disable them during some operations, see https://github.com/rdswift/picard/blob/9be530fc286bd997a709b0a29b6475977c928856/picard/tagger.py#L727

Thanks for the tip.  It seems to be working properly now.

I added the context manager, but found that I also had to specifically set the flag in the `Tagger._file_loaded()` and `Tagger.add_files()` methods (similar to the sorting flag) to get it to work properly.

The latest working version is at https://github.com/rdswift/picard/commit/be713434da0c563f72cb5ffaff2a4551c4a9211f.

jpmsousa03:
> **EDIT:** That appears to be exactly what's happening (the filtering is not being called when a file is dropped on the clustering pane). If I have a previous file visible so that the "Unclustered Files" node is visible, the newly dropped file is displayed, even if it doesn't match the filter criteria.

Yes, the current implementation only updates the filtered items when the query or filter selection changes, something that definitely needs to be addressed. I’ve had to slow down coding for now due to some work stuff, but I’m revisiting this and the other issues over the weekend. While I can’t dive into the code at the moment, I’m still following along and really appreciate everyone’s feedback. (Also a special thanks to Bob for stepping in with the code help)

rdswift:
> Yes, the current implementation only updates the filtered items when the query or filter selection changes, something that definitely needs to be addressed.

I've included the fix based on @zas's comments into https://github.com/jpmsousa03/picard/pull/6.  It seems to work pretty well.

> I’ve had to slow down coding for now due to some work stuff, but I’m revisiting this and the other issues over the weekend. While I can’t dive into the code at the moment, I’m still following along and really appreciate everyone’s feedback.

Not a problem.  Real life happens to all of us.  The changes will still be there when things slow down and you get a chance to get back to it.

picard/ui/filter.py
rdswift: I removed them in one of the commits in the pending https://github.com/jpmsousa03/picard/pull/6.

phw:
I also was more absent over the long weekend we had here.

@rdswift @jpmsousa03 https://github.com/jpmsousa03/picard/pull/6 looks good to me. We should include it, especially as it addresses the issue of filtering new files.

Also let's merge https://github.com/jpmsousa03/picard/pull/4 then, everyone was in favor of it.

Apart from this it is fun to see so much collaboration here. Great PR to start with, now even better. It's seldom that there is such a large PR for a relatively complex new feature that touches many parts of the code base in such high quality. You clearly got a good understanding of Picard's code. Great work @jpmsousa03 and @francescouteiro .

rdswift:
https://github.com/jpmsousa03/picard/pull/6/ rebased and updated to use the `suspend_while_loading` functionality from https://github.com/metabrainz/picard/pull/2672.  Once this and https://github.com/jpmsousa03/picard/pull/4 are merged, we should be pretty close to finalizing this PR.  Should just be minor cleanup and optimizing items remaining.

test/test_ui_filter.py
```diff
@@ -0,0 +1,138 @@
...
+
+class FilterxTest(PicardTestCase):
+    """Test the Filter widget functionality"""
+
+    def test_filter_class_exists(self):
+        """Test that Filter class can be imported and has required attributes"""
+        self.assertTrue(hasattr(Filter, 'filterChanged'))
+        self.assertTrue(hasattr(Filter, '_query_changed'))
+        self.assertTrue(hasattr(Filter, 'clear'))
+        self.assertTrue(hasattr(Filter, 'set_focus'))
```
rdswift: @zas / @phw,

What are your thoughts on including code like this in the test file?  Personally, I really don't see the value in it because if any of these methods were missing there would be an exception raised if they were called.  They are not really testing the code itself.  Further, if changes are made in the code (such as refactoring), it could end up renaming (or removing) one or more of these methods, resulting in a failure of the test even though the functionality of the code still passes.

I'd like to remove them if you agree.

phw: @rdswift Yes, I agree with that

zas: @rdswift I agree too, not useful.

rdswift: Removed in https://github.com/jpmsousa03/picard/pull/6/.  I also added code to revert any filter selection changes when `Cancel` is selected (bug) and added a button to clear all selected filters.

jpmsousa03:
https://github.com/jpmsousa03/picard/pull/4 merged. I'll take a look and do some testing on https://github.com/jpmsousa03/picard/pull/6.

jpmsousa03:
I'll leave comments about the changes in https://github.com/jpmsousa03/picard/pull/6 here to keep the conversation more centralized.

 First of all, I really like the "Clear" buttom, it's definitely a nice addition!

That said, I’m wondering if it might be better positioned alongside the other buttons. In the current implementation, the "Clear All" button can be scrolled out of view, which I don’t think is ideal.

I’m attaching both the current and proposed layouts to help illustrate what I mean.

**Current:** (buttom scrolls)
image
**Proposed implementation:** (I've added a spacer to separate from closing dialog buttons)
image

Let me know your thoughts.

rdswift:
> Let me know your thoughts.

I like it on the bottom because that makes it always visible, but I'm okay with it either way.

picard/ui/itemviews/basetreeview.py
rdswift: @zas,

I added the refactoring that you suggested above into https://github.com/jpmsousa03/picard/pull/6/, and added some tests for the `_matches_file_properties()` and `_matches_metadata()` methods.

zas:
Yes, moving the Clear button is much better.

zas:
It works well though I'm bothered by the All=None filters and Clear All button are not very obvious.
If no checkbox is checked, then we filter on all fields, so Clear All is actually Check All without all checkboxes being checked... and we can't do easily do the reverse (for example, check all but one)


We have a dialog that does more or less that for cover art types:
image

Code is in https://github.com/metabrainz/picard/blob/9c6e49417e8f80ca7ad85945f97af187440171a0/picard/ui/caa_types_selector.py it could be made more generic (making it a reusable widget), and we could use it for selecting filters.

It uses 3 panes: included, available, excluded
And it has buttons to include/exclude all at once, to clear all (=no include/no exclude), and to restore defaults (we may have a limited list of fields to filter on by default instead of all).

Also selected filters should be saved/restored between sessions (=stored in options)

We could merge the PR as is for now and improve later (but before Picard3 release).

What do you think?

jpmsousa03:
> It uses 3 panes: included, available, excluded
> And it has buttons to include/exclude all at once, to clear all (=no include/no exclude), and to restore defaults (we may have a limited list of fields to filter on by default instead of all).

I like the idea of having "include" and "exclude" panels. However, I'm a bit confused about the purpose of the third panel. What does "available" actually mean? If something stays in the "available" panel, is it considered included or excluded? I feel like each item should either be in the "included" or "excluded" panel, having a third one seems unnecessary.

jpmsousa03:
Made some testing, this is functional, but I believe there’s room for discussion around the behavior and user experience (For example, what should happen if all tags are excluded, and how we can clearly communicate that to the user). That said, I agree this could be handled in a separate merge request, since it’s a UI concern that should not afect the logic of the functionality in this PR.


image

rdswift:
> However, I'm a bit confused about the purpose of the third panel.

This is explained in https://picard-docs.musicbrainz.org/en/config/options_cover_art_archive.html as well as in the text below the selection boxes.  There is separate "include" and "exclude" logic built into the coverart type selection because it is filtering on multiple attributes for the same image.  I don't think we need that same logic complexity for the filter selection in this case.

rdswift:
> It works well though I'm bothered by the All=None filters and Clear All button are not very obvious.
> If no checkbox is checked, then we filter on all fields, so Clear All is actually Check All without all checkboxes being checked... and we can't do easily do the reverse (for example, check all but one)

This "no filter selected" actually means "every filter is selected" has always bothered me as being confusing.

The dialog we use for the coverart might be a bit of overkill in this case.  I'm not sure we really need the separate "include" and "exclude" logic that is used for coverart selection.

I think we can get away with the current dialog if:

1. The processing is modified such that "no filters selected" doesn't apply any filtering -- the same as if there is no filter text entered.  This seems clearer to me than the current "no filters" means "all filters" processing.

2. We include buttons (perhaps in a separate button bar above or below the available filters scroll section?) for "Select All", "Clear All" and possibly "Invert Selection".

This will allow us to use the currently designed filter selection dialog (which I actually quite like) with only a couple of small changes.

**EDIT:** If we want to default to filter on everything, then we should indicate that by checking all of the filter checkboxes.

rdswift:
> Made some testing, this is functional, but I believe there’s room for discussion around the behavior and user experience

Functionally, this is the same as the current filter selection dialog, but takes up more space.  As I mentioned before, the dialog that @zas suggested provides the ability for both "include" and "exclude" logic, which I'm not sure we really need for the (relatively simple) filter selection.

The real issue seems to stem from the notion of "no filters selected" actually means "all filters selected".  Personally, I think it would be a lot clearer for the users if the behavior was changed such that if no filters are selected then no filtering occurs.  Then perhaps the default filter selection button text could be something like "Unfiltered" or "0 Filters" to make it clear that one or more filters needs to be selected for the filtering to occur.

zas:
> Functionally, this is the same as the current filter selection dialog, but takes up more space. As I mentioned before, the dialog that @zas suggested provides the ability for both "include" and "exclude" logic, which I'm not sure we really need for the (relatively simple) filter selection.

Yes, after some thoughts, it doesn't fit.

> 
> The real issue seems to stem from the notion of "no filters selected" actually means "all filters selected". Personally, I think it would be a lot clearer for the users if the behavior was changed such that if no filters are selected then no filtering occurs. Then perhaps the default filter selection button text could be something like "Unfiltered" or "0 Filters" to make it clear that one or more filters needs to be selected for the filtering to occur.

Actually:

- when ALL filters are selected, then ALL checkboxes should be checked (so one can easily uncheck one or two), the fact all filters = no checkbox checked is the source of confusion.
- if all are unchecked, filters will not match anything, which doesn't make much sense, but we can display a warning message in this case at dialog exit, it could even suggest a default selection, or All.
- instead of buttons we could have a special checkbox (like the one you have in many tables to select/unselect all rows), it would be perhaps more obvious, it could be always at top (see the GMail screenshot below for example, the top checkbox select/unselect all).

image

rdswift:
> * when ALL filters are selected, then ALL checkboxes should be checked (so one can easily uncheck one or two), the fact all filters = no checkbox checked is the source of confusion.

Agreed 100%.

> * if all are unchecked, filters will not match anything, which doesn't make much sense, but we can display a warning message in this case at dialog exit, it could even suggest a default selection, or All.

In my opinion, it would be easier for the user to understand if we just matched everything if there is no filter selected.  In that case, perhaps the text for the button used to open the filter selection dialog could be set to "Disabled".

> * instead of buttons we could have a special checkbox (like the one you have in many tables to select/unselect all rows), it would be perhaps more obvious, it could be always at top (see the GMail screenshot below for example, the top checkbox select/unselect all).

Agree 100%.  That makes a lot more sense than my original thinking of using buttons.

rdswift:
Further to our discussion, I have gathered my suggested changes and posted them in https://github.com/rdswift/picard/tree/search_feature_proposed_changes, in case anyone wants to review them (or try them out).  My intent is to create a new PR with these changes for consideration once this PR has been merged.

The changes include:

1. **Add Persistent Settings**:
     - Add persistent settings for selected filters, so that the settings are retained between sessions.

3. **Filter Logic Changes**:
    - Return True if no filters selected or no filter text entered.
    - Return True if no tags are found for selected filters.
    - Return True if tag found in selected filters and filter text in tag value.
    - Return False if one or more tags found in selected filters and filter text not in any values of the tags found.

4. **Add Master Selection Checkbox**:
    - Add checkbox at the top of the dialog to select / clear all filters.

The new dialog looks like:

image image

rdswift:
> We could merge the PR as is for now and improve later (but before Picard3 release).

Sounds good to me.

@zas / @phw: The system is showing that each of you have outstanding requested changes.  Have these been addressed sufficiently for now?  Also, I'm not sure what's up with the failing tests.

**EDIT:** Is the failing test related to the Python 3.13.4 bug that @zas addressed in a recent PR?  If so, should @jpmsousa03 be merging the latest `master` branch into their `search_feature` branch to address this?

rdswift (approved):
I think this is good enough for this PR.  It's functional and the code looks clean to me.  As discussed, we can explore changes in a follow-up PR (before the v3 release).

zas:
@rdswift ~~I noticed a bug, which is in your branch and this one, likely introduced by https://github.com/metabrainz/picard/pull/2668/commits/3019cff916f9dcda01de1c24e1c88ab5f5fd62c4~~

If you have **all** filters set (everything cleared, or all checked in your patch) filtering doesn't work properly, it does if, for example, you have one filter (let's say Title).
See:

Peek 16-06-2025 18-55

Actually, not a bug, but a non-intuitive behavior: one tag in all tracks has a `z` in it, so nothing is filtered, but we have no easy way to know which tag matched.


> EDIT: Is the failing test related to the Python 3.13.4 bug that @zas addressed in a recent PR? If so, should @jpmsousa03 be merging the latest master branch into their search_feature branch to address this?

Yes, this PR should be rebased onto master branch

rdswift:
> Actually, not a bug, but a non-intuitive behavior: one tag in all tracks has a `z` in it, so nothing is filtered, but we have no easy way to know which tag matched.

I agree that not knowing which tag matched is confusing.  I wonder if we could put a tooltip on the displayed lines to indicate which tags were matched?  ~Or perhaps a dialog that the user could display that shows all of the match results?~

Do we want to address this in this PR or do a separate one?

**EDIT:**  I tried the tooltip method and it seems to work okay.  If you want to see it in action, it is available at https://github.com/rdswift/picard/tree/filter_match_clarification (`filter_match_clarification` branch in my fork of the repository).  If this is something that we might want to consider, I can add it in a separate PR after this one is merged if you like.

**EDIT 2:**  In testing the tooltip method, something else showed up which is a bit disconcerting (to me at least).  With the current "no filters selected actually means filter on everything" strategy, if there are no filters selected, the matching process does indeed filter on **everything**, including tags that don't appear in the list of available filters.  My proposed changes to the filter logic in https://github.com/rdswift/picard/tree/search_feature_proposed_changes resolves this issue by only filtering on selected filters.  If we wanted to make that behavior available, perhaps it could be done by having an "Unspecified / Other" filter option, so that the user can decide.

zas (approved):

phw (approved):

jpmsousa03:
Thank you, everyone! It's been a really great experience discussing and collaborating with all of you on this PR. Special thanks to @rdswift for all the help with the coding!

This was my first feature contribution to Picard (and a MetaBrainz project), and I really enjoyed it. Once again, thanks to everyone involved. I’m looking forward to contributing to next PRs for this feature and beyond.

Wishing the best for Picard 3!

francescouteiro:
Hi! I apologize for the lack of communication on this PR. To streamline the process and avoid any confusion, I’ve been coordinating privately with @jpmsousa03 since this is part of a collaborative project.

We’d like to thank you all for your valuable guidance and extremely helpful feedback throughout this discussion. We really appreciate being part of such a welcoming community that turned what could be a challenging review process into a positive learning experience. We will love to keep an eye around this repo and to keep contributing!

Thank you again for your time and support! ❤️

@phw @zas @rdswift

rdswift:
@francescouteiro / @jpmsousa03,

Thanks to both of you for your contribution to the project.  It is a definite improvement, and I sincerely hope to see more from you in the future.

Sophist-UK:
> Thank you, everyone! It's been a really great experience discussing and collaborating with all of you on this PR. Special thanks to @rdswift for all the help with the coding!

As someone who has been an  active contributor on several different open source projects - MetaBrainz being only one of the organisations I was involved with - I would like to reinforce what Joao and Francisco have said about the Metabrainz projects, because IME Metabrainz is one of the friendliest, most cooperative, welcoming and supportive open source projects around. PRs are genuinely welcomed, and the leadership and other community members really want to help people to make their contributions a success - and believe me, IME this is not the case on the majority of open source projects out there. 😀

drzraf:
I should mention some filtering abilities (in the Album pane), once requested in https://tickets.metabrainz.org/browse/PICARD-150 but apparently not implemented in the first iteration of the filtering system :

- only show items with no changes / or only items with changes
- only show items which exist on the filesystem
- only show items items with x% match (or color level, or whatever)


If I got 30% of the official tracks (not full albums but custom compilations of selected songs), and that only a subset is to be tagged, and I want to concentrate on reviewing these who got a high match (> 80%), this is the set of filters I'd like to use to go from 1000 items down to 60.

PR #2669, opened 2025-06-04
Add new generators and tests

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Add new generators for hidden, visible and filterable tags.  Add tests.

# Problem

There may be a need for generating a list of hidden, visible or filterable tags only.  This may be the case in https://github.com/metabrainz/picard/pull/2668.

# Solution

Add new `filterable` TagVar attribute.  Add generators and tests.

# Action

Additional actions required:

### DIFF
--- a/picard/tags/__init__.py
+++ b/picard/tags/__init__.py
@@ -96,6 +96,10 @@ def hidden_tag_names():
     yield from ALL_TAGS.names(selector=lambda tv: tv.is_tag and tv.is_hidden)
 
 
+def filterable_tag_names():
+    yield from ALL_TAGS.names(selector=lambda tv: tv.is_tag and tv.is_filterable)
+
+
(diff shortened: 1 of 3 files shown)


rdswift:
I initially set this up so that the default is `True` for the `is_filterable` attribute.  Now that I think about it, should the default be `False` instead?

zas:
> I initially set this up so that the default is `True` for the `is_filterable` attribute. Now that I think about it, should the default be `False` instead?

Yes, perhaps better to default to `False` as the list of filterable tags will be likely smaller.

zas (approved):
LGTM

PR #2671, opened 2025-06-07
Tag filter corrections

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

1. The `filterable_tag_names()` generator only returns items that are also marked as being tags.

2. Sometimes the key used for filtering is different from the str value of a `TagVar` item, such as `filename` and `filepath` which return strings as `~filename` and `~filepath` when the filter requires `filename` and `filepath`.

# Solution

1. Remove the `is_tag` filter from the `filterable_tag_names()` generator.

2. Add a new (optional) `_filter_key` attribute (which defaults to `None`) to the `TagVar` class.  Add a `filter_key` property which returns the `_filter_key` attribute if it exists, otherwise the `__str__()` value of the item.  Add a `filter_key()` method to the `TagVars` class to retrieve the filter key for an item in the collection.  Add filter keys for the `filename` and `filepath` tags.  Update tests.

# Action

Additional actions required:

### DIFF
--- a/picard/const/tags.py
+++ b/picard/const/tags.py
@@ -413,6 +413,7 @@ ALL_TAGS = TagVars(
         is_tag=False,
         is_from_mb=False,
         is_filterable=True,
+        filter_key='filename',
     ),
     TagVar(
         'filepath',
@@ -422,6 +423,7 @@ ALL_TAGS = TagVars(
         is_tag=False,
         is_from_mb=False,
         is_filterable=True,
+        filter_key='filepath'
     ),
     TagVar(
         'filesize',
--- a/picard/const/tags.py
+++ b/picard/const/tags.py
@@ -413,7 +413,6 @@ ALL_TAGS = TagVars(
         is_tag=False,
         is_from_mb=False,
         is_filterable=True,
-        filter_key='filename',
     ),
     TagVar(
         'filepath',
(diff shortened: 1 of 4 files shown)


picard/const/tags.py
```diff
@@ -413,6 +413,7 @@
         is_tag=False,
         is_from_mb=False,
         is_filterable=True,
+        filter_key='filename',
```
zas: I don't think this is needed, especially since this is the value of `_name` property.

picard/tags/__init__.py
```diff
@@ -97,7 +97,7 @@ def hidden_tag_names():
 
 
 def filterable_tag_names():
-    yield from ALL_TAGS.names(selector=lambda tv: tv.is_tag and tv.is_filterable)
+    yield from ALL_TAGS.names(selector=lambda tv: tv.is_filterable)
```
zas, line 101: I agree with this change.

rdswift:
I'll remove the commit with the filter key stuff.  The solution is even simpler than I thought.  See https://github.com/metabrainz/picard/pull/2668#discussion_r2133456240

phw (approved):

PR #2676, opened 2025-06-16
Use public imports for mutagen, when possible

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

Picard in some cases extends mutagen and needs to access its private API. But if possible use the public imports.

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -58,7 +58,7 @@ import re
 import shutil
 import time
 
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

PR #2677, opened 2025-06-16
PICARD-2828: Separate options for standardized instrument/vocal credits

kellnerd:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

PICARD-1354 added an option to use standardized credit attributes for both instruments and vocals or to use no standardization at all.
I would like to standardize instrument credits, but keep vocals as credited because the vocal credit is often used to name roles of actors in audio dramas and operas.

* JIRA ticket (_optional_): PICARD-2828

# Solution

- Introduce a new boolean option (checkbox) `standardize_vocals` in all places where `standardize_instruments` appears.
- The old option affects only instrument credits from now on, while the new option does the same for vocal credits independently.

# Action

Additional actions required:
* [x] Migrate existing user settings, `standardize_vocals` should be initialized with the value of `standardize_instruments` to avoid a change in behavior
* [ ] Backport for Picard 2.x?

### DIFF
--- a/picard/ui/forms/ui_options_metadata.py
+++ b/picard/ui/forms/ui_options_metadata.py
@@ -1,6 +1,6 @@
 # Form implementation generated from reading ui file 'ui/options_metadata.ui'
 #
-# Created by: PyQt6 UI code generator 6.6.1
+# Created by: PyQt6 UI code generator 6.9.0
 #
 # Automatically generated - do not edit.
 # Use `python setup.py build_ui` to update it.
@@ -65,6 +65,9 @@ class Ui_MetadataOptionsPage(object):
         self.standardize_instruments = QtWidgets.QCheckBox(parent=self.metadata_groupbox)
         self.standardize_instruments.setObjectName("standardize_instruments")
         self.verticalLayout_3.addWidget(self.standardize_instruments)
+        self.standardize_vocals = QtWidgets.QCheckBox(parent=self.metadata_groupbox)
+        self.standardize_vocals.setObjectName("standardize_vocals")
+        self.verticalLayout_3.addWidget(self.standardize_vocals)
         self.convert_punctuation = QtWidgets.QCheckBox(parent=self.metadata_groupbox)
         self.convert_punctuation.setObjectName("convert_punctuation")
         self.verticalLayout_3.addWidget(self.convert_punctuation)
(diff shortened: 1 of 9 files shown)


picard/ui/forms/ui_options_metadata.py
```diff
@@ -1,6 +1,6 @@
 # Form implementation generated from reading ui file 'ui/options_metadata.ui'
 #
-# Created by: PyQt6 UI code generator 6.6.1
+# Created by: PyQt6 UI code generator 6.9.0
```
kellnerd, line 3: Running `python setup.py build_ui` changed a couple of other `ui_*.py` files as well, but I ignored them as the only change was this version number.
The file `picard/resources.py` has been completely rewritten, but that seems to be caused by a different PyQt version on my machine rather than by changes in this PR?

test/test_metadata.py
```diff
@@ -61,6 +61,7 @@
     'preferred_release_formats': [],
     'standardize_artists': False,
     'standardize_instruments': False,
+    'standardize_vocals': False,
```
kellnerd, line 64: I don't think this addition was necessary at all, the tests in this module seem to be completely unrelated to the standardization settings here.
I guess we could delete all three of them (and maybe even some others) in a separate commit?

The same goes for `test/test_mbjson.py`, there is half a dozen of tests which unnecessarily define defaults for `standardize_instruments`...

rdswift:
Since this adds a new configuration item, the documentation should also be updated.  I created a sub-task to capture this so we don't forget.  https://tickets.metabrainz.org/browse/PICARD-3082

rdswift (approved):
Looks okay to me.  Thanks.

picard/ui/forms/ui_options_metadata.py
phw: Yes, as you did it is fine. With different Qt versions the output can differ. If needed we can do a full update of all UI files, but that is seldom necessary.

test/test_metadata.py
phw: Those are needed if the option is accessed during the tests, which could be the case here (some tests are using `release_to_metadata` and `track_to_metadata`). Did you try to run the tests without this line?

phw (approved):
Thanks. Looks all good to me

test/test_metadata.py
kellnerd: No, I haven't, but your explanation that the tests may fail when they try to access this setting makes sense (for this test module).
In `test/test_mbjson.py` they are not accessed, otherwise these tests would have failed because I didn't add the new setting there. (Lucky coincidence, I just didn't feel like adding the same unrelated setting seven times.)

PR #2680, opened 2025-06-17
Filter feature proposed changes

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Suggested improvements to the newly added metadata filtering system.

# Problem

The newly added metadata filtering system provides some unexpected results.

# Solution

This PR proposes some changes as follows:

1. **Add Persistent Settings**:
     - Add persistent settings for selected filters, so that the settings are retained between sessions.

2. **Filter Logic Changes**:
    - Return True if no filters selected or no filter text entered.
    - Return True if no tags are found for selected filters.
    - Return True if tag found in selected filters and filter text in tag value.
    - Return False if one or more tags found in selected filters and filter text not in any values of the tags found.

3. **Add Master Selection Checkbox**:
    - Add checkbox at the top of the dialog to select / clear all filters.

4. **Add Logging**:
    - Add debug logging when the set of filterable tags is updated.

The new dialog looks like:

image image

# Action

Additional actions required:

### DIFF
--- a/picard/ui/filter.py
+++ b/picard/ui/filter.py
@@ -25,6 +25,7 @@ from PyQt6 import (
     QtWidgets,
 )
 
+from picard.config import get_config
 from picard.i18n import (
     N_,
     gettext as _,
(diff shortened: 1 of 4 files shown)


picard/ui/filter.py
```diff
@@ -70,6 +75,13 @@ def __init__(self, parent=None):
 
         self.initializing = False
 
+    def _get_saved_selected_filters(self):
+        config = get_config()
+        temp = config.persist[self.saved_filters_key]
+        if temp is not None:
+            return set(temp)
```
zas: We shouldn't load those blindly from config file in case those changed, but rather compare to filterable tags and include only those in config that are also available in running instance.

So basically it should be the intersection of both sets.

picard/ui/filter.py
```diff
@@ -45,20 +47,23 @@ class Filter(QtWidgets.QWidget):
     def __init__(self, parent=None):
         super().__init__(parent)
         Filter.instances.add(self)
+        self.saved_filters_key = f"{type(parent).__name__}FilterItems"
```
zas: Just a minor thing, but for most keys like this in persist we use 'key_Class' format, maybe stick to that.

```suggestion
        self.saved_filters_key = f"filters_{type(parent).__name__}"
```

zas (changes requested):

zas (approved):
Very good improvements. LGTM.

rdswift:
Two questions:

1. Should we also persist the filter text?
2. Should we include an option to filter on ANY metadata tag, including hidden tags and those not found in the `filterable_tags` set?  I can't think of a use case, but there may be one that other users might find necessary.

Both are fairly easy to include, but I don't want to overdo this with things that might not make sense.

zas:
> 1. Should we also persist the filter text?

I'd say no.

> 2\. Should we include an option to filter on ANY metadata tag, including hidden tags and those not found in the `filterable_tags` set?  I can't think of a use case, but there may be one that other users might find necessary.

Nope, keep it simple.

PR #2682, opened 2025-06-28
PICARD-3081: Replace forbidden filenames for Windows

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3081

# Solution

Forbidden names like NUL are replaced with a trailing underscore. Add a new replace_windows_forbidden_names utility function. Call it as part of make_save_path in win_compat mode.

### DIFF
--- a/picard/util/filenaming.py
+++ b/picard/util/filenaming.py
@@ -8,7 +8,7 @@
 # Copyright (C) 2017 Sambhav Kothari
 # Copyright (C) 2017 Ville Skyttä
 # Copyright (C) 2018 Antonio Larrosa
-# Copyright (C) 2019-2022, 2024 Philipp Wolfer
+# Copyright (C) 2019-2022, 2024-2025 Philipp Wolfer
 # Copyright (C) 2022 Bob Swift
 #
 # This program is free software; you can redistribute it and/or
@@ -539,7 +539,10 @@ WINDOWS_FORBIDDEN_NAMES = {
 }
 
 WINDOWS_FORBIDDEN_NAMES_RE = re.compile(
-    r'^(%s)(\.|$)' % '|'.join(re.escape(name) for name in WINDOWS_FORBIDDEN_NAMES),
+    r'(^|[{separators}])({names})(?=$|[\.{separators}])'.format(
+        separators=re.escape(os.path.sep + (os.altsep if os.altsep else '')),
+        names='|'.join(re.escape(name) for name in WINDOWS_FORBIDDEN_NAMES),
+    ),
     re.IGNORECASE
 )
 
@@ -554,17 +557,9 @@ def replace_windows_forbidden_names(path):
 
     Args:
         path: filename or path to clean
-    Returns: sanitized path
+    Returns: Path with escaped forbidden names
     """
-    parts = [
-        WINDOWS_FORBIDDEN_NAMES_RE.sub(r'\1_\2', part)
-        for part in path.split(os.path.normpath(os.path.sep))
-    ]
-    if parts[0] == '':
-        # If the first part is empty, it means the path started with a separator
-        # and is a root path (e.g. "/foo/bar"). Add back the leading separator.
-        parts[0] = os.path.sep
-    return os.path.join(*parts)
+    return WINDOWS_FORBIDDEN_NAMES_RE.sub(r'\1\2_', path)
 
 
 def get_available_filename(new_path, old_path=None):
--- a/picard/util/filenaming.py
+++ b/picard/util/filenaming.py
@@ -499,6 +499,8 @@ def make_save_path(path, win_compat=False, mac_compat=False):
 
     - If win_compat is True, trailing dots in file and directory names will
       be removed, as they are unsupported on Windows (dot is a delimiter for the file extension)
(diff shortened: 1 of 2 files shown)


picard/util/filenaming.py
```diff
@@ -527,6 +530,43 @@ def make_save_path(path, win_compat=False, mac_compat=False):
...
+    """
+    parts = [
+        WINDOWS_FORBIDDEN_NAMES_RE.sub(r'\1_\2', part)
+        for part in path.split(os.path.normpath(os.path.sep))
+    ]
+    if parts[0] == '':
+        # If the first part is empty, it means the path started with a separator
+        # and is a root path (e.g. "/foo/bar"). Add back the leading separator.
+        parts[0] = os.path.sep
+    return os.path.join(*parts)
```
rdswift: According to https://docs.python.org/3/library/os.path.html#os.path.join if there is a Windows drive specified, the `os.path.join()` function will not include the path separator between the drive and the first path element.  I believe that's why the test is failing.

I think you might need to first split off the drive before processing the path, and then add it back within the final `os.path.join()` being returned.  Perhaps something like the following (untested) code:

```suggestion
    _drive, _path = os.path.splitdrive(path)
    parts = [
        WINDOWS_FORBIDDEN_NAMES_RE.sub(r'\1_\2', part)
        for part in _path.split(os.path.normpath(os.path.sep))
    ]
    if parts[0] == '':
        # If the first part is empty, it means the path started with a separator
        # and is a root path (e.g. "/foo/bar"). Add back the leading separator.
        parts[0] = os.path.sep
    return os.path.join(_drive, *parts)
```

rdswift: Actually, since `win_compat` could be True even under a Linux system, you might have to convert `path` to a proper Windows path using `ntpath.normpath` and do the splitting on a backslash, while still using `os.path.normpath(os.path.sep)` for the returned `join()`.  Something like:

```suggestion
    import ntpath  # This should be moved to the imports at the top of the file.
    _drive, _path = ntpath.splitdrive(ntpath.normpath(path))
    parts = [
        WINDOWS_FORBIDDEN_NAMES_RE.sub(r'\1_\2', part)
        for part in _path.split('\\')
    ]
    if parts[0] == '':
        # If the first part is empty, it means the path started with a separator
        # and is a root path (e.g. "/foo/bar"). Add back the leading separator.
        parts[0] = os.path.sep
    return os.path.join(_drive, *parts)
```

rdswift: Note that the `_drive` portion will **not** be included unless the processing is being done on a Windows system (or `ntpath.join()` is used), so you may have to have separate tests for Windows and non-Windows systems in `test/test_util_filenaming.py`.  My assumption is that you want to use `os.path.join()` in the actual return value to ensure that the path is appropriate for the operating system in which Picard is running.

phw: I completely reworked this part. Splitting and joining the path can also alter the path in additional ways, especially  if we also have the need to call normpath.

It's now doing the replacement with a single regex matching the problematic names between start of string or directly after a directory separator, or at the end of string or directly before directory separator or dot.

It is a very similar approach as we use it in `make_save_path` for replacing leading or trailing dots.

picard/util/filenaming.py
```diff
@@ -527,6 +530,39 @@ def make_save_path(path, win_compat=False, mac_compat=False):
...
+}
+
+WINDOWS_FORBIDDEN_NAMES_RE = re.compile(
+    r'(^|[%s])(%s)(?=$|[\.%s])' % (
+        re.escape(os.path.sep + (os.altsep if os.altsep else '')),
+        '|'.join(re.escape(name) for name in WINDOWS_FORBIDDEN_NAMES),
+        re.escape(os.path.sep + (os.altsep if os.altsep else '')),
+    ),
+    re.IGNORECASE
+)
```
rdswift: Really minor point, but I wonder if the duplicate code could be avoided by using `.format()`, something like (untested):

```python
WINDOWS_FORBIDDEN_NAMES_RE = re.compile(
    r'(^|[{separators}])({names})(?=$|[\.{separators}])'.format(
        separators=re.escape(os.path.sep + (os.altsep if os.altsep else '')),
        names='|'.join(re.escape(name) for name in WINDOWS_FORBIDDEN_NAMES),
    ),
    re.IGNORECASE
)
```

**EDIT:**  Personally, I also find this a bit easier to follow what you're doing in the regular expression, although that's an even more minor consideration.  :stuck_out_tongue_winking_eye: 

**EDIT 2:**  Then again, it's not like this is going to save a lot of processing cycles since it is only ever being executed once.

rdswift (approved):
Looks fine to me.  Nice the way you did it all in the one regular expression.

picard/util/filenaming.py
phw: Yes, that's better. I didn't want to introduce yet another variable, as I found this even more convoluted and not easier to read. But using format seems perfect here. Thanks for the input, I updated the PR:

zas (approved):
LGTM

PR #2683, opened 2025-06-30
Make Metadata.length a property and raises exception if set to negati…

zas:
…ve value

# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Metadata.length can be set to a negative value and it shouldn't happen, so ensure we raise an exception in this case.

# Solution

# Action

Additional actions required:


phw (approved):
Makes sense, LGTM

PR #2684, opened 2025-07-03
Only calculate cluster "length" on display

zas:
- adding / subtracting doesn't work well if file metadata length changes over time (it leads to an exception in some cases)

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -133,7 +133,6 @@ class Cluster(FileList):
         if not added_files:
             return
         for file in added_files:
-            self.metadata.length += file.metadata.length
             file._move(self)
             file.update(signal=False)
             if self.can_show_coverart:
@@ -152,7 +151,6 @@ class Cluster(FileList):
 
     def remove_file(self, file, new_album=True):
         self.tagger.window.set_processing(True)
-        self.metadata.length -= file.metadata.length
         self.files.remove(file)
         self.update(signal=False)
         self.ui_item.remove_file(file)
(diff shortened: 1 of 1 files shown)


picard/cluster.py
```diff
@@ -223,7 +221,7 @@ def column(self, column):
         elif self.special and column in {'~length', 'album', 'covercount'}:
             return ''
         elif column == '~length':
-            return format_time(self.metadata.length)
+            return format_time(sum(f.metadata.length for f in self.files))
```
phw: This might be called really a lot of times, and we might have huge clusters.

I think we should have a `length` property on `Cluster` that recalculates the value on demand and gets marked as dirty when files get added or removed.

zas: What do  you think of https://github.com/metabrainz/picard/pull/2684/commits/8951b983f1636b5a2a23ca3121136d4b0a5a8125 approach ?

phw: @zas I like this approach of the list class.

Just the repeated calls to `self.set_dirty()` read a bit strange to me. Instead of a set I think it would be easier to read to have this a property and one can run `self.dirty =  True`.

Also I think we don#t need to mark it dirty on `sort` and `reverse`, as it doesn't change the element count.

zas: > @zas I like this approach of the list class.
> 
> Just the repeated calls to `self.set_dirty()` read a bit strange to me. Instead of a set I think it would be easier to read to have this a property and one can run `self.dirty = True`.
> 

Addressed in https://github.com/metabrainz/picard/pull/2684/commits/0f96974a145d874a0c62dfea70930c20d450163a

> Also I think we don#t need to mark it dirty on `sort` and `reverse`, as it doesn't change the element count.

True, removed in https://github.com/metabrainz/picard/pull/2684/commits/72e8fb0d8a03c7319e938ac5efd7111aaa41e6b3

PR #2687, opened 2025-07-18
Add German translations for debug options

bumblei3:
- Add translation for 'Plugin Fullpath' -> 'Plugin-Vollpfad'
- Add translation for 'Web Service Post Data' -> 'Webservice POST-Daten'
- Add translation for 'Log data of web service post requests' -> 'Daten von Webservice POST-Anfragen loggen'
- Add translation for 'Web Service Replies' -> 'Webservice-Antworten'
- Add translation for 'Log content of web service replies' -> 'Inhalt von Webservice-Antworten loggen'

These translations improve the German localization for debug options.

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/po/de.po
+++ b/po/de.po
@@ -160,7 +160,7 @@ msgstr "Primäres Dateibenennungsskript"
 
 #: picard/debug_opts.py:83
 msgid "Plugin Fullpath"
-msgstr ""
+msgstr "Plugin-Vollpfad"
 
 #: picard/debug_opts.py:83
 msgid "Log plugin full paths"
@@ -168,19 +168,19 @@ msgstr "Vollständige Plugin-Pfade loggen"
 
 #: picard/debug_opts.py:84
 msgid "Web Service Post Data"
-msgstr ""
+msgstr "Webservice POST-Daten"
 
 #: picard/debug_opts.py:84
 msgid "Log data of web service post requests"
-msgstr ""
+msgstr "Daten von Webservice POST-Anfragen loggen"
 
 #: picard/debug_opts.py:85
 msgid "Web Service Replies"
-msgstr ""
+msgstr "Webservice-Antworten"
 
 #: picard/debug_opts.py:85
 msgid "Log content of web service replies"
-msgstr ""
+msgstr "Inhalt von Webservice-Antworten loggen"
 
 #: picard/file.py:216
 msgid ""


rdswift:
@bumblei3,

Although I think you can submit them here via pull requests, in general the translations are managed through the MetaBrainz Weblate server, and I strongly suggest that you consider submitting them there.  See https://translations.metabrainz.org/projects/picard

Obwohl ich denke, dass Sie sie hier über Pull Requests einreichen können, werden die Übersetzungen im Allgemeinen über den MetaBrainz Weblate Server verwaltet, und ich empfehle Ihnen dringend, sie dort einzureichen.  Siehe https://translations.metabrainz.org/projects/picard

rdswift:
@bumblei3 

Please review https://github.com/metabrainz/picard/blob/master/po/README.md which explains the process for contributing translations to the Picard project.

Bitte lesen Sie https://github.com/metabrainz/picard/blob/master/po/README.md, wo das Verfahren für das Einbringen von Übersetzungen in das Picard-Projekt erläutert wird.

zas:
The patch contains totally unrelated changes. It looks to me you're trying to use AI powered editor and don't know what you are doing.
Such PR cannot be accepted.

Sophist-UK:
@bumblei3 Just to expand on what @zas and @rdswift have said, this is NOT how translations work in IT systems at all.

You cannot simply change message languages from English to German - this will:

1. Break all existing translations to other languages which are explicitly from English strings.
2. Change all messages to German for everyone regardless of which languages they understand.

I repeat - this is ***NOT*** how translation systems work ***in any system, anywhere***.

Please try to learn how to do things the right way, because you have some interesting ideas which I would really like to see included, but if you continue creating changes that break rather than enhance Picard, these ideas will never make it into the application.

phw:
I don't really know what to make of this. I didn't have time yesterday to review and comment, but I did see the initial PR with some translation changes. I probably would have had a comment about some string or such, but otherwise this would have been mergable. I would have explained what @rdswift wrote above about the preferred translation process on https://translations.metabrainz.org/projects/picard, but I would have also accepted a PR to update the po file. 

Today this PR has again exploded with a lot of unrelated changes, including a lot of half baked code and changes at the wrong places. Like zas I suspect this is just mostly AI generated code, and there is a lot the AI misses from how the project is structured.

I'm actually not going to review this, but a few general notes:

- The `ui_*.py` files are not supposed to be edited directly. They are automatically generated from the `.ui` files in the ui folder (running `python setup.py build_ui`), which are supposed to be edited with Qt Designer
- Commit messages and comments both must be written in English
- No UI or log strings should be added in German to the code. They must be in English as well and marked for translation (`_("The translatable string")`). The strings then get extracted from the sources and the pot and po files are updated, see https://github.com/metabrainz/picard/blob/master/po/README.md
- Recommended translation workflow is really Weblate
- Updating the po files for attributes, constants and countries is pointless, as those translation projects are shared with the MB server and the files are fetched from Weblate separately. They really should be edited on Weblate
-  I don't think we need a tooltip for every UI element. Tooltips should be used in cases where this extra detail is needed

PR #2688, opened 2025-07-21
Improve global error handling on application startup

bumblei3:
Description: This PR adds a global exception handler to the main entry point in tagger.py. If an unexpected error occurs during startup or runtime, Picard will now display a clear error message (including a traceback) instead of crashing silently. This improves user feedback and makes debugging easier.

Motivation:

Prevents silent crashes and improves user experience.
Provides more helpful error messages for users and developers.
Makes it easier to report and diagnose issues.
How it works:

Wraps the main application logic in a try/except block.
On exception, shows a fatal error dialog (or prints to console) with details and traceback.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1429,39 +1429,48 @@ def setup_translator(tagger):
 
 def main(localedir=None, autoupdate=True):
     """Main entry point to the program"""
-    setup_application()
+    try:
+        setup_application()
 
-    signal.signal(signal.SIGINT, signal.SIG_DFL)
+        signal.signal(signal.SIGINT, signal.SIG_DFL)
 
-    cmdline_args = process_cmdline_args()
+        cmdline_args = process_cmdline_args()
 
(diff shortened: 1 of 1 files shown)


phw:
Thanks. The patch itself looks clean.

But actually we already have this, and even with a UI showing the user the crash details:

grafik

The exception will also be logged to the console.

This is set up in the entry points at https://github.com/metabrainz/picard/blob/master/tagger.py.in and https://github.com/metabrainz/picard/blob/master/scripts/picard.in (the latter is the main one, the `tagger.py` is only for development).

The implementation originally was more similar to yours, with the try...catch. It was changed in #2418 to use `sys.excepthook` instead, as otherwise some exceptions did not get handled.

One reason why this is directly in the entry points and initialized as early as possible is that this way it also handles some cases where importing `picard.tagger` would already fail.

phw:
I'm closing this for now.

PR #2689, opened 2025-07-22
PICARD-2300: Linux dark theme detection

knguyen1:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

  This change ensures that Picard correctly detects and applies the system dark mode on Linux when the “System” UI theme is selected, supporting multiple desktop environments and providing comprehensive tests.

# Problem

Picard did not honor the system’s dark mode setting on Linux when the “System” theme was selected, especially if Qt’s platform theme integration was missing or incomplete. This led to inconsistent user experience for users with dark system themes on GNOME, KDE, XFCE, LXQt, and similar environments.

* JIRA ticket (_optional_): [PICARD-2300](https://tickets.metabrainz.org/browse/PICARD-2300)

# Solution

- Refactored dark mode detection logic into a dedicated module using the Strategy Pattern.
- Added support for detecting dark mode on GNOME, KDE, XFCE, and LXQt desktop environments.
- Modularized detection strategies for easy extension and maintenance.
- Updated and expanded unit tests to cover all supported environments and edge cases, using pytest and proper parameterization.
- Ensured all code is DRY, single-responsibility, and follows Picard’s contribution guidelines.

# Action

Additional actions required:

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -155,8 +155,9 @@ class BaseTheme:
             self._accent_color = palette.color(QtGui.QPalette.ColorGroup.Active, QtGui.QPalette.ColorRole.Highlight)
 
         # Linux-specific: If SYSTEM theme, try to detect system dark mode
+        # Do not apply override if already dark theme
         is_dark_theme = self.is_dark_theme
-        if (not IS_WIN and not IS_MACOS and not IS_HAIKU and self._loaded_config_theme == UiTheme.SYSTEM):
+        if (not self._dark_theme and not IS_WIN and not IS_MACOS and not IS_HAIKU and self._loaded_config_theme == UiTheme.SYSTEM):
             is_dark_theme = self._detect_linux_dark_mode()
             if is_dark_theme:
                 # Apply a dark palette centrally defined
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -43,6 +43,26 @@ from picard.ui.theme_detect import get_linux_dark_mode_strategies
 # DRY: Common dark background color
 DARK_BG_COLOR = QtGui.QColor(51, 51, 51)
 
+# Centralized dark mode palette for Windows and Linux
+DARK_PALETTE_COLORS = {
+    QtGui.QPalette.ColorRole.Window: DARK_BG_COLOR,
+    QtGui.QPalette.ColorRole.WindowText: QtCore.Qt.GlobalColor.white,
+    QtGui.QPalette.ColorRole.Base: QtGui.QColor(31, 31, 31),
+    QtGui.QPalette.ColorRole.AlternateBase: DARK_BG_COLOR,
+    QtGui.QPalette.ColorRole.ToolTipBase: DARK_BG_COLOR,
+    QtGui.QPalette.ColorRole.ToolTipText: QtCore.Qt.GlobalColor.white,
+    QtGui.QPalette.ColorRole.Text: QtCore.Qt.GlobalColor.white,
+    QtGui.QPalette.ColorRole.Button: DARK_BG_COLOR,
+    QtGui.QPalette.ColorRole.ButtonText: QtCore.Qt.GlobalColor.white,
+    QtGui.QPalette.ColorRole.BrightText: QtCore.Qt.GlobalColor.red,
+    (QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Text): QtCore.Qt.GlobalColor.darkGray,
+    (QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Light): QtGui.QColor(0, 0, 0, 0),
+    (QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.ButtonText): QtCore.Qt.GlobalColor.darkGray,
+    (QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Base): QtGui.QColor(60, 60, 60),
+    (QtGui.QPalette.ColorGroup.Inactive, QtGui.QPalette.ColorRole.Highlight): DARK_BG_COLOR,
+    (QtGui.QPalette.ColorGroup.Inactive, QtGui.QPalette.ColorRole.HighlightedText): QtCore.Qt.GlobalColor.white,
+}
+
 
 OS_SUPPORTS_THEMES = True
 AppKit = None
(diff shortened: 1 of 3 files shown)


picard/ui/theme.py
```diff
@@ -118,7 +134,34 @@ def setup(self, app):
         if self._dark_theme:
             self._accent_color = palette.color(QtGui.QPalette.ColorGroup.Active, QtGui.QPalette.ColorRole.Highlight)
 
+        # Linux-specific: If SYSTEM theme, try to detect system dark mode
         is_dark_theme = self.is_dark_theme
+        if (not IS_WIN and not IS_MACOS and not IS_HAIKU and self._loaded_config_theme == UiTheme.SYSTEM):
```
phw: I think this should also check for `not self._dark_theme`, because if this was already detected as a dark color scheme then we don't need to override the colors with hard coded ones.

picard/ui/theme_detect.py
```diff
@@ -0,0 +1,109 @@
...
+                'gsettings', 'get', 'org.gnome.desktop.interface', key,
+            ], capture_output=True, text=True, check=True,
+        )
+        return result.stdout.strip().strip("'\"")
+    except (subprocess.CalledProcessError, FileNotFoundError):
+        log.debug(f"gsettings get {key} failed.")
+        return None
+
+def detect_gnome_color_scheme_dark() -> bool:
+    value = gsettings_get('color-scheme')
```
phw: DE specific color scheme detections should also consider the actually running DE, otherwise the first strategy would win even if I'm running a different DE.

picard/ui/theme.py
```diff
@@ -118,7 +134,34 @@ def setup(self, app):
...
+                palette.setColor(QtGui.QPalette.ColorRole.ButtonText, QtCore.Qt.GlobalColor.white)
+                palette.setColor(QtGui.QPalette.ColorRole.BrightText, QtCore.Qt.GlobalColor.red)
+                palette.setColor(QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Text, QtCore.Qt.GlobalColor.darkGray)
+                palette.setColor(QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Light, QtGui.QColor(0, 0, 0, 0))
+                palette.setColor(QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.ButtonText, QtCore.Qt.GlobalColor.darkGray)
+                palette.setColor(QtGui.QPalette.ColorGroup.Disabled, QtGui.QPalette.ColorRole.Base, QtGui.QColor(60, 60, 60))
+                palette.setColor(QtGui.QPalette.ColorGroup.Inactive, QtGui.QPalette.ColorRole.Highlight, DARK_BG_COLOR)
+                palette.setColor(QtGui.QPalette.ColorGroup.Inactive, QtGui.QPalette.ColorRole.HighlightedText, QtCore.Qt.GlobalColor.white)
+                self._dark_theme = True
+                self._accent_color = palette.color(QtGui.QPalette.ColorGroup.Active, QtGui.QPalette.ColorRole.Highlight)
```
phw, line 171: If we really must use a hard coded palette, why not use the same as we currently do for Windows? We could factor this out and define the palette centrally.

knguyen1: This commit centralises dark colours: https://github.com/metabrainz/picard/pull/2689/commits/2d48f0f17cca757636dff5d6137834d917aad94c

picard/ui/theme_detect.py
knguyen1: This adds de-aware detection: https://github.com/metabrainz/picard/pull/2689/commits/59f5021c2883a55a73c1ec0b3fc59ccf0393bc40

picard/ui/theme.py
knguyen1: This adds `not self._dark_theme`: https://github.com/metabrainz/picard/pull/2689/commits/b16d75df978f5bdf07499be9817f5a293ba48f56

knguyen1:
> Have you looked into the org.freedesktop.appearance.color-scheme interface? That might be another strategy to detect the wanted color scheme and could be the first strategy to try. Currently supported by Gnome at least, see the links at [PICARD-2300](https://tickets.metabrainz.org/browse/PICARD-2300)

Yes.  This is what this PR sets out to do.

> Qt 6 provides a QStyleHints.setColorScheme method, and you can query the current color scheme also.

I'll take a stab at it after this gets merged, or in a separate PR.

picard/ui/theme_detect.py
```diff
@@ -119,13 +120,50 @@ def detect_freedesktop_color_scheme_dark() -> bool:
...
+    if de:
+        return de.lower()
+    # Fallbacks for KDE, XFCE, LXQt
+    if os.environ.get("KDE_FULL_SESSION") == "true":
+        return "kde"
+    if os.environ.get("XDG_SESSION_DESKTOP"):
+        return os.environ["XDG_SESSION_DESKTOP"].lower()
+    if os.environ.get("DESKTOP_SESSION"):
+        return os.environ["DESKTOP_SESSION"].lower()
+    return ""
```
phw, line 167: Just as a side note: We have similar (but less complete) checks also in `icontheme.py`. We could unify this later (by using `get_current_desktop_environment` also in icontheme). But I would keep this out of this PR.

phw (commented):
> > Have you looked into the org.freedesktop.appearance.color-scheme interface? That might be another strategy to detect the wanted color scheme and could be the first strategy to try. Currently supported by Gnome at least, see the links at [PICARD-2300](https://tickets.metabrainz.org/browse/PICARD-2300)
> 
> Yes. This is what this PR sets out to do.

I see :) So it's currently in `detect_freedesktop_color_scheme_dark`, querying this via gsettings. The next step would be to query this via DBus form [org.freedesktop.portal.Settings](https://github.com/flatpak/xdg-desktop-portal/blob/main/data/org.freedesktop.portal.Settings.xml) interface. Then it would also work e.g. when run as a flatpak package. But that is also something we can do in a separate PR.

> > Qt 6 provides a QStyleHints.setColorScheme method, and you can query the current color scheme also.
> 
> I'll take a stab at it after this gets merged, or in a separate PR.

That would be awesome. I really don't yet know how well the new Qt functionality works. But if it does what it promises and it works also on Linux we probably could bring the light / dark theme handling on par with how it already works on macOS and Windows. We even have the potential to simplify things and handle this more generically with less platform specific code.

Because of how fractured the DE landscape is you newly added detection code will very likely still be needed, though :)

This PR actually looks good to me already, but I don't have the time for a more thorough review right now. I'll review this in more detail later.

knguyen1:
I'll address the failures when I get home.

I tested on Windows and Ubuntu. I also have Mac machines I can play with.

knguyen1:
Can you run workflows now?  Should be okay. @zas

picard/ui/theme.py
```diff
@@ -108,7 +144,7 @@ def setup(self, app):
             app.setStyle(MacOverrideStyle(app.style()))
 
         app.setStyleSheet(
-            'QGroupBox::title { /* PICARD-1206, Qt bug workaround */ }'
+            'QGroupBox::title { /* PICARD-1206, Qt bug workaround */ }',
```
phw, line 147: Unrelated for this PR review, but I really wonder if this is still needed. We should check that.

zas (approved):
LGTM

phw:
Merged. Thanks a lot for this contribution

PR #2690, opened 2025-07-24
Modernize dev tooling: switch to uv, pre-commit, PEP 621, update docs

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Modernizes development tooling by migrating to uv for dependency management, adding pre-commit hooks, and adopting PEP 621 project configuration standards.

# Problem

The project was using outdated development tooling and dependency management practices. This made development setup more complex and inconsistent across different environments.

# Solution

- **Migrated to uv**: Replaced pip-tools with uv for faster and more reliable dependency resolution
- **Added pre-commit hooks**: Implemented automated code quality checks with flake8 integration and requirements files generation
- **PEP 621 compliance**: Updated [`pyproject.toml`](pyproject.toml) to use modern Python packaging standards
- **Centralized dependencies**: Consolidated requirements*.txt files into pyproject.toml as single source of truth
- **Updated documentation**: Reorganized dependency installation instructions in [`CONTRIBUTING.md`](CONTRIBUTING.md)

# Action

Additional actions required:

After merging, contributors should install `uv` and run `uv sync` as outlined on CONTRIBUTING.md. `uv sync` is recommended.

```bash
# install all dependencies (main, build, and dev)
uv sync

# alternatively, you can install dependencies manually using `uv pip`
uv pip install -r requirements.txt
uv pip install -r requirements-build.txt -r requirements-dev.txt

# important: run pre-commit install
uv run pre-commit install
```

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,30 +17,37 @@ repos:
     hooks:
       - id: uv-lock
         name: Sync uv.lock with pyproject.toml
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements.txt
         args: [pyproject.toml, -o, requirements.txt, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements-build.txt
         args: [pyproject.toml, -o, requirements-build.txt, --group, build, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements-dev.txt
         args: [pyproject.toml, -o, requirements-dev.txt, --group, dev, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements-plugins.txt
         args: [pyproject.toml, -o, requirements-plugins.txt, --group, plugins, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements-macos-11.0.txt
         args: [pyproject.toml, -o, requirements-macos-11.0.txt, --group, macos_11_0, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
       - id: pip-compile
         name: pip-compile requirements-win.txt
         args: [pyproject.toml, -o, requirements-win.txt, --group, win, --generate-hashes, --universal]
         pass_filenames: false
+        always_run: true
   - repo: local
     hooks:
       - id: pytest


pyproject.toml
```diff
@@ -28,7 +28,72 @@ classifiers = [
...
+    "Markdown==3.8",
+    "mutagen==1.47.0",
+    "PyJWT==2.10.1",
+    "PyQt6==6.9.0",
+    "PyQt6-Qt6==6.9.0",
+    "python-dateutil==2.9.0.post0",
+    "pywin32==310",
+    "PyYAML==6.0.2",
+    "charset-normalizer==3.4.2",
+]
```
phw, line 94: I think we can be less strict with the above, as exact versions are now also pinned in the requirements-(macos|win).txt files. Just for some background, there are essentially two reasons why we have those platform specific requirements files:

1. Generally so that we could pin the versions for our binary builds, so once we tag a release the binary builds are done with well defined dependencies.
2. More specifically because we frequently have the situation, that we need specific versions for some platform. Either because of platform specific bugs in a dependency, or because newer dependencies don't run on the targeted OS version.

The second one is the important one, and the reason why I kept this setup, even though we right now are on the latest versions of all dependencies on all platforms. But e.g. we had this just recently, where we needed to use an older PyQt6 on macOS due to build issues.

More often than not it was all about PyQt6 or PyQt6-Qt6 (the later is a dependency of the former, but sometimes we had to pin this to a bit older version). The rest of the dependencies  I usually just update to the latest version without issue. I'm always a bit wary about the OS specific ones (pywin32 and pyobjc), but I actually cannot recall a single time those caused a broken build.

So yes, let's keep it with those special optional dependency. But let's open it up so it uses the same version ranges as the normal dependencies. Then we can just update the requirements files with `uv pip compile pyproject.toml` from time to time, and if we have a specific need to pin something to a different version only then we do this here in pyproject.toml.

scripts/precommit-flake8.sh
```diff
@@ -0,0 +1,13 @@
...
+  -e "^tagger\\.py$" \
+  -e "^picard/resources\\.py$" \
+  -e "^picard/const/\(attributes\|countries\)\\.py$" \
+  -e "^picard/ui/ui_.*\\.py$" \
+  -e "^scripts/picard\\.in$")
+
+if [ ! -z "$PYFILES" ]; then
+  set -e
+  isort --check-only --diff --quiet $PYFILES
+  flake8 $PYFILES
```
phw: I also run ruff there now. But we can update this if we move over to ruff completely. zas already did most of the preparation, but there are maybe still some differences.

phw (changes requested):
This is great. zas and I had already talked about wanting to do this. Also looks good to me, special thumbs up for also considering the documenation.

I just added a note about handling the macos/win specific requirements files a bit different.

Sophist-UK:
I am not sure whether these changes only impact the packaging process or whether they will impact either new or existing users that are running from source?

I suspect that there are some documentation pages which may need updating - I have found https://picard.musicbrainz.org/docs/build-windows/ and https://picard.musicbrainz.org/docs/build-osx/ but there may be others.

phw:
Those pages are a pain to maintain, are likely outdated (especially the macos one) and sould probably be removed or replaced with a link to https://github.com/metabrainz/picard/blob/master/CONTRIBUTING.md

scripts/precommit-flake8.sh
knguyen1: I left that in there due to the legacy instructions on `CONTRIBUTING.md`.  We could delete `flake8` completely and also have `ruff` handle checking and `isort` (you add rule `I` to `ruff` conffig).  So yeah, `ruff` can handle both checking and isorting.

.pre-commit-config.yaml:9-14
```
  - repo: https://github.com/charliermarsh/ruff-pre-commit
    rev: v0.8.3
    hooks:
      - id: ruff
        args: [--fix]
      - id: ruff-format
```

Thoughts?

pyproject.toml
knguyen1: Okay.  Will address this when I get home.   Good suggestions and thanks for the details!

scripts/precommit-flake8.sh
phw: Not sure, I would leave it for now and make this shift in a separate step. There were some things working a bit differently I think, hence we had ruff in parallel for now. But @zas will know more.

phw:
This PR will also close PICARD-2816. It was proposed to use poetry there, but this proved to be not a good fit for our requirements. zas already brought uv up in the discussion there, which is a much better fit.

scripts/precommit-flake8.sh
zas: I don't remember exactly but I couldn't obtain exact same result with ruff vs isort + flake8, but prolly just some extra configs needed.

@knguyen1 feel free to try, having just ruff handling everything would be great.

phw: I also wouldn't mind if this means a small change to our coding style. Generally ruff has been working well for me, and it is so fast that it's impact on the pre-commit is really low.

I think I remember one difference: flake8 was more strict with double blank lines between certain elements. I'm not sure which way, but either it required double blank lines in an instance where our ruff setup didn't, or the other way around it expected a single line where the ruff setup would allow multiple. I just had this once when I checked the code locally only with ruff, and then the CI build failed with flake8.

knguyen1: Oooff... Adding rule `I` is causing the ruff pipelines to fail.

https://github.com/metabrainz/picard/pull/2690/commits/2d3745a42f983640d16cf18bb914feb9cf899248 This is the commit that I got pretty close to matching current `isort`.  I'll investigate more.

knguyen1: https://github.com/metabrainz/picard/pull/2690/commits/efa2ceb62be95f9250b15e071a6b5ba58f3ce1da @zas @phw This commit is the closest I can get it to match with current files.  It requires one change in `picard/script/__init__.py` (a blank line at 49).  Let me know your thoughts and I'll begin removing `flake` and `isort` in favour of `ruff`.

In the meantime, let me ask `ruff` authors about it...

phw: @knguyen1  Looks good to me and I would be fine with killing the run of isort in `lint.yml` (which now fails because it disagrees on the one changed line). @zas , what do you think?

zas: I would be fine with that too

knguyen1: Okay, removing refs to `flake8` and `isort` now.. Inclulding the CI pipelines.

pyproject.toml
knguyen1: This commit makes these deps less restrictive: https://github.com/metabrainz/picard/pull/2690/commits/b3764f0540be34d5de8bf99da5697a528bb9b0f9

knguyen1:
Ready for review.  Summary of what's been done:

- Removed references to `flake8` and `isort` and their dependencies
- Added settings to make `ruff` isort match with `isort`.
- Collapsed the `ruff`, `isort`, `flake8` CI pipelines to `ruff check --output-format=github`

Note:

To `lint.yml`, we can add a stage for `ruff format --check picard` but that isn't possible right now since there are legacy files not formatted with ruff.  I would leave it for another refactor PR:

> ruff format --check picard
> Would reformat: picard\util\thread.pyy
> ...
> Would reformat: picard\webservice\utils.py
> 182 files would be reformatted, 31 files already formatted
> (exit 1)

phw (approved):
A good step forward, thanks. LGTM

zas (approved):
Very nice improvement, I ran commands from doc with success. LGTM.

zas:
@knguyen1 perhaps you should also update INSTALL.md to use uv.

PR #2692, opened 2025-07-31
PICARD-2300: Add `qtdbus` dark theme detect

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Added dark theme detection using QtD-Bus for Linux systems with comprehensive unit test coverage.

# Problem

Existing linux dark theme detection uses `gsettings`, calling insecure subprocesses and reading files/services that may not exist.  Added support for `qtdbus`, which should be more secure, with fallback to legacy strategies.

Note: Couldn't implement `dbus-python`, as that requires system tools to be installed.

# Solution

**Dark Theme Detection Implementation:**
- Added QtD-Bus integration for detecting dark theme preferences on Linux systems
- Implemented `DBusThemeDetector` class that communicates with freedesktop portal and GNOME dconf services
- Supports both freedesktop portal color scheme detection and GNOME-specific theme detection
- Graceful fallback mechanism when services are unavailable

**Comprehensive Test Coverage:**
- Added extensive test file `test_theme_detect_qtdbus.py` with 734 lines of test coverage
- Tests cover D-Bus connection initialization, error handling, and service availability
- Parametrized tests for different message types, service states, and error conditions
- Integration tests verify detection priority between portal and GNOME methods
- Mock fixtures ensure reliable testing without requiring actual D-Bus services

**Technical Features:**
- Service availability checking to avoid unnecessary D-Bus calls
- Exception handling for various D-Bus communication scenarios
- Support for both freedesktop portal (preferred) and GNOME dconf fallback
- Singleton pattern for detector instance management
- Comprehensive error handling for connection failures and service unavailability

# Action

Additional actions required:

### DIFF
--- a/picard/ui/theme_detect.py
+++ b/picard/ui/theme_detect.py
@@ -28,6 +28,12 @@ import subprocess  # noqa: S404
 
 from picard import log
 
+from picard.ui.theme_detect_qtdbus import (
+    detect_freedesktop_color_scheme_dbus,
+    detect_gnome_color_scheme_dbus,
+    get_dbus_detector,
+)
+
 
 def gsettings_get(key: str) -> str | None:
     """Get a gsettings value as a string or None."""
@@ -51,6 +57,16 @@ def gsettings_get(key: str) -> str | None:
 
 def detect_gnome_color_scheme_dark() -> bool:
     """Detect if GNOME color-scheme is set to dark."""
+    # Try D-Bus first (secure method)
+    try:
+        detector = get_dbus_detector()
+        result = detector.detect_gnome_color_scheme_dbus()
+        if result is not None:
+            return result
+    except Exception:  # noqa: BLE001
+        log.debug("Unable to detect gnome color scheme with dbus.")
+
+    # Fallback to subprocess method (legacy support)
     value = gsettings_get("color-scheme")
     if value and "dark" in value.lower():
         log.debug("Detected GNOME color-scheme: dark")
@@ -127,7 +143,16 @@ def detect_lxqt_dark_theme() -> bool:
 
 def detect_freedesktop_color_scheme_dark() -> bool:
     """Detect dark mode using org.freedesktop.appearance.color-scheme (XDG portal, cross-desktop)."""
-    # Try org.freedesktop.appearance first
+    # Try D-Bus first (secure method)
+    try:
+        detector = get_dbus_detector()
+        result = detector.detect_freedesktop_portal_color_scheme()
+        if result is not None:
+            return result
+    except Exception:  # noqa: BLE001
+        log.debug("Unable to detect `freedesktop` color scheme with dbus.")
+
+    # Fallback to subprocess method (legacy support)
     try:
         result = subprocess.run(  # nosec B603 B607
             [
@@ -201,6 +226,10 @@ def detect_lxqt_dark_wrapper() -> bool:
 def get_linux_dark_mode_strategies() -> list:
     """Return the list of dark mode detection strategies in order of priority."""
     return [
+        # Pure D-Bus methods (will gracefully fail if D-Bus unavailable)
+        detect_freedesktop_color_scheme_dbus,
+        detect_gnome_color_scheme_dbus,
+        # Hybrid methods (D-Bus with subprocess fallback)
         detect_freedesktop_color_scheme_dark,
         detect_gnome_dark_wrapper,
         detect_kde_dark_wrapper,
--- a/picard/ui/theme_detect_qtdbus.py
+++ b/picard/ui/theme_detect_qtdbus.py
@@ -1,0 +1,191 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2019-2022, 2024-2025 Philipp Wolfer
+# Copyright (C) 2020-2021 Gabriel Ferreira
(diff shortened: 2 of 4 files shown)


picard/ui/theme_detect.py
```diff
@@ -127,7 +143,16 @@ def detect_lxqt_dark_theme() -> bool:
...
     """Detect dark mode using org.freedesktop.appearance.color-scheme (XDG portal, cross-desktop)."""
-    # Try org.freedesktop.appearance first
+    # Try D-Bus first (secure method)
+    try:
+        detector = get_dbus_detector()
+        result = detector.detect_freedesktop_portal_color_scheme()
+        if result is not None:
+            return result
+    except Exception:  # noqa: BLE001
+        log.debug("Unable to detect `freedesktop` color scheme with dbus.")
```
rdswift: This largely duplicates the code from Lines 60-67.  If there was a third test using similar code, I would suggest some refactoring, but it's likely not worth it in this case.

picard/ui/theme_detect_qtdbus.py
```diff
@@ -0,0 +1,191 @@
...
+            return False
+        else:
+            return service_name in services
+
+    def detect_freedesktop_portal_color_scheme(self) -> bool | None:
+        """
+        Detect color scheme using org.freedesktop.portal.Settings interface.
+        Returns
+        -------
+            True for dark theme, False for light theme, None if unavailable
```
zas, line 97: According to https://flatpak.github.io/xdg-desktop-portal/docs/doc-org.freedesktop.portal.Settings.html, values are:

        0: No preference
        1: Prefer dark appearance
        2: Prefer light appearance


But below we have:

```python
            if value == 1:
                return True
            if value == 2:
                return False
```

And this comment says " True for dark theme, False for light theme, None if unavailable" -> value = 1 = dark

~~Isn't there an error here? Do I miss something?~~  -> Nope, it is ok, as is, my bad.

Also, if this method returns a boolean, it would be more convenient to have it named something like `freedesktop_portal_color_scheme_is_dark()`.

knguyen1: Renamed the funcs: https://github.com/metabrainz/picard/pull/2692/commits/e6c287e85a6a784b59200000ba7200140a341073

picard/ui/theme_detect.py
knguyen1: Refactored for DRY and removed blind exception catch: https://github.com/metabrainz/picard/pull/2692/commits/e6c287e85a6a784b59200000ba7200140a341073

picard/ui/theme_detect_qtdbus.py
```diff
@@ -0,0 +1,191 @@
...
+        """
+        try:
+            if not self.gnome_interface or not self.gnome_interface.isValid():
+                return None
+            # Get the color-scheme property from org.gnome.desktop.interface using dconf
+            reply = self.gnome_interface.call("Read", "/org/gnome/desktop/interface/color-scheme")
+
+            if reply.type() != QDBusMessage.MessageType.ErrorMessage:
+                value = reply.arguments()[0] if reply.arguments() else None
+                if value and isinstance(value, str) and "dark" in value.lower():
```
zas: Aren't those lines equivalent to:

```python
if value:
    return isinstance(value, str) and "dark" in value.lower()
```

phw: That would read nicer, yes

picard/ui/theme_detect.py
```diff
@@ -201,6 +242,10 @@ def detect_lxqt_dark_wrapper() -> bool:
 def get_linux_dark_mode_strategies() -> list:
     """Return the list of dark mode detection strategies in order of priority."""
     return [
+        # Pure D-Bus methods (will gracefully fail if D-Bus unavailable)
+        detect_freedesktop_color_scheme_dbus,
+        detect_gnome_color_scheme_dbus,
```
phw, line 247: One thing we could optimize is that we don't need to initialize `DBusThemeDetector` twice. It handles both interfaces already.

So either we add a single dbus strategy, that checks both the freedesktop and gnome interfaces, or we initialize `DBusThemeDetector` first and have the detect methods accept it as a parameter (and use partial to apply it here).

knguyen1: So actually, this pattern already uses a singleton.  `DBusThemeDetector` is only ever instantiated once.

picard\ui\theme_detect_qtdbus.py:160-169
```
# Global D-Bus detector instance
_dbus_detector = None


def get_dbus_detector() -> DBusThemeDetector:
    """Get or create the global D-Bus theme detector instance."""
    global _dbus_detector
    if _dbus_detector is None:
        _dbus_detector = DBusThemeDetector()
    return _dbus_detector
```

picard/ui/theme_detect_qtdbus.py
knguyen1: If block simplified: https://github.com/metabrainz/picard/pull/2692/commits/3eafa9e21ab86a4bbeabb999402c0acd075ba15a

phw (approved):

zas:
@knguyen1 PR looks good to me, please resolve conflicts

knguyen1:
> @knguyen1 PR looks good to me, please resolve conflicts

Resolved.

PR #2693, opened 2025-08-01
PICARD-2300: Add qt's stylehints to theme.py for a consistent exp. across platforms

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
  Added Qt6 style hints support to theme system and made appearance menu consistent across all platforms by enabling DEFAULT, LIGHT, and DARK themes everywhere except Haiku.

# Problem

The theme system was inconsistent across platforms - Linux only had DEFAULT and SYSTEM themes while Windows/macOS had DEFAULT, LIGHT, and DARK. Qt6's style hints provide a better way to handle dark themes with automatic fallback to manual color application.

# Solution

Added new style hints methods to `BaseTheme`:
- `_get_style_hints()`: Gets QGuiApplication.styleHints()
- `_set_color_scheme()`: Sets color scheme using style hints
- `_apply_dark_palette_colors()`: Manual dark color application
- `_apply_dark_theme_to_palette()`: Uses Qt color scheme or manual fallback

Updated `AVAILABLE_UI_THEMES` to include DEFAULT, LIGHT, and DARK on all platforms except Haiku, making the appearance menu consistent.

Enhanced setup logic to use Qt's built-in color scheme when available, with automatic fallback to manual color application.

# Action

Additional actions required:
* [x] Update Picard [documentation](https://github.com/metabrainz/picard-docs) (See PICARD-3103)

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -224,10 +224,12 @@ class BaseTheme:
             self._accent_color = palette.color(QtGui.QPalette.ColorGroup.Active, QtGui.QPalette.ColorRole.Highlight)
 
         # Linux-specific: If DEFAULT theme, try to detect system dark mode
-        # Do not apply override if already dark theme
+        # Do not apply override if already dark theme, or if a subclass already
+        # determined a dark theme state (e.g., WindowsTheme via registry on tests).
         is_dark_theme = self.is_dark_theme
         if (
             not self._dark_theme
+            and not is_dark_theme
             and not IS_WIN
             and not IS_MACOS
             and not IS_HAIKU
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -109,12 +109,8 @@ def _style_hints_available() -> bool:
 # Theme availability based on platform capabilities
 if IS_HAIKU:
     # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
-    # Platform Detection: `IS_HAIKU` is detected via sys.platform == 'haiku1'
-    # Feature Flag: `OS_SUPPORTS_THEMES` is set to False for Haiku
-    # Empty Options: `AVAILABLE_UI_THEMES` is set to an empty list [] for Haiku
-    # UI Hiding: The ui_theme_container widget is hidden when `OS_SUPPORTS_THEMES` is False (see `interface.py`)
     AVAILABLE_UI_THEMES = []
-elif not IS_WIN and not IS_MACOS:  # Linux
+elif not IS_WIN and not IS_MACOS:  # Non-Windows and non-macOS platforms
     if _style_hints_available():
         # All themes available when style hints are supported
         AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
@@ -122,7 +118,7 @@ elif not IS_WIN and not IS_MACOS:  # Linux
         # Only DEFAULT theme available when style hints are not available
         AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
 else:
-    # Windows and macOS: consistent structure
+    # All others, like Windows and macOS: consistent structure
     AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
 
 
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -86,7 +86,6 @@ class UiTheme(Enum):
     DEFAULT = 'default'
     DARK = 'dark'
     LIGHT = 'light'
-    SYSTEM = 'system'
 
     def __str__(self):
         return self.value
@@ -197,8 +196,8 @@ class BaseTheme:
         self._loaded_config_theme = UiTheme(config.setting['ui_theme'])
 
         # Use the new fusion style from PyQt6 for a modern and consistent look
-        # across all OSes, except when using system default theme on Linux.
-        if not IS_MACOS and not IS_HAIKU and not (not IS_WIN and self._loaded_config_theme == UiTheme.DEFAULT):
+        # across all OSes, except for macOS and Haiku.
+        if not IS_MACOS and not IS_HAIKU:
             app.setStyle('Fusion')
         elif IS_MACOS:
             app.setStyle(MacOverrideStyle(app.style()))
@@ -215,7 +214,7 @@ class BaseTheme:
             elif self._loaded_config_theme == UiTheme.LIGHT:
                 set_color_scheme(QtCore.Qt.ColorScheme.Light)
             else:
-                # For DEFAULT and SYSTEM themes, let Qt follow system settings
+                # For DEFAULT theme, let Qt follow system settings
                 set_color_scheme(QtCore.Qt.ColorScheme.Unknown)
 
         palette = QtGui.QPalette(app.palette())
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -109,16 +109,11 @@ def _style_hints_available() -> bool:
 if IS_HAIKU:
     # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
     AVAILABLE_UI_THEMES = []
-elif not IS_WIN and not IS_MACOS:  # Non-Windows and non-macOS platforms
-    if _style_hints_available():
-        # All themes available when style hints are supported
-        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
-    else:
-        # Only DEFAULT theme available when style hints are not available
-        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
-else:
-    # All others, like Windows and macOS: consistent structure
+elif IS_WIN or IS_MACOS or _style_hints_available():
     AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
+else:
+    # Use only default theme on platforms without style hints
+    AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
 
 
 class MacOverrideStyle(QtWidgets.QProxyStyle):
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -96,17 +96,33 @@ class UiTheme(Enum):
         return cls.DEFAULT
 
 
+def get_style_hints() -> QtGui.QStyleHints | None:
+    """Get style hints from QGuiApplication, returning None if unavailable."""
+    return QtGui.QGuiApplication.styleHints()
+
+
+def _style_hints_available() -> bool:
+    """Check if style hints are available on the current system."""
+    return get_style_hints() is not None
+
+
 # Theme availability based on platform capabilities
 if IS_HAIKU:
     # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
     # Platform Detection: `IS_HAIKU` is detected via sys.platform == 'haiku1'
     # Feature Flag: `OS_SUPPORTS_THEMES` is set to False for Haiku
-    # Empty Options: `AVAILABLE_UI_THEMES`` is set to an empty list [] for Haiku
+    # Empty Options: `AVAILABLE_UI_THEMES` is set to an empty list [] for Haiku
     # UI Hiding: The ui_theme_container widget is hidden when `OS_SUPPORTS_THEMES` is False (see `interface.py`)
     AVAILABLE_UI_THEMES = []
+elif not IS_WIN and not IS_MACOS:  # Linux
+    if _style_hints_available():
+        # All themes available when style hints are supported
+        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
+    else:
+        # Only DEFAULT theme available when style hints are not available
+        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
 else:
-    # All other platforms: consistent structure
-    # It is now safe to add these to linux since `QtGui.QGuiApplication.styleHints().setColorScheme()` is available
+    # Windows and macOS: consistent structure
     AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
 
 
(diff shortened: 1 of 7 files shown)


picard/ui/theme.py
```diff
@@ -96,11 +96,18 @@ def _missing_(cls, value):
...
+    # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
+    # Platform Detection: `IS_HAIKU` is detected via sys.platform == 'haiku1'
+    # Feature Flag: `OS_SUPPORTS_THEMES` is set to False for Haiku
+    # Empty Options: `AVAILABLE_UI_THEMES`` is set to an empty list [] for Haiku
+    # UI Hiding: The ui_theme_container widget is hidden when `OS_SUPPORTS_THEMES` is False (see `interface.py`)
+    AVAILABLE_UI_THEMES = []
+else:
+    # All other platforms: consistent structure
+    # It is now safe to add these to linux since `QtGui.QGuiApplication.styleHints().setColorScheme()` is available
+    AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
```
knguyen1, line 113: Actually, it think light and dark should only be available when `getStylehints` (see line 145) is not `null`. Will fix.

rdswift (commented):
The code looks fine to me.

It appears that this will change the choices available for the theme setting on Linux systems.  As such, the documentation should be updated -- specifically the "**User interface color theme**" section on the [User Interface Options](https://picard-docs.musicbrainz.org/en/config/options_interface.html) page.  I'll add a sub-task ticket so that this isn't forgotten (see PICARD-3103).

picard/ui/theme.py
```diff
@@ -96,11 +96,34 @@ def _missing_(cls, value):
...
+
+# Theme availability based on platform capabilities
+if IS_HAIKU:
+    # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
+    # Platform Detection: `IS_HAIKU` is detected via sys.platform == 'haiku1'
+    # Feature Flag: `OS_SUPPORTS_THEMES` is set to False for Haiku
+    # Empty Options: `AVAILABLE_UI_THEMES` is set to an empty list [] for Haiku
+    # UI Hiding: The ui_theme_container widget is hidden when `OS_SUPPORTS_THEMES` is False (see `interface.py`)
+    AVAILABLE_UI_THEMES = []
+elif not IS_WIN and not IS_MACOS:  # Linux
```
zas: Use `IS_LINUX` here?

picard/ui/theme.py
```diff
@@ -96,11 +96,34 @@ def _missing_(cls, value):
...
+
+def _style_hints_available() -> bool:
+    """Check if style hints are available on the current system."""
+    return get_style_hints() is not None
+
+
+# Theme availability based on platform capabilities
+if IS_HAIKU:
+    # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
+    # Platform Detection: `IS_HAIKU` is detected via sys.platform == 'haiku1'
```
phw: I think this line comment is not needed here. It is the same for all our `IS_{someos}` constants, which are there to avoid all the specific `sys.platform` tests.

picard/ui/theme.py
phw: No, I think it's better this way. It is actually any platform except for win and mac (and haiku, because handled separately). So this includes not only Linux, but also the BSDs for example.

phw (changes requested):
This is so great! The code looks fine to me overall. Some minor comments, otherwise I'd just like to test this live before merging.

picard/ui/theme.py
zas: Then the comment is a bit misleading, but nvm.

picard/ui/theme.py
knguyen1: That was for provenance; I needed to make sure that themes are indeed hidden for `IS_HAIKU`.  But I'll remove the comment.

knguyen1: Removed/clarified comments: https://github.com/metabrainz/picard/pull/2693/commits/a2d25762df5e10a74396613856c020884eab4726

picard/ui/theme.py
knguyen1: Removed/clarified comments: https://github.com/metabrainz/picard/pull/2693/commits/a2d25762df5e10a74396613856c020884eab4726

phw (approved):

zas:
@knguyen1 PR looks good to me, please resolve conflicts

phw (changes requested):
I tested this on Linux on Gnome. Unfortunately the style hint `setColorScheme` does not seem to have an effect there. It uses whatever I have set system wide.

That also means the "Light" and "Dark" theme settings in Picard don't have the desired effect.

`apply_dark_theme_to_palette` is only run, if the setting is on `UiTheme.DEFAULT`. Also `apply_dark_theme_to_palette` doesn't change the palette, if style hint support is detected. So setting the style to "Dark" does not run.

On the opposite if the system theme is already dark, but I set the Picard setting to `LIGHT` we currently have no way to do this. As the style hint again has no effect, but we also have no logic of explicitly setting a light palette, the app stays dark.

Is anyone currently able to test under a different Linux setup, e.g. Plasma desktop?

knguyen1:
Failing linting step from the GitHub outage. I filed an issue https://github.com/astral-sh/ruff-action/issues/198

phw:
@knguyen1 @zas How shall we proceed here? In general I think this is looking good, and we definitely should move forward with the new APIs provides by Qt6. We will have even some potential to further simplify the code in the future with less differences between platforms, I think. But this PR is overall a good starting point.

Just my comment above shows that Linux stays complicated. As I wrote it does not work for me on my Gnome setup. I currently don't have my KDE Neon VM available to run, so I haven't checked there. Do you guys have done some better results on Linux?

knguyen1:
@phw It's working for me on `gnome`.  Here's a screencast: https://1drv.ms/v/c/0e7c7882e6da112b/EZ5VziUuchBGl03_05izHUIB-MWMCV1kZO-YoVgQ5-S1rQ?e=1m00QB  I tested various permutations light, dark, default following sys theme, etc.

I suspect you needed to have `qt5` \ `qt6` installed

```bash
apt list --installed | grep -i 'qt5\|qt6'

libqt5core5t64/noble,now 5.15.13+dfsg-1ubuntu1 amd64 [installed,automatic]
libqt5dbus5t64/noble,now 5.15.13+dfsg-1ubuntu1 amd64 [installed,automatic]
...
python3-pyqt5.sip/noble,now 12.13.0-1build3 amd64 [installed,automatic]
python3-pyqt5/noble,now 5.15.10+dfsg-1build6 amd64 [installed,automatic]
qt5-gtk-platformtheme/noble,now 5.15.13+dfsg-1ubuntu1 amd64 [installed,automatic]
````

Install with `sudo apt install qt6-base-dev`.  On linux, it's a reasonable expectation for people to have; if not it can be installed.  I would update the docs and be very clear about this little nuance.

Maybe file a bug with `QTBUG` as well?

**Comment by zas** [comment]
On Linux Mint 22 + Cinnamon, if I switch theme to Light, while having a dark theme, save, restart Picard, it stays on dark theme, no way to force it to Light. So it doesn't look to work for me. I'll do more testing later.

Also since it changes option values we need an upgrade hook (I think).

**Comment by knguyen1** [comment]
> On Linux Mint 22 + Cinnamon, if I switch theme to Light, while having a dark theme, save, restart Picard, it stays on dark theme, no way to force it to Light. So it doesn't look to work for me. I'll do more testing later.

Can you try after installing `sudo apt install qt6-base-dev`?

**Comment by phw** [comment]
I have been trying with the PyQt6 / Qt6 from PyPI only. I'm on Ubuntu 24.04, and the Qt6 package provided there is too old (6.4). We have 6.5 defined as minimum, and the `ColorScheme` support makes this more important again.

Anyway, that might as well be the reason. We had the experience already in the past that the PyPI version did not integrate that well with the platform, while the distribution packages provided a much better experience. On Qt5 e.g. you could install qgnomeplatform for a better integration.

Maybe also the `qt6-xdgdesktopportal-platformtheme` package has some impact?

Anyway, if it generally works this is already much of an improvement. A lot of Qt6 is still progressing. I'd vote for getting this PR merged and improve as needed.

UPDATE: Weirdly it works fine for me today with PyQt6 from PyPI. I'm not aware that anything changed since my last attempt, but theme overrides in Picard work both when I have the light and dark theme configured in Gnome settings.

File: picard/ui/theme.py [hunk 4]
```diff
@@ -137,8 +195,8 @@ def setup(self, app):
         self._loaded_config_theme = UiTheme(config.setting['ui_theme'])
 
         # Use the new fusion style from PyQt6 for a modern and consistent look
-        # across all OSes.
-        if not IS_MACOS and not IS_HAIKU and self._loaded_config_theme != UiTheme.SYSTEM:
+        # across all OSes, except when using system default theme on Linux.
+        if not IS_MACOS and not IS_HAIKU and not (not IS_WIN and self._loaded_config_theme == UiTheme.DEFAULT):
```
phw: I think this change is wrong. We do want to set Fusion style by default. We had the special case of allowing "SYSTEM", which would apply whatever Qt6 style was configured on the system, to just suddenly switch to Fusion if you explicitly request DARK or LIGHT.

Currently this change changes behavior. DEFAULT not only adapts to system color, but will also change the Qt6b theme . 

The changes of this PR already effectively remove SYSTEM from all available configurations. But it is still indirectly being considered. I think we should completely remove this from the UiTheme enum.

See my proposed changes in https://github.com/knguyen1/picard/pull/1

zas:
> sudo apt install qt6-base-dev

No change.

Though it seems it sets theme to Light, it doesn't actually switch to light theme.

```
D: 15:27:36,625 tagger._log_startup:314: Starting Picard from '/home/zas/src/picard/picard/tagger.py'
D: 15:27:36,626 tagger._log_startup:315: Platform: Linux-6.8.0-71-generic-x86_64-with-glibc2.39 CPython 3.12.3
D: 15:27:36,634 tagger._log_startup:317: Versions: Picard 3.0.0.dev6, Python 3.12.3, PyQt 6.9.0, Qt 6.9.0, Mutagen 1.47.0, Discid discid 1.2.0, libdiscid 0.6.4, astrcmp C, SSL OpenSSL 3.0.13 30 Jan 2024
D: 15:27:36,635 tagger._log_startup:318: Configuration file path: '/home/zas/.config/MusicBrainz/Picard.ini'
D: 15:27:36,636 tagger._log_startup:320: User directory: '/home/zas/.config/MusicBrainz/Picard'
D: 15:27:36,636 tagger._log_startup:321: System long path support: True
D: 15:27:36,637 tagger._log_startup:324: Qt Env.: QT_SCALE_FACTOR='1.3' QT_ACCESSIBILITY='1' QT_AUTO_SCREEN_SCALE_FACTOR='0'
D: 15:27:36,652 ui/theme.setup:254: Theme: light (BaseTheme) dark=False accent_color=#ffc5a07c
```

phw:
@knguyen1 sorry, I pushed my changes from my PR accidentally directly to your branch with my latest commit. Please double check, we can undo stuff if it turns out to be a problem.

zas:
> @knguyen1 sorry, I pushed my changes from my PR accidentally directly to your branch with my latest commit. Please double check, we can undo stuff if it turns out to be a problem.

There's a leftover in the current branch:

```
picard/ui/options/interface.py:180:            if new_theme_setting == str(UiTheme.SYSTEM):
```

knguyen1:
Haha, no problem.  I'm about to go to the airport for a trip.  But will check when possible...

knguyen1:
> > sudo apt install qt6-base-dev
> 
> 
> 
> No change.
> 
> 
> 
> Though it seems it sets theme to Light, it doesn't actually switch to light theme.
> 
> 
> 
> ```
> 
> D: 15:27:36,625 tagger._log_startup:314: Starting Picard from '/home/zas/src/picard/picard/tagger.py'
> 
> D: 15:27:36,626 tagger._log_startup:315: Platform: Linux-6.8.0-71-generic-x86_64-with-glibc2.39 CPython 3.12.3
> 
> D: 15:27:36,634 tagger._log_startup:317: Versions: Picard 3.0.0.dev6, Python 3.12.3, PyQt 6.9.0, Qt 6.9.0, Mutagen 1.47.0, Discid discid 1.2.0, libdiscid 0.6.4, astrcmp C, SSL OpenSSL 3.0.13 30 Jan 2024
> 
> D: 15:27:36,635 tagger._log_startup:318: Configuration file path: '/home/zas/.config/MusicBrainz/Picard.ini'
> 
> D: 15:27:36,636 tagger._log_startup:320: User directory: '/home/zas/.config/MusicBrainz/Picard'
> 
> D: 15:27:36,636 tagger._log_startup:321: System long path support: True
> 
> D: 15:27:36,637 tagger._log_startup:324: Qt Env.: QT_SCALE_FACTOR='1.3' QT_ACCESSIBILITY='1' QT_AUTO_SCREEN_SCALE_FACTOR='0'
> 
> D: 15:27:36,652 ui/theme.setup:254: Theme: light (BaseTheme) dark=False accent_color=#ffc5a07c
> 
> ```

What about `qt5-base-dev`? I don't know why the diff would be. I have it on Gnome and it works.

In regard to the PR, I'm of the opinion that if the code on our part is technically correct, it's the qt api's fault and we should file a bug with QTBUG.

phw (approved):
I'm fine with the current state in this PR. I'd appreciate some review of my specific changes, but otherwise this did currently work for me and the usage of the APIs as it is makes sense.

I assume the situation on Linux to stay a bit difficult for a while. We already made the experience that the various Qt 6 versions have different issues, and distributions like e.g. Ubuntu 24.04 unfortunately ship older versions. But yes, if we have specific error case we should report it.

picard/ui/theme.py
```diff
@@ -96,11 +95,30 @@ def _missing_(cls, value):
...
+elif not IS_WIN and not IS_MACOS:  # Non-Windows and non-macOS platforms
+    if _style_hints_available():
+        # All themes available when style hints are supported
+        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
+    else:
+        # Only DEFAULT theme available when style hints are not available
+        AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
+else:
+    # All others, like Windows and macOS: consistent structure
+    AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
```
zas, line 113: Isn't that equivalent to:

```python
if IS_HAIKU:
    # Haiku doesn't support themes - UI is hidden anyway, but keep empty for consistency
    AVAILABLE_UI_THEMES = []
elif IS_WIN or IS_MACOS or _style_hints_available():
    AVAILABLE_UI_THEMES = [UiTheme.DEFAULT, UiTheme.LIGHT, UiTheme.DARK]
else:
    AVAILABLE_UI_THEMES = [UiTheme.DEFAULT]
```

zas (dismissed):
LGTM

picard/ui/theme.py
phw: Yes, you are right, this should be the same. The code has clearly grown from various conditions. I'll fix this part and then merge directly.

phw: Fixed in 604940dcbaa359eecd5f867d2045182dbb9138bc

phw (approved):

PR #2696, opened 2025-08-02
PICARD-442: Add match quality column and allow sorting

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds a new "Match Quality" column to the album view that displays match statistics using visual icons and text, showing matched/total tracks, missing tracks, duplicates, and extra files.

# Problem

* JIRA ticket (_optional_): PICARD-442

Users need a quick visual way to assess the quality of their file-to-track matches in albums. Currently, users must manually inspect each track to understand match status, which is time-consuming and error-prone.

# Solution

- Added `MatchQualityColumn` class that displays match statistics at the album level
- Implemented custom delegate (`MatchQualityColumnDelegate`) for rendering icons and text
- Column shows format: "matched/total; missing; duplicates; extra; unmatched"
- Uses existing match icons from FileItem for visual consistency
- Added re-sort trigger in Album._finalize_loading() to ensure accurate sorting after album loads
- Column integrates with existing item views system and respects sorting/filtering

image

# Action

Additional actions required:
* [x] Update Picard [documentation](https://github.com/metabrainz/picard-docs) (PICARD-3102)

```

### DIFF
--- a/picard/album.py
+++ b/picard/album.py
@@ -594,7 +594,11 @@ class Album(MetadataItem):
 
         self.tagger.window.set_statusbar_message(
             N_('Album %(id)s loaded: %(artist)s - %(album)s'),
-            {'id': self.id, 'artist': self.metadata['albumartist'], 'album': self.metadata['album']},
+            {
+                'id': self.id,
+                'artist': self.metadata['albumartist'],
+                'album': self.metadata['album'],
+            },
             timeout=3000,
         )
         for func, _run_on_error in self._after_load_callbacks:
@@ -646,7 +650,10 @@ class Album(MetadataItem):
         if self._requests:
             log.info("Not reloading, some requests are still active.")
             return
-        self.tagger.window.set_statusbar_message(N_("Loading album %(id)s …"), {'id': self.id})
+        self.tagger.window.set_statusbar_message(
+            N_("Loading album %(id)s …"),
+            {'id': self.id},
+        )
         self.loaded = False
         self.status = AlbumStatus.LOADING
         if self.release_group:
--- a/picard/album.py
+++ b/picard/album.py
@@ -578,13 +578,23 @@ class Album(MetadataItem):
             self.match_files(unmatched_files + self.unmatched_files.files)
         self.update_metadata_images()
         self.update()
+
+        # Trigger re-sort after album is fully loaded to ensure accurate match quality sorting
+        # See the module: picard/ui/itemviews/match_quality_column.py
+        # Without this, the sorting calcs never retrigger
+        if self.ui_item:
+            tree_widget = self.ui_item.treeWidget()
+            if tree_widget and tree_widget.isSortingEnabled():
+                # Clear cached sort keys for this item to force recalculation
+                self.ui_item._sortkeys.clear()
+                # Trigger re-sort by calling sortByColumn with current sort column
+                current_sort_column = tree_widget.sortColumn()
+                if current_sort_column >= 0:
+                    tree_widget.sortByColumn(current_sort_column, tree_widget.header().sortIndicatorOrder())
+
         self.tagger.window.set_statusbar_message(
             N_('Album %(id)s loaded: %(artist)s - %(album)s'),
-            {
-                'id': self.id,
-                'artist': self.metadata['albumartist'],
-                'album': self.metadata['album'],
-            },
+            {'id': self.id, 'artist': self.metadata['albumartist'], 'album': self.metadata['album']},
             timeout=3000,
         )
         for func, _run_on_error in self._after_load_callbacks:
@@ -636,10 +646,7 @@ class Album(MetadataItem):
         if self._requests:
             log.info("Not reloading, some requests are still active.")
             return
-        self.tagger.window.set_statusbar_message(
-            N_("Loading album %(id)s …"),
-            {'id': self.id},
-        )
+        self.tagger.window.set_statusbar_message(N_("Loading album %(id)s …"), {'id': self.id})
         self.loaded = False
         self.status = AlbumStatus.LOADING
         if self.release_group:
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -91,6 +91,7 @@ from picard.util import (
 from picard.ui.collectionmenu import CollectionMenu
 from picard.ui.enums import MainAction
 from picard.ui.filter import Filter
+from picard.ui.itemviews.match_quality_column import MatchQualityColumn, MatchQualityColumnDelegate
 from picard.ui.ratingwidget import RatingWidget
 from picard.ui.scriptsmenu import ScriptsMenu
 from picard.ui.util import menu_builder
@@ -409,6 +410,13 @@ class BaseTreeView(QtWidgets.QTreeWidget):
     def _init_header(self):
         header = ConfigurableColumnsHeader(self.columns, parent=self)
         self.setHeader(header)
+
+        # Set up delegate for progress columns
+        progress_delegate = MatchQualityColumnDelegate(self)
+        for i, column in enumerate(self.columns):
+            if isinstance(column, MatchQualityColumn):
+                self.setItemDelegateForColumn(i, progress_delegate)
+
         self.restore_default_columns()
         self.restore_state()
 
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -45,6 +45,7 @@
 
 from PyQt6 import QtCore
 
+from picard.album import AlbumStatus
 from picard.i18n import N_
 from picard.util import icontheme
 
@@ -81,10 +82,16 @@ def _sortkey_match_quality(obj):
     """Sort key for match quality column - sort by completion percentage."""
     if hasattr(obj, 'get_num_matched_tracks') and hasattr(obj, 'tracks'):
         # Album object
+        # Check if album is still loading - if so, return 0 to avoid premature sorting
+        if hasattr(obj, 'status') and obj.status == AlbumStatus.LOADING:
+            return 0.0
+
         matched = obj.get_num_matched_tracks()
         total = len(obj.tracks) if obj.tracks else 0
         if total > 0:
-            return matched / total
+            # Return negative percentage so that higher percentages (closer to 1.0)
+            # appear first in ascending sort order
+            return -(matched / total)
         return 0.0
     # For track objects, return 0 since we don't show icons at track level
     return 0.0
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -46,6 +46,7 @@
 from PyQt6 import QtCore
 
 from picard.album import AlbumStatus
+from picard.const.sys import IS_LINUX
 from picard.i18n import N_
 from picard.util import icontheme
 
@@ -89,7 +90,9 @@ def _sortkey_match_quality(obj):
         matched = obj.get_num_matched_tracks()
         total = len(obj.tracks) if obj.tracks else 0
         if total > 0:
-            return matched / total
+            # Column sorting is reversed on Linux
+            multiplier = -1 if IS_LINUX else 1
+            return matched / total * multiplier
         return 0.0
     # For track objects, return 0 since we don't show icons at track level
     return 0.0
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -193,8 +193,6 @@ _common_columns = (
 FILEVIEW_COLUMNS = Columns(_common_columns, default_width=100)
 
 # Album view columns (with match quality column)
-# Insert `_match_quality_column` at index 4, after Title, Length, Artist, Album Artist
-ALBUMVIEW_COLUMNS = Columns(
-    _common_columns[:4] + (_match_quality_column,) + _common_columns[4:],
-    default_width=100,
-)
+# Insert `_match_quality_column` after Title, Length, Artist, Album Artist
+ALBUMVIEW_COLUMNS = Columns(_common_columns, default_width=100)
+ALBUMVIEW_COLUMNS.insert(ALBUMVIEW_COLUMNS.pos('albumartist') + 1, _match_quality_column)
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -89,9 +89,7 @@ def _sortkey_match_quality(obj):
         matched = obj.get_num_matched_tracks()
         total = len(obj.tracks) if obj.tracks else 0
         if total > 0:
-            # Return negative percentage so that higher percentages (closer to 1.0)
-            # appear first in ascending sort order
-            return -(matched / total)
+            return matched / total
         return 0.0
     # For track objects, return 0 since we don't show icons at track level
     return 0.0
@@ -134,13 +132,15 @@ _fingerprint_column = IconColumn(N_("Fingerprint status"), '~fingerprint')
 _fingerprint_column.header_icon_func = lambda: icontheme.lookup('fingerprint-gray', icontheme.ICON_SIZE_MENU)
 _fingerprint_column.set_header_icon_size(16, 16, 1)
 
-_match_quality_column = MatchQualityColumn(N_("Match Quality"), '~match_quality', width=100)
+_match_quality_column = MatchQualityColumn(N_("Match"), '~match_quality', width=57)
 _match_quality_column.sortable = True
 _match_quality_column.sort_type = ColumnSortType.SORTKEY
 _match_quality_column.sortkey = _sortkey_match_quality
+_match_quality_column.always_visible = True
 
 ITEMVIEW_COLUMNS = Columns(
     (
+        _match_quality_column,
         DefaultColumn(
             N_("Title"), 'title', sort_type=ColumnSortType.NAT, width=250, always_visible=True, status_icon=True
         ),
@@ -180,7 +180,6 @@ ITEMVIEW_COLUMNS = Columns(
         ),
         Column(N_("Genre"), 'genre'),
         _fingerprint_column,
-        _match_quality_column,
         Column(N_("Date"), 'date'),
         Column(N_("Original Release Date"), 'originaldate'),
         Column(N_("Release Date"), 'releasedate'),
--- a/picard/ui/itemviews/columns.py
+++ b/picard/ui/itemviews/columns.py
@@ -56,6 +56,7 @@ from picard.ui.columns import (
     DefaultColumn,
     ImageColumn,
 )
+from picard.ui.itemviews.match_quality_column import MatchQualityColumn
 
 
 def _sortkey_length(obj):
@@ -74,6 +75,19 @@ def _sortkey_bitrate(obj):
         return float(obj.metadata['~bitrate'] or obj.orig_metadata['~bitrate'] or 0)
     except (ValueError, TypeError):
         return 0
+
+
+def _sortkey_match_quality(obj):
+    """Sort key for match quality column - sort by completion percentage."""
+    if hasattr(obj, 'get_num_matched_tracks') and hasattr(obj, 'tracks'):
+        # Album object
+        matched = obj.get_num_matched_tracks()
+        total = len(obj.tracks) if obj.tracks else 0
+        if total > 0:
+            return matched / total
+        return 0.0
+    # For track objects, return 0 since we don't show icons at track level
+    return 0.0
 
 
 class IconColumn(ImageColumn):
@@ -113,6 +127,10 @@ _fingerprint_column = IconColumn(N_("Fingerprint status"), '~fingerprint')
 _fingerprint_column.header_icon_func = lambda: icontheme.lookup('fingerprint-gray', icontheme.ICON_SIZE_MENU)
 _fingerprint_column.set_header_icon_size(16, 16, 1)
 
+_match_quality_column = MatchQualityColumn(N_("Match Quality"), '~match_quality', width=100)
+_match_quality_column.sortable = True
+_match_quality_column.sort_type = ColumnSortType.SORTKEY
+_match_quality_column.sortkey = _sortkey_match_quality
 
 ITEMVIEW_COLUMNS = Columns(
     (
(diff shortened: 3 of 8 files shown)

### CODE REVIEW

**Comment by knguyen1** [comment]
FYI.. changes unrelated to the feature is due to `ruff format`.

File: picard/ui/itemviews/basetreeview.py [hunk 1]
```diff
@@ -91,6 +91,7 @@
 from picard.ui.collectionmenu import CollectionMenu
 from picard.ui.enums import MainAction
 from picard.ui.filter import Filter
+from picard.ui.itemviews.match_quality_column import MatchQualityColumn, MatchQualityColumnDelegate
```
rdswift, line 94: Is this the new style for `import` formatting because of ruff?  Wasn't our standard to format multiple imports as a tuple with each import (tuple item) on a separate line?

rdswift (commented):
The code looks fine to me.

For clarification, I assume this new column is intended to augment (not replace) the current Album / Release status icon.  In that case, it might be better positioned in a column adjacent to the current Album / Release status icon.

My other concern is with the width of this column -- always a minimum of 200 pixels (if I read the code correctly).  Since the details are available for display as a tooltip on this new Match Quality icon, perhaps we can just display the icon and not the summary numbers (which can be viewed from the tooltip).  This would save some of the limited available display space.

Finally, since this changes the standard display for the user, I believe that the documentation should be updated (at least the [Album / Release Icons](https://picard-docs.musicbrainz.org/en/getting_started/status_icons.html#album-release-icons) section, and possibly the images on the [Main Screen](https://picard-docs.musicbrainz.org/en/getting_started/screen_main.html) page).  I'll add a sub-task ticket for this so it doesn't get forgotten.

picard/ui/itemviews/basetreeview.py
knguyen1: Yes.  This is from the `pre-commit` hook `ruff format`, the rule is governed by:

pyproject.toml:157-157
```
line-length = 120
```

This particularly line is 100 chars long, so `ruff/black` will prefer to keep it on one line.

knguyen1:
> For clarification, I assume this new column is intended to augment (not replace) the current Album / Release status icon. In that case, it might be better positioned in a column adjacent to the current Album / Release status icon.

Need to make some decisions here re: `match/total; missing; duplicates; extra; unmatched` details in the col.  My preference is only the icon.

- Display the full details (`match/total; missing; duplicates; extra; unmatched`) in the column?

image

- Only matched/total?

image

- Only the icon?

image

Also `matched/total` is available in the `Title` column already anyway.  So have to make a decision there too...

And if we decide on only the icon, I suggest we rename the column -> `Match`.

rdswift:
> My preference is only the icon.

I think that would be my preference as well, since all the details are available via the tooltip.

> And if we decide on only the icon, I suggest we rename the column -> `Match`.

That would help reduce the space required.  Would it be possible to leave out the column name completely (and only have the sort direction indicator if required)?  I'm thinking that it being next to the status icon might make it clear that it is the quality of the match.  Then again, it might not be clear to other users.

picard/ui/itemviews/basetreeview.py
rdswift: Yes, but there was an isort rule `force_grid_wrap = 2` (along with other formatting rules) in the `pyproject.toml` file that differ from the current ruff configuration.  Thus the reason for my question, because the `include` style is now different.

knguyen1: Ruff prefers `black` style (can't change it). That's another thing noticed that was missing from the previous `isort` configuration.

Anyway that's a `pyproject.toml`/formatting change, unrelated to this PR.

rdswift: > Ruff prefers `black` style (can't change it).

I see that, and it appears that the ruff development team have no intention of supporting isort formatting.  That's unfortunate because the way that we used to format the imports made a lot more sense (in my opinion) when reviewing git diffs.

> Anyway that's a `pyproject.toml`/formatting change, unrelated to this PR.

Agreed.  It's just that this PR happens to be where I noticed the changes.

zas: I think we should review changes (in a separate PR) after running `ruff format`. I just did a test on current master branch: 273 files changed, 5013 insertions(+), 4555 deletions(-)

We need to ensure format is correct and no issue is created by running `ruff format`.
Also we need to update documentation about coding style if we go this way.

I would apply those changes first, and get rid of stylistic changes in PRs. I mean if we agree to use `black` style, I'm not totally happy with it style-wise, but `ruff format` is very convenient and fast.

@phw @rdswift @knguyen1 What do you think?

zas:
> My preference is only the icon.

Mine too, as all infos are available elsewhere (tooltip and/or title).

picard/ui/itemviews/basetreeview.py
phw: I generally agree with @zas. I'm also not really happy with the black style. The changing of import style based on line length is rather annoying and can lead to unnecessary and hard to read changed lines in diffs rather randomly. We had quite lengthy discussions about this when we introduced isort managing the imports, and the current system of having multiple imports on separate lines worked really well.

But if we really can't change it I would prefer the convenience of `ruff format` over my personal taste.

We should do a complete formatting run in a separate PR and check the result. Maybe there is more changes that we need to discuss and probably tweak one way or another.

rdswift: @phw has summarized my thoughts exactly.

phw:
> My preference is only the icon.

> And if we decide on only the icon, I suggest we rename the column -> Match.

+1 for both of those from me, too

knguyen1:
> > My preference is only the icon.
> 
> > And if we decide on only the icon, I suggest we rename the column -> Match.
> 
> +1 for both of those from me, too

This commit: https://github.com/metabrainz/picard/pull/2696/commits/3f7260f20774c9dcee663529c1e568aa71d169ba

- Pull column to left, next to Title/Length/Artist
- Rename column -> `Match`
- Column always shows
- Remove text from column
- Column sort order seems to be different on linux vs. Windows (will confirm)

image

picard/ui/itemviews/basetreeview.py
knguyen1: > I think we should review changes (in a separate PR) after running ruff format. I just did a test on current master branch: 273 files changed, 5013 insertions(+), 4555 deletions(-)

> We need to ensure format is correct and no issue is created by running ruff format.

Yes.  I noted this in this PR comment: https://github.com/metabrainz/picard/pull/2690#issuecomment-3114430963

I also suggest the same: We need a massive `ruff format` refactor so that (a) we can add `ruff format --check` to the CI pipelines and (b) so that we can avoid seeing so much non-PR related formatting noise in future PR's.

knguyen1:
> Column sort order seems to be different on linux vs. Windows (will confirm)

Just tested on Linux and confirmed.  The sorting order icon seems reversed on linux.  Commit: https://github.com/metabrainz/picard/pull/2696/commits/3039a01b417c49647d0f39388caa78221755934b

picard/ui/itemviews/basetreeview.py
zas: I ran `ruff format` on current master branch, and created a PR for us to review changes: https://github.com/metabrainz/picard/pull/2697

zas:
Having `Match` column for middle pane is a bit weird.

image

zas:
@knguyen1 Overall the PR looks good, but UI wise there's likely few improvements we can make. Anyway, please resolve conflicts.

phw:
I think the match column on the right works fine. But as zas pointed out it would be great if we could have no Match column, just the title, on the middle pane.

knguyen1:
Rebased and...

- Split columns: introduced _common_columns, FILEVIEW_COLUMNS (no match column) and ALBUMVIEW_COLUMNS (includes match column).
- Match column placement: _match_quality_column is now in ALBUMVIEW_COLUMNS after albumartist (index 4).
- Removed legacy: deleted ITEMVIEW_COLUMNS and its usage.
- View wiring: MainPanel now creates FileTreeView(FILEVIEW_COLUMNS) and AlbumTreeView(ALBUMVIEW_COLUMNS) instead of a single shared set.
- Dynamic item columns: TreeItem.columns changed from a static class attribute to a @property that returns treeWidget().columns; with a safe fallback to FILEVIEW_COLUMNS when not yet attached (prevents construction-time crashes).

If you've previously tested the code with `Match` all the way to the left, you'll need to **temporarily** disable `restore_state`.

```python
# line 421 of picard\ui\itemviews\basetreeview.py
self.restore_state()  # <-- temporarily comment this out here
```

image

phw:
> If you've previously tested the code with `Match` all the way to the left, you'll need to **temporarily** disable `restore_state`.

Do we maybe need a migration that resets the column state?

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,243 @@
...
+        # Count files per track to detect duplicates
+        track_file_counts = {}
+        for track in obj.tracks:
+            track_file_counts[track] = len(track.files)
+
+        # Count unmatched files
+        unmatched_files = obj.unmatched_files.files if hasattr(obj, "unmatched_files") else []
+
+        # Calculate duplicates (tracks with more than one file)
+        for _, file_count in track_file_counts.items():
```
zas: Please don't use `_` (it is used for gettext, as `_()`):

```suggestion
        for file_count in track_file_counts.values():
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,243 @@
...
+    def paint(self, painter, rect):
+        """Override paint method to prevent NotImplementedError from base class."""
+        # This column is painted by the delegate, not the header
+        # So we just do nothing here to prevent the error
+        pass
+
+    def get_match_icon(self, obj):
+        """Get the appropriate match icon for the given object."""
+        # Only show icons at the release (album) level, not track level
+        if not hasattr(obj, "get_num_matched_tracks") or not hasattr(obj, "tracks"):
```
zas: Just a minor thing, but we used to use single quotes for strings that are keys (and double quotes for textual strings, translatable strings):
 ```suggestion
         if not hasattr(obj, 'get_num_matched_tracks') or not hasattr(obj, 'tracks'):
```

File: picard/ui/itemviews/match_quality_column.py [hunk 4]
```diff
@@ -0,0 +1,243 @@
...
+
+        # Album object
+        matched = obj.get_num_matched_tracks()
+        total = len(obj.tracks) if obj.tracks else 0
+
+        if total == 0:
+            # Use pending icon for zero tracks
+            from picard.ui.itemviews import FileItem
+
+            if hasattr(FileItem, "match_pending_icons") and len(FileItem.match_pending_icons) > 5:
```
zas:
```suggestion
            if hasattr(FileItem, 'match_pending_icons') and len(FileItem.match_pending_icons) > 5:
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,243 @@
...
+        elif percentage >= 0.9:
+            icon_index = 4  # 90% match
+        elif percentage >= 0.8:
+            icon_index = 3  # 80% match
+        elif percentage >= 0.7:
+            icon_index = 2  # 70% match
+        elif percentage >= 0.6:
+            icon_index = 1  # 60% match
+        elif percentage >= 0.5:
+            icon_index = 0  # 50% match
```
zas: What about something like:

```python
percent_match_icon = {
  1.0: 5,
  0.9: 4,
  0.8: 3,
  0.7: 2,
  0.6: 1,
  0.0: 0,
}

for threshold, i in percent_match_icon.items():
    if percentage >= threshold:
        icon_index = i
        break

```

knguyen1:
> Do we maybe need a migration that resets the column state?

We don't in my opinion. It affects people (us devs) who had a col of 85px width there. It causes the Title col to shrink to 85px.

picard/ui/itemviews/match_quality_column.py
knguyen1: See: https://github.com/metabrainz/picard/pull/2696/commits/2a77bdafa653adef0adf196f41cb08a6ba2762d6

picard/ui/itemviews/columns.py
```diff
@@ -76,6 +79,25 @@ def _sortkey_bitrate(obj):
...
 
+def _sortkey_match_quality(obj):
+    """Sort key for match quality column - sort by completion percentage."""
+    if hasattr(obj, 'get_num_matched_tracks') and hasattr(obj, 'tracks'):
+        # Album object
+        # Check if album is still loading - if so, return 0 to avoid premature sorting
+        if hasattr(obj, 'status') and obj.status == AlbumStatus.LOADING:
+            return 0.0
+
+        matched = obj.get_num_matched_tracks()
```
zas: This line can be moved to `if` branch below, as it is used only there.

picard/ui/itemviews/columns.py
```diff
@@ -113,53 +135,66 @@ def paint(self, painter, rect):
...
+)
+
+
+# File view columns (without match quality column)
+FILEVIEW_COLUMNS = Columns(_common_columns, default_width=100)
+
+# Album view columns (with match quality column)
+# Insert `_match_quality_column` at index 4, after Title, Length, Artist, Album Artist
+ALBUMVIEW_COLUMNS = Columns(
+    _common_columns[:4] + (_match_quality_column,) + _common_columns[4:],
```
zas: We can get rid of hard-coded number here:

```python
# Album view columns (with match quality column)
ALBUMVIEW_COLUMNS = Columns(_common_columns, default_width=100)
# Insert `_match_quality_column` after Title, Length, Artist, Album Artist
ALBUMVIEW_COLUMNS.insert(ALBUMVIEW_COLUMNS.pos('albumartist'), _match_quality_column)
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+        # Determine which icon to use based on percentage using thresholds
+        for threshold in sorted(THRESHOLD_TO_ICON_INDEX.keys(), reverse=True):
+            if percentage >= threshold:
+                icon_index = THRESHOLD_TO_ICON_INDEX[threshold]
+                break
+
+        # Get the match icons from FileItem
+        from picard.ui.itemviews import FileItem
+
+        if hasattr(FileItem, "match_icons") and icon_index < len(FileItem.match_icons):
```
zas:
```suggestion
        if hasattr(FileItem, 'match_icons') and icon_index < len(FileItem.match_icons):
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+            return FileItem.match_icons[icon_index]
+
+        return None
+
+    def get_match_stats(self, obj):
+        """Get comprehensive match statistics for the given object."""
+        # Only show stats at the release (album) level, not track level
+        if not hasattr(obj, "get_num_matched_tracks") or not hasattr(obj, "tracks"):
```
zas:
```suggestion
        if not hasattr(obj, 'get_num_matched_tracks') or not hasattr(obj, 'tracks'):
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+        extra = 0
+        missing = 0
+
+        # Count files per track to detect duplicates
+        track_file_counts = {}
+        for track in obj.tracks:
+            track_file_counts[track] = len(track.files)
+
+        # Count unmatched files
+        unmatched_files = obj.unmatched_files.files if hasattr(obj, "unmatched_files") else []
```
zas:
```suggestion
        unmatched_files = obj.unmatched_files.files if hasattr(obj, 'unmatched_files') else []
```

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+        if unmatched_files:
+            extra = len(unmatched_files)
+
+        # Calculate missing tracks (tracks with no files)
+        for track in obj.tracks:
+            if len(track.files) == 0:
+                missing += 1
+
+        return {
+            "matched": matched,
```
zas: Same here, keys, use single quotes for keys.

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+    """Delegate for rendering match quality icons in tree items."""
+
+    def __init__(self, parent=None):
+        super().__init__(parent)
+
+    def _get_item_data(self, index):
+        """Extract item data and validate it's a match quality column.
+
+        Returns:
+            tuple: (obj, column, stats) or (None, None, None) if validation fails
```
zas: That's a good candidate for a NamedTuple.

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+
+        # Core match info
+        if stats["total"] > 0:
+            percentage = (stats["matched"] / stats["total"]) * 100
+            tooltip_parts.append(f"Match: {stats['matched']}/{stats['total']} ({percentage:.1f}%)")
+        else:
+            tooltip_parts.append("Match: 0/0 (0%)")
+
+        # Additional stats with explanations
+        tooltip_parts.append(f"Missing tracks: {stats['missing']}")
```
zas: This should be translatable, as is, it isn't.

picard/ui/itemviews/match_quality_column.py
```diff
@@ -0,0 +1,246 @@
...
+        # Always draw icon, regardless of column width
+        if icon:
+            x = option.rect.x() + icon_margin
+            y = option.rect.y() + (option.rect.height() - icon_size.height()) // 2
+            icon.paint(painter, QtCore.QRect(x, y, icon_size.width(), icon_size.height()))
+
+    def helpEvent(self, event, view, option, index):
+        """Show tooltip with explanation of the stats."""
+        # Get item data
+        _, __, stats = self._get_item_data(index)
```
zas: Do not use `_`, if `_get_item_data()` returns a named tuple then no need for that.

zas (changes requested):

picard/ui/itemviews/match_quality_column.py
knguyen1: Let me see if there is a `ruff` rule for this. Or a pre-commit. This is getting to be quite a lot of effort to have to keep track of.

knguyen1: We don't want this right?

```
# Configure single quotes
[tool.ruff.format]
quote-style = 'single'
```

zas: I don't think so , if it's too complicated ignore this, it's just we used to differentiate keys (as strings) from others

picard/ui/itemviews/columns.py
knguyen1: Moved the if block: https://github.com/metabrainz/picard/pull/2696/commits/e831e1b17deea83f9b697ec2887e93b2722c0c18

picard/ui/itemviews/columns.py
knguyen1: Not hardcoding col index: https://github.com/metabrainz/picard/pull/2696/commits/e831e1b17deea83f9b697ec2887e93b2722c0c18

picard/ui/itemviews/match_quality_column.py
knguyen1: See: https://github.com/metabrainz/picard/pull/2696/commits/8508e563341f40e1f9cf91020d48243083b73f25

picard/ui/itemviews/match_quality_column.py
knguyen1: See: https://github.com/metabrainz/picard/pull/2696/commits/8508e563341f40e1f9cf91020d48243083b73f25

picard/ui/itemviews/match_quality_column.py
knguyen1: See: https://github.com/metabrainz/picard/pull/2696/commits/8508e563341f40e1f9cf91020d48243083b73f25

picard/ui/itemviews/match_quality_column.py
knguyen1: See: https://github.com/metabrainz/picard/pull/2696/commits/8508e563341f40e1f9cf91020d48243083b73f25

zas:
Overall looks good, I noticed we can't hide the Match column in context menu though.

image

knguyen1:
Okay. Can make it hideable.

knguyen1:
> Overall looks good, I noticed we can't hide the Match column in context menu though.

Okay can make it hideable. this commit: https://github.com/metabrainz/picard/pull/2696/commits/7fae0ede816e786dc492de21d078508efbdff0f4

picard/ui/itemviews/columns.py
zas: It wasn't it seems

zas (changes requested):
Just a minor change missed, but the whole patch looks good to me.

picard/ui/itemviews/columns.py
knguyen1: There were two.  Here's the other one: https://github.com/metabrainz/picard/pull/2696/commits/ed555828d55651609893f7ba735f62a511a631a1

picard/album.py
```diff
@@ -578,6 +578,20 @@ def _finalize_loading_album(self):
             self.match_files(unmatched_files + self.unmatched_files.files)
         self.update_metadata_images()
         self.update()
+
+        # Trigger re-sort after album is fully loaded to ensure accurate match quality sorting
+        # See the module: picard/ui/itemviews/match_quality_column.py
```
zas, line 583: Just a minor thing, but this kind of comment may not age well, we may move files around in the future (or rename them). Using a permalink is a bit better (well, even though they use github domain, there are easier to find and replace if we move from it).

Aerozol:
This looks very cool, but does it solve PICARD-442? Which is explicitly about _sorting by album completion_ (e.g. gold disc etc).

The problem and solution in this ticket seems to address something else.
(I'm here because a user asked about sorting by completion and I wanted to link them this PR, but I'm not sure anymore)

PR #2697, opened 2025-08-05
Run `ruff format`

zas:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/installer/i18n/nshutil.py
+++ b/installer/i18n/nshutil.py
@@ -112,7 +112,9 @@ ESCAPE_CHARS = {
     r'$\`': '`',
 }
 
-RE_LANGSTRING_LINE = re.compile(r'LangString\s+(?P<identifier>[A-Za-z0-9_]+)\s+\${LANG_[A-Z]+}\s+["\'`](?P<text>.*)["\'`]$')
+RE_LANGSTRING_LINE = re.compile(
+    r'LangString\s+(?P<identifier>[A-Za-z0-9_]+)\s+\${LANG_[A-Z]+}\s+["\'`](?P<text>.*)["\'`]$'
+)
 
 
 def language_to_code(language):
@@ -139,10 +141,7 @@ def unescape_string(text):
 def parse_langstring(line):
     match_ = RE_LANGSTRING_LINE.match(line)
     if match_:
-        return (
-            match_.group('identifier'),
-            unescape_string(match_.group('text'))
-        )
+        return (match_.group('identifier'), unescape_string(match_.group('text')))
     else:
         return None
 
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -55,8 +55,7 @@ PICARD_VERSION_STR = str(PICARD_VERSION)
 PICARD_VERSION_STR_SHORT = PICARD_VERSION.short_str()
 if PICARD_BUILD_VERSION_STR:
     __version__ = "%s+%s" % (PICARD_VERSION_STR, PICARD_BUILD_VERSION_STR)
-    PICARD_FANCY_VERSION_STR = "%s (%s)" % (PICARD_VERSION_STR_SHORT,
-                                            PICARD_BUILD_VERSION_STR)
+    PICARD_FANCY_VERSION_STR = "%s (%s)" % (PICARD_VERSION_STR_SHORT, PICARD_BUILD_VERSION_STR)
 else:
     __version__ = PICARD_VERSION_STR_SHORT
     PICARD_FANCY_VERSION_STR = PICARD_VERSION_STR_SHORT
@@ -83,6 +82,7 @@ def crash_handler(exc: Exception = None):
     # with minimum chance to fail.
     from tempfile import NamedTemporaryFile
     import traceback
+
     if exc:
         if sys.version_info < (3, 10):
             trace_list = traceback.format_exception(None, exc, exc.__traceback__)
@@ -111,6 +111,7 @@ def crash_handler(exc: Exception = None):
         QApplication,
         QMessageBox,
     )
+
     app = QCoreApplication.instance()
     if not app:
         app = QApplication(sys.argv)
@@ -120,12 +121,13 @@ def crash_handler(exc: Exception = None):
     msgbox.setTextFormat(Qt.TextFormat.RichText)
     msgbox.setText(
         'An unexpected error has caused Picard to crash. '
-        'Please report this issue on the <a href="https://tickets.metabrainz.org/projects/PICARD">MusicBrainz bug tracker</a>.')
+        'Please report this issue on the <a href="https://tickets.metabrainz.org/projects/PICARD">MusicBrainz bug tracker</a>.'
+    )
     if logfile:
         logfile_url = QUrl.fromLocalFile(logfile)
         msgbox.setInformativeText(
-            'A logfile has been written to <a href="{0}">{1}</a>.'
-            .format(logfile_url.url(), logfile))
+            'A logfile has been written to <a href="{0}">{1}</a>.'.format(logfile_url.url(), logfile)
+        )
     msgbox.setDetailedText(trace)
     msgbox.setStandardButtons(QMessageBox.StandardButton.Close)
     msgbox.setDefaultButton(QMessageBox.StandardButton.Close)
@@ -136,6 +138,7 @@ def crash_handler(exc: Exception = None):
 def register_excepthook():
     def _global_exception_handler(exctype, value, traceback):
         from picard import crash_handler
+
         crash_handler(exc=value)
         sys.__excepthook__(exctype, value, traceback)
 
--- a/picard/const/cover_processing.py
+++ b/picard/const/cover_processing.py
@@ -27,10 +27,10 @@ from picard.i18n import N_
 
 
 class ResizeModes(IntEnum):
-    MAINTAIN_ASPECT_RATIO = 0,
-    SCALE_TO_WIDTH = 1,
-    SCALE_TO_HEIGHT = 2,
-    CROP_TO_FIT = 3,
+    MAINTAIN_ASPECT_RATIO = (0,)
+    SCALE_TO_WIDTH = (1,)
+    SCALE_TO_HEIGHT = (2,)
+    CROP_TO_FIT = (3,)
     STRETCH_TO_FIT = 4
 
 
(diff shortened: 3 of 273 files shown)


picard/const/cover_processing.py
```diff
@@ -27,10 +27,10 @@
 
 
 class ResizeModes(IntEnum):
-    MAINTAIN_ASPECT_RATIO = 0,
-    SCALE_TO_WIDTH = 1,
-    SCALE_TO_HEIGHT = 2,
-    CROP_TO_FIT = 3,
+    MAINTAIN_ASPECT_RATIO = (0,)
```
zas: Weird changes

picard/formats/apev2.py
```diff
@@ -176,7 +173,7 @@ def _load(self, filename):
                         if value.endswith(')'):
                             start = value.rfind(' (')
                             if start > 0:
-                                name += ':' + value[start + 2:-1]
+                                name += ':' + value[start + 2 : -1]
```
zas, line 182: Not fan of [a : b ] vs [a:b] but ok.

picard/formats/vorbis.py
```diff
@@ -481,13 +485,7 @@ def OggVideoFile(filename):
...
 def OggContainerFile(filename):
     """Generic Ogg file."""
-    options = [
-        OggFLACFile,
-        OggOpusFile,
-        OggSpeexFile,
-        OggTheoraFile,
-        OggVorbisFile
-    ]
+    options = [OggFLACFile, OggOpusFile, OggSpeexFile, OggTheoraFile, OggVorbisFile]
```
zas: Typical change I'm not fond of.

picard/script/functions.py
```diff
@@ -70,11 +70,14 @@ def _compute_logic(operation, *args):
     return operation(args)
 
 
-@script_function(eval_args=False, documentation=N_(
-    """`$if(if,then,else)`
+@script_function(
+    eval_args=False,
+    documentation=N_(
```
zas, line 75: Does it break translations?

picard/const/cover_processing.py
knguyen1: It's not weird. It's bad input code. Enums should not have trailing commas.  Or maybe original author really meant to use tuples...?

```
class Foo(Enum)
    MY_ENUM = 1,  # creates a tuple
```

picard/formats/vorbis.py
knguyen1: You can add a trailing commas `ruff` will respect it.

Eg

```
ONE = 1
TWO = 2
THREE = 3

numbers = [ONE, TWO, THREE,]
#     trailing comma here ^
```

Will turn into

```
numbers = [
    ONE,
    TWO,
    THREE,
]
```

But yeah. This is typical of Black style.

zas:
Overall it doesn't look too bad, I'm for doing this move, being able to run `ruff format` will be a nice improvement to maintain code style, especially for new contributors.
If we do it, that's now, major version changes, not too many PRs queued

picard/formats/vorbis.py
rdswift: Will this also work with `include` statements?  (I suspect not, but I haven't tried.)  For example, would:

```python
from a import (b, c, d,)
```

be saved as:

```python
from a import (
    b,
    c,
    d,
)
```

and would this remain as such throughout future runs of `ruff format`?  If so, this may partly address my concern with the import formatting.

rdswift:
> Overall it doesn't look too bad, I'm for doing this move, being able to run `ruff format` will be a nice improvement to maintain code style, especially for new contributors.

Although I'm not really a fan of the black style (personal preference), there are some definite advantages to being able to run `ruff format` as you noted.

> If we do it, that's now, major version changes, not too many PRs queued

Agreed.  Now would be the best time (before any version 3.0 releases).  It basically comes down to doing it now and then having to rebase the outstanding PRs, or to merge them first and then do a full conversion (reformatting) before any release.

picard/formats/vorbis.py
knguyen1: Yes. Just confirmed. 

```
from module_a import (a, d, c,)
# trailing AND as a tuple    ^
```

Prints as

```
from module_a import (
    a,
    b,
    c,
)
```

rdswift: Excellent!  Thanks for testing and confirming.

installer/i18n/nshutil.py
```diff
@@ -139,10 +141,7 @@ def unescape_string(text):
 def parse_langstring(line):
     match_ = RE_LANGSTRING_LINE.match(line)
     if match_:
-        return (
-            match_.group('identifier'),
-            unescape_string(match_.group('text'))
-        )
+        return (match_.group('identifier'), unescape_string(match_.group('text')))
```
rdswift: For stuff like this, if we want to retain the separate line formatting should we be adding the trailing commas and re-run `ruff format` before merging?

picard/__init__.py
```diff
@@ -120,12 +121,13 @@ def crash_handler(exc: Exception = None):
...
-        'Please report this issue on the <a href="https://tickets.metabrainz.org/projects/PICARD">MusicBrainz bug tracker</a>.')
+        'Please report this issue on the <a href="https://tickets.metabrainz.org/projects/PICARD">MusicBrainz bug tracker</a>.'
+    )
     if logfile:
         logfile_url = QUrl.fromLocalFile(logfile)
         msgbox.setInformativeText(
-            'A logfile has been written to <a href="{0}">{1}</a>.'
-            .format(logfile_url.url(), logfile))
+            'A logfile has been written to <a href="{0}">{1}</a>.'.format(logfile_url.url(), logfile)
+        )
```
rdswift, line 130: If we're doing such a major reformatting, do we want to take the opportunity to change things like this to use f-strings?  (Of course, that doesn't apply to logging entries where replacements are included as separate arguments.)

picard/coverart/providers/provider.py
```diff
@@ -77,15 +77,15 @@ def title(cls):
...
+    `__init__()` does not have to do anything.
+    `queue_images()` will be called if `enabled()` returns `True`.
+    `queue_images()` must return `FINISHED` when it finished to queue
+    potential cover art downloads (using `queue_put(<CoverArtImage object>).
+    If `queue_images()` delegates the job of queuing downloads to another
+    method (asynchronous) it should return `WAIT` and the other method has to
+    explicitly call `next_in_queue()`.
+    If `FINISHED` is returned, `next_in_queue()` will be automatically called
+    by CoverArt object.
     """
```
rdswift, line 89: This looks odd.  Is there a way to retain indented lines in a docstring?  Would it format correctly if Line 79 were split into two lines as:

```python
    """
    Subclasses of this class need to reimplement at least `queue_images()`.
```

picard/formats/apev2.py
```diff
@@ -84,14 +84,12 @@ def is_valid_key(key):
 
     See http://wiki.hydrogenaud.io/index.php?title=APE_key
     """
-    return (key and 2 <= len(key) <= 255
-            and key not in DISALLOWED_KEYS
-            and INVALID_CHARS.search(key) is None)
+    return key and 2 <= len(key) <= 255 and key not in DISALLOWED_KEYS and INVALID_CHARS.search(key) is None
```
rdswift: Personally, I find the original easier to read and follow.

picard/formats/apev2.py
rdswift: I'm even less a fan of `[a :]` vs `[a:]` as in Line 151 of `picard/extension_points/script_functions.py` but that's just a personal preference.

picard/script/__init__.py
```diff
@@ -216,11 +218,11 @@ def preset_title(number, title):
...
-               "$if($gt(%totaldiscs%,1),%discnumber%-,)\n"
-               "$if($and(%albumartist%,%tracknumber%),$num(%tracknumber%,2) ,)\n"
-               "$if(%_multiartist%,%artist% - ,)\n"
-               "%title%",
+        "$if(%albumartist%,%album%/,)\n"
+        "$if($gt(%totaldiscs%,1),%discnumber%-,)\n"
+        "$if($and(%albumartist%,%tracknumber%),$num(%tracknumber%,2) ,)\n"
+        "$if(%_multiartist%,%artist% - ,)\n"
+        "%title%",
         author=AUTHOR,
```
rdswift: Would this retain the indented formatting if it was:

```python
    script=(
        "$if2(%albumartist%,%artist%)/\n"
        "$if(%albumartist%,%album%/,)\n"
        "$if($gt(%totaldiscs%,1),%discnumber%-,)\n"
        "$if($and(%albumartist%,%tracknumber%),$num(%tracknumber%,2) ,)\n"
        "$if(%_multiartist%,%artist% - ,)\n"
        "%title%"
    ),
    author=AUTHOR,
```

picard/ui/mainwindow/__init__.py
```diff
@@ -1436,24 +1439,29 @@ def _auto_update_check(self):
...
-            'last_check': str(datetime.date.fromordinal(last_update_check)) if last_update_check > 0 else 'never',
-            'check_interval': update_check_days,
-            'update_level': update_level,
-            'update_level_name': PROGRAM_UPDATE_LEVELS[update_level]['name'] if update_level in PROGRAM_UPDATE_LEVELS else 'unknown',
-        })
+        do_auto_update_check = (
+            check_for_updates and update_check_days > 0 and today >= last_update_check + update_check_days
+        )
+        log.debug(
+            "%(check_status)s startup check for program updates.  Today: %(today_date)s, Last check: %(last_check)s (Check interval: %(check_interval)s days), Update level: %(update_level)s (%(update_level_name)s)",
```
rdswift: Perhaps this long line could be split into smaller lines within brackets, something like:

```python
(
    "%(check_status)s startup check for program updates.  Today: %(today_date)s, Last check: %(last_check)s "
    "(Check interval: %(check_interval)s days), Update level: %(update_level)s (%(update_level_name)s)"
),
```

(Unless the original actually fits within the maximum line length.)

knguyen1:
> I wonder if we could retain much of our import formatting by running isort --fix before running ruff format

You can't.  Having both leads to inevitable conflicts.  See thread: https://github.com/astral-sh/ruff/issues/19532
Even `ruff` vs. `black` leads to conflicts (`ruff` only matches up with `black` 99.99%).

picard/__init__.py
knguyen1: You can add this rule to ruff.  Auto-fix is sometimes available: https://docs.astral.sh/ruff/rules/f-string/#f-string-up032

If not auto fixable, the rule will mark the violation until you manually fix it.

picard/formats/apev2.py
knguyen1: Both versions are terrible, imo.

```
if not key:
    return False
if not (2 <= len(key) <= 255):
    return False
if key in DISALLOWED_KEYS:
    return False
return INVALID_CHARS.search(key) is None
```

installer/i18n/nshutil.py
zas: Yes

rdswift:
> > I wonder if we could retain much of our import formatting by running isort --fix before running ruff format
> 
> You can't. Having both leads to inevitable conflicts. See thread: [astral-sh/ruff#19532](https://github.com/astral-sh/ruff/issues/19532)

I'm aware of that, but if the only difference is the addition/removal of blank lines then I still think there might be merit in my suggestion.  For example, a user contributes code that includes:

```python
from PyQt6 import QtCore, QtWidgets, QtGui
import sys, os
```

This is clearly not what we would like to see.  Running `isort --fix` (with our original isort configuration) would reformat this as:

```python
import os
import sys

from PyQt6 import (
    QtCore,
    QtGui,
    QtWidgets,
)

```

This makes more sense for reviewing changes, but it isn't valid in ruff.  If the file was then processed by `ruff format`, the number of blank lines would be adjusted, but the multiple imports would still be retained on separate lines (as is evidenced by `ruff format` not reformatting the imports in this PR other than blank line removal).  This provides a (generally) stable way of cleaning up imports to have them sorted and avoid multiple imports on the same line, while still providing a consistent formatted output prepared by ruff.  This way we use isort as an interim step to get the sorting and multi-line output before the final formatting by ruff.  I understand that this interim step isn't acceptable to ruff (because of the number of blank lines), but that is corrected by following up with `ruff format`.

The end result (after processing initially with `isort --fix` and then with `ruff format`) will yield consistent results acceptable to ruff but still retain most of the isort formatting which facilitates easier review of the changes.

As I said initially, I haven't thought this through (and I'll admit I'm not thoroughly familiar with ruff), but it seems that it might be a way to help automate the preferred formatting of imports to produce them in a (reproducable) sorted manner with no multiple iimports on a single line.  What am I missing?

installer/i18n/nshutil.py
rdswift: I'll start on it and send you a PR.  (https://github.com/zas/picard/pull/18 submitted.)

picard/ui/metadatabox/__init__.py
```diff
@@ -238,8 +241,7 @@ def _on_setting_changed(self, name, old_value, new_value):
             "rename_files",
             "selected_file_naming_script_id",
             "standardize_artists",
-            "user_profile_settings"
-            "user_profiles",
+            "user_profile_settingsuser_profiles",
             "va_name",
```
rdswift: I think this may be a bug, and it should actually be:

```python
            "user_profile_settings",
            "user_profiles",
```

installer/i18n/nshutil.py
knguyen1: There is a python tool for adding trailing commas automatically.  Also a pre-commit: https://github.com/asottile/add-trailing-comma Add it to `.pre-commit-config.yaml`

picard/ui/metadatabox/__init__.py
zas: Missing comma at end of line 241, good catch

picard/__init__.py
rdswift: I would be concerned that the auto-fix may end up breaking some of the translation stuff.  Those can be pretty finicky to get just right.  Unfortunately, "auto" doesn't always mean "correct".

picard/ui/metadatabox/__init__.py
rdswift: I created a new PR https://github.com/metabrainz/picard/pull/2698 for the fix.

picard/formats/apev2.py
rdswift: Updated in https://github.com/zas/picard/pull/18.

picard/__init__.py
zas: > opportunity to change things like this to use f-strings?

Not in auto, too risky imho at this point.

picard/coverart/providers/provider.py
zas: It looks ok to me, as all lines are indented. If only few lines were indented, the relative indentation would have been preserved.

picard/disc/whipperlog.py
```diff
@@ -35,8 +35,5 @@ def toc_from_file(path):
     results on other non-standard cases."""
     with open(path, encoding='utf-8') as f:
         data = yaml.safe_load(f)
-        toc_entries = (
-            TocEntry(num, t['Start sector'], t['End sector'])
-            for num, t in data['TOC'].items()
-        )
+        toc_entries = (TocEntry(num, t['Start sector'], t['End sector']) for num, t in data['TOC'].items())
```
zas, line 38: Is there a way to preserve the style (apart disabling formatting) for such expressions? Not critical IMHO, but I still find the orignal much easier to read.

picard/script/__init__.py
zas: Yes, done in https://github.com/metabrainz/picard/pull/2697/commits/44199a97138fd6cf529cdcdaae93b581928885e5

picard/ui/mainwindow/__init__.py
```diff
@@ -780,7 +788,7 @@ def add_toolbar_action(action_id):
         self.show_toolbar()
 
     def _create_player_toolbar(self):
-        """"Create a toolbar with internal player control elements"""
+        """ "Create a toolbar with internal player control elements"""
```
zas: Fixed in https://github.com/metabrainz/picard/pull/2697/commits/2f533f555ca3c269ff53ee9ba77066e3191fe38b

picard/ui/mainwindow/__init__.py
```diff
@@ -1229,14 +1248,7 @@ def _update_actions(self):
...
                 # Skip further loops if all values now True.
-                if (
-                    can_analyze
-                    and can_autotag
-                    and can_refresh
-                    and can_remove
-                    and can_save
-                    and can_submit
-                ):
+                if can_analyze and can_autotag and can_refresh and can_remove and can_save and can_submit:
```
zas, line 1251: Ok...

picard/ui/mainwindow/__init__.py
zas: Good idea, done in https://github.com/metabrainz/picard/pull/2697/commits/bc9dde2913d9b7a1d0709ecb748ee2a59ed39030

picard/disc/whipperlog.py
rdswift: I was going to fix those (and there are a lot of them) when I was adding the commas, but I couldn't find a way to get ruff to cooperate.

picard/script/__init__.py
rdswift: Note that this only works if the total length of the combined strings exceeds the maximum line length.  Otherwise the strings are still combined.

knguyen1:
Very nice changes, thanks.  Should reduce noise for incoming PR's.  Next step we should add `ruff format check` to CI pipelines.

PR #2701, opened 2025-08-12
PICARD-3085: Add more details to album art info box and tests

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
  Display cover art details (type, size in kB/KiB, dimensions, mime) directly in the cover art view below thumbnails and on hover, showing info for the first (front) image.

# Problem

* JIRA ticket (_optional_): PICARD-3085

Users must open “Show more details” to see image size/dimensions/mime, which slows bulk artwork review.

# Solution

- Update `CoverArtBox` to render a four-line detail block for the first image:
  - `{type}`, `{size kB} ({size KiB})`, `{W} x {H}`, `{mime}`
- Show the same details as tooltip on hover.
- Support both current (new) and original thumbnails; headers shown when both are present and differ.
- Add concise unit tests for formatting and label/tooltip updates; tests are xdist-safe.

image

# Action

Additional actions required:


phw:
More image details make sense. My problem with this is the vertical size this display uses. It's the one reason why I had been hesitant on adding any additional info below the images.

We already sometimes have complaints of the metadatabox height being too large. We had this pane collapsible in the past, which was bad because people lost it completely and did not figure out how to recover it. Suggested was also to make the cover art area scrollable, but I think that's horrible UI.

Things I think could work:

1. Making display of those additional details an option, so it can be disabled to get more vertical space.
2. Showing the image details in the tooltip. This could be done in addition to 1. also
3. Showing the images side by side. This would also match how other new/old metadata is shown. But it takes more space horizontally then.

rdswift:
@phw, I echo your concerns.

> Things I think could work:
> 
> 1. Making display of those additional details an option, so it can be disabled to get more vertical space.
> 2. Showing the image details in the tooltip. This could be done in addition to 1. also
> 3. Showing the images side by side. This would also match how other new/old metadata is shown. But it takes more space horizontally then.

If we're going to offer this, I think it should be optional for the user.  My thinking is Option 2, with Option 1 if deemed necessary.

A possible 4th option would be to superimpose the information on the image itself, but that brings with it a whole new set of potential issues.

knguyen1:
These options are available.

image

image

picard/ui/coverartbox/__init__.py
```diff
@@ -165,6 +227,65 @@ def update_metadata(self):
...
+        The incoming list is expected to be in exactly this order when available.
+        """
+        show_type = CoverArtBox.initialize_covertartbox_settings('show_cover_art_details_type', True)
+        show_size = CoverArtBox.initialize_covertartbox_settings('show_cover_art_details_filesize', True)
+        show_dims = CoverArtBox.initialize_covertartbox_settings('show_cover_art_details_dimensions', True)
+        show_mime = CoverArtBox.initialize_covertartbox_settings('show_cover_art_details_mimetype', True)
+
+        result: list[str] = []
+        # Index mapping: 0 type, 1 size, 2 dimensions, 3 mimetype (if present)
+        if len(lines) > 0 and show_type:
```
zas: It seems a bit hacky to me, what about deciding what to display in `_first_image_info_lines()` instead?

picard/ui/forms/ui_options_cover.py
```diff
@@ -115,11 +134,16 @@ def setupUi(self, CoverOptionsPage):
...
+        self.cb_show_cover_art_details_type.setText(_("Show type (front, back, etc.)"))
+        self.cb_show_cover_art_details_filesize.setText(_("Show file size"))
+        self.cb_show_cover_art_details_dimensions.setText(_("Show dimensions"))
+        self.cb_show_cover_art_details_mimetype.setText(_("Show MIME type"))
         self.save_images_to_tags.setTitle(_("Embed cover images into tags"))
         self.cb_embed_front_only.setText(_("Embed only a single front image"))
         self.cb_dont_replace_with_smaller.setText(_("Never replace cover images with smaller ones"))
         self.cb_never_replace_types.setText(_("Never replace cover images matching selected types"))
-        self.select_types_button.setText(_("Select types…"))
+        self.select_types_button.setText(_("Select Types..."))
```
zas: Unrelated and incorrect change

knguyen1: That's generated by `python setup.py build_ui`. My understanding is we're not to modify these files by hand.

zas: That's weird it made this change, it doesn't when I run this command here

phw: We should fix this separately. The issue was caused by me in https://github.com/metabrainz/picard/commit/3a7fec4bb3e7ad7d7445c25007bb18c3f7c6ddb4 . The problem here is that I searched for UI strings to unify, but didn't notice I change it in the ui_*.py files. Those should have been updated in the *.ui files as well.

To make things consistent I fixed the string in https://github.com/metabrainz/picard/commit/8fac12a3180295783aa1d2ae888eddcb5ed5e779 .

@knguyen1 Can you rebase?

knguyen1: Done.

picard/ui/coverartbox/__init__.py
knguyen1: Done. https://github.com/metabrainz/picard/pull/2701/commits/a90717e83c808ac0ddc4be34d57598d0e35aebef

zas:
> These options are available.
> image image

I would move those options to User Interface > Cover Art Box instead of having those under Cover Art (the pane is already pretty stuffed), after all that's UI-related options.

@phw @rdswift @knguyen1 What do you think?

rdswift:
> I would move those options to User Interface > Cover Art Box instead of having those under Cover Art (the pane is already pretty stuffed), after all that's UI-related options.

I agree completely.

picard/options.py
```diff
@@ -208,6 +208,11 @@
 BoolOption('setting', 'save_images_to_files', False, title=N_("Save cover images as separate files"))
 BoolOption('setting', 'save_images_to_tags', True, title=N_("Embed cover images into tags"))
 BoolOption('setting', 'save_only_one_front_image', False, title=N_("Save only a single front image as separate file"))
+BoolOption('setting', 'show_cover_art_details', False, title=N_("Show cover art details in cover art view"))
+BoolOption('setting', 'show_cover_art_details_type', True, title=N_("Show cover art type in info labels"))
```
phw: I would default this specific setting to `False`, as the cover art box would usually always show the front cover. But no strong opinion, just my view on what is the better default.

picard/ui/coverartbox/__init__.py
```diff
@@ -111,6 +128,24 @@ def __init__(self, parent=None):
     def show_cover_art_info(self):
         self.tagger.window.view_info(default_tab=1)
 
+    @staticmethod
+    def initialize_covertartbox_settings(name: str, default: bool) -> bool:
```
phw: I don't fully understand why this function is needed. The options should be defined, and the settings system already handles the defaults. We don't route reading settings over such functions elsewhere. What do I miss?

phw (changes requested):
I agree with @zas . The configuration should go to a Cover Art Box page below the User Interface settings. Apart from that the PR looks good to me. I just have a question / comment on the `initialize_covertartbox_settings` function.

picard/ui/coverartbox/__init__.py
knguyen1: Nice catch. It was because it was a work-around for an improperly mocked test https://github.com/metabrainz/picard/pull/2701/commits/b3496514f9f81b67e67967a6b349d0c879bc5c19

picard/options.py
knguyen1: Okay, good suggestion.

knguyen1: @phw Default to false: https://github.com/metabrainz/picard/pull/2701/commits/55d60d0c3d01f79ae8f9bf1d34bf20d1d7192b80

knguyen1:
@zas @phw 

- Remove cover art config: https://github.com/metabrainz/picard/pull/2701/commits/ec3d1555cffe5ac837dc5d88fdd12b2d208f45e9
- Add cover art interface config: https://github.com/metabrainz/picard/pull/2701/commits/aefb6b4f3b5ac1f1265e1b56932d03d14768fb88

picard/ui/coverartbox/__init__.py
```diff
@@ -262,7 +370,12 @@ def _try_load_remote_image(self, url, data):
             mode = CoverArtSetterMode.APPEND
 
         setter = CoverArtSetter(mode, coverartimage, self.item)
-        setter.set_coverart()
```
zas: Why is this change needed?

zas (changes requested):

picard/ui/coverartbox/__init__.py
knguyen1: `ty` returned the below error.  So I was wondering of ways to fix it.  I thought being explicit in the types reads better.  Anyway, if you think it's safe to ignore, I'll revert it.

> No argument provided for required parameter `self` of function `set_coverart`ty[missing-argument](https://ty.dev/rules#missing-argument)

> (bound method CoverArtSetter.set_coverart() -> Unknown) | (def set_coverart(self) -> Unknown)

image

zas: Clearly not needed, see https://github.com/metabrainz/picard/blob/c8cbe31866913baeaefd633252144bc8f18b8eba/picard/ui/coverartbox/coverartsetter.py#L60-L69

knguyen1: I'll revert.

At some point, I suggest a refactor of this.  A strategy pattern would have been great here. `ty` is highlighting the fact that dynamically resetting the method leads to confusion.

```python
class CoverArtSetter:
    """Context class that uses different strategies to set cover art."""
    
    # Strategy registry
    _strategies = {
        'Album': AlbumCoverArtStrategy(),
        'FileListItem': FileListCoverArtStrategy(),
        'File': FileCoverArtStrategy(),
    }
    
    def __init__(self, mode: CoverArtSetterMode, coverartimage, source_obj):
        self.mode = mode
        self.coverartimage = coverartimage
        self.source_obj = source_obj
        self.strategy = self._get_strategy(source_obj)
    
    def _get_strategy(self, source_obj) -> CoverArtStrategy:
        """Get the appropriate strategy for the source object."""
        class_name = source_obj.__class__.__name__
        strategy = self._strategies.get(class_name)
        
        if strategy is None:
            raise ValueError(f"No strategy found for object type: {class_name}")
        
        return strategy
    
    def set_coverart(self) -> bool:
        """Set cover art using the appropriate strategy."""
        return self.strategy.set_coverart(self.mode, self.coverartimage, self.source_obj)
```

knguyen1: Or, should we refactor now?  Since we're changing the module anyway...

zas: Since that's not strictly needed, don't mix refactoring & feature PRs. You can submit another PR with suggested refactoring.

knguyen1: Reverted: https://github.com/metabrainz/picard/pull/2701/commits/bbcf543d749e2d2eeec73754bd6db886eaafe8ef

zas (approved):
LGTM

PR #2702, opened 2025-08-13
PICARD-3011: Add ui quick toggles for cover art

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
  Add two quick toggles to Options menu: “Save images to tags” and “Save images to files”, mirroring existing cover art settings for fast on/off control without opening Options.

# Problem

Users frequently switch cover art saving behaviors (embed vs file) and want quick toggles akin to “Save Tags/Rename/Move” to avoid navigating the Options dialog.

* JIRA ticket (optional): PICARD-3011

# Solution

- Add actions to `MainAction`: `ENABLE_SAVE_IMAGES_TO_TAGS`, `ENABLE_SAVE_IMAGES_TO_FILES`.
- Implement actions in `picard/ui/mainwindow/actions.py` and handlers in `picard/ui/mainwindow/__init__.py`.
- Insert both under “Save Tags” in Options menu and keep checked state synced with `save_images_to_tags` and `save_images_to_files`.
- Focused tests in `test/test_ui_cover_quick_toggles.py` validate action state, triggers, and handler config updates (with Qt action mocked).

# Action

Additional actions required:
* [x] Update Picard documentation (add the two new quick toggles under Options)

### DIFF
--- a/picard/ui/enums.py
+++ b/picard/ui/enums.py
@@ -74,6 +74,8 @@ class MainAction(str, Enum):
     ENABLE_MOVING = 'enable_moving_action'
     ENABLE_RENAMING = 'enable_renaming_action'
     ENABLE_TAG_SAVING = 'enable_tag_saving_action'
+    ENABLE_SAVE_IMAGES_TO_TAGS = 'enable_save_images_to_tags_action'
+    ENABLE_SAVE_IMAGES_TO_FILES = 'enable_save_images_to_files_action'
     EXIT = 'exit_action'
     GENERATE_FINGERPRINTS = 'generate_fingerprints_action'
     HELP = 'help_action'
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -294,6 +294,10 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             self.actions[MainAction.ENABLE_MOVING].setChecked(new_value)
         elif name == 'dont_write_tags':
             self.actions[MainAction.ENABLE_TAG_SAVING].setChecked(not new_value)
+        elif name == 'save_images_to_tags':
+            self.actions[MainAction.ENABLE_SAVE_IMAGES_TO_TAGS].setChecked(new_value)
+        elif name == 'save_images_to_files':
+            self.actions[MainAction.ENABLE_SAVE_IMAGES_TO_FILES].setChecked(new_value)
         elif name in {'file_renaming_scripts', 'selected_file_naming_script_id'}:
             self._make_script_selector_menu()
 
@@ -617,11 +621,21 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         config = get_config()
         config.setting['dont_write_tags'] = not checked
 
+    def toggle_save_images_to_tags(self, checked):
+        config = get_config()
+        config.setting['save_images_to_tags'] = checked
+
+    def toggle_save_images_to_files(self, checked):
+        config = get_config()
+        config.setting['save_images_to_files'] = checked
+
     def _reset_option_menu_state(self):
         config = get_config()
         self.actions[MainAction.ENABLE_RENAMING].setChecked(config.setting['rename_files'])
         self.actions[MainAction.ENABLE_MOVING].setChecked(config.setting['move_files'])
         self.actions[MainAction.ENABLE_TAG_SAVING].setChecked(not config.setting['dont_write_tags'])
+        self.actions[MainAction.ENABLE_SAVE_IMAGES_TO_TAGS].setChecked(config.setting['save_images_to_tags'])
+        self.actions[MainAction.ENABLE_SAVE_IMAGES_TO_FILES].setChecked(config.setting['save_images_to_files'])
         self._make_script_selector_menu()
         self._init_cd_lookup_menu()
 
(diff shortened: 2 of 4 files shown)


rdswift:
I'm not against this, but isn't this sort of thing the reason we added the profiles and profile quick selector (enable/disable)?  That way we don't have to have quick selectors for **every** setting.

Aerozol:
I use profiles, but this situation is one where I toggle back and forth _so_ much while editing... much more than the three other quick toggles. I agree that it should be a very high threshold for something to be added there, though.

phw:
Same situation as @Aerozol for me as well.

zas:
> Looks good. I'd like to get @zas' opinion on this before merging, though

Like @rdswift I think that's a bit redundant with quick profile switching, though I understand why such feature is convenient.
I'm a bit worried about adding 2 new menu items that fit only certain use cases.
What about adding a way to select one or more boolean options in a profile and add them to this menu.

For example, in this profile we have:

image

Let's say I'm a user that needs to switch Windows Compatibility often.

Imagine a context menu, with item "Add to quick menu", available for any boolean option in the profile. When selected, the item is added the quick menu, until removed (by the same contextual menu, with item now "Remove from quick menu"). This way we get a totally configurable "quick" menu, and we still use profiles, so selecting another profile can have a different quick menu.
That's just an idea. IMHO current profile stuff is a bit hard to understand for new users, it's not very intuitive, so perhaps we need better in-app documentation, and more straight forward access to profile settings, dunno.

Overall, I'm bothered by the fact this patch adds very specific code, for only a few use cases (not everyone needs to toggle cover art saving,
I could live with this patch, but I would like we put more thoughts in this before merging.

What do you think about a more generic approach (quick access to ANY boolean options at least) ?

rdswift:
> That's just an idea.

I think that's a great idea.  It could provide maximum flexibility for even the most "niche" users.

> IMHO current profile stuff is a bit hard to understand for new users, it's not very intuitive, so perhaps we need better in-app documentation, and more straight forward access to profile settings, dunno.

I think that what makes the profile stuff difficult for most users to understand is the whole "layer" concept.  I suspect that they are more looking for something like the ability to select one profile from a set of (user defined) profiles with each profile containing the complete option settings, similar to the ability to set the configuration to use from the command line.  This is what I had in mind with my original profile settings code, but it was abandoned in favour of the layered concept (which I admit is a better model).  The layered method is quite powerful, flexible, and efficient both in terms of storage and not having to make the same change to multiple configurations.  Unfortunately, it isn't quite as intuitive as we had hoped.

Better in-app help regarding profiles is a good idea, but I'm not sure where to start or what to say.  I wonder if this is something that @Aerozol might be able to provide some insight?

knguyen1:
I'm not really familiar with the Profile system and have never used it.  Let me take some time to digest this.

rdswift:
> I'm not really familiar with the Profile system and have never used it. Let me take some time to digest this.

I tried to document it in the https://picard-docs.musicbrainz.org/en/config/options_profiles.html and https://picard-docs.musicbrainz.org/en/usage/option_profiles.html pages in the Picard User Guide.  Hopefully these will help.

Also, I didn't mean to hijack the discussion of this PR.  Sorry about that.

Aerozol:
What about having it behind a rollover? Then we can add more quick toggles, if they are ever requested, without bogging down the menu. I don't see a downside to this approach. Profiles are awesome, but they are not super simple (more on this below), this would allow for beginner (quick toggles) and advanced (profiles) to coexist.

Rollover illustration:
image

We always want to avoid extra clicks, so it is great that these submenus are rollover already.

---

I like @zas suggestion of allowing users to add things to this menu, but I worry that this may lead to another string of necessary complexity/menus/descriptions, and we end up with another setting more suited to expert users!

If we have something like the above 'more' menu, perhaps that allows for an 'add options' option at the bottom, which then gives a list to select from? I think it would be fine to only add 'popular' or requested options, and expect users to utitilise profiles for the full range of settings.

...in that case we should pre-populate the menu with a few pre-checked examples *cough* cover art toggles *cough* ;P

---

re. better onboarding of users into profiles, I just had another look at them and at @rdswift's documentation, and it seems fine to me.

I think it's one of those situations where adding more text/guidance/etc will just complicate things even more - some users simply won't be able to deal with the complexity no matter what (I am not implying that it is too complex, just that the threshold is very low for some users).

This might be suited to another ticket/discussion but I did think that the bright yellow profile highlights are so strong that they could use more than a mouseover (also: I didn't realise there was a mouseover until reading this ticket!). This would also allow for a little link to the documentation:

image

zas:
What about going for a rollover but keeping the idea of using profiles to set which boolean options are shown there, we can define those 2 cover art related options as being in the default (base) profile, so they can even be removed by adding a new profile (to be replaced by some others).
This way we can have a generic code, let user define its own rollover options, while having 2 common default options for users without forcing them to use profiles.

What do you think?

Aerozol:
I don't think I _entirely_ understand sorry @zas, which makes me worry that it would become a bit technical again! It sounds like a more powerful feature (being able to set those quick options via profile) but also like it would require a user to go into the docs again.

However - selfishly - if it puts the cover art toggles in by default, my ticket and needs are satisfied, so I wouldn't complain!

rdswift:
> This might be suited to another ticket/discussion but I did think that the bright yellow profile highlights are so strong that they could use more than a mouseover (also: I didn't realise there was a mouseover until reading this ticket!). This would also allow for a little link to the documentation:

This should probably be on a separate ticket, and I could take a look at implementing it.

Nice idea, but it's possible that the highlighted options on a page could actually be saved to different profiles (which is why I showed the information in a tooltip).  Any suggestions on how that could / should be displayed at the bottom of the page as in your example?

**EDIT:** Perhaps if there were multiple profiles, the text could say:

***Clicking "Make it So!" will apply the highlighted settings to different profiles.  Placing your mouse pointer over a highlighted item will identify the profile for that item.***

I realize that this is pretty verbose, so there's probably something better.

If you want to pursue this further, please enter a ticket and we can continue the discussion there.

**EDIT 2:** I added ticket PICARD-3108 to capture the request.

zas (approved):
Let's accept this, that's an improvement for the user already. Whether to use profiles or not later is more an internal affair.

PR #2704, opened 2025-08-14
PICARD-3041: Fix `QUrl` drag-n-drop issue; added path canonicalisation

knguyen1:
### Summary

- This is a…
  - [x] Bug fix
  - [ ] Feature addition
  - [ ] Refactoring
  - [ ] Minor / simple change (like a typo)
  - [ ] Other
- Describe this change in 1-2 sentences:
  - Fix drag-and-drop failures for scheme-less paths and address macOS NFC/NFD path mismatches causing accented folders/files to be ignored or fail to open.

### Problem

- JIRA ticket (optional): PICARD-3041
- Drag sources sometimes provide scheme-less paths, making `QUrl.toLocalFile()` return empty and dropping files no-op.
- On macOS (incl. sshfs/FUSE), Unicode normalization (NFC/NFD) mismatches caused ENOENT and “Add Files/Add Folder” to gray out accented directories.

### Solution

- `picard/ui/itemviews/basetreeview.py`: In `drop_urls`, use `url.toLocalFile() or url.path()` and canonicalize the result.
- `picard/util/__init__.py`: Add `canonicalize_path`:
  - Trims trailing NULs, normalizes, resolves paths non-strictly.
  - macOS: component-wise NFC/NFD resolution against filesystem.
  - Windows: preserves long-path handling (`\\?\\` prefix) when needed.
- `picard/ui/mainwindow/__init__.py`: Canonicalize paths from “Add Files” and “Add Folder” before use.
- Tests:
  - `test/test_path_handling.py`: DnD coverage (scheme-less and file://, move vs add, null trimming, HTTP handling).
  - `test/test_canonicalize_path.py`: Canonicalization (NULs, normalization, relative/absolute, Windows long-path behavior).
  - `test/test_resolve_path_components_macos.py` (macOS-only): NFC/NFD resolution per component.

### Action

Additional actions required:
- [ ] Update Picard documentation (optional reference to new path canonicalization behavior)
- [ ] Other (please specify below)

Notes:
- Follow-up PRs can extend canonicalization to other entry points (e.g., cover art local files, `open_local_path`).

### DIFF
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -84,7 +84,6 @@ from picard.track import (
 from picard.util import (
     icontheme,
     iter_files_from_objects,
-    normpath,
     restore_method,
 )
 
@@ -479,12 +478,18 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         for url in urls:
             log.debug("Dropped the URL: %r", url.toString(QtCore.QUrl.UrlFormattingOption.RemoveUserInfo))
             if url.scheme() == 'file' or not url.scheme():
-                filename = normpath(url.toLocalFile().rstrip('\0'))
+                # Prefer a local file if provided; fall back to treating the URL path as a local filesystem path.
+                local: str = url.toLocalFile() or url.path()
+                # canonicalize handles trimming NULs, normalization and macOS NFC/NFD resolution
+                from picard.util import canonicalize_path  # local import to avoid cycles at module import time
+
+                filename: str = canonicalize_path(local) if local else ''
                 file = tagger.files.get(filename)
                 if file:
                     files.append(file)
                 else:
-                    new_paths.append(filename)
+                    if filename:
+                        new_paths.append(filename)
             elif url.scheme() in {'http', 'https'}:
                 file_lookup = tagger.get_file_lookup()
                 file_lookup.mbid_lookup(url.path(), browser_fallback=False)
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -927,6 +927,10 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             filter=";;".join(formats),
         )
         if files:
+            # Canonicalize paths before use
+            from picard.util import canonicalize_path
+
+            files = [canonicalize_path(p) for p in files if p]
             config = get_config()
             config.persist['current_directory'] = os.path.dirname(files[0])
             self.tagger.add_files(files)
@@ -943,12 +947,18 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
                 dir=current_directory,
             )
             if directory:
-                dir_list.append(directory)
+                from picard.util import canonicalize_path
+
+                dir_list.append(canonicalize_path(directory))
         else:
             dir_list = FileDialog.getMultipleDirectories(
                 parent=self,
                 directory=current_directory,
             )
+            if dir_list:
+                from picard.util import canonicalize_path
+
+                dir_list = [canonicalize_path(d) for d in dir_list if d]
 
         dir_count = len(dir_list)
         if dir_count:
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -61,7 +61,7 @@ import json
 import ntpath
 from operator import attrgetter
 import os
-from pathlib import PurePath
+from pathlib import Path, PurePath
 import re
 import subprocess  # nosec: B404
 import sys
@@ -252,6 +252,140 @@ def normpath(path, realpath=True):
     if IS_WIN and not system_supports_long_paths():
         path = win_prefix_longpath(path)
     return path
+
+
+def _resolve_path_components_macos(path: str) -> str:
+    """Best-effort macOS path component resolution.
+
+    Resolve path components against the filesystem on macOS to account for
+    Unicode normalization differences (NFC vs. NFD). Returns a best‑effort
+    on-disk representation, resolving as many components as possible.
+
+    Parameters
+    ----------
+    path : str
+        The input path to resolve. Can be absolute or relative. The final
+        file/directory does not need to exist.
+
+    Returns
+    -------
+    str
+        A path with components matched to existing directory entries where
+        possible. If resolution fails for any component, the original
+        component is kept. Absolute vs. relative form is preserved.
+
+    Notes
+    -----
+    - This function does not change the drive/volume.
+    - It tolerates missing final components (resolves what exists).
+    - Matching is performed using exact, NFC, and NFD comparisons.
+    """
+
+    def _listdir_safe(directory: Path) -> list[str]:
+        try:
+            return [entry.name for entry in directory.iterdir()]
+        except OSError:
+            return []
+
+    def _match_component(candidate: str, entries: list[str]) -> str:
+        if not entries:
+            return candidate
+        if candidate in entries:
+            return candidate
+        cand_nfc = unicodedata.normalize('NFC', candidate)
+        cand_nfd = unicodedata.normalize('NFD', candidate)
+        for entry in entries:
+            if (
+                entry == cand_nfc
+                or entry == cand_nfd
+                or unicodedata.normalize('NFC', entry) == cand_nfc
+                or unicodedata.normalize('NFD', entry) == cand_nfd
+            ):
+                return entry
+        return candidate
+
+    def _join_resolved(parts: list[str], absolute: bool) -> str:
+        if not parts:
+            return os.sep if absolute else ''
+        if absolute:
+            return str(Path('/').joinpath(*parts))
+        return str(Path().joinpath(*parts))
+
+    try:
+        if not path:
+            return path
+        is_abs = Path(path).is_absolute()
+        parts = [p for p in PurePath(path).parts if p not in ('', '.')]
+        resolved_parts: list[str] = []
+        current_dir = Path('/') if is_abs else Path.cwd()
+
+        for part in parts:
+            entries = _listdir_safe(current_dir)
+            chosen = _match_component(part, entries)
+            resolved_parts.append(chosen)
+            current_dir = current_dir / chosen
+
+        return _join_resolved(resolved_parts, is_abs)
+    except Exception as err:  # noqa: BLE001
+        log.debug("canonicalize components failed for %r: %s", path, err)
+        return str(Path(path))
+
+
+def canonicalize_path(path: str) -> str:
+    """Canonicalize a path for robust I/O.
+
+    Produces a sanitized path suitable for file operations by trimming
+    trailing NULs, normalizing separators, and resolving platform-specific
+    quirks (e.g., Unicode normalization on macOS).
+
+    Parameters
+    ----------
+    path : str
+        The input path as a Unicode string.
+
+    Returns
+    -------
+    str
+        A canonicalized path. On macOS, components are matched against the
+        filesystem using normalization-insensitive rules. On all platforms,
+        the result is normalized and, where possible, resolved via
+        ``os.path.realpath``.
+
+    Notes
+    -----
+    - Trailing ``"\0"`` characters are stripped.
+    - If resolution via ``realpath`` fails, the normalized input is returned.
+    - Non-string inputs are returned unchanged.
+    """
+    if not isinstance(path, str):
+        return path  # type: ignore[return-value]
+
+    # Trim any accidental NUL terminators and normalize separators
+    trimmed = path.rstrip('\0')
+    if not trimmed:
+        return ''
+
+    # Avoid realpath here; resolve components manually on macOS first
+    normalized = str(Path(trimmed))
+
+    if IS_MACOS:
+        resolved = _resolve_path_components_macos(normalized)
+        # realpath only if possible, otherwise keep best-effort resolved path
+        try:
+            candidate = str(Path(resolved).resolve(strict=False))
+        except OSError:
+            candidate = resolved
+    else:
+        try:
+            candidate = str(Path(normalized).resolve(strict=False))
+        except OSError:
+            candidate = normalized
+
+    # Carry over Windows long-path handling from normpath
+    if IS_WIN and not system_supports_long_paths():
+        candidate = win_prefix_longpath(candidate)
+
+    return candidate
 
 
 def win_prefix_longpath(path):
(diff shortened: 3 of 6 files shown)


picard/util/__init__.py
```diff
@@ -254,6 +254,148 @@ def normpath(path, realpath=True):
...
+        # unexpectedly change drive letters on Windows runners (e.g. C: → D:).
+        # Prefer an absolute, normalized path without crossing filesystem links.
+        candidate = os.path.abspath(os.path.normpath(normalized))
+
+    # Windows: normalize drive letter to uppercase for stable comparisons
+    if IS_WIN and len(candidate) >= 2 and candidate[1] == ':':
+        candidate = candidate[0].upper() + candidate[1:]
+    # Carry over Windows long-path handling from normpath
+    if IS_WIN and not system_supports_long_paths():
+        candidate = win_prefix_longpath(candidate)
```
zas: Maybe regroup all win-specific code under one `if`:

```python
    # Windows: normalize drive letter to uppercase for stable comparisons
    if IS_WIN:
        if len(candidate) >= 2 and candidate[1] == ':':
            candidate = candidate[0].upper() + candidate[1:]
        # Carry over Windows long-path handling from normpath
        if not system_supports_long_paths():
            candidate = win_prefix_longpath(candidate)
```

picard/util/__init__.py
```diff
@@ -254,6 +254,148 @@ def normpath(path, realpath=True):
...
+    Notes
+    -----
+    - This function does not change the drive/volume.
+    - It tolerates missing final components (resolves what exists).
+    - Matching is performed using exact, NFC, and NFD comparisons.
+    """
+
+    def _listdir_safe(directory: Path) -> list[str]:
+        try:
+            return [entry.name for entry in directory.iterdir()]
```
zas: A list isn't strictly needed, what about returning a tuple instead?

picard/util/__init__.py
```diff
@@ -254,6 +254,148 @@ def normpath(path, realpath=True):
...
+            return os.sep if absolute else ''
+        if absolute:
+            return str(Path('/').joinpath(*parts))
+        return str(Path().joinpath(*parts))
+
+    try:
+        if not path:
+            return path
+        is_abs = Path(path).is_absolute()
+        parts = [p for p in PurePath(path).parts if p not in ('', '.')]
```
zas: This could be merged with `for part in`  loop below:

```python
for part in PurePath(path).parts:
    if part in ('', '.'):
        continue
```

Also we could move most of the code to a generator, something like (totally untested):

```python
def resolved_parts(is_abs): 
        current_dir = Path('/') if is_abs else Path.cwd()

        for part in  PurePath(path).parts:
            if part not in ('', '.'):
                entries = _listdir_safe(current_dir)
                chosen = _match_component(part, entries)
                yield chosen
                current_dir = current_dir / chosen


 return _join_resolved(resolved_parts(is_abs), is_abs)
``

File: picard/ui/itemviews/basetreeview.py [hunk 4]
```diff
@@ -475,16 +483,38 @@ def scrollTo(self, index, scrolltype=QtWidgets.QAbstractItemView.ScrollHint.Ensu
     def drop_urls(urls, target, move_to_multi_tracks=True):
         files = []
         new_paths = []
-        tagger = QtCore.QCoreApplication.instance()
+        # We need to cast to Tagger here to avoid type checking errors below.
+        tagger = cast("Tagger", QtCore.QCoreApplication.instance())
```
zas, line 484: Unrelated to this PR, but there are more places where we should do that.

picard/ui/itemviews/basetreeview.py
```diff
@@ -475,16 +483,38 @@ def scrollTo(self, index, scrolltype=QtWidgets.QAbstractItemView.ScrollHint.Ensu
...
+            # string and treat it as a local path rather than a custom scheme.
+            if (
+                url.scheme() == 'file'
+                or not url.scheme()
+                or (
+                    IS_WIN
+                    and len(url.scheme()) == 1
+                    and re.match(r'^[a-zA-Z]:', url.toString(QtCore.QUrl.UrlFormattingOption.RemoveUserInfo))
+                )
+            ):
```
zas: Not easy to read, perhaps move that to a small help function outside the loop.

picard/ui/mainwindow/__init__.py
```diff
@@ -927,6 +928,8 @@ def add_files(self):
             filter=";;".join(formats),
         )
         if files:
+            # Canonicalize paths before use
+            files = [canonicalize_path(p) for p in files if p]
```
zas: Prefer tuple over list when it's unlikely it will be modified.

picard/ui/mainwindow/__init__.py
```diff
@@ -943,12 +946,14 @@ def add_directory(self):
...
             if directory:
-                dir_list.append(directory)
+                dir_list.append(canonicalize_path(directory))
         else:
             dir_list = FileDialog.getMultipleDirectories(
                 parent=self,
                 directory=current_directory,
             )
+            if dir_list:
+                dir_list = [canonicalize_path(d) for d in dir_list if d]
```
zas: tuple vs list

zas (changes requested):

phw (commented):
I don't have the time right now to review the code in detail. But I have my concerns about the general approach for fixing a very specific issue reported on macOS.

The approach of verifying every path component in different Unicode normalizations against the file system by iterating over the parent directory appears to me to be a rather big performance issue. Specifically when doing so over a virtual filesystem backed by an SSH connection (which is the case for PICARD-3041). And it is not even clear to me, whether this change indeed fixes the reported issue.

In general our file handling is already rather complex, due to considering multiple special cases across operating systems. We already do path normalization at various parts, e.g. with the `normpath` function that also handles Windows long paths.

This change seems to me to add yet another layer to this complexity, while I think the issue behind PICARD-3041 might be related to some path handling we already do. My main suspects would be the `encode_filename`  / `decode_filename` helpers that are not consistently used everywhere and also might lead to those normalization issues.

If we actually end up in situations where we need to try different Unicode normalizations, maybe we should only do this in the error case?

I'm not generally against this change, and if you all convince me that this is all needed to fix path issues on macOS then we maybe need to go this approach. But right now it does seem to add a lot of complexity.

knguyen1:
>And it is not even clear to me, whether this change indeed fixes the reported issue.

It's not just OP's macFUSE/sshfs issue; what this will fix though is for programmes that supply schema-less URL's.  QT people already made it clear they're not going to fix it.  Like you, their concern was complexity and who's really responsible?  I mean yeah, it would be wonderful if all apps provide full URI for files.

```python
>>> from PyQt6.QtCore import QUrl
>>> url = QUrl("/User/music/file.mp3")
>>> url.toLocalFile()
''
>>> url.path()
'/User/music/file.mp3'
>>> url.scheme()
''
>>> 
```

> My main suspects would be the encode_filename / decode_filename helpers that are not consistently used everywhere and also might lead to those normalization issues.

I was going to take on those next.  Introducing a `canonicalisation` concept before encode/decode fn made more sense.

According to the LLM I'm using, there's actually a lot more to do in terms of handling non-unicode systems better.

### Next tasks to standardize non‑Unicode/locale handling

- Replace legacy helpers
  - Deprecate `encode_filename`/`decode_filename`. Introduce small I/O helpers:
    - `to_io_path(x: str|bytes) -> str|bytes` (pass through `str`; use `os.fsencode` only if the callee strictly needs `bytes`).
    - `open_safe(path, ...)` wrapper that logs both original and canonicalized candidates on failure.

- Centralize canonicalization
  - Use `canonicalize_path` at all UI/file entry points: cover art local files, “Open Containing Folder”, drag-and-drop, clipboard paste, remote command handlers, plugin file inputs.
  - Add it to `open_local_path` and any path constructed from `QUrl`.

- Mutagen and formats
  - Ensure formats pass `str` paths to Mutagen whenever possible; remove any lossy `.encode(..., 'replace')`.
  - At write time, if a path fails due to locale/FS constraints, retry with a sanitized fallback name and surface a clear message.

- POSIX bytes path support
  - Allow bytes paths to flow through low-level scanning (`_scan_paths_recursive`) and I/O when needed on non‑UTF‑8 locales; keep UI strings as `str`.

- Locale and normalization strategy
  - On startup, detect problematic locales (e.g., ASCII, ISO‑8859‑1) and warn with actionable guidance.
  - macOS: expose a setting to enable/disable component-wise NFC/NFD resolution and log when it alters a path.

- Error reporting and logging
  - Standardize path logging to include:
    - Original input, canonicalized result, NFC/NFD variants (macOS), `exists/isdir` checks.
    - Use safe repr for undecodable bytes without raising.

- Test coverage and CI
  - Add matrix jobs simulating non‑UTF‑8 locales on Linux; run bytes-path tests on POSIX.
  - Unit tests for all entry points wired to `canonicalize_path`.
  - Mocked sshfs normalization test to emulate NFD exposure without a real mount.

- Documentation and settings
  - Update docs on how Picard handles encodings/normalization, troubleshooting (sshfs, locales), and new settings/toggles if added.

- Type hints and interfaces
  - Annotate path-facing APIs as `str | bytes | os.PathLike` where applicable; keep UI layers `str`-only, I/O wrappers accept both.

- Gradual migration
  - Grep and replace remaining direct `os.path` usage where normalization/resolution is expected; route through `canonicalize_path` and I/O wrappers.

picard/util/__init__.py
knguyen1: Refactor to tuple: https://github.com/metabrainz/picard/pull/2704/commits/037b30c97ee73a1a464baf8beb97154fb59ed9dc

picard/ui/mainwindow/__init__.py
knguyen1: Refactor to tuple: https://github.com/metabrainz/picard/pull/2704/commits/037b30c97ee73a1a464baf8beb97154fb59ed9dc

picard/ui/mainwindow/__init__.py
knguyen1: Refactor to tuple: https://github.com/metabrainz/picard/pull/2704/commits/037b30c97ee73a1a464baf8beb97154fb59ed9dc

picard/util/__init__.py
knguyen1: Merged filter logic: https://github.com/metabrainz/picard/pull/2704/commits/9c73c2b7eadf9c260ed5aafef6217703d430ad20

picard/util/__init__.py
knguyen1: Combine `IS_WIN` block: https://github.com/metabrainz/picard/pull/2704/commits/45c41d08c814ff8c789c4bbdf7460fe5a32cbcee

picard/ui/itemviews/basetreeview.py
knguyen1: Refactor to util methods: https://github.com/metabrainz/picard/pull/2704/commits/0e5a4cce4e4e18cd726a34d383532685eea6bfda

picard/ui/itemviews/basetreeview.py
knguyen1: Revert or no?  It was annoying me to get error intellisense highlights without it.

knguyen1: FYI: I recommend `ty`, when the project comes out of pre-release:  https://github.com/astral-sh/ty It's from the same authors as `ruff` and `uv`.

phw: I wouldn't do this. As Python is not statically typed there is no need for such casts during runtime.

But getting the main application instance is a typical case. Maybe we should provide our own helper method for this. This could have the appropriate type hint set for the return value.

phw:
How exactly is the fact that we get a path and not a file URI the cause of the issue in the ticket? The functionality should generally stay the same, with or without the `file://` scheme.

My main concern with this PR is really the macOS code that tries to resolve a misencoded filename against the filesystem. This feels to me as it is working around the symptoms rather to address the root cause, which is that we end up with a misencoded file path for some reason.

At a minimum we should:

- measure the performance impact on file loading with this change
- reproduce the issue in the ticket and ensure it is fixed

zas:
> My main concern with this PR is really the macOS code that tries to resolve a misencoded filename against the filesystem. This feels to me as it is working around the symptoms rather to address the root cause, which is that we end up with a misencoded file path for some reason.

I have the same concern regarding this patch.

knguyen1:
> The functionality should generally stay the same, with or without the `file://` scheme.

I don't follow. In a schema-less URL, `toLocalFile` (current behaviour) returns `''`, as referenced in https://bugreports.qt.io/browse/QTBUG-22382  Are you saying it's already possible for picard to add schema-less files?  The unit tests in this PR seems to indicate no...

> This feels to me as it is working around the symptoms rather to address the root cause, which is that we end up with a misencoded file path for some reason.

The work for which should be its own epic, to make picard play well with non-unicode FS and locales.  I'm wondering if the juice is even worth the squeeze.  There's a handful user complaints on JIRA that is yet to be resolved (mutagen errors, posix bytes-encoded paths, etc.), well, including this one...

phw:
> I don't follow. In a schema-less URL, `toLocalFile` (current behaviour) returns `''`, as referenced in https://bugreports.qt.io/browse/QTBUG-22382 Are you saying it's already possible for picard to add schema-less files? The unit tests in this PR seems to indicate no...

Generally yes of course (e.g. if passed form command line). Inside the specific code path of dropping in `BaseTreeView.drop_urls` not, and your fixes there to handle this there make totally sense. I just have no indication that this actually happened in practice, nor does it seem to be related to PICARD-3041.

We should of course fix this handling, and accept schema less "URL" in this drop action.

My concerns are more about the path canonicalization code, specifically on macOS, and the PR claiming to solve  PICARD-3041 (which it might or might not do).

knguyen1:
@phw Let me take some time to digest this.  Maybe `canonicalise_path` is too much and too unnecessary for now.  We need OP to do a live testing.

> My concerns are more about the path canonicalization code, specifically on macOS, and the PR claiming to solve [PICARD-3041](https://tickets.metabrainz.org/browse/PICARD-3041) (which it might or might not do).

The reason `canonicalize_path` is part of this PR is because OP complained about the fileviewer greying out his Verite folder.  Which I suspected it had to do with NFC/NFD normalisation.

image

phw:
I am closing this for now. As stated above I don't think the approach here is the way to fix the issue it claims to fix. The issue of handling schema-less files in the drop action also does not seem to be really relevant. I'm not sure how one would even trigger this behavior, as the drag and drop is handled by the OS and Qt, and we do have paths with schema passed in.

Addressing [PICARD-3041](https://tickets.metabrainz.org/browse/PICARD-3041) requires actually understanding why we end up with different encodings. It pretty likely is related to Unicode normalization in some way, but I think this needs some reproduction with an actuall sshfs mount on macOS.

PR #2707, opened 2025-08-16
Change platform specific deps

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

The management of requirements-win.txt and requirements-macos-11.0.txt was moved to uv as separate requirements groups. Unfortunately this doesn't fit our use case. uv expects the groups to not conflict. So you cannot for example fix a requirement in one group without fixing it for the entire project. Also setting conflicting versions in different groups is not allowed.

# Solution

Revert to handling the requirements-win.txt and requirements-macos-11.0.txt manually again. For the majority of use cases we want to use the general requirments in pyproject.toml anyway, and I'll usually just upgrade the platform specific ones to the latest versions we also use in the generated requirements.txt. I'm fine to manage the specific cases for the wind and mac binary builds manually when needed.


zas (approved):
What a pity, let's hope we can revert this in the future.

phw:
Maybe we can even get rid of it. Right now in this setup we actually don't need it (same deps anywhere), but I'm yet hesitant to change it. Just a few months ago we had the need for different Qt6 versions on Win and macOS.

PR #2709, opened 2025-08-18
PICARD-2103: Add API for custom columns (field ref, scripts, transforms, etc.)

knguyen1:
### Summary

- This is a…
  - [ ] Bug fix
  - [x] Feature addition
  - [ ] Refactoring
  - [ ] Minor / simple change (like a typo)
  - [ ] Other
- Describe this change in 1-2 sentences:
  - Introduces a public API to define and register custom columns for File/Album views, including script-based columns, provider protocols, and extensible sorting adapters.

### Problem

- JIRA ticket: PICARD-2103
- Users cannot add custom/computed columns; limited sorting and no script-backed values in the views.

### Solution

- Added `picard.ui.itemviews.custom_columns` package:
  - Protocols (`ColumnValueProvider`, `SortKeyProvider`), `CustomColumn` type.
  - Factory helpers: `make_field_column`, `make_script_column`, `make_callable_column`, `make_transformed_column`.
  - `registry` to register/unregister columns in File/Album views.
  - Script evaluation (`ChainedValueProvider`) with caching/perf thresholds and album-loading cache avoidance.
  - Context/resolver layers for fast simple-var lookup and ScriptParser fallback.
  - Sorting adapters: casefold/descending, numeric/descending, article-insensitive, composite, nulls-first/last, length, random, cached, reverse.
- Comprehensive tests added for API and adapters; type hints and concise NumPy-style docstrings included.

#### Test With

Append to `./picard/ui/itemviews/columns.py`:

```python
# Temporary developer hook to live-test a custom script column.
# Enable by setting environment variable PICARD_DEV_ENABLE_CUSTOM_SCRIPT_COLUMN=1
try:
    import os

    if os.environ.get('PICARD_DEV_ENABLE_CUSTOM_SCRIPT_COLUMN') == '1':  # pragma: no cover - dev only
        from picard.ui.columns import ColumnAlign, ColumnSortType
        from picard.ui.itemviews.custom_columns import (
            CasefoldSortAdapter,
            CustomColumn,
            make_script_column,
            registry,
        )

        # Script-based column
        script = "$if(%title%,$if2(%artist%,Unknown Artist) - $if2(%title%,Unknown Title),$if2(%albumartist%,Unknown Artist) - $if2(%album%,Unknown Album))"
        base_col = make_script_column(
            title="Artist – Title",
            key="artist_title_script",
            script=script,
            width=280,
            align=ColumnAlign.LEFT,
        )

        # Add case-insensitive sorting
        sorted_provider = CasefoldSortAdapter(base_col.provider)
        sorted_col = CustomColumn(
            title=base_col.title,
            key=base_col.key,
            provider=sorted_provider,
            width=base_col.width,
            align=base_col.align,
            sort_type=ColumnSortType.SORTKEY,
        )

        # Register after Title
        registry.register(sorted_col, insert_after_key="title", add_to=["ALBUM_VIEW"])
except Exception:
    # Never break production code due to dev-only snippet
    pass
```

Then

```bash
PICARD_DEV_ENABLE_CUSTOM_SCRIPT_COLUMN=1 python tagger.py
```

#### Demo: See `Artist - Title` custom column

image

### Action

Additional actions required:
- [x] Update Picard documentation (reference this PR)
- [ ] Other:

Note: A follow-up PR is required for custom column management UX (create/edit/delete, ordering, persistence). See: #2714

### DIFF
--- a/picard/ui/itemviews/__init__.py
+++ b/picard/ui/itemviews/__init__.py
@@ -398,6 +398,9 @@ class TreeItem(QtWidgets.QTreeWidgetItem):
         return sortkey
 
     def update_colums_text(self, color=None, bgcolor=None):
+        # Local import to avoid cycles
+        from picard.ui.itemviews.custom_columns import CustomColumn
+
         for i, column in enumerate(self.columns):
             if color is not None:
                 self.setForeground(i, color)
@@ -413,8 +416,6 @@ class TreeItem(QtWidgets.QTreeWidgetItem):
                     self.setTextAlignment(i, QtCore.Qt.AlignmentFlag.AlignRight | QtCore.Qt.AlignmentFlag.AlignVCenter)
                 # Support custom columns with provider evaluation
                 try:
-                    from picard.ui.itemviews.custom_columns import CustomColumn  # Local import to avoid cycles
-
                     if isinstance(column, CustomColumn):
                         self.setText(i, column.provider.evaluate(self.obj))
                         continue
--- a/picard/ui/itemviews/__init__.py
+++ b/picard/ui/itemviews/__init__.py
@@ -411,6 +411,16 @@ class TreeItem(QtWidgets.QTreeWidgetItem):
             else:
                 if column.align == ColumnAlign.RIGHT:
                     self.setTextAlignment(i, QtCore.Qt.AlignmentFlag.AlignRight | QtCore.Qt.AlignmentFlag.AlignVCenter)
+                # Support custom columns with provider evaluation
+                try:
+                    from picard.ui.itemviews.custom_columns import CustomColumn  # Local import to avoid cycles
+
+                    if isinstance(column, CustomColumn):
+                        self.setText(i, column.provider.evaluate(self.obj))
+                        continue
+                except Exception:  # noqa: BLE001
+                    # Fallback to default behavior
+                    pass
                 self.setText(i, self.obj.column(column.key))
 
 
--- a/picard/ui/itemviews/custom_columns/__init__.py
+++ b/picard/ui/itemviews/custom_columns/__init__.py
@@ -31,7 +31,10 @@ from picard.ui.itemviews.custom_columns.factory import (
     make_script_column,
     make_transformed_column,
 )
-from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
+from picard.ui.itemviews.custom_columns.protocols import (
+    ColumnValueProvider,
+    SortKeyProvider,
+)
 from picard.ui.itemviews.custom_columns.registry import registry
 from picard.ui.itemviews.custom_columns.sorting_adapters import (
     ArticleInsensitiveAdapter,
--- a/picard/ui/itemviews/custom_columns/__init__.py
+++ b/picard/ui/itemviews/custom_columns/__init__.py
@@ -1,0 +1,74 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2025 The MusicBrainz Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Public API for custom columns: types, factories and registry."""
+
+from __future__ import annotations
+
+from picard.ui.itemviews.custom_columns.column import CustomColumn
+from picard.ui.itemviews.custom_columns.factory import (
+    _create_custom_column,
+    make_callable_column,
+    make_field_column,
+    make_script_column,
+    make_transformed_column,
+)
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
+from picard.ui.itemviews.custom_columns.registry import registry
+from picard.ui.itemviews.custom_columns.sorting_adapters import (
+    ArticleInsensitiveAdapter,
+    CachedSortAdapter,
+    CasefoldSortAdapter,
+    CompositeSortAdapter,
+    DescendingCasefoldSortAdapter,
+    DescendingNumericSortAdapter,
+    LengthSortAdapter,
+    NullsFirstAdapter,
+    NullsLastAdapter,
+    NumericSortAdapter,
+    RandomSortAdapter,
+    ReverseAdapter,
+)
+
+
+__all__ = [
+    "ColumnValueProvider",
+    "SortKeyProvider",
+    "CustomColumn",
+    "make_field_column",
+    "make_script_column",
+    "make_callable_column",
+    "make_transformed_column",
+    "registry",
+    "_create_custom_column",
+    # Sorting adapters
+    "CasefoldSortAdapter",
+    "DescendingCasefoldSortAdapter",
+    "NumericSortAdapter",
+    "DescendingNumericSortAdapter",
+    "LengthSortAdapter",
+    "RandomSortAdapter",
+    "ArticleInsensitiveAdapter",
+    "CompositeSortAdapter",
+    "NullsLastAdapter",
+    "NullsFirstAdapter",
+    "CachedSortAdapter",
+    "ReverseAdapter",
+]
--- a/picard/ui/itemviews/custom_columns/__init__.py
+++ b/picard/ui/itemviews/custom_columns/__init__.py
@@ -47,7 +47,6 @@ from picard.ui.itemviews.custom_columns.sorting_adapters import (
     NullsFirstAdapter,
     NullsLastAdapter,
     NumericSortAdapter,
-    RandomSortAdapter,
     ReverseAdapter,
 )
 
@@ -69,7 +68,6 @@ __all__ = [
     "NumericSortAdapter",
     "DescendingNumericSortAdapter",
     "LengthSortAdapter",
-    "RandomSortAdapter",
     "ArticleInsensitiveAdapter",
     "CompositeSortAdapter",
     "NullsLastAdapter",
--- a/picard/ui/itemviews/custom_columns/__init__.py
+++ b/picard/ui/itemviews/custom_columns/__init__.py
@@ -27,6 +27,7 @@ from picard.ui.itemviews.custom_columns.factory import (
     _create_custom_column,
     make_callable_column,
     make_field_column,
+    make_provider_column,
     make_script_column,
     make_transformed_column,
 )
@@ -53,6 +54,7 @@ __all__ = [
     "SortKeyProvider",
     "CustomColumn",
     "make_field_column",
+    "make_provider_column",
     "make_script_column",
     "make_callable_column",
     "make_transformed_column",
--- a/picard/ui/itemviews/custom_columns/column.py
+++ b/picard/ui/itemviews/custom_columns/column.py
@@ -22,8 +22,15 @@
 
 from __future__ import annotations
 
-from picard.ui.columns import Column, ColumnAlign, ColumnSortType
-from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
+from picard.ui.columns import (
+    Column,
+    ColumnAlign,
+    ColumnSortType,
+)
+from picard.ui.itemviews.custom_columns.protocols import (
+    ColumnValueProvider,
+    SortKeyProvider,
+)
 
 
 class CustomColumn(Column):
--- a/picard/ui/itemviews/custom_columns/column.py
+++ b/picard/ui/itemviews/custom_columns/column.py
@@ -1,0 +1,76 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2025 The MusicBrainz Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Custom column type that delegates value and optional sort to a provider."""
+
+from __future__ import annotations
+
+from picard.ui.columns import Column, ColumnAlign, ColumnSortType
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
+
+
+class CustomColumn(Column):
+    """A column whose value is computed by a provider for each row object."""
+
+    def __init__(
+        self,
+        title: str,
+        key: str,
+        provider: ColumnValueProvider,
+        width: int | None = None,
+        align: ColumnAlign = ColumnAlign.LEFT,
+        sort_type: ColumnSortType = ColumnSortType.TEXT,
+        always_visible: bool = False,
+    ):
+        """Create custom column.
+
+        Parameters
+        ----------
+        title
+            Display title of the column.
+        key
+            Internal key used to identify the column.
+        provider
+            Provide computation of values.
+        width
+            Optional fixed width in pixels.
+        align
+            Set text alignment.
+        sort_type
+            Set sorting behavior of the column.
+        always_visible
+            If True, hide the column visibility toggle.
+        """
+        sortkey_fn = (
+            provider.sort_key  # type: ignore[attr-defined]
+            if (sort_type == ColumnSortType.SORTKEY and isinstance(provider, SortKeyProvider))
+            else None
+        )
+        super().__init__(
+            title,
+            key,
+            width=width,
+            align=align,
+            sort_type=sort_type,
+            sortkey=sortkey_fn,
+            always_visible=always_visible,
+            status_icon=False,
+        )
+        self.provider = provider
--- a/picard/ui/itemviews/custom_columns/context.py
+++ b/picard/ui/itemviews/custom_columns/context.py
@@ -28,10 +28,8 @@ from abc import (
 )
 from typing import Any
 
-from picard.album import Album
-from picard.cluster import Cluster
 from picard.file import File
-from picard.item import Item
+from picard.item import Item, MetadataItem
 from picard.track import Track
 
 
@@ -68,25 +66,9 @@ class TrackContextStrategy(ContextStrategy):
         return (getattr(obj, 'metadata', None), file_obj)
 
 
-class AlbumContextStrategy(ContextStrategy):
-    def can_handle(self, obj: Item) -> bool:
-        return isinstance(obj, Album)
-
-    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
-        return (getattr(obj, 'metadata', None), None)
-
-
-class ClusterContextStrategy(ContextStrategy):
-    def can_handle(self, obj: Item) -> bool:
-        return isinstance(obj, Cluster)
-
-    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
-        return (getattr(obj, 'metadata', None), None)
-
-
 class DefaultContextStrategy(ContextStrategy):
     def can_handle(self, obj: Item) -> bool:
-        return True
+        return hasattr(obj, 'metadata') or isinstance(obj, MetadataItem)
 
     def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
         return (getattr(obj, 'metadata', None), None)
@@ -97,8 +79,6 @@ class ContextStrategyManager:
         self.strategies = [
             FileContextStrategy(),
             TrackContextStrategy(),
-            AlbumContextStrategy(),
-            ClusterContextStrategy(),
             DefaultContextStrategy(),
         ]
 
--- a/picard/ui/itemviews/custom_columns/context.py
+++ b/picard/ui/itemviews/custom_columns/context.py
@@ -22,7 +22,10 @@
 
 from __future__ import annotations
 
-from abc import ABC, abstractmethod
+from abc import (
+    ABC,
+    abstractmethod,
+)
 from typing import Any
 
 from picard.album import Album
--- a/picard/ui/itemviews/custom_columns/context.py
+++ b/picard/ui/itemviews/custom_columns/context.py
@@ -1,0 +1,106 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2025 The MusicBrainz Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Context extraction strategies for different Picard item types."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from typing import Any
+
+from picard.album import Album
+from picard.cluster import Cluster
+from picard.file import File
+from picard.item import Item
+from picard.track import Track
+
+
+class ContextStrategy(ABC):
+    """Abstract base class for context creation strategies."""
+
+    @abstractmethod
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        """Return (metadata, file_obj) tuple"""
+        raise NotImplementedError
+
+    @abstractmethod
+    def can_handle(self, obj: Item) -> bool:
+        """Check if this strategy can handle the given object type"""
+        raise NotImplementedError
+
+
+class FileContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return isinstance(obj, File)
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        return (getattr(obj, 'metadata', None), obj if isinstance(obj, File) else None)
+
+
+class TrackContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return isinstance(obj, Track)
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        files = getattr(obj, 'files', None)
+        num_linked = getattr(obj, 'num_linked_files', 0)
+        file_obj = files[0] if isinstance(files, (list, tuple)) and num_linked == 1 else None
+        return (getattr(obj, 'metadata', None), file_obj)
+
+
+class AlbumContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return isinstance(obj, Album)
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        return (getattr(obj, 'metadata', None), None)
+
+
+class ClusterContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return isinstance(obj, Cluster)
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        return (getattr(obj, 'metadata', None), None)
+
+
+class DefaultContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return True
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        return (getattr(obj, 'metadata', None), None)
+
+
+class ContextStrategyManager:
+    def __init__(self):
+        self.strategies = [
+            FileContextStrategy(),
+            TrackContextStrategy(),
+            AlbumContextStrategy(),
+            ClusterContextStrategy(),
+            DefaultContextStrategy(),
+        ]
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        for strategy in self.strategies:
+            if strategy.can_handle(obj):
+                return strategy.make_context(obj)
+        return (None, None)
--- a/picard/ui/itemviews/custom_columns/factory.py
+++ b/picard/ui/itemviews/custom_columns/factory.py
@@ -25,6 +25,7 @@ from __future__ import annotations
 from collections.abc import Callable
 
 from picard.item import Item
+from picard.script import ScriptParser
 
 from picard.ui.columns import ColumnAlign, ColumnSortType
 from picard.ui.itemviews.custom_columns.column import CustomColumn
@@ -78,6 +79,10 @@ def _create_custom_column(
         The configured column.
     """
     inferred_sort_type = _infer_sort_type(provider, sort_type)
+    # If explicitly requested SORTKEY but provider cannot supply a sort_key,
+    # downgrade to TEXT to avoid invalid configuration
+    if inferred_sort_type == ColumnSortType.SORTKEY and not isinstance(provider, SortKeyProvider):
+        inferred_sort_type = ColumnSortType.TEXT
     return CustomColumn(
         title,
         key,
@@ -131,6 +136,8 @@ def make_script_column(
     always_visible: bool = False,
     max_runtime_ms: int = 25,
     cache_size: int = 1024,
+    parser: ScriptParser | None = None,
+    parser_factory: Callable[[], ScriptParser] | None = None,
 ) -> CustomColumn:
     """Create column whose value is computed by a script.
 
@@ -144,7 +151,13 @@ def make_script_column(
     CustomColumn
         The script-backed column.
     """
-    provider = ChainedValueProvider(script, max_runtime_ms=max_runtime_ms, cache_size=cache_size)
+    provider = ChainedValueProvider(
+        script,
+        max_runtime_ms=max_runtime_ms,
+        cache_size=cache_size,
+        parser=parser,
+        parser_factory=parser_factory,
+    )
     return _create_custom_column(
         title,
         key,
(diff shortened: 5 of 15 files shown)


knguyen1:
Failing lint is from: https://github.com/metabrainz/picard/commit/03d5c9ddd726963d8f98574b97e9be6962bf5b47
We need a PR to run `ruff check --fix` and push the changes.

Then I'll rebase `knguyen1:feat/PICARD-2103/add-ability-to-add-custom-columns-to-fiewview-and-albumview` to master.  It'll clear the lint failure.

phw:
> We need a PR to run `ruff check --fix` and push the changes.

https://github.com/metabrainz/picard/pull/2710

knguyen1:
> Can you rebase this PR?

Rebased: https://github.com/metabrainz/picard/compare/0df8df5f1f38c22a8f29a37cd53f196930b7a41c..1446569de09c85bf29848cc7edd157229dd0bd69

picard/ui/itemviews/custom_columns/providers.py
```diff
@@ -0,0 +1,79 @@
...
+
+class CallableProvider:
+    def __init__(self, func: Callable[[Item], str]):
+        self._func = func
+
+    def evaluate(self, obj: Item) -> str:
+        try:
+            return str(self._func(obj))
+        except (TypeError, ValueError, AttributeError) as e:
+            debug("CallableProvider failure for %r: %r", getattr(self._func, "__name__", self._func), e)
```
zas: 2 things here:

- while importing just `debug()` is ok, please stick to `log.debug()`, as we do everywhere else in the code.
- Use `self.__class__.__name__` instead of hard-coded class name: it helps when renaming classes, but also when subclassing them.

picard/ui/itemviews/custom_columns/registry.py
```diff
@@ -0,0 +1,247 @@
...
+        # may be modified by plugins or other parts of the application at runtime.
+        # Importing at registration time ensures we always get the current state
+        # of these collections, not a stale reference from module load time.
+        from picard.ui.itemviews.columns import ALBUMVIEW_COLUMNS, FILEVIEW_COLUMNS
+
+        targets = set(str(v).upper() for v in (add_to or {VIEW_FILE, VIEW_ALBUM}))
+        if VIEW_FILE in targets:
+            self._insert_column(FILEVIEW_COLUMNS, column, insert_after_key)
+        if VIEW_ALBUM in targets:
+            self._insert_column(ALBUMVIEW_COLUMNS, column, insert_after_key)
```
zas: A more generic approach:

```python
VIEW_COLUMNS = {
    VIEW_FILE: FILEVIEW_COLUMNS,
    VIEW_ALBUM: ALBUMVIEW_COLUMNS,
}

for target in targets:
    self._insert_column(VIEW_COLUMNS[target], column, insert_after_key)
```

Note: let the exception be raised if `target` is an unknown view, it shouldn't happen anyway.

picard/ui/itemviews/custom_columns/registry.py
```diff
@@ -0,0 +1,247 @@
...
+        column = self._by_key.pop(key, None)
+        if not column:
+            return None
+
+        # Local import: Same reasoning as in register() - we need the current
+        # state of these mutable module-level collections at the time of
+        # unregistration, not what they were at module load time
+        from picard.ui.itemviews.columns import ALBUMVIEW_COLUMNS, FILEVIEW_COLUMNS
+
+        for cols in (FILEVIEW_COLUMNS, ALBUMVIEW_COLUMNS):
```
zas: Using the dict above more widely would help here, it could be something like:

```python
for cols in VIEW_COLUMNS.values():
```

picard/ui/itemviews/custom_columns/shared.py
```diff
@@ -0,0 +1,83 @@
...
+
+    Returns
+    -------
+    str
+        Comma-separated identifiers. Order follows ``DEFAULT_ADD_TO`` for
+        recognized views; any additional (unrecognized) tokens are appended in
+        alphabetical order for forward-compatibility.
+    """
+
+    view_set: set[str] = {v.upper() for v in views if v}
```
zas:
```suggestion
    view_set: set[str] = {v.strip().upper() for v in views if v}
```

picard/ui/itemviews/custom_columns/providers.py
knguyen1: Use `log.debug`: https://github.com/metabrainz/picard/pull/2709/commits/8016277b48dbb97786083abd60a26c3f2f79ca71

Do not hardcode classname:https://github.com/metabrainz/picard/pull/2709/commits/80f64f52b25e5df1e025574bb9e9cf4f71b74f89

picard/ui/itemviews/custom_columns/registry.py
knguyen1: https://github.com/metabrainz/picard/pull/2709/commits/ce9fbe40bae8a15050268c15923010e4eebc7951

picard/ui/itemviews/custom_columns/registry.py
knguyen1: https://github.com/metabrainz/picard/pull/2709/commits/d4f1453934e23a4e38cb3b9fcde5a61e5bb1433d

picard/ui/itemviews/custom_columns/providers.py
```diff
@@ -0,0 +1,84 @@
...
+                "%s failure using %r: %r",
+                self.__class__.__name__,
+                getattr(self._transform, "__name__", self._transform),
+                e,
+            )
+            return ""
+
+    def __repr__(self) -> str:  # pragma: no cover - debug helper
+        transform_name = getattr(self._transform, "__name__", repr(self._transform))
+        return f"TransformProvider(base={self._base!r}, transform={transform_name})"
```
zas:
```suggestion
        return f"{self.__class__.__name__}(base={self._base!r}, transform={transform_name})"
```

picard/ui/itemviews/custom_columns/providers.py
```diff
@@ -0,0 +1,84 @@
...
+    def evaluate(self, obj: Item) -> str:
+        try:
+            return str(self._func(obj))
+        except (TypeError, ValueError, AttributeError) as e:
+            log.debug("%s failure for %r: %r", self.__class__.__name__, getattr(self._func, "__name__", self._func), e)
+            return ""
+
+    def __repr__(self) -> str:  # pragma: no cover - debug helper
+        func_name = getattr(self._func, "__name__", repr(self._func))
+        return f"CallableProvider(func={func_name})"
```
zas:
```suggestion
        return f"{self.__class__.__name__}(func={func_name})"
```

picard/ui/itemviews/custom_columns/script_provider.py
```diff
@@ -0,0 +1,142 @@
...
+        str
+            Computed value (empty on failure).
+        """
+        can_cache = True
+        avoid_cache_for_obj = False
+        try:
+            if obj in self._cache:
+                return self._cache[obj]
+        except TypeError as e:
+            debug("Weak cache lookup failed (non-weakrefable object): %r", e)
```
zas:
```suggestion
            log.debug("Weak cache lookup failed (non-weakrefable object): %r", e)
```

zas: More occurrences to fix

knguyen1: [Use self.__class__.__name__; use log.debug](https://github.com/metabrainz/picard/pull/2709/commits/d12412838837097246cdf5e7495fd5a530cd9b8b)

knguyen1: Hmm, https://github.com/metabrainz/picard/pull/2709/commits/132c7f98b3284ffb99b83709a4b2ec8ca54baf2d causing failures in CI... Need to check locally..

knguyen1: Failures seem unrelated to changes.  I'm going to ignore.

picard/ui/itemviews/__init__.py
```diff
@@ -411,6 +411,16 @@ def update_colums_text(self, color=None, bgcolor=None):
             else:
                 if column.align == ColumnAlign.RIGHT:
                     self.setTextAlignment(i, QtCore.Qt.AlignmentFlag.AlignRight | QtCore.Qt.AlignmentFlag.AlignVCenter)
+                # Support custom columns with provider evaluation
+                try:
+                    from picard.ui.itemviews.custom_columns import CustomColumn  # Local import to avoid cycles
```
zas: Though it would be better if moved outside the loop.

picard/ui/itemviews/custom_columns/registry.py
```diff
@@ -0,0 +1,246 @@
...
+        insert_after_key
+                Insert after this key if present, else append.
+        """
+        self._by_key[column.key] = column
+
+        # Local import: These column collections are mutable module-level objects that
+        # may be modified by plugins or other parts of the application at runtime.
+        # Importing at registration time ensures we always get the current state
+        # of these collections, not a stale reference from module load time.
+        targets = set(str(v).upper() for v in (add_to or {VIEW_FILE, VIEW_ALBUM}))
```
zas: `{VIEW_FILE, VIEW_ALBUM}` can be replaced by `RECOGNIZED_VIEWS`

picard/ui/itemviews/custom_columns/resolve.py
```diff
@@ -0,0 +1,284 @@
...
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
+from collections.abc import Callable
+from typing import Any
+
+from picard.item import Item
+from picard.log import debug
```
zas:
```suggestion
from picard import log
```

picard/ui/itemviews/custom_columns/resolve.py
```diff
@@ -0,0 +1,284 @@
...
+            The resolved value, or ``None`` if no resolver in the chain
+            produced a non-empty result.
+        """
+        if self.can_resolve(obj, simple_var, script, ctx, file_obj):
+            try:
+                result = self.resolve(obj, simple_var, script, ctx, file_obj)
+                if result:
+                    return result
+            except (TypeError, ValueError, AttributeError, KeyError) as e:
+                debug("ValueResolver.handle suppressed error: %r", e)
```
zas:
```suggestion
                log.debug("%s.handle() suppressed error: %r", self.__class__.__name__, e)
```

picard/ui/itemviews/custom_columns/resolve.py
```diff
@@ -0,0 +1,284 @@
...
+    def _compile_script(self, parser: ScriptParser, script: str) -> Any:
+        """Load functions best-effort and parse the script.
+
+        Any exception is handled by the caller; this method only logs failures
+        when loading functions, as this is non-fatal for parsing.
+        """
+        try:
+            parser.load_functions()
+        except Exception as e:  # pragma: no cover - defensive
+            debug("Failed to load script functions: %r", e)
```
zas:
```suggestion
            log.debug("Failed to load script functions: %r", e)
```

picard/ui/itemviews/custom_columns/script_provider.py
```diff
@@ -0,0 +1,143 @@
...
+        can_cache = True
+        avoid_cache_for_obj = False
+        try:
+            if obj in self._cache:
+                return self._cache[obj]
+        except TypeError as e:
+            log.debug("Weak cache lookup failed (non-weakrefable object): %r", e)
+            can_cache = False
+        # Avoid caching for album-like objects that are not fully loaded yet
+        if bool(getattr(obj, "is_album_like", False)) and not getattr(obj, "loaded", True):
```
zas: `bool()` isn't strictly needed here.

picard/ui/itemviews/custom_columns/script_provider.py
```diff
@@ -0,0 +1,143 @@
...
+    - Fallback id-cache: A bounded FIFO cache keyed by ``id(obj)`` for
+      objects that cannot be weakly referenced. Evictions use O(1)
+      ``deque.popleft``.
+    """
+
+    def __init__(
+        self,
+        script: str,
+        max_runtime_ms: int = 25,
+        cache_size: int = 1024,
```
zas: Maybe define those as `None` and use 2 class variables to store a default value:


```python
class ...
    DEFAULT_MAX_RUNTIME_MS = 25

    def ...
        max_runtime_ms: int | None = None,
        ...

        if max_runtime_ms is None:
            max_runtime_ms = self.DEFAULT_MAX_RUNTIME_MS
```

This way defaults can be changed more easily if needed (by subclassing or patching).

picard/ui/itemviews/custom_columns/script_provider.py
```diff
@@ -0,0 +1,143 @@
...
+        self._max_runtime_ms = max_runtime_ms
+
+        self._context_manager = ContextStrategyManager()
+        # Reuse a parser instance or factory through resolver chain
+        self._value_resolver = ValueResolverChain(parser=parser, parser_factory=parser_factory)
+
+        self._cache: WeakKeyDictionary[Item, str] = WeakKeyDictionary()
+        self._id_cache: dict[int, str] = {}
+        self._id_order: deque[int] = deque()
+        self._id_cache_max = max(16, int(cache_size))
```
zas: Default max can be a class variable too.

picard/ui/itemviews/custom_columns/shared.py
```diff
@@ -0,0 +1,102 @@
...
+
+
+# Public identifiers for views used in configuration and API
+VIEW_FILE: str = "FILE_VIEW"
+VIEW_ALBUM: str = "ALBUM_VIEW"
+
+# Default order when both are selected
+DEFAULT_ADD_TO: str = f"{VIEW_FILE},{VIEW_ALBUM}"
+
+RECOGNIZED_VIEWS: set[str] = {VIEW_FILE, VIEW_ALBUM}
```
zas, line 35: Could be `set(get_recognized_view_columns())`

zas (changes requested):
Making columns configurable is a great addition, just few more stylistic comments, but overall it looks good to me.

picard/ui/itemviews/custom_columns/shared.py
knguyen1: I actually disagree with this.  Current value of `RECOGNIZED_VIEWS` is lightweight and maintainable.  Setting `RECOGNIZED_VIEWS: set[str] = set(get_recognized_view_columns())` would cause us to also import `picard.ui.itemviews.columns` (and thus `PyQt`) at import time of `shared.py`.

The current constant is a lightweight, config-level token list; the dynamic mapping is intentionally fetched lazily where needed.

If you still want it, we can make the change, but I recommend against it; due to having a simple constant vs. not.

picard/ui/itemviews/__init__.py
knguyen1: Moved outsite loop: https://github.com/metabrainz/picard/pull/2709/commits/7f5bda463207f3b44367d9edab03e60dc736a448

picard/ui/itemviews/custom_columns/registry.py
knguyen1: [Replace {VIEW_FILE, VIEW_ALBUM} with RECOGNIZED_VIEWS](https://github.com/metabrainz/picard/pull/2709/commits/ea2ab73b0258459c3abd706a6dcb918295d08782)

picard/ui/itemviews/custom_columns/resolve.py
knguyen1: [Use log.debug; use __class__.__name__](https://github.com/metabrainz/picard/pull/2709/commits/dd47c771b07764b7f69a8bd8e5321036e5129e11)

picard/ui/itemviews/custom_columns/resolve.py
knguyen1: [Use log.debug; use __class__.__name__](https://github.com/metabrainz/picard/pull/2709/commits/dd47c771b07764b7f69a8bd8e5321036e5129e11)

picard/ui/itemviews/custom_columns/resolve.py
knguyen1: [Use log.debug; use __class__.__name__](https://github.com/metabrainz/picard/pull/2709/commits/dd47c771b07764b7f69a8bd8e5321036e5129e11)

picard/ui/itemviews/custom_columns/script_provider.py
knguyen1: [Remove bool(); not needed](https://github.com/metabrainz/picard/pull/2709/commits/347852198683fa04e8589be26c94beee185eca9e)

picard/ui/itemviews/custom_columns/script_provider.py
knguyen1: [Default max and min should be injectible class vars](https://github.com/metabrainz/picard/pull/2709/commits/16325c02a8f389f4175a953e111552b976e34c80)

picard/ui/itemviews/custom_columns/script_provider.py
knguyen1: [Default max and min should be injectible class vars](https://github.com/metabrainz/picard/pull/2709/commits/16325c02a8f389f4175a953e111552b976e34c80)

picard/ui/itemviews/custom_columns/shared.py
zas: Ok, makes sense, keep it as is then.

picard/ui/itemviews/custom_columns/context.py
```diff
@@ -0,0 +1,106 @@
...
+    def can_handle(self, obj: Item) -> bool:
+        return isinstance(obj, Cluster)
+
+    def make_context(self, obj: Item) -> tuple[Any | None, Any | None]:
+        return (getattr(obj, 'metadata', None), None)
+
+
+class DefaultContextStrategy(ContextStrategy):
+    def can_handle(self, obj: Item) -> bool:
+        return True
```
phw: Other strategies above do the same and extract `obj.metadata`, relying on our `MetadataItem` implementation. I think it would be fine to have `DefaultContextStrategy.can_handle` check for `hasattr(obj, 'metadata')` or on instance of `MetadataItem` and then get rid of `AlbumContextStrategy` and `ClusterContextStrategy`.

picard/ui/itemviews/custom_columns/factory.py
```diff
@@ -0,0 +1,273 @@
...
+def make_script_column(
+    title: str,
+    key: str,
+    script: str,
+    *,
+    width: int | None = None,
+    align: ColumnAlign = ColumnAlign.LEFT,
+    always_visible: bool = False,
+    max_runtime_ms: int = 25,
+    cache_size: int = 1024,
```
zas: Those can be defined to `None` so the default value is used, after the change in https://github.com/metabrainz/picard/pull/2709/commits/16325c02a8f389f4175a953e111552b976e34c80

zas (changes requested):

picard/ui/itemviews/custom_columns/factory.py
knguyen1: [Make cache_size/max_runtime_ms default to None](https://github.com/metabrainz/picard/pull/2709/commits/8c18b75c17ac6bd9aa57b5996364d237d0d75ed7)

zas:
I reverted Ukrainian translation change in master, since it breaks everything, but there's also a problem with tests unrelated to the reverted change. Tests run ok locally though.

phw:
I don't think it was the translation file per se. Tests passed for that PR: https://github.com/metabrainz/picard/actions/runs/17271183163 . So not sure what changed.

For other failures I suspect there is some issue with some tests not being fully independent. The tests in CI always run randomized. We already had this last week that tests sometimes failed and on retry worked again. We need to investigate and fix the tests.

picard/ui/itemviews/custom_columns/registry.py
```diff
@@ -0,0 +1,249 @@
...
+
+    def __init__(self):
+        self._by_key = {}
+
+    def register(
+        self,
+        column: CustomColumn,
+        *,
+        add_to: set[str] | frozenset[str] | list[str] | tuple[str, ...] | None = None,
+        insert_after_key: str | None = None,
```
phw: I don't think that we need this. Even less so when exposed on the UI. I tested this on the UI branch, and this is super confusing. For once the input to enter the insert after is hard to use. You need to now the internal column names, and even I had to look this up in code.

The second thing is that users can rearange the columns anyway. So if you set this "insert_after_key" the column in code ends up behind the specified one, but this does not necessarily need to be where the column is displayed once the user reorders it.

I would suggest to completely get rid of that concept and add columns to the end.

picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -0,0 +1,296 @@
...
+        Returns a tuple (tail, original_lower) so items compare by the tail
+        (without article) and tie-break by the original lowercased value.
+        """
+        v = (self._base.evaluate(obj) or "").strip()
+        lv = v.casefold()
+        for art in self._articles:
+            prefix = f"{art} "
+            if lv.startswith(prefix):
+                return (lv[len(prefix) :], lv)
+        return (lv, lv)
```
phw, line 162: A bit off-topic to this PR, but we should see that we unify this code with the implementation we have in functions (see the `_delete_prefix` helper). The implementation above looks better and more performant, though.

picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -0,0 +1,296 @@
...
+        super().__init__(base)
+        self._seed = seed
+
+    def __repr__(self) -> str:  # pragma: no cover - debug helper
+        return f"{self.__class__.__name__}(base={self._base!r}, seed={self._seed})"
+
+    def sort_key(self, obj: Item):  # pragma: no cover - thin wrapper
+        """Return deterministic pseudo-random sort key for item."""
+        v = self._base.evaluate(obj) or ""
+        h = hashlib.blake2b(v.encode("utf-8"), digest_size=8, person=str(self._seed).encode("utf-8")).digest()
```
phw: Should this result maybe be cached?

picard/ui/itemviews/custom_columns/__init__.py
```diff
@@ -0,0 +1,76 @@
...
+from picard.ui.itemviews.custom_columns.column import CustomColumn
+from picard.ui.itemviews.custom_columns.factory import (
+    _create_custom_column,
+    make_callable_column,
+    make_field_column,
+    make_provider_column,
+    make_script_column,
+    make_transformed_column,
+)
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
```
rdswift: Shouldn't this have a trailing comma to force the imports to separate lines?

picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -0,0 +1,296 @@
...
+        self._base = base
+
+    def evaluate(self, obj: Item) -> str:
+        """Return evaluated text value for item."""
+        return self._base.evaluate(obj)
+
+    def __repr__(self) -> str:  # pragma: no cover - debug helper
+        return f"{self.__class__.__name__}(base={self._base!r})"
+
+
```
phw, line 67: There are quite a few sort key providers below. Will we actually need them? I'm not much a fan of adding code just in case it might be used in the future. Especially the random sort order seems to be less useful.

Also the inverted providers seem odd, as the user can change the order anyway.

The one I miss actually is for our most common sorting `ColumnSortType.NAT`. Or will this be handled differently?

One thing this change might open up is refactoring the standard columns in the future as well to use those sorting algorithms. The `ArticleInsensitiveAdapter` looks specifically interesting.

picard/ui/itemviews/custom_columns/column.py
```diff
@@ -0,0 +1,76 @@
...
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Custom column type that delegates value and optional sort to a provider."""
+
+from __future__ import annotations
+
+from picard.ui.columns import Column, ColumnAlign, ColumnSortType
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
```
rdswift: Add trailing comma?

picard/ui/itemviews/custom_columns/context.py
```diff
@@ -0,0 +1,106 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Context extraction strategies for different Picard item types."""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/factory.py
```diff
@@ -0,0 +1,273 @@
...
+"""Factory helpers to create common custom column types."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+
+from picard.item import Item
+from picard.script import ScriptParser
+
+from picard.ui.columns import ColumnAlign, ColumnSortType
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/factory.py
```diff
@@ -0,0 +1,273 @@
...
+from picard.ui.itemviews.custom_columns.column import CustomColumn
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/factory.py
```diff
@@ -0,0 +1,273 @@
...
+from picard.ui.itemviews.custom_columns.providers import CallableProvider, FieldReferenceProvider, TransformProvider
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/protocols.py
```diff
@@ -0,0 +1,63 @@
...
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Protocols for custom column value providers and optional sort capability."""
+
+from __future__ import annotations
+
+from typing import Protocol, runtime_checkable
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/resolve.py
```diff
@@ -0,0 +1,284 @@
...
+
+"""Value resolution chain for scripts: object, context, then parser.
+
+Adds support for injecting a reusable script parser and caching compiled
+scripts to improve stability and performance.
+"""
+
+from __future__ import annotations
+
+from abc import ABC, abstractmethod
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/shared.py
```diff
@@ -0,0 +1,102 @@
...
+
+    Returns
+    -------
+    dict[str, Any]
+        Mapping from view id (e.g., ``FILE_VIEW``) to the corresponding
+        columns collection. Performed via a local import to avoid stale
+        references and import-time side effects.
+    """
+
+    from picard.ui.itemviews.columns import ALBUMVIEW_COLUMNS, FILEVIEW_COLUMNS
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -0,0 +1,296 @@
...
+"""Sorting adapters that add `sort_key` support to value providers."""
+
+from __future__ import annotations
+
+from collections.abc import Callable
+import hashlib
+
+from picard.item import Item
+
+from picard.ui.itemviews.custom_columns.protocols import ColumnValueProvider, SortKeyProvider
```
rdswift: Trailing comma?

test/test_custom_columns_sorting_adapters.py
```diff
@@ -0,0 +1,281 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from __future__ import annotations
+
+from collections.abc import Callable
+import dataclasses
+
+import pytest
+
+from picard.ui.columns import ColumnAlign, ColumnSortType
```
rdswift: Trailing comma?

picard/ui/itemviews/custom_columns/resolve.py
```diff
@@ -0,0 +1,284 @@
...
+        try:
+            # Compile once per script; parsing with functions=True requires functions already loaded.
+            if script not in self._compiled_by_script:
+                self._compiled_by_script[script] = self._compile_script(parser, script)
+
+            # Set evaluation context and evaluate compiled expression
+            parser.context = ctx
+            parser.file = file_obj
+            return self._compiled_by_script[script].eval(parser)
+        except Exception as e:
```
phw: It might be fine to specifically handle `ScriptError` here. Or what other exceptions do we expect?

rdswift:
What happens if multiple custom columns are registered with the same key?

What happens if custom columns are registered with the same key as a standard column (e.g. 'title') that is referenced in another custom column's `insert_after_key` parameter?

picard/ui/itemviews/custom_columns/sorting_adapters.py
knguyen1: Can you decide which ones you'd like to keep?  Making changes here will affect documentation: https://github.com/metabrainz/picard-docs/pull/234/files Open to deleting some but I suggest we keep them so as to offer them via API for plugins/extenders.

phw: I didn't consider this. Of course, when we see this as an API more are useful.

I still would remove the random one. This one seems particular complex with the potential of higher performance impact, while overall looking more like a demonstration of the flexibility, but not so much useful. Once we make this part of the API it will be hard to remove, so I'd rather start without it.

I'm also still not quite sure why we need the inverse sorting variants, as the sort direction can be controlled by the user in the UI anyway.

phw:
@knguyen1 Tests on master should now be more stable.

test/test_custom_columns_sorting_adapters.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/sorting_adapters.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/shared.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/resolve.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/protocols.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/factory.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/factory.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/factory.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/context.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/column.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/__init__.py
knguyen1: [Add trailing commas to import](https://github.com/metabrainz/picard/pull/2709/commits/8903ee300ed7e4313d717b864485891a664f1856)

picard/ui/itemviews/custom_columns/sorting_adapters.py
knguyen1: >I'm also still not quite sure why we need the inverse sorting variants, as the sort direction can be controlled by the user in the UI anyway.

Linux weirdly reverses sorting direction (at least for `float`).  See: https://github.com/metabrainz/picard/pull/2696#issuecomment-3153076894

I would keep these reverse sorters in the API.  Happy to remove if you think it's better.

picard/ui/itemviews/custom_columns/context.py
knguyen1: [Remove redundant strategies; apply DefaultContextStrategy for metadataitem](https://github.com/metabrainz/picard/pull/2709/commits/14a0a0b72afcdd1bd3132d779a1ca9a55558c19f)

picard/ui/itemviews/custom_columns/registry.py
knguyen1: [Remove insert_after_key concept](https://github.com/metabrainz/picard/pull/2709/commits/22ff6930baa84e092e8c7e5e705b219558d772b5)

picard/ui/itemviews/custom_columns/sorting_adapters.py
knguyen1: [Remove RandomSortAdapter](https://github.com/metabrainz/picard/pull/2709/commits/d0e437b831533cb5761d3918384cd49b451c9322)

picard/ui/itemviews/custom_columns/sorting_adapters.py
knguyen1: [Remove RandomSortAdapter](https://github.com/metabrainz/picard/pull/2709/commits/d0e437b831533cb5761d3918384cd49b451c9322)

picard/ui/itemviews/custom_columns/resolve.py
knguyen1: [Do not catch blind exceptions](https://github.com/metabrainz/picard/pull/2709/commits/c8e776df563bbcf4ce23554170527cf11cd7cbb6)

zas:
I was thinking... what about using this in a more generic manner, that is for ALL columns (that is standard ones)? And drop redundant code.

@knguyen1 @phw What do you think?

knguyen1:
> what about using this in a more generic manner, that is for ALL columns (that is standard ones)? And drop redundant code.

That's a good idea; can address it via a new PR after merge.  The API will have to be expanded for image columns (`MatchQualityColumn`) and status icons (`Title` column)

Converting all text columns to the custom API is straightforward. Sorting parity is achievable with existing adapters.  Example conversions:

- Album (NAT):

 ```python
make_provider_column(N_("Album"),
    "album",
    FieldReferenceProvider("album"),
    sort_type=ColumnSortType.NAT)
 ```
 
 - Size (numeric sort):
 
 ```python
make_provider_column(
        N_("Size"), "~filesize",
        NumericSortAdapter(FieldReferenceProvider("~filesize")),
        align=ColumnAlign.RIGHT
    )
 ```

It would help unify/simplify the code quite a bit.  e.g. today there is the bespoke sorting key for filesize/bitrate:

```python
def _sortkey_length(obj):
    return obj.metadata.length or 0
# ...
def _sortkey_filesize(obj):
    try:
        return int(obj.metadata['~filesize'] or obj.orig_metadata['~filesize'])
    except ValueError:
        return 0
# ...
def _sortkey_bitrate(obj):
    try:
        return float(obj.metadata['~bitrate'] or obj.orig_metadata['~bitrate'] or 0)
    except (ValueError, TypeError):
        return 0
# ...
def _sortkey_match_quality(obj):
    # percent logic ...
```

Looks like this: `NumericSortAdapter(FieldReferenceProvider("~filesize")),` under the new API.

zas (approved):
LGTM

zas:
@knguyen1It would be convenient if we could chat about Picard on matrix (see https://musicbrainz.org/doc/Communication/ChatBrainz).
https://matrix.to/#/#musicbrainz-picard-dev:chatbrainz.org is dedicated to Picard
https://matrix.to/#/#metabrainz:chatbrainz.org is where we do MeB meetings
I'm `zas` there.

phw (approved):
Thanks for the updates. LGTM

PR #2710, opened 2025-08-18
Explicitly set the "strict" parameter for builtin zip()

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Explicitly set the "strict" parameter for builtin [zip()](https://docs.python.org/3/library/functions.html#zip). This parameter is available since Python 3.10, and Ruff wants it to be used in Python 3.10 mode. Which makes sense to explicitly state the behavior.

Since our current minimum supported version is also 3.10 we should also run Ruff in Python 3.10 mode.

# Solution

Use strict=True for code we internally control. Use strict=False for external data, where we have no control over the input. Specifically for the case of reading `/proc/sys/dev/cdrom/info` on Linux the input should be equal amount of elements. But if it wouldn't be it would still be better to return all devices that had both entries.

For the ID3 SYLT case we could maybe also be strict, but again I thought it better to be more relaxed here and don't change existing behavior.

### DIFF
--- a/picard/util/cdrom.py
+++ b/picard/util/cdrom.py
@@ -83,7 +83,7 @@ def _parse_linux_cdrom_info(f):
             if line.startswith(CAN_PLAY_AUDIO):
                 drive_audio_caps = [v == '1' for v in line[len(CAN_PLAY_AUDIO) :].split()]
                 break
-    yield from zip(drive_names, drive_audio_caps)
+    yield from zip(drive_names, drive_audio_caps, strict=False)
 
 
 if IS_WIN:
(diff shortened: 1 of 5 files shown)


picard/util/cdrom.py
```diff
@@ -83,7 +83,7 @@ def _parse_linux_cdrom_info(f):
             if line.startswith(CAN_PLAY_AUDIO):
                 drive_audio_caps = [v == '1' for v in line[len(CAN_PLAY_AUDIO) :].split()]
                 break
-    yield from zip(drive_names, drive_audio_caps)
+    yield from zip(drive_names, drive_audio_caps, strict=False)
```
zas, line 86: Instead of just using `strict=False`, we could use `strict=True` and catch the `ValueError` exception, log an error, and fallback on `strict=False`.

phw: I don't really think that has any advantage really. Would be a very artificial case.

PR #2711, opened 2025-08-19
PICARD-3107: Add profile save warning to option settings

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds a warning to indicate highlighted settings will be saved to an option profile.

# Problem

Comment by @Aerozol at https://github.com/metabrainz/picard/pull/2702#issuecomment-3193912644 suggested the addition of a warning to indicate that highlighted settings will be saved to an option profile.

* JIRA ticket (_optional_): PICARD-3107

# Solution

Adds a warning to indicate highlighted settings will be saved to an option profile if there are highlighted settings on the option page.  Also include a link to the option profiles help page.

Screen shot with options from the page saved to a single profile:

image

Screen shot with options from the page saved to multiple profiles:

image

# Action

Additional actions required:


ui/options.ui
```diff
@@ -56,19 +56,115 @@
...
+            <property name="text">
+             <string>x</string>
+            </property>
+           </widget>
+          </item>
+          <item>
+           <widget class="QLabel" name="profile_warning_text">
+            <property name="sizePolicy">
+             <sizepolicy hsizetype="Preferred" vsizetype="Preferred">
+              <horstretch>0</horstretch>
```
zas: I think it would look better with:
```suggestion
              <horstretch>1</horstretch>
```

It would modify the rendering:

image

picard/ui/options/dialog.py
```diff
@@ -400,6 +412,34 @@ def set_profiles_button_and_highlight(self, page):
...
+                for item in working_profiles:
+                    if not item['enabled']:
+                        continue
+                    profile_id = item['id']
+                    profile_title = item['title']
+                    if profile_id not in working_settings:
+                        continue
+                    profile_settings = working_settings[profile_id]
+                    if opt.name in profile_settings:
+                        profile_set.add(profile_title)
```
zas:
```suggestion
                        profile_set.add(item['title'])
```

And get rid of `profile_title = item['title']` above.

zas (changes requested):
It looks good to me, just minor stylistic comments.

ui/options.ui
zas: Also, just noticed, the icon (?) isn't really centered on my system (it seems it is on yours),

Aerozol:
That was quick! Instead of 'multiple profiles' can we have:
"The highlighted settings will be applied to: [profile 1], [profile 2], [profile 3]"

If it's only one profile, just list the one. This could simplify things - if that's more complicated than I'm imagining then this looks like a good improvement, as-is, as well!

rdswift:
Revised screen shot with requested changes:

image

rdswift:
> That was quick! Instead of 'multiple profiles' can we have: "The highlighted settings will be applied to: [profile 1], [profile 2], [profile 3]"
> 
> If it's only one profile, just list the one. This could simplify things - if that's more complicated than I'm imagining then this looks like a good improvement, as-is, as well!

Done.  I wasn't sure if you were suggesting to include each profile name in square brackets or were just using them for examples.  For now I have enclosed the profile names in quotes.  Let me know if you would prefer to have them enclosed in square brackets instead.  It's a simple change.

ui/options.ui
rdswift: Vertically centered, or did you mean tight to the right hand side of the text?  With the new horizontal stretch setting, it looks the same here (the help button pushed to the far right) with text that isn't wrapped.

Aerozol:
Yes, the brackets were just for examples. Looks good!
If we are ending each profile with (1), (2), etc, it might not be necessary to have the quotes either, but I am in danger of entering the realm of bikeshedding so I am happy for you to test and make the call.

rdswift:
> If we are ending each profile with (1), (2), etc, it might not be necessary to have the quotes either, but I am in danger of entering the realm of bikeshedding so I am happy for you to test and make the call.

Actually, you can name the profiles however you want the same way you can name scripts.  The "My profile (x)" is just the default name when you create a new one.  For that reason, my personal preference is to enclose them in quotes.  Also, you can include a comma in the profile name, so not having the quotes could make it confusing.  :wink:

Aerozol:
Ah, sorry I forgot about that, because I dropped the (1), (2) etc from all my profile names! So I thought you were numbering them here to show the order. All your reasoning sounds good to me :)

zas (dismissed):
LGTM

rdswift:
This may be way to nit-picky, but when there are multiple profiles I listed them in alphabetical order.  Would this be better in the order that they are applied (top to bottom or bottom to top)?

rdswift:
Also, is this something we should backport to the final 2.x release?  I already have the code changes done, but was waiting for this to be finalized before submitting anything so that, if we do decide to include it in 2.x, it would have the same look-and-feel.

zas:
> This may be way to nit-picky, but when there are multiple profiles I listed them in alphabetical order. Would this be better in the order that they are applied (top to bottom or bottom to top)?

Yes, I think it would be better to keep order they apply

zas:
> Also, is this something we should backport to the final 2.x release? I already have the code changes done, but was waiting for this to be finalized before submitting anything so that, if we do decide to include it in 2.x, it would have the same look-and-feel.

It's a small improvement, if you have the code ready for 2.x then it would be nice to have it in 2.x (not sure when we'll do such release, but it may happen, so better have it).

rdswift:
> It's a small improvement, if you have the code ready for 2.x then it would be nice to have it in 2.x (not sure when we'll do such release, but it may happen, so better have it).

Added https://github.com/metabrainz/picard/pull/2713.

PR #2712, opened 2025-08-19
PICARD-3109: Add in missing profile when saving settings

rdswift:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

Option saving would fail when saving a new enabled profile with settings selected.  Steps to reproduce:

1. Remove all profiles and save the changes by clicking "Make It So!".
2. Re-open the option settings.  Confirm that there are no profiles.
3. Create a new profile with one or more profile settings selected.
4. Enable the new profile.
5. Close the option settings by clicking "Make It So!"
6. The operation will fail with a `KeyError` because the system is trying to save the values to a profile that has not yet been added to the config.settings.

* JIRA ticket (_optional_): PICARD-3109

# Solution

Save the `profiles` page before any of the other option pages.  This will ensure that the proper profile keys are in place when the other option settings are saved.

# Action

Additional actions required:

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -439,7 +439,7 @@ class OptionsDialog(PicardDialog, SingletonDialog):
                 self._show_page_error(page, e)
                 return
 
-        for page in sorted(self.loaded_pages, key=lambda x: x.SORT_ORDER):
+        for page in sorted(self.loaded_pages, key=lambda p: (p.SORT_ORDER, p.NAME)):
             try:
                 page.save()
             except Exception as e:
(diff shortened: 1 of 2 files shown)


picard/ui/options/dialog.py
```diff
@@ -439,14 +439,13 @@ def accept(self):
                 self._show_page_error(page, e)
                 return
 
-        for page in self.loaded_pages:
+        for page in sorted(self.loaded_pages, key=lambda x: x.SORT_ORDER):
```
zas: Just to be sure the sort order is constant, better sort on `SORT_ORDER` and `NAME` as we do elsewhere, it shouldn't have much impact in this case, but it guarantees pages are saved is the same order on each call, if 2 pages have same sort order number (they shouldn't have the same name).

```suggestion
        for page in sorted(self.loaded_pages, key=lambda p: (p.SORT_ORDER, p.NAME)):
```

zas:
@rdswift Maybe create a ticket for this.

rdswift:
> @rdswift Maybe create a ticket for this.

Done.  PICARD-3109 entered.

PR #2716, opened 2025-08-21
PICARD-3108: Add quick settings menu and the ability to select the items to include

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**:  Adds a quick settings menu and alows users to select the (boolean) items that appear in the menu.

# Problem

Users have requested a way to quickly toggle selected option settings directly from the Options menu rather than having to open the Option Settings window.

* JIRA ticket (_optional_): PICARD-3108

# Solution

Add a `Quick settings` menu item under the `Options` menu, and provide a means of selecting the boolean settings items that appear in the `Quick settings` menu.  Selected items are set by profile on the `Profiles` page, through a dialog opened by clicking a new `Quick menu` button.  The list of options available for selection is automatically generated by including all boolean (checkbox) items from all options pages, and items are grouped by options page in the selection tree.

Note that the quick settings items are taken from the top enabled option profile, or the default of `save_images_to_tags` and `save_images_to_files` options.  If there are no quick setting options selected in the top enabled option profile, the `Quick settings` menu will be disabled.

# Action

Additional actions required:


zas:
I didn't do a deep review, but a quick test. I wonder about another approach:

- Move Quick Menu action to a pane in Options > User Interface
- List all possible options there, and let the user selects them from there
- Add selected options to quick menu
- Initial state of the quick option is always the one the applied profiles
- When toggled it automatically applies to last profile
- If the last profile is disabled, the value is set to the one of all previous profiles.


Example:

Options > UI > Quick Menu shows:

A [ ]
B [ ]

There's no profile enabled, A defaults to True, and B to False

User selects A and B to appear in Quick Menu:

A [x]
B [x]

The Quick Menu shows:

A [V]
B [ ]

If the user creates a profile and add A option to it, and toggles it, the profile has A = False, then he disables it A is True again (and Quick Menu reflects that).

What do you think?

Main point is: quick menu is set in UI options rather than profiles, and toggling options in quick menu always applies to active profile (if any, else default config)

rdswift:
> Main point is: quick menu is set in UI options rather than profiles, and toggling options in quick menu always applies to active profile (if any, else default config)

I woke up in the middle of the night thinking something similar.  This option should be moved to the UI options and handled just like any other option in the profiles.  I'll make the change.

rdswift:
Implemented this as a new options page.

image

zas:
I did a quick test with your new implementation and it looks much better to me, more intuitive at least.
I noticed something prolly unrelated about Options wanting to restart Picard after theme change (while I did change it).

Overall, I think this approach works better for end user (and it still uses profiles).

I'll do more extensive tests later.

rdswift:
Updated screen shots:

image

---

image

Aerozol:
Screenshots look EXCELLENT! Wonderful work.

PR #2720, opened 2025-08-28
Translated Ukrainian strings not displayed in Picard

Oleghishak:
Hello!
This is a new pull request for Picard. The translated Ukrainian strings remain untranslated in the Picard application and need to be fixed. Screenshots are attached.
Thanks.

### DIFF
--- a/po/uk.po
+++ b/po/uk.po
@@ -15,19 +15,15 @@ msgstr ""
 "Project-Id-Version: MusicBrainz\n"
 "Report-Msgid-Bugs-To: EMAIL@ADDRESS\n"
 "POT-Creation-Date: 2025-08-17 03:05+0200\n"
-"PO-Revision-Date: 2025-09-01 07:07+0000\n"
-"Last-Translator: oleh_hishak <ohishak@gmail.com>\n"
-"Language-Team: Ukrainian <https://translations.metabrainz.org/projects/"
-"picard/3/app/uk/>\n"
+"PO-Revision-Date: 2025-08-21 13:09+0300\n"
+"Last-Translator: Nerten <alex@nerten.info>\n"
+"Language-Team: Ukrainian <https://translations.metabrainz.org/projects/picard/3/app/uk/>\n"
 "Language: uk\n"
 "MIME-Version: 1.0\n"
 "Content-Type: text/plain; charset=UTF-8\n"
 "Content-Transfer-Encoding: 8bit\n"
-"Plural-Forms: nplurals=4; plural=(n % 1 == 0 && n % 10 == 1 && n % 100 != 11 "
-"? 0 : n % 1 == 0 && n % 10 >= 2 && n % 10 <= 4 && (n % 100 < 12 || n % 100 > "
-"14) ? 1 : n % 1 == 0 && (n % 10 ==0 || (n % 10 >=5 && n % 10 <=9) || (n % "
-"100 >=11 && n % 100 <=14 )) ? 2: 3);\n"
-"X-Generator: Weblate 5.13\n"
+"Plural-Forms: nplurals=4; plural=(n % 1 == 0 && n % 10 == 1 && n % 100 != 11 ? 0 : n % 1 == 0 && n % 10 >= 2 && n % 10 <= 4 && (n % 100 < 12 || n % 100 > 14) ? 1 : n % 1 == 0 && (n % 10 ==0 || (n % 10 >=5 && n % 10 <=9) || (n % 100 >=11 && n % 100 <=14 )) ? 2: 3);\n"
+"X-Generator: Poedit 3.7\n"
 "Generated-By: Babel 2.9.1\n"
 
 #: picard/album.py:191
@@ -66,7 +62,7 @@ msgstr "Немає відповідних випусків для кластер
 #: picard/cluster.py:278
 #, python-format
 msgid "Looking up the metadata for cluster %(album)s…"
-msgstr "Пошук метаданих для кластера%(album)s…"
+msgstr "Пошук метаданих для кластера %(album)s…"
 
 #: picard/cluster.py:335
 msgid "Unclustered Files"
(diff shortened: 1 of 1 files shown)


phw:
Thanks for this again. Will need some time before merging this, though.

We are dealing with some test errors after we had merged the last PR, somehow our test pipeline started to fail processing these specific po files. Not yet sure what happened, locally all seems fine. But I'll make sure the latest translation gets merged soon

Oleghishak:
Strangely, this string is not present even in the English localization file:
"You have changed the interface language. You have to restart Picard in order for the change to take effect."

phw:
@Oleghishak I think you are mixing latest translation files from here with the release version of Picard. Those strings have changed, so if you use the po file for the release version 2.13 the translations won't show up.

Oleghishak:
You are right — the issue is that I used the latest translation files together with Picard version 2.13, so some translations don’t appear.

phw:
@Oleghishak This didn't go well, because we now have changes both from weblate and here. And one often cannot easily solve merge conflicts in po files. Is the file you provide here complete, or would there be some translations only be in Weblate?

If you say this version here is complete I'll use it and overwrite the other changes.

po/uk.po
```diff
@@ -1004,23 +995,23 @@ msgstr "Архів обкладинки"
 
 #: picard/coverart/providers/caa.py:184
 msgid "Cover Art Archive: Release"
-msgstr "Архів обкладинок: реліз"
+msgstr "Архів обкладинок: Реліз"
 
 #: picard/coverart/providers/caa_release_group.py:44
 msgid "Cover Art Archive: Release Group"
-msgstr "Архів обкладинок: група релізів"
+msgstr "Архів обкладинок: Release Group"
```
phw: Sure about this? Most languages find some kind of translation for this expression.

po/uk.po
```diff
@@ -1029,85 +1020,64 @@ msgstr "Пресет %(number)d: %(title)s"
...
 
 #: picard/script/__init__.py:207
 msgid "[album artist]/[album]/[track #]. [title]"
-msgstr "[виконавець альбому]/[альбом]/[трек #]. [назва]"
+msgstr "[album artist]/[album]/[track #]. [title]"
 
 #: picard/script/__init__.py:219
 msgid "[album artist]/[album]/[disc and track #] [artist] - [title]"
-msgstr "[виконавець альбому]/[альбом]/[# диска та треку] [виконавець] - [назва]"
+msgstr "[album artist]/[album]/[disc and track #] [artist] - [title]"
```
phw: Translating the descriptions here is fine. This is shown to the user as a help to understand how the file names will be constructed. So it is helpful to show it in native language.

po/uk.po
```diff
@@ -2423,7 +2092,7 @@ msgstr "Будь ласка, пожертвуйте"
 
 #: picard/ui/aboutdialog.py:114
 msgid "Thank you for using Picard. Picard relies on the MusicBrainz database, which is operated by the MetaBrainz Foundation with the help of thousands of volunteers. If you like this application please consider donating to the MetaBrainz Foundation to keep the service running."
-msgstr "Дякуємо, що користуєтеся Picard. Picard використовує базу даних MusicBrainz, якою керує фонд MetaBrainz за допомогою тисяч волонтерів. Якщо вам подобається ця програма, будь ласка, подумайте про пожертвування фонду MetaBrainz, щоб підтримувати роботу сервісу."
+msgstr ""
```
phw: Is this removal intentional? Also many of the other strings.

phw (changes requested):
@Oleghishak Ok, I did some updates. First I updated the current master to be in sync with Weblate again. So we are now on the latest state in Weblate.

Then I updated your branch here, rebased it against the latest master and fixed the line ending of the file. The file you committed had Windows style line endings, which broke the Weblate processing and also caused this huge diff were every line was shown as changed.

With this change we can now see clearly what has changed in your uk.po file provided here. Please take a look at 
https://github.com/metabrainz/picard/pull/2720/files

There are many additions. But also multiple cases where existing translations get removed. I can't judge whether this is intentional or not. If the translations were so bad that they better get removed, then  it might be fine. But if the translations just need some work it might have been better to mark them as fuzzy instead of deleting.

I can help cleaning this file up and merging it with the existing one, but I need your guidance on what to do and which translations to keep or overwrite.

po/uk.po
Oleghishak: I don’t understand why this is happening.
Ignore these replacements.

Oleghishak: I downloaded the EN .pot file and the UK .po file from Weblate. In Poedit, I loaded the Ukrainian .po file and updated it from the EN .pot file — no changes were added. That means the Ukrainian .po file on Weblate is correct. 
Thanks, you corrected the Ukrainian.

Oleghishak:
Hi! 
I’ve resolved 4 out of 5 conflicts in the `uk.po` on Weblate. I can’t resolve the first one because it doesn’t appear there. I hope this will resolve the conflicts.

phw:
@Oleghishak Thanks. I'll take care of merging the new and changed translations without the deletion.

po/uk.po
```diff
@@ -715,77 +706,77 @@ msgstr "Вилучити позначки APEv2 з файлів AAC"
...
+msgstr "Символ з'єднання ID3v2.3"
 
 #: picard/options.py:466 picard/ui/forms/ui_options_tags_compatibility_id3.py:125
 msgid "ID3v2 text encoding"
 msgstr "Кодування тексту ID3v2"
 
 #: picard/options.py:467 picard/ui/forms/ui_options_tags_compatibility_id3.py:131
 msgid "Save iTunes compatible grouping and work"
-msgstr "Зберігати сумісність із групуванням і роботою iTunes"
+msgstr "Збережіть групи, сумісні з iTunes, та працюйте"
```
phw: A question here: The "work" in the original sentence refers to a [Work as defined for MusicBrainz](https://musicbrainz.org/doc/Work):

> "a work is a distinct intellectual or artistic creation"

So it is a work in the sense of an opus. I think opus was not used, as it was considered to tied to classical music, and the usage in MB is broader. The string describes an option to write the two tags "grouping" and "work" slightly different to ID3 files, to maintain compatibility with different iTunes versions (they changed their internal tag mapping for those at some point in the past).

From what I understand from looking up  "працюйте" it is the imperative of "to work". But also "робота" from the original does not seem to fit, it rather seems to mean "work" as in labor, or maybe operation or function in a computing context, right? The dictionary suggest something like "твір". Would this fit better?

In my cleanup I'll keep the original, but mark it as fuzzy. But please let me know about a better translation :)

phw:
@Oleghishak I updated the strings as announced. Your changes and additions have been merged, but I omitted the deletions. Please check the latest version.

About my comment of this one string and the translation of "work" I marked the translation as fuzzy and added a comment on https://translations.metabrainz.org/translate/picard/3/app/uk/?checksum=792cb0ab7d3e79c5

po/uk.po
Oleghishak: “Thank you for your attention, I’ve corrected it. Normally, ‘work’ translates as ‘робота’ or ‘праця’, but in this context, ‘твір’ is more appropriate.

phw: Thank you very much

PR #2722, opened 2025-08-29
PICARD-2729: Fix/refactor `sanitize_date`; fix bug with shifting unknown month

knguyen1:
# Summary

* This is a…
  * [x] Bug fix
  * [x] Refactoring
  * [ ] Feature addition
  * [ ] Other
  - Fix date sanitization shifting of unknown components and add robust normalization for multiple input formats; refactor into small helpers and add comprehensive pytest coverage.

# Problem

> **Philipp Wolfer**: The cases 2005-00-12" => "2005-12" and "0000-00-12" => "0012" are a bug, though. This should be fixed.

- Dates with unknown month/day were incorrectly shifted:
  - "2005-00-12" → "2005-12" and "0000-00-12" → "0012".
- Tests lacked focused, parametric coverage for date normalization and bug cases.

# Solution

- Implement `picard.util.sanitize_date` that preserves unknown components ("00") without shifting, drops only truly unknown trailing parts, and supports: "YYYY", "YYYY-MM", "YYYY-MM-DD", unambiguous "YYYY-DD-MM", "DD/MM/YYYY" vs "MM/DD/YYYY" (heuristic), and compact "YYYYMMDD"/"YYYYDDMM".
- Refactor into helpers: `_parse_pure_year`, `_parse_iso_like`, `_parse_slash_separated`, `_parse_compact_eight`, `_format_from_components`.
- Add new pytest module `test/test_dateutil_normalize.py` using fixtures and `pytest.mark.parametrize`.
- Update `sanitize_date` docstring (NumPy style) to match behavior and examples.

# Action

Additional actions required:
* [ ] Update Picard documentation (please include a reference to this PR)

**Note**: Will need a follow-up PR to add option UI's and to close #2406.  refactored `sanitize_date` sets up ability to sanitize dates or not.

### DIFF
--- a/picard/util/datestr_util.py
+++ b/picard/util/datestr_util.py
@@ -147,11 +147,13 @@ def _parse_compact_eight(value: str) -> str | None:
     m = re.fullmatch(r"(\d{4})(\d{2})(\d{2})", value)
     if not m:
         return None
+
     y = _clamp_year(int(m.group(1)))
+    if y is None:
+        return ""
+
     a = int(m.group(2))
     b = int(m.group(3))
-    if y in (None, 0):
-        return ""
     if _is_valid_month(a) and _is_valid_day(b) and a != 0:
         return f"{y:04d}-{a:02d}-{b:02d}"
     if _is_valid_month(b) and _is_valid_day(a) and b != 0:
--- a/picard/util/datestr_util.py
+++ b/picard/util/datestr_util.py
@@ -127,8 +127,8 @@ def _parse_slash_separated(value: str) -> str | None:
 
     if a is None or b is None or c is None or not (0 < c <= 9999):
         return None
-    y = _clamp_year(c)
-    if y in (None, 0):
+    y = _clamp_year(c) or None
+    if y is None:
         return ""
     if a > 12 and _is_valid_month(b) and _is_valid_day(a):
         d, m = a, b
--- a/picard/util/datestr_util.py
+++ b/picard/util/datestr_util.py
@@ -63,7 +63,7 @@ def _parse_pure_year(value: str) -> str | None:
 
 def _format_from_components(year: int | None, month: int | None, day: int | None) -> str:
     y = _clamp_year(year if year is not None else -1)
-    if y in (None,):
+    if y is None:
         return ""
     # y could be 0 here for partial unknown dates; treat specially below
     if month is None and day is None:
--- a/picard/util/datestr_util.py
+++ b/picard/util/datestr_util.py
@@ -116,14 +116,15 @@ def _parse_iso_like(value: str) -> str | None:
 
 
 def _parse_slash_separated(value: str) -> str | None:
-    if "/" not in value:
+    # We only expect 3 tokens (two slashes)
+    try:
+        tokens = value.split("/", 2)
+        a = _int_or_none(tokens[0])
+        b = _int_or_none(tokens[1])
+        c = _int_or_none(tokens[2])
+    except IndexError:
         return None
-    tokens = value.split("/")
-    if len(tokens) != 3:
-        return None
-    a = _int_or_none(tokens[0])
-    b = _int_or_none(tokens[1])
-    c = _int_or_none(tokens[2])
+
     if a is None or b is None or c is None or not (0 < c <= 9999):
         return None
     y = _clamp_year(c)
--- a/picard/util/datestr_util.py
+++ b/picard/util/datestr_util.py
@@ -1,0 +1,245 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2025 The MusicBrainz Team
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+"""Utility functions for date str/obj."""
+
+import re
+
+
+def _int_or_none(value: str) -> int | None:
+    value = value.strip()
+    if value == "":
+        return 0
+    try:
+        return int(value)
+    except ValueError:
+        return None
+
+
+def _clamp_year(value: int | None) -> int | None:
+    if value is None:
+        return None
+    if value < 0:
+        return None
+    if value > 9999:
+        return None
+    return value
+
+
+def _is_valid_month(value: int) -> bool:
+    return 0 <= value <= 12
+
+
+def _is_valid_day(value: int) -> bool:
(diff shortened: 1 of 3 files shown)


picard/util/datestr_util.py
```diff
@@ -0,0 +1,245 @@
...
+
+
+def _parse_compact_eight(value: str) -> str | None:
+    m = re.fullmatch(r"(\d{4})(\d{2})(\d{2})", value)
+    if not m:
+        return None
+    y = _clamp_year(int(m.group(1)))
+    a = int(m.group(2))
+    b = int(m.group(3))
+    if y in (None, 0):
```
zas: Move this test just after `y` declaration since it doesn't depend on `a` or `b`.
Also it would perhaps be clearer to have:

```python
y = _clamp_year(int(m.group(1))) or None
if y is None:
    return ""
```

picard/util/datestr_util.py
```diff
@@ -0,0 +1,245 @@
...
+        if m > 12 and d <= 12:
+            m, d = d, m
+        return _format_from_components(y, m, d)
+    return None
+
+
+def _parse_slash_separated(value: str) -> str | None:
+    if "/" not in value:
+        return None
+    tokens = value.split("/")
```
zas: We expect 2 slashes right? Case with more isn't really handled (1/2/3/4) and is catched by `if len(tokens) != 3` below (None is returned)

I think we can simplify things a bit here:

```python
def _parse_slash_separated(value: str) -> str | None:
    try:
        tokens = value.split("/", 2)
        a = _int_or_none(tokens[0])
        b = _int_or_none(tokens[1])
        c = _int_or_none(tokens[2])
    except IndexError:
         return None
```

picard/util/datestr_util.py
```diff
@@ -0,0 +1,245 @@
...
+    if len(tokens) != 3:
+        return None
+    a = _int_or_none(tokens[0])
+    b = _int_or_none(tokens[1])
+    c = _int_or_none(tokens[2])
+    if a is None or b is None or c is None or not (0 < c <= 9999):
+        return None
+    y = _clamp_year(c)
```
zas: I prefer:

```python
y = _clamp_year(c) or None
if y is None:
    return ""
```

picard/util/datestr_util.py
```diff
@@ -0,0 +1,245 @@
...
+        return None
+    year = _clamp_year(int(m.group(1)))
+    if year in (None, 0):
+        return ""
+    return f"{year:04d}"
+
+
+def _format_from_components(year: int | None, month: int | None, day: int | None) -> str:
+    y = _clamp_year(year if year is not None else -1)
+    if y in (None,):
```
zas:
```suggestion
    if y is None:
```

picard/util/datestr_util.py
knguyen1: [Move y check right after clamping](https://github.com/metabrainz/picard/pull/2722/commits/bc6a3e2babce016ef957938b79fb7256ba0237f3)

picard/util/datestr_util.py
knguyen1: [Only expect two slashes](https://github.com/metabrainz/picard/pull/2722/commits/eb3ba250b41be340d894cb65b875ddbf0034482c)

picard/util/datestr_util.py
knguyen1: [Explicitly check if y is None after _clamp_year](https://github.com/metabrainz/picard/pull/2722/commits/a7e7a26ebd494e4d0656e5955ca682ae5136c377)

knguyen1:
> Old code was only bothering with "-" separator, new one handles "/" and no separator, Where are those new features used?

They are common formats I've seen in the wild and added for completeness.  Does MusicBrainz sanitize them upstream?  If so, can be removed.

zas:
> They are common formats I've seen in the wild and added for completeness. Does MusicBrainz sanitize them upstream? If so, can be removed.

AFAIK, `sanitize_date()` in the current code only have to handle `-` as separator. And we parse mainly dates from tags read from files.

https://github.com/metabrainz/picard/blob/57dac026b73c61f40078ad4791025686d7a08099/picard/formats/apev2.py#L165

https://github.com/metabrainz/picard/blob/57dac026b73c61f40078ad4791025686d7a08099/picard/formats/vorbis.py#L150

https://github.com/metabrainz/picard/blob/57dac026b73c61f40078ad4791025686d7a08099/picard/formats/id3.py#L309
https://github.com/metabrainz/picard/blob/57dac026b73c61f40078ad4791025686d7a08099/picard/formats/id3.py#L346

References:
https://www.cl.cam.ac.uk/~mgk25/iso-time.html
https://en.wikipedia.org/wiki/ISO_8601

In ISO_8601 it seems hyphens are optional, though:
_"Although the standard allows both the "YYYY-MM-DD" and YYYYMMDD formats for complete calendar date representations, if the day [DD] is omitted then only the YYYY-MM format is allowed. **By disallowing dates of the form YYYYMM**, the standard avoids confusion with the [truncated representation](https://en.wikipedia.org/wiki/ISO_8601#Truncated_representations)[[25]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-iso8601amd1-25)[[4]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-doc2004-4) YYMMDD (still often used). The 2000 version also allowed writing the truncation "--04-05" to mean "April 5"[[23]](https://en.wikipedia.org/wiki/ISO_8601#cite_note-rfc6350-23) but the 2004 version does not allow omitting the year when a month is present."_

So supporting no hyphen notation is perhaps a good idea but apparently that's only for YYYYMMDD case.
Standard also allows week format (2020-W03) but I don't think we'll ever see that in date tags, so no need to bother.

It is important to note that current code only support hyphen format (YYYY, YYYY-MM, YYYY-MM-DD)
Also it is unclear to me which version of ISO8601 is used for tags.

@phw probably knows more about dates & tags, insight welcome.

Regarding this patch, I still think that's overcomplicated and it doesn't address the issue at all, as https://tickets.metabrainz.org/browse/PICARD-2729 is about "Allow disabling date sanitazation for APE and Vorbis tags" not about sanitizing more (or differently).

We should create a specific ticket for _"The cases 2005-00-12" => "2005-12" and "0000-00-12" => "0012" are a bug, though. This should be fixed."_ and only address this in this PR.

knguyen1:
>I still think that's overcomplicated and it doesn't address the issue at all, as https://tickets.metabrainz.org/browse/PICARD-2729 is about "Allow disabling date sanitazation for APE and Vorbis tags" not about sanitizing more (or differently).

Right, I planned on addressing that next PR, the new `sanitize_date` accepts a "should sanitize" `disable_sanitization` argument.

https://github.com/metabrainz/picard/blob/bc6a3e2babce016ef957938b79fb7256ba0237f3/picard/util/datestr_util.py#L170C1-L184C46

It would close out stale #2406

phw:
Sorry, only jumping in quickly due to lack of time.

> We should create a specific ticket for _"The cases 2005-00-12" => "2005-12" and "0000-00-12" => "0012" are a bug, though. This should be fixed."_ and only address this in this PR.

This is the part I don't understand, because this specific issue had been already fixed in PICARD-2712, see PR #2283. Hence I don't understand why this PR claims to fix those.

knguyen1:
My mistake. I thought it hasn't been addressed yet; was looking at the un-closed ticket. Should we close without merging?

zas:
> My mistake. I thought it hasn't been addressed yet; was looking at the un-closed ticket. Should we close without merging?

Yes, closing, since the ticket in title isn't actually addressed here anyway (and the other issue was already fixed as said by @phw)

PR #2724, opened 2025-08-29
PICARD-2729: Add select formats to disable date sanitization

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
  - Allow disabling Picard’s date sanitization per tag format (Vorbis, APEv2, ID3) via a new settings list and options UI, so users can preserve “unknown” month/day values. Closes #2406.

# Problem

Users reported that Picard always sanitizes dates (e.g., 2005-12-00 → 2005-12, 2005-00-00 → 2005, 0000-00-00 → empty), preventing them from representing unknown month/day values in formats that are essentially free-form (Vorbis comments, APEv2). This could not be overridden.

# Solution

- Added ListOption `setting['disable_date_sanitization_formats']` (list of format keys: `vorbis`, `apev2`, `id3`).
- Metadata options UI: added a multi-select (checkable combo) to choose formats for which sanitization is disabled.
- Conditionally apply sanitization:
  - Vorbis (`picard/formats/vorbis.py`): gate sanitize for `date`, `originaldate`, `releasedate` on load/save.
  - APEv2 (`picard/formats/apev2.py`): gate sanitize when mapping `Year` to `date` on load.
  - ID3 (`picard/formats/id3.py`): gate load-time sanitization and v2.3 date coercion in `format_specific_metadata`.
- Added small helper to check the setting; production code continues to use existing config access patterns.
- Tests:
  - New module `test/test_date_sanitization_setting.py` (pytest, parametrized) covering:
    - Setting-based enable/disable decision.
    - ID3 v2.3 date coercion honoring the new setting.
    - Complaint-driven Vorbis cases (e.g., 2005-12-00, 2005-00-00, 0000-00-00, 2005-00-12, 0000-00-12) for both enabled and disabled scenarios.
  - Where necessary, tests locally mock `picard.config.get_config` to ensure the setting is available without altering common test scaffolding.

# Action

Additional actions required:
* [x] Update Picard documentation (add the new “Do not sanitize dates for these tag formats” option to the Metadata options page; describe behavior per format; reference this PR)  (@rdswift added PICARD-2862 to cover the documentation changes.)

# Demo

image

### DIFF
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -37,7 +37,10 @@ from collections import OrderedDict
 from picard import PICARD_VERSION
 from picard.const import appdirs
 from picard.const.attributes import MB_ATTRIBUTES
-from picard.i18n import N_
+from picard.i18n import (
+    N_,
+    gettext as _,
+)
 
 
 # Config directory
@@ -169,3 +172,13 @@ SCRIPT_LANGUAGE_VERSION = '1.1'
 
 BROWSER_INTEGRATION_LOCALIP = '127.0.0.1'
 BROWSER_INTEGRATION_LOCALHOST = BROWSER_INTEGRATION_LOCALIP
+
+# Introduced by PICARD-2729: Major tagging formats
+MAJOR_TAGGING_FORMATS: tuple[tuple[str, str], ...] = (
+    ('vorbis', _("Vorbis Comments (FLAC, Ogg Vorbis, Opus)")),
+    ('id3', _("ID3 (MP3, AIFF)")),
+    ('apev2', _("APEv2 (Monkey's Audio, WavPack)")),
+    ('mp4', _("MP4/iTunes Tags (M4A, AAC)")),
+    ('asf', _("ASF/WMA Tags")),
+    ('riff', _("RIFF INFO (WAV)")),
+)
--- a/picard/const/__init__.py
+++ b/picard/const/__init__.py
@@ -37,10 +37,7 @@ from collections import OrderedDict
 from picard import PICARD_VERSION
 from picard.const import appdirs
 from picard.const.attributes import MB_ATTRIBUTES
-from picard.i18n import (
-    N_,
-    gettext as _,
-)
+from picard.i18n import N_
 
 
 # Config directory
@@ -172,13 +169,3 @@ SCRIPT_LANGUAGE_VERSION = '1.1'
 
 BROWSER_INTEGRATION_LOCALIP = '127.0.0.1'
 BROWSER_INTEGRATION_LOCALHOST = BROWSER_INTEGRATION_LOCALIP
-
-# Introduced by PICARD-2729: Major tagging formats
-MAJOR_TAGGING_FORMATS: tuple[tuple[str, str], ...] = (
-    ('vorbis', _("Vorbis Comments (FLAC, Ogg Vorbis, Opus)")),
-    ('id3', _("ID3 (MP3, AIFF)")),
-    ('apev2', _("APEv2 (Monkey's Audio, WavPack)")),
-    ('mp4', _("MP4/iTunes Tags (M4A, AAC)")),
-    ('asf', _("ASF/WMA Tags")),
-    ('riff', _("RIFF INFO (WAV)")),
-)
--- a/picard/file.py
+++ b/picard/file.py
@@ -134,6 +134,12 @@ class FileErrorType(Enum):
 
 class File(MetadataItem):
     NAME = None
+    # Logical tag format key and description for the family of this handler.
+    # Subclasses should override for format families (e.g. ID3, Vorbis, APEv2).
+    FORMAT_KEY = None
+    FORMAT_DESCRIPTION = None
+    # Whether date sanitization can be toggled for this format family via settings
+    DATE_SANITIZATION_TOGGLEABLE = False
 
     UNDEFINED = -1
     PENDING = 0
@@ -960,6 +966,41 @@ class File(MetadataItem):
 
     def iterfiles(self, save=False):
         yield self
+
+    # --- Per-format feature toggles ---
+    def is_date_sanitization_enabled(self) -> bool:
+        """Determine whether date sanitization should be applied.
+
+        Returns
+        -------
+        bool
+            True if sanitization should be applied for this file's format, False otherwise.
+
+        Notes
+        -----
+        This respects the per-format user configuration keyed by ``FORMAT_KEY``.
+        If the current format does not support toggling (i.e.,
+        ``DATE_SANITIZATION_TOGGLEABLE`` is False) or ``FORMAT_KEY`` is not set,
+        this method returns True.
+        """
+        # If this format does not support toggling, always treat as enabled
+        if not getattr(self.__class__, 'DATE_SANITIZATION_TOGGLEABLE', False):
+            return True
+        format_key = getattr(self.__class__, 'FORMAT_KEY', None)
+
+        # If the format key is not set, always treat as enabled
+        if not format_key:
+            return True
+
+        # Local import to avoid circular dependencies
+        from picard.config import get_config  # pylint: disable=import-outside-toplevel
+
+        try:
+            disabled = get_config().setting['disable_date_sanitization_formats']
+        except KeyError:
+            return True
+        else:
+            return format_key not in set(disabled or [])
 
 
 file_post_load_processors = PluginFunctions(label='file_post_load_processors')
--- a/picard/file.py
+++ b/picard/file.py
@@ -986,9 +986,9 @@ class File(MetadataItem):
         # If this format does not support toggling, always treat as enabled
         if not getattr(self.__class__, 'DATE_SANITIZATION_TOGGLEABLE', False):
             return True
-        format_key = getattr(self.__class__, 'FORMAT_KEY', None)
 
         # If the format key is not set, always treat as enabled
+        format_key = getattr(self.__class__, 'FORMAT_KEY', None)
         if not format_key:
             return True
 
@@ -1000,7 +1000,7 @@ class File(MetadataItem):
         except KeyError:
             return True
         else:
-            return format_key not in set(disabled or [])
+            return format_key not in disabled
 
 
 file_post_load_processors = PluginFunctions(label='file_post_load_processors')
--- a/picard/formats/util.py
+++ b/picard/formats/util.py
@@ -33,7 +33,6 @@ from picard.extension_points.formats import (
     ext_point_formats,
     ext_to_format,
 )
-from picard.i18n import gettext as _
 
 
 def supported_formats():
@@ -88,6 +87,23 @@ def open_(filename):
         return None
 
 
+def format_key_desc_generator():
+    """Yield (file_format, key, desc) for formats with key and description.
+
+    Ensures each FORMAT_KEY is yielded at most once.
+    """
+    seen = set()
+    for file_format in ext_point_formats:
+        key = getattr(file_format, 'FORMAT_KEY', None)
+        if key is None or key in seen:
+            continue
+        desc = getattr(file_format, 'FORMAT_DESCRIPTION', None)
+        if desc is None:
+            continue
+        seen.add(key)
+        yield file_format, key, desc
+
+
 def date_sanitization_format_entries() -> tuple[tuple[str, str], ...]:
     """Return registered format entries that support date-sanitization toggle.
 
@@ -104,12 +120,9 @@ def date_sanitization_format_entries() -> tuple[tuple[str, str], ...]:
     de-duplicating on ``FORMAT_KEY``.
     """
     entries = []
-    seen = set()
-    for file_format in ext_point_formats:
-        key = getattr(file_format, 'FORMAT_KEY', None)
-        desc = getattr(file_format, 'FORMAT_DESCRIPTION', None)
+    for file_format, key, desc in format_key_desc_generator():
         toggleable = getattr(file_format, 'DATE_SANITIZATION_TOGGLEABLE', False)
-        if toggleable and key and desc and key not in seen:
-            entries.append((key, _(desc)))
-            seen.add(key)
+        if toggleable:
+            # dropping `_()` here as it's done in the UI, e.g. see `tags.py`
+            entries.append((key, desc))
     return tuple(entries)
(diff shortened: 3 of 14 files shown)


zas:
I wonder if those options wouldn't better fit in Tags rather than Metadata (which is a bit crowded already). @phw @rdswift ?
Apart that, it seems to work as expected.

picard/ui/options/metadata.py
```diff
@@ -196,6 +206,61 @@ def open_script_selector(self):
...
+        self.disable_date_sanitization_formats = QtWidgets.QComboBox(self)
+        self.disable_date_sanitization_formats.setObjectName('disable_date_sanitization_formats')
+        self.disable_date_sanitization_formats.setModel(
+            QtGui.QStandardItemModel(self.disable_date_sanitization_formats)
+        )
+        self.disable_date_sanitization_formats.setSizeAdjustPolicy(
+            QtWidgets.QComboBox.SizeAdjustPolicy.AdjustToContents
+        )
+
+        for key, title in self._date_format_entries():
```
zas: Perhaps we should display entries sorted by translated titles (not critical right now, since there are only 3 items and those are mainly common acronyms).

rdswift:
> I wonder if those options wouldn't better fit in Tags rather than Metadata (which is a bit crowded already)

I think so, because it's really a format-specific setting.  Might be best appearing as a (checkbox) setting on each of the tag format setting pages.  **EDIT:** I think that might also potentially simplify the code a bit.

picard/ui/options/metadata.py
rdswift: I think I would have it in the same order, regardless of translation, but that's just my initial thought.  No real preference either way.

Then again, is this a moot point if the settings are moved to the various tags settings pages?

knguyen1: Moved to `tags.py`; added a test to simulate saving/loading vorbis: https://github.com/metabrainz/picard/pull/2724/commits/2988bfb8d110c81b47248b6e9a5b58f7c5833b88

rdswift: I actually suggest having the option to not sanitize placed on the individual format-specific pages rather than on the general Tags page.

knguyen1: Eek. :E That's gonna fragment this setting which currently lives in a single `list[str]`.  This would require refactor.  I'd like to put it up for a vote.

```python
ListOption(
    'setting',
    'disable_date_sanitization_formats',
    [],
    title=N_("Tag formats to not sanitize dates"),
)
```

zas: I'm fine with one option for all formats (rather than one option per format), I don't see any reason to change that (and the checklist is clear and convenient, and easy to add more formats to it if needed).
Having one option per format makes sense when the option is specific to a format.
I still think it should be in Tags panel (general options for tags) rather than Metadata.

rdswift: > I still think it should be in Tags panel (general options for tags) rather than Metadata.

Absolutely agree that Metadata is not the right place.

picard/const/__init__.py
```diff
@@ -169,3 +172,13 @@
...
+
+# Introduced by PICARD-2729: Major tagging formats
+MAJOR_TAGGING_FORMATS: tuple[tuple[str, str], ...] = (
+    ('vorbis', _("Vorbis Comments (FLAC, Ogg Vorbis, Opus)")),
+    ('id3', _("ID3 (MP3, AIFF)")),
+    ('apev2', _("APEv2 (Monkey's Audio, WavPack)")),
+    ('mp4', _("MP4/iTunes Tags (M4A, AAC)")),
+    ('asf', _("ASF/WMA Tags")),
+    ('riff', _("RIFF INFO (WAV)")),
+)
```
zas: Not sure about this, as formats are registered dynamically in https://github.com/metabrainz/picard/blob/596314571a1f92ddfcbed2121822c9a740545f72/picard/formats/__init__.py#L71-L96 it looks weird to me to hard-code this list.

It would be much better to extract such infos from format classes.
For example: https://github.com/metabrainz/picard/blob/596314571a1f92ddfcbed2121822c9a740545f72/picard/formats/id3.py#L129 could contain a generic description and a flag to tell if date sanitizition can be toggled for this format.
 
And I would replace `is_date_sanitization_enabled('id3')` with a method `File.is_date_sanitization_enabled()`.

It is important that such feature is able to work with any new format without changes to core code (that is just registering the format), and that's not the case here.

zas (changes requested):

picard/const/__init__.py
knguyen1: Done.

[refactor: do not hardcode toggle settings list; centralized `is_date_…](https://github.com/metabrainz/picard/pull/2724/commits/763f90f55d90cd2c6f3d7518cf093d286cf03c2f)

[Refactor: remove is_date_sanitization_enabled from util](https://github.com/metabrainz/picard/pull/2724/commits/339885c21a70e053656bfa45b301e01789c7e0cb)

picard/ui/options/metadata.py
knguyen1: Yes, the toggle list is at the bottom of `picard/ui/options/tags.py`.

picard/formats/util.py
```diff
@@ -85,3 +86,30 @@ def open_(filename):
...
+    de-duplicating on ``FORMAT_KEY``.
+    """
+    entries = []
+    seen = set()
+    for file_format in ext_point_formats:
+        key = getattr(file_format, 'FORMAT_KEY', None)
+        desc = getattr(file_format, 'FORMAT_DESCRIPTION', None)
+        toggleable = getattr(file_format, 'DATE_SANITIZATION_TOGGLEABLE', False)
+        if toggleable and key and desc and key not in seen:
+            entries.append((key, _(desc)))
```
zas: For translation to work here, we need to mark `FORMAT_DESCRIPTION` strings for translation with `N_()`:

```python
    FORMAT_DESCRIPTION = N_("ID3 (MP3, AIFF)")
```

Also I think we shouldn't translate here, but just before display instead, so I would drop the `_()` call here and move it to:

```python
    for key, title in self._date_format_entries():
            item = QtGui.QStandardItem(_(title))
```

picard/formats/util.py
```diff
@@ -85,3 +86,30 @@ def open_(filename):
...
+    Notes
+    -----
+    This inspects registered format classes and includes only those where
+    ``DATE_SANITIZATION_TOGGLEABLE`` is True. Duplicates are avoided by
+    de-duplicating on ``FORMAT_KEY``.
+    """
+    entries = []
+    seen = set()
+    for file_format in ext_point_formats:
+        key = getattr(file_format, 'FORMAT_KEY', None)
```
zas: It would be perhaps convenient to split this in two parts and to have a generator for formats having key & desc, something like:

 ```python
def format_key_desc_generator():
    seen = set()
    for file_format in ext_point_formats:
        key = getattr(file_format, 'FORMAT_KEY', None)
        if key is None or key in seen:
            continue
        desc = getattr(file_format, 'FORMAT_DESCRIPTION', None)
        if desc is None:
            continue
        seen.add(key)
        yield file_format, key, desc
```

```python
for file_format, key, desc in format_key_desc_generator():
    toggleable = getattr(file_format, 'DATE_SANITIZATION_TOGGLEABLE', False)
    if toggleable:
        entries.append((key, _(desc)))
```

picard/ui/options/tags.py
```diff
@@ -134,7 +130,8 @@ def _init_disable_date_sanitization_formats_control(self):
 
     @staticmethod
     def _date_format_entries() -> tuple[tuple[str, str], ...]:
```
zas: This method isn't strictly needed IMHO
Also since formats are unlikely to change perhaps we could just call `date_sanitization_format_entries()` once in constructor and store in a property.

picard/const/__init__.py
zas: Much better :)

picard/ui/options/tags.py
knguyen1: [Move _date_format_entries into property](https://github.com/metabrainz/picard/pull/2724/commits/66cace937188a5ca16be04798010e6b746d97523)

picard/formats/util.py
knguyen1: [Refactor: entries to use format_key_desc_generator](https://github.com/metabrainz/picard/pull/2724/commits/9d96f7db7339023be7bce51be2d23a5ed713324b)

picard/formats/util.py
knguyen1: [Mark format desc for translation; _() to menu](https://github.com/metabrainz/picard/pull/2724/commits/703904141b036575e242c78ca3ab5f66f9840f56)

picard/file.py
```diff
@@ -961,6 +967,41 @@ def update_item(self, update_selection=True):
...
+        -----
+        This respects the per-format user configuration keyed by ``FORMAT_KEY``.
+        If the current format does not support toggling (i.e.,
+        ``DATE_SANITIZATION_TOGGLEABLE`` is False) or ``FORMAT_KEY`` is not set,
+        this method returns True.
+        """
+        # If this format does not support toggling, always treat as enabled
+        if not getattr(self.__class__, 'DATE_SANITIZATION_TOGGLEABLE', False):
+            return True
+        format_key = getattr(self.__class__, 'FORMAT_KEY', None)
```
zas: I would move this line after next comment, to keep things together:

```python
        # If the format key is not set, always treat as enabled
        format_key = getattr(self.__class__, 'FORMAT_KEY', None)
        if not format_key:
            return True
```

picard/file.py
```diff
@@ -961,6 +967,41 @@ def update_item(self, update_selection=True):
...
+
+        # Local import to avoid circular dependencies
+        from picard.config import get_config  # pylint: disable=import-outside-toplevel
+
+        try:
+            disabled = get_config().setting['disable_date_sanitization_formats']
+        except KeyError:
+            return True
+        else:
+            return format_key not in set(disabled or [])
```
zas: Since `setting['disable_date_sanitization_formats']` is a `ListOption` the `or []` shouldn't be needed, also I would reformat a bit to:

```python
        # Local import to avoid circular dependencies
        from picard.config import get_config  # pylint: disable=import-outside-toplevel

        try:
            disabled = get_config().setting['disable_date_sanitization_formats']
            return format_key not in disabled
        except KeyError:
            return True
```

The conversion to `set` isn't exactly faster in this case (especially with low number of elements and low number of calls), so I would just drop it:

```bash
➜ python -m timeit -p -n 1000000 -s 'l = ["a", "b", "c"]' '"c" in set(l)'
1000000 loops, best of 5: 99.5 nsec per loop
➜ python -m timeit -p -n 1000000 -s 'l = ["a", "b", "c"]' '"c" in l'     
1000000 loops, best of 5: 34.1 nsec per loop
```

That's much faster only if we don't have to convert on each call (or with very larger number of elements), see the difference when that's a set from start:

```bash
➜ python -m timeit -p -n 1000000 -s 'l = {"a", "b", "c"}' '"c" in l' 
1000000 loops, best of 5: 15.4 nsec per loop
```

picard/formats/util.py
```diff
@@ -85,3 +85,44 @@ def open_(filename):
...
+    ``DATE_SANITIZATION_TOGGLEABLE`` is True. Duplicates are avoided by
+    de-duplicating on ``FORMAT_KEY``.
+    """
+    entries = []
+    for file_format, key, desc in format_key_desc_generator():
+        toggleable = getattr(file_format, 'DATE_SANITIZATION_TOGGLEABLE', False)
+        if toggleable:
+            # dropping `_()` here as it's done in the UI, e.g. see `tags.py`
+            entries.append((key, desc))
+    return tuple(entries)
```
zas: It might be a good idea to sort this to ensure the order is stable:

```suggestion
    return tuple(sorted(entries))
```

picard/ui/options/tags.py
```diff
@@ -72,6 +88,10 @@ def load(self):
         self.ui.preserved_tags.update(config.setting['preserved_tags'])
         self.ui.preserved_tags.set_user_sortable(False)
 
+        # Load disable date sanitization formats
+        disabled = config.setting['disable_date_sanitization_formats'] or []
```
zas:
```suggestion
        disabled = config.setting['disable_date_sanitization_formats']
```

`or []` shouldn't be needed (`ListOption`)

zas (changes requested):
The patch looks good to me, just few stylistic comments, but overall almost ready to go for me.

picard/file.py
knguyen1: I removed the `set()`: https://github.com/metabrainz/picard/pull/2724/commits/00b702450c128ec3bca74113e4d1409105164c01

But I suggest keeping the `try... except... else` pattern.  Here's why: https://docs.astral.sh/ruff/rules/try-consider-else/

picard/file.py
knguyen1: Done: https://github.com/metabrainz/picard/pull/2724/commits/00b702450c128ec3bca74113e4d1409105164c01

zas (approved):
LGTM

phw (commented):
Code looks good to me, the feature also is useful. But not sure about the UI, I think we should improve this more. While it is great that this (rather advanced) option does not take much space on the already crowded tags page, the UI shown in collapsed state is misleading:

image

This looks as if I had disabled sanitize dates for APEv2, while in reality it is ID3 and Vorbis I had selected:

image

I'd be fine with merging this PR for now, but we should refactor the UI of the Tags options.

Some ideas:

- Open a new dialog to select the types
- Maybe keep the UI, but in unopened state show all selected formats comma separated
- We could also completely rework this. We already have some format specific sub pages. We could also add a sub page for each major tag format that has extra options and move the options there. The problem is a bit that sometimes we have options for individual file formats, and sometimes generally for a tagging format. Not sure this will be resolved nicely.

zas:
> I'd be fine with merging this PR for now, but we should refactor the UI of the Tags options.

Yes, I agree the UI isn't perfect but it works, and I doubt many users will have to change that often. The core issue is addressed, so I prefer to merge now.
Please copy your comment above to a new ticket about UI improvement.

phw:
Done, see PICARD-3115

PR #2726, opened 2025-09-01
PICARD-2300: Check for setColorScheme for Qt <6.8 compatibility

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

A small addition to #2693 . this introduced the usage of `QStyleHints.setColorScheme`, which was only added in Qt 6.8 (see https://doc.qt.io/qt-6/qstylehints.html#setColorScheme).

We have Qt 6.6.1 currently defined as the minimum. I think something like the styles is no reason to not maintain compatibility with older versions.

# Solution

Check for existance of `QStyleHints.setColorScheme`. If this is not available then the system is handled as having no style hints.

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -97,7 +97,11 @@ class UiTheme(Enum):
 
 def get_style_hints() -> QtGui.QStyleHints | None:
     """Get style hints from QGuiApplication, returning None if unavailable."""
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #2729, opened 2025-09-05
Use Qt standard buttons.  Only use "Make It So!" on Options page.

rdswift:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

There was inconsistent use of Qt standard buttons and Picard version of standard buttons, and "Make It So!" was being used in multiple places when the original intent was to use it on the Options page only.

# Solution

Remove the Picard standard buttons class and use `Qt.QButtonBox.StandardButton` definitions throughout.  Replace additional use of "Make It So!" buttons with standard "Ok" button.

# Action

Additional actions required:


picard/ui/scripteditor/__init__.py
```diff
@@ -156,14 +156,13 @@ def __init__(self, parent=None, examples=None):
         self.installEventFilter(self)
 
         # Dialog buttons
-        self.reset_button = QtWidgets.QPushButton(_("Reset"))
+        self.reset_button = self.ui.buttonbox.addButton(QtWidgets.QDialogButtonBox.StandardButton.Reset)
         self.reset_button.setToolTip(self.reset_action.toolTip())
         self.reset_button.clicked.connect(self.reload_from_config)
+        # Adding the button with an ActionRole places it with the Cancel and Ok buttons
         self.ui.buttonbox.addButton(self.reset_button, QtWidgets.QDialogButtonBox.ButtonRole.ActionRole)
```
rdswift: I did something odd here which is likely poor practice, in order to get the `Reset` button to group with the `Cancel` and `Ok` buttons, rather than with the `Help` button.  I just thought:

image

looks better than:

image

Let me know if you want me to remove this and allow Qt to place the button as a `ResetRole` next to the `Help` button.

zas: > Let me know if you want me to remove this and allow Qt to place the button as a `ResetRole` next to the `Help` button.

I prefer having only OK/Cancel on one side (so Help & Reset together).

rdswift: > I prefer having only OK/Cancel on one side (so Help & Reset together).

Done.  I can live with it either way.

PR #2730, opened 2025-09-07
Avoid a possible crash when columns index is invalid

zas:
- index < 0 was handled
- index >= len(self._columns) wasn't (it happened when switching between different branches, it shouldn't during normal use)
- so be more defensive and catch IndexError (and log it as warning)

Original error was:

```
  File "/home/zas/src/picard/picard/ui/widgets/configurablecolumnsheader.py", line 130, in on_sort_indicator_changed
    if index < 0 or not self._columns[index].sortable:
                        ~~~~~~~~~~~~~^^^^^^^
  File "/home/zas/src/picard/picard/ui/columns.py", line 167, in __getitem__
    return self._list.__getitem__(index)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
```

# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

# Solution

# Action

Additional actions required:

### DIFF
--- a/picard/ui/widgets/configurablecolumnsheader.py
+++ b/picard/ui/widgets/configurablecolumnsheader.py
@@ -51,6 +51,7 @@ from PyQt6 import (
     QtWidgets,
 )
 
+from picard import log
 from picard.i18n import gettext as _
 
 from picard.ui.columns import ImageColumn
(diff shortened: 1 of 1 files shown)


rdswift (approved):
Good catch!

phw (approved):
LGTM

PR #2733, opened 2025-09-15
PICARD-150: Set default filter columns for tree view filters

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-150

This is extending #2668 to add default columns for the tree view filter. I found that starting with a blank config this filtering is confusing to use. Once you enable the filtering in the View menu it shows the search input. But entering text there doesn't do anything, unless you also select some columns to filter by.

I think we need some default, so filtering works by default for the common cases.

# Solution
By default filter by album, title, albumartist and artist


zas (approved):
LGTM

PR #2714, opened 2025-08-21
PICARD-2103: Add UI to manage custom columns

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:
  Adds a UI to manage custom columns from the header menu: create, edit, duplicate, delete, place, and apply to File/Album views.

# Problem

* JIRA ticket (_optional_): PICARD-2103

Users had no built-in UI to add/manage custom columns; changes required backend code or were not discoverable from the views.

# Solution

- Header context menu: new “Manage Custom Columns…” entry (unlock columns if needed).
- Manager dialog:
  - List of custom columns with Type, Expression, Align, Width.
  - Actions: Add…, Edit…, Duplicate, Delete, “Make It So!” (apply).
- Expression dialog (per column):
  - Field Name, Type (Field or Script), Expression, Width, Align.
  - Add to views: File view / Album view.
  - Insert after key: position by existing column key (e.g. title, artist).
- UX details:
  - Immediate registration into live views on apply; closing applies pending changes.
  - Idempotent updates when editing or duplicating; safe deletion with confirmation.
  - Columns can also be toggled from the header menu like built-ins.

# Demo

image

# Action

Additional actions required:
* [x] Update Picard [documentation](https://github.com/metabrainz/picard-docs) (please include a reference to this PR)
* [ ] Other (please specify below)

Note: API PR #2709 must be merged first; this PR only covers the UI/UX management layer.

PR #2735, opened 2025-09-18
PICARD-2103: Custom columns sorting tweaks

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-2103

This applies some changes to the custom column sorting:

- Removes the "descending" variations. These are not needed and even confusing. Sort order can be changed by the user in the UI. The descending ones then behave exactly the opposite, with no advantage to the user.
- Makes sorting locale aware by default (just like the standard columns). Only special sorting cases, like by length, do not need to be locale aware
- Removes the unused `ReverseAdapter`. This does not work with locale aware sorting keys anyway.

# Solution

# Action

Additional actions required:


picard/ui/itemviews/custom_columns/shared.py
```diff
@@ -296,13 +296,10 @@ def generate_new_key() -> str:
...
+    N_("Default"): "LocaleAwareSortAdapter",
     N_("Case Insensitive"): "CasefoldSortAdapter",
-    N_("Case Insensitive - Descending"): "DescendingCasefoldSortAdapter",
     N_("Numeric"): "NumericSortAdapter",
-    N_("Numeric - Descending"): "DescendingNumericSortAdapter",
     N_("Natural"): "NaturalSortAdapter",
-    N_("Natural - Descending"): "DescendingNaturalSortAdapter",
     N_("By Value Length"): "LengthSortAdapter",
     N_("Article Insensitive"): "ArticleInsensitiveAdapter",
     N_("Empty Values Last"): "NullsLastAdapter",
```
rdswift: At the risk of introducing some scope creep, I wonder if it would be worthwhile to have a tooltip or something for clarification of each of the sort selection items because the selection names themselves might not be clear to the user.  For example, "Article Insensitive" (which actually may not be necessary because I think it could be handled by using `$swapprefix()` in the column specification) or the difference between "Default" and "Natural".

**EDIT:**  If I understand "Article Insensitive" correctly, then using the `$swapprefix()` scripting funtion is actually a much better solution because it allows the user to specify the prefixes used/ignored.

picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -66,27 +66,20 @@ def __repr__(self) -> str:  # pragma: no cover - debug helper
         return f"{self.__class__.__name__}(base={self._base!r})"
 
 
-class CasefoldSortAdapter(_AdapterBase):
+class LocaleAwareSortAdapter(_AdapterBase):
```
knguyen1, line 69: This duplicates the already present `NaturalSortAdapter`.  But `NaturalSortAdapter` does feed `is_numeric`, so...

test/test_custom_columns_sorting_adapters.py
```diff
@@ -137,49 +123,18 @@ def adapter(base):
...
 
 
+@pytest.mark.skipif(IS_WIN, reason="QCollator not used on Windows")
 def test_article_insensitive_adapter() -> None:
+    setup_gettext(None, 'en')
     values = ["The Beatles", "Beatles", "An Artist", "Artist"]
-    expected = ["An Artist", "Artist", "Beatles", "The Beatles"]
+    expected = ["An Artist", "Artist", "The Beatles", "Beatles"]
     result = _sorted_values(ArticleInsensitiveAdapter, values)
     assert result == expected
```
knguyen1, line 139: That's strange... Setting the locale (I assume `setup_gettext(None, 'en')` does this) changes the order?

rdswift:
Also a "heads up" that the documentation changes in https://github.com/metabrainz/picard-docs/pull/234 will need to be updated as well.

picard/ui/itemviews/custom_columns/shared.py
knguyen1: Since you're making changes to this dict, you'll also need to modify `_apply_sorting_adapter` in storage.py.  Something like:

```python
def _apply_sorting_adapter(base_provider: ColumnValueProvider, sorting_adapter_name: str) -> ColumnValueProvider:
    """Apply the specified sorting adapter to a base provider.

    Parameters
    ----------
    base_provider : ColumnValueProvider
        The base provider to wrap with sorting adapter.
    sorting_adapter_name : str
        Name of the sorting adapter class to apply.

    Returns
    -------
    ColumnValueProvider
        The wrapped provider with sorting capability, or the original provider if no adapter.
    """
    if not sorting_adapter_name:
        return base_provider

    # Import the shared mapping to get all available class names
    from picard.ui.itemviews.custom_columns.shared import SORTING_ADAPTER_NAMES
    from picard.ui.itemviews.custom_columns import sorting_adapters

    # Build mapping dynamically from available class names in SORTING_ADAPTER_NAMES
    # This ensures we only try to map classes that are actually defined in the UI
    available_class_names = {name for name in SORTING_ADAPTER_NAMES.values() if name}

    # Create a mapping of class name strings to actual classes
    # Only include classes that are referenced in SORTING_ADAPTER_NAMES
    adapter_classes = {}
    for class_name in available_class_names:
        # Dynamically import and get the class from the sorting_adapters module
        try:
            adapter_class = getattr(sorting_adapters, class_name, None)
            if adapter_class is not None:
                adapter_classes[class_name] = adapter_class
        except (ImportError, AttributeError):
            # Skip classes that don't exist or can't be imported
            continue

    adapter_class = adapter_classes.get(sorting_adapter_name)
    if adapter_class is None:
        log.warning("Unknown sorting adapter: %s", sorting_adapter_name)
        return base_provider

    return adapter_class(base_provider)
```

picard/ui/itemviews/custom_columns/storage.py
```diff
@@ -263,30 +263,24 @@ def _apply_sorting_adapter(base_provider: ColumnValueProvider, sorting_adapter_n
...
-        'DescendingNumericSortAdapter': DescendingNumericSortAdapter,
         'NaturalSortAdapter': NaturalSortAdapter,
-        'DescendingNaturalSortAdapter': DescendingNaturalSortAdapter,
         'LengthSortAdapter': LengthSortAdapter,
+        'LocaleAwareSortAdapter': LocaleAwareSortAdapter,
         'ArticleInsensitiveAdapter': ArticleInsensitiveAdapter,
         'NullsLastAdapter': NullsLastAdapter,
         'NullsFirstAdapter': NullsFirstAdapter,
-        'ReverseAdapter': ReverseAdapter,
     }
```
knguyen1: Make this more generic and not dependent on the dict.  see: https://github.com/metabrainz/picard/pull/2735/files#r2359957886

picard/ui/itemviews/custom_columns/sorting_adapters.py
phw: That's the actual defining difference. The NaturalSortAdapter is specifically about this sorting of numbers inside strings (sorting "10" after "2"). It can be computational more expensive. But in any case we always sort locale aware in Picard. It is otherwise very annoying if "ä" does not show up where you'd expect it (and that's different for e.g. Norwegian or German).

test/test_custom_columns_sorting_adapters.py
phw: No. But we have two implementations of locale aware sorting. Linux and macOS use `QCollator`, while on Windows we use `locale.strxfrm`. The reason for this are different issues we experienced on different platforms. See also the implementation of `i18n.sort_key` in https://github.com/metabrainz/picard/blob/master/picard/i18n.py#L238 .

`QCollator` would also be what would be used if we would use https://doc.qt.io/qt-6/qsortfilterproxymodel.html#isSortLocaleAware-prop . Which we don't, as we want to have our own control over the sorting of different columns.

This case of sorting `("Beatles", "The")` and `("Beatles", "")` differently comes from those separate implementations. No matter the OS, you get the same sorting with those implementations, but they are not consistent between each other. But honestly that's one of those cases I just don't mind. We could try to make it consistent, but likely there are more, locale specific differences anyway.

For this test function I just switched it around so it now skips Windows but runs on macOS and Linux. We have specific `sort_key` tests for all OS anyway, so I guess that's fine.

Regarding setting the locale in the tests: This is always a bit tricky, as it introduces side effects. It's process wide, so it affects other tests. That is also why some of those column sorting tests work fine even without explicitly setting the locale if run isolated, but fail when you run the full test suite.

In other tests that use the old style test classes we often set the locale in the `setUp` function, and probably should unset it in the `tearDown` (but often don't do). That's not easily possible with pytest, though. You can kind of use a `fixture` for it, but you also have to modify each test function to use the fixture, which is about as much change as just setting the locale in the function.

picard/ui/itemviews/custom_columns/shared.py
phw: Will a tooltip for items in a select box work well? Maybe we can rather see that we make the naming better if something is not clear.

> EDIT: If I understand "Article Insensitive" correctly, then using the $swapprefix() scripting funtion is actually a much better solution because it allows the user to specify the prefixes used/ignored.

Yes, it's the same idea. Using `$swappprefix` will do the same, but is more flexible. The sorting setting is easier to use.

But generally both are doing a similar thing. That's why I also said in the other PR that one of the next todos is to unify both the sorting adpater and `$swapprefix` to use the same implementation of extracting the prefixes.

phw: Thanks. Much better to have this generic. I commited your change in b31023c42b79e0b61ce20cf1ff9707ab1b3ae04f

phw:
> Also a "heads up" that the documentation changes in https://github.com/metabrainz/picard-docs/pull/234 will need to be updated as well.

It's good we have this PR already, but I think we should only merge it once we have fully finished the implementation of the new columns (including unifying the code for the predefined columns). There will probably be more changes to this coming.

phw:
> This is the way it should have been done right from the start.

Yes, but sometimes it is easier to get a complex PR like #2714 merged first and make such small adjustments afterwards, where they are much easier to review. We already had 220 comments on the initial PR.

picard/ui/itemviews/custom_columns/shared.py
phw: I added tooltips in f6d5fb150c523b5c70843a26235b58bdbc5db9fc

phw:
I actual wonder if we need both `NumericSortAdapter` and `NaturalSortAdapter`. They are kind of interchangable. `NumericSortAdapter` will first try to convert the whole value to a number, then falls back to natural sorting. Natural sorting will sort full numbers inside as a string by their value. I'm unsure whether the `NumericSortAdapter`'s initial step is an actual optimization or unnecessary extra work. But in the end you'll essentially get the same results with both adapaters.

@knguyen1 your thoughts?

zas (approved):
LGTM, good move

picard/ui/itemviews/custom_columns/storage.py
```diff
@@ -259,35 +259,26 @@ def _apply_sorting_adapter(base_provider: ColumnValueProvider, sorting_adapter_n
...
-        'NullsFirstAdapter': NullsFirstAdapter,
-        'ReverseAdapter': ReverseAdapter,
-    }
+    # Import the shared mapping to get all available class names
+    from picard.ui.itemviews.custom_columns import sorting_adapters
+    from picard.ui.itemviews.custom_columns.shared import SORTING_ADAPTER_NAMES
+
+    # Build mapping dynamically from available class names in SORTING_ADAPTER_NAMES
+    # This ensures we only try to map classes that are actually defined in the UI
+    available_class_names = {name for name in SORTING_ADAPTER_NAMES.keys() if name}
```
zas, line 268: Make this a generator instead?

```python
def available_class_names():
    for name in SORTING_ADAPTER_NAMES.keys():
        if name:
           yield name

....
for class_name in available_class_names():
```

rdswift:
> > Also a "heads up" that the documentation changes in [metabrainz/picard-docs#234](https://github.com/metabrainz/picard-docs/pull/234) will need to be updated as well.
> 
> It's good we have this PR already, but I think we should only merge it once we have fully finished the implementation of the new columns (including unifying the code for the predefined columns). There will probably be more changes to this coming.

I agree completely.  I'm in no rush to merge the documentation PRs for these large UI changes for exactly that reason.

rdswift:
> I actual wonder if we need both `NumericSortAdapter` and `NaturalSortAdapter`. They are kind of interchangable. `NumericSortAdapter` will first try to convert the whole value to a number, then falls back to natural sorting. Natural sorting will sort full numbers inside as a string by their value. I'm unsure whether the `NumericSortAdapter`'s initial step is an actual optimization or unnecessary extra work. But in the end you'll essentially get the same results with both adapaters.

I know you didn't specifically ask for my thoughts on this, but here they are anyway.  :stuck_out_tongue_winking_eye: 

If we can remove one of the selection options and still get the same results from one of the other options, I think we should be doing exactly that.  The simpler we can make this for the user the better.

picard/ui/itemviews/custom_columns/shared.py
```diff
@@ -294,19 +294,46 @@ def generate_new_key() -> str:
...
+        tooltip=N_("Sort ignoring leading articles (e.g. a, an, the)."),
+    ),
+    "NullsLastAdapter": SortingAdapterDisplayInfo(
+        display_name=N_("Empty values first"),
+        tooltip=N_("Empty/whitespace values sort first."),
+    ),
+    "NullsFirstAdapter": SortingAdapterDisplayInfo(
+        display_name=N_("Empty values last"),
+        tooltip=N_("Empty/whitespace values sort last."),
+    ),
```
rdswift, line 336: Ideally, wouldn't it be better if these two (mutually exclusive) options could be applied to any of the other sorting options?  Perhaps not in the selection list, but as a separate radio button as something like:

**Show empty values:**  ○ first,   ◉ last

knguyen1:
>  I actual wonder if we need both `NumericSortAdapter` and `NaturalSortAdapter`. They are kind of interchangable. `NumericSortAdapter` will first try to convert the whole value to a number, then falls back to natural sorting. Natural sorting will sort full numbers inside as a string by their value. I'm unsure whether the `NumericSortAdapter`'s initial step is an actual optimization or unnecessary extra work. But in the end you'll essentially get the same results with both adapaters.

Numeric and Natural are not interchangeable; they encode different semantics.

Below is LLM claim.  See https://github.com/metabrainz/picard/pull/2735#issuecomment-3316012595 for proof.

### What differs
- **NumericSortAdapter**: 
  - Tries to parse the entire value as a number; if successful returns a key like (0, number) so numbers sort first and numerically.
  - If parsing fails, falls back to natural sort key with a (1, ...) flag so non-numeric entries come after numeric ones, still sorted naturally.
  - Supports a **custom parser** (e.g., durations “3:30” → seconds), negative numbers, and decimals consistently.
  - Returns normalized tuple keys to avoid mixed-type comparison issues.

- **NaturalSortAdapter**:
  - Always uses locale-aware natural collation on the string, treating digit runs inside text numerically (e.g., “Track 2” < “Track 10”).
  - Does not force “numbers first” semantics and doesn’t accept a custom numeric parser.
  - Key shape is whatever the collator returns; semantics differ across platforms/locales.

### Why keep both
- **Semantics**: Numeric enforces “numbers-first, numeric order” for whole-value numerics; Natural provides “human” alphanumeric order within text. These yield different outcomes for negatives, decimals, equal numeric values with different formatting (“000”, “01”, “1”), and custom formats (mm:ss).
- **Features**: Only Numeric supports a custom parser and explicit numeric-first grouping.
- **Stability**: Numeric’s tuple keys avoid cross-type comparison issues; tests rely on this behavior.

### Performance note
- The numeric-first parse is not just an optimization; it provides distinct semantics. It may also avoid building a collator key for purely numeric values, which can be marginally faster, but correctness is the main reason.

### Recommendation
- Use **NumericSortAdapter** for primarily numeric columns or when you need numbers-first or custom parsing.
- Use **NaturalSortAdapter** for general text with embedded numbers where “Track 2, Track 10” style ordering is desired.

- Kept both adapters; clarified they serve different semantics:
  - `NumericSortAdapter`: numbers-first, supports custom parser, normalized tuple keys.
  - `NaturalSortAdapter`: locale-aware natural order within strings, no numbers-first guarantee.

knguyen1:
You can use this `python` script to test why `NumericSortAdapter` should still be kept.

- Numbers-first semantics across mixed values: ensures all numeric entries sort before non-numeric text.
- Whole-value numeric parsing: supports negatives, decimals, scientific notation; natural sort does not parse signs/exponents as numbers.
- Domain-specific numeric formats via custom parser: durations (mm:ss), localized numbers (“1,200”), file sizes, bitrates.
- Stable, normalized keys: returns tuple keys that avoid mixed-type comparisons.

```python
#!/usr/bin/env python3
from __future__ import annotations

from typing import Callable

from picard.item import Item
from picard.ui.itemviews.custom_columns.sorting_adapters import (
    NaturalSortAdapter,
    NumericSortAdapter,
)


class _ValueItem(Item):
    def __init__(self, value: str) -> None:
        super().__init__()
        self.value = value


class _Provider:
    def __init__(self, value: str) -> None:
        self._value = value

    def evaluate(self, _obj: Item) -> str:
        return self._value


def _sorted_values_with_adapter(
    adapter_factory: Callable[[object], object], values: list[str]
) -> list[str]:
    adapters = [adapter_factory(_Provider(v)) for v in values]
    keys = [a.sort_key(_ValueItem(v)) for a, v in zip(adapters, values, strict=True)]
    # Pair the original value with its key and sort by the produced key
    pairs = sorted(zip(values, keys, strict=True), key=lambda p: p[1])
    return [v for v, _ in pairs]


def _print_case(title: str, values: list[str]) -> None:
    print(f"\n=== {title} ===")
    nat = _sorted_values_with_adapter(lambda base: NaturalSortAdapter(base), values)
    num = _sorted_values_with_adapter(lambda base: NumericSortAdapter(base), values)
    print("Input:     ", values)
    print("Natural:   ", nat)
    print("Numeric:   ", num)


def _print_case_with_parser(
    title: str,
    values: list[str],
    parser: Callable[[str], float],
    parser_label: str,
) -> None:
    print(f"\n=== {title} ({parser_label}) ===")
    nat = _sorted_values_with_adapter(lambda base: NaturalSortAdapter(base), values)
    num = _sorted_values_with_adapter(lambda base: NumericSortAdapter(base, parser=parser), values)
    print("Input:     ", values)
    print("Natural:   ", nat)
    print("Numeric*:  ", num)


def _mmss_parser(s: str) -> float:
    # e.g. "3:30" -> 210.0
    total = 0
    for part in s.split(":"):
        total = total * 60 + int(part)
    return float(total)


def _comma_number_parser(s: str) -> float:
    # Handle localized/thousands separators like "1,200"
    return float(s.replace(",", "").strip())


def main() -> None:
    # 1) Mixed numerics, negatives, decimals, scientific notation, and text
    #    NaturalSortAdapter treats digits inside text as numbers, but not signs/exponents.
    #    NumericSortAdapter parses the entire value (float), so "-1" and "1e3" are truly numeric.
    _print_case(
        "Mixed: integers, decimals, negatives, scientific notation, and text",
        ["10", "2", "3.5", "x", "-1", "7a", "1e3", "3e2"],
    )

    # 2) Duration-like values need domain parsing. NumericSortAdapter with a custom parser
    #    sorts by actual time; NaturalSortAdapter only sees text + digit runs.
    _print_case_with_parser(
        "Durations requiring domain parsing",
        ["3:30", "2:05", "1:00", "x", "10:00", "0:59"],
        parser=_mmss_parser,
        parser_label="custom mm:ss parser",
    )

    # 3) Localized numbers with thousands separators. NumericSortAdapter can parse them
    #    (with a custom parser); NaturalSortAdapter compares as text with digits.
    _print_case_with_parser(
        "Localized numbers with commas",
        ["1,200", "300", "20", "1,050", "x"],
        parser=_comma_number_parser,
        parser_label="custom comma-aware parser",
    )

    # 4) Whitespace and formatting differences. float() handles surrounding whitespace;
    #    natural sorting treats whitespace as text.
    _print_case(
        "Whitespace and formatting",
        ["  12  ", "2", " 10", "x", "000", "01", "1", "1.0"],
    )


if __name__ == "__main__":
    main()
```

### The Output

```bash
$ uv run python test_sort.py

=== Mixed: integers, decimals, negatives, scientific notation, and text ===
Input:      ['10', '2', '3.5', 'x', '-1', '7a', '1e3', '3e2']
Natural:    ['-1', '1e3', '2', '3.5', '3e2', '7a', '10', 'x']
Numeric:    ['-1', '2', '3.5', '10', '3e2', '1e3', '7a', 'x']

=== Durations requiring domain parsing (custom mm:ss parser) ===
Input:      ['3:30', '2:05', '1:00', 'x', '10:00', '0:59']
Natural:    ['0:59', '1:00', '2:05', '3:30', '10:00', 'x']
Numeric*:   ['0:59', '1:00', '2:05', '3:30', '10:00', 'x']

=== Localized numbers with commas (custom comma-aware parser) ===
Input:      ['1,200', '300', '20', '1,050', 'x']
Natural:    ['1,050', '1,200', '20', '300', 'x']
Numeric*:   ['20', '300', '1,050', '1,200', 'x']

=== Whitespace and formatting ===
Input:      ['  12  ', '2', ' 10', 'x', '000', '01', '1', '1.0']
Natural:    ['  12  ', ' 10', '000', '01', '1', '1.0', '2', 'x']
Numeric:    ['000', '01', '1', '1.0', '2', ' 10', '  12  ', 'x']
```

PR #2739, opened 2025-09-22
refactor: standard columns to use new API

knguyen1:
# Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Refactor standard columns (common cols, size, bitrate, length, match quality) to use the new provider architecture, removing delegate adapter wrappers for cleaner implementation.

# Problem

* JIRA ticket (_optional_): PICARD-3119

Standard columns were using outdated delegate adapter patterns that added unnecessary complexity and didn't align with the new provider architecture.

# Solution

- Refactored standard columns to use the new provider architecture
- Removed delegate adapter wrapper for cleaner implementation  
- Fixed size, bitrate, length columns to be field columns
- Updated match quality column to use provider architecture
- General cleanup and code improvements
- Updated tests to reflect new column architecture

# Action

Additional actions required:


zas:
This PR makes "Unclusted Files" title to disappear (compared to current master branch):

This branch:
Capture d’écran du 2025-09-22 17-09-57

Master:
Capture d’écran du 2025-09-22 17-09-10

knguyen1:
> This PR makes "Unclusted Files" title to disappear (compared to current master branch):

That's actually from this https://github.com/metabrainz/picard/pull/2714#issuecomment-3268555686 and this change https://github.com/metabrainz/picard/pull/2714/commits/4d05b75f75f0e473c571085c38d1ba4b80077f3a

And this is the fix: https://github.com/metabrainz/picard/pull/2739/commits/e95b18238446c45c75214c78560b45e25c7042db

zas (approved):
LGTM

rdswift:
Are we actually going to manage these settings through the option profile system, or should I quit asking about it?

zas:
> Are we actually going to manage these settings through the option profile system, or should I quit asking about it?

What do you mean?
This PR is just about extending the use of CustomColumns, how does it relate to option profile system?

I mean I'm not against "manage these settings through the option profile system" but I fail to see how it is related to this PR.

rdswift:
I've been asking for the custom columns to be managed through the option profile system right from the initial PR that introduced the functionality, but it keeps getting ignored.  It's clear to me that @knguyen1 has no intention of **ever** doing this, so I guess I'll give up and quit asking.

zas:
> I've been asking for the custom columns to be managed through the option profile system right from the initial PR that introduced the functionality, but it keeps getting ignored. It's clear to me that @knguyen1 has no intention of **ever** doing this, so I guess I'll give up and quit asking.

@rdswift Can you clarify your expectations? Did you create ticket for that? Everything is still work in progress.

phw:
@rdswift  I don't think we have decided on making the custom columns part of the option profiles yet. If we decide to do this, we can do this in a separate PR. But I have stated my opinion elsewhere. To me the column configuration is about UI state, and as such does not really fit into the option profiles. It's hence also not fully clear which parts of the column configuration you would see in the profiles.

Only the configuration which custom columns are configured? But then this doesn't make sense if I switch profiles and have changing columns that are also part of the stored column header state (which includes the columns shown and the Qt state of their position, size and sort order). If we make the whole column state part of option profiles, will we also need to make window and pane size states part of the option profiles?

If there is a strong case for having the column state part of the option profiles and people really want this, then I'd be ok with having it implemented. It should just be clear that this is more involved then just making the `custom_columns` option part of the profiles.

phw:
Just to elaborate on the problem I see in more detail:

The state of the header is currently stored in two options per pane, that are both considered state and are inside the "persist" config section:

- `(album_view|file_view)_header_state`: This is an opaque value directly from `QHeaderView`. It encodes the state of which columns are visible, the order they are shown in, their size and their sort order. 
- `(album_view|file_view)_header_locked`: This is just a boolean, indicating whether the header is locked from changes

Previously the list of available columns was hard coded. Now we have another configuration, where the user can extend this list with custom columns. Those are stored in the "setting" config section as `custom_columns`. Now we consider this config, and we could consider this part of the option profiles then. This raises the following issues:

1. If I have different custom columns in different profiles and switch between them I'll break the header state. We would need to reset the header state. Which would be annoying to loose header state on each profile switch.
2. Or we make the header state also part of the profile. But it would be no good to allow the user to configure just the custom columns as part of the profile without the state. So we need a mechanism for users to select the columns being part of the profile, but then it actually affects multiple config keys, some of them in "persist".
3. Actually, if we allow storing the state in profile (which also holds the visible columns) why do we then need to have the `custom_columns` settings in the profile at all? Wouldn't it be nice to have columns globally, but only the state change between profiles?
4. If we store the header state, what about the rest of the view state (window size, pane sizes). After all my column sizes might depend on my view sizes as well. 
5. If we do store all the view state, we need to update it all on profile switch.
6. What about all those other windows, that use similar logic (search dialogs etc.). Is even the pane size of the options dialog now subject to option profiles?

For me, option profiles are clearly about settings and should not also handle window state.

rdswift:
As I mentioned elsewhere, which columns are displayed could be part of a user's workflow that is faciltated through a selected options profile.  In any event, it seems that I'm alone in my request, so I'll drop it.

knguyen1:
Rebased and resolved conflicts.  MacOS tests passing locally.

picard/ui/itemviews/custom_columns/common_columns.py
```diff
@@ -0,0 +1,183 @@
...
+    # Title (status icon column)
+    title_col = make_field_column(
+        N_("Title"),
+        'title',
+        sort_type=ColumnSortType.NAT,
+        width=250,
+        always_visible=True,
+        status_icon=True,
+    )
+    title_col.is_default = True
```
zas: What about using a named parameter to `make_field_column()` so we can drop local variables totally and write:

```python
    return (
        make_field_column(
            N_("Title"),
            'title',
            sort_type=ColumnSortType.NAT,
            width=250,
            always_visible=True,
            status_icon=True,
            is_default=True,
        ),
        make_field_column(
....
        )    
    )
```

knguyen1: Can be done.  It'd look like this.  A gigantic return statement is less readable imo, vs. local variables where we can mix up in-line comments, logical sections, place breakpoints, etc..  Below the code is LLM analyses/pros/cons.  Let me know if you'd like to proceed.

---
# LLM:

### Readability trade-offs

- • Pros of the “gigantic return”:
  - **Fewer names in scope**: avoids 20+ locals.
  - **Tighter grouping**: all items are visible together; order is explicit.
  - **Immutability by construction**: no post-creation attribute mutations.

- • Cons (and why I’d avoid it here):
  - **Harder to scan and annotate**: per-item comments, blank lines, and visual grouping are awkward inside a massive tuple literal.
  - **Debuggability suffers**: you can’t set breakpoints or inspect a problematic column before aggregation; locals make step-by-step inspection trivial.
  - **Diff quality declines**: reordering or tweaking a couple of items churns a large block, increasing merge conflicts and reducing review clarity.
  - **Reuse and naming**: locals give semantic names (`title_col`, `length_col`) that help readers and future refactors; inline calls hide intent.
  - **Line noise and cognitive load**: nested argument lists across many items are tougher to skim than short local assignments + a clean `return (...)`.

### Recommendation
- Keep the local variables, but use the new `is_default` parameter to remove the old post-assignment lines. This achieves brevity without sacrificing readability, debuggability, or diff quality.

Example compromise (shortened for illustration):
```python
title_col = make_field_column(N_("Title"), 'title', sort_type=ColumnSortType.NAT, width=250, always_visible=True, status_icon=True, is_default=True)
length_col = make_numeric_field_column(N_("Length"), '~length', parse_time_format, width=50, align=ColumnAlign.RIGHT, is_default=True)
artist_col = make_field_column(N_("Artist"), 'artist', width=200, is_default=True)
# ...
return (title_col, length_col, artist_col, ...)
```

### Suggested reviewer comment (concise)
- Keeping locals improves readability, debugging, and diff stability. Using `is_default=True` in the factories already removes prior boilerplate. The giant inline tuple makes per-item comments, breakpoints, and reordering harder and increases merge churn without a meaningful runtime benefit. I propose we keep locals + `is_default` for a clean, maintainable middle ground.

I reviewed the staged diff in both files and focused on readability. No code edits needed.

---

```python
def create_common_columns() -> tuple[CustomColumn, ...]:
    """Create the built-in common columns using factories.

    Returns
    -------
    tuple
        Tuple of configured column objects for both views.
    """
    return (
        make_field_column(
            N_("Title"),
            'title',
            sort_type=ColumnSortType.NAT,
            width=250,
            always_visible=True,
            status_icon=True,
            is_default=True,
        ),
        make_numeric_field_column(
            N_("Length"),
            '~length',
            parse_time_format,
            width=50,
            align=ColumnAlign.RIGHT,
            is_default=True,
        ),
        make_field_column(N_("Artist"), 'artist', width=200, is_default=True),
        make_field_column(N_("Album Artist"), 'albumartist'),
        make_field_column(N_("Composer"), 'composer'),
        make_field_column(N_("Album"), 'album', sort_type=ColumnSortType.NAT),
        make_field_column(N_("Disc Subtitle"), 'discsubtitle', sort_type=ColumnSortType.NAT),
        make_field_column(N_("Track No."), 'tracknumber', align=ColumnAlign.RIGHT, sort_type=ColumnSortType.NAT),
        make_field_column(N_("Disc No."), 'discnumber', align=ColumnAlign.RIGHT, sort_type=ColumnSortType.NAT),
        make_field_column(N_("Catalog No."), 'catalognumber', sort_type=ColumnSortType.NAT),
        make_field_column(N_("Barcode"), 'barcode'),
        make_field_column(N_("Media"), 'media'),
        make_numeric_field_column(
            N_("Size"),
            '~filesize',
            parse_file_size,
            align=ColumnAlign.RIGHT,
        ),
        make_field_column(N_("File Type"), '~format', width=120),
        make_numeric_field_column(
            N_("Bitrate"),
            '~bitrate',
            parse_bitrate,
            width=80,
            align=ColumnAlign.RIGHT,
        ),
        make_field_column(N_("Genre"), 'genre'),
        create_fingerprint_status_column(),
        make_field_column(N_("Date"), 'date'),
        make_field_column(N_("Original Release Date"), 'originaldate'),
        make_field_column(N_("Release Date"), 'releasedate'),
        make_field_column(N_("Cover"), 'covercount'),
        make_field_column(N_("Cover Dimensions"), 'coverdimensions'),
    )
```

zas: Ok, let's just add is_default parameter and keep variables

knguyen1: [refactor: enhance custom column creation with is_default parameter](https://github.com/metabrainz/picard/pull/2739/commits/21920ffe8e626d440bf2dbaec6fb7062a90cbf1d)

picard/ui/itemviews/columns.py
```diff
@@ -43,156 +43,24 @@
...
-_fingerprint_column.set_header_icon_size(16, 16, 1)
-
-_match_quality_column = MatchQualityColumn(N_("Match"), '~match_quality', width=57)
-_match_quality_column.sortable = True
-_match_quality_column.sort_type = ColumnSortType.SORTKEY
-_match_quality_column.sortkey = _sortkey_match_quality
-_match_quality_column.is_default = True
+from picard.ui.itemviews.custom_columns.common_columns import (
+    create_common_columns,
+    create_match_quality_column,
```
zas: Instead of adding a new file `custom_columns/common_columns.py` I would move the code here in `columns.py` (and ofc totally remove `common_columns.py`).

phw (approved):

phw (approved):
Changes look good and work well for me. LGTM

PR #2783, opened 2025-12-07
Added script for easy .ui file compilation in plugins

phw:
This also handles the translation system and allows easy translation of UI labels with the plugin3 translation system.

# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

If plugins ship Qt designer .ui files (e.g. for option pages) those needs to be converted to Python code. For plugin authors this should be a simple process.

# Solution
This PR adds a `plugin_compile_ui.py` script to handle two things:

1. Generally simplify the process
2. Rewrite the resulting Python code to use the plugin3 translation system

Plugin authors should use the translation keys instead of actual text when designing the .ui file, e.g. like this:

image

The .ui file can then be compiled to Python like this:

```
plugin_compile_ui.py ui_options.ui
```

The resulting code directly uses `PluginApi.tr` for translations.

Example: https://git.sr.ht/~phw/picard-plugin-xqaf/tree/main/item/ui_options.py

### DIFF
--- a/scripts/plugin_compile_ui.py
+++ b/scripts/plugin_compile_ui.py
@@ -1,0 +1,81 @@
+#!/usr/bin/env python3
+"""Convert Qt6 .ui files into Python files for use by Picard plugins.
+
+Usage:
+    plugin_compile_ui.py --help
+    plugin_compile_ui.py <ui_file>
+
+Options:
+    -h, --help   show this help message and exit
+    --force, -f  Force compilation even if up-to-date
+
+Examples:
+    # Compile ui_options.ui to ui_options.py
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

phw:
@zas Sorry, I misunderstood the existing mechanism for ui file translation in https://github.com/metabrainz/picard/blob/master/docs/PLUGINSV3/TRANSLATIONS.md . The entire translation file handling is not needed.

I'll resubmit this, but only for the compilation without the rewriting.

PR #2868, opened 2025-12-16
Refitem refactor, fix refs cache, handling of annotated tags, and switching to buggy versions

zas:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

# Problem

The old code had many issues in the way it was handling references.
Especially with annotated tags, and when switching from a valid version to a buggy one.

# Solution

Rethink and refactor so we make more use of RefItem which was heavily extended.
Better handling of enabling failures (install/reinstall/switch-ref/update), so the user gets a message, the error is properly logged, and the plugin disabled.

Also redesign the refs cache so it is much more efficient (less I/O by batching, instead of having one cache flush per plugin), and cache much more ref infos to avoid disk or network operations.

Get rid of hacky get_ref_type() and replace it with a resolve_ref() method which should be more reliable.

Add more unit tests.

# Action

Additional actions required:

### DIFF
--- a/picard/git/ops.py
+++ b/picard/git/ops.py
@@ -276,7 +276,7 @@ class GitOperations:
                 branch.upstream = branches.remote[f'origin/{ref}']
                 # Now point HEAD to the branch
                 repo.set_head(f'refs/heads/{ref}')
-                log.info('Switched plugin %s to branch %s', plugin.plugin_id, ref)
+                log.debug('Switched plugin %s to branch %s', plugin.plugin_id, ref)
                 return old_ref, ref, old_commit, commit.id
 
             # Try as tag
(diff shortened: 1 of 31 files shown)


phw:
@zas Something is not working right on update checks. I made new commits with a new tag to my playground repo (local install). When checking with the picard master branch and using "Refresh all" it shows as update available.

When I then switch picard to the refitem_refactor branch in this state upon opening settings Picard will now also show the tag version and commit ref in the update column. But when then doing "Refresh all" it will now no longer detect the update and show the plugin as no update available.

Switching back to Picard master the refresh all will show the update again.

zas:
I will redo that, too much changes, too much new bugs.

PR #2975, opened 2026-01-15
PICARD-3160: custom columns bug fixes

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3160

This fixes the two issues reported in PICARD-3160

# Solution

1. After accepting the custom  columns config dialog, the "Clusters" special item lost its title.

The main cause was "Clusters" being a `ClusterList` object, and it had special handling for setting the text. Unified this by introducing `ClisterList.column` and a `ClisterList` ui item class. This allowed removing the special handling and have Clusters treated mostly like any other item.

2. The columns context menu toggled the first item

This was caused by the mouse release even after opening the context menu. This mouse release already caused the `CheckboxMenuItem` to toggle. Updated the implementation to only toggle if the item was previously "down" (pressed).

3. Also some minor code cleanup in columns implementation

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -389,7 +389,11 @@ class UnclusteredFiles(Cluster):
 class ClusterList(list, Item):
     """A list of clusters."""
 
-    def __init__(self):
+    def __init__(self, name=None):
+        if not name:
+            self._name = _('Clusters')
+        else:
+            self._name = name
         super().__init__()
 
     def __hash__(self):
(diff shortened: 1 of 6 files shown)


rdswift (approved):
Not surprising, the code looks good to me.  Tested it here and both issues have been resolved.  Thanks for this.

PR #2993, opened 2026-01-22
PICARD-2025: Actually set ~filepath in metadata

phw:
# Summary

Complements PICARD-2025

The ~filepath variable was documented as part of the TagVar list, but actually not available. Instead this was only an internal variable name used by the metadatabox and itemview filters.

Make this variable actually available to complement existing ~filename and ~dirname.

See also https://github.com/metabrainz/picard/pull/2563


phw:
Note: The alternative approach is to drop this as a variable from tags.py

rdswift (approved):

zas (approved):

PR #3000, opened 2026-01-23
Replace python-dateutil

phw:
# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [x] Other

# Problem

Picard is directly depending on python-dateutil, but the only usecase is for extracting the year in metadata comparison with `picard.util.extract_year_from_date`.

Given that python-dateutil is rather complex, but does not received regular releases I think we should rather get rid of this dependency.

# Solution

Implement a more basic `parse_date` function in `picard.util` and have `extract_year_from_date` use this new function. `parse_date` attempts to convert a string to `datetime` with `datetime.fromisoformat` and if this fails it tries `datetime.strptime` with a couple of pre-defined formats.

This covers all test cases we had previously for `extract_year_from_date` and I added a couple more.

Generally I think this covers our common use cases of date formats.


phw:
See https://github.com/metabrainz/picard/pull/1509 for some previous discussion when we introduced this dependency.

zas (approved):
Thanks for getting rid of the dependency, I agree, not really needed.

PR #3002, opened 2026-01-23
PICARD-3170: Drop deprecated parameter from $matchedtracks function

phw:
# Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

# Problem

* JIRA ticket (_optional_): PICARD-3170

`$matchedtracks()` was never supposed to accept arguments. Due to a mistake in the original definition however it does accept an arbitrary number of arguments. For backward compatibility this was kept, see PICARD-637.

With Picard 3 we can finally clean this up and fix the function signature.

# Solution

Drop the `*args` from `func_matchedtracks`

### DIFF
--- a/picard/script/functions.py
+++ b/picard/script/functions.py
@@ -755,7 +755,7 @@ def func_performer(parser, pattern="", join=", "):
 _Since Picard 0.12_"""
     ),
 )
-def func_matchedtracks(parser, *args):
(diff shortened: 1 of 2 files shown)


rdswift (approved):
Works for me.

PR #3026, opened 2026-02-03
Make external file change detection explicit in save pipeline

iron-prog:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
Make external file change detection explicit in the save pipeline and prevent silent overwrites by aborting the save and marking the file as error when the on-disk file has changed after loading.
## Problem
PR #2963 introduced FileIdentity to detect external file changes.
However, when a file was modified or replaced on disk after being loaded, Picard would only log a warning and continue, potentially overwriting external changes silently.
This could lead to unintended data loss without user awareness.

## Solution
This change:
	•	Reuses FileIdentity to detect external modifications before saving.
	•	Raises ExternalFileModifiedError if the file was modified or deleted.
	•	Aborts the save operation.
	•	Marks the file as ERROR state with a clear message:
“Cannot save file: it was modified externally after being loaded.”
	•	Ensures no silent overwrite occurs.
	•	Keeps behavior batch-safe (no blocking dialogs per file).

This keeps the logic deterministic and avoids UI complexity for batch operations.
## Limitation
	•	No conflict resolution UI is implemented in this PR.
	•	When multiple files are modified externally, each file is marked as error individually.
	•	No batch policy (“overwrite all”, “reload all”) is included yet.
	•	No diff view of tag differences.

These can be explored in a future PR if desired.

## Action

Additional actions required:

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -606,8 +606,7 @@ class File(MetadataItem):
             # Report summary of skipped files (once batch finishes)
             if self.tagger._external_change_count > 0 and not self.is_pending():
                 log.warning(
-                    "%d files were skipped because they were modified externally.",
-                    self.tagger._external_change_count
+                    "%d files were skipped because they were modified externally.", self.tagger._external_change_count
                 )
                 self.tagger._external_change_count = 0
             self._update_filesystem_metadata(self.orig_metadata)
--- a/picard/file.py
+++ b/picard/file.py
@@ -461,6 +461,12 @@ class File(MetadataItem):
         return self.state == File.State.ERROR
 
     def save(self):
+        # Allow retry if previous save failed due to external modification
+        if self.state == File.State.ERROR:
+            change = self._external_file_change(self.filename)
+            if change == ExternalChange.MODIFIED:
+                self._loaded_identity = FileIdentity(self.filename)
+                self.clear_errors()
         self.set_pending()
         run_file_pre_save_processors(self)
         metadata = Metadata()
@@ -568,7 +574,9 @@ class File(MetadataItem):
             if isinstance(error, ExternalFileModifiedError):
                 self.clear_pending(signal=False)
                 self.state = File.State.ERROR
-                self.error_append(_("Cannot save file: it was modified externally after being loaded."))
+                self.error_append(_("Cannot save file: it was modified externally after loading."))
+                # Track batch external changes
+                self.tagger._external_change_count += 1
                 self.update()
                 return
             self._set_error(error)
@@ -595,6 +603,13 @@ class File(MetadataItem):
             self.orig_metadata.update(temp_info)
             self.clear_errors()
             self.clear_pending(signal=False)
+            # Report summary of skipped files (once batch finishes)
+            if self.tagger._external_change_count > 0 and not self.is_pending():
+                log.warning(
+                    _("%d files were skipped because they were modified externally.")
+                    % self.tagger._external_change_count
+                )
+                self.tagger._external_change_count = 0
             self._update_filesystem_metadata(self.orig_metadata)
             if images_changed:
                 self.metadata_images_changed.emit()
--- a/picard/file.py
+++ b/picard/file.py
@@ -50,7 +50,6 @@ from enum import (
 )
 import fnmatch
 from functools import partial
-from enum import Enum, auto
 import hashlib
 import os
 import os.path
@@ -68,6 +67,8 @@ from mutagen import (
     FileType,
     MutagenError,
 )
+
+from PyQt6.QtWidgets import QMessageBox
 
 from picard import (
     PICARD_APP_NAME,
@@ -143,6 +144,14 @@ FILE_COMPARISON_WEIGHTS = {
 class ExternalChange(Enum):
     MISSING = auto()
     MODIFIED = auto()
+
+
+class ExternalFileModifiedError(Exception):
+    """Raised when file changed externally before save."""
+
+    pass
+
+
 class FileIdentityError(Exception):
     pass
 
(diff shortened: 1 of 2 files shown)


iron-prog:
This change is limited to detection/logging only.
If this direction looks reasonable, I can follow up with user-visible handling in a separate PR; otherwise I’m happy to adjust or drop this.

picard/file.py
```diff
@@ -494,11 +514,13 @@ def _save_and_rename(self, old_filename, metadata):
         new_filename = old_filename
         if config.setting['enable_tag_saving']:
             # Detect source changes before saving (debug only)
```
zas, line 528: With this patch, that's not anymore "debug only" right?

zas (changes requested):
The main issue is that the PR adds return statements that abort saves when external changes are detected, whereas the original code only logged warnings and continued. This needs to be clearly documented as it's a significant behavioral change. Users will see no UI feedback when saves are silently aborted.

So I think, to be accepted, the user interface should be updated too, and user should be able to choose what he wants when a change was detected.

It seems that's what you suggest in your comment, and I think that's the way to go before we think about merging this.

picard/file.py
iron-prog: Thanks @zas for the review.
You’re right  silently aborting without user feedback isn’t a good experience. My intention was to prevent unintended overwrites, but I agree that proper UI handling is needed.
I’ll work on adding a dialog so that when an external change is detected, the user is informed and can choose how to proceed. I’ll update the once it's ready.

picard/file.py
```diff
@@ -453,6 +462,27 @@ def keep_original_images(self):
     def has_error(self):
         return self.state == File.State.ERROR
 
+    def _handle_external_modification(self):
+        msg = QMessageBox(self.tagger.window)
```
zas: What happens if multiple files changed?

iron-prog: right now it would show one dialog per file during batch saves, which isn’t ideal.
For this PR, my goal was mainly to prevent silent overwrites. I can simplify it so that externally modified files are just marked with an error state (without showing a dialog per file), and then we can design a proper batch-friendly UI policy in a follow-up change is it ideal.

iron-prog: I have updated it as per requirement

iron-prog: @zas For this PR I’ve kept the behavior minimal : if a file was modified externally after being loaded, Picard now aborts the save and marks the file with an error instead of silently overwriting it. There is no per-file dialog anymore, so batch saves won’t trigger multiple popups.

For future improvements, we could consider a more complete conflict-resolution design, for example:
	•	A batch-aware conflict policy (Ask / Overwrite / Skip)
	•	A single summary dialog listing all externally modified files after a batch save
	•	A per-file “Resolve Conflict” action (Reload / Overwrite) from the UI
	•	Possibly a diff-style metadata comparison before overwriting (longer-term idea)

I think that would deserve a separate, properly designed change so we can handle multi-file scenarios cleanly without complicating the core save pipeline. I want your opinion on this does it sound good to you

zas: @iron-prog ok for me, it makes sense to have improvements in future PRs

zas:
@iron-prog ensure all tests pass (ruff)

iron-prog:
@zas Update on implementation:
	•	Added retry support after external modification
	•	Added batch summary logging for skipped files
	•	Simplified error message

picard/file.py
```diff
@@ -563,6 +603,13 @@ def _saving_finished(self, result=None, error=None):
             self.orig_metadata.update(temp_info)
             self.clear_errors()
             self.clear_pending(signal=False)
+            # Report summary of skipped files (once batch finishes)
+            if self.tagger._external_change_count > 0 and not self.is_pending():
+                log.warning(
+                    _("%d files were skipped because they were modified externally.")
```
zas: Minor remark, but we don't usually translate log output (log should stay in English to ease parsing and ease reading when shared.

iron-prog: I have fixed it ,any other improvement or suggestion

picard/file.py
```diff
@@ -539,6 +571,14 @@ def _saving_finished(self, result=None, error=None):
             return
         old_filename = new_filename = self.filename
         if error is not None:
+            if isinstance(error, ExternalFileModifiedError):
+                self.clear_pending(signal=False)
+                self.state = File.State.ERROR
+                self.error_append(_("Cannot save file: it was modified externally after loading."))
+                # Track batch external changes
+                self.tagger._external_change_count += 1
```
zas, line 579: Where is `_external_change_count` initialized?

picard/file.py
```diff
@@ -494,11 +526,11 @@ def _save_and_rename(self, old_filename, metadata):
...
             # Detect source changes before saving (debug only)
-            current = FileIdentity(old_filename)
-            if current and current != self._loaded_identity:
-                log.warning("File externally modified.")
-            elif not current:
-                log.warning("File missing!")
+            change = self._external_file_change(old_filename)
+            if change == ExternalChange.MISSING:
+                raise FileNotFoundError(f"{old_filename} no longer exists")
+            elif change:
```
zas:
```suggestion
            elif change == ExternalChange.MODIFIED:
```

Right?

picard/file.py
```diff
@@ -450,6 +461,12 @@ def has_error(self):
         return self.state == File.State.ERROR
 
     def save(self):
+        # Allow retry if previous save failed due to external modification
+        if self.state == File.State.ERROR:
+            change = self._external_file_change(self.filename)
+            if change == ExternalChange.MODIFIED:
+                self._loaded_identity = FileIdentity(self.filename)
```
zas, line 468: Between checking for ExternalChange.MODIFIED and creating the new FileIdentity, the file could be modified again. Not sure if that's a real problem though.

picard/file.py
```diff
@@ -563,6 +603,12 @@ def _saving_finished(self, result=None, error=None):
             self.orig_metadata.update(temp_info)
             self.clear_errors()
             self.clear_pending(signal=False)
+            # Report summary of skipped files (once batch finishes)
+            if self.tagger._external_change_count > 0 and not self.is_pending():
+                log.warning(
+                    "%d files were skipped because they were modified externally.", self.tagger._external_change_count
+                )
+                self.tagger._external_change_count = 0
```
zas: I think there's a logic flaw here:

- This assumes the last file to finish saving will reset the counter, but with a single-threaded save pool, this might work accidentally
- If multiple files are saved and some fail with external modifications, the counter might be reset prematurely by a successful save
- The logic ties the counter to individual file state rather than batch completion

zas (changes requested):

picard/file.py
iron-prog: sorry it's my mistake I have created this in tagger.py in init_tagger_entities which I does not uploaded

picard/file.py
iron-prog: there is indeed a small race window there. Since the retry simply refreshes the file identity before the next save attempt, I think the behavior remains consistent with the current optimistic checks.

picard/file.py
iron-prog: Thanks for this

picard/file.py
iron-prog: You’re right — my initial approach incorrectly tied the counter reset to individual file completion instead of batch completion. I see now that this makes the logic fragile even with a single-threaded save pool, because a successful save could reset the counter before all failures are accounted for.

iron-prog: @zas if any other changes requred I'm happy to make and  if it's good then should I work on a prototype of giving option to user for batch changes (overwrite , skip, reload)?

iron-prog:
Hi @zas,
Just checking if there are any remaining changes you'd like before this can move forward.
I’ve addressed the earlier comments.
Happy to make further adjustments if needed.

PR #3030, opened 2026-02-05
PICARD-3186: Improve plugin options handling plugins being unavailable

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3186

Picard can be started with plugins being disabled for currently two reasons: Either pygit2 is not available, or Picard is started with `--no-plugins`.

Both cases would cause issues in the plugin options:

1. If pygit2 was unavailable, the plugin options would fail to initialize and the default error fallback options page would be shown
2. If plugins got disabled with `--no-plugins` this only disabled loading plugins, but still kept the plugin manager. The options UI would load and generally work, but e.g. fail on plugin install.

## Solution

Treat both cases, poygit2 missing or `--no-plugins` parameter, the same when loading Picard and completely disabled initializing the plugin system.

In plugin options show an warning like message with a reason for why the plugins are unavailable:

grafik

grafik

For consistency the warning message is shown the same as the warning in general options when choosing a non-default server.

Note: I first considered not showing the plugin options at all. But I think showing the reason makes it easier for the user to understand why plugins are not available.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -280,7 +280,7 @@ class Tagger(QtWidgets.QApplication):
         self._debug_opts = cmdline_args.debug_opts
         self._debug = cmdline_args.debug or 'PICARD_DEBUG' in os.environ
         self._no_player = cmdline_args.no_player
-        self._no_plugins = cmdline_args.no_plugins
+        self._no_plugins = not HAS_PLUGIN3 or cmdline_args.no_plugins
         self._no_restore = cmdline_args.no_restore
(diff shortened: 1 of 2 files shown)


rdswift (approved):
This looks good to me.  The color selection choice makes sense considering that we don't have user-defined color selections for the warning lines.

zas (approved):
LGTM

PR #3032, opened 2026-02-06
PICARD-3187: Add config upgrade hook to update usage of $matchedtracks()

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

## Problem

* JIRA ticket (_optional_): PICARD-3187

Config upgrade to remove the (previously optional) parameter from $matchedtracks() calls in scripts. This does not change the function of scripts, as the parameter was previously optional and unused.

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -42,7 +42,7 @@ PICARD_APP_NAME = "Picard"
 PICARD_DISPLAY_NAME = "MusicBrainz Picard"
 PICARD_APP_ID = "org.musicbrainz.Picard"
 PICARD_DESKTOP_NAME = PICARD_APP_ID + ".desktop"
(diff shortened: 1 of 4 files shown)


zas (dismissed):
LGTM

phw:
@zas I changed the update slightly. It will now not change `$matchedtracks` if it contains nested funciton calls, e.g. `$matchedtracks($get(foo))` will not be changed.

I think it is rather unlikely to have this. But the previous code would have replaced this to `$matchedtracks())`, breaking the parantheses. Fully handling this would require a proper syntax parsing. But I think for our case here it is ok to leave such cases to the user to manually fix.

zas (approved):

PR #3042, opened 2026-02-15
PICARD-1044: Allow the usage of scripting for loading local cover art

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

Adds a matching mode to the local cover art provider where the entered value is
evaluated as a tagger script producing a file name pattern (with `*`, `?`
wildcards and `{jpg,png}` alternation), so users can load cover art saved with a
scripting pattern.

## Problem

The local cover art provider only supports regular expressions for file name
matching. Users who save cover art with a naming script (e.g.
`%albumartist% - %album% [%date%].ext`) have no way to load it back, because
reconstructing a regex that matches the script-generated filename for arbitrary
metadata is impractical — particularly when many covers share a single directory
(the "singular track" case described by the reporter in [this comment](https://github.com/metabrainz/picard/pull/3042#issuecomment-5282660420)).

* JIRA ticket: PICARD-1044

### Design decisions (resulting from review feedback)

**Single input field, mode toggle (per phw's review):** The UI stays unchanged
except for one checkbox that switches how the value is evaluated. No separate
script editor, preview pane, or second input field. The description label and
note adapt to the active mode.

**One value per mode, not a shared value (profiles & revert safety):** Each mode
stores its value in its own config option (`local_cover_regex`,
`local_cover_script`). This avoids data loss when switching modes: toggling back
restores the other mode's expression. It also makes option profiles clean — each
key holds an unambiguous value type that can be overridden independently.

**String enum for the mode, not a boolean (extensibility):** The active mode is
stored as `LocalCoverMatchMode` (a string enum: `"regex"` / `"script"`),
following the `StandardizeArtistNames` precedent. This lets a future mode (e.g.
full relative-path matching as explored in #3329) be added with just a new enum
member + descriptor table row + option, with no configuration migration. Missing
key → `REGEX` default → backward-compatible.

**Descriptor table (`LOCAL_COVER_MODES`):** An internal dict maps each mode to
its config key, UI strings (description, note, example), so the options page is
table-driven and adding a mode requires no page-code changes.

**Reuses `picard.util.wildcards_to_regex_pattern` (per phw's inline comment):**
The shared utility is extended with backward-compatible, keyword-only flags
(`allow_char_class`, `allow_alternation`, `anchored`) so existing callers keep
their behavior. The local provider uses `allow_char_class=False` (brackets
literal — essential for filenames like `Album [2007].jpg`) and
`allow_alternation=True` (`{jpg,png,gif}` for easy extension matching).
`pattern_as_regex` gains matching pass-through parameters.

**Walk each directory once (performance):** In script mode, tracks of an album
usually produce the same evaluated pattern. `(directory, pattern)` pairs are
deduplicated so each is walked only once, matching the regex mode's efficiency.

### Implementation

- `picard/options.py`: `LocalCoverMatchMode` enum, `LOCAL_COVER_MODES` descriptor
  table, `local_cover_script` + `local_cover_match_mode` options.
- `picard/coverart/providers/local.py`: `ProviderOptionsLocal` (table-driven,
  lossless mode switching, adaptive labels, `_build_matcher` playground check),
  `CoverArtProviderLocal` (mode dispatch, `_eval_script`, `_pattern_to_re`,
  `find_local_images_by_script`, directory walk dedup).
- `picard/util/__init__.py`: `wildcards_to_regex_pattern` extended with
  `allow_char_class`, `allow_alternation`, `anchored` flags.
- `ui/provider_options_local.ui`: one added checkbox widget.
- `picard/const/defaults.py`: `DEFAULT_LOCAL_COVER_ART_SCRIPT`.

### Verified against the reporter's real-world scenario

The example pattern shipped in the SCRIPT-mode help:
```
%albumartist% - %album%$if(%date%, [%date%],).{jpg,png}
```
correctly resolves each of the reporter's actual filenames in a shared `Cover/`
directory (including `Various Artists - Monstercat 001 - Launch Week [2011].jpg`
with internal ` - ` separators and literal brackets). A dedicated regression test
guards this.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

AI was used for the implementation, test development, and PR description based on
design decisions made interactively with the PR author.

## Action

Additional actions required:


phw (commented):
Generally code looks fine and this seems to work.
I just wonder how useful this is, as it only works with exact matches. I initially thought we might be able to allow scripting but still have regex matching.

But this will become difficult, as it would be extremely tricky to have the script return a valid regex. We could escape all variables beforehand, but this then makes the variables difficult to use.

But could we have the resulting filename allow glob matching? So users could have `*` or `?` returned as part of the filename for partial matches. It could be an option.

Also we could allow matching the file name case insensitive.

zas:
Not sure what is the exact use case for this, maybe ask to the 2 people asking for it in the ticket.
At least they can test this code to see if it fits their needs.
Not sure it should be merged (hence the draft)

rdswift:
A bit of thinking out loud here, but rather than using `script_to_filename()`, perhaps you could use `ScriptParser.eval()` to generate the regular expression to be used for checking the file names.  That way the user could include reserved characters in the expression by escaping them (simlar to how it's done in the `$rsearch()` and `$rreplace()` scripting functions.

**EDIT:** There could even be a text box or label showing the regular expression created so the user could check that it made sense for them?

**EDIT 2:**  For example, if the user entered something like:

```taggerscript
.*%album%\\.\(png|jpg\)\$
```
would yield the regular expression:

```
.*Abbey Road\.(png|jpg)$
```

**EDIT 3:** Might need a new scripting function like `$rsanitize()` to replace invalid filename characters with regular expression wildcards to give you something like (assuming the album is "Back in Black" by AC/DC):

```taggerscript
.*$rsanitize(%albumartist% - %album%)\\.\(png|jpg\)\$
```
would yield the regular expression:

```
.*AC.DC - Back in Black\.(png|jpg)$
```

rdswift:
Just thinking that this might even allow the users to set track-specific cover art.  Not something that I would want to do, but I've seen some users ask about this.

rdswift:
> **EDIT 3:** Might need a new scripting function like `$rsanitize()` to replace invalid filename characters with regular expression wildcards

Actually, that could already be accomplished using the `$rreplace()` scripting function.

zas:
> to generate the regular expression

I think this approach is too complex, the glob approach covers most cases, regex generator approach would be more complete but much more complex to implement and to use.
We could add it as a 3rd mode though: raw regex (current), script to glob (this PR), script to regex (your proposal)

phw (dismissed):
I think this makes the provider overly complicated, and I don't really see the use case it solves. For matching certain patterns the regex approach seems more flexible.

From the example given in PICARD-1044 I rather suspect the problem lies elsewhere: The user saves images in a subdirectory as something like `Cover/Album Name.jpg`, then probably expects to be able to load it back from local provider with a regex such as `Cover/.*\.jpg$`. But this doesn't work, as the regex is not evaluated against the full path, but only the filename (but including in subdirectory). Currently you could only set a pattern like `.*\.jpg$`, but this would not limit the match to the "Cover" directory.

As a solution we could extend the logic to match the full relative path as well, see PR #3329 for an (experimental) proposal into this direction.

If the problem is that regex is too complicated to handle, we could also just use the glob approach from this PR and use it as default, with a toggle to change to regex evaluation. But this wouldn't need two separate settings, it would just change how the entered value gets evaluated. The UI would otherwise stay unchanged.

picard/coverart/providers/local.py
```diff
@@ -113,20 +191,90 @@ class CoverArtProviderLocal(CoverArtProvider):
...
+
+        new_metadata = Metadata()
+        for name in metadata:
+            new_metadata[name] = [str(v).replace(os.sep, '_') for v in metadata.getall(name)]
+        script = script.replace('\t', '').replace('\n', '')
+        result = ScriptParser().eval(script, new_metadata)
+        return result.replace('\x00', '')
+
+    @staticmethod
+    def _glob_to_re(pattern):
```
phw: Note: If we keep the globbing logic here we should not re-implement this, but reuse `picard.util.wildcards_to_regex_pattern`. It does not seem to support `{a|b|...}`, but could be extended.

Also `pattern_as_regex` would be interesting if we want to support both globbing and regex.

culinko:
> From the example given in [PICARD-1044](https://tickets.metabrainz.org/browse/PICARD-1044) I rather suspect the problem lies elsewhere: The user saves images in a subdirectory as something like `Cover/Album Name.jpg`, then probably expects to be able to load it back from local provider with a regex such as `Cover/.*\.jpg$`. But this doesn't work, as the regex is not evaluated against the full path, but only the filename (but including in subdirectory). Currently you could only set a pattern like `.*\.jpg$`, but this would not limit the match to the "Cover" directory.

Hi there :) I appreciate all the work that you've all been doing on this!

Unfortunately, using `Cover/.*\.jpg$` wouldn't work in my case if the track was a "singular track" (which I mentioned in my reply in the jira ticket) as there are multiple cover arts in that directory. Let me provide you more examples of how things are structured for me.

example covers for full albums - only one (front) cover image in "Cover" directory

Music\Alicia Keys - The Diary Of Alicia Keys [2003]\Cover\Alicia Keys - The Diary Of Alicia Keys [2003].jpg
Music\blink-182 - Take Off Your Pants And Jacket [2001]\Cover\blink-182 - Take Off Your Pants And Jacket [2001].jpg

example covers for singular tracks - each track that has unique album has (front) cover art in this directory

Music\Cover\The Ting Tings - We Started Nothing [2008].jpg
Music\Cover\Various Artists - Monstercat 001 - Launch Week [2011].jpg
Music\Cover\Various Artists - Spotlight Compilation Vol. 1 [2013].jpg

Also, I found out that the Picard version that I'm currently using (2.12) apparently searches subdirectories automatically, as I was able to find a test cover at Cover\cover.jpg with just the default regex `^(?:cover|folder|albumart)(.*)\.(?:jpe?g|png|gif|tiff?)$`. So I guess I don't need to include the directory part in the cover art loading script at all.

Additionally, I could simplify some parts of my saving script (and therefore also from the loading script) if Picard supported some of these operations natively. For example when saving files (afaik this option is also used for cover art), in File Naming/Compatibility, the Windows compatibility options already support replacing some invalid characters with your own, however the backslash is currently missing. If it was added, I could then drop the $replace part from the scripts. That is, if loading script also accounted for the invalid characters being replaced, which I don't know if it currently does.

Another way could be separating the loading cover art string into 3 different gui elements: (optional) subdirectory, filename and extension. There could also be an option for users to create a list of cover art extensions instead, so they wouldn't need to use regex for this part or the script. However I don't want to complicate your work on this, just throwing around some ideas.

phw:
@culinko Thanks for the additional input. Yes, this case totally makes sense and will require access to the script.

> So I guess I don't need to include the directory part in the cover art loading script at all.

Actually more so, you **can not** currently include the directory part. The pattern matches only filenames (but in subdirectories). That is something I propose to address in #3329, which would allow matching the directory.

> For example when saving files (afaik this option is also used for cover art), in File Naming/Compatibility, the Windows compatibility options already support replacing some invalid characters with your own, however the backslash is currently missing.

The replacement is configurable, it is just not Windows specific. You can specify the replacement for the directory separator in the compatibility options. Both forward and backward slashes are considered directory separators on all platforms.

There is also a difference on how the replacements are applied: The Windows compatibility replacements are applied to the file path generated by the script. The directory separator is applied to tag values before they are passed to the script. This is necessary as otherwise either the script would not be able to create directories or artists like "AC/DC" would result in extra directories being generated.

If you have any code in your naming script that tries to replace slashes it won't have any affect.

culinko:
Thanks for the quick response.

I migh've been confused about the backslash part as I remember (very long time ago) I was either trying to save a file or a cover art that had a backslash in it, which produced an extra directory instead. I think it was long before the options for "Windows compatibility" and "Replace directory separators with:" even existed.

I've checked and I have both of these options configured to use the underscore as the replacement character and you are correct, there's no need for the $replace part in my cover art saving script at all. I just did a quick test using `Cover\\%albumartist% - %title%$if(%date%, [%date%],)` (I changed album into title as I don't have any albums with a backslash) and it correctly saved cover art for Röyksopp's track "40 Years Back\Come" into `Cover\Röyksopp - 40 Years Back_Come [2001].jpg` :)

phw:
> Both forward and backward slashes are considered directory separators on all platforms.

Short correction to this: Forward slash is considered a path separator on all platforms. Backslash only on Windows or if Windows compatibility is enabled.

zas:
@culinko I reworked this, I think it should now answer to your needs, but please test and confirm (or not)
@phw I tried to follow your feedback, though keeping 2 different options has many advantages: one can switch from one mode to another without loosing the value, and it is a good thing too when using option profiles (2 profiles can use different modes/values). Also I went for a checkbox in the UI, but a string for the mode option, it will ease adding more modes (like matching the full path).

picard/coverart/providers/local.py
zas: I extended it and re-used.

zas:
The fixes were merged as https://github.com/metabrainz/picard/pull/3384

The rest isn't really worth at this point, it can be revisited later (post 3.0)
Closing

PR #3040, opened 2026-02-12
PICARD-2670: Add columns file directory, name and extension

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

## Problem

* JIRA ticket (_optional_): PICARD-2670

Add file path related columns to be selectable by default. While they can also be added as custom columns, those are typical use cases for data to display.

## AI Usage

In accordance with the AI use policy portion of the [MetaBrainz Contribution Guidelines](https://github.com/metabrainz/guidelines/blob/master/README.md), the level of AI/LLM use in the development of this Pull Request is:

* [x] No AI/LLM use
* [ ] Minimal use (e.g. autocompletion)
* [ ] Moderate use (e.g. suggestions regarding code fragments)
* [ ] Significant use (e.g. code structure, tests development, etc.)
* [ ] Primarily AI developed
* [ ] Other (please specify below)

## Action

Additional actions required:
* [ ] Update Picard [documentation](https://github.com/metabrainz/picard-docs) (please include a reference to this PR)
* [ ] Other (please specify below)

PR #3043, opened 2026-02-15
PICARD-3196: Implement custom persistence for column state in ConfigurableColumnsHeader

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3196

Changing the default columns for the main views can break the column display as the persisted state does not match the column model anymore. See e.g. #3040 and the example screenshot there.

## Solution
Instead of relying on an opaque binary structure based on logical column indices use a custom structure which stores the state of each column based on its name. This fixes issues with column state breaking if the indices of the default columns change, e.g. when new default columns are addded or removed.

As there is not clean upgrade path there is a config upgrade to remove the old states. The columns will reset to their default state on upgrade.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -385,24 +385,18 @@ class BaseTreeView(QtWidgets.QTreeWidget):
     @restore_method
     def restore_state(self):
         config = get_config()
-        header_state = config.persist[self.header_state]
         header = self.header()
-        if header_state and header.restoreState(header_state):
-            log.debug("Restoring state of %s" % header)
-            for i in range(0, self.columnCount()):
-                header.show_column(i, not self.isColumnHidden(i))
-
+        header_state = config.persist[self.header_state]
+        if header_state:
+            header.restore_columns_state(header_state)
         header.lock(config.persist[self.header_locked])
 
     def save_state(self):
         config = get_config()
         header = self.header()
-        if header.prelock_state is not None:
-            state = header.prelock_state
-        else:
-            state = header.saveState()
-        log.debug("Saving state of %s" % header)
-        config.persist[self.header_state] = state
+        if not header:
+            return
+        config.persist[self.header_state] = header.get_columns_state()
         config.persist[self.header_locked] = header.is_locked
 
     def _set_header_labels(self, update_column_count=False):
--- a/picard/ui/widgets/configurablecolumnsheader.py
+++ b/picard/ui/widgets/configurablecolumnsheader.py
@@ -271,6 +271,36 @@ class ConfigurableColumnsHeader(LockableHeaderView):
         menu.exec(event.globalPos())
         event.accept()
 
+    def get_columns_state(self) -> dict[str, dict]:
+        state = {}
+        sorted_column = self.sortIndicatorSection()
+        sort_order = self.sortIndicatorOrder()
+        for i, column in enumerate(self._columns):
+            column_state = {
+                'visible': not self.isSectionHidden(i),
+                'position': self.visualIndex(i),
+                'width': self.sectionSize(i),
+            }
+            if i == sorted_column:
+                column_state['sorted'] = sort_order.value
+            state[column.key] = column_state
+        return state
+
+    def restore_columns_state(self, state: dict[str, dict]):
+        for i, column in enumerate(self._columns):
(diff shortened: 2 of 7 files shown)


picard/ui/itemviews/basetreeview.py
```diff
@@ -385,24 +385,18 @@ def add_actions(*args):
...
+        header_state = config.persist[self.header_state]
+        if header_state:
+            header.restore_columns_state(header_state)
         header.lock(config.persist[self.header_locked])
 
     def save_state(self):
         config = get_config()
         header = self.header()
-        if header.prelock_state is not None:
-            state = header.prelock_state
```
zas, line 401: I noticed a small issue likely related to this removal, the sort indicator on a column disappears when locking/unlocking.

picard/ui/widgets/configurablecolumnsheader.py
```diff
@@ -271,6 +271,36 @@ def contextMenuEvent(self, event):
...
+            self.show_column(i, column_state.get('visible', column.always_visible))
+            self.resizeSection(i, column_state.get('width', column.width or 100))
+            current_pos = self.visualIndex(i)
+            new_pos = column_state.get('position', -1)
+            if current_pos > 0 and new_pos > 0 and current_pos != new_pos:
+                self.moveSection(current_pos, new_pos)
+            sorted = column_state.get('sorted', None)
+            if sorted is not None:
+                self.setSortIndicator(i, QtCore.Qt.SortOrder(sorted))
+
```
zas, line 359: Adding this method fixes the issue with sort indicator and column locking for me:

```python
    def lock(self, is_locked):
        """Override parent's lock() to use custom dictionary-based state.

        This ensures the sort indicator and other column state is preserved
        across lock/unlock cycles using the same format as save/restore.
        """
        if is_locked:
            self.prelock_state = self.get_columns_state()
            super().lock(True)
        else:
            super().lock(False)
            if self.prelock_state is not None:
                self.restore_columns_state(self.prelock_state)
                self.prelock_state = None
```

picard/ui/itemviews/basetreeview.py
zas: See https://github.com/metabrainz/picard/pull/3043#discussion_r2821625898

picard/ui/widgets/configurablecolumnsheader.py
phw: It's a bit more complicated. The primary reason the prelock state exists in the  `LockableHeaderView` is that it was needed to cleanly restore the column resize mode. Locking sets all columns to fixed, so when we store this state and restore it all columns remain fixed. Hence the state was saved before locking, and resetting the resize mode was restoring the state. We had no other way, since the persisted Qt data is opaque.

But we can do better now, as `ConfigurableColumnsHeader` has proper knowlegde about all columns and we can store the resize mode of each column separately. I think the proper fix is to:

1. Override `lock` completely to disable the default state restore
2. Persists the resize mode

phw:
I found some other issue: After unlocking and then adding / removing columns the header does not directly update. Only after clicking on the header again. It's rather strange, but I cannot reproduce on master.

phw:
@zas Please test again. I fixed the issues we discussed above. I re-implemented the lock handling to not interfere with the default Qt based state restore.

The issue about not being able to add new columns after restore was caused by setting the width for all columns. If you run `resizeSection` on a column that has resize mode set to "stretch" (the last column in the view) Qt gets confused.

zas:
> @zas Please test again. I fixed the issues we discussed above. I re-implemented the lock handling to not interfere with the default Qt based state restore.
> 
> The issue about not being able to add new columns after restore was caused by setting the width for all columns. If you run `resizeSection` on a column that has resize mode set to "stretch" (the last column in the view) Qt gets confused.

Actually found some issues when reviewing and testing.
They should be addressed by https://github.com/phw/picard/pull/29
I added extensive comments and split the code in smaller methods since this code is rather obscure (as it depends on Qt internals).
Tested heavily, I tried to move columns, lock, remove, change sort indicator, etc, and it seems to work in all cases now.

zas (approved):
LGTM

PR #3093, opened 2026-03-16
PICARD-2496: Add option to skip moving files if destination already exists

iron-prog:
## Summary
This PR introduces a new option allowing Picard to skip moving files if the destination file already exists.

New option:
"Skip moving files if destination already exists"

Location:
Options → File Naming → Move files when saving

Behavior priority:
1. Skip move if destination exists (new option)
2. Overwrite existing files
3. Rename with (1), (2) suffix

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
Adds a new configuration option allowing Picard to skip moving files if a file with the same name already exists in the destination directory.

## Problem
When moving files during tagging, Picard currently either renames the file by adding a numbered suffix (e.g., `(1)`) or overwrites the existing file depending on user settings.  
Some users prefer to leave the file untouched if a file with the same name already exists in the destination directory.

* JIRA ticket (_optional_): PICARD-2496

## Solution
This PR introduces a new setting `move_skip_existing_files`.

If enabled, Picard checks whether the destination file already exists before moving a file.  
If it does exist, the move operation is skipped and the file remains in its original location.

Implementation details:

- Added a new `BoolOption` (`move_skip_existing_files`) in `options.py`
- Updated the file renaming logic in `picard/file.py` to skip the move when the option is enabled
- Added a checkbox to the **File Naming → Move files when saving** options page
- Ensured mutual exclusivity with the existing **Overwrite existing files** option to avoid conflicting behavior

Default behavior remains unchanged if the option is disabled.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/ui/forms/ui_options_renaming.py
rdswift, line 6: The changes made in this file should be made in the `ui/options_renaming.ui` file (and the file regenerated using `python setup.py build_ui`), so that they are not lost when the ui files are regenerated in the future.

I guess your AI agent doesn't read the instructions in the file headers.

rdswift (changes requested):
I haven't reviewed this in great depth to try to understand whether it is actually necessary or if this is the best solution.  If it is necessary and appropriate, the documentation will absolutely need to be updated to explain the new configuration setting, but I see that you have indicated that there are no documentation changes required.

I may be wrong, but I suspect that the AI Usage in developing this PR is actually higher than you indicated.

picard/ui/forms/ui_options_renaming.py
iron-prog: AI was used for discussion about implementation approach,
UI integration, and preparing the PR description.
All code was manually reviewed and tested(not regressively but at a significant level).

rdswift:
For clarity for the users, perhaps there could be (mutually exclusive) radio buttons for the three options:

- Do not move duplicate files
- Move and rename file with numeric suffix
- Move and overwrite existing files

That way the preference could be stored in a single setting, rather than multiple binary settings relying on code to maintain their mutual exclusivity.

iron-prog:
thanks for this suggestion it's my bad for relying on ai

rdswift:
> thanks for this suggestion it's my bad for relying on ai

This was a suggestion only, based on my initial cursory review.  Others may have different opinions and different suggested preferences.  I would wait until at least @phw and @zas have had a chance to comment before making any changes.

zas:
> For clarity for the users, perhaps there could be (mutually exclusive) radio buttons for the three options:
> 
>     * Do not move duplicate files
> 
>     * Move and rename file with numeric suffix
> 
>     * Move and overwrite existing files
> 
> 
> That way the preference could be stored in a single setting, rather than multiple binary settings relying on code to maintain their mutual exclusivity.

Yes, I think it would be clearer. 
Also we need to have same thing for additional files for consistency (I think).
Also we have warning dialogs: perhaps they should clearly state what is going to happen (move where? rename how?)

picard/options.py
```diff
@@ -540,6 +539,7 @@ def make_default_toolbar_layout():
     '',
     title=N_("Sessions directory"),
 )
+TextOption('setting', 'move_conflict_strategy', 'rename', title=N_("File conflict handling strategy"))
```
rdswift: This should be moved up into the `picard/ui/options/renaming.py` section (around Line 430 above).

picard/ui/options/renaming.py
```diff
@@ -75,9 +75,9 @@ class RenamingOptionsPage(OptionsPage):
     OPTIONS = (
         ('move_files', ['move_files']),
         ('move_files_to', ['move_files_to']),
-        ('move_overwrite_existing_files', ['move_overwrite_existing_files']),
         ('move_additional_files', ['move_additional_files']),
         ('move_additional_files_pattern', ['move_additional_files_pattern']),
+        ('move_conflict_strategy', ['move_conflict_strategy']),
```
rdswift: The list doesn't appear to contain ui elements that can be highlighted, as explained in the instructions (Item 2) starting on line 78 of `picard/options.py`.

picard/ui/options/renaming.py
```diff
@@ -250,8 +250,11 @@ def load(self):
         self.ui.move_files_to.setCursorPosition(0)
         self.ui.move_additional_files.setChecked(config.setting['move_additional_files'])
         self.ui.move_additional_files_pattern.setText(config.setting['move_additional_files_pattern'])
+        strategy = config.setting['move_conflict_strategy']
+        self.ui.move_conflict_skip.setChecked(strategy == "skip")
+        self.ui.move_conflict_rename.setChecked(strategy == "rename")
+        self.ui.move_conflict_overwrite.setChecked(strategy == "overwrite")
```
rdswift: What happens if the value of 'move_conflict_strategy' doesn't happen to be one of these choices?  That shouldn't happen, but the string in the configuration could be accidentally changed in a plugin or something.  I assume this will result in none of the radio boxes being checked.  Will this be a problem when saving?  It might be better to use an "if -> elif -> else" strategy (either here or in the `save()` method, or both).

picard/ui/options/renaming.py
```diff
@@ -285,9 +288,14 @@ def save(self):
         config.setting['move_files'] = self.ui.move_files.isChecked()
         config.setting['move_files_to'] = os.path.normpath(self.ui.move_files_to.text())
         config.setting['move_additional_files'] = self.ui.move_additional_files.isChecked()
+        if self.ui.move_conflict_skip.isChecked():
+            config.setting['move_conflict_strategy'] = "skip"
+        elif self.ui.move_conflict_rename.isChecked():
+            config.setting['move_conflict_strategy'] = "rename"
+        elif self.ui.move_conflict_overwrite.isChecked():
+            config.setting['move_conflict_strategy'] = "overwrite"
```
rdswift: What happens if none of the radio buttons is checked?  (See my comment in the `load()` method above.)

picard/file.py
```diff
@@ -676,8 +676,16 @@ def _rename(self, old_filename, metadata, settings):
...
+        config = get_config()
+        strategy = config.setting.get('move_conflict_strategy', 'rename')
+        if os.path.exists(new_filename):
+            if strategy == "skip":
+                log.warning("Destination exists, skipping move of %r", old_filename)
+                return old_filename
+            elif strategy == "rename":
+                new_filename = get_available_filename(new_filename, old_filename)
+            elif strategy == "overwrite":
+                pass
```
rdswift: This will get "rename" as the value if there is no 'move_conflict_strategy' setting, but will actually process as "overwrite" if there is a value other than "skip", "rename" or "overwrite".

Also, this code is repeated below in the `_apply_additional_file_moves()` method below.  You might want to consider making this a separate helper method called from the other methods.

rdswift (changes requested):
@zas / @phw would it make sense to define the choices in an Enum or IntEnum class rather than storing / retrieving / comparing a string value?

iron-prog:
@rdswift  Thanks for the detailed review!
I’ve made the changes as your suggestion:
	•	Refactored duplicate conflict-handling logic into a helper method
	•	Added validation and fallback handling for unexpected strategy values
	•	Updated the UI logic to ensure a default option is always selected
	•	Cleaned up related code paths for consistency

rdswift:
Please provide meaningful descriptions for your commits.  The commit message "ruff fail" on everything does not provide any useful information as to what reviewers can expect to see when they review the commit.

Also, please don't continually force push each time.  That destroys the revision history for the PR and causes the reviewers to have to go through all of the changes each time, rather than what has changed incrementally since their last review (causing much more work for the reviewers and making them less likely to review and accept the PR).

I suggest that you review the following best practices and consider incorporating them into your workflow:

- https://www.w3schools.com/git/git_best_practices.asp
- https://www.ojambo.com/git-best-practices-commit-often-commit-early

iron-prog:
there is some problem going on authentication in  GitHub in my terminal like I have one previous git  account which I deleted and second which I m working right now but some times it's just does not give access(or switch randomly although no sign it is in my terminal or workflow) due to which I have to force push after many changes I have tried many ways to fix it because of this I use force push and for commit message I use what I changes but sometimes in hurry or simple failure I message this

rdswift:
In that case, I suggest that you get this sorted out and working properly before continuing to make any PR submissions.  I plan to avoid any further reviews until this has been addressed.

iron-prog:
Thanks for pointing this out, and apologies for the workflow issues.
Going forward, I’ll use incremental commits with clear messages to make review easier.

picard/options.py
```diff
@@ -425,9 +425,9 @@ def make_default_toolbar_layout():
 BoolOption('setting', 'delete_empty_dirs', True, title=N_("Delete empty directories"))
 BoolOption('setting', 'move_additional_files', False, title=N_("Move additional files"))
 TextOption('setting', 'move_additional_files_pattern', "*.jpg *.png", title=N_("Additional file patterns"))
+TextOption('setting', 'move_conflict_strategy', 'rename', title=N_("File conflict handling strategy"))
 BoolOption('setting', 'move_files', False, title=N_("Move files"))
 TextOption('setting', 'move_files_to', DEFAULT_MUSIC_DIR, title=N_("Destination directory"))
-BoolOption('setting', 'move_overwrite_existing_files', False, title=N_("Overwrite existing files"))
```
rdswift: Removing the old setting should also be done in the `config_upgrade.py` module, which should also be used to set the initial value of the new setting based on the value of the old setting.

rdswift (changes requested):

picard/options.py
iron-prog: I have implemented as your suggestion

picard/config_upgrade.py
```diff
@@ -726,3 +726,18 @@ def upgrade_config(config):
...
+            old_value = config.setting['move_overwrite_existing_files']
+
+            if old_value:
+                config.setting['move_conflict_strategy'] = "overwrite"
+            else:
+                config.setting['move_conflict_strategy'] = "rename"
+
+            del config.setting['move_overwrite_existing_files']
+        else:
+            config.setting['move_conflict_strategy'] = "rename"
```
rdswift: This won't work as written.  Please actually review and test the code that your AI agent provides.

picard/config_upgrade.py
```diff
@@ -650,6 +650,20 @@ def upgrade_to_v3_0_0a3(config):
     """Remove persisted column configuration"""
     config.persist.remove('album_view_header_state')
     config.persist.remove('file_view_header_state')
+    print("UPGRADE RUNNING")
```
rdswift: Is this a left over from testing?  There should not be any `print()` statements in the final code.

picard/config_upgrade.py
```diff
@@ -650,6 +650,20 @@ def upgrade_to_v3_0_0a3(config):
...
+            old_value = config.setting['move_overwrite_existing_files']
+
+            if old_value:
+                config.setting['move_conflict_strategy'] = "overwrite"
+            else:
+                config.setting['move_conflict_strategy'] = "rename"
+
+            config.setting.remove('move_overwrite_existing_files')
+        else:
+            config.setting['move_conflict_strategy'] = "rename"
```
rdswift: Since v3.0.0alpha3 has already been made publicly available, this update code should be included in a new method, and the Picard version number updated appropriately (v3.0.0alpha4 ?).

picard/config_upgrade.py
iron-prog: Thanks for the feedback and your patience.
I’m still in the process of learning the project, and your comments helped me understand the expected approach better.
I’ve now addressed the issues, especially around config migration and edge cases, and tested the changes with different config scenarios.
Please let me know if anything else needs improvement.

rdswift: > Please let me know if anything else needs improvement.

You still need to address my comment in https://github.com/metabrainz/picard/pull/3093#discussion_r2971609131 : The list doesn't appear to contain ui elements that can be highlighted, as explained in the instructions (Item 2) starting on line 78 of `picard/options.py`.

picard/ui/options/renaming.py
```diff
@@ -38,7 +38,7 @@
 from PyQt6 import QtWidgets
 from PyQt6.QtGui import QPalette
 
-from picard.config import get_config
+from picard.config import get_config, log
```
rdswift: Although this should work, I believe that `log` should actually be imported directly from `picard`.

picard/ui/options/renaming.py
rdswift: Your new option seems to be missing from this tuple.

rdswift (changes requested):

rdswift:
Please DO NOT keep using the same commit message for everything.  The commit message should indicate what has changed in that commit.

Also, you requested my review again, but you still haven't addressed all of my previous comments.

iron-prog:
Thanks for the review I was busy with my sem examination so it's my mistake I m going to address all concerns now with precision and as you suggested.

picard/config_upgrade.py
iron-prog: yes it's just a left over

picard/config_upgrade.py
iron-prog: I m going to use v3.0.0alpha4

picard/config_upgrade.py
rdswift: Rather than checking for the existence of 'move_conflict_strategy', I believe that you should set the value of 'move_conflict_strategy' based on the value of 'move_overwrite_existing_files' (as you have done) and then remove the 'move_overwrite_existing_files' setting (because it has been superceded by the new 'move_conflict_strategy' setting).  If you don't remove the old setting in the upgrade method, it will be left as an orphan in the INI file.

rdswift (changes requested):

picard/config_upgrade.py
iron-prog: addressed this improvement

iron-prog:
@rdswift any update on this

rdswift:
I don't think you have addressed @zas's comments yet.

**EDIT:** Just to clarify, I'm not saying that all the code changes have to have been made, but there should at least be some discussion as to how the comments could be addressed (or even if there should be code changes to accommodate).  The best scenario would be to discuss the options as to how the comments could be addressed, and ultimately arrive at some consensus before commiting time and energy to actual code changes.

iron-prog:
I m going with this , 
* skip - Do not move duplicate files
* rename - Move and rename file with numeric suffix
* overwrite - Move and overwrite existing files

does it's good or some more option

rdswift:
Have you thought about @zas's comment about the warning dialogs?

iron-prog:
> Have you thought about @zas's comment about the warning dialogs?

Yes I am going to address those also

picard/config_upgrade.py
```diff
@@ -652,6 +652,20 @@ def upgrade_to_v3_0_0a3(config):
     config.persist.remove('file_view_header_state')
 
 
+def upgrade_to_v3_0_0a4(config):
```
zas, line 655: A matching test is required, should be in `test/test_config_upgrade.py`

picard/ui/forms/ui_options_renaming.py
```diff
@@ -55,9 +55,15 @@ def setupUi(self, RenamingOptionsPage):
...
-        self.verticalLayout_4.addWidget(self.move_overwrite_existing_files)
+        self.move_conflict_skip = QtWidgets.QRadioButton(parent=self.move_files)
+        self.move_conflict_skip.setObjectName("move_conflict_skip")
+        self.verticalLayout_4.addWidget(self.move_conflict_skip)
+        self.move_conflict_rename = QtWidgets.QRadioButton(parent=self.move_files)
+        self.move_conflict_rename.setObjectName("move_conflict_rename")
+        self.verticalLayout_4.addWidget(self.move_conflict_rename)
+        self.move_conflict_overwrite = QtWidgets.QRadioButton(parent=self.move_files)
+        self.move_conflict_overwrite.setObjectName("move_conflict_overwrite")
+        self.verticalLayout_4.addWidget(self.move_conflict_overwrite)
```
zas, line 71: The three `QRadioButton` widgets (`move_conflict_skip`, `move_conflict_rename`, `move_conflict_overwrite`) are children of the move_files `QGroupBox`, which means Qt will auto-group them since they share a parent. This works, but it's fragile, if another radio button is ever added to that group box, it'll get lumped in.
Adding an explicit `QButtonGroup` in the .ui file is safer and more conventional.

picard/file.py
```diff
@@ -449,6 +449,22 @@ def keep_original_images(self):
     def has_error(self):
         return self.state == File.State.ERROR
 
+    def _handle_conflict(self, new_path, old_path):
+        config = get_config()
```
zas: The method calls `get_config()` and reads `move_conflict_strategy` each time it's invoked. During         
  `_apply_additional_files_moves`, this is called in a loop for every additional file. The strategy won't change mid-save, so it would be cleaner to read it once and pass it in, or at least read it once at the caller level.

picard/file.py
```diff
@@ -449,6 +449,22 @@ def keep_original_images(self):
...
+        strategy = config.setting['move_conflict_strategy']
+        if strategy not in ("skip", "rename", "overwrite"):
+            strategy = "rename"
```
zas: A log.warning() would be great here, instead of silently defaulting to rename strategy.

test/test_savewarningdialog.py
```diff
@@ -0,0 +1,148 @@
...
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from picard import config
+
+import pytest
```
zas: The standard pattern in this project is: 
  - Subclass PicardTestCase (from test.picardtestcase) 
  - Use setUp / tearDown for setup
  - Use self.assert* for assertions
  - Use unittest.mock.patch instead of monkeypatch

This bare pytest isn't matching that, see other test files.

picard/ui/options/renaming.py
```diff
@@ -250,8 +251,17 @@ def load(self):
         self.ui.move_files_to.setCursorPosition(0)
         self.ui.move_additional_files.setChecked(config.setting['move_additional_files'])
         self.ui.move_additional_files_pattern.setText(config.setting['move_additional_files_pattern'])
+        strategy = config.setting['move_conflict_strategy'] or "rename"
+        if strategy == "skip":
```
zas: For strategies, it would be preferable to use an enum:

```python
  class MoveConflictStrategy(str, Enum):
      SKIP = "skip"                                                                                              
      RENAME = "rename"
      OVERWRITE = "overwrite"
```

We could use [StrEnum](https://docs.python.org/3/library/enum.html#enum.StrEnum), but that's Python 3.11+

When reading config, you can use:

```python
strategy = MoveConflictStrategy(config.setting['move_conflict_strategy'])
```

Or even:

```python
  class MoveConflictStrategy(str, Enum):
      SKIP = "skip"
      RENAME = "rename"
      OVERWRITE = "overwrite"
                                                                                                                                                           
      @classmethod
      def from_config(cls, value, default=None):                                                                                                           
          if default is None:
              default = cls.RENAME
          try: 
              return cls(value)
          except ValueError: 
              log.warning("Invalid move_conflict_strategy: %r, defaulting to %r", value, default.value) 
              return default
```

Then every call site becomes just:
```python
strategy = MoveConflictStrategy.from_config(config.setting['move_conflict_strategy'])
```

It would simplify the code in many places.

picard/file.py
```diff
@@ -449,6 +449,22 @@ def keep_original_images(self):
     def has_error(self):
         return self.state == File.State.ERROR
 
+    def _handle_conflict(self, new_path, old_path):
```
zas: `_handle_conflict` is on the `File` instance but doesn't use `self`, it could be a static method or a module level function.

zas (changes requested):
A lot of small issues in the PR, and a messy commit history.
As is, it cannot be merged.

Also be sure to respect project conventions.

Feature-wise, no problem, but the code needs to be polished.

picard/file.py
iron-prog: sorry! I m going to solve all of this

picard/ui/options/renaming.py
iron-prog: thanks for this

picard/file.py
iron-prog: should I use force push to clear the messy commit?

iron-prog: any other changes? then I clear the messy history

zas: You can re-organize your branch as you wish, in general we prefer to keep a nice commit history, with clear testable/reversible commits, whenever possible. This also helps reviewing changes (commit per commit).

zas: A good example of what you shouldn't do is your last commit titled "[WIP](https://github.com/metabrainz/picard/pull/3093/commits/422b69297953136a48a1f00ce645c0eb3ec39470)", instead I would expect one commit per logical change so we can reference them: "Fix ... ", "Improve ...", "Implement ..." titles and clear commit messages explaining what's done.

Small logical commits are also very important when it comes to use `git bisect` to find a bug or regression in the git history. So better have many small commits than a big one. A balance has to be found between too many and not enough ;)

iron-prog: Thanks 😊 I m going to keep this in mind

Jormangeud:
I hope you took into account that renaming/moving files is usually release based, so the skip should activate for all the files if the target directory already exists. Otherwise you may get some files moved and some left in the original place.
If there is not yet a way for distinguishing whether saving wanted file(s) or releases, there should be a big fat warning on the option.
Having the UI color the skipped files would help, but then again, if some files from a "complete" release get moved and some not, it likely should not happen.

ui/options_renaming.ui
```diff
@@ -298,14 +321,19 @@
...
+  <tabstop>move_conflict_rename</tabstop>
+  <tabstop>move_conflict_overwrite</tabstop>
   <tabstop>rename_files</tabstop>
   <tabstop>naming_script_selector</tabstop>
   <tabstop>open_script_editor</tabstop>
   <tabstop>example_filename_before</tabstop>
-  <tabstop>example_filename_after</tabstop>
-  <tabstop>example_filename_sample_files_button</tabstop>
+ <tabstop>example_filename_after</tabstop>
+ <tabstop>example_filename_sample_files_button</tabstop>
```
zas: Minor indentation issue in the 2 lines above

picard/file.py
```diff
@@ -674,9 +687,12 @@ def _rename(self, old_filename, metadata, settings):
             return old_filename
 
         new_dirname = os.path.dirname(new_filename)
-        os.makedirs(new_dirname, exist_ok=True)
-        if not settings['move_overwrite_existing_files']:
-            new_filename = get_available_filename(new_filename, old_filename)
+        if not os.path.isdir(new_dirname):
+            os.makedirs(new_dirname)
```
zas: `os.makedirs(new_dirname, exist_ok=True)` does the same as the 2 lines above

picard/config_upgrade.py
```diff
@@ -652,6 +652,22 @@ def upgrade_to_v3_0_0a3(config):
...
+    """Rename move_overwrite_existing_files to move_conflict_strategy."""
+    if 'move_overwrite_existing_files' in config.setting:
+        with temp_option(BoolOption, 'setting', 'move_overwrite_existing_files', False) as old_opt:
+            old_value = config.setting.value(old_opt, False)
+
+        if old_value:
```
zas: Since this depends on `old_value` this if/else should be in `with` too.

picard/ui/options/renaming.py
```diff
@@ -285,9 +292,16 @@ def save(self):
...
         config.setting['move_files_to'] = os.path.normpath(self.ui.move_files_to.text())
         config.setting['move_additional_files'] = self.ui.move_additional_files.isChecked()
+        if self.ui.move_conflict_skip.isChecked():
+            config.setting['move_conflict_strategy'] = MoveConflictStrategy.SKIP.value
+        elif self.ui.move_conflict_rename.isChecked():
+            config.setting['move_conflict_strategy'] = MoveConflictStrategy.RENAME.value
+        elif self.ui.move_conflict_overwrite.isChecked():
+            config.setting['move_conflict_strategy'] = MoveConflictStrategy.OVERWRITE.value
+        else:
+            config.setting['move_conflict_strategy'] = MoveConflictStrategy.RENAME.value
```
zas: We should probably have a `MoveConflictStrategy.default()` and use it to define the default action, instead of assuming that's RENAME.

picard/ui/savewarningdialog.py
```diff
@@ -31,9 +31,50 @@
...
+                "move the files to <em>{move_to}</em> (skip moving if destination file exists)",
+                file_count,
+            ).format(move_to=move_to)
+        if strategy == MoveConflictStrategy.OVERWRITE:
+            return ngettext(
+                "move the file to <em>{move_to}</em> (overwrite destination file if it exists)",
+                "move the files to <em>{move_to}</em> (overwrite destination file if it exists)",
+                file_count,
+            ).format(move_to=move_to)
+        return ngettext(
```
zas: This assumes the default is `RENAME` but I would prefer we actually don't make such assumption.

picard/config_upgrade.py
```diff
@@ -652,6 +652,22 @@ def upgrade_to_v3_0_0a3(config):
...
+            config.setting['move_conflict_strategy'] = "overwrite"
+        else:
+            config.setting['move_conflict_strategy'] = "rename"
+
+        config.setting.remove('move_overwrite_existing_files')
+    else:
+        config.setting['move_conflict_strategy'] = "rename"
```
zas: Here too, we assume the default is rename.

picard/move_conflict.py
```diff
@@ -0,0 +1,39 @@
...
+
+class MoveConflictStrategy(str, Enum):
+    SKIP = "skip"
+    RENAME = "rename"
+    OVERWRITE = "overwrite"
+
+    @classmethod
+    def from_config(cls, value, default=None):
+        if default is None:
+            default = cls.RENAME
```
zas: Instead let's define the default, it could be a method or a property, but a method would avoid a chicken-egg caveat (I think):

```python

@classmethod
def default(cls):
    return cls.RENAME
```

```python
        if default is None:
            default = cls.default()
```

picard/move_conflict.py
```diff
@@ -0,0 +1,39 @@
...
+        try:
+            return cls(value)
+        except ValueError:
+            log.warning("Invalid move_conflict_strategy: %r, defaulting to %r", value, default.value)
```
zas, line 41: We only need `default` here, so the definition can be moved here.

picard/move_conflict.py
iron-prog: that's good a way

picard/ui/options/renaming.py
iron-prog: I agree, this can more user friendly than renaming

iron-prog: should we ask to user for any action or set it to default?

picard/config_upgrade.py
```diff
@@ -652,6 +653,20 @@ def upgrade_to_v3_0_0a3(config):
     config.persist.remove('file_view_header_state')
 
 
+def upgrade_to_v3_0_0a4(config):
```
rdswift: Now that Picard 3.0beta1 is out, will this method need to be renamed?

iron-prog: I’ll keep upgrade_to_v3_0_0a4 as-is because upgrade hooks are versioned by the release where the migration was introduced, and they’re executed in version order. Renaming it now could change migration history.
If we need any additional migration for beta1, I’ll add a separate upgrade_to_v3_0_0beta1 does this right to you?.

rdswift: > If we need any additional migration for beta1, I’ll add a separate upgrade_to_v3_0_0beta1 does this right to you?.

Not really, since I suspect that it will never be executed for people already running v3.0b1, which is currently publicly available.  Remember that the Picard version number is stored in the configuration file.

iron-prog: I agree "since I suspect that it will never be executed for people already running v3.0b1, which is currently publicly available." I am going to use higher version like `upgrade_to_v3_0_0b2` so it work fine on current and future version also and I am using a version bump from b1 -> b2 does this right or I have to go with b1 problem with b1  3.0.0b1 wouldn’t execute this migration if it remained at b1

iron-prog:
@zas Since the PR has been open for a while and master has changed considerably, I’m currently extracting the feature into a clean branch and resolving conflicts. before I spend more time rebasing and porting the remain changes, I wanted to ask whether the proposed move conflict strategy (skip / rename / overwrite) is still the preferred direction for this feature.
Would you recommend continuing to update this PR, or would you prefer a fresh PR based on current master?

zas:
@iron-prog you can update this PR, but you have to clean your branch history, and of course address last remarks.

picard/__init__.py
```diff
@@ -49,6 +49,12 @@
 PICARD_VERSION = Version(3, 0, 0, 'beta', 4)
 
 
+# optional build version
+# it should be in the form '<platform>_<YYMMDDHHMMSS>'
+# ie. win32_20140415091256
+PICARD_BUILD_VERSION_STR = ""
+
```
zas, line 52: Duplicated and useless block of code.

You should self-review to detect such issues.

ui/options_renaming.ui
```diff
@@ -80,9 +80,32 @@
...
       </item>
       <item>
-       <widget class="QCheckBox" name="move_overwrite_existing_files">
+       <widget class="QRadioButton" name="move_conflict_skip">
+        <attribute name="buttonGroup">
+         <string notr="true">move_conflict_group</string>
+        </attribute>
         <property name="text">
-         <string>Overwrite existing files</string>
+          <string>Do not move duplicate files</string>
```
zas: Extra leading space here

picard/ui/savewarningdialog.py
```diff
@@ -31,9 +31,54 @@
...
+                "move the files to <em>{move_to}</em> (overwrite destination file if it exists)",
+                file_count,
+            ).format(move_to=move_to)
+        if strategy == MoveConflictStrategy.RENAME:
+            return ngettext(
+                "move the file to <em>{move_to}</em> (rename with a numeric suffix if destination file exists)",
+                "move the files to <em>{move_to}</em> (rename with a numeric suffix if destination file exists)",
+                file_count,
+            ).format(move_to=move_to)
+
```
zas, line 58: `_get_move_action` can implicitly return `None`: The last `if` should be an `else` (or return the `RENAME` case as the final fallback), consistent with how `_get_additional_files_action` handles it.
Inconsistent style between the two helper methods: `_get_move_action` uses early `if/if/if` returns; `_get_additional_files_action` uses` if/elif/else`. It should be unified to use the latter. Same for  `_handle_conflict`.

zas (changes requested):
Commit history is still messy, you can do better on this:
  
- Duplicate messages: two commits named "Remove leftover migration code and fix unused imports in config_upgrade"
- Non-descriptive: "renaming the method from 0a4->b2"
- Commits that undo/redo previous ones
  
This should be squashed into 2-3 logical commits (e.g.: 1. Add MoveConflictStrategy enum + file logic, 2. Update UI and options page, 3. Add config migration + tests).


Also [Jormangeud's concern](https://github.com/metabrainz/picard/pull/3093#issuecomment-4286781557) was never addressed, answer at least.

iron-prog:
> I hope you took into account that renaming/moving files is usually release based, so the skip should activate for all the files if the target directory already exists. Otherwise you may get some files moved and some left in the original place. If there is not yet a way for distinguishing whether saving wanted file(s) or releases, there should be a big fat warning on the option. Having the UI color the skipped files would help, but then again, if some files from a "complete" release get moved and some not, it likely should not happen.

@Jormangeud — great point, thanks for raising it.
You're right that file moves in Picard are release-scoped: if a release has 12 tracks and only some get moved while others are skipped, it creates a split state that's confusing and hard to recover from.
After thinking about it, the current implementation applies skip per-file independently, which can result in exactly the partial-move scenario you described.
I see a few options for how to handle this:

Warn in the UI — Add a note next to the "Do not move duplicate files" radio button explaining that this operates per-file and can result in a release being split across directories.
Skip the entire release — If any file in a release would be skipped (destination exists), skip all files in the release and show a warning. This is safer but requires release-level coordination in the save logic.
Color/mark skipped files — As you suggested, visually distinguish skipped files in the file browser.

My current implementation does option 1 (warn in UI) as the minimum. If the maintainers agree option 2 is worth implementing, I'm happy to do it, but it's a larger change. What's the preferred direction?

PR #3126, opened 2026-04-17
Fix initial load of plugin registry and improve related code/documentation

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

- On a fresh install, the plugin registry list is empty — no remote fetch is triggered when cache doesn't exist, affecting both UI (Install Plugin dialog) and CLI (`--browse`, `--search`, `--install by ID`) 
- `find_similar_plugin_ids` crashes with TypeError due to dict-style access on `RegistryPlugin` — no test covered this path                                  
- Registry cache uses an unnecessary JSON wrapper around TOML data, adding complexity
- `AsyncPluginRegistry` is dead code that duplicates `PluginRegistry.fetch_registry()` with wrong format (JSON instead of TOML)
- `InstallPluginDialog` crashes when registry loads from cache because the fetch callback fires before UI widgets are created
- "Registry not loaded" debug message is logged dozens of times during startup (once per registry lookup per plugin)
- When a fallback registry URL succeeds, the cache is saved under the primary URL's hash but `_load_from_cache` only checks the primary — so the cache is  never found on next startup
- Registry cache can be corrupted: invalid TOML from remote could be cached, and partial writes on disk full leave truncated files

## Solution

- Auto-fetch registry when cache is empty — CLI `_ensure_registry()` helper with sync fetch; UI reloads plugin list when async fetch completes
- Fix `p['id']` → `p.id` in `find_similar_plugin_ids`, add regression test
- Cache raw TOML bytes directly — remove JSON layer, REGISTRY_CACHE_VERSION, RegistryCacheError; change cache extension to .toml; update docs
- Remove `AsyncPluginRegistry`, its test, and export
- Move `setup_ui()` before `fetch_registry()` in `InstallPluginDialog.__init__`
- Log "registry not loaded" once with `_logged_not_loaded` flag
- Add `_cache_path_for_url()` helper; `_load_from_cache()` tries all registry URLs; `_update_cache_path() `sets correct path on fallback success
- Parse TOML before caching (only save valid data); use atomic writes (temp file + rename) via new shared `atomic_write()` utility extracted from session_manager

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


phw (approved):
LGTM

rdswift (approved):
Looks good, and seems to work as expected when I tested it here.

PR #3134, opened 2026-04-23
Use consistent short version format in `__version__`, ensure PEP440 compliance

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

## Problem

When we have a build version, `__version__` is built with `PICARD_VERSION_STR` but with `PICARD_VERSION_STR_SHORT` when not.
`PICARD_VERSION_STR` isn't [PEP 440](https://peps.python.org/pep-0440/) compliant.

## Solution

Use `PICARD_VERSION_STR_SHORT` in both cases and ensure it is PEP 440-compliant by matching generated versions against [provided regex](https://peps.python.org/pep-0440/#appendix-b-parsing-version-strings-with-regular-expressions).

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/__init__.py
```diff
@@ -45,20 +45,17 @@
...
 
 
+PICARD_VERSION_STR = str(PICARD_VERSION)
+__version__ = PICARD_VERSION.short_str()
+PICARD_FANCY_VERSION_STR = __version__
+
 # optional build version
-# it should be in the form '<platform>_<YYMMDDHHMMSS>'
-# ie. win32_20140415091256
+# it should be in the form '<platform>.<YYYYMMDDHHMMSS>'
```
phw, line 53: As a side note, we don't actually do this. We exclusively use `PICARD_BUILD_VERSION_STR` for non-tagged builds, and always set it to the commit short hash for both macOS and Windows build

phw (commented):
Looks good. I think we should do this.

picard/__init__.py
zas: https://github.com/metabrainz/picard/blob/17c4eabcf619446b8cda18a0cbbea31fcce7e2cc/setup.py#L668

Isn't that used?

phw: Your are right. But we don't us the platform that way. Currently patch_version is being called with platform set to the commit count since last release tag + the commit short hash. So e.g. something like this "3.0.0.alpha4+152.34c856761.20260422232144" . Which is more useful than the OS (you know this anyway when running). See https://github.com/metabrainz/picard/blob/master/.github/workflows/package-macos.yml#L56C9-L56C100

Maybe we need to rename the parameter.

zas: Yes, probably. But that's not directly related to this PR.

phw (approved):

PR #3136, opened 2026-04-23
Refactor WebService / APIHelper to not default to mblogin=True

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

The WebService / APIHelper classes support the `mblogin` parameter, which enables MusicBrainz authentication for a request. This is enabled by default for post / put / delete.

But we use those classes for all web requests. That means every non-MB use, including plugins, needs to set `mblogin=False`.

## Solution

Revert the logic and default to `mblogin=False`. Only enable it by default in `MBAPIHelper`. Other uses that rely on MB calls, such as get requests, already set `mblogin=True` explicitly, when needed.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/webservice/api_helpers/apihelper.py
+++ b/picard/webservice/api_helpers/apihelper.py
@@ -66,7 +66,6 @@ class APIHelper:
         kwargs['url'] = self.url_from_path(path)
         kwargs['handler'] = handler
         kwargs['data'] = data
-        kwargs['mblogin'] = kwargs.get('mblogin', True)
         return self._webservice.post_url(**kwargs)
 
     def put(self, path: str, data: str | None, handler: ReplyHandler, **kwargs):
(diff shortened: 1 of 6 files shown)


picard/webservice/api_helpers/apihelper.py
zas: Shouldn't this one be removed too?

zas (changes requested):

picard/webservice/api_helpers/apihelper.py
phw: Yes, done

zas (approved):
LGTM, good move

PR #3143, opened 2026-04-30
Patch version helper

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [x] Other

## Problem

For development build (not build on a release tag) we are adding the commit count since last release tag + the current commit hash to the version using the `setup.py patch_version` script. For this we have code for generating this version information from git duplicated both inside Picard (for macos and win builds) and outside (in snap and daily appimage builds).

## Solution

Add `scripts/tools/patch-version.sh` to encapsulate this version update. Use this in both the macOS and Windows CI builds. Also update the PyPI builds to also make use of patch-version.

Once merged the same script can be used in both the Snapcraft and AppImage builds to replace https://github.com/metabrainz/picard-snap/blob/5acf858b64e42f6c1359de1d27ec29469d338d11/snap/snapcraft.yaml#L94-L103 and https://github.com/phw/picard-daily/blob/1cef008d81c875e4a10242be5451c834329b83bd/patch-version.sh

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/.github/workflows/package-windows.yml
+++ b/.github/workflows/package-windows.yml
@@ -58,12 +58,7 @@ jobs:
     - name: Install dependencies
       run: uv sync --group build --group dev --extra plugins
     - name: Patch build version
-      if: startsWith(github.ref, 'refs/tags/') != true
-      run: |
-        $ReleaseTag = $(git describe --match "release-*" --abbrev=0 --always HEAD)
-        $BuildNumber = $(git rev-list --count "$ReleaseTag..HEAD")
-        Add-Content $Env:GITHUB_ENV "BUILD_NUMBER=$BuildNumber"
-        uv run python setup.py patch_version --platform=$BuildNumber.$(git rev-parse --short HEAD)
+      run: uv run bash -c "./scripts/tools/patch-version.sh"
     - name: Run tests
       timeout-minutes: 30
       run: uv run pytest --verbose
@@ -73,12 +68,14 @@ jobs:
         uv run python setup.py clean
     - name: Build TaskbarLib.tlb
       run: |
-          call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
-          midl TaskbarLib.idl /tlb TaskbarLib.tlb
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        midl TaskbarLib.idl /tlb TaskbarLib.tlb
       shell: cmd
     - name: Build
       run: |
-        uv run python setup.py build --build-number=$Env:BUILD_NUMBER --disable-locales
+        $ReleaseTag = $(git describe --match "release-*" --abbrev=0 --always HEAD)
+        $BuildNumber = $(git rev-list --count "$ReleaseTag..HEAD")
+        uv run python setup.py build --build-number=$BuildNumber --disable-locales
         uv run python setup.py build_locales
         uv run python setup.py build_ext -i
         uv run pyinstaller --noconfirm --clean picard.spec
--- a/scripts/tools/patch-version.sh
+++ b/scripts/tools/patch-version.sh
@@ -1,0 +1,17 @@
+#!/bin/bash
+set -e
+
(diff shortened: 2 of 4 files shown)


.github/workflows/package-windows.yml
```diff
@@ -73,12 +69,14 @@ jobs:
...
-          midl TaskbarLib.idl /tlb TaskbarLib.tlb
+        call "C:\Program Files\Microsoft Visual Studio\2022\Enterprise\VC\Auxiliary\Build\vcvars64.bat"
+        midl TaskbarLib.idl /tlb TaskbarLib.tlb
       shell: cmd
     - name: Build
       run: |
-        uv run python setup.py build --build-number=$Env:BUILD_NUMBER --disable-locales
+        $ReleaseTag = $(git describe --match "release-*" --abbrev=0 --always HEAD)
+        $BuildNumber = $(git rev-list --count "$ReleaseTag..HEAD")
+        uv run python setup.py build --build-number=$Env:BuildNumber --disable-locales
```
zas: Not sure, but shouldn't it be:

```suggestion
        uv run python setup.py build --build-number=$BuildNumber --disable-locales
```

Since `$BuildNumber` is now a local variable.

.github/workflows/package-windows.yml
```diff
@@ -58,12 +58,8 @@ jobs:
...
       run: uv sync --group build --group dev --extra plugins
     - name: Patch build version
-      if: startsWith(github.ref, 'refs/tags/') != true
-      run: |
-        $ReleaseTag = $(git describe --match "release-*" --abbrev=0 --always HEAD)
-        $BuildNumber = $(git rev-list --count "$ReleaseTag..HEAD")
-        Add-Content $Env:GITHUB_ENV "BUILD_NUMBER=$BuildNumber"
-        uv run python setup.py patch_version --platform=$BuildNumber.$(git rev-parse --short HEAD)
+      run: uv run bash -c "./scripts/tools/patch-version.sh"
+      shell: bash
```
zas: Since `shell` is set to `bash` , do we still need `uv run bash -c`?

scripts/tools/patch-version.sh
```diff
@@ -0,0 +1,17 @@
...
+cd "$(dirname "$0")/../../" || exit
+
+RELEASE_TAG=$(git describe --match "release-*" --abbrev=0 --always HEAD)
+CHANGE_COUNT=$(git rev-list --count "$RELEASE_TAG..HEAD")
+COMMIT_HASH=$(git rev-parse --short HEAD)
+
+if [ "$CHANGE_COUNT" -ne 0 ]; then
+    VERSION_EXTRA="$CHANGE_COUNT-$COMMIT_HASH"
+    echo "Patching version: $VERSION_EXTRA"
+    python setup.py patch_version --platform="$VERSION_EXTRA"
```
zas, line 13: Wouldn't be simpler/safer to use `uv run python`  in this script?

.github/workflows/package-windows.yml
phw: Yes, because it is run with uv, and uv on Windows doesn't know about shebangs (or rather Windows doesn't). Eventually we don't need to set `shell: bash` then.

scripts/tools/patch-version.sh
phw: That limits it then to environments where we have uv (easily) available. Which e.g. is not the case for the appimage build.

.github/workflows/package-windows.yml
phw: Good catch. Fixed in the recent push

.github/workflows/package-windows.yml
phw: I removed `shell: bash`.

zas (approved):
LGTM

phw:
Uses in:

snap: https://github.com/metabrainz/picard-snap/commit/299608e6547145c29970f0bd26aadb713cf906c4
daily: https://github.com/phw/picard-daily/commit/3ad2359eac6510ce607f069376d2a31f8f119fbd

PR #3144, opened 2026-05-01
New plugin dir template, --init interactive and non-interacting CLI

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3262

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/.github/workflows/run-tests.yml
+++ b/.github/workflows/run-tests.yml
@@ -69,7 +69,7 @@ jobs:
         os: [ubuntu-latest]
         python-version: ['3.10']
         dependencies: [
-          "PyQt6==6.5.3 PyQt6-Qt6==6.5.3 mutagen==1.45 PyYAML==5.1 tomlkit==0.13.3 pygit2==1.18.2",  # minimal versions, minimum dependencies
+          "PyQt6==6.5.3 PyQt6-Qt6==6.5.3 mutagen==1.45 PyYAML==5.1 tomlkit==0.13.3",  # minimal versions, minimum dependencies
           "PyQt6>=6.5.3 PyQt6-Qt6>=6.5.3,!=6.10.* mutagen~=1.45 PyYAML~=6.0 discid==1.0 tomlkit==0.13.3 pygit2==1.18.2",
           "PyQt6>=6.5.3 PyQt6-Qt6>=6.5.3,!=6.10.* mutagen~=1.45 PyYAML~=6.0 python-libdiscid tomlkit==0.13.3 pygit2==1.18.2",
           "PyQt6>=6.5.3 PyQt6-Qt6>=6.5.3,!=6.10.* mutagen~=1.45 PyYAML~=6.0 charset-normalizer==2.0.6 tomlkit==0.13.3 pygit2==1.18.2",
--- a/docs/PLUGINSV3/CLI.md
+++ b/docs/PLUGINSV3/CLI.md
@@ -1114,9 +1114,111 @@ categories = ["metadata"]
 
 
 ### Update Workflow
+
 ```bash
 picard-plugins --check-updates
 picard-plugins --update-all
+```
+
+### Create a New Plugin
+
+**Command:** `picard-plugins --init [NAME]`
+
+**Description:** Create a new plugin project with all required files and a git
+repository.
+
+**Use cases:**
+
+- Scaffold a new plugin project interactively (no argument)
+- Create a plugin project non-interactively with a name
+
+**Generated files:**
+
+- `MANIFEST.toml` — Plugin metadata with a generated UUID
+- `__init__.py` — Plugin code with `enable()`/`disable()` stubs
+- `README.md` — Basic readme with plugin name
+- `.gitignore` — Python-specific ignore rules
+
+With `--with-translations`, the following are also generated or modified:
+
+- `MANIFEST.toml` includes `source_locale` and commented `[name_i18n]`/`[description_i18n]` sections
+- `__init__.py` includes `t_` and `api.tr()` usage examples
+- `locale/en.toml` — Source locale translation file
+
+A git repository is always initialized. In interactive mode, you are asked
+whether to create an initial commit. In non-interactive mode, the initial
+commit is created automatically.
+
+**Interactive mode prompts:**
+
+When run without a name (`picard-plugins --init`), the following prompts
+appear in order:
+
+1. Plugin name (required)
+2. Destination directory — shows computed default, can be overridden
+3. Author name — defaults to last used value or git config
+4. Author email — shown only if author is provided
+5. Short description
+6. Categories — comma-separated numbers for multiple (e.g. `1,3`)
+7. License — defaults to last used value or GPL-2.0-or-later
+8. Create initial git commit? — yes/no (default: yes)
+
+Author name, email, and license are persisted across runs.
+
+**Options:**
+
+| Flag | Description |
+|------|-------------|
+| `--target-dir NAME` | Override project directory name (relative to `--parent-dir`, default: `picard-plugin-<slug>`) |
+| `--parent-dir DIR` | Parent directory where the project is created (default: current directory) |
+| `--author NAME` | Author name for MANIFEST.toml |
+| `--category CATEGORY` | Plugin category (metadata, coverart, ui, etc.) |
+| `--with-translations` | Include translation support (locale files and examples) |
+| `--no-commit` | Skip initial git commit (works in both interactive and non-interactive modes) |
+
+**Examples:**
+
+```bash
+# Interactive mode - prompts for all fields
+picard-plugins --init
+
+# Non-interactive with just a name
+picard-plugins --init "My Cool Plugin"
+
+# Non-interactive with all options
+picard-plugins --init "My Cool Plugin" --author "Jane Doe" --category metadata
+
+# Create in a specific parent directory
+picard-plugins --init "My Cool Plugin" --parent-dir ~/dev
+
+# Override the directory name
+picard-plugins --init "My Cool Plugin" --target-dir my-plugin
+
+# Both: custom name inside custom parent (creates ~/dev/my-plugin)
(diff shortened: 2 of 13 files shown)


picard/plugin3/init_templates.py
```diff
@@ -0,0 +1,316 @@
...
+
+from picard.plugin3.api import PluginApi
+
+
+def enable(api: PluginApi) -> None:
+    """Called when the plugin is enabled.
+
+    Use api to register plugin hooks and access essential Picard APIs.
+    """
+
```
rdswift, line 188: Perhaps include a similar logging action (without translation) similar to the `with_i18n` example so that the plugin is actually doing something.

picard/plugin3/init_templates.py
```diff
@@ -0,0 +1,316 @@
...
+    value = value.replace('\n', '\\n')
+    value = value.replace('\f', '\\f')
+    value = value.replace('\r', '\\r')
+    return value
+
+
+def generate_manifest(
+    name, description='', authors=None, categories=None, license_id='', license_url='', with_i18n=False
+):
+    """Generate a filled-in MANIFEST.toml for a new plugin.
```
rdswift, line 83: If you have an email address, you might want to add something into the MANIFEST like:

```toml
# report_bugs_to = "https://url.report.to/issues"
report_bugs_to = "mailto:name@address.com"
```

rdswift (commented):
This looks good, and seems to work well.  I'm going to use some of the code and templates for my "Plugin Framework Creator" plugin.

picard/plugin3/init_templates.py
```diff
@@ -0,0 +1,316 @@
...
+    """
+    # Translate a string at runtime
+    greeting = api.tr(GREETING, "Hello from the plugin!")
+    api.logger.info(greeting)
+
+
+def disable() -> None:
+    """Called when the plugin is disabled."""
+'''
+    return '''"""Basic Picard 3 plugin."""
```
phw, line 177: I would recommend to put an `else` here. The long template strings make this a bit harder to follow and an explicit else helps to understand that these are two alternative templates.

Alternatively put each template into a separate function (`generate_plugin_init_py_with_i18n`, `generate_plugin_init_py_no_i18n`)

picard/plugin3/init_templates.py
```diff
@@ -0,0 +1,316 @@
...
+    """Generate a .gitignore for a new plugin.
+
+    Returns:
+        str: .gitignore content
+    """
+    return '''__pycache__/
+*.py[cod]
+*.egg-info/
+dist/
+'''
```
phw, line 254: I currently use this for all my plugins: 

```
# Byte-compiled / optimized / DLL files
*.py[cod]
__pycache__/

# Environments and development tools
.venv/
.ruff_cache/
```

test/plugins3/test_cli.py
```diff
@@ -1178,3 +1192,401 @@ def test_color_enabled_by_default(self):
...
+        dirty = {f for f, s in status.items() if s != GitStatusFlag.IGNORED}
+        self.assertEqual(dirty, set(), f'Uncommitted files: {dirty}')
+        repo.free()
+
+    def test_init_commit_uses_provided_author(self):
+        """--init --author uses the provided name for the git commit."""
+        target = self.tmpdir / 'test-plugin'
+        exit_code, stdout, stderr = run_cli(MockPluginManager(), init='Test', output=str(target), author='Jane Doe')
+        self.assertEqual(ExitCode.SUCCESS, exit_code)
+        repo = pygit2.Repository(str(target))
```
phw: This gets run even if `pygit2` is unavailable. I think the test should be marked with `@unittest.skipUnless(pygit2, "requires pygit2")`

phw (commented):
This is a great addition to help getting unified plugins. I have a personal bash script `init-picard-plugin` I used for all of my plugins with a similar setup, but a bite more opinionated. It additionally generates:

A `.pre-commit-config.yaml`:

```yaml
default_install_hook_types: [pre-commit, pre-push]
repos:
  - repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v6.0.0
    hooks:
      - id: check-toml
      - id: check-yaml
      - id: trailing-whitespace
      - id: end-of-file-fixer
      - id: mixed-line-ending
  - repo: https://github.com/astral-sh/ruff-pre-commit
    rev: v0.14.8
    hooks:
      - id: ruff-check
        args: [--fix]
      - id: ruff-format
 ```

A `.ruff.toml`:

```toml
exclude = ["ui_*.py"]
```

It also sets up a venv as `.venv` and installs pre-commit and picard (for type checking and auto-completion). Not sure how useful this is as a general thing, but I think it encourages good defaults.

This setup is also e.g. in https://github.com/metabrainz/picard-plugin-replaygain2

picard/plugin3/init_templates.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3144/commits/332ec074db9b118959a4f6f662c4326b44cc8ca7

picard/plugin3/init_templates.py
zas: Done in https://github.com/metabrainz/picard/pull/3144/commits/ff91ec1e91509e6e286c59ef53b7f94a87bbad10

picard/plugin3/init_templates.py
zas: Done in https://github.com/metabrainz/picard/pull/3144/commits/ff91ec1e91509e6e286c59ef53b7f94a87bbad10

picard/plugin3/init_templates.py
zas: Done in https://github.com/metabrainz/picard/pull/3144/commits/d5b9f326e3cfaba429042a745843e5e4e10e1e24

test/plugins3/test_cli.py
zas: Actually I implemented the missing backend (a bit more complex, but cleaner). Done in https://github.com/metabrainz/picard/pull/3144/commits/8682d0add89679afe1dbd0805d2b8349529ebe8a and few following commits.

.github/workflows/run-tests.yml
```diff
@@ -62,6 +62,32 @@ jobs:
         name: coverage-${{ runner.os }}-${{ steps.setup-python.outputs.python-version }}
         path: htmlcov/
 
+  test-no-pygit2:
```
phw: I'd instead add a line without pygit2 to the test-requirements test cases. With pygit2 now actually being optional the best solution would be probably to just drop it from the first test case (`# minimal versions, minimum dependencies`).

.github/workflows/run-tests.yml
phw: Drop pygit2 here to test without it

phw (changes requested):
This is looking good to me, definitely a very helpful addition. Suggested some changes to the CI workflow.

.github/workflows/run-tests.yml
zas: Done in https://github.com/metabrainz/picard/pull/3144/commits/69b7dc1e0ab639ad973095032b7c3b4e0a1acbf9

phw (dismissed):
Looks good to me

docs/PLUGINSV3/CLI.md
```diff
@@ -1114,11 +1114,113 @@ categories = ["metadata"]
...
+
+- Scaffold a new plugin project interactively (no argument)
+- Create a plugin project non-interactively with a name
+
+**Generated files:**
+
+- `MANIFEST.toml` — Plugin metadata with a generated UUID
+- `__init__.py` — Plugin code with `enable()`/`disable()` stubs
+- `README.md` — Basic readme with plugin name
+- `.gitignore` — Python-specific ignore rules
```
rdswift, line 1140: Should this list include the pre-commit and ruff files recently added?

picard/plugin3/init_templates.py
rdswift: I'm not suggesting that you include this, but with all of my plugins I include a symlink to the `picard` directory in my local clone of the Picard source code repo.  This way, as I do my development work on the plugin all of the includes are valid and available for inspection and auto-completion.  As a result, I also include "picard" in my `.gitignore` file (along with ".project" which is used by my GUI git manager).

rdswift (dismissed):
Looks good.  Nice addition.  Thanks.

docs/PLUGINSV3/CLI.md
zas: Actually I still ponder whether we should add those to the template directory, as that's very specific to dev env rather than plugin code.

rdswift: Good point.  I never thought of it that way.

phw: Yes, no strong opinion from my side on that. It's my setup, and I think the setup we should have in official plugins. But it also needs a bit to be usable (use of a venv with pre-commit and picard installed, actually setting up the pre-commit hook).

I'd be fine with omitting both files from the plugin template.

picard/plugin3/init_templates.py
phw: That's also an interesting approach. I do something similar in Picard, where I symlink to the latest mutagen development version.

But for plugins I take a slightly different approach: I create a venv inside the plugin directory:

```
python -m venv .venv`
```

Then I install Picard there. With the current beta releases this can be installed from pypi with `pip install picard` inside the venv. I actually often install from my git checkout of Picard directly with something like `pip install ../picard/`.

docs/PLUGINSV3/CLI.md
zas: I removed them, I want to keep the --init output minimal and not tied to dev env.
I also renamed few options and rebased the PR (as code was added/removed).

phw (approved):

PR #3147, opened 2026-05-06
PICARD-3261: better non unicode encoding of tags

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3261

When storing tags with non-Unicode encoding, e.g. ISO-8859-1/latin1 in ID3 or Windows-1252 in WAV, any character not supported by the encoding gets replaced with a "?" by using `string.encode(..., errors="replace")`. No attempt is made to use any more readable character.

E.g. the band name "Ēbn-Ōzn" gets stored as "?bn-?zn", since both Ē and Ō cannot be encoded in neither ISO-8859-1 nor Windows-1252. But it would be more helpful to store it as "Ebn-Ozn".

We do have extensive functionality to convert Unicode to a more simplified format. But functions like `replace_non_ascii` assume ASCII compatibility, and will even replace characters that would be encodable in the target encoding.

## Solution

Register a "simplify" encoding error handler. This error handler attempts to first simplify the Unicode characters, it falls back to encoding with the "replace" error handler.

E.g.

```python
"Ēbn-Ōzn".encode("latin1", errors="replace")  # This would output "?bn-?zn"
"Ēbn-Ōzn".encode("latin1", errors="simplify")  # This would output "Ebn-Ozn"
```

By having this an error handler:

- It is universally usable in every call of `encode`
- Encoding aware: Any character already encodable in the target encoding does not trigger the error handler
- A simplified character is used, if possible. If not the use of the "replace" error handler when encoding the replacement ensures we fall back to the default replacement behavior

Use `errors="simplify"` when encoding WAVE and ID3 tags.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -112,7 +112,7 @@ def id3text(text, encoding):
     """
 
     if encoding == Id3Encoding.LATIN1:
-        return text.encode('latin1', 'replace').decode('latin1')
+        return text.encode('latin1', errors='simplify').decode('latin1')
     return text
 
 
--- a/picard/formats/wav.py
+++ b/picard/formats/wav.py
@@ -144,7 +144,7 @@ class RiffListInfo(MutableMapping):
         return value.rstrip('\0')
 
     def __encode_data(self, value):
-        return value.encode(self.encoding, errors='replace') + b'\x00'
+        return value.encode(self.encoding, errors='simplify') + b'\x00'
 
     def __contains__(self, name):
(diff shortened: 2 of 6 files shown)


zas (approved):
Very clean solution.

LGTM

PR #3150, opened 2026-05-08
PICARD-3264: Improved alias selection for artist, album and track translations and refactoring

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3264

When enabling artist / album / track translations the locale selection sometimes gives unexpected results.

- For album and track titles a non-primary alias is used instead of the primary one for the locale
- If there is no exact match for the preferred root locale, but a full locale would be available, a secondary preferred locale might be used.

Also the alias translation code was inconsistent and partially duplicated between artist locales and album / track title translation.

## Solution

- Refactor mbjson to reuse the same logic for both artist and album/track title translation
- Rename the option `artist_locales` to `translation_locales`, since it is no longer limited to artist translation
- Filter for primary aliases also for album / track title
- Use matching full locale if a root locale is preferred. E.g. if the user prefers "en" over "de", return "en_US" even if there is an exact match for a "de" locale.

There is still a fallback if the user prefers a full locale (e.g. "en_US") it will also check for "en". But only after all explicitly preferred locales have been tried.

## AI Usage

* [ ] No AI/LLM use
* [x] Minimal use (e.g. autocompletion): only trivial autocompletions when updating recurring test code
* [ ] Primarily AI developed

## Action

Additional actions required:


rdswift (dismissed):
I got lost in the logic a couple of times because I've never looked closely at the matching stuff before, but after going through it again, this looks good to me.

phw:
Yes, it also took me a while to understand this.  Also untangling the somewhat duplicated code for the album/track title aliases. I think it showed how that part of the code was at least partly LLM generated and instead of reusing and adapting the already existent functionality it just added separate implementations for extracting the translation from aliases.

phw:
If you are looking at a test case have a look at https://musicbrainz.org/release/13e91f6b-2456-4567-a3b2-e2e22ac3cbdd . It has Chinese and English aliases for all of artist, album title and recording titles

picard/mbjson.py
```diff
@@ -295,61 +310,27 @@ def check_higher_score(locale_dict, locale, score):
...
-def _build_alias_locale_maps(aliases: list[dict[str, Any]]) -> tuple[dict[str, str], dict[str, str]]:
-    """
-    Build maps of full and root locales from alias entries.
-
-    Parameters
-    ----------
-    aliases : list[dict[str, Any]]
-        Alias dictionaries that can include ``name`` and ``locale`` keys.
+        if check_higher_score(locales, root_locale, comb):
+            locales[root_locale] = AliasMatch(comb, Alias(alias['name'], alias.get('sort-name', None)))
```
zas: Do we really need `alias.get('sort-name', None)`, old code was using `alias['sort-name']`, it means we'll have a None value instead of an exception (that shouldn't happen), so basically it hides a potential issue (normally no, because MB API returns sort-name for artist, but who knows...)

test/test_mbjson_translations.py
```diff
@@ -64,60 +69,91 @@ def mock_get_config(monkeypatch: pytest.MonkeyPatch, config: Any) -> Callable[[]
...
     ],
 )
-def test_build_alias_locale_maps(
-    aliases: list[dict[str, str]], expected_full: dict[str, str], expected_root: dict[str, str]
-) -> None:
-    full, root = mbjson._build_alias_locale_maps(aliases)
-    assert full == expected_full
-    assert root == expected_root
+def test_build_alias_locale_maps(aliases: list[dict[str, str]], expected: dict[str, mbjson.AliasMatch]) -> None:
+    aliase_matches = mbjson._locales_from_aliases(aliases)
```
zas: Typo, it should be `alias_matches`, right?

picard/mbjson.py
```diff
@@ -583,18 +553,18 @@ def artist_credit_from_node(node):
         if artist and 'id' in artist and artist['id']:
             # Add artist's country code if specified, otherwise 'XX' (Unknown Country)
             artist_countries.append(artist['country'] if 'country' in artist and artist['country'] else 'XX')
-        translated_name, sort_name = _translate_artist_node(artist, config=config)
-        has_translation = translated_name != artist['name']
+        translation_alias = _translate_artist_node(artist, config=config)
```
zas: `translation_alias` -> `translated_alias` (it seems more correct to me, but perhaps @rdswift could tell)

picard/mbjson.py
```diff
@@ -540,34 +520,24 @@ def _should_skip_translation_due_to_scripts(text: str | None, config: Any | None
...
-            for locale in config.setting['artist_locales']:
-                if locale in full_locales:
-                    return full_locales[locale][1]
-
-            # Second pass to match root locale if available
-            for locale in config.setting['artist_locales']:
-                lang = locale.split('_')[0]
-                if lang in root_locales:
-                    return root_locales[lang][1]
+            alias = _find_localized_alias_name(node['aliases'], config.setting['translation_locales'])
```
zas: Perhaps stick to `translated_alias` ?

zas (changes requested):
Overall, it looks good to me, code is cleaner. Just few comments about naming.

picard/mbjson.py
phw: I actually changed this to treat this optionally, as this avoided that I had to add the sort-name to all test cases. But I changed it back now to do `alias['sort-name']` and extended the tests in 7e320c1b00c15af595617214980d0f81dfd723da . MB server will actually always return the field, and even have it always be a non-empty string (it gets filled with the alias name, if not sort-name is given explicitly, and that even for cases like search hints).

Having said this, should, for whatever reason, the sort-name be missing, we'll get a bug report. And as I see it the client side fix would be to use  `alias.get('sort-name', '')` or `alias.get('sort-name', None)`

test/test_mbjson_translations.py
phw: Good catch, fixed

picard/mbjson.py
phw: I think I use `translated_alias` now consistently

picard/mbjson.py
phw: I think I use translated_alias now consistently

picard/mbjson.py
zas: I guess that should be handled in mbjson (we can default to empty string, and log an error)

picard/mbjson.py
```diff
@@ -119,6 +121,18 @@
 _BLANK_SPECIAL_RELTYPES = {'vocal': 'vocals'}
 
 
+@dataclass
+class Alias:
```
zas, line 125: Minor but perhaps add a doctstring explaining what's the purpose of those 2 classes

picard/mbjson.py
```diff
@@ -148,12 +162,12 @@ def _relation_attributes(relation):
...
     artist = relation['artist']
-    translated_name, sort_name = _translate_artist_node(artist, config=context.config)
-    has_translation = translated_name != artist['name']
+    translated_alias = _translate_artist_node(artist, config=context.config)
+    has_translation = translated_alias.name != artist['name']
     if not has_translation and context.use_credited_as and 'target-credit' in relation:
         credited_as = relation['target-credit']
         if credited_as:
-            translated_name = credited_as
+            translated_alias.name = credited_as
```
zas, line 174: This modifies the object directly, perhaps we should instantiate a new Alias() instead?
We could even prevent such change using `@dataclass(frozen=True)`

Not critical though, but perhaps cleaner, dunno.

zas (dismissed):
It looks good to me, just few comments, but those aren't blocking the merge.

picard/mbjson.py
phw: I don't think we need to have this class immutable. The only reason it exists is really that it gives more structure then just passing around a tuple.

For this specific case I'd rather introduce a `artist_name` variable again. But I don't think it helps much in regards of code readability or such. Rather the opposite.

PR #3153, opened 2026-05-10
PICARD-2867: barcode matching

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

## Problem

Barcode isn't used in matching, so Picard fails to differentiate releases based on it (for files having a barcode).

* JIRA ticket (_optional_): PICARD-2867

## Solution

Add barcode to weighted linear comparison.
A weight of 6 was chosen to be enough to differentiate without being more important than other criteria.

## AI Usage

In accordance with the AI use policy portion of the [MetaBrainz Contribution Guidelines](https://github.com/metabrainz/guidelines/blob/master/README.md), the level of AI/LLM use in the development of this Pull Request is:

* [ ] No AI/LLM use
* [ ] Minimal use (e.g. autocompletion)
* [ ] Moderate use (e.g. suggestions regarding code fragments)
* [x] Significant use (e.g. code structure, tests development, etc.)
* [ ] Primarily AI developed
* [ ] Other (please specify below)

## Action

Additional actions required:
* [ ] Update Picard [documentation](https://github.com/metabrainz/picard-docs) (please include a reference to this PR)
* [ ] Other (please specify below)

PR #3154, opened 2026-05-10
Scripts and corpus to evaluate cluster and file matching

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

We have no tools to evaluate changes in matching functions (and associated weights).

Typical case is something like https://github.com/metabrainz/picard/pull/3153 
We are rather blind when it comes to such changes, because we don't have any reproducible mean to evaluate the code changes.

## Solution

Provide a stable corpus and scripts to be able to evaluate changes.
See `scripts/eval_matching/README.md` for details.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/scripts/eval_matching/README.md
+++ b/scripts/eval_matching/README.md
@@ -9,8 +9,16 @@ from a set of confusable candidates when file metadata is degraded.
 # Run full evaluation (all configs, cluster + file level)
 python scripts/eval_matching/eval_matching.py
 
-# Focus on a specific problem
+# Focus on a specific problem with diagnostics
 python scripts/eval_matching/eval_matching.py -v -s non_latin --cluster-only -p neutral
+
+# Filter by degradation pattern
+python scripts/eval_matching/eval_matching.py -d combo --cluster-only -p neutral
+
+# Save baseline, make changes, compare
+python scripts/eval_matching/eval_matching.py --cluster-only -p neutral --save baseline.json
+# ... edit matching code ...
+python scripts/eval_matching/eval_matching.py --cluster-only -p neutral --compare baseline.json
 
 # Refresh corpus from MusicBrainz (after API changes)
 python scripts/eval_matching/refresh_corpus.py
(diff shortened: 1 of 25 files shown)


phw:
I think this will turn out to be very useful. I also like the general approach to this with starting with the full data and degrading it. I think with the test cases, scenarios and config settings this is a very extensible setup.

But as already mentioned in our chat, there are some cases that won't perform well, especially the ones relying on preferred releases settings. Test cases for this will very often only succeed with the proper settings, and expectedly fail else. One example in the current corpus is the "non_latin_editions" scenario currently always reporting 0%. It's actually not an issue with non-latin matching in general, but with the specific test cases.

There are only two test cases, and they match two releases that are identical except for the media format (digital vs CD):

```python
    # --- Non-Latin scripts: same album, different editions ---
    {
        "target": "eval_release_3ac4a81e.json",  # 椎名林檎 三毒史 digital
        "distractors": [
            "eval_release_4fdf1514.json",  # 椎名林檎 三毒史 CD
        ],
        "scenario": "non_latin_editions",
    },
    {
        "target": "eval_release_4fdf1514.json",  # 椎名林檎 三毒史 CD
        "distractors": [
            "eval_release_3ac4a81e.json",  # 椎名林檎 三毒史 digital
        ],
        "scenario": "non_latin_editions",
    },
```

Without any preferred releases settings these tests both will be always ambiguous, as both releases match the same. This is actually expected.

With a config preset that prefers digital format that first test will be "correct", while the second one is reported as "wrong". Both results are actually expected and correct. With a  config preset that prefers CD format the result will be exactly reversed, also as expected.

That means such a combination of test cases will always report as 0% correct, no matter the config settings.

Not sure how to improve this. Maybe we need to extend the test cases with an expected outcome based on config? E.g. we could, optionally, define expectations for a test case, such as:

 ```python
{
    "target": "eval_release_3ac4a81e.json",  # 椎名林檎 三毒史 digital
    "distractors": [
        "eval_release_4fdf1514.json",  # 椎名林檎 三毒史 CD
    ],
    "scenario": "non_latin_editions",
    "expectations": {
        "prefer_jp_digital": "correct"
        "prefer_us_cd": "wrong",
        "neutral": "ambiguous",
        "compilations_low": "ambiguous",
        "prefer_eu_vinyl": "ambiguous",
    }
}
```

Or maybe with a `"*": "ambiguous"` wildcard match.

If not specified the default expectation would of course be "correct".

zas:
@phw I implemented your suggestion about expectations in https://github.com/metabrainz/picard/pull/3154/commits/fa1c09bbd13678f71999d1e1de61c45651db6933
Tell me if it works for you.

phw:
@zas Yes, this looks good and works as expected. This was just a quick idea yesterday that I got while writing down my thoughts. But after thinking about this more and seeing it in action I really like it.

It can mean for some tests that we need to do some fine tuning, also if we add more config presets. But it really allows us to specify expected results and get more reliable indicators if some changes break our matching.

phw (approved):

PR #3155, opened 2026-05-11
PICARD-3163: Make translation sortname fallback optional, disable by default

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3163

Historically Picard was using the sort name as a simple way to "translate" artists names. Initially this functionality was mostly meant as a way to allow users to have artist in non-Latin scripts shown in Latin script, since the sort name usually contains the name in Latin script.

We kept this behavior as a fallback to the alias based translations. This has caused several confusions, see e.g. PICARD-758, PICARD-831 and partially also PICARD-1768.

## Solution

Disable the sort name fallback by default. Since this can break some users tags that relied on this behavior, add a `translate_from_sortname` option to re-enable this behavior.

As part of this also restructure Options > Metadata to group all translation related functions into a separate groupbox:

grafik

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/mbjson.py
```diff
@@ -599,10 +598,13 @@ def _translate_artist_node(node: Node, config: Config | None = None) -> Alias:
...
         sort_name = node['sort-name']
-        translated_name = translate_from_sortname(node['name'] or '', sort_name)
+        if config.setting['translate_from_sortname']:
+            translated_name = translate_from_sortname(node['name'] or '', sort_name)
+        else:
+            translated_name = node['name']
     else:
         translated_name, sort_name = node['name'], node['sort-name']
-    return Alias(translated_name, sort_name)
+    return Alias(translated_name or '', sort_name or '')
```
phw, line 607: To anticipate the question why the empty string fallback was added: We have a test case `recording_null.json`, that explicitly tests the behavior when all fields are `null`. However, the test case for this only did test with translations enabled, and that always used the sort name fallback. And that code path already did `node['name'] or ''`.

Now this test would fail by default. `Alias` currently is supposed to hold only strings, which makes things better to handle further down. So I think it is best to handle the null case here.

Also the corresponding test was extended and we now test for both with and without translation.

rdswift (approved):
Looks good to me.  The only thing is to remember to update the documentation to include the new option setting.

zas (approved):
LGTM, much cleaner :)

PR #3165, opened 2026-05-17
PICARD-3278: fix load all image types

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3278

When configuring Picard to not only save a single front image but download all cover art images from CAA, the first non-front image loaded will replace the already loaded front image.

For example:

1. Enable cover art saving to tags or files, but disable to embed / save only a single front image
2. Enable the CAA provider as primary image source, but disable to load only selected images (or select a wide range of image types to load)
3. Load e.g. https://musicbrainz.org/release/cd16cf3b-86b0-4b6f-bd7e-73e8b2451a83/cover-art

Current result will be that the front image shows up after getting loaded, but gets removed again as soon as any other image got loaded.

## Solution

The reason is that the `CoverArtSetter` gets run in REPLACE mode, which triggers `strip_front_images()`, even if the replacing image was not a front image.

Fix by only performing the replace if the replacing image is a front image. Added a test.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/coverart/setters/__init__.py
+++ b/picard/coverart/setters/__init__.py
@@ -91,7 +91,7 @@ class CoverArtSetter:
         obj
             The object to set the image on
         """
-        if self.mode == CoverArtSetterMode.REPLACE:
+        if self.mode == CoverArtSetterMode.REPLACE and self.coverartimage.is_front_image():
             obj.metadata.images.strip_front_images()
             log.debug("Replacing images with %r in %r", self.coverartimage, obj)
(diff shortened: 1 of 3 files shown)


zas (approved):
LGTM

PR #3166, opened 2026-05-18
PICARD-3284: Improve release matching with tiered scoring and identifier support

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

Picard's release matching algorithm treats all signals equally in a flat weighted average. This means exact identifiers like barcodes (weight 6 out of ~52 total) only contribute ~11% to the final score — not enough to be decisive.
As a result:
  
- Two releases with different barcodes can still tie if their titles and artists match
- Catalog number + label information is not used for matching at all
- ISRC is not used for track-level matching
- The totalalbumtracks comparison silently fails for most releases (the track-count key is absent from MB API responses at the release level)
- The trackcount_score function is binary (0.0 or 0.3 or 1.0), so a 1-track difference scores the same as a 10-track difference
- Cluster lookup only sends title/artist/tracks to the search API, missing barcode/catno/date that could narrow results to the correct edition
  
The eval harness shows 72.4% pass rate (239/330) on the neutral profile.

* JIRA ticket (_optional_): PICARD-3284

### 3-tier matching architecture

Reorganize matching into 3 tiers with a structured `ReleaseMatchParts` dataclass and nested `TieredWeights` dict:

**Tier 1 — Exact identifiers** (high weight, decisive when present):
  
- Barcode: weight 6 → 28. A matching barcode now dominates the score.
- Catalog number + label (new, weight 22): matches file's catalognumber/label tags against the release's label-info.
- ISRC (new, weight 28, track-level only): matches file's ISRC against recording ISRCs.
  
**Tier 2 — Fuzzy similarity** (moderate weight, core matching):
  
- Album title, artist, track count, date — unchanged weights.
  
**Tier 3 — Preferences/discriminators** (low weight, tie-breaking):
  
- Country, format, release type — unchanged weights.

### Bug fixes
  
- `totalalbumtracks`: compute from media track-counts when release-level `track-count` is absent
- `trackcount_score`: gradual ratio-based scoring instead of binary thresholds
- Barcode "no release barcode" case: 0.5 → 0.0 (file clearly came from a barcoded release)
- Fixed early return in `compare_to_release_parts` when no Qt app (was skipping similarity/preferences)

### Cluster lookup improvement

Include barcode, catalog number, and date in the cluster lookup search query (when available in file metadata). This narrows MB search results to the correct edition instead of relying solely on title/artist/track count.

### Code quality & performance

- Extracted `_date_score` method to reduce nesting (5 levels → early returns)
- Merged duplicate barcode branches
- Reordered tiers: identifiers first (cheap, decisive), similarity second, preferences last
- `itertools.chain` instead of list concatenation in the release loop
- `assert` instead of `if/raise ValueError` in `linear_combination_of_weights` (stripped with `-O`)
- Pass config to `compare_to_release_parts` instead of calling `get_config()` per release
- Normalize file catno/label once before comparison loop
- Unified `compare_to_track` to use `ReleaseMatchParts` for track-level signals
- Cache `track_parts.all_parts` before release loop
- `_get_total_release_weight` uses `_TRACK_SIMILARITY_KEYS` frozenset instead of hardcoded key list

### Eval corpus extended

- Added `labels` and `isrcs` to MB API includes in `refresh_corpus.py`
- 8 new scenarios (31 releases total): remaster vs original, same-title-different-artist, multi-disc vs single-disc, self-titled albums
- Added `wrong_isrc` degradation (shifted ISRC from adjacent track)
  
### Results

Cluster-level (neutral profile):
- Before: 239/330 pass (72.4%)
- After: 574/594 pass (96.6%) — 12 scenarios, all profiles ≥96.3%

File-level (neutral profile):
- 1232/1242 pass (99.2%)

0 regressions. Remaining failures are all genuinely ambiguous (same barcode, stripped identifiers) or wrong_isrc on very similar tracks.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

## Action

Additional actions required:


picard/cluster.py
```diff
@@ -305,11 +319,17 @@ def lookup_metadata(self):
...
+        if catno:
+            catno = catno.split(MULTI_VALUED_JOINER)[0]
         self._lookup_task = self.tagger.mb_api.find_releases(
             self._lookup_finished,
             artist=self.metadata['albumartist'],
             release=self.metadata['album'],
             tracks=str(len(self.files)),
+            barcode=self.metadata.get('barcode', ''),
+            catno=catno,
+            date=self.metadata.get('date', ''),
```
phw, line 378: Just a thing to think about: the way the queries currently get constructed it does a OR search for all the criteria. The title has a heavy weight server side already, so this heavily influences the results and closely matching titles usually are at the top. 

Just as an example take the below query, which uses a title matching [this RG](https://musicbrainz.org/release-group/2d13c628-282c-4391-82a3-cc5612aa7074), but barcode and year matching a [different one](https://musicbrainz.org/release-group/549cb4dc-bccc-3576-9b04-cba691eb2c59). The title match still scores 100, despite different barcode and year, while the first one matching by barcode and year only scores 32.

https://musicbrainz.org/ws/2/release?limit=100&query=release%3A(Lightning%20at%20the%20Door)%20barcode%3A(5016583115227)%20date%3A(1993)

Anyway, my point is just that we might also want to look into how adding additional criteria influences the search results, because more OR criteria might return a broader range of results (which is not necessarily better or worse)..

I also investigated this part in the past and tried to change the query by using explicit AND. The results where considerable worse, for not already perfectly tagged files, though. The AND approach more often returns no matches, while the OR gives Picard a broader list of potential matches.

picard/metadata.py
```diff
@@ -86,6 +91,68 @@
...
+    preferences: list[tuple[float, int]] = field(default_factory=list)
+
+    @property
+    def all_parts(self) -> list[tuple[float, int]]:
+        """Flat list of all (score, weight) tuples for linear combination."""
+        return self.identifiers + self.similarity + self.preferences
+
+
+# Type for the tiered weights dict structure
+TieredWeights = dict[str, dict[str, int]]
```
phw: Not strictly needed, but:

```
TieredWeights: TypeAlias = dict[str, dict[str, int]]
```

would make the intent clear (or the `type` keyword once we use Python 3.12).

picard/metadata.py
```diff
@@ -261,29 +368,60 @@ def compare(self, other: 'Metadata', ignored: Iterable[str] | None = None):
 
         return linear_combination_of_weights(parts)
 
-    def compare_to_release(self, release: dict, weights: dict[str, int]):
+    def compare_to_release(self, release: dict, weights: 'TieredWeights'):
```
phw: No need for the quotes around `TieredWeights`, the type is declared in the same file.

picard/metadata.py
```diff
@@ -261,29 +368,60 @@ def compare(self, other: 'Metadata', ignored: Iterable[str] | None = None):
...
-        parts = self.compare_to_release_parts(release, weights)
-        sim = linear_combination_of_weights(parts) * get_score(release)
+        config = get_config()
+        parts = self.compare_to_release_parts(release, weights, config)
+        sim = linear_combination_of_weights(parts.all_parts) * get_score(release)
         return SimMatchRelease(similarity=sim, release=release)
 
-    def compare_to_release_parts(self, release: dict, weights: dict[str, int]):
-        parts = []
+    def compare_to_release_parts(self, release: dict, weights: 'TieredWeights', config=None):
```
phw: No need for the quotes around `TieredWeights`, the type is declared in the same file.

phw (commented):
@zas This looks very promising. I only looked over the code so far, haven't run it. But the approach makes sense to me. Adding new criteria and setting the weights always had this difficulty that it was difficult to balance exactly because different values have different roles, that is not fully captured by weight only.

As a side note (but not necessarily as part of this PR): I think we should move all the matching logic out of `metadata.py`, the `Metadata` class should just hold the data and handle updating etc. But the release / recording matching for metadata should be a separate module.

picard/metadata.py
zas: Fixed in https://github.com/metabrainz/picard/pull/3166/commits/b30c7ffa99a639bd3b60d757425a7e8cc3de76b8

picard/metadata.py
zas: Fixed in https://github.com/metabrainz/picard/pull/3166/commits/b30c7ffa99a639bd3b60d757425a7e8cc3de76b8

picard/metadata.py
zas: Fixed in https://github.com/metabrainz/picard/pull/3166/commits/b30c7ffa99a639bd3b60d757425a7e8cc3de76b8

picard/cluster.py
zas: We could have a 2 queries approach, if we have clear identifiers (like a barcode or a catno), do a first narrow query, then a broader one if needed. And keep current behavior if we have no clear identifiers.

rdswift (commented):
The matching has always been a bit of "black magic" to me.  From what I understand from my review, I don't think I have anything to add to what @phw has already noted.

phw:
@zas Overall this works very well in my tests. With the recent changes it also identifies nearly all of the recording lookups. There is one remaining issue, though: The scoring prefers standalone-recordings.

We had this issue before: Because without a release a recording would be compared only for the track weights. But since additional values usually lower the score any recording with releases would score lower. The previous attempt to fix this added an artificial value to the similarity, based on the preference for type "Other" and the total weight of all release weights, see https://github.com/metabrainz/picard/pull/2348

The code is generally still there (https://github.com/zas/picard/blob/cea16f91c8646ad2ea9838455571c325c5ce689f/picard/metadata.py#L578-L583), but due to the now different way the final score gets calculated it is no longer as effective.

I have one example recorded at https://docs.google.com/spreadsheets/d/1E6tronu0qQC2ql3syAkD1ELOmWQSf8ioG3RgRjXTi0Y/edit?usp=sharing with the metadata used and the lookup result. The [MB search](https://musicbrainz.org/ws/2/recording?limit=100&query=track%3A%28Big%20in%20Japan%29%20artist%3A%28Alphaville%29%20release%3A%28First%20Harvest%201984%5C-92%29%20tnum%3A%281%29%20tracks%3A%2815%29%20qdur%3A%28117%29%20isrc%3A%28DEA620101063%3B%20DEA629260030%29) gives both the [standalone recording 190f8ff0-2587-43a9-a121-55c2b4383c52](https://musicbrainz.org/recording/190f8ff0-2587-43a9-a121-55c2b4383c52)  and the [expected match f11939cf-9ad1-45b8-b927-a89b009e50e9](https://musicbrainz.org/recording/f11939cf-9ad1-45b8-b927-a89b009e50e9).

The debug shows:

```
D: 07:52:04,491 log.debug_if:250: match_to_track: file='/home/phw/Musik/Library/Alphaville/First Harvest 1984-92/01 Big in Japan.mp3', 100 candidates, min_sim=0.250, min_margin=0.020
D: 07:52:04,491 file._match_to_track:1094:   #1 sim=0.3453  'Big In Japan' (190f8ff0-2587-43a9-a121-55c2b4383c52)
D: 07:52:04,491 file._match_to_track:1094:   #2 sim=0.2111  'Big in Japan' (f11939cf-9ad1-45b8-b927-a89b009e50e9)
D: 07:52:04,491 file._match_to_track:1094:   #3 sim=0.2030  'Big in Japan (culture mix)' (fdef3080-c741-47b2-8f97-86faf5cab5dd)
D: 07:52:04,491 file._match_to_track:1094:   #4 sim=0.1762  'Big in Japan' (911a0ee2-8256-4793-a551-60db9faf54bc)
D: 07:52:04,491 file._match_to_track:1094:   #5 sim=0.1696  'Big in Japan' (771da62e-99ed-4f5d-a5a4-bc5a0e5b773a)
```

Outputting the track_parts just before they get returned:

```
190f8ff0-2587-43a9-a121-55c2b4383c52 ReleaseMatchParts(identifiers=[], similarity=[(1.0, 13), (1.0, 4), (0.07440000000000002, 10)], preferences=[(0, 2), (0.5, 111)])
f11939cf-9ad1-45b8-b927-a89b009e50e9 ReleaseMatchParts(identifiers=[(0.0, 28)], similarity=[(1.0, 13), (1.0, 4), (0.9287, 10)], preferences=[(1, 2)])
```

I'll provide a PR with proposed fixes.

phw (dismissed):

phw (approved):

PR #3167, opened 2026-05-18
PICARD-3272: simplify "never replace embedded image" type selection

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3272

The option "Never replace selected cover image types" for embedded images currently uses the CAA type selector dialog, with separate include / exclude lists. There are some issues with that:

- The default settings are copied from CAA types, where lower quality types (raw, watermark, matrix) are by default excluded from download. This makes sense. But here it means that Picard will never replace the front, but will do so anyway if there is some watermark or unedited version available.
- More generally, I think the logic is hard to grasp, and I don't really see a need for the exclude list. I think it should just be a list of types to select from.

## Solution

Simplify this by having only a single list of types not to overwrite. This seems to be sufficient for the use case and avoids the complexity. Use a different UI for selection.

New dialog:

grafik

Previous dialog:

grafik

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

We should eventually rename the config key used, this requires a config migration. See my comment below.

### DIFF
--- a/picard/ui/dialogs/cover_types_selector.py
+++ b/picard/ui/dialogs/cover_types_selector.py
@@ -24,6 +24,7 @@ from PyQt6 import (
     QtWidgets,
 )
 
+from picard.const.defaults import DEFAULT_CA_NEVER_REPLACE_TYPES
 from picard.coverart.utils import (
     CAA_TYPES,
     translate_caa_type,
@@ -44,17 +45,13 @@ class CoverTypesSelectorDialog(PicardDialog):
 
         self._types_list = QtWidgets.QListWidget(self)
 
-        # Ensure selected_types is a set
-        selected_types = set(selected_types or ())
-
         for type in CAA_TYPES:
             name = type['name']
             item = QtWidgets.QListWidgetItem(translate_caa_type(name))
             item.setData(QtCore.Qt.ItemDataRole.UserRole, name)
-            item.setCheckState(
-                QtCore.Qt.CheckState.Checked if name in selected_types else QtCore.Qt.CheckState.Unchecked
-            )
             self._types_list.addItem(item)
+
+        self._update_checked_items(selected_types)
 
         self._layout.addWidget(self._types_list)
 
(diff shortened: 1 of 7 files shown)


phw:
@zas About the config migration: I omitted this in this PR, just reusing the `dont_replace_included_types` parameter to still hold the types to ignore and silently stop registering the `dont_replace_excluded_types` option.

But I think we should migrate this and rename  the key to e.g. `keep_embedded_cover_types`, and remove the `dont_replace_excluded_types` key from config.

My problem right now: Each new migration requires bumping the version, but we already bumped to beta 2. Now bumping to beta 3 is confusing, because we skip beta 2. This all looks like we actually should have a separate config schema version or such.

What do you think we should do for now? Just not migrate for beta 2, and do the migration in the next beta? Or implement some different approach for config upgrades?

rdswift (approved):
I agree that this is simpler and easier to understand.  Code looks fine to me.

I also noted that the documentation needs to be updated when this is merged.

picard/ui/dialogs/cover_types_selector.py
```diff
@@ -0,0 +1,99 @@
...
+
+from picard.ui import PicardDialog
+
+
+class CoverTypesSelectorDialog(PicardDialog):
+    defaultsize = QtCore.QSize(400, 460)
+
+    def __init__(self, selected_types: Iterable[str] | None = None, parent=None):
+        super().__init__(parent)
+        self.setWindowTitle(_("Cover art types"))
```
rdswift, line 42: Not a big deal, but to make sure it's clear you might consider changing the title to something like "Don't replace cover art types" or "Never replace cover art types".

phw: That's very long and the window is rather narrow

phw:
> I also noted that the documentation needs to be updated when this is merged.

Not sure we really need a change. It currently already just says

> **Never replace selected cover image types**
> When this is enabled, Picard will not replace existing embedded cover images of the types specified. The types to retain are selected from a list in the selection dialog.

And I think that's fine. The dialog is now even easier to handle.

phw:
Ok, let's merge this. I'll do a config upgrade in a separate PR.

rdswift:
> > I also noted that the documentation needs to be updated when this is merged.
> 
> Not sure we really need a change. It currently already just says
> 
> > **Never replace selected cover image types**
> > When this is enabled, Picard will not replace existing embedded cover images of the types specified. The types to retain are selected from a list in the selection dialog.
> 
> And I think that's fine. The dialog is now even easier to handle.

Sorry, I assumed that there was a screen shot of the selection dialog, but didn't check.  I agree, no changes needed.

picard/ui/dialogs/cover_types_selector.py
rdswift: Good point.  Also, people rarely read the title of a dialog anyway.  :wink:

PR #3173, opened 2026-05-20
PICARD-3286: Log view rework, filtering, performance

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

The Log View dialog uses QTextBrowser/QTextDocument which becomes slow with many log entries (rebuilds the entire document on verbosity change). The highlight feature is not intuitive for finding strings in thousands of lines — it highlights but does not filter.

* JIRA ticket: PICARD-3286

## Solution

Rewrite using QListView with model/view architecture:

- `LogItemModel` + `LogFilterProxyModel` + `LogItemDelegate` in new `logviewmodel.py`
- Instant filtering by level and/or text regex (no document rebuild)
- Merged highlight and filter: matching rows shown with border highlight, non-matching hidden
- Filter UI: QLineEdit with embedded clear button + checkable toggle (Enter toggles on/off)
- No word wrap: `setUniformItemSizes(True)` for performance — word wrap forces Qt to measure every item on each insert (50s vs 9s for 1000 files). Long lines truncated at 500 chars; "View Detail…" in context menu opens scrollable dialog with full text
- 100ms batched updates: coalesces rapid log signals (~9.5s with log view open vs ~9s without)
- Context menu: View Detail, Copy, Select All, Clear Log (with icons)
- Plain substring filter by default (case-insensitive); optional regex mode via context menu toggle
- Debug Options menu stays open when toggling entries
- Compact View toggle in context menu (hides level prefix and source location)
- Regex Filter toggle in context menu for power users
- Clear Log moved from toolbar to context menu (declutters UI)
- Visible/total line count at bottom right
- Auto-scroll follows tail; scrolling up preserves position across filter/resize changes
- Explicit QueuedConnection for cross-thread log signal

image
image
image
image

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/logview.py
+++ b/picard/ui/logview.py
@@ -252,6 +252,10 @@ class LogView(LogViewCommon):
         self._delegate = LogItemDelegate(parent=self.list_view)
         self.list_view.setItemDelegate(self._delegate)
 
+        clear_log_action = QtGui.QAction(QtGui.QIcon.fromTheme("edit-clear"), _("Clear &Log…"), self.list_view)
+        clear_log_action.triggered.connect(self._clear_log_do)
+        self.list_view.addAction(clear_log_action)
+
         self.hbox = QtWidgets.QHBoxLayout()
         self.vbox.addLayout(self.hbox)
 
@@ -289,12 +293,6 @@ class LogView(LogViewCommon):
         self.filter_button.setToolButtonStyle(QtCore.Qt.ToolButtonStyle.ToolButtonTextOnly)
         self.filter_button.toggled.connect(self._on_filter_toggled)
         self.hbox.addWidget(self.filter_button)
-
-        # clear log
-        self.clear_log_button = QtWidgets.QPushButton(_("Clear Log…"))
-        self.clear_log_button.setAutoDefault(False)
-        self.hbox.addWidget(self.clear_log_button)
-        self.clear_log_button.clicked.connect(self._clear_log_do)
 
         # save as
         self.save_log_as_button = QtWidgets.QPushButton(_("Save As…"))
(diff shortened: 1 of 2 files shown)


phw (commented):
This is noticeable faster (I guess primarily because it is no longer using `QTextDocument`), and the filtering is more useful than only the previous highlighting.

Maybe someone will notice that text selection now works differently, being now line based instead of a full text document with selectable text. But for me the line based selection is easier to use than the previous text document. Also the speed and filtering advantages would IMHO outweigh this concern.

zas:
> Maybe someone will notice that text selection now works differently, being now line based instead of a full text document with selectable text. But for me the line based selection is easier to use than the previous text document. Also the speed and filtering advantages would IMHO outweigh this concern.

Yes, that's intentional tradeoff, but I have now added a View Detail dialog, you can select a bunch of lines, open the dialog and you can select whatever you want, and lines are properly wrapped.

Still work in progress, but the word wrap during logging is a real performance issue, I'm trying to minimizing the impact of the log view on overall perfomance, even with high rate of log output or a lot of lines in log. Not really possible without tradeoffs.

picard/ui/logview.py
```diff
@@ -78,24 +78,22 @@ def __init__(self, title, parent=None):
...
+        view_detail_action.triggered.connect(self._show_detail)
+        self.list_view.addAction(view_detail_action)
+
+        copy_action = QtGui.QAction(QtGui.QIcon.fromTheme("edit-copy"), _("&Copy"), self.list_view)
         copy_action.setShortcut(QtGui.QKeySequence.StandardKey.Copy)
         copy_action.triggered.connect(self._copy_selection)
         self.list_view.addAction(copy_action)
 
-        select_all_action = QtGui.QAction(_("Select &All"), self.list_view)
+        select_all_action = QtGui.QAction(QtGui.QIcon.fromTheme("edit-select-all"), _("Select &All"), self.list_view)
```
phw: In the main window itemviews it is `_("Select &all")`. Let's change this here for consistency

phw (changes requested):
This works well for me, and code looks good. I have a minor comment for string consistency

phw (approved):
LGTM

PR #3174, opened 2026-05-21
Refactor metadata matching into separate module

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

The `Metadata` class has become quite complex, as it includes all the logic for comparing metadata to releases and tracks. This includes a lot of functionality that is beyond the core scope of that class and is very dependent on the mbjson structure.

The matching logic is refactored into a separate module.

## Solution

Move the compare_to_release and compare_to_track functionality from Metadata into a separate module matching.py, including all the scoring helpers. Add extended tests for the scoring helpers.

The `Metadata` class retains the simple `Metadata.compare` and related weights, as this is more specific to the metadata itself and useful without other classes involved.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/matching.py
```diff
@@ -0,0 +1,544 @@
...
+                result.identifiers.append((score, id_w['catno']))
+
+    if 'release-group' in release:
+        tagger = PyQt6.QtCore.QCoreApplication.instance()
+        if tagger is not None:
+            rg = tagger.get_release_group_by_id(release['release-group']['id'])  # type: ignore[attr-defined]
+            if release['id'] in rg.loaded_albums:
+                result.identifiers.append((1.0, 6))
+
+    # Tier 2: Similarity — fuzzy matching core
```
zas, line 220: In the old code, compare_to_release_parts was a method on Metadata and acquired the read lock, this was removed here.

Perhaps wrap in `with metadata._lock.lock_for_read():`, this is a thread-safety issue with `if 'album' in metadata` followed by access to `metadata['album']`.

picard/matching.py
```diff
@@ -0,0 +1,544 @@
...
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from collections.abc import Iterable
+from dataclasses import dataclass, field
+from typing import (
+    TYPE_CHECKING,
+    TypeAlias,
+)
+
+import PyQt6.QtCore
```
zas: In most places we use `from PyQt6 import QtCore`, there's only one use here, but perhaps stick to the "standard" way.

picard/matching.py
```diff
@@ -0,0 +1,544 @@
...
+        return _DATE_MATCH_FACTORS['year']
+    if abs(release_year - metadata_year) <= 2:
+        return _DATE_MATCH_FACTORS['close_year']
+    return _DATE_MATCH_FACTORS['differed']
+
+
+def _catno_label_score(file_catno: str, file_label: str, release_label_info: list[dict]) -> float:
+    """Score catalog number + label match against release label-info.
+
+    file_catno and file_label should be pre-normalized (stripped, lowercased).
```
zas: Not anymore "pre-normalized", it was a performance optimization, removed here, not much impact I think but the docstring is now incorrect.

picard/matching.py
```diff
@@ -0,0 +1,544 @@
...
+# equal lengths will give a score of 1.0
+# example
+# a     b     score
+# 20000 0     0.333333333333
+# 20000 10000 0.666666666667
+# 20000 20000 1.0
+# 20000 30000 0.666666666667
+# 20000 40000 0.333333333333
+# 20000 50000 0.0
+LENGTH_SCORE_THRESHOLD_MS = 30000
```
zas, line 71: It was renamed from `LENGTH_SCORE_THRES_MS` but perhaps some plugins use the old name, not sure, I prefer the new name though.

picard/matching.py
```diff
@@ -0,0 +1,544 @@
...
+    Tier 3 (preferences): User preferences like country, format, release type.
+        Low weight — tie-breaking discriminators.
+    """
+
+    identifiers: list[ScoreWeightPair] = field(default_factory=list)
+    similarity: list[ScoreWeightPair] = field(default_factory=list)
+    preferences: list[ScoreWeightPair] = field(default_factory=list)
+
+    @property
+    def all_parts(self) -> list[ScoreWeightPair]:
```
zas: It doesn't look used anywhere, remove?

scripts/eval_matching/eval_matching.py
```diff
@@ -31,6 +31,11 @@
...
 )
 
+from picard.matching import (
+    compare_to_release,
+    compare_to_track,
+)
+
 
 # Add project root to path
 sys.path.insert(0, str(Path(__file__).parent.parent.parent))
```
zas, line 41: Not introduced in this PR, but this should probably be removed, I don't think it is needed.

zas (changes requested):
Overall that's a good and needed move, just few comments.

picard/matching.py
phw: Yes, I intentionally renamed it. I don't think it is otherwise clear what a "THRES" is. The import location changed anyway. I don't think any plugin is using this, but if it does, it must be updated.

scripts/eval_matching/eval_matching.py
phw: The imports will fail without it, unless you have the `picard` package installed (e.g. when running with uv). But I would keep this. It allows running the script directly from the source tree.

picard/matching.py
phw: You are right. I thought it was not needed, but missed the dependent metadata access. Added back in 53b838826db59ff20c447a34f04435bea58f151e

picard/matching.py
phw: Changed it to the common version

picard/matching.py
phw: I removed that line from the comment

picard/matching.py
phw: I don't think we need it, removed in b9312552026dbdb4e29adcb1a2e6c3dfa699d598

scripts/eval_matching/eval_matching.py
zas: Then it should be moved at top (before other imports)

phw: I think it is intentionally where it is, as the imports that follow it are the ones relative to the root.

zas (approved):
Much needed move, thanks! LGTM

PR #3176, opened 2026-05-21
Add tagger_instance() function for type safe access to the Tagger instance

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

The code needs access to the main application `Tagger` instance at many cases. Currently this means we have calls to `QtCore.QCoreApplication.instance()` scattered over the codebase. This has some issues:

1. Accessing the main tagger instance requires using a Qt specific function. If something changes in how this class is represented, we need to change all places where this gets accessed.
2. This approach does not work well with type checking. The returned instance is assumed to be of type `QCoreApplication`. But what is usually expected is `Tagger`, as the code needs access to the specific methods of that class. We know, that in our code the main application is always of type `Tagger`. We could cast or assert in all places where it is needed, but this increases code duplication.

## Solution

Add a util function `tagger_instance()` that wraps QCoreApplication.instance(). Assert the expected type to catch wrong app instance errors early in development.

Adjust some failing tests to pass the `isinstance` assertion.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -55,7 +55,10 @@ from PyQt6 import (
     QtWidgets,
 )
 
-from picard import log
+from picard import (
+    log,
+    tagger_instance,
+)
 from picard.album import (
     Album,
     NatAlbum,
@@ -86,7 +89,6 @@ from picard.util import (
     iter_files_from_objects,
     normpath,
     restore_method,
-    tagger_instance,
 )
 
 from picard.ui.collectionmenu import CollectionMenu
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -86,6 +86,7 @@ from picard.util import (
     iter_files_from_objects,
     normpath,
     restore_method,
+    tagger_instance,
 )
 
 from picard.ui.collectionmenu import CollectionMenu
@@ -180,7 +181,7 @@ class BaseTreeView(QtWidgets.QTreeWidget):
         self.columns = columns
         self.setAccessibleName(_(self.NAME))
         self.setAccessibleDescription(_(self.DESCRIPTION))
-        self.tagger = QtCore.QCoreApplication.instance()
+        self.tagger = tagger_instance()
         self.window = window
 
         # Subscribe to header update events
@@ -602,7 +603,7 @@ class BaseTreeView(QtWidgets.QTreeWidget):
     def drop_urls(urls, target, move_to_multi_tracks=True):
         files = []
         new_paths = []
-        tagger = QtCore.QCoreApplication.instance()
+        tagger = tagger_instance()
         for url in urls:
             log.debug("Dropped the URL: %r", url.toString(QtCore.QUrl.UrlFormattingOption.RemoveUserInfo))
             if url.scheme() == 'file' or not url.scheme():
--- a/picard/ui/itemviews/basetreeview.py
+++ b/picard/ui/itemviews/basetreeview.py
@@ -104,9 +104,7 @@ from picard.ui.itemviews.events import header_events
 from picard.ui.ratingwidget import RatingWidget
 from picard.ui.scriptsmenu import ScriptsMenu
 from picard.ui.util import menu_builder
-from picard.ui.widgets.configurablecolumnsheader import (
-    ConfigurableColumnsHeader,
-)
+from picard.ui.widgets.configurablecolumnsheader import ConfigurableColumnsHeader
 
 
 FILE_FILTERS = {'~filename', '~filepath'}
--- a/picard/util/__init__.py
+++ b/picard/util/__init__.py
@@ -42,22 +42,6 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from typing import (
-    TYPE_CHECKING,
-    Any,
-)
-
-from PyQt6.QtCore import QByteArray
-from PyQt6.QtNetwork import QNetworkReply
-
-
-try:
-    from charset_normalizer import detect  # type: ignore[unresolved-import]
-except ImportError:
-    try:
-        from chardet import detect  # type: ignore[unresolved-import,no-redef]
-    except ImportError:
-        detect = None  # type: ignore[assignment]
 from collections import (
     defaultdict,
     namedtuple,
@@ -85,12 +69,18 @@ import subprocess  # nosec: B404
 import sys
 import tempfile
 from time import monotonic
+from typing import Any
 import unicodedata
 
 from PyQt6 import QtCore
+from PyQt6.QtCore import QByteArray
 from PyQt6.QtGui import QDesktopServices
+from PyQt6.QtNetwork import QNetworkReply
 
-from picard import log
+from picard import (
+    log,
+    tagger_instance,
+)
 from picard.const import (
     MUSICBRAINZ_SERVERS,
     PICARD_DOCS_URLS,
@@ -108,12 +98,18 @@ from picard.i18n import (
 )
 
 
+try:
+    from charset_normalizer import detect  # type: ignore[unresolved-import]
+except ImportError:
+    try:
+        from chardet import detect  # type: ignore[unresolved-import,no-redef]
+    except ImportError:
+        detect = None  # type: ignore[assignment]
+
+
 winreg = None
 if IS_WIN:
     import winreg  # type: ignore[assignment]
-
-if TYPE_CHECKING:
-    from picard.tagger import Tagger
 
 # Windows path length constraints
 # See https://docs.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation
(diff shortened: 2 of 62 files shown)


picard/util/__init__.py
```diff
@@ -1498,3 +1505,11 @@ def atomic_write(path, data):
             with suppress(OSError, PermissionError):
                 temp_path.unlink()
         raise
+
+
+def tagger_instance() -> 'Tagger':
+    instance = QtCore.QCoreApplication.instance()
+    from picard.tagger import Tagger
```
zas: Just add a comment about avoiding circular imports here.
Also I wonder if having it in `picard/__init__.py`  would work, since this is imported everywhere I find it weird for it to be in util.py, dunno.

zas (changes requested):

picard/util/__init__.py
phw: You are right, I'll move it to `__init__.py`

phw: @zas I moved it in 020a129758718f2d5e769b35cb5c82ec0c3d5774

There is an important detail: We must not import Qt6 on this top level module. It must be possible to import e.g. `picard.VERSION` without the Qt6 dependency being installed. Hence this now also became an inline import.

picard/ui/itemviews/basetreeview.py
```diff
@@ -81,12 +84,7 @@
     NonAlbumTrack,
     Track,
 )
-from picard.util import (
-    icontheme,
-    iter_files_from_objects,
-    normpath,
-    restore_method,
-)
+from picard.util import icontheme, iter_files_from_objects, normpath, restore_method
```
zas: This change is unrelated

test/picardtestcase.py
```diff
@@ -47,47 +47,35 @@
...
 
 
 class FakeThreadPool(QtCore.QObject):
     def start(self, runnable, priority):
         runnable.run()
 
 
-class FakeTagger(QtCore.QObject):
-    tagger_stats_changed = QtCore.pyqtSignal()
+cleanup_tasks = []
```
zas: Not needed anymore I think but see comment below

test/picardtestcase.py
```diff
@@ -47,47 +47,35 @@
...
 
 class PicardTestCase(unittest.TestCase):
     def setUp(self):
         super().setUp()
         log.set_verbosity(logging.DEBUG)
         setup_gettext(None, 'C')
-        self.tagger = FakeTagger()
+        self.tagger = MockTagger()
         QtCore.QCoreApplication.instance = lambda: self.tagger
         self.addCleanup(self.tagger.run_cleanup)
```
zas: Since `self.tagger` is now a `MagicMock(spec=Tagger)`, calling `self.tagger.run_cleanup()` just calls a mock method that does nothing.
The old `FakeTagger.run_cleanup()` actually iterated `exit_cleanup` and called registered functions.
If any test relies on cleanup functions being executed (registered via `register_cleanup`), they'll silently stop running. Not sure there's any, but better check.

test/plugins3/test_registry.py
```diff
@@ -70,17 +70,15 @@ def get_url_mock(url, handler, **kwargs):
...
+        self.tagger.webservice.get_url = mock_webservice_fetch(response_data, error)
 
         result = {}
 
         def callback(success, err):
             result['success'] = success
             result['error'] = err
 
-        with patch('picard.plugin3.registry.QtCore.QCoreApplication.instance', return_value=mock_tagger):
+        with patch('picard.plugin3.registry.QtCore.QCoreApplication.instance', return_value=self.tagger):
```
zas: Patch targets should be `picard.tagger_instance` not `picard.plugin3.registry.QtCore.QCoreApplication.instance`

There are multiple occurences of those, so better double check.

test/plugins3/test_registry.py
```diff
@@ -616,25 +606,22 @@ def test_registry_parse_error_stops_fallback(self):
         url2 = 'https://second.example.com/registry.toml'
         registry = PluginRegistry(registry_url=[url1, url2])
 
-        mock_tagger = Mock()
-        mock_tagger.webservice = Mock()
-
```
zas, line 621: There are more of those below (see `test_registry_graceful_fallback_on_blacklist_check` for example) I guess you missed them.

zas (changes requested):
Good move, few more comments about things I noticed

test/plugins3/test_registry.py
phw: I'll check, but not necessarily. The plugins3 tests make heavy use of different mocks for tagger. There is another `MockTagger` in the helpers, and many tests just create a `Mock`.

I changed those that failed, because they make use of `tagger_instance`. But this stuff is tricky. If `tagger_instance` is involved we want the mock to pass as a `Tagger instance`.

But if the mock reaches https://github.com/metabrainz/picard/blob/master/picard/plugin3/manager/__init__.py#L104 , we don't want it to be a `QObject` (which it is, if it is also `Tagger`). Otherwise Qt will fail.

But I can try to convert as many as possible to `picardtestcase.MockTagger`.

test/picardtestcase.py
phw: Yes, this should be handled.

test/plugins3/test_registry.py
phw: This is changed now

test/picardtestcase.py
phw: @zas We used to need this primarily to cleanup the cover image temp files created during tests. But since we refactored this and images get cleaned on object destruction, I don't think we need this currently. I removed the cleanup call. Should be handled by tests that need it if it arises.

test/plugins3/test_registry.py
phw: @zas I changed a few more to use `PicardTestCase.tagger`. But everything using `PluginManager` needs a separate approach. Generally, there are some cases where you want the mocking to be done differently. But calls to getting the instance are so common, hence we have some default handling.

picard/ui/itemviews/basetreeview.py
phw: Fixed in the last rebase

test/plugins3/test_registry.py
phw: @zas as another note, it would probably be a good idea to leave the mocking of the `picard.tagger_instance` result (or `QtCore.QCoreApplication.instance` in the few cases it might be needed) to the individual tests. Currently we have some generic:

```
QtCore.QCoreApplication.instance = lambda: self.tagger
```

in `PicardTestCase.setUp`. But changing that is a major disruption to a large amount of our tests. If we do this, this should be an entirely separate PR.

picard/util/thread.py
```diff
@@ -40,7 +40,7 @@
     QRunnable,
 )
 
-from picard import log
+from picard import log, tagger_instance
```
zas: Minor, but:

```python
from picard import (
    log,
    tagger_instance,
)
```

zas (changes requested):
Just one minor change missed, else LGTM

picard/util/thread.py
phw: Fixed, also found another instance. I also included an additional commit 4646c0a026abe5e9fe43de81a28fe363b353b258 that fixes more imports (but otherwise unrelated to the tagger_instance change).

phw:
@zas Just to add to the discussion about the tests: We sometimes (I have so far only seen this on macOS) get an error like this during tests:

```
  File "/Users/runner/work/picard/picard/picard/util/thread.py", line 76, in run
    to_main(self.next_func, result=result)
  File "/Users/runner/work/picard/picard/picard/util/thread.py", line 133, in to_main
    QCoreApplication.postEvent(QCoreApplication.instance(), ProxyToMainEvent(func, *args, **kwargs))
TypeError: postEvent(receiver: QObject|None, event: QEvent|None, priority: int = Qt.NormalEventPriority): argument 1 has unexpected type 'MagicMock'
Fatal Python error: Aborted

Current thread 0x000000016b2f7000 (most recent call first):
  <no Python frame>

Thread 0x00000001f01dd0c0 (most recent call first):
  File "/Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/pathlib.py", line 573 in _cparts
  File ???, line 84625536 in ???

Extension modules: PyQt6.QtCore, objc._objc, CoreFoundation._CoreFoundation, CoreFoundation._inlines, Foundation._Foundation, Foundation._inlines, PyQt6.QtGui, PyQt6.QtNetwork, charset_normalizer.md, charset_normalizer.cd, PyQt6.QtWidgets, PyQt6.QtDBus, AppKit._AppKit, AppKit._inlines, yaml._yaml, pygit2._pygit2, _cffi_backend, PyQt6.QtMultimedia, PyQt6.QtTest (total: 19)
/Users/runner/work/_temp/e8332a93-bcd9-48e5-9686-0e3cfd1b4be3.sh: line 1: 19950 Abort trap: 6           pytest
test/test_util_time.py .
```

I have now just rerun the tests, this only randomly happens. But this is exactly caused by us messing with `QtCore.QCoreApplication.instance` in https://github.com/metabrainz/picard/blob/master/test/picardtestcase.py#L91

I tried to reduce this as part of this PR. But it touches so many tests that it is a major task and should be done separately.

But generally, we should properly work with the unittest `patch` function and isolate such mocking to the tests where it is really needed. And now patch only `picard.tagger_instance` instead of the Qt function itself.

zas (approved):
LGTM!

zas:
> But generally, we should properly work with the unittest `patch` function and isolate such mocking to the tests where it is really needed. And now patch only `picard.tagger_instance` instead of the Qt function itself.

This was done in https://github.com/metabrainz/picard/pull/3182

PR #3183, opened 2026-05-22
PICARD-3289: Allow user to set the default log level used when Picard is started

rdswift:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other
* **Describe this change in 1-2 sentences**:  Automatically start Picard with log level INFO by default.

## Problem

During review of https://github.com/metabrainz/picard-docs/pull/264 the idea was raised of always starting Picard with the default log level INFO to avoid excessive logging, unless it was started with the `--debug` command line option.

In general, it is recommended to set the log level to INFO for normal use to avoid excessive logging which can make finding specific information difficult in the log (and potentially impacting Picard's performance).  The DEBUG level is only recommended when troubleshooting specific issues to ensure that all log entries are captured.

Picard currently remembers the last log level used and automatically starts with that level.  Unfortunately many users don't remember to reset the log level to INFO after saving a DEBUG log for troubleshooting or reporting an issue.

* JIRA ticket (_optional_): PICARD-3289

## Solution

~Remove the `log_verbosity` setting and start Picard with log level INFO unless started with the `--debug` command line option or the PICARD_DEBUG environment variable is set.  Update version number and add configuration update test.~

The solution evolved during the process of this PR.  The end result was to allow the user to specifically set the default startup log level through the Options settings.  A new "Startup" options page was created as a sub-page of the "General" options page, and other startup-specific settings were moved to this new page.

The session log level can still be changed in the log viewer.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


zas (commented):
I think that's a good move, we don't need to switch from info, warning and error levels are very quiet.
If advanced users want to always start in debug mode they can use --debug or PICARD_DEBUG env var.

Though since that's a behavior change, I think it deserves a matching ticket for the changelog.

rdswift:
> Though since that's a behavior change, I think it deserves a matching ticket for the changelog.

Good point.  Ticket PICARD-3289 added, and the PR has been updated to refer to the ticket.

zas (dismissed):

phw:
I'm not sure this is an improvement. Remembering the last log level seems useful to me.

zas:
> I'm not sure this is an improvement. Remembering the last log level seems useful to me.

The point is that most users would set debug mode and forget about it, and it will slow down the app until they change it, but the current Log View doesn't really explain it is a "sticky" setting.
Since debug mode is mainly a dev thing (or asked by devs when it comes to debug), and advanced users have multiple ways to start Picard in debug mode (env var/cmd line option).

Also, error and warning levels are hiding most useful log lines from info level (which isn't exactly verbose), I think it is simpler and avoid issues for basic users.

We could also move the setting of the permanent option elsewhere, but that's just moving the problem elsewhere.

That said, not sure either about that yet, the core issue we are trying to fix is that debug mode isn't really meant to be permanent (but for devs and they just have to set PICARD_DEBUG=1 in their env).

rdswift:
> I'm not sure this is an improvement. Remembering the last log level seems useful to me.

How about a compromise (based on one of @zas's comments)...  Set the default start-up log level in the option settings somewhere, and not update the setting in the log viewer.  The selection in the log viewer would only be used for filtering the log view.  Initial setting for the default logging would be INFO.  This way the user could set their logging preference (including WARNING or ERROR), without it being changed accidentally in the log viewer.

zas:
> How about a compromise (based on one of @zas's comments)... Set the default start-up log level in the option settings somewhere, and not update the setting in the log viewer. The selection in the log viewer would only be used for filtering the log view. Initial setting for the default logging would be INFO. This way the user could set their logging preference (including WARNING or ERROR), without it being changed accidentally in the log viewer.

Yes, perhaps a better approach, we keep the option, but move it where all options are set usually, and clearly label it as default logging level. And Log View selector can became a list where one can filter log levels displayed. It would be more consistent.

phw:
I agree with the suggestions. That way it can still be changed permanently, but it doesn't happen accidentally in the log view.

Shall we merge this change then and do re-adding this as an option in a separate PR?

zas:
> Shall we merge this change then and do re-adding this as an option in a separate PR?

I prefer we fix it here, as this PR remove the option, etc... we keep it as is, but modify UI

rdswift:
> > Shall we merge this change then and do re-adding this as an option in a separate PR?
> 
> I prefer we fix it here, as this PR remove the option, etc... we keep it as is, but modify UI

I agree.  Fixing it here removes the need for increasing the version number because the setting is being removed, and then re-added in another PR.

rdswift:
Now another question...  I was originally planning on adding the setting to the "General" options page (because it really didn't seem to belong anywhere else), but that page is already starting to get a bit lengthy.  What do you think of adding a new "Startup" settings page, and also move the startup update checking settings from the General page to the new page?

rdswift:
The revised / new option pages would look like:

image


image

picard/ui/options/startup.py
```diff
@@ -0,0 +1,120 @@
...
+)
+
+from picard.ui.forms.ui_options_startup import Ui_StartupOptionsPage
+from picard.ui.options import OptionsPage
+
+
+class StartupOptionsPage(OptionsPage):
+    NAME = 'startup'
+    TITLE = N_("Startup")
+    PARENT = None
```
zas: Perhaps add this as a sub-page of 'General'?

zas (changes requested):

picard/ui/options/startup.py
phw: I like the idea of the Startup options, and grouping other startup related options there. But yes, grouping under General makes sense to me.

zas: @rdswift I created https://github.com/rdswift/picard/pull/32 with minor fixes

rdswift: Just merged it.  Thanks.

phw (dismissed):

zas (dismissed):
LGTM

zas (approved):
LGTM

zas (approved):

PR #3184, opened 2026-05-23
PICARD-3291: Aggregate file metadata in cluster for release lookup

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

Loading already-tagged files (e.g. Black Sabbath "The Dio Years" tagged from the DE release) and performing a cluster lookup would pick an arbitrary release  among tied candidates (e.g. the AU edition) because:

1. The barcode was not included in the MusicBrainz search query
2. The barcode-based scoring in compare_to_release was skipped (no barcode in cluster metadata)

* JIRA ticket (_optional_): [PICARD-3291](https://tickets.metabrainz.org/browse/PICARD-3291)

## Solution

When looking up a cluster, Picard now aggregates barcode, catalog number, date, and label from the files in the cluster. Previously these fields were never  propagated to the cluster metadata, so the lookup query and release scoring could not distinguish between releases that differ only by barcode/country.

`picard/cluster.py`: Add `_compute_aggregate_tags()` and `_aggregate_metadata()` to extract consensus values from files before lookup. Uses majority threshold and multi-valued intersection for safety.
`test/test_clustering.py`: Unit tests for the aggregation logic.
`scripts/eval_matching/`: Fix tagger_instance mocking, add corpus scenario reproducing the issue, add `--cluster-aggregated` mode testing the full aggregation→scoring pipeline.

Tag is set only when >50% of files have it AND all agree on the same value
Multi-valued tags (catalognumber, label): uses intersection across files (e.g. "Rhino; Warner" / "Rhino" = "Rhino")
If files disagree, the tag is not set (safer than guessing wrong)
  
Before: 5 candidates tied at 0.9106, result "ambiguous", arbitrary pick
After: correct release scores 0.9750, next best 0.5782, result "identified!"

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -207,6 +207,68 @@ class Cluster(FileList):
         if signal and self.ui_item:
             self.ui_item.update()
 
+    _AGGREGATE_TAGS = ('barcode', 'catalognumber', 'date', 'label')
+    _MULTI_VALUED_TAGS = ('catalognumber', 'label')
+
+    def _aggregate_metadata(self):
+        """Populate cluster metadata with album-level tags from files.
+
+        Called once before lookup to provide barcode, catalognumber, date,
+        and label for the search query and release scoring.
+
+        A tag is set on the cluster only when:
+          1. More than half the files carry a non-empty value for it.
+          2. All those values agree.
+
+        For single-valued tags (barcode, date), agreement means all
+        raw strings are identical.
+
+        For multi-valued tags (catalognumber, label), agreement is
+        based on the intersection of values across files. For example,
+        if 5 files have "L1; L2" and 5 have "L1", the common value
+        "L1" is used. If the intersection is empty, the tag is not set.
+
+        Tags that don't meet the conditions are removed from cluster
+        metadata, ensuring stale values from a previous aggregation
+        are not reused.
+        """
+        result = self._compute_aggregate_tags()
+        for tag in self._AGGREGATE_TAGS:
+            if tag in result:
+                self.metadata[tag] = result[tag]
+            elif tag in self.metadata:
+                del self.metadata[tag]
+
+    def _compute_aggregate_tags(self):
+        """Compute aggregated tag values from files.
+
+        Returns a dict of tag -> value for tags that meet the
+        consensus criteria.
+        """
+        result = {}
+        threshold = len(self.files) / 2
+        for tag in self._AGGREGATE_TAGS:
+            values = []
+            for f in self.files:
+                v = f.metadata.get(tag, '')
+                if v:
+                    values.append(v)
+            if not values or len(values) <= threshold:
+                continue
+            # Fast path: all raw strings identical (common case).
+            unique = set(values)
+            if len(unique) == 1:
+                result[tag] = values[0]
+                continue
+            # For multi-valued tags, find the intersection of values
+            # across all files that have the tag.
+            if tag in self._MULTI_VALUED_TAGS:
+                sets = [set(v.split(MULTI_VALUED_JOINER)) for v in unique]
+                common = sets[0].intersection(*sets[1:])
+                if common:
+                    result[tag] = MULTI_VALUED_JOINER.join(sorted(common))
+        return result
+
     def get_num_files(self):
         return len(self.files)
 
@@ -361,6 +423,7 @@ class Cluster(FileList):
         """Try to identify the cluster using the existing metadata."""
         if self._lookup_task:
             return
+        self._aggregate_metadata()
         self.tagger.window.set_statusbar_message(
             N_("Looking up the metadata for cluster %(album)s…"),
             {'album': self.metadata['album']},
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -247,12 +247,18 @@ class Cluster(FileList):
         """
         result = {}
         threshold = len(self.files) / 2
-        for tag in self._AGGREGATE_TAGS:
-            values = []
-            for f in self.files:
+        if not threshold:
+            return result
+
+        # Single pass over files: collect non-empty values for all tags
+        tag_values: dict[str, list[str]] = {tag: [] for tag in self._AGGREGATE_TAGS}
+        for f in self.files:
+            for tag, values in tag_values.items():
                 v = f.metadata.get(tag, '')
                 if v:
                     values.append(v)
+
+        for tag, values in tag_values.items():
             if not values or len(values) <= threshold:
                 continue
             # Fast path: all raw strings identical (common case).
--- a/picard/cluster.py
+++ b/picard/cluster.py
@@ -62,7 +62,6 @@ from picard.matching import (
     SimMatchRelease,
     compare_to_release,
 )
-from picard.metadata import MULTI_VALUED_JOINER
 from picard.track import Track
 from picard.util import (
     album_artist_from_path,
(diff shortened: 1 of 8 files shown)


rdswift (commented):
Matching is still a bit of "Black magic" to me, but this looks good and seems to improve the match quality.  Nice.

picard/cluster.py
```diff
@@ -207,6 +207,74 @@ def update(self, signal=True):
...
+        Returns a dict of tag -> value for tags that meet the
+        consensus criteria.
+        """
+        result = {}
+        threshold = len(self.files) / 2
+        if not threshold:
+            return result
+
+        # Single pass over files: collect non-empty values for all tags
+        tag_values: dict[str, list[str]] = {tag: [] for tag in self._AGGREGATE_TAGS}
```
phw: This could be a `defaultdict(list)`, to avoid the iteration over the tags. Then below the for loop can iterate over the tags and just append to the defaultdict.

picard/cluster.py
```diff
@@ -207,6 +207,74 @@ def update(self, signal=True):
...
+            if len(unique) == 1:
+                result[tag] = values[0]
+                continue
+            # For multi-valued tags, find the intersection of values
+            # across all files that have the tag.
+            if tag in self._MULTI_VALUED_TAGS:
+                sets = [set(v.split(MULTI_VALUED_JOINER)) for v in unique]
+                common = sets[0].intersection(*sets[1:])
+                if common:
+                    result[tag] = MULTI_VALUED_JOINER.join(sorted(common))
```
phw: I'm not sure about the use of `MULTI_VALUED_JOINER`. I kind of understand that we try to be permissive, and handle the case where a potential multi-value variable is using a single value with joiner. Usually we only support this for specific formats, and handle that automatically, or use the joiner for display purposes.

But if we have split such a tag we should not write it back with joiner to the resulting metadata, but actually write it back as multi valued. So here just:

```python
result[tag] = sorted(common)
```

Also I'm not sure, how well such a lookup will work. Does the lookup query handle multiple values here properly?

picard/cluster.py
```diff
@@ -361,6 +429,7 @@ def lookup_metadata(self):
         """Try to identify the cluster using the existing metadata."""
         if self._lookup_task:
             return
+        self._aggregate_metadata()
```
phw: We should rather aggregate the metadata when the cluster is created / has files added. If we do it on `lookup_metadata` the cluster will suddenly change the metadata on this action. This also affects displayed values.

test/test_clustering.py
```diff
@@ -261,3 +261,79 @@ def test_multi(self):
...
+        self.assertNotIn('barcode', result)
+
+    def test_no_files(self):
+        result = self.cluster._compute_aggregate_tags()
+        self.assertEqual(result, {})
+
+    def test_multi_valued_identical(self):
+        self._add_files('label', ['Rhino; Warner', 'Rhino; Warner', 'Rhino; Warner'])
+        result = self.cluster._compute_aggregate_tags()
+        self.assertEqual(result['label'], 'Rhino; Warner')
```
phw: This is the case I discussed in the code above. A lookup for `label:"Rhino; Warner"` likely performs worse then checking a single valie `label: "Rhino"`.

phw (changes requested):
Generally the right thing to do here, but a few comments / questions on the implemented logic.

test/test_clustering.py
zas: In this case all files agree on a set of labels, so it makes sense to query only that, but if the database has only one of them the search may give too restricted results. Should we just ignore multivalue in this case?

picard/cluster.py
zas: Yes, it makes sense.

picard/cluster.py
zas: Not sure about that either, I'll test.

test/test_clustering.py
phw: I'm not sure. But with the current way lookup queries are constructed I think we should only use a single value for the lookup. Currently we do a lookup like `label:(The Label Name)`, which would fit labels that have the given words in them (not necessarily in that order). If we want to have proper multi-value lookup we should probably OR multiple Dutch statements.

I would try focusing on one label. So probably count all values individually and use only the most used one, or the first one if the count is the same?

picard/cluster.py
zas: Done in https://github.com/metabrainz/picard/pull/3184/commits/05e7b6279c310034e437f998bdc1d36d5560e6b8

picard/cluster.py
zas: Done in https://github.com/metabrainz/picard/pull/3184/commits/3cee09335175eb13e5fc4f327da5bdb4f1e32619

picard/cluster.py
zas: > But if we have split such a tag we should not write it back with joiner to the resulting metadata, but actually write it back as multi valued. So here just:
> 
> ```python
> result[tag] = sorted(common)
> ```

I changed it in https://github.com/metabrainz/picard/pull/3184/commits/83603c48f2a415345e09ec177273b79ca06c0f89

test/test_clustering.py
zas: I did tests and Lucene escapes the semicolon (for multi-values), turning it into an OR query. In my test case it leads to ~60K results, noisy ranking, and the correct release gets buried.

So the code was modified to pick the single most frequent individual value across all files (with quorum).
This produces much less results (~11K), where the correct release ranks higher because the search term matches exactly.

The use of quorum is to help when files disagree about a tag, if 3 files have different labels for example, then it will be ignored for the search. So for multi-values we count the number of occurrences among all files and only retain those above the threshold. It seems reasonable to me.

Also I noticed the search on label was missing, added.

See https://github.com/metabrainz/picard/pull/3184/commits/7f6f45bafa942d77b2a881c07c09e3ae015cfd5a

picard/cluster.py
phw: This really reads nicer now.

picard/cluster.py
```diff
@@ -204,9 +204,103 @@ def remove_file(self, file: File, new_album=True):
...
+        Single pass over all files. Only returns a value for a tag if it
+        meets the quorum (more than half the files).
+        """
+        threshold = len(self.files) / 2
+        counts: dict[str, Counter[str]] = {tag: Counter() for tag in tags}
+        for f in self.files:
+            for tag in tags:
+                v = f.metadata.get(tag, '')
+                if v:
+                    for part in set(v.split(MULTI_VALUED_JOINER)):
```
phw: Wouldn't a `metadata.getall` above be better? Currently it uses `f.metadata.get(tag, '')`, which will return all values joined, then here string split it again.

Also instead of `f.metadata.get(tag, '')` only `f.metadata[tag]` could be used, same result.

It looks to me the LLM is really confused about the proper use of `Metadata` and considers `MULTI_VALUE_JOINER` overly important to use. Not sure we can fix this somehow, e.g. with better documentation of the methods.

phw (changes requested):

picard/cluster.py
zas: Yes, and I didn't notice it either (so that's my fault).

I fixed it in https://github.com/metabrainz/picard/pull/3184/commits/a2c94b8a0d0b76b4b6952ca26ecd58b26a2a8878

I also open a new PR to add docstrings to Metadata access methods, see https://github.com/metabrainz/picard/pull/3190

phw (approved):
LGTM, good work.

As a side effect we also have some additional data available in columns for a cluster :)

PR #3185, opened 2026-05-24
PICARD-3290: Fix file write access blocked by locked files on Windows

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: On Windows, files cannot be saved or moved when held open by the internal player or an external application. This PR fixes both cases: releasing the player's file handle before save/move, and retrying on transient PermissionError from external locks.

## Problem

* JIRA ticket: [PICARD-3290](https://tickets.metabrainz.org/browse/PICARD-3290)

On Windows, when a file is being played by Picard's built-in player, `QMediaPlayer` keeps the file handle open even after `stop()` is called. This prevents Picard from writing tags or moving the file, resulting in `PermissionError: [WinError 32]`. Since there is no "stop" button (only pause), the user has no way to release the file short of restarting Picard.

Additionally, external applications (antivirus scanners, file indexers, other media players) can temporarily hold file locks on Windows, causing the same PermissionError during save/move operations.

## Solution

Two separate fixes (one commit each):

**1. Internal player releases file handle before save/move**

- `Player.stop()` now calls `setSource(QUrl())` to fully release the file handle (previously it only called `stop()` which keeps the source open on Windows).
- Added `Player.release_file(filename)` slot that stops playback if the player currently holds the given file.
- `File.save()` (which runs on the main thread) calls `_release_file_from_player()` before dispatching the save task to the worker thread. Since both `save()` and the player live on the main thread, this is a simple direct call with no cross-thread complexity.

**2. Retry mechanism for externally locked files**

- Added `File._retry_on_permission_error(func)` that retries a callable up to 3 times with a 1-second delay on `PermissionError`. On non-Windows platforms it is a no-op (calls func directly).
- Wrapped the save and rename operations in `_save_and_rename()` with this retry.
- Runs in the save worker thread so it does not block the UI.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed with AI assistance (Kiro CLI), with human review and direction.

## Action

Additional actions required:

### DIFF
--- a/picard/file.py
+++ b/picard/file.py
@@ -463,6 +463,10 @@ class File(MetadataItem):
     def save(self):
         self.set_pending()
         run_file_pre_save_processors(self)
+        # On Windows, ask the internal player to release this file before
+        # writing or moving it (PICARD-3290).
+        if IS_WIN:
+            self._release_file_from_player(self.filename)
         metadata = Metadata()
         metadata.copy(self.metadata)
         thread.run_task(
@@ -490,6 +494,16 @@ class File(MetadataItem):
             errmsg = "Couldn't preserve timestamps for %r: %s" % (filename, why)
             raise self.PreserveTimesUtimeError(errmsg) from None
         return (st.st_atime_ns, st.st_mtime_ns)
+
+    def _release_file_from_player(self, filename):
+        """Ask the internal player to release the file handle if it holds it.
+
+        Must be called from the main thread (e.g. from save()).
+        """
+        window = getattr(self.tagger, 'window', None)
+        player = getattr(window, 'player', None) if window else None
+        if player and hasattr(player, 'release_file'):
+            player.release_file(filename)
 
     def _save_and_rename(self, old_filename, metadata):
         """Save the metadata."""
--- a/picard/file.py
+++ b/picard/file.py
@@ -123,6 +123,8 @@ from picard.ui.filter import Filter
 
 
 if TYPE_CHECKING:
+    from collections.abc import Callable
+
     from picard.cluster import Cluster
     from picard.track import Track
 
@@ -495,7 +497,32 @@ class File(MetadataItem):
             raise self.PreserveTimesUtimeError(errmsg) from None
         return (st.st_atime_ns, st.st_mtime_ns)
 
-    def _release_file_from_player(self, filename):
+    _PERMISSION_ERROR_RETRIES = 3
+    _PERMISSION_ERROR_RETRY_DELAY = 1.0  # seconds
+
+    def _retry_on_permission_error(self, func: 'Callable') -> None:
+        """Call func, retrying on PermissionError (Windows file locking).
+
+        This is safe to call from a worker thread.  On non-Windows platforms
+        or when retries are exhausted, the exception propagates normally.
+        """
+        if not IS_WIN:
+            return func()
+        for attempt in range(1, self._PERMISSION_ERROR_RETRIES + 1):
+            try:
+                return func()
+            except PermissionError:
+                if attempt >= self._PERMISSION_ERROR_RETRIES:
+                    raise
+                log.warning(
+                    "Permission denied on %r, retrying (%d/%d)…",
+                    self.filename,
+                    attempt,
+                    self._PERMISSION_ERROR_RETRIES,
+                )
+                time.sleep(self._PERMISSION_ERROR_RETRY_DELAY)
+
+    def _release_file_from_player(self, filename: str) -> None:
         """Ask the internal player to release the file handle if it holds it.
 
         Must be called from the main thread (e.g. from save()).
@@ -527,11 +554,11 @@ class File(MetadataItem):
             save = partial(self._save, old_filename, metadata)
             if config.setting['preserve_timestamps']:
                 try:
-                    self._preserve_times(old_filename, save)
+                    self._retry_on_permission_error(partial(self._preserve_times, old_filename, save))
                 except self.PreserveTimesUtimeError as why:
                     log.warning(why)
             else:
-                save()
+                self._retry_on_permission_error(save)
         # Rename files
         if config.setting['rename_files'] or config.setting['move_files']:
             new_filename = self._rename(old_filename, metadata, config.setting)
@@ -703,8 +730,36 @@ class File(MetadataItem):
         if not settings['move_overwrite_existing_files']:
             new_filename = get_available_filename(new_filename, old_filename)
         log.debug("Moving file %r => %r", old_filename, new_filename)
-        move_ensure_casing(old_filename, new_filename)
+        self._move_with_retry(old_filename, new_filename)
         return new_filename
+
+    def _move_with_retry(self, old_filename: str, new_filename: str) -> None:
+        """Move a file, retrying on PermissionError (Windows).
+
+        On cross-drive moves, shutil.move copies then deletes. If the delete
+        fails (source locked), a copy is left at the destination. Clean it up
+        before retrying.
+        """
+        if not IS_WIN:
+            move_ensure_casing(old_filename, new_filename)
+            return
+        for attempt in range(1, self._PERMISSION_ERROR_RETRIES + 1):
+            try:
+                move_ensure_casing(old_filename, new_filename)
+                return
+            except PermissionError:
+                if attempt >= self._PERMISSION_ERROR_RETRIES:
+                    raise
+                # Clean up partial copy at destination from failed cross-drive move
+                if os.path.exists(new_filename) and os.path.exists(old_filename):
+                    os.remove(new_filename)
+                log.warning(
+                    "Permission denied on %r, retrying (%d/%d)…",
+                    old_filename,
+                    attempt,
+                    self._PERMISSION_ERROR_RETRIES,
+                )
+                time.sleep(self._PERMISSION_ERROR_RETRY_DELAY)
 
     def _save_images(self, dirname, metadata):
         """Save the cover images to disk."""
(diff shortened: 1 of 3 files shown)


rdswift (commented):
I only took  a quick look at the code so far, but it looks fine to me.

I tested on Windows 10 with some interesting results:

1. Playing using Picard's internal player: Cluster and lookup while playing worked fine.  After clustering, the cover art "Show more details" occasionally failed to open the dialog. This failure was not consistently repeatable, and there was nothing in the log that indicated there was a problem.  Saving the release (write tags and move) stopped playback, and the files were successfully tagged and relocated.  This is what I expected, other than the random cover art issue which I suspect is not related to these changes.

2. Playing using third-party player VLC: Cluster and lookup while playing worked fine.  Saving the release (write tags and move) continued playback with only a slight hiccup, and the files were successfully tagged and relocated.  Attempting to replay the file in VLC failed with a file not found error.  This was a pleasant surprise, that playback continued even while the file was being updated and relocated.

3. Playing using Windows Media Player (WMP):  Cluster and lookup while playing worked fine.  Saving the release (write tags and move) continued playback, and the files were successfully tagged and relocated other than the file that was playing (which showed an error and logged "Permission denied").  Stopping playback and trying to save the failed file still failed and logged another "Permission denied" error.  After closing WMP, retrying the save worked successfully.  This is what I expected.

picard/file.py
```diff
@@ -491,6 +497,41 @@ def _preserve_times(self, filename, func):
...
+            except PermissionError:
+                if attempt + 1 >= self._PERMISSION_ERROR_RETRIES:
+                    raise
+                log.warning(
+                    "Permission denied on %r, retrying (%d/%d)…",
+                    self.filename,
+                    attempt + 1,
+                    self._PERMISSION_ERROR_RETRIES,
+                )
+                time.sleep(self._PERMISSION_ERROR_RETRY_DELAY)
```
rdswift: Rather than repeating `attemp + 1` throughout the loop, perhaps something like:

```python
for attempt in range(1, self._PERMISSION_ERROR_RETRIES + 1):
```

picard/file.py
```diff
@@ -689,9 +730,37 @@ def _rename(self, old_filename, metadata, settings):
...
+        """Move a file, retrying on PermissionError (Windows).
+
+        On cross-drive moves, shutil.move copies then deletes. If the delete
+        fails (source locked), a copy is left at the destination. Clean it up
+        before retrying.
+        """
+        if not IS_WIN:
+            move_ensure_casing(old_filename, new_filename)
+            return
+        for attempt in range(self._PERMISSION_ERROR_RETRIES):
```
rdswift: Similar comment to above.

rdswift (dismissed):
Code looks fine to me, and it seems to work as expected when I tested it.

picard/file.py
zas: Yes, cleaner, done and rebased.

picard/file.py
zas: Done, rebased.

rdswift (dismissed):

zas:
chaban confirmed it works as expected on the ticket

picard/ui/player/player.py
```diff
@@ -223,8 +223,20 @@ def stop(self):
         # hard stop, not just end of track
         self._playback_state = Player.PlaybackState.STOPPED
         self._player.stop()
+        self._player.setSource(QtCore.QUrl())
```
phw: I can't test this on Windows currently, and I'm not sure whether this explicit source unsetting is needed after a `_player.stop()`.

But if it is needed I think it would be better to call `self._player.setSource(QtCore.QUrl())`  in `_on_playback_state_changed` in case of `QMediaPlayer.PlaybackState.StoppedState`. This way it gets triggered in all cases if the player stops, not only when `stop()` gets explicitly called.

picard/ui/player/player.py
```diff
@@ -223,8 +223,20 @@ def stop(self):
...
+    @QtCore.pyqtSlot(str)
+    def release_file(self, filename: str):
+        """Release the file handle if the player is currently holding the given file.
+
+        This stops playback and clears the media source so the file can be
+        written to or moved.  Must be called from the main thread.
+        """
+        if self._current_file and self._current_file.filename == filename:
+            log.debug("Internal player: releasing file %r for save/move", filename)
+            self.stop()
```
phw, line 237: We could consider calling `self.play_next()` instead, so skipping the currently playing file if it gets saved, instead of stopping playback completely.

This requires moving the `setSource` to the state change signal as mentioned above.

picard/file.py
```diff
@@ -491,6 +497,41 @@ def _preserve_times(self, filename, func):
...
+                )
+                time.sleep(self._PERMISSION_ERROR_RETRY_DELAY)
+
+    def _release_file_from_player(self, filename: str) -> None:
+        """Ask the internal player to release the file handle if it holds it.
+
+        Must be called from the main thread (e.g. from save()).
+        """
+        window = getattr(self.tagger, 'window', None)
+        player = getattr(window, 'player', None) if window else None
```
phw: `player = self.tagger.player` is enough. `self.tagger` is always a `Tagger`, which has always a `window`, which has always a `player` attribute (which might be `None`).

picard/file.py
```diff
@@ -491,6 +497,41 @@ def _preserve_times(self, filename, func):
...
+        if player and hasattr(player, 'release_file'):
```
phw: Again, `Player` always has a `release_file` method, it got implemented above.

picard/file.py
```diff
@@ -689,9 +730,37 @@ def _rename(self, old_filename, metadata, settings):
...
+        for attempt in range(1, self._PERMISSION_ERROR_RETRIES + 1):
+            try:
+                move_ensure_casing(old_filename, new_filename)
+                return
+            except PermissionError:
+                if attempt >= self._PERMISSION_ERROR_RETRIES:
+                    raise
+                # Clean up partial copy at destination from failed cross-drive move
+                if os.path.exists(new_filename) and os.path.exists(old_filename):
+                    os.remove(new_filename)
```
phw, line 759: This looks dangerous. If `new_filename == old_filename` we remove the file we try to move. Also if the paths differ, but resolve to the same file also. Not sure if we have at this point normalized the paths enough. But we have the `picard.util.samefile` to check this case.

phw (approved):
I had hoped the new player would not lock the file as the old player did, and I even did some tests and closed PICARD-1616. But if it still happens, the approach here generally seems fine.

But I think it needs a couple of changes.

phw (changes requested):
See requested changes

picard/ui/player/player.py
zas: If I understand correctly, it is needed to release the file lock, but yes, in any case it makes sense to call it in _on_playback_state_changed.

picard/ui/player/player.py
zas: Hmm, yes, play_next() makes sense but it may be also locked, etc... If one plays an album, and save the whole album, it will hit locked files one after the other, no?

zas: I did check and we may end with more issues with play_next(), either because next target error because it moved, or just because of the sequence:

  1. Save track 3 (playing) → skip to track 4
  2. Save track 4 (now playing) → skip to track 5
  3. Save track 5 (now playing) → skip to track 6
  4. ...

So, I'm for keeping stop() approach for now, it is predictable at least.

picard/file.py
zas: Oh yes, good catch, and damn Windows...

zas: Fixed in https://github.com/metabrainz/picard/pull/3185/commits/7bf19a557d830b8c1602d628c20a4725b2713f9c

picard/ui/player/player.py
zas: Done in https://github.com/metabrainz/picard/pull/3185/commits/3bf14f86c00156978f82f71da9f8a0b4c7aa4dd4

picard/file.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3185/commits/d9633374bd7cf4b77d087423293b2ef3af235465 (and yes, it's annoying, I'll be more careful with that)

picard/file.py
zas: Addressed in https://github.com/metabrainz/picard/commit/d9633374bd7cf4b77d087423293b2ef3af235465

picard/ui/player/player.py
phw: Ok, let's keep stop then

picard/file.py
```diff
@@ -491,6 +498,40 @@ def _preserve_times(self, filename, func):
...
+            except PermissionError:
+                if attempt >= self._PERMISSION_ERROR_RETRIES:
+                    raise
+                log.warning(
+                    "Permission denied on %r, retrying (%d/%d)…",
+                    self.filename,
+                    attempt,
+                    self._PERMISSION_ERROR_RETRIES,
+                )
+                time.sleep(self._PERMISSION_ERROR_RETRY_DELAY)
```
phw, line 524: I found that we have a `RetryHelper.retry_until` helper, that also handles max retries and a delay. Not sure it helps simplifying the code much in this case, as it expects separate callables for the condition and the action to take, while here the condition is the write error of the action. But maybe worth investigating how to bring this together.

The `RetryHelper` currently lives in the `session` package, but could be moved if it gets used outside it.

phw (approved):

picard/file.py
zas: In https://github.com/metabrainz/picard/pull/3189 we agreed to remove `RetryHelper.retry_until`, I think the `sleep()` approach is much simpler in this case.

PR #3186, opened 2026-05-24
PICARD-3297: Fix cover art 'Show more details' failing after clustering

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Fix the cover art "Show more details" button failing to open the info dialog after clustering and lookup.

## Problem

* JIRA ticket: PICARD-3297

After clustering and looking up an album, clicking "Show more details" in the cover art box does nothing. Two issues:

1. `show_cover_art_info()` called `view_info()` without passing the item, relying on `selected_objects` which can be empty or stale after operations that move items in the tree (clustering, lookup).

2. `view_info()` had no `isinstance` check for `File` objects. When the cover art box's `self.item` was a File (common after lookup), it fell into the `else` branch which tried to extract a file from `selected_objects` instead of using the passed item — silently failing when `selected_objects` was empty.

## Solution

- `show_cover_art_info()` now passes `self.item` directly to `view_info()`, so the dialog always opens for the item the cover art box is displaying.
- `view_info()` accepts an optional `item` parameter. When provided, it uses that item directly instead of reading from `selected_objects`.
- Extracted `_get_info_dialog(item)` static method that resolves any item to the appropriate dialog class. The `else` branch now uses `iter_files_from_objects([selected])` (the resolved item) instead of falling back to `selected_objects`.
- Added debug log when `view_info()` is called with no selection.
- Added unit tests for `_get_info_dialog` covering all object types: Album, NatAlbum, Cluster, UnclusteredFiles, Track, File, FileList, and unknown objects.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed with AI assistance (Kiro CLI), with human review and direction.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -96,6 +96,7 @@ from picard.i18n import (
     gettext as _,
     ngettext,
 )
+from picard.item import Item
 from picard.options import (
     Option,
     get_option_title,
@@ -1682,7 +1683,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
                 return None
             return (FileInfoDialog, file)
 
-    def view_info(self, default_tab: int = 0, item: object | None = None) -> None:
+    def view_info(self, default_tab: int = 0, item: Item | None = None) -> None:
         """Open the info dialog for the given item or the current selection.
 
         Args:
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -1663,23 +1663,47 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         if obj and obj.release_group:
             AlbumSearchDialog.show_releasegroup_search(obj.release_group.id, obj)
 
-    def view_info(self, default_tab=0):
-        try:
-            selected = self.selected_objects[0]
-        except IndexError:
-            return
-        if isinstance(selected, Album):
-            dialog_class = AlbumInfoDialog
-        elif isinstance(selected, Cluster):
-            dialog_class = ClusterInfoDialog
-        elif isinstance(selected, Track):
-            dialog_class = TrackInfoDialog
+    @staticmethod
+    def _get_info_dialog(item: object) -> tuple[type, object] | None:
+        """Resolve an item to the appropriate info dialog class and target object.
+
+        Returns a (dialog_class, item) tuple, or None if no dialog can be shown.
+        """
+        if isinstance(item, Album):
+            return (AlbumInfoDialog, item)
+        elif isinstance(item, Cluster):
+            return (ClusterInfoDialog, item)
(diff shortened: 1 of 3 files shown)


rdswift (dismissed):
Code looks good to me, and it seemed to work quite well when I tested it.  Thanks.

picard/ui/mainwindow/__init__.py
```diff
@@ -1663,23 +1663,47 @@ def show_album_other_versions(self):
...
+            return (TrackInfoDialog, item)
         else:
             try:
-                selected = next(iter_files_from_objects(self.selected_objects))
+                file = next(iter_files_from_objects([item]))
             except StopIteration:
+                return None
+            return (FileInfoDialog, file)
+
+    def view_info(self, default_tab: int = 0, item: object | None = None) -> None:
```
phw: Instead of `object` the type hint could be for a `picard.item.Item`.

picard/ui/mainwindow/__init__.py
```diff
@@ -1663,23 +1663,47 @@ def show_album_other_versions(self):
...
+        """Open the info dialog for the given item or the current selection.
+
+        Args:
+            default_tab: Index of the tab to show initially.
+            item: Item to show info for.  If None, uses the first item
+                from selected_objects.
+        """
+        if item:
+            if not getattr(item, 'can_view_info', True):
```
phw: If we expect the `item` to be of type `Item` this `getattr` is not really needed and could be just `if not item.can_view_info`.

phw (changes requested):
Looks good. Some code suggestions

phw:
As a general note I really dislike the overuse of `getattr` for objects where we know what they are. This seems to be a typical AI thing, which creates overly defensive code. I also noticed this pattern in some previous PRs with heavy AI use.

If we follow this we should never call any method directly or access any attribute, but wrap anything into `getattr`. Which will make things unnecessarily hard to read.

We should really use this only if we:

1. Need this as duck typing, as we don't know what object we are dealing with
2. Now that this attribute is only present under certain circumstances (but in most such cases it would likely be better to define it with a default value)

picard/ui/mainwindow/__init__.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3186/commits/5c6d23c5d4d3d8d6fdda0a375e498debe7b737c6

picard/ui/mainwindow/__init__.py
zas: Done in https://github.com/metabrainz/picard/pull/3186/commits/5a2c803d5db8ba0bdeeaf0df2067efa3448e3264

phw (approved):
LGTM

PR #3188, opened 2026-05-24
PICARD-3281: Reduce UI blocking when loading many files

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Reduce UI blocking when loading many files by throttling both file load dispatching and background task callback processing.

## Problem

* JIRA ticket: [PICARD-3281](https://tickets.metabrainz.org/browse/PICARD-3281)

When loading many files (e.g. 2000 singletons), the UI becomes completely unresponsive — users cannot scroll, click menus, or interact with the application until all files are loaded. This is reported on Windows 10/11 but also reproducible on Linux.

**Root cause analysis:**

1. All `file.load()` calls were dispatched to the thread pool at once. With 8+ threads completing in parallel, hundreds of completion callbacks (`ProxyToMainEvent`) queue up on the main thread.

2. `Tagger.event()` processed each `ProxyToMainEvent` immediately and synchronously. Qt delivers all queued events of the same type sequentially, so the main thread executes hundreds of callbacks back-to-back without ever yielding to process paint or input events.

3. Calling `processEvents()` from inside the event handler doesn't help — it recursively processes more `ProxyToMainEvent`s, making the problem worse.

**Testing methodology:**

- Tested with ~2000 audio files on Linux/SSD
- Added timing instrumentation showing the main thread was blocked for up to 2+ seconds between event loop yields
- Verified that `processEvents()` inside callbacks causes recursion (processes more callbacks, not paint events)
- Tested batch sizes of 20, 25, and 50 — settled on 25 as a good balance between responsiveness and loading speed

## Solution

Two complementary changes:

**1. Batch file load dispatching** (commit 1)

Instead of dispatching all file loads at once, dispatch them in batches of 25 using `QTimer.singleShot(0)` between batches. This naturally limits how many completion callbacks can arrive in a burst.

**2. Throttle callback processing** (commit 2)

Instead of running each `ProxyToMainEvent` immediately in `Tagger.event()`, queue them internally and process 25 at a time. Between batches, `QTimer.singleShot(0)` yields to the event loop, allowing Qt to process paint and input events.

This second change is generic — it benefits all background task callbacks (file loading, AcoustID lookups, album loading, cover art downloads), not just file loading.

**Result:** The UI remains responsive during bulk file loading. Users can scroll, open menus, and interact with the application while files load in the background. There is a minor increase in total loading time due to the yielding, but the perceived performance is dramatically better.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

Code was developed with AI assistance (Kiro CLI), with human review, direction, and manual testing.

## Action

Additional actions required:

**Requesting feedback:** This is a draft for discussion. The batch size (25) may need tuning based on testing on Windows with large collections. The callback throttling approach changes how all background task results are delivered to the main thread, so thorough testing across different workflows is recommended.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -822,9 +822,16 @@ class Tagger(QtWidgets.QApplication):
         else:
             self.browser_integration.stop()
 
+    _CALLBACK_BATCH_SIZE = 25
+    _callback_queue: list = []
+    _callback_timer_running: bool = False
+
     def event(self, event):
         if isinstance(event, thread.ProxyToMainEvent):
-            event.run()
+            self._callback_queue.append(event)
+            if not self._callback_timer_running:
+                self._callback_timer_running = True
+                QtCore.QTimer.singleShot(0, self._process_callback_batch)
         elif event.type() == QtCore.QEvent.Type.FileOpen:
             file = event.file()
             self.add_paths([file])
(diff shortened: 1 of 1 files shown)


rdswift (commented):
This looks like it should work well.  Nice improvement.

phw (approved):
I can't reproduce the UI blocking when loading many files. But the changes look plausible and I didn't realize any regression.

sakgoyal:
Defender exclusion did not help. This PR did not help either. I still have the same issue.

PR #3193, opened 2026-05-26
PICARD-3294: Move 'never replace' cover art filters to file assignment time

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Move the "Never replace cover images with smaller ones" and "Never replace selected cover image types" filters from download time to file assignment time, so they work regardless of whether files are matched before or after cover art is loaded.

## Problem

* JIRA ticket: PICARD-3294

The "never replace" filters ran at download time comparing against `album.orig_metadata.images`. This only worked if files were already matched to the album when cover art was downloaded. If files were matched later, the filters had already passed (since `album.orig_metadata.images` was empty), and images were assigned to files without filtering.

## Solution

- Remove `bigger_previous_image_filter` and `image_types_filter` from the download-time filter chain
- Refactor both filters to accept `(coverartimage, previous_images)` — comparing against a specific file's embedded images rather than the album aggregate
- Apply per-file filtering in `CoverArtSetter._set_image` so it covers all code paths (`_handle_album`, `_handle_filelist`, `_handle_file`)
- Apply per-file filtering in `track.update_file_metadata()` for files matched after cover art is loaded
- Apply the same filter in the cover art info dialog so "New Embedded" accurately shows what will be saved
- Filters only apply when `save_images_to_tags` is enabled (nothing to protect if not embedding)
- Add `DebugOpt.COVERART` logging for filter accept/reject decisions

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI investigated the codebase, proposed the fix approach, and implemented the changes. Human reviewed, tested manually with real files (Black Sabbath - The Dio Years), and validated the debug log output.

## Action

Additional actions required:

### DIFF
--- a/picard/coverart/setters/handlers.py
+++ b/picard/coverart/setters/handlers.py
@@ -26,8 +26,6 @@ from functools import singledispatch
 from picard import log
 from picard.album import Album
 from picard.cluster import Cluster
-from picard.config import get_config
-from picard.coverart.processing.filters import filter_image_for_file
 from picard.file import File
 from picard.item import (
     FileListItem,
@@ -88,11 +86,8 @@ def _handle_album(album: Album, setter) -> bool:
             setter._set_image(track)
 
         for file in album.iterfiles():
-            if get_config().setting['save_images_to_tags']:
-                if not filter_image_for_file(setter.coverartimage, file.orig_metadata.images):
-                    continue
-            setter._set_image(file)
-            file.update(signal=False)
+            if setter._set_image(file):
+                file.update(signal=False)
 
     album.update(update_tracks=True)
     return True
@@ -129,8 +124,8 @@ def _handle_filelist(filelist: FileListItem, setter) -> bool:
                 stack.enter_context(parent.suspend_metadata_images_update)
                 parents.add(parent)
 
-            setter._set_image(file)
-            file.update(signal=False)
+            if setter._set_image(file):
+                file.update(signal=False)
 
         for parent in parents:
             if isinstance(parent, Album):
@@ -163,8 +158,8 @@ def _handle_file(file: File, setter) -> bool:
     """
     log.debug("set_coverart_file %r", file)
 
-    setter._set_image(file)
-    file.update()
+    if setter._set_image(file):
+        file.update()
     return True
 
 
--- a/picard/coverart/setters/handlers.py
+++ b/picard/coverart/setters/handlers.py
@@ -26,6 +26,8 @@ from functools import singledispatch
 from picard import log
 from picard.album import Album
 from picard.cluster import Cluster
+from picard.config import get_config
+from picard.coverart.processing.filters import filter_image_for_file
 from picard.file import File
 from picard.item import (
     FileListItem,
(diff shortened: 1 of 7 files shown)


picard/coverart/setters/handlers.py
```diff
@@ -86,6 +88,9 @@ def _handle_album(album: Album, setter) -> bool:
             setter._set_image(track)
 
         for file in album.iterfiles():
+            if get_config().setting['save_images_to_tags']:
+                if not filter_image_for_file(setter.coverartimage, file.orig_metadata.images):
+                    continue
```
phw: Shouldn't this check be implemented also for `_handle_filelist` and `_handle_file`? Maybe move the logic to `CoverArtSetter._set_image` (check if the passed obj is a `File`, apply the filter if it is).

phw (commented):
Looks good and works with applying the filters correctly. I have some comment about completeness in the cover art setters.

We could consider making the filters for embedded images also an extension point, but this can be done in a future update.

picard/coverart/setters/handlers.py
zas: Yes, it makes sense. Done in https://github.com/metabrainz/picard/pull/3193/commits/f5b846f7d2c2dcc9a059c0e2a4ca9fe916d8cd8b

docs/CoverArt.md
```diff
@@ -0,0 +1,219 @@
...
+
+The most common provider. The CAA stores images uploaded by MusicBrainz editors
+with metadata: image types (front, back, booklet, etc.), approval status, and
+available thumbnail sizes.
+
+**How it works:**
+
+1. Picard fetches the CAA JSON metadata for the release (list of available images)
+2. For each image, it checks:
+   - Is it approved? (if "approved only" is enabled)
```
rdswift: Not sure if it's necessary, but you might consider clarifying that "approved" means that the edit to add the image to CAA was approved and is no longer pending.

docs/CoverArt.md
```diff
@@ -0,0 +1,219 @@
...
+When the user saves a file, Picard writes images based on two independent
+options:
+
+### Save to Tags (embedding)
+
+Images from `file.metadata.images` are embedded directly in the audio file's
+tags (ID3 APIC frames, Vorbis METADATA_BLOCK_PICTURE, MP4 covr atoms, etc.).
+
+Related options:
+- **Embed only one front image** — Only embed the first front image, skip others
```
rdswift: Should this clarify how the "first front image" is determined if there are multiple front images?

rdswift (commented):
Looks good to me.

I have a personal preference with markdown files not to introduce unnecessary newlines just to make the source files look pretty by manually word wrapping, but I guess that's just how the AI generates the file.  Not a big deal, just a personal preference (and it makes for cleaner patches when there are changes and things need to be re-wrapped).

docs/CoverArt.md
zas: Doc removed (will serve as basis to update official documentation instead)

docs/CoverArt.md
zas: Doc removed (will serve as basis to update official documentation instead)

phw (approved):

PR #3196, opened 2026-05-26
Plugin UI quality-of-life improvements

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

A collection of small plugin UI improvements: security warning on install, better visual feedback, bug fixes, and keyboard accessibility.

## Problem

Several small UX issues in the plugin management UI:
- No general security warning when installing plugins
- Hardcoded colors breaking dark themes
- Double separator in context menu
- No keyboard shortcut for enable/disable
- "New Version" column showing "Available" instead of actual version info
- Status log limited to 3 messages with fixed pixel height

## Solution

1. **Security warning** on plugin install dialog with "don't show again" checkbox
2. **Bold plugin name** and clickable repository URL in install confirm dialog
3. **Default ref** shown more prominently in ref selector (bold "Recommended:" prefix)
4. **Fix double separator** in plugin list context menu
5. **Replace hardcoded color** in "plugins unavailable" warning with theme-aware icon + bold text
6. **Space key shortcut** to toggle plugin enabled/disabled state
7. **Fix update version display** for UpdateCheck objects (regression from c9a616f0f)
8. **Improve status log**: font-scaled height, full history with scroll, show plugin count only on change

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI proposed changes; human reviewed, tested interactively, and directed design decisions.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/options/plugins.py
+++ b/picard/ui/options/plugins.py
@@ -58,6 +58,7 @@ class Plugins3OptionsPage(OptionsPage):
 
         # Cache plugin manager for performance
         self.plugin_manager = self.tagger.get_plugin_manager()
+        self._last_plugin_count = 0
 
         self.setup_ui()
         if self.plugin_manager:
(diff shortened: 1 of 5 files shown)


rdswift (commented):
Looks good, and seems to work well when I tested it here.  The only thing I noticed was that there doesn't seem to be any way to reset the `show_plugin_install_warning` flag once it has been disabled.  I wonder if there should be something like the "Reset tutorial and setup wizard..." button in the User Interface options page, or perhaps something to allow resetting all those type of UI warnings to enabled?

zas:
> Looks good, and seems to work well when I tested it here. The only thing I noticed was that there doesn't seem to be any way to reset the `show_plugin_install_warning` flag once it has been disabled. I wonder if there should be something like the "Reset tutorial and setup wizard..." button in the User Interface options page, or perhaps something to allow resetting all those type of UI warnings to enabled?

Yes, it will be in another PR if we decide to do that, one button to reset them all ;)

rdswift (dismissed):
Looks good.

One really minor thing I noticed while testing is that the status log always shows a blank line at the bottom.  I couldn't see the reason for that in the code, unless it has something to do with setting the scroll bar value to the maximum.  Definitely not a show-stopper.

picard/ui/widgets/pluginlistwidget.py
```diff
@@ -295,8 +297,31 @@ def _setup_update_column(self, item, plugin):
...
+        and UpdateResult (from actual updates, has new_ref_item).
+        """
+        # UpdateResult has new_ref_item
+        new_ref_item = getattr(update, 'new_ref_item', None)
+        if new_ref_item:
+            return new_ref_item.format() or _("Available")
+
+        # UpdateCheck has new_ref and new_commit
+        ref = getattr(update, 'new_ref', None) or getattr(update, 'old_ref', 'main')
+        commit = getattr(update, 'new_commit', None)
```
phw, line 312: These `getattr` again suspiciously look like they shouldn't be needed. We should always deal with a `UpdateCheck` here. Might hide issues.

picard/ui/options/plugins.py
```diff
@@ -157,36 +161,35 @@ def setup_ui(self):
...
         self.status_messages.append(formatted_message)
-        if len(self.status_messages) > 3:
-            self.status_messages.pop(0)
 
         # Update display
         self.status_log.setPlainText("\n".join(self.status_messages))
         # Scroll to bottom to show latest message
-        self.status_log.verticalScrollBar().setValue(self.status_log.verticalScrollBar().maximum())
+        sb = self.status_log.verticalScrollBar()
+        sb.setValue(sb.maximum())
```
phw: Changing this to `sb.setValue(sb.maximum() - 1)` makes it show always the last three lines, without that extra empty line. The user can still technically scroll down a bit further, as that extra empty line is still present. But the default display updates more nicely.

phw (changes requested):
Nice improvements, looks good. I think I found a solution for the empty line issue in the log, see my comment in the code.

picard/ui/options/plugins.py
zas: Ah lol, I was implementing it, I just pushed the change and then I read your comment... well, that's exactly how I fixed it ;)

picard/ui/widgets/pluginlistwidget.py
zas: Yes, but I'll review later,  this part is reverting the code incorrectly removed in another PR.

phw (approved):
LGTM

picard/ui/widgets/pluginlistwidget.py
zas: Addressed as part of https://github.com/metabrainz/picard/commit/5e0145dfa8b5bf754f01a3e67a47d45d611efbf8

PR #3199, opened 2026-05-28
PICARD-3281: Time-budgeted batch processing for UI responsiveness

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Replace fixed-size batch processing with a time-budgeted approach (50ms per batch) for callback processing, file load dispatching, and clustering. Also adds a TIMINGS debug option for diagnosing UI responsiveness issues.

## Problem

* JIRA ticket: PICARD-3281

The previous fix (#3188) batched callbacks in groups of 25, but each batch could still block the main thread for 100-330ms on fast machines (and likely 500ms+ on slower Windows machines). The user reported the fix didn't help.

Root cause: with 2000+ pre-tagged files, each callback does tree widget operations (~2-4ms each). A fixed batch of 25 means 50-100ms of blocking per batch. Additionally, clustering was running synchronously without yielding at all.

## Solution

**1. Time-budgeted batch processing** (replaces fixed batch size)

Instead of processing a fixed number of items per batch, process items until a 50ms time budget is exhausted, then yield to the event loop. This adapts to both fast and slow machines:
- Light callbacks (API responses): many per batch
- Heavy callbacks (tree updates): fewer per batch
- Guarantees ~20 yields/sec regardless of per-item cost

**2. Generic `_process_in_batches()` utility**

Reusable method for any list that needs time-budgeted processing with event loop yielding. Supports an `on_complete` callback for post-processing.

**3. Time-budgeted clustering**

`_clustering_finished` now uses the same approach instead of processing all clusters synchronously.

**4. TIMINGS debug option**

New `--debug-opts=timings` option with a `timing()` context manager on `DebugOptEnum`. Logs elapsed time for instrumented code blocks. Used to measure callback batch duration, file load dispatch, clustering, and exit cleanup.

**Testing results (2025 files, Linux/SSD):**

| Metric | Old (fixed 25) | New (50ms budget) |
|--------|---------------|-------------------|
| Items/batch during load | 25 | 4-14 (adaptive) |
| Batch blocking time | 100-330ms | ≤50ms |
| UI yields/sec | ~4-8 | ~20 |
| Total load time | ~10-12s | ~18s |
| UI responsive | No | Yes |

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed with AI assistance (Kiro CLI), with human direction, testing, and review.

## Action

Additional actions required:

**Requesting feedback:** The 50ms time budget may need tuning. Testing on Windows with large collections of singletons (the reporter's scenario) would be valuable.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -832,7 +832,9 @@ class Tagger(QtWidgets.QApplication):
     _callback_queue: list = []
     _callback_timer_running: bool = False
 
-    def _process_in_batches(self, items: list, process_func, label: str, time_budget: float | None = None) -> None:
+    def _process_in_batches(
+        self, items: list, process_func, label: str, time_budget: float | None = None, on_complete=None
+    ) -> None:
         """Process items with a time budget, yielding to the event loop between batches.
 
         Args:
@@ -840,6 +842,7 @@ class Tagger(QtWidgets.QApplication):
             process_func: Callable that processes a single item.
             label: Label for debug timing output.
             time_budget: Max seconds per batch. Defaults to _BATCH_TIME_BUDGET.
+            on_complete: Optional callable invoked when all items are processed.
         """
         if time_budget is None:
             time_budget = self._BATCH_TIME_BUDGET
@@ -850,7 +853,11 @@ class Tagger(QtWidgets.QApplication):
             count += 1
         log.debug_if(DebugOpt.TIMINGS, "%s: %d items, %d remaining", label, count, len(items))
         if items:
-            QtCore.QTimer.singleShot(0, partial(self._process_in_batches, items, process_func, label, time_budget))
+            QtCore.QTimer.singleShot(
+                0, partial(self._process_in_batches, items, process_func, label, time_budget, on_complete)
+            )
+        elif on_complete:
+            on_complete()
 
     def event(self, event):
         if isinstance(event, thread.ProxyToMainEvent):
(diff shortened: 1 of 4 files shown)


picard/tagger.py
```diff
@@ -728,9 +728,15 @@ def exit(self):
...
-        DataHash.remove_all_files()
+        with DebugOpt.TIMINGS.timing("run_cleanup"):
+            self.run_cleanup()
+        with DebugOpt.TIMINGS.timing("DataHash.remove_all_files"):
+            DataHash.remove_all_files()
         QtCore.QCoreApplication.processEvents()
+        if sys.stdout:
+            sys.stdout.flush()
+        if sys.stderr:
+            sys.stderr.flush()
```
phw, line 739: Why is this needed?

picard/tagger.py
```diff
@@ -857,7 +893,8 @@ def _file_loaded(self, file, target=None, remove_file=False, unmatched_files=Non
         config = get_config()
         self._pending_files_count -= 1
         if self._pending_files_count == 0:
-            self.window.suspend_while_loading_exit()
+            with DebugOpt.TIMINGS.timing("suspend_while_loading_exit"):
+                self.window.suspend_while_loading_exit()
```
phw, line 897: Just some side notes: This should not have any significant timing issue, it currently just sets two booleans. But generally I think we have overengineered this, with the `register_suspend_while_loading` registering callbacks and calling them.

IMHO we should get rid of that again and just implement setting the few things we want in `suspend_while_loading_enter` and `suspend_while_loading_exit`.

phw (approved):
Looks good. Some questions / points for discussion, but nothing blocking this PR

picard/tagger.py
zas: In some cases, log output was flushed when exiting and I was missing last timings (those just above).

PR #3202, opened 2026-05-29
PICARD-3304: Fix picard-plugins CLI crash on startup

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

The picard-plugins CLI crashed with 'Expected an instance of Tagger' on every command since commit 3f188e0c2 (2026-05-21), which replaced `QCoreApplication.instance() `with `tagger_instance()` in `WebService.__init__`.

The CLI's `minimal_init()` creates a plain `QCoreApplication` (not a Tagger), so the assertion in `tagger_instance()` always fails in the CLI context.

* JIRA ticket (_optional_): PICARD-3300, PICARD-3304

## Solution

- Remove `self.tagger` from `WebService.__init__`: it became dead code when commit 9bf514b82 replaced self.tagger.tagger_stats_changed.emit()` with the dedicated `WebService.pending_requests_changed` signal.
- Replace `tagger_instance() `calls in cli.py with `QtCore.QCoreApplication.instance()`, since the CLI only needs `processEvents()` and never requires a full Tagger.
- Add a subprocess integration test that catches this class of regression by running `picard-plugins` in a real process without mocking.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/plugin3/cli.py
+++ b/picard/plugin3/cli.py
@@ -40,7 +40,6 @@ from picard import (
     PICARD_FANCY_VERSION_STR,
     PICARD_ORG_NAME,
     log,
-    tagger_instance,
 )
 from picard.config import (
     Option,
@@ -185,7 +184,7 @@ class PluginCLI:
 
         self._manager._registry.fetch_registry(callback=callback)
 
-        app = tagger_instance()
+        app = QtCore.QCoreApplication.instance()
         while not result['done']:
             app.processEvents()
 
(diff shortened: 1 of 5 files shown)


phw (approved):
The patch makes sense, even independently of the app crash:

- `WebService.tagger` was not used, so does not need to be there
- `plugin3/cli.py` only needed Qt specific functionality  (`app.processEvents()`), so again no need for `tagger_instance`

PR #3205, opened 2026-05-31
PICARD-3300: Exclude picard-plugins CLI from macOS app bundle

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3300

macOS does not support multiple binary targets in the same app bundle. Previously we had the picard-plugins marked as `console=True`, which made macOS launch even the GUI program in console mode without window. Now it runs both the GUI and CLI on every app start.

Given that the CLI is hidden in the app folder anyway and hard to access, completely exclude it from the bundle. Users wanting the CLI can install Picard via pypi.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard.spec
+++ b/picard.spec
@@ -148,56 +148,48 @@ else:
         strip=False,
         upx=False,
         icon='picard.ico',
-        version='win-version-info.txt',
         console=False,
-        # macOS specific
-        target_arch=os.environ.get('TARGET_ARCH', None),
-        codesign_identity=os.environ.get('CODESIGN_IDENTITY', None),
-        entitlements_file='./scripts/package/entitlements.plist',
-    )
-
-    # The picard-plugins CLI tool
-    a_plugins = Analysis(
-        ['picard/plugin3/cli.py'],
-        pathex=['picard'],
-        binaries=[],
-        datas=[],
-        hiddenimports=['cffi'],
-        hookspath=None,
-        runtime_hooks=[],
-        excludes=excludes,
-    )
-    pyz_plugins = PYZ(a_plugins.pure, a_plugins.zipped_data)
-    exe_plugins = EXE(
-        pyz_plugins,
-        a_plugins.scripts,
-        exclude_binaries=True,
-        name='picard-plugins',
-        debug=False,
-        strip=False,
-        upx=False,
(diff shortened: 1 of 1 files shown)


picard.spec
```diff
@@ -134,62 +142,73 @@ else:
...
+    else:
+        coll = COLLECT(
+            exe,
+            a.binaries,
+            a.zipfiles,
+            a.datas,
+            strip=False,
+            upx=False,
+            name='picard',
+        )
```
zas: Not blocking, but perhaps we could avoid some redundancy using something like:

```python
collect_items = [exe, a.binaries, a.zipfiles, a.datas]
if os_name != 'Darwin':
    # ... build a_plugins, pyz_plugins, exe_plugins ...
    collect_items.extend([exe_plugins, a_plugins.binaries, a_plugins.zipfiles, a_plugins.datas])
     
coll = COLLECT(*collect_items, strip=False, upx=False, name='picard')
```

picard.spec
```diff
@@ -134,62 +142,73 @@ else:
...
+            a_plugins.scripts,
+            exclude_binaries=True,
+            name='picard-plugins',
+            debug=False,
+            strip=False,
+            upx=False,
+            icon='picard.ico',
+            console=True,
+            # Windows specific
+            version='win-version-info.txt',
```
zas, line 188: Shouldn't we test for windows then? Currently that's `os_name != 'Darwin'` so that's not exactly windows-specific

zas (dismissed):
Just suggestions about the code style, but apart that LGTM

picard.spec
phw: Well, no. I just wanted to group the parameters a bit for readability, as I did the same with the macOS specific ones. But it's fine to set it here, PyInstaller ignores it on other platforms. We don't need multiple if else for different platforms.

picard.spec
phw: @zas Makes sense, fixed in latest push

zas (approved):

PR #3206, opened 2026-06-01
PICARD-3302: Allow focusing Log View dialog while Options dialog is open

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Fix dialog modality so that Log View and History View remain focusable and interactive while the Options dialog (or any other modal dialog) is open. Standardize modality handling across all dialogs.

## Problem

* JIRA ticket: PICARD-3302

When the Options dialog is open, the Log View dialog cannot receive focus. Clicking the Log View in the taskbar briefly highlights it but focus returns to the Options dialog. This happens because the Options dialog uses `ApplicationModal` modality, which blocks interaction with all other application windows, including independent top-level windows like the Log View.

## Solution

**Infrastructure changes:**
- `PicardDialog` now auto-applies `WindowModal` when a parent is given, `NonModal` otherwise. A class-level `modality` attribute allows subclasses to opt out.
- Added `show_modal()` / `show_nonmodal()` convenience methods to `PicardDialog`.
- Fixed `SingletonDialog.show_instance()` which was overriding modality with `NonModal`.
- Made `LogView` and `HistoryView` parentless windows so they are never in any modality chain. Set the app icon on `QApplication` so parentless windows inherit it, and explicitly close them on app quit.

**Per-dialog fixes:**
- Removed explicit `ApplicationModal` / `setModal(True)` from: `OptionsDialog`, `AttachedProfilesDialog`, `ScriptDetailsEditor`, `InstallPluginDialog`, `InstallConfirmDialog`, `PluginInfoDialog`, `SwitchRefDialog`, `UninstallPluginDialog`, `EditTagDialog`, `AllowRtdUpdatesDialog`, `NewUserDialog`, `SaveWarningDialog`.
- Added explicit `WindowModal` to `SetupWizard` (QWizard, not PicardDialog).
- On macOS only: `OptionsDialog` sets `Qt.WindowType.Window` flag to prevent sheet rendering (on Linux/Windows this flag breaks `WindowModal` enforcement).
- Cleaned up `.ui` file artifacts (removed `NonModal` on child widget in `passworddialog.ui`).

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed collaboratively with an AI assistant (Kiro/Claude), with human review, testing, and direction at each step.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -1311,14 +1311,11 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         options_dialog = OptionsDialog.show_instance(page, self)
         options_dialog.finished.connect(self._options_closed)
         if not modal_options():
-            self.setEnabled(False)
-            options_dialog.setEnabled(True)
+            self._disable_while_dialog_shown(options_dialog)
 
         return options_dialog
 
     def _options_closed(self):
-        if not modal_options():
-            self.setEnabled(True)
         if self.script_editor_dialog is not None:
             self.open_file_naming_script_editor()
             self.script_editor_dialog.show()
@@ -1327,6 +1324,24 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         self._make_profile_selector_menu()
         self._make_script_selector_menu()
         self._make_settings_selector_menu()
+
+    def _disable_while_dialog_shown(self, dialog):
+        """Disable MainWindow while dialog is shown, re-enable when it closes.
+
+        Connects to both `finished` (normal close path) and `destroyed`
+        (safety net if the dialog is destroyed without emitting `finished`,
+        e.g. via deleteLater or parent destruction). The isEnabled() guard
+        prevents double-action when both signals fire in sequence.
+        """
+
+        def re_enable():
+            if not self.isEnabled():
+                self.setEnabled(True)
+
+        dialog.destroyed.connect(re_enable)
+        dialog.finished.connect(re_enable)
+        self.setEnabled(False)
+        dialog.setEnabled(True)
 
     def show_help(self):
         webbrowser2.open('documentation')
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -264,6 +264,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
                 ":/images/{size}x{size}/{app_id}.png".format(size=size, app_id=PICARD_APP_ID), QtCore.QSize(size, size)
             )
         self.setWindowIcon(icon)
+        QtWidgets.QApplication.instance().setWindowIcon(icon)
 
         self.show_close_window = IS_MACOS
 
@@ -287,8 +288,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
             self.file_browser.hide()
         self.panel.insertWidget(0, self.file_browser)
 
-        self.log_dialog = LogView(self)
-        self.history_dialog = HistoryView(self)
+        self.log_dialog = LogView()
+        self.history_dialog = HistoryView()
 
         self.metadata_box = MetadataBox(parent=self)
         self.cover_art_box = CoverArtBox(parent=self)
@@ -432,6 +433,8 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
                 # Silently close the script editor without displaying the confirmation a second time.
                 self.script_editor_dialog.loading = True
         event.accept()
+        self.log_dialog.close()
+        self.history_dialog.close()
 
     def _setup_desktop_status_indicator(self):
         if DesktopStatusIndicator:
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -1303,12 +1303,12 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
     def show_options(self, page=None):
         ReadTheDocs.update_documentation_items()  # Retry updates if required
-        options_dialog = OptionsDialog.show_instance(page, self)
-        options_dialog.finished.connect(self._options_closed)
         if self.script_editor_dialog is not None:
             # Disable signal processing to avoid saving changes not processed with "Make It So!"
             for signal in self.script_editor_signals:
                 signal.disconnect()
+        options_dialog = OptionsDialog.show_instance(page, self)
+        options_dialog.finished.connect(self._options_closed)
 
         return options_dialog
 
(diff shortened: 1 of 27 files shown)


picard/ui/mainwindow/__init__.py
```diff
@@ -258,12 +260,6 @@ def suspend_while_loading_exit(self):
 
     def setupUi(self):
         self.setWindowTitle(_("MusicBrainz Picard"))
-        icon = QtGui.QIcon()
-        for size in (16, 24, 32, 48, 128, 256):
-            icon.addFile(
-                ":/images/{size}x{size}/{app_id}.png".format(size=size, app_id=PICARD_APP_ID), QtCore.QSize(size, size)
-            )
-        self.setWindowIcon(icon)
```
phw, line 266: Why does this get removed? I don't think it has any relevance on my system (Gnome doesn't make use of window icons), but wouldn't it affect systems that do show the icon in the window taskbar?

picard/ui/mainwindow/__init__.py
```diff
@@ -1302,20 +1299,23 @@ def close_active_window(self):
         self.tagger.activeWindow().close()
 
     def show_about(self):
-        return AboutDialog.show_instance(self)
+        return AboutDialog.show_instance()
```
phw: I think the about dialog should stay window modal. Having it attached to the window or a sheet on macOS would make sense to me for this case.

phw (changes requested):
I tested this only on Gnome / X11 so far, but there it works well, with one bug regarding the About dialog. Especially the script editor works much better now, and generally the concept of non-modal windows in many cases is useful. I tested the smaller dialogs for confirmations and subdialogs in options, they are still modal as expected.

I have a few comments for discussion. And when closing the main window after having opened and closed the about dialog I get:

```
Picard terminated unexpectedly
  File "/home/phw/devel/musicbrainz/picard/picard/ui/mainwindow/__init__.py", line 431, in closeEvent
    PicardDialog.close_all_parentless()
  File "/home/phw/devel/musicbrainz/picard/picard/ui/__init__.py", line 215, in close_all_parentless
    dialog.close()
RuntimeError: wrapped C/C++ object of type AboutDialog has been deleted
```

PicardDialog does not seem to unregister a dialog once it got closed.

picard/ui/mainwindow/__init__.py
zas: It was moved to application level instead, we were only registering icon for main window (and its children), but when we have parentless dialogs, they don't have an icon associated, this was to fix that (all windows/dialogs have same icon).

zas:
> 
> PicardDialog does not seem to unregister a dialog once it got closed.

I'll have a look, this is in last changes I made, perhaps missed smt. Thanks for noticing

picard/ui/mainwindow/__init__.py
zas: Well, if it is it will be disabled as main window, and since it's doing nothing special it can be parentless. But I can make it model again if you prefer.

phw: I just don't see the point for this dialog, and maintaining many separate window makes the application overall harder to use. For activities or logs the separate window is totally useful. But this is just a application wide non-interactive window

picard/ui/mainwindow/__init__.py
phw: I see, that makes sense

zas:
> ```
> Picard terminated unexpectedly
>   File "/home/phw/devel/musicbrainz/picard/picard/ui/mainwindow/__init__.py", line 431, in closeEvent
>     PicardDialog.close_all_parentless()
>   File "/home/phw/devel/musicbrainz/picard/picard/ui/__init__.py", line 215, in close_all_parentless
>     dialog.close()
> RuntimeError: wrapped C/C++ object of type AboutDialog has been deleted
> ```
> 
> PicardDialog does not seem to unregister a dialog once it got closed.

It should be addressed by https://github.com/metabrainz/picard/pull/3206/commits/e50344d89c78657e3df0024c47ea66a7b354b89a

picard/ui/mainwindow/__init__.py
zas: I reverted About dialog to modal in https://github.com/metabrainz/picard/pull/3206/commits/5ee1e7d576d578bd5060426b687d76e1d8c82f1a

julian45:
So far so good on this branch on Win11 and Fedora 44 KDE; I'll update this comment in-place as I test with more OSes. (I have macOS Sequoia and Fedora 44 KDE devices to test with)

Flows tested:
- Open Options > Options... and confirm both that window and the main window are usable (including switching focus) while open
- Go back to main window while options window is open, then *also* open the log viewer
- Confirm all three windows will remain open/usable without panics while switching focus between them
- Close main window and confirm that ancillary windows (options & logs in this case) gracefully close as well

~~Edit: partial no-joy on macOS Sequoia. Flows 2 and 3 above aren't quite possible, as the options dialog is not a fully separate window on macOS. That said, I can at least switch between the main-with-options-forcibly-superimposed window and the log viewer window without issue, and closing the main window gracefully also closes the log viewer window. Also, I'm having trouble reproducing this in new attempts, but sometime during my initial round of testing on macOS, this showed up in the console _without_ causing a panic:
`2026-06-04 20:38:15.084 python[49228:2775211] TSMSendMessageToUIServer: CFMessagePortSendRequest FAILED(-1) to send to port com.apple.tsm.uiserver`~~
(addressed by @zas [here](https://github.com/metabrainz/picard/pull/3206#issuecomment-4629616509))
Edit 2: Everything works as expected & desired on Fedora 44 KDE!

zas:
> partial no-joy on macOS Sequoia. Flows 2 and 3 above aren't quite possible, as the options dialog is not a fully separate window on macOS.

This is one of the limitations we encountered on macos, we had to make the dialog modal again on this platform, so the behavior differs from Win & Linux. It is related to the specific way macos managed windows and Preferences.



> Also, I'm having trouble reproducing this in new attempts, but sometime during my initial round of testing on macOS, this showed up in the console _without_ causing a panic:
> `2026-06-04 20:38:15.084 python[49228:2775211] TSMSendMessageToUIServer: CFMessagePortSendRequest FAILED(-1) to send to port com.apple.tsm.uiserver`

Yes, it seems a common "warning" with Qt on mac, but not sure if we can even identify the exact source, and if it is worth "fixing" it.

aredridel:
Works for me: MacOS 26.5 ARM64, Macbook Neo.

zas:
> Works for me: MacOS 26.5 ARM64, Macbook Neo.

Thank you @aredridel for the feedback, much appreciated.

phw (approved):
LGTM

PR #3219, opened 2026-06-13
PICARD-3315: Refactor options with in_profile field and plugin profile support

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

Refactor options to add explicit `in_profile` field on `Option` class, convert page `OPTIONS` from tuples to dicts, add full plugin profile support, improve developer warnings for misconfigured options, and fix several pre-existing profile bugs.

## Problem

* JIRA ticket: PICARD-3315

The current options/profiles architecture has several issues:
- No clear separation between "this option can be in a profile" vs UI concerns (highlights/widgets)
- Plugins cannot participate in the profiles system
- Profile settings for different sections could collide in the shared storage dict
- `SettingConfigSection` duplicates profile logic that should live in base `ConfigSection`
- Misconfigured options (missing titles, non-existent widgets) fail silently
- Highlight styling didn't distinguish between "tracked" and "actively overridden" options

### Core refactoring
- `Option` gains `in_profile` field — single source of truth for profile eligibility
- Page `OPTIONS` converted from tuples to dicts with `{'widgets': [...]}` format
- `SettingConfigSection` delegates profile read/write to base `ConfigSection` (removed ~50 lines of duplicated logic)
- `register_option()` gains `title` and `in_profile` parameters

### Plugin profile support
- Plugin options auto-register in profile groups under "Plugins > Plugin Name" hierarchy
- Namespaced profile keys (`plugin.<uuid>/option_name`) eliminate collision risk
- `OPTION_SECTION` attribute on pages identifies the config section (set automatically for plugins)
- `restore_defaults()` uses `OPTION_SECTION` for correct section lookup
- Profile groups removed on plugin disable, profile settings cleaned on plugin purge
- Profiles page refreshes when plugins are enabled/disabled

### Developer warnings
- Warning when `in_profile=True` but no `title` set (falls back to option name)
- Warning logged when widgets declared for an option without `in_profile=True` (widgets ignored)
- Warning when profile highlight references a non-existent widget on a page

### UI improvements
- Two-level highlights: alpha=50 (tracked in profile) vs alpha=120 (actively overridden)
- QComboBox uses palette override instead of stylesheet (preserves native rendering)
- QAbstractItemView skipped from stylesheet highlights (breaks checkable items)
- Tooltip appends profile info, preserving original tooltip across highlight passes
- `seen_widgets` set prevents duplicate tooltip lines for shared widgets

### Bug fixes
- Fixed fpcalc validation race with `_fpcalc_checking` flag
- Fixed `profile_groups_remove_group` corrupting `_known_settings` on name collision
- Fixed `_set_profile_override` bypassing `settings_override` when dialog is open
- Fixed save order: clear overrides between profiles page and other pages
- Stale profile settings cleaned on save (preserving plugin keys)

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed collaboratively using Kiro CLI with human direction on architecture, design decisions, and iterative review/testing.

## Action

Additional actions required:


picard/ui/options/tags.py
```diff
@@ -49,17 +49,17 @@ class TagsOptionsPage(OptionsPage):
...
+    OPTIONS: dict[str, dict] = {
+        'enable_tag_saving': {'widgets': ['write_tags']},
+        'preserve_timestamps': {'widgets': ['preserve_timestamps']},
+        'clear_existing_tags': {'widgets': ['clear_existing_tags']},
+        'preserve_images': {'widgets': ['preserve_images']},
+        'remove_id3_from_flac': {'widgets': ['remove_id3_from_flac']},
+        'remove_ape_from_mp3': {'widgets': ['remove_ape_from_mp3']},
+        'fix_missing_seekpoints_flac': {'widgets': ['fix_missing_seekpoints_flac']},
+        'preserved_tags': {'widgets': ['preserved_tags']},
+        'disable_date_sanitization_formats': {},
```
rdswift: This is how the page looks currently, with all settings on the page selected in the active profile:

Image

Note that the 'preserved_tags' widget is a custom widget that does not appear to support highlighting.  For that reason, in my initial PR, I had to set the highlighted widget to 'preserved_tags_label' so that something was highlighted.

I also revised the UI file to include the tag format for the sanitization checkboxes (rather than adding that section within the code) so that they could be highlighted.

rdswift: Also note that the "Cover Art Archive" and "Local Files" options do not appear in the option profile for selection.  This is also something that I had addressed in my PR.

picard/ui/options/fingerprinting.py
```diff
@@ -65,14 +65,14 @@ class FingerprintingOptionsPage(OptionsPage):
...
-        ('fpcalc_threads', None),
-    )
+    OPTIONS: dict[str, dict] = {
+        'fingerprinting_system': {},
+        'acoustid_fpcalc': {},
+        'acoustid_apikey': {},
+        'ignore_existing_acoustid_fingerprints': {},
+        'save_acoustid_fingerprints': {},
+        'fpcalc_threads': {},
+    }
```
rdswift: Suggested widgets for highlighting:

```suggestion
    OPTIONS: dict[str, dict] = {
        'fingerprinting_system': {'widgets': ['disable_fingerprinting', 'use_acoustid']},
        'acoustid_fpcalc': {},
        'acoustid_apikey': {},
        'ignore_existing_acoustid_fingerprints': {'widgets': ['ignore_existing_acoustid_fingerprints']},
        'save_acoustid_fingerprints': {'widgets': ['save_acoustid_fingerprints']},
        'fpcalc_threads': {},
    }
```

I also suggest that we include the 'ignore_existing_acoustid_fingerprints' setting in the option profiles.

zas: Addressed in https://github.com/metabrainz/picard/pull/3219/commits/d439ca142a83533095fa94cb8842f0d90a44ce32

picard/ui/options/tags.py
zas: Those should be fixed now

ui/options_tags.ui
```diff
@@ -95,21 +95,49 @@
        </spacer>
       </item>
       <item>
-       <widget class="QLabel" name="preserved_tags_label">
-        <property name="text">
+       <widget class="QGroupBox" name="preserved_tags_groupbox">
```
phw: I'm not really a fan of the nested group boxes now. Can we get rid of that again? Or maybe Move the new, currently nested groupbox, to after the "Before tagging" groupbox?

picard/ui/options/cdlookup.py
```diff
@@ -49,7 +49,9 @@ class CDLookupOptionsPage(OptionsPage):
     ACTIVE = True
     HELP_URL = "/config/options_cdlookup.html"
 
-    OPTIONS = (('cd_lookup_device', None),)
+    OPTIONS: dict[str, dict] = {
+        'cd_lookup_device': {},
```
phw: Is there any specific reason we exclude this from profiles? Or could we include it as well? Could be handy if one has multiple CD drives.

picard/ui/options/advanced.py
```diff
@@ -38,21 +38,18 @@ class AdvancedOptionsPage(OptionsPage):
...
-            ['ignore_track_duration_difference_under', 'label_track_duration_diff'],
-        ),
-        ('query_limit', ['query_limit', 'label_query_limit']),
-        ('completeness_ignore_videos', ['completeness_ignore_videos']),
-        ('completeness_ignore_pregap', ['completeness_ignore_pregap']),
-        ('completeness_ignore_data', ['completeness_ignore_data']),
-        ('completeness_ignore_silence', ['completeness_ignore_silence']),
-        ('compare_ignore_tags', ['groupBox_ignore_tags']),
-    )
+    OPTIONS: dict[str, dict] = {
```
phw: I wonder if we should define a type alias for this somewhere. Maybe even with a `TypedDict` for the dict part. Something like:

```python
from typing import (
    TypeAlias,
    TypedDict,
)


class OptionConfig(TypedDict):
    widgets: list[str]


Options: TypeAlias = dict[str, OptionConfig]
```

Not sure about proper names, though.

picard/plugin3/manager/clean.py
```diff
@@ -57,5 +60,21 @@ def _clean_plugin_config(self, plugin_uuid):
...
+        """Remove profile settings entries for a plugin."""
+        try:
+            profile_settings = config.profiles[SettingConfigSection.SETTINGS_KEY]
+        except AttributeError:
+            return
+        if not profile_settings:
+            return
+        profile_prefix = f'plugin.{plugin_uuid}/'
+        for settings in profile_settings.values():
+            keys_to_remove = [k for k in settings if k.startswith(profile_prefix)]
```
phw, line 77: This could either be a generator comprehension (`keys_to_remove = (k for k in settings if k.startswith(profile_prefix))`) or the next for loop is so simple anyway that I think it would be more readable to combine both into a single loop:

```python
for k in settings:
    if k.startswith(profile_prefix):
        del settings[k]
```

picard/plugin3/api_impl.py
```diff
@@ -193,6 +193,7 @@ def __init__(self, manifest: PluginManifest, tagger: 'Tagger') -> None:
         full_name = f'plugin.{self._manifest.uuid}'
         self._logger = getLogger(f'main.plugin.{self._manifest.module_name}')
         self._api_config = ConfigSection(get_config(), full_name)
```
phw: If we want to have plugins support profiles this should become `SettingConfigSection`.

picard/config.py
```diff
@@ -292,7 +408,19 @@ def register_option(self, name: str, default: ConfigValueType) -> Option:
...
 
-        return option_type(self.__name, name, default)
+        opt = option_type(self.__name, name, default, title=title, in_profile=in_profile)
+        if in_profile:
+            group_name = self.__name
+            group_title = self.display_name or self.__name
+            profile_groups_add_setting(
+                group_name,
+                name,
+                (),
```
phw, line 447: A list is expected here (`[]`).

picard/config.py
```diff
@@ -205,9 +222,69 @@ def __getitem__(self, name: str) -> Any:
         opt = Option.get(self.__name, name)
         if opt is None:
             return None
+        if opt.in_profile:
+            override = self._get_profile_override(name)
+            if override is not _SENTINEL:
+                return override
```
phw, line 312: I don't fully understand this part. Previously we had `ConfigSection` as the base class, without the profile specific part, and `SettingConfigSection` which added all the profile handling (and the potential overhead it brings). Maybe `SettingConfigSection` is not the best name, but its description still says:

> Custom subclass to automatically accommodate saving and retrieving values based on user profile settings.

Now part of the profile functionality gets moved over to `ConfigSection`, and parts remains in `SettingConfigSection`.

I think we should either maintain the separation (which I'm in favor of) or completely merge the classes. But the current separation IMHO makes no sense anymore.

phw (changes requested):
This is difficult to review for me as I'm personally not a heavy user of option profiles. From testing this it seems to work well, though. And overall it makes things more consistent and options better maintainable. Allowing plugins to opt-in into this also makes a lot of sense.

A couple of comments for things I noticed in the code, though.

picard/ui/options/cdlookup.py
zas: We could include it if you think it's useful, it wasn't before but it can be easily changed.

ui/options_tags.ui
zas: Without it, the profile highlighting doesn't work well, so that's totally intentional.

picard/plugin3/manager/clean.py
zas: We can't mutate the dict (due to `del`), and dropping the intermediate variable isn't really more readable, so I kept this as is.

picard/ui/options/cdlookup.py
zas: Done in https://github.com/metabrainz/picard/pull/3219/commits/9ff85632eeb2019f67aa451ff90f9835e2363ce9

picard/config.py
zas: Done in https://github.com/metabrainz/picard/pull/3219/commits/691e4d6a7b63261aaa48fa91ad24697f2136c7f0

picard/ui/options/advanced.py
zas: Done in https://github.com/metabrainz/picard/pull/3219/commits/b33ecbcb2f111df478ff6e4eae09d98da224e717

Check if it is what you had in mind

picard/config.py
zas: Addressed by https://github.com/metabrainz/picard/pull/3219/commits/a351e50e7f0b94fff79ec455ea03b66d8e5df82f

picard/plugin3/api_impl.py
zas: Addressed by https://github.com/metabrainz/picard/pull/3219/commits/a351e50e7f0b94fff79ec455ea03b66d8e5df82f

docs/PLUGINSV3/API.md
```diff
@@ -241,12 +241,57 @@ class MyOptionsPage(OptionsPage):
...
+
+Plugin options can participate in Picard's user profiles system. When an option is profile-eligible, users can override its value per-profile from the Profiles options page.
+
+To make an option profile-eligible, pass `title` and `in_profile=True` to `register_option`:
+
+```python
+def enable(api):
+    api.plugin_config.register_option(
+        'greeting', 'Hello World',
+        title='Greeting message',
```
rdswift, line 260: Am I correct in assuming that we can also use a translated title here?  Something like:

```python
title=api.tr('opt.title.greeting', 'Greeting message'),
```

picard/ui/options/advanced.py
phw: Yes, perfect. Also good naming. That way we can easily extend the definition if needed and have some type safety check when defining new entries.

picard/config.py
phw: @zas When I wrote my comment the third parameter of `profile_groups_add_setting` was typed as `list[str]`. But since https://github.com/zas/picard/commit/6553b9edc211cce895898693aaad4960ae9f6815 is is now typed as `tuple`, so this is wrong again 😛

What about changing the typedef to `Iterable[str]`?

```python
def profile_groups_add_setting(
    group: str,
    option_name: str,
    highlights: Iterable[str],
    title: str | None = None,
    parent: str | None = None,
    section: str = 'setting',
) -> None:
    ...
```

ui/options_tags.ui
phw: @zas Please see https://github.com/zas/picard/pull/29

picard/config.py
zas: Done in https://github.com/metabrainz/picard/pull/3219/commits/6394bda7ec5bcfc1619e94a466966a30b80309a5 and I dropped the change to `[]` then rebased.

picard/config.py
phw: @zas I'm not fully sold to the three class approach. `SettingConfigSection` implements the methods to set the `settings_overrides`, but `ProfileConfigSection` already uses those.

Also regarding plugin options, shouldn't the overrides also be applied there?

I like the name of `ProfileConfigSection`, though. Let's keep that.

zas: The 3 class approach was looking fine to me, but I'm not sure to understand what you expect here, perhaps just make a small PR over this branch (faster than explaining often), and if it works I'm not against ;)

phw: Ok, I'll look into this

phw: @zas I misunderstood how `ProfileConfigSection` makes use of `settings_override` in the settings section. After looking over the code more closely this makes sense now.

picard/plugin3/api_impl.py
```diff
@@ -104,7 +104,10 @@
 )
 from picard.webservice.api_helpers import MBAPIHelper
 
-from picard.ui.options import OptionsPage as _OptionsPage
+from picard.ui.options import (
+    OptionsPage as _OptionsPage,
+    PageOptionConfigs,  # noqa: F401 (re-exported via plugin3.api)
```
phw: As this is not used here it could also be just directly be imported in `api.py`. It is not mandatory to import everything in `api.py` from `api_impl.py`, `api_impl.py` is only there to prevent other stuff getting exported.

phw (changes requested):
A very minor change request.

Otherwise I reviewed the code and tested the changes, also with a plugin. As I said before I'm myself not a heavy profile user, so I have not that much usage experience and might miss something. But as far I as I could tell this behaves as expected. Settings get properly stored to configured profiles, turning off a profile makes the original value visible again, resetting settings works.

picard/plugin3/api_impl.py
zas: Fixed in https://github.com/metabrainz/picard/pull/3219/commits/78ffc6523bc72dca59dd8fe4b1f90f343e96d192

phw (approved):

phw:
Great work

PR #3223, opened 2026-06-18
Refactor tagger_instance helper to use Tagger.instance

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

This refactors the `tagger_instance` helper to directly use `Tagger.instance`.

Advantage: We avoid the assert and the trouble this can cause when trying to mock the tagger instance.

Why not have code use `Tagger.instance` directly? The `tagger_instance` is intentionally placed at top level and importable everywhere in the code. Importing `Tagger` often causes circular dependencies.

## Solution

- Rework `Tagger.instance` to avoid typing warnings due to invalid override of `QCoreApplication.instance`
- Change `tagger_instance` implementation to directly return `Tagger.instance()`

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/__init__.py
+++ b/picard/__init__.py
@@ -72,13 +72,9 @@ def tagger_instance() -> 'Tagger':
 
     Tagger is the main application instance and a subclass of QApplication.
     """
-    from PyQt6.QtCore import QCoreApplication  # avoid top level Qt import
-
     from picard.tagger import Tagger  # avoid circular imports
 
-    instance = QCoreApplication.instance()
-    assert isinstance(instance, Tagger), 'Expected an instance of Tagger'
-    return instance
+    return Tagger.instance()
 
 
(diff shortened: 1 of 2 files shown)


zas (dismissed):
We could default `__instance` to `None` as it was, but it doesn't change anything in practice (it still fails if used too early).

phw:
> We could default __instance to None as it was, but it doesn't change anything in practice (it still fails if used too early).

Not defaulting to None makes this easier to define with type hints. For our case it is supposed to be set very early on, and trying to use it before the tagger was instantiated will hard fail. I think that's fine

zas (approved):

PR #3226, opened 2026-06-19
PICARD-3308: allow setting collation algorithm used

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3308

This is a preparation for debugging PICARD-3308 (sorting broken on macOS), but does not in itself resolve the issue.

Due to various bugs with sorting on different platforms Picard currently implement two different collation algorithms for locale aware sorting. It can either use `locale.strxfrm` or Qt's `QCollator`.

`strxfrm` is the current default on Windows, since QCollator used to be broken in the Windows build (see PICARD-2953 and https://qt-project.atlassian.net/browse/QTBUG-88704).

On the other hand we had issues with `strxfrm` causing crashes on macOS for some locales (PICARD-2914) and also on Haiku (fixed on the OS side since then) and some other sorting issues. So currently the default for all other platforms is using `QCollator`.

Right now sorting does not work at all on macOS, at least not on ARM builds, this is PICARD-3308.

In order to be able to test the various implementations with a binary build and to give users some workaround if sorting is broken this PR introduces support for a `PICARD_COLLATOR` environment variable, which can be set to one of three values:

- `strxfrm`: Use `locale.strxfrm`
- `qt`: Use `QCollator`
- `string`: No locale aware collation, just sort by normal string comparison

Without `PICARD_COLLATOR` being set or set to an unknown value the platform specific default will be used as before.

This PR also refactors the previous `i18n.py` into a `i18n` package. 

## Solution

- Add support for defining the collation algorithm to use by setting the `PICARD_COLLATOR` environment variable
- Refactor the previous `i18n.py` into a `i18n` package, separate the gettext and collation code into separate files
- Move the `Transators` and `Translator` classes, that handle setting up `QTranslator` instances, from `tagger.py` to `i18n/qt.py`.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/i18n/collate.py
+++ b/picard/i18n/collate.py
@@ -114,9 +114,14 @@ def _strxfrm(string: str) -> str:
         return string.lower()
 
 
+def _sort_key_string(string: str, numeric: bool = False) -> Comparable:
+    return string
+
+
 AVAILABLE_COLLATORS = {
     'strxfrm': _sort_key_strxfrm,
     'qt': _sort_key_qt,
+    'string': _sort_key_string,
 }
 
 # QCollator.sortKey is broken on Windows in some Qt builds,
--- a/picard/i18n/collate.py
+++ b/picard/i18n/collate.py
@@ -1,0 +1,129 @@
+# -*- coding: utf-8 -*-
+#
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2024, 2026 Philipp Wolfer
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, write to the Free Software
+# Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
+
+from collections.abc import Callable
+import locale
+import os
+import re
+from typing import Protocol
+
+from PyQt6.QtCore import QCollator
+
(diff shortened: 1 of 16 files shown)


rdswift (approved):
Looks okay to me.

picard/i18n/collate.py
```diff
@@ -0,0 +1,134 @@
...
+from picard.const.sys import (
+    IS_MACOS,
+    IS_WIN,
+)
+
+
+_qcollator = QCollator()
+_qcollator_numeric = QCollator()
+_qcollator_numeric.setNumericMode(True)
+
```
zas, line 38: These are created at import time (before setup_collator is called), then potentially re-created inside `setup_collator` if the active collator is 'qt'
On import, `QLocale.setDefault() `hasn't been called yet, so these initial `QCollator` instances use the system default  locale.
They're then replaced in `setup_collator` after locale is configured but only if `ACTIVE_COLLATOR == 'qt'`.
  
If we import `sort_key` before` setup_i18n()` is called (e.g. at module level elsewhere), they'd get a collator with the wrong locale.
The old code had the same issue but recreated collators unconditionally inside `setup_gettext`.

Not sure what we should do, or even if it is a real issue, but I think we should avoid module-level initialization in general if we can avoid them, especially since we may want to allow changing locale during runtime at some point (without requiring a restart).

zas (approved):

picard/i18n/collate.py
phw: I don't think it is a real issue, using the setup function is required before use. The condition has no impact here, because if that's false the collator is not used.

We can also initially set them to None. Might be some tests then need to be adjusted to initialize the i18n beforehand.

phw: Upon reviewing the code I decided to keep it as is. The current behavior means that just using `_sort_key_qt` without running `setup_gettext` works, but uses the default locale. This is consistent to how `_sort_key_strxfrm` also works and gives a predictable behavior.

The code is supposed to run `setup_collator` after the default locales have been changed, which is done by `setup_i18n` calling `setup_gettext` beforehand.

PR #3229, opened 2026-06-20
PICARD-3322: Display all profiles in the Attached Profiles dialog

rdswift:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

The title on the Attached Profiles dialog displayed from the options pages shows as "Profiles Attached to Options in {page title} Section".  This is incorrect because the dialog displays only the currectly active profiles, regardless of whether or not they are actually attached to any settings on the page.  If a profile that is attached to settings on the page isn't currently set as active, it will not be displayed in the dialog.

* JIRA ticket (_optional_): PICARD-3322

## Solution

Rather than trying to find an appropriate title, because the Attached Profiles dialog now allows settings on the page to be added to or removed from profiles, the dialog was modified to display all profiles (regardless of whether or not they are currently active).  In addition, ensure that the dialog is disabled if there are no profiles defined.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -988,13 +988,6 @@ class AttachedProfilesDialog(PicardDialog):
                 return QtCore.Qt.CheckState.Checked
             if in_count > 0:
                 return QtCore.Qt.CheckState.PartiallyChecked
-        # Check if any other enabled profile has it
-        selected_set = set(selected_ids)
-        for profile in self._enabled_profiles:
-            if profile['id'] in selected_set:
-                continue
-            if pkey in self.profile_page.profile_settings.get(profile['id'], {}):
-                return QtCore.Qt.CheckState.PartiallyChecked
         return QtCore.Qt.CheckState.Unchecked
 
     def _build_tooltip(self, pkey: str) -> str:
--- a/picard/ui/options/dialog.py
+++ b/picard/ui/options/dialog.py
@@ -559,11 +559,9 @@ class OptionsDialog(PicardDialog, SingletonDialog):
 
     def set_profiles_button_and_highlight(self, page):
         option_group = profile_groups_group_from_page(page)
-        enabled = False
-        if option_group:
-            working_profiles, _ = self.get_working_profile_data()
-            enabled = any(p['enabled'] for p in working_profiles)
-        self.ui.attached_profiles_button.setEnabled(enabled)
+        defined_profiles = self.get_page('profiles')._clean_and_get_all_profiles()
+        # Only enable the dialog button if there are profiles defined and an option group on the page
+        self.ui.attached_profiles_button.setEnabled(bool(option_group) and bool(defined_profiles))
         self.update_profile_save_warning(page)
 
     def update_profile_save_warning(self, page):
@@ -883,8 +881,8 @@ class AttachedProfilesDialog(PicardDialog):
 
         self.setWindowTitle(_("Profiles Attached to Options in %s Section") % self.option_group['title'])
 
-        # Only show enabled profiles
-        self._enabled_profiles = [p for p in self.profile_page._clean_and_get_all_profiles() if p['enabled']]
+        # Show all profiles because even disabled profiles can be attached to options
+        self._enabled_profiles = self.profile_page._clean_and_get_all_profiles()
 
         self._populate_profile_list()
         self.ui.profile_list.setIndentation(0)
@@ -921,7 +919,11 @@ class AttachedProfilesDialog(PicardDialog):
         marker_height = fm.boundingRect(self.MARKER_ACTIVE).height()
         marker_size = QtCore.QSize(marker_width, marker_height)
         for profile in self._enabled_profiles:
-            item = QtWidgets.QTreeWidgetItem([self.MARKER_INACTIVE, profile['title']])
+            # Identify profiles that are disabled but still attached to options with the "(disabled)" suffix,
+            # but keep them selectable so users can manage their attached options from the page.
+            item = QtWidgets.QTreeWidgetItem(
+                [self.MARKER_INACTIVE, profile['title'] + (_(" (disabled)") if not profile['enabled'] else "")]
+            )
             item.setData(0, QtCore.Qt.ItemDataRole.UserRole, profile['id'])
             item.setSizeHint(0, marker_size)
             self.ui.profile_list.addTopLevelItem(item)


picard/ui/options/dialog.py
```diff
@@ -986,13 +988,6 @@ def _compute_check_state(self, pkey: str, selected_ids: list[str]) -> QtCore.Qt.
                 return QtCore.Qt.CheckState.Checked
             if in_count > 0:
                 return QtCore.Qt.CheckState.PartiallyChecked
-        # Check if any other enabled profile has it
-        selected_set = set(selected_ids)
-        for profile in self._enabled_profiles:
-            if profile['id'] in selected_set:
-                continue
-            if pkey in self.profile_page.profile_settings.get(profile['id'], {}):
-                return QtCore.Qt.CheckState.PartiallyChecked
```
zas, line 995: This seems to me this change is a regression.
The "other profiles" `PartiallyChecked` state was correct and intentional, especially given that the `_build_tooltip` method relies on it to show which profiles contain the option.
What's the idea behind this change?

picard/ui/options/dialog.py
```diff
@@ -921,7 +919,11 @@ def _populate_profile_list(self) -> None:
         marker_height = fm.boundingRect(self.MARKER_ACTIVE).height()
         marker_size = QtCore.QSize(marker_width, marker_height)
         for profile in self._enabled_profiles:
-            item = QtWidgets.QTreeWidgetItem([self.MARKER_INACTIVE, profile['title']])
+            # Identify profiles that are disabled but still attached to options with the "(disabled)" suffix,
+            # but keep them selectable so users can manage their attached options from the page.
+            item = QtWidgets.QTreeWidgetItem(
+                [self.MARKER_INACTIVE, profile['title'] + (_(" (disabled)") if not profile['enabled'] else "")]
```
zas, line 925: We should avoid concatenation for i18n, better use:

`_("%s (disabled)") % profile['title'] if not profile['enabled'] else profile['title']`

picard/ui/options/dialog.py
```diff
@@ -883,8 +881,8 @@ def __init__(self, option_group: dict, profile_page: 'ProfilesOptionsPage', pare
 
         self.setWindowTitle(_("Profiles Attached to Options in %s Section") % self.option_group['title'])
 
-        # Only show enabled profiles
-        self._enabled_profiles = [p for p in self.profile_page._clean_and_get_all_profiles() if p['enabled']]
+        # Show all profiles because even disabled profiles can be attached to options
+        self._enabled_profiles = self.profile_page._clean_and_get_all_profiles()
```
zas, line 885: `_enabled_profiles` should be renamed since it now contains all profiles

picard/ui/options/dialog.py
zas, line 937: This comment is now incorrect with those changes.
The list now includes disabled profiles too, which may change behavior in single-profile mode. If there's exactly one profile and it's disabled, the dialog will still show it and allow interaction.

picard/ui/options/dialog.py
```diff
@@ -559,11 +559,9 @@ def get_page(self, pagename):
 
     def set_profiles_button_and_highlight(self, page):
         option_group = profile_groups_group_from_page(page)
-        enabled = False
-        if option_group:
-            working_profiles, _ = self.get_working_profile_data()
-            enabled = any(p['enabled'] for p in working_profiles)
-        self.ui.attached_profiles_button.setEnabled(enabled)
+        defined_profiles = self.get_page('profiles')._clean_and_get_all_profiles()
```
zas, line 562: This is accessing a private method from another page.
Also `_clean_and_get_all_profiles() ` is not just a getter, it mutates state by adding missing profile settings and removing orphaned ones.
This method is called every time a page is selected in the options dialog (via `set_profiles_button_and_highlight`).
Previously the method was only called when opening the `AttachedProfilesDialog`.
Calling it on  every page navigation may be fine in practice (it's idempotent after the first  call), but it's worth confirming this doesn't cause performance issues or  unexpected state changes if the profile page hasn't been loaded yet.

zas (changes requested):
Why fixing the dialog title requires to change the whole logic?

Displaying disabled profiles isn't needed, they are disabled and therefore do not apply to any option (=exactly the same if they didn't exist). 
What is the motivation for this change?

If we want to change the overall logic, all side effects have to be handled, variables should be renamed, and comments updated.

phw:
I tried the PR and compared to previous behavior. I personally really have no strong opinion on whether the dialog should show only active profiles or all. I use myself the profile feature not that much.

But I tend toward keeping the behavior of only showing active ones, as I think this makes it less complex. For full management of the profiles and included options there is the dedicated profile management page. But inside individual option pages you already have the highlighting that focuses on active profiles. Hence I think also the attached profiles dialog can focus on this part.

rdswift:
As I see it, there are two things that make the "Attached Profiles" dialog confusing for the user, and possibly misleading.

**Concern 1**

As a user, based on the title I would expect to see only the profiles that have settings on the page that are linked to the profile.  This is not the case in two ways:
1. The list contains profiles that are not linked to any of the settings on the page (but appear because they are currently active).
2. The list does not contain profiles that are linked to settings on the page but are not currently active.

Consider the scenario where there is one setting on the page, and there are five profiles as:

|   | Profile 1 | Profile 2 | Profile 3 | Profile 4 | Profile 5 |
| :--- | :---: | :---: | :---: | :---: | :---: |
| Active: | Yes | Yes | Yes | No | No |
| Linked to setting: | Yes | No | No | Yes | No |

Currently Profiles 1, 2 and 3 will be displayed, even though Profiles 2 and 3 are not linked to the setting on the page, and Profile 4 is not displayed even though it is linked to the page's setting.

Because the dialog now includes the functionality of linking/unlinking a setting on the page with a profile, I can sort of understand showing profiles that are not linked to any settngs on the page because the user can now add the settings to a profile where they are not already linked.  In our scenario, the user can change the linked state between the setting and Profiles 1, 2 and 3 from this dialog, but cannot change the linked state for Profiles 4 and 5. 

I believe that the current functionality to allow changing the setting/profile links from the dialog on the option page is an overall improvement, because it allows the user to directly manage which profiles are linked to the settings on that page, without having to return to the Profiles page.  What I don't understand is why we are restricting that functionality only to profiles that are currently active.  My thinking is that if we are going to allow management of links between profiles and settings on the page (which again I think is a good thing), it makes sense to allow this consistently with all defined profiles, not just the ones that are currently marked as active.


**Concern 2**

When the user selects one or more profiles in the list, the dialog does not consistently show the linked status of the settings as they apply to the selected profiles.  Consider the scenario where there is one setting on the page, and there are five profiles as:

|   | Profile 1 | Profile 2 | Profile 3 | Profile 4 | Profile 5 |
| :--- | :---: | :---: | :---: | :---: | :---: |
| Active: | Yes | Yes | Yes | No | No |
| Linked to setting: | Yes | No | No | Yes | No |

Currently only Profiles 1, 2 and 3 are displayed.  The linked state of the setting is displayed as follows based on which profiles are selected in the list:

| Selected Profile(s) | Displayed Linked State | Expected Linked State | Comment |
| :--- | :---: | :---: | :--- |
| 1 | Full | Full | Properly shows the setting linked to the selected profile |
| 1, 2 | Partial | Partial | Properly shows the setting linked to some of the selected profiles |
| 1, 2, 3 | Partial | Partial | Properly shows the setting linked to some of the selected profiles |
| 1, 3 | Partial | Partial | Properly shows the setting linked to some of the selected profiles |
| 2 | Partial | None | Setting is not linked to the selected profile but shows as partial rather than none |
| 2, 3 | Partial | None | Setting is not linked to the selected profiles but shows as partial rather than none |
| 3 | Partial | None | Setting is not linked to the selected profile but shows as partial rather than none |

Showing a linked state of Partial when the setting is not linked to any of the selected profiles just seems wrong to me, and could be confusing or misleading to the user.


**Suggested Changes**

In order to address Concern 1, it is suggested that all defined profiles be displayed (with an indication of whether the profile is currently active).  This will clearly indicate all of the profiles which are linked to the settings on the page (regardless whether or not they are active), and allow linking/unlinking a setting from any of the defined profiles.

In order to address Concern 2, it is suggested that the link status displayed properly reflect the linked states of only the selected profiles.

**Next Steps**

If there is no agreement to the suggested changes, the PR and corresponding ticket should be closed as "Won't Fix".

If there is agreement, then the code concerns noted in the reviews will have to be addressed.  Additionally, the ticket and PR title should be modified to reflect the discussion which extends beyond simply a misleading title of the dialog.

zas:
**Concern 1:**

You assume the title is incorrect, I actually fail to understand why, at best, it is a bit misleading.

The button says "Attached Profiles" and the dialog title says "Profiles Attached to Options in X Section". The dialog shows all active profiles, and for each option on the page it shows whether that option is attached (linked) to those profiles.
The ★/☆ markers on profiles and the checkboxes on options show exactly which options are attached to which  profiles.
  
"Attached" describes the relationship between options and profiles and that's exactly what the dialog displays and lets you manage. The profiles shown are the active ones (because inactive profiles are irrelevant, they don't attach to anything).

In any case, the correct way to address the ticket is to change the title to something better, rather than incorrectly changing the code logic to fit a misleading title.

I agree the title is a bit misleading, as the dialog actually shows all active profiles regardless of whether they have any attachment to this page's options. A profile with zero options linked on this page still appears (but that's to be able to attach them).

A more accurate title would be something like "Profile Options for X" or simply "Profiles — X", framing it as a management view rather than a filtered list.
  
That said, it's a minor wording issue, not a functional problem.

**Concern 2:**

_This isn't a bug._ It's an intentional tri-state design where "Partial" means "this option is managed by another active profile" (with a tooltip explaining which one). The behavior is correct and useful.


@rdswift: I think you are expecting the checkbox to reflect only the selected profile's state, but the design _purposely_ gives a global view.

phw:
@rdswift @zas

About concern 2, I didn't notice that it is possible to select multiple profiles on the left of the dialog. I think we should just limit it to select one profile at a time.

For concern 1 I think only showing the active ones is consistent with the way we also highlight only active profiles on individual setting pages. Full profile management can be done on the profile page, but when I'm on a individual options page I see the profiles that are currently active and will be saved to.

zas:
> About concern 2, I didn't notice that it is possible to select multiple profiles on the left of the dialog. I think we should just limit it to select one profile at a time.

It would remove the possibility to enable or disable one option in multiple profiles at once. It was designed like this specifically to allow quick changes: if you have 3 profiles and want to enable or disable one option in each, that would be select 1, click the checkbox, select 2, click the checkbox, select 3, click the checkbox (6 clicks), while with this you ctrl or shift + click once to select all profiles and click once on the checkbox. That's quite usual and expected.

zas:
@rdswift @phw  we should discuss again about this.

I prefer to close this PR for now, we can re-open if we find a correct way to improve that, probably post 3.0.

PR #3231, opened 2026-06-21
PICARD-3328: Support TOML-based exportable/importable profiles

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Implement shareable profiles: TOML-based profile export/import for sharing configurations on forums and backing up profiles. Includes CLI (`picard-cli profiles`), GUI buttons, version-aware settings upgrades on import, and a comprehensive design document.

## Problem

* JIRA ticket: PICARD-3328

Users of media servers like Navidrome, Plex, or Jellyfin need specific Picard configurations (scripts + settings) that are currently scattered across forum posts and documentation. There's no way to share or import a complete profile configuration.

### New `picard/profiles/` package:

- **`exporter.py`** — `export_profile()`: exports a profile as self-contained TOML with settings, naming script, and tagger scripts. Share mode excludes sensitive data; backup mode includes everything + profile UUID.
- **`importer.py`** — `import_profile()`: imports a TOML file, validates settings, registers scripts, handles collisions, deduplicates tagger scripts, supports `--replace` to update existing profiles.
- **`settings_upgrades.py`** — Version-keyed transform registry that upgrades old option names/formats in imported profiles (leverages PICARD-3330 helpers).
- **`cli.py`** — `picard-cli profiles` subcommand with `list`, `export`, `import` verbs. Supports partial matching on title/UUID, colored output, and confirmation prompts (overwrite/replace).

### Other changes:

- `shareable` flag on `Option` (9 options marked as non-shareable: credentials, paths, platform-specific)
- `tomlkit` promoted to runtime dependency
- Export/Import buttons on profiles options page with replace detection
- Settings upgrade registry: infrastructure only (no old-version hooks needed since profiles can only be exported from this version forward)
- Design document: `docs/ShareableProfiles.md`

### CLI usage:

```bash
picard-cli profiles list
picard-cli profiles export "Navidrome" -o navidrome.toml
picard-cli profiles import navidrome.toml --enable
picard-cli profiles import updated.toml --replace "Navidrome"
```

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Design analysis, code implementation, tests, and documentation were developed with AI assistance (Kiro CLI). All decisions were reviewed and directed by the developer.

## Action

Additional actions required:


docs/ExportableProfiles.md
```diff
@@ -0,0 +1,477 @@
...
+
+## Design Decisions
+
+### Why TOML?
+
+**Decision:** Use TOML as the profile format.
+
+**Rationale:**
+- Already used in the project (plugin `MANIFEST.toml`, `pyproject.toml`)
+- `tomlkit` (comment-preserving read/write) is already a dependency
```
phw: Not quite true, it is not yet a dependency, except for development scripts. If we need write support, I'd be not opposed to include it, though.

But I also would consider then to get rid of yaml and change the format for the script export also to toml.

docs/ShareableProfiles.md
```diff
@@ -0,0 +1,477 @@
...
+and GitHub gists. Users must manually piece together scripts and settings, which
+is error-prone and discouraging for newcomers.
+
+### Real-World Example: Navidrome
+
+Navidrome's [tagging documentation](https://www.navidrome.org/docs/usage/library/tagging/#picard-specific-tips)
+recommends specific Picard scripts for multi-valued artist tags, date handling,
+and album version metadata. Their docs also recommend:
+
+- Using standardized artist names (to prevent duplicate entries)
```
phw, line 50: We could consider making this the default

docs/ExportableProfiles.md
zas: I updated the document, it seems we can't totally get rid of YAML, but at least we could make a wider use of TOML.

docs/ShareableProfiles.md
zas: Yes, but it has no impact on shareable profiles (we still need to be able to override an option if it was modified). It could be done separately though.

docs/ExportableProfiles.md
phw: Yes, agreed. We need to keep the YAML support around at least for compatibility to be able to import old data.

phw (commented):
Overall I think this reads sensible. The feature is definitely useful, and would cause wider use of the profile feature.

The details have to be figured out. E.g. it's important that only relevant settings are exported. But as I understand this is handled by the profile.

zas:
> But as I understand this is handled by the profile.

Only options attached to the profile are exported (and only if marked as shareable if that's not a "backup").

phw (commented):
From the code side I find the settings_upgrades basically duplicates config_upgrade. I'm not sure how this can be fully avoided, but maybe we need to just extend the hooks that they always get the settings to modify passed. The current hooks either change the sections `settings` or `persist`. We could see whether it is possible to see that the hooks always got those settings passed an manipulate it there. Then it could probably be reused for profiles as well.

Also the PR here introduces profile upgrade hooks for old versions. This won't be necessary, nobody will have a profile export for those. We only need those going forward.

phw:
Import and export seem to work well. Here is an exported profile for the Navidrome recommendations:

```toml
# Picard Profile
# https://picard.musicbrainz.org/

[profile]
title = "Navidrome"
picard_version = "3.0.0.beta5"
created = "2026-07-03"

[settings]
save_images_to_files = true
save_images_to_tags = true
standardize_artists = true

[[scripts.tagging]]
title = "Navidrome: Multiple Artists"
script = '''$setmulti(albumartists,%_albumartists%)
$setmulti(albumartistssort,%_albumartists_sort%)
$setmulti(artistssort,%_artists_sort%)'''

[[scripts.tagging]]
title = "Navidrome: Album Version"
script = '''$set(musicbrainz_albumcomment,%_releasecomment%)
$if(%_recordingcomment%, $set(subtitle,%_recordingcomment%))
'''

[[scripts.tagging]]
title = "Navidrome: Release and Original dates"
script = '''$set(releasedate,%date%)
$set(date,%_recording_firstreleasedate%)
$set(originaldate,%originaldate%)
$delete(originalyear)
'''
```

One thing is a bit annoying with profiles for scripts: You only get to use the full set of scripts for one profile. It would be cool if individual scripts could be put into profiles. But that's a different story I think :)

zas:
> Also the PR here introduces profile upgrade hooks for old versions. This won't be necessary, nobody will have a profile export for those. We only need those going forward.

Yes, I used those to figure out how to implement profile upgrade hooks, I'll remove them.

docs/ShareableProfiles.md
```diff
@@ -0,0 +1,1016 @@
...
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `title` | string | yes | Display name for the profile |
+| `description` | string | no | What this profile does and why |
+| `author` | string | no | Who created/maintains it |
+| `version` | string | no | Profile version (for user reference) |
+| `created` | string | no | ISO date of creation |
+| `picard_version` | string | yes | Picard version that generated this file (serves as format identifier) |
+| `picard_version_min` | string | no | Minimum Picard version required to use this profile |
```
phw: I understand the intention behind `picard_version_min`, but I don't see how this would be set. What determines the `picard_version_min`?

docs/ShareableProfiles.md
```diff
@@ -0,0 +1,1016 @@
...
+In backup mode, all profile options are included.
+Unknown keys are silently ignored on import.
+
+#### `[scripts.naming]` — File Naming Script (optional)
+
+| Field | Type | Required | Description |
+|-------|------|----------|-------------|
+| `id` | string | no | Script ID (generated if missing) |
+| `title` | string | yes | Display name |
+| `script` | string | yes | The naming script content |
```
phw, line 383: Naming script, when exported separately, have additional metadata (that the user can edit). I think this should be included in the profile export to allow full restoration. These are: author, version, description, license and last update time.

Exported YAML files also include the `script_language_version`. This is not editable, but only the value of `SCRIPT_LANGUAGE_VERSION` at the time of export. Not sure how relevant this is here.

docs/ShareableProfiles.md
zas: It was in the initial design, but since we added `picard_version` so it's obsolete, I'll remove it (and the associated code), good catch.

docs/ShareableProfiles.md
zas: Addressed and force-pushed.

zas:
> : You only get to use the full set of scripts for one profile. It would be cool if individual scripts could be put into profiles.

In share mode, only enabled scripts are exported

picard/profiles/exporter.py
```diff
@@ -0,0 +1,230 @@
...
+
+        # Skip script-related options (handled separately)
+        if key in _SCRIPT_OPTIONS:
+            continue
+
+        # Handle plugin options separately
+        if is_plugin_profile_key(key):
+            # Plugin settings are only exported if explicitly requested
+            # (mode doesn't apply here — they're always opt-in)
+            continue
```
phw, line 111: I don't fully understand this part. According to the comment above plugin options are only exported if explicitly opted in. But I don't see a way to opt-in to the export, they always seem to get skipped. What do I miss?

picard/profiles/exporter.py
```diff
@@ -0,0 +1,230 @@
...
+    if isinstance(value, Enum):
+        return value.value
+    if isinstance(value, dict):
+        table = tomlkit.inline_table()
+        table.update(value)
+        return table
+    if isinstance(value, tuple):
+        return list(value)
+    if isinstance(value, list):
+        return [_export_value(item) for item in value]
```
phw, line 147: A `tuple` gets converted to a `list`, without further processing. But an actual `list` gets the values transformed.

Should this be changed to:

```python
if isinstance(value, tuple) or isinstance(value, list):
    return [_export_value(item) for item in value]
```

Also should transformation of individual values allso be applied to dicts?

picard/profiles/importer.py
```diff
@@ -0,0 +1,368 @@
...
+
+    return result
+
+
+def _make_unique_title(config: Config, title: str) -> str:
+    """Ensure the profile title is unique, appending (copy) if needed."""
+    existing_titles = {p['title'] for p in config.profiles['user_profiles']}
+    if title not in existing_titles:
+        return title
+    candidate = f"{title} (copy)"
```
phw: Copy is actually misleading here. If I import a profile from someone else and the name happens to conflict with one of my profiles that doesn't make the other a copy, they might be totally unrelated. Hence I would only append numbers.

phw (changes requested):

picard/profiles/exporter.py
zas: Actually I tried to implement those, but finally went back, so the comment is now misleading, I'll fix it.

picard/profiles/exporter.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3231/commits/02f5921df8aadd64b11f694d607896d36f783d9c

picard/profiles/importer.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3231/commits/a84c843aa1daa8c893542ea94e1ddabf941b0cc5

picard/profiles/exporter.py
zas: Clarified in https://github.com/metabrainz/picard/pull/3231/commits/79bf86df1d11095f1a966ef0fc05d7697f6a6b4f

picard/profiles/importer.py
```diff
@@ -0,0 +1,366 @@
...
+                "%d duplicate script skipped (same title and content already exist).",
+                "%d duplicate scripts skipped (same title and content already exist).",
+                duplicate_count,
+            )
+            % duplicate_count
+        )
+
+    if existing_scripts:
+        profile_settings['list_of_scripts'] = existing_scripts
+        profile_settings['enable_tagger_scripts'] = True
```
phw: Do we rally want to do this? `enable_tagger_scripts` itself is an exportable option, so I think it should be up to the profile whether it is enabled or not. A profile might provide script, but intentionally come with scripting disabled.

picard/profiles/importer.py
```diff
@@ -0,0 +1,366 @@
...
+
+def _import_tagger_scripts(
+    config: Config,
+    profile_settings: dict,
+    tagging_section: list,
+    result: ProfileImportResult,
+):
+    """Import tagger scripts into the profile settings."""
+    # Get existing scripts from the profile (if any) or start fresh
+    existing_scripts = profile_settings.get('list_of_scripts', [])
```
phw, line 288: Just noting that right now the existing scripts will always be empty, as on import a new profile is created and populated.

But still I agree that the way this function behaves does make sense, as it prevents any surprises if we extend the usage in the future.

phw (changes requested):

picard/profiles/importer.py
zas: Yes, you're right, I remove it in https://github.com/metabrainz/picard/pull/3231/commits/f8b567eb3efaee2726d92c0eb1b9648834a655de

phw (approved):
From what I have tested and from the code this does work as expected. Definitely a good enhancement, I'm curious how this might get used.

We need to update the documentation for this change then.

rdswift:
> We need to update the documentation for this change then.

I guess by "we", you actually mean me.  :wink:  Done.

PR #3237, opened 2026-06-26
PICARD-3325: Fix tag view double-counting selected tracks after refresh

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Fix the metadata box showing inflated item counts (e.g. "different across 6 items" instead of 3) after pressing Ctrl+R when both an album node and its track nodes are selected.

## Problem

* JIRA ticket: PICARD-3325

After metadata refresh (Ctrl+R), album tracks are re-created as new Python objects. However, `window.selected_objects` still held references to the old Track instances. The metadatabox then counted both the stale tracks from `selected_objects` and the new tracks obtained via the Album node, resulting in double-counting.

## Solution

Added `TreeItem._notify_selection_updated()` which rebuilds `selected_objects` from the view's current `selectedItems()` (whose `.obj` references are already updated by `AlbumItem.update()`). Replaced all call sites that previously called `update_selection(new_selection=False)` without passing objects.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Investigation, root cause analysis, and fix implementation were done with AI assistance.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/itemviews/__init__.py
+++ b/picard/ui/itemviews/__init__.py
@@ -483,7 +483,7 @@ class ClusterItem(TreeItem):
         if self.obj.special and album and album.loaded:
             album.ui_item.update(update_tracks=False)
         if update_selection and self.isSelected():
-            TreeItem.window.update_selection(new_selection=False)
+            TreeItem.window.panel.update_current_view()
 
     def add_file(self, file):
         self.add_files([file])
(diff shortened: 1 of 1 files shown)


zas:
@phw Totally untested, but have a look

phw:
@zas I tested and it fixes the issue. But I'm not sure about the implementation. The `TreeItem._notify_selection_updated` duplicates partly the functionality of `MainPanel.update_current_view`, which I think we could call as well. Alternative implementation hence would be:

```
class TreeItem:
    ...

    def _notify_selection_updated(self):
        self.window.panel.update_current_view()
```

(or directly calling that at the corresponding places).

zas:
> ```
> class TreeItem:
>     ...
> 
>     def _notify_selection_updated(self):
>         self.window.panel.update_current_view()
> ```
> 
> (or directly calling that at the corresponding places).

Done in https://github.com/metabrainz/picard/pull/3237/commits/fdcb77c668ece104c74f5dfe4334ee39a8f93763 (calling directly since that's 1 line)

phw (approved):
Tested, works as expected.

PR #3238, opened 2026-06-26
PICARD-507: Remove loaded-album boost from release matching

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Remove the "loaded-album boost" in `_compare_to_release_parts()` that gave already-loaded releases an unfair advantage in the identifier tier during cluster matching, causing files to be matched to the wrong release.

## Problem

* JIRA ticket: PICARD-507

When a user has multiple similar releases to look up (e.g., different issues of a magazine sampler series), the second cluster lookup can match to the *wrong*, already-loaded release instead of its correct release.

**Root cause:** `_compare_to_release_parts()` in `matching.py` added a `(1.0, 6)` entry to the identifier tier when a candidate release was already loaded in Picard. Because cluster lookups typically have no barcode or catalog number in file metadata, this was the *only* identifier signal — giving `id_score = 1.0` and triggering the "strong identifier match" branch in `combine_tiers()`, which floors the score at ~0.95 regardless of how poorly the release actually matches the file metadata.

This was originally reported in 2013 with Paste Magazine Sampler `#48` and `#41`. The bug still reproduces with current code: with the boost active and slightly degraded metadata (missing date — common for CD rips), the wrong release scores 0.9602 vs the correct release at 0.9256.

## Solution

Remove the 6-line loaded-album boost entirely. The similarity signals (album title, artist, date, track count) are sufficient to identify the correct release without this boost.

**Verification:**
- `scripts/eval_matching/eval_matching.py --compare baseline.json` shows **0 regressions** across all 850 test cases and 5 config profiles — scores are identical before and after since the boost only fires with a live tagger instance
- A new `scripts/eval_matching/simulate_picard507.py` script reproduces the original bug scenario with a mocked tagger and confirms the fix: all degradation variants now match correctly with healthy margins (~0.16)
- Full pytest suite passes (5331 tests, 0 failures)

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Analysis of the bug, the fix, the simulation script, and the PR were developed with AI assistance (Kiro CLI).

## Action

Additional actions required:

### DIFF
--- a/picard/matching.py
+++ b/picard/matching.py
@@ -50,7 +50,6 @@ from typing import (
     TypeVar,
 )
 
-from picard import tagger_instance
 from picard.config import Config, get_config
 from picard.mbjson import artist_credit_from_node, get_score
 from picard.similarity import similarity2
@@ -232,13 +231,6 @@ def _compare_to_release_parts(
                 file_label = metadata.get('label', '') or ''
                 score = _catno_label_score(file_catno, file_label, release_label_info)
                 result.identifiers.append((score, id_w['catno']))
-
-    if 'release-group' in release:
-        tagger = tagger_instance()
-        if tagger is not None:
-            rg = tagger.get_release_group_by_id(release['release-group']['id'])  # type: ignore[attr-defined]
-            if release['id'] in rg.loaded_albums:
-                result.identifiers.append((1.0, 6))
 
     # Tier 2: Similarity — fuzzy matching core
     with metadata._lock.lock_for_read():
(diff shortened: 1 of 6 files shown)


phw (approved):
Yes, makes sense to me. If we want to have improvements into this direction we probably rather need a processing step that analysis matches for multiple files lookups and tries to reduce the number of releases from the same RG. But the current approach of just preferring the first loaded hit leads to more false matches overall.

I had previously tried to extend our eval matching tests to generally test for this condition, but this did not really work. But having the separate test script is easier.

PR #3240, opened 2026-06-28
PICARD-3327: fix auto match with restored session

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3327

If the option "No auto-matching on load" is enabled in session options auto-matching files with MBIDs or auto-analyzing added files does not work for the entire session.

These get disabled during the session restoring process, but never get enabled after the session was restored.

## Solution

The root cause was that `SessionLoader` is not a `QObject` and hence can not have Qt slots, but when passing a method to `singleShot` a slot is expected, otherwise it is not called.

Turning  `SessionLoader` into a proper `QObject` ensures `_unset_restoring_flag_when_idle` gets called. Also removed the condition to only unset `tagger._restoring_session` if the `session_safe_restore` option is enabled. The option is not guaranteed to stay unchanged during session restore, so there is a chance of it getting disabled during session restore, which would cause `_restoring_session` to be stuck. There is also no harm setting it to `False` regardless.

This PR also contains a small change to always emit a message after session restore ended. Otherwise progress report was stuck with the last message when finalization was pending.

Note: Instead of turning `SessionLoader` into a `QObject` the callback could also have been defined as an inline function. But this complicates testing.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/session/session_loader.py
+++ b/picard/session/session_loader.py
@@ -29,6 +29,7 @@ from dataclasses import dataclass
 import gzip
 from pathlib import Path
 from typing import (
+    TYPE_CHECKING,
     Any,
     Protocol,
 )
@@ -53,6 +54,10 @@ from picard.session.session_data import (
     GroupedItems,
 )
 from picard.session.track_mover import TrackMover
+
+
+if TYPE_CHECKING:
+    from picard.tagger import Tagger
 
 
 class ProgressReporter(Protocol):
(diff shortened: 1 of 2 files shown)


zas (approved):
LGTM

zas (approved):
LGTM

PR #3242, opened 2026-07-01
PICARD-3330: Fix config upgrade hooks to also update profile overrides

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Config upgrade hooks that transform option values (not just rename keys) now also update those values in user profile overrides. Previously, profile overrides were left with stale/broken values after upgrade.

## Problem

* JIRA ticket: PICARD-3330

Several config upgrade hooks modify `in_profile=True` option values in the base config but do not update the same options when overridden in user profiles. This means profile overrides retain stale/broken values after upgrade.

Affected hooks:
- `v2_5_0dev1`: `ca_providers` — "Whitelist" never renamed to "UrlRelationships" in profiles
- `v3_0_0dev10`: `cover_tags/file_convert_to_format` — never lowercased in profiles
- `v3_0_0a2`: `list_of_scripts` — broken `$matchedtracks()` never fixed in profiles
- `v3_0_0b5`: `quick_menu_items` — new default items never added in profiles

The existing `rename_option()` helper already handles profiles correctly, but there was no equivalent for value transforms.

## Solution

1. Add `upgrade_option_value(config, name, transform)` helper that applies a transform function to both the base config and all profile overrides (analogous to `rename_option()`).

2. Add reusable `rename_option_in_settings()` and `upgrade_option_value_in_settings()` that operate on plain settings dicts. `rename_option()` is refactored to use the former internally. These helpers are also designed for reuse by future profile import functionality.

3. Fix the 4 affected hooks to use `upgrade_option_value()`.

4. Add unit tests for the new helpers and profile-aware integration tests for each fixed hook.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code, tests, and analysis were developed with AI assistance (Kiro CLI). The bug was identified during design analysis of the shareable profiles feature, and the fix was implemented and tested with AI guidance.

## Action

Additional actions required:

### DIFF
--- a/picard/config_upgrade.py
+++ b/picard/config_upgrade.py
@@ -64,6 +64,39 @@ def temp_option(option_type: type[Option], section: str, name: str, default: Con
     opt.unregister()
 
 
+def _rename_option_in_settings(
+    settings: dict,
+    old_name: str,
+    new_name: str,
+    reverse: bool = False,
+) -> None:
+    """Rename an option key in a settings dict (profile override or imported settings).
+
+    If reverse is True, the boolean value is also inverted.
+    Settings with None value (tracked but not overridden) are renamed without transformation.
+    """
+    if old_name in settings:
+        value = settings[old_name]
+        if reverse and value is not None:
+            value = not value
+        settings[new_name] = value
+        del settings[old_name]
+
+
+def _upgrade_option_value_in_settings(
+    settings: dict,
+    name: str,
+    transform: Callable,
+) -> None:
+    """Apply a value transform to an option in a settings dict (profile override or imported settings).
+
+    The transform function receives the current value and must return the new value.
+    Settings with None value (tracked but not overridden) are left unchanged.
+    """
+    if name in settings and settings[name] is not None:
+        settings[name] = transform(settings[name])
+
+
 def rename_option(
     config: Config,
     old_opt: str,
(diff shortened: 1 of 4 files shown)


picard/config_upgrade.py
```diff
@@ -64,6 +64,39 @@ def temp_option(option_type: type[Option], section: str, name: str, default: Con
     opt.unregister()
 
 
+def rename_option_in_settings(
```
phw: Both `rename_option_in_settings` and `upgrade_option_value_in_settings` are very specific helpers. They are for internal use by `rename_option` and `upgrade_option_value` and specified to be used on a dict, which only as part of profiles makes really sense.

Callers should really use `rename_option` and `upgrade_option_value`. To make this clear, let's rename the `*_in_settings` helpers to `_rename_option_in_settings` and `_upgrade_option_value_in_settings`, prefixed with underscore.

Or maybe even `_rename_option_in_profile` and `_upgrade_option_value_in_profile`, if we consider them really profile specific (as they are currently used).

test/test_config_upgrade_hooks.py
```diff
@@ -647,3 +647,142 @@ def test_upgrade_to_v3_0_0b5(self):
         self.assertEqual(items[:3], ['rename_files', 'move_files', 'enable_tag_saving'])
         self.assertIn('save_images_to_tags', items)
         self.assertIn('save_images_to_files', items)
+
+    def test_upgrade_to_v2_5_0dev1_profiles(self):
```
phw: Let's sort these tests to follow the existing `test_upgrade_to_v2_5_0dev1` instead of tucking them to the end of file. Applies accordingly to the following tests.

phw (changes requested):
Great work. This issue is really kind of a big oversight with the profiles. The solution here looks clean and easy to use.

Some minor comments on code structure.

picard/config_upgrade.py
zas: Done in 3ab3c1894

test/test_config_upgrade_hooks.py
zas: Done in 947fac139

zas:
@phw I did the changes and squashed to first commit.

phw (approved):
LGTM

PR #3244, opened 2026-07-02
PICARD-3331: Fix encoding error saving file metadata (NFC/NFD normalization)

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Replace `encode_filename()` with `resolve_fs_path()` — a new function that resolves file paths against the filesystem using Unicode NFC/NFD normalization fallback. This fixes the `MutagenError: [Errno 2] No such file or directory` when saving files with accented characters on byte-exact filesystems (macOS → SMB → Linux ext4).

## Problem

* JIRA ticket: PICARD-3331 (also related: PICARD-644, PICARD-2055)

On macOS, `encode_filename()` always converted file paths from Python `str` to `bytes` before passing them to Mutagen. When files reside on a byte-exact filesystem (e.g. SMB mount backed by Linux ext4), and the stored filename uses a different Unicode normalization (NFC vs NFD) than what the filesystem has, the bytes don't match and the file cannot be found.

Example: A file stored as NFD `Franc\xcc\xa7oise` (decomposed ç) on the server fails to open when Picard produces NFC bytes `Fran\xc3\xa7oise` (composed ç).

This also affects Linux where `os.path.supports_unicode_filenames` is `False`, causing `encode_filename` to return bytes there as well.

## Solution

1. **Add `resolve_fs_path(filename: str | bytes | Path) -> str`** — resolves a path against the filesystem, trying NFC and NFD normalization alternatives if the file isn't found. Zero overhead in the common case (file exists → immediate return).

2. **Call `resolve_fs_path` at external path entry points** — where paths come from Qt (drag & drop, file dialogs, URLs) or CLI arguments. Called once before `normpath`. Also called after file save/rename to ensure `self.filename` matches the on-disk representation.

3. **Remove `encode_filename()` from all format handlers** — Mutagen receives `str` paths directly, which the OS handles correctly on all platforms.

4. **Remove `encode_filename()` and `decode_filename()`** entirely — they have no remaining callers. The root cause of the bug (forcing bytes) is eliminated.

5. **Fix `_shorten_to_bytes_length`** — uses `os.fsencode()` / `os.fsdecode()` for byte-length calculation.

### Convention established

- `resolve_fs_path()` is called **once** when a path enters the pipeline from an external source (Qt dialogs, drag & drop, CLI). Not needed for paths from `os.scandir`/`os.listdir` or files we just created.
- All downstream code (format handlers, stat calls, cover art) receives an already-resolved `str`
- No more `str` → `bytes` → `str` round-trips for file I/O

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Analysis, reproduction, implementation, and tests were developed with AI assistance (Kiro CLI). Design decisions were made collaboratively with the maintainer.

## Action

Additional actions required:
* [x] Other: This is a draft — needs testing on macOS with actual SMB mounts to confirm the fix works end-to-end in the real scenario.

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -176,6 +176,7 @@ from picard.util import (
     periodictouch,
     pipe,
     process_events_iter,
+    resolve_filename,
     system_supports_long_paths,
     thread,
     versions,
@@ -1014,6 +1015,7 @@ class Tagger(QtWidgets.QApplication):
         new_files = []
         for filename in filenames:
             filename = normpath(filename)
+            filename = resolve_filename(filename)
             if ignore_hidden and is_hidden(filename):
                 log.debug("File ignored (hidden): %r", filename)
                 continue
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -175,7 +175,7 @@ from picard.util import (
     periodictouch,
     pipe,
     process_events_iter,
-    resolve_filename,
+    resolve_fs_path,
     system_supports_long_paths,
     thread,
     versions,
@@ -1014,7 +1014,7 @@ class Tagger(QtWidgets.QApplication):
         new_files = []
         for filename in filenames:
             filename = normpath(filename)
-            filename = resolve_filename(filename)
+            filename = resolve_fs_path(filename)
             if ignore_hidden and is_hidden(filename):
                 log.debug("File ignored (hidden): %r", filename)
                 continue
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -1013,8 +1013,8 @@ class Tagger(QtWidgets.QApplication):
         ignore_hidden = config.setting["ignore_hidden_files"]
         new_files = []
         for filename in filenames:
-            filename = normpath(filename)
             filename = resolve_fs_path(filename)
+            filename = normpath(filename)
             if ignore_hidden and is_hidden(filename):
                 log.debug("File ignored (hidden): %r", filename)
                 continue
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -168,7 +168,6 @@ from picard.track import (
 from picard.util import (
     check_io_encoding,
     cli,
-    encode_filename,
     is_hidden,
     iter_files_from_objects,
     mbid_validate,
@@ -1307,9 +1306,7 @@ class Tagger(QtWidgets.QApplication):
     def run_lookup_cd(self, device):
         disc = Disc()
         self.set_wait_cursor()
-        thread.run_task(
-            partial(disc.read, encode_filename(device)), partial(self._lookup_disc, disc), traceback=log.is_debug()
-        )
+        thread.run_task(partial(disc.read, device), partial(self._lookup_disc, disc), traceback=log.is_debug())
 
     def lookup_discid_from_logfile(self):
         file_chooser = FileDialog(parent=self.window)
--- a/picard/util/filenaming.py
+++ b/picard/util/filenaming.py
@@ -122,7 +122,7 @@ def _shorten_to_bytes_length(text: str, length: int) -> str:
     when encoded in the "filesystem encoding".
     """
     assert isinstance(text, str), "This function only works on unicode"
-    raw = text.encode(_io_encoding, 'replace')
+    raw = os.fsencode(text)
     # maybe there's no need to truncate anything
     if len(raw) <= length:
         return text
@@ -137,12 +137,12 @@ def _shorten_to_bytes_length(text: str, length: int) -> str:
         # so ord(char) & 0b11000000 = 0b10000000
         while i > 0 and (raw[i] & 0xC0) == 0x80:
             i -= 1
-        return raw[:i].decode(_io_encoding)
+        return os.fsdecode(raw[:i])
     # finally, a brute force approach
     i = length
     while i > 0:
         try:
-            return raw[:i].decode(_io_encoding)
+            return os.fsdecode(raw[:i])
         except UnicodeDecodeError:
             pass
         i -= 1
--- a/picard/util/filenaming.py
+++ b/picard/util/filenaming.py
@@ -48,8 +48,6 @@ from picard.util import (
     WIN_MAX_FILEPATH_LEN,
     WIN_MAX_NODE_LEN,
     _io_encoding,
-    decode_filename,
-    encode_filename,
     is_unc_path,
     samefile,
 )
@@ -124,7 +122,7 @@ def _shorten_to_bytes_length(text: str, length: int) -> str:
     when encoded in the "filesystem encoding".
     """
     assert isinstance(text, str), "This function only works on unicode"
-    raw = encode_filename(text)
+    raw = text.encode(_io_encoding, 'replace')
     # maybe there's no need to truncate anything
     if len(raw) <= length:
         return text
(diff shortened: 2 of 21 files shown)


picard/tagger.py
```diff
@@ -1014,6 +1014,7 @@ def add_files(self, filenames, target=None):
         new_files = []
         for filename in filenames:
             filename = normpath(filename)
+            filename = resolve_filename(filename)
```
phw: I'm a bit unsure here, but I think we should call `resolve_filename` before `normpath`. `normpath` (by defaullt) also does a `os.path.realpath(path)`, and this actually accesses the file. So having the path properly resolved beforehand seems necessary.

picard/util/filenaming.py
```diff
@@ -124,7 +122,7 @@ def _shorten_to_bytes_length(text: str, length: int) -> str:
     when encoded in the "filesystem encoding".
     """
     assert isinstance(text, str), "This function only works on unicode"
-    raw = encode_filename(text)
+    raw = text.encode(_io_encoding, 'replace')
```
phw: Could we use `os.fsencode` instead? (and accordingly `os.fddecode` below?)

test/test_utils.py
```diff
@@ -757,37 +757,81 @@ def test_encoded_queryargs(self):
...
+        """Returns existing str path unchanged."""
+        with NamedTemporaryFile(suffix='.flac') as f:
+            result = resolve_filename(f.name)
+            self.assertIsInstance(result, str)
+            self.assertEqual(f.name, result)
+
+    def test_bytes_existing_file(self):
+        """Decodes bytes path and returns str."""
+        with NamedTemporaryFile(suffix='.flac') as f:
+            result = resolve_filename(f.name.encode('utf-8'))
```
phw: This test might fail if run on a system where `sys.getfilesystemencoding()` is not `utf-8`, right? Skip the test unless `sys.getfilesystemencoding() == 'utf-8'`?

test/test_utils.py
```diff
@@ -757,37 +757,81 @@ def test_encoded_queryargs(self):
...
+        nfc_name = 'Voilà.flac'
+        nfd_name = unicodedata.normalize('NFD', nfc_name)
+        with tempfile.TemporaryDirectory() as d:
+            # Create file with NFD name
+            nfd_path = os.path.join(d, nfd_name)
+            with open(nfd_path, 'w') as f:
+                f.write('test')
+            # Try resolving via NFC path
+            nfc_path = os.path.join(d, nfc_name)
+            result = resolve_filename(nfc_path)
```
phw: Verify the result equals the nfd_path:

```
self.assertEqual(nfd_path, result)
```

test/test_utils.py
```diff
@@ -757,37 +757,81 @@ def test_encoded_queryargs(self):
...
+            # Create file with NFC name
+            nfc_path = os.path.join(d, nfc_name)
+            with open(nfc_path, 'w') as f:
+                f.write('test')
+            # Try resolving via NFD path
+            nfd_path = os.path.join(d, nfd_name)
+            result = resolve_filename(nfd_path)
+            self.assertTrue(os.path.exists(result))
+
+    def test_ascii_path_no_normalization_needed(self):
```
phw: I don't think this test is needed, this is already fully covered by `test_str_existing_file`.

test/test_utils.py
```diff
@@ -757,37 +757,81 @@ def test_encoded_queryargs(self):
...
+        nfd_name = unicodedata.normalize('NFD', nfc_name)
+        with tempfile.TemporaryDirectory() as d:
+            # Create file with NFC name
+            nfc_path = os.path.join(d, nfc_name)
+            with open(nfc_path, 'w') as f:
+                f.write('test')
+            # Try resolving via NFD path
+            nfd_path = os.path.join(d, nfd_name)
+            result = resolve_filename(nfd_path)
+            self.assertTrue(os.path.exists(result))
```
phw:
```
self.assertEqual(nfc_path, result)

```

test/test_util_filenaming.py
```diff
@@ -354,18 +354,8 @@ def test_shorten_bytes_invalid_mode(self):
...
-    def test_shorten_bytes_fs_unicode_support(self):
-        self.assertEqual('ä' * 10, shorten_filename('ä' * 11, 10, ShortenMode.BYTES))
-
-    @unittest.skipIf(
-        os.path.supports_unicode_filenames and not IS_MACOS,
-        'for filesystem without Unicode support',
-    )
-    def test_shorten_bytes_fs_no_unicode_support(self):
+    def test_shorten_bytes_multibyte(self):
+        # 'ä' is 2 bytes in UTF-8, so 10 bytes fits 5 'ä' characters
```
phw, line 358: Why are these tests removed? I think they should be kept.

scripts/tools/setup_picard_3331.sh
```diff
@@ -0,0 +1,393 @@
+#!/usr/bin/env bash
+#
```
phw: This way this script sets up filesystems for local testing is very useful. Not only file system encoding, but we could also use it e.g. for testing ext4 with case-insensitivity (we do support that, but it is again a tricky situation).

But the script here with all the specific reproducer code is already broken with the newer commits. I would keep this out of this PR and leave it as a testing code (maybe put into a gist).

We can add a virtual file system setup script in a separate PR. Maybe with some options for encodings and other options for the generated file system. What do you think?

phw (changes requested):
Good job, this looks very good. This fits my assumption that for the most part this encode / decode work should not be needed and we pass around the path as string. But we address the Unicode normalization issue on load.

For audio files I think this PR already covers everything. We likely need to also run `resolve_fs_path` on a few other places where we receive paths via drag and drop or also from Qt. From the top of my head:

- Drag and drop of images into the image box
- Selecting a cover art image via file open dialog
- Ripper log files selected via dialog
- Script files loaded via file naming script editor load dialog

Maybe also when we reload an audio file after saving. When saving we call `make_save_path`, which on macOS does NFD normalization. But we might have other cases where the filesystem applies Unicode normalization. So after saving an audio file it would be save to run `resolve_fs_path` again on the resulting path.

test/test_util_filenaming.py
zas: The old tests were removed because they tested platform-dependent behavior that was a side effect of `encode_filename`'s inconsistency.
The new `test_shorten_bytes_multibyte` test verifies the correct behavior on all platforms.

scripts/tools/setup_picard_3331.sh
zas: Yes, I'll remove it.

picard/tagger.py
zas: Addressed in b9c13f030

picard/util/filenaming.py
zas: Addressed in 149dcebf8

test/test_utils.py
zas: Addressed in 3e3b74fee

test/test_utils.py
zas: Addressed in 3e3b74fee

test/test_utils.py
zas: Addressed in 3e3b74fee

test/test_utils.py
zas: Addressed in 3e3b74fee

scripts/tools/setup_picard_3331.sh
zas: Removed from PR (rebased out).

phw:
@zas This is great. It gets rid of so many uncertainties surrounding the proper use  of the encoding utility functions and for all I can tell should fix the linked tickets. Code looks clean. Since the changes mostly remove a lot of `encode_filename` / `decode_filename` calls they are less dramatic then I have expected.

In my testing this branch is working well. There is some potential that we might get some encoding related bug report for some odd case. But even if we do I think it will be easier to deal with with the changes here in place.

Would be great to get some feedback from dear_misery, but I'm all for LGTM.

phw (dismissed):

zas:
Closed due to branch being incorrectly pushed to upstream instead of fork. Re-opened as a new PR from the correct source.

PR #3246, opened 2026-07-02
PICARD-3331: Fix encoding error saving file metadata (NFC/NFD normalization)

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Replace `encode_filename()` with `resolve_fs_path()` — a new function that resolves file paths against the filesystem using Unicode NFC/NFD normalization fallback. This fixes the `MutagenError: [Errno 2] No such file or directory` when saving files with accented characters on byte-exact filesystems (macOS → SMB → Linux ext4).

## Problem

* JIRA ticket: PICARD-3331 (also related: PICARD-644, PICARD-2055)

On macOS, `encode_filename()` always converted file paths from Python `str` to `bytes` before passing them to Mutagen. When files reside on a byte-exact filesystem (e.g. SMB mount backed by Linux ext4), and the stored filename uses a different Unicode normalization (NFC vs NFD) than what the filesystem has, the bytes don't match and the file cannot be found.

Example: A file stored as NFD `Franc\xcc\xa7oise` (decomposed ç) on the server fails to open when Picard produces NFC bytes `Fran\xc3\xa7oise` (composed ç).

This also affects Linux where `os.path.supports_unicode_filenames` is `False`, causing `encode_filename` to return bytes there as well.

## Solution

1. **Add `resolve_fs_path(filename: str | bytes | Path) -> str`** — resolves a path against the filesystem, trying NFC and NFD normalization alternatives if the file isn't found. Zero overhead in the common case (file exists → immediate return).

2. **Call `resolve_fs_path` at external path entry points** — where paths come from Qt (drag & drop, file dialogs, URLs) or CLI arguments. Called once before `normpath`. Also called after file save/rename to ensure `self.filename` matches the on-disk representation.

3. **Remove `encode_filename()` from all format handlers** — Mutagen receives `str` paths directly, which the OS handles correctly on all platforms.

4. **Remove `encode_filename()` and `decode_filename()`** entirely — they have no remaining callers. The root cause of the bug (forcing bytes) is eliminated.

5. **Fix `_shorten_to_bytes_length`** — uses `os.fsencode()` / `os.fsdecode()` for byte-length calculation.

### Convention established

- `resolve_fs_path()` is called **once** when a path enters the pipeline from an external source (Qt dialogs, drag & drop, CLI). Not needed for paths from `os.scandir`/`os.listdir` or files we just created.
- All downstream code (format handlers, stat calls, cover art) receives an already-resolved `str`
- No more `str` → `bytes` → `str` round-trips for file I/O

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Analysis, reproduction, implementation, and tests were developed with AI assistance (Kiro CLI). Design decisions were made collaboratively with the maintainer.

## Action

Additional actions required:
* [x] Other: Needs testing on macOS with actual SMB mounts to confirm the fix works end-to-end in the real scenario.

### DIFF
--- a/picard/coverart/__init__.py
+++ b/picard/coverart/__init__.py
@@ -59,6 +59,7 @@ from picard.i18n import N_
 from picard.metadata import Metadata
 from picard.util import (
     imageinfo,
+    resolve_fs_path,
     thread,
 )
 
@@ -216,7 +217,7 @@ class CoverArt:
         # local files, load image data from filesystem
         elif image.url and image.url.scheme() == 'file':
             try:
-                path = image.url.toLocalFile()
+                path = resolve_fs_path(image.url.toLocalFile())
                 with open(path, 'rb') as file:
                     data = file.read()
                     image_info = imageinfo.identify(data)
--- a/picard/coverart/image.py
+++ b/picard/coverart/image.py
@@ -56,8 +56,6 @@ from picard.coverart.utils import (
 )
 from picard.metadata import Metadata
 from picard.util import (
-    decode_filename,
-    encode_filename,
     imageinfo,
     is_absolute_path,
     periodictouch,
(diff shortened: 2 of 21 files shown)


phw (approved):
As on the previous PR, LGTM

PR #3249, opened 2026-07-03
Columns util cleanup

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

This is a refactoring for the length and filesize custom columns sorting. Both used custom sort key parsing with some wrong assumptions.

## Solution

The `NumericSortAdapter` would always call `_base.evaluate` to get a columns value. This returns the display value. For numeric values this usually involves a call to formatting the numeric value, which `NumericSortAdapter` then has to parse again.

Hence `NumericSortAdapter` got changed to use the raw value from metadata first, if available. While always being a string, the value is usually already numeric, hence the default parser of using `float` might already work.

The filesize column used a helper `parse_file_size`. This had several issues:
- The column `~filesize` already contains the size in bytes. So the default conversion in `NumericSortAdapter` of using `float` already is enough
- `parse_file_size` did not handle `MiB` (as used by bytes2human and hence by the display value) nor localized decimal separators. Hence it always failed.
- The fallback of using natural number sorting already works reasonable well

Hence remove the `parse_file_size` helper and rely on parsing the raw metadata value as float.

The length column again used unnecessary parsing. Specifically the length aka audio duration has an efficient way of sorting, as `Metadata.length` already is an integer value of the duration in milliseconds. Add a new `MetadataDurationSortAdapter` with tests.R

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/itemviews/custom_columns/sorting_adapters.py
+++ b/picard/ui/itemviews/custom_columns/sorting_adapters.py
@@ -102,7 +102,7 @@ class NumericSortAdapter(_AdapterBase):
 
     def __init__(self, base: ColumnValueProvider, parser: Callable[[str], float] | None = None):
         super().__init__(base)
-        self._parser = parser or (lambda s: float(s))
+        self._parser = parser or float
 
     def __repr__(self) -> str:  # pragma: no cover - debug helper
         parser_name = getattr(self._parser, "__name__", repr(self._parser))
--- a/picard/ui/itemviews/custom_columns/sorting_adapters.py
+++ b/picard/ui/itemviews/custom_columns/sorting_adapters.py
@@ -98,7 +98,10 @@ class CasefoldSortAdapter(_AdapterBase):
 
 
 class NumericSortAdapter(_AdapterBase):
-    """Provide numeric sort using a parser (default: float) on evaluated value."""
+    """Provide numeric sort using a parser (default: float) on value.
+    Sorting will happen on the raw metadata value, if set. Otherwise falls back
+    to base provider evaluate.
+    """
 
     def __init__(self, base: ColumnValueProvider, parser: Callable[[str], float] | None = None):
         super().__init__(base)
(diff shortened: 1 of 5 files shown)


picard/ui/itemviews/custom_columns/sorting_adapters.py
```diff
@@ -124,6 +132,21 @@ def sort_key(self, obj: Item) -> Comparable:  # pragma: no cover - thin wrapper
...
+    """Sort items by their metadata.length (audio duration).
+
+    Fall back to numeric sort adapter of the field value.
+    """
+
+    def sort_key(self, obj: Item) -> Comparable:
+        """Return length-based sort key for item."""
+        metadata = getattr(obj, 'metadata', None)
+        if metadata:
+            return metadata.length
```
zas: Just a minor thing, but here we return an `int` and according to above `getattr(obj, 'metadata', None)`,  `metadata` == None leads to executing `else` branch that calls `NumericSortAdapter.sort_key() ` which always return a tuple, see https://github.com/metabrainz/picard/blob/a8adf1dd5f0f26b4c56fd7f2e393a27d3ac26a45/picard/ui/itemviews/custom_columns/sorting_adapters.py#L165-L166

For consistency, we should also return a tuple here:

```suggestion
            return (0, metadata.length)
```

Without that, we may (but in practice I don't think it happens) have a `TypeError` raised:

```
>>> a = int(1)
>>> b = (0, 2,)
>>> a < b
TypeError: '<' not supported between instances of 'int' and 'tuple'
```

Not sure if this is something that should (or could) be detected by type checking (`Comparable` doesn't enforce a consistent type between compared items in this case).
Also it seems unit tests assume that `metadata` always exists (so we never test the `else` branch).

zas (changes requested):
Just a minor comment, the changes make total sense.

picard/ui/itemviews/custom_columns/sorting_adapters.py
phw: I addresses both issues in https://github.com/metabrainz/picard/pull/3249/changes/a17efe99ba98ea4df697bde178d8b056c64edacb

The adapter now returns a tuple and the test checks the mixed case of both variations being in the sorted values.

zas (approved):
LGTM

PR #3250, opened 2026-07-04
PICARD-3336: Make the about window non-modal

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3336

The about dialog is shown on macOS as a modal sheet, with no window controls and hence no UI element to close it.

I did some research how other applications on different platforms handle about / info dialogs, see https://tickets.metabrainz.org/browse/PICARD-3336?focusedId=81318&page=com.atlassian.jira.plugin.system.issuetabpanels:comment-tabpanel#comment-81318. This varies highly from application to application, there is not really a "right way" to do it.

But very common is the model where the about dialog is non-modal, so other windows still can get focus. But also the about dialog is shown on top, so it does not easily get lost. This is typical for all KDE applications and several other Linux applications, but also many applications on Windows.

On macOS the about dialogs of Apple's own software is also non-modal, but allows getting hidden by the main window (even thought macOS makes it rather difficult to reach hidden windows).

## Solution

Make the about dialog `WindowModality.NonModal`, but still set a parent. This makes the Window on most platforms stay on top, but not prevent other windows to gain focus. This matches behavior of many other applications across different platforms (see above).

It also fixes the dialog on macOS being shown as a sheet without a window close button.

grafik

On macOS the dialog does not stay on top, though. But again this matches behavior of Apple's own software, so I think is fine.

This also is now the same behavior as on Picard 2.x again.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/aboutdialog.py
+++ b/picard/ui/aboutdialog.py
@@ -42,6 +42,8 @@ from picard.ui.forms.ui_aboutdialog import Ui_AboutDialog
 
 
 class AboutDialog(PicardDialog, SingletonDialog):
+    modality = QtCore.Qt.WindowModality.NonModal
+
     def __init__(self, parent=None):
         super().__init__(parent=parent)
         self.setAttribute(QtCore.Qt.WidgetAttribute.WA_DeleteOnClose)


zas (approved):
LGTM

PR #3260, opened 2026-07-06
Wizard updates: Add update setup page, UI tweaks, menu item to re-run wizard

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-2999

This PR implements several changes to the current wizard. It has no intention to consider the wizard complete, but rather move it some steps into the right direction. If wanted, we can also split this PR up and merge only parts.

The focus is on adding a new page to configure the update checks to perform on startup. This would fix PICARD-2999 and make the update check opt-in.

This also involves removing the existing RTD update check dialog, as it becomes redundant. For nicer display this also adds a special wizard checkbox widget that allows showing more context for an option. Since this is also useful for other pages, apply it there as well.

Also add a Help menu item to allow re-running the wizard later again.

## Solution

1. Updates wizard page

grafik

The `WizardCheckbox` widget allows showing an additional description and applies some mouse over highlight to the checkbox widget. The option  can be toggled by clicking anywhere. Updates will be triggered when saving.

2. Apply the `WizardCheckbox` also to other option pages:

grafik

grafik

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/ui/setupwizard.py
```diff
@@ -20,19 +21,90 @@
...
+
+        self.setAttribute(QtCore.Qt.WidgetAttribute.WA_StyledBackground, True)
+        self.setStyleSheet(f"""
+            WizardCheckbox:hover {{
+                background-color: {highlight_bg.name()};
+            }}
+            * {{
+                color: {highlight_text.name()};
+            }}
+        """)
```
zas, line 66: The highlight is a bit weird, and on light theme the text is almost invisible:

Image

zas (changes requested):
I found various minor issues, mainly visual but also functional.
Also I tried to simplify texts even more as we aim for new comers/beginners.
Check https://github.com/phw/picard/pull/35

phw:
@zas Thanks for the changes. I cherry-picked most of your changes, with two exceptions:

1. The stylesheet changes: that change missed the point. What was broken was setting the hover text color (for ensuring contrast to the highlight color). What the commit changed was setting the general text color to WindowText as a kind of random choice. Should have been rather Text really, but it's pointless since that's the default color and Text and WindowText are often the same (and if they aren't, the text in WindowText color would probably not be well readable in our case).

I opted instead of adding alpha (~30%) to the highlight color. That way it is more subtle and blends in with the background. And I think we don't need a hover text color in this case.

2. The click changes: Text input and button already handle click events themselves. The outer widget in that case does not even receive the click event. You can click into the input and select text etc. without triggering a toggle [Bildschirmaufzeichnung vom 2026-07-07 08-13-36.webm](https://github.com/user-attachments/assets/3664c247-dea8-4d97-b1e5-4b49a521d23c) .

zas:
> @zas Thanks for the changes. I cherry-picked most of your changes, with two exceptions:
> 
>     1. The stylesheet changes: that change missed the point. What was broken was setting the hover text color (for ensuring contrast to the highlight color). What the commit changed was setting the general text color to WindowText as a kind of random choice. Should have been rather Text really, but it's pointless since that's the default color and Text and WindowText are often the same (and if they aren't, the text in WindowText color would probably not be well readable in our case).
> 
> 
> I opted instead of adding alpha (~30%) to the highlight color. That way it is more subtle and blends in with the background. And I think we don't need a hover text color in this case.
> 
>     2. The click changes: Text input and button already handle click events themselves. The outer widget in that case does not even receive the click event. You can click into the input and select text etc. without triggering a toggle [Bildschirmaufzeichnung vom 2026-07-07 08-13-36.webm](https://github.com/user-attachments/assets/3664c247-dea8-4d97-b1e5-4b49a521d23c) .

I tested and everything seems to work as expected now.

zas (approved):

phw:
I also tested this again, and it's overall an improvement. Let's merge this and next step we could do some improvements to the options offered here for selection. I liked you idea of offering some grouped options, where you say to use tags and artwork close to as it is on the release, or more standardize things.

I also added PICARD-3343 to track specifically the feature to restart the wizard on demand as implemented here.

PR #3263, opened 2026-07-08
PICARD-3344: Fix crash when collection menu callback fires after widget deletion

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Fix a crash (`RuntimeError: wrapped C/C++ object of type CollectionCheckBox has been deleted`) that occurs when the async web service response arrives after the user has closed the collection menu.

## Problem

* JIRA ticket: PICARD-3344

When a user adds/removes a release from a collection via the context menu and closes the menu before the web service response arrives, the callback `_update_text` tries to call `setText()` on a destroyed Qt widget, causing a crash.

## Solution

Wrap the `setText()` call in `CollectionCheckBox._update_text()` with a `try/except RuntimeError` to gracefully handle the case where the underlying C++ widget has already been deleted. This is the only async entry point to the widget, and `_label()` only accesses Python-side attributes, so the exception is precisely scoped.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/collectionmenu.py
+++ b/picard/ui/collectionmenu.py
@@ -143,7 +143,12 @@ class CollectionCheckBox(QtWidgets.QCheckBox):
             self.setCheckState(QtCore.Qt.CheckState.Unchecked)
 
     def _update_text(self):
-        self.setText(self._label())
+        try:
+            self.setText(self._label())
+        except RuntimeError:
+            # The underlying C++ widget may have been deleted if the menu
+            # was closed before the async web service response arrived.
+            pass
 
     def _label(self):
         c = self.collection


phw (approved):
I tried to reproduce this, but couldn't trigger it. Even when on my system the `_update_text` got called after the menu was already closed, the exception  would not trigger. Could you actually reproduce?

This is probably really timing related. Anyway, the situation generally is clear and the fix makes sense. The alternative would maybe be to make the widget itself part of the callback (using `partial`). But not sure if this might lead to a circular reference and potential memory leak.

Let's god with the exception handling.

PR #3266, opened 2026-07-09
PICARD-3330: Unified config upgrade framework with decorator-based hooks

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

Replace the old version-encoded `upgrade_to_v*` hook system with a decorator-based framework (`@upgrade_settings`, `@upgrade_config`) that cleanly separates settings transforms from non-settings operations, enabling automatic profile override propagation and profile import support.

## Problem

* JIRA ticket (_optional_): PICARD-3330

The old config upgrade system conflated settings transforms (option renames, value changes) with non-settings operations (persist cleanup, UI state resets). Settings transforms need to apply in three contexts: base config at startup, existing profile overrides, and imported profile data. The old system required manual profile iteration in each hook and had no import support. Adding these features would have required duplicating all transforms in a parallel registry.

## Solution

**Two decorators** register into a single ordered registry:
- `@upgrade_settings(version)` — settings transforms, automatically applied to base config + all profile overrides + imported profiles
- `@upgrade_config(version)` — non-settings operations (persist, UI state), startup only

**Polymorphic helpers** work on both plain dicts and `SettingConfigSection`:
- `rename_option()` — rename a key, optionally invert boolean
- `upgrade_option_value()` — apply a value transform
- `get_option()` — read without removing (for complex transforms)
- `write_option()` — write with automatic Enum serialization for dicts
- `remove_option()` — remove a key

**Key properties:**
- Single source of truth — each transform defined once, works everywhere
- Declaration order determines execution within same version
- All hooks are idempotent — safe on retry after failure
- Profile import: `apply_settings_upgrades_for_import(settings, version_str)`
- `NamedTuple` registry entries with proper type annotations

**All 42 existing hooks converted** — zero old-style `upgrade_to_v*` functions remain. Old infrastructure removed (`autodetect_upgrade_hooks`, `UPGRADE_FUNCTION_PREFIX`, etc.).

Design document: `docs/UnifiedConfigUpgrades.md`

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Design decisions were made collaboratively. Implementation, tests, and documentation were primarily AI-generated with human review and direction at each step.

## Action

Additional actions required:


picard/config_upgrade.py
```diff
@@ -28,170 +28,414 @@
...
+
+    def decorator(func):
+        version = Version.from_string(version_str)
+        _UPGRADES_REGISTRY.append(_UpgradeEntry(version, _UpgradeType.SETTINGS, func))
+        return func
+
+    return decorator
+
+
+def upgrade_config(version_str: str):
```
phw: Let's add a typehint for the return type here:

```python
 -> Callable[[Callable[[Config], None]], None]
```

This will then trigger a type error when trying this decorator on an incompatible function.

Or maybe we should define some type alias for the functions:

```python
SettingsUpgradeFunc: TypeAlias = Callable[[Settings], None]
ConfigUpgradeFunc: TypeAlias = Callable[[Config], None]
```

Then we can annotate `upgrade_settings` and `upgrade_config` with:

```python
def upgrade_settings(version_str: str) -> Callable[[SettingsUpgradeFunc], None]:
    ...

def upgrade_config(version_str: str) -> Callable[[ConfigUpgradeFunc], None]:
    ...
```

picard/config_upgrade.py
```diff
@@ -28,170 +28,414 @@
...
+    Returns:
+        The option value (deserialized), or default if the key is not present.
+    """
+    if isinstance(settings, dict):
+        return settings.get(name, default)
+    else:
+        assert option_type is not None
+        assert default is not None
+        if name not in settings:
+            return default
```
phw, line 271: Handling the default here would not be strictly needed, as it would be handled by `settings.value(opt, default)`. But maybe a shortcut, not sure it is worth it.

picard/config_upgrade.py
```diff
@@ -28,170 +28,414 @@
...
+        option_type: Option class (e.g., BoolOption). Required for ConfigSection,
+                     ignored for plain dicts.
+        default: Default value if the key is missing. Required for ConfigSection,
+                 ignored for plain dicts (returns None if missing and no default).
+
+    Returns:
+        The option value (deserialized), or default if the key is not present.
+    """
+    if isinstance(settings, dict):
+        return settings.get(name, default)
```
phw: If `option_type` is given, we need a `opt.convert` here to mimic https://github.com/metabrainz/picard/blob/master/picard/config.py#L312

picard/config_upgrade.py
```diff
@@ -28,170 +28,414 @@
...
 ) -> None:
-    """Apply a value transform to an option in a settings dict (profile override or imported settings).
+    """Apply a value transform to an option in settings.
+
+    Polymorphic: works on both plain dicts (profile overrides, imported
+    settings) and ConfigSection objects (base config at startup).
 
-    The transform function receives the current value and must return the new value.
-    Settings with None value (tracked but not overridden) are left unchanged.
+    The transform function receives the current value and must return the
```
phw: Clarify, that it receives the current "raw value", since it uses `raw_value` for settings. I think this is actually good, because it avoids that we need to apply the Option.convert in two directions.

test/test_config_upgrade.py
```diff
@@ -23,139 +23,370 @@
...
-            Version(major=1, minor=0, patch=0, identifier='final', revision=0),
-            Version(major=2, minor=0, patch=0, identifier='dev', revision=1),
-            Version(major=2, minor=0, patch=0, identifier='final', revision=0),
-        )
-        self.assertEqual(tuple(hooks), expected_keys)
-
-
-class TestRenameOptionInSettings(PicardTestCase):
     def test_rename_existing_key(self):
+        from picard.config_upgrade import rename_option
```
phw: A lot of duplicate inline imports in this file, should be used to top.

phw (changes requested):
@zas This is I think generally the right approach. But I think we need to check closer if the value handling is working correctly in the polymorphic handlers.

In the normal code setting values are accessed via `ConfigSection.__getitem__` or `ConfigSection.value`, which handle all necessary transforms, and profile values are also need special handling implemented in `ProfileConfigSection`.

In the upgrade helpers we use `ConfigSection` access, but direct access to the dict values. This works well in the standard cases. But we have special types handled differently. E.g. we have `Enum` and `dict` values, which are specified with the `Option` type. And `IntEnum` specified as `IntOption` type.

The exact storage differs between JSON and Qt's ini. But the readback value should be the same in the end.

We probably also need to extend the tests for the polymorphic helpers to test both dict and settings access.

picard/config_upgrade.py
phw: Done in affc4a4a362f1acc90ab023c1c392f591caa1c75

picard/config_upgrade.py
phw: Done in d1cc143e3be62f6ed4978a69ab27979031aeadfe

picard/config_upgrade.py
phw: Done in ed15305cb720355b72ecf0d6b0d60939ab80d667

test/test_config_upgrade.py
phw: Done in a27160694f106004b3df6a2e7d543ea1725b12a1

picard/config_upgrade.py
phw: But also does not harm, I left it as is in my PR

phw (approved):
This looks good to me, both in code and from testing. Important is, that the changed hooks still pass the previous test conditions, which they do.

PR #3271, opened 2026-07-13
PICARD-3358: Add metadata tag context menu extension point and script variable title

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add a new plugin API extension point (`register_metadata_tag_action`) that allows plugins to register context menu actions for metadata tags. Also adds a `title` parameter to `register_script_variable` so plugins can provide display names for custom tags in the metadata box.

## Problem

* JIRA ticket: [PICARD-3358](https://tickets.metabrainz.org/browse/PICARD-3358)

Plugins currently cannot add per-tag actions to the metadata view context menu. This blocks use cases like the proposed "Pinned Tags" feature (see [community discussion](https://community.metabrainz.org/t/artist-folder-naming-how-do-you-handle-artists-with-identical-names/814857/14)) where users need to pin/unpin specific tags from the right-click menu.

Additionally, plugin-registered script variables show their raw name (e.g., `pinned_tags`) in the metadata box instead of a user-friendly title.

### New extension point: `register_metadata_tag_action`
- `MetadataTagAction` base class (inherits `HasDisplayTitle` for translation support)
- `callback(self, tags, objects)` receives list of selected tags + selected objects
- `is_visible(self, tags, objects)` controls menu visibility
- Actions appear for both single and multi-tag selections
- UI automatically refreshes after callback via `update_selection()`

### Enhanced `register_script_variable`
- New `title` parameter for metadata box display name
- Refactored internal storage to `PluginVariable` dataclass (was a positional tuple)
- `display_tag_name()` checks plugin variable titles as fallback

### Real-world plugin using this API
- https://github.com/zas/picard-plugin-pinned-tags

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Implementation and iteration done with Kiro CLI, with human design decisions and review throughout.

## Action

Additional actions required:
* [x] Other: Create matching PICARD JIRA tickets


test/test_metadata_tag_actions.py
```diff
@@ -0,0 +1,148 @@
...
+                pass
+
+            def is_visible(self, tags, objects):
+                return all(not tag.startswith('~') for tag in tags)
+
+        action = MyAction()
+        self.assertTrue(action.is_visible(['artist'], set()))
+        self.assertTrue(action.is_visible(['artist', 'albumartist'], set()))
+        self.assertFalse(action.is_visible(['~filename'], set()))
+        self.assertFalse(action.is_visible(['artist', '~filename'], set()))
```
phw, line 131: Why do all these asserts have the `msg` set to an empty `set()`?

phw (commented):
This looks good already. This can be really useful. I can see how various plugins could provide actions to apply transformations on specific tags. Or allow opening links for certain tags.

As an example I'll likely update the opencc plugin to provide such an action so it can do Chinese script conversion on selected tags only.

How do we handle updating the metadata view after the action completed? The action needs to call `obj.update()` in case it changed something on a tag and needs the view to update?

test/test_metadata_tag_actions.py
zas: That's not the `msg`, but the `objects` parameter of `is_visible()`

zas:
> How do we handle updating the metadata view after the action completed? The action needs to call `obj.update()` in case it changed something on a tag and needs the view to update?

Currently the core calls` update_selection(new_selection=False, drop_album_caches=True)` unconditionally after every action callback.
I initially tried to call `obj.update()` but it was causing a runtime error (because that's in a separate thread).

With core changes and my example plugin it seems to work as expected though.

test/test_metadata_tag_actions.py
phw: 🤦

phw:
> Currently the core calls` update_selection(new_selection=False, drop_album_caches=True)` unconditionally after every action callback.

But that'll only work as long as the processing is synchronous, though. Should we offer some clear way to trigger an update from within the action?

zas:
> Should we offer some clear way to trigger an update from within the action?

Probably, but I'm not sure how, if you have an idea I'll happily accept a PR over this PR ;)

phw:
@zas I'll look into this and provide a PR

phw (approved):
Looks good to me, we only need to add a ticket

zas:
> Looks good to me, we only need to add a ticket

Ticket added as https://tickets.metabrainz.org/browse/PICARD-3358

PR #3275, opened 2026-07-14
Fix crash in profile options when copying without selection

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

The profile view would update on the list's current item changing, but act on the list's selected item. Those are distinct in Qt. If the user cleared the selection, this would cause there being a current but no selected item.

This would cause the actions to misbehave. Specifically the "Copy" button would crash.

## Solution

Change the logic to always act on the current item, not the selection. If there is no current item, the buttons are disabled.

Also change the `currentItemChanged` handler not to again call `setCurrentItem`. The signal indicates that the item changed, and the handler gets the new value. It does not need to call `setCurrentItem` again. This just triggered a callback loop, which the handler then tried to workaround by setting the loading flag.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/options/profiles.py
+++ b/picard/ui/options/profiles.py
@@ -396,10 +396,7 @@ class ProfilesOptionsPage(OptionsPage):
         Returns:
             ProfileListWidgetItem: Currently selected item
         """
-        items = self.ui.profile_list.selectedItems()
-        if items:
-            return items[0]
-        return None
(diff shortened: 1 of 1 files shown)


zas (approved):
LGTM

PR #3278, opened 2026-07-15
ISRC support: lookup, CD extraction, and submission

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Adds comprehensive ISRC support to Picard — extraction from CD via libdiscid, lookup by ISRC (explicit user action), and submission to MusicBrainz with confirmation dialog.

## Problem

* JIRA tickets: PICARD-163, PICARD-2482, PICARD-2961

Picard reads ISRCs from files and includes them in search queries, but does not:
- Extract ISRCs from physical CDs during disc read
- Offer a direct "Lookup by ISRC" action
- Submit ISRCs to MusicBrainz

The external tool `musicbrainz-isrcsubmit` provides CD ISRC submission but requires a separate workflow. Users have requested this functionality be integrated into Picard (PICARD-163 has 30 votes, open since 2012).

### CD ISRC Extraction
- `Disc.read()` now requests the `isrc` feature from libdiscid (when available and enabled)
- Extracted ISRCs are validated, stored on the `Disc` object, and logged in debug mode
- Duplicate ISRCs (same ISRC on multiple tracks — a known drive issue) are detected and skipped
- A "View ISRCs" button in the CD lookup dialog shows per-track ISRCs
- When disc ID is not found in MB, a "Lookup by ISRC" button searches all disc ISRCs in a single Lucene OR query (1 API call)
- ISRC reading is user-configurable (enabled by default, can be disabled for slow drives)

### ISRC Lookup
- "Lookup by ISRC" context menu action for files
- Uses recording search API with `isrc:` query (workaround for MBS-10185)
- Uses file metadata (album, artist) for release disambiguation when multiple results
- Rate-limiting aware: explicit user action only, not part of automatic lookup
- Matching logic in `picard/util/isrc.py`, not in mainwindow

### ISRC Submission
- `MBAPIHelper.submit_isrcs()` — POST XML to `/ws/2/recording/`
- `ISRCSubmitManager` tracks pending ISRCs (from files or CD)
- Detects submittable ISRCs on file-to-track match and after CD lookup
- Confirmation dialog grouped by release showing track, title, existing ISRCs, new ISRCs
- Option in General settings (disabled by default)
- ISRCs from CD are added to track metadata regardless of submission setting
- Multi-disc releases handled correctly (matches by medium disc ID)

### Utilities
- `picard/util/isrc.py`: `valid_isrc()`, `normalize_isrc()`, `format_isrc()`, `best_release_from_isrc_results()`
- ISRCs displayed in formatted form (CC-XXX-YY-NNNNN) in UI and logs

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

## Action

Additional actions required:

image
image


docs/ISRCSupport.md
```diff
@@ -0,0 +1,367 @@
...
+ISRC"), not part of the automatic flow:
+
+- Add a `lookup_isrc()` method to `MBAPIHelper` in
+  `picard/webservice/api_helpers/musicbrainz.py`. It performs a GET to
+  `/isrc/<isrc>` with optional `inc` parameters (same as recording lookups).
+- Add a "Lookup by ISRC" context menu action for files and clusters (relates
+  to PICARD-2961).
+- For a **single file**: one API call to look up its ISRC.
+- For a **cluster**: look up one ISRC to identify the release from the
+  recording's release list, then load that release normally (2 calls total,
```
phw, line 122: From a conceptional point, without having yet checked what has been implemented, that doesn't sound very useful really. Loading some release of one recording from the cluster is very different from how one would expect a cluster lookup to work. If ISRC lookup does not work for a cluster, I'd rather limit the functionality to file lookups.

docs/ISRCSupport.md
```diff
@@ -0,0 +1,367 @@
...
+  regardless of album size).
+- If the ISRC is not found in MB, fall back to the existing metadata search.
```
phw: I don't think a ISRC lookup should fall back to anything else. It is a separate action, it should do what it is called.

docs/ISRCSupport.md
```diff
@@ -0,0 +1,367 @@
...
+serve as a reference for building the submission request.
+
+#### ISRC Sources
+
+1. **From file tags:** After a file is matched to a track, compare the file's
+   ISRCs with the recording's ISRCs (already fetched via `inc=isrcs`). ISRCs
+   present in the file but absent from MB are candidates for submission.
+
+2. **From CD:** After a CD lookup matches tracks to recordings, the extracted
+   ISRCs can be compared with what's already in MB.
```
phw, line 159: Note for the implementation: Submitting for a matched CD should be possible directly from the CD lookup. The match for a disc ID is usually good and it is rather safe to submit ISRCs here. Submitting from metadata for matched tracks not so much.

picard/album.py
```diff
@@ -350,6 +351,28 @@ def add_discid(self, discid):
...
 
+    def _submit_disc_isrcs(self):
+        """Add disc ISRCs to track metadata and register for submission."""
+        if not self._disc_isrcs:
+            return
+        config = get_config()
+        submit = config.setting['submit_isrcs']
+        for track in self.tracks:
+            track_number = int(track.metadata.get('tracknumber', 0))
+            disc_isrc = self._disc_isrcs.get(track_number)
```
phw: I wouldn't base this on the tracknumber tag. It is a tag commonly modified by users (also with scripts). I would rather use actual track position. `self.tracks` is already ordered. `self._disc_isrcs` can also be filled accordingly if coming from a disc lookup.

picard/ui/options/general.py
```diff
@@ -79,6 +80,20 @@ def __init__(self, parent=None):
...
+        self.ui.submit_isrcs = QtWidgets.QCheckBox(_("Submit ISRCs to MusicBrainz when tagging"), submission_group)
+        self.ui.submit_isrcs.setToolTip(
+            _(
+                "When enabled, ISRCs found in files that are not yet associated "
+                "with the matched recording in MusicBrainz can be submitted."
+            )
+        )
+        submission_layout.addWidget(self.ui.submit_isrcs)
+        # Insert after the general_box (index 3 in the vboxlayout)
+        self.ui.vboxlayout.insertWidget(3, submission_group)
```
phw: The above should be added to the .ui file instead

picard/disc/__init__.py
```diff
@@ -75,8 +80,11 @@ def read(self, device=None):
         if device is None:
             device = discid.get_default_device()
         log.debug("Reading CD using device: %r", device)
+        features = ['mcn']
+        if 'isrc' in discid.FEATURES:
+            features.append('isrc')
         try:
-            disc = discid.read(device, features=['mcn'])
+            disc = discid.read(device, features=features)
```
phw, line 90: ISRC read should be user configurable. We could have this directly toggleable in the CD lookup dialog. But ISRC reading slows down the CD reading significantly and shouldn't be enabled without an option to disable.it.

picard/ui/mainwindow/__init__.py
```diff
@@ -715,6 +719,102 @@ def _on_submit_acoustid(self):
...
+            if file_album and title:
+                score += similarity2(file_album, title) * 2
+            artist_credit = release.get('artist-credit', [])
+            if file_artist and artist_credit:
+                release_artist = artist_credit_from_node(artist_credit).name
+                score += similarity2(file_artist, release_artist)
+            if score > best_score:
+                best_score = score
+                best_id = release['id']
+        return best_id
```
phw: All the matching logic itself should be put into a separate module, not be part of the mainwindow implementation.

phw (changes requested):
Generally a good move. I couldn't fully test this with hardware yet, but the approach generally makes sense. My main concern would be that this PR tries to do too many things at once. We could separate this into e.g. only having the reading from CD implemented first, or separating the ISRC lookup.

Some things I found during review:

- ISRC lookup for a file always fails for me with a message that not releases were found. Further testing shows that e.g. https://musicbrainz.org/ws/2/isrc/US27Q1763711?inc=artist-credits+releases&fmt=json does not return releases, while https://musicbrainz.org/ws/2/isrc/US27Q1763711?inc=artist-credits+releases&fmt=xml does. This looks like a server bug to me.
- A common problem with reading ISRCs from disc always was that sometimes (varying on drive and disc) it will fail read the ISRCs for the proper track, resulting in the same ISRC being returned for multiple tracks. Often times retrying the read can solve this. But if we do this we should implement some handling for it. [isrcsubmit](https://pypi.org/project/isrcsubmit/) handled this by showing a warning when it detects the same ISRC being returned for multiple track.
- I'm not quite sure how submitting ISRCs from disc and from tags are handled. I couldn't get the submit ISRC to activate. But I think we should offer a dedicated path to submit from disc, it is where we can have higher confidence (if we handle the reading issue from above properly).

Also about the linked tickets: PICARD-604 is not really related to this PR (and generally I think out-of-scope for Picard, but that's a different discussion). The PR touches aspects of PICARD-2961, but this needs to be implemented separately.

zas:
Agreed this does a lot. The commits are self-contained and could be split into separate PRs if preferred (e.g., CD extraction alone, lookup alone, submission alone).

About PICARD-604: you're right, it's not really related — it was listed as a dependency on PICARD-163 but that link seems outdated. I'll remove the reference from the design doc.

Regarding the ISRC lookup always failing: that's MBS-10185 (`/ws/2/isrc/<isrc>?fmt=json` ignores `inc=releases`). Worked around in e1aeaa6d3 by using the search API instead.

For the duplicate ISRCs from drives: detection and warning added in cdaddfbc4 — duplicated ISRCs are skipped.

Submit button not activating: the `submit_isrcs` option is disabled by default. Once enabled in Options → General → Submission, it activates after a CD lookup matches a release. Tested end-to-end, works.

Separate CD submission path: the CD flow already uses a distinct code path (disc ISRCs → `_submit_disc_isrcs()` in `album.py`), separate from the file metadata path. The confidence level is indeed higher for disc-matched ISRCs.

docs/ISRCSupport.md
zas: Addressed in 61022b572 — removed cluster support from Lookup by ISRC, limited to files only.

docs/ISRCSupport.md
zas: Addressed in b9291bb77 — removed fallback mentions from the design doc (was never implemented in code).

docs/ISRCSupport.md
zas: Addressed in 8fcefc513 — disc ISRCs are matched by medium position using disc ID, separate from the file-based path. Submission is triggered after the album loads from a CD lookup.

picard/album.py
zas: Addressed in 8fcefc513 — uses enumerate position and filters by medium disc ID instead of the tracknumber metadata tag. Also handles multi-disc releases correctly.

picard/ui/options/general.py
zas: Addressed in 5f7d95003 — moved Submission group to the .ui file with proper tab order.

picard/disc/__init__.py
zas: Addressed in d8991ab80 — added `read_isrcs_from_disc` option (default: enabled), only requests ISRC feature when the option is on and libdiscid supports it.

picard/ui/mainwindow/__init__.py
zas: Addressed in 6c8263e8c — moved to `picard/util/isrc.py` as `best_release_from_isrc_results()`.

rdswift:
I haven't had a chance to review this yet, but there might be some things that you could use from the Submit ISRCs plugin, such as checking for duplicate ISRCs between tracks and not submitting already existing ISRCs.

zas:
> such as checking for duplicate ISRCs between tracks and not submitting already existing ISRCs.

Implemented in https://github.com/metabrainz/picard/pull/3278/commits/83de8fa51fad37241e05afd59651b2084e108126
Also the submit ISRC confirmation dialog shows reasons ISRCs are disabled for submission.

zas:
Superseded by individual PRs split by feature:

- #3280 — ISRC submission from file metadata (standalone)
- #3281 — CD ISRC extraction during disc lookup (depends on #3280)
- #3282 — Lookup by ISRC context menu action (depends on #3280)

Closing this combined PR in favor of the smaller, reviewable pieces.

PR #3280, opened 2026-07-16
PICARD-163: Add ISRC submission from file metadata

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add ISRC submission support from file metadata. When a file with an ISRC is matched to a track, detect if the ISRC is new to MusicBrainz and offer to submit it via a confirmation dialog.

## Problem

* JIRA ticket: PICARD-163

Users who have files with ISRCs from authoritative sources (e.g., CD ripping software) cannot submit those ISRCs to MusicBrainz from within Picard. This requires external tools or plugins.

## Solution

- Add `picard/util/isrc.py` with validation, normalization, and formatting utilities
- Add `submit_isrcs()` method to MBAPIHelper (POST XML to `/recording/`)
- Add `ISRCSubmitManager` to track pending ISRCs as files are matched to tracks
- Add `isrc_update()` to `File`, called alongside `acoustid_update()` on match
- Add "Submit ISRCs" action in File menu and toolbar options with confirmation dialog
- Add `DebugOpt.ISRC` for detailed ISRC processing logging (`--debug-opts=isrc`)

Safety checks inspired by rdswift/picard-plugin-submit-isrc:
- Cross-recording duplicates are detected and shown as disabled in the dialog
- ISRCs already assigned to another track on the same release are flagged
- Invalid ISRCs are logged as warnings

Files with multiple ISRCs (e.g., from a previous tagging session that wrote MB ISRCs) are handled correctly — the manager compares each ISRC against MusicBrainz to identify genuinely new ones.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

## Action

Additional actions required:


picard/ui/options/interface_toolbar.py
```diff
@@ -122,6 +122,10 @@ class InterfaceToolbarOptionsPage(OptionsPage):
             N_("Submit AcoustIDs"),
             'acoustid-fingerprinter',
         ),
+        MainAction.SUBMIT_ISRC: ToolbarButtonDesc(
+            N_("Submit ISRCs"),
+            'document-save',
```
phw: I think we need a different icon here, not the disk icon. If we want to keep this as a placeholder for now I'm happy with that as well, we just shouldn't forget about this and ask aerozol or monkey for help.

picard/util/isrc.py
```diff
@@ -0,0 +1,71 @@
...
+
+    Converts `CCXXXYYNNNNN` to `CC-XXX-YY-NNNNN`.
+
+    Args:
+        value: Normalized 12-character ISRC string.
+
+    Returns:
+        Hyphenated display form, or the original value if not 12 characters.
+    """
+    if len(value) != 12:
```
phw: We should also use `if _ISRC_RE.match(normalized):` here and return it unchanged if it is not a valid ISRC.

picard/util/isrc.py
```diff
@@ -0,0 +1,71 @@
...
+def normalize_isrc(value: str) -> str:
+    """Normalize an ISRC value by removing hyphens/spaces and uppercasing.
+
+    Args:
+        value: Raw ISRC string, possibly with hyphens or spaces.
+
+    Returns:
+        Normalized uppercase ISRC string with separators removed.
+    """
+    return value.strip().replace('-', '').replace(' ', '').upper()
```
phw, line 41: From how this function works we should probably consider this internal to this module and prefix it with an underscore.

The real function that should be used is `valid_isrc`, as this validates the result's syntax and returns accordingly.

But `normalize_isrc` just removes hyphens and spaces from any string. The input, and hence also the output, might or might not be an ISRC

phw (changes requested):
This looks pretty good already. I have a few comments on the code.

Also I noticed that submission entries stay in the ISRCSubmitManager even if the album or recording got removed. I think we should remove entries in this case (as it is for AcoustID submissions.

I like the submission dialog. Something similar to this we need for AcosustIDs as well.

zas:
Thanks for the review!

> submission entries stay in the ISRCSubmitManager even if the album or recording got removed

Addressed in 670e62d07 — `file.remove()` now calls `isrc_submit_manager.remove()`, and `tagger.remove_album()` calls `isrc_submit_manager.remove_album()` to clean up track-keyed entries (from disc ISRCs).

picard/ui/options/interface_toolbar.py
zas: Addressed in a388507d8 — changed to a dedicated `'isrc-submit'` icon name as a placeholder. The theme lookup will fall back gracefully until a proper icon is designed.

picard/util/isrc.py
zas: Addressed in 15b28fcb1 — `format_isrc()` now uses `_ISRC_RE.match()` to validate before formatting, returning the value unchanged if it's not a valid ISRC.

picard/util/isrc.py
zas: Addressed in 3d02b3392 — renamed to `_normalize_isrc()` to indicate it's internal to the module. Callers should use `valid_isrc()` which validates the result.

rdswift (dismissed):
Now that you've addressed my concerns regarding missing some ISRC processing lines, this is working as expected.

You might want to reconsider if the new configuration setting about enabling the "Submit ISRCs" action is actually necessary.

Not sure how all the markdown file formatting applies to this issue, but whatever.

Once this is implemented, it appears that the "Submit ISRCs" plugin will be rendered obsolete, and can be removed from the registry.

zas:
> Not sure how all the markdown file formatting applies to this issue, but whatever.

This is a side effect of upgrade to ruff 0.16.0 which now formats code examples in the markdown files. I pushed the change separately and drop the changes in those PRs as it was unrelated.

> You might want to reconsider if the new configuration setting about enabling the "Submit ISRCs" action is actually necessary.

Yes, I think I'll remove it.

rdswift (dismissed):
Looks good, and works as expected during testing.  I like that you removed the configuration setting to enable/disable the menu action.

scripts/tools/generate_icons.sh
```diff
@@ -0,0 +1,61 @@
+#!/bin/bash
```
phw, line 1: I think the script should be put under scripts/tools

picard/ui/isrcsubmitdialog.py
```diff
@@ -0,0 +1,115 @@
...
+
+    def __init__(self, details, parent=None):
+        """
+        Args:
+            details: dict keyed by (album, albumartist) with values being
+                lists of (track_number, title, existing_isrcs, new_isrcs).
+        """
+        super().__init__(parent=parent)
+        self.setWindowTitle(_("Submit ISRCs"))
+        self.resize(700, 450)
```
phw: Better set this as `defaultsize` on the dialog? See PicardDialog

picard/isrcsubmit.py
```diff
@@ -0,0 +1,371 @@
...
+                sorted(new_isrcs),
+            )
+        else:
+            del self._entries[file]
+            log.debug_if(
+                DebugOpt.ISRC,
+                "ISRC update: %r → recording %s, all ISRCs now in MB, removed entry",
+                file,
+                recording_id,
+            )
```
phw, line 131: The above if/else block for updating the entry with new_isrcs is the same in add and update and could be put into a helper (e.g. `def _update_entry(self, file, new_isrcs`)

phw (changes requested):
Looks good. A few comments, but I think then we should merge this (and go on with the next PRs).

picard/isrcsubmit.py
zas: Done in https://github.com/metabrainz/picard/pull/3280/commits/b0112704f3032b4b1be3a357ad17090d7f51cac6

picard/ui/isrcsubmitdialog.py
zas: Yes, done in https://github.com/metabrainz/picard/pull/3280/commits/35aea0c73cdb500a649426013b1088def022e88f

scripts/tools/generate_icons.sh
zas: Done in https://github.com/metabrainz/picard/pull/3280/commits/5f3903c4e350485ba4946c51f8d6fae273c46da3

phw (approved):
LGTM

PR #3281, opened 2026-07-16
PICARD-163, PICARD-2482: Add CD ISRC extraction during disc lookup

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Extract ISRCs from physical CDs via libdiscid during disc lookup. Display them in the CD lookup dialog and register them for submission after the release loads.

## Problem

* JIRA tickets: PICARD-163, PICARD-2482

When looking up a CD, Picard reads the disc ID and MCN but not the ISRCs embedded in the disc's subcode. These ISRCs are authoritative and could be submitted to MusicBrainz (PICARD-2482).

## Solution

Depends on #3280 (ISRC submission from files).

- Read ISRCs from CD via libdiscid's `isrc` feature during `Disc.read()`
- Validate and deduplicate ISRCs (duplicate ISRCs across tracks are a known drive issue)
- Add "Read ISRCs from CD" option on the CD Lookup options page (defaults to on) — some drives are slow with ISRC reading
- Add "View ISRCs (N)" button to CD lookup dialog showing per-track ISRCs
- Pass disc ISRCs to Album on load; after album finishes loading, add ISRCs to track metadata and register for submission
- Handle multi-disc releases by filtering to the correct medium via disc ID

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

## Action

Additional actions required:


jesus2099:
> * Validate and deduplicate ISRCs (duplicate ISRCs across tracks are a known drive issue)

Sometimes, duplicate ISRC are legit, though. For some cases of split tracks, or for x musical parts within the same work, split over x tracks, for example.

Mistake duplicate ISRC is mostly a problem with many slim CD drives.

The best way to avoid mistake duplicate ISRC was implemented by mediatools.exe, slow but secure:

Read 10+ consecutive ISRC for each track and wait for 10 consecutive same ISRC, to avoid empty ISRC or previous track's ISRC sliding down to next.

Maybe implement this as a second pass, only each time duplicate ISRC occurs? But like this, it wouldn't be as secure.

zas:
> Sometimes, duplicate ISRC are legit, though

Yes, but in this case we depend on libdiscid, and duplicate ISRCs is often due to a CD read issue, for now it seems safer to avoid submission (it can be fixed on MB directly if required).

jesus2099:
Oh yes, you're right! 👍

phw (approved):
@zas This works well. I even found a CD lying around here that had unsubmitted ISRCs (still has those, I keep this for testing for now).

Image

One thing that could be useful would be to show the track title in the dialog listing a CDs ISRCs. As a third column (track number and ISRC are the most important infos).

zas:
> One thing that could be useful would be to show the track title in the dialog listing a CDs ISRCs. As a third column (track number and ISRC are the most important infos).

Yes, possible, let's merge this for now.

PR #3282, opened 2026-07-16
PICARD-163, PICARD-2961: Add Lookup by ISRC in a 'Lookup by...' submenu

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add a "Lookup by…" submenu to the file context menu with "ISRC" as the first entry. Right-click a file and look up its recording via the track search dialog using a MusicBrainz ISRC query.

## Problem

* JIRA tickets: PICARD-163, PICARD-2961

Users with files that have ISRCs but no other MusicBrainz identifiers have no way to use the ISRC for lookup within Picard. They must manually search on the website. PICARD-2961 describes the broader need for a "Lookup by..." submenu with various identifier-based lookups.

## Solution

Depends on #3280 (ISRC submission from files, for the `picard/util/isrc` module).

### Commit 1: Refactor TrackSearchDialog to accept a file in constructor

- Add `matched_file` parameter to `TrackSearchDialog.__init__()` so a file can be passed directly at construction time (instead of only via `show_similar_tracks()`)
- Rename `file_` to `_file` (private attribute), add type hints, rename `search()` parameter to `query`

### Commit 2: Add 'Lookup by ISRC' context menu action for files

- Add a "Lookup by…" submenu to the file context menu (PICARD-2961), with "ISRC" as the first entry
- Use the search API (`isrc:<value>` query in the track search dialog) as a workaround for MBS-10185 (the `/isrc/` endpoint ignores `inc=releases` in JSON mode)
- Action enabled when any files are selected; uses the first valid ISRC found
- Debug logging (via `--debug-opts=isrc`) for the ISRC being searched
- The submenu structure prepares for future additions (barcode, catalogue number, disc ID)

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -757,6 +757,13 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
         dialog.search(f'isrc:{escape_lucene_query(isrc)}')
         dialog.exec()
 
+    def create_lookup_by_menu(self, parent):
+        """Create a 'Lookup by…' submenu for use in context menus."""
+        lookup_menu = QtWidgets.QMenu(_("Lookup by…"), parent)
+        lookup_menu.setIcon(icontheme.lookup('system-search'))
+        lookup_menu.addAction(self.action_map[MainAction.LOOKUP_ISRC])
+        return lookup_menu
+
     def _create_actions(self):
         self.action_map = dict(create_actions(self))
 
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -86,6 +86,7 @@ from picard.const.sys import (
     IS_MACOS,
     IS_WIN,
 )
+from picard.debug_opts import DebugOpt
 from picard.extension_points.plugin_tools_menu import (
     ext_point_plugin_tools_items,
     signaler,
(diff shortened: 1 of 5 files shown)


picard/ui/mainwindow/__init__.py
```diff
@@ -725,6 +730,77 @@ def _on_submit_isrc(self):
...
+            return
+        # Pick the best release using file metadata for disambiguation
+        release_id = best_release_from_isrc_results(releases, source_metadata)
+        log.debug_if(
+            DebugOpt.ISRC,
+            "ISRC lookup: %d releases found, selected %s",
+            len(releases),
+            release_id,
+        )
+        self.tagger.load_album(release_id)
```
phw: The alternative here would be to utilize the existing search dialogs. We already have "Search for similar tracks". I imagined ISRC lookup to work similar, just with a different query, then showing the results. It would fit here, since it only looks up a single ISRC anyway.

Not sure what is better. But it is something we might also want to consider when implementing PICARD-2961 fully: Is the result automatched, or shown to the user to select from.

phw (approved):
LGTM

PR #3283, opened 2026-07-18
PICARD-1444: Redesign release preferences UI

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other

Redesign the "Preferred Releases" options page to replace confusing sliders and dual-list pickers with simple priority lists. Users now add items from a dropdown and reorder them — no more guessing which side is "preferred."

## Problem

* JIRA ticket: PICARD-1444

The preferred releases UI has been a source of confusion for years (reported in 2019, again in July 2026). Users can't tell:
- Which column in the dual-list picker is "preferred" vs "available"
- What the slider values mean or what direction increases preference
- Whether their sliders are at 50% or 49%

See: https://community.metabrainz.org/t/confusing-settings-preferred-release-countries-formats-is-the-left-the-preferred-or-the-right/423486

## Solution

Replace the entire page with 4 `PreferenceListWidget` instances:

- **Preferred release types** — ordered priority list
- **Discouraged release types** — types to avoid when matching
- **Preferred release countries** — ordered priority list
- **Preferred medium formats** — ordered priority list

Each widget provides:
- A combobox dropdown to add items (with locale-aware sorting)
- Up/down buttons and drag-and-drop to reorder
- Remove button to deselect items
- Cross-exclusion between preferred/discouraged type lists

The scoring logic uses position-based weights (higher in list = higher score), matching the existing pattern used for countries and formats. Evaluation shows identical pass/fail results (827/850) with the matching test corpus.

Config migration hook (`3.0.0b8`) converts old `release_type_scores` to the new `preferred_release_types` / `discouraged_release_types` format.

image

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed collaboratively with an AI assistant (Kiro CLI), with human review, testing, and design decisions at each step.

## Action

Additional actions required:

The options page documentation at https://picard-docs.musicbrainz.org/en/config/options_releases.html will need screenshots and description updated.

### DIFF
--- a/picard/ui/options/releases.py
+++ b/picard/ui/options/releases.py
@@ -10,6 +10,7 @@
 # Copyright (C) 2017 Suhas
 # Copyright (C) 2018-2022, 2024-2025 Philipp Wolfer
 # Copyright (C) 2024 joncrall
+# Copyright (C) 2026 Laurent Monin
 #
 # This program is free software; you can redistribute it and/or
 # modify it under the terms of the GNU General Public License
@@ -26,19 +27,8 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 
-from collections.abc import (
-    Callable,
-    Mapping,
-)
-from functools import partial
-from operator import itemgetter
+from PyQt6 import QtWidgets
 
-from PyQt6 import (
-    QtCore,
-    QtWidgets,
-)
-
-from picard import tagger_instance
 from picard.config import get_config
 from picard.const import (
     RELEASE_FORMATS,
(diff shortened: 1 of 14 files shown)


picard/ui/options/releases.py
```diff
@@ -4,12 +4,13 @@
 #
 # Copyright (C) 2006 Lukáš Lalinský
 # Copyright (C) 2011-2014 Michael Wiencek
-# Copyright (C) 2012 Frederik “Freso” S. Olesen
+# Copyright (C) 2012 Frederik "Freso" S. Olesen
```
rdswift: This really doesn't matter to me, but I noticed that the quotes were changed.  Interesting.

rdswift (commented):
I wasn't sure at first, but now I'm starting to like the changed UI. One question though... Is the order of the "discouraged" items relevant? If not, perhaps that list could be sorted alphabetically (or in the same order the items appear in the selection drop-down) to display consistently?

Also, if the order is not relevant, could the up/down buttons be hidden and the drag'n'drop ordering disabled? Perhaps this could be managed within the `PreferenceListWidget` and the widget's `__init__()` could have an optional parameter to disable the manual ordering?

I would also be tempted to include your "What's this?" text as a tooltip on each of the section's title labels, for example the "Preferred release types (higher priority first)" label.  I never thought to right-click to get the "What's this?" until I saw it in the code.

I don't think "What's this?" is used anywhere else in Picard (but perhaps it could/should be added in other places... on a different PR).

picard/ui/options/releases.py
zas: Yes, annoying I'll revert that

rdswift (dismissed):
I like your recent changes.  I think you've addressed all of the things I noticed.  Thanks.

rdswift (dismissed):
This is looking really good.  I didn't spot any ui issues with my latest round of testing, including proper highlighting when selected on an option profile.

picard/ui/widgets/preferencelistwidget.py
```diff
@@ -0,0 +1,263 @@
...
+            self._up_btn = self._make_button(buttons_layout, ":/images/16x16/go-up.png", _("Move up"))
+            self._down_btn = self._make_button(buttons_layout, ":/images/16x16/go-down.png", _("Move down"))
+        self._remove_btn = self._make_button(buttons_layout, ":/images/16x16/list-remove.png", _("Remove"))
+
+        layout.addLayout(buttons_layout)
+
+    def _make_button(self, layout, icon_path, tooltip) -> QtWidgets.QToolButton:
+        btn = QtWidgets.QToolButton(self)
+        btn.setIcon(QtGui.QIcon(icon_path))
+        btn.setToolTip(tooltip)
```
phw, line 99: As the buttons only have an icon, please also do set `btn.setAccessibleDescription`. With the current texts it can I think be identical to the tooltip.

phw (changes requested):
Well done. The new UI is much clearer and the behavior much easier to understand and handle. It covers what I think users would expect, without the micro management the previous scoring offered. Also preferred countries / formats is now clear to understand.

I have a small change request to set the accessible description on the buttons, otherwise this is good to go.

Also just a note: I think it is fine, just noting that the migration can change results in an important way. Previously only types with a score of 0.0 where really scored with a high penalty so they would usually be excluded, while everything > 0 but < 0.5 would just score relatively. The migration will now treat everything below 0.5 as discouraged, hence more similar to previous 0.0.

I still think it makes sense as a migration. Smaller score usually indicated that the user does not want this.

picard/ui/widgets/preferencelistwidget.py
zas: Addressed in https://github.com/metabrainz/picard/pull/3283/commits/fe83525c92272af60cd93a89cb4a2b3f94767e71

phw (approved):
LGTM

PR #3284, opened 2026-07-20
PICARD-1164: Tag diff highlighting in metadata box

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Add word/character-level diff highlighting to the metadata box. Changed tag values now show exactly which parts differ using colored background highlights, making it much easier to spot changes in long tag values.

## Problem

* JIRA ticket: PICARD-1164
* Related: PICARD-1163 (partially addressed)

When tag values change, the metadata box shows the old and new values side by side, but for long values (e.g. artist names, titles) it's hard to see exactly what part changed. The entire row was colored in a status color (golden for changed, red for removed, green for added) without indicating which specific characters or words differ.

## Solution

A hybrid word/character-level diff highlighting system:

1. **Placeholder color for unchanged values** — New Value column uses muted/placeholder text when unchanged, reducing visual noise.
2. **Hybrid diff algorithm** — Tokenizes into words, punctuation, and whitespace. Diffs at token level first, then refines with character-level diff within similar tokens (ratio >= 0.5). This handles both word-level changes (artist reordering) and subtle character changes (typographic apostrophes).
3. **Full-string highlight for opaque identifiers** — MBIDs/AcoustIDs get full red/green background since char-level diff is meaningless for UUIDs.
4. **Status color on tag name only** — The tag name column carries the status color (golden/red/green) for quick scanning; value columns use normal text with background highlights.
5. **Removed tags** — Old value with red background, new column empty (no more strikethrough).
6. **Added tags** — New value with green background.
7. **Whitespace preservation** — Uses `white-space: pre` in HTML so extra spaces are visible.
8. **Dark theme support** — Higher highlight alpha (90) for dark theme, lower (60) for light.
9. **Background thread** — Diff computation runs in the background thread pool, keeping the UI responsive.

Architecture:
- `tagdiffhtml.py` — Pure diff logic, no Qt dependency, fully testable (26 tests)
- `difftextdocument.py` — Thin Qt display layer (color resolution, QTextDocument creation)
- `__init__.py` — Delegate paint override + integration in `_update_items`

Dark theme:
image

Light theme:
image

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed iteratively using Kiro CLI (Claude) with continuous human review and testing at each step. All design decisions (hybrid diff approach, color strategy, MBID handling, dark theme, whitespace preservation) were driven by human feedback on visual results.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -911,32 +911,60 @@ class MetadataBox(QtWidgets.QTableWidget):
             tag_status = self.tag_diff.tag_status(tag)
             color = colors.get(tag_status, colors[TagStatus.UNCHANGED])
 
+            # Tag name column gets the status color
+            tag_item.setForeground(color)
+
             orig_item = get_table_item(row, self.COLUMN_ORIG)
             self._set_item_value(orig_item, self.tag_diff.old, tag, color)
 
             new_item = get_table_item(row, self.COLUMN_NEW)
             if not self.tag_diff.is_readonly(tag):
                 new_item.setFlags(editable_item_flags)
-            strikeout = tag_status == TagStatus.REMOVED
             new_color = placeholder_color if tag_status == TagStatus.UNCHANGED else color
-            self._set_item_value(new_item, self.tag_diff.new, tag, new_color, strikeout=strikeout)
+            if tag_status == TagStatus.REMOVED:
+                # For removed tags, leave the new value column empty
+                new_item.setText("")
+                new_item.setData(QtCore.Qt.ItemDataRole.UserRole, tag)
+                font = new_item.font()
+                font.setItalic(False)
+                font.setStrikeOut(False)
+                new_item.setFont(font)
+            else:
+                self._set_item_value(new_item, self.tag_diff.new, tag, new_color)
 
             # Compute diff highlights for changed tags with concrete values.
             # For opaque identifiers (MBIDs), use full-string highlight.
             # For other tags, use hybrid word/character-level diff.
+            # For removed tags, highlight the entire old value.
+            # When diff HTML is active, use normal text color since the
+            # background highlights carry the change information.
             old_diff_html = None
             new_diff_html = None
-            if tag_status == TagStatus.CHANGED:
+            if tag_status == TagStatus.REMOVED:
+                old_status = self.tag_diff.old.status(tag)
+                if not old_status.is_grouped:
+                    old_text = MULTI_VALUED_JOINER.join(self.tag_diff.old[tag])
+                    normal_text_color = self.palette().color(QtGui.QPalette.ColorRole.Text)
+                    old_diff_html, new_diff_html = compute_full_diff_html(old_text, "", normal_text_color)
+                    new_diff_html = None
+            elif tag_status == TagStatus.ADDED:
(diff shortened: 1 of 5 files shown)


rdswift (commented):
I haven't gone over the code in detail, but this looks good to me so far.  A very nice improvement.

The only thing I noticed, is that if a track time length is different it isn't highlighted.  I don't think it's a show-stopper, but it is something that I noticed.

zas:
> The only thing I noticed, is that if a track time length is different it isn't highlighted. I don't think it's a show-stopper, but it is something that I noticed.

Yes, the length tag is special, it is read-only and it has a difference tolerance.
In last commits I improved it, showing the difference, but the tag title stays black (=no change) because it isn't written.
Also fixed the highlight.

image

picard/ui/metadatabox/__init__.py
```diff
@@ -818,6 +874,60 @@ def _add_tracks_to_tag_diff(self, tracks, tag_diff, config):
...
+            tag_status = tag_diff.tag_status(tag)
+            old_diff_html = None
+            new_diff_html = None
+
+            if tag == '~length':
+                # Always show diff for length when display values differ,
+                # regardless of tolerance-based status (it's read-only/informational).
+                # Skip when values are grouped (multiple files with different lengths).
+                # Use placeholder color for new value since this tag is never written.
+                if not tag_diff.old.status(tag).is_grouped and not tag_diff.new.status(tag).is_grouped:
```
phw, line 914: Not sure, but I think it should consider `ignore_track_duration_difference_under`. At least the metadatabox will not list the length as a changed track if the difference is below that threshold.

picard/ui/metadatabox/__init__.py
```diff
@@ -842,6 +952,7 @@ def _update_items(self, result=None, error=None):
             TagStatus.ADDED: QtGui.QBrush(interface_colors.get_qcolor('tagstatus_added')),
             TagStatus.CHANGED: QtGui.QBrush(interface_colors.get_qcolor('tagstatus_changed')),
         }
+        placeholder_color = self.palette().color(QtGui.QPalette.ColorRole.PlaceholderText)
```
phw, line 969: Good move of using the placeholder text here.

phw (commented):
Code looks good and it works well in my tests. The visual presentation works really good.

A questions for discussion about considering `ignore_track_duration_difference_under`

picard/ui/metadatabox/__init__.py
zas: > Not sure, but I think it should consider `ignore_track_duration_difference_under`. At least the metadatabox will not list the length as a changed track if the difference is below that threshold.

Actually the diff is shown, because it was rather unintuitive to have 3:04 and 3:05 and no difference shown (they are different), but it doesn't change the matching (tolerance is used). That was a bit tricky, and I tried both, also based on rdswift feedback regarding this, and the current code is the result of that process. I would not change that.
That's a compromise: this read-only field is a bit special.

phw: Ok, yes. The explanation makes sense. Let's go with this behavior

phw (dismissed):
LGTM

PR #3287, opened 2026-07-26
PICARD-3351: Make Options dialog scrollable and constrain to screen

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Make the Options dialog always fit on screen by wrapping each page in a scroll area and constraining the dialog to available screen geometry.

## Problem

* JIRA ticket: PICARD-3351

On systems where the Options dialog is taller than the screen (e.g. small screens, HiDPI, certain window managers), the "Make It So!" button becomes inaccessible, and the dialog cannot be resized down or moved up.

## Solution

- Each options page is lazily wrapped in a `QScrollArea` when first navigated to, so pages taller than the viewport become scrollable.
- Added `_constrain_to_screen()` in `showEvent` to cap dialog size to available screen geometry and reposition if off-screen (handles restored geometry from a previously larger monitor).
- Changed stacked widget vertical size policy from `Preferred` to `Ignored` so it no longer forces the dialog to grow to the tallest page.
- Removed the internal `QScrollArea` from the colors options page to avoid double scrollbars (the per-page wrapper handles scrolling generically for all pages, including plugin pages).

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed using Kiro CLI with Claude, with direction and review from the developer.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/forms/ui_options.py
+++ b/picard/ui/forms/ui_options.py
@@ -1,6 +1,6 @@
 # Form implementation generated from reading ui file 'ui/options.ui'
 #
-# Created by: PyQt6 UI code generator 6.9.1
+# Created by: PyQt6 UI code generator 6.11.0
 #
 # Automatically generated - do not edit.
 # Use `python setup.py build_ui` to update it.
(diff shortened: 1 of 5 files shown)


rdswift (dismissed):
This is the reason that I started getting into the habit of wrapping all of my plugin option pages in a QScrollArea.  I never thought to patch Picard instead.

Tested on Linux Mint and Windows 10, and both seemed to work as expected.

phw (changes requested):
To be honest, I'm not too much of a fan of the behavior. I think ideally the option pages should not scroll. Maybe it makes sense to have this scrolling behavior in place for situations where the options are too small, but we should aim for options being available without scrolling.

So for this PR I see two things to improve:

1. The default size of the dialog is way too small. We need to set a `defaultsize` attribute large enough so that it allows at least most dialogs to be shown without scrollbar.
2. The vertical scrollbar for many dialogs, e.g. Metadata or User Inrerface, shows up too early, when there is still plenty of room and all the UI elements are visible (see screenshots).

Image

Image

phw (approved):
Works well now, LGTM

PR #3295, opened 2026-07-28
Simplify cover art image selection flow

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Simplify the cover art image selection logic by extracting a clear early-termination method and adding a symmetric `to_be_saved_to_files()` method on `ImageList`.

## Problem

The cover art download early-termination logic in `_start_queue` was an inline condition that only handled the "tags only, one front" case. The external file saving logic in `File._save_images` used ad-hoc inline filtering rather than the same pattern as `to_be_saved_to_tags()`.

## Solution

**Commit 1: Refactor early termination into `_all_images_satisfied`**

Extract the inline condition into a dedicated method that considers both embedding and external file settings symmetrically. This also extends the optimization to cover cases not previously handled (e.g., both tags and files configured to save only the front image, or only files enabled with `save_only_one_front_image`).

**Commit 2: Add `ImageList.to_be_saved_to_files()`**

Add `to_be_saved_to_files()` as the symmetric counterpart to `to_be_saved_to_tags()`. Both methods follow the same pattern: check if saving is enabled, apply the "only one front" filter if configured, and yield matching images. Refactors `File._save_images` to use it, removing inline filtering logic.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed collaboratively with an AI assistant (Kiro CLI), with human review and design decisions by the author.

## Action

Additional actions required:

No documentation update needed — internal refactoring only.

### DIFF
--- a/picard/coverart/__init__.py
+++ b/picard/coverart/__init__.py
@@ -133,15 +133,26 @@ class CoverArt:
             self._queue_generator = None
             self.image_processing.wait_for_processing()
 
+    def _all_images_satisfied(self, config=None) -> bool:
+        """Check if all configured image needs have been met.
+
+        Returns True if no more images need to be downloaded, considering
+        both embedding and external file settings symmetrically.
+        """
+        if not self.front_image_found:
+            return False
+        if config is None:
+            config = get_config()
+        need_more_for_tags = config.setting['save_images_to_tags'] and not config.setting['embed_only_one_front_image']
+        need_more_for_files = config.setting['save_images_to_files'] and not config.setting['save_only_one_front_image']
+        return not need_more_for_tags and not need_more_for_files
+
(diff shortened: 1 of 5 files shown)


phw (approved):
I think this should fix PICARD-3352, from the description it really seems to be because of early termination of image loading.

zas:
> I think this should fix [PICARD-3352](https://tickets.metabrainz.org/browse/PICARD-3352), from the description it really seems to be because of early termination of image loading.

Not sure it actually fixes it, the issue is unclear. I asked for a debug log.

PR #3296, opened 2026-07-28
Make tomlkit optional, disable profile export if tomlkit is not available

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [x] Other

## Problem

Since tomlkit is only strictly needed for profile export, and both profiles and more so their export is non-essential functionality, we can keep this as an optional, but highly recommended, dependency.

We do the same with e.g. PyJWT or charset-normalizer, which we really recommend being installed, but Picard will run when missing (and disable the corresponding functionality).

We still have it as a normal dependency in `pyrpoject.toml`. When installing via pypi this dependency is easily resolved, but it might be missing elsewhere.

## Solution

Conditionally import tomlkit. If not available, disable the export button (but show a explaining tooltip) and in picard-cli show a proper error message.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -15,7 +15,6 @@ For development or building from source, follow the instructions below.
 * [Mutagen 1.45 or newer](https://mutagen.readthedocs.io/)
 * [PyYAML 5.1 or newer](https://pyyaml.org/)
 * [pygit2 1.18.2 or newer](https://www.pygit2.org/) - For plugin system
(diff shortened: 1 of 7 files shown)


zas (approved):
LGTM

PR #3298, opened 2026-07-28
PICARD-3355: Allow plugins to unregister script variables

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Add `unregister_script_variable()` and `unregister_all_script_variables()` to the Plugin API, and prevent duplicate documentation entries when a plugin re-registers the same variable name.

## Problem

* JIRA ticket: PICARD-3355

The plugin API assumes variables are registered once in `enable()` and never change. But plugins with dynamic, config-dependent variable names (e.g. Simple Classical) need to re-register variables at runtime. Currently there is no clean way to unregister variables, and calling `register_script_variable` twice for the same name creates duplicate entries in documentation.

See https://community.metabrainz.org/t/simple-classical-picard-3-plugin-for-tagging-classical-music/815330/8

## Solution

- Add `ExtensionPoint.unregister(module, match)` — a generic method for selective item removal using a callable predicate.
- Add `unregister_script_variable(name, api)` — removes a single variable registered by the calling plugin.
- Add `unregister_all_script_variables(api)` — removes all variables registered by the calling plugin.
- `register_script_variable()` now automatically replaces an existing entry with the same name from the same plugin (deduplication), so plugins can safely call it multiple times without needing to unregister first.
- Expose both unregister functions as `PluginApi` methods.
- Cross-plugin isolation: unregistering only affects the calling plugin's own registrations.

### Plugin usage example

A plugin whose variable names depend on user configuration can now cleanly
re-register them whenever settings change:

```python
def register_variables(api):
    # Clear all previously registered variables for this plugin.
    # This removes stale entries from a previous configuration.
    api.unregister_all_script_variables()

    # Re-register based on current config.
    # Each variable will appear in script/tag editor autocomplete.
    for name in api.plugin_config['tag_names']:
        api.register_script_variable(name, "Custom tag from My Plugin")


def enable(api):
    # Set up a default list of tag names the plugin writes to
    api.plugin_config.register_option('tag_names', ['composer', 'conductor'])

    # Register variables on startup
    register_variables(api)
    api.register_options_page(MyOptionsPage)


class MyOptionsPage(OptionsPage):
    def save(self):
        # User changed tag names in settings — persist them
        self.api.plugin_config['tag_names'] = self.tag_names_input.text().split(', ')

        # Update registered variables so autocomplete reflects
        # the new names immediately, without restarting Picard.
        register_variables(self.api)
```

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Implementation, tests, and documentation were developed with AI assistance under human review and direction.

## Action

Additional actions required:

### DIFF
--- a/picard/extension_points/__init__.py
+++ b/picard/extension_points/__init__.py
@@ -34,7 +34,10 @@
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
 from collections import defaultdict
-from collections.abc import Iterator
+from collections.abc import (
+    Callable,
+    Iterator,
+)
 from typing import Any
 import uuid
 
@@ -67,6 +70,20 @@ class ExtensionPoint:
             # uncomment to debug internal extensions loaded at startup
             # print("ExtensionPoint: %s register <- item=%r" % (self.label, item))
         self.__dict[name].append(item)
+
+    def unregister(self, module: str, match: Callable[[Any], bool]) -> None:
+        """Remove items registered by *module* for which *match* returns True.
+
+        Args:
+            module: Full module path (same value passed to :meth:`register`).
+            match: A callable receiving an item; return True to remove it.
+        """
+        if module.startswith(PLUGIN_MODULE_PREFIX):
+            name = module[PLUGIN_MODULE_PREFIX_LEN:].split('.')[0]
+        else:
+            name = None
+        if name in self.__dict:
+            self.__dict[name] = [item for item in self.__dict[name] if not match(item)]
 
     def unregister_module(self, name: str | None) -> None:
         try:
--- a/picard/extension_points/__init__.py
+++ b/picard/extension_points/__init__.py
@@ -61,14 +61,17 @@ class ExtensionPoint:
         self.__dict = defaultdict(list)
         _extension_points.append(self)
 
-    def register(self, module: str, item: Any) -> None:
+    @staticmethod
+    def _derive_key(module: str) -> str | None:
+        """Derive the internal storage key from a full module path."""
         if module.startswith(PLUGIN_MODULE_PREFIX):
-            name = module[PLUGIN_MODULE_PREFIX_LEN:].split('.')[0]
+            return module[PLUGIN_MODULE_PREFIX_LEN:].split('.')[0]
+        return None
+
+    def register(self, module: str, item: Any) -> None:
+        name = self._derive_key(module)
+        if name is not None:
             log.debug("ExtensionPoint: %s register <- plugin=%r item=%r", self.label, name, item)
-        else:
-            name = None
-            # uncomment to debug internal extensions loaded at startup
-            # print("ExtensionPoint: %s register <- item=%r" % (self.label, item))
         self.__dict[name].append(item)
 
     def unregister(self, module: str, match: Callable[[Any], bool]) -> None:
@@ -78,10 +81,7 @@ class ExtensionPoint:
             module: Full module path (same value passed to :meth:`register`).
             match: A callable receiving an item; return True to remove it.
         """
-        if module.startswith(PLUGIN_MODULE_PREFIX):
-            name = module[PLUGIN_MODULE_PREFIX_LEN:].split('.')[0]
-        else:
-            name = None
+        name = self._derive_key(module)
         if name in self.__dict:
             self.__dict[name] = [item for item in self.__dict[name] if not match(item)]
 
--- a/docs/PLUGINSV3/API.md
+++ b/docs/PLUGINSV3/API.md
@@ -680,6 +680,10 @@ def enable(api):
 
 Register a variable name for script autocomplete.
 
+If the same variable name has already been registered by this plugin, the
+existing entry is replaced (no duplicate is created). This allows plugins to
+safely re-register variables without needing to unregister first.
+
 ```python
 def enable(api):
     api.register_script_variable(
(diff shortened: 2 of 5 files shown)


phw (approved):
LGTM

picard/extension_points/__init__.py
```diff
@@ -58,16 +61,30 @@ def __init__(self, label: str | None = None) -> None:
...
-            # print("ExtensionPoint: %s register <- item=%r" % (self.label, item))
         self.__dict[name].append(item)
 
+    def unregister(self, module: str, match: Callable[[Any], bool]) -> None:
+        """Remove items registered by *module* for which *match* returns True.
+
+        Args:
+            module: Full module path (same value passed to :meth:`register`).
+            match: A callable receiving an item; return True to remove it.
+        """
```
Sophist-UK, line 83: Is there any security to prevent a rogue plugin from adding or removing variables for another plugin (or a phantom plugin)?

Sophist-UK (commented):
As @zas and @phw know, I have previously had concerns about security.

In Picard v2, security was effectively provided by plugins offered by Picard being restricted to the official repository and having each change reviewed by knowledgeable and trusted reviewers. If users wanted to install 3rd party plugins, there were some technical hoops to jump through.

I haven't really kept up with the security thinking around the V3 federated plugin infrastructure, but I believe that:

1. Picard users cannot be expected to be Python programmers capable of reviewing 3rd party plugins. If Picard v3 is going to offer to download 3rd party plugins and remove the technical manual steps that v2 requires for this, then I believe that Picard needs to take greater responsibility for ensuring that the user does NOT install a rogue plugin (such as - worse case - a ransomware virus that encrypts all the user's data!!). Remember, a plugin can run arbitrary python code (though usually in user mode rather than root) and can do literally anything within those constraints. I raised this some time ago, and I believe that this has been taken into account though I haven't kept up and am not sure exactly what the actual solution does.

2. History shows that if an exploit is possible, at some point someone will try it.

3. Consistent with the objective of enabling wider plugin development (and reducing the workload on Picard devs) Picard should therefore do everything it reasonably can (given that Python does **not** have flexible sandbox functionality) to restrict the functionality available to plugins to inflict damage outside its own boundaries.

Looking at the variable registration functionality, I would suggest that the plugin source should be identified using the calling context (stack trace or reflection) rather than by allowing the plugin itself to give its context (which could be faked). I haven't looked at all the other plugin functions to see whether this applies elsewhere - this is simply an example.

Sophist-UK:
Hmmm ... this was merged as I was doing my review. So long as this comments is still reviewed for applicability, I **don't** see this as a problem.

picard/extension_points/__init__.py
phw: We can't effectively prevent a plugin from calling this function directly, but it is not part of the public plugin API. Plugins are supposed to use the functions exposed in `picard/plugin3/api.py`, specifically the ones defined in the [`PluginApi`](https://github.com/metabrainz/picard/blob/master/picard/plugin3/api_impl.py#L190) class.

The signature of the functions there is:

```
def register_script_variable(self, name: str, documentation: str | None = None) -> None:
    ...

def unregister_script_variable(self, name: str) -> None:
    ...

def unregister_all_script_variables(self) -> None:
    ...
```

Needing the module name for registering on the extension point is an implementation detail hidden from plugins. On the discussion in the forums the plugin was using the extension point directly to unregister all variables, and then it needed to set the proper module name. But this is exactly the issue addressed here: There is a use case for a plugin unregistering variables, now there is an official API to do so.

Also `register_script_variable` isn't as special as you seem to believe it is. From all the extension points it is probably the least interesting one. All it does is making the name and a description known to Picard so it can show documentation and help for that variable and autocomplete it in the script editor. It is not needed to set metadata, `Metadata` is still a key / value store accepting any string.

But given the nature of Python we can't prevent a plugin to access other parts of the code or even monkey patch functions. But we won't allow that for any plugin in the registry.

phw:
> Picard being restricted to the official repository

It never was. It was always possible to install plugins from elsewhere, there are several plugins in the wild not provided by the official registry and people do use them.

picard/extension_points/__init__.py
Sophist-UK: I am not sure how easy it would be to make methods and attributes private (which I think effectively means making names dunder), but even then I believe they can still be accessed by name mangling.

The only "firewall"-like approach I can think of is for each and every method to make a call to a security method that uses a stack trace/reflection based approach to confirm that: A) dunder methods have not been called by name mangling; and B) public non-API methods have not been called by plugins. I have never tried to implement this type of approach, so I am not sure whether there would still be ways around this (like monkey patching), and I would also worry about potential performance impacts of such an approach. Furthermore, whilst this could catch name-mangled calls to private methods, I am pretty sure it can't catch name-mangling reads and / or writes to private attributes. 

Ultimately, I think this comes down to Guido's "open kimono" comment whereby he essentially said that Python was intentionally insecure and unsuitable for secure environments.

Ultimately, all security mitigations comes down to risk management - what is the level of risk and what is the effort required to close it? I think that AI both increases the level of risk (I could ask AI to code a rogue plugin and then market it) and decreases the possible costs of mitigation (I can ask AI to create and code a mitigation strategy but only if the language allows a secure approach and Python probably doesn't).

So, in the end I think that it has to come down to human reviews.

I don't know - perhaps I am being paranoid and over estimating the risks and consequences of a rogue plugin and it will either never happen or the consequences will be very limited - but equally I wouldn't want even a single user to lose their lifetime's data to a ransomware attack that is introduced by a rogue plugin downloaded and installed by Picard.

Sophist-UK:
Yes - but to install them you had to:

* Go to the github repo and download the code
* Find the plugins directory and manually copy the files in there.

This implies two things:

1. You can review the source code (if you have the technical skills)
2. You have to do something technical that is well outside the normal UI plugin workflow

My worry is that by making it easy to install plugins that have not been verified as safe by trusted Picard developers, you are making it easier for malware to be introduced.

picard/extension_points/__init__.py
phw: There are some interesting approaches like https://simonwillison.net/2026/Jun/6/micropython-in-a-sandbox/ . This is something to look into.

But anyway, nothing for Picard 3.0. Privilege wise the 3.0 plugin system will be the same as in Picard 2. The advantages are that it does not run plugin code for disabled plugins on startup (which Picard 2 must do for reading the metadata, and it could have side effects) and that it actually has a defined plugin API, that we will extend as needed. There are still some cases where plugins will import other code from Picard, but these can be evaluated and if it is a common use case the API will be extended.

Sophist-UK: Yes - micropython sandbox is interesting, but it is essentially a small stand-alone environment implemented in wasm, so almost certainly not directly applicable here.

My guess is that for our environment it would probably be more sensible to use an o/s sandbox implementation and run plugins in a sub-process with an API wrapper that only exposes the API and not other methods.

PR #3299, opened 2026-07-29
config: Separate volatile UI state from valuable application state

zas:
## Summary

Preparatory work toward separating volatile UI state (`persist`) from valuable application state, motivated by [PICARD-3354](https://tickets.metabrainz.org/browse/PICARD-3354).

Introduces a new `state` config section for data that is valuable (should survive config resets and be syncable across machines) but is not user-configurable settings. Migrates 13 keys from `persist` to `state` and exposes `plugin_state`/`plugin_persist` in the plugin v3 API.

### Mechanical prep (commits 1-4, no behavior change)
- Add `ConfigSection.keys()` method for section key enumeration
- Add `Config.sync()` override as extension point for future multi-store sync
- Refactor `clear_qt5_state` upgrade hook to use section API instead of raw `allKeys()`
- Decouple `ConfigSection` from `Config` type (accept `QSettings`, rename internal attribute)

### Feature work (commits 5-7)
- Add `state` ConfigSection to `Config` for valuable non-volatile data
- Add `plugin_state` and `plugin_persist` properties to `PluginApi`
- Migrate 13 keys from `persist` → `state` with upgrade hook at 3.0.0b10

### Keys migrated to `state`
- OAuth credentials: `oauth_access_token`, `oauth_access_token_expires`, `oauth_refresh_token`, `oauth_refresh_token_scopes`, `oauth_username`
- Progress flags: `tutorial_steps_shown`, `tutorial_disabled`, `setup_wizard_completed`, `show_plugin_install_warning`
- Session management: `last_session_path`, `recent_sessions`, `session_autosave_path`
- CLI state: `cli_plugins_init`

### What remains in `persist` (volatile, per-machine)
Window geometry, splitter positions, column widths/lock state, panel visibility toggles, media player volume/rate, file browser path, filter columns, metadata box state.

## Problem

Picard writes user configuration and volatile UI state into a single INI file. The `[persist]` section holds window geometry, splitter positions, etc. (rewritten on nearly every quit) alongside OAuth tokens, tutorial flags, and session paths that have real value to the user. This makes the config file impossible to share across machines and noisy to back up or version-control.

## Solution

Introduce semantic separation now (while in beta) so the sections have clean, well-defined purposes:
- `setting`: user preferences, profile-aware
- `state`: valuable application state, non-profile, syncable  
- `persist`: volatile per-machine UI state, disposable

This also prepares for a future file split (PICARD-3354) where `persist` would live in its own file.

## AI Usage

This PR was developed with AI assistance (Kiro CLI) for code audit, implementation, and test writing. All changes were verified by the existing test suite (5633 tests passing).

## Tickets to create

The following follow-up tickets should be created on [tickets.metabrainz.org](https://tickets.metabrainz.org/projects/PICARD):

- **Separate persist section into its own file** — The actual file split proposed in PICARD-3354. With the semantic separation done, this becomes: construct a second `QSettings` for persist, extend `Config.sync()`, update backup/restore to handle two files, add `PICARD_PERSIST_FILE` env var / `--persist-file` CLI option.
- **Document plugin_state and plugin_persist in plugin v3 docs** — Update `docs/PLUGINSV3/API.md` with usage examples and guidelines for when to use each storage type.
- **Consider OS keychain for OAuth tokens** — The tokens are now in `state` (syncable), but ideally credentials would use the OS keychain (libsecret/Keychain/Credential Manager). This would be a further improvement on top of the current separation.
- **Add PICARD_VERSION bump to 3.0.0b10** — The upgrade hook is registered at `3.0.0b10` but `PICARD_VERSION` is still at `3.0.0b9`. Version bump needed before release.

### DIFF
--- a/picard/collection.py
+++ b/picard/collection.py
@@ -203,7 +203,7 @@ def add_release_to_user_collections(release_node: dict[str, Any]) -> None:
     if 'collections' in release_node:
         release_id = release_node['id']
         config = get_config()
-        username = config.persist['oauth_username'].lower()
+        username = config.state['oauth_username'].lower()
         for node in release_node['collections']:
             if node['editor'].lower() == username:
                 collection = get_user_collection(node['id'])
--- a/picard/config.py
+++ b/picard/config.py
@@ -677,6 +677,14 @@ class Config(QtCore.QSettings):
             return False
         return self._save_backup(backup_path)
 
+    def sync(self):
+        """Write all pending configuration changes to disk.
+
+        Override of QSettings.sync() to provide a single extension point
+        for syncing additional backing stores in the future.
+        """
+        super().sync()
+
 
 class OptionError(Exception):
     def __init__(self, message, section, name):
--- a/picard/config.py
+++ b/picard/config.py
@@ -214,9 +214,9 @@ class ConfigSection(QtCore.QObject):
     # Signal emitted when the value of a setting has changed.
     setting_changed = QtCore.pyqtSignal(str, object, object)
 
-    def __init__(self, config: 'Config', name: str):
+    def __init__(self, config: QtCore.QSettings, name: str):
         super().__init__()
-        self.__qt_config = config
+        self._config_store = config
         self.__name = name
         self.__prefix = self.__name + '/'
         self._memoization: dict[str, Memovar] = defaultdict(Memovar)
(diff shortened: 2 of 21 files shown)


zas:
Closing per review feedback from @outsidecontext. The core config system changes (state section, key migration, prep refactoring) are out of scope for 3.0. Will open a smaller PR that only exposes plugin_persist to the plugin v3 API.

PR #3302, opened 2026-08-01
Generate match icons using a script, rather than using static SVGs

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Replace hand-crafted SVG/PNG match quality icons with script-generated PNGs and simplify the runtime to a threshold-based lookup.

## Problem

The old approach used 12 hand-crafted SVGs to produce 24 static PNGs (6 levels × normal/pending × 1×/2×). Adjusting colours, fill levels, or adding new thresholds required editing SVGs in a vector editor, exporting, and manually replacing files. The runtime code had a fragile index mapping (`int(similarity * 5 + 0.5)`) and circular import workarounds to access the icons.

## Solution

- New generator script (`scripts/tools/generate_match_icons.py`) renders bookmark-shaped icons using QPainter with configurable colours, thresholds, and fill levels. Run with `--sheet` to produce a visual reference PNG.
- 6 distinct levels (same count as before), each with 2 px fill height difference at 16 px, ensuring clear visual separation.
- Similarities below 0.5625 are squashed into one "poor match" red icon.
- Only exact 1.0 gets the fully-filled green "perfect match" icon, distinguishing it from near-perfect matches.
- Runtime module (`picard/ui/match_icons.py`) is ~80 lines: `similarity_to_level()` + `load_icons()` + two icon lists. No QPainter at runtime, no caching, no threading concerns.
- `match_quality_column.py` simplified: `THRESHOLD_TO_ICON_INDEX` mapping and `FileItem` circular-import workaround removed.
- Removes 12 SVG source files; replaces 24 old PNGs with 24 new optimized ones (3.8 KB total).
- Adds `scripts/tools/README.md` documenting all tool scripts.
- 23 unit tests for the level mapping logic.

match_icons_sheet

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

Developed with Kiro (AI assistant).

## Action

Additional actions required:


phw (changes requested):
I don't think we should do this. This is calculating a new icon for each row in the list, where we could use pre-calculated images just as well. The main change here seems to visualize matches lower than 50 also more clearly. But we could extend the images to go lower (but probably cap it still at 30% or such, because below the color coding becomes less visible). In my testing of this branch I didn't get results that differed much from the previous Image. I got some that maybe looked like 40%.

If we want to automate the image creation we could put the generation in a script and run it to generate the static icons and replace the SVGs with it. Or we calculate on startup and reuse the same QIcon. But again it would be unclear why we don't just statically provide the image data then.

> e.g., a 91 % match from a 99 % match, and — more importantly — the 100 % icon was visually similar to the saved checkmark

I think both of these statements have problems. I don't think it is necessary to have a tiny difference in images because of some single digit percentage difference. Between the current 100% and 90% match icons there is 1px difference, which doesn't leave much room for visually difference of 91% and 99% match.

The difference calculation is a rough indicator anyway. Also I don't think the rectangle of 100% match is visually similar to the checkbox.

zas:
> This is calculating a new icon for each row in the list

No, that's calculating a new icon for each value, rounded to 5%, and only if needed.

So if we have 100 rows with values 91, 92, 94 % it will be ONE icon.

Also, my benchmarks did show that generating the icon is actually faster than loading it from disk (because it is small and simple).
Note: in practice, there are loaded as part of resources in memory once. So not from disk.

 Time needed to generate one is negligible and it happens once per session only. So there's zero advantage in static icons (for this).

Pre-calculation (at start) isn't needed, the only gain would the cache check time, totally negligible.
Also cache could be improve even more (but not worth imho).

But you are missing the goal here: experimenting with colors and shape for this indicator. Some people don't see colors well, so perhaps we can even make those colors configurable (that's not possible with static icons).

To me, SVGs for this is overkill, and has no added value. 

> Between the current 100% and 90% match icons there is 1px difference, which doesn't leave much room for visually difference of 91% and 99% match.

There is now more difference than before, because 99% and 100% were identical, now they differ.
Also < 50% now differs from 50%.
We had 10% steps, now 5% (but I could revert back to 10% easily, as this is dynamic).

Still experimental.

zas:
## Generated icon sheet

All 6 levels (normal + pending), with similarity ranges as column headers.

Generated with:
```
python scripts/tools/generate_match_icons.py --sheet /tmp/match_icons_sheet.png
```

Configuration (in `picard/ui/match_icons.py`):
- `LOW_THRESHOLD = 0.5625` — below this → single red icon
- `NUM_STEPS = 4` — levels between poor and perfect
- 2 px fill height difference per level at 16 px
- 24 optimized PNGs, 3.8 KB total

match_icons_sheet

phw (approved):
LGTM

phw:
@zas I added PICARD-3366 for the changelog

PR #3307, opened 2026-08-04
PICARD-3367: Add shell completion generation for picard-cli

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add shell completion generation for `picard-cli` using `shtab`. Completions are derived from the argparse parser structure and support all shells provided by the installed shtab version.

## Problem
- JIRA ticket (optional): PICARD-3367

`picard-cli` has a growing number of subcommands (`plugins`, `profiles`, and more to come) but no shell completion support, making discoverability harder for users.

## Solution

- Add `picard-cli completions <shell>` subcommand that generates completion scripts for any shell supported by shtab (currently bash, zsh, tcsh, fish).
- Add `python setup.py build_completions` command for packagers to generate static completion scripts at build time (output to `build/completions/`).
- Supported shells are dynamically derived from `shtab.SUPPORTED_SHELLS` — no hardcoding.
- `shtab` is fully optional: everything works without it, the error is clear when it's needed but missing.
- Adjust `_register_subcommand` to not override `run_command` if the module's `setup_parser()` already set one, enabling subcommands without verb subparsers.
- `shtab` added as a dev dependency.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed through interactive AI-assisted session with human review and direction at each step.

## Action

Additional actions required:


pyproject.toml
```diff
@@ -80,6 +80,7 @@ dev = [
     "pytest-randomly>=4.0.1",
     "pytest-xdist>=3.8.0",
     "ruff>=0.13.0",
+    "shtab>=1.7.0",
```
phw: I think we should put this into an optional dependency group instead. Let's have one called `cli` or such. That way it is available for install using pip with `pip install "picard[cli]"`.

phw (changes requested):
Thanks a lot. I already wanted to look into this myself. Looks good to me. I'd only argue to put shtab into an optional dependency. Otherwise this is good to go.

pyproject.toml
zas: Done, and rebased

phw (approved):
LGTM

PR #3308, opened 2026-08-04
PICARD-3363: Allow automatically removal of completed albums after save

rdswift:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

Picard does not have an option to automatically remove fully completed albums after they have been saved.  This would allow automatically cleaning up the albums list when processing large nnumbers of albums.

* JIRA ticket (_optional_): PICARD-3363

## Solution

- Add a new `remove_complete_albums_after_save` configuration option under the "General" section.
~~- Add a signal to the `Album` class indicating that the album has been updated.~~
~~- Emit the new `album_updated` signal when a related file is saved.~~
~~- Monitor album changes to trigger removal of the album it is complete and all related files have been saved.~~
- Add a post file save processor to trigger removal of the album it is complete and all related files have been saved.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:


picard/ui/mainwindow/__init__.py
```diff
@@ -1391,6 +1396,25 @@ def open_support_forum(self):
     def open_donation_page(self):
         webbrowser2.open('donate')
 
+    def file_post_save(self, obj: File):
```
zas: `_remove_album_if_complete` would be a more descriptive name IMHO.

picard/ui/mainwindow/__init__.py
```diff
@@ -1391,6 +1396,25 @@ def open_support_forum(self):
...
+        if not config.setting['remove_complete_albums_after_save']:
+            return
+
+        # Try getting the related album
+        try:
+            album = obj.parent_item.album
+        except AttributeError:
+            return
+
+        if album.is_complete() and album.get_num_unsaved_files() == 0:
```
zas: There's a small risk that the album was already removed by a precedent call.

We could add something like:

```python
      if album.id not in self.tagger.albums:
          return
```

Once `panel.remove([album])` triggers `tagger.remove_album()`, the album is deleted from the dict, so this check will avoid any further remove call.

zas (changes requested):

picard/ui/mainwindow/__init__.py
rdswift: Changed in https://github.com/metabrainz/picard/pull/3308/commits/306fa788744876d1e9a919b4bc3c5200b7a60a74

picard/ui/mainwindow/__init__.py
rdswift: LOL, I was just looking at exactly that same issue.  Also had to add a check for a valid `self.tagger` to allow the proper execution of `test_preferencelistwidget.py`. 

Changes in https://github.com/metabrainz/picard/pull/3308/commits/7a8a951b5a23f15e57822ac05a2455bd8c25167d

I suspect there may still be an issue because this test is occasionally aborting randomly when run locally.  It usually passes, but occasionally fails without any changes to the code.  I'm wondering if it has somethhing to do with the order that the tests are run or if tests are being run concurrently.

picard/ui/mainwindow/__init__.py
```diff
@@ -1391,6 +1396,32 @@ def open_support_forum(self):
...
+
+        # Try getting the related album
+        try:
+            album = obj.parent_item.album
+        except AttributeError:
+            return
+
+        # Check for existing tagger to allow proper execution of test_preferencelistwidget.py
+        if not self.tagger:
+            return
```
zas: I don't understand why this is needed.
What do you mean by "allow proper execution of test_preferencelistwidget.py" ?
I noticed a problem with tests, sometimes crashing, but I don't think this is related.

zas (changes requested):
Yes, it looks much simpler

picard/ui/mainwindow/__init__.py
zas: @rdswift this isn't needed, the issue was elsewhere, it should be fixed by https://github.com/metabrainz/picard/pull/3310

picard/ui/mainwindow/__init__.py
```diff
@@ -249,6 +249,11 @@ def __init__(self, parent=None, disable_player=False):
         webservice_manager.authenticationRequired.connect(self._show_password_dialog)
         webservice_manager.proxyAuthenticationRequired.connect(self._show_proxy_dialog)
 
+        # Imported here to allow proper execution of test_preferencelistwidget.py
+        from picard.extension_points.event_hooks import register_file_post_save_processor
+
+        register_file_post_save_processor(self._remove_album_if_complete)
```
phw: Instead of the main window this should be better put into the `Tagger` application instance. There is nothing UI specific to this registration.

phw (changes requested):
Looks good, but the code should not be in mainwindow but inside of Tagger. That then also makes the check for whether `self.tagger` is defined obsolete.

picard/ui/mainwindow/__init__.py
rdswift: That makes perfect sense.  Changes made in https://github.com/metabrainz/picard/pull/3308/commits/7eca8b4b0443662d80940798f6d50e59c41cc8e1

phw (approved):
Clean, simple and very useful

zas (approved):
LGTM!

PR #3312, opened 2026-08-07
Disable ISRC extraction if not supported by libdiscid on the platform

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

libdiscid does not support ISRC extraction on all platforms. If the ISRC feature is unavailable, it does not make sense to allow the user to enable the feature.

Note: From currently realistically used platforms this affects only Haiku. 

## Solution

If ISRC is not supported by libdiscid, disable the checkbox in options and show a tooltip explaining it. Also set the default value of the "read_isrcs_from_disc" option to `False`.

Add a new `cdrom.has_isrc_support` helper with tests.

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/disc/__init__.py
+++ b/picard/disc/__init__.py
@@ -31,7 +31,6 @@
 
 from functools import partial
 import traceback
-from types import ModuleType
 
 from picard import (
     log,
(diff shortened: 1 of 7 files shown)


zas (approved):
LGTM

rdswift:
Should I include a note in the documentation that this option might not be available on some platforms?

phw:
> Should I include a note in the documentation that this option might not be available on some platforms?

@rdswift  Yes, I think mentioning it in some way would be good. But it is important that this is really a niche case, all three major operating systems and the BSDs support it. The only platform that doesn't and might be realistically used is Haiku. So I'm not quite sure how to phrase it.

PR #3313, opened 2026-08-07
PICARD-3368: Batch object updates when removing multiple tags

cryobry:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

Removing multiple tags from multiple tracks is slow and freezes the UI for an unnecessarily long time.

Previously, removing N tags from M selected objects called obj.update() NxM times. Each call performs a full metadata comparison and UI refresh, making bulk tag removal noticeably slow with large selections.

* JIRA ticket (_optional_): PICARD-3368

## Solution

Introduce `_remove_tags()` which applies all metadata deletions first using the existing `_set_tag_values_delayed_updates()` generator, accumulates the affected objects, then calls obj.update() exactly once per object.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

Used to determine if there were existing batching methods and to implement objects_to_update.update() batching.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/metadatabox/__init__.py
+++ b/picard/ui/metadatabox/__init__.py
@@ -623,7 +623,7 @@ class MetadataBox(QtWidgets.QTableWidget):
         Adds actions for removing, restoring, and merging tags to the context menu.
         """
         remove_tag_action = QtGui.QAction(_("Remove"), self)
-        remove_tag_action.triggered.connect(partial(self._apply_update_funcs, removals))
+        remove_tag_action.triggered.connect(partial(self._remove_tags, removals))
         remove_tag_action.setShortcut(self.remove_tag_shortcut.key())
         remove_tag_action.setEnabled(bool(removals))
         menu.addAction(remove_tag_action)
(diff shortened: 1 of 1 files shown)


picard/ui/metadatabox/__init__.py
```diff
@@ -767,11 +767,17 @@ def _set_tag_values(self, tag, values, objects=None):
     def _remove_tag(self, tag):
         self._set_tag_values(tag, [])
 
-    def remove_selected_tags(self):
-        for tag in self._selected_tags(filter_func=self._tag_is_removable):
-            self._remove_tag(tag)
```
zas, line 772: `_remove_tag()` isn't called anywhere after this change right? If so, I think it can be totally removed.

zas (changes requested):
Apart the now unused method we can remove, the patch looks good to me.
It also reveals we need a better code separation and unit tests for those metadata operations. I may provide a refactoring patch after this is merged.

zas (approved):
LGTM

phw (approved):
Thanks for this patch

PR #3315, opened 2026-08-08
PICARD-1638: Show authorization required dialog only once

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: When Picard makes multiple requests requiring MusicBrainz authentication (e.g. with ratings or collections enabled), the "Authentication Required" dialog is now shown only once instead of for every failed request. If the user declines to log in, albums still load without the user-specific data.

## Problem

* JIRA ticket: [PICARD-1638](https://tickets.metabrainz.org/browse/PICARD-1638)

When features like track ratings or collections are enabled and the user is not logged in, each album lookup triggers an `authenticationRequired` signal from Qt, resulting in a flood of modal "Authentication Required" dialogs — one per request. This makes Picard unusable (users reported needing to click through 6-7 dialogs, sometimes causing crashes as reported in PICARD-3260).

The root cause, as identified by @phw in ticket comments, is that Qt's `authenticationRequired` signal blocks the event loop synchronously. The OAuth login flow (which requires network requests) cannot work inside that handler. Queued requests each trigger a new dialog as soon as the previous one closes.

## Solution

Instead of relying on Qt's `authenticationRequired` signal for the MB server, we handle HTTP 401 responses in `WebService._handle_reply`:

### Core mechanism
- When an `mblogin` request gets a 401, it's queued in `_awaiting_authorization` and a new `authorization_required` signal is emitted (only once, regardless of how many requests fail).
- `MainWindow` connects to this signal and shows the auth dialog exactly once.
- **On successful login**: all queued requests are retried with the new token.
- **On declined/failed login**: GET requests are retried without user-specific `inc` params (`user-ratings`, `user-collections`, `user-tags`, `user-genres`) so albums still load with public data. Non-GET requests (submissions) are failed with an auth error.

### Additional improvements
- **Race condition handling**: If a 401 arrives on a stale request (sent before login) but the user has since authenticated, the request is retried immediately rather than triggering a new dialog.
- **Queue cap**: The authorization queue is limited to 100 requests to prevent unbounded growth.
- **Type safety**: A `ReplyLike` protocol is introduced so that the handler interface is explicit and type-checked.
- **User feedback**: Statusbar messages show the auth flow state (waiting, success, failure, declined).
- **Improved dialog message**: Dynamically lists which enabled features require login (ratings, collections, genres) based on the user's current settings.
- **Consistent UI**: All auth-requiring options now show "(requires login)" in their label. The user collections checkbox is no longer disabled when not logged in — it gracefully degrades like ratings and genres.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code was developed collaboratively with an AI assistant (Claude via Kiro CLI). The human provided direction on architecture (following @phw's guidance from the ticket comments), reviewed behavior through manual testing against the live MusicBrainz API, caught bugs (URL encoding issue with QUrlQuery), guided UX decisions (dialog wording, i18n approach, graceful degradation behavior), and verified the fix works in real usage scenarios.

## Action

Additional actions required:

Follow-up considerations:
- The "Enable managing user collections" option might be better placed on the Metadata page rather than the General/Account page, but that's a separate UX discussion.


rdswift (dismissed):
This looks okay to me.  Much nicer user experience.

picard/ui/options/general.py
```diff
@@ -119,27 +121,32 @@ def update_server_host(self):
...
             return
+        oauth_manager = self.tagger.webservice.oauth_manager
+        log.debug(
+            "update_login_logout: is_logged_in=%s, is_authorized=%s, username=%r, error_msg=%r",
+            oauth_manager.is_logged_in(),
+            oauth_manager.is_authorized(),
+            oauth_manager.username,
+            error_msg,
+        )
         if self.tagger.webservice.oauth_manager.is_logged_in():
```
phw: Use variable `oauth_manager` here directly

picard/webservice/__init__.py
```diff
@@ -82,6 +84,7 @@
 COUNT_REQUESTS_DELAY_MS = 250
 
 TEMP_ERRORS_RETRIES = 5
+MAX_PENDING_AUTHORIZATION_REQUESTS = 100
```
phw: Any big downside of having this higher? I agree we should have a limit, but the cost seems reasonable small that we could also have it set to 1000 or such?

picard/oauth.py
```diff
@@ -197,6 +197,7 @@ def _on_revoke_token_finished(self, callback, data, http, error):
     def forget_refresh_token(self):
         del self.refresh_token
         del self.refresh_token_scopes
+        self.webservice.authorization_state_changed.emit()
```
phw: It's a bit bad practice to trigger the signal on a different object. I think it would be cleaner if the oauth manager class provided `authorization_state_changed` signal. Webservice could connect this to its own `authorization_state_changed` signal.

phw (changes requested):
This is a much needed change. Tested, works as expected. A few minor comments on the code, but otherwise ready to go.

picard/ui/options/general.py
zas: Done in https://github.com/metabrainz/picard/pull/3315/commits/32c7b22815afee7d3ae9edb4c8b61f126bbdc5b5

picard/webservice/__init__.py
zas: Set to 1000 in https://github.com/metabrainz/picard/pull/3315/commits/a2e37fe74a1e533ef6feccfd3fbb1907fae1e497

picard/oauth.py
zas: Done in https://github.com/metabrainz/picard/pull/3315/commits/1361e711918d9fdc8d4eb36cc9c005a9df4751be
Since we add a signal to it, `OAuthManager` is now a `QObject`.

phw (approved):
LGTM

PR #3322, opened 2026-08-10
PICARD-3373: Reorganize options pages

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Reorganize the Options dialog to reduce top-level clutter, create a new Lookup page grouping identification settings, and consolidate matching/completeness options on the Matching page.

## Problem

* JIRA ticket: [PICARD-3373](https://tickets.metabrainz.org/browse/PICARD-3373)

The Options dialog has too many top-level items, some commonly-used settings are buried under "Advanced", and related options are scattered across unrelated pages.

## Solution

Each commit is independently revertible.

### Page reorganization

* Move Option Profiles to bottom of the tree (sort order 95)
* Move Network from Advanced to General
* Move Audio Player from top-level to General
* Fix duplicate sort orders in Interface and Metadata children

### New Lookup page

* Create a top-level **Lookup** page grouping identification/search settings
* Move `analyze_new_files`, `cluster_new_files`, `ignore_file_mbids` from General
* Move `query_limit` from Advanced
* Move `builtin_search`, `use_adv_search_syntax` from Interface
* Re-parent CD Lookup and Fingerprinting under Lookup

### Consolidate Matching page

* Move `ignore_track_duration_difference_under` from Advanced
* Move `completeness_ignore_*` from Advanced
* Move `compare_ignore_tags` from Advanced

All "how strictly does Picard compare/evaluate?" settings now live together.

### Resulting tree structure

```
General (1)
├── Startup (5)
├── Network (20)
└── Audio Player (25)
Metadata (20)
├── Preferred Releases (10)
├── Genres (20)
└── Ratings (25)
Lookup (25)
├── CD Lookup (10)
└── Fingerprinting (15)
Tags (30)
├── ID3 (30)
├── AAC (40)
├── AC3 (50)
└── WAVE (60)
Cover Art (35)
└── Processing (0)
File Naming (40)
└── Compatibility
Plugins (70)
Scripting (75)
User Interface (80)
├── Colors (20)
├── Top Tags (30)
├── Quick Settings Menu (40)
├── Cover Art Box (50)
└── Action Toolbar (60)
Advanced (90)
├── Matching (30)
├── Sessions (90)
└── Maintenance (99)
Option Profiles (95)
```

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI-assisted (Kiro) with human design decisions and review feedback integration.

## Action

Additional actions required:


rdswift:
In general, I think your changes are good in they they seem to make the option tree a bit more organized.  There are a couple of things that I want to raise for consideration:

1. There is already confusion with meny users about the difference between a tagging script and a file naming script.  Putting "Scripting" under "File Naming" may actually increase the confusion.  Perhaps consider putting "Scripting" under "Tags" instead, because it deals with tagger scripts.

2. I wonder if "Preferred Releases" and "Matching" should be combined into a single page because they both deal with matching.  The down side is that the page could be rather lengthy.

3. The "Ignore the following tracks when determining whether a release is complete:" and "Tags to ignore for comparison:" settings should probably be on the same page because both deal with completion marking in the Album pane.  Both seem out of place where they are located.

4. Not a big deal, but I might be tempted to put the "Processing" page at the bottom of the "Cover Art" tree items because it deals more with something that happens after an image is selected based one of the other provider settings.

5. The items on the "Compatibility" page could possibly be incorporated into a new section on the main "File Naming" page because there seems to be a lot of space currenly used for the "Files will be named like this:" section.  I'm actually on the fence about that, because it may take up too much room and end up forcing the page to scroll.

Not necessarily related to this PR, I also noticed a few formatting issues.  For example:

image
Text line cut off.

image
Needs a spacer at the bottom.

Once the final placement is determined, we might also consider adding a test playground for the regex entered here.

zas:
@rdswift I took your feedback in account in last commits, see if it works better for you now. The spacer issue was fixed, but the long text issue isn't, as it was preexisting, this is something we should address in all pages anyway.

phw (changes requested):
Generally good to rework this a bit. But there are several things that I think we shouldn't do.

- Scripting should not be below filenaming. This would only serve to confuse this even more and gives the impression that scripting is about filenaming, which it isn't. You could argue it could go below Metadata, but I actually would keep it top-level, as it is one of the core functionalities.
- I don't think we should move the "Ignore the following tracks..." options from advanced to metadata, for several reasons:
  1. It actually doesn't impact metadata
  2. The metadata page is already full. When we introduced scrolling on all option pages I already said that IMHO option pages in general should be usable without scrolling. If at all, we should split this page, e.g. by extracting the translations part into a separate page.
  3. It really is an advanced feature.
- Similar to the "Ignore the following tracks", "Tags to ignore for comparison" should I think also not go into tags. Similar to the other option, it doesn't affect actual tags.
- Also, "Ignore the following tracks...", "Tags to ignore for comparison" and "Ignore track durations under the following seconds" all serve related purposes. Tearing those options apart I don't see as useful. What could work is placing all three in the Matching dialog, below but separate to the thresholds.
- I think Matching should stay under advanced as well. This is really something you usually should not mess around with, at least the thresholds.
- ~~The Advanced page is now broken, missing a stretcher at the end~~
- If you actually want to move something from Advanced I think Sessions could go under General

What I really like is the "Lookup" option and moving CD and fingerprinting options below it.

Also if we want to do this we should also update the documentation before the release this change is included in.

rdswift:
> Also if we want to do this we should also update the documentation before the release this change is included in.

Good idea.  I'll probably need a day or two to get all the rearranging done before you guys have a chance to review it.

phw:
While writing my previous review changes where pushed that might address some of the points. Please ignore those.

For the Scripting it is good that it was moved, but I think it should be Metadata, not Tags, if we want to put it below something. Tags is about the file format details, Metadata generally about metadata handling, which is what scripting is concerned about. My personal opinion is still keeping it a top level, though.

zas:
> My personal opinion is still keeping it a top level, though.

I moved it back to top level. That's better I agree.

About Sessions, moving to General would make it easier to discover, but that's quite advanced use, so dunno. Keeping where it is for now.

Do you see further moves we should do?

rdswift:
No "must have" changes from me.  Looks good.

phw (changes requested):
One last thing: The "Remove complete albums after saving" does not fit into "Tags". I'm not fully sure where to put it, though. Probably still in General, or Advanced.

If we keep it in Tags (but I wouldn't, it really is a bit lost and unexpected there) we need to fix the tab order on this page. "Remove complete albums after saving" is currently outside the expected order.

Apart from this I agree this is an improvement and ready to be merged.

rdswift:
> One last thing: The "Remove complete albums after saving" does not fit into "Tags". I'm not fully sure where to put it, though. Probably still in General, or Advanced.
> 
> If we keep it in Tags (but I wouldn't, it really is a bit lost and unexpected there) we need to fix the tab order on this page. "Remove complete albums after saving" is currently outside the expected order.

I agree that it doesn't really belong where it is.  I also am not exactly where to put it.  If we had enough related options, perhaps a new "File Saving" settings page.  Otherwise, my vote would be to place it under "Advanced".

zas:
"Remove complete albums after saving" doesn't belong anywhere, in Advanced it looks weird, in Tags it doesn't fit, creating a new section "Saving" would be very thin (not much to put there) and re-add another section. 
So for now, I'll move it back to where it was, it will be part of a future revamp perhaps.

phw (approved):

rdswift (approved):
Looks good to me.  Thanks.

PR #3334, opened 2026-08-11
PICARD-3374: Pass PluginApi instance to ripper log callback

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-3374

`PluginApi.register_disc_log_reader` was defined as:

```python
def register_disc_log_reader(self, function: Callable[[str], 'TocNumbers']) -> None:
    ...
```

Unlike other PluginApi registration functions that take a callback, the callback here does not receive the PluginApi instance. This is inconsistent, compare e.g.:

```python
def register_file_post_load_processor(
        self, function: Callable[['PluginApi', File], None], priority: int = 0
    ) -> None:
```

Also `register_disc_log_reader` being a new API we can use the chance and pass a `pathlib.Path`.

## Solution

Change `register_disc_log_reader` to accept a `PluginApi` instance as first parameter and update the code to pass a `pathlib.Path` as second parameter:

```python
def register_disc_log_reader(self, function: Callable[['PluginApi', Path], 'TocNumbers']) -> None:
    ...
```

Update all internal implementations accordingly. Also relax the type hints on the `detect_file_encoding` util function to also accept `Path` (no functional change needed).

Note: This is a breaking change compared to beta 7, but I think we need to get this right before Picard 3 final release.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/docs/PLUGINSV3/API.md
+++ b/docs/PLUGINSV3/API.md
@@ -1122,7 +1122,7 @@ from picard.disc.utils import (
 )
 
 
-def my_ripper_toc_from_file(path):
+def my_ripper_toc_from_file(api, path):
     """Parse MyRipper log files for disc ID lookup."""
     entries = []
     with open(path, encoding='utf-8') as f:
--- a/picard/disc/cyanriplog.py
+++ b/picard/disc/cyanriplog.py
@@ -23,6 +23,7 @@ from collections.abc import (
     Iterable,
     Iterator,
 )
+from pathlib import Path
 import re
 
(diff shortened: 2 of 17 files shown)


zas (approved):
Good move, that's a breaking change but only between beta releases, and that's definitively better this way, LGTM.

PR #3335, opened 2026-08-12
PluginApi: _plugin_dir and _plugin_module must not be optional

phw:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

## Problem

The plugin API exposes a `plugin_dir` property, which was defined to return `str | None`, indicating that the plugin_dir is optional and might no exist.

```
@property
    def plugin_dir(self) -> Path | None:
```

This is actually not true. At the point the `PluginApi` is created the plugin must exist and be loaded. `plugin_dir` should never be None. To make this clear to plugins using this property, the type hint needs to be changed.

A similar situation exists for plugin_module.

## Solution

When initializing PluginApi both plugin_dir and plugin_module are known and present. Change `PluginApi` to accept both in the constructor as mandatory arguments and adjust the type hints.

Update tests as needed.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/plugin3/api_impl.py
+++ b/picard/plugin3/api_impl.py
@@ -205,10 +205,10 @@ class PluginApi:
     _module_cache: ClassVar[dict[str, 'PluginApi']] = {}  # Maps module name -> PluginApi instance (for faster lookup)
     _deprecation_warnings_emitted: ClassVar[set[tuple[str, str, int]]] = set()  # Track emitted deprecation warnings
 
-    def __init__(self, manifest: PluginManifest, tagger: 'Tagger') -> None:
+    def __init__(self, manifest: PluginManifest, tagger: 'Tagger', module: types.ModuleType, plugin_dir: Path) -> None:
         self._tagger: Tagger = tagger
         self._manifest = manifest
-        self._plugin_module: types.ModuleType | None = None  # Will be set when plugin is enabled
+        self._plugin_module: types.ModuleType = module
         self._plugin_id = manifest.module_name
         full_name = f'plugin.{self._manifest.uuid}'
         self._logger = getLogger(f'main.plugin.{self._manifest.module_name}')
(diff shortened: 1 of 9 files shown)


test/plugins3/helpers.py
```diff
@@ -482,7 +482,7 @@ def create_test_plugin_api(plugin_dir, locale='en', module_name='test'):
 
     with open(manifest_path, 'rb') as f:
         manifest = PluginManifest(module_name, f)
-        api = PluginApi(manifest, Mock())
+        api = PluginApi(manifest, Mock(), Mock(), plugin_dir)
         api._plugin_dir = plugin_dir
```
zas: This line is now redundant:

```suggestion
```

picard/plugin3/api_impl.py
```diff
@@ -249,7 +249,7 @@ def _get_caller_info(frame_depth=2):
         return plugin_name, filename, lineno
 
     @classmethod
-    def deprecation_warning(cls, message, *args, frame_depth=3):
+    def deprecation_warning(cls, message: str, *args: tuple[Any, ...], frame_depth: int = 3):
```
zas:
```suggestion
    def deprecation_warning(cls, message: str, *args: Any, frame_depth: int = 3):
```

https://peps.python.org/pep-0484/#arbitrary-argument-lists-and-default-argument-values

According to this document, `*args: str `means each arg is a `str`, not that `args` is a `str`.
By the same logic, `*args: tuple[Any, ...]` means each individual arg must be a `tuple[Any, ...]`, which is wrong in this case.

It is called only once in: https://github.com/metabrainz/picard/blob/e8e0219fcb0228246f1abb3605f5e1589e7e4205/picard/album.py#L303-L312

zas (changes requested):
Overall it looks good to me, just 2 minor things to address before merging

picard/plugin3/api_impl.py
phw: Fixed

test/plugins3/helpers.py
phw: Fixed

zas (approved):
LGTM

PR #3337, opened 2026-08-12
PICARD-3378: Add $get_new() and $get_original() script functions

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Add `$get_new(name)` and `$get_original(name)` script functions to explicitly access new (track/MB) or original (file) metadata values.

## Problem

* JIRA ticket: PICARD-3378

Users writing tagger scripts to fill in missing MusicBrainz data run into a confusing situation: `$if(%composer%,,...)` sees file's existing tags through the `MultiMetadataProxy` fallback, making it impossible to check "did MusicBrainz actually provide this tag?"

For example, this common pattern doesn't work as expected:
```
$if(%composer%,,$set(composer,$if2(%lyricist%,%writer%,%arranger%)))
```

If the file already has a composer tag but MB didn't provide one, `%composer%` is non-empty (from the file fallback), so the `$set` never fires.

Related community post: https://community.metabrainz.org/t/tagger-script-not-executing-automatically/815447

## Solution

Add two getter functions that provide explicit access to each metadata source:

- `$get_new(name)`: returns the value from new metadata only (MusicBrainz or set by a script), ignoring the file's original tags.
- `$get_original(name)`: returns the value from the file's original tags only, ignoring new metadata from MusicBrainz or scripts.

The user's script becomes:
```
$if($get_new(composer),,$set(composer,$if2(%lyricist%,%writer%,%arranger%)))
```

Implementation:
- Add `get_new()` and `get_original()` methods to both `Metadata` (trivial pass-through to `get()`) and `MultiMetadataProxy` (checks only the writable or readonly metadata respectively).
- Store the readonly metadata reference in `MultiMetadataProxy` for `get_original()` access.
- 12 tests covering both functions with proxy and plain metadata contexts.

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code developed through interactive AI-assisted session with human design decisions, review, and testing.

## Action

Additional actions required:

### DIFF
--- a/picard/metadata.py
+++ b/picard/metadata.py
@@ -280,6 +280,25 @@ class Metadata(MutableMapping[str, str | list[str] | None]):
         with self._lock.lock_for_read():
             return self._store.__contains__(self.normalize_tag(name))
 
+    def get_new(self, name, default=''):
+        """Get a value only from the new metadata.
+
+        On a plain Metadata object this is the same as get().
+        On a MultiMetadataProxy this ignores the readonly fallback.
+        Provides a consistent interface for script functions.
+        """
+        return self.get(name, default)
+
+    def get_original(self, name, default=''):
+        """Get a value from the original (file) metadata.
+
+        On a plain Metadata object this is the same as get() (there is no
+        separate original source).
+        On a MultiMetadataProxy this reads only from the readonly fallback.
+        Provides a consistent interface for script functions.
+        """
+        return self.get(name, default)
+
     def _del(self, name):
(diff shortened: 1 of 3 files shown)


rdswift (approved):
Looks good to me, and should provide a solution for the user's issue.

phw (approved):

PR #3338, opened 2026-08-12
PICARD-3382: Expose debug_if() via plugin API as api.logger.debug_if

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Expose `debug_if()` to plugin authors via `api.logger.debug_if()`, and export the `DebugOpt` enum from the plugin API so plugins can reduce log noise by gating verbose debug output on debug options.

## Problem

* JIRA ticket: [PICARD-3382](https://tickets.metabrainz.org/browse/PICARD-3382)

Plugin authors have no way to use conditional debug logging (`debug_if`) through the plugin API. The only option is unconditional `api.logger.debug()`, which creates noise in the log output even when the user is not interested in plugin-level debug details.

## Solution

- Add a `PluginLogger` wrapper class in `api_impl.py` that wraps the standard `logging.Logger` and adds a `debug_if()` method delegating to `picard.log.debug_if()`.
- Export `DebugOpt` from `picard.plugin3.api` so plugins can import it directly.
- Add a `stacklevel` parameter to `log.debug_if()` (default unchanged at 3) so wrappers can report correct source locations.
- Update `docs/PLUGINSV3/API.md` with usage documentation.
- Add tests in `test/plugins3/test_plugin_logger.py`.

Usage:
```python
from picard.plugin3.api import DebugOpt, PluginApi

def enable(api: PluginApi):
    api.logger.debug_if(DebugOpt.PLUGIN_DEVELOPMENT, "Details: %s", data)

    if dbg := api.logger.debug_if(DebugOpt.PLUGIN_DEVELOPMENT):
        for item in items:
            dbg("item: %r", item)
```

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

Code, tests, and documentation were developed with AI assistance (Kiro CLI), with human review and direction throughout.

## Action

Additional actions required:

### DIFF
--- a/docs/PLUGINSV3/API.md
+++ b/docs/PLUGINSV3/API.md
@@ -77,6 +77,7 @@ The following classes are available through the `api` module:
 - `Track` - Track object
 - `File` - File base class (for custom formats)
 - `Cluster` - Cluster object
+- `DebugOpt` - Debug option enum for conditional logging
 - `Metadata` - Metadata container for tags
 - `BaseAction` - Base class for UI actions
 - `MetadataTagAction` - Base class for metadata tag context menu actions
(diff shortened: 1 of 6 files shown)


rdswift (dismissed):
Looks good, and seems to work as expected.  Thanks for this.

phw (changes requested):
Generally yes. But now that `api.logger` exposes the type `PluginLogger`, which itself only defines `debug_if` explicitly, editors can't provide meaningful code completion or type checking.

Can we inherit from `logging.Logger` instead? Or explicitly define the relevant logging functions?

phw (approved):
LGTM

zas:
> Generally yes. But now that `api.logger` exposes the type `PluginLogger`, which itself only defines `debug_if` explicitly, editors can't provide meaningful code completion or type checking.
> 
> Can we inherit from `logging.Logger` instead? Or explicitly define the relevant logging functions?

`logging.Logger` isn't designed to be subclassed directly, its `__init__` wires the instance into the global logger hierarchy by name, and there's no clean way to wrap an existing logger instance via inheritance. This is something I wanted to do initially, but I gave up.

So I explicitly defined the relevant logging functions (again, because that's actually what I did first, but then removed as it wasn't strictly needed. I didn't think about editors/completions.). See https://github.com/metabrainz/picard/pull/3338/commits/75284362449a50ac4b74a6d11207b916ac14426d

PR #2377, opened 2024-03-24
PICARD-187: Support for manually removing cover art

ShubhamBhut:
PICARD-187: Support for manually removing cover art

finishing PICARD-187

# Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

# Problem

* JIRA ticket (_optional_): PICARD-187

# Solution

Add support in all formats to allow to delete cover art tags, and modify ImageList to allow to remove desired image from ImageList

# Action

PR #3340, opened 2026-08-12
PICARD-3380: Add "Remove cover images from tags" option

cryobry:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

**Describe this change in 1-2 sentences**:

This commit adds a new option to the Picard UI to remove cover images from tags for users that only wish to use external cover art.

The option is guarded by "Save cover images as separate files" so tagged artwork is saved to an external file before being removed from the tags.

## Problem

* JIRA ticket (_optional_): PICARD-3380

There is currently not a way to automatically remove cover art from tags for users that wish to exclusively use external cover images.

## Solution

Add a new option to the Picard UI to "Remove cover images from tags" when being saved to a separate file (so erroneous removal is potentially less destructive).

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

* UI insertion
* Tagging syntax
* Tests

## Action

Additional actions required:

Not tested across all tag formats, only id3 and vorbis.

### DIFF
--- a/picard/coverart/__init__.py
+++ b/picard/coverart/__init__.py
@@ -84,6 +84,7 @@ class CoverArt:
         'save_images_to_tags',
         'save_images_to_files',
         'save_images_overwrite',
+        'remove_images_from_tags',
         'cover_image_filename',
         'image_type_as_filename',
         'filter_cover_by_size',
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -203,8 +203,8 @@ class APEv2File(File):
             tags = mutagen.apev2.APEv2(filename)
         except mutagen.apev2.APENoHeaderError:
             tags = mutagen.apev2.APEv2()
-        images_to_save = list(metadata.images.to_be_saved_to_tags())
-        remove_images = metadata.images.should_remove_images_from_tags()
+        images_to_save = list(metadata.images.to_be_saved_to_tags(previous_images=self.orig_metadata.images))
+        remove_images = metadata.images.should_remove_images_from_tags(previous_images=self.orig_metadata.images)
         if config.setting['clear_existing_tags']:
             preserved = []
             if not remove_images and config.setting['preserve_images']:
--- a/picard/formats/apev2.py
+++ b/picard/formats/apev2.py
@@ -204,14 +204,15 @@ class APEv2File(File):
         except mutagen.apev2.APENoHeaderError:
             tags = mutagen.apev2.APEv2()
         images_to_save = list(metadata.images.to_be_saved_to_tags())
+        remove_images = metadata.images.should_remove_images_from_tags()
         if config.setting['clear_existing_tags']:
             preserved = []
-            if config.setting['preserve_images']:
+            if not remove_images and config.setting['preserve_images']:
                 preserved = list(self._iter_cover_art_tags(tags))
             tags.clear()
             for name, value in preserved:
                 tags[name] = value
-        elif images_to_save:
+        elif images_to_save or remove_images:
             for name, _value in self._iter_cover_art_tags(tags):
                 del tags[name]
         temp = {}
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -763,9 +763,9 @@ class ID3File(File):
 
     def _save_images_to_tags(self, tags, metadata):
         """Save cover art images to tags."""
-        images_to_save = list(metadata.images.to_be_saved_to_tags())
+        images_to_save = list(metadata.images.to_be_saved_to_tags(previous_images=self.orig_metadata.images))
         if not images_to_save:
-            if metadata.images.should_remove_images_from_tags():
+            if metadata.images.should_remove_images_from_tags(previous_images=self.orig_metadata.images):
                 tags.delall('APIC')
             return
 
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -765,6 +765,8 @@ class ID3File(File):
         """Save cover art images to tags."""
         images_to_save = list(metadata.images.to_be_saved_to_tags())
         if not images_to_save:
+            if metadata.images.should_remove_images_from_tags():
+                tags.delall('APIC')
             return
 
         tags.delall('APIC')
--- a/picard/ui/coverartbox/__init__.py
+++ b/picard/ui/coverartbox/__init__.py
@@ -52,6 +52,7 @@ from picard import (
     log,
     tagger_instance,
 )
+from picard.cluster import Cluster
 from picard.config import get_config
 from picard.coverart.image import (
     CoverArtImage,
@@ -276,8 +277,13 @@ class CoverArtBox(QtWidgets.QGroupBox):
 
         # Predict whether saving will strip cover art already embedded in tags,
         # using the same rules the format save code applies (see ImageList).
+        # Only show the removal indicator for items in the right pane (matched
+        # to an album/track). Files still in a cluster (left pane) haven't been
+        # matched yet, so the warning would be premature.
+        is_unmatched = isinstance(getattr(self.item, 'parent_item', None), Cluster)
         self._removal_predicted = bool(
-            metadata
+            not is_unmatched
+            and metadata
             and orig_metadata
             and orig_metadata.images
             and metadata.images.should_remove_images_from_tags(previous_images=orig_metadata.images)
--- a/picard/ui/coverartbox/__init__.py
+++ b/picard/ui/coverartbox/__init__.py
@@ -69,7 +69,10 @@ from picard.util import (
 )
 from picard.util.lrucache import LRUCache
 
-from .coverartthumbnail import CoverArtThumbnail
+from .coverartthumbnail import (
+    THUMBNAIL_WIDTH,
+    CoverArtThumbnail,
+)
 from .imageurldialog import ImageURLDialog
 
 from picard.ui.util import FileDialog
@@ -109,6 +112,15 @@ class CoverArtBox(QtWidgets.QGroupBox):
             QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter
         )
         self.orig_cover_art_info_label.setWordWrap(True)
+        self.cover_removal_warning_label = QtWidgets.QLabel('')
+        self.cover_removal_warning_label.setAlignment(
+            QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter
+        )
+        self.cover_removal_warning_label.setWordWrap(True)
+        # Keep wrapping within the thumbnail column instead of widening the box
+        self.cover_removal_warning_label.setMaximumWidth(THUMBNAIL_WIDTH)
+        self.cover_removal_warning_label.setStyleSheet('QLabel { color: #a00000; font-weight: bold; }')
+        self.cover_removal_warning_label.setHidden(True)
         self.show_details_button = QtWidgets.QPushButton(_('Show more details'), self)
         self.show_details_shortcut = QtGui.QShortcut(
             QtGui.QKeySequence(_("Ctrl+Shift+I")), self, self.show_cover_art_info
@@ -119,6 +131,7 @@ class CoverArtBox(QtWidgets.QGroupBox):
         self.layout.addWidget(self.orig_cover_art_label)
         self.layout.addWidget(self.orig_cover_art)
         self.layout.addWidget(self.orig_cover_art_info_label)
+        self.layout.addWidget(self.cover_removal_warning_label)
         self.layout.addWidget(self.show_details_button)
         self.layout.addSpacerItem(spacerItem)
         self.setLayout(self.layout)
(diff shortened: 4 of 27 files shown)


picard/formats/id3.py
```diff
@@ -765,6 +765,8 @@ def _save_images_to_tags(self, tags, metadata):
         """Save cover art images to tags."""
         images_to_save = list(metadata.images.to_be_saved_to_tags())
         if not images_to_save:
+            if get_config().setting['remove_images_from_tags']:
```
zas: The UI "guard" is purely visual: the checkbox is a child of the `save_images_to_files` groupbox, so it's only clickable when that groupbox is checked, but the format handlers read the setting directly without verifying `save_images_to_files`.

If a user enables the option, then unchecks "Save as separate files", the persisted `remove_images_from_tags=True` will still strip images on next save,  silently destroying cover art with no backup.

That's a serious flaw.

Check `save_images_to_files` alongside `remove_images_from_tags` in each format handler (or ensure it is False when save_images_to_files is False, or add a property should_remove_cover_art to File, dunno what's best).

ui/options_cover.ui
```diff
@@ -134,6 +134,13 @@
         </property>
        </widget>
       </item>
+      <item>
+       <widget class="QCheckBox" name="remove_images_from_tags">
+        <property name="text">
+         <string>Remove cover images from tags</string>
```
zas, line 143: Missing tooltip on the new checkbox

test/formats/coverart.py
```diff
@@ -148,6 +148,23 @@ def test_cover_art_clear_tags_preserve_images_no_existing_images(self):
...
+
+        @skipUnlessTestfile
+        def test_cover_art_remove_images_from_tags(self):
+            image = CoverArtImage(data=self.pngdata, types=['front'])
+            file_save_image(self.format_registry, self.filename, image)
+            config.setting['save_images_to_tags'] = False
+            config.setting['remove_images_from_tags'] = True
+            metadata = save_and_load_metadata(self.format_registry, self.filename, Metadata())
+            self.assertEqual(0, len(metadata.images))
+
```
zas, line 185: Tests don't cover the dangerous scenario where `save_images_to_tags = False` and `remove_images_from_tags = True`, this is a must-have.

picard/coverart/__init__.py
```diff
@@ -84,6 +84,7 @@ def __repr__(self):
         'save_images_to_tags',
         'save_images_to_files',
         'save_images_overwrite',
+        'remove_images_from_tags',
```
zas, line 87: Since that's the inverse of `save_images_to_tags` perhaps keep them together.

zas (changes requested):
Thanks for contributing. Such feature was asked in the past, I thought we had a ticket for it somewhere, but I didn't find it.

To me this PR has a major flaw: UI "guard" is insufficient for such destructive operation.
Also `resouces.py` is just noise (it shouldn't be committed).

There is no visual feedback anywhere in the UI that images will be stripped from tags on save. The user enables the option in settings and then saves, the removal happens silently.
  
- At minimum, the "Existing Cover" column in the info dialog (picard/ui/infodialog/dialog.py) could show a strikethrough, overlay icon, or text indicating removal when `remove_images_from_tags=True`.
- Alternatively or additionally, the cover art box in the main window could show a status label or icon indicating that embedded images will be removed on save.

Perhaps a confirmation too (when enabling the option, clearly state what it does, ask the user if sure).

picard/formats/apev2.py
```diff
@@ -204,14 +204,15 @@ def _save(self, filename, metadata):
         except mutagen.apev2.APENoHeaderError:
             tags = mutagen.apev2.APEv2()
         images_to_save = list(metadata.images.to_be_saved_to_tags())
+        remove_images = config.setting['remove_images_from_tags']
         if config.setting['clear_existing_tags']:
             preserved = []
-            if config.setting['preserve_images']:
+            if config.setting['preserve_images'] and not remove_images:
```
zas: In general, test the variable first, and the config value after, slightly faster.

phw:
> Thanks for contributing. Such feature was asked in the past, I thought we had a ticket for it somewhere, but I didn't find it.

We do have (PICARD-187), and also an open PR #2377 . The last approach would go further, allowing the user to choose images to remove. Which in theory would be great, but also brings some complications. The open PR still has some issues with cover art display. I recently rebased the branch in https://github.com/phw/picard/commits/PICARD-187-delete-coverart/ and tried to address some issues in https://github.com/phw/picard/commit/7e9a1b2b3db9a1e3a600e596b2a5917ccfa073d0 , but there was still some issues with the cover art box.

I still think we should address the open PR. But the PR here provides a simpler approach, and I'm no opposed to merging this for now, because it avoids a lot of the issues the other PR raised.

picard/formats/id3.py
cryobry: Added a reusable `should_remove_images_from_tags()` method in imagelist.py that checks both.

phw: This is actually very nice. Because it allows us later to combine this with individual image removal (but please don't start anything like this in this PR, I rather get something mergable now ;-) ).

One of the issues to solve in #3340 was, that removing one image out of several is no problem, as the images get rewritten, automatically getting rid of the removed ones. But if you remove all of them (empty ImageList) it needs to indicate somehow that images need to be removed or left alone (that's analog to `MetaData`, which can indicate tag removal vs. tag just not being overwritten). By having `should_remove_images_from_tags` and the implementation in formats already in place it makes it easier to later have `ImageList` integrate with individual image removal.

cryobry:
Latest push made the Existing Cover (to be removed) warning overlay slightly pixelated w/ QtCore.Qt.BrushStyle.Dense4Pattern to differentiate it from a (common) cover art tint difference.

rdswift:
Please don't force push the commits each time.  It makes reviewing much more difficult because you can't see what has been changed since your last review.

ui/options_cover.ui
zas: Please don't close conversation, especially in this case, this wasn't implemented.
Add a tooltip explaining what this checkbox does in details (especially since that's destructive)

picard/ui/coverartbox/__init__.py
```diff
@@ -109,6 +112,15 @@ def __init__(self, parent=None):
...
         )
         self.orig_cover_art_info_label.setWordWrap(True)
+        self.cover_removal_warning_label = QtWidgets.QLabel('')
+        self.cover_removal_warning_label.setAlignment(
+            QtCore.Qt.AlignmentFlag.AlignTop | QtCore.Qt.AlignmentFlag.AlignHCenter
+        )
+        self.cover_removal_warning_label.setWordWrap(True)
+        # Keep wrapping within the thumbnail column instead of widening the box
+        self.cover_removal_warning_label.setMaximumWidth(THUMBNAIL_WIDTH)
+        self.cover_removal_warning_label.setStyleSheet('QLabel { color: #a00000; font-weight: bold; }')
```
zas: Don't use hardcoded color, use `interface_colors.get_color()`, select one that could work or add a new interface color if strictly needed, this supports both dark/light theme. 

See https://github.com/metabrainz/picard/blob/a6b1558775063f629b0f09f68e87aab2b3926670/picard/ui/dialogs/installplugin.py#L280 for an example.

picard/ui/coverartbox/__init__.py
```diff
@@ -182,10 +195,18 @@ def update_display(self, force=False):
         self.cover_art.setToolTip("<br/>".join(tooltip_cover_lines))
         self.orig_cover_art.setToolTip("<br/>".join(tooltip_orig_lines))
 
+        removal_warning_label = getattr(self, 'cover_removal_warning_label', None)
```
zas: The label is created unconditionally in `__init__()`  so this guard can never trigger. Just use `self.cover_removal_warning_label`.

test/test_ui_options_cover.py
```diff
@@ -0,0 +1,61 @@
...
+
+@pytest.fixture(autouse=True)
+def _patch_tagger_instance(monkeypatch):
+    import picard.ui.options as options_module
+
+    monkeypatch.setattr(options_module, 'tagger_instance', lambda: MagicMock())
+
+
+def test_save_forces_remove_images_from_tags_false_when_save_to_files_disabled():
+    page = CoverOptionsPage()
```
zas, line 32: Not strictly needed but better be explicit about `qapp` here since it creates a widget

```python
@pytest.fixture()
def cover_options_page(qapp):
    page = CoverOptionsPage()
    page.load()
    return page

def test_save_forces_remove_images_from_tags_false_when_save_to_files_disabled(cover_options_page):
    cover_options_page.ui.save_images_to_files.setChecked(False)
    cover_options_page.ui.remove_images_from_tags.setChecked(True)
    cover_options_page.save()
    ...
```

It works only because `conftest.py` defines `qapp` as `autouse=True` at session scope, but making it explicit help to document code.

test/test_ui_options_cover.py
```diff
@@ -0,0 +1,61 @@
...
+
+
+@pytest.fixture(autouse=True)
+def _patch_tagger_instance(monkeypatch):
+    import picard.ui.options as options_module
+
+    monkeypatch.setattr(options_module, 'tagger_instance', lambda: MagicMock())
+
+
+def test_save_forces_remove_images_from_tags_false_when_save_to_files_disabled():
```
zas: A bit too long for a function name.

test/test_ui_options_cover.py
```diff
@@ -0,0 +1,61 @@
...
+import pytest
+
+from picard.ui.options.cover import CoverOptionsPage
+
+
+@pytest.fixture(autouse=True)
+def _patch_tagger_instance(monkeypatch):
+    import picard.ui.options as options_module
+
+    monkeypatch.setattr(options_module, 'tagger_instance', lambda: MagicMock())
```
zas: You can use `MockTagger`, something like:

```python
from test.picardtestcase import MockTagger
  
@pytest.fixture(autouse=True)
def _patch_tagger_instance(monkeypatch):
    import picard.ui.options as options_module

    tagger = MockTagger()
    monkeypatch.setattr(options_module, 'tagger_instance', lambda: tagger)
```

But I think I'll implement it as a fixture, this pattern comes back quite often.

zas (changes requested):

picard/ui/coverartbox/__init__.py
cryobry: I'm going to remove this text, it doesn't look good and I've landed on a better-looking design, but it's proving tricky to implement due to how cover art is filtered.

cryobry:
From my testing this feature is now working as intended, with the proper guarding (also disabled when _Embed cover images into tags_ is enabled), even on settings mismatches (e.g. restores). Cover art to be removed is designated with a red overlay and red strikethrough text in both the CoverArtBox and AlbumInfoDialog, which works well visually. I had to make some wider changes to preserve the cover art state for tagged images set for removal to be available for saving as external cover art. Also needed to cache removed cover art so we can continue to display it after it's removed from the tags (without affecting metadata state). With the scope expansion I'm sure there's plenty to cleanup but this is a better starting spot having now dug into it.

test/test_ui_options_cover.py
phw: @cryobry  zas' #3344 was merged, which provides a global `patch_tagger_instance` fixture you could use.

phw:
Thanks a lot. I'll give this a full testing and review later today. Overall this is looking solid

test/test_ui_options_cover.py
cryobry: Done!

picard/ui/util.py
```diff
@@ -54,6 +54,21 @@ def find_starting_directory():
     return find_existing_path(path)
 
 
+def apply_removal_overlay(pixmap):
+    """Draw a dotted translucent red overlay signaling the image will be removed from tags."""
+    overlaid = QtGui.QPixmap(pixmap)
+    painter = QtGui.QPainter(overlaid)
+    brush = QtGui.QBrush(QtGui.QColor(200, 0, 0, 200), QtCore.Qt.BrushStyle.Dense4Pattern)
```
zas: This still uses hard-coded color.

picard/ui/util.py
```diff
@@ -54,6 +54,21 @@ def find_starting_directory():
...
+    painter.fillRect(overlaid.rect(), brush)
+    painter.end()
+    return overlaid
+
+
+def strikethrough_removal_text(text):
+    """Style rich text as struck-through dark red, signaling removal from tags."""
+    return f'<span style="color:#a00000; text-decoration:line-through;">{text}</span>'
```
zas: This still uses hard-coded color.

picard/util/imagelist.py
```diff
@@ -86,42 +86,102 @@ def get_front_image(self) -> 'CoverArtImage | None':
...
+        Only true when explicitly requested and this list actually has at
+        least one image that will be saved to an external file, so cover art
+        is never deleted without being kept somewhere else.
+        """
+        if settings is None:
+            config = get_config()
+            settings = config.setting
+        if not settings['remove_images_from_tags']:
+            return False
+        return any(self.to_be_saved_to_files(settings, previous_images))
```
zas, line 186: Perhaps evaluate the performance of this (not a blocker)

zas (changes requested):
The recent changes look very good, just a few more minor comments

zas:
@cryobry see https://github.com/cryobry/picard/pull/1

zas (dismissed):
LGTM

rdswift (dismissed):
I did some quick testing and it seemed to work as advertised.  The UI changes look fine.  Thanks.

phw (changes requested):
Thanks again. Code looks good and it worked in my testing as expected. I found only one minor quirk:

If you all of "save_images_to_tags", "save_images_to_files" and "remove_images_from_tags" enabled and have a release loaded (with old and new cover), then reload that release, the original cover images shows the red overlay during loading. Once the loading finished, the overlay is gone (because "save_images_to_tags" is enabled).

In order to get into this situation, you first need to enable both  "save_images_to_files" and "remove_images_from_tags", and afterwards enable "save_images_to_tags".

It probably should never show the overlay, if "save_images_to_tags" is enabled.

phw (approved):
Thanks for the final fix. Works well for me.

Great to have finally support for removing images. LGTM.

zas (approved):
Good job, thanks for your contribution!

LGTM

PR #3341, opened 2026-08-12
resources/compile: Modernize script and use atomic write pattern

zas:
## Summary

* This is a...
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other

Modernize `resources/compile.py` by removing deprecated distutils usage, switching to pathlib and subprocess, and using an atomic write pattern so `resources.py` is never left in a partial state.

## Problem

* `resources/compile.py` relied on `distutils` (deprecated, removed from stdlib in Python 3.12) and used fragile file writing that could leave `resources.py` corrupted if the build was interrupted.

## Solution

- Replace `distutils.log` and `distutils.spawn` with `logging` and `subprocess`
- Replace `setuptools.modified.newer` with a simple `os.path.getmtime` comparison
- Use `pathlib.Path` for path handling
- Extract `find_rcc()` function for clearer rcc binary search (covers RHEL lib64, Debian lib, and system PATH)
- Check for rcc availability early, before checking timestamps
- Log which rcc binary is being used
- Have rcc write to a temp file, fix the PySide6 import there, then atomically rename to the target
- Remove the unnecessary `coding: utf-8` declaration

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used as a coding assistant for suggestions and reviewing the implementation.

## Action

Additional actions required:

### DIFF
--- a/resources/compile.py
+++ b/resources/compile.py
@@ -1,10 +1,9 @@
 #!/usr/bin/env python
-# -*- coding: utf-8 -*-
 #
 # Picard, the next-generation MusicBrainz tagger
 #
 # Copyright (C) 2006 Lukáš Lalinský
-# Copyright (C) 2013-2014, 2018, 2020 Laurent Monin
+# Copyright (C) 2013-2014, 2018, 2020, 2026 Laurent Monin
 # Copyright (C) 2014 Shadab Zafar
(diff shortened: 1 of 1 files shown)


phw (approved):
LGTM

PR #3347, opened 2026-08-15
PICARD-3385: add tags for called folk dances

SamWhited:
## Summary

[1]: https://datatracker.ietf.org/doc/draft-swhited-contra-tags/

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

This patch adds tags for traditional called folk dances from the IETF Internet-Draft (I-D) [draft-swhited-contra-tags-04][1]. It also supports mapping these tags to Vorbis Comment namespaces as recommended in the document when saving a file that uses Vorbis Comments.

* JIRA ticket (_optional_): PICARD-3385

## Solution

This is a relatively simple patch adding the various tags from the document linked above to the tags list found at `picard/const/tags.py`. In addition it adds translations to the format the document recommends for Vorbis Comments to `picard/formats/vorbis.py`. That said, it may be worth removing this second part and updating the document to not make this distinction as I don't see that it provides any value (I simply did this originally because I was told this is how Vorbis tags should be namespaced, but most other tags in more common use don't seem to do this, so we could likely change this and drop this part of the commit, I would be happy to get opinions here).

## AI Usage

* [x] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

I don't believe this requires specific updates to the documentation, but if there is a page that exhaustively lists tags like this somewhere that I'm missing I am happy to submit a patch there as well. Thanks!

### DIFF
--- a/picard/const/tags.py
+++ b/picard/const/tags.py
@@ -335,6 +335,164 @@ ALL_TAGS = TagVars(
         is_from_mb=False,
     ),
     TagVar(
+        'dance_caller',
+        shortdesc=N_('Caller'),
+        longdesc=N_('The name of the dance caller(s) heard in the track.'),
+        doc_links=(
+            DocumentLink(
+                N_('I-D: Metadata for Called Folk Dances'),
+                'https://datatracker.ietf.org/doc/draft-swhited-contra-tags/',
+            ),
+        ),
+    ),
+    TagVar(
+        'dance_choreographer',
+        shortdesc=N_('Dance Choreographer'),
+        longdesc=N_('The name of the author(s) of the danc(es) being called in the track for called folk dances.'),
+        doc_links=(
+            DocumentLink(
+                N_('I-D: Metadata for Called Folk Dances'),
+                'https://datatracker.ietf.org/doc/draft-swhited-contra-tags/',
+            ),
+        ),
+        is_multi_value=True,
+    ),
+    TagVar(
+        'dance_choreography',
+        shortdesc=N_('Dance Choreography'),
+        longdesc=N_(
+            'The moves of the dance as text for called folk dances. This should not include the dance title, author, or other information covered by other tags.'
+        ),
+        doc_links=(
+            DocumentLink(
+                N_('I-D: Metadata for Called Folk Dances'),
+                'https://datatracker.ietf.org/doc/draft-swhited-contra-tags/',
+            ),
+        ),
+    ),
+    TagVar(
+        'dance_crooked',
+        shortdesc=N_('Crooked'),
+        longdesc=N_('Whether a traditional called folk dance tune is "crooked" (ie. not in perfect dance form).'),
+        doc_links=(
+            DocumentLink(
+                N_('I-D: Metadata for Called Folk Dances'),
+                'https://datatracker.ietf.org/doc/draft-swhited-contra-tags/',
+            ),
+        ),
+    ),
+    TagVar(
+        'dance_form',
+        shortdesc=N_('Dance Form'),
+        longdesc=N_(
+            'The form of a called dance with no particular format, eg. "contra" or "square dance" or "duple minor improper contra".'
+        ),
+        doc_links=(
+            DocumentLink(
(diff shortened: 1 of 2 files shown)


rdswift:
Looking at this, I wonder if it would be more appropriate as a plugin.  It seems pretty "niche" to include in the base code.

zas:
I tend to agree this would better fit a plugin, because that's very niche, but also because it is based on a recent draft.

SamWhited:
That makes sense to me (though in a selfish way it makes me sad, I was hoping to just have auto complete and not have to install more stuff, but given how few people would likely use these tags I get it), but I also don't see a way to do that from glancing at the plugin API, there doesn't appear to be anything about the metadata autocomplete list: https://github.com/metabrainz/picard/blob/master/docs/PLUGINSV3/API.md

rdswift:
I've taken the liberty of creating a plugin for this (for Picard v3).  It is available at https://github.com/rdswift/picard-plugin-called-folk-dance-tags so please have a look and let me know if that is okay.

SamWhited:
Oh wow, I was still looking through the plugin docs in the time it took you to make that, that's amazing, thank you!

SamWhited:
I gave it a try, this appears to work in that if I add eg. a "dance_caller" tag it shows up with the nice localized name "Caller" in Picard, but it doesn't add the tag to the auto-complete list like the change to Picard does. Eg. if I just add a new tag and type "dance_" I don't get the nice list that I can select from. I'm not sure if this is a bug with the plugins API, a limitation of how it works, or some other issue.

rdswift:
> Oh wow, I was still looking through the plugin docs in the time it took you to make that, that's amazing, thank you!

Don't thank me yet.  It might not do exactly what you want.  :stuck_out_tongue_winking_eye: 

Besides, I've done at least 16 plugins for Picard 3 so far, so I'm pretty familiar with the process.  If you're okay with this, I'll add it to the public registry, and set it up for translations to other languages on Weblate.  I'll also add it to my plugin documentation information on ReadTheDocs at https://picard-plugins-user-guides.readthedocs.io/en/latest/index.html

rdswift:
> I gave it a try, this appears to work in that if I add eg. a "dance_caller" tag it shows up with the nice localized name "Caller" in Picard, but it doesn't add the tag to the auto-complete list like the change to Picard does. Eg. if I just add a new tag and type "dance_" I don't get the nice list that I can select from. I'm not sure if this is a bug with the plugins API, a limitation of how it works, or some other issue.

If I understand correctly, this is when you're adding the tag in the metadata box, right?  In that case, it's likely something within Picard.  Would you mind entering a ticket in the [MusicBrainz issue tracker](https://tickets.metabrainz.org/browse/PICARD) so we remember to look into it?

SamWhited:
That's correct. I'll submit a new ticket. And thank you for your implementation, you're of course welcome to publish it however you'd like as the author 

**EDIT:** Done: https://tickets.metabrainz.org/browse/PICARD-3387

rdswift:
> you're of course welcome to publish it however you'd like as the author

Well, all I did was rearrange your original work.  From my perspective, this makes the plugin at least as much yours as mine.  I'm okay publishing it, but am also fine if you want to take over the repo and publish it yourself instead.  That way it would be easier for you to update and make changes as required.  Just let me know.

SamWhited:
@rdswift up to you, feel free to add me as a collaborator and if there are changes made to the document I'm happy to update it as well, and thank you again!

rdswift:
Okay, I had already included you in the copyright note in the code and in the README file.  I'll look into publishing this in the public plugin registry when I get a chance.  Thanks.

phw:
As I already wrote on the ticket, I agree this should be a plugin (and thanks to rdswift already is). That plugin registered variables should be available for autocomplete definitely was the intention and that it doesn't work was an oversight.

Let's close this PR. And if we need some more functionality for this then rather let's think about how to improve the plugin API to cover the case.

PR #3349, opened 2026-08-16
Replace hardcoded colors with theme-aware alternatives

zas:
## Summary
* This is a…
  * Bug fix
  * Feature addition
  * Refactoring
  * Minor / simple change (like a typo)
  * Other

Replace hardcoded color values with theme-aware alternatives: validation
indicators adapt to dark/light without user configuration, `row_highlight`
is removed in favor of the system palette, and `match_similarity_low` is
registered as a configurable color.

## Problem

Several UI components use hardcoded color values (`#f55`, `#292`,
`(223, 125, 125)`) that don't adapt to the current theme. Additionally,
the `row_highlight` configurable color is overly specific (used in one
place) and hard to justify as a user-facing option.

## Solution

Three commits:

1. **Replace validation stylesheets with theme-aware functions** —
   `STYLESHEET_SUCCESS` / `STYLESHEET_ERROR` class constants become
   module-level functions `stylesheet_validation_success()` and
   `stylesheet_validation_error()` that pick dark/light color variants
   at call time. These are *not* added to interface_colors since they
   are standard UX patterns (green/red) that don't need user
   configuration.

2. **Remove `row_highlight` from interface_colors** — Replace with a
   semi-transparent variation of `QPalette.ColorRole.Highlight`, which
   automatically adapts to any theme or accent color.

3. **Make `match_similarity_low` configurable** — Register the
   low-similarity endpoint of the match quality gradient (`#DF7D7D`) in
   interface_colors. Match quality colors are functional (users
   distinguish match status by color), so this is appropriate.

## AI Usage

* No AI/LLM use
* Minimal use (e.g. autocompletion)
* Moderate use (e.g. suggestions regarding code fragments)
* Significant use (e.g. code structure, tests development, etc.)
* Primarily AI developed
* Other (please specify below)

AI was used to identify hardcoded colors, evaluate which should be
configurable vs theme-derived, and implement the changes.

## Action

Additional actions required:
* Update Picard [documentation](https://github.com/metabrainz/picard-docs) (please include a reference to this PR)
* Other (please specify below)

### DIFF
--- a/picard/ui/options/__init__.py
+++ b/picard/ui/options/__init__.py
@@ -53,6 +53,31 @@ from picard.profile import (
 )
 from picard.util.display_title_base import HasDisplayTitle
 
+from picard.ui.theme import theme
+
+
+# Theme-aware validation indicator colors.
+# These are not user-configurable; they are standard UX patterns
+# (green = success, red = error) that adapt to dark/light themes.
+_VALIDATION_COLORS = {
+    'light': {'success_bg': '#229922', 'error_bg': '#ff5555'},
+    'dark': {'success_bg': '#2d8a2d', 'error_bg': '#cc4444'},
+}
+
+
+def _validation_colors():
+    return _VALIDATION_COLORS['dark'] if theme.is_dark_theme else _VALIDATION_COLORS['light']
+
+
+def stylesheet_validation_success():
+    bg = _validation_colors()['success_bg']
+    return f"QWidget {{ background-color: {bg}; color: white; padding: 2px; }}"
+
+
+def stylesheet_validation_error():
+    bg = _validation_colors()['error_bg']
+    return f"QWidget {{ background-color: {bg}; color: white; font-weight: bold; padding: 2px; }}"
+
 
 class OptionsCheckError(Exception):
     def __init__(self, title, info):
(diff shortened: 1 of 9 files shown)


phw (commented):
I'm not too sure about this. We are moving into a strange direction with arbitrary parts of the UI becoming configurable colors. We originally started with customizable syntax highlighting, and now end up offering color selection for some, from a user perspective, rather arbitrary and obscure interface elements (while not allowing to change other parts).

I think we should limit color configuration to the parts where it is functional: syntax highlighting, colors for match quality, log levels. I don't think we should have some random interface elements have color controlled here.

I even think the existing "row_highlight" color was a mistake to add and should be removed again. Hard to explain what this actually is. This color is very specific and used exactly for one special case currently (highlighting the row of the current release in "other versions" result dialog). Instead it probably should use a variation of the default `ColorRole.Highlight`.

picard/ui/options/__init__.py
```diff
@@ -53,6 +53,31 @@
...
+
+
+def stylesheet_validation_success():
+    bg = _validation_colors()['success_bg']
+    return f"QWidget {{ background-color: {bg}; color: white; padding: 2px; }}"
+
+
+def stylesheet_validation_error():
+    bg = _validation_colors()['error_bg']
+    return f"QWidget {{ background-color: {bg}; color: white; font-weight: bold; padding: 2px; }}"
```
phw: Maybe define those color functions, even if not changeable, in colors.py regardless? It would fit, and if it needs to be used in some dialog one does not need to import the options dialog for the color support.

phw (dismissed):
Maybe moving the error / success color helper functions from the option pages. Not sure, no really strong opinion about this one. Otherwise LGTM.

picard/ui/options/__init__.py
zas: Done in https://github.com/metabrainz/picard/pull/3349/commits/0ff0d88594b4bd5dd9e8a2034f01a9dc52cdcf38

phw (approved):
LGTM

PR #3350, opened 2026-08-16
PICARD-3388: Add interface colors API to Plugin v3

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Add interface colors API to Plugin v3, allowing plugins to register
their own configurable colors and read Picard's core interface colors.

## Problem

* JIRA ticket: PICARD-3388

Plugins that create custom UI (options pages, dialogs) have no way to:
- Register their own user-configurable colors
- Read Picard's core interface colors for consistent styling
- Know whether the current theme is dark or light

This forces plugins to hardcode colors that may not adapt to the user's
theme.

## Solution

Add new methods to `PluginApi`:

- `register_color(name, title, light_value, dark_value)` — register a
  plugin-specific color that automatically appears in the Interface
  Colors options page, grouped under the plugin's name
- `get_plugin_color(name)` / `get_plugin_qcolor(name)` — read own colors
- `get_color(name)` / `get_qcolor(name)` — read core interface colors
- `is_dark_theme` — property for theme awareness
- `unregister_color(name)` / `unregister_all_colors()` — manual cleanup
  (automatic on plugin disable)

Plugin colors are internally namespaced as `plugin:<uuid>:<name>` to
prevent collisions with core colors or other plugins. They are persisted
in the same config keys as core colors, so user customizations survive
plugin disable/re-enable cycles.

Infrastructure changes:
- `picard/ui/colors.py`: added `unregister_color()` and updated
  `register_color()` to accept an optional `description` parameter
- `picard/plugin3/plugin.py`: cleanup on disable

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used for API design discussion, implementation, test development,
and documentation writing.

## Action

Additional actions required:

### DIFF
--- a/picard/plugin3/api_impl.py
+++ b/picard/plugin3/api_impl.py
@@ -39,6 +39,7 @@ from typing import (
 )
 
 from PyQt6.QtCore import QLocale
+from PyQt6.QtGui import QColor
 
 from picard import log
 from picard.album import Album
@@ -1721,8 +1722,12 @@ class PluginApi:
         """
         if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', name):
             raise ValueError(f"Invalid color name '{name}': must be alphanumeric with underscores")
+        if light_value is not None and not QColor(light_value).isValid():
+            raise ValueError(f"register_color('{name}'): invalid light_value '{light_value}'")
+        if dark_value is not None and not QColor(dark_value).isValid():
+            raise ValueError(f"register_color('{name}'): invalid dark_value '{dark_value}'")
         if light_value is None and dark_value is None:
-            raise ValueError(f"register_color('{name}'): at least one of light_value or dark_value must be provided")
+            raise ValueError(f"register_color('{name}'): no valid color values provided")
         if light_value is None:
             light_value = dark_value
         if dark_value is None:
--- a/picard/plugin3/api_impl.py
+++ b/picard/plugin3/api_impl.py
@@ -118,6 +118,12 @@ from picard.webservice import (
 )
 from picard.webservice.api_helpers import MBAPIHelper
 
+from picard.ui.colors import (
+    ColorDescription,
+    interface_colors,
+    register_color as _register_color,
+    unregister_color as _unregister_color,
+)
 from picard.ui.options import OptionsPage as _OptionsPage
 
 
@@ -295,6 +301,7 @@ class PluginApi:
         self._plugin_dir: Path = plugin_dir
         self._qt_translator: PluginTranslator | None = None
         self._mb_api: MBAPIHelper | None = None
+        self._registered_colors: list[str] = []
 
     @staticmethod
     def _get_caller_info(frame_depth=2):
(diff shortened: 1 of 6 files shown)


picard/plugin3/api_impl.py
```diff
@@ -1658,6 +1665,153 @@ def enable(api):
...
+        """
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', name):
+            raise ValueError(f"Invalid color name '{name}': must be alphanumeric with underscores")
+        if light_value is None and dark_value is None:
+            raise ValueError(f"register_color('{name}'): at least one of light_value or dark_value must be provided")
+        if light_value is None:
+            light_value = dark_value
+        if dark_value is None:
+            dark_value = light_value
+
```
rdswift: Is it worth checking that the color values are not both `None` before saving?  Also that the values are valid, perhaps with some sort of regex like `^(\#[a-fA-F0-9]{6}|[a-zA-Z]{3,})$`.  This could avoid problems later if trying to use an invalid value.

picard/plugin3/api_impl.py
```diff
@@ -1658,6 +1665,153 @@ def enable(api):
...
+                )
+        """
+        if not re.match(r'^[a-zA-Z_][a-zA-Z0-9_]*$', name):
+            raise ValueError(f"Invalid color name '{name}': must be alphanumeric with underscores")
+        if light_value is None and dark_value is None:
+            raise ValueError(f"register_color('{name}'): at least one of light_value or dark_value must be provided")
+        if light_value is None:
+            light_value = dark_value
+        if dark_value is None:
+            dark_value = light_value
```
rdswift, line 1734: Is it worth checking that the values are valid, perhaps with some sort of regex like `^(\#[a-fA-F0-9]{6}|\#[a-fA-F0-9]{8}|[a-zA-Z]{3,})$`.  This could avoid problems later if trying to use an invalid value.

zas: Good point, but I didn't use regex to check, see https://github.com/metabrainz/picard/pull/3350/commits/44cb82b4caf8e3016fc4c6157de8a750cde92bb7

rdswift: I think that's actually a better approach than using a regex.  Thanks.

rdswift (approved):

picard/plugin3/api_impl.py
```diff
@@ -1658,6 +1666,157 @@ def enable(api):
...
+        if key in self._registered_colors:
+            self._registered_colors.remove(key)
+
+    def unregister_all_colors(self) -> None:
+        """Unregister all colors registered by this plugin."""
+        for key in list(self._registered_colors):
+            _unregister_color(key)
+        self._registered_colors.clear()
+
+    def get_plugin_color(self, name: str) -> str:
```
phw, line 1765: There is some inbalance in the naming. You use `register_color` to register a color, but `get_color` won't give you the color again. Instead you have to use `get_plugin_color`.

I'd turn that around and have `register_color` and `get_color` be for the plugin's colors. Then add something else for the Picard defined colors (`get_default_color`, `get_theme_color`?).

picard/plugin3/api_impl.py
```diff
@@ -1658,6 +1666,157 @@ def enable(api):
...
+        Returns:
+            Hex color string (e.g., '#c80000').
+
+        Example::
+
+            error_color = api.get_color('entity_error')
+        """
+        return interface_colors.get_color(name)
+
+    def get_qcolor(self, name: str):
```
phw, line 1809: For the pre-defined colors we should define an enum instead of the string. That makes the API more discoverable.

picard/plugin3/api_impl.py
```diff
@@ -1658,6 +1666,157 @@ def enable(api):
...
+        Returns:
+            Hex color string (e.g., '#ff0000').
+
+        Example::
+
+            color = api.get_plugin_color('highlight')
+        """
+        return interface_colors.get_color(self._color_key(name))
+
+    def get_plugin_qcolor(self, name: str):
```
phw, line 1780: Do we need the separate qcolor, or would it be sufficient to always return `QColor`?

phw (changes requested):
I'm again not sure about this. Is there a specific use case we want to cover?

Anyway, I would do a bit separate API. I would expose a `theme` property in the `PluginApi`, which then gives an object that gives access to theme related functionality. It exposes the color related functions and the `is_dark` property.

One reason for this is that we have things more isolated. Currently the additions blow up the PluginApi surface significantly with a lot of very similar, but rather specific functions, that only a few plugins will ever need. Also a separate class allows us to derive it from `QObject` and provide a signal for theme change (once we implement this fully in Picard).

zas:
> I'm again not sure about this. Is there a specific use case we want to cover?

Yes, providing same colors to plugins UI for consistency, and let plugin devs use colors that the user can modify through existing interface colors UI.

> Anyway, I would do a bit separate API. I would expose a `theme` property in the `PluginApi`, which then gives an object that gives access to theme related functionality. It exposes the color related functions and the `is_dark` property.
> 
> One reason for this is that we have things more isolated. Currently the additions blow up the PluginApi surface significantly with a lot of very similar, but rather specific functions, that only a few plugins will ever need. Also a separate class allows us to derive it from `QObject` and provide a signal for theme change (once we implement this fully in Picard).

Yes, I tend to agree, but I just following existing approach, and we have rather "flat" API, so I kept the same logic.
If we go in this direction (which makes total sense), we should perhaps revisit existing API functions to see if we can group them.

picard/plugin3/api_impl.py
zas: Yes, it makes sense.

picard/plugin3/api_impl.py
zas: If we go with the `api.theme.` idea, it should be `api.theme.color.get` no? So we could keep get_color for core Picard.

phw: I don't fully follow :) You would make `theme.color` a separate object?

zas:
I'll close this PR, this needs more thoughts. I still think we need to expose theme/colors somehow for plugins.
I'll create a new PR later.

phw:
I agree, on both. We should expose this somehow, but also get the API right. One thought: Maybe we should implement live dark / light switching first. That isn't too far off. Qt does already most of it, and also exposes a signal. It is a few places where colors don't update properly. Some is our colors, some might be Qt issues we might need to workaround. Our theme class could expose a signal when color mode changes.

PR #3351, opened 2026-08-16
PICARD-3387: Use TagVar for plugin-registered script variables

zas:
## Summary
* This is a…
  * Bug fix
  * Feature addition
  * Refactoring
* **Describe this change in 1-2 sentences**: Replace `PluginVariable` with `TagVar` in the script variables extension point, giving plugin-registered variables the same capabilities as built-in tags. This makes `tag_names()` include plugin variables, fixing autocomplete in tag dropdowns and restoring plugin attribution in the scripting documentation.

## Problem
* JIRA ticket: [PICARD-3387](https://tickets.metabrainz.org/browse/PICARD-3387)

When a plugin registers custom tags using `api.register_script_variable()`, those tags appear in the script editor autocomplete but do NOT appear in:
* The edit tag dialog dropdown (metadata box)
* The tag list editor autocomplete (options pages)

The root cause is that `tag_names()` only queries `ALL_TAGS` (the static built-in registry), while plugin variables lived in a separate `ExtensionPoint[PluginVariable]` dataclass that was only consumed by the script editor.

## Solution

Instead of maintaining a separate `PluginVariable` dataclass, plugin-registered variables now use `TagVar` instances stored in the `ExtensionPoint`. This allows:

1. **`tag_names()` extended** to yield from both `ALL_TAGS` and `ext_point_script_variables` — all UI consumers of `tag_names()` (edit tag dialog, tag list editor) automatically get plugin variables.
2. **Full TagVar capabilities** for plugin variables: `is_hidden`, `is_multi_value`, proper `script_name()` formatting, tooltip/documentation support.
3. **Plugin attribution via `plugin_id`** — resolved to a display name at render time via the plugin manager. The scripting documentation panel shows `[Plugin: <name>]` for plugin-registered variables, consistent with how script functions handle attribution.
4. **New API parameters** — `register_script_variable()` now accepts `title` and `is_multi_value`. Hidden status is derived from the `_` prefix convention (names starting with `_` are hidden).
5. **Unified iteration** — A single `_all_tag_vars()` helper iterates both built-in and plugin TagVars, simplifying `tag_names()`, `script_variable_tag_names()`, and the `CompletionProvider`.
6. **Simplified CompletionProvider** — Plugin variables flow through the standard tag iteration path; no separate `get_plugin_variable_names` callable injection needed.
7. **Registration restriction** — A plugin cannot register the same base name as both hidden and non-hidden (e.g., both `foo` and `_foo`). Raises `ValueError` if attempted. This keeps lookup functions simple (name-only matching).

## AI Usage

* Significant use (e.g. code structure, tests development, etc.)

Investigation, implementation, and testing developed with AI assistance. Design decisions made collaboratively with human review.

## Action

Additional actions required:
* Update Picard [documentation](https://github.com/metabrainz/picard-docs) (please include a reference to this PR)

The plugin v3 API docs (`docs/PLUGINSV3/API.md`) have been updated in this PR to document the `title`, `is_multi_value` parameters, the `_` prefix convention, and the `ValueError` raised for conflicting registrations.

### DIFF
--- a/picard/extension_points/script_variables.py
+++ b/picard/extension_points/script_variables.py
@@ -18,28 +18,19 @@
 # along with this program; if not, write to the Free Software
 # Foundation, Inc., 51 Franklin Street, Fifth Floor, Boston, MA 02110-1301, USA.
 
-from dataclasses import dataclass
 from typing import TYPE_CHECKING
 
-from picard.i18n import _
 from picard.plugin import ExtensionPoint
 from picard.script.variable_pattern import VARIABLE_NAME_FULLMATCH_RE
 from picard.tags import script_variable_tag_names
+from picard.tags.tagvar import TagVar
 
 
 if TYPE_CHECKING:
     from picard.plugin3.api_impl import PluginApi
 
 
-@dataclass(frozen=True)
-class PluginVariable:
-    name: str
-    documentation: str
-    plugin_name: str
-    title: str | None = None
-
-
-ext_point_script_variables = ExtensionPoint[PluginVariable](label='script_variables')
+ext_point_script_variables = ExtensionPoint[TagVar](label='script_variables')
 
 
 def _check_if_duplicate_variable_name(name: str) -> str | None:
@@ -49,7 +40,10 @@ def _check_if_duplicate_variable_name(name: str) -> str | None:
 
     for var in ext_point_script_variables:
         if name == var.name:
-            sources.append(f'"{var.plugin_name}"')
+            if var.plugin_id:
+                sources.append(f'"{var.plugin_id}"')
+            else:
+                sources.append('"Unknown"')
 
     return ', '.join(sources) if sources else None
 
@@ -80,7 +74,7 @@ def register_script_variable(
     api : PluginApi, optional
         The plugin API instance
     title : str, optional
-        Display title for the metadata box (e.g., "Pinned Tags").
+        Display title for the metadata box (e.g., "Caller").
         If provided, the tag will show this title instead of the raw name.
 
     Examples
@@ -97,25 +91,22 @@ def register_script_variable(
 
     if api:
         module_name = api.module_path
+        plugin_id = api.plugin_id
     else:
         module_name = 'unknown'
-
-    plugin_name = api.manifest.name_i18n() if api else _("Unknown Plugin")
-    plugin_documentation = documentation or ""
-    if plugin_documentation and plugin_name:
-        plugin_documentation += "\n\n"
-    if plugin_name:
-        plugin_documentation += _("Plugin: %s") % plugin_name
+        plugin_id = None
 
     # Remove any existing entry with the same name from this plugin to avoid duplicates
     ext_point_script_variables.unregister(module_name, lambda item: item.name == name)
     ext_point_script_variables.register(
         module_name,
-        PluginVariable(
+        TagVar(
             name=name,
-            documentation=plugin_documentation,
-            plugin_name=plugin_name,
-            title=title,
+            shortdesc=title,
+            longdesc=documentation,
+            is_from_mb=False,
+            is_populated_by_picard=False,
+            plugin_id=plugin_id,
         ),
     )
 
@@ -170,7 +161,7 @@ def get_plugin_variable_documentation(name: str) -> str | None:
     """
     for var in ext_point_script_variables:
         if var.name == name:
-            return var.documentation
+            return var._longdesc
     return None
 
 
@@ -189,5 +180,5 @@ def get_plugin_variable_title(name: str) -> str | None:
     """
     for var in ext_point_script_variables:
         if var.name == name:
-            return var.title
+            return var._shortdesc
     return None
--- a/picard/extension_points/script_variables.py
+++ b/picard/extension_points/script_variables.py
@@ -20,9 +20,9 @@
 
 from typing import TYPE_CHECKING
 
+from picard.const.tags import ALL_TAGS
 from picard.plugin import ExtensionPoint
 from picard.script.variable_pattern import VARIABLE_NAME_FULLMATCH_RE
-from picard.tags import script_variable_tag_names
 from picard.tags.tagvar import TagVar
 
 
@@ -35,7 +35,9 @@ ext_point_script_variables = ExtensionPoint[TagVar](label='script_variables')
 
 def _check_if_duplicate_variable_name(name: str) -> str | None:
     sources = []
-    if name in set(script_variable_tag_names()):
+    # Check against built-in system variables only (not plugin-registered ones)
+    builtin_names = {tagvar.script_name() for tagvar in ALL_TAGS if tagvar.is_script_variable}
+    if name in builtin_names:
         sources.append("System Variables")
 
     for var in ext_point_script_variables:
--- a/picard/extension_points/script_variables.py
+++ b/picard/extension_points/script_variables.py
@@ -119,6 +119,12 @@ def register_script_variable(
             longdesc=documentation,
             is_hidden=is_hidden,
             is_multi_value=is_multi_value,
+            # Locked attributes for plugin-registered variables
+            is_preserved=False,
+            is_script_variable=True,
+            is_tag=True,
+            is_calculated=False,
+            is_file_info=False,
             is_from_mb=False,
             is_populated_by_picard=False,
             plugin_id=plugin_id,
--- a/picard/extension_points/script_variables.py
+++ b/picard/extension_points/script_variables.py
@@ -113,8 +113,21 @@ def register_script_variable(
         module_name = 'unknown'
         plugin_id = None
 
+    # Reject registering the same base name with a different is_hidden status.
+    # A plugin cannot register both 'foo' and '_foo' — the base name must be unique.
+    for var in ext_point_script_variables:
+        if var.name == name and var.is_hidden != is_hidden:
+            prefix = '_' if is_hidden else ''
+            existing_prefix = '_' if var.is_hidden else ''
+            msg = (
+                f"Cannot register '{prefix}{name}': "
+                f"'{existing_prefix}{name}' is already registered "
+                f"with different hidden status."
+            )
+            raise ValueError(msg)
+
     # Remove any existing entry with the same name from this plugin to avoid duplicates
-    ext_point_script_variables.unregister(module_name, lambda item: item.name == name and item.is_hidden == is_hidden)
+    ext_point_script_variables.unregister(module_name, lambda item: item.name == name)
     ext_point_script_variables.register(
         module_name,
         TagVar(
@@ -166,7 +179,7 @@ def get_plugin_variable_documentation(name: str) -> str | None:
     Parameters
     ----------
     name : str
-        The variable name
+        The variable name (bare name without prefix)
 
     Returns
     -------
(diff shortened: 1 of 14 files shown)


rdswift (commented):
Looks good.

I wonder if there's any value in extending the plugin api registration function to include some of the other `TagVar` attributes such as `is_multi_value`.

**EDIT:**  I know...  Not part of this PR.

picard/tags/__init__.py
```diff
@@ -90,6 +90,15 @@ def tag_names():
     yield from ALL_TAGS.names(selector=lambda tv: tv.is_tag)
 
 
+def all_tag_names():
```
phw: I think we should instead extend `tag_names` to return all variable names.

picard/tags/__init__.py
```diff
@@ -90,6 +90,15 @@ def tag_names():
...
+    """All tag names available for user assignment: built-in tags + plugin-provided tags."""
+    # Inline import to avoid circular dependency: script_variables imports from this module.
+    from picard.extension_points.script_variables import get_plugin_variable_names
+
+    yield from tag_names()
+    yield from get_plugin_variable_names()
```
phw: `tag_names` returns only variables that are considered tags (not hidde). So this function would need to exclude hidden variables.

phw (changes requested):
I don't think we should add another function. Instead we should extend `tag_names` and in extension also `visible_tag_names` and `hidden_tag_names`. That way it works directly in all auto completes, and I think it matches expectations. The current PR replaces all uses if `tag_names` with `all_tag_names` anyway.

It also makes for a better internal API, without two confusing similar names.

picard/extension_points/script_variables.py
```diff
@@ -64,7 +60,12 @@ def _is_valid_plugin_variable_name(name: str | None) -> bool:
 
 
 def register_script_variable(
-    name: str, documentation: str | None = None, api: 'PluginApi | None' = None, title: str | None = None
+    name: str,
+    documentation: str | None = None,
+    api: 'PluginApi | None' = None,
+    title: str | None = None,
+    is_hidden: bool = False,
+    is_multi_value: bool = False,
```
rdswift, line 67: I'm wondering if this will break existing plugins that register names beginning wih an underscore (such as "Additional Artists Variables").  Is it worth including some clean-up code to provide backwards compatibility?  Perhaps something like:

```python
if name.startswith('_') and not is_hidden:
    is_hidden = True
    name = name[1:]  # to preserve tags starting with multiple underscores
```

**EDIT:**  Thinking about this a bit more, it may be more appropriate to put this check into the `TagVar.__init__()` method instead.  That way it would make the correction regardless of where the tag definition originated.  PR https://github.com/metabrainz/picard/pull/3352 added.

rdswift (commented):
I like the change to using `TagVar`.  This should provide better compatibility between plugin-added variables and system scripting variables for processing.

picard/extension_points/script_variables.py
zas: @rdswift I handled it in https://github.com/metabrainz/picard/pull/3351/commits/9899406a6cc31d29f960253b73af68a2fb415dbf

picard/extension_points/script_variables.py
```diff
@@ -91,31 +98,42 @@ def register_script_variable(
         msg = "Invalid script variable name; use letters, digits, underscores."
         raise ValueError(msg)
 
+    # Backward compatibility: plugins may register hidden variables using the
+    # script syntax prefix "_". Normalize to bare name + is_hidden=True.
+    if name.startswith('_'):
+        name = name[1:]
+        is_hidden = True
```
rdswift, line 99: Is it possible to have a hidden variable start with two underscores so that it is accessed as `%__something%`?  If so, then I think this patch will be a problem if the variable is registered as `name='_something', is_hidden=True` because the `is_hidden` value provided would have no meaning.  In this case the variable would be accessed as `%_something%`, which would be the same as if it were registered as `name='something', is_hidden=true`.  If they wanted the variable accessed as `%__something%`, they would have to register it as `name='__something', is_hidden=False` or `name='__something', is_hidden=True` (because in both cases the `is_hidden` value provided would have no meaning).

If the test was `if name.startswith('_') and not is_hidden:` as I originally proposed, then the variable would be properly registered with either `name='__something, is_hidden=False` or `name='_something', is_hidden=True`, and would be accessed as `%__something%`.  To me, this seems clearer and has no ambiguity in the way the registration information is processed.

Also, I'm not 100% sure but I think this name processing test may have to be moved ahead of the `if not _is_valid_plugin_variable_name(name):` test to catch the case when someone tries to register the variable as `name='_', is_hidden=False`, which I believe would end up with the name being an empty string.

picard/extension_points/script_variables.py
rdswift: Okay, I'll cancel my PR.

picard/extension_points/script_variables.py
rdswift: Another alternative, and this may actually be the simplest solution, is to remove the `is_hidden` setting from the `register_script_variable()` arguments available to the plugin author and always default it to False.  This way, the plugin author simply registers the name as the way they expect it to be accessed between percent signs (as they do now), and your code properly strips the first underscore and sets `is_hidden` to True as appropriate.

This way, if the plugin registers the name as `'_something'` it will get the name `'something'` in the `TagVar` with `is_hidden` set to True, and it will be accessed as `%_something%`.  If the plugin registers the name as `'__something'` then it will get the name `'_something'` in the `TagVar` with `is_hidden` set to True, and it will be accessed as `%__something%`.  Everything as expected.

I still think you might have to move the name processing before the `_is_valid_plugin_variable_name(name)` check though for the same reason I mentioned earlier.

picard/extension_points/script_variables.py
```diff
@@ -91,31 +98,42 @@ def register_script_variable(
...
-            plugin_name=plugin_name,
-            title=title,
+            shortdesc=title,
+            longdesc=documentation,
+            is_hidden=is_hidden,
+            is_multi_value=is_multi_value,
+            # Locked attributes for plugin-registered variables
+            is_preserved=False,
+            is_script_variable=True,
+            is_tag=True,
```
phw: Should this be `is_tag=not is_hidden`? As I understand it doesn't make sense to have a variable marked as both is_tag and is_hidden.

picard/tags/__init__.py
```diff
@@ -87,7 +87,14 @@ def parse_subtag(name):
 
 
 def tag_names():
+    """Tag names available for user assignment: built-in tags + plugin-provided tags."""
+    # Inline import to avoid circular dependency: script_variables imports from this module.
+    from picard.extension_points.script_variables import ext_point_script_variables
+
     yield from ALL_TAGS.names(selector=lambda tv: tv.is_tag)
```
phw: Side question, shouldn't this also check for `not tv.is_hidden`? I understand that built-in variables don't have both set, though. So maybe not needed.

picard/extension_points/script_variables.py
```diff
@@ -91,31 +98,42 @@ def register_script_variable(
...
-    plugin_name = api.manifest.name_i18n() if api else _("Unknown Plugin")
-    plugin_documentation = documentation or ""
-    if plugin_documentation and plugin_name:
-        plugin_documentation += "\n\n"
-    if plugin_name:
-        plugin_documentation += _("Plugin: %s") % plugin_name
+        plugin_id = None
 
     # Remove any existing entry with the same name from this plugin to avoid duplicates
     ext_point_script_variables.unregister(module_name, lambda item: item.name == name)
```
phw, line 132: Existing issue but, but it wasn't possible for a plugin to register both `~foo` and `foo` as variable. We should check for `is_hidden` also. Fixed in https://github.com/zas/picard/pull/34

picard/extension_points/script_variables.py
zas: @rdswift I went for the second solution (removing is_hidden from the plugin API), as it is cleaner and simpler for the user. I also moved the check for the name, and added a test to verify it works as expected.
See https://github.com/metabrainz/picard/pull/3351/commits/505918000a14750428ab7f0ab904dbf0bfed269a

phw (changes requested):
The change really simplifies things. Generally this looks good. I found two smaller issues and have fixes and some additional cleanup in https://github.com/zas/picard/pull/34

picard/extension_points/script_variables.py
rdswift: Good catch.

picard/extension_points/script_variables.py
rdswift: I never noticed that before, but I think you're right.

phw: I added the change to my PR as 16182e3184141ddf43d762e601474e6cfaf4ac94

phw:
@zas @rdswift One thing I noticed just now is that the PR removes showing the info about which plugin registered the variable in the docs. Was this intentional or should we keep the info?

rdswift:
> @zas @rdswift One thing I noticed just now is that the PR removes showing the info about which plugin registered the variable in the docs. Was this intentional or should we keep the info?

I believe that we should identify in the documentation generated for the script editor the plugin that registered the tag/variable.  This should be able to be added automatically to the `TagVar` item created either by appending the information to the description or (my preference) setting it in the `additionaldesc` parameter, without the plugin author specifying the argument.

picard/tags/__init__.py
```diff
@@ -123,7 +130,13 @@ def file_info_tag_names():
 
 def script_variable_tag_names():
     """Tag names available to scripts (used by script editor completer)"""
+    # Inline import to avoid circular dependency: script_variables imports from this module.
+    from picard.extension_points.script_variables import ext_point_script_variables
+
     yield from (tagvar.script_name() for tagvar in ALL_TAGS if tagvar.is_script_variable)
+    for var in ext_point_script_variables:
+        if var.is_script_variable:
+            yield var.script_name()
```
phw: Maybe use the same syntax as the line before for consistency:

```python
yield from (var.script_name() for var in ALL_TAGS if var.is_script_variable)
yield from (var.script_name() for var in ext_point_script_variables if var.is_script_variable)
```

phw: See cb1bf2a52f536206dbbf68658cacb11013ff40a0

phw (approved):

picard/tags/__init__.py
phw: This was done as part of cb1bf2a52f536206dbbf68658cacb11013ff40a0

picard/extension_points/script_variables.py
phw: Now decided to disallow same name as hidden and non-hidden, changed with 9393e7d8c675c17a12aee70322b1956e2a37828f

phw (dismissed):

rdswift (commented):
Just testing this locally and I think I've run into a couple of problems.  I'm using the "Additional Artists Variables" plugin to provide the plugin variables for my testing.

---

1. The tooltip on hovering over a variable in the script editor shows "No description available."

Image

---

2. The plugin indicator in the description in the documentation is not indented like the rest of the description and there is no space between it and the following variable name. This makes it a bit confusing as to whether the "[Plugin:...]" line is associated with the variable above or the one below.

Image

---

Both display correctly in the current `master` branch.

Image

Image

rdswift (approved):
Just tested it here using the "Additional Artists Variables" plugin to provide the variables, and everything seems to work okay and display as expected.  Once this is released, I'll need to add the `is_multi_value` setting to some of the variables registered in the plugin.

Looks good.  Thanks.

phw (approved):

PR #3356, opened 2026-08-21
Improve scripts/tools/fix-header.py: exclusions, performance, reliability

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [ ] Feature addition
  * [x] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Improve `scripts/tools/fix-header.py` with path exclusions, better performance (batch git log, git ls-files for discovery), reliability improvements (error handling), and developer UX additions (--dry-run, --verbose).

## Problem

The fix-header script had several issues when run on larger codebases:
- No directory exclusion: it would descend into `.venv/`, `build/`, `node_modules/` etc.
- Poor performance: one `git log` subprocess per file (hundreds of subprocess spawns)
- No error handling for git failures, encoding issues, or file I/O errors
- Noisy debug output by default
- No way to preview changes before applying them
- Hardcoded `.venv310` instead of a generic pattern

## Solution

Applied as a series of focused commits:

1. **Path exclusions & batch git log** — default exclusion list, `--exclude`/`--no-default-excludes` options, single `git log --name-only` call for all files
2. **Code simplification** — extracted helpers (`_resolve_alias`, `_run_git`), use sets for year tracking, `fnmatch` for dir matching, removed `defaultdict` usage
3. **git ls-files for file discovery** — automatically respects `.gitignore`, with `--no-git-ls` fallback
4. **Module-level regex compilation** — avoid recompiling patterns per-file
5. **Pre-compiled exclusion regex** — single combined regex via `fnmatch.translate()`
6. **Batch edge case fix** — distinguish "batch ran, no data" from "batch didn't run"
7. **--dry-run flag** — show unified diffs without writing
8. **Logging cleanup** — default to INFO, `-v/--verbose` for DEBUG

## AI Usage

* [ ] No AI/LLM use
* [x] Primarily AI developed

All code changes were authored by AI (Kiro CLI) with human review and direction at each step.

## Action

Additional actions required:

### DIFF
--- a/scripts/tools/fix-header.py
+++ b/scripts/tools/fix-header.py
@@ -32,7 +32,7 @@ import sys
 logging.basicConfig(
     force=True,
     format="%(asctime)s:%(levelname)s: %(message)s",
-    level=logging.DEBUG,
+    level=logging.INFO,
     stream=sys.stderr,
 )
 
@@ -447,6 +447,7 @@ def main():
         help='Show unified diff of changes without writing files',
     )
     parser.add_argument('-r', '--recursive', action='store_true', default=False, help='Search through subfolders')
+    parser.add_argument('-v', '--verbose', action='store_true', default=False, help='Enable debug logging')
     parser.add_argument('--encoding', default='utf-8', help='File encoding of the source files')
     parser.add_argument(
         '--exclude',
@@ -476,6 +477,9 @@ def main():
         help='Disable git ls-files for file discovery (use os.walk with exclusions instead)',
     )
     args = parser.parse_args()
+
+    if args.verbose:
+        logging.getLogger().setLevel(logging.DEBUG)
 
     # Build the exclusion set
     if args.no_default_excludes:
--- a/scripts/tools/fix-header.py
+++ b/scripts/tools/fix-header.py
@@ -19,7 +19,7 @@
 
 
 import argparse
-from collections import defaultdict
+from fnmatch import fnmatch
 import itertools
 import logging
 import os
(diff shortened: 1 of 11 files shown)


scripts/tools/fix-header.py
```diff
@@ -56,19 +56,103 @@
...
+        '.eggs',
+        '__pycache__',
+        'build',
+        'dist',
+        'htmlcov',
+        'locale',
+        'node_modules',
+        '*.egg-info',
+    }
+)
```
phw, line 81: I wonder whether we should also get this from `.gitignore`

scripts/tools/fix-header.py
```diff
@@ -56,19 +56,103 @@
     'Stevil Knevil': 'StevilKnevil',
     'deepakss-74': 'Deepak Kumar',
     'Thuna': 'Thuna-Cing',
-    'Frederik "Freso" S. Olesen': 'Frederik “Freso” S. Olesen',
+    'Frederik "Freso" S. Olesen': 'Frederik "Freso" S. Olesen',
```
phw: This was to unify both variations used in the comments. If both left and right side are the same this makes no sense.

phw (changes requested):
Overall this looks good to go. Some minor comments to check

scripts/tools/fix-header.py
zas: It is only for the os.walk() mode, since by default it now uses git ls-files (which uses .gitignore).

scripts/tools/fix-header.py
zas: Yes, fixed!

phw (approved):
LGTM

PR #3359, opened 2026-08-21
PICARD-2442: Runtime theme switching without restart

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences:**

Enable runtime dark/light theme switching and interface color changes without requiring a restart.

## Problem

* JIRA ticket: [PICARD-2442](https://tickets.metabrainz.org/browse/PICARD-2442)

Currently switching theme (dark/light) or modifying interface colors requires restarting Picard. This is a poor UX especially when experimenting with colors or when the system theme changes.

Also related: [PICARD-2160](https://tickets.metabrainz.org/browse/PICARD-2160) (macOS dark mode switching not updating list elements).

## Solution

Infrastructure:
* Make `BaseTheme` a `QObject` with `theme_changed` and `colors_changed` signals
* Make `apply_theme()` re-entrant (starts from fresh `standardPalette()`, handles both directions)
* Add `InterfaceColors.reload()` to switch color sets after theme change
* Emit signals after palette update, auto-reload interface colors
* Force `unpolish`/`polish`/`update` on all widgets for immediate visual refresh
* Add `UiTheme.from_color_scheme()` and `to_color_scheme()` helpers
* Skip redundant `apply_theme` on system color scheme change
* Allow system style changes if `ui_theme` is "default"

Widget fixes:
* Log view: connect existing `refresh_colors()` to signal
* MainPanel tree: refresh cached class-level color attributes
* Syntax highlighter: rebuild `HighlightFormat` objects and rehighlight
* Toolbar: swap dark/light icon variants and extension button stylesheet
* Options dialog: regenerate profile highlight stylesheets
* About dialog: re-apply styling on theme change
* Cover art box: already dynamic (reads colors at paint time, no change needed)

Options pages:
* Theme change in Interface options now applies live
* Color changes emit `colors_changed` instead of showing restart warning
* Use `QSignalBlocker` context manager for theme combo

Debug:
* `Ctrl+Shift+D` shortcut toggles theme at runtime for testing
* `--debug-opts=theme` for theme switching diagnostics

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:
* [ ] Further testing on macOS and Windows
* [ ] Plugin API for theme awareness (separate PR, see [PICARD-3388](https://tickets.metabrainz.org/browse/PICARD-3388))

### DIFF
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -230,6 +230,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
         self.toolbar = None
         self.player = None
+        self.plugin_updates_button = None
         self.status_indicators = []
         if DesktopStatusIndicator:
             self.ready_for_display.connect(self._setup_desktop_status_indicator)
@@ -367,12 +368,13 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
 
     def _update_statusbar_plugin_icon(self):
         """Swap the plugin updates icon for the current theme."""
-        if hasattr(self, 'plugin_updates_button'):
-            if theme.is_dark_theme:
-                icon = QtGui.QIcon(":/images/plugin-dark.png")
-            else:
-                icon = QtGui.QIcon(":/images/plugin.png")
-            self.plugin_updates_button.setIcon(icon)
+        if not self.plugin_updates_button:
+            return
+        if theme.is_dark_theme:
+            icon = QtGui.QIcon(":/images/plugin-dark.png")
+        else:
+            icon = QtGui.QIcon(":/images/plugin.png")
+        self.plugin_updates_button.setIcon(icon)
 
     def _setup_player(self):
         # Local import: player depends on QtMultimedia which is an optional runtime dependency
--- a/picard/ui/mainwindow/__init__.py
+++ b/picard/ui/mainwindow/__init__.py
@@ -338,6 +338,7 @@ class MainWindow(QtWidgets.QMainWindow, PreserveGeometry):
     def _setup_debug_shortcuts(self):
         """Set up keyboard shortcuts for development/debugging."""
         shortcut = QtGui.QShortcut(QtGui.QKeySequence("Ctrl+Shift+D"), self)
+        shortcut.setContext(QtCore.Qt.ShortcutContext.ApplicationShortcut)
         shortcut.activated.connect(self._toggle_theme)
 
     def _toggle_theme(self):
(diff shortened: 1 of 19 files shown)


picard/ui/mainwindow/__init__.py
```diff
@@ -329,6 +333,50 @@ def setupUi(self):
         # Initially set the plugin updates available status in the status bar.
         self._update_statusbar_plugin_updates_available()
 
+        theme.theme_changed.connect(self._on_theme_changed)
+        self._setup_debug_shortcuts()
```
phw: Do we remove this again before merge? I'd be in favor of it.

phw (changes requested):
We are close :) Code so far looks good. I found two open issues.

First, the interface color options need to handle the theme switch as well. Currently they show the colors for the theme on load time. While testing this I also found an existing bug with this option page: Changed colors are directly written to the global `InterfaceColors` instance, but only saved on option saving. This means if the user changes a color and cancels the dialog, the color will still be used in the current session of picard.

I think we need to change this interface, so it holds the loaded colors temporarily and modifies the copy. And it should hold a set for light and dark. On theme switch it can then update the buttons and allow editing. On saving the colors will be applied.

It looks like a behavior similar to this was once intended, but never used. There is a `self.new_colors = {}` variable defined that looks like it was supposed to hold the temporary colors.

Another issue was the script documentation resetting scroll position on theme change. See https://github.com/zas/picard/pull/37 for a fix.

picard/ui/mainwindow/__init__.py
zas: Done in https://github.com/metabrainz/picard/pull/3359/commits/15ec57dcbe167b2d20e789afa0b0774bd9254079

zas:
> I think we need to change this interface, so it holds the loaded colors temporarily and modifies the copy. And it should hold a set for light and dark. On theme switch it can then update the buttons and allow editing. On saving the colors will be applied.
> 
> It looks like a behavior similar to this was once intended, but never used. There is a `self.new_colors = {}` variable defined that looks like it was supposed to hold the temporary colors.
> 

This was addressed by https://github.com/metabrainz/picard/pull/3359/commits/3180c279fa5fd460edcea55b625760e78e413e91 and https://github.com/metabrainz/picard/pull/3359/commits/723558b033b8497bc405a9b2748075d310991320

> Another issue was the script documentation resetting scroll position on theme change. See [zas#37](https://github.com/zas/picard/pull/37) for a fix.

Thanks, merged.

phw (approved):
Great work. Touches quite a few places with custom color handling. But glad to see that quite a few of them had the color updating already more or less isolated and just needed to be hooked up to the theme change signal.

In my testing this is working really smooth.

PR #3362, opened 2026-08-25
PICARD-2442: Fix switching UI theme to Default having no effect until restart

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

  Fix the runtime theme switcher so that selecting **Default** actually takes
  effect immediately when the previously applied theme was Dark (or Light),
  instead of only after a restart.

## Problem

* JIRA ticket (_optional_): PICARD-2442

Reported by rdswift (running from source on Linux): switching from **Light** or
**Dark** to **Default** had no visible effect until Picard was restarted.
Switching between Light/Dark, or from Default to Light/Dark, worked fine.

Root cause: `BaseTheme.get_system_theme()` resolved the system theme from the
live application palette via `palette_is_dark(app.palette())`. At runtime that
palette reflects whatever theme Picard last applied, not the operating system
setting. After the user selected **Dark** and then switched back to **Default**,
`app.palette()` was dark, so `get_system_theme()` incorrectly returned `DARK`.
`apply_theme(DARK)` then early-returned (because dark was already applied) and
nothing repainted.

This matches the reported log line:

```
ui/theme.get_system_theme: No system dark mode detected, but palette is already dark.
```

The palette heuristic is only reliable at startup, before Picard has overridden
the application palette.

## Solution

`get_system_theme()` now determines the OS theme independently of any palette
Picard applied at runtime:

* Prefer `QStyleHints.colorScheme()` when available (Qt 6.5+), which reports the
  operating system color scheme directly.
* Otherwise fall back to the style's `standardPalette()` rather than the live
  `app.palette()`, so a runtime theme override no longer skews detection.

The OS-specific dark-mode detection strategies (freedesktop portal, GNOME, KDE,
XFCE, LXQt) remain the first and highest-priority signal and are unchanged.

Added regression tests covering both detection paths:

* `test_get_system_theme_ignores_runtime_applied_dark_palette` — reproduces
  rdswift's scenario (light OS, dark runtime palette, no strategy reporting
  dark) and asserts `LIGHT` is returned.
* `test_get_system_theme_uses_style_hints_color_scheme` — verifies the color
  scheme hint is trusted over a dark runtime palette.

`pytest test/test_theme.py` passes (48 tests); `ruff format`, `ruff check`, and
`ty check` are clean on the changed files.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

Investigation of the root cause, the fix in `get_system_theme()`, and the
regression tests were developed with AI assistance and reviewed by the author.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -397,8 +397,31 @@ class BaseTheme(QtCore.QObject):
         if any(strategy() for strategy in self._dark_mode_strategies):
             return UiTheme.DARK
 
-        if palette_is_dark(app.palette()):
-            log.debug("No system dark mode detected, but palette is already dark.")
+        # Ask Qt directly for the OS color scheme when available (Qt 6.5+).
+        # Unlike app.palette(), this reflects the operating system setting and
+        # is not affected by any palette Picard applied at runtime.
+        # get_style_hints() only returns non-None when setColorScheme (Qt 6.8+)
+        # is present, which implies colorScheme() (Qt 6.5+) is available too.
+        style_hints = get_style_hints()
+        if style_hints is not None:
+            color_scheme = style_hints.colorScheme()
+            if color_scheme == QtCore.Qt.ColorScheme.Dark:
+                log.debug("System color scheme reported as dark.")
+                return UiTheme.DARK
+            if color_scheme == QtCore.Qt.ColorScheme.Light:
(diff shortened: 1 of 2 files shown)


phw (approved):
LGTM

PR #3365, opened 2026-08-25
PICARD-700: Add albumartists tag

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Adds a first-class, multi-valued `albumartists` tag, written to files by
default, mirroring the existing `artists` tag for track artists.

## Problem

* JIRA ticket: PICARD-700

Picard already exposes the individual album artists internally as the hidden
`~albumartists` variable, but never writes them to files. This means albums
with multiple album artists can only be stored as the joined `albumartist`
string, which cannot be reliably split back into individual artists.

Several players read a dedicated `ALBUMARTISTS` tag and recommend writing it:

* Kodi reads `ALBUMARTISTS` (Vorbis/ID3/APE/MP4); its wiki tells users to add a
  script to write it because Picard does not do so yet.
* Navidrome maps `albumartists` directly and prefers it over splitting the
  singular `albumartist` string.
* Jellyfin reads it when the "Prefer nonstandard artists tag" library option is
  enabled (otherwise it splits the singular tag).

## Solution

Promotes `albumartists` to a real, multi-valued written tag, populated by
default just like `artists`. Tag mapping follows the existing `artists` tag:

* Vorbis (FLAC/Ogg): `ALBUMARTISTS` (native)
* ID3 (MP3): `TXXX:ALBUMARTISTS`
* MP4: `----:com.apple.iTunes:ALBUMARTISTS`
* ASF/WMA: `WM/ALBUMARTISTS`

The hidden `~albumartists` variable is kept for backward compatibility with
existing scripts using `%_albumartists%`. It is never written to files, so it
cannot drift out of sync with the new tag.

Scope note: only the plural `albumartists` (names) is added. No plural sort tag,
since sorting uses the singular `ALBUMARTISTSORT` and there are no consumers for
a plural one. Per-artist IDs are already covered by `MUSICBRAINZ_ALBUMARTISTID`.

Open question: for ASF/WMA I used `WM/ALBUMARTISTS` (all caps) to match the
existing `WM/ARTISTS`. The sibling official attributes use PascalCase
(`WM/AlbumArtist`, `WM/AlbumArtistSortOrder`), so `WM/AlbumArtists` would also
be defensible — happy to change it if preferred.

Tests cover round-trip for all four formats. The full test suite passes.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI assistance was used to investigate the existing tag handling, design the
backward-compatible approach for `~albumartists`, implement the format mappings
and tests, and verify against downstream player behaviour. All changes were
reviewed by the author.

## Action

Additional actions required:

### DIFF
--- a/picard/const/tags.py
+++ b/picard/const/tags.py
@@ -82,9 +82,24 @@ ALL_TAGS = TagVars(
             'The artists primarily credited on the release. These could be either "standardized" or '
             '"as credited" depending on whether the "Use standardized artist names" metadata option is enabled.'
         ),
+        is_multi_value=True,
+        see_also=('albumartist', '_albumartists'),
+    ),
+    TagVar(
+        'albumartists',
+        shortdesc=N_('Album Artists'),
+        longdesc=N_(
+            'The artists primarily credited on the release. These could be either "standardized" or '
+            '"as credited" depending on whether the "Use standardized artist names" metadata option is enabled.'
+        ),
+        additionaldesc=N_(
+            'This hidden variable is retained for backward compatibility with existing scripts. '
+            'It is never written to files; use the "albumartists" tag instead.'
+        ),
         is_hidden=True,
         is_tag=False,
         is_multi_value=True,
+        see_also=('albumartists',),
     ),
     TagVar(
         'albumartists_countries',
--- a/picard/formats/id3.py
+++ b/picard/formats/id3.py
@@ -227,6 +227,7 @@ class ID3File(File):
             'ASIN': 'asin',
             'MusicMagic Fingerprint': 'musicip_fingerprint',
             'ARTISTS': 'artists',
+            'ALBUMARTISTS': 'albumartists',
             'DIRECTOR': 'director',
             'WORK': 'work',
             'Writer': 'writer',
--- a/test/test_mbjson.py
+++ b/test/test_mbjson.py
@@ -119,6 +119,7 @@ class ReleaseTest(MBJSONTest):
         release_to_metadata(self.json_doc, m, a)
         self.assertEqual(m['album'], 'The Dark Side of the Moon')
         self.assertEqual(m['albumartist'], 'Pink Floyd')
+        self.assertEqual(m['albumartists'], 'Pink Floyd')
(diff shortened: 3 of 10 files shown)


test/test_mbjson.py
phw, line 489: I'd add a check here, since this- test is actually multi-artist:

```python
self.assertEqual(m.getall('~albumartists'), m.getall('albumartists'))
```

picard/formats/id3.py
```diff
@@ -255,6 +256,7 @@ class ID3File(File):
     __rename_freetext = MappingProxyType(
         {
             'Artists': 'ARTISTS',
+            'Album Artists': 'ALBUMARTISTS',
```
phw: No, this is specifically for updating tags Picard wrote differently in the past. There is no widespread use of a `TXXX:Album Artists` being written

phw (changes requested):

test/test_mbjson.py
zas: Amended & force-pushed

picard/formats/id3.py
zas: Amended & force-pushed

picard/const/tags.py
```diff
@@ -75,16 +75,32 @@
         ),
         is_filterable=True,
     ),
+    TagVar(
+        'albumartists',
+        shortdesc=N_('Album Artists'),
+        longdesc=N_(
+            'The artists primarily credited on the release, as a multi-value list. These could be either '
```
phw: Just noticed: The description differs slightly from the one for the hidden variable ("..., as a multi-value list."). Do we want this difference? Or should we use the same longdesc on both, to ease translations?

zas: Yes, better re-use, I'll fix it

zas: Done, rebased, in https://github.com/metabrainz/picard/pull/3365/commits/5078ca75565de160aebd23a6cda8ef743f7f62ff

phw (approved):
LGTM

PR #3366, opened 2026-08-25
PICARD-2442: Fix theme switch to Default staying on the last applied theme

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
* **Describe this change in 1-2 sentences**: Switching the UI theme option from Dark (or Light) back to Default had no effect until restart on Qt 6.8+. This makes the switch take effect immediately.

## Problem

* JIRA ticket: PICARD-2442

Follow-up to dfd21e3 (PICARD-2442). @rdswift reported the issue was still
reproducible on the current master (Linux Mint / X-Cinnamon, Qt 6.11, `qt5ct`
platform theme): starting on the light system theme, selecting **Dark**, then
switching back to **Default** left the UI stuck on dark until the next restart.

His debug log showed `get_system_theme()` reporting the system scheme as
*dark* right after a Dark theme had been applied, even though the OS theme was
light — so `apply_theme()` early-returned and nothing switched back.

Root cause: `apply_theme()` calls `QStyleHints.setColorScheme()`, which mutates
what `QStyleHints.colorScheme()` subsequently returns. When the option is set to
Default, `get_system_theme()` read `colorScheme()` back to detect the OS theme,
but it now reported the *last theme Picard applied* rather than the operating
system's theme. A restart cleared this, which is why it only worked after
restarting.

This only affects Qt 6.8+ (where `setColorScheme` exists); on older Qt the
runtime theme-switching path is not used.

## Solution

Capture the OS color scheme once at setup, **before** Picard ever calls
`setColorScheme()`, and refresh it only on genuine `colorSchemeChanged` signals.
`get_system_theme()` now trusts this cached value instead of the live,
pollutable `colorScheme()`.

The self-triggered `colorSchemeChanged` echo from our own `setColorScheme()`
call is delivered synchronously (verified with a standalone script on Qt
6.11/xcb), so it is already filtered by the existing `_applying_theme` guard and
does not re-pollute the cache.

Includes regression tests covering the pollution scenario (Dark → Default) and
continued tracking of genuine runtime OS color-scheme changes.

A separate commit adds the desktop/session environment variables
(`XDG_CURRENT_DESKTOP`, etc.) to the startup debug log, since the dark-mode
detection strategies depend on them and they make theme issues easier to
diagnose from a user's log (this is what pinpointed the X-Cinnamon case).

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

The root-cause analysis, the fix, the regression tests, and a standalone
verification script (used to confirm the `setColorScheme`/`colorScheme`
pollution and that the signal echo is synchronous) were developed with
significant AI assistance, reviewed by the author.

## Action

Additional actions required:

### DIFF
--- a/picard/tagger.py
+++ b/picard/tagger.py
@@ -381,6 +381,17 @@ class Tagger(QtWidgets.QApplication):
 
         # log interesting environment variables
         log.debug("Qt Env.: %s", " ".join("%s=%r" % (k, v) for k, v in os.environ.items() if k.startswith('QT_')))
+        desktop_env_vars = (
+            'XDG_CURRENT_DESKTOP',
+            'XDG_SESSION_DESKTOP',
+            'XDG_SESSION_TYPE',
+            'DESKTOP_SESSION',
+            'KDE_FULL_SESSION',
+        )
+        log.debug(
+            "Desktop Env.: %s",
+            " ".join("%s=%r" % (k, os.environ[k]) for k in desktop_env_vars if k in os.environ),
+        )
 
     def _init_gettext(self, config, localedir):
         """Initialize gettext"""
--- a/picard/ui/theme.py
+++ b/picard/ui/theme.py
@@ -275,6 +275,13 @@ class BaseTheme(QtCore.QObject):
         self._accent_color_is_system: bool = False
         self._applying_theme: bool = False
         self._dark_mode_strategies: list[Callable[[], bool]] = []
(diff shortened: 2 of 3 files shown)


phw (approved):
LGTM

rdswift (approved):
Works perfectly when I tested it here.  Thanks.

PR #3369, opened 2026-08-27
PICARD-3398: Handle cancelled OAuth in browser integration

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Handle the OAuth `error` parameter in the browser integration `/auth` endpoint so that cancelled authentication properly releases pending requests instead of leaving them stuck.

## Problem

* JIRA ticket: PICARD-3398

When a release loading requires authentication and the user cancels the OAuth flow in the browser, MusicBrainz redirects to `/auth?error=access_denied&state=xxx`. The handler only checked for the `code` parameter — if missing, it returned HTTP 400 without invoking the state callback. This left pending requests permanently stuck in "loading album information" state, with no way to recover (reload fails, removing and re-adding the release also fails).

## Solution

The `_auth` method now handles the `error` query parameter from the OAuth redirect. When present, it verifies the state token, retrieves the associated callback, and invokes it with `successful=False` and the error message. This triggers the normal failure path (`on_mb_authorization_finished` → `discard_authorized_requests`), which releases the stuck requests.

Also refactored `_auth` to extract the shared state verification logic, eliminating code duplication between the success and error paths.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used for drafting the initial implementation and tests, with manual review and refinement.

## Action

Additional actions required:

### DIFF
--- a/picard/browser/server.py
+++ b/picard/browser/server.py
@@ -245,25 +245,38 @@ class RequestHandler(BaseHTTPRequestHandler):
         else:
             self._response(400, 'Missing parameter "token".')
 
+    _OAUTH_CALLBACK_PARAMS = frozenset(('code', 'error'))
+
     def _auth(self, args):
-        if 'code' in args and args['code']:
-            tagger = tagger_instance()
-            oauth_manager = tagger.webservice.oauth_manager
-            try:
-                state = args.get('state', [''])[0]
-                callback = oauth_manager.verify_state(state)
-            except OAuthInvalidStateError:
-                self._response(400, 'Invalid "state" parameter.')
-                return
+        if not args.keys() & self._OAUTH_CALLBACK_PARAMS:
+            self._response(400, 'Invalid OAuth callback parameters.')
+            return
+
+        def qs_value(key):
+            return args.get(key, [''])[0]
+
+        tagger = tagger_instance()
+        oauth_manager = tagger.webservice.oauth_manager
+        try:
+            callback = oauth_manager.verify_state(qs_value('state'))
+        except OAuthInvalidStateError:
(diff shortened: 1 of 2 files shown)


phw (approved):
Tested and works. LGTM

picard/browser/server.py
```diff
@@ -245,25 +245,38 @@ def _add_release(self, args):
...
                 scopes='profile tag rating collection submit_isrc submit_barcode',
                 callback=callback,
             )
             self._response(200, "Authentication successful, you can close this window now.", 'text/html')
         else:
-            self._response(400, 'Missing parameter "code".')
+            error_msg = qs_value('error_description') or qs_value('error')
+            log.info("%s: Authorization denied: %s", LOG_PREFIX, error_msg)
+            to_main(callback, successful=False, error_msg=error_msg)
+            self._response(200, "Authentication cancelled, you can close this window now.", 'text/html')
```
phw, line 279: Sorry, minor thing I noticed after initial review: Remove the content-type parameter here. It defaults to `text/plain`, which is what we currently return.

I think we discussed this before, but we could have a small HTML template that gets returned here instead, for a nicer user experience. But right now, we return just text.

phw (changes requested):

picard/browser/server.py
zas: Yes, I planned to do that after this is merged, and even created a ticket for it yesterday, as https://tickets.metabrainz.org/projects/PICARD/issues/PICARD-3399

phw (approved):

PR #3370, opened 2026-08-28
PICARD-3400: Use consistent OAuth scopes for browser-integration login

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Minor / simple change (like a typo)
  * [ ] Other
* **Describe this change in 1-2 sentences**:

Fix the browser-integration OAuth login exchanging the authorization code with
the wrong (unprefixed) scopes, which left the token without the requested
permissions so collections failed to load with HTTP 401 right after login.

## Problem

* JIRA ticket: PICARD-3400

After logging in to MusicBrainz via the browser integration flow, Picard
immediately re-prompted for authentication and failed to load collections:

```
E: Network request error for https://musicbrainz.org/ws/2/collection ->
   AuthenticationRequiredError (HTTP 401)
I: MusicBrainz authentication declined
E: Error loading collections: Authorization required
```

Picard requests OAuth scopes in two places and they had drifted apart:

- `picard/tagger.py` (`mb_login`, the authorization request) used the current
  `musicbrainz:`-prefixed scopes.
- `picard/browser/server.py` (the `/auth` callback that exchanges the
  authorization code) still used the old, unprefixed scopes
  (`profile tag rating collection submit_isrc submit_barcode`).

`OAuthManager.on_exchange_authorization_code_finished()` persists the scopes
passed to `exchange_authorization_code()` together with the refresh token, so
the stored token was associated with the wrong scope set and authorized
requests such as `GET /ws/2/collection` returned HTTP 401.

The server migrated scope names to the `musicbrainz:` prefix; PICARD-2772
updated the scopes in `tagger.py` but did not update `browser/server.py`.

## Solution

Define the requested scopes once as `METABRAINZ_OAUTH_SCOPES` (in
`picard/const`) and use it in both the authorization request
(`tagger.mb_login`) and the authorization-code exchange (`browser/server._auth`)
so the two paths can no longer drift. Added a regression test asserting the
`/auth` code exchange requests the shared, prefixed scopes (including
`musicbrainz:collection`).

## AI Usage

In accordance with the AI use policy portion of the [MetaBrainz Contribution Guidelines](https://github.com/metabrainz/guidelines/blob/master/README.md), the level of AI/LLM use in the development of this Pull Request is:

* [ ] No AI/LLM use
* [ ] Minimal use (e.g. autocompletion)
* [ ] Moderate use (e.g. suggestions regarding code fragments)
[...]

PR #3371, opened 2026-08-28
PICARD-3399: Styled browser-integration OAuth pages, ui_locales, favicon

zas:
## Summary

* This is a…
  * [ ] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

Replace the plain-text browser-integration OAuth callback responses with simple
styled HTML pages (Picard logo, clear success/failure status, auto-close
attempt), request the authorization pages in Picard's UI language via
`ui_locales`, and serve the logo as a favicon.

## Problem

* JIRA ticket: PICARD-3399

After completing (or cancelling) MusicBrainz authorization in the browser, the
`/auth` endpoint returned bare plain-text responses such as
"Authentication successful, you can close this window now." This is a poor user
experience and, for a French (or other non-English) Picard, the authorization
pages on the server were shown in English regardless of the configured UI
language.

## Solution

* **Styled response pages** (`picard/browser/auth_responses.py`): success,
  cancelled, and error pages with the Picard logo, a clear status line
  (green check / cross), and a best-effort `window.close()` (browsers such as
  Firefox refuse to close tabs not opened by script, so the page always shows
  an explicit "You can close this window." message rather than promising
  automatic closing). Cancellation/error detail is derived from the stable
  OAuth error code and translated via gettext, falling back to the server's
  raw description for unknown codes. Invalid/expired `state` (e.g. reloading
  the callback) now renders the styled error page instead of raw text.

image

image

* **UI language** (`picard/oauth.py`): add the OpenID Connect `ui_locales`
  parameter to the authorization URL, derived from the configured UI language
  (falling back to the system locale, converting POSIX `fr_FR` to BCP 47
  `fr-FR`). This asks the MetaBrainz server to render its login/consent pages
  in Picard's language. Honored by the MetaBrainz server side
  (MEB-190 / metabrainz.org#626) and ignored gracefully by servers that do not
  support it, as required by the spec.

* **Favicon**: serve the Picard logo SVG at `/favicon.ico` (browsers request it
  automatically; previously it 404'd) and reference it from the auth pages.

image

Notes:
* The new user-facing strings are marked for translation and will be picked up
  by the normal Weblate workflow.
* The related OAuth scope-mismatch bug is fixed separately in PICARD-3400
  (#3370); this PR intentionally leaves the requested scopes untouched to avoid
  overlap.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used for the page design/implementation, the `ui_locales` handling, and
test development, with manual review and iteration.

## Action

Additional actions required:

Server-side counterparts on metabrainz.org:
* MEB-190 (metabrainz.org#626): honor `ui_locales` and localize OAuth error
  messages — required for the authorization/error pages to be shown in Picard's
  language.
* MEB-191: localize the OAuth scope descriptions on the consent page (still
  English even when the page chrome is localized).

### DIFF
--- a/picard/browser/auth_responses.py
+++ b/picard/browser/auth_responses.py
@@ -123,7 +123,7 @@ def _render(*, title, status, detail, icon, status_class, hint):
         status=escape(status),
         detail=escape(detail),
         icon=escape(icon),
-        status_class=status_class,
+        status_class=escape(status_class),
         colors=COLORS,
         logo=_PICARD_LOGO_SVG,
         hint=escape(hint),
--- a/picard/browser/auth_responses.py
+++ b/picard/browser/auth_responses.py
@@ -1,0 +1,218 @@
+# Picard, the next-generation MusicBrainz tagger
+#
+# Copyright (C) 2026 Laurent Monin
+#
+# This program is free software; you can redistribute it and/or
+# modify it under the terms of the GNU General Public License
+# as published by the Free Software Foundation; either version 2
+# of the License, or (at your option) any later version.
+#
+# This program is distributed in the hope that it will be useful,
+# but WITHOUT ANY WARRANTY; without even the implied warranty of
+# MERCHANTABILITY or FITNESS FOR A PARTICULAR PURPOSE.  See the
+# GNU General Public License for more details.
+#
+# You should have received a copy of the GNU General Public License
+# along with this program; if not, see <https://www.gnu.org/licenses/>.
+
+"""Styled HTML response pages for the browser integration OAuth callback.
+
+These pages replace the previous plain-text responses shown to the user in
+their browser after completing (or cancelling) the MusicBrainz authorization
+flow. They embed the Picard logo, attempt to close the browser tab
+automatically, and clearly distinguish success from failure.
+"""
+
+from collections import namedtuple
+from html import escape
+
+from picard import PICARD_APP_NAME
+from picard.i18n import gettext as _
+
+
+# Page colors, by role. Change here to restyle the pages.
+Colors = namedtuple('Colors', ('ok', 'failed', 'background', 'text', 'hint'))
+COLORS = Colors(ok="#771b85", failed="#eb743b", background="#fffedb", text="#2b2b2b", hint="#6b6b6b")
+
+# The Picard logo, inlined from resources/img-src/Picard_logo_small_no_text.svg
+# (editor metadata stripped, whitespace collapsed). Inlined so no extra asset
+# needs to be shipped or served. Regenerate by hand if the source logo changes.
+_PICARD_LOGO_SVG = (
+    '<svg xmlns="http://www.w3.org/2000/svg" viewBox="0 0 100 100">'
+    '<g transform="translate(5.5,0)">'
+    '<polygon points="46,98 87,74 87,26 46,2" style="fill:#eb743b"/>'
+    '<polygon points="2,26 2,74 43,98 43,2" style="fill:#771b85"/>'
+    '<path style="fill:#fffedb" d="m 78.281,58.087 c -1.239,-1.97 -3.38,-3.148 -5.718,-3.148 -0.666,0 -1.324,0.103'
+    ' -1.956,0.295 -2.632,-3.125 -5.819,-4.775 -8.602,-5.636 2.235,-1.714 4.208,-3.771 5.909,-6.153 1.74,0.394'
+    ' 3.63,0.127 5.162,-0.832 3.158,-2.006 4.103,-6.197 2.104,-9.348 -1.249,-1.975 -3.39,-3.152 -5.724,-3.152'
+    ' -1.283,0 -2.533,0.363 -3.616,1.05 -3.146,1.991 -4.093,6.181 -2.105,9.336 0.13,0.207 0.27,0.404 0.422,0.596'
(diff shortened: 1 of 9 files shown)


picard/browser/auth_responses.py
```diff
@@ -0,0 +1,224 @@
...
+<p class="status {status_class}"><span class="icon" aria-hidden="true">{icon}</span>{status}</p>
+<p class="detail">{detail}</p>
+<p class="hint">{hint}</p>
+</main>
+<script>
+// Best-effort: try to close the tab. Browsers only allow this for windows
+// opened by a script (e.g. Firefox refuses to close a tab reached by a
+// redirect), so the page always shows an explicit "you can close this
+// window" message rather than promising automatic closing.
+setTimeout(function () {{ window.close(); }}, {close_delay_ms});
```
phw: Do we need this then at all? I don't really see a case right now where this would actually close the tab.

picard/browser/auth_responses.py
```diff
@@ -0,0 +1,224 @@
...
+'''
+
+
+def _render(*, title, status, detail, icon, status_class, hint, close_delay_ms=1000):
+    return _PAGE_TEMPLATE.format(
+        title=escape(title),
+        status=escape(status),
+        detail=escape(detail),
+        icon=escape(icon),
+        status_class=status_class,
```
phw: Missing `escape`?

picard/browser/server.py
```diff
@@ -261,7 +266,7 @@ def qs_value(key):
         try:
             callback = oauth_manager.verify_state(qs_value('state'))
         except OAuthInvalidStateError:
-            self._response(400, 'Invalid "state" parameter.')
+            self._response(400, auth_responses.error_page(), 'text/html')
```
phw, line 275: Previously we had a bit more specific message here. Should we pass an error string to `error_page`, or log the more detailed message? I'm afraid we loose some sometimes important debug information. Specifically invalid state can give some hints. We had cases where people somehow had Picard restarted in between, and invalid state is a sign that a different Picard instance is responding. While generally invalid parameters is rather a sign that something else called the endpoint.

picard/oauth.py
```diff
@@ -264,6 +281,14 @@ def get_authorization_url(self, scopes, callback: Callable):
...
             'access_type': 'offline',
         }
+        # Ask the authorization server to render its pages in Picard's UI
+        # language, using the OpenID Connect `ui_locales` parameter (a
+        # space-separated list of BCP 47 tags). Honored by the MetaBrainz
+        # server (MEB-190); ignored gracefully by servers that do not support
+        # it, as required by the spec.
+        ui_locales = _get_ui_locales()
+        if ui_locales:
+            params['ui_locales'] = ui_locales
```
phw, line 277: Is this already working, or is it a proposal for the server side to implement?

phw (commented):
This is great. Simple but effective design. A couple of comments, but nothing actually a blocker to me.

picard/oauth.py
phw: Sorry, found it. This is MEB-190 / https://github.com/metabrainz/metabrainz.org/pull/626

picard/oauth.py
```diff
@@ -54,6 +55,22 @@
...
 
+def _get_ui_locales():
+    """Return Picard's UI language as a BCP 47 tag for OAuth `ui_locales`.
+
+    Uses the configured UI language when set, otherwise the system locale.
+    Returns an empty string when no meaningful locale can be determined
+    (e.g. the "C" locale), so the caller can omit the parameter.
+    """
+    config = get_config()
+    language = config.setting['ui_language'] or QLocale.system().name()
```
phw: I wonder whether we should add a `get_locale` to `picard.i18n` instead. This here is actually querying more the side effects.

The "ui_language" option is used to force a language, but often it is empty. QLocale is initialized in `setup_gettext`, but our primary way to change the locale is the Python `locale` package.

zas:
Thanks for the review! Addressed each point in separate commits:

- **`window.close()` (auth_responses.py):** you're right, it never actually closes a redirect-opened tab (Firefox refuses, Chrome only closes script-opened windows). Removed the script and the `close_delay_ms` plumbing; the page just shows "You can close this window." — `dce1482a2`.
- **Missing `escape` (_render):** `status_class` is now escaped (it's injected into a `class` attribute) — `6331e39ff`.
- **Lost debug info on invalid state vs invalid params (server.py):** kept the generic user-facing page but added distinct `log.warning` messages for the two branches, so the "invalid state → possibly a different Picard instance" hint is preserved — `1420edfd2`.
- **`ui_locales` — is it working / `get_locale` in i18n:** yes, this is the client side of MEB-190 (metabrainz.org#626). And good call on the locale source: added `picard.i18n.get_locale_bcp47()` which derives the BCP 47 tag from the active locale (the Python `locale` package, our primary mechanism) rather than re-deriving from the often-empty `ui_language` option + QLocale. oauth now uses that — `9a657394b`.

picard/browser/auth_responses.py
zas: Dropped, I agree, it doesn't work well anyway.

picard/browser/auth_responses.py
zas: Yes, solved.

picard/browser/server.py
zas: Restored in last iteration.

picard/oauth.py
zas: Yes, better, done in last iteration.

phw (approved):
This is a real nice improvement. Makes the login process more friendly

PR #3372, opened 2026-08-28
PICARD-3401: Do not submit ListenBrainz listens while disabled

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Stop Picard from submitting queued ListenBrainz listens when ListenBrainz
submission is disabled in the settings.

## Problem

* JIRA ticket: PICARD-3401

With ListenBrainz submission disabled, Picard still submitted queued listens on
startup and kept retrying every couple of minutes:

```
E: Network request error for https://api.listenbrainz.org/1/submit-listens
   -> AuthenticationRequiredError (HTTP 401)
E: ListenBrainz batch submission failed: ...
```

`ListenBrainzSubmissionService.__init__` loaded the persisted queue, and
`ListenQueue.load()` started the submission timer. That timer submitted queued
listens independently of the `listenbrainz_enabled` setting — the
`enable()`/`disable()` gating only connected/disconnected the player signals,
not the queue's submission timer.

## Solution

Load the persisted queue without starting submission, and gate submission on
the service state:

* `ListenQueue.load()` no longer schedules submission.
* Added `ListenQueue.start()` / `stop()`.
* `ListenBrainzSubmissionService.enable()` starts the queue (flushing any
  listens queued in a previous session); `disable()` stops it.

The queue is still loaded and persisted while disabled, so no listens are lost.

Added regression tests covering load-without-submit and the enable/disable
start/stop gating.

Note: an invalid token can still cause the batch submission to retry in a loop
*while enabled*; that separate robustness issue is tracked in its own ticket.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used to diagnose the issue from runtime logs and implement the fix and
tests, with manual review.

## Action

Additional actions required:

### DIFF
--- a/picard/ui/player/listenbrainz.py
+++ b/picard/ui/player/listenbrainz.py
@@ -98,6 +98,10 @@ class ListenBrainzSubmissionService:
         self._current: PreparedSubmission | None = None
         self._lbapi = ListenBrainzAPIHelper(webservice)
         self._queue = ListenQueue(self._lbapi)
+        # Load any persisted queue so listens are not lost, but do not start
+        # submitting: submission is only scheduled once the service is enabled
+        # (see enable()). Otherwise queued listens would be submitted even when
+        # ListenBrainz submission is disabled in the settings.
         self._queue.load()
         tagger.register_cleanup(self._queue.save)
 
@@ -111,9 +115,15 @@ class ListenBrainzSubmissionService:
         self._player.playback_state_changed.connect(self.on_playback_state_changed)
         self._player.position_changed.connect(self.on_position_changed)
 
+        # Flush any listens queued from a previous session.
+        self._queue.start()
+
     def disable(self):
         if not self._enabled:
             return
+
+        # Stop submitting queued listens while disabled.
+        self._queue.stop()
 
         self._player.media_changed.disconnect(self.on_media_changed)
         self._player.playback_state_changed.disconnect(self.on_playback_state_changed)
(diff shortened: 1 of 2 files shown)


phw (changes requested):
I actually don't think this change is correct, and the issue reported in the ticket is invalid. Disabling listen submission should still submit anything queued up to this point. It is currently working exactly as intended: If disabled, no new listens are queued up for submission. If there are still unsubmitted listens in the queue, they will get submitted. Afterwards the timer suspends.

I have for example the listen submission as a quick setting. I want to be able to enable / disable it at any time when playing audio. I do expect everything that I listened to before disabling the setting to be submitted.

The issue this tries to solve is the submission timer running forever even with the setting disabled. But the timer only running while there are listens to submit. Once the submission was possible it will suspend the timer.

PICARD-3402 is a separate issue, and suspending on permanent errors makes sense to avoid the endless retry without a chance of success.

phw (changes requested):
I tested this again, with and without this patch. Currently the timer is really working as intended. If listens are disabled and nothing is queued up, the timer doesn't run. If there are queued items, they gets submitted and the timer gets suspended.

I think the added `start` and `stop` functions show the underlying misunderstanding: The way the `ListenQueue` it does not need a `start` function nor does it need to be scheduled when the option is enabled.

The actual interface to use are just `add` and `load`. Usually you just `add` a listen. If possible, it gets submitted right away. If that's not possible, only then the listen is added to the queue and the timer is started. After all listens have been submitted, the timer suspends again.

That means usually there is not much of a queue, and the timer is most of the time not needed, it is only used in the case of network issues or misconfiguration.

The only other case is `load`. It is only used in case there were still queued listens when Picard was closed. Then the timer is scheduled on load (and again suspends again after the submission).

If listens are disabled, `add` is never called. As I wrote previously, submission is intentional of queued listens. They were recorded when the function was enabled. That allows the user to enable and disable listen submission in between. While it is on, listens are recorded, when turned off they aren't. And it works offline and online.

zas:
> I tested this again, with and without this patch. Currently the timer is really working as intended. If listens are disabled and nothing is queued up, the timer doesn't run. If there are queued items, they gets submitted and the timer gets suspended.


Then I don't get why I had tons of attempts to submit with nothing in the cache and listenbrainz disabled.
I suggest we close this PR and the matching ticket as invalid, until I eventually reproduce it.


[PICARD-3402](https://tickets.metabrainz.org/browse/PICARD-3402) was based on this one though;

zas:
You're right, thanks for testing this thoroughly. I misdiagnosed the cause: the timer already suspends once the queue drains, and submitting listens that were queued while enabled (even after disabling) is the intended, desirable behavior — it supports toggling the setting during playback. There's no bug here.

The endless-retry symptom I saw was really the permanent-error (401) case, which is PICARD-3402. Closing this as invalid; the retry-loop fix continues in #3373.

PR #3373, opened 2026-08-28
PICARD-3402: Stop ListenBrainz retry loop on permanent submission errors

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other

Stop the ListenBrainz submission retry loop when the failure is permanent
(e.g. HTTP 401 from an invalid token), and resume submitting once the token is
fixed.

> Note: this branch builds on PICARD-3401 (#3372); it uses the queue
> start()/stop() lifecycle introduced there. Review/merge #3372 first.

## Problem

* JIRA ticket: PICARD-3402

With ListenBrainz submission enabled but an invalid user token, batch
submission failed with HTTP 401 and retried indefinitely every couple of
minutes, hammering the API:

```
E: ListenBrainz batch submission failed:
   data=b'{"code":401,"error":"Provided Authorization header is invalid."}'
```

`ListenQueue._on_submit_batch_response` only logged the error and left the
submission timer running, so a permanent failure never stopped retrying. The
single-listen handler also treated 401 as retryable.

## Solution

Add a helper `_is_temporary_submission_error()` that classifies only 429 (rate
limit) and 5xx (server error) as transient; everything else, notably 401, is
permanent.

* On a permanent batch error the retry timer is stopped (no more looping).
* The same classifier is applied to single-listen submissions, so 401 is no
  longer re-queued.
* **Recovery:** submission is re-armed when the `listenbrainz_token` setting
  changes (while enabled), so the queued listens are submitted again once the
  user fixes their token. Submission also resumes on re-enable or a newly
  queued listen.

Added tests for the classifier, the batch handler stopping on a permanent
error (and continuing on a transient one), and the token-change re-arm.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

AI was used to diagnose the loop from runtime logs and implement the fix and
tests, with manual review.

## Action

Additional actions required:

Depends on / builds upon PICARD-3401 (#3372).

### DIFF
--- a/picard/ui/player/listenbrainz.py
+++ b/picard/ui/player/listenbrainz.py
@@ -100,6 +100,14 @@ class ListenBrainzSubmissionService:
         self._queue = ListenQueue(self._lbapi)
         self._queue.load()
         tagger.register_cleanup(self._queue.save)
+        # Resume submitting queued listens when the token changes: a permanent
+        # auth failure (e.g. 401) suspends the retry timer, so fixing the token
+        # should give the queued listens another chance.
+        get_config().setting.setting_changed.connect(self._on_setting_changed)
+
+    def _on_setting_changed(self, name, old_value, new_value):
+        if name == 'listenbrainz_token':
+            self._queue.schedule_submission()
 
     def enable(self):
         if self._enabled:
@@ -195,7 +203,7 @@ class ListenQueue:
                         self._queue = json.load(f, object_hook=from_json)
                         if count := len(self._queue):
                             log.debug("Loaded %d listens from queue", count)
-                            self._schedule_submission()
+                            self.schedule_submission()
                 except json.JSONDecodeError as e:
                     log.error("Failed to load listen queue: %s", e)
                     self._queue = []
@@ -224,14 +232,18 @@ class ListenQueue:
     ):
         if error:
             status_code = reply.attribute(QNetworkRequest.Attribute.HttpStatusCodeAttribute)
-            if not status_code or int(status_code) in {401, 429} or int(status_code) >= 500:
+            # Keep the listen queued regardless of the error, so it is not lost
+            # (e.g. a wrong token can be fixed later). Only schedule a retry for
+            # transient errors; for permanent ones the timer stays suspended
+            # until submission is retriggered (e.g. on a token change).
+            self._append(payload)
+            if _is_temporary_submission_error(status_code):
                 log.warning(
                     'Temporary ListenBrainz submission error (queued for retry): data=%s, error=%s', data, error
                 )
-                self._append(payload)
-                self._schedule_submission()
+                self.schedule_submission()
             else:
-                log.error('ListenBrainz submission failed: data=%s, error=%s', data, error)
+                log.error('ListenBrainz submission failed (queued, retry suspended): data=%s, error=%s', data, error)
         else:
             log.debug('ListenBrainz submission successful: data=%s', data)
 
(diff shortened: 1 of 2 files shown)


picard/ui/player/listenbrainz.py
```diff
@@ -105,6 +105,18 @@ def __init__(self, player: Player, webservice: WebService, tagger: 'Tagger'):
         self._queue.load()
         tagger.register_cleanup(self._queue.save)
 
+        # Re-arm submission when the token changes: a permanent auth failure
+        # (e.g. 401) stops the retry timer, so fixing the token should resume
+        # submitting the queued listens.
+        setting = get_config().setting
+        setting_changed = getattr(setting, 'setting_changed', None)
```
phw: Unneeded use of getattr, setting_changed is always defined in ConfigSection.

picard/ui/player/listenbrainz.py
```diff
@@ -224,7 +255,7 @@ def _on_submit_listen_response(
     ):
         if error:
             status_code = reply.attribute(QNetworkRequest.Attribute.HttpStatusCodeAttribute)
-            if not status_code or int(status_code) in {401, 429} or int(status_code) >= 500:
+            if _is_temporary_submission_error(status_code):
```
phw, line 240: This now completely drops the listen on a 401 instead of queuing it. Is this actually wanted?

picard/ui/player/listenbrainz.py
```diff
@@ -98,9 +98,25 @@ def __init__(self, player: Player, webservice: WebService, tagger: 'Tagger'):
...
+        # (e.g. 401) stops the retry timer, so fixing the token should resume
+        # submitting the queued listens.
+        setting = get_config().setting
+        setting_changed = getattr(setting, 'setting_changed', None)
+        if setting_changed is not None:
+            setting_changed.connect(self._on_setting_changed)
+
+    def _on_setting_changed(self, name, old_value, new_value):
+        if name == 'listenbrainz_token' and self._enabled:
+            self._queue.start()
```
phw: When the first commit gets dropped it is still needed to schedule submission on setting change. `start` is a bit misleading, it implies that this needs to be called for having the queue work at all. But the only reason to submit is to check whether there are queued listens that failed because of wrong token, and hence fixed settings might heal the situation.

I would rename `_schedule_submission` to `schedule_submission` and call `self._queue.schedule_submission` here directly. That makes the intention clearer.

phw (changes requested):
Except for the changes from #3372, which are also included here and should be dropped, this looks mostly good.

My only concern is that the changed code now drops listens completely if the token is set wrongly, instead of queuing them up.

picard/ui/player/listenbrainz.py
zas: It is needed for tests

zas:
Thanks, addressed all points:

- **Dropped the #3372 changes** — that PR is closed as invalid. This branch is now a single commit off master with just the permanent-error handling.
- **No longer drops listens on 401** — good catch, it did (the single-listen path skipped `_append` on a non-temporary error). Now the listen is always kept in the queue on any error; only the retry *timer* is suspended for permanent errors (both the single-listen and batch paths), so nothing is lost and there's no endless retry.
- **Renamed** `_schedule_submission` → `schedule_submission` and call it directly.
- **Recovery** — the service reschedules submission when `listenbrainz_token` changes, so fixing a wrong token gives the queued listens another chance. Connected `setting_changed` directly (dropped the `getattr` guard); the tests provide a config with the signal instead.

picard/ui/player/listenbrainz.py
zas: Fixed, good catch.

picard/ui/player/listenbrainz.py
zas: Done.

phw (approved):
LGTM

PR #3374, opened 2026-08-28
PICARD-2844: Support language for lyrics tags, make language mandatory in comment tags with description

phw:
## Summary

* This is a…
  * [x] Bug fix
  * [x] Feature addition
  * [ ] Refactoring
  * [ ] Other

## Problem

* JIRA ticket (_optional_): PICARD-2844, PICARD-1877

This change makes the tags lyrics, syncedlyrics and comments consistent to follow the syntax `{tagname}:{lang}:{description}`. While both "lang" and "description" are optional, a missing language but non-empty description is indicated by a double colon like `comment::withdescription`.

The general syntax for language and description was already possible for syncedlyrics (since this was newly implemented specifically to follow this syntax) and in parts for comments. The PR applies the following changes:

- The `lyrics` tag did not support language at all, hence language set in ID3 tags was lost. This is a bug, see PICARD-1877.
- The `comment` tag supported both e.g. `comment:lang:description` and also just `comment:description`. Now the latter form gets stanardized to `comment::description` to avoid the ambiguity with only language.
- Unify the implementation for all of `comment`, `lyrics` and `syncedlyrics`

## Solution

- Combine  helper functions `parse_subtag` and `parse_comment_tag` into `parse_lang_desc_tag` and use it for all three tags in the ID3 implementation
- Fix `parse_lang_desc_tag` to handle special cases:
  - Treat language code part with not exactly 3 letters (as required by ID3) as part of the description. This gives a fallback to cases that use a single colon
  - The description part could contain additional colons
- Add corresponding `create_lang_desc_tag`
- Added a migration to update scripts to add the double colon syntax for `lyrics` and `comment` tags.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

## Action

Additional actions required:

### DIFF
--- a/picard/tags/__init__.py
+++ b/picard/tags/__init__.py
@@ -52,7 +52,11 @@ def parse_lang_desc_tag(name: str, default_language: str = 'xxx') -> tuple[str, 
     else:
         lang = default_language
 
-    if len(split) > 2:
+    if len(lang) != 3:
+        # If language is not 3 chars, assume default and use remaining parts as description
+        lang = default_language
+        desc = ':'.join(split[1:])
+    elif len(split) > 2:
         desc = split[2]
     else:
         desc = ''
--- a/picard/tags/__init__.py
+++ b/picard/tags/__init__.py
@@ -41,27 +41,41 @@ RE_COMMENT_LANG = re.compile('^([a-zA-Z]{3}):')
 
 
 def parse_lang_desc_tag(name: str, default_language: str = 'xxx') -> tuple[str, str]:
-    """
-    Parses a tag name like "lyrics:XXX:desc", where XXX is the language.
-    If language is not set, the colons are still mandatory, and "eng" is
-    assumed by default.
-    """
-    split = name.split(':')
-    if len(split) > 1 and split[1]:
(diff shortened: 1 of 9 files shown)


picard/tags/__init__.py
```diff
@@ -77,16 +50,32 @@ def parse_subtag(name):
...
-        lang = 'eng'
+        lang = default_language
 
-    if len(split) > 2:
+    if len(lang) != 3:
+        # If language is not 3 chars, assume default and use remaining parts as description
+        lang = default_language
+        desc = ':'.join(split[1:])
+    elif len(split) > 2:
         desc = split[2]
```
zas: This is not working if description contains ':'.
The whole function is rather hard to understand, I went for another approach in phw/picard#39 (which also fixes the issue and add tests to ensure it is the case).

zas (changes requested):

picard/tags/__init__.py
phw: Thanks, I merged your changes

zas (approved):
LGTM

PR #3378, opened 2026-08-30
PICARD-3404: Fix a rating of 0 being discarded when loading ASF/WMA/WMV files

zas:
## Summary

* This is a…
  * [x] Bug fix
  * [ ] Feature addition
  * [ ] Refactoring
  * [ ] Other
  A rating of 0 was written to ASF/WMA/WMV files correctly but silently
  dropped on load, so saving a 0 rating and reloading returned no rating.
  This restores 0 on load, making ASF consistent with every other format.

## Problem

* JIRA ticket: PICARD-3404

`ASFFile._load()` normalizes `WM/SharedUserRating` from the 0-99 WMA scale to
Picard's 0-5 scale and leaves the result as an `int`. A few lines later every
tag value passes through a truthiness filter:

```python
values = [str(value) for value in values if value]
```

The rating is still an `int` at that point, so a value of 0 is falsy and is
silently discarded. The rating is written to the file correctly (verified as
`ASFDWordAttribute(0)`); only reading it back loses it. ID3, Vorbis, MP4 and
APEv2 all preserve a rating of 0, so ASF was the odd one out.

The bug has been latent since ratings were added to ASF in 2009, and became
live with the Python 3 port (commit `8834ec3384`, shipped in Picard 2.0),
which changed the code to filter the still-`int` value on truthiness before
converting it to a string.

## Solution

Normalize the rating to `str` immediately, so 0 becomes `'0'` and survives the
truthiness filter. Empty strings are still filtered out as before.

The first commit is the failing reproduction: `CommonTests.test_ratings` was
only ever testing a single rating value because the loop overwrote its own
loop variable. Fixing the loop to genuinely test all six values (0-5) as
subtests exposes `SUBFAILED(rating=0)` for `ASFTest`, `WMATest` and `WMVTest`.
The second commit applies the one-line fix and the test passes.

## AI Usage

* [ ] No AI/LLM use
* [ ] Primarily AI developed

Written with Kiro CLI and carefully reviewed by a human. The failing test and
the fix were each verified from tool output during the session: the test fails
with `SUBFAILED(rating=0)` at the first commit and passes at the second.

## Action

Additional actions required:

### DIFF
--- a/picard/formats/asf.py
+++ b/picard/formats/asf.py
@@ -242,8 +242,16 @@ class ASFFile(File):
 
                 continue
             elif name == 'WM/SharedUserRating':
-                # Rating in WMA ranges from 0 to 99, normalize this to the range 0 to 5
-                values[0] = int(round(int(str(values[0])) / 99.0 * (config.setting['rating_steps'] - 1)))
+                # Rating in WMA ranges from 0 to 99, normalize this to the range 0 to 5.
+                # The result is converted to str because the values below are
+                # filtered on truthiness, which would drop a rating of 0.
+                rating_steps = config.setting['rating_steps']
+                try:
+                    rating = int(values[0])
+                except ValueError:
+                    log.warning('Invalid rating value in %r: %s', filename, values[0])
+                else:
+                    values[0] = str(round(rating / 99 * (rating_steps - 1)))
             elif name == 'WM/PartOfSet':
                 disc = str(values[0]).split("/")
                 if len(disc) > 1:
--- a/test/formats/common.py
+++ b/test/formats/common.py
@@ -498,13 +498,15 @@ class CommonTests:
             if not self.supports_ratings:
                 raise unittest.SkipTest("Ratings not supported")
             for rating in range(6):
-                rating = 1
-                metadata = Metadata()
-                metadata['~rating'] = rating
-                loaded_metadata = save_and_load_metadata(self.format_registry, self.filename, metadata)
-                self.assertEqual(
-                    int(loaded_metadata['~rating']), rating, '~rating: %r != %r' % (loaded_metadata['~rating'], rating)
-                )
+                with self.subTest(rating=rating):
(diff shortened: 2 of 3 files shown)


picard/formats/asf.py
```diff
@@ -242,8 +242,10 @@ def _load(self, filename):
 
                 continue
             elif name == 'WM/SharedUserRating':
-                # Rating in WMA ranges from 0 to 99, normalize this to the range 0 to 5
-                values[0] = int(round(int(str(values[0])) / 99.0 * (config.setting['rating_steps'] - 1)))
+                # Rating in WMA ranges from 0 to 99, normalize this to the range 0 to 5.
+                # The result is converted to str because the values below are
+                # filtered on truthiness, which would drop a rating of 0.
+                values[0] = str(int(round(int(str(values[0])) / 99.0 * (config.setting['rating_steps'] - 1))))
```
phw: I think the `int` around `round` is no longer needed, Python 3 returns an int already. We can simplify this line to

```
values[0] = str(round(int(str(values[0])) / 99 * (config.setting['rating_steps'] - 1)))
```

Overall this looks rather messy, and might raise a ValueError if `WM/SharedUserRating` is not numeric (the value is just a string after all). When we are at it, shall we make it more robust and catch the error?

```python
                try:
                    values[0] = str(round(int(values[0]) / 99 * (config.setting['rating_steps'] - 1)))
                except ValueError:
                    log.warning('Invalid rating value in %r: %s', filename, values[0])
```

This is consistent with how it is handled in vorbis.py also

zas: Fixed and rebased in https://github.com/metabrainz/picard/pull/3378/commits/b5c6810230a8f797216c7cf8acdd9252f1c0f926

phw: Even better.

The old code was dealing with some Python 2 quirks around round and division.

phw (approved):

